PR #11765, opened 2021-08-09
Fix implementation of BYTE

chriseth:


test/formal/byte_big.py
```diff
@@ -0,0 +1,18 @@
...
+rule = Rule()
+
+n_bits = 256
+
+# Input vars
+X = BitVec('X', n_bits)
+A = BitVec('A', n_bits)
+
+rule.require(A >= n_bits / 8)
+rule.check(BYTE(A, X), 0)
```
chriseth, line 18: The old implementation of BYTE finds a counterexample here

test/formal/opcodes.py
```diff
@@ -58,5 +58,10 @@ def SAR(x, y):
 	return y >> x
 
 def BYTE(i, x):
+	bitInt = (BV2Int(i) + 1) * 8
 	bit = (i + 1) * 8
-	return If(UGT(bit, x.size()), BitVecVal(0, x.size()), (LShR(x, (x.size() - bit))) & 0xff)
+	return If(
+		bitInt > x.size(),
```
ekpyron: Would
```suggestion
		i > x.size() / 8 - 1,
```
also work?

ekpyron: (I guess we can assume ``x.size()`` to be a multiple of 8 and greater than 1 and then it's the same without any casting or non-constant arithmetic?)

chriseth:
Rebased.

ekpyron (dismissed):

test/formal/opcodes.py
```diff
@@ -59,4 +59,8 @@ def SAR(x, y):
 
 def BYTE(i, x):
 	bit = (i + 1) * 8
-	return If(UGT(bit, x.size()), BitVecVal(0, x.size()), (LShR(x, (x.size() - bit))) & 0xff)
+	return If(
+		i > x.size() / 8 - 1,
```
hrkrshnn: Isn't this a signed comparison?

For example:

```py
from z3 import *

def fun(i, x):
    return i > x.size() / 8 - 1

x = BitVecVal(200, 32)
i = BitVecVal(2**32 - 1, 32)

simplify(fun(i, x))
# returns False
```

chriseth: Fixed

ekpyron (approved):

PR #11773, opened 2021-08-10
An equivalence check for SIGNEXTEND opcode

hrkrshnn:
Checking the implementation of SIGNEXTEND using Z3's native SignExt and Extract

### DIFF
--- a/test/formal/signextend_equivalence.py
+++ b/test/formal/signextend_equivalence.py
@@ -1,0 +1,24 @@
+from rule import Rule
+from opcodes import *
+
+"""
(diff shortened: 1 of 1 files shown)


hrkrshnn:
In `t = 256 −8(μs[0] + 1)`, calculation, none of the overflows are accounted. If `μs[0] >= 31` implementations simply return `μs[1]`.

For reference: signextend in [Geth](https://github.com/holiman/uint256/blob/master/uint256.go#L1124) and [evmone](https://github.com/ethereum/evmone/blob/master/lib/evmone/instructions.hpp#L161).

chriseth (approved):

PR #11352, opened 2021-05-05
Redundant store eliminator.

chriseth:

PR #11774, opened 2021-08-10
Refactor redundant assign

chriseth:
Needed for https://github.com/ethereum/solidity/pull/11352

### DIFF
--- a/libsolutil/CommonData.h
+++ b/libsolutil/CommonData.h
@@ -295,6 +295,31 @@ decltype(auto) mapTuple(Callable&& _callable)
 	return detail::MapTuple<Callable>{std::forward<Callable>(_callable)};
 }
 
+/// Merges map @a _b into map @a _a. If the same key exists in both maps,
+/// calls @a _conflictSolver to combine the two values.
+template <class K, class V, class F>
+void joinMap(std::map<K, V>& _a, std::map<K, V>&& _b, F _conflictSolver)
+{
+	auto ita = _a.begin();
+	auto aend = _a.end();
+	auto itb = _b.begin();
+	auto bend = _b.end();
+
+	for (; itb != bend; ++ita)
+	{
+		if (ita == aend)
+			ita = _a.insert(ita, std::move(*itb++));
+		else if (ita->first < itb->first)
+			continue;
+		else if (itb->first < ita->first)
+			ita = _a.insert(ita, std::move(*itb++));
+		else
+		{
+			_conflictSolver(ita->second, std::move(itb->second));
+			++itb;
+		}
+	}
+}
 
 // String conversion functions, mainly to/from hex/nibble/byte representations.
 
--- a/libyul/optimiser/RedundantAssignEliminator.cpp
+++ b/libyul/optimiser/RedundantAssignEliminator.cpp
@@ -38,7 +38,7 @@ void RedundantAssignEliminator::run(OptimiserStepContext& _context, Block& _ast)
 	RedundantAssignEliminator rae{_context.dialect};
 	rae(_ast);
 
-	AssignmentRemover remover{rae.m_pendingRemovals};
+	StatementRemover remover{rae.m_pendingRemovals};
 	remover(_ast);
 }
 
@@ -49,7 +49,7 @@ void RedundantAssignEliminator::operator()(Identifier const& _identifier)
 
 void RedundantAssignEliminator::operator()(VariableDeclaration const& _variableDeclaration)
 {
-	ASTWalker::operator()(_variableDeclaration);
+	RedundantStoreBase::operator()(_variableDeclaration);
 
 	for (auto const& var: _variableDeclaration.variables)
 		m_declaredVariables.emplace(var.name);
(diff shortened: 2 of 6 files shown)


libsolutil/CommonData.h
```diff
@@ -295,6 +295,30 @@ decltype(auto) mapTuple(Callable&& _callable)
 	return detail::MapTuple<Callable>{std::forward<Callable>(_callable)};
 }
 
+template <class K, class V, class F>
+void joinMap(std::map<K, V>& _a, std::map<K, V>&& _b, F _conflictSolver)
```
leonardoalt, line 301: How's this different/better than `map::merge`?

leonardoalt: Hm I guess the conflict part

leonardoalt: Ah this was just moved around, nevermind

libyul/optimiser/RedundantAssignEliminator.cpp
```diff
@@ -104,35 +104,25 @@ void RedundantAssignEliminator::operator()(Switch const& _switch)
...
-	swap(m_forLoopInfo, forLoopInfo);
+		finalize(retParam.name, UseState::Used);
 }
 
 void RedundantAssignEliminator::operator()(ForLoop const& _forLoop)
 {
-	ForLoopInfo outerForLoopInfo;
-	swap(outerForLoopInfo, m_forLoopInfo);
+	ScopedSaveAndRestore outerForLoopInfo(m_forLoopInfo, {});
 	++m_forLoopNestingDepth;
```
ekpyron:
```suggestion
	ScopedSaveAndRestore loopNestingDepthIncrement(m_forLoopNestingDepth, m_forLoopNestingDepth + 1);
```
not sure that's really nicer, but it's at least more local than having the decrement at the very end. But also fine keeping as is.

libyul/optimiser/RedundantAssignUtils.h
```diff
@@ -0,0 +1,84 @@
...
+		_a.m_value =  Value(std::max(int(_a.m_value), int(_b.m_value)));
+	}
+private:
+	Value m_value = Undecided;
+};
+
+
+/// Working data for traversing for-loops.
+template <class TrackedAssignments>
+struct ForLoopInfo
```
ekpyron: That's quite a generic name for a top level construct... and also not very descriptive...

chriseth:
@leonardoalt  Is that an "approve"? :)

leonardoalt:
not yet

leonardoalt (dismissed):

ekpyron (dismissed):

libyul/optimiser/RedundantAssignUtils.h
chriseth: I know... Let me see what I can do.

chriseth:
Did a complete makeover and create a common base class.

chriseth:
Sorry, this is probably difficult to review, something like a three-way diff is probably the best way...

libyul/optimiser/RedundantStoreBase.h
```diff
@@ -0,0 +1,116 @@
...
+		static inline void join(State& _a, State const& _b)
+		{
+			// Using "max" works here because of the order of the values in the enum.
+			_a.m_value =  Value(std::max(int(_a.m_value), int(_b.m_value)));
+		}
+	private:
+		Value m_value = Undecided;
+	};
+
+	using TrackedStores = std::map<YulString, std::map<Statement const*, State>>;
```
leonardoalt, line 71: Might make sense to use a compare function on the inner map based on Statement id or something similar

libyul/optimiser/RedundantStoreBase.h
```diff
@@ -0,0 +1,116 @@
...
+	static void merge(TrackedStores& _target, TrackedStores&& _source);
+	static void merge(TrackedStores& _target, std::vector<TrackedStores>&& _source);
+	void changeUndecidedTo(YulString _variable, State _newState);
+	/// Called when a variable goes out of scope. Sets the state of all still undecided
+	/// assignments to the final state. In this case, this also applies to pending
+	/// break and continue TrackedStores.
+	void finalize(YulString _variable, State _finalState);
+
+	Dialect const& m_dialect;
+	std::set<Statement const*> m_pendingRemovals;
```
leonardoalt, line 88: Same here (re compare function)

libyul/optimiser/RedundantStoreBase.h
chriseth: You mean to ensure determinism? I don't think this is needed, to be honest. Furthermore, there are no IDs in yul.

leonardoalt: Yea, because there was a TODO somewhere about determinism

libyul/optimiser/RedundantAssignEliminator.cpp
```diff
@@ -215,101 +81,79 @@ void RedundantAssignEliminator::operator()(Leave const&)
...
 	}
 	for (auto& continueAssignments: m_forLoopInfo.pendingContinueStmts)
 	{
-		joinMap(assignments, std::move(continueAssignments[_variable]), State::join);
+		util::joinMap(stores, std::move(continueAssignments[_variable]), State::join);
 		continueAssignments.erase(_variable);
 	}
 
-	for (auto const& assignment: assignments)
+	for (auto const& store: stores)
```
ekpyron:
```suggestion
	for (auto&& [statement, state]: stores)
```
?

libyul/optimiser/RedundantStoreBase.h
ekpyron: `YulString` comparison is "weird", but deterministic. (Ah, sorry you meant the inner one... but still probably fine, I'll keep an eye on it)

libyul/optimiser/RedundantStoreBase.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include <variant>
+
+
+namespace solidity::yul
+{
+struct Dialect;
+
+/**
+ * Base class for both RedundantAsssignEliminator and RedundantStoreEliminator.
```
ekpyron: May state the prerequisites induced by the base here already as well?

libyul/optimiser/RedundantAssignEliminator.cpp
```diff
@@ -60,151 +60,17 @@ void RedundantAssignEliminator::operator()(Assignment const& _assignment)
 	visit(*_assignment.value);
 	for (auto const& var: _assignment.variableNames)
 		changeUndecidedTo(var.name, State::Unused);
-
-	if (_assignment.variableNames.size() == 1)
-		// Default-construct it in "Undecided" state if it does not yet exist.
-		m_assignments[_assignment.variableNames.front().name][&_assignment];
```
ekpyron, line 66: Note: this is actually removed from what I can see; still need to make sure that's fine.
Ah, it just moved to ``Statement``.

libyul/optimiser/RedundantAssignEliminator.cpp
chriseth: considered this, but there is another local variable called `state`...

chriseth: How can I enforce `statement` and `state` to be non-references? Is `auto [statement, state]` enough?

libyul/optimiser/RedundantStoreBase.cpp
```diff
@@ -0,0 +1,166 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+/**
+ * Base class for both RedundantAsssignEliminator and RedundantStoreEliminator.
```
ekpyron:
```suggestion
 * Base class for both RedundantAssignEliminator and RedundantStoreEliminator.
```
``Ass_sign`` is a nice typo :-).

libyul/optimiser/RedundantAssignEliminator.cpp
```diff
@@ -215,101 +81,79 @@ void RedundantAssignEliminator::operator()(Leave const&)
...
 			assignment.second = _newState;
 }
 
 void RedundantAssignEliminator::finalize(YulString _variable, RedundantAssignEliminator::State _finalState)
 {
-	std::map<Assignment const*, State> assignments;
-	joinMap(assignments, std::move(m_assignments[_variable]), State::join);
-	m_assignments.erase(_variable);
+	std::map<Statement const*, State> stores;
+	util::joinMap(stores, std::move(m_stores[_variable]), State::join);
```
ekpyron: Isn't this just
```
std::map<Statement const*, State> stores = std::move(m_stores[_variable]);
```
?

libyul/optimiser/RedundantAssignEliminator.cpp
```diff
@@ -215,101 +81,79 @@ void RedundantAssignEliminator::operator()(Leave const&)
...
-		else
-		{
-			_conflictSolver(ita->second, std::move(itb->second));
-			++itb;
-		}
-	}
+	if (auto const* assignment = get_if<Assignment>(&_statement))
+		if (assignment->variableNames.size() == 1)
+			// Default-construct it in "Undecided" state if it does not yet exist.
+			m_stores[assignment->variableNames.front().name][&_statement];
```
ekpyron, line 99: It has always worked like that apparently, but I'm not entirely clear on why only single assignments are relevant here...

ekpyron: On the other hand: why exactly did this move, since it still only applies to the `Assignment` case?

libyul/optimiser/RedundantAssignEliminator.cpp
ekpyron: The condition below can just be
```
if ((state == State::Unused || (state == State::Undecided && _finalState == State::Unused)) && ...
```
or the inner one could just be renamed... but also fine keeping it.

Forcing ``auto [statement, state]`` to be of value type is something I wouldn't try myself :-). (The problem there not so much being ``auto``, but the structured binding working via weird intermediate tuple types by spec ;-))

ekpyron (commented):
Some more comments, but I'm rather confident that it does what it did before.

Why don't we remove multi-assignments? As in both: what prevents them to be removed in the code and why is removing them prevented at all...

libyul/optimiser/RedundantAssignEliminator.cpp
chriseth: Yes, `m_stores` needs to have `Statement` now instead of `Assignment` to generalize it, so this had to be moved to `visit(Statement)`

libyul/optimiser/RedundantAssignEliminator.cpp
chriseth: Explanation about moving is here: https://github.com/ethereum/solidity/pull/11774/files#r688454105

single-assignments are fine because of SSA: If this is a multi-assignment and turns out to be redundant, the actual SSA assignment will be redundant and removed, while this assignment has to stay unless the right-hand-side is side-effect-free. We have to assign the value of the righ-hand-side somewhere, so this will turn into a variable declaration and the variable will not be used at all.

libyul/optimiser/RedundantAssignEliminator.cpp
chriseth: Yes, not sure why I did it that way, maybe because of conformity? I can change it to use move it is easier to understand.

libyul/optimiser/RedundantStoreBase.h
ekpyron: Also says ``Ass_sign`` by the way ;-).

libyul/optimiser/RedundantAssignEliminator.cpp
ekpyron: Ah right, hadn't thought about the type in the map.

chriseth:
Addressed all comments.

libyul/optimiser/RedundantAssignEliminator.cpp
ekpyron: Yul needs a ``pop<n>`` :-).

ekpyron: And the reason for multi-assignments to generally not be removed by any of this here, is that they are never actually entered in the map in the first place? Might be worth an assert in ``finalize``... but also fine.

ekpyron (dismissed):
Can be squashed and merged.

ekpyron (approved):

PR #11659, opened 2021-07-14
Fix broken bullet points in docs caused by a bug in dependencies of `sphinx-rtd-theme`

cameel:
Styling in our documentation is slightly broken due to https://github.com/readthedocs/sphinx_rtd_theme/issues/1115. Most noticeably, bullet points are not displayed for unordered lists.

This does not affect docs built using the `scripts/docs.sh` script but the docs at https://docs.soliditylang.org are not built using that.

Here are the commands to reproduce this problem locally (based on [build log #14231164 from readthedocs](https://readthedocs.org/api/v2/build/14231164.txt)):

```
virtualenv venv-solidity-docs/
source venv-solidity-docs/bin/activate
pip install --upgrade --no-cache-dir "mock==1.0.1" "pillow==5.4.1" "alabaster>=0.7,<0.8,!=0.7.5" "commonmark==0.8.1" "recommonmark==0.5.0" "sphinx<2" "sphinx-rtd-theme<0.5" "readthedocs-sphinx-ext<2.2"
pip install --exists-action=w --no-cache-dir -r docs/requirements.txt
cd docs
sphinx-build -n -b html -d _build/doctrees . _build/html
```
The commands should be executed from repo root.

PR #11775, opened 2021-08-10
[DOCS] Update resources.rst

franzihei:
## Changes Done

Updates the resources section of the docs:
- Deletes all outdated tools and 404 links
- Adds missing IDEs and tools
- Adds some new relevant links to the "general" section

## Remaining Issue

For some reason, the formatting looks different now. 

It used to show the IDEs like this:

image

while now for me locally it looks like this: 

image

I don't know why. Can anybody help me?

### DIFF
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -1,45 +1,49 @@
+#########
 Resources
----------
+#########
 
-General
-~~~~~~~
+General Resources
+=================
 
-* `Ethereum <https://ethereum.org>`_
+* `Ethereum.org Developer Portal <https://ethereum.org/en/developers/>`_
+* `Ethereum StackExchange <https://ethereum.stackexchange.com/>`_
+* `Solidity Portal <https://soliditylang.org/>`_
+* `Solidity Changelog <https://github.com/ethereum/solidity/blob/develop/Changelog.md>`_
+* `Solidity Source Code on GitHub <https://github.com/ethereum/solidity/>`_
+* `Solidity Language Users Chat <https://matrix.to/#/#ethereum_solidity:gitter.im>`_
+* `Solidity Compiler Developers Chat <https://matrix.to/#/#ethereum_solidity-dev:gitter.im>`_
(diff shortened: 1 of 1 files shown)


cameel:
Looks like that's an intentional change in newer versions of the RTD theme: https://github.com/readthedocs/sphinx_rtd_theme/issues/979#issuecomment-663231386. Apparently that look was inconsistent with other stuff. Weird, I thought it looked nice. We could bring it back with a bit of custom CSS (https://github.com/readthedocs/sphinx_rtd_theme/issues/979#issuecomment-662288540).

BTW, It must have changed after I updated `sphinx-rtd-theme` in #11659 but the weird thing that I don't see the old style in rendered docs at docs.soliditylang.org, even for very old versions like 0.5.0. I thought those never get updated/rebuilt.

franzihei:
Ok I just thought I did something wrong. I'm fine with keeping it in the new style.

docs/resources.rst
```diff
@@ -1,45 +1,49 @@
...
+=================
 
-* `Ethereum <https://ethereum.org>`_
+* `Ethereum.org Developer Portal <https://ethereum.org/en/developers/>`_
+* `Ethereum StackExchange <https://ethereum.stackexchange.com/>`_
+* `Solidity Portal <https://soliditylang.org/>`_
+* `Solidity Changelog <https://github.com/ethereum/solidity/blob/develop/Changelog.md>`_
+* `Solidity Source Code on GitHub <https://github.com/ethereum/solidity/>`_
+* `Solidity Language Users Chat <https://gitter.im/ethereum/solidity/>`_
+* `Solidity Compiler Developers Chat <https://gitter.im/ethereum/solidity-dev/>`_
```
leonardoalt: Is there a matrix version for these instead of gitter? They might deprecate the gitter version at some point

franzihei: changed to Matrix (tried to display both which was confusing and I think eventually they will move everything to Matrix anyways)

leonardoalt (approved):

PR #11776, opened 2021-08-10
[SMTChecker] Fix ICE when running on Uniswap V2

leonardoalt:
https://github.com/Uniswap/uniswap-v2-core

When running the SMTChecker on UniswapV2Pair.sol from the repo above, we get an ICE. This is caused by a weird combination of
- a library having a constant
- a contract that imports that library imports another contract as well
- libraries' constants are considered file level constants
- the EncodingContext is reset before each source is analyzed, clearing the created variables
- at some point the sequence here goes wrong and the used library constant is removed

I managed to reproduce the same behavior with 3 very minimal files, but when I put them in our multi-file isoltest format the ICE doesn't happen, probably because of some naming/order. But locally this PR does fix the issue.


ekpyron (approved):
Could get a bugfix changelog entry.

leonardoalt:
> Could get a bugfix changelog entry.

Yea I wanted to add one but I can't really tell what *exactly* is causing the bug, so I found it hard to write the summary

Issue #11764, opened 2021-08-09
What is the naming convention in terms of leading/trailing underscore?


Ahmed-Ali:
## Page

https://docs.soliditylang.org/en/latest/solidity-by-example.html

## Abstract
I am a bit confused about solidity naming convention when it comes to leading/trailing underscores.
Where and how they should be used?
I saw contract members with and without trailing underscore, parameters with and without trailing underscore, in other examples parameters with and without leading underscore.

Is this a consistency issue, or there is a convention that I just don't know yet?

hrkrshnn:
There is a style guide https://docs.soliditylang.org/en/latest/style-guide.html which doesn't seem to mention anything about leading underscores in the name.

To be fair, I think the leading underscore for function parameters comes from the style guide of the solidity compiler, written in c++. 

> I saw contract members with and without trailing underscore

Can you point out such an example?

Ahmed-Ali:
> Can you point out such an example?

I am actually wrong. Seems like I mixed the constructor parameters (with leading underscore) and the member without trailing underscore.

However, the parameters themselves does have two variations without underscores, with leading underscore. Here are few examples

Without underscore
```
/// Create a new ballot to choose one of `proposalNames`.
    constructor(bytes32[] memory proposalNames) {
        chairperson = msg.sender;
        voters[chairperson].weight = 1;

        // For each of the provided proposal names,
        // create a new proposal object and add it
        // to the end of the array.
        for (uint i = 0; i < proposalNames.length; i++) {
            // `Proposal({...})` creates a temporary
            // Proposal object and `proposals.push(...)`
            // appends it to the end of `proposals`.
            proposals.push(Proposal({
                name: proposalNames[i],
                voteCount: 0
            }));
        }
    }

    // Give `voter` the right to vote on this ballot.
    // May only be called by `chairperson`.
    function giveRightToVote(address voter) public {
        // If the first argument of `require` evaluates
        // to `false`, execution terminates and all
        // changes to the state and to Ether balances
        // are reverted.
        // This used to consume all gas in old EVM versions, but
        // not anymore.
        // It is often a good idea to use `require` to check if
        // functions are called correctly.
        // As a second argument, you can also provide an
        // explanation about what went wrong.
        require(
            msg.sender == chairperson,
            "Only chairperson can give right to vote."
        );
        require(
            !voters[voter].voted,
            "The voter already voted."
        );
        require(voters[voter].weight == 0);
        voters[voter].weight = 1;
    }
```

[...]

PR #11777, opened 2021-08-10
[DOCS] Improving the naming consistency in Solidity by Example documentation

Ahmed-Ali:
According to Solidity naming convention: https://docs.soliditylang.org/en/latest/style-guide.html#naming-conventions

There is no need to use underscore except when there is a naming collision. In which case, a trailing underscore is used to avoid the collision.
So in this change, I am removing all underscores, except for the ones that could shadow other symbols in their context (none of the changed names is a reserved keyword: https://docs.soliditylang.org/en/latest/cheatsheet.html?highlight=reserved#reserved-keywords )

For ref, this is a follow up PR with reference to issue: #11764

### DIFF
--- a/docs/examples/blind-auction.rst
+++ b/docs/examples/blind-auction.rst
@@ -64,15 +64,15 @@ to receive their money - contracts cannot activate themselves.
         /// The function auctionEnd has already been called.
         error AuctionEndAlreadyCalled();
 
-        /// Create a simple auction with `_biddingTime`
+        /// Create a simple auction with `biddingTime`
         /// seconds bidding time on behalf of the
-        /// beneficiary address `_beneficiary`.
+        /// beneficiary address `beneficiaryAddress`.
         constructor(
-            uint _biddingTime,
-            address payable _beneficiary
+            uint biddingTime,
+            address payable beneficiaryAddress
         ) {
-            beneficiary = _beneficiary;
-            auctionEndTime = block.timestamp + _biddingTime;
+            beneficiary = beneficiaryAddress;
+            auctionEndTime = block.timestamp + biddingTime;
         }
 
         /// Bid on the auction with the value sent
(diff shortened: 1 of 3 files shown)


docs/examples/micropayment.rst
```diff
@@ -346,11 +346,11 @@ The full contract
         address payable public recipient;   // The account receiving the payments.
         uint256 public expiration;  // Timeout in case the recipient never closes.
 
-        constructor (address payable _recipient, uint256 duration)
+        constructor (address payable recipient_, uint256 duration)
```
chriseth: Maybe it would be better to make a larger change to the parameter name to distinguish it from the state variable, a single '_' is easy to overlook.

Ahmed-Ali: @chriseth yeah makes sense; I thought about something like `recipientAddress` for the parameter name or something like that. But then it seemed that using trailing underscore will be more of the current naming guideline.

On it

docs/examples/safe-remote.rst
```diff
@@ -36,8 +36,8 @@ you can use state machine-like constructs inside a contract.
         // The state variable has a default value of the first member, `State.created`
         State public state;
 
-        modifier condition(bool _condition) {
-            require(_condition);
+        modifier condition(bool condition) {
```
hrkrshnn: I think this would lead to a shadowing warning. Can you please fix this. That should fix one of the CI issues.

hrkrshnn:
@Ahmed-Ali thanks for the PR and following up.

docs/examples/safe-remote.rst
Ahmed-Ali: thanks for spotting it. I don't think this was the intended change. On it

docs/examples/blind-auction.rst
```diff
@@ -260,13 +260,13 @@ invalid bids.
...
         /// place multiple bids.
-        function bid(bytes32 _blindedBid)
+        function bid(bytes32 blindedBid)
             public
             payable
             onlyBefore(biddingEnd)
         {
             bids[msg.sender].push(Bid({
-                blindedBid: _blindedBid,
+                blindedBid: blindedBid,
```
Ahmed-Ali, line 269: I am not 100% sure whether this would cause shadowing warning

chriseth:
Please rebase your changes instead of merging develop into your branch.

hrkrshnn:
I'll squash and remove the commits.

docs/examples/blind-auction.rst
hrkrshnn: This should be fine, because the struct member `blindedBid` is not visible by itself.

hrkrshnn (dismissed):
@Ahmed-Ali Thanks for the PR. Looks good.

hrkrshnn (approved):

Ahmed-Ali:
Thanks @hrkrshnn and apologies for not rebasing it earlier; had been pulled in other stuff that kept me away for a bit!

PR #11782, opened 2021-08-11
[SMTChecker] Make show unproved CLI a flag

leonardoalt:
Currently this takes a value `yes|no`, `true|false`, etc, but it should rather just be a flag "use this or not".
This was done for `divModNoSlacks` but not this option. This PR just changes that to a flag instead of value option.


ekpyron (approved):

test/cmdlineTests/model_checker_show_unproved_false_chc/args
```diff
@@ -1 +0,0 @@
---model-checker-engine chc --model-checker-show-unproved false
```
hrkrshnn, line 1: In theory, the test can stay after removing `--model-checker-show-unproved false`. But fine either way.

hrkrshnn (approved):

test/cmdlineTests/model_checker_show_unproved_false_chc/args
leonardoalt: But there's already a test with that

leonardoalt: (without the option completely)

PR #11783, opened 2021-08-11
Authorized Proxies

chriseth:


hrkrshnn:
Is this related to the poly chain hack?

chriseth:
Yes, at least this was the trigger for me to write it.

leonardoalt (dismissed):

docs/security-considerations.rst
```diff
@@ -193,6 +193,39 @@ Note that ``.send()`` does **not** throw an exception if the call stack is
...
+
+        function callOther(address _addr, bytes memory _payload) public
+                returns (bool, bytes memory) {
+            return proxy.callOther(_addr, _payload);
+        }
+        // Other functions and other functionality
+    }
+
+    // This is the full contract, it has no other functionality and no way
+    // to assume any privileges.
```
leonardoalt: Well, the privilege could be given by the target contract, assigning this as the owner for example. This is ofc the fault of the target contract, but not sure the sentence is precise in that case

chriseth: True. So maybe "has no other functionality and requires no privileges to work."?

leonardoalt (approved):

PR #11789, opened 2021-08-11
Add hirsute to release_ppa script.

ekpyron:
Should we add an item to the release checklist to verify that this is up-to-date?

### DIFF
--- a/scripts/release_ppa.sh
+++ b/scripts/release_ppa.sh
@@ -55,9 +55,9 @@ keyid=379F4801D622CDCF
 email=builds@ethereum.org
 packagename=solc
 
-static_build_distribution=focal
+static_build_distribution=hirsute
 
-DISTRIBUTIONS="focal groovy"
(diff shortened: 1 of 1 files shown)


scripts/release_ppa.sh
```diff
@@ -55,9 +55,9 @@ keyid=379F4801D622CDCF
 email=builds@ethereum.org
 packagename=solc
 
-static_build_distribution=focal
```
ekpyron, line 58: We could argue to stay with focal for the static builds, since we run our CI against focal... on the other hand we should update those to hirsute anyways.

scripts/release_ppa.sh
```diff
@@ -55,9 +55,9 @@ keyid=379F4801D622CDCF
...
+static_build_distribution=hirsute
 
-DISTRIBUTIONS="focal groovy"
```
ekpyron, line 60: groovy is EOL

leonardoalt (approved):

Issue #5773, opened 2019-01-09
[Yul] Expression simplifier not working in specific case


chriseth:
Why isn't `add(headStart44, value4)` simplified to `value4`?
Especially since we already use the fact that `value4` is zero in the revert call above...

_Originally posted by @chriseth in https://github.com/ethereum/solidity/pull/5764/files_

PR #5864, opened 2019-01-24
Some optimizer tuning.

chriseth:
Fixes https://github.com/ethereum/solidity/issues/5773

Unfortunately, I cannot really write a test, but this change eliminated an "add 0" in another (complicated) setup.

PR #11794, opened 2021-08-12
Make use of inner loops feature in optimizer step description.

chriseth:


ekpyron:
The expression simplifier should stabilize, shouldn't it? Then we could also just run it until it does that...

chriseth:
Thought about that, too, but the simplifier needs to run the splitter again, because if the replacement is not just a single opcode, it cannot use the variable anymore, since it is not in SSA form.

We can of course just run the splitter just before the simplifier whenever it is run and then stabilize.

chriseth:
Hm, we could also support nested brackets in the optimiser suite and change the sequence to include
`[sxu]` (simplifier, splitter, unused pruner) - the suite will run the bracketed expression until it is stable.

ekpyron:
> Thought about that, too, but the simplifier needs to run the splitter again, because if the replacement is not just a single opcode, it cannot use the variable anymore, since it is not in SSA form.
> 
> We can of course just run the splitter just before the simplifier whenever it is run and then stabilize.

Yeah, I know, I meant just running them both until stable.

chriseth:
@cameel what do you think, how much work would it be to allow nested `[...]` sequences inside the optimizer sequences?

cameel:
@chriseth Not much as long as nothing fancy is needed. The processing is inside a single function that loops over the string so it's just a matter of either giving it a stack or making it recursive.

libsolidity/interface/OptimiserSettings.h
```diff
@@ -44,7 +44,7 @@ struct OptimiserSettings
 	static char constexpr DefaultYulOptimiserSteps[] =
 		"dhfoDgvulfnTUtnIf"            // None of these can make stack problems worse
 		"["
-			"xarrscLM"                 // Turn into SSA and simplify
+			"xarrsxuscLM"              // Turn into SSA and simplify
```
chriseth: Why is it `sc` and not `cs` by the way? `c` is common subexpression eliminator and `s` is expression simplifier and the expression simplifier docs say it is better if `c` is run beforehand.

cameel: I checked to make sure I did not introduce this by mistake when refactoring the optimizer suite to use a sequence but looks like it was an intentional tuning that you did a long time ago in #5864 to fix #5773 :)

> Why isn't `add(headStart44, value4)` simplified to `value4`?
> Especially since we already use the fact that `value4` is zero in the revert call above...

> Unfortunately, I cannot really write a test, but this change eliminated an "add 0" in another (complicated) setup.

test/libsolidity/semanticTests/externalContracts/ramanujan_pi.sol
```diff
@@ -35,10 +35,10 @@ contract test {
 // compileViaYul: also
 // ----
 // constructor()
-// gas irOptimized: 528041
+// gas irOptimized: 528905
```
chriseth: This is weird - maybe because we run remat multiple times now?

chriseth: Running `m` only once did not change anything.

chriseth: It's weird because I don't see a big difference in the assembly. The version after the PR looks better to me - maybe we have to wait and see what the new evm code transform makes of it?

chriseth: I think this change is fine - it is just 4 bytes, right?

chriseth:
It is weird that this does not have a real impact. I will undo the changes except for the cases where we already have a repetition.

hrkrshnn (approved):

cameel:
Yeah, that's really weird. Just to make sure it's not a bug in the code handling the brackets, I locally replaced the brackets with steps manually repeated a few times and ran tests. There were no failures so apparently repeating these steps multiple times really does not affect the final output. At least on the code samples we have in the repo.

chriseth:
Ah, but I'm actually not surprised that the current version has no changes in the test: If running the step one more time would make a difference, it is just that the outer loop is run once again. So adding inner loops was just to have fewer executions overall.

PR #11795, opened 2021-08-12
[SMTChecker] Fix false positive on external calls from constructors

leonardoalt:
We shouldn't add the nondeterministic interface constraint for external calls from constructors since they can't re-enter anyway.

### DIFF
--- a/libsolidity/formal/CHC.cpp
+++ b/libsolidity/formal/CHC.cpp
@@ -756,6 +756,9 @@ void CHC::externalFunctionCall(FunctionCall const& _funCall)
 	for (auto var: function->returnParameters())
 		m_context.variable(*var)->increaseIndex();
 
+	if (!m_currentFunction || m_currentFunction->isConstructor())
+		return;
+
 	auto preCallState = vector<smtutil::Expression>{state().state()} + currentStateVariables();
 	bool usesStaticCall = kind == FunctionType::Kind::BareStaticCall ||
 		function->stateMutability() == StateMutability::Pure ||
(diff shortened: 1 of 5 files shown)


test/libsolidity/smtCheckerTests/external_calls/external_call_from_constructor_reentrancy_1.sol
```diff
@@ -0,0 +1,15 @@
+interface D {
+	function ext(C c) external returns (uint);
+}
+
+contract C {
+	uint x;
+	function s(uint _x) public { x = _x; }
+	constructor(D d) {
+		uint a = d.ext(this);
+		assert(x == 0); // should hold because there's no reentrancy from the constructor
```
leonardoalt, line 10: Before this PR this used to fail

libsolidity/formal/CHC.cpp
```diff
@@ -756,6 +756,9 @@ void CHC::externalFunctionCall(FunctionCall const& _funCall)
 	for (auto var: function->returnParameters())
 		m_context.variable(*var)->increaseIndex();
 
+	if (!m_currentFunction || m_currentFunction->isConstructor())
```
ekpyron, line 759: ``!m_currentFunction`` is stuff like state variable initialization and thereby always constructor context I guess?

ekpyron (dismissed):

libsolidity/formal/CHC.cpp
leonardoalt: Yep!

leonardoalt: If `!m_currentFunction` here, it means there was a state variable init that made an external call

leonardoalt:
@ekpyron can you reapprove? had to fix test expectations that I forgot

ekpyron (approved):

PR #11615, opened 2021-07-05
Yul Stack Layout Generator (New Code Transform Step 2)

ekpyron:
Second step involved in #11493. ~~Depends on https://github.com/ethereum/solidity/pull/11525.~~
Immediate prerequesite for https://github.com/ethereum/solidity/pull/11797

PR #11493, opened 2021-06-07
Enable new Yul->EVM code generation. (New Code Transform Step 4)

ekpyron:
Depends on https://github.com/ethereum/solidity/pull/11974

I'm keeping a general description of the process here in https://hackmd.io/3xjDedzmRJWuyfK7RKUwQw

PR #11797, opened 2021-08-12
Optimized EVM Code Transform. (New Code Transform Step 3) 

ekpyron:
Depends on https://github.com/ethereum/solidity/pull/11615.
Prerequesite for https://github.com/ethereum/solidity/pull/11493 (which runs it universally and thereby universally tests it).

The idea here is to primarily validate that whatever https://github.com/ethereum/solidity/pull/11615 did made sense and leads to code consistent with the Yul Control Flow Graph.

### DIFF
--- a/libyul/backends/evm/OptimizedEVMCodeTransform.cpp
+++ b/libyul/backends/evm/OptimizedEVMCodeTransform.cpp
@@ -1,0 +1,513 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+#include <libyul/backends/evm/OptimizedEVMCodeTransform.h>
+
+#include <libyul/backends/evm/ControlFlowGraphBuilder.h>
+#include <libyul/backends/evm/StackHelpers.h>
+#include <libyul/backends/evm/StackLayoutGenerator.h>
+
+#include <libyul/Utilities.h>
+
+#include <libsolutil/Visitor.h>
+#include <libsolutil/cxx20.h>
+
+#include <range/v3/view/drop.hpp>
+#include <range/v3/view/enumerate.hpp>
+#include <range/v3/view/filter.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/map.hpp>
+#include <range/v3/view/reverse.hpp>
+#include <range/v3/view/take_last.hpp>
+
+using namespace solidity;
+using namespace solidity::yul;
+using namespace std;
+
+vector<StackTooDeepError> OptimizedEVMCodeTransform::run(
+	AbstractAssembly& _assembly,
+	AsmAnalysisInfo& _analysisInfo,
+	Block const& _block,
+	EVMDialect const& _dialect,
+	BuiltinContext& _builtinContext,
+	bool _useNamedLabelsForFunctions
+)
+{
+	std::unique_ptr<CFG> dfg = ControlFlowGraphBuilder::build(_analysisInfo, _dialect, _block);
+	StackLayout stackLayout = StackLayoutGenerator::run(*dfg);
+	OptimizedEVMCodeTransform optimizedCodeTransform(
+		_assembly,
+		_builtinContext,
+		_useNamedLabelsForFunctions,
+		*dfg,
+		stackLayout
+	);
+	// Create initial entry layout.
+	optimizedCodeTransform.createStackLayout(stackLayout.blockInfos.at(dfg->entry).entryLayout);
+	optimizedCodeTransform(*dfg->entry);
+	for (Scope::Function const* function: dfg->functions)
+		optimizedCodeTransform(dfg->functionInfo.at(function));
+	return move(optimizedCodeTransform.m_stackErrors);
+}
+
+void OptimizedEVMCodeTransform::operator()(CFG::FunctionCall const& _call)
+{
+	// Validate stack.
+	{
+		yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+		yulAssert(m_stack.size() >= _call.function.get().arguments.size() + 1, "");
+		// Assert that we got the correct arguments on stack for the call.
+		for (auto&& [arg, slot]: ranges::zip_view(
+			_call.functionCall.get().arguments | ranges::views::reverse,
+			m_stack | ranges::views::take_last(_call.functionCall.get().arguments.size())
+		))
+			validateSlot(slot, arg);
+		// Assert that we got the correct return label on stack.
+		auto const* returnLabelSlot = get_if<FunctionCallReturnLabelSlot>(
+			&m_stack.at(m_stack.size() - _call.functionCall.get().arguments.size() - 1)
+		);
+		yulAssert(returnLabelSlot && &returnLabelSlot->call.get() == &_call.functionCall.get(), "");
+	}
+
+	// Emit code.
+	{
+		m_assembly.setSourceLocation(locationOf(_call));
+		m_assembly.appendJumpTo(
+			getFunctionLabel(_call.function),
+			static_cast<int>(_call.function.get().returns.size() - _call.function.get().arguments.size()) - 1,
+			AbstractAssembly::JumpType::IntoFunction
+		);
+		m_assembly.appendLabel(m_returnLabels.at(&_call.functionCall.get()));
+	}
+
+	// Update stack.
+	{
+		// Remove arguments and return label from m_stack.
+		for (size_t i = 0; i < _call.function.get().arguments.size() + 1; ++i)
+			m_stack.pop_back();
+		// Push return values to m_stack.
+		for (size_t index: ranges::views::iota(0u, _call.function.get().returns.size()))
+			m_stack.emplace_back(TemporarySlot{_call.functionCall, index});
+		yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+	}
+}
+
+void OptimizedEVMCodeTransform::operator()(CFG::BuiltinCall const& _call)
+{
+	// Validate stack.
+	{
+		yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+		yulAssert(m_stack.size() >= _call.arguments, "");
+		// Assert that we got a correct stack for the call.
+		for (auto&& [arg, slot]: ranges::zip_view(
+			_call.functionCall.get().arguments |
+			ranges::views::enumerate |
+			ranges::views::filter(util::mapTuple([&](size_t idx, auto&) -> bool {
+				return !_call.builtin.get().literalArgument(idx);
+			})) |
+			ranges::views::reverse |
+			ranges::views::values,
+			m_stack | ranges::views::take_last(_call.arguments)
+		))
+			validateSlot(slot, arg);
+	}
+
+	// Emit code.
+	{
+		m_assembly.setSourceLocation(locationOf(_call));
+		static_cast<BuiltinFunctionForEVM const&>(_call.builtin.get()).generateCode(
+			_call.functionCall,
+			m_assembly,
+			m_builtinContext
+		);
+	}
+
+	// Update stack.
+	{
+		// Remove arguments from m_stack.
+		for (size_t i = 0; i < _call.arguments; ++i)
+			m_stack.pop_back();
+		// Push return values to m_stack.
+		for (size_t index: ranges::views::iota(0u, _call.builtin.get().returns.size()))
+			m_stack.emplace_back(TemporarySlot{_call.functionCall, index});
+		yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+	}
+}
+
+void OptimizedEVMCodeTransform::operator()(CFG::Assignment const& _assignment)
+{
+	yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+
+	// Invalidate occurrences of the assigned variables.
+	for (auto& currentSlot: m_stack)
+		if (VariableSlot const* varSlot = get_if<VariableSlot>(&currentSlot))
+			if (util::contains(_assignment.variables, *varSlot))
+				currentSlot = JunkSlot{};
+
+	// Assign variables to current stack top.
+	yulAssert(m_stack.size() >= _assignment.variables.size(), "");
+	for (auto&& [currentSlot, varSlot]: ranges::zip_view(
+		m_stack | ranges::views::take_last(_assignment.variables.size()),
+		_assignment.variables
+	))
+		currentSlot = varSlot;
+}
+
+OptimizedEVMCodeTransform::OptimizedEVMCodeTransform(
+	AbstractAssembly& _assembly,
+	BuiltinContext& _builtinContext,
+	bool _useNamedLabelsForFunctions,
+	CFG const& _dfg,
+	StackLayout const& _stackLayout
+):
+	m_assembly(_assembly),
+	m_builtinContext(_builtinContext),
+	m_useNamedLabelsForFunctions(_useNamedLabelsForFunctions),
+	m_dfg(_dfg),
+	m_stackLayout(_stackLayout)
+{
+}
+
+void OptimizedEVMCodeTransform::assertLayoutCompatibility(Stack const& _currentStack, Stack const& _desiredStack)
+{
+	yulAssert(_currentStack.size() == _desiredStack.size(), "");
+	for (auto&& [currentSlot, desiredSlot]: ranges::zip_view(_currentStack, _desiredStack))
+		yulAssert(holds_alternative<JunkSlot>(desiredSlot) || currentSlot == desiredSlot, "");
+}
+
+AbstractAssembly::LabelID OptimizedEVMCodeTransform::getFunctionLabel(Scope::Function const& _function)
+{
+	CFG::FunctionInfo const& functionInfo = m_dfg.functionInfo.at(&_function);
+	if (!m_functionLabels.count(&functionInfo))
+		m_functionLabels[&functionInfo] = m_useNamedLabelsForFunctions ? m_assembly.namedLabel(
+			functionInfo.function.name.str(),
+			functionInfo.function.arguments.size(),
+			functionInfo.function.returns.size(),
+			{}
+		) : m_assembly.newLabelId();
+	return m_functionLabels[&functionInfo];
+}
+
+void OptimizedEVMCodeTransform::validateSlot(StackSlot const& _slot, Expression const& _expression)
+{
+	std::visit(util::GenericVisitor{
+		[&](yul::Literal const& _literal) {
+			auto* literalSlot = get_if<LiteralSlot>(&_slot);
+			yulAssert(literalSlot && valueOfLiteral(_literal) == literalSlot->value, "");
+		},
+		[&](yul::Identifier const& _identifier) {
+			auto* variableSlot = get_if<VariableSlot>(&_slot);
+			yulAssert(variableSlot && variableSlot->variable.get().name == _identifier.name, "");
+		},
+		[&](yul::FunctionCall const& _call) {
+			auto* temporarySlot = get_if<TemporarySlot>(&_slot);
+			yulAssert(temporarySlot && &temporarySlot->call.get() == &_call && temporarySlot->index == 0, "");
+		}
+	}, _expression);
+}
+
+void OptimizedEVMCodeTransform::createStackLayout(Stack _targetStack)
+{
+	static constexpr auto slotVariableName = [](StackSlot const& _slot) {
+		return std::visit(util::GenericVisitor{
+			[](VariableSlot const& _var) { return _var.variable.get().name; },
+			[](auto const&) { return YulString{}; }
+		}, _slot);
+	};
+
+	yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+	// ::createStackLayout asserts that it has successfully achieved the target layout.
+	::createStackLayout(
+		m_stack,
+		_targetStack | ranges::to<Stack>,
+		// Swap callback.
+		[&](unsigned _i)
+		{
+			yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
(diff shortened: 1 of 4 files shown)


libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,431 @@
...
+					m_returnLabels[&_returnLabel.call.get()] = m_assembly.newLabelId();
+				m_assembly.appendLabelReference(m_returnLabels.at(&_returnLabel.call.get()));
+			},
+			[&](VariableSlot const& _variable)
+			{
+				if (m_currentFunctionInfo)
+					if (util::contains(m_currentFunctionInfo->returnVariables, _variable))
+					{
+						m_assembly.appendConstant(0);
+						return;
```
ekpyron, line 320: This is a blind spot in the verification, I guess. In practice, this will only happen on the first use of the return variable and the stack layout generator will propagate any use e.g. through loops, so it should be fine, but I don't see an easy way to verify that in here.

ekpyron: But it might warrant some dedicated test cases at least.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,433 @@
...
+				// - distinctive, since it is not generated elsewhere
+				// Disadvantages:
+				// - static analysis might get confused until it realizes that these are always popped
+				// Alternatives:
+				// - any other opcode with cost 2
+				// - unless the stack is empty: DUP1
+				// - the constant 0
+				// Note: it might even make sense to introduce a specific assembly item for this, s.t.
+				//       the peephole optimizer can deal with this (e.g. POP PUSHJUNK can be removed).
+				m_assembly.appendInstruction(evmasm::Instruction::PC);
```
ekpyron:
```suggestion
				m_assembly.appendInstruction(evmasm::Instruction::CODESIZE);
```
may be a good alternative.

ekpyron:
[rebased]

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,433 @@
...
+	::createStackLayout(m_stack, _targetStack | ranges::to<Stack>, [&](unsigned _i) {
+		yulAssert(_i > 0, "");
+		if (_i > 16)
+		{
+			int deficit = static_cast<int>(_i - 16);
+			StackSlot const& deepSlot = m_stack.at(m_stack.size() - _i - 1);
+			YulString varNameDeep = slotVariableName(deepSlot);
+			YulString varNameTop = slotVariableName(m_stack.back());
+			string msg = "Cannot swap " + (varNameDeep.empty() ? "Slot " + stackSlotToString(deepSlot) : "Variable " + varNameDeep.str()) +
+						 " with " + (varNameTop.empty() ? "Slot " + stackSlotToString(m_stack.back()) : "Variable " + varNameTop.str()) +
```
stackenbotten, line 204: Coding style error

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,433 @@
...
+						 ": too deep in the stack by " + to_string(deficit) + " slots in " + stackToString(m_stack);
```
stackenbotten, line 205: Coding style error

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	BuiltinContext& _builtinContext,
+	bool _useNamedLabelsForFunctions,
+	CFG const& _dfg,
+	StackLayout const& _stackLayout
+):
+m_assembly(_assembly),
+m_builtinContext(_builtinContext),
+m_useNamedLabelsForFunctions(_useNamedLabelsForFunctions),
+m_dfg(_dfg),
+m_stackLayout(_stackLayout)
```
chriseth:
```suggestion
	m_assembly(_assembly),
	m_builtinContext(_builtinContext),
	m_useNamedLabelsForFunctions(_useNamedLabelsForFunctions),
	m_dfg(_dfg),
	m_stackLayout(_stackLayout)
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+		// Swap callback.
+		[&](unsigned _i)
+		{
+			yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
+			yulAssert(_i > 0, "");
+			if (_i <= 16)
+				m_assembly.appendInstruction(evmasm::swapInstruction(_i));
+			else
+			{
+				int deficit = static_cast<int>(_i - 16);
```
chriseth:
```suggestion
				int deficit = static_cast<int>(_i) - 16;
```
(although I'm actually not sure which one is better)

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+	// ::createStackLayout asserts that it has successfully achieved the target layout.
+	::createStackLayout(
+		m_stack,
+		_targetStack | ranges::to<Stack>,
+		// Swap callback.
+		[&](unsigned _i)
+		{
+			yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
+			yulAssert(_i > 0, "");
```
chriseth:
```suggestion
			yulAssert(0 < _i && _i < m_stack.size(), "");
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+				m_assembly.appendInstruction(evmasm::swapInstruction(_i));
+			else
+			{
+				int deficit = static_cast<int>(_i - 16);
+				StackSlot const& deepSlot = m_stack.at(m_stack.size() - _i - 1);
+				YulString varNameDeep = slotVariableName(deepSlot);
+				YulString varNameTop = slotVariableName(m_stack.back());
+				string msg = "Cannot swap " + (varNameDeep.empty() ? "Slot " + stackSlotToString(deepSlot) : "Variable " + varNameDeep.str()) +
+					" with " + (varNameTop.empty() ? "Slot " + stackSlotToString(m_stack.back()) : "Variable " + varNameTop.str()) +
+					": too deep in the stack by " + to_string(deficit) + " slots in " + stackToString(m_stack);
```
chriseth:
```suggestion
				string msg =
					"Cannot swap " + (varNameDeep.empty() ? "slot " + stackSlotToString(deepSlot) : "variable " + varNameDeep.str()) +
					" with " + (varNameTop.empty() ? "slot " + stackSlotToString(m_stack.back()) : "variable " + varNameTop.str()) +
					": too deep in the stack by " + to_string(deficit) + " slots in " + stackToString(m_stack);
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: Best would probably be to just make this ``size_t`` in ``StackTooDeepError`` :-).

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+				{
+					m_assembly.appendInstruction(evmasm::dupInstruction(static_cast<unsigned>(*depth + 1)));
+					return;
+				}
+				else if (!canBeFreelyGenerated(_slot))
+				{
+					int deficit = static_cast<int>(*depth - 15);
+					YulString varName = slotVariableName(_slot);
+					string msg = (varName.empty() ? "Slot " + stackSlotToString(_slot) : "Variable " + varName.str())
+						+ " is " + to_string(*depth - 15) + " too deep in the stack " + stackToString(m_stack);
```
chriseth:
```suggestion
					string msg =
						(varName.empty() ? "Slot " + stackSlotToString(_slot) : "Variable " + varName.str()) +
						" is " + to_string(*depth - 15) + " slots too deep in the stack " + stackToString(m_stack);
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+					m_stackErrors.emplace_back(StackTooDeepError(
+						m_currentFunctionInfo ? m_currentFunctionInfo->function.name : YulString{},
+						varName,
+						deficit,
+						msg
+					));
+					m_assembly.markAsInvalid();
+					m_assembly.appendConstant(u256(0xDEADBEEF));
```
chriseth:
```suggestion
					m_assembly.appendConstant(u256(0xCAFFEE));
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+				},
+				[&](VariableSlot const& _variable)
+				{
+					if (m_currentFunctionInfo)
+						if (util::contains(m_currentFunctionInfo->returnVariables, _variable))
+						{
+							m_assembly.setSourceLocation(locationOf(_variable));
+							m_assembly.appendConstant(0);
+							return;
+						}
```
chriseth:
```suggestion
					if (m_currentFunctionInfo && util::contains(m_currentFunctionInfo->returnVariables, _variable))
					{
							m_assembly.setSourceLocation(locationOf(_variable));
						m_assembly.appendConstant(0);
						return;
					}
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+					yulAssert(false, "Variable not found on stack.");
+				},
+				[&](TemporarySlot const&)
+				{
+					yulAssert(false, "Function call result requested, but not found on stack.");
+				},
+				[&](JunkSlot const&)
+				{
+					// Note: this will always be popped, so we can push anything.
+					// TODO: discuss if PC is in fact a good choice here.
```
chriseth: I'm not convinced :)
PC is evil, it could actually break some static analysis tools.
I think CHAINID would be a better deterministic opcode, but of course this cannot be used in all evm versions.
DUP1 would be best IMO (and push1 0 for empty stack)
CODESIZE might be another option.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+		_builtinContext,
+		_useNamedLabelsForFunctions,
+		*dfg,
+		stackLayout
+	);
+	// Create initial entry layout.
+	optimizedCodeTransform.createStackLayout(stackLayout.blockInfos.at(dfg->entry).entryLayout);
+	optimizedCodeTransform(*dfg->entry);
+	for (Scope::Function const* function: dfg->functions)
+		optimizedCodeTransform(dfg->functionInfo.at(function));
```
chriseth, line 63: Why don't we need to call `createStackLayout` here? What actually does "Create initial entry layout" do? I would assume the entry layout to be empty.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	return move(optimizedCodeTransform.m_stackErrors);
+}
+
+void OptimizedEVMCodeTransform::operator()(CFG::FunctionCall const& _call)
+{
+	// Validate stack.
+	{
+		yulAssert(m_assembly.stackHeight() == static_cast<int>(m_stack.size()), "");
+		yulAssert(m_stack.size() >= _call.function.get().arguments.size() + 1, "");
+		// Assert that we got a correct arguments on stack for the call.
```
chriseth:
```suggestion
		// Assert that we got the correct arguments on stack for the call.
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: ``operator()(CFG::FunctionInfo const& _functionInfo)`` does it (the block visit can't because it will handle nested blocks differently)

ekpyron: The initial layout can contain constants and call return labels even in the outermost block.

ekpyron: At least in theory, in practice it'll probably be empty :-).

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+		if (VariableSlot const* varSlot = get_if<VariableSlot>(&currentSlot))
+			if (util::contains(_assignment.variables, *varSlot))
+				currentSlot = JunkSlot{};
+
+	// Assign variables to current stack top.
+	yulAssert(m_stack.size() >= _assignment.variables.size(), "");
+	for (auto&& [currentSlot, varSlot]: ranges::zip_view(
+		m_stack | ranges::views::take_last(_assignment.variables.size()),
+		_assignment.variables)
+	)
```
chriseth:
```suggestion
		_assignment.variables
	))
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	for (auto&& [currentSlot, desiredSlot]: ranges::zip_view(_currentStack, _desiredStack))
+		yulAssert(holds_alternative<JunkSlot>(desiredSlot) || currentSlot == desiredSlot, "");
+}
+
+AbstractAssembly::LabelID OptimizedEVMCodeTransform::getFunctionLabel(Scope::Function const& _function)
+{
+	CFG::FunctionInfo const& functionInfo = m_dfg.functionInfo.at(&_function);
+	if (!m_functionLabels.count(&functionInfo))
+		m_functionLabels[&functionInfo] = m_useNamedLabelsForFunctions ?
+			m_assembly.namedLabel(
```
chriseth:
```suggestion
		m_functionLabels[&functionInfo] = m_useNamedLabelsForFunctions ? m_assembly.namedLabel(
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: I'm not sure ``DUP1`` is less confusing actually. For something like ``CODESIZE`` that otherwise only rarely occurs you can learn and then have in mind that it's just stack alignment. For ``DUP1`` that's much more hidden.
And it's slightly more expensive...
But all is fine and the cost doesn't differ by much: https://gist.github.com/ekpyron/391401bf29ef1bc1a2a13c11d5b75d7f

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	auto const& blockInfo = m_stackLayout.blockInfos.at(&_block);
+
+	// Assert that the stack is valid for entering the block.
+	assertLayoutCompatibility(m_stack, blockInfo.entryLayout);
+	m_stack = blockInfo.entryLayout; // Might set some slots to junk, if not required by the block.
+	yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
+
+	// Emit jump label, if required.
+	if (auto label = util::valueOrNullptr(m_blockLabels, &_block))
+		m_assembly.appendLabel(*label);
```
chriseth, line 358: source location?

ekpyron: That's what I mentioned in the channel yesterday: those are missing from the CFG so far and my plan was to add them there separately and then all the way through to here as small change after this PR.
I can also do it now here, though.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+		// Assert that the operation produced its proclaimed output.
+		yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
+		yulAssert(m_stack.size() == baseHeight + operation.output.size(), "");
+		yulAssert(m_stack.size() >= operation.output.size(), "");
+		assertLayoutCompatibility(
+			m_stack | ranges::views::take_last(operation.output.size()) | ranges::to<Stack>,
+			operation.output
+		);
+	}
+
```
chriseth, line 386: Set the source location again.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+	yulAssert(static_cast<int>(m_stack.size()) == m_assembly.stackHeight(), "");
+
+	// Emit jump label, if required.
+	if (auto label = util::valueOrNullptr(m_blockLabels, &_block))
+		m_assembly.appendLabel(*label);
+
+	for (auto const& operation: _block.operations)
+	{
+		// Create required layout for entering the operation.
+		createStackLayout(m_stackLayout.operationEntryLayout.at(&operation));
```
chriseth, line 363: Actually this might also need a source location to be set.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+			// If this is the only jump to the block, we do not need a label and can directly continue with the target block.
+			if (!m_blockLabels.count(_jump.target) && _jump.target->entries.size() == 1)
+			{
+				yulAssert(!_jump.backwards, "");
+				(*this)(*_jump.target);
+			}
+			else
+			{
+				// Generate a jump label for the taret, if not already present.
```
chriseth:
```suggestion
				// Generate a jump label for the target, if not already present.
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -0,0 +1,527 @@
...
+				if (!m_blockLabels.count(_jump.target))
+					m_blockLabels[_jump.target] = m_assembly.newLabelId();
+
+				// If we already have generated the target block, jump to it, otherwise generate it in place.
+				if (m_generated.count(_jump.target))
+					m_assembly.appendJumpTo(m_blockLabels[_jump.target]);
+				else
+					(*this)(*_jump.target);
```
chriseth, line 414: This creates at least one C++ stack frame per jump, right? Could this lead to stack exhaustion?

ekpyron: Theoretically, yes, but I don't think you can get there in practice.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
axic: I brought up the idea a few days ago that since this uses `m_assembly`, we could just introduce `pushJunk` and at least for someone inspecting the assembly stream, it would be clear that this is a special push.

The assembler then could translate that into `DUP1` or `PUSH 1` depending on the stack height.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: I'd have said that part is the job of ``::createStackLayout``. But I can add it, doesn't hurt.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: I'm not sure... I generally don't like having an assembly item for it - that's strain on tooling and seems like overkill.
I also said I wouldn't want assembling to depend on evm version (you did suggest to translate the assembly item to codesize, chainid, resp. push0 once maybe available depending on version) - but that won't apply for just emitting DUP1 and PUSH 1 of course - but it'd still be an annoying assembly item, since it wouldn't have fixed code size :-).

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
chriseth: We had this problem in the old cpp-ethereum evm interpreter a lot, but maybe the stack frames then were also just very large.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -246,15 +246,19 @@ void ControlFlowGraphBuilder::operator()(Block const& _block)
...
 {
-	auto&& [ifBranch, afterIf] = makeConditionalJump(std::visit(*this, *_if.condition));
-	m_currentBlock = ifBranch;
+	auto& ifBranch = m_graph.makeBlock(debugDataOf(_if.body));
+	auto& afterIf = m_graph.makeBlock(debugDataOf(*m_currentBlock));
+	makeConditionalJump(debugDataOf(_if), std::visit(*this, *_if.condition), ifBranch, afterIf);
+	m_currentBlock = &ifBranch;
 	(*this)(_if.body);
-	jump(*afterIf);
+	jump(debugDataOf(_if), afterIf);
```
ekpyron, line 254: Or is it part of the body :-)?

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -286,90 +290,93 @@ void ControlFlowGraphBuilder::operator()(Switch const& _switch)
...
-		auto&& [caseBranch, elseBranch] = makeConditionalJump(makeValueCompare(*switchCase.value));
-		m_currentBlock = caseBranch;
+		auto& caseBranch = m_graph.makeBlock(debugDataOf(switchCase));
+		auto& elseBranch = m_graph.makeBlock(debugDataOf(_switch));
+		makeConditionalJump(debugDataOf(*switchCase.value), makeValueCompare(*switchCase.value), caseBranch, elseBranch);
+		m_currentBlock = &caseBranch;
 		(*this)(switchCase.body);
-		jump(afterSwitch);
-		m_currentBlock = elseBranch;
+		jump(debugDataOf(_switch), afterSwitch);
```
ekpyron, line 303: Or is it part of ``switchCase``?

ekpyron:
I just removed the source location changes - I'll push them as a separate PR in a bit.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: I can always rewrite it to something non-recursive, but my guess would be that it'd become much harder to read...

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: Have it in https://github.com/ethereum/solidity/pull/11922

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: But yeah, the stack frame at the recursions here is largely empty, not sure it's an issue.

ekpyron:
[rebased]

chriseth (approved):

Issue #7720, opened 2019-11-14
Increase use of C++ constexpr constant expressions in the code base.


ekpyron:
Came up in https://github.com/ethereum/solidity/pull/7655#discussion_r343720759 with one minor example for this.
But in general I'd expect a lot of places in the compiler could be refactored to benefit from constant expressions. One prominent example would be the simplification rule list, which is entirely compile time constant and with a bit of care could be constexpr initialized to a ``std::array`` which could live statically in the data segment. That might especially have beneficial consequences for the emscripten builds and the stack issues.
There's probably a lot of other places as well.

chriseth:
This requires rather modern compilers, though.

ekpyron:
For a lot of stuff like this compilers that are *actually* C++17 standard compliant should be sufficient... but yeah - for more complex stuff like the rule list this would be rather tricky and would probably also require quite some restructuring, so that we'd have to evaluate separately.

christianparpart:
I think one may want to start with the obvious stuff, and create smaller PRs for each iteration (whenever they come up), and by the time, evaluate whether or not it makes sense to also migrate the more complex stuff.

However, I'd highly recommend to apply `constexpr` as much as possible, since it's also ensuring we have no undefined behavior.

PR #11799, opened 2021-08-14
Increase use of C++ constexpr

benldrmn:
As described in issue #7720.
Changed the qualifiers to `static constexpr` instead of `static const` whenever possible. Some types don't have a `constexpr` constructor, e.g. `std::set<T>` or `std::regex`, so I couldn't `constexpr` them without changing the type (e.g. adding a `set` implementation with a `constexpr` constructor, which is an overkill).  `test/tools/ossfuzz/yulOptimizerFuzzDictionary.h` has a `static const std::vector<std::string>` variable defined, which can be made `static constexpr` in C++20.

Note that in `libsolidity/analysis/DocStringTagParser.cpp` I changed the `string static const` to `string_view static constexpr` which has a `constexpr` constructor, but requires C++17.

I refrained from changing `libsolutil/picosha2.h:43` `static const size_t` to `size_t static constexpr` since it's an external dependency, so I believe minimal changes are preferred for easier updates from `picosha`'s repo down the road.

I also changed a coding style guideline accordingly.

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -81,7 +81,7 @@ tuple<float, float> myNamespace::meanAndSigma(vector<float> const& _v)
    - Copyright
    - License (e.g. see COPYING)
 2. Never use `#ifdef`/`#define`/`#endif` file guards. Prefer `#pragma` once as first line below file comment.
-3. Prefer static const variable to value macros.
+3. Prefer static constexpr variables to value macros.
 4. Prefer inline constexpr functions to function macros.
 5. Split complex macro on multiple lines with `\`.
 
--- a/libsolidity/analysis/DocStringTagParser.cpp
+++ b/libsolidity/analysis/DocStringTagParser.cpp
@@ -34,6 +34,7 @@
 #include <boost/algorithm/string.hpp>
 
 #include <regex>
+#include <string_view>
 
 using namespace std;
 using namespace solidity;
@@ -233,7 +234,7 @@ void DocStringTagParser::parseDocStrings(
 
 	for (auto const& [tagName, tagValue]: _annotation.docTags)
 	{
-		string static const customPrefix("custom:");
+		string_view static constexpr customPrefix("custom:");
 		if (tagName == "custom" || tagName == "custom:")
 			m_errorReporter.docstringParsingError(
 				6564_error,
--- a/test/ExecutionFramework.h
+++ b/test/ExecutionFramework.h
@@ -49,9 +49,8 @@ namespace solidity::test
 using rational = boost::rational<bigint>;
 
 // The ether and gwei denominations; here for ease of use where needed within code.
-static const u256 gwei = u256(1) << 9;
-static const u256 ether = u256(1) << 18;
-
+static u256 const gwei = u256(1) << 9;
+static u256 const ether = u256(1) << 18;
 class ExecutionFramework
 {
 
--- a/test/tools/fuzzer_common.cpp
+++ b/test/tools/fuzzer_common.cpp
@@ -78,7 +78,7 @@ void FuzzerUtil::testCompilerJsonInterface(string const& _input, bool _optimize,
 void FuzzerUtil::forceSMT(StringMap& _input)
 {
 	// Add SMT checker pragma if not already present in source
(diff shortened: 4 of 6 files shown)


libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -233,7 +234,7 @@ void DocStringTagParser::parseDocStrings(
 
 	for (auto const& [tagName, tagValue]: _annotation.docTags)
 	{
-		string static const customPrefix("custom:");
+		string_view static constexpr customPrefix("custom:");
```
hrkrshnn, line 237: C++ 20 seems to allow constexpr strings. http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0980r1.pdf. Maybe something to change once we upgrade to C++20.

test/tools/fuzzer_common.cpp
```diff
@@ -78,7 +78,7 @@ void FuzzerUtil::testCompilerJsonInterface(string const& _input, bool _optimize,
 void FuzzerUtil::forceSMT(StringMap& _input)
 {
 	// Add SMT checker pragma if not already present in source
-	static const char* smtPragma = "pragma experimental SMTChecker;";
+	auto static constexpr smtPragma = "pragma experimental SMTChecker;";
```
hrkrshnn: I think the codebase tends to do `static auto` instead of `auto static`, although I can't find an explicit rule from https://github.com/ethereum/solidity/blob/develop/CODING_STYLE.md

Similarly for evmhost.cpp

hrkrshnn (commented):
Looks good. Some minor comment.

@benldrmn Thanks for the PR.

CODING_STYLE.md
```diff
@@ -81,7 +81,7 @@ tuple<float, float> myNamespace::meanAndSigma(vector<float> const& _v)
    - Copyright
    - License (e.g. see COPYING)
 2. Never use `#ifdef`/`#define`/`#endif` file guards. Prefer `#pragma` once as first line below file comment.
-3. Prefer static const variable to value macros.
+3. Prefer static constexpr variable to value macros.
```
cameel: Looks like there's an old typo here. Can you fix it while changing it?
```suggestion
3. Prefer static constexpr variables to value macros.
```

libsolidity/analysis/DocStringTagParser.cpp
benldrmn: True. Also a constexpr vector implementation that we can use.

I couldn't find a place where the project C++ standard is specified, is it documented somewhere?

test/tools/fuzzer_common.cpp
benldrmn: Fixed.

CODING_STYLE.md
benldrmn: Thanks, fixed.

libsolidity/analysis/DocStringTagParser.cpp
hrkrshnn: I believe we use c++ standard 17. https://github.com/ethereum/solidity/blob/develop/cmake/EthToolchains.cmake#L3

test/ExecutionFramework.h
```diff
@@ -49,8 +49,8 @@ namespace solidity::test
 using rational = boost::rational<bigint>;
 
 // The ether and gwei denominations; here for ease of use where needed within code.
-static const u256 gwei = u256(1) << 9;
-static const u256 ether = u256(1) << 18;
+u256 static constexpr gwei = u256(1) << 9;
+u256 static constexpr ether = u256(1) << 18;
```
hrkrshnn:
```suggestion
static u256 constexpr gwei = u256(1) << 9;
static u256 constexpr ether = u256(1) << 18;
```

hrkrshnn:
Will squash the commits now

hrkrshnn (dismissed):

hrkrshnn:
@benldrmn Thanks for the PR!

leonardoalt:
@hrkrshnn some builds are failing

Marenz:
I rebased & fixed the compiler error shown in the CI.

Marenz (approved):

Issue #11732, opened 2021-08-03
[CLI] Consolidate exception handlers in `CommandLineInterface`


cameel:
Related to #10295 and #11731.

Currently there are lots of independent `try`/`catch` statements inside various functions of `CommandLineInterface`. In most (if not all) cases they just print the error message and exit. In some cases this results in exceptions being silenced too eagerly, in others exceptions are not caught at all and can terminate the application without a sensible error message.

There should really be just a single error handler of this kind and it should reside at the very top-level of the hierarchy where printing and exiting is the default action for stuff that was not handled at lower levels. At lower levels the exceptions should be free to escape and interrupt the processing. Most `try`/`catch` blocks should only perform cleanup (if any is necessary at all) and then rethrow the exception so that the top-level handler can deal with it.

The top-level handler should be placed inside `main()` (like in #10295).

s0sharma:
Hi, I would like to work on this, can you suggest me ways to get started?

cameel:
Great!

- Take a look at [`CommandLineInterface.cpp`](https://github.com/ethereum/solidity/blob/develop/solc/CommandLineInterface.cpp). There are lots of `catch` handlers there.
- Find and remove all the generic handlers for things like `...`, `std::exception`, `util::Exception` or `InternalCompilerError`.
    - If a handler is doing more than just printing (e.g. it modifies a variable) remember to preserve it and just rethrow the exception.
- Leave the more specific handlers ones like `Error`, `CompilerError` or `UnimplementedFeature` alone. You could still reduce the scope of their `try` block though - `try` blocks for those should cover as little code as possible.
- Add a `try`/`catch` block in [`main.cpp`](https://github.com/ethereum/solidity/blob/develop/solc/main.cpp). You can take that part from #10295. Handle all these generic cases there by printing a message and any diagnostic info you can find.
- There might be some cases that need to handled specially but don't worry about that. If you miss any, we'll sort that out in the review of the PR.
- Make sure tests pass.
    - This code is covered mostly by command-line tests that you can run with `test/cmdlineTests.sh` script.

s0sharma:
Thanks for your reply I will be working on this.

s0sharma:
While testing file I am getting errors, so where can I ask for help or communicate with other members of the community?

cameel:
[...]

PR #11800, opened 2021-08-15
Separate control flow from error reporting in command line parser

### LINKED ISSUE #11731: [CLI] Separate control flow from error reporting in `CommandLineParser`
This is a continuation of the CLI refactor from #11518. Getting rid of `exit()` was suggested in the review.

Currently [`CommandLineParser`](https://github.com/ethereum/solidity/blob/develop/solc/CommandLineParser.h) uses the mechanism inherited from `CommandLineInterface` to report errors - it prints a message to the error stream and returns `false`.

- This is too inflexible in some situations - for example it can't return a value that will make the application exit without reporting an error so when it wants ro do so, it just calls `exit()` which terminates the application.
- Sometimes errors are returned in nested calls and the return value has to be bubbled up, which is error prone (see for example https://github.com/ethereum/solidity/pull/11544#discussion_r669767594).
- It also makes the code overly verbose - returning an error takes several lines of code (braces, printing to `serr()`, returning). Throwing exceptions would be more concise.

Here's my proposal on how to refactor this:

1) Modify `CommandLineInterface` so that the return value from the parser only indicates whether it should continue processing (`true`) or immediately exit without error (`false`).
    - Modify the functions that print version and license info in `CommandLineParser` to make them not use `exit()`.
    - Consider moving all of the printing logic to `CommandLineInterface`. The parser should only signal that printing was requested on the command line, not perform the action. This would be more in line with the purpose of the class and would let us get rid of the `sout()` stream form it.
2) Define an exception type that represents a command line validation error.
    - Replace all the code that prints to `serr()` and returns `false` with throwing that exception.
        - Remember to use `BOOST_THROW_EXCEPTION` macro.
    - Remove the `serr()` stream from the parser.
    - Consider defining a whole exception hierarchy and moving message formatting to specific exception classes. The parser could then be leaner and simply pass in the raw info needed to construct the message (file names, option names, option values, etc).
3) Catch the validation error in `CommandLineInterface` and handle it by printing the message to stderr and exiting with an error code.
    - Move the existing handler for `boost::program_options::error` from `CommandLineParser::parse()` to `CommandLineInterface` too. They're the same kind of error and should really be handled the same way.

neelriyer:
Partially Resolves #11731

I did the following:
- Modified the functions that print version and license info in `CommandLineParser` to make them not use `exit()`.
- Replaced all the code that prints to `serr()` and returns `false` with an exception (uses the `BOOST_THROW_EXCEPTION` macro)

Doesn't change too much. This is my first commit to this repository. Let me know what you think.

EDIT by cameel:
- ~Depends on #12125.~ Merged.
- ~Depends on #12118.~ Merged.

It's not based on these two PRs to avoid github closing it when they're merged. Instead it contains commits from these PRs. I'll rebase it to remove them when those PRs are merged.

### DIFF
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -23,6 +23,8 @@
  */
 #include <solc/CommandLineInterface.h>
 
+#include <solc/Exceptions.h>
+
 #include "license.h"
 #include "solidity/BuildInfo.h"
 
@@ -592,9 +594,16 @@ bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
 		return false;
 	}
 
-	bool success = parser.parse(_argc, _argv);
-	if (!success)
+	try
+	{
+		parser.parse(_argc, _argv);
+	}
+	catch (CommandLineValidationError const& _exception)
+	{
+		serr() << _exception.what() << endl;
 		return false;
+	}
+
 	m_hasOutput = m_hasOutput || parser.hasOutput();
 	m_options = parser.options();
 
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -546,19 +546,12 @@ void CommandLineInterface::createFile(string const& _fileName, string const& _da
 
 	string pathName = (m_options.output.dir / _fileName).string();
 	if (fs::exists(pathName) && !m_options.output.overwriteFiles)
-	{
-		serr() << "Refusing to overwrite existing file \"" << pathName << "\" (use --overwrite to force)." << endl;
-		m_outputFailed = true;
-		return;
-	}
+		solThrow(CommandLineOutputError, "Refusing to overwrite existing file \"" + pathName + "\" (use --overwrite to force).");
+
 	ofstream outFile(pathName);
 	outFile << _data;
 	if (!outFile)
-	{
-		serr() << "Could not write to file \"" << pathName << "\"." << endl;
-		m_outputFailed = true;
-		return;
-	}
+		solThrow(CommandLineOutputError, "Could not write to file \"" + pathName + "\".");
 }
 
 void CommandLineInterface::createJson(string const& _fileName, string const& _json)
@@ -643,9 +636,6 @@ void CommandLineInterface::processInput()
 		compile();
 		outputCompilationResults();
 	}
-
-	if (m_outputFailed)
-		solThrow(CommandLineOutputError, "Failed to write all output files to disk.");
 }
 
 void CommandLineInterface::printVersion()
@@ -969,10 +959,7 @@ void CommandLineInterface::writeLinkedFiles()
 			ofstream outFile(src.first);
 			outFile << src.second;
 			if (!outFile)
-			{
-				m_outputFailed = true;
 				solThrow(CommandLineOutputError, "Could not write to file " + src.first + ". Aborting.");
-			}
 		}
 	sout() << "Linking completed." << endl;
 }
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -583,7 +583,7 @@ void CommandLineInterface::createJson(string const& _fileName, string const& _js
 
 bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
 {
-	CommandLineParser parser(serr(/* _markAsUsed */ false));
+	CommandLineParser parser;
 
 	if (isatty(fileno(stdin)) && _argc == 1)
 	{
@@ -604,7 +604,6 @@ bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
 		return false;
 	}
 
-	m_hasOutput = m_hasOutput || parser.hasOutput();
 	m_options = parser.options();
 
 	return true;
@@ -1002,6 +1001,8 @@ string CommandLineInterface::objectWithLinkRefsHex(evmasm::LinkerObject const& _
 bool CommandLineInterface::assemble(yul::AssemblyStack::Language _language, yul::AssemblyStack::Machine _targetMachine)
 {
 	solAssert(m_options.input.mode == InputMode::Assembler, "");
+
+	serr() << "Warning: Yul is still experimental. Please use the output with care." << endl;
 
 	bool successful = true;
 	map<string, yul::AssemblyStack> assemblyStacks;
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -420,16 +420,10 @@ bool CommandLineInterface::readInputFiles()
 	if (m_fileReader.basePath() != "")
 	{
 		if (!boost::filesystem::exists(m_fileReader.basePath()))
-		{
-			serr() << "Base path does not exist: " << m_fileReader.basePath() << endl;
-			return false;
-		}
+			solThrow(CommandLineValidationError, "Base path does not exist: \"" + m_fileReader.basePath().string() + '"');
 
 		if (!boost::filesystem::is_directory(m_fileReader.basePath()))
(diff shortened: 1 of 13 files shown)


liblangutil/Exceptions.h
```diff
@@ -44,6 +44,7 @@ struct InternalCompilerError: virtual util::Exception {};
 struct FatalError: virtual util::Exception {};
 struct UnimplementedFeatureError: virtual util::Exception {};
 struct InvalidAstError: virtual util::Exception {};
+struct CommandLineValidationError: virtual util::Exception {};
```
cameel: I'd create a new `Exceptions.h` file in `solc/` and put the exception there. `liblangutil/` is for much more generic stuff and this exception will only be used by classes inside the `solc/` directory.

solc/CommandLineInterface.cpp
```diff
@@ -387,14 +387,12 @@ bool CommandLineInterface::readInputFiles()
 	{
 		if (!boost::filesystem::exists(m_fileReader.basePath()))
 		{
-			serr() << "Base path does not exist: " << m_fileReader.basePath() << endl;
-			return false;
+			BOOST_THROW_EXCEPTION(CommandLineValidationError() << errinfo_comment("Base path does not exist: "));
 		}
```
cameel: These braces are no longer needed. Please remove them here and in other places. Making the code less verbose is one of the goals of this change after all.

solc/CommandLineInterface.cpp
```diff
@@ -387,14 +387,12 @@ bool CommandLineInterface::readInputFiles()
 	{
 		if (!boost::filesystem::exists(m_fileReader.basePath()))
 		{
-			serr() << "Base path does not exist: " << m_fileReader.basePath() << endl;
-			return false;
+			BOOST_THROW_EXCEPTION(CommandLineValidationError() << errinfo_comment("Base path does not exist: "));
```
cameel: The new message no longer contains `m_fileReader.basePath()`.

Same below and some other places. Please make sure that the exceptions still contain all the information that used to be printed to stderr.

solc/CommandLineInterface.cpp
```diff
@@ -505,15 +501,15 @@ void CommandLineInterface::createFile(string const& _fileName, string const& _da
 	string pathName = (m_options.output.dir / _fileName).string();
 	if (fs::exists(pathName) && !m_options.output.overwriteFiles)
 	{
-		serr() << "Refusing to overwrite existing file \"" << pathName << "\" (use --overwrite to force)." << endl;
+		BOOST_THROW_EXCEPTION(CommandLineValidationError() << errinfo_comment("Refusing to overwrite existing file \"" << pathName << "\" (use --overwrite to force).");
```
cameel: I don't think this will work. You're using the `<<` operator on a string.

solc/CommandLineInterface.cpp
```diff
@@ -649,32 +644,20 @@ bool CommandLineInterface::compile()
 	}
 	catch (InternalCompilerError const& _exception)
 	{
-		serr() <<
-			"Internal compiler error during compilation:" <<
-			endl <<
-			boost::diagnostic_information(_exception);
-		return false;
+		BOOST_THROW_EXCEPTION(CommandLineValidationError() << errinfo_comment("Internal compiler error during compilation:" << boost::diagnostic_information(_exception));
 	}
```
cameel: I just noticed that you're modifying `CommandLineInterface`. We should refactor it eventually too but we really need to have #11732 first - otherwise nothing is catching these exceptions. That's why this issue is only about `CommandLineParser` for now. Please remove changes from this file and start with `CommandLineParser` instead. The only changes here should be in `CommandLineInterface::parseArguments()` - that's where the exceptions should be caught.

solc/CommandLineParser.cpp
```diff
@@ -189,15 +189,13 @@ void CommandLineParser::printVersionAndExit()
 		"Version: " <<
 		solidity::frontend::VersionString <<
 		endl;
-	exit(EXIT_SUCCESS);
```
cameel: It's not as simple as just removing the `exit()` :) The compiler must still exit after printing this info, just not like this. You should instead make `CommandLineParser::parse()` return `false` after it runs one of these functions. But now returning `false` means that there was an error - which is why you first need to ensure that returning with an error is only done by raising an exception.

cameel (dismissed):
This needs more work :) The biggest problem is that you're modifying the wrong file - the issue is about `CommandLineParser` and there are only small adjustments needed in `CommandLineInterface`.

Let me know if you need some more hints on how to continue.

s0sharma:
I'll make the changes.

neelriyer:
Thanks @cameel. I'm kind of embarrassed to have submitted this now :). I'll try to work on this and get back to you. Thanks for your help!

cameel:
@spiyer99 Are you working on this? Do you need help?

We have a ton of pending PRs right now and we'd like trim that number down a bit. If you don't have time for it, I can take over and finish it.

cameel:
This is almost finished. The one remaining thing is to update docstrings in `CommandLineInterface` and `CommandLineParser`.

Also #12125 and #12118 need to be merged first so I'm keeping it as a draft for now.

cameel:
This is now ready but still waiting for #12118 to be merged.

cameel:
In https://github.com/ethereum/solidity/pull/12125#issuecomment-942121559 @chriseth said it's ok to slightly change the control flow so that output errors cause an immediate failure (instead of just setting a flag and failing later after the output stage). I added a commit that changes it.

neelriyer:
> @spiyer99 Are you working on this? Do you need help?
> 
> We have a ton of pending PRs right now and we'd like trim that number down a bit. If you don't have time for it, I can take over and finish it.

Hey @cameel sorry for the late reply. I actually forgot about this, but good to see that you got this done in my absence 😊.

cameel:
All dependencies have been merged so this is now reviewable.

test/solc/CommandLineParser.cpp
```diff
@@ -75,12 +73,9 @@ BOOST_AUTO_TEST_CASE(no_options)
 	expectedOptions.modelChecker.initialize = true;
 	expectedOptions.modelChecker.settings = {};
 
-	stringstream serr;
-	optional<CommandLineOptions> parsedOptions = parseCommandLine(commandLine, serr);
+	CommandLineOptions parsedOptions = parseCommandLine(commandLine);
 
-	BOOST_TEST(serr.str() == "");
-	BOOST_REQUIRE(parsedOptions.has_value());
-	BOOST_TEST(parsedOptions.value() == expectedOptions);
```
chriseth, line 83: Not relevant here, but you can actually compare `optional<T>` with `T` directly, so this and the previous line would be equivalent to
`BOOST_TEST(parsedOptions == expectedOptions);`

solc/CommandLineInterface.cpp
```diff
@@ -1003,6 +1002,8 @@ bool CommandLineInterface::assemble(yul::AssemblyStack::Language _language, yul:
 {
 	solAssert(m_options.input.mode == InputMode::Assembler, "");
 
+	serr() << "Warning: Yul is still experimental. Please use the output with care." << endl;
```
chriseth, line 988: Are you sure this is the right place?

chriseth: Yes, it seems it is.

chriseth (approved):

test/solc/CommandLineParser.cpp
cameel: I think that it won't work with this macro though because boost requires both types to be printable so I'd have to define `<<` also for the optional version. Well, maybe a generic `<<` for printing `optional<T>` you had in one of your PRs would work for that.

PR #11802, opened 2021-08-16
Add fuzzer test harness

bshastry:


test/tools/ossfuzz/YulToEvmDifferentialFuzzer.cpp
```diff
@@ -0,0 +1,251 @@
...
+	YulOptimizerTestCommon optimizerTest(
+		subObject,
+		dialect
+	);
+	// Run circular references pruner and then stack limit evader.
+	string step = "fullSuite";
+	optimizerTest.setStep(step);
+	shared_ptr<solidity::yul::Block> astBlock = optimizerTest.run();
+	bool recursiveFunction = recursiveFunctionHasUnreachableVariables(dialect, *subObject);
+	string optimisedSubObject = AsmPrinter{}(*astBlock);
```
ekpyron: Ah, ok, so you do run the full Yul optimizer for the second run, but not the first... I'll need to check, if given that, I can reprodue the stack error. But not sure we should do that in general, at least for a first comparison.
For a first comparison it might be good to just run the code transform with and without stack optimization on the very same yul code.

Then we can check, if we want to force the old stack optimizations of the old code transform as reference (that will allow comparing more cases in which the old transform will otherwise throw stack too deep), that would involve replacing ``YulAssembler{...}.assemble();`` by something more manual for the first case.

Then we can check what happens with the optimizer enabled in both cases.

bshastry: > For a first comparison it might be good to just run the code transform with and without stack optimization on the very same yul code.

Okay, I'll refactor the Yul (full suite) optimized into an unused local function for now (essentially, comment out the said lines) and rerun the fuzzer. Does that sound reasonable?

ekpyron: Yep

ekpyron: Still, what worries me a bit, is that for the test case you posted earlier, I don't seem to be able to get a stack-too-deep error no matter what combination I try :-).

ekpyron: Also I'd still be interested in any new exceptions thrown that are not stack-too-deep errors.

bshastry: Can I run stack optimizer on Yul objects of the format 

```
object "main" {
  code {
    <deployObject>
  }
}
object "deployedObject" {
  code {
    <FuzzerGeneratedCode>
  }
}
}
```

ekpyron: Should be fine, yeah.

ekpyron:
I don't think we need this anymore at this point.

ekpyron:
[I haven't removed the branch yet, though, in case you still need it]

PR #11805, opened 2021-08-16
Remove the code for disabling `prepublish` hook before solc-js tests

ekpyron:


cameel:
I have updated https://github.com/ethereum/solc-js/pull/542 so that this change is no longer needed.

But now the `npm config set` line is redundant anymore so I'll push here in a moment to remove it. It was a terrible hack anyway, since it modifiers your global npm config in `~/.npmrc`.

ekpyron:
"Approved" (again I'm original author, so I can't officially :-))

cameel (approved):

PR #11806, opened 2021-08-16
User defined value types

### LINKED ISSUE #11531: User-defined value types
Originally proposed by @maurelian 

It should be possible to define a user-defined type that is identical to a provided value type (including operators, members, etc.), but cannot be implicitly converted to any other type, and can only be explicitly converted to and from the underlying value type.

This has the effect that arithmetic operations are impossible to perform across these types.

Example (syntax under discussion):

```
typedef DistanceInMeters uint;
typedef DistanceInInch uint;
typedef Price fixed128x10;

DistanceInMeters distanceToDestination;
Price buyerPrice;
Price sellerPrice;
```

In the above, you can do `buyerPrice - sellerPrice`, but you cannot do `buyerPrice * distanceToDestination`.

hrkrshnn:
Closes https://github.com/ethereum/solidity/issues/11531

TODO
- [x] Cleanup all TODOs
- [x] Implement AST import and Export
- [x] Write more tests
- [x] Documentation
- [x] ~NatSpec~?
- [x] ANTLR grammar


libsolidity/analysis/DeclarationContainer.cpp
```diff
@@ -50,7 +50,8 @@ Declaration const* DeclarationContainer::conflictingDeclaration(
 	if (
 		dynamic_cast<FunctionDefinition const*>(&_declaration) ||
 		dynamic_cast<EventDefinition const*>(&_declaration) ||
-		dynamic_cast<MagicVariableDeclaration const*>(&_declaration)
+		dynamic_cast<MagicVariableDeclaration const*>(&_declaration) ||
```
chriseth: Isn't this the list of symbols that can be overloaded? I'm not sure `UserDefinedValueType` belongs here.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -158,6 +158,20 @@ void DeclarationTypeChecker::endVisit(UserDefinedTypeName const& _typeName)
 		_typeName.annotation().type = TypeProvider::enumType(*enumDef);
 	else if (ContractDefinition const* contract = dynamic_cast<ContractDefinition const*>(declaration))
 		_typeName.annotation().type = TypeProvider::contract(*contract);
+	else if (auto userDefinedValueType = dynamic_cast<UserDefinedValueType const*>(declaration))
+	{
+		TypeName const* typeName = userDefinedValueType->typeName();
+		solAssert(typeName, "");
+		if (!dynamic_cast<ElementaryTypeName const*>(typeName))
```
chriseth: Don't we also want to restrict these to value types? (`string` and `bytes` are also elementary type names)

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -158,6 +158,20 @@ void DeclarationTypeChecker::endVisit(UserDefinedTypeName const& _typeName)
...
+			m_errorReporter.fatalTypeError(
+				8657_error,
+				_typeName.location(),
+				"The type has to be an elementary value type."
+			);
+		Type const* actualType = userDefinedValueType->typeName()->annotation().type;
```
chriseth: "underlyingType"?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1700,21 +1700,33 @@ Type const* TypeChecker::typeCheckTypeConversionAndRetrieveReturnType(
...
 		);
 	else if (!isPositionalCall)
-		m_errorReporter.typeError(
-			5153_error,
-			_functionCall.location(),
-			"Type conversion cannot allow named arguments."
-		);
+		m_errorReporter
+			.typeError(5153_error, _functionCall.location(), "Type conversion cannot allow named arguments.");
+	else if (auto userDefinedValueTypeType = dynamic_cast<UserDefinedValueTypeType const*>(expressionType))
```
chriseth: Do we need a special case here? Isn't this handled by `isExplicitlyConvertibelTo`?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2421,6 +2433,15 @@ bool TypeChecker::visit(FunctionCall const& _functionCall)
 		break;
 	}
 
+	case Type::Category::UserDefinedValueType:
```
chriseth: The conversion should be handled by the case above (`TypeType`).

libsolidity/parsing/Parser.h
```diff
@@ -108,6 +108,7 @@ class Parser: public langutil::ParserBase
 	ASTPointer<Identifier> parseIdentifier();
 	ASTPointer<Identifier> parseIdentifierOrAddress();
 	ASTPointer<UserDefinedTypeName> parseUserDefinedTypeName();
+	ASTPointer<UserDefinedValueType> parseUserDefinedValueType();
```
chriseth: Shouldn't this be placed closer to StructDefinition and EnumDefinition?

chriseth: Also maybe `parseUserDefinedValueTypeDefinition`?

libsolidity/parsing/Parser.cpp
```diff
@@ -386,6 +392,12 @@ ASTPointer<ContractDefinition> Parser::parseContractDefinition()
 				subNodes.push_back(parseEventDefinition());
 			else if (currentTokenValue == Token::Using)
 				subNodes.push_back(parseUsingDirective());
+			else if (
+				currentTokenValue == Token::Type &&
```
chriseth: Isn't `type` enough of a trigger?

libsolidity/ast/Types.h
```diff
@@ -1454,6 +1455,54 @@ class TypeType: public Type
...
 };
 
+/**
+ * The type of a UserDefinedValueType.
+ */
+class UserDefinedValueTypeType: public Type
+{
+public:
+	explicit UserDefinedValueTypeType(
+		Type const& _actualType,
```
chriseth: `_underlyingType`?

libsolidity/ast/Types.h
```diff
@@ -1454,6 +1455,54 @@ class TypeType: public Type
 	Type const* m_actualType;
 };
 
+/**
+ * The type of a UserDefinedValueType.
+ */
+class UserDefinedValueTypeType: public Type
```
chriseth: Why TypeType?
```
type NewType is int;
NewType a = NewType(2);
a;
```

In the above, the type of `a` (the one in isolation) is `UserDefinedValueType(int)`, while the type of `NewType` (as sub-expression of `NewType(2)` is `TypeType(UserDefinedValueType(int))`

If you treat the newly introduced type class more like the `Enum` or `Struct` class, then most of the other things should be automatic.

docs/types/value-types.rst
```diff
@@ -617,6 +617,49 @@ subsequent unsigned integer values starting from ``0``.
...
+.. index:: ! user defined value type, custom type
+
+.. _userDefinedValueTypes:
+
+User Defined Value Types
+------------------------
+
+A user defined value type allows creating a zero cost abstraction over a value type. This is similar
+to an alias, but with stricter type requirements.
+
```
chriseth, line 640: Maybe insert:

A user-defined value type is defined using ``type C is V``, where `C` is the name of the newly introduced type and `V` has to be a built-in value type.

docs/types/value-types.rst
```diff
@@ -617,6 +617,49 @@ subsequent unsigned integer values starting from ``0``.
...
+For a custom type ``C``, defined by ``type C is V;`` where ``V`` is the underlying value type, the
+following type requirements apply:
+
+- Implicit conversions of ``C`` are only allowed with ``C``.
```
chriseth: I would not list this, since it is not really a conversion

docs/types/value-types.rst
```diff
@@ -617,6 +617,49 @@ subsequent unsigned integer values starting from ``0``.
...
+- Explicit conversions of ``C`` are only allowed with itself and the underlying type, i.e., ``C``
```
chriseth: Same here

docs/types/value-types.rst
```diff
@@ -617,6 +617,49 @@ subsequent unsigned integer values starting from ``0``.
...
+  and ``V`` respectively.
+- The type ``C`` does not have any operators or bound member functions. In particular, the operator
+  ``==`` is disallowed. One can use operators or bound member functions by explicitly converting it
```
chriseth:
```suggestion
- The type ``C`` does not have any operators or bound member functions. In particular, even the operator
  ``==`` is not defined. One can use operators or bound member functions by explicitly converting it
```

docs/types/value-types.rst
```diff
@@ -617,6 +617,49 @@ subsequent unsigned integer values starting from ``0``.
...
+  to the underlying type.
+
+The following example demonstrates a custom type ``MyAddress`` which is an abstraction over the
```
chriseth: Shouldn't we use an example that also shows the usefulness of this concept?

libsolidity/ast/AST.h
```diff
@@ -726,6 +726,37 @@ class EnumValue: public Declaration
 	Type const* type() const override;
 };
 
+/**
+ * User defined value types, i.e., custom types, for example, `type MyInt is int`. Allows creating a
+ * zero cost abstraction over value type with stricter type requirements.
+ */
+class UserDefinedValueType: public Declaration
```
chriseth:
```suggestion
class UserDefinedValueTypeDefinition: public Declaration
```

libsolidity/ast/Types.cpp
```diff
@@ -410,6 +420,8 @@ BoolResult AddressType::isExplicitlyConvertibleTo(Type const& _convertTo) const
 {
 	if ((_convertTo.category() == category()) || isImplicitlyConvertibleTo(_convertTo))
 		return true;
+	else if (toUserDefinedValueType(_convertTo, *this))
```
chriseth: Maybe implement this in the Type class?

chriseth: Or in the UserDefinedValueType class?

libsolidity/ast/Types.cpp
```diff
@@ -3780,6 +3804,31 @@ BoolResult TypeType::isExplicitlyConvertibleTo(Type const& _convertTo) const
 	return isImplicitlyConvertibleTo(_convertTo);
 }
 
+string UserDefinedValueTypeType::richIdentifier() const
+{
+	return "t_userDefinedValueType" + identifierList(&underlyingType());
```
chriseth:
```suggestion
	return "t_userDefinedValueType" + parenthesizedIdentifier(underlyingType());
```

libsolidity/analysis/DeclarationTypeChecker.cpp
hrkrshnn: This is done in the TypeChecker.

test/libsolidity/syntaxTests/userDefinedValueType/all_value_types.sol
```diff
@@ -0,0 +1,21 @@
...
+// TODO add fixed point type
+type MyFixedBytes32 is bytes32;
+type MyFixedBytes1 is bytes1;
+type MyBool is bool;
+type MyFunction is function() external;
+type MyFunctionWithParm is function(uint) returns (uint);
+type MyFunctionWithCustomParam is function(MyAddress) returns (MyInt);
+
+
+// TODO should these be disallowed?
```
hrkrshnn: @chriseth Should these be disallowed?

test/libsolidity/syntaxTests/userDefinedValueType/explicit_conversions.sol
```diff
@@ -0,0 +1,10 @@
+type MyUint is uint;
+type MyAddress is address;
+
+function f() pure {
+    // TODO Should we allow explicit conversions from literals?
```
hrkrshnn: I think yes. This is similar to how newtype semantics work in haskell and rust (?)

libsolidity/analysis/DeclarationTypeChecker.cpp
chriseth: But we could do it already here, can't we? Shouldn't we do it as early as possible? Otherwise we might get into trouble if we have a forward reference and the type checker somehow assumes that the underlying type is a value type.

libsolidity/ast/Types.h
```diff
@@ -176,7 +176,7 @@ class Type
 		Address, Integer, RationalNumber, StringLiteral, Bool, FixedPoint, Array, ArraySlice,
 		FixedBytes, Contract, Struct, Function, Enum, Tuple,
 		Mapping, TypeType, Modifier, Magic, Module,
-		InaccessibleDynamic
+		InaccessibleDynamic, UserDefinedValueType
```
chriseth: Maybe move it closer to struct and enum?

libsolidity/analysis/DeclarationTypeChecker.cpp
hrkrshnn: Sure, I can move it here.

libsolidity/ast/Types.h
```diff
@@ -1076,6 +1077,56 @@ class EnumType: public Type
...
+		UserDefinedValueTypeDefinition const& _userDefinedValueType
+	):
+		m_underlyingType(_underlyingType),
+		m_userDefinedValueType(_userDefinedValueType)
+	{
+	}
+
+	Category category() const override { return Category::UserDefinedValueType; }
+	Type const& underlyingType() const { return m_underlyingType; }
+	UserDefinedValueTypeDefinition const& userDefinedValueType() const { return m_userDefinedValueType; }
```
chriseth: Do you need this? If yes maybe call the function `definition()`?

libsolidity/ast/Types.h
```diff
@@ -1076,6 +1077,56 @@ class EnumType: public Type
...
+
+	bool leftAligned() const override { return m_underlyingType.leftAligned(); }
+	bool canBeStored() const override { return m_underlyingType.canBeStored(); }
+	u256 storageSize() const override { return m_underlyingType.storageSize(); }
+	bool isValueType() const override
+	{
+		solAssert(m_underlyingType.isValueType(), "");
+		return true;
+	}
+	bool nameable() const override { return m_underlyingType.nameable(); }
```
chriseth: I think we should assert that the underlying type is nameable.

libsolidity/ast/Types.h
```diff
@@ -1076,6 +1077,56 @@ class EnumType: public Type
...
+
+	std::string toString(bool _short) const override { return "UserDefinedValueType(" + m_underlyingType.toString(_short) + ")"; }
```
chriseth: Shouldn't this also include the name of the type? (as we do for struct)

libsolidity/ast/Types.h
```diff
@@ -1076,6 +1077,56 @@ class EnumType: public Type
...
+
+	BoolResult isExplicitlyConvertibleTo(Type const& _convertTo) const override;
+	///	A helper function to reason about UserDefinedValueType to ValueType explicit conversion
```
chriseth:
```suggestion
	/// helper function to reason about UserDefinedValueType to ValueType explicit conversion
```

libsolidity/ast/Types.h
```diff
@@ -1076,6 +1077,56 @@ class EnumType: public Type
...
+	static bool convertTo(Type const& _userDefined, Type const& _valueType);
+protected:
+	std::vector<std::tuple<std::string, Type const*>> makeStackItems() const override;
+
```
chriseth, line 1127: I think we should also overwrite the following two functions:
 - canonicalName (assert false)
 - signatureInExternalFunction (assert false)

libsolidity/codegen/CompilerUtils.cpp
```diff
@@ -772,6 +772,33 @@ void CompilerUtils::convertType(
...
 	Type::Category targetTypeCategory = _targetType.category();
 
+	if (stackTypeCategory == Type::Category::UserDefinedValueType)
+	{
+		solAssert(_typeOnStack.isExplicitlyConvertibleTo(_targetType), "");
+		solAssert(_targetType.isValueType(), "");
+		return
+			convertType(
+				_targetType,
+				dynamic_cast<UserDefinedValueTypeType const&>(_typeOnStack).underlyingType(),
```
chriseth: shouldn't this be
```suggestion
				dynamic_cast<UserDefinedValueTypeType const&>(_typeOnStack).underlyingType(),
				_targetType,
```
?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3431,6 +3443,15 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 			body = "converted := value";
 			break;
 		}
+		case Type::Category::UserDefinedValueType:
```
chriseth: Can this still happen?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -869,8 +869,10 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
 
 	if (functionCallKind == FunctionCallKind::TypeConversion)
 	{
+		Type::Category category = _functionCall.expression().annotation().type->category();
 		solAssert(
-			_functionCall.expression().annotation().type->category() == Type::Category::TypeType,
+			category == Type::Category::TypeType ||
+			category == Type::Category::UserDefinedValueType,
```
chriseth: I don't think this special case is needed if the type checker properly assigns a typetype to `MyType` in the expression `MyType(t)`

libsolidity/parsing/Parser.cpp
```diff
@@ -134,6 +134,10 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
 					nodes.push_back(parseVariableDeclaration(options));
 					expectToken(Token::Semicolon);
 				}
+				else if (
+					m_scanner->currentToken() == Token::Type
+				)
```
```suggestion
				else if (m_scanner->currentToken() == Token::Type)
```

chriseth: Actually why not move this to a `case`? The part here should only be the fallback / workaround case

libsolidity/parsing/Parser.cpp
```diff
@@ -386,6 +390,10 @@ ASTPointer<ContractDefinition> Parser::parseContractDefinition()
 				subNodes.push_back(parseEventDefinition());
 			else if (currentTokenValue == Token::Using)
 				subNodes.push_back(parseUsingDirective());
+			else if (
+				currentTokenValue == Token::Type
+			)
```
```suggestion
			else if (currentTokenValue == Token::Type)
```

test/libsolidity/syntaxTests/userDefinedValueType/explicit_conversions.sol
chriseth: I'm not too sure about it - wouldn't it create ambiguities? We have to at least check that implicit conversion from the literal to the underlying type is possible.

test/libsolidity/syntaxTests/userDefinedValueType/all_value_types.sol
chriseth: Yes, I would disallow them, at least for now.

test/libsolidity/semanticTests/userDefinedValueType/erc20.sol
```diff
@@ -0,0 +1,146 @@
...
+// gas legacyOptimized: 24077
+// transfer(address,uint256): 2, 14 -> true
+// ~ emit Transfer(address,address,uint256): #0x1212121212121212121212121212120000000012, #0x02, 0x0e
+// gas irOptimized: 28614
+// gas legacy: 29586
+// gas legacyOptimized: 28675
+// transfer(address,uint256): 2, 2 -> FAILURE, hex"4e487b71", 0x11
+// gas irOptimized: 24052
+// gas legacy: 24492
+// gas legacyOptimized: 24074
```
hrkrshnn, line 147: For a gas diff, compared to our other erc20 test: https://gist.github.com/hrkrshnn/f7e6e69edfe62d5f94648a7b4f4abb26

I think the difference is because of inlining for the library based safemath.

docs/abi-spec.rst
```diff
@@ -89,6 +89,10 @@ The following non-fixed-size types exist:
 
 - ``<type>[]``: a variable-length array of elements of the given type.
 
+For :ref:`user defined value types<user-defined-value-types>`:
```
chriseth: I think this belongs into the "Mapping Solidity to ABI types".

docs/grammar/SolidityParser.g4
```diff
@@ -247,6 +249,11 @@ structMember: type=typeName name=identifier Semicolon;
  * Definition of an enum. Can occur at top-level within a source unit or within a contract, library or interface.
  */
 enumDefinition:	Enum name=identifier LBrace enumValues+=identifier (Comma enumValues+=identifier)* RBrace;
+/**
+ * Definition of a user defined value type. Can occur at top-level within a source unit or within a contract, library or interface.
+ */
+userDefinedValueTypeDefinition:
+	Type name=identifier Is elementaryTypeName[true] Semicolon;
```
chriseth, line 256:
```suggestion
	Type name=identifier Is elementaryTypeName[false] Semicolon;
```
I would not allow `address payable` here since it serves no purpose and makes the grammar more complicated.

docs/types/value-types.rst
```diff
@@ -617,6 +617,51 @@ subsequent unsigned integer values starting from ``0``.
 .. note::
     Enums can also be declared on the file level, outside of contract or library definitions.
 
+.. index:: ! user defined value type, custom type
+
+.. _user-defined-value-types:
+
+User Defined Value Types
```
chriseth, line 635: Is it "user defined" or "user-defined"?

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -140,6 +140,36 @@ bool DeclarationTypeChecker::visit(StructDefinition const& _struct)
 	return false;
 }
 
+bool DeclarationTypeChecker::visit(UserDefinedValueTypeDefinition const& _userDefined)
```
chriseth: Why can't we do this both in `endVisit`?

chriseth: Ok, I see why you did it this way: In `DeclarationTypeChecker::endVisit(UserDefinedTypeName`, the underlying type needs to be available, but I think this is a more serious problem and cannot be solved by changing the visitation order: We can also have forward references to other user-defined types at that point. I think the solution is the same as for struct types: Change the type constructor so that it only takes the AST definition, but do not analyze the type further.

chriseth: Indeed, the following code still triggers the assertion:
```
function f(MyIntB x) pure {}
type MyIntB is MyIntB;
```

test/libsolidity/semanticTests/userDefinedValueType/calldata.sol
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// test_f() -> true
+// gas irOptimized: 96989275
+// gas legacy: 96989774
+// gas legacyOptimized: 96989223
```
chriseth: This is weird - it got almost 100 gas more expensive after my changes.

ekpyron: If we can't get around cost otherwise, I'd say we just add the functions to the type instead.
I mean - they should be there anyways - otherwise you'll have to write these wrapping functions all the time for each type. And then we could actually more easily also allow function types, etc.

docs/grammar/SolidityParser.g4
hrkrshnn: Do you want to disallow `type MyPayableAddress is address payable`? This is currently allowed. I think there is no good reason to disallow it.

ekpyron: I'd say it should be allowed.

test/libsolidity/syntaxTests/userDefinedValueType/recursive_err.sol
```diff
@@ -0,0 +1,4 @@
+type MyInt1 is MyInt2;
+type MyInt2 is MyInt1;
+// ----
+// TypeError 8657: (15-21): The underlying type for a user defined value type has to be an elementary value type.
```
ekpyron, line 4: This is a bad restriction. We can of course keep this for now, but we *do* plan to lift this eventually, right :-)?

test/libsolidity/syntaxTests/userDefinedValueType/non_value_type_function_err.sol
```diff
@@ -0,0 +1,3 @@
+type MyFunction is function(uint) returns (uint);
+// ----
+// TypeError 8657: (19-49): The underlying type for a user defined value type has to be an elementary value type.
```
ekpyron, line 3: Since we now generate the wrapping and unwrapping functions internally, is there any reason for this not to be allowed? But also fine to keep disallowed for now.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1051,6 +1051,20 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
+	case FunctionType::Kind::Wrap:
+	case FunctionType::Kind::Unwrap:
+	{
+		solAssert(arguments.size() == 1, "");
+		FunctionType::Kind kind = functionType->kind();
+		auto& userDefinedType = kind == FunctionType::Kind::Wrap ?
+			dynamic_cast<UserDefinedValueType const&>(type(_functionCall)) :
+			dynamic_cast<UserDefinedValueType const&>(type(*arguments.at(0)));
+		if (kind == FunctionType::Kind::Wrap)
+			solAssert(type(*arguments.at(0)).isImplicitlyConvertibleTo(userDefinedType.underlyingType()), "");
```
ekpyron: Not sure this is necessary, but if we want it, then

```suggestion
		if (kind == FunctionType::Kind::Wrap)
			solAssert(type(*arguments.at(0)).isImplicitlyConvertibleTo(
				dynamic_cast<UserDefinedValueType const&>(type(_functionCall))
			);
```
would be enough.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3696,6 +3706,9 @@ string YulUtilFunctions::arrayConversionFunction(ArrayType const& _from, ArrayTy
 
 string YulUtilFunctions::cleanupFunction(Type const& _type)
 {
+	if (auto userDefinedValueType = dynamic_cast<UserDefinedValueType const*>(&_type))
+		return cleanupFunction(userDefinedValueType->underlyingType());
```
ekpyron, line 3710: It'd probably be more consistent to actually generate a function that's a thin wrapper around the underlying cleanup function similarly to the other proxy cases below.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -957,6 +957,23 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+			solAssert(arguments.size() == 1, "");
+			Type const* argumentType = arguments.at(0)->annotation().type;
+			solAssert(argumentType, "");
+			FunctionType::Kind kind = functionType->kind();
+			auto userDefinedType = kind == FunctionType::Kind::Wrap ?
+				dynamic_cast<UserDefinedValueType const*>(_functionCall.annotation().type) :
+				dynamic_cast<UserDefinedValueType const*>(argumentType);
+			solAssert(userDefinedType, "");
+			if (kind == FunctionType::Kind::Wrap)
+				solAssert(argumentType->isImplicitlyConvertibleTo(userDefinedType->underlyingType()), "");
```
ekpyron:
```suggestion
			if (kind == FunctionType::Kind::Wrap)
			{
				auto userDefinedType = dynamic_cast<UserDefinedValueType const*>(_functionCall.annotation().type);
				solAssert(userDefinedType, "");
				solAssert(argumentType->isImplicitlyConvertibleTo(userDefinedType->underlyingType()), "");
			}
```

ekpyron:
How about dropping the ``Value`` part and calling these things just plain ``UserDefinedType`` everywhere in the sources, docs and messages and rather note that this is *currently* restricted to non-function value types :-)?

test/libsolidity/ABIJson/user_defined_value_type.sol
```diff
@@ -0,0 +1,173 @@
...
+//     "name": "setMyInt",
+//     "outputs": [],
+//     "stateMutability": "nonpayable",
+//     "type": "function"
+//   },
+//   {
+//     "inputs":
+//     [
+//       {
+//         "internalType": "user defined type MyUInt8",
```
ekpyron, line 163: Ah, nice, so here the ``value`` thing is already skipped, that's good.

test/libsolidity/ASTJSON/userDefinedValueType.json
```diff
@@ -0,0 +1,264 @@
...
+  },
+  "id": 22,
+  "nodeType": "SourceUnit",
+  "nodes":
+  [
+    {
+      "id": 2,
+      "name": "MyAddress",
+      "nameLocation": "5:9:1",
+      "nodeType": "UserDefinedValueTypeDefinition",
```
ekpyron, line 30:
```suggestion
      "nodeType": "UserDefinedValueTypeDefinition",
```
As for names: This I would rather have say ``UserDefinedTypeDefinition`` or even just plainly ``TypeDefinition``.

docs/abi-spec.rst
```diff
@@ -113,6 +113,9 @@ them.
 +-------------------------------+-----------------------------------------------------------------------------+
 |:ref:`enum<enums>`             |``uint8``                                                                    |
 +-------------------------------+-----------------------------------------------------------------------------+
+|:ref:`user defined value types |same as its underlying value type                                            |
```
chriseth: "same as" sounds like it is not exactly the underlying value type

libsolidity/ast/Types.cpp
```diff
@@ -3754,6 +3786,34 @@ MemberList::MemberMap TypeType::nativeMembers(ASTNode const* _currentScope) cons
 		for (ASTPointer<EnumValue> const& enumValue: enumDef.members())
 			members.emplace_back(enumValue.get(), enumType);
 	}
+	else if (m_actualType->category() == Category::UserDefinedValueType)
+	{
+		auto& userDefined = dynamic_cast<UserDefinedValueType const&>(*m_actualType);
```
chriseth, line 3791:
```suggestion
		auto const& userDefined = dynamic_cast<UserDefinedValueType const&>(*m_actualType);
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -957,6 +957,23 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+
+			acceptAndConvert(*arguments[0], *function.parameterTypes()[0], true);
```
chriseth: Do we really need to do cleanup?

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -957,6 +957,23 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
 			break;
 		}
+		case FunctionType::Kind::Wrap:
+		case FunctionType::Kind::Unwrap:
+		{
+			solAssert(arguments.size() == 1, "");
+			Type const* argumentType = arguments.at(0)->annotation().type;
+			solAssert(argumentType, "");
+			FunctionType::Kind kind = functionType->kind();
+			auto userDefinedType = kind == FunctionType::Kind::Wrap ?
```
chriseth: Wouldn't it be easier and less confusing to retrieve this from `_functionCall.expression`? Maybe not.

libsolidity/codegen/ExpressionCompiler.cpp
chriseth: Can you assert that the implicit conversion is possible, please?

libsolidity/codegen/ExpressionCompiler.cpp
ekpyron: On that note test cases for stuff like ``(MyType).wrap`` and the like would be good.

docs/abi-spec.rst
hrkrshnn: Changed it to "exactly as its underlying value type"

libsolidity/codegen/ExpressionCompiler.cpp
hrkrshnn: I've removed the cleanup and added one more assert.

libsolidity/ast/Types.cpp
hrkrshnn: But the `const` here is redundant, isn't it?

chriseth: It is, but may be nice to mention it - we usually do.

docs/abi-spec.rst
```diff
@@ -113,6 +113,9 @@ them.
 +-------------------------------+-----------------------------------------------------------------------------+
 |:ref:`enum<enums>`             |``uint8``                                                                    |
 +-------------------------------+-----------------------------------------------------------------------------+
+|:ref:`user defined value types |exactly as its underlying value type                                         |
```
chriseth: Sorry, the "exactly" was not meant literally. The text above table says "on the right column the ABI types that represent them", so it should be enough to say "its underlying value type" or "the underlying value type".

docs/grammar/SolidityParser.g4
chriseth: Ok!

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2481,6 +2481,13 @@ bool TypeChecker::visit(FunctionCall const& _functionCall)
 			returnTypes = functionType->returnParameterTypes();
 			break;
 		}
+		case FunctionType::Kind::Wrap:
+		case FunctionType::Kind::Unwrap:
+		{
+			typeCheckFunctionGeneralChecks(_functionCall, functionType);
```
chriseth, line 2487: Ah nice!

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -957,6 +957,34 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+			solAssert(argumentType, "");
+			solAssert(functionCallType, "");
+			FunctionType::Kind kind = functionType->kind();
+			if (kind == FunctionType::Kind::Wrap)
+			{
+				solAssert(
+					argumentType->isImplicitlyConvertibleTo(
+					dynamic_cast<UserDefinedValueType const&>(*functionCallType).underlyingType()),
+					""
+				);
```
chriseth:
```suggestion
				solAssert(
					argumentType->isImplicitlyConvertibleTo(
						dynamic_cast<UserDefinedValueType const&>(*functionCallType).underlyingType()
					),
					""
				);
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1051,6 +1051,23 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
 	}
+	case FunctionType::Kind::Wrap:
+	case FunctionType::Kind::Unwrap:
+	{
+		solAssert(arguments.size() == 1, "");
+		FunctionType::Kind kind = functionType->kind();
+		if (kind == FunctionType::Kind::Wrap)
+			solAssert(
+				type(*arguments.at(0)).isImplicitlyConvertibleTo(
+					dynamic_cast<UserDefinedValueType const&>(type(_functionCall)).underlyingType()),
```
chriseth:
```suggestion
					dynamic_cast<UserDefinedValueType const&>(type(_functionCall)).underlyingType()
				),
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1051,6 +1051,23 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
+				""
+			);
+		else
+			solAssert(type(*arguments.at(0)).category() == Type::Category::UserDefinedValueType, "");
+
+		define(_functionCall, *arguments.at(0));
```
chriseth, line 1069: This actually skips one conversion - but would be fine for me, since the conversion is a no-op (and I hope it stays like that).

docs/abi-spec.rst
hrkrshnn: Fixed

chriseth (approved):

libsolidity/ast/AST.h
```diff
@@ -726,6 +726,37 @@ class EnumValue: public Declaration
 	Type const* type() const override;
 };
 
+/**
+ * User defined value types, i.e., custom types, for example, `type MyInt is int`. Allows creating a
+ * zero cost abstraction over value type with stricter type requirements.
+ */
+class UserDefinedValueTypeDefinition: public Declaration
```
ekpyron, line 733: I *still* think this is a *really* bad name for the node :-D - but well.

test/libsolidity/syntaxTests/userDefinedValueType/in_parenthesis.sol
```diff
@@ -0,0 +1,7 @@
+type MyInt is int;
+function f() pure {
+    (MyInt).wrap;
+    (MyInt).wrap(5);
+    (MyInt).unwrap;
+    (MyInt).unwrap(MyInt.wrap(5));
```
ekpyron, line 6: Syntactically this is really not interesting - this should be a semantics test!

hrkrshnn: I can move this.

hrkrshnn: Did it here: https://github.com/ethereum/solidity/pull/11914/

fulldecent (commented):
The fixed math library claims the division function is to the "nearest integer" but actually it rounds down.

hrkrshnn:
@fulldecent Thanks for noticing. Will update the comment.

fulldecent:
Thank you! Looks good now

Issue #11629, opened 2021-07-07
[CLI] Report errors for command line options that have no effect


cameel:
## Abstract
The compiler is pretty lax about command-line options whose values it does not use and just ignores many of them. It should report errors instead.

Over time we've started adding checks against this (e.g. #9075, #9364) they they're still very incomplete.

## Motivation
Not getting an error when an option is invalid makes it harder for the user to realize that. It's especially problematic with options that seem to make sense together:
- `--pretty-json` sounds like it would affect `--standard-json` but it only works with `--combined-json` (#11583.) 
- `--bin` should be supported in the assembler mode but currently isn't (#3870).

## Specification
-  There are several input modes (compiler, linker, assembler, standard JSON) and many options are not valid in one of these modes. The invalid ones should produce an error.
    - Take a look at [tests for input modes in `CommandLineParser`](https://github.com/ethereum/solidity/blob/develop/test/solc/CommandLineParser.cpp). Basically, all options do not result in a change in `CommandLineOptions` in a specific mode should be invalid in that mode.
- Options that have equivalents and Standard JSON should be errors when `--standard-json` is used. For example:
    - Model checker settings.
    - Remappings.
    - `--color`/`--no-color` - Standard JSON output is not colorized and even when it will be (#11507), this option should not affect it.
    - `--libraries`.
    - `--error-codes`.
- Please also check options whose values are not ignored by `CommandLineParser` but are still not used by `CommandLineInterface`.

## Backwards Compatibility
This does not introduce any backwards-incompatible changes to the language itself but any command-line tools relying on the current behavior of options being ignored will be affected.

CRYONE7:
TEACH ME MOTHER FUCKER!
[...]

PR #11807, opened 2021-08-16
Add support for importing asm-json.

### LINKED ISSUE #11787: Compiler mode to reimport assembly json.
The compiler can output the state of ``Assembly`` containing evmasm assembly items (``AssemblyItem``) in json using ``solc --asm-json``. (The main exporting logic is in ``Assembly::assemblyJSON``)
Since there are projects that use this assembly json, externally optimize it, and output a new, compatible assembly json, we want to be able to reimport assembly in this format.

So we want a command line interface mode, e.g. ``solc --asm-json-import`` that can parse the output of ``solc --asm-json`` and can continue compilation down to bytecode.

Exporting assembly json and then importing it, should produce exactly the same bytecode as compiling directly to bytecode.

The assembly json will contain nested subassemblies, auxiliary data and tags, etc., which should be sufficient to fill all members of ``Assembly`` correctly.

aarlt:
Closes #11787.

- [X] the import can be done with `--import-asm-json`.
- [X]  support for `--opcodes`, `--asm` and `--bin` and `--bin-runtime`, if `--import-asm-json` is used.
- [x] add additional tests that reimport the following assembly items:
  - [x] ~`PUSH tag`~
  - [x] `PUSH [ErrorTag]`
  - [x] `PUSHLIB`
  - [x] `PUSHDEPLOYADDRESS`
  - [x] `PUSHIMMUTABLE`
  - [x] `VERBATIM`
 - [x] support `combinedJson` for `--import-asm-json`
   - [x] `srcmap`, `srcmap-runtime`
     - [x] add import tests
   - [x] check for invalid `combinedJson` parameters
 - [x] support optimisation
 - [ ] `AsmJsonImportTest.sh`: also compare optimiser output

Note: 
- `PUSH tag` got removed, see #11948


aarlt:
The generated byte code after importing asm-json file is not yet the same. 

E.g.:

- exporting asm json `solc ../test/libsolidity/semanticTests/constructor_with_params.sol --asm-json > /tmp/test.json`. (Ensure that `/tmp/test.json` was edited, so it holds a valid json file)

- compare `solc ../test/libsolidity/semanticTests/constructor_with_params.sol --opcodes` with `solc --import-asm-json /tmp/test.json --opcodes`

`
PUSH1 0x80 PUSH1 0x40 MSTORE CALLVALUE DUP1 ISZERO PUSH2 0x10 JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP PUSH1 0x40 MLOAD PUSH2 0x1D1 CODESIZE SUB DUP1 PUSH2 0x1D1 DUP4 CODECOPY DUP2 DUP2 ADD PUSH1 0x40 MSTORE DUP2 ADD SWAP1 PUSH2 0x32 SWAP2 SWAP1 PUSH2 0x5C JUMP JUMPDEST DUP2 PUSH1 0x0 DUP2 SWAP1 SSTORE POP DUP1 PUSH1 0x1 DUP2 SWAP1 SSTORE POP POP POP PUSH2 0xC2 JUMP JUMPDEST PUSH1 0x0 DUP2 MLOAD SWAP1 POP PUSH2 0x56 DUP2 PUSH2 0xAB JUMP JUMPDEST SWAP3 SWAP2 POP POP JUMP JUMPDEST PUSH1 0x0 DUP1 PUSH1 0x40 DUP4 DUP6 SUB SLT ISZERO PUSH2 0x73 JUMPI PUSH2 0x72 PUSH2 0xA6 JUMP JUMPDEST JUMPDEST PUSH1 0x0 PUSH2 0x81 DUP6 DUP3 DUP7 ADD PUSH2 0x47 JUMP JUMPDEST SWAP3 POP POP PUSH1 0x20 PUSH2 0x92 DUP6 DUP3 DUP7 ADD PUSH2 0x47 JUMP JUMPDEST SWAP2 POP POP SWAP3 POP SWAP3 SWAP1 POP JUMP JUMPDEST PUSH1 0x0 DUP2 SWAP1 POP SWAP2 SWAP1 POP JUMP JUMPDEST PUSH1 0x0 DUP1 REVERT JUMPDEST PUSH2 0xB4 DUP2 PUSH2 0x9C JUMP JUMPDEST DUP2 EQ PUSH2 0xBF JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP JUMP JUMPDEST` 

~**`PUSH2 0x102`**~ **`PUSH1 0xA7`** 

`DUP1 PUSH2 0xD0 PUSH1 0x0 CODECOPY PUSH1 0x0 RETURN INVALID PUSH1 0x80 PUSH1 0x40 MSTORE CALLVALUE DUP1 ISZERO PUSH1 0xF JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP PUSH1 0x4 CALLDATASIZE LT PUSH1 0x32 JUMPI PUSH1 0x0 CALLDATALOAD PUSH1 0xE0 SHR DUP1 PUSH4 0xB4F40C61 EQ PUSH1 0x37 JUMPI DUP1 PUSH4 0xE5AA3D58 EQ PUSH1 0x51 JUMPI JUMPDEST PUSH1 0x0 DUP1 REVERT JUMPDEST PUSH1 0x3D PUSH1 0x6B JUMP JUMPDEST PUSH1 0x40 MLOAD PUSH1 0x48 SWAP2 SWAP1 PUSH1 0x84 JUMP JUMPDEST PUSH1 0x40 MLOAD DUP1 SWAP2 SUB SWAP1 RETURN JUMPDEST PUSH1 0x57 PUSH1 0x71 JUMP JUMPDEST PUSH1 0x40 MLOAD PUSH1 0x62 SWAP2 SWAP1 PUSH1 0x84 JUMP JUMPDEST PUSH1 0x40 MLOAD DUP1 SWAP2 SUB SWAP1 RETURN JUMPDEST PUSH1 0x1 SLOAD DUP2 JUMP JUMPDEST PUSH1 0x0 SLOAD DUP2 JUMP JUMPDEST PUSH1 0x7E DUP2 PUSH1 0x9D JUMP JUMPDEST DUP3 MSTORE POP POP JUMP JUMPDEST PUSH1 0x0 PUSH1 0x20 DUP3 ADD SWAP1 POP PUSH1 0x97 PUSH1 0x0 DUP4 ADD DUP5 PUSH1 0x77 JUMP JUMPDEST SWAP3 SWAP2 POP POP JUMP JUMPDEST PUSH1 0x0 DUP2 SWAP1 POP SWAP2 SWAP1 POP JUMP` 

~**`INVALID`**~ 

 `LOG2 PUSH5 0x6970667358 0x22 SLT KECCAK256 0x2D LOG4 0x4B DUP16 SMOD 0xF6 SWAP1 0xD JUMPI MULMOD MLOAD 0xF6 AND 0x5E 0xED STOP MUL 0xD8 0x4B 0xEF 0xD 0xEB DUP14 SUB 0xDD 0xC3 0xC JUMPI 0xD8 0xBF LOG4 0x28 PUSH5 0x736F6C6378 0x27 ADDRESS 0x2E CODESIZE 0x2E CODESIZE 0x2D PUSH5 0x6576656C6F PUSH17 0x2E323032312E382E31362B636F6D6D6974 0x2E CODECOPY PUSH4 0x61663734 CALLDATALOAD PUSH7 0x580000000000
`

libevmasm/Assembly.cpp
```diff
@@ -227,6 +229,214 @@ string Assembly::toStringInHex(u256 _value)
...
+			"ASSIGNIMMUTABLE",
+			_sourceIndex,
+			_item.location().start,
+			_item.location().end,
+			m_immutables.at(h256(_item.data()))));
+		break;
+	case Tag:
+		result.emplace_back(
+			createJsonValue("tag", _sourceIndex, _item.location().start, _item.location().end, toString(_item.data())));
+//		result.emplace_back(createJsonValue("JUMPDEST", _sourceIndex, _item.location().start, _item.location().end));
```
aarlt: It looks like that this `JUMPDEST` is not really needed, if I generated this here, I will see `JUMPDEST JUMPDEST` in the resulting byte code that seem to be not correct.

libevmasm/Assembly.cpp
```diff
@@ -242,89 +452,8 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
-				sourceIndex,
-				i.location().start,
-				i.location().end,
-				m_immutables.at(h256(i.data()))
-			));
-			break;
-		case Tag:
-			collection.append(
-				createJsonValue("tag", sourceIndex, i.location().start, i.location().end, toString(i.data())));
-			collection.append(
```
aarlt, line 329: Was this really correct?

solc/CommandLineParser.cpp
```diff
@@ -578,6 +579,11 @@ General Information)").c_str(),
 			"Supported Inputs is the output of the --" + g_strStandardJSON + " or the one produced by "
 			"--" + g_strCombinedJson + " " + g_strAst + "," + g_strCompactJSON).c_str()
 		)
+		(
+			g_strImportAssemblyJson.c_str(),
+			("Import assembly to be used, assumes input holds assembly in JSON format. "
+			 "Supported Input is the output of the --" + g_strAsmJson + ".").c_str()
```
chriseth: This has to be clarified a bit more, I think.

libevmasm/AssemblyItem.cpp
```diff
@@ -179,6 +179,18 @@ string AssemblyItem::getJumpTypeAsString() const
...
+void AssemblyItem::setJumpTypeFromString(std::string const& _jumpType)
+{
+	if (_jumpType == "[in]")
+		m_jumpType = JumpType::IntoFunction;
+	else if (_jumpType == "[out]")
+		m_jumpType = JumpType::OutOfFunction;
+	else if (_jumpType.empty())
+		m_jumpType = JumpType::Ordinary;
+	else
+		assertThrow(!_jumpType.empty(), AssemblyException, "Invalid jump type.");
```
chriseth:
```suggestion
		assertThrow(false, AssemblyException, "Invalid jump type.");
```

libevmasm/Assembly.cpp
chriseth: What do you mean? It seems that every tag has to be followed by a jumpdest.

aarlt: Sure. This will generate something like:
```json
[{
  "begin": 1079,
  "end": 1201,
  "name": "tag",
  "source": 1,
  "value": "11"
},
{
  "begin": 1079,
  "end": 1201,
  "name": "JUMPDEST",
  "source": 1
}]
```
in the resulting json file. 

However, the problem is that `Assembly::assemble()` also adds a `JUMPDEST`, if it hits `Tag`.

```c++
LinkerObject const& Assembly::assemble() const
{
  ...
		case Tag:
		{
			assertThrow(i.data() != 0, AssemblyException, "Invalid tag position.");
			assertThrow(i.splitForeignPushTag().first == numeric_limits<size_t>::max(), AssemblyException, "Foreign tag.");
			size_t tagId = static_cast<size_t>(i.data());
			assertThrow(ret.bytecode.size() < 0xffffffffL, AssemblyException, "Tag too large.");
			assertThrow(m_tagPositionsInBytecode[tagId] == numeric_limits<size_t>::max(), AssemblyException, "Duplicate tag position.");
			m_tagPositionsInBytecode[tagId] = ret.bytecode.size();
			ret.bytecode.push_back(static_cast<uint8_t>(Instruction::JUMPDEST));
			break;
		}
   ...
}
```

Not sure how to deal with this in a non-breaking way.

aarlt: I think we should not generate the `JUMPDEST` within in the asm json. But that could break some external tools (is it used anywhere?)

solc/CommandLineParser.cpp
aarlt: Maybe, what should be clarified a bit more?

libevmasm/Assembly.cpp
aarlt: https://github.com/ethereum/solidity/pull/11807#discussion_r690519736

libevmasm/Assembly.cpp
```diff
@@ -348,6 +477,61 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+			{
+				AssemblyItem item = assemblyItemFromJSON(e);
+				if (item.type() == AssemblyItemType::Tag)
+				{
+					auto tag = static_cast<unsigned>(item.data());
+					if (subassembly->m_usedTags <= tag)
+						subassembly->m_usedTags = tag + 1;
+				}
+				subassembly->m_items.emplace_back(item);
+			}
```
ekpyron: This is pretty much the same as a few lines up, isn't it? Might be nice to unify...
Also doing sth like
```suggestion
			for (auto it = code.begin(); it != code.end(); ++it)
			{
			    auto const& e = *it;
				AssemblyItem item = assemblyItemFromJSON(e);
				if (item.type() == AssemblyItemType::Tag)
				{
					auto tag = static_cast<unsigned>(item.data());
					if (subassembly->m_usedTags <= tag)
						subassembly->m_usedTags = tag + 1;
					subassembly->m_items.emplace_back(item);
					if (std::next(it) != code->end() && *std::next(it) == AssemblyItem(Instruction::JumpDest)) // or however those are compared
					   ++it;
				}
				subassembly->m_items.emplace_back(item);
			}
```
would be *a* solution to the jumpdest issue, wouldn't it? Not sure it's the best one, though.

libevmasm/Assembly.cpp
aarlt: Ok, I just ignore `JUMPDEST` for now, if loaded from json. Not (yet) sure whether it can be always ignored.

aarlt:
Ok, finally I found out what created the mismatching bytecode. I added the auxiliary data to the wrong assembly object.

libevmasm/Assembly.cpp
aarlt: Ah nice thx! I also ended up just ignoring all `JUMPDEST`, if they where coming from json. I'm not sure whether this is always correct. I guess I will know after implementing reimport tests for all semantic tests.

libevmasm/Assembly.cpp
```diff
@@ -348,6 +474,58 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	solAssert(m_auxiliaryData.empty(), "");
+	solAssert(m_usedTags == 1, "");
+
+	Json::Value assemblyJson;
+	bool success = util::jsonParseStrict(_source, assemblyJson);
+	if (success)
+	{
+		forEachJSONAssemblyItem(assemblyJson[".code"], *this);
+		Json::Value const& data = assemblyJson[".data"];
+		for(Json::ValueConstIterator itr = data.begin(); itr != data.end(); itr++)
```
stackenbotten, line 508: Coding style error

solc/CommandLineParser.cpp
stackenbotten, line 585: Coding style error

libevmasm/Assembly.cpp
ekpyron: That was mainly my point: if you just skip the ``JUMPDEST``s immediately following tags in this loop, then we can be sure that it's fine, since you'll still get any further stray ``JUMPDEST``s :-). (Although I could very well imagine that there are none anyways.)

ekpyron:
If exporting and importing anything really works and produces identical bytecode everywhere, it would probably be nice to add a test run similar to the AST import tests, that just compiles all sources in the repo once directly to bytecode and once via exporting assembly and then importing it back first, comparing the resulting bytecode :-).
If we do that, then one more dedicated test with an artificial assembly JSON containing one of each possible assembly items would probably be enough.

solc/CommandLineParser.cpp
stackenbotten, line 585: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -580,6 +581,11 @@ General Information)").c_str(),
...
```
stackenbotten, line 587: Coding style error

aarlt:
After checking what assembly items where exactly imported by `AsmJsonImportTest.sh`, it turned out that the following assembly items never get imported due to missing tests.

- `PUSH tag`
- `PUSH [ErrorTag]`
- `PUSHLIB`
- `PUSHDEPLOYADDRESS`
- `PUSHIMMUTABLE`
- `VERBATIM`

TODO: Add specific tests for these assembly items.

libevmasm/Assembly.cpp
```diff
@@ -342,12 +476,59 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	solAssert(_code.isArray(), "");
+	for (auto it = _code.begin(); it != _code.end(); ++it)
+	{
+		AssemblyItem current = assemblyItemFromJSON(*it, _assembly);
+		_assembly.m_items.emplace_back(current);
+
+		auto nextIt = std::next(it);
+		if (nextIt != _code.end())
+		{
+			AssemblyItem next = assemblyItemFromJSON(*nextIt, _assembly);
```
aarlt: A comment here describing why this was done can not hurt.

libevmasm/Assembly.cpp
```diff
@@ -342,12 +476,59 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
 	}
 
-	if (m_auxiliaryData.size() > 0)
+	if (!m_auxiliaryData.empty())
 		root[".auxdata"] = toHex(m_auxiliaryData);
 
 	return root;
 }
 
+void Assembly::forEachJSONAssemblyItem(Json::Value const& _code, Assembly& _assembly)
```
aarlt: Maybe rename this function.

aarlt:
Added json re-import - and byte code equivalence test to `test/libevmasm/Assembler.cpp`.
Still missing assembly items:
- PUSH tag
- PUSH [ErrorTag]
- VERBATIM

libevmasm/Assembly.cpp
```diff
@@ -227,6 +229,221 @@ string Assembly::toStringInHex(u256 _value)
...
+	auto updateImmutables = [&](u256 const& data) -> h256 {
+		std::string value = toStringInHex(data);
+		h256 hash(util::keccak256(value));
+		_assembly.m_immutables[hash] = asString(fromHex(value));
+		return hash;
+	};
+
+	auto updateLibraries = [&](string const& value) -> h256 {
+		h256 hash(util::keccak256(value));
+		_assembly.m_libraries[hash] = value;
```
cameel: We might want to check that it's a valid address (length, chars, maybe checksum?)

libevmasm/Assembly.cpp
cameel: `addAssemblyItemsFromJSON()`?

libevmasm/Assembly.cpp
```diff
@@ -342,12 +480,57 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	{
+		AssemblyItem current = assemblyItemFromJSON(*it, _assembly);
+		_assembly.m_items.emplace_back(current);
+
+		auto nextIt = std::next(it);
+		if (nextIt != _code.end())
+		{
+			AssemblyItem next = assemblyItemFromJSON(*nextIt, _assembly);
+			if (current.type() == AssemblyItemType::Tag &&
+				next.type() == AssemblyItemType::Operation && next.instruction() == Instruction::JUMPDEST)
```
cameel:
```suggestion
			if (
				current.type() == AssemblyItemType::Tag &&
				next.type() == AssemblyItemType::Operation &&
				next.instruction() == Instruction::JUMPDEST
			)
```

libevmasm/Assembly.cpp
```diff
@@ -227,6 +229,221 @@ string Assembly::toStringInHex(u256 _value)
 	return hexStr.str();
 }
 
+AssemblyItem Assembly::assemblyItemFromJSON(Json::Value const& _json, Assembly& _assembly)
```
cameel: At first I thought this just returns a new item but then I noticed that the function actually modifies `_assembly`. I think the name should reflect that.

Actually, maybe a better design would be to have two functions - one that just gives you a list of items (and does not modify the assembly) and another one that takes the list and updates the assembly based on it? If you had to explicitly run the update function, it would be clearer that an update is taking place.

libevmasm/Assembly.cpp
```diff
@@ -342,12 +480,57 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
 	}
 
-	if (m_auxiliaryData.size() > 0)
+	if (!m_auxiliaryData.empty())
 		root[".auxdata"] = toHex(m_auxiliaryData);
 
 	return root;
 }
 
+void Assembly::forEachJSONAssemblyItem(Json::Value const& _code, Assembly& _assembly)
```
cameel: With the change I suggested above this function would get a list of assembly items. You would not have to call `assemblyItemFromJSON()` twice for each item (which is especially bad given that `assemblyItemFromJSON` has side-effects - it modifies the assembly).

libevmasm/Assembly.h
```diff
@@ -169,6 +171,10 @@ class Assembly
 
 	unsigned bytesRequired(unsigned subTagSize) const;
 
+	static AssemblyItem assemblyItemFromJSON(Json::Value const& _json, Assembly& _assembly);
+	static void forEachJSONAssemblyItem(Json::Value const& _code, Assembly& _assembly);
```
cameel: `forEachJSONAssemblyItem()` looks like an implementation detail that should be private.

libevmasm/AssemblyItem.h
```diff
@@ -160,6 +160,7 @@ class AssemblyItem
 	langutil::SourceLocation const& location() const { return m_location; }
 
 	void setJumpType(JumpType _jumpType) { m_jumpType = _jumpType; }
+	void setJumpTypeFromString(std::string const& _jumpType);
```
cameel: These two have essentially the same effect (setting jump type). I'd use an overload.
```suggestion
	void setJumpType(std::string const& _jumpType);
```

liblangutil/SourceLocation.h
```diff
@@ -102,6 +103,7 @@ struct SourceLocation
 	int start = -1;
 	int end = -1;
 	std::shared_ptr<std::string const> sourceName;
+	std::optional<int> source = std::nullopt;
```
cameel: `source` sounds like something that would contain the source code.
```suggestion
	std::optional<int> sourceIndex = std::nullopt;
```

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
```
cameel, line 1: There's  a ton of duplication between this and `ASTImportTest.sh`. Maybe we could extract common parts?

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
...
+
+set -e
```
cameel:
```suggestion
set -eu
```

Or even this (if it works):
```suggestion
set -euo pipefail
```

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
...
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if $SOLC "$nth_input_file" "${all_input_files[@]}" > /dev/null 2>&1
+    then
+        # save exported json as expected result (silently)
+        $SOLC --combined-json asm,bin,compact-format --pretty-json "$nth_input_file" "${all_input_files[@]}" > expected.json 2> /dev/null
+        for contract in $(jq '.contracts | keys' expected.json | jq '.[]' 2> /dev/null)
```
cameel: This adds a dependency on `jq`. We should list that in the docs about setting up dev environment and also ensure that we have it available in all images in CI. I guess we do since I do not see CI failing on Windows or macOS for this PR.

It's nice though. Now I'll be able to use it in other shell scripts :)

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
...
+        do
+            jq ".contracts.${contract}.asm" expected.json > asm.json 2> /dev/null
+            jq ".contracts.${contract}.bin" expected.json | jq --raw-output '.' > expected.bin 2> /dev/null
```
cameel:
```suggestion
        for contract in $(jq '.contracts | keys | .[]' expected.json 2> /dev/null)
        do
            jq ".contracts.${contract}.asm" expected.json > asm.json 2> /dev/null
            jq --raw-output ".contracts.${contract}.bin" expected.json > expected.bin 2> /dev/null
```

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
...
+            expected_bin=$(cat expected.bin)
+            if [ -z "$expected_bin" ]
```
cameel:
```suggestion
            if [[ $expected_bin == "" ]]
```

scripts/AsmJsonImportTest.sh
```diff
@@ -0,0 +1,160 @@
...
+            then
+              continue
+            fi
+
+            if ! $SOLC --import-asm-json asm.json --bin > obtained.bin 2> /dev/null
```
cameel:
```suggestion
            if ! "$SOLC" --import-asm-json asm.json --bin > obtained.bin 2> /dev/null
```

solc/CommandLineInterface.cpp
```diff
@@ -828,14 +839,15 @@ void CommandLineInterface::handleAst()
 
 bool CommandLineInterface::actOnInput()
 {
-	if (m_options.input.mode == InputMode::StandardJson || m_options.input.mode == InputMode::Assembler)
+	if (m_options.input.mode == InputMode::StandardJson ||
+		m_options.input.mode == InputMode::Assembler)
```
```suggestion
	if (
		m_options.input.mode == InputMode::StandardJson ||
		m_options.input.mode == InputMode::Assembler
	)
```

solc/CommandLineInterface.cpp
```diff
@@ -1087,61 +1099,91 @@ bool CommandLineInterface::assemble(
...
 }
 
+bool CommandLineInterface::assembleFromAssemblyJson()
+{
+	solAssert(m_fileReader.sourceCodes().size() == 1, "");
+
+	Json::Value assemblyJson;
+	m_assembly = make_unique<evmasm::Assembly>();
+	return util::jsonParseStrict(m_fileReader.sourceCodes().begin()->second, assemblyJson) &&
+				m_assembly->loadFromAssemblyJSON(assemblyJson, *m_assembly);
```
```suggestion
	return
		util::jsonParseStrict(m_fileReader.sourceCodes().begin()->second, assemblyJson) &&
		m_assembly->loadFromAssemblyJSON(assemblyJson, *m_assembly);
```

cameel: I think it would be better not to create an empty assembly if the function fails.

solc/CommandLineInterface.cpp
```diff
@@ -1087,61 +1099,91 @@ bool CommandLineInterface::assemble(
...
 
-	vector<string> contracts = m_compiler->contractNames();
-	for (string const& contract: contracts)
+	if (m_compiler)
 	{
-		if (needsHumanTargetedStdout(m_options))
-			sout() << endl << "======= " << contract << " =======" << endl;
+		solAssert(m_assembly == nullptr, "");
+		if (!m_compiler->compilationSuccessful()
+			&& m_options.output.stopAfter == CompilerStack::State::CompilationSuccessful)
```
```suggestion
		if (
			!m_compiler->compilationSuccessful() &&
			m_options.output.stopAfter == CompilerStack::State::CompilationSuccessful
		)
```

solc/CommandLineInterface.cpp
```diff
@@ -1087,61 +1099,91 @@ bool CommandLineInterface::assemble(
...
+					ret = jsonPrettyPrint(removeNullMembers(m_compiler->assemblyJSON(contract)));
+				else
+					ret = m_compiler->assemblyString(contract, m_fileReader.sourceCodes());
+
+				if (!m_options.output.dir.empty())
+				{
+					createFile(
+						m_compiler->filesystemFriendlyName(contract)
+							+ (m_options.compiler.outputs.asmJson ? "_evm.json" : ".evm"),
+						ret);
```
```suggestion
						ret
					);
```

solc/CommandLineInterface.cpp
```diff
@@ -1087,61 +1099,91 @@ bool CommandLineInterface::assemble(
...
 	handleCombinedJSON();
 
 	// do we need AST output?
 	handleAst();
 
-	if (
-		!m_compiler->compilationSuccessful() &&
-		m_options.output.stopAfter == CompilerStack::State::CompilationSuccessful
-	)
+	if (m_assembly)
```
cameel: Can `m_assembly` and `m_compiler` both be present at the same time?

solc/CommandLineParser.cpp
```diff
@@ -1187,7 +1195,7 @@ General Information)").c_str(),
 	if (m_options.input.mode == InputMode::Compiler)
 		m_options.input.errorRecovery = (m_args.count(g_strErrorRecovery) > 0);
 
-	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
+	solAssert(m_options.input.mode == InputMode::AssemblyJson || m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
```
cameel: Does this new input mode support all of the arguments that the compiler does? If not, I think it should make the function return at a much earlier point.

Also, which other command-line flags are usable in this mode? There's an issue underway to report errors for flags invalid in other modes (#11629) and I think we should start rejecting invalid flags in this new mode from the very beginning.

Each input mode has a test in `test/solc/CommandLineParser.cpp` that checks all the options valid in that mode. This new mode should have such a test too.

solc/CommandLineParser.h
```diff
@@ -49,6 +49,7 @@ enum class InputMode
 	StandardJson,
 	Linker,
 	Assembler,
+	AssemblyJson,
```
cameel: Assembly is a bit ambiguous. On the CLI it usually refers to Yul. Maybe we should call it `EVMAssemblyJson` instead? Or rename `Assembler` mode to `Yul`?

liblangutil/SourceLocation.h
chriseth: Is this redundancy required? Can't we just store the source name?

aarlt: Yeah that would be perfect. The problem is that we don't have the source name available in the json, only the source index.

aarlt: But maybe we could just change the structure of the assembly json. But I guess that would be breaking.

scripts/AsmJsonImportTest.sh
aarlt: interesting that `shellcheck` was not seeing this

solc/CommandLineInterface.cpp
aarlt: No, only `m_assembly` or `m_compiler` is possible at the same time. Change this to `else if`.

libevmasm/Assembly.cpp
```diff
@@ -342,12 +480,54 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	if (!m_auxiliaryData.empty())
 		root[".auxdata"] = toHex(m_auxiliaryData);
 
 	return root;
 }
 
+void Assembly::addAssemblyItemsFromJSON(Json::Value const& _code)
+{
+	solAssert(_code.isArray(), "");
+	for (auto const& it : _code)
```
stackenbotten: Coding style error

libevmasm/Assembly.cpp
aarlt: This is not using any addresses here. `value` is just a name of a library (I will change the parameter name). The hash is just used as a placeholder, where linking will replace the placeholder with the actual library address.

aarlt: I guess linking will do some checks on the supplied addresses, but I haven't checked this yet.

scripts/AsmJsonImportTest.sh
cameel: Yeah. In some cases quotes can be omitted, e.g. `x=$y`, but this one seems like a clear case where they should be required.

libevmasm/Assembly.cpp
cameel: Ah, right, these are the placeholders.

Still, the placeholders need to have a specific length, start and end with `__`, etc. Linking is done in [`LinkerObject.cpp`](https://github.com/ethereum/solidity/blob/develop/libevmasm/LinkerObject.cpp) and it does not seem to have any validations (or even assertions) against that. AFAIK all the validation is in the CLI and `StandardCompiler`.

BTW, some other parts of this JSON structure might need some validation too.

solc/CommandLineParser.h
aarlt: What about `ImportAssemblyJson`?

cameel: The problem I see is just that it might not be clear that `Assembly` refers to EVM assembly rather than Yul here. On the other hand the class is named `Assembly`, so maybe it's not such a big problem...

aarlt:
removed:
- `PUSH tag`
  - removed, see #11948
 
simple tests added:

- `PUSH [ErrorTag]`
- `VERBATIM`

solc/CommandLineParser.cpp
```diff
@@ -1197,19 +1236,20 @@ General Information)").c_str(),
...
-		m_args.count(g_strModelCheckerContracts) ||
-		m_args.count(g_strModelCheckerDivModNoSlacks) ||
-		m_args.count(g_strModelCheckerEngine) ||
-		m_args.count(g_strModelCheckerShowUnproved) ||
-		m_args.count(g_strModelCheckerSolvers) ||
-		m_args.count(g_strModelCheckerTargets) ||
-		m_args.count(g_strModelCheckerTimeout);
+	m_options.modelChecker.initialize
+		= m_args.count(g_strModelCheckerContracts) || m_args.count(g_strModelCheckerDivModNoSlacks)
+		  || m_args.count(g_strModelCheckerEngine) || m_args.count(g_strModelCheckerShowUnproved)
```
stackenbotten: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -1197,19 +1236,20 @@ General Information)").c_str(),
...
+		  || m_args.count(g_strModelCheckerSolvers) || m_args.count(g_strModelCheckerTargets)
```
stackenbotten: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -1197,19 +1236,20 @@ General Information)").c_str(),
...
+		  || m_args.count(g_strModelCheckerTimeout);
```
stackenbotten: Coding style error

solc/CommandLineParser.cpp
aarlt: I added a very simple check in `solc/CommandLineParser.cpp`. Basically with the new import mode only `--asm`, `--bin` and `--opcodes` are right now supported.

solc/CommandLineParser.h
aarlt: I just changed that now to `ImportEVMAssemblyJson`.

leonardoalt:
Needs rebase. Besides that, is it ready for review again?

solc/CommandLineParser.cpp
```diff
@@ -641,8 +657,12 @@ General Information)").c_str(),
 		(
 			g_strImportAst.c_str(),
 			("Import ASTs to be compiled, assumes input holds the AST in compact JSON format. "
-			"Supported Inputs is the output of the --" + g_strStandardJSON + " or the one produced by "
-			"--" + g_strCombinedJson + " " + CombinedJsonRequests::componentName(&CombinedJsonRequests::ast)).c_str()
+			 "Supported Inputs is the output of the --" + g_strStandardJSON + " or the one produced by "
```
stackenbotten, line 660: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -641,8 +657,12 @@ General Information)").c_str(),
...
+																			  "--" + g_strCombinedJson + " " + CombinedJsonRequests::componentName(&CombinedJsonRequests::ast)).c_str()
```
stackenbotten, line 661: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
 			ranges::views::keys |
 			ranges::to<set>()
 		) - set<string>{CompilerOutputs::componentName(&CompilerOutputs::ewasmIR)};
-		static set<string> const assemblerModeOutputs = {
+	  static set<string> const assemblerModeOutputs = {
```
stackenbotten, line 471: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+		  CompilerOutputs::componentName(&CompilerOutputs::asm_),
```
stackenbotten, line 472: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+		  CompilerOutputs::componentName(&CompilerOutputs::binary),
```
stackenbotten, line 473: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+		  CompilerOutputs::componentName(&CompilerOutputs::irOptimized),
```
stackenbotten, line 474: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+		  CompilerOutputs::componentName(&CompilerOutputs::ewasm),
```
stackenbotten, line 475: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+		  CompilerOutputs::componentName(&CompilerOutputs::ewasmIR),
```
stackenbotten, line 476: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,12 +468,18 @@ bool CommandLineParser::parseOutputSelection()
...
+	  };
```
stackenbotten, line 477: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -1329,7 +1367,14 @@ bool CommandLineParser::parseCombinedJsonOption()
 
 	set<string> requests;
 	for (string const& item: boost::split(requests, m_args[g_strCombinedJson].as<string>(), boost::is_any_of(",")))
-		if (CombinedJsonRequests::componentMap().count(item) == 0)
+		if (m_options.input.mode == InputMode::CompilerWithEvmAssemblyJsonImport &&
+			CombinedJsonRequests::componentMapAssemblyJsonImport().count(item) == 0)
+		{
+			serr() << "Invalid option to --" << g_strCombinedJson << ": " << item << ", for --"
+				   << g_strImportEvmAssemblerJson << endl;
```
stackenbotten, line 1374: Coding style error

aarlt:
Replaced by https://github.com/ethereum/solidity/pull/12704.

Issue #9582, opened 2020-08-06
Remove or fix install_deps.sh


ekpyron:
It's horribly outdated for most distributions. And we don't ever test it.

We just had a report that it doesn't recognize CentOS, even though there is a case for it in the script - but even if it *was* recognized, it would install weird versions of things from weird sources, not install smt checkers, etc.

I vote for removing it and instead making sure to properly document the dependencies (should be the case though).

Somewhat related to https://github.com/ethereum/solidity/issues/8860 - and getting rid of boost :-).

chriseth:
I agree, we should not have any scripts that are fully untested. Furthermore, I always find it scary to run scripts that modify my system.

Zachinquarantine:
can I have this assigned to me? I'd like to work on it.

cameel:
@Zachinquarantine Great! This might not be the easiest task to finish because it's a bit open-ended. We need to agree on what we want to do here. It should not be technically hard though which is why I marked it as `good first issue` anyway.

So, the first question is if we want to keep the script at all. We're getting PRs from people using it from time to time so I guess it must useful to someone. So my vote would be for keeping it but making it more maintainable. Reducing its scope would be a good step in that direction and #11808 looks good in these terms.

I think that one uncontroversial part is that we'd accept PRs that improve the script on specific systems where the submitter can run it locally and verify his changes.

A big open question is how to test it as a whole. I don't think we can realistically maintain CI checks for the whole range of systems it's meant to support but we could be testing it at least on the ones we do have. So adding CI jobs for it would be another thing to do here if we agree on that.

The rest mostly depends on whether this script is even worth more effort. Maybe we could crowd-source it and find some people running various OSes who would be willing to at least periodically check that it works for them and report back? Or maybe periodically create issues asking a larger group of people to verify it on their systems. First though we should somehow gauge interest in that. I suspect that the vast majority of users get the compiler either through tools/frameworks or directly from solc-bin and the script is not relevant to them at all.

ekpyron:
As already stated in the issue I strongly vote for just removing it :-). People are only using it, since we mention it in the docs, if we just specifically list the dependencies there trusting that people are able to install them on their respective platform, then people won't use it anymore. And expecting users to just run a huge intransparent script with root privileges is rather insane anyways :-).

cameel:
> And expecting users to just run a huge intransparent script with root privileges is rather insane anyways :-).

[...]

PR #11808, opened 2021-08-16
Removes support for old OS versions in install_deps.sh and adds support for new ones...

tqpcharlie:
I know that with #9582, this file will probably get removed, but still...

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -267,30 +267,9 @@ case $(uname -s) in
                         echo "Installing solidity dependencies on Ubuntu Trusty Tahr (14.04)."
                         echo "Or, you may also be running Linux Mint Qiana / Rebecca / Rafaela / Rosa (base: Ubuntu Trusty Tahr (14.04).)"
                         ;;
-                    utopic)
-                        echo "Installing solidity dependencies on Ubuntu Utopic Unicorn (14.10)."
-                        ;;
-                    vivid)
-                        echo "Installing solidity dependencies on Ubuntu Vivid Vervet (15.04)."
-                        ;;
-                    wily)
-                        echo "Installing solidity dependencies on Ubuntu Wily Werewolf (15.10)."
-                        ;;
                     xenial|sarah|serena|sonya|sylvia)
                         echo "Installing solidity dependencies on Ubuntu Xenial Xerus (16.04)."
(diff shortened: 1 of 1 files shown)


cameel:
Looks good overall but I wonder if we really do want to remove all those old Ubuntu versions. Out of all systems, Ubuntu is probably the easiest for us to support since we use it a lot in CI and if they're not broken, we could just keep them there.

Also, not sure why Windows jobs are failing in CI. Its only in some PRs. I've restarted them and hopefully they will pass. The failure is definitely not related to the PR anyway.

leonardoalt:
I think this PR makes sense. IMO we should remove even more :p
@ekpyron wdyt?

cameel:
In any case this needs a rebase on `develop` for CI to pass. The windows/boost problem is fixed there.

These removed versions are very old so I don't have all that much against merging this but overall I think the next step in this task should be listing these dependencies in the docs (https://github.com/ethereum/solidity/issues/9582#issuecomment-900265173).

ekpyron:
As for ubuntu versions: we also don't support building on Trusty, Xenial and Bionic (we do provide static PPA builds for those, but we don't support building on them)... but yeah, as mentioned elsewhere, I think we should remove the entire script anyways and I'm not sure why we don't just do that :-). But if we don't, reducing the versions here is fine, but yeah: Trusty, Xenial and Bionic are also unsupported.

chriseth:
I'm also voting for removing the script.

cameel:
@chriseth Do we want to at least get a list of deps for each platform in the docs before removing the file?

If we're just going to remove the script then I guess we can just close this PR.

chriseth:
I kind of have the impression that if you compile yourself, you should also know how to install basic dependencies. Since we don't test the script, I don't think it makes sense to keep it. We can give some recommendations in the documentation, but not sure if we should be too detailed, like we should not tell how to use the package manager, but that you need boost and z3 and here is where you can get z3 and so on.

leonardoalt:
Ok so do we just remove the script?

cameel (dismissed):

cameel:
Rebased on latest `develop`.

@leonardoalt I think you said you're going to remove the whole script. Are you still working on that?

cameel (approved):

PR #11810, opened 2021-08-17
Fix parsing fixed point values.

chriseth:


test/libsolidity/util/BytesUtils.cpp
```diff
@@ -94,12 +94,22 @@ bytes BytesUtils::convertNumber(string const& _literal)
...
+	o_fractionalDigits = dotPos < _literal.size() ? _literal.size() - dotPos : 0;
+	bool negative = !_literal.empty() && _literal.at(0) == '-';
+	// remove decimal point
+	string valueInteger = _literal.substr(0, dotPos) + _literal.substr(dotPos + 1);
+	// erase leading zeros to avoid parsing as octal.
+	while (!valueInteger.empty() && (valueInteger.at(0) == '0' || valueInteger.at(0) == '-'))
+		valueInteger.erase(valueInteger.begin());
+	if (valueInteger.empty())
+		valueInteger = "0";
+	cout << valueInteger << endl;
```
Marenz: Did you mean to keep the `cout` here?

test/libsolidity/util/BytesUtils.cpp
```diff
@@ -94,12 +94,22 @@ bytes BytesUtils::convertNumber(string const& _literal)
...
 	try
 	{
-		return util::toBigEndian(u256(valueInteger + valueFraction));
+		u256 value(valueInteger);
+		if (negative)
+			value = s2u(-u2s(value));
```
Marenz, line 110: Is `s2u()` signed to unsigned? And are we in this case converting a negative signed value to an unsigned one?

test/libsolidity/util/BytesUtils.cpp
hrkrshnn:
```suggestion
```

test/libsolidity/util/BytesUtils.cpp
hrkrshnn, line 110:
```suggestion
			value = -value;
```
?

chriseth: it converts between two's complement and "real" signed integers. Writing `value = -value` would be equivalent, but the compiler complains that it can result in undefined behaviour.

chriseth: actually I only tried `0-value`, let me see if this works.

chriseth: Nope, does not work.

Marenz (approved):

hrkrshnn (approved):

Issue #10269, opened 2020-11-12
Remove unneeded includes in files


axic:
It seems there is a very large number of includes which are imported without any reason. Most of these are remnants of old code.

Removing them could allow for nicer separation of concerns, but could also speed up compilation.

Is there a simple tool which could show includes whose declarations are not really used in a given file?

KishkinJ10:
i am a beginner how can i contribute on this??

hrkrshnn:
@KishkinJ10 

Probably the easiest way to proceed is to find a tool that would automatically do this. Googling lead me to https://stackoverflow.com/questions/1301850/tools-to-find-included-headers-which-are-unused. You could also do some research into it, since the stack overflow answer might be outdated.

So try to pick one such tool and then run it on the repo, then try compiling to see if it works.

Might be useful: https://docs.soliditylang.org/en/v0.8.3/installing-solidity.html#building-from-source

Feel free to ask questions if you are stuck. Please make sure you understand the problem before working on this. There is no pressure to actually do it, so please take your time.

TerranCivilian:
@chriseth I can work on this if it still needs to be done

TerranCivilian:
Letting the CLion inspector run now. I'm having it analyze header files as well, not sure if that matters.

cameel:
Do we consider it done after #11491 or do we want to have it automated?

TerranCivilian:
I can take a crack at automating it. @cameel do you have anything in mind for that?

alice-dfi:
Hello,
Please do not do anything to my github accounts; https://github.com/terra-project along with https://github.com/Alice_finance
These account have been hacked for months and are being ran by hackers who have been stealing my money everyday.  nothing, I mean nothing should get done without notifying me first. I’ve been contacting github support daily with no response. I recommend github contact me so we can solve this issue. If am currently working with U.S. Attorney Craig Carpentino along with FBI special agent John R. Tafur and my personal lawyer Yelena Sharova from Sharova Law Firm in NYC. Please feel free to contact them if you have any questions or concerns.
You can reach me at ***@***.***, ***@***.***, ***@***.***, please try all 3 accounts like I said, I’ve been hacked and they block most emails. Looking forward to hearing from you, hopefully we can resolve this matter and continue a great business relationship like we started  before I was hacked and my identity stolen.
Best,
Engjell Valteri

cameel:
[...]

PR #11812, opened 2021-08-17
Add missing header.

### LINKED ISSUE #11811: ‘list’ in namespace ‘std’ does not name a template type
## Description

Build failure due to missing include

## Environment

- Compiler version: g++ Debian 8.3.0-6, CMake 3.13.4
- Operating system: Debian

## Steps to Reproduce

This looks like the exact same thing as in https://github.com/ethereum/solidity/issues/11103 but I figured I would make a new issue since it's a different file. Feel free to close and just use that issue if that's better

At least on my system,

```
$ git clone --recursive https://github.com/ethereum/solidity.git
$ git checkout v0.8.7
$ mkdir build && cd build
$ cmake .. && make
```

I get the following output:
image

ekpyron:
Closes https://github.com/ethereum/solidity/issues/11811.

I wonder if our policy should be to directly include all STL headers required by a class, independently of them being included by other headers already. Otherwise, I see little chance of reviewing against these kinds of things and STL implementations will pull in whatever they like, so CI won't avoid this either.

### DIFF
--- a/libyul/optimiser/StackToMemoryMover.h
+++ b/libyul/optimiser/StackToMemoryMover.h
@@ -27,6 +27,8 @@
 #include <liblangutil/SourceLocation.h>
 #include <libsolutil/Common.h>
 
+#include <list>
+
 namespace solidity::yul
 {
 


cameel (approved):

cameel:
> I wonder if our policy should be to directly include all STL headers required by a class, independently of them being included by other headers already. Otherwise, I see little chance of reviewing against these kinds of things and STL implementations will pull in whatever they like, so CI won't avoid this either.

I thought that *was* our policy :) That's what I usually do. I include everything that contains symbols I explicitly refer to in the file. Though it's hard to keep these includes up to date because the compiler does not complain when a header is already included through another.

Maybe we should try https://include-what-you-use.org and add it to CI if it's reliable? It looked promising for superfluous includes (https://github.com/ethereum/solidity/issues/10269#issuecomment-869074661) but that wasn't very important. It could help with cases like this, it would be more worthwhile.

ekpyron:
> > I wonder if our policy should be to directly include all STL headers required by a class, independently of them being included by other headers already. Otherwise, I see little chance of reviewing against these kinds of things and STL implementations will pull in whatever they like, so CI won't avoid this either.
> 
> I thought that _was_ our policy :) That's what I usually do. I include everything that contains symbols I explicitly refer to in the file. Though it's hard to keep these includes up to date because the compiler does not complain when a header is already included through another.

At a quick glance, I hardly see it adhered to anywhere :-).
I mean, the file here would additionally need at least ``<map>``, ``<vector>`` and ``<optional>``. But moving to a random next one ``OptimiserStep.h``, it's e.g. missing ``<memory>`` and my guess is it's true for more files than for which it isn't that they're missing *something* relative to that policy.

> 
> Maybe we should try https://include-what-you-use.org and add it to CI if it's reliable? It looked promising for superfluous includes ([#10269 (comment)](https://github.com/ethereum/solidity/issues/10269#issuecomment-869074661)) but that wasn't very important. It could help with cases like this, it would be more worthwhile.

Does that really work for STL headers at all?

ekpyron:
If there is something that works reliably that we can use in CI, then I'm all for doing so, but I'd be surprised :-).

cameel:
Well, let's find out. I created an issue, very likely that someone will pick that up: #11813.
It's already mentioned in the issue about removing includes, but it's buried under a ton of comments so I think a new, more focused issue will bring more attention to it.

PR #11815, opened 2021-08-18
New Code Transform: Add optimizer setting for fuzzing.

ekpyron:
@bshastry This adds ``forceOldPipeline`` to ``OptimiserSettings``.

Note that this both affects code transform *and* the Yul optimizer (since the Yul optimizer will query the respective code transform about stack issues), so you will have to run the Yul optimizer each time separately.
Since it looks like you're running the optimizer via ``YulOptimizerTestCommon``, I also added another boolean flag ``_forceOldPipeline`` to its constructor.

The idea is to run the full optimizer and code transform in both cases with ``optimizeStackOptimizations == true``, but one time with ``forceOldPipeline = false`` and one time with ``forceOldPipeline = true``, the former using the new code transform, the latter forcing the use of the old code transform.


ekpyron:
Closing this now. If we want to fuzz it again like this, we can recreate something similar.

PR #11817, opened 2021-08-18
Extend knowledge base

chriseth:


test/libyul/KnowledgeBaseTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+		let a := calldataload(0)
+		let b := calldataload(0)
+		let zero := 0
+		let c := add(b, 0)
+		let d := mul(b, 0)
+		let e := sub(a, b)
+	})");
+
+	BOOST_CHECK(!kb.knownToBeDifferent("a"_yulstring, "b"_yulstring));
+	BOOST_CHECK(kb.valueIfKnownConstant("zero"_yulstring) == u256(0));
```
chriseth, line 89: This fails and I have no idea why.

ekpyron: Even worse it fails only on osx or do I see that incorrectly?

ekpyron (commented):
Looks good per se, but the osx test failure is scary.

test/libyul/KnowledgeBaseTest.cpp
chriseth: My assumption is that it has to do with YulStringRepository.

chriseth:
Ah, the reference to the value of the expression `zero` is a lingering reference, I think...

chriseth:
Sorry, I got on a tangent here, but now the knowledge base and maybe also the optimizer is much more powerful.

ekpyron (commented):
Looks good and makes a lot of sense, but before fully approving, I'd want to check for global interactions with existing rules (tomorrow).

libevmasm/RuleList.h
```diff
@@ -681,6 +681,22 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart8(
 			// X - (A + Y) -> (X - Y) + (-A)
 			Builtins::SUB(X, Builtins::ADD(A, Y)),
 			[=]() -> Pattern { return Builtins::ADD(Builtins::SUB(X, Y), 0 - A.d()); }
+		}, {
+			// (X - A) - Y -> (X - Y) - A
+			Builtins::SUB(Builtins::SUB(X, A), Y),
+			[=]() -> Pattern { return Builtins::SUB(Builtins::SUB(X, Y), A); }
```
ekpyron, line 687: Will this rule ever be matched rather than first hitting ``X - A -> X + (-A)``? I'd need to look through the matching mechanism to be able to tell for sure, but maybe a few test cases for these would be nice in general.
We should probably in general have `isoltest`-tests for single and multiple applications of the rule list, but may be overkill for this...

ekpyron: Without dedicated rule list tests, probably hard to test, since
``X - A`` -> ``X + (-A)``
followed by
``(X + (-A)) - Y -> (X - Y) + (-A)``
will be indistinguishable from

``(X - A) - Y) -> (X - Y) - A``
followed by
``(X - Y) - A -> (X - Y) + (-A)``

But in that sense this rule is probably redundant, isn't it?

libevmasm/RuleList.h
```diff
@@ -681,6 +681,22 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart8(
...
+		}, {
+			// (A - X) - Y -> A - (X + Y)
+			Builtins::SUB(Builtins::SUB(A, X), Y),
+			[=]() -> Pattern { return Builtins::SUB(A, Builtins::ADD(X, Y)); }
+		}, {
+			// X - (Y - A) -> (X - Y) + A
```
ekpyron, line 693: Similar as above:
Shouldn't ``Y - A`` be matched to ``Y + (-A)`` first, and then
``X - (Y + (-A))`` matches ``X - (Y + B)`` and should reduce to ``(X - Y) + (--A)`` which should also yield ``(X - Y) + A``?

test/libyul/KnowledgeBaseTest.cpp
```diff
@@ -0,0 +1,134 @@
...
+//		let b := calldataload(0)
+//		let zero := 0
+//		let c := add(b, 0)
+//		let d := mul(b, 0)
+//		let e := sub(a, b)
+//	})");
+
+//	BOOST_CHECK(!kb.knownToBeDifferent("a"_yulstring, "b"_yulstring));
+//	// This only works if the variable names are the same.
+//	// It assumes that SSA+CSE+Simplifier actually replaces the variables.
```
hrkrshnn: I think this can be improved by slightly modifying the pattern matching engine: https://github.com/ethereum/solidity/commit/8a94f173adf4ca154ef91b8bf7bad4b004efdc01

For example, for 
```
let a := calldataload(0)
let b := calldataload(0)
let c := b
let d := c
let e := a
let f := e
```
In this case, it can reason equality of `f` and `d`, without needing to run CSE first.

chriseth: The thing is that I actually don't want the pattern matching engine to be more complicated.

libevmasm/RuleList.h
chriseth: Yes, some rules are redundant, but I think I need them anyway because of the way the KnowledgeBase works.

chriseth: Actually no, they are really redundant.

chriseth:
Ok, it turns out the rules are indeed required.

ekpyron:
> Ok, it turns out the rules are indeed required.

Why?

chriseth:
> > Ok, it turns out the rules are indeed required.
> 
> Why?

I commented them out and the tests failed.

libyul/optimiser/KnowledgeBase.cpp
```diff
@@ -51,39 +50,58 @@ bool KnowledgeBase::knownToBeDifferent(YulString _a, YulString _b)
...
+{
+	if (m_variableValues.count(_a))
+		if (Literal const* literal = get_if<Literal>(m_variableValues.at(_a).value))
+			return valueOfLiteral(*literal);
+	return {};
+}
+
 Expression KnowledgeBase::simplify(Expression _expression)
 {
-	bool startedRecursion = (m_recursionCounter == 0);
```
chriseth, line 71: Actually this change is quite dramatic: It is now checking the depth of the search tree, not the size. I think I'll revert this again (or fix it in a different way).

libyul/optimiser/KnowledgeBase.cpp
```diff
@@ -51,39 +50,58 @@ bool KnowledgeBase::knownToBeDifferent(YulString _a, YulString _b)
...
 		return _expression;
-	else
-		--m_recursionCounter;
 
 	if (holds_alternative<FunctionCall>(_expression))
 		for (Expression& arg: std::get<FunctionCall>(_expression).arguments)
-			arg = simplify(arg);
+			arg = simplifyRecursively(arg, _recursionCounter + 1);
 
 	if (auto match = SimplificationRules::findFirstMatch(_expression, m_dialect, m_variableValues))
```
ekpyron, line 104: Maybe the reason for the knowledge base needing the additional rule is that this needs to be a ``while``?

chriseth: there is a recursive call: https://github.com/ethereum/solidity/pull/11817/files#diff-689e75ac518209a32f0ffa72f01340bebfa32b36ac25f5ff1cc17b99ed876b0eR104

ekpyron (dismissed):
Looks good if tests pass.

hrkrshnn (approved):

test/libsolidity/semanticTests/array/copying/copy_byte_array_in_struct_to_storage.sol
```diff
@@ -37,11 +37,11 @@ contract C {
 // compileViaYul: also
 // ----
 // f() -> 0x40, 0x80, 6, 0x6162636465660000000000000000000000000000000000000000000000000000, 0x49, 0x3132333435363738393031323334353637383930313233343536373839303120, 0x3132333435363738393031323334353637383930313233343536373839303120, 0x3132333435363738390000000000000000000000000000000000000000000000
-// gas irOptimized: 179963
```
chriseth, line 40: These gas savings come from the fact that the load resolver can now reason that two memory operations are unrelated:
```
let zero_struct_S_mpos := allocate_and_zero_memory_struct_struct_S()
mstore(zero_struct_S_mpos,  0x07)
let _1 := add(zero_struct_S_mpos, 96)
mstore(_1, copy_literal_to_memory_e3fc88296de5f3bc29e06c9130a9b57250926f606654c95ecdac75bb501a42ff())
let _2 := 32
let _3 := add(zero_struct_S_mpos, _2)
let memPtr := allocate_memory_array_string_1990()
mstore(add(memPtr,  _2),  "abcdef")
mstore(_3, memPtr)
let _4 := add(zero_struct_S_mpos, 64)
mstore(_4,  0x09)
update_storage_value_offsett_uint16_to_uint16_1991(and(mload(zero_struct_S_mpos), 0xffff))
let _5 := mload(_3) /* XXX */
let newLen := mload(_5)
```

This is the old code. In the new code, the load marked by `XXX` is replaced by `memPtr`.

ekpyron (approved):

PR #11818, opened 2021-08-18
Fix incorrect Bash highlighting in a Yul code snippet

cameel:
Just a tiny formatting problem I noticed while reading through the docs.


franzihei (approved):
makes sense!

PR #11823, opened 2021-08-19
Add UBSan build + test nightly CI.

### LINKED ISSUE #11822: [UBSan] CI run
While running the undefined behavior sanitizer (UBSan) fuzzer, I noticed that several isoltests fail with a null pointer dereference. I don't know why this does not happen outside of the UBSan build.

Wonder if we should have another couple of CIs: (1) UBSan build; (2) UBSan tests

Example error

```
semanticTests/array/function_array_cross_calls.sol: /src/solidity/libyul/ScopeFiller.cpp:56:55: runtime error: member access within null pointer of type 'std::shared_ptr<const solidity::yul::DebugData>::element_type' (aka 'const solidity::yul::DebugData')
```

bshastry:
Fixes #11822 

- Adds 3 nightly runs (UBSan build, UBSan soltest, UBSan CLI test)

Since the UBSan run will be on a nightly basis, we will only start seeing failures such as those in https://github.com/ethereum/solidity/issues/11822 after this PR is merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,18 +121,7 @@ defaults:
       no_output_timeout: 30m
       command: ./.circleci/soltest_all.sh
 
-  - run_cmdline_tests: &run_cmdline_tests
-      name: command line tests
-      no_output_timeout: 30m
-      command: ./test/cmdlineTests.sh
-
-  - run_docs_pragma_min_version: &run_docs_pragma_min_version
-      name: docs pragma version check
-      command: ./scripts/docs_version_pragma_check.sh
-
-  - test_ubuntu1604_clang: &test_ubuntu1604_clang
-      docker:
-        - image: << pipeline.parameters.ubuntu-1604-clang-ossfuzz-docker-image >>
+  - run_soltest_steps: &run_soltest_steps
       steps:
         - checkout
         - attach_workspace:
@@ -141,21 +130,7 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
-  - test_ubuntu2004_clang: &test_ubuntu2004_clang
-      docker:
-        - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
-      steps:
-        - checkout
-        - attach_workspace:
-            at: build
-        - run: *run_soltest
-        - store_test_results: *store_test_results
-        - store_artifacts: *artifacts_test_results
-
-  - test_ubuntu2004: &test_ubuntu2004
-      docker:
-        - image: << pipeline.parameters.ubuntu-2004-docker-image >>
-      parallelism: 6
+  - run_soltest_all_steps: &run_soltest_all_steps
       steps:
         - checkout
         - attach_workspace:
@@ -164,27 +139,48 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
+  - run_cmdline_tests: &run_cmdline_tests
+      name: command line tests
+      no_output_timeout: 30m
+      command: ./test/cmdlineTests.sh
+
+  - run_cmdline_tests_steps: &run_cmdline_tests_steps
+      steps:
+        - checkout
+        - attach_workspace:
+            at: build
+        - run: *run_cmdline_tests
+        - store_test_results: *store_test_results
+        - store_artifacts: *artifacts_test_results
+
+  - run_docs_pragma_min_version: &run_docs_pragma_min_version
+      name: docs pragma version check
+      command: ./scripts/docs_version_pragma_check.sh
+
+  - test_ubuntu1604_clang: &test_ubuntu1604_clang
+      docker:
+        - image: << pipeline.parameters.ubuntu-1604-clang-ossfuzz-docker-image >>
+      <<: *run_soltest_steps
+
+  - test_ubuntu2004_clang: &test_ubuntu2004_clang
+      docker:
+        - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
+      <<: *run_soltest_steps
+
+  - test_ubuntu2004: &test_ubuntu2004
+      docker:
+        - image: << pipeline.parameters.ubuntu-2004-docker-image >>
+      parallelism: 6
+      <<: *run_soltest_all_steps
+
   - test_asan: &test_asan
       <<: *test_ubuntu2004
-      steps:
-      - checkout
-      - attach_workspace:
-          at: build
-      - run:
-          <<: *run_soltest
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
+      <<: *run_soltest_steps
 
-  - test_asan_clang: &test_asan_clang
-      <<: *test_ubuntu2004_clang
-      steps:
-      - checkout
-      - attach_workspace:
-          at: build
-      - run:
-          <<: *run_soltest
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
+  - test_ubuntu2004_clang_cli: &test_ubuntu2004_clang_cli
+      docker:
+        - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
+      <<: *run_cmdline_tests_steps
 
   # --------------------------------------------------------------------------
   # Workflow Templates
@@ -238,6 +234,11 @@ defaults:
       <<: *workflow_trigger_on_tags
       requires:
         - b_ubu_asan_clang
+
+  - workflow_ubuntu2004_ubsan_clang: &workflow_ubuntu2004_ubsan_clang
+      <<: *workflow_trigger_on_tags
+      requires:
+        - b_ubu_ubsan_clang
 
   - workflow_emscripten: &workflow_emscripten
       <<: *workflow_trigger_on_tags
@@ -447,14 +448,27 @@ jobs:
       - store_artifacts: *artifacts_solc
       - persist_to_workspace: *artifacts_executables
 
-
-  b_ubu_asan_clang: &build_ubuntu2004_clang
+  b_ubu_asan_clang: &b_ubu_asan_clang
     docker:
       - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
     environment:
       CC: clang
       CXX: clang++
       CMAKE_OPTIONS: -DSANITIZE=address
+      MAKEFLAGS: -j 3
+    steps:
+      - checkout
+      - run: *run_build
+      - store_artifacts: *artifacts_solc
+      - persist_to_workspace: *artifacts_executables
+
+  b_ubu_ubsan_clang: &b_ubu_ubsan_clang
+    docker:
+      - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
+    environment:
+      CC: clang
+      CXX: clang++
+      CMAKE_OPTIONS: -DSANITIZE=undefined
       MAKEFLAGS: -j 3
     steps:
       - checkout
@@ -768,13 +782,7 @@ jobs:
       - image: << pipeline.parameters.ubuntu-2004-docker-image >>
     environment:
       TERM: xterm
-    steps:
-      - checkout
-      - attach_workspace:
-          at: build
-      - run: *run_cmdline_tests
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
+    <<: *run_cmdline_tests_steps
 
   t_ubu_release_cli: &t_ubu_release_cli
     <<: *t_ubu_cli
@@ -784,14 +792,7 @@ jobs:
     environment:
       TERM: xterm
       ASAN_OPTIONS: check_initialization_order=true:detect_stack_use_after_return=true:strict_init_order=true:strict_string_checks=true:detect_invalid_pointer_pairs=2
-    steps:
-      - checkout
-      - attach_workspace:
-          at: build
-      - run:
-          <<: *run_cmdline_tests
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
+    <<: *run_cmdline_tests_steps
 
   t_ubu_asan_constantinople:
     <<: *test_asan
@@ -802,12 +803,18 @@ jobs:
       ASAN_OPTIONS: check_initialization_order=true:detect_stack_use_after_return=true:strict_init_order=true:strict_string_checks=true:detect_invalid_pointer_pairs=2
 
   t_ubu_asan_constantinople_clang:
-    <<: *test_asan_clang
+    <<: *test_ubuntu2004_clang
     environment:
       EVM: constantinople
       OPTIMIZE: 0
       SOLTEST_FLAGS: --no-smt
       ASAN_OPTIONS: check_initialization_order=true:detect_stack_use_after_return=true:strict_init_order=true:strict_string_checks=true:detect_invalid_pointer_pairs=2
+
+  t_ubu_ubsan_clang:
+    <<: *test_ubuntu2004_clang
+
+  t_ubu_ubsan_clang_cli:
+    <<: *test_ubuntu2004_clang_cli
 
   t_ems_solcjs:
     docker:
@@ -1207,6 +1214,11 @@ workflows:
       - t_ubu_asan_constantinople: *workflow_ubuntu2004_asan
       - t_ubu_asan_constantinople_clang: *workflow_ubuntu2004_asan_clang
       - t_ubu_asan_cli: *workflow_ubuntu2004_asan
+
+      # UBSan build and tests
+      - b_ubu_ubsan_clang: *workflow_trigger_on_tags
+      - t_ubu_ubsan_clang: *workflow_ubuntu2004_ubsan_clang
+      - t_ubu_ubsan_clang_cli: *workflow_ubuntu2004_ubsan_clang
 
       # Emscripten build and tests that take more than 15 minutes to execute
       - b_ems: *workflow_trigger_on_tags
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -434,7 +434,7 @@ jobs:
           name: Python unit tests
           command: python.exe test/pyscriptTests.py
 
-  b_ubu_clang: &build_ubuntu2004_clang
+  b_ubu_clang: &b_ubu_clang
     resource_class: xlarge
     docker:
       - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
@@ -473,10 +473,11 @@ jobs:
     steps:
       - checkout
       - run: *run_build
+      - run: *gitter_notify_failure
       - store_artifacts: *artifacts_solc
       - persist_to_workspace: *artifacts_executables
 
-  b_ubu: &build_ubuntu2004
+  b_ubu: &b_ubu
     resource_class: xlarge
     docker:
       - image: << pipeline.parameters.ubuntu-2004-docker-image >>
@@ -489,14 +490,14 @@ jobs:
       - store_artifacts: *artifacts_tools
       - persist_to_workspace: *artifacts_executables
 
-  b_ubu_release: &build_ubuntu2004_release
-    <<: *build_ubuntu2004
+  b_ubu_release: &b_ubu_release
+    <<: *b_ubu
     environment:
       FORCE_RELEASE: ON
       MAKEFLAGS: -j 10
 
   b_ubu_static:
-    <<: *build_ubuntu2004
+    <<: *b_ubu
     environment:
       MAKEFLAGS: -j 10
       CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
@@ -509,7 +510,7 @@ jobs:
       - store_artifacts: *artifacts_solc
 
   b_ubu_codecov:
-    <<: *build_ubuntu2004
+    <<: *b_ubu
     environment:
       COVERAGE: ON
       CMAKE_BUILD_TYPE: Debug
@@ -543,7 +544,7 @@ jobs:
   # Builds in C++20 mode and uses debug build in order to speed up.
   # Do *NOT* store any artifacts or workspace as we don't run tests on this build.
   b_ubu_cxx20:
-    <<: *build_ubuntu2004
+    <<: *b_ubu
     environment:
       CMAKE_BUILD_TYPE: Debug
       CMAKE_OPTIONS: -DCMAKE_CXX_STANDARD=20 -DUSE_CVC4=OFF
@@ -552,7 +553,7 @@ jobs:
       - checkout
       - run: *run_build
 
-  b_ubu_ossfuzz: &build_ubuntu1604_clang
+  b_ubu_ossfuzz: &b_ubu_ossfuzz
     docker:
       - image: << pipeline.parameters.ubuntu-1604-clang-ossfuzz-docker-image >>
     environment:
@@ -697,7 +698,7 @@ jobs:
 
   # x64 ASAN build, for testing for memory related bugs
   b_ubu_asan: &b_ubu_asan
-    <<: *build_ubuntu2004
+    <<: *b_ubu
     environment:
       CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 10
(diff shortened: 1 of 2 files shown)


cmake/EthCompilerSettings.cmake
```diff
@@ -197,6 +197,8 @@ if (SANITIZE)
 	# -fsanitize-address-use-after-scope throws an error when a variable is used beyond its scope
 	if (san STREQUAL "address")
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer -fsanitize=address -fsanitize-address-use-after-scope")
+	elseif (san STREQUAL "undefined")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr")
```
cameel: You specify `unsigned-integer-overflow` for `-fno-sanitize-recover` but not for `-fsanitize`. Is it a mistake and you just forgot to add it for `-fsanitize` or is it on purpose? If on purpose, this definitely needs a comment.

cameel: This line is very long and you also have two lists that seem identical at a glance but are not. I think this would be a much better way to format it:
```suggestion
		string(JOIN "," _sanitizers
			array-bounds
			bool
			builtin
			enum
			float-divide-by-zero
			function
			integer-divide-by-zero
			null
			object-size
			return
			returns-nonnull-attribute
			shift
			signed-integer-overflow
			unreachable
			vla-bound
			vptr
		)
		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=${_sanitizers} -fno-sanitize-recover=${_sanitizers},unsigned-integer-overflow")
```
I did not test it in practice so it may need some tweaks (not sure if the variable in JOIN needs to be declared first for example) but something long these lines should work.

cameel: From [Instrumentation Options](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) in GCC docs:
> `-fsanitize=float-divide-by-zero`
>
> Detect floating-point division by zero. Unlike other similar options, `-fsanitize=float-divide-by-zero` is not enabled by `-fsanitize=undefined`, since floating-point division by zero can be a legitimate way of obtaining infinities and NaNs.

So technically it could cause false-positives. I hope we're not relying on that behavior though - if that's the case, it's fine.

cameel: I did not delve too deep into Clang's [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) docs but I see some unused sanitizers that look like it would not hurt to include them: `alignment`, `pointer-overflow`, `unsigned-shift-base`. Any reason we do not want them?

There's also `nonnull-attribute` and `float-cast-overflow` but I think the former is only relevant for some non-standard "non-null" attribute which we are not using and the latter is documented in GCC docs as not playing well with other sanitizers so I guess these were excluded on purpose?

.circleci/config.yml
```diff
@@ -1208,6 +1241,10 @@ workflows:
       - t_ubu_asan_constantinople_clang: *workflow_ubuntu2004_asan_clang
       - t_ubu_asan_cli: *workflow_ubuntu2004_asan
 
+      # UBSan build and tests
+      - b_ubu_ubsan_clang: *workflow_trigger_on_tags
+      - t_ubu_ubsan_clang: *workflow_ubuntu2004_ubsan_clang
```
cameel, line 1239: Maybe a CLI run would be useful for UBSan too?

.circleci/config.yml
```diff
@@ -186,6 +186,17 @@ defaults:
...
+      <<: *test_ubuntu2004_clang
+      steps:
+      - checkout
+      - attach_workspace:
+          at: build
+      - run:
+          <<: *run_soltest
+      - store_test_results: *store_test_results
+      - store_artifacts: *artifacts_test_results
+
```
cameel: `test_ubsan_clang` looks identical to `test_asan_clang`. The sanitizer is selected at build time so that's not surprising. Maybe it would be better to just rename `test_asan_clang` to something like `test_san_clang` and use it for both? Then also rename `test_asan` to `test_san`.

.circleci/config.yml
```diff
@@ -462,6 +478,20 @@ jobs:
...
+    environment:
+      CC: clang
+      CXX: clang++
+      CMAKE_OPTIONS: -DSANITIZE=undefined
+      MAKEFLAGS: -j 3
+    steps:
+      - checkout
+      - run: *run_build
+      - store_artifacts: *artifacts_solc
+      - persist_to_workspace: *artifacts_executables
```
cameel: You could parametrize it and merge it with `b_ubu_asan_clang`:

```yaml
  b_ubu_san_clang: &build_ubuntu2004_clang
    parameters:
      sanitizer:
        type: string
    docker:
    environment:
      CC: clang
      CXX: clang++
      CMAKE_OPTIONS: -DSANITIZE=<< parameters.sanitizer >>
      MAKEFLAGS: -j 3
    steps:
      - checkout
      - run: *run_build
      - store_artifacts: *artifacts_solc
      - persist_to_workspace: *artifacts_executables
```

Then, in `jobs`:
```yaml
      # ASan build and tests
      - b_ubu_asan: *workflow_trigger_on_tags
      - b_ubu_san_clang:
          <<: *workflow_trigger_on_tags
          sanitizer: address
      - t_ubu_asan_constantinople: *workflow_ubuntu2004_asan
      - t_ubu_asan_constantinople_clang: *workflow_ubuntu2004_asan_clang
      - t_ubu_asan_cli: *workflow_ubuntu2004_asan

      # UBSan build and tests
      - b_ubu_san_clang:
          <<: *workflow_trigger_on_tags
          sanitizer: undefined
      - t_ubu_ubsan_clang: *workflow_ubuntu2004_ubsan_clang
```

Or, alternatively, to avoid making `jobs` more verbose, you could still define `b_ubu_asan_clang` and `b_ubu_ubsan_clang` as standalone dicts with defaults (as they are now) but using this parameterized definition so that you only have to specify the parameter there:
```yaml
  b_ubu_asan_clang:
    <<: &b_ubu_san_clang
    sanitizer: address

  b_ubu_ubsan_clang:
    <<: &b_ubu_san_clang
    sanitizer: undefined
```

.circleci/config.yml
```diff
@@ -186,6 +186,17 @@ defaults:
...
+  - test_ubsan_clang: &test_ubsan_clang
+      <<: *test_ubuntu2004_clang
+      steps:
+      - checkout
+      - attach_workspace:
+          at: build
+      - run:
+          <<: *run_soltest
+      - store_test_results: *store_test_results
+      - store_artifacts: *artifacts_test_results
```
cameel: Looking at the definition of `test_ubuntu2004_clang`, the `steps` here are not needed. They're already defined in `test_ubuntu2004_clang` and this only overwrites them with the same content. The whole dict could be reduced to just this:

```suggestion
  - test_ubsan_clang: &test_ubsan_clang
      <<: *test_ubuntu2004_clang
```

It's different with `test_asan` above because it's based on `test_ubuntu2004`, which differs slightly - it has `run: *run_soltest_all` instead of `run: *run_soltest`. I wonder if it's intentional though because others all use `run_soltest`.

cameel: This sequence of steps also looks like something that's repeated often and could be easily extracted into a separate dict. We already have dicts named `run_soltest` and `run_soltest_all` (that represent just the single step) but this could be named `run_soltest_steps` and `run_soltest_all_steps`.

cmake/EthCompilerSettings.cmake
bshastry: > You specify unsigned-integer-overflow for -fno-sanitize-recover but not for -fsanitize

I think it is the other way round, right?

>  If on purpose, this definitely needs a comment.

Yup, agree. Will do.

cameel: > I think it is the other way round, right?

Ah, true. Sorry. I must have looked at modified version that I've been in the process of editing locally to create the suggestion below :)

Makes sense then. But is that really the only one we want recovery for? Are the others too spammy if you let them continue?

bshastry: Good observation. I don't think we are relying on this since the fuzzer would have alerted me otherwise, and I have not received such a bug report so far.

Actually, the motivation for this PR are null dereferences on shared_ptr that were reported by the fuzzer on a very regular basis.

bshastry: > Any reason we do not want them?

No reason. Actually, the set of flags used in this PR are an identical copy of the flags used by Google's fuzzing engine.

.circleci/config.yml
bshastry: Ah nice, thanks for the tip!

.circleci/config.yml
bshastry: I have addressed the first comment. I don't fully understand the second comment, so leaving it unaddressed until further clarification.

.circleci/config.yml
bshastry: I have done the `test_san_clang` refactoring but not `test_asan` because the UBSan run is only executed on the clang build image.

.circleci/config.yml
bshastry: CLI run would also benefit. I was a bit cautious because that would introduce another CI and lately we have been debating cost-benefit tradeoff and I wasn't sure where this would tip should CLI have a dedicated UBSan run.

bshastry:
There seems to be a circle ci config failure after I performed refactoring. I will post an update once I have understood how to pass a parameter correctly to a parameterized build job.

cmake/EthCompilerSettings.cmake
cameel: Ok, sounds reasonable. If it does not find any false-positives now then I guess we put off worrying about it until we actually run into them. We can disable this particular sanitizer then.

.circleci/config.yml
cameel: I think it's not a big deal for a nightly run. I'd only worry about that if it was a PR check because I suspect most of our usage comes from them.

cmake/EthCompilerSettings.cmake
cameel: Well, they proably have some reasons for not including them.

In that case I'd add them but with a comment that the fuzzing engine it's based on is not using them and to remove them if they cause us trouble.

.circleci/config.yml
cameel: I really meant just renaming `test_asan` to match the naming here. This way the next person that needs to change this will see that it can be used with sanitizers in general and not just the address sanitizer.

.circleci/config.yml
cameel: I meant that this:
```yaml
      steps:
      - checkout
      - attach_workspace:
          at: build
      - run:
          <<: *run_soltest
      - store_test_results: *store_test_results
```

is used in several places, with just a slight variation in the `run` part. I.e. sometimes it's `run: *run_soltest` (which means the same thing), and sometimes it's modified to run `run_soltest_all` instead.

You could add an entry for this under the `defaults:` as we already do with some other reusable pieces of config:
```yaml
  - run_soltest_steps: &run_soltest_steps
      steps:
      - checkout
      - attach_workspace:
          at: build
      - run: *run_soltest
      - store_test_results: *store_test_results
```

And then some other dicts could be simplified. For example `test_ubuntu2004_clang` could be shortened:
```yaml
  - test_ubuntu2004_clang: &test_ubuntu2004_clang
      docker:
      <<: *run_soltest_steps
```
Same with `run_soltest_all_steps`.

I'd be inclined to call them just `run_soltest`/`run_soltest_all` but we already have dicts that use these names (in fact that's exactly what `run: *run_soltest` is referring to):
```yaml
  - run_soltest: &run_soltest
      name: soltest
      no_output_timeout: 30m
      command: ./.circleci/soltest.sh

  - run_soltest_all: &run_soltest_all
      name: soltest_all
      no_output_timeout: 30m
      command: ./.circleci/soltest_all.sh
```

cameel:
> There seems to be a circle ci config failure after I performed refactoring. I will post an update once I have understood how to pass a parameter correctly to a parameterized build job.

I did not test them unfortunately (can't really without actually running CI) so I might have made a mistake somewhere but that's based on what I did for `t_ems_ext` and the general concept should work with some minimal adjustments.

cmake/EthCompilerSettings.cmake
```diff
@@ -192,11 +192,37 @@ endif ()
...
 	# Perform case-insensitive string compare
-	string(TOLOWER "${SANITIZE}" san)
+	string(TOLOWER "${SANITIZE}" sanitizer)
 	# -fno-omit-frame-pointer gives more informative stack trace in case of an error
 	# -fsanitize-address-use-after-scope throws an error when a variable is used beyond its scope
-	if (san STREQUAL "address")
+	if (sanitizer STREQUAL "address")
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer -fsanitize=address -fsanitize-address-use-after-scope")
+	elseif (sanitizer STREQUAL "undefined")
+		list(APPEND undefinedSanitizerFlags
```
cameel: Just a tiny nitpick: "flags" makes me think of a set of parameters, like the whole `-fsanitize=... -fsanitize-address-use-after-scope` part.

[UndefinedBehaviorSanitizer in clang docs](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) refers to these sanitizer options as "checks" instead.
```suggestion
		list(APPEND undefinedSanitizerChecks
```

bshastry:
> I did not test them unfortunately (can't really without actually running CI) so I might have made a mistake somewhere but that's based on what I did for `t_ems_ext` and the general concept should work with some minimal adjustments.

Sadly, I couldn't figure out a way to parameterize the sanitizer build job while using it as a dependent job for the sanitizer test job. The `t_ems_ext` job is slightly different because it is a standalone job.

.circleci/config.yml
bshastry: Got it, I will add a UBSAN CLI nightly job in that case.

cameel:
Too bad. At last the step refactor makes the config a bit shorter so there's still some reuse :)

leonardoalt:
What's the state here?

bshastry:
> What's the state here?

I think I have incorporated most review comments by @cameel , waiting for approval/re-review.

.circleci/config.yml
```diff
@@ -578,10 +583,9 @@ jobs:
             git clone https://github.com/ethereum/solidity-fuzzing-corpus /tmp/solidity-fuzzing-corpus
             mkdir -p test_results
             scripts/regressions.py -o test_results
-      - run: *gitter_notify_failure
-      - run: *gitter_notify_success
       - store_test_results: *store_test_results
       - store_artifacts: *artifacts_test_results
+    <<: *run_gitter_notify_steps
```
cameel: Does it actually work? AFAIK nested lists do not get merged. I think this will work like in Python, where if you have two dicts that look like this:
- `t_ubu_ossfuz` = `{"steps": ["checkout", {"attach_workspace": ...}, {"run": ...}, ...]}`
- `run_gitter_notify_steps` = `{"steps": [{"run": ...}, {"run": ...}]}`

and add them, `steps` list from one of them will just overwrite `steps` in the other. If `run_gitter_notify_steps` had other keys, they would be added but matching keys just replace each other.

.circleci/config.yml
```diff
@@ -784,14 +796,7 @@ jobs:
...
-      - attach_workspace:
-          at: build
-      - run:
-          <<: *run_cmdline_tests
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
+    <<: *run_cmdline_tests_steps
 
   t_ubu_asan_constantinople:
     <<: *test_asan
```
cameel, line 799: I see you replaced `test_asan_clang` with just `test_ubuntu2004_clang`.
You could the same way replace `test_asan` with `test_ubuntu2004`.

.circleci/config.yml
```diff
@@ -784,14 +792,7 @@ jobs:
     environment:
       TERM: xterm
       ASAN_OPTIONS: check_initialization_order=true:detect_stack_use_after_return=true:strict_init_order=true:strict_string_checks=true:detect_invalid_pointer_pairs=2
```
cameel, line 795: Maybe it would be better to move these options into the `CMakeFile.txt`? We now have the undefined sanitizer options there and I did not even realize that we had that config spread over multiple places.

Also I wonder if the `TERM` variable is actually needed here. You don't have it in `t_ubu_ubsan_clang_cli`. Maybe it should be added there too?

`t_ubu_asan_cli` and `t_ubu_ubsan_clang_cli` only differ with these two variables (and the notification).

cmake/EthCompilerSettings.cmake
```diff
@@ -192,11 +192,44 @@ endif ()
...
+			vla-bound
+			vptr
+		)
+		list(JOIN undefinedSanitizerChecks "," sanitizerChecks)
+		list(REMOVE_ITEM undefinedSanitizerChecks unsigned-integer-overflow)
+		# The fuzzer excludes reports of unsigned-integer-overflow. Hence, we remove it
+		# from the -fno-sanitize-recover checks. Consider reducing this list if we do not
+		# want to be notified about other failed checks.
+		list(JOIN undefinedSanitizerChecks "," dontRecoverFromChecks)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=${sanitizerChecks} -fno-sanitize-recover=${dontRecoverFromChecks}")
```
cameel, line 232: I just realized that we do not have `-fsanitize=undefined` here while the address sanitizer has an explicit `-fsanitize=address` option passed to it above. Does the `SANITIZE` variable on its own already enable the sanitizer (and `-fsanitize=address` is therefore superfluous) or is it a bug?

.circleci/config.yml
```diff
@@ -802,13 +807,21 @@ jobs:
...
       SOLTEST_FLAGS: --no-smt
       ASAN_OPTIONS: check_initialization_order=true:detect_stack_use_after_return=true:strict_init_order=true:strict_string_checks=true:detect_invalid_pointer_pairs=2
 
+  t_ubu_ubsan_clang:
+    <<: *test_ubuntu2004_clang
+    <<: *run_gitter_notify_steps
+
+  t_ubu_ubsan_clang_cli:
+    <<: *test_ubuntu2004_clang_cli
+    <<: *run_gitter_notify_steps
```
cameel: Here, again, I think that `run_gitter_notify_steps` will overwrite steps present in `test_ubuntu2004_clang`/`test_ubuntu2004_clang_cli`.

.circleci/config.yml
bshastry: > Maybe it would be better to move these options into the CMakeFile.txt?

These are contents of a runtime environment variable (see https://clang.llvm.org/docs/AddressSanitizer.html) and hence cannot be moved to CMakeFile, since the latter defines compile-time options only.

> Also I wonder if the TERM variable is actually needed here. You don't have it in t_ubu_ubsan_clang_cli. Maybe it should be added there too?

I actually don't know why TERM is required but I'm guessing it was added for a reason. Since we won't be seeing a UBSan run until this PR is merged, may be I will see the need for a term :-)

In any case, this can be addressed in a follow-up PR imo.

.circleci/config.yml
bshastry: > Does it actually work? AFAIK nested lists do not get merged

Good point. I assumed it would be merged, but I am not sure. I will revert this change to be safe.

cmake/EthCompilerSettings.cmake
bshastry: Good observation. I just double checked to make sure: 

See 
  - https://github.com/google/oss-fuzz/blob/878c32419cdf89457d1f1be616d4896fc2586160/infra/base-images/base-builder/Dockerfile#L131-L134
  - https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

```
-fsanitize=undefined: All of the checks listed above other than float-divide-by-zero, unsigned-integer-overflow, implicit-conversion, local-bounds and the nullability-* group of checks.
```

So `-fsanitize=undefined` is a shorthand for a pre-defined default set of checks. In this PR, we are essentially adding more checks that what `-fsanitize=undefined` would execute

.circleci/config.yml
bshastry: Actually, I'm confused. The following config seems to work (multiple `-run`s at the same nesting level). So this change is doing the same except via an alias, right?  

See: https://github.com/ethereum/solidity/blob/6e6bbb2f8383f10b913ad1c5fd15ee1d065f3b4c/.circleci/config.yml#L517-L526

.circleci/config.yml
bshastry: > I see you replaced test_asan_clang with just test_ubuntu2004_clang.

This is because they are identical (i.e., build and run `soltest_steps`)

> You could the same way replace test_asan with test_ubuntu2004.

But that would mean, we run `soltest_all_steps` instead of `soltest_steps` right?

.circleci/config.yml
cameel: Oh, I thought they were used at compile time. Fine then, they have to stay here.

And yeah, we can deal with `TERM` in a separate PR. Actually might be a `good first issue`.

cmake/EthCompilerSettings.cmake
cameel: Ok. Good.

.circleci/config.yml
cameel: It's a list of dicts, each with a single key called `"run"` and another, nested dict under that key in each dict. And there's no merging here. All list items are declared explicitly. Even `- run: *run_soltest` is equivalent to declaring a list item that contains `{"run": run_soltest}`, where you can think of `run_soltest` as a variable containing some dict.

In your change with `<<: *run_gitter_notify_steps`, on the other hand, you're taking a whole structure that looks like `{"steps": [{"run": ...}, {"run": ...}, ...]}` and trying to combine it with another such structure stored in `run_gitter_notify_steps` "variable" in a way that would result in the lists under the `"steps"` key being added together. There's no syntax for that in Yaml as far as I know.

.circleci/config.yml
cameel: Ah true, forgot that it had different steps.

This is really tripping me up :) We should probably rename `test_ubuntu2004` to `test_ubuntu2004_soltest_all` or something but, well, it's probably something better done in a separate cleanup PR.

cameel:
Looks like the only thing left to do here is to fix merging of `"steps"` and after that the PR should be good to go into `develop`.

.circleci/config.yml
bshastry: I've just unmerged this. Please feel free to unresolve this comment if you think otherwise.

.circleci/config.yml
```diff
@@ -802,13 +803,25 @@ jobs:
...
+    <<: *test_ubuntu2004_clang
+    steps:
+      - run: *gitter_notify_failure
+      - run: *gitter_notify_success
+
+  t_ubu_ubsan_clang_cli:
+    <<: *test_ubuntu2004_clang_cli
+    steps:
+      - run: *gitter_notify_failure
+      - run: *gitter_notify_success
```
cameel: This still overwrites `steps`, just now it does it without the `<<: *` syntax :)

Unfortunately you have to include these on the original list of steps. In `t_ems_ext` I solved it by making them conditional. You could do the same in `run_cmdline_tests_steps` and the other reusable step lists but then you would have to include `parameters` there and it gets complicated.

I'd say, let's go back to repeating the list of steps here. It's not the prettiest solution and unfortunately undoes some of the refactors you did here but at least will let us get this PR merged quickly :) We can then create and issue for someone to try to refactor it at leisure.

bshastry: Sounds good to me.

bshastry: Pushed an update, resolving this comment.

.circleci/config.yml
```diff
@@ -447,8 +448,7 @@ jobs:
       - store_artifacts: *artifacts_solc
       - persist_to_workspace: *artifacts_executables
 
-
-  b_ubu_asan_clang: &build_ubuntu2004_clang
+  b_ubu_asan_clang: &b_ubu_asan_clang
```
cameel, line 451: Looks like a good change. I only now realized that these `build_` and `b_` names refer to the same things.

Could you rename the others too for consistency? E.g. ` b_ubu_clang: &build_ubuntu2004_clang`.

cameel (dismissed):
I have one comment about the `b_`/`build_` naming that I think should be cleaned up but functionally the PR seems fine so I'm approving already.

.circleci/config.yml
```diff
@@ -121,18 +121,7 @@ defaults:
...
-      command: ./test/cmdlineTests.sh
-
-  - run_docs_pragma_min_version: &run_docs_pragma_min_version
-      name: docs pragma version check
-      command: ./scripts/docs_version_pragma_check.sh
-
-  - test_ubuntu1604_clang: &test_ubuntu1604_clang
-      docker:
-        - image: << pipeline.parameters.ubuntu-1604-clang-ossfuzz-docker-image >>
+  - run_soltest_steps: &run_soltest_steps
```
cameel, line 124: Actually, one more thing. I'm creating an issue for someone to do some cleanup in this file and I noticed that these `run_` dicts do not follow our current convention:
- The ones that contain `steps:` should actually start with `steps_`, not `run_`. E.g. `steps_soltest` and `steps_soltest_all`.
- There's a section at the beginning called `Build Templates`. It seems like the place where the `steps_` and `run_` dicts should be in.

I'm going to list that as one of the things to clean up in the new issue but you might want to do that already in this PR.

cameel: Here's the issue: #11846.

.circleci/config.yml
bshastry: Done.

cameel (approved):

Issue #11715, opened 2021-07-28
[Legacy/Sol->Yul] No compilation mode except optimized Sol->Yul issues a revert due to arithmetic overflow for exponentiation of very large numbers via a simple recursive algorithm


bshastry:
```
function exp(uint base, uint exponent) pure returns (uint power) {
  if (exponent == 0)
    return 1;
  power = exp(base, exponent / 2);
  power *= power;
  if (exponent & 1 == 1)
    power *= base;
}

contract C {
  function g(uint base, uint exponent) public pure returns (uint) {
    return exp(base, exponent);
  }
}
// ====
// compileViaYul: also
// ----
// g(uint256, uint256): 55558612478609019528521988286241949159871555179327291270583879839857001820084,15045494086288292508913825510347384281681719338099058151627027489948519117802 -> FAILURE
```

passes but with different EVMC status codes

Legacy: Stack underflow
Sol->Yul: Revert

Edit:

Reproduction

```
$ cp test.sol test/libsolidity/semanticTests/testStackUnderflow.sol
$ isoltest --optimize --show-messages -t semanticTests/testStackUnderflow
```

Ideally, legacy optimizer should also revert for this large exponentiation operation but it does not. Also note that the gas consumption in the legacy optimized call is unusually large (ideally, revert due to arithmetic overflow should occur in an earlier recursion call frame).

Gas table

| Gas consumption | Legacy optimized call | Sol->Yul optimized call |
| --- | --- | --- |
|  | 9223372036854775807  | 34583 |

_Originally posted by @bshastry in https://github.com/ethereum/solidity/issues/11709#issuecomment-887795572_

bshastry:
I'm not able to reproduce this issue independently of the fuzzer. Will reopen if I find a way to. Until then, treating this as a false positive.

bshastry:
Reopening this since it is reproducible via isoltest's `--optimize` flag. See edited OP.

bshastry:
Here's a summary of outcomes for legacy/Sol->Yul with/without opt

| Configuration | Gas cost | EVMC status code (Description) |
| --- | --- | --- |
| Legacy without optimizer | 9223372036854775807  | 6 (Stack underflow) |
| Sol->Yul without optimizer | 9223372036854775807  | 6 (Stack underflow) |
| Legacy with optimizer | 9223372036854775807  | 6 (Stack underflow) |
| Sol->Yul with optimizer | 34583  | 2 (Revert, panic code 0x11) |

[...]

PR #11824, opened 2021-08-19
Mark recursive calls in yul control flow graph.

ekpyron:
Adds an additional annotation to function call operations in the yul control flow graph: ``recursive``, which is true, iff the call is recursive.

Background: this information will be useful for opting for more aggressive stack compression before any recursive calls, s.t. we are less likely to hit the global 1024 stack slot limit of the EVM during deep recursions.

Related: https://github.com/ethereum/solidity/issues/11715
With this change and opting for aggressive stack compression for recursive calls, the maximal amount of recursions can be reached for the function in the issue.

### DIFF
--- a/libyul/backends/evm/ControlFlowGraph.h
+++ b/libyul/backends/evm/ControlFlowGraph.h
@@ -131,6 +131,9 @@ struct CFG
 		std::shared_ptr<DebugData const> debugData;
 		std::reference_wrapper<Scope::Function const> function;
 		std::reference_wrapper<yul::FunctionCall const> functionCall;
+		/// True, if the call is recursive, i.e. entering the function involves a control flow path (potentially involving
+		/// more intermediate function calls) that leads back to this very call.
+		bool recursive = false;
 	};
 	struct Assignment
 	{
--- a/libyul/backends/evm/ControlFlowGraphBuilder.cpp
+++ b/libyul/backends/evm/ControlFlowGraphBuilder.cpp
@@ -47,22 +47,14 @@ using namespace solidity;
 using namespace solidity::yul;
 using namespace std;
 
-std::unique_ptr<CFG> ControlFlowGraphBuilder::build(
-	AsmAnalysisInfo const& _analysisInfo,
-	Dialect const& _dialect,
-	Block const& _block
-)
+namespace
 {
-	auto result = std::make_unique<CFG>();
-	result->entry = &result->makeBlock();
-
-	ControlFlowGraphBuilder builder(*result, _analysisInfo, _dialect);
-	builder.m_currentBlock = result->entry;
-	builder(_block);
-
+// Removes edges to blocks that are not reachable.
+void cleanUnreachable(CFG& _cfg)
+{
 	// Determine which blocks are reachable from the entry.
-	util::BreadthFirstSearch<CFG::BasicBlock*> reachabilityCheck{{result->entry}};
-	for (auto const& functionInfo: result->functionInfo | ranges::views::values)
+	util::BreadthFirstSearch<CFG::BasicBlock*> reachabilityCheck{{_cfg.entry}};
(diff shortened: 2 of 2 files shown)


libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -85,6 +77,71 @@ std::unique_ptr<CFG> ControlFlowGraphBuilder::build(
 		cxx20::erase_if(node->entries, [&](CFG::BasicBlock* entry) -> bool {
 			return !reachabilityCheck.visited.count(entry);
 		});
+}
+// Sets the ``recursive`` member to ``true`` for all recursive function calls.
+void markRecursiveCalls(CFG& _cfg)
```
ekpyron, line 82: This is different from the call graph generator and the recursive function finder in the yul optimizer in two ways:
- It works on non-disambiguated ``Scope::Function``s.
- It determines which *calls* are recursive rather than which *functions* are recursive.

Both of them combined justify keeping this separate, I'd say.

libyul/backends/evm/ControlFlowGraph.h
```diff
@@ -131,6 +131,7 @@ struct CFG
 		std::shared_ptr<DebugData const> debugData;
 		std::reference_wrapper<Scope::Function const> function;
 		std::reference_wrapper<yul::FunctionCall const> functionCall;
+		bool recursive = false;
```
leonardoalt, line 136: Is it only direct? Or indirect too?

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -85,6 +77,71 @@ std::unique_ptr<CFG> ControlFlowGraphBuilder::build(
...
+			util::BreadthFirstSearch<CFG::FunctionCall*> breadthFirstSearch{{call}};
+			breadthFirstSearch.run([&](CFG::FunctionCall* _call, auto _addChild) {
+				auto& calledFunctionInfo = _cfg.functionInfo.at(&_call->function.get());
+				if (&calledFunctionInfo == &functionInfo)
+				{
+					call->recursive = true;
+					breadthFirstSearch.abort();
+					return;
+				}
+				for (CFG::FunctionCall* recursiveCalls: findCalls(_cfg.functionInfo.at(&_call->function.get()).entry))
```
leonardoalt: These are not necessarily `recursive` I guess?

libyul/backends/evm/ControlFlowGraph.h
ekpyron: I guess indirect, too - it's a call, s.t. entering it can end up at the same call again, i.e. a recursive one.
I.e.
```
function f() {
   f() // this call has recursive = true
  g() // this call has recursive = false
  h() // this call has recursive = true
}
function g()
{
  i() // this call has recursive = false - it calls something that's recursive, but the call is not recursive.
}
function h() {
  f() // this call has recursive = true
  g() // this call has recursive = false
}
function i() {
  i() // this call has recursive = true
}
```

libyul/backends/evm/ControlFlowGraphBuilder.cpp
ekpyron: Yeah, bad name indeed. More like ``nested``.

libyul/backends/evm/ControlFlowGraph.h
leonardoalt: Right

leonardoalt:
I would approve this, but it has no tests (understandably) and the new internal feature isn't used yet. I'll give it some time for someone else to complain, and will approve after otherwise.

ekpyron:
For the record: this is the use case: https://github.com/ethereum/solidity/pull/11493/files#r692197511

leonardoalt (approved):

Issue #10749, opened 2021-01-12
[SMTChecker] Add balance/block constraints between transactions


leonardoalt:
Between the end of a tx and the beginning of the next tx run by a contract:
- it's balance needs to be >= the previous balance
- block.timestamp must be >=
- block.number must be >=

Currently the SMTChecker does not add those constraints.
Anything else?

hrkrshnn:
> it's balance needs to be >= the previous balance

Why does this have to be true?

leonardoalt:
Because it can't decrease, but it can increase if it was used as the argument of a `selfdestruct`

hrkrshnn:
Does the block number and block time stamp have to be strictly greater?

axic:
> Because it can't decrease, but it can increase if it was used as the argument of a selfdestruct

Do you want to clarify this?

leonardoalt:
> Does the block number and block time stamp have to be strictly greater?

For new transactions I *think* so, for reentrant calls they should be the same.

leonardoalt:
> Do you want to clarify this?

This issue is about balance changes where the contract is not called. The only way for that to happen is when its address is used as the `selfdestruct` argument, and therefore the balance can only increase while the contract is idle.
Or what exactly is unclear?

axic:
It is not only `selfdestruct`, but the "coinbase transaction too" (e.g. being the mining recipient in a block).

leonardoalt:
Can you assign that to a contract?

leonardoalt:
> Can you assign that to a contract?

Confirmed yes.

djppdjpp325:
Предлагаю конкретное решение для SMTChecker, чтобы корректно учитывать ограничения на баланс и свойства блока между транзакциями:

🔧 Логика для SMTChecker:
Добавить инварианты:

balance_after(tx_i) ≤ balance_before(tx_{i+1}), с исключениями для selfdestruct, coinbase.

block.timestamp_i ≤ block.timestamp_{i+1}

block.number_i ≤ block.number_{i+1}

Ввести флаг isReentrantCall, чтобы разрешить равенство timestamp и block.number при реентерабельных вызовах.

Поддерживать аннотации вида @custom:smt-invariant для явного указания допустимых изменений.

🧪 Примеры для тестирования:
Контракт с инвариантами:

solidity
contract InvariantTracker {
    uint256 public lastBalance;
    uint256 public lastTimestamp;
    uint256 public lastBlock;

    function check() external {
[...]

PR #11828, opened 2021-08-20
[SMTChecker] Add proper constraints on the contract's balance

leonardoalt:
Ref https://github.com/ethereum/solidity/issues/10749

This PR is more easily reviewed by commit.

This PR correlates `address(this).balance` and `msg.sender` properly, following:
- At contract deployment, the address of the contract might have funds already, so we add the constraint `address(this).balance >= msg.value`.
- The balance of the contract might change between transactions because of potentially being used as `selfdestruct` or `coinbase` targets, so at the external entry point of every public function, we add the constraint `address(this).balance' = address(this).balance + msg.value + k` where `k` is an unbounded non-negative value, representing these potential increases that do not trigger calls to the contract.

The second item required a small but nice refactoring in the way that the transaction entry points were encoded.
For each function we have a function summary that represents the summary of the function body. To represent external calls into the contract, we had the rule
`interface && summary_f => interface'`
However, now we need to add more things here, so that we have something like
`interface && update_balance && summary_f => interface'`
and this can get complicated very easily.

What this PR does is:
- Create a new rule to separate these extra things done right before the transaction goes into the function, using a new type of summary "external summary":
```
true => external_f_entry
external_f_entry && update_balance && summary_f => external_summary_f
```
This predicate is then used as a constraint in the interface part:
`interface && external_summary_f => interface'`

It's nice that this is done here because the trusted mode PR also needs to do that.

It is also used by the nondeterministic interface rules instead of the normal summary, because that one is also related only to the external interface.


test/libsolidity/smtCheckerTests/types/address_transfer_3.sol
```diff
@@ -0,0 +1,15 @@
+contract C
+{
+	function f(address payable a) public {
+		require(1000 == address(this).balance);
+		require(100 == a.balance);
```
leonardoalt, line 5: This example is nice, the solver infers that `a == this` is not possible because the balances are different.

libsolidity/formal/CHC.cpp
```diff
@@ -169,7 +169,13 @@ void CHC::endVisit(ContractDefinition const& _contract)
 	smtutil::Expression zeroes(true);
 	for (auto var: stateVariablesIncludingInheritedAndPrivate(_contract))
 		zeroes = zeroes && currentValue(*var) == smt::zeroValue(var->type());
-	addRule(smtutil::Expression::implies(initialConstraints(_contract) && zeroes, predicate(entry)), entry.functor().name);
+	// The contract's address might already have funds before deployment,
+	// so the balance must be at least `msg.value`, but not equals.
+	auto initialBalance = state().balance(state().thisAddress()) >= state().txMember("msg.value");
```
hrkrshnn: Maybe changing the variable name to `initialBalanceConstraint`?

libsolidity/formal/CHC.cpp
```diff
@@ -270,21 +270,13 @@ bool CHC::visit(FunctionDefinition const& _function)
...
 
-	// The contract may have received funds through a selfdestruct or
-	// block.coinbase, which do not trigger calls into the contract.
-	// So the only constraint we can add here is that the balance of
-	// the contract grows by at least `msg.value`.
-	SymbolicIntVariable k{TypeProvider::uint256(), TypeProvider::uint256(), "funds_" + to_string(m_context.newUniqueId()), m_context};
-	state().addBalance(state().thisAddress(), k.currentValue());
-	m_context.addAssertion(k.currentValue() >= state().txMember("msg.value"));
-
-	auto bodyPred = predicate(*bodyBlock);
```
hrkrshnn: Wasn't this change made in the last commit. Perhaps these two changes can be removed?

docs/smtchecker.rst
```diff
@@ -830,6 +830,25 @@ are located in storage, even though they also have type ``uint[]``.  However,
...
+However, the contract's address may already have funds before deployment,
+which are kept by the contract.
+Therefore, the SMTChecker assumes that ``address(this).balance >= msg.value``
+in the constructor in order to be consistent with the EVM rules.
+The contract's balance may also increase without triggering any calls to the
+contract, if
+
+- ``selfdestruct`` is executed by another contract with the analyzed contract
+  as target of remaining funds
+- the contract is the ``coinbase`` of some block.
```
hrkrshnn:
```suggestion
- the contract address is the coinbase (i.e., ``block.coinbase``) of some block.
```

docs/smtchecker.rst
```diff
@@ -840,4 +859,6 @@ One of such cases is the length of a dynamic storage array overflowing during a
 push: If the ``push`` operation is applied to an array of length 2^256 - 1, its
```
hrkrshnn: Maybe it's a good idea to point to EIP-1985 in this section? It formalizes some upper bounds.

test/libsolidity/smtCheckerTests/types/address_transfer_3.sol
```diff
@@ -0,0 +1,15 @@
...
+		a.transfer(600);
```
hrkrshnn, line 6: But `transfer` can forward the eth to another contract, can't it?

hrkrshnn: Not sure about this. There is the 2300 gas stipend, and `call` is `2600` cold. But it can be made `100` by using access lists. I'll try to create an example case where that works.

libsolidity/formal/CHC.cpp
leonardoalt: Ah yes, the first commit introduced the idea, and this refactored it into its own function. I thought these were already in a single commit. I'll move the code changes from this commit to the first one.

test/libsolidity/smtCheckerTests/types/address_transfer_3.sol
leonardoalt: Not sure what exactly you mean. The goal of `a.transfer(600)` is exactly to send 600 wei to `a`

hrkrshnn: I meant that the condition `assert(a.balance == 700);` doesn't have to be true.

leonardoalt: Ohh that's true

leonardoalt:
@hrkrshnn updated review comments and merged the first two commits getting rid of the problem you mentioned.
I still need to add (or remove) some constraints to take care of the transfer case you mentioned.

test/libsolidity/smtCheckerTests/types/address_transfer_3.sol
hrkrshnn: But I'm not able to test it properly: https://gist.github.com/hrkrshnn/0066038f2e740a999290a1861d1118b5

leonardoalt: and that's not working?

hrkrshnn: We concluded [here](https://matrix.to/#/!poXqlbVpQfXKWGseLY:gitter.im/$812or6L6BR2cUPPCQ9aPIqOknLfm7MMo6l8ImZR6s20?via=gitter.im&via=matrix.org&via=ekpyron.org) that this is not a problem :)

docs/smtchecker.rst
```diff
@@ -830,6 +830,25 @@ are located in storage, even though they also have type ``uint[]``.  However,
...
+deployment transaction.
+However, the contract's address may already have funds before deployment,
+which are kept by the contract.
+Therefore, the SMTChecker assumes that ``address(this).balance >= msg.value``
+in the constructor in order to be consistent with the EVM rules.
+The contract's balance may also increase without triggering any calls to the
+contract, if
+
+- ``selfdestruct`` is executed by another contract with the analyzed contract
+  as target of remaining funds
```
cameel:
```suggestion
  as the target of the remaining funds,
```

test/libsolidity/smtCheckerTests/blockchain_state/transfer.sol
```diff
@@ -1,4 +1,5 @@
 contract C {
+	constructor() payable {}
```
cameel: Why are these constructors needed now?

test/libsolidity/smtCheckerTests/blockchain_state/balance_receive_4.sol
```diff
@@ -0,0 +1,21 @@
+contract C {
+	uint c;
+	function f() public payable {
+		require(msg.value > 10);
+		++c;
+	}
+	function inv() public view {
+		assert(address(this).balance >= c * 10); // should hold
```
cameel: This could be stronger because `msg.value` is guaranteed to be greater than 10: `address(this).balance > c * 10`. Or even `address(this).balance >= c * 11` since `msg.value` is an integer.

libsolidity/formal/CHC.cpp
```diff
@@ -315,11 +321,16 @@ void CHC::endVisit(FunctionDefinition const& _function)
 		shouldAnalyze(*m_currentContract)
```
cameel, line 321: Can it handle public library functions too? I.e. will `address(this).balance` be correctly treated as the balance of the contract rather than a separate thing? What about transfers to the contract initiated by the library?

If these are expected to work, can you add some tests that use libraries?

libsolidity/formal/CHC.cpp
```diff
@@ -1143,12 +1157,41 @@ void CHC::defineInterfacesAndSummaries(SourceUnit const& _source)
...
+	m_context.addAssertion(state().txTypeConstraints() && state().txFunctionConstraints(_function));
+
+	// The contract may have received funds through a selfdestruct or
+	// block.coinbase, which do not trigger calls into the contract.
+	// So the only constraint we can add here is that the balance of
+	// the contract grows by at least `msg.value`.
+	SymbolicIntVariable k{TypeProvider::uint256(), TypeProvider::uint256(), "funds_" + to_string(m_context.newUniqueId()), m_context};
+	m_context.addAssertion(k.currentValue() >= state().txMember("msg.value"));
+	// Assume that address(this).balance cannot overflow.
+	m_context.addAssertion(smt::symbolicUnknownConstraints(state().balance(state().thisAddress()) + k.currentValue(), TypeProvider::uint256()));
```
cameel, line 1187: Is there any particular reason why it's not `state().balance(state().thisAddress()) + state().txMember("msg.value") + k.currentValue()`? With the `k >= msg.value` constraint above it looks like it should amount to the same thing but I'm just curious why it differs from the PR description.

cameel (commented):
Seems good overall. I only have a few minor comments.

I do not know enough about how blocks and rules work in `CHC.cpp` to review changes done there properly but as far as I can tell, they seem to match your description. Your tests seem sensible so if they pass, it should be ok.

test/libsolidity/smtCheckerTests/blockchain_state/transfer.sol
leonardoalt: Otherwise the require two lines below is never true, and I didn't want to change the function.

test/libsolidity/smtCheckerTests/blockchain_state/balance_receive_4.sol
leonardoalt: True, I think `c * 11` should hold.

libsolidity/formal/CHC.cpp
leonardoalt: Hmm now I'm not sure. `internal` library functions should be the same as free functions, and those should take the context contract as the contract for all that stuff.
I'll check public library functions.

libsolidity/formal/CHC.cpp
leonardoalt: Exactly, since `k` is fresh and `k >= msg.value`, adding `+ msg.value` here is redundant. Symbolically `k` can already be the largest possible value, and here we're just saying "for however large `k` might be, the new balance must not overflow".

libsolidity/formal/CHC.cpp
leonardoalt: Ok you just found a bug in the way we encode `delegatecall` :)
```
library L {
	function l(address payable a) public {
		a.transfer(1);
	}
}

contract C {
	using L for address payable;
	uint x;
	function f(address payable a) public payable {
		require(msg.value > 1);
		uint b1 = address(this).balance;
		a.l();
		uint b2 = address(this).balance;
		assert(b1 == b2);
		assert(x == 0);
	}
}
```

When CHC sees a `delegatecall` is correctly erases all knowledge about state variables, since we don't know what the called code is doing, that's why the last assertion fails here. However, it incorrectly *does not* erase all balances, so the first assertion holds even though it shouldn't. I'll quickly fix that first in another PR.

leonardoalt: The fix: https://github.com/ethereum/solidity/pull/11841

test/libsolidity/smtCheckerTests/blockchain_state/transfer.sol
cameel: Wasn't this PR meant to change that assumption though? The contract could have some initial balance so this `require` could pass even if the contract has no payable constructor or functions.

leonardoalt: That's true. I need to check why the deployment procedure is always reverting in that case.

leonardoalt: You're right, the current code does work without the constructors. I must have added them with the first implementation which didn't take that into account yet. Just removed the 3 such cases.

leonardoalt:
Rebased and fixed comments.
Will squash when it's approved.

libsolidity/formal/SymbolicState.cpp
```diff
@@ -104,6 +104,14 @@ smtutil::Expression SymbolicState::blockhash(smtutil::Expression _blockNumber) c
 	return smtutil::Expression::select(m_tx.member("blockhash"), move(_blockNumber));
 }
 
+void SymbolicState::newBalances()
+{
+	auto stateS = dynamic_pointer_cast<TupleSort>(stateSort());
```
cameel: For a moment I started wondering what this mysterious state `S` is :) How about this?
```suggestion
	auto tupleSort = dynamic_pointer_cast<TupleSort>(stateSort());
```

cameel (dismissed):
Other than one minor naming suggestion, I don't see any problems in the PR and it passes tests so I think it's good to go.

libsolidity/formal/SymbolicState.cpp
leonardoalt: Yea I wanted to call it `stateSort` but the function is already called like that. Will change

leonardoalt:
@cameel I wanted to squash anyway, so fixed the variable name in the same push.

leonardoalt:
> @cameel I wanted to squash anyway, so fixed the variable name in the same push.

Squash a few of the extra commits I meant. The commits list is fine now.

cameel (approved):

Issue #10738, opened 2021-01-11
Replace loops by range-v3 loops where applicable.


ekpyron:
After https://github.com/ethereum/solidity/pull/10737 we should be able to replace nearly all non-range-based for loops by range-based loops using appropriate ranges.

chriseth:
oh yes!

axic:
And also replace the `boost::algorithm::reverse` iterator? I think ranges-v3 has some functionality for that?

ekpyron:
Especially if there is a ``ranges::cpp20::*`` version of any boost range algorithm or boost range adaptor, it'd make sense to switch to the ranges-v3 version.

cameel:
Two things that should be easy to find and replace are uses of `boost::adaptors::map_keys` and `boost::adaptors::map_values`. That would be `views::keys` and `views::values` now.

ekpyron:
Also to be clear in case any external contributor wants to take this: there's no need to find all cases where this can be done at once. We'll gladly accept PRs that change one or two for loops at a time.

ekpyron:
``vector_ref``, resp. ``bytesRef`` should also be replaced by ``ranges::span``.

axic:
> vector_ref, resp. bytesRef should also be replaced by ranges::span.

So I did this work but using `std::basic_string_view<uint8_t>`. It is quite a few changes, but simplifies a lot of things. Withheld those PRs are people were saturated with "technical debt" PRs after the "remove u160" thing :wink:

ekpyron:
> > vector_ref, resp. bytesRef should also be replaced by ranges::span.
> 
> So I did this work but using `std::basic_string_view<uint8_t>`. It is quite a few changes, but simplifies a lot of things. Withheld those PRs are people were saturated with "technical debt" PRs after the "remove u160" thing wink

That's immutable, though, isn't it? So while this may work for ``bytesConstRef`` it won't for ``bytesRef``, will it?
``ranges::span``, which will become C++20 ``std::span`` is the better choice in any case... and it'll work in the mutable and immutable case.

axic:
Yes, but we didn't actually need mutability in most of the places.

anurag-git:
Fixed following issues in #11193 
Replaced boost::adaptors::map_keys with ranges::views::keys and boost::adaptors::map_values with ranges::views::values.
Replaced boost::adaptors::reverse with ranges::views::reverse.

axic:
The cases remaining:
```
[...]

PR #11831, opened 2021-08-23
Replace boost::adaptors with range-v3

soroosh-sdi:
Ref #10738 

According to the above issue, I replaced two usages of boost::adaptors with range-v3 counterpart functions. The only thing that we don't have in cpp20 namespace is ranges::to_vector, but I couldn't resist not to use it.

### DIFF
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -41,11 +41,11 @@
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/algorithm/string/split.hpp>
-#include <boost/range/adaptor/sliced.hpp>
 #include <boost/range/algorithm/copy.hpp>
 
 #include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/reverse.hpp>
+#include <range/v3/view/tail.hpp>
 #include <range/v3/view/transform.hpp>
 
 #include <limits>
@@ -3694,7 +3694,7 @@ MemberList::MemberMap TypeType::nativeMembers(ASTNode const* _currentScope) cons
 			auto bases = contract.annotation().linearizedBaseContracts;
 			solAssert(bases.size() >= 1, "linearizedBaseContracts should at least contain the most derived contract.");
 			// `sliced(1, ...)` ignores the most derived contract, which should not be searchable from `super`.
-			for (ContractDefinition const* base: bases | boost::adaptors::sliced(1, bases.size()))
+			for (ContractDefinition const* base: bases | ranges::views::tail)
 				for (FunctionDefinition const* function: base->definedFunctions())
 				{
 					if (!function->isVisibleInDerivedContracts() || !function->isImplemented())
--- a/solc/CommandLineParser.cpp
+++ b/solc/CommandLineParser.cpp
@@ -25,6 +25,8 @@
 #include <boost/algorithm/string.hpp>
 
 #include <range/v3/view/transform.hpp>
+#include <range/v3/view/filter.hpp>
+#include <range/v3/range/conversion.hpp>
 
 using namespace std;
 using namespace solidity::langutil;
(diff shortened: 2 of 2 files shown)


libsolidity/ast/Types.cpp
```diff
@@ -47,6 +47,7 @@
 #include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/reverse.hpp>
 #include <range/v3/view/transform.hpp>
+#include <range/v3/view/drop.hpp>
```
hrkrshnn: Line 44 can be removed, can't it?

solc/CommandLineParser.cpp
```diff
@@ -961,7 +963,7 @@ General Information)").c_str(),
 		if (countEnabledOptions(nonAssemblyModeOptions) >= 1)
 		{
 			auto optionEnabled = [&](string const& name){ return m_args.count(name) > 0; };
-			auto enabledOptions = boost::copy_range<vector<string>>(nonAssemblyModeOptions | boost::adaptors::filtered(optionEnabled));
+			auto enabledOptions = nonAssemblyModeOptions | ranges::cpp20::views::filter(optionEnabled) | ranges::to_vector;
```
hrkrshnn: Is the cpp20 prefix needed here? Right now, we don't care about using things from the cpp20 namespace.

soroosh-sdi: no, it is not needed. I saw [this comment](https://github.com/ethereum/solidity/issues/10738#issuecomment-758266471) and I thought may be you care about it.

libsolidity/ast/Types.cpp
soroosh-sdi: yeah, it is removed now

solc/CommandLineParser.cpp
soroosh-sdi: I removed `cpp20 `

cameel: Well, AFAIK `cpp20` is what ended up in C++20 so sticking to it would make a transition easier in cases where the library evolved further.

But yeah, we don't do that consistently and the full namespaced names are already too long so I'd drop it and only worry about that when we upgrade to C++20 :)

libsolidity/ast/Types.cpp
```diff
@@ -41,12 +41,12 @@
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/algorithm/string/split.hpp>
-#include <boost/range/adaptor/sliced.hpp>
 #include <boost/range/algorithm/copy.hpp>
 
 #include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/reverse.hpp>
 #include <range/v3/view/transform.hpp>
+#include <range/v3/view/drop.hpp>
```
cameel: Just a tiny nitpick: we try to put imports in alphabetical order in each group. This should be above `enumerate.hpp`.

soroosh-sdi: OK, 
I used clang-format for this kind of thing, Why you don't use it on this repo?

hrkrshnn: We had a hard time fitting our coding style into clang-format, and changing the coding style was controversial. So here we are :)

soroosh-sdi: @cameel 
I pushed last changes

cameel:
@soroosh-sdi Please rebase your changes on top of #11832. It fixes the CI breakage. Once the CI passes on your PR, we will be able to merge it.

soroosh-sdi:
@cameel Thanks to you, now all of CIs are passed

libsolidity/ast/Types.cpp
```diff
@@ -3694,7 +3694,7 @@ MemberList::MemberMap TypeType::nativeMembers(ASTNode const* _currentScope) cons
 			auto bases = contract.annotation().linearizedBaseContracts;
 			solAssert(bases.size() >= 1, "linearizedBaseContracts should at least contain the most derived contract.");
 			// `sliced(1, ...)` ignores the most derived contract, which should not be searchable from `super`.
-			for (ContractDefinition const* base: bases | boost::adaptors::sliced(1, bases.size()))
+			for (ContractDefinition const* base: bases | ranges::views::drop(1))
```
hrkrshnn:
```suggestion
			for (ContractDefinition const* base: bases | ranges::views::tail)
```

hrkrshnn (commented):
Just a minor comment. Looks good otherwise.

libsolidity/ast/Types.cpp
soroosh-sdi: OK I'll push it in a minute.
to me both are the same,  why do you prefer `tail`?
I admit it looks nicer though.

hrkrshnn: I just like tail more because it's more obvious that `drop(1)`. With `drop`, I'll have to look up if it drops from the beginning or the end.

hrkrshnn (approved):
@soroosh-sdi Thanks for the PR. Looks good.

PR #11832, opened 2021-08-23
Fix pylint warnings about encoding in `open()`

cameel:
Fixes the failing [`chk_pylint` job](https://app.circleci.com/pipelines/github/ethereum/solidity/18273/workflows/9a7347b4-cbbd-4b99-a714-f017e128dd19/jobs/814903) in CI. [Pylint 2.10](https://pylint.pycqa.org/en/stable/whatsnew/2.10.html) emitts the `unspecified-encoding` warning if `open()` does not specify encoding and we had quite a few of these.

### DIFF
--- a/scripts/endToEndExtraction/remove-testcases.py
+++ b/scripts/endToEndExtraction/remove-testcases.py
@@ -61,7 +61,7 @@ def get_checks(content, sol_file_path):
             constructors.append(line)
         if line.startswith("ABI_CHECK") or line.startswith("BOOST_REQUIRE"):
             checks.append(line)
-    with open(sol_file_path, "r") as sol_file:
+    with open(sol_file_path, "r", encoding='utf8') as sol_file:
         sol_constructors = []
         sol_checks = []
         inside_expectations = False
@@ -118,7 +118,7 @@ def get_tests(e2e_path):
 
 def process_input_file(e2e_path, input_file, interactive):
     tests = get_tests(e2e_path)
-    with open(input_file, "r") as cpp_file:
+    with open(input_file, "r", encoding='utf8') as cpp_file:
         inside_test = False
         test_name = ""
         inside_extracted_test = False
--- a/scripts/fix_homebrew_paths_in_standalone_zip.py
+++ b/scripts/fix_homebrew_paths_in_standalone_zip.py
@@ -44,21 +44,20 @@ import subprocess
 import sys
 
 def readDependencies(fname):
-    with open(fname) as f:
-        with subprocess.Popen(['otool', '-L', fname], stdout=subprocess.PIPE) as o:
-            for line in o.stdout:
-                if line[0] == '\t':
-                    library = line.split(' ', 1)[0][1:]
-                    if (library.startswith("/usr/local/lib") or
-                            library.startswith("/usr/local/opt") or
-                            library.startswith("/Users/")):
-                        if (os.path.basename(library) != os.path.basename(fname)):
-                            command = "install_name_tool -change " + \
-                                library + " @executable_path/./" + \
-                                os.path.basename(library) + " " + fname
-                            print(command)
-                            os.system("chmod +w " + fname)
-                            os.system(command)
+    with subprocess.Popen(['otool', '-L', fname], stdout=subprocess.PIPE) as o:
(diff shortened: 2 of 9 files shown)


cameel:
Note that UTF-8 is not always the default encoding. From [Built-in Functions > open()](https://docs.python.org/3/library/functions.html#open):

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever `locale.getpreferredencoding()` returns), but any text encoding supported by Python can be used. See the `codecs` module for the list of supported encodings.

It should be the default at least in some cases so I think we should be ok just setting it to `'utf8'` though.

scripts/endToEndExtraction/remove-testcases.py
```diff
@@ -61,7 +61,7 @@ def get_checks(content, sol_file_path):
             constructors.append(line)
         if line.startswith("ABI_CHECK") or line.startswith("BOOST_REQUIRE"):
             checks.append(line)
-    with open(sol_file_path, "r") as sol_file:
+    with open(sol_file_path, "r", encoding='utf8') as sol_file:
```
cameel, line 64: I could not test this and `verify-testcases.py` but the only thing we have to worry about here are test cases for broken unicode sequences. I think `SolidityEndToEndTest` does not have any of these and they're all in standalone files.

scripts/fix_homebrew_paths_in_standalone_zip.py
```diff
@@ -44,21 +44,20 @@
 import sys
 
 def readDependencies(fname):
-    with open(fname) as f:
```
cameel, line 47: This opens a file but does not use it directly. It passes its name to the commands instead.

scripts/wasm-rebuild/docker-scripts/isolate_tests.py
```diff
@@ -36,7 +36,7 @@ def extract_test_cases(path):
 
 def extract_and_write(f, path):
     if f.endswith('.sol'):
-        with open(path, 'r') as _f:
+        with open(path, 'r', encoding='utf8') as _f:
```
cameel, line 39: Note: this script is unmaintained and uses Python 2. In Python 2 `open()` does not have the `encoding` argument so this will make it raise an exception. I added it anyway because there are other places in this file that already do this (so the script must be broken already) and it silences pylint.

scripts/endToEndExtraction/remove-testcases.py
hrkrshnn: What was the default encoding previously?

hrkrshnn: It seems to use `locale.getpreferredencoding(False)`. Which is `utf-8` in my system.

hrkrshnn (dismissed):
Looks good. Can be merged if CI works.

leonardoalt:
> Looks good. Can be merged if CI works.

You can use the auto-merge feature for that ;)

hrkrshnn:
Yeah, I didn't do anything since there seem to be some dependency with the windows PR.

scripts/endToEndExtraction/remove-testcases.py
cameel: Yeah, see https://github.com/ethereum/solidity/pull/11832#issuecomment-903999899. it's system-dependent. Which is why pylint now warns about it :)

It's only a problem if you expect byte sequences that are invalid UTF-8 in that file.

cameel:
~Closing, since the PR has been merged as a part of #11835.~

cameel:
Or not. Looks like the merge went into this branch rather than into `develop`. We can merge this one then since it should now pass all the checks.

leonardoalt:
but now this PR has a Merge commit from the other one. Do we wanna keep that?

cameel:
I don't think it's a problem. The merge commits used to update a branch are bad because you can potentially get "hidden" changes from conflict resolutions in them. But this one is just an empty merge commit that marks a point where two branches were joined, just like the ones on `develop`. I'd even say that it's a good thing because it shows that here were two branches.

But if you think it's better not to have it, I can submit a new PR.

leonardoalt (approved):

PR #11833, opened 2021-08-23
Enable more pylint checks

cameel:
~Depends on #11832.~ Merged.

This is an old branch I prepared while working on the bytecode compare script when I discovered that some simple mistakes like unused imports or variables go unnoticed by pylint. I never submitted though because we had a big PR backlog.

Well, we still have a backlog, but having to fix #11832 seems like a good occasion to finally push it through :)

### DIFF
--- a/scripts/bytecodecompare/prepare_report.py
+++ b/scripts/bytecodecompare/prepare_report.py
@@ -413,7 +413,7 @@ def commandline_parser() -> ArgumentParser:
         action='store_true',
         help="Immediately exit and print compiler output if the compiler exits with an error.",
     )
-    return parser;
+    return parser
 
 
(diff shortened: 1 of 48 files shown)


cameel:
Looks like my import tweaks broke `chk_proofs`. It's because some files import stuff from `z3` indirectly, via `opcodes.py`. Switching to draft now. I'll get back to it when I'm done fixing the other CI problems.

cameel:
This is finally reviewable. My lazy way of ignoring the warnings in `test/formal` didn't work out and in the end it was simpler to just go other all the files and add explicit imports than to look for another workaround.

chriseth (approved):

PR #11834, opened 2021-08-23
Improve `pylint_all.sh`

cameel:
This PR refactors the `pylint_all.sh` script and fixes a few annoying things about it:
- The script used to run pylint separately on each file, which is **slow**. Now it passes them all in one go.
    - It reverts to the old behavior if you want it to stop on first error (because pylint does not have an option for that).
    - Now pylint can detect `duplicate-code`. Unfortunately we have a lot of duplicated code in `test/formal/`, which is why I had to disable it.
- I enabled pylint's colored output and disabled the score summary. Especially when running on files individually, the summary makes the output hard to read.
- It now uses `subprocess` instead `os.system()`, which means that it actually stops on Ctrl+C. It also no longer prints the huge stack trace if you kill it that way (well, mostly - if you do it at the wrong time, it still prints it but even then it's shorter :)).
- It now uses `argparse` for sane command-line processing.

### DIFF
--- a/scripts/pylint_all.py
+++ b/scripts/pylint_all.py
@@ -1,14 +1,13 @@
 #! /usr/bin/env python3
 
 """
-Performs pylint on all python files in the project repo's {test,script,docs} directory recursively.
-
-This script is meant to be run from the CI but can also be easily in local dev environment,
-where you can optionally pass `-d` as command line argument to let this script abort on first error.
+Runs pylint on all Python files in project directories known to contain Python scripts.
 """
 
+from argparse import ArgumentParser
 from os import path, walk, system
-from sys import argv, exit as exitwith
+from sys import exit as exitwith
+from textwrap import dedent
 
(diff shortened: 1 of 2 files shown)


scripts/pylint_all.py
```diff
@@ -1,57 +1,102 @@
...
+    parser = ArgumentParser(description=script_description)
+    parser.add_argument(
+        '-d', '--dev-mode',
+        dest='dev_mode',
+        default=False,
+        action='store_true',
+        help=(
+            "Abort on first error. "
+            "In this mode every script is passed to pylint separately. "
+        )
```
cameel, line 77: Not sure if anyone actually uses this mode but I kept it just in case. I also kept the same short option (`-d`) even though I think a better name would be something like `--fail-fast`.

hrkrshnn (approved):
Looks good. Didn't check thoroughly, however fine to merge if it works.

m14253647g (commented):
Y

PR #11835, opened 2021-08-23
Use empty user agent when downloading Boost from jfrog in powershell

cameel:
This fixes the [`b_win`](https://app.circleci.com/pipelines/github/ethereum/solidity/18272/workflows/f6534c77-5660-4de4-af2a-f2f9abc3db43/jobs/814852) and `b_win_release` jobs in CI. They have been failing intermittently for some time and now they seem to be failing in all PRs.

Looks like it's a server-side problem: [[RTFACT-26216] Native browser redirect prevents windows Power Shell (Invoke-WebRequest command) from downloading a file](https://www.jfrog.com/jira/si/jira.issueviews:issue-html/RTFACT-26216/RTFACT-26216.html). One of the workarounds is to use a different user agent. An empty one does the job.

EDIT: I just realized that this and #11832 can't be merged separately because the failing checks in both of them are marked as `required`. I'm putting this PR on top of #11832 to make it mergable. Please review them both and then merge just this one.

### DIFF
--- a/scripts/install_deps.ps1
+++ b/scripts/install_deps.ps1
@@ -10,7 +10,9 @@ if ( -not (Test-Path "$PSScriptRoot\..\deps\boost") ) {
   tar -xf cmake.zip
   mv cmake-3.18.2-win64-x64 "$PSScriptRoot\..\deps\cmake"
 
-  Invoke-WebRequest -URI "https://boostorg.jfrog.io/artifactory/main/release/1.74.0/source/boost_1_74_0.zip" -OutFile boost.zip
+  # FIXME: The default user agent results in Artifactory treating Invoke-WebRequest as a browser
+  # and serving it a page that requires JavaScript.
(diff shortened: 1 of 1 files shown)


hrkrshnn (approved):
Can be merged if CI passes.

leonardoalt:
But merging this one still requires the other one to be merged

cameel:
> But merging this one still requires the other one to be merged

Interestingly, it doesn't :) I got a few cases where my PR that depended on another was merged without merging that other one.

I mean, github might since then have introduced some safeguard for PRs that have a different base branch. But in that case changing it to `develop` would work around that.

An in the wost case I can always just create a new PR, but I wanted to avoid having too many of them.

cameel:
Wow, it even let me merge it myself :)

cameel:
Oh, wait it went into the base branch, not into `develop`. So looks like the behavior did change. ~Oh well, I'll create a new PR then.~

cameel:
Or we can just merge #11832 as @leonardoalt suggested.

PR #11311, opened 2021-04-24
Use range-v3 loops(remove_erase, remove_erase_if)

anurag-git:
Partial fix for #10738
Replaced boost::remove_erase with ranges::actions::remove
Replaced boost::remove_erase_if with ranges::actions::remove_if

PR #11837, opened 2021-08-24
Using range-v3 instead of boost

soroosh-sdi:
Ref #10738 

### DIFF
--- a/libsolidity/analysis/ControlFlowAnalyzer.cpp
+++ b/libsolidity/analysis/ControlFlowAnalyzer.cpp
@@ -20,7 +20,8 @@
 
 #include <liblangutil/SourceLocation.h>
 #include <libsolutil/Algorithms.h>
-#include <boost/range/algorithm/sort.hpp>
+
+#include <range/v3/algorithm/sort.hpp>
 
 #include <functional>
 
@@ -141,7 +142,7 @@ void ControlFlowAnalyzer::checkUninitializedAccess(CFGNode const* _entry, CFGNod
 			exitInfo.uninitializedVariableAccesses.begin(),
 			exitInfo.uninitializedVariableAccesses.end()
 		);
-		boost::range::sort(
+		ranges::sort(
 			uninitializedAccessesOrdered,
 			[](VariableOccurrence const* lhs, VariableOccurrence const* rhs) -> bool
 			{
--- a/libsolidity/ast/ASTJsonConverter.cpp
+++ b/libsolidity/ast/ASTJsonConverter.cpp
@@ -34,7 +34,8 @@
 #include <libsolutil/UTF8.h>
 
 #include <boost/algorithm/string/join.hpp>
-#include <boost/range/algorithm/sort.hpp>
+
+#include <range/v3/algorithm/sort.hpp>
 
 #include <utility>
 #include <vector>
@@ -563,8 +564,9 @@ bool ASTJsonConverter::visit(InlineAssembly const& _node)
 
 	Json::Value externalReferencesJson = Json::arrayValue;
 
-	for (auto&& it: boost::range::sort(externalReferences))
-		externalReferencesJson.append(std::move(it.second));
+	ranges::sort(externalReferences);
+	for (Json::Value& it: externalReferences | ranges::views::values)
+		externalReferencesJson.append(std::move(it));
 
 	setJsonNode(_node, "InlineAssembly", {
 		make_pair("AST", Json::Value(yul::AsmJsonConverter(sourceIndexFromLocation(_node.location()))(_node.operations()))),
--- a/libsolidity/parsing/DocStringParser.cpp
+++ b/libsolidity/parsing/DocStringParser.cpp
@@ -24,8 +24,9 @@
 #include <liblangutil/ErrorReporter.h>
 #include <liblangutil/Exceptions.h>
 
-#include <boost/range/algorithm/find_first_of.hpp>
-#include <boost/range/irange.hpp>
+#include <range/v3/algorithm/find_first_of.hpp>
+#include <range/v3/algorithm/find_if_not.hpp>
+#include <range/v3/view/subrange.hpp>
 
 using namespace std;
 using namespace solidity;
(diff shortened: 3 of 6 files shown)


libsolidity/analysis/ControlFlowAnalyzer.cpp
```diff
@@ -141,7 +142,7 @@ void ControlFlowAnalyzer::checkUninitializedAccess(CFGNode const* _entry, CFGNod
 			exitInfo.uninitializedVariableAccesses.begin(),
 			exitInfo.uninitializedVariableAccesses.end()
 		);
-		boost::range::sort(
+		ranges::sort(
```
leonardoalt, line 145: @cameel @ekpyron I don't know who said it, or if I'm remembering correctly, but didn't someone say these are not equivalent?

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -563,7 +564,8 @@ bool ASTJsonConverter::visit(InlineAssembly const& _node)
 
 	Json::Value externalReferencesJson = Json::arrayValue;
 
-	for (auto&& it: boost::range::sort(externalReferences))
+	ranges::sort(externalReferences);
```
leonardoalt, line 567: Does this change it in place? It looked like the previous one was pure

soroosh-sdi: yes, both of them sort the container in place. look at this [example](http://coliru.stacked-crooked.com/a/18c07a3c138d6218)

soroosh-sdi: boost version returns a reference of the container. it is simpler to use but confusing; at first glance it looks like it is pure an sort lazy, but we don't have lazy sort in boost or std or range-v3.

libsolidity/parsing/DocStringParser.cpp
```diff
@@ -52,8 +53,7 @@ string::const_iterator firstNonIdentifier(
 	if (currPos == _pos && isIdentifierStart(*currPos))
 	{
 		currPos++;
-		while (currPos != _end && isIdentifierPart(*currPos))
-			currPos++;
+		currPos = ranges::find_if_not(ranges::subrange(currPos, _end), [](char const& it) { return isIdentifierPart(it); });
```
soroosh-sdi: a CI failed
maybe I should use make_subrange

libsolidity/analysis/ControlFlowAnalyzer.cpp
cameel: I only remember something like that about `boost::remove_erase_if()` vs `ranges::actions::remove_if()`? (https://github.com/ethereum/solidity/pull/11311#issuecomment-826132661).

I think that we do not have this problem with `sort()`. Both sort in place.

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -563,7 +564,8 @@ bool ASTJsonConverter::visit(InlineAssembly const& _node)
...
+	for (auto&& it: externalReferences)
 		externalReferencesJson.append(std::move(it.second));
```
cameel: I think this could be refactored using `values`:
```suggestion
	for (Json::Value& identifier: externalReferences | ranges::views::values)
		externalReferencesJson.append(std::move(identifier));
```

libsolidity/parsing/DocStringParser.cpp
```diff
@@ -52,8 +53,7 @@ string::const_iterator firstNonIdentifier(
 	if (currPos == _pos && isIdentifierStart(*currPos))
 	{
 		currPos++;
-		while (currPos != _end && isIdentifierPart(*currPos))
-			currPos++;
+		currPos = ranges::find_if_not(ranges::make_subrange(currPos, _end), [](char const& it) { return isIdentifierPart(it); });
```
cameel: `isIdentifierPart` is already a function. Does it really need to be wrapped in a lambda?
```suggestion
		currPos = ranges::find_if_not(ranges::make_subrange(currPos, _end), isIdentifierPart);
```

libsolidity/parsing/DocStringParser.cpp
```diff
@@ -63,7 +63,7 @@ string::const_iterator firstWhitespaceOrNewline(
 	string::const_iterator _end
 )
 {
-	return boost::range::find_first_of(make_pair(_pos, _end), " \t\n");
+	return ranges::find_first_of(ranges::make_subrange(_pos, _end),  " \t\n");
```
cameel:
```suggestion
	return ranges::find_first_of(ranges::make_subrange(_pos, _end), " \t\n");
```

libsolidity/parsing/DocStringParser.cpp
```diff
@@ -72,10 +72,7 @@ string::const_iterator skipWhitespace(
 	string::const_iterator _end
 )
 {
-	auto currPos = _pos;
-	while (currPos != _end && (*currPos == ' ' || *currPos == '\t'))
-		currPos += 1;
-	return currPos;
+	return ranges::find_if_not(ranges::make_subrange(_pos, _end), [](char const& it) { return (it == ' ' || it == '\t'); });
```
cameel: These expressions are much less readable if the lambda is defined inline.
```suggestion
	auto isWhitespace = [](char const& c) { return (c == ' ' || c == '\t'); };
	return ranges::find_if_not(ranges::make_subrange(_pos, _end), isWhitespace);
```

libsolidity/ast/ASTJsonConverter.cpp
soroosh-sdi: done

libsolidity/parsing/DocStringParser.cpp
soroosh-sdi: done

libsolidity/parsing/DocStringParser.cpp
soroosh-sdi: done

libsolidity/parsing/DocStringParser.cpp
soroosh-sdi: done

soroosh-sdi:
@cameel Thanks for the precise review. I pushed a new version

cameel (approved):
Thanks. Looks good now.

libsolidity/ast/ASTJsonConverter.cpp
leonardoalt: Ok!

PR #11839, opened 2021-08-24
Don't create empty expectation files when updating test expectations

### LINKED ISSUE #11753: Do not create empty superfluous `err`, `output` and `exit` files when auto-updating command-line test expectations
This came up in https://github.com/ethereum/solidity/pull/11738#discussion_r684159827

When the [`cmdlineTests.sh`](https://github.com/ethereum/solidity/blob/develop/test/cmdlineTests.sh) script gets no output from the compiler, it accepts both an empty `output` file as well as the file not being there at all. Same with stderr and the `err` file. In case of `exit` file, it not being present is equivalent with expecting a return code of `0`.

If the compiler output does not match the expectations, the script can update the content of these files but then it always creates them even if they are empty. It would be better if instead it did not and if it deleted them if they already exist.

Command-line tests can be found in [test/cmdlineTests/](https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests).

priyansh786:
Fixes: https://github.com/ethereum/solidity/issues/11753
Pinging @cameel for reviews

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -71,6 +71,15 @@ function update_expectation {
     local newExpectation="${1}"
     local expectationFile="${2}"
 
+    if [[ $newExpectation == "" || $newExpectation -eq 0 && $expectationFile == */exit ]]
+    then
+        if [[ -f $expectationFile ]]
+        then
+            rm "$expectationFile"
+        fi
+        return
+    fi
+
     echo "$newExpectation" > "$expectationFile"
     printLog "File $expectationFile updated to match the expectation."
 }


test/cmdlineTests.sh
```diff
@@ -71,6 +71,15 @@ function update_expectation {
     local newExpectation="${1}"
     local expectationFile="${2}"
 
+    if [[ $newExpectation == "" || $newExpectation -eq 0 ]]
```
cameel: `0` only applies to the `exit` file. In `err` and `output` it would be valid output and should not result in the file being omitted/deleted.

Please add an extra condition that `$expectationFile` also ends with `/exit`.

test/cmdlineTests.sh
```diff
@@ -71,6 +71,15 @@ function update_expectation {
...
+    then
+    	if [[ -f $expectationFile ]]
+    	then
```
cameel: Looks like you're mixing tabs and spaces here. Note that the file uses spaces for indentation.

test/cmdlineTests.sh
```diff
@@ -71,6 +71,15 @@ function update_expectation {
...
+    	    rm -f "$expectationFile"
```
cameel: Why do you need `--force` here?
```suggestion
    	    rm "$expectationFile"
```

cameel (changes requested):
Thanks for the PR!

I have some remarks but it looks good overall.

cameel:
Please also rebase your branch on top of `develop` to make the CI pass. The three failing jobs are unrelated to this PR and were fixed in #11832 and #11835 merged yesterday.

test/cmdlineTests.sh
```diff
@@ -71,6 +71,15 @@ function update_expectation {
     local newExpectation="${1}"
     local expectationFile="${2}"
 
+    if [[ $newExpectation == "" || $newExpectation -eq 0 && $expectationFile == "*/exit" ]]
```
cameel: I think that globbing won't work inside a quoted string.
```suggestion
    if [[ $newExpectation == "" || $newExpectation -eq 0 && $expectationFile == */exit ]]
```

cameel (commented):
Just one more change needed. Other than that it looks good!

cameel:
The CI is failing because your PR is based on an older commit where we had some breakage. Please rebase on top of latest `develop`. Also, please squash the commits with review fixes.

cameel (approved):
Thanks! This looks good to go now.

Issue #11249, opened 2021-04-13
[SMTChecker] Take function options into account


leonardoalt:
I think this wasn't done yet

PR #11841, opened 2021-08-25
[SMTChecker] Erase balances when delegatecall is seen

leonardoalt:
Found in https://github.com/ethereum/solidity/pull/11828

Storage is being erased properly when we see a `delegatecall`, but balances are not. This PR fixes that.
Note that the TODO line added here and some of the tests will change in https://github.com/ethereum/solidity/pull/11828, when proper support connecting `msg.value` and `balance` will make the internal library/free functions work. The public library functions should remain the same.

### DIFF
--- a/libsolidity/formal/BMC.cpp
+++ b/libsolidity/formal/BMC.cpp
@@ -587,6 +587,11 @@ void BMC::internalOrExternalFunctionCall(FunctionCall const& _funCall)
 			_funCall.location(),
 			"BMC does not yet implement this type of function call."
 		);
+	else if (funType.kind() == FunctionType::Kind::BareStaticCall)
+	{
+		// Do nothing here.
+		// Neither storage nor balances should be modified.
+	}
 	else
 	{
 		m_externalFunctionCallHappened = true;
--- a/libsolidity/formal/BMC.cpp
+++ b/libsolidity/formal/BMC.cpp
@@ -591,6 +591,7 @@ void BMC::internalOrExternalFunctionCall(FunctionCall const& _funCall)
 	{
 		m_externalFunctionCallHappened = true;
 		resetStorageVariables();
+		resetBalances();
 	}
 }
 
--- a/libsolidity/formal/SMTEncoder.cpp
+++ b/libsolidity/formal/SMTEncoder.cpp
@@ -2281,6 +2281,13 @@ void SMTEncoder::resetStorageVariables()
 	});
 }
 
+void SMTEncoder::resetBalances()
+{
+	// TODO this should be changed to `balances` only
+	// when `state` gets more members.
+	state().newState();
+}
+
 void SMTEncoder::resetReferences(VariableDeclaration const& _varDecl)
 {
 	m_context.resetVariables([&](VariableDeclaration const& _var) {
--- a/test/libsolidity/smtCheckerTests/blockchain_state/library_public_1.sol
+++ b/test/libsolidity/smtCheckerTests/blockchain_state/library_public_1.sol
@@ -1,0 +1,23 @@
+library L {
+	function l(address payable a) public {}
+}
+
+contract C {
+	using L for address payable;
+	uint x;
+	function f(address payable a) public payable {
+		require(msg.value > 1);
+		uint b1 = address(this).balance;
+		a.l();
+		uint b2 = address(this).balance;
+		assert(b1 == b2); // should fail because the called library can transfer with `this`s balance
+		assert(x == 0); // should fail because of `delegatecall`
+	}
+}
+// ====
+// SMTEngine: all
+// ----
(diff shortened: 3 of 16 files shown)


libsolidity/formal/SMTEncoder.cpp
```diff
@@ -2281,6 +2281,13 @@ void SMTEncoder::resetStorageVariables()
 	});
 }
 
+void SMTEncoder::resetBalances()
+{
+	// TODO this should be changed to `balances` only
+	// when `state` gets more members.
```
leonardoalt, line 2287: This will happen in the next PR.

test/libsolidity/smtCheckerTests/blockchain_state/library_public_1.sol
```diff
@@ -0,0 +1,23 @@
+library L {
+	function l(address payable a) public {}
```
hrkrshnn, line 2: What happens if it returns `msg.value`? Can it correctly assert that this is same as `msg.value` for `C.f(..)`? (Something like the recent sushi bug.)

leonardoalt: Nop, because there's no support at all to `delegatecall` currently, we just erase everything and report false positives.
I'll add a test with that and msg.sender.

leonardoalt: Also this is the same situation as external calls to code that is available at compile time, but not compiled together so you can't know for sure that the library's code is the one you're going to execute in runtime.
In the future we could try to support that
- if the `delegatecall` is to `address(this)`, like the recent sushi bug
- in the trusted mode

leonardoalt:
@hrkrshnn added more tests

cameel (dismissed):
Looks fine.

cameel:
Looks like counterexamples are slightly different on macOS though and CI fails.

leonardoalt:
> Looks like counterexamples are slightly different on macOS though and CI fails.

As always...
just pushed a fix, let's see

leonardoalt:
@cameel can you reapprove?

libsolidity/formal/BMC.cpp
```diff
@@ -591,6 +591,7 @@ void BMC::internalOrExternalFunctionCall(FunctionCall const& _funCall)
 	{
 		m_externalFunctionCallHappened = true;
 		resetStorageVariables();
+		resetBalances();
```
hrkrshnn, line 599: This is the balance of the current contract, isn't it? Do we have to erase it for all calls other than `internal`? If the value field is not set for regular `call`, we don't need this. Similarly, for `staticcall`.

hrkrshnn: I see that maybe a callback can change balances for a regualar `call`. But still for `staticall`, we shouldn't change this.

leonardoalt: Why would we erase it for `internal` calls? Those get inlined at the beginning of the function.
But yea, we could filter it better.
I would do the value field in a different PR, targeting https://github.com/ethereum/solidity/issues/11249
Maybe it makes sense to do both that and `staticcall` in the same PR?

I kind of want to keep this one here minimal, even if it introduces false positives (but still sound), to move on with the other PR which is more important.

leonardoalt: I only saw your second comment after writing the above.
I'd personally prefer doing the `staticcall` part in a different PR, but don't have too strong of an opinion here. If you think it should be in this one I'll change it

hrkrshnn: I meant other than `internal` in the sense that that was the condition for this else branch.

> Maybe it makes sense to do both that and staticcall in the same PR

Yeah, isn't that change simple though? adding another `else if (funType.kind() == FunctionType::Kind::Staticcall) { // dont' reset value or state }`

leonardoalt: It's simple, but I'm annoyed by the tests duration and blocking the other PR for another 2h (at least) which is in turn blocking another PR. But I'll change it.

hrkrshnn: If CI is the problem, then I don't mind merging this and fixing staticcall later.

leonardoalt: It's fine, I'll just do some other stuff in the meantime. Running the tests locally, will probably add some more to enforce that BMC is running on staticcall

leonardoalt:
@hrkrshnn changed code and added tests specifically to BMC (CHC should be correct already)

cameel (approved):
Changes look good. Reapproving.

hrkrshnn (approved):

PR #11843, opened 2021-08-25
Disallow delete on types containing nested mappings.

### LINKED ISSUE #8535: Disallow deleting mappings either directly or via array.pop
* [x] Any operation that causes a delete on a mapping in storage should be disallowed.
* [x] Any assignment to something in storage that contains a mapping is disallowed.
  * [x] assignment to struct that contains mapping is disallowed
* [x] Array pop and (non-empty) push on an array that (recursively) contains a mapping is disallowed.
  * [x] push
  * [x] pop 
* [x] Delete on anything that (recursively) contains a mapping is disallowed.
Empty push on such arrays is fine.

Marenz:
fixes #8535


Changelog.md
```diff
@@ -1,8 +1,10 @@
 ### 0.9.0 (unreleased)
 
 Breaking changes:
- * `error` is now a keyword that can only be used for defining errors.
+ * Disallow .pop() on arrays containing nested mappings.
```
```suggestion
 * Disallow ``.pop()`` on arrays containing nested mappings.
```

Changelog.md
```diff
@@ -1,8 +1,10 @@
...
+ * Disallow delete on types that contain nested mappings.
  * Inline Assembly: Consider functions, function parameters and return variables for shadowing checks.
+ * `error` is now a keyword that can only be used for defining errors.
```
hrkrshnn: Just realized that this should also be double backtick.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2839,6 +2846,17 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
+			funType->kind() == FunctionType::Kind::ArrayPop &&
+			exprType->nameable() &&
+			exprType->containsNestedMapping()
+		)
+			m_errorReporter.typeError(
+				6298_error,
+				_memberAccess.location(),
+				"Storage arrays with nested mappings do not support .pop()."
+			);
+
```
hrkrshnn: Extra line?

test/libsolidity/semanticTests/storage/mappings_array_pop_delete.sol
```diff
@@ -1,36 +0,0 @@
...
-
-	function map(uint key) public view returns (uint) {
-		return a[a.length - 1][key];
-	}
-
-	function p() public {
-		a.pop();
-	}
-
-	function d() public returns (uint) {
```
hrkrshnn, line 21: Can we salvage this test by removing just this function? Similarly for the other test.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2839,6 +2846,17 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
 			);
 
 		if (
+			funType->kind() == FunctionType::Kind::ArrayPop &&
+			exprType->nameable() &&
```
hrkrshnn: I know that `nestedMapping` has an assert for `nameable`. Wondering if we need the `exprType->nameable()` condition here? Do you have an example that throws otherwise?

test/libsolidity/semanticTests/storage/mappings_array_pop_delete.sol
Marenz: Well, given that the whole point of the test is to test the deleting and popping (see the name) I didn't see much use in keeping it as both is not possible anymore now..

libsolidity/analysis/TypeChecker.cpp
Marenz: You're right it is not needed at this particular location. Only at the other place where I added it. There it would throw in `signed_rational_modulus.sol`. Will remove it here.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1613,6 +1613,13 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 	_operation.annotation().isPure = !modifying && *_operation.subExpression().annotation().isPure;
 	_operation.annotation().isLValue = false;
 
+	if (subExprType->nameable() && subExprType->containsNestedMapping())
```
hrkrshnn: What about something like
```solidity
contract C {
    mapping (uint => uint) m;
    function f() external {
        m[1]++;
    }
}
```
Wouldn't that also make it here?

Marenz: uh. You're right I forgot to actually check for the delete operation :sweat_smile:

Marenz: Actually no. Your example works fine because `m[1]` does NOT contain a nested mapping. It would need to be
```
contract C {
    mapping (uint => uint) m;
    function f() external {
        m++;
    }
}
```
which isn't supported..

Marenz: However, I should still add a check for the `Delete` token

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2838,16 +2845,24 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
-			funType->kind() == FunctionType::Kind::ArrayPush &&
-			arguments.value().numArguments() != 0 &&
-			exprType->containsNestedMapping()
-		)
-			m_errorReporter.typeError(
-				8871_error,
-				_memberAccess.location(),
-				"Storage arrays with nested mappings do not support .push(<arg>)."
-			);
+		if (exprType->containsNestedMapping())
```
hrkrshnn: I'm not sure if this should be the first condition, becuase that would throw if the type is not nameable:

```
contract C {
    function f() external {
        C.f();
    }
}
```

```
: Internal compiler error during compilation:
: /solidity/libsolidity/ast/Types.h(278): Throw in function virtual bool solidity::frontend::Type::containsNestedMapping() const
: Dynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>
: std::exception::what: Called for a non nameable type.
: [solidity::util::tag_comment*] = Called for a non nameable type.
```

Maybe check in here if `kind` is `ArrayPop` or `ArrayPush`? That should fix this problem.

Marenz: Meh. I think my original solution was the cleanest here.. reverting to it..

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1613,6 +1613,13 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 	_operation.annotation().isPure = !modifying && *_operation.subExpression().annotation().isPure;
 	_operation.annotation().isLValue = false;
 
+	if (op == Token::Delete && subExprType->nameable() && subExprType->containsNestedMapping())
```
hrkrshnn: With the `delete` check, I feel that `nameable` condition can be removed. But not sure :(

Marenz: I removed it and checked `isoltest` locally. If the CI is happy I'd say it's safe?

Marenz: I looked a bit through the code and tested some cases and it seems this is good.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1613,6 +1613,13 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 	_operation.annotation().isPure = !modifying && *_operation.subExpression().annotation().isPure;
 	_operation.annotation().isLValue = false;
 
+	if (op == Token::Delete && subExprType->containsNestedMapping())
```
chriseth: Is it more convenient to move this check inside `Type::unaryOperatorResult`? We would only need it on reference types.

Marenz: Could could move it, but the error reporter is not available in the type and in this particular visit, the string part of the return value of that function is also not used for the error message

Marenz: I could change the call here to use the err message returned.. it would change a few.. like this:

```
syntaxTests/negation.sol: FAIL
  Contract:
    contract test {
        function f() public pure {
            int x;
            uint y = uint(-x);
            -y;
        }
    }

  Expected result:
    TypeError 4907: (97-99): Unary operator - cannot be applied to type uint256
  Obtained result:
    TypeError 4907: (97-99): Unary operator - cannot be applied to type uint256: Unary negation is only allowed for signed integers.


```

chriseth: That actually sounds good!

Changelog.md
```diff
@@ -1,7 +1,10 @@
 ### 0.9.0 (unreleased)
 
 Breaking changes:
+ * Disallow ``.pop()`` on arrays containing nested mappings.
+ * Disallow ``delete`` on types that contain nested mappings.
  * Inline Assembly: Consider functions, function parameters and return variables for shadowing checks.
+ * ``error`` is now a keyword that can only be used for defining errors.
```
ekpyron: That's a rebase artifact, isn't it?

Marenz: oh. Yes, I must have missed that

Changelog.md
```diff
@@ -1,6 +1,8 @@
 ### 0.9.0 (unreleased)
 
 Breaking changes:
+ * Disallow ``.pop()`` on arrays containing nested mappings.
+ * Disallow ``delete`` on types that contain nested mappings.
```
ekpyron, line 5: Is this in alphabetic order :-D? I'm honestly not sure :-D. [no need to do anything as far as I'm concerned]

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1594,11 +1594,18 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 		requireLValue(_operation.subExpression(), false);
 	else
 		_operation.subExpression().accept(*this);
+
 	Type const* subExprType = type(_operation.subExpression());
-	Type const* t = type(_operation.subExpression())->unaryOperatorResult(op);
-	if (!t)
+	TypeResult result = subExprType->unaryOperatorResult(op);
+	Type const* t = result;
```
ekpyron, line 1600: Can't you just remove this and replace every ``t`` below with ``result``? If not: also fine.

ekpyron (approved):
Looks fine to me.

Changelog.md
Marenz: It is. `.` before `d`, also originally done by vim using `:sort` (that's why we had that rebase artefact, it reordered another change, too)

libsolidity/analysis/TypeChecker.cpp
Marenz: I remember I actually tried that and that there was a problem with that. Something with `TypeResult` vs `Type`..

Marenz: Ah yes. We're assigning to `t` in line 1615 and that's why. Assigning to a `TypeResult` is at the very least confusing here

ekpyron: In my mind assigning to ``TypeResult`` is not that big of a deal (although not sure if it just works and jumping through hoops for being able to would definitely be too much). I'm fine with keeping the ``t``.

PR #11844, opened 2021-08-25
find_package z3 4.8.12 in CMake

soroosh-sdi:
we use z3 4.8.12, so I update the CMakefile. 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 include(EthOptions)
 configure_project(TESTS)
 
-find_package(Z3 4.8.0)
+find_package(Z3 4.8.12)
 if(${USE_Z3_DLOPEN})
   add_definitions(-DHAVE_Z3)
   add_definitions(-DHAVE_Z3_DLOPEN)


soroosh-sdi:
> leo @leonardoalt:matrix.org [m] Aug 21 13:42
Yea, unfortunately the ubuntu version is way too old. The requirement for our tests to pass is always the latest release, which currently is 4.8.12.
If you want you can install libz3-static-dev from our deps PPA: https://launchpad.net/~ethereum/+archive/ubuntu/cpp-build-deps

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 include(EthOptions)
 configure_project(TESTS)
 
-find_package(Z3 4.8.0)
+find_package(Z3 4.8.12)
```
cameel, line 67: I don't think we can do this. Z3 is a bleeding-edge version and might not be available everywhere. We do support earlier versions and some people are using them. It's just that some results are slightly different (but still valid) between Z3 versions. Instead of creating an elaborate test harness that could handle all of them, we assume that tests are running with the latest Z3. So 4.8.12 is the requirement only if you absolutely need SMT tests to pass.

soroosh-sdi: just wanted to make it more clear for other newbie who can't bear a test failure .
So I close this PR

cameel: Well, thanks for trying anyway :) It is a bit of a problem indeed, though the solution is unfortunately not that simple :)

I wonder if we could do something different. @leonardoalt what do you think about one of these?
- Make `soltest` and `isoltest` print a warning when linked with a Z3 version other than the latest one and `--no-smt` option is not used. This would require us to store the latest version in some build-time constant and update it every time we update Z3, but we already had people asking what that version is so making that explicit would probably be a good thing anyway.
- Keep requiring >= 4.8.0 for `solc` but for `soltest` and `isoltest` bump the requirement to 4.8.12. This way anyone running tests would have to either have latest Z3 or no Z3 at all. If an older version is installed one can always build with `-DUSE_Z3=OFF`. Here I wonder if CMake can actually handle having different version requirements for different executables.

leonardoalt: If it's possible with CMake, then I think the second option is cleaner.

cameel: ok. I'll create an issue then.

cameel: #11845

PR #11688, opened 2021-07-24
Fix and document `--allow-paths`

cameel:
Fixes #4623.
Fixes #10059.
~Depends on #11686 (draft until that one's merged).~ Merged.

Cases that this PR fixes:
- Neither allowed paths nor imports were being normalized before being compared. This broke:
    - Even very simple cases like whitelisting `../dir/` or `./dir/`.
    - Non-normalized paths like `dir1/../dir2/` `dir1/../dir2/`, `dir1///dir2/`.
    - Paths going beyond root like `/../../dir/`.
    - Symbolic links inside allowed paths were not resolved (only the ones in the imports were).
    - Whitelisted relative paths could only match imports using relative paths as well. Same for absolute paths.
    - All of this applied also to implicit whitelisting of directories due to being remapping targets or base path.
- Base path was not being automatically whitelisted.
- Empty paths would just whitelist the whole filesystem. There were several ways they could appear:
    - Given explicitly (e.g. `--allow-paths=a/,,b/` or `--allow-paths ""`). These are now ignored.
    - Resulting from a remapping with an empty target (e.g. `x=`). These are now ignored. After #11685 they will not be possible at all.
    - Resulting from a remapping to a file located at the VFS root (`x=contract.sol`). These now whitelist the working dir.
- When a remapping target ended with `/..`, wrong directory was whitelisted (e.g. for `x=/a/b/..` the whitelisted dir was `/a/b/`, not `/a/`).
    - Note that this is just a special case. In cases like `x=/a/b/c.sol` the right directory is actually `/a/b/`.

Issue #11409, opened 2021-05-19
Search path for libraries (path spec v3)


cameel:
Depends on #11408.
Related to #9353.
Part of a set of issues that replaces #11138.

## Abstract
Add `--include-paths` option which specifies a list of directories from which files can be loaded using relative paths (in addition to `--base-path`). This is meant to be similar to `$PATH` variable in the shell or `$PYTHONPATH` in Python.

The new feature makes it possible to reliably use relative paths in imports without relying on remappings. This is the first step towards disallowing absolute import paths, though for now it will still be possible to use an empty path as `--base-path` (in fact it remains the default) to keep using them.

## Motivation
Currently the compiler provides the `--base-path` option, which is a way to specify how import paths should be resolved by the default file loader. By default base path is empty, which means that relative paths are relative to the current working directory while absolute paths remain absolute. This makes imports change their meaning depending on the directory from which the compiler is invoked. Specifying a custom base path, on the other hand, results in both relative and absolute paths being interpreted as relative to the base directory. This prevents the file loader from accessing files outside of that directory and thus breaks the common usage pattern where libraries are installed using a package manager and can reside in different locations depending on how they were installed (e.g. local vs global `node_nodules/`).

From what I have seen, frameworks generally ignore `--base-path` and have their own mechanisms for dealing with importing libraries: they either don't rely on the default file loader at all (Truffle, Hardhat) or use import remappings to redirect relative paths to the right directories (Brownie, dapp.tools). The goal of this proposal is to extend the base path mechanism to be actually usable in direct CLI usage and optionally also for those frameworks that do rely on the default file loader. It might may also make it easier to use LSP with projects that rely on frameworks that do path resolution on their own (https://github.com/ethereum/solidity/issues/7763#issuecomment-832008294).

## Specification
1. Add a command-line option called `--include-path`.
    - The value is a single directory path. The option can be used multiple times to specify multiple directories.
    - The path must not be empty and must point at an existing directory.
    - The path is made canonical according to the same rules as `--base-path` (see #11408).
    - The option is allowed in combination with `--standard-json`, just like `--base-path` currently is.
2. When the default file loader receives a source unit name to load, it first tries to prepend base path to it and see if a file exists under that path. Then it does the same with all directories passed to `--include-path` in the order they were specified.
    - The loader returns the content of the first matching file.
    - If there is more than one matching file, the compiler prints a warning.
[...]

Issue #11413, opened 2021-05-19
Better path-related error messages (path spec v3)


cameel:
Related to #11042.
Part of a set of issues that replaces #11138.

## Abstract
Provide clearer error messages when the compiler fails to load a file. The current ones can be cryptic and hard to make sense of without knowing how the compiler internally handles paths and imports.

## Motivation
The best example of a confusing error message is probably `Cannot import url ("test.sol"): File import callback not supported`. The message simply means that the compiler did not find the file in its virtual filesystem and that it could not invoke the file loader callback to look for it because one was not provided. To the user it suggests that there's a problem with the callback rather than with the file and also the user usually is not even aware what callback it's referring to. There are many instances of this problem in Truffle's and Hardhat's bug trackers: https://github.com/nomiclabs/hardhat/issues/916, https://github.com/nomiclabs/hardhat/issues/689, https://github.com/trufflesuite/truffle/issues/3579, https://github.com/trufflesuite/truffle/issues/593, https://github.com/trufflesuite/truffle/issues/402, https://github.com/ethereum/solc-js/issues/522

Another one is the error you get when the file is not in `--allowed-paths`: `Source "test.sol" not found: File outside of allowed directories.`. It suggests that the file does not exist which may or may not be true.

## Specification
- Reword the "callback not supported" message so that it's clear that the cause of the problem is a missing file. Only mention the missing callback as an extra detail.
- When the path is outside of the allowed directories, say just that. Make it clear that the actual filesystem has not really been checked.
- In general when a file is missing, make it clear where the compiler was looking for it (only in the VFS or also on disk; if on disk then under which paths).
- When the file loader fails to find a file and the path it received looks like an URL (starts with a protocol name followed by `://`), say that the CLI interface does not support downloading files from the internet.

Those are just the ones I found so far. There may be more cases that can be improved.

## Backwards Compatibility
Fully backwards-compatible. Only wording of the error messages changes.

cameel:
We decided to put off work on this for now. We want to deal with #11408, #11409 and #11410 first and go back to this once that's done and working.

cameel:
Actually, there's nothing to discuss in this one - it's just a straightforward cleanup of error messages (@chriseth even said it looked good earlier on gitter) so I think I'll keep this one in the implementation backlog.

PR #11848, opened 2021-08-26
`--include-path` option

cameel:
Implements the `solc` part of #11409.
`solc-js` part is in https://github.com/ethereum/solc-js/pull/544.
Depends on #11688. ~In draft mode until that PR is merged.~

### DIFF
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -339,6 +339,15 @@ of the compiler.
 
     Include paths cannot have empty values and must be used together with a non-empty base path.
 
+.. note::
+
+    Include paths and base path can overlap as long as it does not make import resolution ambiguous.
+    For example, you can specify a directory inside base path as an include directory or have an
+    include directory that is a subdirectory of another include directory.
+    The compiler will only issue an error if the source unit name passed to the Host Filesystem
+    Loader represents an existing path when combined with multiple include paths or an include path
+    and base path.
+
 .. _cli-path-normalization-and-stripping:
 
 CLI Path Normalization and Stripping
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -394,6 +394,16 @@ The resulting file path becomes the source unit name.
 
 .. note::
 
+    The relative path produced by stripping must remain unique within the base path and include paths.
+    For example the compiler will issue an error for the following command if both
+    ``/project/contract.sol`` and ``/lib/contract.sol`` exist:
+
+    .. code-block:: bash
+
+        solc /project/contract.sol --base-path /project --include-path /lib
+
+.. note::
+
     Prior to version 0.8.8, CLI path stripping was not performed and the only normalization applied
     was the conversion of path separators.
     When working with older versions of the compiler it is recommended to invoke the compiler from
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -72,8 +72,9 @@ The initial content of the VFS depends on how you invoke the compiler:
        solc contract.sol /usr/local/dapp-bin/token.sol
 
    The source unit name of a file loaded this way is constructed by converting its path to a
-   canonical form and making it relative to the base path if it is located inside.
-   See :ref:`Base Path Normalization and Stripping <base-path-normalization-and-stripping>` for
+   canonical form and, if possible, making it relative to either the base path or one of the
+   include paths.
+   See :ref:`CLI Path Normalization and Stripping <cli-path-normalization-and-stripping>` for
    a detailed description of this process.
 
    .. index:: standard JSON
@@ -295,16 +296,36 @@ Here are some examples of what you can expect if they are not:
 
     The use of relative imports containing leading ``..`` segments is not recommended.
     The same effect can be achieved in a more reliable way by using direct imports with
-    :ref:`base path <base-path>` and :ref:`import remapping <import-remapping>`.
+    :ref:`base path and include paths <base-and-include-paths>`.
 
-.. index:: ! base path, ! --base-path
-.. _base-path:
+.. index:: ! base path, ! --base-path, ! include paths, ! --include-path
+.. _base-and-include-paths:
 
-Base Path
-=========
+Base Path and Include Paths
+===========================
 
-The base path specifies the directory that the Host Filesystem Loader will load files from.
-It is simply prepended to a source unit name before the filesystem lookup is performed.
+The base path and include paths represent directories that the Host Filesystem Loader will load files from.
+When a source unit name is passed to the loader, it prepends the base path to it and performs a
+filesystem lookup.
+If the lookup does not succeed, the same is done with all directories on the include path list.
+
+It is recommended to set the base path to the root directory of your project and use include paths to
+specify additional locations that may contain libraries your project depends on.
+This lets you import from these libraries in a uniform way, no matter where they are located in the
+filesystem relative to your project.
+For example, if you use npm to install packages and your contract imports
+``@openzeppelin/contracts/utils/Strings.sol``, you can use these options to tell the compiler that
+the library can be found in one of the npm package directories:
+
+.. code-block:: bash
+
+    solc contract.sol \
+        --base-path . \
+        --include-path node_modules/ \
+        --include-path /usr/local/lib/node_modules/
+
+Your contract will compile (with the same exact metadata) no matter whether you install the library
+in the local or global package directory or even directly under your project root.
 
 By default the base path is empty, which leaves the source unit name unchanged.
 When the source unit name is a relative path, this results in the file being looked up in the
@@ -314,10 +335,14 @@ interpreted as absolute paths on disk.
 If the base path itself is relative, it is interpreted as relative to the current working directory
 of the compiler.
 
-.. _base-path-normalization-and-stripping:
+.. note::
 
-Base Path Normalization and Stripping
--------------------------------------
+    Include paths cannot have empty values and must be used together with a non-empty base path.
+
+.. _cli-path-normalization-and-stripping:
+
+CLI Path Normalization and Stripping
+------------------------------------
 
 On the command line the compiler behaves just as you would expect from any other program:
 it accepts paths in a format native to the platform and relative paths are relative to the current
@@ -326,7 +351,7 @@ The source unit names assigned to files whose paths are specified on the command
 should not change just because the project is being compiled on a different platform or because the
 compiler happens to have been invoked from a different directory.
 To achieve this, paths to source files coming from the command line must be converted to a canonical
-form, and, if possible, made relative to the base path.
+form, and, if possible, made relative to the base path or one of the include paths.
 
 The normalization rules are as follows:
 
@@ -357,7 +382,8 @@ The normalization rules are as follows:
     You can avoid such situations by ensuring that all the files are available within a single
     directory tree on the same drive.
 
-Once canonicalized, the base path is stripped from all source file paths that start with it.
+After normalization the compiler attempts to make the source file path relative.
+It tries the base path first and then the include paths in the order they were given.
 If the base path is empty or not specified, it is treated as if it was equal to the path to the
 current working directory (with all symbolic links resolved).
 The result is accepted only if the normalized directory path is the exact prefix of the normalized
@@ -388,11 +414,11 @@ locations that are considered safe by default:
   - The directories used as :ref:`remapping <import-remapping>` targets.
     If the target is not a directory (i.e does not end with ``/``, ``/.`` or ``/..``) the directory
     containing the target is used instead.
-  - Base path.
+  - Base path and include paths.
 
 - In Standard JSON mode:
 
-  - Base path.
+  - Base path and include paths.
 
 Additional directories can be whitelisted using the ``--allow-paths`` option.
 The option accepts a comma-separated list of paths:
@@ -403,6 +429,7 @@ The option accepts a comma-separated list of paths:
     solc token/contract.sol \
         lib/util.sol=libs/util.sol \
         --base-path=token/ \
+        --include-path=/lib/ \
         --allow-paths=../utils/,/usr/libraries
 
 When the compiler is invoked with the command shown above, the Host Filesystem Loader will allow
@@ -410,6 +437,7 @@ importing files from the following directories:
 
 - ``/home/user/project/token/`` (because ``token/`` contains the input file and also because it is
   the base path),
+- ``/lib/`` (because ``/lib/`` is one of the include paths),
 - ``/home/user/project/libs/`` (because ``libs/`` is a directory containing a remapping target),
 - ``/home/user/utils/`` (because of ``../utils/`` passed to ``--allow-paths``),
 - ``/usr/libraries/`` (because of ``/usr/libraries`` passed to ``--allow-paths``),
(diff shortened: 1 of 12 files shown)


docs/path-resolution.rst
```diff
@@ -314,10 +335,23 @@ interpreted as absolute paths on disk.
 If the base path itself is relative, it is interpreted as relative to the current working directory
 of the compiler.
 
-.. _base-path-normalization-and-stripping:
+.. note::
+
+    Include paths cannot have empty values and must be used together with a non-empty base path.
```
cameel, line 340: Since the default for `--base-path` is an empty value, this basically means that you cannot use `--include-paths` without also specifying `--base-path`.

Empty base path is a valid prefix of any import so it "neutralizes" the include paths (they will never match). I might be able to come up with some magic workaround and treat empty base path specially, but I think we should just switch to a more sensible default for base path.

libsolidity/interface/FileReader.cpp
```diff
@@ -104,14 +150,12 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
-
-		if (!boost::filesystem::is_regular_file(canonicalPath))
+		if (!boost::filesystem::is_regular_file(candidates[0]))
 			return ReadCallback::Result{false, "Not a valid file."};
 
 		// NOTE: we ignore the FileNotFound exception as we manually check above
-		auto contents = readFileAsString(canonicalPath);
+		auto contents = readFileAsString(candidates[0]);
+		solAssert(m_sourceCodes.count(_sourceUnitName) == 0, "");
 		m_sourceCodes[_sourceUnitName] = contents;
```
cameel, line 164: Storing the file content loaded by the import callback in `FileReader.m_sourceCodes` is pointless - it's never used because the compiler just uses the return value of the callback.

IIRC @christianparpart wanted it for the LSP, though that's a flawed solution (files included in Standard JSON never get here).

libsolidity/interface/FileReader.cpp
```diff
@@ -87,15 +107,41 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
+			return ReadCallback::Result{false, "File not found."};
+
+		if (candidates.size() >= 2)
+			return ReadCallback::Result{
+				false,
+				"Ambiguous import. "
+				"Multiple matching files found inside base path and/or include paths: " +
+				joinHumanReadable(candidates | ranges::views::transform(pathToQuotedString), ", ") +
+				"."
+			};
```
cameel, line 142: This differs from what we decided in the issue (https://github.com/ethereum/solidity/issues/11409#issuecomment-849764873). The callback does not provide any way to report a warning. And printing it to stderr is not a good solution because that won't mesh well with Standard JSON (the warning should be included in the JSON).

For this reason it's an error rather than a warning.

solc/CommandLineParser.cpp
```diff
@@ -503,6 +505,15 @@ General Information)").c_str(),
 			po::value<string>()->value_name("path"),
 			"Use the given path as the root of the source tree instead of the root of the filesystem."
 		)
+		(
+			g_strIncludePath.c_str(),
+			po::value<vector<string>>()->value_name("path"),
```
cameel, line 531: To specify multiple values for `--include-path` you just use it multiple times, while in `--allow-paths` you specify paths separated with commas. This is a bit inconsistent but it was explicitly requested in https://github.com/ethereum/solidity/issues/11409#issuecomment-844891942.

Maybe we should make `--allow-paths` work that way too? Or I could leave it as is for backwards-compatibility but also add `--allow-path` that could be used multiple times like `--include-path`.

solc/CommandLineInterface.cpp
```diff
@@ -398,9 +398,30 @@ bool CommandLineInterface::readInputFiles()
 		}
 	}
 
+	for (boost::filesystem::path const& includePath: m_options.input.includePaths)
+		m_fileReader.addIncludePath(includePath);
```
cameel, line 402: Note that we process base path and include paths in all modes, not just compiler and standard json. The `.yul` files passed to the assembler, `.bin` files passed to the linker and even the file passed to `--import-ast` are affected by these options as well. The only visible consequence of that is that for example if you do 
```bash
solc --base-path test/ --strict-assembly test/input.yul
```
you will see `input.yul` rather than `test/input.yul` in the error messages.

Since it's harmless I did not bother forbidding that. It would require hacks in `FileReader` because all input files (not just Solidity files) currently go through it.

test/solc/CommandLineInterface.cpp
```diff
@@ -868,6 +870,421 @@ BOOST_AUTO_TEST_CASE(cli_paths_to_source_unit_names_base_path_and_stdin)
...
+		"solc",
+		"--no-color",
+		"--base-path=base/",
+		"--include-path=include/",
+		"-",
+	};
+
+	string expectedMessage =
+		"Error: Source \"contract.sol\" not found: Ambiguous import. "
+		"Multiple matching files found inside base path and/or include paths: \"" +
```
cameel, line 1276: The first part of the message seems to be generic and prepended to any error from the callback. It does not fit this situation well (files were found, actually too many of them). Perhaps something to change in #11413.

docs/path-resolution.rst
leonardoalt: Right, that makes sense.

solc/CommandLineParser.cpp
leonardoalt: Yea I agree what allowing `--allow-path` multiple times too would be nice.

test/solc/CommandLineInterface.cpp
leonardoalt: I guess in this case it will mean "I didn't find the one you actually meant" :p

docs/path-resolution.rst
```diff
@@ -295,16 +296,36 @@ Here are some examples of what you can expect if they are not:
...
-The base path specifies the directory that the Host Filesystem Loader will load files from.
-It is simply prepended to a source unit name before the filesystem lookup is performed.
+The base path and include paths represent directories that the Host Filesystem Loader will load files from.
+When a source unit name is passed to the loader, it prepends the base path to it and performs a
+filesystem lookup.
+If the lookup does not succeed, the same is done with all directories on the include path list.
+
+It is recommended to set the base path to the root directory of your project and use include paths to
+specify additional locations that may contain libraries your project depends on.
+This let you import from these libraries in a uniform way, no matter where they are located in the
```
leonardoalt:
```suggestion
This lets you import from these libraries in a uniform way, no matter where they are located in the
```

libsolidity/interface/FileReader.h
```diff
@@ -125,6 +143,9 @@ class FileReader
 	/// Base path, used for resolving relative paths in imports.
 	boost::filesystem::path m_basePath;
 
+	/// Additional directories used fror resolving relative paths in imports.
```
leonardoalt:
```suggestion
	/// Additional directories used for resolving relative paths in imports.
```

libsolidity/interface/FileReader.cpp
```diff
@@ -21,33 +21,56 @@
...
+#include <range/v3/view/transform.hpp>
 
 using solidity::frontend::ReadCallback;
 using solidity::langutil::InternalCompilerError;
 using solidity::util::errinfo_comment;
 using solidity::util::readFileAsString;
+using solidity::util::joinHumanReadable;
+using std::map;
 using std::string;
+using std::vector;
```
leonardoalt, line 41: These 3 seem a bit non conventional for our code base. Why are they used like this?

libsolidity/interface/FileReader.cpp
```diff
@@ -21,33 +21,56 @@
...
 		solAssert(!allowedDir.empty(), "");
 }
 
 void FileReader::setBasePath(boost::filesystem::path const& _path)
 {
-	m_basePath = (_path.empty() ? "" : normalizeCLIPathForVFS(_path));
+	if (_path.empty())
+	{
+		// Empty base path is a special case that does not make sense when include paths are used.
+		solAssert(m_includePaths.empty(), "");
```
leonardoalt, line 67: If a user tries that, does it end up here? Or is there a more graceful warning somewhere?

libsolidity/interface/FileReader.cpp
```diff
@@ -21,33 +21,56 @@
...
+		m_basePath = "";
+	}
+	else
+		m_basePath = normalizeCLIPathForVFS(_path);
+}
+
+void FileReader::addIncludePath(boost::filesystem::path _path)
```
leonardoalt: Why not `const&`?

libsolidity/interface/FileReader.cpp
```diff
@@ -87,15 +107,41 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
+
+		for (boost::filesystem::path const& prefix: prefixes)
+		{
+			boost::filesystem::path canonicalPath = normalizeCLIPathForVFS(
+				prefix / strippedSourceUnitName,
+				true /* resolveSymlinks */
+			);
+
+			if (boost::filesystem::exists(canonicalPath))
+				candidates.push_back(canonicalPath);
```
leonardoalt: Does this create a copy?

libsolidity/interface/FileReader.cpp
```diff
@@ -87,15 +107,41 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
+		}
+
+		auto pathToQuotedString = [](boost::filesystem::path const& _path){ return "\"" + _path.string() + "\""; };
+
+		if (candidates.size() == 0)
```
leonardoalt:
```suggestion
		if (candidates.empty())
```

solc/CommandLineParser.cpp
```diff
@@ -909,6 +920,25 @@ General Information)").c_str(),
...
+		}
+
+		for (string const& includePath: m_args[g_strIncludePath].as<vector<string>>())
+		{
+			if (includePath.empty())
+			{
+				serr() << "Empty values are not allowed in --" << g_strIncludePath << "." << endl;
+				return false;
+			}
+			m_options.input.includePaths.push_back(includePath);
```
leonardoalt, line 938: Could `includePath` be moved here?

test/solc/CommandLineInterfaceAllowPaths.cpp
```diff
@@ -508,6 +514,37 @@ BOOST_FIXTURE_TEST_CASE(allow_path_automatic_whitelisting_work_dir, AllowPathsFi
 	BOOST_TEST(checkImport("import 'a/X/c.sol'", {"--base-path", ""}));
 }
 
+BOOST_FIXTURE_TEST_CASE(allow_path_automatic_whitelisting_include_paths, AllowPathsFixture)
+{
+	// Relative include path whitelists its content
+	BOOST_TEST(checkImport("import 'b/c.sol'", {"--base-path=a/b/c", "--include-path=../code/a"}));
```
leonardoalt, line 520: I don't really understand what these are testing.

solc/CommandLineParser.cpp
cameel: OK, I'll add it.

libsolidity/interface/FileReader.cpp
cameel: This the style preferred by @christianparpart and, personally, I like it a lot too. But yeah, it's not the style we use. I was going to refactor it into our usual style, but I'm still holding back with a lot of such refactors until this pile of path PRs goes through. Dependencies between PRs are not fun :)

libsolidity/interface/FileReader.cpp
cameel: There's a validation against this in `CommandLineParser`.

solc/CommandLineParser.cpp
cameel: Probably not. I think that `.as()` returns a reference to a string stored by `program_options`. But I need to check to say for sure.

test/solc/CommandLineInterfaceAllowPaths.cpp
cameel: Overall these tests check if this particular source file (the import is the only thing in it) compiled with these specific command-line options compiles fine or if it reports some kind of error (path not allowed or file missing). It's just wrapped in a helper function to allow me to test a lot of such combinations. There are tons of weird corner cases related to allowed paths because the feature interacts with all the other path-related options.

You might wan to look at `AllowPathsFixture` to see a list of files which are automatically created by test setup. Basically, there's a `work/` directory that is the current working directory when the compilation is executed and `code/`, which is specifically meant to be outside of the work dir. Both contain `a/b/c.sol`.

This particular test adds `code/a/` as an include path and `work/a/b/c/` as the base path. Then it checks importing `b/c.sol` will make the compiler find `code/a/b/c.sol` instead of complaining that the path is disallowed or that the file does not exist.

solc/CommandLineParser.cpp
leonardoalt: Ah I didn't mean to do that in this PR, can also be after. I think it's fine without it, to not make this PR touch too much stuff.

libsolidity/interface/FileReader.cpp
leonardoalt: Sure, that's fine.

libsolidity/interface/FileReader.cpp
leonardoalt: Yea, found it later too.

solc/CommandLineParser.cpp
leonardoalt: ok!

leonardoalt:
@cameel let me know when I should take another look

libsolidity/interface/FileReader.cpp
cameel: Probably. It's a local variable but compiler might not move it automatically like it does with return values. Adding `move` won't hurt.

I tried to also make `candidates` a vector of references but I can't because `canonicalPath` is a new object, created based on the original path and prefix. But I managed to replace `prefixes` with a vector of references instead.

solc/CommandLineParser.cpp
cameel: It does return a reference so I left it as is.

cameel:
Thanks @leonardoalt. Fixes applies. Ready for another round,

libsolidity/interface/FileReader.cpp
```diff
@@ -129,6 +178,41 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
+	map<string, FileReader::FileSystemPathSet> nameToPaths;
+	for (boost::filesystem::path const& cliPath: _cliPaths)
+	{
+		string sourceUnitName = cliPathToSourceUnitName(cliPath);
+		boost::filesystem::path normalizedPath = normalizeCLIPathForVFS(cliPath);
+		nameToPaths[sourceUnitName].insert(normalizedPath);
+	}
+
+	map<string, FileReader::FileSystemPathSet> collisions;
+	for (auto const& [sourceUnitName, cliPaths]: nameToPaths)
```
leonardoalt:
```suggestion
	for (auto&& [sourceUnitName, cliPaths]: nameToPaths)
```
Wouldn't this have to be like this to move `cliPaths` below?

cameel: Right. Thanks for spotting this! Fixed.

leonardoalt (approved):

axic:
> Depends on #11688. ~In draft mode until that PR is merged.~~

Description says this, yet it was merged? :)

cameel:
Yeah :) It got merged into the base branch rather than into `develop`.
That's why I had to create #12007.

Issue #11597, opened 2021-06-29
[SMTChecker] Improve usability


leonardoalt:
The SMTChecker's usability is still not great, especially for new users.
When running the SMTChecker on large contracts with the default settings, users are often bombarded with warnings about targets the solver couldn't prove, and verifying all contracts at the same time makes the encoding very large.
A few items that could be improved:

- [x] There is a lot of noise about targets that the solver could not prove. One idea is to compress all of those into a single warning by default, saying `128 verification conditions could not be proved. Consider choosing a single contract to be verified. To see the full list of unproved targets, enable model checker verbosity to 2` for example. This would also highlight problems that are actually confirmed.
- [x] Underflow/overflow verification conditions create A LOT of targets which really bloats the solver. IMO, these should not be checked by default for Solidity >=0.8.0 since most people do rely on the compiler's built-in checks. Conclusion, I think we should not check for underflow/overflow by default for Solidity >=0.8.0.
- [ ] Choosing a single contract to be verified increases performance a lot. Somehow deciding on a single contract (or a few) by default could also improve usability a lot.
- [x] Filter unsupported warnings by target contract reachability https://github.com/ethereum/solidity/issues/11703

dtopia-me:
Hey there, not sure if this is the right place to raise this up, but throughout the whole Solidity document, neither definition, explanation nor the full name of SMTChecker is given anywhere. As you mentioned, not friendly for new users. It would be great if some background of this module or tool can be explained a bit further. Thanks.

cameel:
@DeMeForYou It's there: [SMTChecker and Formal Verification](https://docs.soliditylang.org/en/latest/smtchecker.html). It has been extended and made a top-level section only recently though so make sure you're reading the latest docs. Before then it was buried inside [Security Considerations](https://docs.soliditylang.org/en/latest/security-considerations.html) section.

dtopia-me:
@cameel Thanks for this. What I meant was the fact that SMT stands for `Satisfiability Modulo Theories` is completely hidden in the current doc (I checked the entire doc and did not find what it stands for). New users and those who are not familiar with the term can google it for more details if the full name can be given.

cameel:
There's a link to a Wikipedia article about SMT:

[...]

PR #11850, opened 2021-08-27
[SMTChecker] Remove redundant unsupported warnings

leonardoalt:
Ref https://github.com/ethereum/solidity/issues/11597

The SMTChecker generates a lot of redundant "unsupported" warnings, because it visits every contract separately in its own context, with 2 different engines. So for example if a contract that is inherited by 2 contracts has some assembly in it, we're going to see the message "inline assembly unsupported" possibly 6 times: each time a contract is visited, by each engine.
This PR removes all the redundant messages.

In the no solver case (2), for example when only using the smtlib2 solver, it's now fine to also report the unproved stuff, because there will be only one message bundling all the unproved targets anyway, and all targets will be unproved.


libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
+#include <liblangutil/ErrorReporter.h>
```
cameel: License boilerplate and `#pragma once` are missing in this file.

libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
...
+			errorReporter.warning(_error, _location, _description);
+		set(_error, _location, _description);
+	}
+
+	void warning(
+			langutil::ErrorId _error,
+			langutil::SourceLocation const& _location,
+			std::string const& _description,
+			langutil::SecondarySourceLocation const& _secondaryLocation
+			)
```
cameel:
```suggestion
	void warning(
		langutil::ErrorId _error,
		langutil::SourceLocation const& _location,
		std::string const& _description,
		langutil::SecondarySourceLocation const& _secondaryLocation
	)
```

libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
+#include <liblangutil/ErrorReporter.h>
+
+namespace solidity::frontend
+{
+
+struct UniqueErrorReporter
```
cameel: This looks like a generic utility not tied specifically to SMTChecker. Why not move it to `langutil`?

libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
...
+		seenErrors.insert({_error, _location, _description});
+	}
+
+	langutil::ErrorList const& errors() const { return errorReporter.errors(); }
+
+	void clear() { errorReporter.clear(); }
+
+	langutil::ErrorReporter errorReporter;
+	langutil::ErrorList uniqueErrors;
+	std::set<std::tuple<langutil::ErrorId, langutil::SourceLocation, std::string>> seenErrors;
```
cameel: Why not a class? This stuff looks like it should be private.

libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
...
+			errorReporter.warning(_error, _description);
+		set(_error, {}, _description);
+	}
+
+	bool has(langutil::ErrorId _error, langutil::SourceLocation const& _location, std::string const& _description)
+	{
+		return seenErrors.count({_error, _location, _description});
+	}
+
+	void set(langutil::ErrorId _error, langutil::SourceLocation const& _location, std::string const& _description)
```
cameel: I think that the name `set` does not make it clear enough whether it also reports the error or not.
```suggestion
	void markAsSeen(langutil::ErrorId _error, langutil::SourceLocation const& _location, std::string const& _description)
```
I'd also rename `has()` to `seen()`.

libsolidity/formal/UniqueErrorReporter.h
```diff
@@ -0,0 +1,55 @@
...
+	void warning(langutil::ErrorId _error, std::string const& _description)
+	{
+		if (!has(_error, {}, _description))
+			errorReporter.warning(_error, _description);
+		set(_error, {}, _description);
+	}
+
+	bool has(langutil::ErrorId _error, langutil::SourceLocation const& _location, std::string const& _description)
+	{
+		return seenErrors.count({_error, _location, _description});
```
cameel: Should the `_description` actually be a part of the key? I'd expect the code and location to uniquely identify an error.

libsolidity/formal/ModelChecker.h
```diff
@@ -75,6 +75,8 @@ class ModelChecker
 private:
 	langutil::ErrorReporter& m_errorReporter;
 
+	UniqueErrorReporter m_uniqueErrorReporter;
```
cameel: Why not just use `m_uniqueErrorReporter` for everything? If there's a good reason, I think this needs a comment explaining which reporter should be used in which situation.

test/libsolidity/smtCheckerTests/imports/imported_fail_3.sol
```diff
@@ -23,4 +23,5 @@ contract C is A {
 // SMTEngine: all
 // ----
 // Warning 6328: (B.sol:71-85): CHC: Assertion violation happens here.\nCounterexample:\nx = 0\n_x = 0\n\nTransaction trace:\nB.constructor()\nState: x = 0\nB.g(0)
+// Warning 6328: (B.sol:71-85): CHC: Assertion violation happens here.\nCounterexample:\nx = 0\n_x = 0\n\nTransaction trace:\nB.constructor()\nState: x = 0\nB.g(0)
```
cameel: What's wrong here? It actually added a duplicate error where there was no duplicate before.

test/cmdlineTests/model_checker_solvers_smtlib2/err
```diff
@@ -1,3 +1,7 @@
+Warning: CHC: 1 verification condition(s) could not be proved. Enable the model checker option "show unproved" to see all of them. Consider choosing a specific contract to be verified in order to reduce the solving problems. Consider increasing the timeout per query.
+
 Warning: CHC analysis was not possible. No Horn solver was available. None of the installed solvers was enabled.
 
+Warning: BMC: 1 verification condition(s) could not be proved. Enable the model checker option "show unproved" to see all of them. Consider choosing a specific contract to be verified in order to reduce the solving problems. Consider increasing the timeout per query.
+
 Warning: BMC analysis was not possible. No SMT solver (Z3 or CVC4) was available. None of the installed solvers was enabled.
```
cameel, line 7: I don't see anything obvious in the code that would be responsible for these messages showing up. Were they just going to the wrong error reporter and being discarded or something?

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: hehe

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: Not sure if other components would like to use it, so didn't want to bloat the library.

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: Yea started as struct and then kept it

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: You're right, I can't think of a counterexample to that right now, will check.

libsolidity/formal/ModelChecker.h
leonardoalt: Ah yes, I forgot to change a few in ModelChecker.cpp. But we still need `m_errorReporter`, cus that the ones that comes from `CompilerStack`. So we'd have only one use of `m_errorReporter` in `ModelChecker.cpp`, to append the ones from `m_uniqueErrorReporter`.

test/libsolidity/smtCheckerTests/imports/imported_fail_3.sol
leonardoalt: Yea, I noticed that but don't have an answer yet.

test/cmdlineTests/model_checker_solvers_smtlib2/err
leonardoalt: That's the change that I meant in the end of my first comment in the PR. These were ignored because no solver was available. This was introduced when we didn't bundle the unproved messages yet, so it was really full of messages, and at the end you'd have "no solver available", which felt a bit hidden. So back then we decided to ignore everything else and just issue "no solver available". But now since all the unproved are bundle anyway, I think it's fine to have cleaner code and issue these 2 extra messages.

test/libsolidity/smtCheckerTests/imports/imported_fail_3.sol
leonardoalt: I think what happened here is that the `UniqueErrorReporter` is used per source (and its dependencies), but here `CompilerStack` calls `ModelChecker` mutliple times, so I'd understand the duplicate being there. But I don't understand yet why it wasn't there in the first place.

leonardoalt: Ah ok, figured it out. CompilerStack calls ModelChecker::analyze multiple times, so we actually need to clear m_uniqueErrorReporter every time we append it to the original m_errorReporter, which makes sense.

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: So there are a few warnings without location:
- ModelChecker.cpp: 2 warnings with SourceLocation{}
- BMC.cpp: 2 warnings without location, 3 warnings with Sourcelocation{}
- CHC.cpp: 3 warnings with SourceLocation{}

All of these *might* be unique anyway, so could be fine to remove `description` for the key. I'll update the other comments and try this after.

libsolidity/formal/UniqueErrorReporter.h
cameel: In my opinion it's generic enough to be included there.

And if it's not in the library, it will be harder for others to realize that it exists when they do need it :)

libsolidity/formal/UniqueErrorReporter.h
cameel: But they should either have different codes or same description, right?

I think I'd add an assert that the description must be the same. This way we'll catch situations where someone breaks this assumption.

test/cmdlineTests/model_checker_solvers_smtlib2/err
cameel: But which specific change in the code is responsible for them being issued now?

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: I removed `description` from the key and nothing changed, so that's more evidence that the warnings mentioned above are indeed unique anyway, so we can keep it like this (without `description` in the key)

libsolidity/formal/BMC.cpp
```diff
@@ -124,10 +123,6 @@ void BMC::analyze(SourceUnit const& _source, map<ASTNode const*, set<Verificatio
 			);
 		}
 	}
-	else
-		m_outerErrorReporter.append(m_errorReporter.errors());
-
-	m_errorReporter.clear();
```
leonardoalt, line 130: This here @cameel /\\
(and same in CHC.cpp)

cameel: Oh. So it's this after all. Thanks!

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: > But they should either have different codes or same description, right?
> 
> I think I'd add an assert that the description must be the same. This way we'll catch situations where someone breaks this assumption.

That would require storing the description in the `seen` set though, which it's not doing at the moment.

liblangutil/UniqueErrorReporter.h
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	void warning(ErrorId _error, std::string const& _description)
+	{
+		if (!seen(_error, {}))
+			m_errorReporter.warning(_error, _description);
+		markAsSeen(_error, {});
+	}
+
+	bool seen(ErrorId _error, SourceLocation const& _location)
```
cameel:
```suggestion
	bool seen(ErrorId _error, SourceLocation const& _location) const
```

libsolidity/formal/ModelChecker.h
```diff
@@ -33,6 +33,8 @@
 #include <libsmtutil/SolverInterface.h>
 #include <liblangutil/ErrorReporter.h>
 
+#include <liblangutil/UniqueErrorReporter.h>
```
cameel:
```suggestion
#include <liblangutil/ErrorReporter.h>
#include <liblangutil/UniqueErrorReporter.h>
```

cameel (dismissed):
Two minor tweaks needed but overall looks good so approving already.

libsolidity/formal/UniqueErrorReporter.h
cameel: Ah, true. Noticed that while re-reviewing now. It's fine without an assert then.

leonardoalt:
@cameel updated last comments + added `description` back into the `map` but not as key, simply to assert that equal keys must have equal descriptions as suggested.

cameel (dismissed):

libsolidity/formal/UniqueErrorReporter.h
leonardoalt: Now it has both description and assert haha.

leonardoalt:
@cameel just rebased on top of your merged PRs, can you reapprove?

cameel (approved):

PR #11852, opened 2021-08-27
Fix ICE on multi-source use of abi.*

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/11587
Fixes https://github.com/ethereum/solidity/issues/11710
Depends on https://github.com/ethereum/solidity/pull/11850

Easier to review with ` Hide whitespace changes ` set

PR #11854, opened 2021-08-27
[SMTChecker] Fix false negative after `push` on storage references

### LINKED ISSUE #11027: [SMTChecker] Invalid analysis of `push` operation on lvalue expression.
In the following example:
```
pragma experimental SMTChecker;
contract C {
    bytes x;

    function getX() internal returns (bytes storage) {
        return x;
    }

    function s() public {
        require(x.length == 0);
        getX().push("a");
        assert(x.length == 1); // should hold
        assert(x.length == 0); // should fail
    }
}
```
SMTChecker incorrectly reports assertion violation for the first assert (false positive) and also incorrectly reports the second assertion as safe (false negative).

True both in v0.8.1 and current develop.

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/11027
Depends on https://github.com/ethereum/solidity/pull/11852

We need to erase knowledge about storage references when we push an array, because of aliasing.
This was already done when pushing variables directly, but wasn't done when the lvalue is a function return.

It's easier to review this PR with ` Hide whitespace changes ` set.

### DIFF
--- a/libsolidity/formal/SMTEncoder.cpp
+++ b/libsolidity/formal/SMTEncoder.cpp
@@ -1977,26 +1977,32 @@ void SMTEncoder::assignment(
 		indexOrMemberAssignment(*left, _right);
 	else if (auto const* funCall = dynamic_cast<FunctionCall const*>(left))
 	{
-		if (
-			auto funType = dynamic_cast<FunctionType const*>(funCall->expression().annotation().type);
-			funType && funType->kind() == FunctionType::Kind::ArrayPush
-		)
+		if (auto funType = dynamic_cast<FunctionType const*>(funCall->expression().annotation().type))
 		{
-			auto memberAccess = dynamic_cast<MemberAccess const*>(&funCall->expression());
-			solAssert(memberAccess, "");
-			auto symbArray = dynamic_pointer_cast<smt::SymbolicArrayVariable>(m_context.expression(memberAccess->expression()));
-			solAssert(symbArray, "");
+			if (funType->kind() == FunctionType::Kind::ArrayPush)
+			{
+				auto memberAccess = dynamic_cast<MemberAccess const*>(&funCall->expression());
+				solAssert(memberAccess, "");
+				auto symbArray = dynamic_pointer_cast<smt::SymbolicArrayVariable>(m_context.expression(memberAccess->expression()));
+				solAssert(symbArray, "");
 
-			auto oldLength = symbArray->length();
-			auto store = smtutil::Expression::store(
-				symbArray->elements(),
-				symbArray->length() - 1,
-				_right
-			);
-			symbArray->increaseIndex();
-			m_context.addAssertion(symbArray->elements() == store);
-			m_context.addAssertion(symbArray->length() == oldLength);
-			assignment(memberAccess->expression(), symbArray->currentValue());
+				auto oldLength = symbArray->length();
+				auto store = smtutil::Expression::store(
+					symbArray->elements(),
+					symbArray->length() - 1,
+					_right
+				);
+				symbArray->increaseIndex();
+				m_context.addAssertion(symbArray->elements() == store);
+				m_context.addAssertion(symbArray->length() == oldLength);
+				assignment(memberAccess->expression(), symbArray->currentValue());
+			}
+			else if (funType->kind() == FunctionType::Kind::Internal)
+			{
+				for (auto type: funType->returnParameterTypes())
+					if (type->category() == Type::Category::Mapping || dynamic_cast<ReferenceType const*>(type))
+						resetReferences(type);
+			}
 		}
 	}
 	else
--- a/test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
+++ b/test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
@@ -1,0 +1,20 @@
+contract C {
+	bytes x;
+
+	function getX() internal view returns (bytes storage) {
+		return x;
+	}
+
+	function s() public {
+		require(x.length == 0);
+		getX().push("a");
+		assert(x.length == 1); // should hold but knowledge is erased due to pushing a reference
+		assert(x.length == 0); // should fail
+	}
+}
+// ====
+// SMTEngine: all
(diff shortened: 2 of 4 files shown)


test/libsolidity/smtCheckerTests/array_members/storage_pointer_push_1.sol
```diff
@@ -17,4 +17,5 @@ contract C {
 // SMTEngine: all
 // SMTIgnoreCex: yes
 // ----
+// Warning 6368: (159-169): CHC: Out of bounds access happens here.
```
leonardoalt, line 20: This is a false positive but it's correct that it reports it, since we don't really know what `s()` returns at that point (storage reference).

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
```diff
@@ -0,0 +1,20 @@
+contract C {
+	bytes x;
```
cameel, line 2: I see that `resetReferences()` only looks at the type of the whole variable. What happens if the variable is a struct containing an array and the function returns a reference to that array? Or if it's an array of arrays and one of the nested arrays is returned? Or a mapping where the value type is an array? Mapping nested in a mapping? Can you add test cases covering these?

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1977,26 +1977,32 @@ void SMTEncoder::assignment(
...
+				symbArray->increaseIndex();
+				m_context.addAssertion(symbArray->elements() == store);
+				m_context.addAssertion(symbArray->length() == oldLength);
+				assignment(memberAccess->expression(), symbArray->currentValue());
+			}
+			else if (funType->kind() == FunctionType::Kind::Internal)
+			{
+				for (auto type: funType->returnParameterTypes())
+					if (type->category() == Type::Category::Mapping || dynamic_cast<ReferenceType const*>(type))
+						resetReferences(type);
```
cameel, line 2004: So this basically resets all the state variables of this type because we don't know which ones can actually be returned by the function?

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
```diff
@@ -0,0 +1,20 @@
...
+
+	function getX() internal view returns (bytes storage) {
+		return x;
+	}
+
+	function s() public {
+		require(x.length == 0);
+		getX().push("a");
+		assert(x.length == 1); // should hold but knowledge is erased due to pushing a reference
```
cameel, line 11: The reset is due to it being returned from a function rather than due to the push, right?
```suggestion
		assert(x.length == 1); // should hold but knowledge is erased due to returning a reference
```

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
```diff
@@ -0,0 +1,20 @@
+contract C {
+	bytes x;
+
+	function getX() internal view returns (bytes storage) {
+		return x;
+	}
```
cameel, line 6: What about functions taking `storage` arguments? I don't see any cases covering that in `smtCheckerTests/array_members/` (I haven't inspected all of them but none of those whose names refer to functions or arguments seem to have that).

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
leonardoalt: Nested between array/mappings is handled by `SMTEncoder::sameTypeOrSubtype`. Maybe structs should be included there too.

libsolidity/formal/SMTEncoder.cpp
leonardoalt: Not only of this type, but also of every type that can be a subtype of this type, as in nested arrays/mappings for example.

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
leonardoalt: Yea there are a bunch of tests with storage/memory arguments in `types`, for example `types/array_aliasing_storage_1.sol`

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
leonardoalt: Not really, the resetting is due to the push, since we don't know exactly what it is. If we don't assign/push the returned reference then nothing needs to be reset.

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
leonardoalt: Interestingly the example below already works as intended, not sure why.

```
contract C {
	struct S { uint x; uint[] a; }
	S s;
	uint[] b;

	function f(bool c) public {
		require(b.length == 0);
		require(s.a.length == 0);

		b.push();
		assert(b.length == 1); // should hold

		s.a.push();
		assert(s.a.length == 1); // should hold

		uint[] storage ref = s.a;

		ref.push();

		assert(b.length == 1); // should hold but fails due to aliasing
		assert(s.a.length == 2); // should hold but fails due to aliasing

	}
}
```

cameel: Isn't it because the reference is just a variable and not something returned by a function??

libsolidity/formal/SMTEncoder.cpp
cameel: Oh, so `resetReferences()` recursively visits the variables rather than just looping over them? That explains why https://github.com/ethereum/solidity/pull/11854#discussion_r698413463 is not a problem.

test/libsolidity/smtCheckerTests/array_members/push_function_return_value.sol
cameel: Ok, thought that the assign referred to assigning the returns inside the function. Now it makes more sense.

cameel (approved):

PR #11856, opened 2021-08-28
UBSan CI: Remove unsigned-shift-base sanitizer flag.

bshastry:
The newly added UBSan nightly build was failing because of a newly introduced sanitizer flag that is not available in clang docker image we use (clang v10.0).

This PR removes the said flag.


cameel (approved):
Ok, so now we know why `unsigned-shift-base` was not used originally :)

PR #11218, opened 2021-04-07
Change `error` from directive to keyword.

christianparpart:
Closes #11214.

Makes `error` a keyword, prohibiting constructs like mentioned in the above ticket.

## Checklist

- [x] Changelog entry

PR #11859, opened 2021-08-30
Revert making `error` a keyword

### LINKED ISSUE #11743: undo making "error" a keyword
@axic proposed to not make it a keyword, and I slightly agree it can lead to serious problems with upgrading.

note: Making "error" a keyword has already been implemented in the breaking branch.

cameel:
Fixes #11743.

Just a direct revert of the original commit in #11218. I kept some minor changes like adding the doc page about breaking changes or changing a `catch` variable name from `error` to `_error` in a test unrelated to errors.

### DIFF
--- a/docs/grammar/SolidityParser.g4
+++ b/docs/grammar/SolidityParser.g4
@@ -13,7 +13,6 @@ sourceUnit: (
 	pragmaDirective
 	| importDirective
 	| contractDefinition
-	| errorDefinition
 	| interfaceDefinition
 	| libraryDefinition
 	| functionDefinition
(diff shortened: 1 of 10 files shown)


docs/grammar/SolidityParser.g4
```diff
@@ -13,7 +13,6 @@ sourceUnit: (
 	pragmaDirective
 	| importDirective
 	| contractDefinition
-	| errorDefinition
```
chriseth, line 16: Is this change correct?

test/libsolidity/syntaxTests/errors/weird_name.sol
```diff
@@ -1,8 +0,0 @@
-error error(uint a);
```
chriseth, line 1: Does it make sense to keep this test?

docs/grammar/SolidityParser.g4
cameel: It seems so. Apparently #11218 added a second `errorDefinition` here - note that there's another one at the end of this list. I think it must have been a bug so reverting it is what we want.

test/libsolidity/syntaxTests/errors/weird_name.sol
cameel: Maybe. I think @hrkrshnn added a syntax test only because the semantic one got removed. It's pretty much identical to `semanticTests/errors/weird_name.sol`. But I can add it back if you think we should keep both.

cameel:
`chk_antlr_grammar` is failing because #11851 is only on `develop`. We need to merge `develop` into breaking again.

chriseth:
Rebased after having merged develop into breaking.

test/libsolidity/syntaxTests/errors/weird_name.sol
chriseth: Ah right!

chriseth (approved):

PR #11860, opened 2021-08-30
[SMTChecker] Add static array length constraint

### LINKED ISSUE #11679: [SMTChecker] Length of static array not taken into account in type constraints
```
contract C {
	function f(address[2] memory a) public pure {
		assert(a.length == 2);
	}
}
```
This assertion should hold but it fails.

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/11679

### DIFF
--- a/libsolidity/formal/SymbolicTypes.cpp
+++ b/libsolidity/formal/SymbolicTypes.cpp
@@ -551,6 +551,11 @@ smtutil::Expression symbolicUnknownConstraints(smtutil::Expression _expr, fronte
 	solAssert(_type, "");
 	if (isEnum(*_type) || isInteger(*_type) || isAddress(*_type) || isFixedBytes(*_type))
 		return _expr >= minValue(_type) && _expr <= maxValue(_type);
+	else if (
+		auto arrayType = dynamic_cast<ArrayType const*>(_type);
+		arrayType && !arrayType->isDynamicallySized()
+	)
+		return smtutil::Expression::tuple_get(_expr, 1) == arrayType->length();
 	else if (isArray(*_type) || isMapping(*_type))
 		/// Length cannot be negative.
 		return smtutil::Expression::tuple_get(_expr, 1) >= 0;
--- a/test/libsolidity/smtCheckerTests/types/array_static_aliasing_memory_5.sol
+++ b/test/libsolidity/smtCheckerTests/types/array_static_aliasing_memory_5.sol
@@ -1,9 +1,6 @@
 contract C
 {
 	function f(uint[2] memory a, uint[2] memory b, uint[2] memory c) public pure {
-		require(a.length > 0);
-		require(b.length > 0);
-		require(c.length > 0);
 		require(c[0] == 42);
 		require(a[0] == 2);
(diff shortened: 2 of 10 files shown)


libsolidity/formal/SymbolicTypes.cpp
```diff
@@ -551,6 +551,11 @@ smtutil::Expression symbolicUnknownConstraints(smtutil::Expression _expr, fronte
 	solAssert(_type, "");
 	if (isEnum(*_type) || isInteger(*_type) || isAddress(*_type) || isFixedBytes(*_type))
 		return _expr >= minValue(_type) && _expr <= maxValue(_type);
+	else if (
+		auto arrayType = dynamic_cast<ArrayType const*>(_type);
+		arrayType && !arrayType->isDynamicallySized()
+	)
+		return smtutil::Expression::tuple_get(_expr, 1) == arrayType->length();
```
axic, line 558: Hm, wouldn't it be useful adding `&& (arrayType->length() >= 0)` (as below) here too?

leonardoalt: but that would just be redundant, here we're already assigning the exact value

axic: Ah `->length()` is the actual Solidity type, not a symbolic variable as the function name suggests.

leonardoalt: since `arrayType->length()` is not symbolic, but rather a concrete value at this point

test/libsolidity/smtCheckerTests/types/array_static_aliasing_memory_5.sol
```diff
@@ -1,9 +1,6 @@
 contract C
 {
 	function f(uint[2] memory a, uint[2] memory b, uint[2] memory c) public pure {
-		require(a.length > 0);
-		require(b.length > 0);
-		require(c.length > 0);
```
axic, line 6: Why remove these? To ensure this case tests aliasing only?

test/libsolidity/smtCheckerTests/types/static_array_length_2.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
+	function f() public pure returns (address[2] memory a) {
+		assert(a.length == 2); // should hold
+		assert(a.length < 2); // should fail
```
axic, line 4: Could also add `a.length > 2` for fail.

test/libsolidity/smtCheckerTests/types/array_static_aliasing_memory_5.sol
leonardoalt: They're redundant since the type ensures that length == 2

test/libsolidity/smtCheckerTests/types/static_array_length_2.sol
leonardoalt: Added

leonardoalt:
@axic was that approval? :grimacing:

axic (approved):

PR #11861, opened 2021-08-30
[SMTChecker] Support `value` in CHC for external function calls

### LINKED ISSUE #11249: [SMTChecker] Take function options into account

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/11249

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,7 @@ Compiler Features:
  * Immutable variables can be read at construction time once they are initialized.
  * SMTChecker: Support low level ``call`` as external calls to unknown code.
  * SMTChecker: Add constraints to better correlate ``address(this).balance`` and ``msg.value``.
+ * SMTChecker: Support the ``value`` option for external function calls.
  * Commandline Interface: Disallowed the ``--experimental-via-ir`` option to be used with Standard Json, Assembler and Linker modes.
 
 
--- a/libsolidity/formal/CHC.cpp
+++ b/libsolidity/formal/CHC.cpp
@@ -34,6 +34,7 @@
 
 #include <range/v3/algorithm/for_each.hpp>
 
+#include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/reverse.hpp>
 
 #ifdef HAVE_Z3_DLOPEN
@@ -743,13 +744,15 @@ void CHC::externalFunctionCall(FunctionCall const& _funCall)
 	/// so we just add the nondet_interface predicate.
 
 	solAssert(m_currentContract, "");
-	if (isTrustedExternalCall(&_funCall.expression()))
+	auto [callExpr, callOptions] = functionCallExpression(_funCall);
+
+	if (isTrustedExternalCall(callExpr))
 	{
 		externalFunctionCallToTrustedCode(_funCall);
 		return;
 	}
 
-	FunctionType const& funType = dynamic_cast<FunctionType const&>(*_funCall.expression().annotation().type);
+	FunctionType const& funType = dynamic_cast<FunctionType const&>(*callExpr->annotation().type);
 	auto kind = funType.kind();
 	solAssert(
 		kind == FunctionType::Kind::External ||
@@ -772,6 +775,19 @@ void CHC::externalFunctionCall(FunctionCall const& _funCall)
 
 	if (!m_currentFunction || m_currentFunction->isConstructor())
 		return;
+
+	if (callOptions)
+	{
+		optional<unsigned> valueIndex;
+		for (auto&& [i, name]: callOptions->names() | ranges::views::enumerate)
+			if (name && *name == "value")
+			{
+				valueIndex = i;
+				break;
+			}
+		solAssert(valueIndex, "");
+		state().addBalance(state().thisAddress(), 0 - expr(*callOptions->options().at(*valueIndex)));
+	}
 
 	auto preCallState = vector<smtutil::Expression>{state().state()} + currentStateVariables();
 
@@ -829,6 +845,8 @@ void CHC::externalFunctionCallToTrustedCode(FunctionCall const& _funCall)
 	state().newTx();
 	// set the transaction sender as this contract
 	m_context.addAssertion(state().txMember("msg.sender") == state().thisAddress());
+	// set the transaction value as 0
+	m_context.addAssertion(state().txMember("msg.value") == 0);
 	// set the origin to be the current transaction origin
 	m_context.addAssertion(state().txMember("tx.origin") == txOrigin);
 
@@ -1451,10 +1469,12 @@ smtutil::Expression CHC::predicate(FunctionCall const& _funCall)
 		return smtutil::Expression(true);
 
 	auto contractAddressValue = [this](FunctionCall const& _f) {
-		FunctionType const& funType = dynamic_cast<FunctionType const&>(*_f.expression().annotation().type);
+		auto [callExpr, callOptions] = functionCallExpression(_f);
+
+		FunctionType const& funType = dynamic_cast<FunctionType const&>(*callExpr->annotation().type);
 		if (funType.kind() == FunctionType::Kind::Internal)
 			return state().thisAddress();
-		if (MemberAccess const* callBase = dynamic_cast<MemberAccess const*>(&_f.expression()))
+		if (MemberAccess const* callBase = dynamic_cast<MemberAccess const*>(callExpr))
 			return expr(callBase->expression());
 		solAssert(false, "Unreachable!");
 	};
--- a/libsolidity/formal/SMTEncoder.h
+++ b/libsolidity/formal/SMTEncoder.h
@@ -71,6 +71,10 @@ public:
 	/// otherwise _expr.
 	static Expression const* innermostTuple(Expression const& _expr);
 
+	/// @returns {_funCall.expression(), nullptr} if function call option values are not given, and
+	/// {_funCall.expression().expression(), _funCall.expression()} if they are.
+	static std::pair<Expression const*, FunctionCallOptions const*> functionCallExpression(FunctionCall const& _funCall);
+
 	/// @returns the expression after stripping redundant syntactic sugar.
 	/// Currently supports stripping:
 	/// 1. 1-tuple; i.e. ((x)) -> x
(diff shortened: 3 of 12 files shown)


Changelog.md
```diff
@@ -5,10 +5,11 @@ Language Features:
 
 Compiler Features:
  * Commandline Interface: Normalize paths specified on the command line and make them relative for files located inside base path.
+ * Commandline Interface: Disallowed the ``--experimental-via-ir`` option to be used with Standard Json, Assembler and Linker modes.
```
chriseth: Can we avoid this practice of re-sorting other changelog entries as part of PRs?

chriseth: It's always nice to `git blame` the changelog and find out which PR introduced a feature...

leonardoalt: Sure, just wanted to avoid the Changelog looking messy already, since there are still only a few lines

leonardoalt: Fixed

libsolidity/formal/CHC.cpp
```diff
@@ -749,7 +750,12 @@ void CHC::externalFunctionCall(FunctionCall const& _funCall)
 		return;
 	}
 
-	FunctionType const& funType = dynamic_cast<FunctionType const&>(*_funCall.expression().annotation().type);
+	FunctionCallOptions const* callOptions = nullptr;
+	Expression const* callExpr = &_funCall.expression();
+	if ((callOptions = dynamic_cast<FunctionCallOptions const*>(callExpr)))
```
cameel:
```suggestion
	if (callOptions = dynamic_cast<FunctionCallOptions const*>(callExpr))
```

cameel: I'd either move the `callOptions` variable into the `if` or move the assignment out of the `if`. Current version is the worst of both worlds in my opinion :) Initially I mistook it for `==`.

```suggestion
	Expression const* callExpr = &_funCall.expression();
	auto const* callOptions =  dynamic_cast<FunctionCallOptions const*>(callExpr);
	if (callOptions)
```

Same in `SMTEncoder.cpp`.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
```diff
@@ -0,0 +1,14 @@
+contract C {
+	function g(address payable i) public {
+		require(address(this).balance == 100);
+		i.call{value: 10}("");
+		// Disabled due to Spacer nondeterminism
+		//assert(address(this).balance == 90); // should hold
+		// Disabled due to Spacer nondeterminism
+		//assert(address(this).balance == 100); // should fail
```
cameel, line 8: Do we have some plan for dealing with this nondeterminism long-term?

test/libsolidity/smtCheckerTests/external_calls/call_with_value_2.sol
```diff
@@ -0,0 +1,15 @@
...
+		i.call{value: 0}("");
+		assert(address(this).balance == 100); // should hold
+		assert(address(this).balance == 20); // should fail
+	}
+}
+// ====
+// SMTEngine: all
+// ----
+// Warning 9302: (96-116): Return value of low-level calls not used.
+// Warning 6328: (120-156): CHC: Assertion violation might happen here.
```
cameel, line 13: False positive? It's the line where the comment says "should hold".

test/libsolidity/smtCheckerTests/external_calls/call_with_value_3.sol
```diff
@@ -0,0 +1,15 @@
...
+		// Disabled due to Spacer nondeterminism
+		//assert(address(this).balance == 0); // should fail
+	}
+}
+// ====
+// SMTEngine: all
+// ----
+// Warning 9302: (95-116): Return value of low-level calls not used.
+// Warning 6328: (120-153): CHC: Assertion violation might happen here.
+// Warning 4661: (120-153): BMC: Assertion violation happens here.
```
cameel, line 15: Again, 120 is the line with "should hold".

libsolidity/formal/CHC.cpp
leonardoalt: The compiler complains if I don't put the extra pair.

leonardoalt: The idea was to update and test it at the same time, saving some lines :p because it might also be `nullptr`, in which case we don't want to update `callExpr`.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_2.sol
leonardoalt: `might happen` means the solver couldn't solve it, so it's just an ack here

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
leonardoalt: 99% of the problems are OSX vs rest, so we could disable the smt tests on osx :p

test/libsolidity/smtCheckerTests/external_calls/call_with_value_3.sol
leonardoalt: same, chc `might happen` means it couldn't solve it, bmc `happens` is a false positive

libsolidity/formal/CHC.cpp
cameel: But the code inside the `if`'s parenthesis will always be executed, won't it? It was actually less efficient because of the extra `nullptr` assignment :)

cameel: It probably just tells you that using `=` inside `if` is evil :P

test/libsolidity/smtCheckerTests/external_calls/call_with_value_3.sol
cameel: So, how can we know the feature even works? :P I think all tests have false positives or non-determinism.

libsolidity/formal/CHC.cpp
leonardoalt: It actually asked if I meant `==`

leonardoalt: Why would it always execute?

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
cameel: Maybe we could introduce some soltest setting for having a test skipped on some platforms? Would be better than having to skip it everywhere as we do now.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_3.sol
leonardoalt: Not all though. Combining all the new tests there are a few correct `holds` and a few correct `fails`. But before disabling all the ones that I did, all the tests actually passed without disabling any of them, except for osx.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
leonardoalt: Yea, could be too. I'd still vote for removing osx, from Earth.

libsolidity/formal/CHC.cpp
cameel: Because it can't be tested without peforming the assignment.

Maybe I'm misunderstading what you meant but I'm just saying that this:
```c++
auto const* x = nullptr;
if (x = f())
```
is equivalent to this:
```c++
auto const* x = nullptr;
x = f();
if (x)
```
the assignment and `f()` get executed in both cases.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_3.sol
cameel: ok

libsolidity/formal/CHC.cpp
leonardoalt: Ah right, I misread "inside the if's parenthesis" and thought you meant the branch.

leonardoalt: I'll move the cast out of the if's condition.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
cameel: #11870

libsolidity/formal/CHC.cpp
cameel: That's the same thing :)

test/libsolidity/smtCheckerTests/external_calls/external_call_with_value_3.sol
```diff
@@ -0,0 +1,18 @@
+interface I {
+	function f() external payable;
+}
+
+contract C {
+	function g(I i) public {
+		require(address(this).balance > 100);
+		i.f{value: 20}();
```
cameel, line 8: Maybe test with a call to `this.g{value: ...}()` would be useful too.

leonardoalt: Nice catch, that's somehow broken.

leonardoalt:
@cameel fixed the external calls to `this` issue. Needed to collect the function call expression through a FunctionCallOption in another place as well. Extracted that piece of code into its own function since it was used in 3 places now.

libsolidity/formal/SMTEncoder.h
```diff
@@ -71,6 +71,12 @@ class SMTEncoder: public ASTConstVisitor
 	/// otherwise _expr.
 	static Expression const* innermostTuple(Expression const& _expr);
 
+	/// @returns {_funCall.expression(), nullptr} if function call option
+	/// values are not given, and
+	/// {_funCall.expression().expression(), _funCall.expression()}
+	/// if they are.
+	static std::pair<Expression const*, FunctionCallOptions const*> functionCallExpression(FunctionCall const& _funCall);
```
cameel: Why wrap the comment in an unnatural way like that if the code is longer anyway?
```suggestion
	/// @returns {_funCall.expression(), nullptr} if function call option values are not given, and
	/// {_funCall.expression().expression(), _funCall.expression()} if they are.
	static std::pair<Expression const*, FunctionCallOptions const*> functionCallExpression(FunctionCall const& _funCall);
```

cameel (dismissed):

libsolidity/formal/SMTEncoder.h
leonardoalt: cus that was my editor's width limit

leonardoalt:
@cameel fixed the command and the osx test

cameel (approved):

PR #11863, opened 2021-08-31
Circle CI: Introduce default evm version parameter and change the default to `london`

bshastry:
; remove hard-coded setting.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,9 @@ parameters:
     type: string
     # solbuildpackpusher/solidity-buildpack-deps:emscripten-6
     default: "solbuildpackpusher/solidity-buildpack-deps@sha256:092da5817bc032c91a806b4f73db2a1a31e5cc4c066d94d43eedd9f365df7154"
+  evm-version:
+    type: string
+    default: london
 
 orbs:
   win: circleci/windows@2.2.0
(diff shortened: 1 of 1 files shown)


cameel (dismissed):
Looks good.

@chriseth suggested also changing `soltest.sh` so that we can just use `latest` instead of a specific version. This would save us having to update it. Might be done in a separate PR though.

.circleci/config.yml
```diff
@@ -824,6 +829,8 @@ jobs:
       - store_artifacts: *artifacts_test_results
 
   t_ubu_ubsan_clang_cli:
+    environment:
+      UBSAN_OPTIONS: silence_unsigned_overflow=1
```
cameel: This seems unrelated to the PR.

bshastry: Yeah. I can remove it but felt it would be another PR just for this change. The context is that the nightly UBSan CLI job failed because UBSan printed superfluous unsigned integer overflows in boost onto stdout.

cameel: But the PRs would be independent, wouldn't they? One would not have to wait for the other.

I really think that it's better to have it in its own PR because if I wanted to get back to it, I would not even think about looking in this PR. Or at least to adjust the PR name to mention it.

bshastry: Sounds good, I will create a new PR for it.

cameel (approved):

PR #11668, opened 2021-07-15
No utility for via ir

chriseth:

PR #11661, opened 2021-07-14
Support function debug info via IR.

chriseth:
Part of #10152

PR #11864, opened 2021-08-31
Add protection against mixing "viaIR" and "non-viaIR".

chriseth:
Extracted from #11668

### DIFF
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -741,9 +741,10 @@ Json::Value CompilerStack::generatedSources(string const& _contractName, bool _r
 	return sources.init([&]{
 		Json::Value sources{Json::arrayValue};
 		// If there is no compiler, then no bytecode was generated and thus no
-		// sources were generated.
+		// sources were generated (or we compiled "via IR").
 		if (c.compiler)
 		{
+			solAssert(!m_viaIR, "");
 			string source =
 				_runtime ?
 				c.compiler->runtimeGeneratedYulUtilityCode() :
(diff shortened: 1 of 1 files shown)


libsolidity/interface/CompilerStack.cpp
```diff
@@ -1065,6 +1066,7 @@ size_t CompilerStack::functionEntryPoint(
 	shared_ptr<Compiler> const& compiler = contract(_contractName).compiler;
 	if (!compiler)
 		return 0;
+	solAssert(!m_viaIR, "");
```
cameel: I see that `gasEstimates()` calls this. Does this mean that calling `solc` with both `--gas` and `--experimental-via-ir` will ~not~ now cause an ICE?

chriseth: Ah, interesting! At least `--gas --experimental-via-ir` is an unimplemented feature then, since `compiler->functionEntryLabel` does not use the IR (`compiler` is a non-IR component).

cameel: In that case I think we should have a validation against this option combination in `CommandLineParser`.

chriseth: No, we should implement it.

chriseth:
So it seems we do need https://github.com/ethereum/solidity/pull/11661 - do you want to take a look at it, @cameel ?

libsolidity/interface/CompilerStack.cpp
cameel: Sure but I'd still add a validation until that's implemented.

Or at least the assert here should be changed to `solUnimplementedAssert`.

cameel:
ok. I'm currently working on printing the source snippets but I'll take a look when I'm done with that.

chriseth:
Depends on https://github.com/ethereum/solidity/pull/11661

chriseth:
Changed base.

Marenz (approved):
LGTM, assuming @cameel concerns were addressed.

PR #11865, opened 2021-08-31
Only list used source names.

chriseth:
Replaces https://github.com/ethereum/solidity/pull/11668


chriseth:
This PR will still not "compress" source indices, i.e. there will be gaps in the indices, but at least it does not list all files supplied to the compiler as is currently the case.

libsolidity/codegen/ir/Common.cpp
```diff
@@ -127,9 +127,10 @@ string IRNames::zeroValue(Type const& _type, string const& _variableName)
 	return "zero_" + _type.identifier() + _variableName;
 }
 
-string sourceLocationComment(langutil::SourceLocation const& _location, IRGenerationContext const& _context)
+string sourceLocationComment(langutil::SourceLocation const& _location, IRGenerationContext& _context)
 {
 	solAssert(_location.sourceName, "");
+	_context.markSourceUsed(*_location.sourceName);
```
cameel: This might need a better name. I'd not expect side-effects from it.

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -211,6 +200,7 @@ string IRGenerator::generate(
 	bool creationInvolvesAssembly = m_context.inlineAssemblySeen();
 	t("memoryInitCreation", memoryInit(!creationInvolvesAssembly));
 
+	set<string> usedSourceNames = m_context.usedSourceNames();
 	resetContext(_contract, ExecutionContext::Deployed);
```
cameel: `resetContext()` should clear `usedSourceNames()`.

cameel (commented):
Looks good overall. Just two minor remarks.

libsolidity/codegen/ir/IRGenerator.cpp
chriseth: It does.

libsolidity/codegen/ir/Common.cpp
chriseth: I don't understand - `mark` is an action, so it should have side-effects. Can you suggest a name?

cameel: Sorry I meant `sourceLocationComment()` but commented on a bigger fragment to make the context visible.

How about `dispenseLocationComment()`? It would suggest that getting this comment updates some internal state and that you should expect some side-effects.

libsolidity/codegen/ir/IRGenerator.cpp
cameel: Ah, right. It just creates a new context instead of clearing it field by field :)

cameel (dismissed):
The name is a minor issue so I'm approving anyway but I think we should change it.

libsolidity/codegen/ir/Common.cpp
Marenz: I like that name suggestion

chriseth:
Rebased.

cameel (dismissed):
What do you think about the name I proposed in https://github.com/ethereum/solidity/pull/11865#discussion_r699389587?

chriseth:
Renamed.

chriseth:
If you really want this naming change, I'll submit a different PR first.

chriseth:
Updated.

cameel (approved):

PR #11866, opened 2021-08-31
Clarify function call options example.

chriseth:


docs/control-structures.rst
```diff
@@ -105,8 +105,10 @@ otherwise, the ``value`` option would not be available.
 .. warning::
   Be careful that ``feed.info{value: 10, gas: 800}`` only locally sets the
   ``value`` and amount of ``gas`` sent with the function call, and the
-  parentheses at the end perform the actual call. So in this case, the
-  function is not called and the ``value`` and ``gas`` settings are lost.
+  parentheses at the end perform the actual call. So
+  ``feed.info{value: 10, gas: 800}`` does not call the function and
+  the ``value`` and ``gas`` settings are lost, only
+  ``feed.info{value: 10, gas: 800}()`` performs the function call.
```
cameel, line 111: Perhaps we should warn about this? Or outright disallow it. Is there a case where including the `{}` part when referring to a function (without `()`) is actually useful and has some effect?

cameel (approved):

PaulRBerg:
The new wording is better. @chriseth thanks for clarifying on Matrix and creating this PR.

Issue #11858, opened 2021-08-30
Print code snippets next to source locations in IR


cameel:
With `--asm` option the compiler prints not only source locations but also small snippets of the Solidity code from those locations. For example:

```
tag_5:
    /* "test.sol":111:117  uint x */
  0x00
    /* "test.sol":120:121  1 */
  0x01
    /* "test.sol":111:121  uint x = 1 */
  swap1
  pop
    /* "test.sol":101:128  {... */
  pop
    /* "test.sol":81:128  function f() public {... */
  jump      // out
```

In IR we only get locations:
```yul
/// @src 0:81:128
function fun_f_21() {

    /// @src 0:120:121
    let expr_18 := 0x01
    /// @src 0:111:121
    let var_x_17 := convert_t_rational_1_by_1_to_t_uint256(expr_18)

}
/// @src 0:0:130

function identity(value) -> ret {
    ret := value
}
```

`AsmPrinter` should be extended to add snippets in IR too.

cameel:
Also, there should be a command-line flag for selecting debug info format:
> there should be at least three modes: no debug info comments in the IR, regular debug info (as we have it now), debug info with source snippets

cameel:
@chriseth What do you think about calling the new option `--inline-debug-info`? Or `--debug-tags`?

It would have three values for now:
- `--inline-debug-info=disabled`
- `--inline-debug-info=location`
- `--inline-debug-info=location,snippet`

This could map to either an enum or to a struct with a set of `bool` flags. For now I'm making it an enum. Not sure where to put it though - if I create a single enum for both `libyul::AsmPrinter` and for `libevmasm::Assembly` it will create a dependency between them. But having two separate enums does not sound all that good either.

chriseth:
Why "inline"? Maybe `--ir-debug=...` with the following options: `astid` (future PR), `src` (the snippet), `loc` (the location)

`--ir-debug-info` would also be fine for me

cameel:
> Why "inline"?

Well, I wanted to distinguish it from the full debug info that could be printed by the compiler when requested as its own artifact but I guess these comments are the only way to output any debug info for now? In that case `--ir-debug` sounds ok.

Should I allow `--ir-debug=src` (i.e just the snippet without location)? Currently the `loc` part in the parser is not optional. Not sure if I should make it optional or just disallow this combination. Does printing only snippets sound like a useful feature?

cameel:
[...]

PR #11867, opened 2021-08-31
Solidity snippets next to source locations in IR

cameel:
Depends on #11913.

First part of #11858. No options to control this behavior yet - for now the snippets are always included.

TODO:
- [x] Make sure snippets are present in all kinds of IR output. For now I checked only `--ir` and `--ir-optimized` flags.
- [x]
    > Please check that the AsmParser correctly parses the `@src` comment under all circumstances, even if the original source code contains a string `"something @src something"`.

### DIFF
--- a/libsolidity/codegen/ir/Common.cpp
+++ b/libsolidity/codegen/ir/Common.cpp
@@ -138,7 +138,8 @@ string dispenseLocationComment(langutil::SourceLocation const& _location, IRGene
 	return AsmPrinter::formatSourceLocationComment(
 		_location,
 		_context.sourceIndices(),
-		true /* _statement */
+		true /* _statement */,
+		_context.soliditySourceProvider()
 	);
 }
 
--- a/libsolidity/codegen/ir/Common.cpp
+++ b/libsolidity/codegen/ir/Common.cpp
@@ -22,9 +22,13 @@
 
 #include <libsolutil/CommonIO.h>
 
+#include <libyul/AsmPrinter.h>
+
 using namespace std;
-using namespace solidity::util;
+using namespace solidity::langutil;
 using namespace solidity::frontend;
+using namespace solidity::util;
+using namespace solidity::yul;
 
 namespace solidity::frontend
 {
@@ -131,12 +135,11 @@ string dispenseLocationComment(langutil::SourceLocation const& _location, IRGene
 {
 	solAssert(_location.sourceName, "");
 	_context.markSourceUsed(*_location.sourceName);
-	return "/// @src "
-		+ to_string(_context.sourceIndices().at(*_location.sourceName))
-		+ ":"
-		+ to_string(_location.start)
-		+ ":"
-		+ to_string(_location.end);
+	return AsmPrinter::formatSourceLocationComment(
+		_location,
+		_context.sourceIndices(),
+		true /* _statement */
+	);
 }
 
 string dispenseLocationComment(ASTNode const& _node, IRGenerationContext& _context)
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -1329,7 +1329,7 @@ void CompilerStack::generateIR(ContractDefinition const& _contract)
 	for (auto const& pair: m_contracts)
 		otherYulSources.emplace(pair.second.contract, pair.second.yulIR);
 
-	IRGenerator generator(m_evmVersion, m_revertStrings, m_optimiserSettings, sourceIndices());
+	IRGenerator generator(m_evmVersion, m_revertStrings, m_optimiserSettings, sourceIndices(), this);
 	tie(compiledContract.yulIR, compiledContract.yulIROptimized) = generator.run(
 		_contract,
 		createCBORMetadata(compiledContract),
--- a/libyul/AsmParser.cpp
+++ b/libyul/AsmParser.cpp
@@ -135,7 +135,8 @@ void Parser::fetchSourceLocationFromComment()
 		regex_constants::ECMAScript | regex_constants::optimize
 	);
 	static regex const srcTagArgsRegex = regex(
-		R"~~(^(-1|\d+):(-1|\d+):(-1|\d+)(?:\s+|$))~~", // index and location, e.g.: 1:234:-1
+		R"~~(^(-1|\d+):(-1|\d+):(-1|\d+)(?:\s+|$))~~"  // index and location, e.g.: 1:234:-1
+		R"~~(("(?:[^"\\]|\\.)*"?)?)~~",                // optional code snippet, e.g.: "string memory s = \"abc\";..."
 		regex_constants::ECMAScript | regex_constants::optimize
 	);
 
@@ -164,8 +165,21 @@ void Parser::fetchSourceLocationFromComment()
 				return;
 			}
 
-			solAssert(srcTagArgsMatch.size() == 4, "");
+			solAssert(srcTagArgsMatch.size() == 5, "");
 			position += srcTagArgsMatch.position() + srcTagArgsMatch.length();
+
+			if (srcTagArgsMatch[4].matched && (
+				!boost::algorithm::ends_with(srcTagArgsMatch[4].str(), "\"") ||
+				boost::algorithm::ends_with(srcTagArgsMatch[4].str(), "\\\"")
+			))
+			{
+				m_errorReporter.syntaxError(
+					1544_error,
+					commentLocation,
+					"Invalid code snippet in source location mapping. Quote is not terminated."
+				);
+				return;
+			}
 
 			optional<int> const sourceIndex = toInt(srcTagArgsMatch[1].str());
 			optional<int> const start = toInt(srcTagArgsMatch[2].str());
--- a/libyul/AsmPrinter.cpp
+++ b/libyul/AsmPrinter.cpp
@@ -38,6 +38,7 @@
 
 using namespace std;
 using namespace solidity;
+using namespace solidity::langutil;
 using namespace solidity::util;
 using namespace solidity::yul;
 
@@ -256,7 +257,33 @@ string AsmPrinter::appendTypeName(YulString _type, bool _isBoolLiteral) const
 		return ":" + _type.str();
 }
 
-string AsmPrinter::formatSourceLocationComment(shared_ptr<DebugData const> const& _debugData,  bool _statement)
+string AsmPrinter::formatSourceLocationComment(
+	SourceLocation const& _location,
+	map<string, unsigned> const& _nameToSourceIndex,
+	bool _statement
+)
+{
+	yulAssert(!_nameToSourceIndex.empty(), "");
+
+	string sourceIndex = "-1";
+	if (_location.sourceName)
+		sourceIndex = to_string(_nameToSourceIndex.at(*_location.sourceName));
+
+	string sourceLocation =
+		"@src " +
+		sourceIndex +
+		":" +
+		to_string(_location.start) +
+		":" +
+		to_string(_location.end);
+
+	return
+		_statement ?
+		"/// " + sourceLocation :
+		"/** " + sourceLocation + " */ ";
+}
+
+string AsmPrinter::formatSourceLocationComment(shared_ptr<DebugData const> const& _debugData, bool _statement)
 {
 	if (
 		!_debugData ||
@@ -267,19 +294,9 @@ string AsmPrinter::formatSourceLocationComment(shared_ptr<DebugData const> const
 
 	m_lastLocation = _debugData->location;
 
-	string sourceIndex = "-1";
-	if (_debugData->location.sourceName)
-		sourceIndex = to_string(m_nameToSourceIndex.at(*_debugData->location.sourceName));
-
-	string sourceLocation =
-		"@src " +
-		sourceIndex +
-		":" +
-		to_string(_debugData->location.start) +
-		":" +
-		to_string(_debugData->location.end);
-	return
-		_statement ?
-		"/// " + sourceLocation + "\n" :
-		"/** " + sourceLocation + " */ ";
+	return formatSourceLocationComment(
+		_debugData->location,
+		m_nameToSourceIndex,
+		_statement
+	) + (_statement ? "\n" : "");
 }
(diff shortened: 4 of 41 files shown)


libsolidity/interface/CompilerStack.cpp
```diff
@@ -1329,7 +1329,7 @@ void CompilerStack::generateIR(ContractDefinition const& _contract)
 	for (auto const& pair: m_contracts)
 		otherYulSources.emplace(pair.second.contract, pair.second.yulIR);
 
-	IRGenerator generator(m_evmVersion, m_revertStrings, m_optimiserSettings, sourceIndices());
+	IRGenerator generator(m_evmVersion, m_revertStrings, m_optimiserSettings, sourceIndices(), this);
```
cameel, line 1332: This now passes the `CompilerStack` instance (as `CharStreamProvider`) all the way down through `IRGenerator`, `IRGeneratorContext`, `AssemblyStack` and `Object` down to `AsmPrinter`. All the methods on the way need the provider as an optional argument. Not sure if there's a way to make it nicer.This now passes the `CompilerStack` instance (as `CharStreamProvider`) all the way down through `IRGenerator`, `IRGeneratorContext`, `AssemblyStack` and `Object` down to `AsmPrinter`. All the methods on the way need the provider as an optional argument. Not sure if there's a way to make it nicer.

libsolidity/codegen/ir/Common.cpp
```diff
@@ -127,15 +131,15 @@ string IRNames::zeroValue(Type const& _type, string const& _variableName)
...
-		+ to_string(_location.start)
-		+ ":"
-		+ to_string(_location.end);
+	return AsmPrinter::formatSourceLocationComment(
+		_location,
+		_context.sourceIndices(),
+		true /* _statement */,
+		_context.soliditySourceProvider()
+	);
 }
```
cameel: This is a piece of code duplicated from `AsmParser`. Because of this, my changes in `AsmParser` added snippets for `--ir-optimized` but not for `--ir`.

I moved this from both places into a generic static method in `AsmParser`.

test/cmdlineTests/optimizer_array_sload/output
```diff
@@ -31,27 +31,27 @@ object "Arraysum_34" {
...
+                        /** @src 0:380:397  i < values.length */ lt(var_i, _2)
+                        /// @src 0:368:378  uint i = 0
                         {
-                            /// @src 0:80:429
+                            /// @src 0:80:429  contract Arraysum {...
                             if eq(var_i, not(0)) { panic_error_0x11() }
-                            /// @src 0:399:402
-                            var_i := /** @src 0:80:429 */ add(var_i, 1)
+                            /// @src 0:399:402  i++
+                            var_i := /** @src 0:80:429  contract Arraysum {... */ add(var_i, 1)
```
cameel: These inline comments are probably not very useful when they point at the whole contract. Maybe we should have a heuristic that cuts them out?

test/cmdlineTests/yul_string_format_ascii_long/output.json
```diff
@@ -152,18 +155,21 @@ object \"C_11\" {
                 mstore(64, newFreePtr)
             }
 
-            /// @src 0:91:241
+            /// @src 0:91:241  function f() external pure returns (string memory) { return \"abcdabcdcafecafeabcdabcdcafecafeffffzzzzoooo0123456789,.<,>.?:;'[{]}|`~!@#$%^&*()-_=+\"; }
+
             function fun_f_10() -> var__5_mpos {
```
cameel: Having the original declaration next to the function is nice.

cameel:
Looks like my change produces invalid IR on Windows - `...` gets dropped to the next line. I think must be because of `\r\n` line endings (the function that prepares snippets looks for `\n` only).

liblangutil/SourceLocation.cpp
```diff
@@ -27,6 +27,34 @@ using namespace solidity;
...
+	auto it = _sourceCodes.find(*this->sourceName);
+	if (it == _sourceCodes.end())
+		return {};
+
+	return singleLineSnippet(it->second);
+}
+
+string SourceLocation::singleLineSnippet(string const& _sourceCode) const
+{
+	if (!this->hasText())
```
chriseth: We usually never use explicit `this`...

chriseth:
Please check that the AsmParser correctly parses the `@src` comment under all circumstances, even if the original source code contains a string `"something @src something"`.

test/cmdlineTests/exp_base_literal/output
```diff
@@ -10,7 +10,8 @@ IR:
 /// @use-src 0:"exp_base_literal/input.sol", 1:"#utility.yul"
 object "C_81" {
     code {
-        /// @src 0:82:370
+        /// @src 0:82:370  contract C {...
```
chriseth: Something is adding additional newlines here

liblangutil/SourceLocation.cpp
cameel: Oh, true. I did not notice that `this` is pointless here when moving this :)

liblangutil/SourceLocation.cpp
```diff
@@ -27,6 +27,34 @@ using namespace solidity;
 using namespace solidity::langutil;
 using namespace std;
 
+string SourceLocation::singleLineSnippet(StringMap const& _sourceCodes) const
```
chriseth: I don't fully remember, but I think I specifically took the functionality to extract the source snippet out of this class, and I think it is nicer if this class just stays a data container without extra functionality, but at least we should find this place and unify it if possible.

liblangutil/SourceLocation.cpp
cameel: No, wait. I replaced them myself :) Sorry, not sure how I did not notice this earlier :)

Anyway, removed.

chriseth:
The fact that we have to drag the source provider all the way down is really not nice, but I currently don't see a better way either, but I did not take a close look yet. It gets even more complicated once we add the option about how to format the source references.

liblangutil/SourceLocation.cpp
cameel: I could make it a standalone function in that file.

~Or move it to `liblangutil/Common.h`.~ No it would have to depend on source location. A new file specifically for this function then.

chriseth: It's `CharStream::text`

cameel: Ah, that's what you meant. Right. I can add it next to that helper.

test/cmdlineTests/exp_base_literal/output
cameel: Fixed. The helper in `Common.h` was not expected to print them while `AsmParser` is. To avoid adding an extra argument in `formatSourceLocationComment()` just to control newlines, some of the overloads now skip this newline while some do not. Which is not great but at least preserves the current behavior.

liblangutil/SourceLocation.cpp
cameel: Moved to `CharStream`.

cameel:
> Please check that the AsmParser correctly parses the `@src` comment under all circumstances, even if the original source code contains a string `"something @src something"`.

It uses this regex on the result of `currentCommentLiteral()` so it looks like it will catch `@src` in the middle of some text as long as it's followed by at least one space:
```c++
static regex const lineRE = std::regex(
	R"~~~((^|\s*)@src\s+(-1|\d+):(-1|\d+):(-1|\d+)(\s+|$))~~~",
	std::regex_constants::ECMAScript | std::regex_constants::optimize
);
```

Also, nested comments will likely be a problem too.

This means that snippets will have to be escaped or sanitized in some way.

I'm looking for a way to test it though. I tried feeding the output of `--ir-optimized` into `--strict-assembly` but looks like that does not show snippets in the assembly (even on `develop` so it's not because my changes broke it :)).

cameel:
Snippet escaping done. Now `@src` parsing will ignore properly escaped snippets next to the tag so that tags inside them will not affect locations.

I dealt with `*/` inside snippets by replacing `/` with `\/`, but only in that particular sequence. Other `/` are not escaped to reduce noise.

This is now ready for another review.

libyul/AsmParser.cpp
```diff
@@ -131,7 +131,9 @@ void Parser::fetchSourceLocationFromComment()
 		return;
 
 	static regex const lineRE = std::regex(
-		R"~~~((^|\s*)@src\s+(-1|\d+):(-1|\d+):(-1|\d+)(\s+|$))~~~",
```
chriseth: Since we plan to parse more tags, would it be possible to use an instance of Scanner for this?

cameel: Probably. I'd have to create a completely new scanner, just for the single comment that is being processed?

But I'd rather leave it as is here and do it in a new PR. It looks like it will completely change how this function works and I have this already tested and working so I'd rather not drag this PR out too much.

chriseth: Yes, but Scanner is a very simple class. We already do exactly the same thing for parsing `@use-src`,

cameel: After trying to use the scanner, looks like ensuring that tags nested in code snippets are not treated as actual tags is problematic.

I could first use a regex to find all substrings containing `@src` tags with parameters and then process each with a scanner. But to be able to determine where each substring ends, the regex would look pretty much like the one I have now. It's not what you wanted, is it? A simpler regex would not skip nested `@src` in code snippets.

I could also run the scanner over the whole comment and only start processing a new tag once the current one ends. This looks like the right solution to me but the problem here is that the scanner does not have a token for `@` and I think that adding it would affect how the language is parsed (some `@` inside other stuff would now become standalone tokens?).

I see that the Natspec parser just manually finds `@` inside the string and only then processes the part between `@` and the end of a line as a natspec tag. I could do that if I could assume that there can be only one tag per line. Can we make such an assumption here? Currently multiple `@src` tags on a single line are accepted and the last one wins.

chriseth: Ok, this starts to get very complicated :)
I think we can make our own assumptions/restrictions as long as printing and parsing is in line.
Forcing the tags to be at the start of a line is probably a good idea. I hope there are no tools that do reformatting.
Somewhere I saw that the line breaks inside are not preserved - but maybe that was just the natspec parser.

cameel: What about one tag per line assumption? Without either that or adding `@` token to the scanner it's still problematic.

> Somewhere I saw that the line breaks inside are not preserved - but maybe that was just the natspec parser.

The natspec parser stops at the end of the line as far as I can tell:
https://github.com/ethereum/solidity/blob/4615e625142b9c8366c87bc56b3c6f0fad2091d5/libsolidity/parsing/DocStringParser.cpp#L90-L95

cameel:
Rebased on `develop` to resolve conflicts with other PRs that got merged in the meantime.

libyul/AsmParser.cpp
cameel: For reference, here's what we agreed on on the chat:
> can't we use a generic regex to search for tags first and then when we find the `@src` tag, we skip over the string?
> (and with tags I mean anything starting with `@`)

> no I mean just search for tags starting with `@`
> if you find one, but the tag is unknown, we ignore it and just search for the next `@`
> if there is a string, then it is wrong
> but it's also unknown tag
> so I think it's fine to properly consider the string for `@src`, and do nothing for the other tags (that don't also have strings as arguments)

libyul/AsmPrinter.cpp
```diff
@@ -256,30 +257,46 @@ string AsmPrinter::appendTypeName(YulString _type, bool _isBoolLiteral) const
 		return ":" + _type.str();
 }
 
-string AsmPrinter::formatSourceLocationComment(shared_ptr<DebugData const> const& _debugData,  bool _statement)
```
chriseth, line 259: Soon, this will have to handle all debug data, not only source locations. Should we undo this change in this PR or later?

cameel:
Looks like it was not actually rebased. Only base in the PR was changed.

libyul/AsmPrinter.cpp
cameel: This was discussed on chat so marking as resolved. The overload that uses `DebugData` has not been removed, it's below.

libyul/AsmParser.cpp
cameel: Here's a new PR that does this refactor: #11913.
The PR does not really have anything to do with snippets and I had a bunch of other refactors related to parsing `@src` tags to I extracted it all to that PR. For this reason I rebased the snippet PR on top of it so now it's a dependency.

cameel:
This PR now depends on #11913.

test/libyul/Parser.cpp
```diff
@@ -642,6 +642,164 @@ BOOST_AUTO_TEST_CASE(customSourceLocations_reference_original_sloc)
...
+	ErrorList errorList;
+	ErrorReporter reporter(errorList);
+	auto const sourceText = R"(
+		/// @src 0:111:222 "abc"def
+		{}
+	)";
+	EVMDialectTyped const& dialect = EVMDialectTyped::instance(EVMVersion{});
+	shared_ptr<Block> result = parse(sourceText, dialect, reporter);
+	BOOST_REQUIRE(!!result && errorList.size() == 0);
+	CHECK_LOCATION(result->debugData->location, "source0", 111, 222);
```
cameel, line 744: The regex for location strictly requires whitespace after the numbers. I did not change that aspect but for the quoted snippet, which is a new element, I made it more lenient - you don't have to add whitespace after it. Is that ok?

cameel: From @chriseth:
> of course no strict requirement for whitespace after something
> whitespace is only required to separate stuff

I'm leaving it as is then.

cameel: Actually, I think it wasn't clear from my question on the chat that I mean requiring whitespace in the case above, i.e. `"abc" def` and not requiring trailing whitespace after the quote . So I guess I should not allow this after all?

chriseth (dismissed):

chriseth:
Test seems to be failing :(

cameel:
Yeah, I rebased it on top of #11913 today but looks like I need to fix one of the tests that's no longer valid due to changes in that lower PR.

cameel:
Tests should be fixed now.

test/libyul/Parser.cpp
cameel: I left it as is after all. I don't think that missing whitespace after `"` should be a syntax error and ignoring the whole tag because of it does not seem like a good idea either.

chriseth (approved):

Issue #11846, opened 2021-08-26
Simplify and deduplicate CircleCI config


cameel:
This came up in #11823.

Our [.circleci/config.yml](https://github.com/ethereum/solidity/blob/develop/.circleci/config.yml) file is pretty long. I went through the file and here's a bunch of changes that we could do to improve it.

Deduplication:
- [x] `t_archlinux_soltest`, `t_ubu_soltest_enforce_yul`, `t_ubu_ubsan_clang` and `t_ubu_ubsan_clang_cli` could use `run_soltest_steps`.
    - Note that `run_soltest_steps` cannot be added to an existing list of steps but it you could probably embed inside a step by using a [`when` parameter of a `run` step](https://circleci.com/docs/2.0/configuration-reference/#run).
- Some jobs look very similar. Perhaps they could be replaced by a single job with parameters (see `t_ems_ext` job for an example; also [Reusable Config Reference Guide](https://circleci.com/docs/2.0/reusing-config/)). If not, it might at least be possible to extract reusable steps from them:
    - [ ] `b_ubu_clang`, `b_ubu_asan_clang` and `b_ubu_ubsan_clang`.
    - [ ] `t_osx_soltest` and `t_osx_cli`.
    - [ ] `b_bytecode_ubu`, `b_bytecode_osx` and `b_bytecode_win`. 
- [x] Many jobs set the `TERM` variable to `xterm`. I think it would not hurt to just have it set once, globally, so that we do not have to worry about it.
    - EDIT: This probably does not have a good solution (see https://github.com/ethereum/solidity/pull/11868#discussion_r699430288) but I'm leaving it open. Maybe someone will figure out something.
    - EDIT2: I found a good compromise in #12173 - now the variable is defined in the `base_` templates.
- [ ] Jobs that differ only by one or two steps could be simplified by [passing these steps as arguments](https://circleci.com/docs/2.0/reusing-config/#steps).

Refactors:
[...]

PR #11868, opened 2021-08-31
CleanUp/Deduplication in .circleci/config.yml

whhygee:
Implements part of  #11846.

Cleanup
- Dicts containing steps: key have been renamed to start with `steps_` instead of `run_`
- All of `run_` steps have been moved to `Build Templates`
- `b_ubu_asan` has been moved next to `b_ubu_asan_clang` (Line 455)
- Branch `develop_060` doesn't exist at the moment. 

Deduplication
- The `xterm` variable in job environment has been declared globally in pipeline parameters (Line 26) as `environment-term`.

Following your suggestion, I intend on breaking down the tasks into smaller PRs, this being the first one if the changes seem appropriate. Let me know about any changes needed in the current commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,25 @@ defaults:
       name: Correctness proofs for optimization rules
       command: scripts/run_proofs.sh
 
+  - run_soltest: &run_soltest
+      name: soltest
+      no_output_timeout: 30m
+      command: ./.circleci/soltest.sh
+
+  - run_soltest_all: &run_soltest_all
+      name: soltest_all
+      no_output_timeout: 30m
+      command: ./.circleci/soltest_all.sh
+
+  - run_cmdline_tests: &run_cmdline_tests
+      name: command line tests
+      no_output_timeout: 30m
+      command: ./test/cmdlineTests.sh
+
+  - run_docs_pragma_min_version: &run_docs_pragma_min_version
+      name: docs pragma version check
+      command: ./scripts/docs_version_pragma_check.sh
+
   # --------------------------------------------------------------------------
   # Artifacts Templates
 
@@ -114,17 +133,7 @@ defaults:
   - store_test_results: &store_test_results
       path: test_results/
 
-  - run_soltest: &run_soltest
-      name: soltest
-      no_output_timeout: 30m
-      command: ./.circleci/soltest.sh
-
-  - run_soltest_all: &run_soltest_all
-      name: soltest_all
-      no_output_timeout: 30m
-      command: ./.circleci/soltest_all.sh
-
-  - run_soltest_steps: &run_soltest_steps
+  - steps_soltest: &steps_soltest
       steps:
         - checkout
         - attach_workspace:
@@ -133,7 +142,7 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
-  - run_soltest_all_steps: &run_soltest_all_steps
+  - steps_soltest_all: &steps_soltest_all
       steps:
         - checkout
         - attach_workspace:
@@ -142,12 +151,7 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
-  - run_cmdline_tests: &run_cmdline_tests
-      name: command line tests
-      no_output_timeout: 30m
-      command: ./test/cmdlineTests.sh
-
-  - run_cmdline_tests_steps: &run_cmdline_tests_steps
+  - steps_cmdline_tests: &steps_cmdline_tests
       steps:
         - checkout
         - attach_workspace:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -306,7 +309,7 @@ jobs:
     docker:
       - image: circleci/node
     environment:
-      TERM: xterm
+      TERM: << pipeline.parameters.environment-term >>
```
cameel: What I meant was to have the variable automatically declared for all jobs.

Maybe that's not possible - part of the task is to check that. Looking at [Environment variable usage options > Order of precedence](https://circleci.com/docs/2.0/env-vars/#order-of-precedence), the highest level that is still possible via `.config.yml` file be the `job` dict, which we are already doing.

If it's not possible, then I'd just leave the variable is it was because doing it via parameters is actually more verbose and still requires you to remember to put it on a job.

cameel (commented):
Other than the `TERM` variable, it looks good.

cameel:
> * Branch `develop_060` doesn't exist at the moment.

Even more reason to remove it from the nightly run :) I think it stopped being relevant after 0.6.0 and just no one remembered to get rid of it from the config.

cameel:
```
# Unable to parse YAML
# found undefined alias b_ubu
#  in 'string', line 456, column 9:
#         <<: *b_ubu
#             ^
```
So this explains why `b_ubu_asan` was not placed next to the other sanitizer jobs. In that case let's move those other jobs so that they are declared after `b_ubu`.

whhygee:
Noted! I'll make the necessary changes in the next commit, and then we could continue with the rest of the tasks.

whhygee:
### Cleanup
- `b_ubu_asan`, `b_ubu_clang`, `b_ubu_asan_clang`, `b_ubu_ubsan_clang` have been moved after `b_ubu` . (Line 454)
- Removed the nightly workflow on the `develop_060` branch, pursuant to its deletion/irrelevance

### Deduplication
- `xterm` variable has been restored to its original state

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ parameters:
     default: "solbuildpackpusher/solidity-buildpack-deps@sha256:092da5817bc032c91a806b4f73db2a1a31e5cc4c066d94d43eedd9f365df7154"
   evm-version:
     type: string
-    default: london
+    default: london\
```
cameel: Looks like an accidental change:
```suggestion
    default: london
```

cameel (commented):
Looks like external PRs do not run under our CI plan. You're probably running it under your own account that has run out of free credits ([`t_pyscripts_win`](https://app.circleci.com/pipelines/github/yatharthagoenka/solidity/7/workflows/d9b3dd92-ed16-4123-8099-41339ec393f3/jobs/70)).

Let's see if I can make it run manually by adding a dummy commit.

cameel:
To work around the CI limitations, I have created a separate PR: #11880. I see that you can still run some CI jobs so please continue working on it here. When you're done I'll pull changes into my PR so that it can complete the whole run.

yatharthagoenka:
### Deduplication
`run_soltest_steps` was renamed to `steps_soltest` in #11868.
- `t_archlinux_soltest` was incorporated with `steps_soltest` using the when parameter, setting the condition to `true`.
- A direct reference for the same was added to `t_ubu_soltest_enforce_yul`
- `t_ubu_ubsan_clang` and `t_ubu_ubsan_clang_cli` have a different set of steps, thus I'm not sure if we could use `steps_soltest` instead, as it would disrupt the original flow of commands.

whhygee:
### Deduplication
> `run_soltest_steps` was renamed to `steps_soltest` #11868.
- `steps_soltest` was incorporated into `t_archlinux_soltest` using the when parameter of run, setting the condition to true.
- The same was directly references into `t_ubu_soltest_enforce_yul` as it had the same set of steps.
- `t_ubu_ubsan_clang` and `t_ubu_ubsan_clang_cli` had a differing set of steps as `steps_soltest`, thus I'm not sure if we could refer to them in this regard, as it would disrupt the original flow of commands

cameel:
> * `t_ubu_ubsan_clang` and `t_ubu_ubsan_clang_cli` had a differing set of steps as `steps_soltest`, thus I'm not sure if we could refer to them in this regard, as it would disrupt the original flow of commands

The only difference is `- run: *gitter_notify_failure`, which can be safely moved to the end of the list of steps. It's just a notification posted to the gitter channel if the job fails.

By the way, this set of changes looks fine overall. After you do `t_ubu_ubsan_clang` and `t_ubu_ubsan_clang_cli` I'll pull your changes into #11880 and merge them if the CI passes. For other subtasks from #11846 please start a new PR.

whhygee:
Sure, Noted!
I've completed the said tasks in the last commit (28d0a50).
> `t_ubu_ubsan_clang_cli ` had another differing step i.e. `- run: *run_cmdline_tests`, which was also moved after the `steps_soltest`

.circleci/config.yml
```diff
@@ -813,25 +804,20 @@ jobs:
...
-      - checkout
-      - attach_workspace:
-          at: build
+      - when:
+          condition: true
+          <<: *steps_soltest
       - run: *run_cmdline_tests
       - run: *gitter_notify_failure
-      - store_test_results: *store_test_results
-      - store_artifacts: *artifacts_test_results
```
cameel: Why `steps_soltest`? It has `run_cmdline_tests` so you should use `steps_cmdline_tests`.

whhygee:
Might have confused it with another test. Thank you for pointing it out.
- .`steps_cmdline_tests` have been incorporated into `t_ubu_ubsan_clang_cli`

cameel:
Ok, this looks good now. Could you squash commits with review fixes?

yatharthagoenka:
Commits have been squashed into c0dd58d.

whhygee:
Commits were squashed into c0dd58d. 

Should I open up a new PR for the remaining changes, or do we still have work to do in this one?

cameel:
I'll need someone to review and approve #11880 so there's a small chance that some problem will be spotted and minor changes will be needed but other than that I think we're finished here.

Please open new PRs for the remaining changes.

yatharthagoenka:
Sure, Noted! 

I'll open another PR with the next lead, while this gets reviewed.

whhygee:
Sure, Noted! 

A new PR #11896 has been opened to work on the remaining tasks, while this gets reviewed.

PR #11869, opened 2021-08-31
Allow and require use-src to be repeated for each object.

chriseth:
This is needed for `new` to work properly without having to re-parse the IR code of the created object.

Will combine this with https://github.com/ethereum/solidity/pull/11865

### DIFF
--- a/libyul/ObjectParser.cpp
+++ b/libyul/ObjectParser.cpp
@@ -45,14 +45,15 @@ shared_ptr<Object> ObjectParser::parse(shared_ptr<Scanner> const& _scanner, bool
 	{
 		shared_ptr<Object> object;
 		m_scanner = _scanner;
-		m_sourceNameMapping = tryParseSourceNameMapping();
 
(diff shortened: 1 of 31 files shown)


libyul/ObjectParser.cpp
```diff
@@ -103,13 +107,13 @@ shared_ptr<Object> ObjectParser::parseObject(Object* _containingObject)
...
 
-shared_ptr<Block> ObjectParser::parseCode()
+shared_ptr<Block> ObjectParser::parseCode(optional<SourceNameMap> _sourceNames)
 {
 	if (currentToken() != Token::Identifier || currentLiteral() != "code")
 		fatalParserError(4846_error, "Expected keyword \"code\".");
 	advance();
 
-	return parseBlock();
+	return parseBlock(_sourceNames);
```
cameel:
```suggestion
	return parseBlock(move(_sourceNames));
```

ekpyron (approved):

Issue #11845, opened 2021-08-26
Require latest Z3 in soltest


cameel:
Extracted from https://github.com/ethereum/solidity/pull/11844#discussion_r696541700

We support linking with a range of versions of Z3 (the minimum is currently Z3 4.8.0) but our SMTChecker tests are only guaranteed to pass with the latest one (currently 4.8.12). This is because different versions of Z3 sometimes give different (but still valid) results and it's hard to create a test harness that could account for these differences. Instead we compare results verbatim. Because of this we sometimes get reports about broken tests when people build from source and use older Z3.

We don't want to raise the minimum version for `solc` but we could improve the situation by doing it for `soltest` and `isoltest`. Then anyone wanting to use them would have to have the latest Z3 or none at all (if an older version is installed it can always be disabled by passing `-DUSE_Z3=OFF` to cmake).

soroosh-sdi:
I dig in this issue, and recreated the problem, this is what I've found:
- In `scripts/tests.sh` , we have comandline tests that actually run solc with an input file and expect specific output. these test are created by looping over all folders in `test/cmdlineTests/`. when we don't have any model checker OR z3's version < 4.8.12 we should ignore `/test/cmdlineTests/*model_checker*` in that loop. unfortunately we cannot do something about them in CMake.
- Also, in the tessuite we have `SMTCheckerTest` which should be disabled if we don't have any model checker OR z3's version < 4.8.12; fortunately we can do something about it in Cmake
- I compiled project with `-DUSE_Z3=OFF` but both of above issues exist.

for the first issue I should recognize SMT is disabled  and if it's enabled what version of Z3 solc is using; maybe we could add an option to solc to print this info. what do you  think @cameel ? 
 for the second issue I could implement something in CMake.

leonardoalt:
@soroosh-sdi the problem with recognizing it (for the first issue) is that `solc` can also use z3 dynamically even if it's not compiled with it. So you'd need to query that part as well. Should be doable anyway, just more work

cameel:
@leonardoalt Do you mean automatically disabling these commandline tests when Z3 is not available? We do not do that in `soltest` and `--no-smt` is required instead. I think we should just add support for this flag in `cmdlineTests.sh` as well (and, by extension, in `tests.sh`) and require user to explicitly disable them.

@soroosh-sdi 
[...]

PR #11712, opened 2021-07-27
Running only selected command-line tests

cameel:
A simple modification that makes `cmdlineTests.sh` run tests only from the directories that match positional arguments.

For example this command:
```bash
test/cmdlineTests.sh --update *standard* stdin/ too_long_line*/
```
will run only tests from:
- `test/cmdlineTests/*standard*/`
- `test/cmdlineTests/stdin/`
- `test/cmdlineTests/stdin/too_long_line*/`

By default - if no pattern is specified - it runs all tests.

Note that when the script is not executed from inside `test/cmdlineTests/` and a pattern matches some files in the current dir, it needs to be quoted.  For example running `test/cmdlineTests.sh *` from repo root will try to use `Changelog.md` as the test name so you have to use `test/cmdlineTests.sh "*"` instead.

PR #6891, opened 2019-06-04
Build System: Try to find Z3 using CONFIG_MODE and stock cmake files before manual search.

ekpyron:
If Z3 is built using cmake it does produce its own cmake files which can be used to find it without a manual search - that's preferable, if available, and should help in https://github.com/ethereum/solidity/pull/6884, since the Z3 provided cmake files should be aware that the build is static and automatically declare pthreads as its dependency.

We still need to fall back to manually searching for Z3 for the case that there are no Z3 provided cmake configuration files.

PR #11873, opened 2021-08-31
Require latest Z3 by default and allow relaxing the requirement with `STRICT_Z3_VERSION=OFF`

soroosh-sdi:
Ref #11845 
- add --no-smt in scripts/tests.sh and test/cmdlineTests.sh

### DIFF
--- a/.circleci/build_win.ps1
+++ b/.circleci/build_win.ps1
@@ -18,7 +18,7 @@ else {
 mkdir build
 cd build
 $boost_dir=(Resolve-Path $PSScriptRoot\..\deps\boost\lib\cmake\Boost-*)
-..\deps\cmake\bin\cmake -G "Visual Studio 16 2019" -DBoost_DIR="$boost_dir\" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DCMAKE_INSTALL_PREFIX="$PSScriptRoot\..\upload" ..
+..\deps\cmake\bin\cmake -G "Visual Studio 16 2019" -DBoost_DIR="$boost_dir\" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DCMAKE_INSTALL_PREFIX="$PSScriptRoot\..\upload" -DUSE_Z3=OFF ..
 if ( -not $? ) { throw "CMake configure failed." }
 msbuild solidity.sln /p:Configuration=Release /m:5 /v:minimal
 if ( -not $? ) { throw "Build failed." }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ endif()
 
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
+option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
 
 # Setup cccache.
 include(EthCcache)
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 
 include(EthOptions)
 configure_project(TESTS)
+set(LATEST_Z3_VERSION "4.8.12")
+set(MINIMUM_Z3_VERSION "4.8.0")
+find_package(Z3)
+if (${Z3_FOUND})
+  if (${STRICT_Z3_VERSION})
+    if (NOT ("${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION}))
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use a different version. \
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+      )
+    endif()
+  else()
+    if ("${Z3_VERSION_STRING}" VERSION_LESS ${MINIMUM_Z3_VERSION})
+      message(
+        FATAL_ERROR
+        "Solidity requires Z3 ${MINIMUM_Z3_VERSION} or newer. You can also use -DUSE_Z3=OFF to build without Z3."
+      )
+    endif()
+  endif()
+endif()
 
-find_package(Z3 4.8.0)
 if(${USE_Z3_DLOPEN})
   add_definitions(-DHAVE_Z3)
   add_definitions(-DHAVE_Z3_DLOPEN)
--- a/docs/installing-solidity.rst
+++ b/docs/installing-solidity.rst
@@ -335,6 +335,16 @@ The following are dependencies for all builds of Solidity:
 
     Starting from 0.5.10 linking against Boost 1.70+ should work without manual intervention.
 
+.. note::
+    The default build configuration requires a specific Z3 version (the latest one at the time the
+    code was last updated). Changes introduced between Z3 releases often result in slightly different
+    (but still valid) results being returned. Our SMT tests do not account for these differences and
+    will likely fail with a different version than the one they were written for. This does not mean
+    that a build using a different version is faulty. If you pass ``-DSTRICT_Z3_VERSION=OFF`` option
+    to CMake, you can build with any version that satisfies the requirement given in the table above.
+    If you do this, however, please remember to pass the ``--no-smt`` option to ``scripts/tests.sh``
+    to skip the SMT tests.
+
 Minimum Compiler Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 


CMakeLists.txt
```diff
@@ -87,6 +87,9 @@ if(${USE_Z3_DLOPEN})
 elseif (${Z3_FOUND})
   add_definitions(-DHAVE_Z3)
   message("Z3 SMT solver found. This enables optional SMT checking with Z3.")
+  if("${Z3_VERSION}" VERSION_LESS "4.8.12.0")
+	message(WARNING "SMT tests need Z3 4.8.12 or higher, SMT tests are expected to fail!")
```
Marenz: missing indentation here

Marenz (changes requested):
The title of the PR and commit is misleading as the commit does a lot more. 
Otherwise, looks like a good change to me.

Please rename the commit and maybe separate the warning print into an extra commit?

CMakeLists.txt
Marenz: Ah, not missing, just used TABS when it's spaces.

soroosh-sdi: indentation for the line 91 looks good, but I used tab I will change it to 4 spaces.
and I should add a space after `if`

soroosh-sdi:
> The title of the PR and commit is misleading as the commit does a lot more.
> Otherwise, looks like a good change to me.
> 
> Please rename the commit and maybe separate the warning print into an extra commit?

yeah I wanted to point out the most important one. I'll change it.

Marenz (dismissed):

CMakeLists.txt
```diff
@@ -87,6 +87,9 @@ if(${USE_Z3_DLOPEN})
 elseif (${Z3_FOUND})
   add_definitions(-DHAVE_Z3)
   message("Z3 SMT solver found. This enables optional SMT checking with Z3.")
+  if ("${Z3_VERSION}" VERSION_LESS "4.8.12.0")
+    message(WARNING "SMT tests need Z3 4.8.12 or higher, SMT tests are expected to fail!")
+  endif()
```
cameel: Not sure if we want this warning in a general case. In the PR I mentioned a warning that would be issued only when your build includes `isoltest` and only if you have older Z3 installed but disabled with `-DUSE_Z3=OFF`.

@leonardoalt What do you think? Are we fine with a broad warning like this one?

scripts/tests.sh
```diff
@@ -52,17 +52,31 @@ cleanup() {
...
+                echo "Usage: $0 [--junit_report <report_directory>] [--no-smt]"
+                exit 1
+            else
+                log_directory="$2"
+            fi
+            shift
+            shift
+            ;;
+        --no-smt)
+            no_smt="--no-smt"
```
cameel: We already have a variable for SMT arguments.
```suggestion
            SMT_FLAGS+=(--no-smt)
```

cameel: Ah, I see you're passing it down to `cmdlineTests.sh` too. Then I'd do this:
```suggestion
            no_smt="--no-smt"
            SMT_FLAGS+=(--no-smt)
```
I think it's better to have all SMT-related flags for soltest in `SMT_FLAGS`.

test/cmdlineTests.sh
```diff
@@ -38,7 +38,20 @@ source "${REPO_ROOT}/scripts/common.sh"
 source "${REPO_ROOT}/scripts/common_cmdline.sh"
 
 AUTOUPDATE=false
-[[ $1 == --update ]] && AUTOUPDATE=true && shift
+no_smt=false
+while [[ $# -gt 0 ]]
```
cameel: If we do this, we can drop the part later in this file that reports an error when `--update` is mixed in with other arguments (line 313).

test/cmdlineTests.sh
```diff
@@ -38,7 +38,20 @@ source "${REPO_ROOT}/scripts/common.sh"
...
+do
+    case "$1" in
+        --update)
+            AUTOUPDATE=true
```
cameel: If you're using lowercase for your new variable, please also adjust the other one:
```suggestion
            autoupdate=true
```

test/cmdlineTests.sh
```diff
@@ -312,6 +325,13 @@ printTask "Running general commandline tests..."
 
         # Strip trailing slash from $tdir.
         tdir=$(basename "${tdir}")
+        if [[ $no_smt == true ]]
+        then
+            if [[ "$tdir" =~ .*model_checker.* ]]; then
```
cameel: I'd include the underscore in the condition to make sure we only match full words.
```suggestion
            if [[ $tdir =~ .*model_checker_.* ]]; then
```

Also, I think quotes are not needed on the left side inside `[[ ]]`.

cameel (changes requested):
Looks good overall, but I have a few cleanup suggestions.

Also, not sure about the warning in CMake.

cameel:
@soroosh-sdi Have you managed to find a way to require Z3 4.8.12 just for soltest?

soroosh-sdi:
> @soroosh-sdi Have you managed to find a way to require Z3 4.8.12 just for soltest?

I worked on it but I got lost. I couldn't solve the following obstacles:
 - why does somebody keep two versions of libz3-dev? It is not supported by package managers like apt.
 - soltest and solc have common dependencies like smtutil and if I wanted to use different z3 version I should define someting like this:
  target_link_libraries(smtutil-oldz3 PUBLIC z3::libz3-old) 
  target_link_libraries(smtutil-latestz3 PUBLIC z3::libz3-latest)
I don't know it's possible in cmake or not, even if we can I think it's over-complication for such a temporary issue. ( I assume we finally stop supporting old versions )
- command-line tests directly execute solc and some tests will fail.

CMakeLists.txt
soroosh-sdi: I can make it more precise and print the warning if TESTS are ON.

scripts/tests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: Ah, cmdlineTests get test names as argument. In the first version I didn't support that. So I add another variable, `tdirs`, to store list of received test names.
by the way I don't know how this piece of code: `${*:-*/}` works :)
please take another look.

cameel:
> why does somebody keep two versions of libz3-dev? It is not supported by package managers like apt.

No, it's not like that. We don't want to support building with multiple versions. It's just that older versions are acceptable for `solc` but not `soltest` and for such versions you should be able to build just `solc` successfully but building `soltest` should error out.

> command-line tests directly execute solc and some tests will fail.

That's fine. Someone building with older Z3 will be expected to use the `--no-smt` flag when running command-line tests. Just like you are expected to use it now with `soltest` if you don't have the latest Z3.

scripts/tests.sh
```diff
@@ -124,7 +139,7 @@ do
             [ "${vm}" = "byzantium" ] && [ "${optimize}" = "" ] && EWASM_ARGS="--ewasm"
 
             set +e
-            "${SOLIDITY_BUILD_DIR}"/test/soltest --show-progress "${log[@]}" -- ${EWASM_ARGS} --testpath "$REPO_ROOT"/test "$optimize" --evm-version "$vm" "${SMT_FLAGS[@]}" "${force_abiv1_flag[@]}"
+            "${SOLIDITY_BUILD_DIR}"/test/soltest --show-progress "${log[@]}" -- ${EWASM_ARGS} --testpath "$REPO_ROOT"/test "$optimize" --evm-version "$vm" "${SMT_FLAGS[@]}" "${force_abiv1_flag[@]}" "$no_smt"
```
cameel: You no longer need the new variable here. The flag is present in `$SMT_FLAGS`, which is already used here.
```suggestion
            "${SOLIDITY_BUILD_DIR}"/test/soltest --show-progress "${log[@]}" -- ${EWASM_ARGS} --testpath "$REPO_ROOT"/test "$optimize" --evm-version "$vm" "${SMT_FLAGS[@]}" "${force_abiv1_flag[@]}"
```

scripts/tests.sh
```diff
@@ -52,17 +52,32 @@ cleanup() {
...
-fi
+log_directory=""
+no_smt=""
+while [[ $# -gt 0 ]]
+do
+    case "$1" in
+        --junit_report)
+            if [ -z "$2" ]
+            then
+                echo "Usage: $0 [--junit_report <report_directory>] [--no-smt]"
```
cameel: Just noticed this. Does `$0` still hold the program name after you use `shift`?

test/cmdlineTests.sh
cameel: > by the way I don't know how this piece of code: `${*:-*/}` works :)

Magic :P See https://github.com/ethereum/solidity/pull/11712#discussion_r678289856.

test/cmdlineTests.sh
```diff
@@ -37,8 +37,26 @@ source "${REPO_ROOT}/scripts/common.sh"
 # shellcheck source=scripts/common_cmdline.sh
 source "${REPO_ROOT}/scripts/common_cmdline.sh"
 
-AUTOUPDATE=false
-[[ $1 == --update ]] && AUTOUPDATE=true && shift
+autoupdate=false
+no_smt=false
+declare -a tdirs
```
cameel:
```suggestion
declare -a selected_tests
```

test/cmdlineTests.sh
```diff
@@ -302,25 +320,28 @@ test_solc_behaviour "${0}" "ctx:=/some/remapping/target" "" "" 1 "" "Invalid rem
 printTask "Running general commandline tests..."
 (
     cd "$REPO_ROOT"/test/cmdlineTests/
-    for tdir in ${*:-*/}
+    if [[ ${#tdirs[@]} == 0 ]]
+    then
+        tdirs=(*)
+    fi
```
cameel: This can be shortened a bit.
```suggestion
    (( ${#tdirs[@]} >= 1 )) || tdirs=(*)
```

CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ if(${USE_Z3_DLOPEN})
 elseif (${Z3_FOUND})
   add_definitions(-DHAVE_Z3)
   message("Z3 SMT solver found. This enables optional SMT checking with Z3.")
+  if (TESTS AND NOT EMSCRIPTEN)
+    if ("${Z3_VERSION}" VERSION_LESS "4.8.12.0")
+      message(WARNING "SMT tests need Z3 4.8.12 or higher, SMT tests are expected to fail!")
```
cameel: It's not `Z3 4.8.12 or higher`. It's exactly 4.8.12. Newer versions won't work either.

```suggestion
      message(WARNING "SMTChecker tests require Z3 4.8.12 for all tests to pass. When using an older Z3 version you will need to use the --no-smt flag.")
```

cameel: Could you add a variable/constant for the required Z3 version?
```suggestion
    if ("${Z3_VERSION}" VERSION_LESS ${LATEST_Z3_VERSION})
      message(WARNING "SMT tests need Z3 ${LATEST_Z3_VERSION}, SMT tests are expected to fail!")
```

scripts/tests.sh
soroosh-sdi: yeah, only arguments are `shift `ed in the loop.

CMakeLists.txt
soroosh-sdi: > No, it's not like that. We don't want to support building with multiple versions. It's just that older versions are acceptable for `solc` but not `soltest` and for such versions you should be able to build just `solc` successfully but building `soltest` should error out.


I guess we can do that
```
if (TESTS AND NOT EMSCRIPTEN)
  finde_package(Z3,  4.8.12)
else
  finde_package(Z3,  4.8.0)
endif()
```

also I need to print an informative error message

scripts/tests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: done

soroosh-sdi:
> No, it's not like that. We don't want to support building with multiple versions. It's just that older versions are acceptable for `solc` but not `soltest` and for such versions you should be able to build just `solc` successfully but building `soltest` should error out.


I guess we can do that:
```
set(LATEST_Z3_VERSION "4.8.12")
if (TESTS AND NOT EMSCRIPTEN)
  find_package(Z3 ${LATEST_Z3_VERSION} REQUIRED)
  if (NOT ${Z3_FOUND})
    message(WARNING "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\
                   \nuse -DTESTS=OFF, if you want to use older version of z3")
  endif()
else()
  find_package(Z3 4.8.0)
endif()
```
 
It worked fine, even when I called it like this: `cmake ../ -DTESTS=ON -DUSE_Z3=OFF`
I didn't expect that :)

CMakeLists.txt
soroosh-sdi: done

soroosh-sdi: done

CMakeLists.txt
leonardoalt: @cameel Sorry I missed your message. Yea I think you're right that we shouldn't output this in the general case. I'll take a look at the current code.

cameel: @leonardoalt What do you think about what @soroosh-sdi proposed in https://github.com/ethereum/solidity/pull/11873#issuecomment-913024021?

I think that's actually pretty neat. I thought we'll have to make it depend on build targets but that actually solves the problem without too much magic.

~@soroosh-sdi One thing I'd change is that the warning should only be printed if the library is not found. `TESTS` is enabled by default so otherwise we'll always be seeing it.~ **EDIT**: I just realized that it's doing exactly that. Great!

leonardoalt: Yea I think that sounds good too. We only have to update that a couple times a year, so the benefit is bigger than the overhead I'd say.

cameel:
@soroosh-sdi I really like this solution. Much better than I expected. Can you modify the code to use it?

soroosh-sdi:
> @soroosh-sdi I really like this solution. Much better than I expected. Can you modify the code to use it?

yeah, sure.
Do you want me to remove `--no-smt` from those scripts too? I guess it could be helpful if somebody doesn't want to test SMT.
@cameel

soroosh-sdi:
Do we run soltest on window?
If not, we can disable TESTS on b_win and b_win_release

cameel:
> Do you want me to remove `--no-smt` from those scripts too? I guess it could be helpful if somebody doesn't want to test SMT.

The option is useful so let's keep it. It's still needed if you choose to build without Z3 and then run command line tests.

.circleci/build_win.ps1
```diff
@@ -18,7 +18,7 @@ else {
 mkdir build
 cd build
 $boost_dir=(Resolve-Path $PSScriptRoot\..\deps\boost\lib\cmake\Boost-*)
-..\deps\cmake\bin\cmake -G "Visual Studio 16 2019" -DBoost_DIR="$boost_dir\" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DCMAKE_INSTALL_PREFIX="$PSScriptRoot\..\upload" ..
+..\deps\cmake\bin\cmake -G "Visual Studio 16 2019" -DBoost_DIR="$boost_dir\" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DCMAKE_INSTALL_PREFIX="$PSScriptRoot\..\upload" -DUSE_Z3=OFF ..
```
cameel, line 21: I took a look at CI output and looks like we build with Z3 enabled but it's not installed so CMake does not find it. [`soltest.ps1`](https://github.com/ethereum/solidity/blob/develop/.circleci/soltest.ps1) runs tests with `--no-smt` anyway so it's not a problem.

Is this actually needed? I see nothing in your PR that would break because of it. We could add `-DUSE_Z3=OFF` here but it worked without it too.

cameel: BTW, @leonardoalt Is SMT disabled on Windows because of non-determinism? I guess it's the reason why it's only macOS that's breaking your PRs :)

CMakeLists.txt
```diff
@@ -63,8 +63,16 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 
 include(EthOptions)
 configure_project(TESTS)
-
-find_package(Z3 4.8.0)
+set(LATEST_Z3_VERSION "4.8.12")
+if (TESTS AND NOT EMSCRIPTEN)
```
cameel: Why `NOT EMSCRIPTEN`? Does this break the emscripten build?

.circleci/build_win.ps1
soroosh-sdi: we have another option, we can disable TESTS instead.

CMakeLists.txt
```diff
@@ -63,8 +63,16 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+  find_package(Z3 ${LATEST_Z3_VERSION} REQUIRED)
```
cameel: Won't `REQUIRED` cause an error if you don't have Z3 installed at all? We do want to support people without Z3. It's just that if it's installed, you need specifically 4.8.12.

.circleci/build_win.ps1
soroosh-sdi: b_win and b_win_release got failed, so I have to add this

CMakeLists.txt
cameel: I'd use `EXACT` here because tests will likely fail on a newer version too when it's released.

CMakeLists.txt
soroosh-sdi: I just add this because we add `tests` folder with that condition; if we don't build `test` directory  we don't need the latest version of z3.

CMakeLists.txt
```diff
@@ -63,8 +63,16 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+  if (NOT ${Z3_FOUND})
+    message(WARNING "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\
+                   \nuse -DTESTS=OFF, if you want to use older version of z3")
```
cameel: I think that the whitespace before `\n` will be included in the message, which could make it wrap in a weird way on a very narrow terminal.

I'd indent and wrap it this way:
```suggestion
    message(
      WARNING
      "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
Build with -DTESTS=OFF if you want to use an older version. \
You can also use -DUSE_Z3=OFF to build without Z3 and then use --no-smt when running tests."
    )
```

.circleci/build_win.ps1
cameel: But what was the error? Maybe it's unrelated?

CMakeLists.txt
soroosh-sdi: > It's just that older versions are acceptable for `solc` but not `soltest` and for such versions you should be able to build just `solc` successfully but building `soltest` should error out.

yeah, that's what you told me earlier.
Actually, I want to inform users to disable TESTS

soroosh-sdi: I'll fix it

CMakeLists.txt
soroosh-sdi: good suggestion.

.circleci/build_win.ps1
leonardoalt: I think it's disabled on windows because it's annoying to install (mainly).

soroosh-sdi: > But what was the error? Maybe it's unrelated?

It was failed because of my changes in cmake, the error was:
```
CMake Error at deps/cmake/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find Z3 (missing: Z3_LIBRARY Z3_INCLUDE_DIR) (Required is at
  least version "4.8.12")
  deps/cmake/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindZ3.cmake:39 (find_package_handle_standard_args)
  CMakeLists.txt:68 (find_package)
```

CMakeLists.txt
cameel: ok. It's good then. Perhaps `EMSCRIPTEN=ON`+ `TEST=ON` should just be an error but if it isn't then I guess this condition makes sense.

I'd just add a comment above the condition explaining this: "NOTE: With EMSCRIPTEN=ON we do not build code from test/ dir and that code is our only reason to require latest Z3."

CMakeLists.txt
cameel: ok, I meant that building `soltest` should error out with older Z3. But without Z3 (or with `-DUSE_Z3=OFF`) it should still build.

CMakeLists.txt
soroosh-sdi: OK, I'll add the comment to make it clear.

.circleci/build_win.ps1
cameel: Ah, so it's exactly what I pointed out in https://github.com/ethereum/solidity/pull/11873#discussion_r705408383 :)

Without `REQUIRED` it should work.

cameel: > I think it's disabled on windows because it's annoying to install (mainly).

@leonardoalt Should I create an issue to add it eventually? Or are there other reasons why we might not want it? E.g. performance or that it might not add all that much value given we run on all other platforms?

CMakeLists.txt
soroosh-sdi: you only want to print a warning and build the tests even when we don't have latest z3. so I'll remove `REQUIRED`.

cameel: Ah, I see where the problem is. I thought that with older Z3 installed cmake would report an error even without `REQUIRED` and that the only thing that `REQUIRED` changed was the case where Z3 is not installed at all. But looks like I was wrong. It continues and just does not use the older version.

What I wanted to achieve is this:
- With `TEST=OFF`
    - Z3 missing -> continue without Z3
    - Z3 < 4.8.0 -> error (but looks like currently it continues without Z3 so it's fine to leave it at that)
- With `TEST=ON`
    - Z3 missing -> continue without Z3
    - Z3 < 4.8.12 -> error

cameel: I think that in the `TEST=ON` + Z3 < 4.8.12 case printing a warning and continuing without Z3 would be fine. But we should not report an error just because Z3 is not installed. It's not required to run tests.

soroosh-sdi: with `REQUIRED` :
> * With `TEST=ON`
>   * Z3 < 4.8.12 -> error

without `REQUIRED` :
* With `TEST=ON`
   * Z3 < 4.8.12 -> continue without Z3 and print a warning

I like the `REQUIRED` version because it is more clear to me. but it is a problem because it will change the default behavior. b_win is an example. I guess we have a lot of users that build solidity without z3 and without `TEST=OFF`.

cameel: So how about this: use `find_package()` with 4.8.0 as a requirement in all cases. But then use `if`s to check if "tests are enabled" + "Z3 was found"  + "Z3 version is not 4.8.12". If all of these are true, manually report an error.

I think this would handle missing Z3 well and still give us an error if Z3 is installed.

cameel:
@soroosh-sdi https://github.com/ethereum/solidity/pull/11873#discussion_r705542022 seems to be the last remaining thing to resolve so after you do this, please squash your commits and rebase on `develop`.

CMakeLists.txt
soroosh-sdi: good idea

CMakeLists.txt
```diff
@@ -63,8 +63,21 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
 include(EthOptions)
 configure_project(TESTS)
-
+set(LATEST_Z3_VERSION "4.8.12")
 find_package(Z3 4.8.0)
+# NOTE: With EMSCRIPTEN=ON we do not build code from test/ dir and that
+# code is our only reason to require latest Z3.
+if (TESTS AND NOT EMSCRIPTEN)
+  if (${Z3_FOUND})
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION} ) )
```
soroosh-sdi: > Xxx_VERSION
The full version string of the package found, if any. Note that many existing modules provide Xxx_VERSION_STRING instead.

I tested Z3_VERSION and Z3_VERSION_STRING with different z3 packages that provided by ethereum ppa and ubuntu repo. The ubuntu version only provides  Z3_VERSION_STRING.

cameel: Let's use `Z3_VERSION_STRING` then. It seems to be the variable that [`Z3Config.cmake` generated by Z3](https://github.com/Z3Prover/z3/blob/master/cmake/Z3Config.cmake.in) sets and our custom [`FindZ3.cmake`](https://github.com/ethereum/solidity/blob/develop/cmake/FindZ3.cmake) also sets that variable consistently, unlike `Z3_VERSION`, which it sets only when the stock find script manages to find Z3 on its own.

I actually don't know why we don't consistently set `Z3_VERSION`. Maybe @ekpyron could shed some light here - he added that part in #6891. I think the expectation might have been that it will be set automatically due to CMake's [project()](https://cmake.org/cmake/help/latest/command/project.html) but then I don't understand why we set it when the stock script finds Z3 rather than in the opposite case.

```suggestion
if (
  TESTS AND
  NOT EMSCRIPTEN AND
  ${Z3_FOUND} AND
  NOT ("${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION})
)
```

CMakeLists.txt
```diff
@@ -63,8 +63,21 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DTESTS=OFF if you want to use an older version. \
+You can also use -DUSE_Z3=OFF to build without Z3 and then use --no-smt when running tests.")
```
```suggestion
You can also use -DUSE_Z3=OFF to build without Z3 and then use --no-smt when running tests."
      )
```

CMakeLists.txt
cameel: Also, some variables are quoted, some are not.

cameel (commented):
There are some minor formatting issues but other than that the PR seems ready.

I think that @leonardoalt and @ekpyron should give it a a final look.

CMakeLists.txt
soroosh-sdi: It seems Z3_VERSION and Z3_VERSION_STRING can be null even when z3 is found. so I quoted this variable to be more robust against any upcoming changes in z3's find module.
Other variables cannot be null.

CMakeLists.txt
```diff
@@ -63,8 +63,21 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
 find_package(Z3 4.8.0)
+# NOTE: With EMSCRIPTEN=ON we do not build code from test/ dir and that
+# code is our only reason to require latest Z3.
+if (TESTS AND NOT EMSCRIPTEN)
+  if (${Z3_FOUND})
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION} ) )
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DTESTS=OFF if you want to use an older version. \
```
ekpyron: I don't think that's a valid suggestion and not having the latest z3 version, but wanting to run tests is a valid thing to do, so it is no reason for a fatal error (you most definitely should be able to run e.g. ``isoltest --no-smt`` if you don't have the latest Z3 version).

I would suggest to just issue a warning about the Z3 version to be too old for all tests to pass and leave it at that.

test/cmdlineTests.sh
```diff
@@ -302,25 +320,25 @@ test_solc_behaviour "${0}" "ctx:=/some/remapping/target" "" "" 1 "" "Invalid rem
 printTask "Running general commandline tests..."
 (
     cd "$REPO_ROOT"/test/cmdlineTests/
-    for tdir in ${*:-*/}
+    (( ${#selected_tests[@]} == 0 )) || selected_tests=(*)
```
ekpyron: What's the purpose of this? If there are no selected tests, select all tests?
Shouldn't it be ``(( ${#selected_tests[@]} == 0 )) && selected_tests=(*)`` or ``(( ${#selected_tests[@]} > 0 )) || selected_tests=(*)`` for that?

ekpyron (dismissed):
As a matter of fact - looking at the CI run - due to https://github.com/ethereum/solidity/pull/11873#discussion_r706090278 this effectively disables all general command line tests in CI ;-).

CMakeLists.txt
ekpyron: If we want to be more fancyful, we could set a flag ``OLD_Z3_VERSION`` and have ``soltest`` and ``isoltest`` error out on running themselves without ``--no-smt``.

ekpyron: The command line test change is independent of this here anyways, though, so maybe we should split this up into two separate PRs anyways, i.e. reduce this PR to only change the command line tests and then we can separately decide whether any of this should in fact be an error and if so, then how exactly.

test/cmdlineTests.sh
soroosh-sdi: Ah, yeah it should be `&&` instead of `||`.

CMakeLists.txt
soroosh-sdi: > I don't think that's a valid suggestion and not having the latest z3 version, but wanting to run tests is a valid thing to do, so it is no reason for a fatal error (you most definitely should be able to run e.g. `isoltest --no-smt` if you don't have the latest Z3 version).
> 
> I would suggest to just issue a warning about the Z3 version to be too old for all tests to pass and leave it at that.

whole point of this PR is to notify the user about the test's requirement to latest z3. I think its better to notify the user as soon as possible, we can change the `FATAL_ERROR` to `WARNING` here and also pass some flag to `soltest` and `isoltest` to error out as you said.

soroosh-sdi: > The command line test change is independent of this here anyways, though, so maybe we should split this up into two separate PRs anyways, i.e. reduce this PR to only change the command line tests and then we can separately decide whether any of this should in fact be an error and if so, then how exactly.

good idea, I'll create another PR for that.

cameel: > not having the latest z3 version, but wanting to run tests is a valid thing to do

Technically yes, but I think that we're better off forcing people to use a specific version of Z3 in this case. In the cases I've seen where we got reports from people getting test failures because of older Z3 they always wanted to run the full suite and were fine with updating Z3 but not with `--no-smt`. They simply did not know that the problems were due to Z3 version. I completely agree with @soroosh-sdi that an error at build stage makes the situation much clearer and easier to resolve on their own for these people.

I suspect that the set of people who care about tests but not Z3 tests and still want to built with Z3 is pretty small. I think that it's more likely that they either do not care about SMTChecker at all (and would be fine with `-DUSE_Z3=OFF`) or do not care about tests at all (and would be fine with `-DTESTS=OFF`).

> good idea, I'll create another PR for that.

Yeah, please create a separate PR for the `--no-smt` flag and rebase this one on top of it. The `--no-smt` change seems to be much less controversial so it'll likely go in without much discussion.

test/cmdlineTests.sh
cameel: > What's the purpose of this? If there are no selected tests, select all tests?

Yes. This is the original behavior of `cmdlineTests.sh` if there are no arguments.

> Shouldn't it be `(( ${#selected_tests[@]} == 0 )) && selected_tests=(*)` or `(( ${#selected_tests[@]} > 0 )) || selected_tests=(*)` for that?

Good catch! My original suggestion (https://github.com/ethereum/solidity/pull/11873#discussion_r702162319) was for `${#selected_tests[@]} >= 1` and then I did not notice in the review that it was different.

CMakeLists.txt
cameel: I discussed it with @ekpyron on the chat. Here are our conclusions:
1. A build-time check is not enough because you can build with Z3 linked dynamically and then the version used at build time does not matter - you can still get failing tests if you use such a binary with older dynanamic Z3 installed in your system. We need a runtime check in soltest/isoltest that prints an error when `--no-smt` is not used.
    - The build-time check against 4.8.0 is valid because it enforces a version that's still ABI-compatible. That's not the case with the check for latest version, which is rather arbitrary.
2. Turns out there are valid reasons to want older Z3. This changes my view on this check because I thought there's no reason not to always use the latest version. So we should support building with tests enabled and older Z3.
    - But we both agreed that we do not have to make that case the default. We're fine with making it even more stringent. We could require Z3 4.8.12 by default and just add a CMake option that allows the user to override it.
3. We do support `-DTESTS=OFF` but we do not want to encourage users to use it so this part should be removed from the warning.

@soroosh-sdi So here's my suggestion on how to proceed here:
- Create a separate PR for `--no-smt` and we can merge that right away.
- Modify this PR to add a CMake option called `STRICT_Z3_VERSION`.
    - It should be `ON` by default. When it's on, the requirement is **exactly** 4.8.12, otherwise **at least** 4.8.0. And it can be a general requirement, not just for tests (you can drop the `TESTS` check).
        - I think it's not a breaking change but probably should be mentioned in the changelog? @chriseth?
    - In both cases finding an older version should be an error. User should just use `-DUSE_Z3=OFF` if he wants to ignore the installed version. This will require adding back the change you had for Windows CI. This is different than the current behavior (older version right now is treated as if it was not installed at all) but I think it will be better because it's more explicit.
- Let's add the runtime check in `soltest`/`isoltest` in a new PR.

CMakeLists.txt
cameel: So going with https://github.com/ethereum/solidity/pull/11873#discussion_r706342315, we need to change it to:
```suggestion
        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
If you want to build without Z3, pass -DUSE_Z3=OFF to cmake and later use --no-smt when running tests.\n\
If you are aware of potential issues and still want use a different Z3 version, use -DSTRICT_Z3_VERSION=OFF to override this error."\
      )
```

cameel: Oh, this should also mention the `-DSTRICT_Z3_VERSION` flag. Suggestion updated.

CMakeLists.txt
cameel: Also there's a question if we should still require 4.8.12 when building with dynamic Z3 (i.e. when `USE_Z3_DLOPEN` is enabled). I think that an error in that case is still useful because the user is likely to still run tests in the same system that was used to build the compiler (even if the compiler executable will later be also used on other systems).

ekpyron: An interesting question is also what to do if the Z3 version is *too new* (i.e. we haven't yet updated to the latest version or somebody uses a pre-release build from the git tree or somebody builds an old compiler version against more recent Z3). That's just as likely to break the tests as an older version is. So it should probably also require ``-DSTRICT_Z3_VERSION=OFF``.
EDIT: I just see that the PR already uses ``VERSION_EQUAL``, so that's indeed what we want - I'm just re-mentioning it since ``When it's on, the requirement is 4.8.12, otherwise 4.8.0`` is not completely clear in that the requirement will be ``exactly 4.8.12`` with the flag on and ``at least 4.8.0`` with the flag off.
EDIT: Ok, now it's perfectly clear ;-).

cameel: Yes. I agree. This is exactly why I did not name the option `OLD_Z3_VERSION` as you suggested above, but looks like I forgot to mention that the check for 4.8.12 should be an exact comparison. Thanks for pointing this out!

cameel: Description updated.

soroosh-sdi:
@cameel 
Please take a look.

CMakeLists.txt
soroosh-sdi: done

test/cmdlineTests.sh
soroosh-sdi: solved in #11933

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
```diff
@@ -63,8 +64,21 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+    if (${USE_Z3} AND NOT ${Z3_FOUND})
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+            )
+    endif()
+else()
+    find_package(Z3 4.8.0 REQUIRED)
```
cameel: Well, we did not say we want an error with `STRICT_Z3_VERSION=OFF` when Z3 is missing. Only that we want an error when it's present but too old.

But I'd personally be fine with this. Requiring `-DUSE_Z3=OFF` when you do not have Z3 seems reasonable to me. @leonardoalt what do you think?

CMakeLists.txt
```diff
@@ -63,8 +64,21 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+endif()
```
cameel: Indents are a bit off. This file uses 2 spaces so this code should too:

```suggestion
if (${STRICT_Z3_VERSION})
  find_package(Z3 ${LATEST_Z3_VERSION} EXACT)
  if (${USE_Z3} AND NOT ${Z3_FOUND})
    message(
      FATAL_ERROR
        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
    )
  endif()
else()
  find_package(Z3 4.8.0 REQUIRED)
endif()
```
Though we should really reformat it with tabs to match our [style](https://github.com/ethereum/solidity/blob/develop/.editorconfig#L21-L23) at some point. 2 space indents are just awful for readability.

cameel (commented):
Looks fine, just needs small whitespace tweaks.

Please also add a changelog entry that mentions the new requirement.

We still need someone to take a final look at it (@leonardoalt? @chriseth?). Especially regarding the requirement for Z3 to be installed if you do not set `-DUSE_Z3=OFF` because we did not discuss such a variant originally.

CMakeLists.txt
leonardoalt: Is that the default though?
I'd say it should work fine if there's no z3 available by default.

cameel:
Also, this needs rebase on latest `develop` because it's currently on top of my branch with a pylint fix (which has already been merged) and those changes show up in the diff.

CMakeLists.txt
cameel: `-DUSE_Z3` is `ON` by default so without Z3 you would get an error unless you set it to `OFF`.

soroosh-sdi: > Well, we did not say we want an error with `STRICT_Z3_VERSION=OFF` when Z3 is missing. Only that we want an error when it's present but too old.
> 
@cameel 
You're right, that way we would have two different behaviors in case of setting the `STRICT_Z3_VERSION` to off and on:
off :: if z3 is not installed we error out. If it's installed it must be ='4.8.12'
on :: if z3 is not installed it's ok. if it's installed it must be >='4.8.0'
So I implemented  it this way:
off :: if z3 is not installed we error out, If it's installed it must be ='4.8.12'
on :: if z3 is not installed we error out, if it's installed it must be >='4.8.0'

I see @leonardoalt  prefers the following:
off :: if z3 is not installed it's ok. If it's installed it must be ='4.8.12'
on :: if z3 is not installed it's ok. if it's installed it must be >='4.8.0'

I'm ok with the solution as long as we have same behavior when z3 is not installed in both cases of STRICT_Z3_VERSION is on or off.

cameel: There's still a problem with the current version:
```cmake
if (${STRICT_Z3_VERSION})
  find_package(Z3 ${LATEST_Z3_VERSION} EXACT)
  if (${USE_Z3} AND NOT ${Z3_FOUND})
    message(
      WARNING
      "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
          )
  endif()
else()
  find_package(Z3 4.8.0)
endif()
```

We actually have 3 cases, not 2 and I think that this code will not produce an error in cases 1.2 and 2.2 below:
1. `STRICT_Z3_VERSION=ON`
    1. Z3 not installed: OK or ERROR (I'm fine with both, @leonardoalt prefers OK)
    2. Z3 != 4.8.12: ERROR
    3. Z3 == 4.8.12: OK
2. `STRICT_Z3_VERSION=OFF`
    1. Z3 not installed: OK or ERROR (I'm fine with both, @leonardoalt prefers OK)
    2. Z3 < 4.8.0: ERROR
    3. Z3 >= 4.8.0: OK

It will give us a warning in 1.2 but will still let us continue.

soroosh-sdi: I don't follow,  I didn't call find_package with `REQUIRED` so I think both cases 1.2 and 2.2 does not error out.
Maybe you thought find_package gives error by default, but it only gives error and stop the configuration process if we call it with `REQUIRED`.

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 
 include(EthOptions)
 configure_project(TESTS)
+set(LATEST_Z3_VERSION "4.8.12")
+set(LOWEST_Z3_VERSION_SUPPORTED "4.8.0")
```
cameel: We could shorten this name a bit:
```suggestion
set(MINIMUM_Z3_VERSION "4.8.0")
```

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+            )
+    endif()
+  else()
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_GREATER_EQUAL ${LOWEST_Z3_VERSION_SUPPORTED} ) )
+      message(
+        FATAL_ERROR
+        "Solidity requires Z3 ${LOWEST_Z3_VERSION_SUPPORTED} or greater. You can also use -DUSE_Z3=OFF to build without Z3."
```
```suggestion
        "Solidity requires Z3 ${LOWEST_Z3_VERSION_SUPPORTED} or newer. You can also use -DUSE_Z3=OFF to build without Z3."
```

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+set(LATEST_Z3_VERSION "4.8.12")
+set(LOWEST_Z3_VERSION_SUPPORTED "4.8.0")
+find_package(Z3)
+if (${USE_Z3} AND ${Z3_FOUND})
+  if (${STRICT_Z3_VERSION})
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION} ) )
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
```
cameel: This applies to newer versions as well:
```suggestion
Build with -DSTRICT_Z3_VERSION=OFF if you want to use a different version. \
```

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+            )
+    endif()
+  else()
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_GREATER_EQUAL ${LOWEST_Z3_VERSION_SUPPORTED} ) )
```
cameel: This could be more concise:
```suggestion
    if ("${Z3_VERSION_STRING}" VERSION_LESS ${LOWEST_Z3_VERSION_SUPPORTED}))
```

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+find_package(Z3)
+if (${USE_Z3} AND ${Z3_FOUND})
+  if (${STRICT_Z3_VERSION})
+    if ( NOT ( "${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION} ) )
+      message(
+        FATAL_ERROR
+        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+            )
```
cameel: Our convention is to line up the closing parenthesis with the beginning of the line that contains the opening parenthesis. Not with the opening parenthesis itself. This makes it easier to see at what nesting level the current statement is. In this particular case the string content slightly breaks the flow but I think this makes it even more important to have it positioned properly:
```suggestion
    if (NOT ("${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION}))
      message(
        FATAL_ERROR
        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
Build with -DSTRICT_Z3_VERSION=OFF if you want to use an older version. \
You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
      )
```
Same in the other message below.

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
 
 include(EthOptions)
 configure_project(TESTS)
+set(LATEST_Z3_VERSION "4.8.12")
+set(LOWEST_Z3_VERSION_SUPPORTED "4.8.0")
+find_package(Z3)
+if (${USE_Z3} AND ${Z3_FOUND})
```
cameel: `find_package()` already checks `USE_Z3` so you can reduce this to:
```suggestion
if (${Z3_FOUND})
```

cameel (commented):
The logic seems good now. I have some stylistic comments but other than that it seems to be exactly what we wanted.

@leonardoalt can you take a final look?

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
soroosh-sdi: done

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
+Build with -DSTRICT_Z3_VERSION=OFF if you want to use a different version. \
+You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
+      )
+    endif()
+  else()
+    if ("${Z3_VERSION_STRING}" VERSION_LESS ${MINIMUM_Z3_VERSION})
+      message(
+        FATAL_ERROR
+        "Solidity requires Z3 ${MINIMUM_Z3_VERSION} or newer. You can also use -DUSE_Z3=OFF to build without Z3."
+             )
```
```suggestion
      message(
        FATAL_ERROR
        "Solidity requires Z3 ${MINIMUM_Z3_VERSION} or newer. You can also use -DUSE_Z3=OFF to build without Z3."
      )
```

cameel (dismissed):
I'm approving because I think it's done but since this was a bit controversial, I think this still needs approvals from @leonardoalt and @ekpyron before we merge it.

CMakeLists.txt
soroosh-sdi: done

soroosh-sdi:
> I'm approving because I think it's done but since this was a bit controversial, I think this still needs approvals from @leonardoalt and @ekpyron before we merge it.

Thanks, 
I squashed all commit and rebased it on the latest version of develop.

cameel (dismissed):

cameel:
Actually, there's one more thing. We have version requirements documented in [Building from Source > Prerequisites - All Operating Systems](https://docs.soliditylang.org/en/latest/installing-solidity.html#prerequisites-all-operating-systems). This needs to be modified to say that without `STRICT_Z3_VERSION=OFF` the compiler will expect the latest version (but do not mention that it's specifically 4.8.12 right now).

soroosh-sdi:
> Actually, there's one more thing. We have version requirements documented in [Building from Source > Prerequisites - All Operating Systems](https://docs.soliditylang.org/en/latest/installing-solidity.html#prerequisites-all-operating-systems). This needs to be modified to say that without `STRICT_Z3_VERSION=OFF` the compiler will expect the latest version (but do not mention that it's specifically 4.8.12 right now).

I change the table as follow:
```
+-----------------------------------+-------------------------------------------------------+
| `z3`_ (latest version, Optional)  | For use with SMT checker.                             |
+-----------------------------------+-------------------------------------------------------+
```
and I can add a note like this:
```
.. note::
    The latest version of Z3 is required to pass all tests, if you don't want to run tests you can
    relax this requirement by calling cmake with `-DSTRICT_Z3_VERSION=OFF` and use Z3 version 4.8+.
```

cameel:
> `z3`_ (latest version, Optional)

The minimum requirement is still 4.8.0 so I'd leave the table unchanged. Instead I'd add only the note:

```rst
.. note::
    The default build configuration requires a specific Z3 version (the latest one at the time the code was last updated).
    Changes introduced between Z3 releases often result in slightly different (but still valid) results being returned.
    Our SMT tests do not account for these differences and will likely fail with a different version than the one they were written for.
    This does not mean that a build using a different version is faulty.
    If you pass ``-DSTRICT_Z3_VERSION=OFF`` option to CMake, you can build with any version that satisfies the requirement given in the table above.
    If you do this, however, please remember to pass the ``--no-smt`` option to ``scripts/tests.sh`` to skip the SMT tests.
```

soroosh-sdi:
> ```
> .. note::
>     The default build configuration requires a specific Z3 version (the latest one at the time the code was last updated).
>     Changes introduced between Z3 releases often result in slightly different (but still valid) results being returned.
>     Our SMT tests do not account for these differences and will likely fail with a different version than the one they were written for.
>     This does not mean that a build using a different version is faulty.
>     If you pass ``-DSTRICT_Z3_VERSION=OFF`` option to CMake, you can build with any version that satisfies the requirement given in the table above.
>     If you do this, however, please remember to pass the ``--no-smt`` option to ``scripts/tests.sh`` to skip the SMT tests.
> ```

it's 200% better than my version, thank you.

cameel (approved):

soroosh-sdi:
this PR hasn't merged yet, please give me feed back. @cameel

leonardoalt:
Sorry for the delay on this, I'll take a look today.

CMakeLists.txt
```diff
@@ -63,8 +64,29 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/templates/license.h.in" include/licens
...
 include(EthOptions)
 configure_project(TESTS)
+set(LATEST_Z3_VERSION "4.8.12")
+set(MINIMUM_Z3_VERSION "4.8.0")
+find_package(Z3)
+if (${Z3_FOUND})
+  if (${STRICT_Z3_VERSION})
+    if (NOT ("${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION}))
+      message(
+        FATAL_ERROR
```
leonardoalt, line 74: I'm afraid this might be too harsh but I'm also fine with it. We do warn if no solver is found so yea, fine.

leonardoalt (approved):

PR #11874, opened 2021-09-01
Fix UBSan CI

bshastry:
CMake: Remove unsigned-integer-overflow check from UBSan compiler flags because it is too noisy.

This should fix the UBSan CLI failing run that is currently (incorrectly) failing due to noisy prints originating from the check disabled by this PR.

The C++ standard does not officially consider unsigned integer overflows to be undefined behavior since it is implementation independent.

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -200,6 +200,10 @@ if (SANITIZE)
 	elseif (sanitizer STREQUAL "undefined")
 		# The following flags not used by fuzzer but used by us may create problems, so consider
 		# disabling them: alignment, pointer-overflow.
+		# The following flag is not used by us to reduce terminal noise
+		# i.e., warnings printed on stderr: unsigned-integer-overflow
+		# Note: The C++ standard does not officially consider unsigned integer overflows
+		# to be undefined behavior since they are implementation independent.
 		# Flags are alphabetically sorted and are for clang v10.0
 		list(APPEND undefinedSanitizerChecks
 			alignment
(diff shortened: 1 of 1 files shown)


cmake/EthCompilerSettings.cmake
```diff
@@ -200,6 +200,8 @@ if (SANITIZE)
 	elseif (sanitizer STREQUAL "undefined")
 		# The following flags not used by fuzzer but used by us may create problems, so consider
 		# disabling them: alignment, pointer-overflow.
+		# The following flag is not used by us to reduce terminal noise
+		# i.e., warnings printed on stderr: unsigned-integer-overflow
```
cameel, line 204: > The C++ standard does not officially consider unsigned integer overflows to be undefined behavior since it is implementation independent.

I think it would be good to include this extra bit of info here.

cameel (dismissed):

Marenz (dismissed):
I'd title the commit something like "Fix UBSan CI" or so (I think commit titles should describe the motivation, not the content). But I won't insist on this.

cmake/EthCompilerSettings.cmake
bshastry: Added a comment.

bshastry:
> I'd title the commit something like "Fix UBSan CI" or so (I think commit titles should describe the motivation, not the content). But I won't insist on this.

Done!

cameel (approved):
The failing `b_archlinux` CI job seems unrelated to the PR (timing out while downloading packages).

PR #11880, opened 2021-09-01
CircleCI config cleanup and deduplication

cameel:
Closes #11868.

This is code from #11868 submitted as a new non-external PR so that we can run full CI checks on it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,25 @@ defaults:
       name: Correctness proofs for optimization rules
       command: scripts/run_proofs.sh
 
+  - run_soltest: &run_soltest
+      name: soltest
+      no_output_timeout: 30m
+      command: ./.circleci/soltest.sh
+
+  - run_soltest_all: &run_soltest_all
+      name: soltest_all
+      no_output_timeout: 30m
+      command: ./.circleci/soltest_all.sh
+
+  - run_cmdline_tests: &run_cmdline_tests
+      name: command line tests
+      no_output_timeout: 30m
+      command: ./test/cmdlineTests.sh
+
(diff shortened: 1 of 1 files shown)


cameel:
OK, this now passes tests. I need someone to review and approve. I can't approve myself since I created the PR (even though the commits are not mine).

.circleci/config.yml
```diff
@@ -813,25 +804,19 @@ jobs:
     docker:
       - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
     steps:
-      - checkout
-      - attach_workspace:
-          at: build
-      - run: *run_soltest
+      - when:
+          condition: true
+          <<: *steps_soltest
```
ekpyron, line 809: This I don't quite understand... is this just because ``<<: *steps_soltest`` on the outermost level would be invalid for some reason?

ekpyron (approved):
Looks reasonable to me.

.circleci/config.yml
cameel: The reason is that putting it at the same level as `steps` would make it completely ovewrite `steps` rather than add keys to it.

I tried to explain it in one of the recent PRs: https://github.com/ethereum/solidity/pull/11823#discussion_r696042121

ekpyron: That makes sense, but there is no other ``steps`` key here, is there? So what's the problem with overwriting what does not exist :-)? But I may be missing something and I don't mind in any case :-).

ekpyron: Ah sorry, I just missed the ``- run: *gitter_notify_failure`` thing below, so indeed makes perfect sense.

leonardoalt (approved):

PR #11881, opened 2021-09-01
Properly export symbols from aliased imports.

### LINKED ISSUE #11855: Importing an aliased import results in unusable symbol
## Description

An aliased import can be imported into other files but subsequently is not available for use.  I find this confusing.  Either the import should fail or it should succeed, with the symbol being available for use.

## Environment

- Compiler version: 0.6.11
- Target EVM version (as per compiler settings): London
- EVM execution environment / backend / blockchain client:  N/A
- Operating system: OS X

## Steps to Reproduce

Here are three contracts:

1. a dummy contract to refer to
2. a file that imports the dummy contract with an alias
3. another file that imports the alias from the previous file and tries to use it

```solidity
// Dummy.sol
pragma solidity 0.6.11;

contract Dummy {
    string public constant FOO = "FOO";
}

// HasAlias.sol
pragma solidity 0.6.11;

import {Dummy as AliasedDummy} from "./Dummy.sol";

// UseAlias.sol
pragma solidity 0.6.11;

import {AliasedDummy} from "./HasAlias.sol";

contract TestAlias is AliasedDummy {}
```

Output:
```
contracts/UseAlias.sol:5:23: DeclarationError: Identifier not found or not unique.
contract TestAlias is AliasedDummy {}
                      ^----------^

Error HH600: Compilation failed
```

**Note:** removing line 5 of `UseAlias.sol` results in successful compilation!

## Expected Result and Discussion

I expect the identifier `AliasedDummy` to be found, as compilation succeeds without line 5 in `UseAlias.sol`.  The resulting behavior seems to be that the alias can be imported without error but cannot be used.  This is confusing.  Either the symbol doesn't exist in `UseAlias.sol` so it can't be imported, or it does exist and can be imported for use.  There should not be an "in-between" state of being able to import symbols that are then unavailable for use.  (this is somewhat reminiscent of Javascript, but even the case of Javascript, the imported symbol exists with the value of `undefined`).

Marenz:
fixes #11855

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Name Resolver: Fix that when importing an aliased symbol using ``import {AliasedName} from "a.sol"`` it would use the original name of the symbol and not the aliased one.
  * SMTChecker: Fix false negative caused by ``push`` on storage array references returned by internal functions.
  * SMTChecker: Fix false positive in external calls from constructors.
  * SMTChecker: Fix internal error on some multi-source uses of ``abi.*``, cryptographic functions and constants.
--- a/libsolidity/analysis/NameAndTypeResolver.cpp
+++ b/libsolidity/analysis/NameAndTypeResolver.cpp
@@ -102,7 +102,12 @@ bool NameAndTypeResolver::performImports(SourceUnit& _sourceUnit, map<string, So
 					else
 						for (Declaration const* declaration: declarations)
 							if (!DeclarationRegistrationHelper::registerDeclaration(
-								target, *declaration, alias.alias.get(), &alias.location, false, m_errorReporter
+								target,
+								*declaration,
+								alias.alias ? alias.alias.get() : &alias.symbol->name(),
+								&alias.location,
+								false,
+								m_errorReporter
 							))
 								error = true;
 				}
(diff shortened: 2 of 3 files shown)


Changelog.md
```diff
@@ -17,6 +17,7 @@ Bugfixes:
  * SMTChecker: Fix false positive in external calls from constructors.
  * SMTChecker: Fix internal error on some multi-source uses of ``abi.*``, cryptographic functions and constants.
  * Type Checker: Disallow modifier declarations and definitions in interfaces.
+ * Type Checker: Properly export symbols from aliased imports.
```
chriseth: Type checker is not the right component - maybe `Name Resolver`?

libsolidity/analysis/NameAndTypeResolver.cpp
```diff
@@ -102,7 +102,12 @@ bool NameAndTypeResolver::performImports(SourceUnit& _sourceUnit, map<string, So
 					else
 						for (Declaration const* declaration: declarations)
 							if (!DeclarationRegistrationHelper::registerDeclaration(
-								target, *declaration, alias.alias.get(), &alias.location, false, m_errorReporter
+								target,
+								*declaration,
+								alias.alias ? alias.alias.get() : &alias.symbol->name(),
```
chriseth, line 107: So the problem was that if you have `import {X} from "y.sol"`, then `alias.alias` would be `nullptr` and it would use the "native name" of the symbol? So this only happens if you have two imports and the name is changed in the first import, right?

chriseth: Can you maybe clarify this a bit more in the changelog, i.e. say that it fixes `import {name} from "y.sol";` if `name` is not the "native name" of the symbol?

chriseth (approved):

PR #11887, opened 2021-09-02
[SMTChecker] Fix ICE on unique error reporting

### LINKED ISSUE #11878: [SMTChecker] ICE in UniqueErrorReporter::seen
Create three files named like so

`a.sol`

```
contract A{uint x;}
```

`b.sol`

```
import"a.sol";
contract B is A {
  function g() public { assert(x>x); }
}
```

`c.sol`
```
import "b.sol";
pragma experimental SMTChecker;
contract C is B {
  function h(uint x) public { assert(x<x); }
}
```

```
$ solc a.sol b.sol c.sol
Internal compiler error during compilation:
/solidity/liblangutil/UniqueErrorReporter.h(72): Throw in function bool solidity::langutil::UniqueErrorReporter::seen(solidity::langutil::ErrorId, const solidity::langutil::SourceLocation &, const std::string &) const
std::exception::what: 
```

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/11878

The bug hasn't been released yet, so no need for a Changelog entry.
Thanks for finding it so quickly @bshastry !

### DIFF
--- a/liblangutil/UniqueErrorReporter.h
+++ b/liblangutil/UniqueErrorReporter.h
@@ -58,11 +58,7 @@ public:
 
 	void warning(ErrorId _error, std::string const& _description)
 	{
-		if (!seen(_error, {}, _description))
-		{
-			m_errorReporter.warning(_error, _description);
-			markAsSeen(_error, {}, _description);
-		}
+		m_errorReporter.warning(_error, _description);
 	}
 
 	bool seen(ErrorId _error, SourceLocation const& _location, std::string const& _description) const
@@ -77,8 +73,8 @@ public:
 
 	void markAsSeen(ErrorId _error, SourceLocation const& _location, std::string const& _description)
 	{
-		solAssert(!seen(_error, _location, _description), "");
-		m_seenErrors[{_error, _location}] = _description;
+		if (_location != SourceLocation{})
+			m_seenErrors[{_error, _location}] = _description;
 	}
 
 	ErrorList const& errors() const { return m_errorReporter.errors(); }
--- a/test/libsolidity/smtCheckerTests/imports/duplicated_errors_1.sol
+++ b/test/libsolidity/smtCheckerTests/imports/duplicated_errors_1.sol
@@ -1,0 +1,24 @@
+==== Source: a.sol ====
+contract A {
+	uint x;
+}
+==== Source: b.sol ====
+import "a.sol";
+contract B is A {
+	function g() public view { assert(x > x); }
+}
+==== Source: c.sol ====
+import "b.sol";
+contract C is B {
+	function h(uint x) public pure { assert(x < x); }
+}
+// ====
+// SMTEngine: all
+// SMTSolvers: smtlib2
+// ----
+// Warning 6328: (b.sol:62-75): CHC: Assertion violation might happen here.
+// Warning 3996: CHC analysis was not possible. No Horn solver was available. None of the installed solvers was enabled.
+// Warning 7812: (b.sol:62-75): BMC: Assertion violation might happen here.
+// Warning 8084: BMC analysis was not possible. No SMT solver (Z3 or CVC4) was available. None of the installed solvers was enabled.
+// Warning 6328: (c.sol:68-81): CHC: Assertion violation might happen here.
+// Warning 7812: (c.sol:68-81): BMC: Assertion violation might happen here.
--- a/test/libsolidity/SMTCheckerTest.cpp
+++ b/test/libsolidity/SMTCheckerTest.cpp
@@ -35,6 +35,7 @@ SMTCheckerTest::SMTCheckerTest(string const& _filename): SyntaxTest(_filename, E
 	else
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid SMT \"show unproved\" choice."));
 
+	m_modelCheckerSettings.solvers = smtutil::SMTSolverChoice::None();
 	auto const& choice = m_reader.stringSetting("SMTSolvers", "any");
 	if (choice == "any")
 		m_modelCheckerSettings.solvers = smtutil::SMTSolverChoice::All();


cameel (dismissed):
Are those all errors without location or only some of them? Looks like they are a bigger problem that I thought. Good that we at least have the assert to detect them.

cameel:
Looks like some extracted tests need to have their messages updated.

leonardoalt:
Not only messages, there's a problem I'm not sure how to solve.
`test/SyntaxTest.cpp` assumes the only errors with `location->start == 0` are the SPDX and pragma warnings. But since these changes that this PR does report warnings on `_source.location()`, it goes through the same conditions as the ones above and the things go wrong from there. Still need to figure out what to do there.

cameel:
Maybe there's code that filters it out and assumes that the message contains no location info?

leonardoalt:
@cameel https://github.com/ethereum/solidity/blob/develop/test/libsolidity/SyntaxTest.cpp#L132
This is the code I'm talking about. Not sure if that's what you meant, but yea it assumes that the license and version lines are there, and adjusts the location of every warning with `location = location - (version+license)`. I'll have to change this bit to allow my warnings that start at 0, but it might get a bit hacky

leonardoalt:
This is ready for review now

test/cmdlineTests/model_checker_solvers_smtlib2/err
```diff
@@ -1,7 +1,23 @@
...
+ --> model_checker_solvers_smtlib2/input.sol:2:1:
+  |
+2 | pragma solidity >=0.0;
+  | ^ (Relevant source part starts here and spans across multiple lines).
 
 Warning: BMC analysis was not possible. No SMT solver (Z3 or CVC4) was available. None of the installed solvers was enabled.
+ --> model_checker_solvers_smtlib2/input.sol:2:1:
+  |
+2 | pragma solidity >=0.0;
+  | ^ (Relevant source part starts here and spans across multiple lines).
```
cameel: I can see now why these had no location :) Printing this does seem a bit useless. Maybe we could deal with it in a more generic way - by not printing the source if it spans the whole contract and has multiple lines?

Alternatively - maybe the mechanism for cutting out repeated errors should just ignore errors without location?

test/libsolidity/SyntaxTest.cpp
```diff
@@ -132,8 +132,13 @@ void SyntaxTest::filterObtainedErrors()
...
-				locationStart = location->start - preambleSize;
+				if (location->start >= preambleSize)
+					locationStart = location->start - preambleSize;
+				else
+				{
+					solAssert(location->start == 0, "");
+					locationStart = 0;
+				}
 			}
 			if (location->end != -1)
```
cameel: Not adjusting `end` will result in the error not covering the whole file. This might look a bit weird, but I don't have a good solution other than making the SMTChecker aware of the preamble and that sounds like bad idea.

test/libsolidity/SyntaxTest.cpp
```diff
@@ -132,8 +132,13 @@ void SyntaxTest::filterObtainedErrors()
 			// ignore the version & license pragma inserted by the testing tool when calculating locations.
 			if (location->start != -1)
 			{
-				solAssert(location->start >= preambleSize, "");
-				locationStart = location->start - preambleSize;
+				if (location->start >= preambleSize)
```
cameel: It would be a good idea to add a comment explaining the weirdness happening here:
```suggestion
				// Preamble is not in the original input so error locations need to be shifted back.
				// Normally this is safe because errors should not happen inside preamble. Errors
				// covering the whole file are a special case because they start at 0, not after the preamble.
				if (location->start >= preambleSize)
```

cameel (commented):
Looks ok but the hack is not pretty. The whole preamble thing is a hack though and I it's hard to suggest something clean here.

Maybe ignoring errors without location during duplicate removal would be a better solution to the problem overall?

leonardoalt:
I mean, in the end I felt like this wasn't a hack anymore, just another rule "if the error actually is about position 0, just keep it"

test/cmdlineTests/model_checker_solvers_smtlib2/err
leonardoalt: Yea I think I prefer the second idea. The first one might touch parts outside the smtchecker that I'm not aware of.

leonardoalt:
I'll fix it on the error duplicate removal layer, so we don't have to touch the tests part hopefully.

leonardoalt:
@cameel fixed it by ignoring the check for errors without SourceLocation

liblangutil/UniqueErrorReporter.h
```diff
@@ -77,7 +77,8 @@ class UniqueErrorReporter
 
 	void markAsSeen(ErrorId _error, SourceLocation const& _location, std::string const& _description)
 	{
-		solAssert(!seen(_error, _location, _description), "");
+		if (_location != SourceLocation{})
+			solAssert(!seen(_error, _location, _description), "");
 		m_seenErrors[{_error, _location}] = _description;
```
cameel: Wouldn't it be better not to mark these errors as seen at all?

leonardoalt: Updated

liblangutil/UniqueErrorReporter.h
```diff
@@ -58,11 +58,7 @@ class UniqueErrorReporter
 
 	void warning(ErrorId _error, std::string const& _description)
 	{
-		if (!seen(_error, {}, _description))
-		{
-			m_errorReporter.warning(_error, _description);
-			markAsSeen(_error, {}, _description);
-		}
+		markAsSeen(_error, {}, _description);
```
cameel: I think you may have left the wrong line here :)
```suggestion
		m_errorReporter.warning(_error, _description);
```

test/libsolidity/smtCheckerTests/imports/duplicated_errors_1.sol
```diff
@@ -0,0 +1,19 @@
...
+==== Source: c.sol ====
+import "b.sol";
+contract C is B {
+	function h(uint x) public pure { assert(x < x); }
+}
+// ====
+// SMTEngine: all
+// ----
+// Warning 6328: (b.sol:62-75): CHC: Assertion violation happens here.\nCounterexample:\nx = 0\n\nTransaction trace:\nB.constructor()\nState: x = 0\nB.g()
+// Warning 6328: (c.sol:68-81): CHC: Assertion violation happens here.\nCounterexample:\nx = 0\nx = 0\n\nTransaction trace:\nC.constructor()\nState: x = 0\nC.h(0)
```
cameel: Actually, why do these have locations? Wasn't the error triggered by errors with different descriptions and no location?

cameel (commented):
Looks much neater overall.

Weirdly, tests are failing due to gas differences. The PR does not touch anything that could trigger that though. Is `develop` broken?

liblangutil/UniqueErrorReporter.h
leonardoalt: ah damn true

test/libsolidity/smtCheckerTests/imports/duplicated_errors_1.sol
leonardoalt: These are triggered by the assertions themselves, that's why they have locations.

leonardoalt:
> Looks much neater overall.
> 
> Weirdly, tests are failing due to gas differences. The PR does not touch anything that could trigger that though. Is `develop` broken?

let me rebase and run it locally

cameel:
I just got these in my PRs. So yeah, `develop` must be broken.

test/libsolidity/smtCheckerTests/imports/duplicated_errors_1.sol
cameel: So there should be more errors here? Perhaps they did not get printed due to the bug in `warning()`?

leonardoalt: Why should there be more errors? I think this is correct.

cameel: In that case I don't understand why the ICE got triggered in the first place.

leonardoalt: Good q, let me think

leonardoalt: Weird, I'm re-running the issue on develop and it doesn't happen anymore.

leonardoalt: Ahh ok it happens only when there are no solvers available

leonardoalt:
@cameel now it all makes much more sense :)

cameel (dismissed):

leonardoalt:
@cameel can you reapprove?

cameel (approved):

PR #10653, opened 2020-12-18
Add more tests for the SPDX-License-Identifier

axic:
Part of #10145.

PR #11888, opened 2021-09-02
Properly detect multiple licenses and validate them.

### LINKED ISSUE #10145: Multiple SPDX-License-Identifier in the same section of comments doesn't generate an error
## Description
Multiple SPDX-License-Identifier in the same section of comments doesn't generate an error and compiles.

## Environment

- Compiler version: 0.6.8 & 0.7.4

## Steps to Reproduce

The following with two SPDX License Identifiers doesn't generate an error and compiles.
```solidity


// SPDX-License-Identifier: Something Else

pragma solidity ^0.7.0;

contract MyContract {
}
```

Whilst two SPDX-License-Identifiers separated by code generates the expected parser error
`browser/MyContract.sol: ParserError: Multiple SPDX license identifiers found in source file. Use "AND" or "OR" to combine multiple licenses. Please see https://spdx.org for more information.`
```solidity


pragma solidity ^0.7.0;

contract MyContract {
}


// SPDX-License-Identifier: Something Else
```

Marenz:
fixes #10145

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -22,6 +22,7 @@ Bugfixes:
  * Commandline Interface: Disallow the ``--experimental-via-ir`` option in Standard JSON, Assembler and Linker modes.
  * Opcode Optimizer: Prevent the optimizer from running multiple times to avoid potential bytecode differences for referenced code.
  * Name Resolver: Fix that when importing an aliased symbol using ``import {AliasedName} from "a.sol"`` it would use the original name of the symbol and not the aliased one.
+ * Parser: Properly check for multiple SPDX license identifiers next to each other and validate them.
  * SMTChecker: Fix false negative caused by ``push`` on storage array references returned by internal functions.
  * SMTChecker: Fix false positive in external calls from constructors.
  * SMTChecker: Fix internal error on some multi-source uses of ``abi.*``, cryptographic functions and constants.
--- a/libsolidity/parsing/Parser.cpp
+++ b/libsolidity/parsing/Parser.cpp
@@ -34,6 +34,7 @@
 #include <libyul/backends/evm/EVMDialect.h>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/algorithm/string/replace.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 #include <cctype>
 #include <vector>
 #include <regex>
@@ -2078,7 +2079,8 @@ bool Parser::variableDeclarationStart()
 optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> const& _nodes)
 {
 	// We circumvent the scanner here, because it skips non-docstring comments.
-	static regex const licenseRegex("SPDX-License-Identifier:\\s*([a-zA-Z0-9 ()+.-]+)");
+	static regex const licenseNameRegex("([a-zA-Z0-9 ()+.-]+)");
+	static regex const licenseDeclarationRegex("SPDX-License-Identifier:\\s*(.+?)([\n\r]|(\\*/))");
 
 	// Search inside all parts of the source not covered by parsed nodes.
 	// This will leave e.g. "global comments".
@@ -2093,21 +2095,33 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
 			sequencesToSearch.emplace_back(source.begin() + node->location().end, source.end());
 		}
 
-	vector<string> matches;
+	vector<string> licenseNames;
 	for (auto const& [start, end]: sequencesToSearch)
 	{
-		smatch match;
-		if (regex_search(start, end, match, licenseRegex))
-		{
-			string license{boost::trim_copy(string(match[1]))};
-			if (!license.empty())
-				matches.emplace_back(std::move(license));
-		}
+		auto declarationsBegin = std::sregex_iterator(start, end, licenseDeclarationRegex);
+		auto declarationsEnd = std::sregex_iterator();
+
+		for (std::sregex_iterator declIt = declarationsBegin; declIt != declarationsEnd; ++declIt)
+			if (!declIt->empty())
+			{
+				string license = boost::trim_copy(string((*declIt)[1]));
+				licenseNames.emplace_back(std::move(license));
+			}
 	}
 
-	if (matches.size() == 1)
-		return matches.front();
-	else if (matches.empty())
+	if (licenseNames.size() == 1)
+	{
+		string const& license = licenseNames.front();
+		if (regex_match(license, licenseNameRegex))
+			return license;
+		else
+			parserError(
+				1114_error,
+				{-1, -1, m_scanner->currentLocation().sourceName},
+				"Invalid SPDX license identifier."
+			);
+	}
+	else if (licenseNames.empty())
 		parserWarning(
 			1878_error,
 			{-1, -1, m_scanner->currentLocation().sourceName},
--- a/test/libsolidity/Metadata.cpp
+++ b/test/libsolidity/Metadata.cpp
@@ -457,28 +457,6 @@ BOOST_AUTO_TEST_CASE(metadata_license_gpl3_or_apache2)
 	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0 OR Apache-2.0");
 }
 
-BOOST_AUTO_TEST_CASE(metadata_license_ignored_unicode)
-{
-	char const* sourceCode = R"(
-		// SPDX-License-Identifier: ⡉⡊⡋⡌⡍⡎⡏⡐⡑⡒
-		pragma solidity >=0.0;
-		contract C {
-		}
-	)";
-	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == nullopt);
-}
-
-BOOST_AUTO_TEST_CASE(metadata_license_ignored_stray_unicode)
-{
-	char const* sourceCode = R"(
-		// SPDX-License-Identifier: GPL-3.0 ⡉⡊⡋⡌⡍⡎⡏⡐⡑⡒
-		pragma solidity >=0.0;
-		contract C {
-		}
-	)";
-	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
-}
-
 BOOST_AUTO_TEST_CASE(metadata_license_bidi_marks)
 {
 	char const* sourceCode =
@@ -570,15 +548,6 @@ BOOST_AUTO_TEST_CASE(metadata_license_natspec_multiline)
 	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
 }
 
-BOOST_AUTO_TEST_CASE(metadata_license_no_whitespace)
-{
-	char const* sourceCode = R"(
-		//SPDX-License-Identifier:GPL-3.0
-		contract C {}
-	)";
-	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
-}
-
 BOOST_AUTO_TEST_CASE(metadata_license_no_whitespace_multiline)
 {
 	char const* sourceCode = R"(
@@ -592,6 +561,15 @@ BOOST_AUTO_TEST_CASE(metadata_license_nonempty_line)
 {
 	char const* sourceCode = R"(
 		pragma solidity >= 0.0; // SPDX-License-Identifier: GPL-3.0
+		contract C {}
+	)";
+	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
+}
+
+BOOST_AUTO_TEST_CASE(metadata_license_no_whitespace)
+{
+	char const* sourceCode = R"(
+		//SPDX-License-Identifier:GPL-3.0
 		contract C {}
 	)";
 	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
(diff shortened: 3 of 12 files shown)


libsolidity/parsing/Parser.cpp
```diff
@@ -2057,7 +2058,8 @@ bool Parser::variableDeclarationStart()
 optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> const& _nodes)
 {
 	// We circumvent the scanner here, because it skips non-docstring comments.
-	static regex const licenseRegex("SPDX-License-Identifier:\\s*([a-zA-Z0-9 ()+.-]+)");
+	static regex const licenseRegexStrict("^([a-zA-Z0-9 ()+.-]+)$");
+	static regex const licenseRegex("SPDX-License-Identifier:\\s*([^\n]+)");
```
cameel: How about this?
```suggestion
	static regex const licenseNameRegex("^([a-zA-Z0-9 ()+.-]+)$");
	static regex const licenseDeclarationRegex("SPDX-License-Identifier:\\s*([^\n]+)");
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
+		auto words_end = std::sregex_iterator();
+
+		for (std::sregex_iterator i = words_begin; i != words_end; ++i)
 		{
-			string license{boost::trim_copy(string(match[1]))};
-			if (!license.empty())
+			std::smatch match = *i;
+			if (!match.empty())
+			{
+				string license{string(match[1])};
```
cameel: Does it actually need a this extra string conversion?
```suggestion
				string license = match[1];
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
+				// Remove trailing comment terminator if present
+				if (boost::algorithm::ends_with(license, "*/"))
+					license = license.substr(0, license.size() - 2);
+
+				license = boost::trim_copy(license);
```
cameel: You could get `*/` stripped and whitespace trimmed automatically by including it in the regex:

```c++
"SPDX-License-Identifier:\\s*([^\n]+)\s*(?:\*/)?"
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
 	for (auto const& [start, end]: sequencesToSearch)
 	{
 		smatch match;
```
cameel: This seems no longer used and gets shadowed by another variable inside the loop below.

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
-		if (regex_search(start, end, match, licenseRegex))
+		auto words_begin =
+			std::sregex_iterator(start, end, licenseRegex);
+		auto words_end = std::sregex_iterator();
+
+		for (std::sregex_iterator i = words_begin; i != words_end; ++i)
```
cameel: The naming bothers me a bit here. I was initially a bit confused about why you match single words if the regex is for the whole declaration :) Also, it does not follow our camel case convention.

```suggestion
		auto declarationsBegin = std::sregex_iterator(start, end, licenseRegex);
		auto declarationsEnd = std::sregex_iterator();

		for (std::sregex_iterator declarationIt = declarationsBegin; declarationIt != declarationsEnd; ++declarationIt)
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
+			if (!match.empty())
+			{
+				string license{string(match[1])};
+				// Remove trailing comment terminator if present
+				if (boost::algorithm::ends_with(license, "*/"))
+					license = license.substr(0, license.size() - 2);
+
+				license = boost::trim_copy(license);
+
 				matches.emplace_back(std::move(license));
```
cameel: Those are just strings now. `matches` sounds is if they were still the match objects.
```suggestion
				licenseNames.emplace_back(std::move(license));
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2076,16 +2078,39 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
 	if (matches.size() == 1)
-		return matches.front();
+	{
+		string const& license = matches.front();
+		if (regex_search(license.begin(), license.end(), licenseRegexStrict))
+			return license;
+		else
+			parserError(
+				1114_error,
+				{-1, -1, m_scanner->currentLocation().sourceName},
```
cameel, line 2120: Would be nice if we could show the location. It might not be obvious which declaration has the malformed name in it.

test/libsolidity/syntaxTests/license/license_double2.sol
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: GPL-3.0
+// SPDX-License-Identifier: MIT
```
cameel, line 2: I'd add a test with multiple declarations in the same multi-line comment and also another with declarations on the same line.

test/libsolidity/Metadata.cpp
```diff
@@ -552,15 +530,6 @@ BOOST_AUTO_TEST_CASE(metadata_license_natspec_multiline)
...
 
-BOOST_AUTO_TEST_CASE(metadata_license_no_whitespace)
-{
-	char const* sourceCode = R"(
-		//SPDX-License-Identifier:GPL-3.0
-		contract C {}
-	)";
-	BOOST_CHECK(compileAndCheckLicenseMetadata("C", sourceCode) == "GPL-3.0");
-}
-
```
cameel: Was this removed because we already have this test in a .sol file? Note that the .sol file has a comment:
```
// This test is actually useless, as the test suite adds the automatic preamble.
```

See https://github.com/ethereum/solidity/pull/10653#discussion_r545974629 and https://github.com/ethereum/solidity/pull/10653#discussion_r558422143.

test/libsolidity/syntaxTests/license/license_double2.sol
cameel: Also another with two identical rather than different licenses (it will still be treated as error, right?)

libsolidity/parsing/Parser.cpp
Marenz: I don't understand the `?:` part of your regex suggestion

Marenz: Also, doesn't the `([^\n]+)` eat all the characters anyway so what follows will be ignored?

Marenz: At least with this, the regex doesn't properly work so far :)

cameel: `(?:)` is a non-capturing group. Same as `()` but the content does not appear on the list of matches.

Sorry, looks like I did not escape the regex properly (I should have tried it on something like https://regex101.com first). Should have been:
```c++
"SPDX-License-Identifier:\\s*([^\n]+)\\s*(?:\*\/)?"
```

> Also, doesn't the `([^\n]+)` eat all the characters anyway so what follows will be ignored?

It does capture the `*/` but you get it as a part of your group and then you have to cut it out manually. My regex would match it and then discard it.

Marenz: I saw the escaping problems and did that myself, but the regex still won't work. The `([^\n]+)` will simply capture everything until the end of the line...

Marenz: See also https://javascript.info/regexp-greedy-and-lazy

cameel: Ah, you're right. It would work if `*/` or whitespace was not optional. With the current structure the `([^\n]+)` part grabs them instead.

Ok then. I guess it might be possible to come up with a more complex regex to handle it but at this point cutting it out manually starts looking like the better approach :)

Marenz: That was my thinking ^^ however, by turning my `([^\n]+)` into a `([^\n])+?` it actually works (switching it from greedy to lazy)

cameel: But won't `([^\n])+?`, being non-greedy, grab only the first character of the license name and stop?

I mean, it should work with regex match  where the whole string has to be matched but with regex search I'd expect it to stop after it gets the smallest possible match.

test/libsolidity/Metadata.cpp
Marenz: I am actually not sure anymore at this point. I added it back and it works. I suspect the whitespace parts were not working at some point during this PR and only successful tests are useful in the Metadata.cpp

libsolidity/parsing/Parser.cpp
Marenz: Well, you should only have one declaration anyway.. It is actually not trivial to find the location as we're operating on the source file and not AST nodes here..

ekpyron:
[just randomly looked through PRs and saw the antlr grammar failing here - without really looking at the PR otherwise: it might be a reasonable case for the blacklist in ``scripts/test_antlr_grammar.sh``, if it's too much hassle to adjust the grammar to fail on those]

libsolidity/parsing/Parser.cpp
cameel: Ah, ok then.

Changelog.md
```diff
@@ -16,6 +16,7 @@ Compiler Features:
 
 Bugfixes:
  * Name Resolver: Fix that when importing an aliased symbol using ``import {AliasedName} from "a.sol"`` it would use the original name of the symbol and not the aliased one.
+ * Parser: Properly check for multiple SPDX license indentifiers and validate them.
```
```suggestion
 * Parser: Properly check for multiple SPDX license identifiers and validate them.
```

cameel: Also maybe it would be good to clarify a bit:
```suggestion
 * Parser: Properly check for multiple SPDX license identifiers next to each other and validate them.
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2072,21 +2074,40 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
+			{
+				string license = boost::trim_copy(string((*declIt)[1]));
+
+				// Remove trailing comment terminator if present
+				if (boost::algorithm::ends_with(license, "*/"))
+					license = license.substr(0, license.size() - 2);
+
+				license = boost::trim_copy(license);
+
+				licenseNames.emplace_back(std::move(license));
```
cameel, line 2108:
```suggestion
				licenseNames.emplace_back(move(license));
```

test/libsolidity/syntaxTests/license/license_double5.sol
```diff
@@ -0,0 +1,4 @@
+/* SPDX-License-Identifier: GPL-3.0 OR GPL-2.0 */ /* SPDX-License-Identifier: GPL-3.0 OR GPL-2.0 */
+contract C {}
+// ----
+// ParserError 1114: Invalid SPDX license identifier.
```
cameel: Since this is invalid, it probably sees `GPL-3.0 OR GPL-2.0 */ /* SPDX-License-Identifier: GPL-3.0 OR GPL-2.0` as the license name? I think that having something after the comment should be legal. I could imagine someone doing something like this:
```solidity
/* SPDX-License-Identifier: GPL-3.0 OR GPL-2.0 */ pragma solidity >=0.0;
```

cameel: Also, this shows that it would really be nice to have a location for the error. Maybe we could at least include the invalid identifer in the error message? On the other hand if it contains invalid UTF-8 it might be a problem...

libsolidity/parsing/Parser.cpp
Marenz: Hmm I thought we wanted ot use `std::` for move calls? Though, looking at the source, it seems wildly inconsistent, but with a clear majority winner:

```
> git grep -o -E "(std::)?move\("  | grep -c std
619
> git grep -o -E "(std::)?move\("  | grep -c std -v
348

```

cameel: We did? Ok then. It's just that in this file `move` is mostly used without `std::` and the `std::` prefix is generally superfluous if you have `using namespace std` in the file.

chriseth: We had it at a point where `boost::move` was still widespread. It is not anymore.

Marenz: I am still unclear at this point if you are saying we should use it or not @chriseth :)

test/libsolidity/syntaxTests/license/license_double5.sol
Marenz: You are right, this is a completely wrong test result. I had to fix the regex again to make this properly work, and it's so much better now. Good call!

libsolidity/parsing/Parser.cpp
chriseth: No need to use it, but also does not hurt.

libsolidity/parsing/Parser.cpp
```diff
@@ -2093,21 +2095,33 @@ optional<string> Parser::findLicenseString(std::vector<ASTPointer<ASTNode>> cons
...
+			}
 	}
 
-	if (matches.size() == 1)
-		return matches.front();
-	else if (matches.empty())
+	if (licenseNames.size() == 1)
+	{
+		string const& license = licenseNames.front();
+		if (regex_search(license.begin(), license.end(), licenseNameRegex))
```
```suggestion
		if (regex_search(license, licenseNameRegex))
```

chriseth: Why not use `regex_match` instead and remove the `^` / `$`? It is confusing because `$` can also sometimes mean "end of line".

chriseth (approved):

PR #11890, opened 2021-09-03
Report stack too deep and conditional stack compression in StackLayoutGenerator.

ekpyron:
Reporting stack-too-deep errors is required for guiding the ``StackCompressor`` and ``StackLimitEvader`` in https://github.com/ethereum/solidity/pull/11493.
Simultanously, it can be used to guide the stack compression during stack layout generation itself.

This PR contains the changes to the StackLayoutGenerator in https://github.com/ethereum/solidity/pull/11493.

### DIFF
--- a/libyul/backends/evm/StackLayoutGenerator.cpp
+++ b/libyul/backends/evm/StackLayoutGenerator.cpp
@@ -28,6 +28,7 @@
 #include <libsolutil/Visitor.h>
 
 #include <range/v3/algorithm/any_of.hpp>
+#include <range/v3/algorithm/find.hpp>
 #include <range/v3/range/conversion.hpp>
 #include <range/v3/view/all.hpp>
 #include <range/v3/view/concat.hpp>
@@ -38,6 +39,7 @@
 #include <range/v3/view/map.hpp>
 #include <range/v3/view/reverse.hpp>
 #include <range/v3/view/take.hpp>
+#include <range/v3/view/take_last.hpp>
 #include <range/v3/view/transform.hpp>
 
 using namespace solidity;
@@ -55,12 +57,83 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
 	return stackLayout;
 }
 
+map<YulString, vector<StackLayoutGenerator::StackTooDeep>> StackLayoutGenerator::reportStackTooDeep(CFG const& _cfg)
+{
+	map<YulString, vector<StackLayoutGenerator::StackTooDeep>> stackTooDeepErrors;
+	stackTooDeepErrors[YulString{}] = reportStackTooDeep(_cfg, YulString{});
+	for (auto const& function: _cfg.functions)
+		if (auto errors = reportStackTooDeep(_cfg, function->name); !errors.empty())
+			stackTooDeepErrors[function->name] = move(errors);
+	return stackTooDeepErrors;
+}
+
+vector<StackLayoutGenerator::StackTooDeep> StackLayoutGenerator::reportStackTooDeep(CFG const& _cfg, YulString _functionName)
+{
+	StackLayout stackLayout;
+	CFG::FunctionInfo const* functionInfo = nullptr;
+	if (!_functionName.empty())
+	{
+		functionInfo = &ranges::find(
+			_cfg.functionInfo,
+			_functionName,
+			util::mapTuple([](auto&&, auto&& info) { return info.function.name; })
+		)->second;
+		yulAssert(functionInfo, "Function not found.");
+	}
+
+	StackLayoutGenerator generator{stackLayout};
+	CFG::BasicBlock const* entry = functionInfo ? functionInfo->entry : _cfg.entry;
+	generator.processEntryPoint(*entry);
+	return generator.reportStackTooDeep(*entry);
+}
+
 StackLayoutGenerator::StackLayoutGenerator(StackLayout& _layout): m_layout(_layout)
 {
 }
 
 namespace
 {
+/// @returns all stack too deep errors that would occur when shuffling @a _source to @a _target.
+vector<StackLayoutGenerator::StackTooDeep> findStackTooDeep(Stack const& _source, Stack const& _target)
+{
+	Stack currentStack = _source;
+	vector<StackLayoutGenerator::StackTooDeep> stackTooDeepErrors;
+	auto getVariableChoices = [](auto&& _range) {
+		vector<YulString> result;
+		for (auto const& slot: _range)
+			if (auto const* variableSlot = get_if<VariableSlot>(&slot))
+				if (!util::contains(result, variableSlot->variable.get().name))
+					result.push_back(variableSlot->variable.get().name);
+		return result;
+	};
+	::createStackLayout(
+		currentStack,
+		_target,
+		[&](unsigned _i)
+		{
+			if (_i > 16)
+				stackTooDeepErrors.emplace_back(StackLayoutGenerator::StackTooDeep{
+					_i - 16,
+					getVariableChoices(currentStack | ranges::views::take_last(_i + 1))
+				});
+		},
+		[&](StackSlot const& _slot)
+		{
+			if (canBeFreelyGenerated(_slot))
+				return;
+			if (
+				auto depth = util::findOffset(currentStack | ranges::views::reverse, _slot);
+				depth && *depth >= 16
+			)
+				stackTooDeepErrors.emplace_back(StackLayoutGenerator::StackTooDeep{
+					*depth - 15,
+					getVariableChoices(currentStack | ranges::views::take_last(*depth + 1))
+				});
+		},
+		[&]() {}
+	);
+	return stackTooDeepErrors;
+}
 
 /// @returns the ideal stack to have before executing an operation that outputs @a _operationOutput, s.t.
 /// shuffling to @a _post is cheap (excluding the input of the operation itself).
@@ -191,13 +264,16 @@ Stack createIdealLayout(Stack const& _operationOutput, Stack const& _post, Calla
 }
 }
 
-Stack StackLayoutGenerator::propagateStackThroughOperation(Stack _exitStack, CFG::Operation const& _operation)
+Stack StackLayoutGenerator::propagateStackThroughOperation(Stack _exitStack, CFG::Operation const& _operation, bool _aggressiveStackCompression)
 {
+	// Enable aggressive stack compression for recursive calls.
+	if (auto const* functionCall = get_if<CFG::FunctionCall>(&_operation.operation))
+		if (functionCall->recursive)
+			_aggressiveStackCompression = true;
+
 	// This is a huge tradeoff between code size, gas cost and stack size.
-	auto generateSlotOnTheFly = [&](StackSlot const&) {
-		//return stack.size() > 12 && canBeFreelyGenerated(_slot);
-		// return canBeFreelyGenerated(_slot);
-		return false;
+	auto generateSlotOnTheFly = [&](StackSlot const& _slot) {
+		return _aggressiveStackCompression && canBeFreelyGenerated(_slot);
 	};
 
 	// Determine the ideal permutation of the slots in _exitLayout that are not operation outputs (and not to be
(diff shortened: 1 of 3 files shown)


libyul/backends/evm/StackLayoutGenerator.h
```diff
@@ -86,6 +96,10 @@ class StackLayoutGenerator
 	/// stack shuffling when starting from the returned layout.
 	static Stack combineStack(Stack const& _stack1, Stack const& _stack2);
 
+	/// Walks through the CFG and reports and stack too deep errors that would occur when generating code for it
```
cameel:
```suggestion
	/// Walks through the CFG and reports any stack too deep errors that would occur when generating code for it
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
+			return;
+		if (
+			auto depth = util::findOffset(currentStack | ranges::views::reverse, _slot);
+			depth && *depth >= 16
+		)
+			stackTooDeepErrors.emplace_back(StackLayoutGenerator::StackTooDeep{
+				*depth - 15,
+				getVariableChoices(currentStack | ranges::views::take_last(*depth + 1))
+			});
+	}, [&]() {});
```
cameel: This way of wrapping makes it hard for me to see at a glance what is nested in what and where one parameter to `createStackLayout` starts and another begins. Having each parameter start on a new line would make it much more apparent that the function takes two normal parameters and 3 lambdas:
```suggestion
	::createStackLayout(
		currentStack,
		_target,
		[&](unsigned _i)
		{
			if (_i > 16)
				stackTooDeepErrors.emplace_back(StackLayoutGenerator::StackTooDeep{
					_i - 16,
					getVariableChoices(currentStack | ranges::views::take_last(_i + 1))
				});
		},
		[&](StackSlot const& _slot)
		{
			if (canBeFreelyGenerated(_slot))
				return;
			if (
				auto depth = util::findOffset(currentStack | ranges::views::reverse, _slot);
				depth && *depth >= 16
			)
				stackTooDeepErrors.emplace_back(StackLayoutGenerator::StackTooDeep{
					*depth - 15,
					getVariableChoices(currentStack | ranges::views::take_last(*depth + 1))
				});
		},
		[&]() {}
);
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
+		functionInfo = &ranges::find(
+			_cfg.functionInfo,
+			_functionName,
+			util::mapTuple([](auto&&, auto&& info) { return info.function.name; })
+		)->second;
+		yulAssert(functionInfo, "Function not found.");
+	}
+
+	StackLayoutGenerator generator{stackLayout};
+	CFG::BasicBlock const* entry = functionInfo ? functionInfo->entry : _cfg.entry;
```
cameel, line 85: Oh, so an empty `YulString` is just a way to use the main entry point of the graph rather than one belonging to any particular function? And the graph is guaranteed not to have any functions with empty names? I think it should be mentioned in function's docstring.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -549,6 +620,55 @@ Stack StackLayoutGenerator::combineStack(Stack const& _stack1, Stack const& _sta
 	return commonPrefix + bestCandidate;
 }
 
+vector<StackLayoutGenerator::StackTooDeep> StackLayoutGenerator::reportStackTooDeep(CFG::BasicBlock const& _entry) const
+{
+	vector<StackTooDeep> stackTooDeepErrors;
+	util::BreadthFirstSearch<CFG::BasicBlock const*> breadthFirstSearch{{&_entry}};
+	breadthFirstSearch.run([&](CFG::BasicBlock const* _block, auto _addChild) {
+		Stack stack;
+		stack = m_layout.blockInfos.at(_block).entryLayout;
```
cameel:
```suggestion
		Stack stack = m_layout.blockInfos.at(_block).entryLayout;
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -549,6 +620,55 @@ Stack StackLayoutGenerator::combineStack(Stack const& _stack1, Stack const& _sta
...
+
+		for (auto const& operation: _block->operations)
+		{
+			Stack& operationEntry = m_layout.operationEntryLayout.at(&operation);
+
+			stackTooDeepErrors += findStackTooDeep(stack, operationEntry);
+			stack = operationEntry;
+			for (size_t i = 0; i < operation.input.size(); i++)
+				stack.pop_back();
+			stack += operation.output;
```
cameel: I originally wanted to say that `stack` gets overwritten in every cycle so this looks pointless but then I noticed that it still gets passed to `findStackTooDeep()` in the next loop cycle. Maybe it would be a good idea to rename it to make its purpose clearer? E.g. `previousOperationExitLayout`?

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -549,6 +620,55 @@ Stack StackLayoutGenerator::combineStack(Stack const& _stack1, Stack const& _sta
...
+		}
+		// Do not attempt to create the exit layout here, since the code generator will directly move to the
+		// target entry layout.
```
cameel: I do not understand this comment. Don't we have the exit layout already created in `stack` variable at this point?

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -235,18 +303,21 @@ Stack StackLayoutGenerator::propagateStackThroughOperation(Stack _exitStack, CFG
...
-	for (auto& operation: _block.operations | ranges::views::reverse)
-		stack = propagateStackThroughOperation(stack, operation);
+	Stack stack = _exitStack;
+	for (auto&& [idx, operation]: _block.operations | ranges::views::enumerate | ranges::views::reverse)
+	{
+		Stack newStack = propagateStackThroughOperation(stack, operation, _aggressiveStackCompression);
+		if (!_aggressiveStackCompression && !findStackTooDeep(newStack, stack).empty())
+			// If we had stack errors, run again with aggressive stack compression.
+			return propagateStackThroughBlock(move(_exitStack), _block, true);
+		stack = std::move(newStack);
```
cameel:
```suggestion
		stack = move(newStack);
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -191,13 +256,16 @@ Stack createIdealLayout(Stack const& _operationOutput, Stack const& _post, Calla
 }
 }
 
-Stack StackLayoutGenerator::propagateStackThroughOperation(Stack _exitStack, CFG::Operation const& _operation)
+Stack StackLayoutGenerator::propagateStackThroughOperation(Stack _exitStack, CFG::Operation const& _operation, bool _aggressiveStackCompression)
 {
+	// Enable aggressive stack compression for recursive calls.
+	if (auto const* functionCall = get_if<CFG::FunctionCall>(&_operation.operation))
+		if (functionCall->recursive)
+			_aggressiveStackCompression = true;
```
cameel, line 272: This part means that a call with `_aggressiveStackCompression = false` sometimes actually runs with `true` so trying to rerun with `true` in `propagateStackThroughBlock()` is pointless. Moving it to `propagateStackThroughBlock()` would let you avoid this.

Or is `createIdealLayout()` light enough that rerunning this is not an issue?

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
 }
 
 namespace
 {
+/// @returns all stack too deep error that would occur when shuffling @a _source to @a _target.
+vector<StackLayoutGenerator::StackTooDeep> findStackTooDeep(Stack const& _source, Stack const& _target)
+{
+	Stack currentStack = _source;
+	vector<StackLayoutGenerator::StackTooDeep> stackTooDeepErrors;
+	auto getVariableChoices = [](auto&& range) {
```
cameel:
```suggestion
	auto getVariableChoices = [](auto&& _range) {
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
+	return generator.reportStackTooDeep(*entry);
+}
+
 StackLayoutGenerator::StackLayoutGenerator(StackLayout& _layout): m_layout(_layout)
 {
 }
 
 namespace
 {
+/// @returns all stack too deep error that would occur when shuffling @a _source to @a _target.
```
cameel:
```suggestion
/// @returns all stack too deep errors that would occur when shuffling @a _source to @a _target.
```

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
+vector<StackLayoutGenerator::StackTooDeep> findStackTooDeep(Stack const& _source, Stack const& _target)
+{
+	Stack currentStack = _source;
+	vector<StackLayoutGenerator::StackTooDeep> stackTooDeepErrors;
+	auto getVariableChoices = [](auto&& range) {
+		vector<YulString> result;
+		for (auto const& slot: range)
+			if (auto const* variableSlot = get_if<VariableSlot>(&slot))
+				if (!util::contains(result, variableSlot->variable.get().name))
+					result.push_back(variableSlot->variable.get().name);
```
cameel, line 106: Is the order important? If not, you could just use a `set`.

Also this function could then basically be replaced with `ranges::views::filter()`.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -55,12 +57,75 @@ StackLayout StackLayoutGenerator::run(CFG const& _cfg)
...
+		return result;
+	};
+	::createStackLayout(currentStack, _target, [&](unsigned _i) {
+		if (_i > 16)
```
cameel: Shouldn't this be `_i >= 16`? I'm assuming that `_i` is an index of a stack slot you're swapping with that it's 0-based (hard to tell because `createStackLayout()` does not document callback arguments). So 16 would already mean that we have 17 slots in use.

cameel (commented):
If the `_i > 16` part mentioned in https://github.com/ethereum/solidity/pull/11890#discussion_r702148877 is not a bug then this looks correct.

I have some suggestions but nothing that would affect functionality.

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: Renamed it to ``currentStack`` now, maybe that's clearer?

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: I made the comment more precise - it was meant to make it clear that not considering ``m_layout.blockInfos.at(_block).exitLayout`` is on purpose.

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: This will only set ``_aggressiveStackCompression`` to true for the specific operation (if it's a recursive function call), while rerunning the entire block with ``true`` in ``propagateStackThroughBlock`` will attempt to rerun *every* operation in the block with aggressive stack compression.
So this is only pointless, if the entire block is only a single recursive function call (and recursive calls are expected to be rare anyways) - and rerunning this stuff is not overly expensive anyways, so I'd say it's fine.

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: The order is relevant, yes - the choice of variables to compress/rematerialize or move to memory will depend on this order (at least sometimes, if there is no reason to prefer one over the other) - and we generally try to avoid stuff depending on the ``YulString``-comparison-induced order (that's based on hash comparison, so while deterministic, it's a rather weird order).

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: I added some comments to ``createStackLayout`` about this. You're right, this is 0-based indices, but ``SWAP16`` swaps the stack top with the sixteenth slot from the top (since ``SWAP1`` already swaps the top-most with the first slot from top), so the boundary is still fine. Or put differently: the argument ``_i`` is as in ``SWAP<_i>`` and we do have ``SWAP16``.

ekpyron:
The ``_i > 16`` part is fine and I think I addressed all other comments.

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: Documented it.

libyul/backends/evm/StackLayoutGenerator.cpp
cameel: Well, part of the problem for me was that I was treating it as the current stack but at some points in the code (especially just before this) it's easier to actually think about it as the previous stack.

At least it's better than just `stack` :)

libyul/backends/evm/StackLayoutGenerator.cpp
cameel: Ah, the function is supposed to generate `m_layout.blockInfos.at(_block).exitLayout`?

Well, the confusing part for me was that gathering info about stack too deep errors happens on the fly, while the stack layout is being generated, but the added code is not all that much different from one that would be doing it after the fact. It only shows through in a few places and at some point I was even wondering why some of these functions are not `const` :)

I guess this is caused by the fact that I'm reviewing the diff rather than at the class as a whole so I'm only aware of how that reviewed part works. Still, despite problems like this, I think these smaller, more focused PRs make reviewing easier overall.

libyul/backends/evm/StackLayoutGenerator.cpp
cameel: Oh, you're right. I missed the fact that it's rerunning the whole block, not just an operation.

libyul/backends/evm/StackLayoutGenerator.cpp
cameel: Oh, so it means that actually there are 17 reachable slots on the stack. I didn't realize that until now, but it makes sense that `SWAP16` would swap slots 0 and 16 (with 16 being the 17-th slot).

cameel (approved):

PR #11892, opened 2021-09-03
Prevent the optimizer from removing ``memoryguard``.

ekpyron:
Disallows removing the ``memoryguard`` instruction, since that will prevent stack limit evasion.

Thinking about that, I'm wondering if we should also set ``movable`` to ``false``, since allowing to copy ``memoryguard`` without being able to remove it might otherwise result in multiple stray ``pop(memoryguard(...))``s... on the other hand that would also be fine, since the peephole optimizer would ultimately get rid of those later anyways...

The proper side-effect for ``memoryguard`` would be "cannot be removed, if it is the last occurrence", but I'd say it's not worth introducing that specifically as special case... if we end up actually seeing a lot of duplicated ``pop(memoryguard(...))``s, we can still specifically remove all but one of those during the cleanup phase.

This will probably have some effects together with https://github.com/ethereum/solidity/pull/11493, i.e. without this PR, some rare cases might fail, since they can't run the stack limit evader due to a removed memoryguard.

### DIFF
--- a/libyul/backends/evm/EVMDialect.cpp
+++ b/libyul/backends/evm/EVMDialect.cpp
@@ -179,7 +179,17 @@ map<YulString, BuiltinFunctionForEVM> createBuiltins(langutil::EVMVersion _evmVe
 			"memoryguard",
 			1,
 			1,
-			SideEffects{},
+			// Cannot be removed, but otherwise no side effects.
+			SideEffects{
+				true,
+				true,
+				false,
+				false,
+				true,
+				SideEffects::None,
+				SideEffects::None,
+				SideEffects::None
+			},
 			{LiteralKind::Number},
 			[](
 				FunctionCall const& _call,


libyul/backends/evm/EVMDialect.cpp
```diff
@@ -179,7 +179,17 @@ map<YulString, BuiltinFunctionForEVM> createBuiltins(langutil::EVMVersion _evmVe
...
-			SideEffects{},
+			// Cannot be removed, but otherwise no side effects.
+			SideEffects{
+				true,
+				true,
+				false,
+				false,
+				true,
+				SideEffects::None,
+				SideEffects:: None,
```
hrkrshnn:
```suggestion
				SideEffects::None,
```

hrkrshnn (commented):
Looks good. Tiny suggestion.

libyul/backends/evm/EVMDialect.cpp
ekpyron: Argh - no idea how stuff like this happens, thanks :-).

hrkrshnn (approved):
Looks good to me.

I think keeping it as `movable` is fine. I'm not sure if  optimizer currently makes extra copies of `memoryguard()` in the first place. Even if it does, as you said, the bytecode optimizer should deal with it.

ekpyron:
> Looks good to me.
> 
> I think keeping it as `movable` is fine. I'm not sure if optimizer currently makes extra copies of `memoryguard()` in the first place. Even if it does, as you said, the bytecode optimizer should deal with it.

I guess it's possible...
You start from
```
mstore(0x40, memoryguard(0x80))
let x := mload(0x40)
```
then potentially split this to
```
let _1 := memoryguard(0x80)
mstore(0x40, _1)
let x := mload(0x40)
```
resolve the load
```
let _1 := memoryguard(0x80)
mstore(0x40, _1)
let x := _1
```
And rematerialize and then you got two of them...
```
let _1 := memoryguard(0x80)
mstore(0x40, memoryguard(0x80))
let x := memoryguard(0x80)
```
and that might cascade... not sure if anything like that will actually happen, but I wouldn't be sure that it doesn't :-).

But yeah - bytecode should be unaffected and if we get weird and ugly ``pop(memoryguard(...))`` in the optimized Yul sources, we can counteract then...

For now it won't be emitted anyways, so it won't affect anything, but maybe better to already change this rather than wondering why stack-to-memory randomly doesn't work in some cases...

ekpyron:
Or actually, I'm growing less sure about this...
Does CSE actually check the ``removable`` condition? I mean, for this case it would actually be good, if it ignores it... but it may so far be an assumption of our side effects that ``movable`` always implies ``removable`` and maybe it's not a good idea to break that...

ekpyron:
Yeah... also ``let y := sub(memoryguard(0x80), memoryguard(0x80))`` is still happily reduced to zero after this change, without adding (two!) ``pop(memoryguard(0x80))``s...
This may still end up doing what we really want - I'm not sure I can construct a case, in which the ``memoryguard`` as we would emit it would actually be removed entirely after this... but for rather weird reasons :-).

ekpyron:
Closing this for now for further consideration. It's not particularly urgent, since (1) we don't emit the memoryguard yet anyways, (2) even if we do as long as we don't remove stores, it won't vanish anyways, (3) even if it vanishes, it will only be rare cases that will get stack errors and nothing bad happens otherwise.

PR #11893, opened 2021-09-03
Do not use the expression callback in memoryguard builtin code generation.

ekpyron:
The new code transform ignores the expression callback and handles function arguments on its own.

However, contrary to all other builtins with literal arguments, the literal argument of ``memoryguard`` actually still ends up on stack despite being literal. Hence instead of visiting the literal argument, I directly push the constant in the builtin.

An extension of this is https://github.com/ethereum/solidity/pull/11894, which removes the expression visit callback from the builtins entirely - if we agree on that one, we can close this one, but merging one of them will be required for https://github.com/ethereum/solidity/pull/11493

### DIFF
--- a/libyul/backends/evm/EVMDialect.cpp
+++ b/libyul/backends/evm/EVMDialect.cpp
@@ -26,8 +26,8 @@
 #include <libyul/Object.h>
 #include <libyul/Exceptions.h>
 #include <libyul/AsmParser.h>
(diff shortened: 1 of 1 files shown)


ekpyron:
Closing this in favour of https://github.com/ethereum/solidity/pull/11894, since we seem to agree that that's better.

PR #11894, opened 2021-09-03
Remove expression visit function from Yul Builtins.

ekpyron:
~~Extension of https://github.com/ethereum/solidity/pull/11893.~~
Entirely removes the visit function from Yul Builtins, shifting the responsibility of pushing the non-literal arguments to the code transform.

The builtins have no business defining how exactly argument evaluation takes place - and this actually simplifies things a lot.

Either this or https://github.com/ethereum/solidity/pull/11893 is required for https://github.com/ethereum/solidity/pull/11493

### DIFF
--- a/libyul/backends/evm/EVMCodeTransform.cpp
+++ b/libyul/backends/evm/EVMCodeTransform.cpp
@@ -239,9 +239,13 @@ void CodeTransform::operator()(FunctionCall const& _call)
 
 	m_assembly.setSourceLocation(extractSourceLocationFromDebugData(_call.debugData));
 	if (BuiltinFunctionForEVM const* builtin = m_dialect.builtin(_call.functionName.name))
-		builtin->generateCode(_call, m_assembly, m_builtinContext, [&](Expression const& _expression) {
-			visitExpression(_expression);
-		});
+	{
+		for (auto&& [i, arg]: _call.arguments | ranges::views::enumerate | ranges::views::reverse)
+			if (!builtin->literalArgument(i))
+				visitExpression(arg);
+		m_assembly.setSourceLocation(extractSourceLocationFromDebugData(_call.debugData));
+		builtin->generateCode(_call, m_assembly, m_builtinContext);
+	}
 	else
 	{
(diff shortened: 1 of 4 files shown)


libyul/backends/evm/EVMCodeTransform.cpp
```diff
@@ -239,9 +239,12 @@ void CodeTransform::operator()(FunctionCall const& _call)
...
 	m_assembly.setSourceLocation(extractSourceLocationFromDebugData(_call.debugData));
 	if (BuiltinFunctionForEVM const* builtin = m_dialect.builtin(_call.functionName.name))
-		builtin->generateCode(_call, m_assembly, m_builtinContext, [&](Expression const& _expression) {
-			visitExpression(_expression);
-		});
+	{
+		for (auto&& [i, arg]: _call.arguments | ranges::views::enumerate | ranges::views::reverse)
+			if (!builtin->literalArgument(i))
+				visitExpression(arg);
+		builtin->generateCode(_call, m_assembly, m_builtinContext);
```
chriseth, line 247: reset source location?

ekpyron: The builtin code looked like it's meant to take care of source locations - but it doesn't hurt to also set it here again as well, so I'll add it in any case. We could consider removing that part from ``generateCode`` then, but we can also keep it on both sides.

chriseth (dismissed):
Could move the source location setting, but fine like this as well.

libyul/backends/evm/EVMCodeTransform.cpp
ekpyron: Added

ekpyron:
> Could move the source location setting, but fine like this as well.

I just added it to the user-end of ``generateCode`` - should I remove it from the ``generateCode`` implementations or also keep it there?

ekpyron:
Removed setting the source location in the builtin code generator callback and added a comment about it being the caller's responsibility.

chriseth (dismissed):

chriseth:
Do you want to squash?

ekpyron:
Squashed.

chriseth (approved):

Issue #9583, opened 2020-08-06
[CLI] Rewrite the command line interface (solc) and consider using StandardCompiler as backend.


ekpyron:
CommandLineInterface.cpp is just one single huge mess right now :-). See other related issues #9579, #8722, #5306. 

For an initial implementation see #11960.

Tasks:
- [x] #6900 (to avoid the need to rewrite this JSON code)
- [ ] #10278 (to remove the need creating a translator to `--combined-json`)
- [x] #11507
- [ ] Support displaying error ids (this is optional, we could manually craft the error output, but if we want to use `formattedMessage` then this is needed)
- [ ] After doing this transition, the back-and-forth conversion done in `CommandLineParser` is no longer needed as the JSON should just expect the input strings

The rewrite is also a good opportunity to fix confusing options:
- [ ] #14364
- [ ] #13714

axic:
Strange there wasn't an issue for this, it has been one goal for me at least :+1:

cameel:
Of course there was :) #7813.
I'm closing it in favor of this one here. As short as its description is, it's still more detailed than in the old one :)

ekpyron:
IMHO another argument for this: as far as I'm concerned https://github.com/ethereum/solidity/pull/9954 should not have been done like that, but rather the command line interface be fixed.

chriseth:
@ekpyron we should do both.

axic:
Need to also keep #8847 in mind, such as:
> Some options available to the standard-json format do not seem to be available as CLI flags. I'm missing in particular source mappings for yul which was introduced only for the json options in #8378. It would be nice if feature parity between CLI flags and json opts was pursued as a general rule.

cameel:
From https://github.com/ethereum/solidity/pull/10241#pullrequestreview-526581636: I think there should be just one top-level try...catch block around everything in the CLI interface that catches exceptions not properly handled below, reports any diagnostic info it finds and exits. Currently there are multiple ad-hoc try...catch with very limited scope that look as if they were going to handle errors but all they don't actually do anything, they just report and exit and that should be delegated to the top level.

cameel:
I think we should really do #5303 before this and get full test coverage of all existing options. Otherwise we'll be dealing with random bugs that nobody noticed for months to come.

I have actually tried to start this recently by adding a boost test suite for `CommandLineInterface` class but it requires either making a lot of stuff public or splitting it into smaller components that can be tested independently.

ekpyron:
[...]

PR #11895, opened 2021-09-03
Do not run evm codegen if not asked.

ekpyron:
It's absolutely driving me nuts that the CLI always runs old codegen evm bytecode generation, even if I just ask for ``--ir``.
I have the CLI locally patched almost permanently because of that... (it's especially infuriating, if there are stack-too-deep errors, but I cannot even query the code that triggered it because of this)

Generally, I'd have said, we should get rid of the current CLI implementation entirely and replace it by a shiny fresh new (decidedly not generally backwards-compatible to crazy behaviour like this here ;-)) StandardCompiler-wrapper... but since for some reason we don't just do that, maybe we can at least patch this part :-).

I haven't checked that this is the exhaustive set of conditions that require evm bytecode generation, so it'd be good if someone double-checked.

### DIFF
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -594,6 +594,26 @@ bool CommandLineInterface::compile()
 
 		m_compiler->enableIRGeneration(m_options.compiler.outputs.ir || m_options.compiler.outputs.irOptimized);
 		m_compiler->enableEwasmGeneration(m_options.compiler.outputs.ewasm);
+		m_compiler->enableEvmBytecodeGeneration(
+			m_options.compiler.estimateGas ||
+			m_options.compiler.outputs.asm_ ||
+			m_options.compiler.outputs.asmJson ||
+			m_options.compiler.outputs.opcodes ||
+			m_options.compiler.outputs.binary ||
+			m_options.compiler.outputs.binaryRuntime ||
+			(m_options.compiler.combinedJsonRequests && (
+				m_options.compiler.combinedJsonRequests->binary ||
+				m_options.compiler.combinedJsonRequests->binaryRuntime ||
+				m_options.compiler.combinedJsonRequests->opcodes ||
+				m_options.compiler.combinedJsonRequests->asm_ ||
+				m_options.compiler.combinedJsonRequests->generatedSources ||
+				m_options.compiler.combinedJsonRequests->generatedSourcesRuntime ||
+				m_options.compiler.combinedJsonRequests->srcMap ||
+				m_options.compiler.combinedJsonRequests->srcMapRuntime ||
+				m_options.compiler.combinedJsonRequests->funDebug ||
+				m_options.compiler.combinedJsonRequests->funDebugRuntime
+			))
+		);
 
 		OptimiserSettings settings = m_options.optimizer.enabled ? OptimiserSettings::standard() : OptimiserSettings::minimal();
 		if (m_options.optimizer.expectedExecutionsPerDeployment.has_value())
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@ Language Features:
 
 
 Compiler Features:
+ * Commandline Interface: Do not implicitly run evm bytecode generation unless needed for the requested output.
  * Commandline Interface: Normalize paths specified on the command line and make them relative for files located inside base path.
  * Immutable variables can be read at construction time once they are initialized.
  * SMTChecker: Support low level ``call`` as external calls to unknown code.


solc/CommandLineInterface.cpp
```diff
@@ -594,6 +594,29 @@ bool CommandLineInterface::compile()
...
+					m_options.compiler.combinedJsonRequests->asm_ ||
+					m_options.compiler.combinedJsonRequests->generatedSources ||
+					m_options.compiler.combinedJsonRequests->generatedSourcesRuntime ||
+					m_options.compiler.combinedJsonRequests->srcMap ||
+					m_options.compiler.combinedJsonRequests->srcMapRuntime ||
+					m_options.compiler.combinedJsonRequests->funDebug ||
+					m_options.compiler.combinedJsonRequests->funDebugRuntime
+				)
+			)
+		);
```
cameel: Is that really enough? I think that for example `--gas` requires EVM compilation too. At least in Standard JSON `evm.gasEstimates` does enable the old codegen:

https://github.com/ethereum/solidity/blob/8447b32d52d5fb965e1a0ff696e8c52634dd4b76/libsolidity/interface/StandardCompiler.cpp#L278-L294

Also, in Standard JSON this check does not include `opcodes` and `metadata`. Is that a bug?

I think it might be a good idea to unify these two.

solc/CommandLineInterface.cpp
```diff
@@ -594,6 +594,29 @@ bool CommandLineInterface::compile()
...
 		m_compiler->enableIRGeneration(m_options.compiler.outputs.ir || m_options.compiler.outputs.irOptimized);
 		m_compiler->enableEwasmGeneration(m_options.compiler.outputs.ewasm);
+		m_compiler->enableEvmBytecodeGeneration(
+			m_options.compiler.outputs.asm_ ||
+			m_options.compiler.outputs.asmJson ||
+			m_options.compiler.outputs.opcodes ||
+			m_options.compiler.outputs.binary ||
+			m_options.compiler.outputs.binaryRuntime ||
+			m_options.compiler.outputs.metadata ||
+			(
```
```suggestion
			m_options.compiler.outputs.metadata || (
```

solc/CommandLineInterface.cpp
ekpyron: ``metadata`` is indeed not needed - ``opcodes`` might just not be available with Standard JSON, but I'm not sure.
And yes, for sure we should unify this - that's why I keep saying, that the CLI should be rewritten as thin wrapper around ``StandardCompiler`` ;-).

ekpyron: Also the gas estimates indeed need this - I added them to the list.

solc/CommandLineInterface.cpp
ekpyron: Not sure I correctly understand the comment, but it looks inconsistent with the coding style to me, doesn't it?
The block in parenthesis is basically the same as any of the previous clauses, why should I *not* break before it here, even though I have broken all other ``||`` clauses?
Funnily I would have skipped these parentheses entirely, but the compiler won't let me, even though it'd be correct :-).

axic:
😨

ekpyron:
> 😨 

What's up ;-)?

ekpyron:
Anyone confident enough to be able to tell if this is all the cases now :-)?

chriseth:
I would be more worried about this not being all cases at some point in the future.

ekpyron:
> I would be more worried about this not being all cases at some point in the future.

I'd hope that in the future this will all go through ``StandardCompiler`` :-). But yeah, there might very well be time in between...

solc/CommandLineInterface.cpp
cameel: > `opcodes` might just not be available with Standard JSON

I just tried `["evm.bytecode.opcodes"]` and it seems to work fine. Looks like there are some extra components are hidden in `evmObjectComponents()`:
- For `bytecode`: `object`, `opcodes`, `sourceMap`, `functionDebugData`, `generatedSources`, `linkReferences`
- For  `deployedBytecode`: `object`, `opcodes`, `sourceMap`, `functionDebugData`, `generatedSources`, `linkReferences`,  `immutableReferences`

So it does generate EVM bytecode for `opcodes` after all. `linkReferences` and  `immutableReferences` seem to be the only things you are missing but CLI does not support printing these currently as far as I can tell. So at least your list now matches Standard JSON, but, as @chriseth has pointed out, if will be very easy to forget about updating this if we ever add these extra keys in the CLI.

By the way, it  looks like output selection in Standard JSON is not completely validated. Invalid values like like `["evm.opc"]` or `["evm.abcdef"]` is not an error. Is this for compatibility or should we start validating them?

solc/CommandLineInterface.cpp
cameel: >  why should I not break before it here, even though I have broken all other `||` clauses?

Well, that's the style that feels intuitively correct to me and you even used it yourself on the `m_options.compiler.combinedJsonRequests && (` line, but if you want a more concrete justification, here's what the [CODING_STYLE](https://github.com/ethereum/solidity/blob/develop/CODING_STYLE.md) says:

> > If lines are broken, a list of elements enclosed with parentheses (of any kind) and separated by a separator (of any kind) are formatted such that there is exactly one element per line, followed by the separator, the opening parenthesis is on the first line, followed by a line break and the closing parenthesis is on a line of its own unindented). See example below.

This matches how we wrap function calls, `if`s and other parentheses as far as I can tell.

> Funnily I would have skipped these parentheses entirely, but the compiler won't let me, even though it'd be correct :-).

Yeah, I hate that too. I can see how operator precedence can be confusing in more obscure cases, and I sometimes add extra parentheses myself, but I don't understand why nowadays many compiler and linters insist on requiring parentheses in very clear cases like `&&` vs `||`. I've seen that tendency in Python too and it's really annoying :)

cameel:
It matches Standard JSON now so I think it covers all cases but I'm also worried about this getting out of sync in the future. I would be much more confident about approving this if `CompilerStack` had asserts against trying to retrieve stuff that was not generated. Or if we had tests covering all possible outputs. Then we could easily detect any mistakes. As it is now, something can easily slip through.

solc/CommandLineInterface.cpp
ekpyron: I mean, I really don't mind moving the parentheses here :-).
But I do in fact read the coding style differently here and the other line you're quoting is actually the one that's incorrect.

I.e. I start from:

```
m_compiler->enableEvmBytecodeGeneration(m_options.compiler.estimateGas || m_options.compiler.outputs.asm_ || m_options.compiler.outputs.asmJson || m_options.compiler.outputs.opcodes || m_options.compiler.outputs.binary || 	m_options.compiler.outputs.binaryRuntime || (m_options.compiler.combinedJsonRequests && (m_options.compiler.combinedJsonRequests->binary || m_options.compiler.combinedJsonRequests->binaryRuntime || m_options.compiler.combinedJsonRequests->opcodes || m_options.compiler.combinedJsonRequests->asm_ || m_options.compiler.combinedJsonRequests->generatedSources || m_options.compiler.combinedJsonRequests->generatedSourcesRuntime || m_options.compiler.combinedJsonRequests->srcMap || m_options.compiler.combinedJsonRequests->srcMapRuntime || m_options.compiler.combinedJsonRequests->funDebug || m_options.compiler.combinedJsonRequests->funDebugRuntime)));
```
Now I break the outermost parentheses. That means, the initial ``(`` is on one line, then one element each line followed by the separator and a line break. So I get:
```
m_compiler->enableEvmBytecodeGeneration(
	m_options.compiler.estimateGas ||
	m_options.compiler.outputs.asm_ ||
	m_options.compiler.outputs.asmJson ||
	m_options.compiler.outputs.opcodes ||
	m_options.compiler.outputs.binary ||
	m_options.compiler.outputs.binaryRuntime ||
	(m_options.compiler.combinedJsonRequests && (m_options.compiler.combinedJsonRequests->binary || m_options.compiler.combinedJsonRequests->binaryRuntime || m_options.compiler.combinedJsonRequests->opcodes || m_options.compiler.combinedJsonRequests->asm_ || m_options.compiler.combinedJsonRequests->generatedSources || m_options.compiler.combinedJsonRequests->generatedSourcesRuntime || m_options.compiler.combinedJsonRequests->srcMap || m_options.compiler.combinedJsonRequests->srcMapRuntime || m_options.compiler.combinedJsonRequests->funDebug || m_options.compiler.combinedJsonRequests->funDebugRuntime))
);
```

The inner parentheses are one element in the ``||``-separated list, so they do get their own line.

Now I break the ``&&`` separated list in the inner parentheses. Same thing again: the ``(`` stays on the first line, then one element each line followed by the separator and line break:

```
m_compiler->enableEvmBytecodeGeneration(
	m_options.compiler.estimateGas ||
	m_options.compiler.outputs.asm_ ||
	m_options.compiler.outputs.asmJson ||
	m_options.compiler.outputs.opcodes ||
	m_options.compiler.outputs.binary ||
	m_options.compiler.outputs.binaryRuntime ||
	(
		m_options.compiler.combinedJsonRequests &&
		(m_options.compiler.combinedJsonRequests->binary || m_options.compiler.combinedJsonRequests->binaryRuntime || m_options.compiler.combinedJsonRequests->opcodes || m_options.compiler.combinedJsonRequests->asm_ || m_options.compiler.combinedJsonRequests->generatedSources || m_options.compiler.combinedJsonRequests->generatedSourcesRuntime || m_options.compiler.combinedJsonRequests->srcMap || m_options.compiler.combinedJsonRequests->srcMapRuntime || m_options.compiler.combinedJsonRequests->funDebug || m_options.compiler.combinedJsonRequests->funDebugRuntime)
	)
);
```




Now I further break the innermost ``||``-separated list and get this:


```
m_compiler->enableEvmBytecodeGeneration(
	m_options.compiler.estimateGas ||
	m_options.compiler.outputs.asm_ ||
	m_options.compiler.outputs.asmJson ||
	m_options.compiler.outputs.opcodes ||
	m_options.compiler.outputs.binary ||
	m_options.compiler.outputs.binaryRuntime ||
	(
		m_options.compiler.combinedJsonRequests &&
		(
			m_options.compiler.combinedJsonRequests->binary ||
			m_options.compiler.combinedJsonRequests->binaryRuntime ||
			m_options.compiler.combinedJsonRequests->opcodes ||
			m_options.compiler.combinedJsonRequests->asm_ ||
			m_options.compiler.combinedJsonRequests->generatedSources ||
			m_options.compiler.combinedJsonRequests->generatedSourcesRuntime ||
			m_options.compiler.combinedJsonRequests->srcMap ||
			m_options.compiler.combinedJsonRequests->srcMapRuntime ||
			m_options.compiler.combinedJsonRequests->funDebug ||
			m_options.compiler.combinedJsonRequests->funDebugRuntime
		)
	)
);
```

Granted that's not exactly what I have in the PR, but it's not ``m_options.compiler.outputs.binaryRuntime || (\n`` either.

I could also *not* have broken the ``&&``-separated list, then I'd get:

```
m_compiler->enableEvmBytecodeGeneration(
	m_options.compiler.estimateGas ||
	m_options.compiler.outputs.asm_ ||
	m_options.compiler.outputs.asmJson ||
	m_options.compiler.outputs.opcodes ||
	m_options.compiler.outputs.binary ||
	m_options.compiler.outputs.binaryRuntime ||
	(m_options.compiler.combinedJsonRequests && (
		m_options.compiler.combinedJsonRequests->binary ||
		m_options.compiler.combinedJsonRequests->binaryRuntime ||
		m_options.compiler.combinedJsonRequests->opcodes ||
		m_options.compiler.combinedJsonRequests->asm_ ||
		m_options.compiler.combinedJsonRequests->generatedSources ||
		m_options.compiler.combinedJsonRequests->generatedSourcesRuntime ||
		m_options.compiler.combinedJsonRequests->srcMap ||
		m_options.compiler.combinedJsonRequests->srcMapRuntime ||
		m_options.compiler.combinedJsonRequests->funDebug ||
		m_options.compiler.combinedJsonRequests->funDebugRuntime
	))
);
```

I'll probably change the PR to that.

But I don't think your suggestion would actually follow the coding style - though I also wouldn't say that the coding style is overly clear on this :-).

ekpyron:
I don't think that worry is overly reasonable.
Whenever we would change anything here or add a new mode that would require evm bytecode generation, we'd add tests for it and/or at least somebody would run the thing once locally and would notice that it breaks due to not being added here.
And also, again: we should remove this entire set of logic anyways and reconstruct it as a thin wrapper of ``StandardCompiler`` anyways, which would solve the issue anyways.

Or  put differently: what precise realistic scenario are you worried about that would break anything due to this in the future?

solc/CommandLineInterface.cpp
ekpyron: I wonder how many inconsistencies in the code in these matters would be noticed by some kind of linting tool if we had any that supported this stuff properly :-).

chriseth:
True.

solc/CommandLineInterface.cpp
```diff
@@ -594,6 +594,26 @@ bool CommandLineInterface::compile()
 
 		m_compiler->enableIRGeneration(m_options.compiler.outputs.ir || m_options.compiler.outputs.irOptimized);
 		m_compiler->enableEwasmGeneration(m_options.compiler.outputs.ewasm);
+		m_compiler->enableEvmBytecodeGeneration(
+			m_options.compiler.estimateGas ||
```
chriseth, line 598: Not needed after https://github.com/ethereum/solidity/pull/11903 if experimental via ir is activated...

chriseth (approved):

axic:
> > 😨
>
> What's up ;-)?

That this is a monster ought to break. Really should just do #9583.

This wasn't meant as saying "not to merge it", but rather "I can't be convinced to review this" 😉

ekpyron:
Fair enough :-). Yeah, the sooner #9583 gets done the better, most definitely :-).

cameel:
@axic 
> That this is a monster ought to break. Really should just do #9583.

I got and idea today that we could break `CommandLineInterface` into a set of of smaller classes, each one dedicated to a single input mode (one for compiler, one for assembly, one for linker, etc.) and inheriting from a single base class with virtual methods like `processInput()`, `actOnInput()`, `formatOutput()`, etc. This would be a pretty easy refactor and then each part would be isolated enough to make rewriting it one part at a time pretty doable without making it a revolution. What do you think?

axic:
@cameel I actually just started to look into it, and the complexity does not seem to be in what you say, but the way all the parts are split into `handleXYZ`. Plus we are actually missing some features which are used by the CLI.

However your proposal definitely would improve the code and the readability of it.

cameel:
Well, the `handleXYZ` is all compiler mode and it's just output formatting. Splitting would not reduce the complexity of that part but it would at least let us deal with that part alone without other distractions (like assembly mode) so rewriting that part would still be less work than rewriting all of it. For me one of the complications has always been that it was not clear which parts run under which conditions and that changing stuff in one place could potentially affect stuff I'm not aware about. With smaller parts you can reasonably expect to be able to understand each part completely before you rewrite it :)

I mean, my point is that I don't think that any part of the CLI is very complex or hard to rewrite. It's just too big to do it in one go in reasonable time.

PR #11896, opened 2021-09-04
Simplifying CircleCi Config: Job aliases

whhygee:
Implements part of #11846.

> This PR is in continuation to #11868 (i.e. contains changes made in the mentioned PR)

Deduplication
- `steps_b_ubu` template has been created amongst Test Templates (Line 163), to store the steps duplicated in `b_ubu_asan`, `b_ubu_clang`, `b_ubu_asan_clang` and `b_ubu_ubsan_clang`.
- `b_ubu_clang` has been used as the alias in `b_ubu_asan_clang` and `b_ubu_ubsan_clang`, in order to deduplicate the steps, while overriding and merging necessary discrepancies.
- `test_osx` template has been created amongst Test Templates (Line 195), to store the commands duplicated in `t_osx_soltest` and `t_osx_cli`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -604,6 +604,15 @@ defaults:
       nodejs_version: '14'
       resource_class: medium
 
+  - job_b_ubu_asan_clang: &job_b_ubu_asan_clang
+      <<: *workflow_trigger_on_tags
+      name: b_ubu_asan_clang
+      cmake_options: -DSANITIZE=address
+  - job_b_ubu_ubsan_clang: &job_b_ubu_ubsan_clang
+      <<: *workflow_trigger_on_tags
+      name: b_ubu_ubsan_clang
+      cmake_options: -DSANITIZE=address
+
 # -----------------------------------------------------------------------------------------------
 jobs:
 
@@ -770,24 +779,18 @@ jobs:
       MAKEFLAGS: -j 10
     <<: *steps_build
 
-  b_ubu_asan_clang: &b_ubu_asan_clang
+  b_ubu_san_clang:
     # This runs a bit faster on large and xlarge but on nightly efficiency matters more.
+    parameters:
+      cmake_options:
+        type: string
     <<: *base_ubuntu2004_clang
     environment:
+      TERM: xterm
       CC: clang
       CXX: clang++
-      CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 3
-    <<: *steps_build
-
-  b_ubu_ubsan_clang: &b_ubu_ubsan_clang
-    # This runs a bit faster on large and xlarge but on nightly efficiency matters more.
-    <<: *base_ubuntu2004_clang
-    environment:
-      CC: clang
-      CXX: clang++
-      CMAKE_OPTIONS: -DSANITIZE=undefined
-      MAKEFLAGS: -j 3
+      CMAKE_OPTIONS: << parameters.cmake_options >>
     <<: *steps_build
 
   b_ubu_release: &b_ubu_release
@@ -1506,13 +1509,13 @@ workflows:
 
       # ASan build and tests
       - b_ubu_asan: *workflow_trigger_on_tags
-      - b_ubu_asan_clang: *workflow_trigger_on_tags
+      - b_ubu_san_clang: *job_b_ubu_asan_clang
       - t_ubu_asan_soltest: *workflow_ubuntu2004_asan
       - t_ubu_asan_clang_soltest: *workflow_ubuntu2004_asan_clang
       - t_ubu_asan_cli: *workflow_ubuntu2004_asan
 
       # UBSan build and tests
-      - b_ubu_ubsan_clang: *workflow_trigger_on_tags
+      - b_ubu_san_clang: *job_b_ubu_ubsan_clang
       - t_ubu_ubsan_clang_soltest: *workflow_ubuntu2004_ubsan_clang
       - t_ubu_ubsan_clang_cli: *workflow_ubuntu2004_ubsan_clang
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,6 +214,14 @@ defaults:
             command: ./test/lsp.py ./build/solc/solc
         - gitter_notify_failure_unless_pr
 
+  - steps_build: &steps_build
+      steps:
+        - checkout
+        - run: *run_build
+        - store_artifacts: *artifacts_solc
+        - persist_to_workspace: *artifacts_executables
+        - gitter_notify_failure_unless_pr
+
   - steps_soltest_all: &steps_soltest_all
       steps:
         - checkout
@@ -233,6 +241,14 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
         - gitter_notify_failure_unless_pr
+
+  - steps_install_dependencies_osx: &steps_install_dependencies_osx
+      steps:
+        - restore_cache:
+            keys:
+              - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
+        - attach_workspace:
+            at: .
 
   # --------------------------------------------------------------------------
   # Base Image Templates
@@ -748,12 +764,7 @@ jobs:
       CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 3
       CMAKE_BUILD_TYPE: Release
-    steps:
-      - checkout
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
-      - gitter_notify_failure_unless_pr
+    <<: *steps_build
 
   b_ubu_clang: &b_ubu_clang
     <<: *base_ubuntu2004_clang_large
@@ -762,12 +773,7 @@ jobs:
       CC: clang
       CXX: clang++
       MAKEFLAGS: -j 10
-    steps:
-      - checkout
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
-      - gitter_notify_failure_unless_pr
+    <<: *steps_build
 
   b_ubu_asan_clang: &b_ubu_asan_clang
     # This runs a bit faster on large and xlarge but on nightly efficiency matters more.
@@ -777,12 +783,7 @@ jobs:
       CXX: clang++
       CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 3
-    steps:
-      - checkout
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
-      - gitter_notify_failure_unless_pr
+    <<: *steps_build
 
   b_ubu_ubsan_clang: &b_ubu_ubsan_clang
     # This runs a bit faster on large and xlarge but on nightly efficiency matters more.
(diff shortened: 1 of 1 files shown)


cameel:
@yatharthagoenka Please rebase this on top of #11880. I changed the PR base to the branch from that PR so that github will itnore commits from it when showing the diff.

whhygee:
Branch has been rebased on top of #11880.

@cameel 
Also, in the jobs `b_bytecode_ubu`, `b_bytecode_osx` and `b_bytecode_win`, while the commands are same, all of them have differing values so if we use an anchor/alias like we did in (664c5dd), it would use up almost the same amount of space in order to override the alias' values to their respective ones.
Could you suggest a different method to approach deduplication in this task?

cameel:
Looks like github automatically closed this when the base PR was merged. Reopening.

I've been busy debugging the issues without binary hosting so I did not manage to review this (and some other PRs) yet. I'll try to get back to this tomorrow.

.circleci/config.yml
```diff
@@ -471,40 +488,21 @@ jobs:
...
 
   b_ubu_asan_clang: &b_ubu_asan_clang
-    docker:
-      - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
+    <<: *b_ubu_clang
     environment:
-      CC: clang
-      CXX: clang++
-      CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 3
```
cameel: This won't work. The `environment` dict will completely overwrite the same dict in `b_ubu_clang` instead of adding keys to it.

.circleci/config.yml
```diff
@@ -471,40 +488,21 @@ jobs:
...
   b_ubu_ubsan_clang: &b_ubu_ubsan_clang
-    docker:
-      - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
+    <<: *b_ubu_clang
     environment:
-      CC: clang
-      CXX: clang++
-      CMAKE_OPTIONS: -DSANITIZE=undefined
       MAKEFLAGS: -j 3
+      CMAKE_OPTIONS: -DSANITIZE=undefined
```
cameel: Same here. `CC` and `CXX` from `b_ubu_clang` will get overwritten.

.circleci/config.yml
```diff
@@ -471,40 +488,21 @@ jobs:
...
     steps:
-      - checkout
-      - run: *run_build
       - run: *gitter_notify_failure
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
```
cameel: `steps` will overwrite `steps` from `b_ubu_clang` as well. You need the trick with adding nested steps using `when`.

.circleci/config.yml
```diff
@@ -652,36 +650,19 @@ jobs:
...
+    <<: *test_osx
     steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - dependencies-osx-{{ checksum ".circleci/osx_install_dependencies.sh" }}
-      - attach_workspace:
-          at: .
       - run: *run_cmdline_tests
       - store_artifacts: *artifacts_test_results
```
cameel: Again, "inherited"  `environment` and `steps` will be overwritten.

cameel (dismissed):
Sorry, for not getting back to this earlier. The general idea is good but unfortunately the `<<: *` syntax in YAML does not merge dicts recursively so you need another solution to deduplicate these. Maybe parameterized jobs as I suggested in the issue.

.circleci/config.yml
whhygee: Oh alright, Noted! Thanks for pointing it out. I'll see to the necessary changes and revert back asap.

whhygee:
A parameterized job called `b_ubu_au_clang` was created to replace `b_ubu_asan_clang` and `b_ubu_ubsan_clang`.
The following references of the jobs were updated: 
- Workflow -> main -> jobs (Line 1106)
- Workflow -> nightly -> jobs (Line 1221 & 1229)

( @cameel Please verify the current implementation for parameterized jobs.)

Help Required: 
`b_ubu_asan_clang` and `b_ubu_ubsan_clang` have been referenced in the Workflow templates (239 and 244). How would we go about this reference using the parametrized job `b_ubu_au_clang`? (The current reference returns an error)

Aliases for `test_osx` were removed and the corresponding steps were nested using `when` condition.

.circleci/config.yml
```diff
@@ -471,40 +487,30 @@ jobs:
...
-      CMAKE_OPTIONS: -DSANITIZE=address
-      MAKEFLAGS: -j 3
-    steps:
-      - checkout
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
-
-  b_ubu_ubsan_clang: &b_ubu_ubsan_clang
+  b_ubu_au_clang:
```
cameel: Let's use `san` here. The job is meant to be used for sanitizers.
```suggestion
  b_ubu_san_clang:
```

.circleci/config.yml
```diff
@@ -471,40 +487,30 @@ jobs:
...
+      gitter_notify:
+        type: boolean
+        default: false
     docker:
       - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
     environment:
       CC: clang
       CXX: clang++
-      CMAKE_OPTIONS: -DSANITIZE=undefined
+      CMAKE_OPTIONS: << parameters.cmake_option >>
```
cameel: I'd make the parameter more specific.
```suggestion
      CMAKE_OPTIONS: -DSANITIZE=<< parameters.sanitizer >>
```
 `cmake_options` as a parameter would make sense too if this job was more generic but for now it's meant only for sanitizers.

.circleci/config.yml
```diff
@@ -471,40 +487,30 @@ jobs:
...
+      cmake_option:
+        type: string
+      gitter_notify:
+        type: boolean
+        default: false
     docker:
       - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
     environment:
       CC: clang
       CXX: clang++
```
cameel, line 791: We could also make this a parameter. This way we'd have one job for both GCC and clang:

```suggestion
      CC: <<# parameters.clang >> clang <</ parameters.clang >><<^ parameters.clang >> gcc <</ parameters.clang >>
      CXX: <<# parameters.clang >> clang++ <</ parameters.clang >><<^ parameters.clang >> g++ <</ parameters.clang >>
```

I wonder if CircleCI has more concise syntax for [boolean parameters](https://circleci.com/docs/2.0/reusing-config/#boolean) though.

cameel: This would also require an option for `MAKEFLAGS: -j 3`. It's 10 for the clang job because it runs on `resource_class: xlarge` so it can benefit from more cores.

In the worst case though, I think that having 10 in both cases would not be that bad. It would not get any speedup over 3 but it should not make it noticeably slower.

.circleci/config.yml
```diff
@@ -471,40 +487,30 @@ jobs:
...
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
-
-  b_ubu_ubsan_clang: &b_ubu_ubsan_clang
+  b_ubu_au_clang:
+    parameters:
+      cmake_option:
+        type: string
+      gitter_notify:
```
cameel: I'd call it `gitter_notify_on_failure` because, unlike in `t_ems_ext`, it does not notify on success.

.circleci/config.yml
```diff
@@ -1102,6 +1102,12 @@ workflows:
       - b_ubu_clang: *workflow_trigger_on_tags
       - t_ubu_clang_soltest: *workflow_ubuntu2004_clang
 
+      - b_ubu_au_clang:
+          cmake_option: -DSANITIZE=address
+      - b_ubu_au_clang:
+          cmake_option: -DSANITIZE=undefined
+          gitter_notify: true
+
```
cameel: It's fine to add these jobs here to test the PR, but please remember to remove them before we merge this. They're included only in nightly runs because we do not want to wait for them in every PR.

Since you have them here, could you look at the execution times with `MAKEFLAGS: -j3` and `MAKEFLAGS: -j10`? If they're similar, we could just use 10 everywhere and remove more boilerplate this way.

.circleci/config.yml
```diff
@@ -236,12 +252,15 @@ defaults:
   - workflow_ubuntu2004_asan_clang: &workflow_ubuntu2004_asan_clang
       <<: *workflow_trigger_on_tags
       requires:
-        - b_ubu_asan_clang
+        - b_ubu_au_clang:
+            cmake_option: -DSANITIZE=address
```
cameel: > `b_ubu_asan_clang` and `b_ubu_ubsan_clang` have been referenced in the Workflow templates (239 and 244). How would we go about this reference using the parametrized job `b_ubu_au_clang`? (The current reference returns an error)

In [Reusable Config Reference Guide > Invoking the same job multiple times](https://circleci.com/docs/2.0/reusing-config/#invoking-the-same-job-multiple-times) I see that in such case the solution is to explicitly assign names to jobs on the `jobs` list. Then refer to them using that name.

So if you do this:
```yaml
      - b_ubu_au_clang:
          <<: *workflow_trigger_on_tags
          name: b_ubu_asan_clang
          cmake_option: -DSANITIZE=address
```
you won't have to change anything in `workflow_ubuntu2004_asan_clang`.

.circleci/config.yml
```diff
@@ -457,11 +476,8 @@ jobs:
       CMAKE_OPTIONS: -DSANITIZE=address
       MAKEFLAGS: -j 10
       CMAKE_BUILD_TYPE: Release
-    steps:
-      - checkout
-      - run: *run_build
-      - store_artifacts: *artifacts_solc
-      - persist_to_workspace: *artifacts_executables
+    <<: *steps_b_ubu
```
cameel: You could use these steps for `b_ubu` as well. `store_artifacts: *artifacts_tools` can be done separately after these steps or you could just include it in `steps_b_ubu`. Storing built tools from sanitizer runs is not necessary but if it simplifies the config, I think it won't hurt.

.circleci/config.yml
cameel: Oh, I see that clang and gcc jobs use different base images. So it would have to be another parameter.

Or you could leave it as is but instead unify `b_ubu_au_clang` and `b_ubu_clang`. Adding an empty `CMAKE_OPTIONS` in `b_ubu_clang` should not hurt.

.circleci/config.yml
cameel: Unless you decide to go for https://github.com/ethereum/solidity/pull/11896#discussion_r711180445. Then leaving it as is would make sense.

cameel:
@yatharthagoenka Are you still working on this? Do you need any help?

whhygee:
Hey @cameel, I apologise for the inactivity. I was at a remote location for the while, would be resuming the work right away.

cameel:
@yatharthagoenka Do you mind if I take over and finish this PR?

Let me know if you still want to finish it yourself but I did some conflicting changes in #12173 so if you want to continue, please rebase on that PR.

chriseth:
@cameel is this still useful to keep open?

cameel:
It's unfortunately very outdated after all the CI changes that went in in the meantime. I think a thing or two could be salvaged from it though.

I'll take it over.

.circleci/config.yml
cameel: Done.

.circleci/config.yml
cameel: I left the more generic version for now.

.circleci/config.yml
cameel: We now notify by default, even in the main workflow, so this parameter is no longer needed. I removed it.

.circleci/config.yml
cameel: Done.

cameel:
This is now ready. Marking as draft though because I still need to remove the debug commit from the top once I make sure taht the CI passes.

I thought there wouldn't be that much left from this after rebasing it but suprisingly all the changes are still relevant and this makes the config much neater for some jobs. The refactors that went into `develop` in the meantime also make the diff here much more readable than it used to be. Thanks for the PR @yatharthagoenka!

chriseth (approved):

PR #11897, opened 2021-09-04
docs: delete stale note about reading code

PaulRBerg:
As per the docs on [Members of Address Types](https://docs.soliditylang.org/en/v0.8.7/units-and-global-variables.html#members-of-address-types), it is now possible to read the code of a contract using high-level Solidity.


hrkrshnn (approved):
Thanks for the PR!

PR #11903, opened 2021-09-06
Print @ast-id

chriseth:


chriseth:
This adds `@ast-id` debug info, which is yet unused. For the AsmParser we currently have the following behaviour:
If `@use-src` is present on the object, the source location is taken from the `@src` annotations.

Do we also want to use this discriminator for whether or not we process `@ast-id` annotations? I think we should always process them. Although I think this means they could also be used in inline assembly - maybe we should at least prevent that.

chriseth:
Rebased - the parser could still be refactored a bit.

libyul/AsmPrinter.cpp
```diff
@@ -261,7 +261,6 @@ string AsmPrinter::appendTypeName(YulString _type, bool _isBoolLiteral) const
 string AsmPrinter::formatSourceLocationComment(
```
cameel: It no longer produces a comment so I think it should be renamed:
```suggestion
string AsmPrinter::formatSourceLocation(
```

libyul/AsmPrinter.cpp
```diff
@@ -294,27 +293,38 @@ string AsmPrinter::formatSourceLocationComment(
...
+		));
+	}
 
-	return formatSourceLocationComment(
-		_debugData->location,
-		m_nameToSourceIndex,
-		_statement,
-		m_soliditySourceProvider
-	) + (_statement ? "\n" : "");
+	string comment = joinHumanReadable(items, " ");
```
cameel:
```suggestion
	string commentBody = joinHumanReadable(items, " ");
```

libsolidity/codegen/ir/Common.cpp
```diff
@@ -135,10 +135,9 @@ string dispenseLocationComment(langutil::SourceLocation const& _location, IRGene
 {
 	solAssert(_location.sourceName, "");
 	_context.markSourceUsed(*_location.sourceName);
-	return AsmPrinter::formatSourceLocationComment(
+	return "/// " + AsmPrinter::formatSourceLocationComment(
```
cameel: Note that `formatSourceLocationComment()` does not include the `@ast-id`. Only `formatDebugData()` does. Is that intentional?

If you only print location here, you will not get `@ast-id` in `--ir` output. It will be present only in `--ir-optimized` (and possibly inline assembly) because only it comes from `AsmPrinter`.

Unless of course `@ast-id` is added to IR output manually. And I see it is but only for functions. I think that because of this `--ir` will at least have much fewer `@ast-id` tags than `--ir-optimized`.

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -340,6 +340,7 @@ string IRGenerator::generateFunction(FunctionDefinition const& _function)
 	return m_context.functionCollector().createFunction(functionName, [&]() {
 		m_context.resetLocalVariables();
 		Whiskers t(R"(
+			/// @ast-id <astID>
```
cameel, line 343: What about `generateConstructors()`? Don't we want AST ID for the functions it generates?

test/cmdlineTests/yul_source_locations/output.json
```diff
@@ -325,6 +325,7 @@ object \"C_54\" {
                 sum := add(x, y)
             }
 
+            /// @ast-id 30
```
cameel, line 349: It seems that only `--ir` output was affected in tests. There are no `@ast-id` annotations added in `--ir-optimized`. I think that's because overall very few tests enable `--ir-optimized`. But still, from `AsmPrinter::formatDebugData()` I'd expect basically every `@src` comment in optimized output to get the `@ast-id` annotation and this does not seem to be the case. Why?

Can you enable `--ir-optimized` in this test? Also, it's a Standard JSON test so I think it should be named `standard_yul_source_locations`. I actually had to do this in my own PR to be able to create a non-JSON equivalent.

libsolidity/codegen/ir/Common.cpp
chriseth: This is intentional, the ast-id is printed only for specific solidity ast nodes and is done in the IRGenerator directly. Or did I misunderstand something here?

test/cmdlineTests/yul_source_locations/output.json
chriseth: It seems the optimizer prunes most of the ast ids, I can investigate a little if we can prevent it, but I think this was to be expected.

chriseth: Oh, this is actually due to the fact that the ast-id parsing is not implemented in this PR, so the optimized will always remove it.

libsolidity/codegen/ir/Common.cpp
cameel: ~But it means that if you read the IR and then print it again with `AsmPrinter`, you won't ever see `@ast-id`s, even after you start parsing them from IR. Unless of course you change this part in the PR that adds parsing but I see no reason not to do that already here.~

EDIT: Ah, It's fine after all. It's the parsed and printed output that has them and IR has them inserted manually.

cameel: You're right. it's fine.

cameel: > I think that because of this `--ir` will at least have much fewer `@ast-id` tags than `--ir-optimized`.

With this I assumed that the AST ID would be printed whenever `@src` gets printed. But now I realize that it will only be there if it was explicitly added to debug data.

chriseth:
Based this on top of #11971  to make the tests work.

test/cmdlineTests/yul_source_locations/output.json
cameel: Ok, then it's fine I guess.

chriseth:
@cameel now the annotations are preserved after optimization.

cameel (approved):
Looks fine now.

Maybe it would make sense to add some getters to the test to check that case too though?

chriseth:
@cameel I'll add getters to the "function debug info" PR.

chriseth:
actually it's not useful to do this in the "function debug info" because the getter is not analyzed internally, and it is fully inlined in the optimizer, so I'll add something here.

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -819,6 +830,10 @@ void IRGenerator::generateConstructors(ContractDefinition const& _contract)
 				for (ASTPointer<VariableDeclaration> const& varDecl: contract->constructor()->parameters())
 					params += m_context.addLocalVariable(*varDecl).stackSlots();
 
+			if (contract->constructor())
+				t("astIDComment", "/// @ast-id " + to_string(contract->constructor()->id()));
+			else
+				t("astIDComment", "");
```
Marenz, line 835: Would it make sense to add the "\n" for the `astIDComment` so that we only have a new line when there is content?

Marenz (dismissed):

libsolidity/codegen/ir/IRGenerator.cpp
chriseth: Yeah, thought about that, too. Will change.

cameel (dismissed):

PR #11905, opened 2021-09-06
[SMTChecker] Cleanup spurious messages about TypeTypes

### LINKED ISSUE #11884: [SMTChecker] Support abi.decode completely
For
```solidity
require(abi.decode(returndata, (bool)), "SafeERC20: ERC20 operation did not succeed");
```
the SMTChecker reports `Warning: Assertion checker does not yet implement type type(bool)`.

leonardoalt:
Fix https://github.com/ethereum/solidity/issues/11884

### DIFF
--- a/libsolidity/formal/SMTEncoder.cpp
+++ b/libsolidity/formal/SMTEncoder.cpp
@@ -560,6 +560,24 @@ bool SMTEncoder::visit(Conditional const& _op)
 	return false;
 }
 
+bool SMTEncoder::visit(FunctionCall const& _funCall)
+{
+	auto functionCallKind = *_funCall.annotation().kind;
+	if (functionCallKind != FunctionCallKind::FunctionCall)
+		return true;
+
+	FunctionType const& funType = dynamic_cast<FunctionType const&>(*_funCall.expression().annotation().type);
+	// We do not want to visit the TypeTypes in the second argument of `abi.decode`.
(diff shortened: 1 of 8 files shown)


leonardoalt:
TypeTypes are used only in `abi.decode` calls, where they are visited directly and it does not end up in `visit(Identifier)`. So I think the cleanest way to avoid trying to SMT visit those TypeTypes is just to not visit them when we see a `abi.decode`.

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -560,6 +560,23 @@ bool SMTEncoder::visit(Conditional const& _op)
...
 }
 
+bool SMTEncoder::visit(FunctionCall const& _funCall)
+{
+	auto functionCallKind = *_funCall.annotation().kind;
+	if (functionCallKind != FunctionCallKind::FunctionCall)
+		return true;
+
+	FunctionType const& funType = dynamic_cast<FunctionType const&>(*_funCall.expression().annotation().type);
+	// We do not want to visit the TypeTypes in the second argument of `abi.decode`.
```
axic, line 570: If it is visited/inspected somewhere else, please leave a note to that effect here.

leonardoalt: Done

axic (approved):

PR #11908, opened 2021-09-07
Run the optimizer only once.

### LINKED ISSUE #11898: Non-deterministic source maps for the same contract when other contract imports it
## Description

Solidity generates slightly different source maps when:

1. The contract is not imported by any other contract in the code base.
2. The exact same contract is imported by other contract in the code base, which is compiled as well.

In the second case, there is an extra semicolon in the source maps. As per the [docs](https://docs.soliditylang.org/en/v0.8.7/internals/source_mappings.html), this *should* refer to an additional assembly instruction in the contract. But there is no such instruction, thus this may be a bug. I expected the source maps to be identical between the two circumstances, given that all else is equal (compiler settings, source code, whitespaces, etc).

Here are two GitHub Gists with the different source maps:

- [1st case](https://gist.github.com/paulrberg/b250fbb6d5aa92ae27381071aa1a8e61)
- [2nd case](https://gist.github.com/paulrberg/7f70bf4415d7ffd51645507fbba63e93)

If you run these on [text-compare.com](https://text-compare.com), you will notice that there are more diffs than just an extra semicolon. But you can ignore all of them, except the semicolon, because they refer to the different file ids that get assigned to `PRBProxy` and `#utility.yul` by Solidity. Check out characters 317-320th; in the 1st case, you find `4;;;2`, and in the 2nd case `8;;;;`.

## Environment

- Compiler version: solc-macosx-amd64-v0.8.7+commit.e28d00a7
- Target EVM version (as per compiler settings): Compiler default (London)
- EVM execution environment / backend / blockchain client: Hardhat Network (hardhat@2.6.2)
- Operating system: macOS@11.5.2

## Steps to Reproduce

1. Git clone [prb-proxy](https://github.com/paulrberg/prb-proxy/commit/93d337b3fa66fe03ccf3e355b5da490bf5bba273) at commit `93d337b` 
2. Install dependencies with `yarn install`
3. Fill a `.env` file as shown by the `.env.example` file
4. Generate the TypeChain bindings with `yarn typechain`
5. Run the tests with `yarn test --verbose`

You should get the following error:

```
  hardhat:core:hardhat-network:node Could not generate stack trace. Please report this to help us improve Hardhat.
  hardhat:core:hardhat-network:node  Error: This shouldn't happen: a call trace has no functionJumpdest but has already jumped into a function
    at ErrorInferrer._getEntryBeforeFailureInModifier (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/error-inferrer.ts:1476:13)
    at ErrorInferrer._checkRevertOrInvalidOpcode (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/error-inferrer.ts:370:16)
    at ErrorInferrer._checkLastInstruction (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/error-inferrer.ts:554:44)
    at ErrorInferrer.inferAfterTracing (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/error-inferrer.ts:162:12)
    at SolidityTracer._rawTraceEvmExecution (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/solidityTracer.ts:211:61)
    at SolidityTracer._traceEvmExecution (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/solidityTracer.ts:149:29)
    at SolidityTracer._getCallMessageStackTrace (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/solidityTracer.ts:72:17)
    at SolidityTracer.getStackTrace (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/stack-traces/solidityTracer.ts:56:19)
    at HardhatNode._manageErrors (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:1705:41)
    at HardhatNode.estimateGas (/Users/paulrberg/workspace/pad/prb-proxy-clone/node_modules/hardhat/src/internal/hardhat-network/provider/node.ts:661:27)
```

## Implications

This is likely what causes https://github.com/nomiclabs/hardhat/issues/1185. The error above is the same as the one reported by everyone else there.

## More Details

I spent three days debugging this. What I know thus far:

- The issue goes away when I manually remove the extra semicolon the source maps
- The issue goes away when I delete the `PRBProxyFactory.sol` and `PRBProxyRegistry.sol` contracts and compile just `PRBProxy.sol` and its dependencies
- The issue goes away when I rename the `execute` function to something else, such as `fooBarBaz`
- The issue goes away when I turn off the compiler optimizer
- Seems this has something to do with the modifier (`onlyOwnable`) and the `JUMPTDEST` opcode

chriseth:
Fixes https://github.com/ethereum/solidity/issues/11898

### DIFF
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -409,18 +409,21 @@ Assembly& Assembly::optimise(OptimiserSettings const& _settings)
 	return *this;
 }
 
-map<u256, u256> Assembly::optimiseInternal(
+map<u256, u256> const& Assembly::optimiseInternal(
 	OptimiserSettings const& _settings,
 	std::set<size_t> _tagsReferencedFromOutside
 )
 {
+	if (m_tagReplacements)
+		return *m_tagReplacements;
+
 	// Run optimisation for sub-assemblies.
 	for (size_t subId = 0; subId < m_subs.size(); ++subId)
 	{
 		OptimiserSettings settings = _settings;
 		// Disable creation mode for sub-assemblies.
 		settings.isCreation = false;
-		map<u256, u256> subTagReplacements = m_subs[subId]->optimiseInternal(
+		map<u256, u256> const& subTagReplacements = m_subs[subId]->optimiseInternal(
 			settings,
 			JumpdestRemover::referencedTags(m_items, subId)
 		);
@@ -546,7 +549,8 @@ map<u256, u256> Assembly::optimiseInternal(
 			*this
 		);
 
-	return tagReplacements;
+	m_tagReplacements = move(tagReplacements);
+	return *m_tagReplacements;
 }
 
 LinkerObject const& Assembly::assemble() const
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Opcode Optimizer: Prevent the optimizer from running multiple times to avoid potential bytecode differences for referenced code.
  * Name Resolver: Fix that when importing an aliased symbol using ``import {AliasedName} from "a.sol"`` it would use the original name of the symbol and not the aliased one.
  * SMTChecker: Fix false negative caused by ``push`` on storage array references returned by internal functions.
  * SMTChecker: Fix false positive in external calls from constructors.
--- a/libevmasm/Assembly.h
+++ b/libevmasm/Assembly.h
@@ -165,7 +165,7 @@ protected:
 	/// Does the same operations as @a optimise, but should only be applied to a sub and
 	/// returns the replaced tags. Also takes an argument containing the tags of this assembly
 	/// that are referenced in a super-assembly.
-	std::map<u256, u256> optimiseInternal(OptimiserSettings const& _settings, std::set<size_t> _tagsReferencedFromOutside);
+	std::map<u256, u256> const& optimiseInternal(OptimiserSettings const& _settings, std::set<size_t> _tagsReferencedFromOutside);
 
 	unsigned bytesRequired(unsigned subTagSize) const;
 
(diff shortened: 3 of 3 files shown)


ekpyron (approved):

libevmasm/Assembly.cpp
```diff
@@ -409,18 +409,21 @@ Assembly& Assembly::optimise(OptimiserSettings const& _settings)
...
 }
 
-map<u256, u256> Assembly::optimiseInternal(
+map<u256, u256> const& Assembly::optimiseInternal(
 	OptimiserSettings const& _settings,
 	std::set<size_t> _tagsReferencedFromOutside
 )
 {
+	if (m_tagReplacements)
+		return *m_tagReplacements;
```
hrkrshnn, line 418: Not sure about this. If there are  2 sub assemblies, after the first subassembly is optimized, the second subassembly will never be optimized; also the second one would return replacements from the first. Is that right?

chriseth: This member is assigned from `tagReplacements` defined in line 431/434, which is unrelated to subassemblies.

Of course the assumption here is that you do not add or remove sub-assemblies once the optimizer has been called...

hrkrshnn (approved):

PR #11909, opened 2021-09-07
Template code for disallowing options in input modes that do not support them (handles `--error-recovery`)

Midhun07:
I have disallowed `--error-recovery` option with input modes Standard json, Assembly and Linker modes. Also this is a template for all the other invalid options mentioned in the issue #11629.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -17,6 +17,7 @@ Bugfixes:
  * Commandline Interface: Fix extra newline character being appended to sources passed through standard input, affecting their hashes.
  * Commandline Interface: Report output selection options unsupported by the selected input mode instead of ignoring them.
  * Commandline Interface: Don't return zero exit code when writing linked files to disk fails.
+ * Commandline Interface: Disallow ``--error-recovery`` option outside of the compiler mode.
  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
  * TypeChecker: Fix internal error when using user defined value types in public library functions.
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
--- a/solc/CommandLineParser.cpp
+++ b/solc/CommandLineParser.cpp
@@ -898,13 +898,21 @@ bool CommandLineParser::processArgs()
 	else
 		m_options.input.mode = InputMode::Compiler;
 
-	if (
-		m_args.count(g_strExperimentalViaIR) > 0 &&
-		m_options.input.mode != InputMode::Compiler &&
-		m_options.input.mode != InputMode::CompilerWithASTImport
-	)
+	map<string, set<InputMode>> validOptionInputModeCombinations = {
+		// TODO: This should eventually contain all options.
+		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+	};
+	vector<string> invalidOptionsForCurrentInputMode;
+	for (auto const& [optionName, inputModes]: validOptionInputModeCombinations)
 	{
-		serr() << "The option --" << g_strExperimentalViaIR << " is only supported in the compiler mode." << endl;
+		if (m_args.count(optionName) > 0 && inputModes.count(m_options.input.mode) == 0)
+			invalidOptionsForCurrentInputMode.push_back(optionName);
+	}
+
+	if (!invalidOptionsForCurrentInputMode.empty())
+	{
+		serr() << "The following options are not supported in the current input mode: " << joinOptionNames(invalidOptionsForCurrentInputMode) << endl;
 		return false;
 	}
 
--- a/test/solc/CommandLineParser.cpp
+++ b/test/solc/CommandLineParser.cpp
@@ -267,7 +267,6 @@ BOOST_AUTO_TEST_CASE(assembly_mode_options)
 			"--include-path=/home/user/include",
 			"--allow-paths=/tmp,/home,project,../contracts",
 			"--ignore-missing",
-			"--error-recovery",            // Ignored in assembly mode
 			"--overwrite",
 			"--evm-version=spuriousDragon",
 			"--revert-strings=strip",      // Accepted but has no effect in assembly mode
@@ -356,7 +355,6 @@ BOOST_AUTO_TEST_CASE(standard_json_mode_options)
 		"--include-path=/home/user/include",
 		"--allow-paths=/tmp,/home,project,../contracts",
 		"--ignore-missing",
-		"--error-recovery",                // Ignored in Standard JSON mode
 		"--output-dir=/tmp/out",           // Accepted but has no effect in Standard JSON mode
 		"--overwrite",                     // Accepted but has no effect in Standard JSON mode
 		"--evm-version=spuriousDragon",    // Ignored in Standard JSON mode
@@ -413,30 +411,25 @@ BOOST_AUTO_TEST_CASE(standard_json_mode_options)
 	BOOST_TEST(parsedOptions.value() == expectedOptions);
 }
 
-BOOST_AUTO_TEST_CASE(experimental_via_ir_invalid_input_modes)
+BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
 {
-	static array<string, 5> const inputModeOptions = {
-		"--assemble",
-		"--yul",
-		"--strict-assembly",
-		"--standard-json",
-		"--link",
+	map<string, vector<string>> invalidOptionInputModeCombinations = {
+		// TODO: This should eventually contain all options.
+		{"--error-recovery", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--experimental-via-ir", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}}
 	};
-	for (string const& inputModeOption: inputModeOptions)
-	{
-		stringstream sout, serr;
-		vector<string> commandLine = {
-			"solc",
-			"--experimental-via-ir",
-			"file",
-			inputModeOption,
-		};
-		optional<CommandLineOptions> parsedOptions = parseCommandLine(commandLine, sout, serr);
 
-		BOOST_TEST(sout.str() == "");
-		BOOST_TEST(serr.str() == "The option --experimental-via-ir is only supported in the compiler mode.\n");
-		BOOST_REQUIRE(!parsedOptions.has_value());
-	}
+	for (auto const& [optionName, inputModes]: invalidOptionInputModeCombinations)
+		for (string const& inputMode: inputModes)
+		{
+			stringstream sout, serr;
+			vector<string> commandLine = {"solc", optionName, "file", inputMode};
+			optional<CommandLineOptions> parsedOptions = parseCommandLine(commandLine, sout, serr);
+
+			BOOST_TEST(sout.str() == "");
+			BOOST_TEST(serr.str() == "The following options are not supported in the current input mode: " + optionName + "\n");
+			BOOST_REQUIRE(!parsedOptions.has_value());
+		}
 }
 
 BOOST_AUTO_TEST_SUITE_END()


solc/CommandLineParser.cpp
```diff
@@ -927,13 +927,22 @@ General Information)").c_str(),
 	else
 		m_options.input.mode = InputMode::Compiler;
 
-	if (
-		m_args.count(g_strExperimentalViaIR) > 0 &&
-		m_options.input.mode != InputMode::Compiler &&
-		m_options.input.mode != InputMode::CompilerWithASTImport
-	)
+	map<string, vector<InputMode>> validOption_InputModeCombinations = {
```
```suggestion
	map<string, vector<InputMode>> validOptionInputModeCombinations = {
```

solc/CommandLineParser.cpp
```diff
@@ -927,13 +927,22 @@ General Information)").c_str(),
...
+		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+	};
+	string invalidOptions;
+	for (auto item: validOption_InputModeCombinations)
```
```suggestion
	for (auto const& [optionName, inputModes]: validOptionInputModeCombinations)
```

solc/CommandLineParser.cpp
```diff
@@ -927,13 +927,22 @@ General Information)").c_str(),
...
+	{
+		if (m_args.count(item.first) > 0 && find(item.second.begin(), item.second.end(), m_options.input.mode) == item.second.end())
+		{
+			invalidOptions.append("--" + item.first + " ");
+		}
+	}
```
```suggestion
		if (m_args.count(item.first) > 0 && find(item.second.begin(), item.second.end(), m_options.input.mode) == item.second.end())
			invalidOptions.append("--" + item.first + " ");
```

solc/CommandLineParser.cpp
cameel: I'd use a `set` here. Checking if option is in there will be more efficient and also more concise to write (you can just call `count()` on the set).
```suggestion
	map<string, set<InputMode>> validOption_InputModeCombinations = {
```

solc/CommandLineParser.cpp
```diff
@@ -927,13 +927,22 @@ General Information)").c_str(),
...
+
+	if (!invalidOptions.empty())
 	{
-		serr() << "The option --" << g_strExperimentalViaIR << " is only supported in the compiler mode." << endl;
+		serr() << "The options " << invalidOptions << "is only supported in the compiler mode." << endl;
 		return false;
 	}
```
cameel: Currently this prints a separate message for each option. I think it would be better to have one message that just lists all the options which were used but are invalid in the current mode.

test/solc/CommandLineParser.cpp
```diff
@@ -429,20 +427,29 @@ BOOST_AUTO_TEST_CASE(experimental_via_ir_invalid_input_modes)
 		"--standard-json",
```
cameel: Test name should be adjusted. Currently it's `experimental_via_ir_invalid_input_modes`.

test/solc/CommandLineParser.cpp
```diff
@@ -429,20 +427,29 @@ BOOST_AUTO_TEST_CASE(experimental_via_ir_invalid_input_modes)
...
+				"file",
+				inputModeOption,
+			};
+			optional<CommandLineOptions> parsedOptions = parseCommandLine(commandLine, sout, serr);
 
-		BOOST_TEST(sout.str() == "");
-		BOOST_TEST(serr.str() == "The option --experimental-via-ir is only supported in the compiler mode.\n");
-		BOOST_REQUIRE(!parsedOptions.has_value());
+			BOOST_TEST(sout.str() == "");
+			BOOST_TEST(serr.str() == "The options " + invalidOption + " is only supported in the compiler mode.\n");
```
```suggestion
			BOOST_TEST(serr.str() == "The option " + invalidOption + " is only supported in the compiler mode.\n");
```

solc/CommandLineParser.cpp
```diff
@@ -927,13 +927,22 @@ General Information)").c_str(),
...
+		if (m_args.count(item.first) > 0 && find(item.second.begin(), item.second.end(), m_options.input.mode) == item.second.end())
+		{
+			invalidOptions.append("--" + item.first + " ");
+		}
+	}
+
+	if (!invalidOptions.empty())
 	{
-		serr() << "The option --" << g_strExperimentalViaIR << " is only supported in the compiler mode." << endl;
+		serr() << "The options " << invalidOptions << "is only supported in the compiler mode." << endl;
```
```suggestion
		serr() << "The option " << invalidOptions << " is only supported in the compiler mode." << endl;
```

cameel: I think we'll have easier time testing this if we change this message to say that the option is unsupported in the current mode. Otherwise the test will have to make assumptions about modes where the function is supported.

cameel (changes requested):
Some changes are needed.

Also, please remember about the changelog entry. Since the previous PR has not been released yet, I'd just update the previous entry and list the new option here.

cameel:
@Midhun07 Are you working on this? Do you need help?

We have a ton of pending PRs right now and we'd like trim that number down a bit. If you don't have time for it, I can take over and finish it.

Midhun07:
@cameel Sorry was caught up due to personal reasons I will start working on this and submit the changes by today

solc/CommandLineParser.cpp
Midhun07: This prints all the invalid options in same message. invalidOptions is a string with all the invalid options

Midhun07:
Hi @cameel I have made the requested changes. Please review.

solc/CommandLineParser.cpp
```diff
@@ -840,13 +840,20 @@ bool CommandLineParser::processArgs()
 	else
 		m_options.input.mode = InputMode::Compiler;
 
-	if (
-		m_args.count(g_strExperimentalViaIR) > 0 &&
-		m_options.input.mode != InputMode::Compiler &&
-		m_options.input.mode != InputMode::CompilerWithASTImport
-	)
+	map<string, set<InputMode>> validOptionInputModeCombinations = {
+		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
```
cameel: Please add a comment to make it clear that this is not complete yet.
```suggestion
	map<string, set<InputMode>> validOptionInputModeCombinations = {
		// TODO: This should eventually contain all options.
		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
```

solc/CommandLineParser.cpp
```diff
@@ -840,13 +840,20 @@ bool CommandLineParser::processArgs()
...
+		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+	};
+	string invalidOptions;
```
cameel: We have a helper called `joinOptionNames()`. You can store options in a vector and then use that function to join them when you print a message.
```suggestion
	vector<string> invalidOptionsForCurrentInputMode;
```

solc/CommandLineParser.cpp
```diff
@@ -840,13 +840,20 @@ bool CommandLineParser::processArgs()
...
+	for (auto const& [optionName, inputModes]: validOptionInputModeCombinations)
+	{
+		if (m_args.count(optionName) > 0 && inputModes.count(m_options.input.mode) == 0)
+			invalidOptions.append("--" + optionName + " ");
+	}
+
+	if (!invalidOptions.empty())
 	{
-		serr() << "The option --" << g_strExperimentalViaIR << " is only supported in the compiler mode." << endl;
+		serr() << "The option " << invalidOptions << "is not supported in the current mode." << endl;
```
cameel: This message only works in case there's one invalid option.
```suggestion
		serr() << "The following options are not supported in the current input mode: " << invalidOptions << endl;
```

test/solc/CommandLineParser.cpp
```diff
@@ -436,20 +434,29 @@ BOOST_AUTO_TEST_CASE(experimental_via_ir_invalid_input_modes)
 		"--standard-json",
 		"--link",
 	};
-	for (string const& inputModeOption: inputModeOptions)
+
+	static array<string, 2> const invalidOptions = {
+		"--error-recovery",
+		"--experimental-via-ir",
+	};
```
cameel: It will work like this in the current PR but only because both options are supported in the same modes. To handle options from different modes you'll later need to store option-mode combinations so you could already do it here.

cameel (commented):
A few more changes needed but it already looks pretty good.

Please also add a changelog entry.

solc/CommandLineParser.cpp
cameel: Ah, true! I got confused by the fact that your message looks as if you were going to print it separately for each option. Here's a suggestion of a better message: https://github.com/ethereum/solidity/pull/11909#discussion_r723584829.

Midhun07:
@cameel I have added the latest suggestions please review.

test/solc/CommandLineParser.cpp
Midhun07: I have implemented a map similar to the one in `CommandLineParser.cpp`.

Midhun07:
Some issues with b_osx which I don't seem to understand. I'm not sure if it is originating from my changes.

cameel:
> Some issues with b_osx which I don't seem to understand. I'm not sure if it is originating from my changes.

It's unrelated to your changes. We're getting this in external PRs lately and I'm not sure why. Tried to fix it in #12106 but that did not help. In one PR creating a branch on our side helped so I'll try to do it here too.

test/solc/CommandLineParser.cpp
```diff
@@ -427,29 +425,47 @@ BOOST_AUTO_TEST_CASE(standard_json_mode_options)
...
 {
-	static array<string, 5> const inputModeOptions = {
-		"--assemble",
-		"--yul",
-		"--strict-assembly",
-		"--standard-json",
-		"--link",
+	map<string, vector<string>> invalidOptionInputModeCombinations = {
+		// TODO: This should eventually contain all options.
+		{"--error-recovery",
```
```suggestion
		{
			"--error-recovery",
```

test/solc/CommandLineParser.cpp
```diff
@@ -427,29 +425,47 @@ BOOST_AUTO_TEST_CASE(standard_json_mode_options)
...
+			{
+				"--assemble",
+				"--yul",
+				"--strict-assembly",
+				"--standard-json",
+				"--link",
+			}
+		},
```
cameel: That's actually short enough that I'd put it on a single line. The number of mode options is pretty limited.

cameel (commented):
It's good now, just formatting is a bit off.

Also, please add a changelog entry (in the bugfix section).

test/solc/CommandLineParser.cpp
cameel: I just pushed a change that fixes this.

test/solc/CommandLineParser.cpp
cameel: I just pushed a change that fixes this.

Midhun07:
@cameel I have added the changelog for the current update

Midhun07:
@cameel Shall I start pushing further fixes to this branch?

Changelog.md
```diff
@@ -12,6 +12,7 @@ Compiler Features:
 
 Bugfixes:
  * Commandline Interface: Fix extra newline character being appended to sources passed through standard input, affecting their hashes.
+ * Commandline Interface: Disallowed following options in those input modes that do not support them: --error-recovery, --experimental-via-ir
```
cameel: `--experimental-via-ir` was already disallowed in 0.8.8.

```suggestion
 * Commandline Interface: Disallow ``--error-recovery`` option outside of the compiler mode.
```

cameel:
It's fine. The changelog entry needs some tweaking but other than that the PR is ready.

If you want the tests to pass, you can rebase it on #12106 but it's probably going to be merged pretty quickly anyway.

cameel:
I mean, if you want to build on this PR, my suggestion would be to open a new one on top of it.

Changelog.md
cameel: I just pushed a commit that fixes this.

cameel (dismissed):
I fixed the changelog entry and rebased it on latest `develop` so that it can pass tests.

Please submit any further changes in a new PR.

cameel (approved):

PR #11910, opened 2021-09-07
Sort yul functions by creation time.

chriseth:
Still todo:
- [x] check for nondeterminism in IR code generator (all files)
- [x] check optimizer stages to run in call-order.


chriseth:
This results in higher gas costs :)

ekpyron:
> This results in higher gas costs :)

Crazy :-).

ekpyron (dismissed):

ekpyron:
It affects ABIv2 code in old codegen as well, doesn't it? Might be worth a changelog entry then...

chriseth:
Yes it does. I would also like to wait for the bytecode comparison :)

Also there is still a todo box to be ticked.

chriseth:
@ekpyron what would you suggest for the changelog?

ekpyron:
Something like ``Bugfixes: Code Generation: Use stable source order for ABI coder v2 helper functions.`` or the like?
Maybe ``, fixing potential non-determinism.``?

ekpyron (dismissed):

ekpyron:
Changelog conflict and unrelatedly breaking CI...

ekpyron (approved):

chriseth:
Waiting for https://github.com/ethereum/solidity/pull/11911 before merging this.

ekpyron:
> Waiting for #11911 before merging this.

Not sure why we have to wait, but #11911 should also be fine to merge.

PR #11911, opened 2021-09-08
Change RematCandidateSelector to not depend on variable name sorting.

ekpyron:
This only happened in rare cases, i.e. stack too deep and multiple candidates for elimination with the same cost.
But still. But I think it can live without changelog entry...

### DIFF
--- a/libyul/optimiser/StackCompressor.cpp
+++ b/libyul/optimiser/StackCompressor.cpp
@@ -31,6 +31,8 @@
 
 #include <libyul/AST.h>
 
+#include <libsolutil/CommonData.h>
+
 using namespace std;
 using namespace solidity;
 using namespace solidity::yul;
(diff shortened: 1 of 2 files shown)


ekpyron:
Ah, seems like I missed the one affected test earlier - fixed.

test/libyul/yulOptimizerTests/fullSuite/aztec.yul
```diff
@@ -284,15 +284,17 @@
 //             let _5 := 0x40
 //             calldatacopy(0xe0, add(_3, 164), _5)
 //             calldatacopy(0x20, add(_3, 100), _5)
-//             mstore(0x120, sub(_2, c))
-//             mstore(0x60, k)
+//             let _6 := 0x120
+//             mstore(_6, sub(_2, c))
+//             let _7 := 0x60
+//             mstore(_7, k)
```
ekpyron, line 290: That's weird - actually looks like *fewer* variables are eliminated.

ekpyron: But ok - it might just be that choosing the "right" one first made it unnecessary to eliminate further ones in a second iteration... maybe wise to verify that that's what's happening let me check.

ekpyron: Yes, as a matter of fact, this seems to be what's happening.
``develop`` used two iterations on the case, first eliminating 11 variables:
``_1982 _1986 _1989 _1993 _1949 _1943 _1955 _1932 _1930 _2020 _2048``
and then two more in a second iteration:
``_1990 _1925``

while after this PR, one iteration suffices, which eliminates these 11 instead:
``_1982 _1986 _1993 _1949 _1943 _1955 _1925 _1932 _1930 _2020 _2048``
i.e. it doesn't choose _1989, but chooses _1925 in the first pass, which seems to solve the problem right away.

And it's probably not unlikely for this change to be beneficial like that - variables declared earlier have more space to be CSEd into, so may be more likely to be used more often and put more strain on the stack.

Or in summary: seems fine ;-).

libyul/optimiser/StackCompressor.cpp
```diff
@@ -105,12 +114,40 @@ class RematCandidateSelector: public DataFlowAnalyzer
...
+				return varsToEliminate;
+			// If a variable we would like to eliminate references another one
+			// we already selected for elimination, then stop selecting
+			// candidates. If we would add that variable, then the cost calculation
+			// for the previous variable would be off. Furthermore, we
+			// do not skip the variable because it would be better to properly re-compute
+			// the costs of all other variables instead.
+			for (YulString const& referencedVar: references)
+				if (varsToEliminate.count(referencedVar))
+					return varsToEliminate;
```
hrkrshnn, line 145: Why are we returning early instead of continuing in the main loop like in the old implementation?

ekpyron: The old implementation didn't continue for this case either, did it?
It had
```
		if (referencesVarToEliminate)
			break;
```
in the outer loop.

hrkrshnn: Ah, right.

hrkrshnn (approved):

libyul/optimiser/StackCompressor.cpp
ekpyron: The old implementation just didn't ``return``, because it wants to hit the part after the loop - which is why I extracted this to a function, s.t. we don't need these weird flags anymore and *can* just return from nested loops.

test/libyul/yulOptimizerTests/fullSuite/aztec.yul
ekpyron: By the way I don't think the old version aborted here due to a variable being referenced, but the stack error just wasn't fixed by the choice it made (there's no guarantee that it works in the old code transform - it's more like guessing what might help and then running again to see if there's still issues :-) - the new code transform will be able to tell more accurately what helps and what doesn't)

PR #11913, opened 2021-09-08
More generic source location parsing + more errors

cameel:
~Depends on #11915.~ Merged.

This is the refactor requested in https://github.com/ethereum/solidity/pull/11867#discussion_r701143568 and a bunch of other small tweaks that apply generally to `@src` tags even without code snippets.

- `Parser` tests were not checking for errors reported during parsing.
- A few more tests for corner cases related to whitespace.
- More general parsing. Tags are found first and then for each tag parameters are parsed separately.
    - This allows us to report an error in cases where the invalid tags were just being ignored until now. Note that this part is not strictly a refactor since it changes error behavior.

### DIFF
--- a/libyul/AsmParser.cpp
+++ b/libyul/AsmParser.cpp
@@ -131,7 +131,8 @@ void Parser::fetchSourceLocationFromComment()
 		return;
 
 	static regex const lineRE = std::regex(
-		R"~~~((^|\s*)@src\s+(-1|\d+):(-1|\d+):(-1|\d+)(\s+|$))~~~",
+		R"~~(\s*@src\s+)~~"                           // tag: @src
+		R"~~((-1|\d+):(-1|\d+):(-1|\d+)(?:\s+|$))~~", // index and location, e.g.: 1:234:-1
 		std::regex_constants::ECMAScript | std::regex_constants::optimize
 	);
 
@@ -141,11 +142,11 @@ void Parser::fetchSourceLocationFromComment()
 
 	for (auto const& matchResult: ranges::make_subrange(from, to))
 	{
-		solAssert(matchResult.size() == 6, "");
+		solAssert(matchResult.size() == 4, "");
 
-		auto const sourceIndex = toInt(matchResult[2].str());
-		auto const start = toInt(matchResult[3].str());
-		auto const end = toInt(matchResult[4].str());
+		auto const sourceIndex = toInt(matchResult[1].str());
+		auto const start = toInt(matchResult[2].str());
+		auto const end = toInt(matchResult[3].str());
 
 		auto const commentLocation = m_scanner->currentCommentLocation();
 		m_debugDataOverride = DebugData::create();
--- a/libyul/AsmParser.cpp
+++ b/libyul/AsmParser.cpp
@@ -131,46 +131,72 @@ void Parser::fetchSourceLocationFromComment()
 		return;
 
 	static regex const tagRegex = regex(
-		R"~~(\s*@src\s+)~~"                           // tag: @src
-		R"~~((-1|\d+):(-1|\d+):(-1|\d+)(?:\s+|$))~~", // index and location, e.g.: 1:234:-1
+		R"~~(\s*(@[a-zA-Z0-9\-_]+)(?:\s+|$))~~",       // tag, e.g: @src
+		regex_constants::ECMAScript | regex_constants::optimize
+	);
+	static regex const srcTagArgsRegex = regex(
+		R"~~(^(-1|\d+):(-1|\d+):(-1|\d+)(?:\s+|$))~~", // index and location, e.g.: 1:234:-1
 		regex_constants::ECMAScript | regex_constants::optimize
 	);
 
 	string const commentLiteral = m_scanner->currentCommentLiteral();
 	SourceLocation const commentLocation = m_scanner->currentCommentLocation();
-	auto from = sregex_iterator(commentLiteral.begin(), commentLiteral.end(), tagRegex);
-	auto to = sregex_iterator();
+	smatch tagMatch;
+	string::const_iterator position = commentLiteral.begin();
 
-	for (auto const& tagMatch: ranges::make_subrange(from, to))
+	while (regex_search(position, commentLiteral.end(), tagMatch, tagRegex))
 	{
-		solAssert(tagMatch.size() == 4, "");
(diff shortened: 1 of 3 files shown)


libyul/AsmParser.cpp
```diff
@@ -130,37 +130,73 @@ void Parser::fetchSourceLocationFromComment()
...
+		solAssert(tagMatch.size() == 2, "");
+		position += tagMatch.position() + tagMatch.length();
+
+		if (tagMatch[1] == "@src")
 		{
-			shared_ptr<string const> sourceName = m_sourceNames->at(static_cast<unsigned>(*sourceIndex));
-			solAssert(sourceName, "");
-			m_debugDataOverride = DebugData::create(SourceLocation{*start, *end, move(sourceName)});
+			smatch srcTagArgsMatch;
+			if (!regex_search(position, commentLiteral.end(), srcTagArgsMatch, srcTagArgsRegex))
```
cameel, line 155: The only thing stopping us from using `Scanner` for parsing tag args is that there's no way to check where the scanner has stopped and use that to update `position`.

test/libyul/Parser.cpp
```diff
@@ -511,6 +517,9 @@ BOOST_AUTO_TEST_CASE(customSourceLocations_invalid_suffix)
 	EVMDialectTyped const& dialect = EVMDialectTyped::instance(EVMVersion{});
 	shared_ptr<Block> result = parse(sourceText, dialect, reporter);
 	BOOST_REQUIRE(!!result);
+	BOOST_REQUIRE(errorList.size() == 1);
+	BOOST_TEST(errorList[0]->type() == Error::Type::SyntaxError);
+	BOOST_TEST(errorList[0]->errorId() == 8387_error);
```
cameel, line 522: `/// @src 0:420:680foo` used to result in the tag being ignored. Now it's a syntax error.

test/libyul/Parser.cpp
```diff
@@ -44,6 +44,9 @@ using namespace solidity;
 using namespace solidity::util;
 using namespace solidity::langutil;
 
+BOOST_TEST_DONT_PRINT_LOG_VALUE(ErrorId)
+BOOST_TEST_DONT_PRINT_LOG_VALUE(Error::Type)
```
cameel, line 48: We don't have `<<` for these defined. I can either define it or tell boost not to try to print it. For now I did the latter.

libyul/AsmParser.cpp
```diff
@@ -130,37 +130,73 @@ void Parser::fetchSourceLocationFromComment()
...
+			{
+				m_errorReporter.syntaxError(
+					8387_error,
+					commentLocation,
```
cameel, line 159: I think it would be easy to make it point just at the tag and its arguments rather than the whole comment. This was not done originally so maybe for some reason we prefer it to point at the whole comment here?

libyul/AsmParser.cpp
```diff
@@ -130,37 +130,73 @@ void Parser::fetchSourceLocationFromComment()
 	if (m_scanner->currentCommentLiteral().empty())
 		return;
 
-	static regex const lineRE = std::regex(
-		R"~~~((^|\s*)@src\s+(-1|\d+):(-1|\d+):(-1|\d+)(\s+|$))~~~",
-		std::regex_constants::ECMAScript | std::regex_constants::optimize
+	static regex const tagRegex = regex(
+		R"~~(\s*(@[a-zA-Z0-9\-_]+)\s+)~~",             // tag, e.g: @src
```
chriseth: What prevents this from matching `a@src`? There should be a space in front of the `@`.

libyul/AsmParser.cpp
chriseth: No, I don't think we prefer that.

libyul/AsmParser.cpp
cameel: Nothing. It was accepted as a valid tag before my changes and the PR does not change that. I added a test case to show this.

Should I change it to require whitespace before?

cameel:
Rebased on top of #11915.

libyul/AsmParser.cpp
cameel: I just tried to make locations more precise and that's actually harder than I thought due to how `Scanner` works. `Scanner::currentCommentLocation()` returns the location of the whole comment, including the leading `//`, `/*`, `///`, etc.  `Scanner::currentCommentLiteral()` only returns the text that's inside the comment though so I can't determine how my position in that text relates to the position of the start of the comment and my location are off a bit.

I guess I'll just leave this as is.

libyul/AsmParser.cpp
```diff
@@ -130,37 +130,73 @@ void Parser::fetchSourceLocationFromComment()
 	if (m_scanner->currentCommentLiteral().empty())
 		return;
 
-	static regex const lineRE = std::regex(
-		R"~~~((^|\s*)@src\s+(-1|\d+):(-1|\d+):(-1|\d+)(\s+|$))~~~",
-		std::regex_constants::ECMAScript | std::regex_constants::optimize
+	static regex const tagRegex = regex(
+		R"~~(\s*(@[a-zA-Z0-9\-_]+)(?:\s+|$))~~",       // tag, e.g: @src
+		regex_constants::ECMAScript | regex_constants::optimize
+	);
```
cameel: I just added one more change to the regex here. `\s+` after the tag was replaced with `(?:\s+|$)`. This way in the example below `@tag2` will not be ignored:

```yul
/// @src 1:3:4 some text, more text @tag1 @tag2
```

It does not matter all that match currently because there are no tags without args but it might become an unpleasant surprise if someone wanted to add them at some point.

libyul/AsmParser.cpp
cameel: Changed to require whitespace.

chriseth (approved):

libyul/AsmParser.cpp
chriseth: Actually I think I merged this too early: Shouldn't this be `regex_match` instead? Otherwise it would match `@src something 1:1:1`

chriseth: But maybe the `^` already does it...

cameel: Yeah, both would work (`regex_match()` with `.*` added at the end). It starts with `^` exactly for this purpose.

PR #11920, opened 2021-09-09
User defined value type: changes in documentation and test case

hrkrshnn:
Changed the name of a function from `truncate` to `floor`, since that is more appropriate; updated
the inaccurate description on the rounding behaviour. Also modified the respective semantic test.

In response to a comment by @fulldecent  https://github.com/ethereum/solidity/pull/11806#pullrequestreview-750278655

### DIFF
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -676,8 +676,9 @@ type with 18 decimals and a minimal library to do arithmetic operations on the t
         function mul(UFixed256x18 a, uint256 b) internal pure returns (UFixed256x18) {
             return UFixed256x18.wrap(UFixed256x18.unwrap(a) * b);
         }
-        /// Truncates UFixed256x18 to the nearest uint256 number.
-        function truncate(UFixed256x18 a) internal pure returns (uint256) {
+        /// Take the floor of a UFixed256x18 number.
+        /// @return the largest integer that does not exceed `a`.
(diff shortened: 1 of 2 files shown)


docs/types/value-types.rst
```diff
@@ -676,8 +676,9 @@ type with 18 decimals and a minimal library to do arithmetic operations on the t
         function mul(UFixed256x18 a, uint256 b) internal pure returns (UFixed256x18) {
             return UFixed256x18.wrap(UFixed256x18.unwrap(a) * b);
         }
-        /// Truncates UFixed256x18 to the nearest uint256 number.
-        function truncate(UFixed256x18 a) internal pure returns (uint256) {
+        /// Take the floor of a UFixed256x18 number.
+        /// @return the largest integer that does not exceed `a`.
```
hrkrshnn, line 680: Description  from https://en.wikipedia.org/wiki/Rounding#Rounding_down

ekpyron (approved):

chriseth:
But the two operations are actually identical on unsigned types, aren't they?

hrkrshnn:
@chriseth The main issue was the old description "truncates to the nearest uint256 number". That was an incorrect description (well if you look at the "the nearest uint256 number" part independently)

I changed `truncate` to `floor` since that's more appropriate for decimal fixed points, although both should work.

chriseth:
Ah ok, `truncate` and `nearest` do not really fit together!

PR #11923, opened 2021-09-09
Clarify `block.*` and `tx.*` values in off-chain calls

miohtama:
Looks like this may vary between different EVM implementations and confuses users.

### DIFF
--- a/docs/cheatsheet.rst
+++ b/docs/cheatsheet.rst
@@ -131,6 +131,12 @@ Global Variables
 - ``type(T).max`` (``T``): the maximum value representable by the integer type ``T``, see :ref:`Type Information<meta-type>`.
 
 .. note::
+    When contracts are evaluated off-chain rather than in context of a transaction included in a
+    block, you should not assume that ``block.*`` and ``tx.*`` refer to values from any specific
+    block or transaction. These values are provided by the EVM implementation that executes the
+    contract and can be arbitrary.
+
+.. note::
     Do not rely on ``block.timestamp`` or ``blockhash`` as a source of randomness,
     unless you know what you are doing.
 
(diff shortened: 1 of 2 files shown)


docs/cheatsheet.rst
```diff
@@ -90,7 +90,7 @@ Global Variables
 - ``block.difficulty`` (``uint``): current block difficulty
 - ``block.gaslimit`` (``uint``): current block gaslimit
 - ``block.number`` (``uint``): current block number
-- ``block.timestamp`` (``uint``): current block timestamp
+- ``block.timestamp`` (``uint``): current block timestamp of the block where the transaction was mined; for no-transaction calls this value can be anything from zero to a random number depending on the EVM implementation
```
chriseth: If we do that much in detail here (which I think we shouldn't) then we should also do the same change to all the other `block` members.

axic: Wouldn't a "warning"/"notice" block be enough for explaining that these parameters can have arbitrary value in case of off-chain evaluation?

miohtama: Happy to amend the pull request and make it a generic warning for all `block` members if you believe this is a better approach.

hrkrshnn: @miohtama You could add it as a new note before the note "Do not rely on ``block.timestamp`` or ``blockhash`` as a source of randomness, unless you know what you are doing."

chriseth:
@miohtama are you still working on this?

miohtama:
Hi, sorry, this was dropped too deep on my todo list. I will revisit.

cameel:
I have updated the description. Does that help, @miohtama?

chriseth (approved):

PR #11924, opened 2021-09-10
Expand description of literals.

acoglio:
How ASCII strings evaluate to integers was not immediately clear to me from reading the current documentation, before the clarifying discussions on Gitter. This PR contains (my understanding of) those clarifications. Please feel free to ask me to make changes/corrections, or to ignore this PR completely if you'd rather expand the documentation of literals in a different way.

### DIFF
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -174,9 +174,37 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
 Literals
 --------
 
-As literals, you can use integer constants in decimal or hexadecimal notation
-or strings as ASCII (`"abc"`) or HEX strings (`hex"616263"`) of up to
-32 bytes length.
+As literals, you can use:
+
+- Integer constants in decimal or hexadecimal notation.
+
+- ASCII strings (e.g. ``"abc"``), which may contain hex escapes ``\xNN`` and Unicode escapes ``\uNNNN`` where ``N`` are hexadecimal digits.
+  A string of this form represents an integer as follows:
+
+  1. The string is viewed as a byte sequence, by viewing
+     a non-escape ASCII character as a single byte whose value is the ASCII code,
+     an escape ``\xNN`` as single byte with that value, and
+     an escape ``\uNNNN`` as the UTF-8 sequence of bytes for that code point.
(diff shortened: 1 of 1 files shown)


docs/yul.rst
```diff
@@ -174,9 +174,37 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
...
+     a non-escape ASCII character as a single byte whose value is the ASCII code,
+     an escape ``\xNN`` as single byte with that value, and
+     an escape ``\uNNNN`` as the UTF-8 sequence of bytes for that code point.
+
+  2. The byte sequence must not exceed 32 bytes.
+
+  3. The byte sequence is padded with zeros on the right to reach the number of bytes appropriate for the type of the literal.
+     In order words, the string is stored left-aligned.
+
+  4. The byte sequence represents an unsigned integer in base 256 in bit endian form.
```
ekpyron, line 194:
```suggestion
  4. The byte sequence represents an unsigned integer in base 256 in big endian form.
```
?
Also this is just another way of formulating 3, isn't it?
Same thing for the hex string part.

docs/yul.rst
```diff
@@ -174,9 +174,37 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
 Literals
 --------
 
-As literals, you can use integer constants in decimal or hexadecimal notation
```
chriseth, line 177: I think this was kept generic, ignoring the dialect, but in practice, we enforced aspects of the EVM dialect more and more over time, so that there is currently no yul dialect that allows words larger than 32 bytes.

The actual semantics highly depend on the dialect and this mostly includes the interpretation of literals and bulit-in functions. The built-in functions are not explained in detail, so maybe we should also not go that much in detail here?

On the other hand, we could also add some more details in the built-in functions part like explain how the builtins that actually interpret the data (add, sub, ...) instead of just moving data around.

Further down it says "When compiling for the EVM..." - so maybe we should combine that part with the part here that explain how the syntax of literals are interpreted and just explain the syntax in a dialect-free first part?

acoglio:
This is replaced by #11977.

PR #11926, opened 2021-09-10
[docs] Add sections for state mutability and special functions

nikitastupin:
Hi,

In this PR I'm organizing state mutability and special functions parts of the documentation to corresponding sections.

Also I think that we should split [Visibility and Getters](https://docs.soliditylang.org/en/latest/contracts.html#visibility-and-getters) section to `Functions -> Visibility` and `State Variables -> Visibility` sections, again to add a bit more structure. I'll do this another PR later if you're OK with these changes.


ekpyron (approved):
I definitely like the ``state mutability`` section.
Whether we need/want the "special functions" section, I'm not entirely sure, but I'd be fine with it.

But it'd be nice to have a second opinion on this.

chriseth:
Looks good!

Issue #11900, opened 2021-09-06
DOCS: Adjust rdt-theme colors by adding a "Solidity style" custom CSS file


franzihei:
# Problem Statement
The current docs theme we are using is the plain readthedocs sphinx theme. Some colors are hard to read (e.g. the search highlights). Additionally, it would be nice if the docs would reflect the overall Solidity brand identity.

# To Do

- [x] Add custom CSS file called "Solidity light mode" with the color changes described blow ([docs](https://docs.readthedocs.io/en/latest/guides/adding-custom-css.html))
- [x] In the dark mode CSS file, adjust the search highlight color to be soft red: #F0978B (yellow is currently not readable)
- [x] [optional] Add Solidity logo in the menu bar and delete logo from the index page
- [ ] [optional] Adjust font to Work Sans to be consistent with the font across blog, portal and docs

**Proposed color changes:**
- Make code block background Solidity light grey: #E5E5E5
- Make titles / highlights Solidity light blue: #65AFFF
- Make hyperlinks Solidity electric blue: #002fa7
- Make highlights for searched words warmer yellow: #EAC545

**Rough sketch of how it would look with the proposed colors:**

![image](https://user-images.githubusercontent.com/41991517/132236639-27d544d6-9602-433e-a0bd-3e78b527fe94.png)

4molybdenum2:
Hi, I made a PR for this issue. Can you please review? @franzihei

cameel:
> Make code block background Solidity light grey: #E5E5E5

I think I liked the green code blocks a bit more than the gray ones. They were not perfect so changing them is probably a good idea (for example green and light blue text does not stand out enough on that background) but the new ones look a bit dull:

### Before
![yul-stand-alone-usage-green](https://user-images.githubusercontent.com/137030/133292176-d1cc5f83-8354-4814-8fe3-c3459be9aff7.png)

### After
![yul-stand-alone-usage-gray](https://user-images.githubusercontent.com/137030/133292343-e651ae4f-5b6a-477e-b79e-49032531af95.png)

axic:
@franzihei @cameel Could something like a thin border around the box with a different colour work?

cameel:
Maybe? Highlighted code looks good on white background but I think that the block itself does not stand out enough.

![yul-stand-alone-usage-white](https://user-images.githubusercontent.com/137030/133293948-c0246a6d-76b2-4a21-9837-6f48c0275dde.png)

axic:
[...]

PR #11927, opened 2021-09-10
Adjust Colors for the Solidity Docs

4molybdenum2:
Made the following changes in this PR  (#11900): 

- [x] Add custom CSS file called "light.css" with the color changes as follows:
- Make code block background Solidity light grey: #E5E5E5
- Make titles / highlights Solidity light blue: #65AFFF
- Make hyperlinks Solidity electric blue: #002fa7
- Make highlights for searched words warmer yellow: #EAC545

- [x] In the dark mode CSS file, adjust the search highlight color to be soft red: #F0978B (yellow is currently not readable)
- [x] [optional] Add Solidity logo in the menu bar and delete logo from the index page
- [ ] [optional] Adjust font to Work Sans to be consistent with the font across blog, portal and docs

### DIFF
--- a/docs/_static/css/dark.css
+++ b/docs/_static/css/dark.css
@@ -1,10 +1,8 @@
 /* links */
 
-a,
-a:visited {
-    color: #aaddff;
+.rst-content a:not(:visited) {
+    color: #aaddff !important;
 }
-
 
 /* code directives */
 
@@ -46,6 +44,9 @@ em.property {
     background-color: #5a5a5a;
 }
 
+.rst-content pre {
+    background: none;
+}
 
 /* inlined code highlights */
 
@@ -70,6 +71,12 @@ code.docutils.literal.notranslate {
     color: #ddd;
 }
 
+/* highlight color search text */
+
+.rst-content .highlighted {
+    background: #ff5722;
+    box-shadow: 0 0 0 2px #f0978b;
+}
 
 /* notes, warnings, hints */
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -238,7 +238,7 @@ latex_elements = {
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-        ('index', 'solidity.tex', 'Solidity Documentation', 'Ethereum', 'manual'),
+    ('index', 'solidity.tex', 'Solidity Documentation', 'Ethereum', 'manual'),
 ]
 
 # The name of an image file (relative to this directory) to place at the top of
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -128,7 +128,11 @@ html_theme = 'sphinx_rtd_theme'
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
+html_theme_options = {
+    'logo_only': True,
+    'style_nav_header_background': '#65afff',
+    'display_version': True,
+}
 
 # Add any paths that contain custom themes here, relative to this directory.
 #html_theme_path = []
@@ -142,7 +146,7 @@ html_theme = 'sphinx_rtd_theme'
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-#html_logo = None
+html_logo = "logo.svg"
 
 # The name of an image file (within the static path) to use as favicon of the
 # docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -1,14 +1,14 @@
 pre {
-  white-space: pre-wrap;       /* css-3 */
-  white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
-  white-space: -pre-wrap;      /* Opera 4-6 */
-  white-space: -o-pre-wrap;    /* Opera 7 */
-  word-wrap: break-word;
+    white-space: pre-wrap;       /* css-3 */
+    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+    white-space: -pre-wrap;      /* Opera 4-6 */
+    white-space: -o-pre-wrap;    /* Opera 7 */
+    word-wrap: break-word;
 }
 
 .wy-table-responsive table td, .wy-table-responsive table th {
-  white-space: normal;
+    white-space: normal;
 }
 .rst-content table.docutils td {
-  vertical-align: top;
+    vertical-align: top;
 }
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -12,3 +12,37 @@ pre {
 .rst-content table.docutils td {
     vertical-align: top;
 }
+
+/* links */
+.rst-content a:not(:visited) {
+    color: #002fa7;
+}
+
+.rst-content .highlighted {
+    background: #eac545;
+}
+
+/* code block highlights */
+.rst-content pre {
+    background: #fafafa;
+}
+
+.wy-side-nav-search img.logo {
+    width: 100px !important;
+    padding: 0;
+}
+
+.wy-side-nav-search > a {
+    padding: 0;
+    margin: 0;
+}
+
+/* project version (displayed under project logo) */
+.wy-side-nav-search .version {
+    color: #272525 !important;
+}
+
+/* menu section headers */
+.wy-menu .caption {
+    color: #65afff !important;
+}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,6 +1,6 @@
 {% extends "!layout.html" %}
 
-  {% block menu %}
+{% block menu %}
     {{ super() }}
-	<a href="{{ pathto('genindex') }}">Keyword Index</a>
-  {% endblock %}
+    <a href="{{ pathto('genindex') }}">Keyword Index</a>
+{% endblock %}
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,11 +1,6 @@
 Solidity
 ========
 
-.. image:: logo.svg
-    :width: 120px
-    :alt: Solidity logo
-    :align: center
-
 Solidity is an object-oriented, high-level language for implementing smart
 contracts. Smart contracts are programs which govern the behaviour of accounts
 within the Ethereum state.


ekpyron:
The CI failure is independent of the PR and can be ignored (that's just CircleCI acting up again).

As for the changed colors: one thing I'm noticing is that the new darker blue for links makes the links quite hard to read if switched to dark (using the toggle in the bottom left).

4molybdenum2:
So should I change it back to the original color currently or maybe a different color for dark mode? 🤔

ekpyron:
Switching back to the old link color only in dark mode would seem fine with me, but I'm not a designer, so I'd defer that question to you or @franzihei :-).

4molybdenum2:
Sure👍

franzihei:
Good point I did not think about the hyperlink being too dark in dark mode. I think keeping the original hyperlink color in the dark mode would work great!

docs/_static/css/light.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.rst-content .highlighted {
+	background: #eac545;
+}
+
+/* inline code highlights */
+.xref,
+.py-meth,
+.rst-content code {
```
franzihei: Here I meant to change the color of the code blocks, as in the background color in the example contracts (currently a greenish yellow), not the inline code highlights. Sorry if that wasn't clear!

franzihei (commented):
Thanks so much for picking up this issue! 

A couple of things:
- In the dark version now all text in the menu sidebar seems to be light blue. This shouldn't be the case, it should remain in the style it was before
Now:
image
Before:
image

- The background of the code blocks (the big boxes with code) should be light grey, not the background of inline code.
- In the header below the logo it now only says `0.8.8` as opposed to `v0.8.8`. The `v`seems to be missing.
- Is it possible to make the logo a bit smaller/ have less free space around it?

docs/_static/css/light.css
4molybdenum2: Oh ok, I will fix these things and make a commit then..

4molybdenum2: Was unable to resolve the **version issue where 'v' was missing**. Seems to be existing. I think we need to pass the version as v0.8.8 in the version variable (which I think is the **CMakeLists.txt**)  file.

franzihei (commented):
Thanks for the changes! One thing that is not optimal are the "visited" hyperlinks in the menu on the left. The menu items should not change their color to light blue when they were clicked/visited, they should keep the same color (that applies for both light mode and dark mode).

Also I think that simply making the logo smaller does not work, it looks like there is too much "empty space" surrounding it. That might be an issue with the original image used. 

image

4molybdenum2:
Hmm, the thing that I couldn't figure out was how to remove the "empty space" around the symbol because the space is in the **png** and we can't remove that with CSS itself.

4molybdenum2:
Hmm...I couldn't figure out the "empty space" scenario.

docs/_static/css/light.css
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/* code blocks highlights */
+.rst-content pre {
+	background: #e5e5e5;
+}
+
+.wy-side-nav-search img{
+	width: 80px !important;
+}
```
cameel: I think this should be more specific. We only want this height specifically for the logo:
```suggestion
.wy-side-nav-search img.logo {
	width: 80px !important;
}
```

cameel: Also, why is this in `light.css`? Don't we need this rule in dark mode as well?

docs/conf.py
```diff
@@ -142,7 +145,7 @@ def setup(sphinx):
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-#html_logo = None
+html_logo = "./logo.svg"
```
cameel:
```suggestion
html_logo = "logo.svg"
```

docs/conf.py
```diff
@@ -154,7 +157,7 @@ def setup(sphinx):
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
-html_css_files = ["css/toggle.css"]
+html_css_files = ["css/toggle.css", "css/light.css"]
```
cameel: Won't this apply `light.css` in both modes rather than only in the light mode as the name would suggest?

If your additions there were meant for both modes then I think that `custom.css` would be a better place.

docs/conf.py
```diff
@@ -124,11 +124,14 @@ def setup(sphinx):
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 html_theme = 'sphinx_rtd_theme'
-
```
cameel: I'd leave this line in. The file separates options with empty lines for readability.

docs/conf.py
```diff
@@ -124,11 +124,14 @@ def setup(sphinx):
...
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
+html_theme_options = {
+    'logo_only': True,
+    'style_nav_header_background': '#65afff',
+    'display_version': True,
```
cameel, line 134: > Was unable to resolve the **version issue where 'v' was missing**. Seems to be existing. I think we need to pass the version as v0.8.8 in the version variable (which I think is the **CMakeLists.txt**)  file.

I think that version gets overridden by ReadTheDocs. If you build docs locally on develop you don't get the `v` either. And ReadTheDocs can also display something completely different there - for it's `latest` for https://docs.soliditylang.org/en/latest/ and `breaking` for https://docs.soliditylang.org/en/breaking/ even though these do have 0.8.8 in `CMakeLists.txt`.

I think that the `v` will be there when the PR gets merged and docs get deployed.

docs/_static/css/dark.css
```diff
@@ -46,7 +45,9 @@ em.property {
     background-color: #5a5a5a;
 }
 
-
+.rst-content pre{
+    background: none;
+}
```
cameel:
```suggestion
.rst-content pre {
    background: none;
}

```

docs/_static/css/light.css
```diff
@@ -0,0 +1,31 @@
+/* links */
+.rst-content a{
+    color: #002fa7;
+}
+
+.rst-content a:visited {
+    color: #002fa7;
+}
```
cameel: I think the visited color was meant to be identical to normal color only in the menu. Currently links in the text do get highlighted when visited and I'd personally count that as a feature :)

> One thing that is not optimal are the "visited" hyperlinks in the menu on the left. The menu items should not change their color to light blue when they were clicked/visited, they should keep the same color (that applies for both light mode and dark mode).

BTW, I can understand why designers might not like menu items having different colors when visited but personally I find that feature very useful so I'd keep the visited colors even in the menu. The light blue was definitely too flashy but maybe we could have at least a subtle color change? What do you think @franzihei?

docs/_static/css/light.css
```diff
@@ -0,0 +1,31 @@
...
+.wy-side-nav-search img{
+	width: 80px !important;
+}
+
+/* version sidebar */
+.wy-side-nav-search .version {
+	color: #272525 !important;
+}
+
+/* title color */
```
cameel: "title color" sounds as if these were titles in the text
```suggestion
/* menu section headers */
```

docs/_static/css/light.css
```diff
@@ -0,0 +1,31 @@
...
+/* code blocks highlights */
+.rst-content pre {
+	background: #e5e5e5;
+}
+
+.wy-side-nav-search img{
+	width: 80px !important;
+}
+
+/* version sidebar */
```
cameel:
```suggestion
/* project version (displayed under project logo) */
```

docs/_static/css/light.css
```diff
@@ -0,0 +1,31 @@
+/* links */
+.rst-content a{
```
cameel:
```suggestion
.rst-content a {
```

docs/_static/css/light.css
cameel: > Hmm...I couldn't figure out the "empty space" scenario.

- The `<a>` element containing the logo has a bottom margin of `0.809em` that could be removed.
- Then the `<div>` that contains that link has `0.809em` of padding.
- Something adds padding of `5px` around `.wy-side-nav-search img` - if you remove it, logo gets slightly bigger.
- `.wy-side-nav-search > div.version` has a negative top margin. If the above are not enough, you could use that to get version as close to the logo as needed.

I think that the above bring us pretty close to the spacing in the design.

docs/_static/css/dark.css
```diff
@@ -70,6 +71,12 @@ code.docutils.literal.notranslate {
     color: #ddd;
 }
 
+/* highlight color search text */
+
+.rst-content .highlighted {
+    background: #f0978b;
+    box-shadow: 0 0 0 2px #f0978b;
+}
```
cameel: > In the dark mode CSS file, adjust the search highlight color to be soft red: #F0978B (yellow is currently not readable)

This is more readable now but I think it could still be improved. The contrast between the text and the highlighted background seems a bit low so it might still be hard to read in some conditions:

contract-metadata-dark-mode

docs/_static/css/dark.css
```diff
@@ -1,6 +1,5 @@
 /* links */
-
-a,
+.rst-content a,
 a:visited {
     color: #aaddff;
 }
```
cameel: This made links in page header and footer dark blue:

contract-metadata-header-dark-mode
contract-metadata-footer-dark-mode

cameel (dismissed):
Thanks for the PR @4molybdenum2! Franzi is off this week so here's some feedback from me. Looks like there are only tiny details to iron out at this point.

docs/_static/css/dark.css
cameel: One more thing, I'd change the background here to white instead and make the border thicker (https://github.com/ethereum/solidity/issues/11900#issuecomment-919297759, https://github.com/ethereum/solidity/issues/11900#issuecomment-919311493).

4molybdenum2:
Sure, I will get on it and make updates by tomorrow...

4molybdenum2:
So, the changes I made acc. to review:
- refactored code wherever necessary
- updated the code blocks background to white and thickened the border
- made sure tho reduce space around logo as much as possible by removing the paddings and margins
- added a different color for the highlights in dark mode which provides better contrast
- deleted the light.css file and put everything in custom.css

cameel:
@4molybdenum2 Looks like all the visual issues have been resolved :) I think that the final result looks good. Now @franzihei just needs to approve :)

There's just some general cleanup needed before this can be merged:
- I see that `custom.css` has 2-space indents and you followed that style. Which is good overall, but the file is very short so maybe you could just convert it to 4 so that we have it consistent in all CSS files?
- There are some files like `toggle.css`, where the only change seems to be adding a newline at the end. Could you remove these unrelated changes from the PR?
- There are a few places with inconsistent whitespace, e.g. missing space before `{`, empty line inside `{}` or no empty line between CSS definitions. Could you fix those? I pointed some of them in comments above.
- Commits should be squashed. We generally perfer one logical change per commit so review fixes should be squashed in to the commits they fix rather than left on top.

docs/_static/css/dark.css
cameel: @axic What do you think? I think that green ones looked slightly better but white ones are not bad either so I'm fine with them.

docs/_static/css/light.css
cameel: @4molybdenum2 Can you fix this one too?

4molybdenum2:
Sorry, I have been away for a while and could not get to working on the above changes.....will do so whenever I get the time.

cameel:
@4molybdenum2 If you don't have time for it, I can take over the PR and finish it. You already did 99% of the work so it just needs some cleanup :)

4molybdenum2:
> @4molybdenum2 If you don't have time for it, I can take over the PR and finish it. You already did 99% of the work so it just needs some cleanup :)

Sorry I couldn't get back to it. I am currently away for the time being and lost track of it😅 .....I would really appreciate it if you could take over

cameel:
I did the cleanup and also adjusted the link style so that visited link color is overridden only in the menu.

@franzihei Please take a look at this. Also, any feedback on https://github.com/ethereum/solidity/pull/11927#discussion_r708379674?

franzihei:
Cool looks really good overall!

A few points:
- I personally really don't like the white-on-white in the code blocks. I'd prefer it to be the very light gray (#FAFAFA) we also use for code blocks in the blog (shown [here](https://github.com/ethereum/solidity/issues/11900#issuecomment-925089250)). Also, what color is currently being used for the outline? It looks like black, which is a bit hard on the eyes imo. I'd prefer it to be similar to the blog, a slightly darker gray than the background. That being said, if everyone else likes it like this I'm also fine to keep it this way.
- Is it possible to change the "v0.x.x" back to how it was displayed before the changes? Now it is missing the "v" and is displayed in hard black, compared to a gray shade in the previous version. I can live with the black, but I find it strange that the "v" is gone - were any changes made on that front?
image
image
- Wrt to the "visited links" question: I really like how it's implemented right now. You can see what you visited in the documents itself, but not in the menu. Having a difference between visited and not visited in the menu will be highly confusing for the eyes and we also didn't have it previously, so I'd like to keep it that way. I think the menu in the index/home page still gives a great overview of what's been visited and what not (see screenshot). :)
image

cameel:
@franzihei 
- Code block background changed to `#fafafa`.
- Border around code blocks was `0.01em` `solid` `#4a4a4a`. Tried to change it to `1px` `solid` `#cccccc` as on the blog but turns out that there's another border on the element the block is wrapped in (`.highlight-solidity`). The problem with overriding it is that I would have to do it for every language separately. Instead I just removed our border - the default one (`1px` `solid` `#e1e4e5`) looks pretty close to the one from the blog. Just has no rounded corners.
- About the `v`, see https://github.com/ethereum/solidity/pull/11927#discussion_r708373582. In short: the `v` should appear there once docs are deployed to Read The Docs.
- I don't see anything here or in the issue about what the color of the `v0.x.x` part was meant to be but it looks to me like it matches the colors in your design. It's currently `#272525`. Is that too black? I can change it to anything you want.

franzihei (dismissed):
Great, I think everything can stay like that! :)

chriseth:
please squash

cameel:
Squashed.

chriseth (approved):

PR #11930, opened 2021-09-10
Fix inline assembly assignments to calldata structs and statically-sized calldata arrays.

### LINKED ISSUE #11680: ICE when assigning to calldata structs and static arrays using inline assembly + `FixedPointType not implemented` error
## Description

In the contract below there seems to be an Internal Compiler Error without a specific message:

pragma abicoder v2;
contract C {
    struct S {
        uint16[] m;
    }
    function f(S calldata s) public pure returns (bool correct) {
        int8 x = int8(int16(s.m[0]));
        uint r;
        assembly {
            s := x
        }
        correct = r == 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff80;
    }
}

while in the contract :

pragma abicoder               v2;
contract C {
    function f(fixed a, int16 b, address c, bytes3 d, bool e)
            public pure returns (uint v, uint w, uint x, uint y, uint z) {
        assembly { v := a  w := b x := c y := d z := e}
    }
}

there is the internal compiler error FixedPointType not implemented.

## Environment

- Compiler version: 0.8.6
- Target EVM version (as per compiler settings): 0.8.6+commit11564f7e
- Framework/IDE (e.g. Truffle or Remix): Remic
- Operating system: Ubuntu 16.04 LTS

### LINKED ISSUE #11929: [Codegen] ICE in ContractCompiler visit of inline assembly
```
contract C{
  struct S{int m;}
  function f(S calldata s) public returns(int r) {assembly{s:=r} }
}
```

throws

https://github.com/ethereum/solidity/blob/dea1b9ec7995ff7f16cdee7925a5858c5ce33982/libsolidity/codegen/ContractCompiler.cpp#L864-L867

```
$ solc test.sol
```

ekpyron:
Fixes https://github.com/ethereum/solidity/issues/11680. Fixes https://github.com/ethereum/solidity/issues/11929.


ekpyron:
I, by the way, wonder if we should disallow this and require accessing these with ``.offset`` in 0.9...

ekpyron:
Must have been too late on Friday, I'll need another look at this.
EDIT: ok, the fix was ok, but the test was wrong. Also added a fix for the statically-sized calldata array case.

test/libsolidity/semanticTests/inlineAssembly/calldata_struct_assign_and_return.sol
```diff
@@ -0,0 +1,17 @@
+contract C{
+  struct S{ int8 x; int8 y; }
+  function f() internal pure returns(S calldata s) {
+    assembly {
+      s := 0x24
+    }
+  }
```
ekpyron, line 8: This could actually be quite a useful pattern and replace using abi decoding from ``msg.data``...
We do allow this for storage references, so I see no reason for not allowing it for calldata.

ekpyron: Although... hm... I'm not entirely sure, this might bypass validation on the outermost layer, let me see...

ekpyron: Well, we do allow it for dynamic calldata arrays anyways, so...

test/libsolidity/semanticTests/inlineAssembly/calldata_struct_assign_and_return.sol
```diff
@@ -0,0 +1,21 @@
...
+  function h() public pure returns(uint256) { f(); return 0x42; }
+  function i() public pure returns(uint256) { abi.decode(msg.data[4:], (S)); return 0x42; }
+}
+// ====
+// compileViaYul: also
+// ----
+// g(): 0xCAFFEE, 0x42, 0x21 -> 0x42, 0x21
+// g(): 0xCAFFEE, 0x4242, 0x2121 -> FAILURE
+// h() -> 0x42
+// i() -> FAILURE
```
ekpyron, line 23: Yeah... these two are basically the same, but ``h()`` skips validation of the outermost layer in calldata, so it succeeds even with empty calldata, while ``i()`` performs this validation and reverts...

However, the same thing will apply for cases like the ``calldata_assign_form_nowhere`` test (https://github.com/ethereum/solidity/blob/dea1b9ec7995ff7f16cdee7925a5858c5ce33982/test/libsolidity/semanticTests/inlineAssembly/calldata_assign_from_nowhere.sol) and we do not appear to care for that case either...

ekpyron: Or to be more precise: https://github.com/ethereum/solidity/blob/dea1b9ec7995ff7f16cdee7925a5858c5ce33982/test/libsolidity/semanticTests/inlineAssembly/calldata_assign_from_nowhere.sol will also succeed, if it sets 
``x.length := 64``... i.e. it will happily read beyond ``calldatasize()``...

test/libsolidity/semanticTests/inlineAssembly/calldata_struct_assign_and_return.sol
```diff
@@ -0,0 +1,22 @@
...
+  }
+  function h() public pure returns(uint256) { f(); return 0x42; }
+  function i() public pure returns(uint256) { abi.decode(msg.data[4:], (S)); return 0x42; }
+}
+// ====
+// compileViaYul: also
+// ----
+// g(): 0xCAFFEE, 0x42, 0x21 -> 0x42, 0x21
+// g(): 0xCAFFEE, 0x4242, 0x2121 -> FAILURE
+// g(): 0xCAFFEE, 0x42 -> 0x42, 0
```
ekpyron, line 21: Added another test case for the phenomenon.
While the inner layer, i.e. the struct members, are validated on access, the check that calldata is large enough in the first place is not performed.

Changelog.md
```diff
@@ -16,6 +16,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Code Generator: Fix ICE on assigning calldata structs and statically-sized calldata arrays in inline assembly.
```
```suggestion
 * Code Generator: Fix ICE on assigning to calldata structs and statically-sized calldata arrays in inline assembly.
```
(just to clarify)

chriseth:
1) No code change for IR needed?
2) Can you check that we detect a missing assignment to a statically-sized calldata array and struct that is a return variable of a function? (no need for a new test if we already have one)

ekpyron:
> 2. Can you check that we detect a missing assignment to a statically-sized calldata array and struct that is a return variable of a function? (no need for a new test if we already have one)

Apparently, we don't have any such test even for dynamic calldata arrays - I guess we said we test it for storage pointers and the check is identical in the code, so if it's fine for storage, it's fine for calldata (and it indeed behaves identically).
I can basically duplicate ``test/libsolidity/syntaxTests/controlFlow/storageReturn`` as a ``calldataReturn`` version, though.

ekpyron:
> 1. No code change for IR needed?

The tests are passing - apparently in IR codegen this all falls under ``suffix.empty() && varDecl->isLocalVariable()`` in IRGeneratorForStatements.cpp:L115

chriseth (approved):

chriseth (approved):

PR #11931, opened 2021-09-10
Different metadata for IR

### LINKED ISSUE #11876: Always include `--experimental-via-ir` flag in metadata generated from IR
Part of #11494.

Currently the metadata embedded in Yul sources generated with `--ir` or `--ir-optimized` is the same as the one that goes into the binary. This means that if you do not use the `--experimental-via-ir` but use `--ir` or `--ir-optimized` to get the output from the new codegen, printed Yul sources  have metadata that is not marked as generated using the experimental pipeline even though it was.

Change this so that the metadata in printed Yul code always looks as if the `--experimental-via-ir` option was used.

cameel:
Fixes #11876.

### DIFF
--- a/libsolidity/interface/CompilerStack.h
+++ b/libsolidity/interface/CompilerStack.h
@@ -326,11 +326,16 @@ public:
 	/// @returns a JSON representing a map of method identifiers (hashes) to function names.
 	Json::Value methodIdentifiers(std::string const& _contractName) const;
 
-	/// @returns the Contract Metadata
-	std::string const& metadata(std::string const& _contractName) const;
+	/// @returns the Contract Metadata matching the pipeline selected using the viaIR setting.
+	std::string const& metadata(std::string const& _contractName) const { return metadata(contract(_contractName)); }
 
-	/// @returns the cbor-encoded metadata.
-	bytes cborMetadata(std::string const& _contractName) const;
+	/// @returns the CBOR-encoded metadata matching the pipeline selected using the viaIR setting.
+	bytes cborMetadata(std::string const& _contractName) const { return cborMetadata(_contractName, m_viaIR); }
+
+	/// @returns the CBOR-encoded metadata.
+	/// @param _forIR If true, the metadata for the IR codegen is used. Otherwise it's the metadata
+	///               for the EVM codegen
+	bytes cborMetadata(std::string const& _contractName, bool _forIR) const;
 
 	/// @returns a JSON representing the estimated gas usage for contract creation, internal and external functions
 	Json::Value gasEstimates(std::string const& _contractName) const;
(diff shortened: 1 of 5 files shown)


test/libsolidity/Metadata.cpp
```diff
@@ -344,24 +344,52 @@ BOOST_AUTO_TEST_CASE(metadata_viair)
...
-		string metadata_str = compilerStack.metadata("test");
-		Json::Value metadata;
-		util::jsonParseStrict(metadata_str, metadata);
-		BOOST_CHECK(solidity::test::isValidMetadata(metadata_str));
-		BOOST_CHECK(metadata.isMember("settings"));
-		if (_viair)
+
+		checkMetadata(compilerStack.metadata("test"), _viaIR);
+		checkMetadata(compilerStack.evmMetadata("test"), false);
+		checkMetadata(compilerStack.irMetadata("test"), true);
```
cameel: This really only checks the metadata stored in `CompilerStack` was generated correctly, not that the correct metadata was actually included in IR output. Checking that would require using `ObjectParser` to decode the generated IR so I only tested it manually.

libsolidity/interface/CompilerStack.h
```diff
@@ -367,7 +373,8 @@ class CompilerStack: public langutil::CharStreamProvider
 		std::string yulIROptimized; ///< Optimized experimental Yul IR code.
 		std::string ewasm; ///< Experimental Ewasm text representation
 		evmasm::LinkerObject ewasmObject; ///< Experimental Ewasm code
-		util::LazyInit<std::string const> metadata; ///< The metadata json that will be hashed into the chain.
+		util::LazyInit<std::string const> evmMetadata; ///< The metadata json that will be hashed into the chain. For EVM codegen.
+		util::LazyInit<std::string const> irMetadata; ///< The metadata json that will be hashed into the chain. For IR codegen.
```
chriseth: Why do we need two different metadatas here, but only one `evmAssembly`?

cameel: Because we only produce one assembly but we might need to print two different metadata hashes. Each metadata json is appropriate for the bytecode from a different codegen but the assembly can only be generated by one of them.

If you prefer I could store only one metadata (matching `viaIR` is it was until this PR) and generate metadata for IR on the fly and discard it immediately after giving the hash to the IR codegen.

chriseth: I think it is less confusing to only store the metadata json that is printed for `--metadata` - unless it is a big performance drawback. And even if it is, it only will be for `--experimental-via-ir` or `--ir`, right?

cameel: ok. Either is fine with me. I thought that we'd want to store it to be able to print it later without generating it again but if we don't need that then generating it on the fly should be fine.

cameel: Done.

chriseth (approved):

PR #11933, opened 2021-09-10
Add switch for command line & test.sh to skip smt

soroosh-sdi:
- add --no-smt in scripts/tests.sh and test/cmdlineTests.sh

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -37,8 +37,26 @@ source "${REPO_ROOT}/scripts/common.sh"
 # shellcheck source=scripts/common_cmdline.sh
 source "${REPO_ROOT}/scripts/common_cmdline.sh"
 
-AUTOUPDATE=false
-[[ $1 == --update ]] && AUTOUPDATE=true && shift
+autoupdate=false
+no_smt=false
+declare -a selected_tests
+while [[ $# -gt 0 ]]
+do
+    case "$1" in
+        --update)
+            autoupdate=true
+            shift
+            ;;
+        --no-smt)
+            no_smt=true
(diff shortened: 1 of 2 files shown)


test/cmdlineTests.sh
```diff
@@ -302,25 +320,25 @@ test_solc_behaviour "${0}" "ctx:=/some/remapping/target" "" "" 1 "" "Invalid rem
 printTask "Running general commandline tests..."
 (
     cd "$REPO_ROOT"/test/cmdlineTests/
-    for tdir in ${*:-*/}
+    (( ${#selected_tests[@]} > 0 )) || selected_tests=(*)
```
soroosh-sdi, line 323: > What's the purpose of this? If there are no selected tests, select all tests?
Shouldn't it be (( ${#selected_tests[@]} == 0 )) && selected_tests=(*) or (( ${#selected_tests[@]} > 0 )) || selected_tests=(*) for that?

sorry for that, it is fixed now.

soroosh-sdi:
@cameel 
this is the PR we talked about in #11873.

cameel (approved):

leonardoalt (approved):

leonardoalt:
I guess we don't have tests for this option? Should be fine as long as the normal ones still work, and we only use it locally.

cameel:
Well, I think we don't generally test the test scripts :)

And yeah, the option is meant mostly for local use. We do not really have a use case for it in CI so far, which is why no one added it so far.

leonardoalt:
Yea my point was more to be sure we're not breaking the test script for the general use.

PR #11673, opened 2021-07-16
Update SMTChecker tests with z3 4.8.12

leonardoalt:

PR #11672, opened 2021-07-16
Bump docker images to z3 4.8.12

leonardoalt:
After these are built, their hashes will be used in the PR that updates the SMTChecker tests. When those tests pass, we can merge this.

PR #11938, opened 2021-09-13
Docker: Update ossfuzz docker image

### LINKED ISSUE #373: Evmone static library clang build fails while llvm-ar reads mri script
The Solidity fuzzers on oss-fuzz build evmone static libraries (see https://github.com/google/oss-fuzz/blob/8851460f3d7e712a406909acdf8b3aa2882e011d/projects/solidity/Dockerfile#L56-L61) as part of their base docker image. During a [recent update]() from `v0.4.0` -> `0.8.0` (see https://github.com/google/oss-fuzz/pull/6207) it was noticed that evmone-standalone.a build fails (failing log: https://gist.github.com/bshastry/e81e76cc57c393b07accee7599548c88).

The error occurs here

```
2021-08-23T08:54:56.0397220Z [1/2] Linking CXX static library lib/x86_64-linux-gnu/libevmone-standalone.a
2021-08-23T08:54:56.0470727Z FAILED: : && /usr/local/bin/cmake -E rm -f lib/x86_64-linux-gnu/libevmone-standalone.a && /usr/local/bin/llvm-ar qc lib/x86_64-linux-gnu/libevmone-standalone.a   && /usr/local/bin/llvm-ranlib lib/x86_64-linux-gnu/libevmone-standalone.a && cd /src/evmone/build/lib/evmone && /usr/local/bin/llvm-ar -M < evmone-standalone.mri
2021-08-23T08:54:56.0472769Z /usr/local/bin/llvm-ar: error: script line 1: unknown command: OPEN
2021-08-23T08:54:56.0473431Z ninja: build stopped: subcommand failed.
```

Sadly, this is stalling Solidity fuzzing :( 

Pointers to fix the problem welcome :)

bshastry:
The update does the following
  - pulls in clang-14 that is currently used by ossfuzz
  - updates cmake
  - updates evmone (in order to fix https://github.com/ethereum/evmone/issues/373)

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -33,7 +33,7 @@ RUN apt-get update; \
 		ninja-build git wget \
 		libbz2-dev zlib1g-dev git curl uuid-dev \
 		pkg-config openjdk-8-jdk liblzma-dev unzip mlton m4; \
-    apt-get install -qy python3-pip sphinx-doc sphinx-common;
+    apt-get install -qy python3-pip;
 
 # Install cmake 3.21.2 (minimum requirement is cmake 3.10)
 RUN wget https://github.com/Kitware/CMake/releases/download/v3.21.2/cmake-3.21.2-Linux-x86_64.sh; \
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -19,7 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with solidity.  If not, see <http://www.gnu.org/licenses/>
 #
-# (c) 2016-2019 solidity contributors.
+# (c) 2016-2021 solidity contributors.
 #------------------------------------------------------------------------------
 FROM gcr.io/oss-fuzz-base/base-clang as base
 LABEL version="12"
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -33,7 +33,7 @@ RUN apt-get update; \
 		ninja-build git wget \
 		libbz2-dev zlib1g-dev git curl uuid-dev \
 		pkg-config openjdk-8-jdk liblzma-dev unzip mlton m4; \
-    apt-get install -qy python-pip python-sphinx;
+    apt-get install -qy python3-pip sphinx-doc sphinx-common;
 
 # Install cmake 3.21.2 (minimum requirement is cmake 3.10)
 RUN wget https://github.com/Kitware/CMake/releases/download/v3.21.2/cmake-3.21.2-Linux-x86_64.sh; \
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
 #------------------------------------------------------------------------------
 FROM gcr.io/oss-fuzz-base/base-clang as base
-LABEL version="11"
+LABEL version="12"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


bshastry:
The failing build should be fixed once https://github.com/ethereum/solidity/pull/11943 and https://github.com/ethereum/solidity/pull/11941 have been merged and this PR rebased on them.

scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
```diff
@@ -33,12 +33,13 @@ RUN apt-get update; \
 		ninja-build git wget \
 		libbz2-dev zlib1g-dev git curl uuid-dev \
 		pkg-config openjdk-8-jdk liblzma-dev unzip mlton m4; \
-    apt-get install -qy python-pip python-sphinx;
+    apt-get install -qy python3-pip sphinx-doc sphinx-common;
```
cameel: Oh, this is actually a good chance to try removing the old version of sphinx from the image again. Basically this: https://github.com/ethereum/solidity/pull/11672#discussion_r671321645 but last time we tried it my PR that fixes doc warnings was not yet merged and we could not do that (https://github.com/ethereum/solidity/pull/11673#issuecomment-882073364).

```suggestion
    apt-get install -qy python3-pip;
```

cameel: I see that his is actually a different image. But then Sphinx should not be needed in it at all.

scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
```diff
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
 #------------------------------------------------------------------------------
 FROM gcr.io/oss-fuzz-base/base-clang as base
```
cameel, line 24: I guess this pulls latest `base-clang` because there's no tag used here? I think we should use a tag with a specific version - to prevent this layer from being taken from cache when we want version to change and also to make it obvious that we're relying on a specific version that should be bumped from time to time.

scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
```diff
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
```
cameel:
```suggestion
# (c) 2016-2021 solidity contributors.
```
Maybe it would be better to get rid of this comment though? Seems pointless to me.

scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
bshastry: This is a good point. I notice that Google uses the tag `latest` to refer to the most recent base-clang image. I'm not sure why if pull by tag, even the tag is cached or if the newer tag is fetched.

```
FROM gcr.io/oss-fuzz-base/base-clang:latest as base
```

So, if the `latest` one month later points to an image with a different digest, the change may not be fetched (or reflected in the dockerfile).

As a solution to make it explicit we rely on a specific version, we could pull by digest. Pulling the current digest would look like

```
FROM gcr.io/oss-fuzz-base/base-clang@sha256:a964dcd3a66bd3898e82869c720a74553d7fe77cfa5d0b87618d59ec7d4f0314 as base
```

List of base-clang images hosted by Google here: https://console.cloud.google.com/gcr/images/oss-fuzz-base/GLOBAL/base-clang

Notice that they are updated almost on a daily basis!

cameel: ok, looks like they are not tagging them with versions.

I think that our CI builds without cache so it's not a big issue. Might be a bigger one if image was meant to be built locally. The bigger problem for me is that the base image really gets updated on any modification to the Dockerfile but it's not very obvious. It gets updated even if you just bump the version. Hard-coding a hash is not great because it's tedious to update and it does not really tell you what you are updating to. So I guess not much we can do here unless Google decides to start tagging it.

In this case I think I'd just add `latest` to make this more obvious.

cameel (approved):
I'd add `latest` but looks find apart from that.

cameel:
What about a PR updating `.circleci/config.yml`?

bshastry:
> I'd add `latest` but looks find apart from that.

This PR got merged before I could update it. Will create a small PR for it later.

> What about a PR updating .circleci/config.yml?

Will do.

cameel:
> This PR got merged before I could update it. Will create a small PR for it later.

Well, at this point I'd leave it as is. Rebuilding these images is a hassle and I think it's not important enough to go through that again :)

PR #11942, opened 2021-09-13
Allow UserDefinedValueType.uwrap (and wrap) as RHS of constant decl

### LINKED ISSUE #11925: Permit user-defined type constant values
```
==== Source: su0.sol ====
type T0 is int224;
pragma solidity >= 0.0.0;
contract C0 {
  T0 constant s0 = T0.wrap(int224(165521356710917456517261742455526507355687727119203895813322792776));
}
// ----
// TypeError 8349: (su0.sol:78-161): Initial value for constant variable has to be compile-time constant.
```

The above could be permitted. Right now it is not. Found by fuzzing.

hrkrshnn:
Closes https://github.com/ethereum/solidity/issues/11925

Needed to make `MyType.unwrap` and `MyType.unwrap` as pure in the process. This change affected some
existing tests ("statement has no effect").


libsolidity/ast/Types.cpp
```diff
@@ -3499,7 +3499,9 @@ bool FunctionType::isPure() const
 		m_kind == Kind::ABIEncodeWithSelector ||
 		m_kind == Kind::ABIEncodeWithSignature ||
 		m_kind == Kind::ABIDecode ||
-		m_kind == Kind::MetaType;
+		m_kind == Kind::MetaType ||
+		m_kind == Kind::Wrap ||
+		m_kind == Kind::Unwrap;
```
hrkrshnn, line 3504: The `FunctionType` already has the state mutability set to `Pure` (see https://github.com/ethereum/solidity/blob/develop/libsolidity/ast/Types.cpp#L3813), but turns out that seem to be ignored. There seems to be a TODO on top to fix this.

ekpyron (approved):
Seems like this is all a bit of a mess and could use some general cleanup, but good enough for the issue at hand!

libsolidity/ast/Types.cpp
chriseth: These might be the two different concepts of pureness in the compiler.

PR #11946, opened 2021-09-13
UserDefinedValueType: test to validate zero-cost-abstraction claim.

hrkrshnn:
Seems that the optimizer can indeed make it a zero-cost-abstraction!


chriseth (approved):
Actually, do we have a test for `MyInt public x;` ?

hrkrshnn:
@chriseth we do: https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/userDefinedValueType/ownable.sol#L7

PR #11947, opened 2021-09-13
Added an abi function test for UserDefinedValueType

hrkrshnn:
A test to see if `abi.encode*(..., (CustomType))` and `abi.decode(..., (CustomType))` works as
intended.

### DIFF
--- a/test/libsolidity/semanticTests/userDefinedValueType/abicodec.sol
+++ b/test/libsolidity/semanticTests/userDefinedValueType/abicodec.sol
@@ -1,0 +1,36 @@
+// A test to see if `abi.encodeWithSelector(..., (CustomType))` works as intended.
+contract C {
+    type MyInt is int;
+    function f(MyInt x) external returns(MyInt a, MyInt b, MyInt c, MyInt d) {
+        a = MyInt.wrap(-1);
+        b = MyInt.wrap(0);
+        c = MyInt.wrap(1);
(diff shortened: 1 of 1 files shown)


test/libsolidity/semanticTests/userDefinedValueType/abicodec.sol
```diff
@@ -0,0 +1,36 @@
...
+contract C {
+    type MyInt is int;
+    function f(MyInt x) external returns(MyInt a, MyInt b, MyInt c, MyInt d) {
+        a = MyInt.wrap(-1);
+        b = MyInt.wrap(0);
+        c = MyInt.wrap(1);
+        d = x;
+    }
+    function g() external returns(bool) {
+        (bool success1, bytes memory ret1) =  address(this).call(abi.encodeWithSelector(this.f.selector, MyInt.wrap(5)));
```
chriseth, line 11: Why does `abi.encodeWithSelector` work with both `MyInt` and `int`?

hrkrshnn: Why is that an issue?

chriseth: Ah, we don't do any type checking, right?

chriseth (approved):

PR #11949, opened 2021-09-13
Make semantics tests use the proper via-IR-pipeline and make yul optimizer requirements explicit.

ekpyron:
Not sure we want to do this like that and this will still break in the non-extracted tests as is, but we should probably do something like this.


ekpyron:
[The remaining test failures are weird and potentially due to buggy library handling in our tests, need to have a closer look]

ekpyron:
Closing this... I still think we should do this, but since nobody picked it up in the past half year, I don't expect anyone will any time soon...

ekpyron:
Opened https://github.com/ethereum/solidity/issues/12668 instead.

Assawal:
Opened #12668 instead.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitH­ub Mobile for iOS or Android.
You are receiving th­is because you are subscribed to this th­read. Monday, 14 February 2022, 11:26pm +05:00 from Daniel Kirchner  ***@***.*** :

PR #11951, opened 2021-09-14
Use locationOf helper instead of accessing debugData directly.

ekpyron:
We had at least one null pointer dereference detected by nightly CI: https://app.circleci.com/pipelines/github/ethereum/solidity/18848/workflows/75f0fbf6-47ba-4cc1-8ab7-ee0883570fcb/jobs/838283.

This way, even if the idea is that eventually every node always has valid ``debugData`` (which is not the case currently), we can just change the helper to assert that instead of using an empty location as default.

I think this PR covers all occurrences of ``debugData->location`` in ``libyul`` that aren't already specifically guarded.

### DIFF
--- a/libyul/AST.h
+++ b/libyul/AST.h
@@ -83,32 +83,28 @@ struct Continue { std::shared_ptr<DebugData const> debugData; };
 /// Leave statement (valid within function)
 struct Leave { std::shared_ptr<DebugData const> debugData; };
 
-struct LocationExtractor
-{
-	template <class T> langutil::SourceLocation operator()(T const& _node) const
-	{
-		return _node.debugData ? _node.debugData->location : langutil::SourceLocation{};
-	}
-};
-
 /// Extracts the source location from a Yul node.
 template <class T> inline langutil::SourceLocation locationOf(T const& _node)
 {
-	return std::visit(LocationExtractor(), _node);
+	return _node.debugData ? _node.debugData->location : langutil::SourceLocation{};
 }
 
-struct DebugDataExtractor
+/// Extracts the source location from a Yul node.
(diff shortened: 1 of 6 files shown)


libyul/AST.h
```diff
@@ -83,32 +83,28 @@ struct Continue { std::shared_ptr<DebugData const> debugData; };
...
+	return _node.debugData ? _node.debugData->location : langutil::SourceLocation{};
+}
 
 /// Extracts the source location from a Yul node.
-template <class T> inline langutil::SourceLocation locationOf(T const& _node)
+template <class... Args> inline langutil::SourceLocation locationOf(std::variant<Args...> const& _node)
 {
-	return std::visit(LocationExtractor(), _node);
+	return std::visit([](auto const& _arg) { return locationOf(_arg); }, _node);
 }
```
ekpyron, line 97: We could also add stuff like
```C++
template <class... Args> inline langutil::SourceLocation locationOf(std::unique_ptr<Args...> const& _node)
{
	return _node ? locationOf(*_node): langutil::SourceLocation{};
}
```
and the like...

libyul/AST.h
```diff
@@ -83,32 +83,28 @@ struct Continue { std::shared_ptr<DebugData const> debugData; };
...
-	template <class T> std::shared_ptr<DebugData const> const& operator()(T const& _node) const
-	{
-		return _node.debugData;
-	}
-};
+	return _node.debugData;
+}
 
 /// Extracts the debug data from a Yul node.
-template <class T> inline std::shared_ptr<DebugData const> const& debugDataOf(T const& _node)
```
ekpyron, line 109: const ref to ``shared_ptr`` is generally an anti-pattern

chriseth (approved):

PR #11954, opened 2021-09-14
Added a multisource test for UserDefinedValueTypes and imports

hrkrshnn:
Testing if `import {MyType} from "source";` works 

Depends on https://github.com/ethereum/solidity/pull/11965

### DIFF
--- a/test/libsolidity/semanticTests/userDefinedValueType/multisource.sol
+++ b/test/libsolidity/semanticTests/userDefinedValueType/multisource.sol
@@ -1,0 +1,14 @@
+==== Source: A ====
+type MyInt is int;
+type MyAddress is address;
+==== Source: B ====
+import {MyInt, MyAddress as OurAddress} from "A";
+contract A {
+    function f(int x) external view returns(MyInt) { return MyInt.wrap(x); }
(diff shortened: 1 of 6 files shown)


test/libsolidity/semanticTests/userDefinedValueType/multisource.sol
```diff
@@ -0,0 +1,14 @@
+==== Source: A ====
+type MyInt is int;
+type MyAddress is address;
+==== Source: B ====
+import {MyInt, MyAddress as OurAddress} from "A";
+contract A {
+    function f(int x) external view returns(MyInt) { return MyInt.wrap(x); }
+    function f(address x) external view returns(OurAddress) { return OurAddress.wrap(x); }
```
hrkrshnn, line 8: Should renaming still have `.wrap` and `.unwrap`?

chriseth:
Can you add another test where we use `ModuleName.TypeName` to access the user-defined type name? And then please check if wrap / unwrap is available.

ekpyron:
Do we have tests for importing stuff as modules? That will probably suffer from the same error message issue as in the contract case...
EDIT: damn chris was faster :-).

chriseth (approved):

PR #11956, opened 2021-09-14
Allow accessing user defined value type members via contract name.

### LINKED ISSUE #11955: User defined type constructors not visible across non-inherited contracts
The following program compiles,

```
contract C { type T0 is uint; }
contract D { C.T0 x; }
```

but the next one does not.

```
contract C { type T0 is uint; }
contract D { C.T0 x = C.T0.wrap(uint(1)); }
```

Found by fuzzing.

ekpyron:
Seems to be as simple as this, ~~but I'll still add a few semantics tests just to be sure.~~ (added one)
Fixes https://github.com/ethereum/solidity/issues/11955.

### DIFF
--- a/test/libsolidity/syntaxTests/userDefinedValueType/wrap_unwrap_via_contract_name_different.sol
+++ b/test/libsolidity/syntaxTests/userDefinedValueType/wrap_unwrap_via_contract_name_different.sol
@@ -1,0 +1,8 @@
+contract C { type T is uint; }
+library L { type T is uint; }
+contract D
+{
+    C.T x = L.T.wrap(uint(1));
+}
+// ----
+// TypeError 7407: (86-103): Type user defined type T is not implicitly convertible to expected type user defined type T.
--- a/libsolidity/ast/AST.h
+++ b/libsolidity/ast/AST.h
@@ -751,6 +751,7 @@ public:
 	Type const* type() const override;
 
 	TypeName const* underlyingType() const { return m_underlyingType.get(); }
+	bool isVisibleViaContractTypeAccess() const override { return true; }
 
 private:
 	/// The name of the underlying type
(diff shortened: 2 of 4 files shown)


hrkrshnn (dismissed):

test/libsolidity/syntaxTests/userDefinedValueType/wrap_unwrap_via_contract_name_different.sol
```diff
@@ -0,0 +1,8 @@
+contract C { type T is uint; }
+library L { type T is uint; }
+contract D
+{
+    C.T x = L.T.wrap(uint(1));
+}
+// ----
+// TypeError 7407: (86-103): Type user defined type T is not implicitly convertible to expected type user defined type T.
```
ekpyron, line 8: Yeah, ok, this works, but the error message sucks :-D.

hrkrshnn: How would you change it?

hrkrshnn: Maybe should also include `L.T` and `C.T` in the `Type::toString()`.

chriseth (approved):

test/libsolidity/syntaxTests/userDefinedValueType/wrap_unwrap_via_contract_name_different.sol
ekpyron: If I were to change it, I'd do the same as we currently do for enums... but @chriseth and I just said, we don't really care, people should do this on file-level with nice names anyways :-).

hrkrshnn: Here's the PR for this: https://github.com/ethereum/solidity/pull/11957

PR #11957, opened 2021-09-14
Canonicalname for user defined value types

hrkrshnn:
Came up in https://github.com/ethereum/solidity/pull/11956#discussion_r708400289

### DIFF
--- a/libsolidity/analysis/NameAndTypeResolver.cpp
+++ b/libsolidity/analysis/NameAndTypeResolver.cpp
@@ -612,13 +612,31 @@ bool DeclarationRegistrationHelper::visitNode(ASTNode& _node)
 
 	if (auto* declaration = dynamic_cast<Declaration*>(&_node))
 		registerDeclaration(*declaration);
+
+	if (auto* annotation = dynamic_cast<TypeDeclarationAnnotation*>(&_node.annotation()))
+	{
+		string canonicalName = dynamic_cast<Declaration const&>(_node).name();
+		solAssert(!canonicalName.empty(), "");
+
+		for (
+			ASTNode const* scope = m_currentScope;
+			scope != nullptr;
+			scope = m_scopes[scope]->enclosingNode()
+		)
+			if (auto decl = dynamic_cast<Declaration const*>(scope))
+			{
+				solAssert(!decl->name().empty(), "");
+				canonicalName = decl->name() + "." + canonicalName;
+			}
+
+		annotation->canonicalName = canonicalName;
+	}
+
 	if (dynamic_cast<ScopeOpener const*>(&_node))
 		enterNewSubScope(_node);
 
 	if (auto* variableScope = dynamic_cast<VariableScope*>(&_node))
 		m_currentFunction = variableScope;
-	if (auto* annotation = dynamic_cast<TypeDeclarationAnnotation*>(&_node.annotation()))
-		annotation->canonicalName = currentCanonicalName();
 
 	return true;
 }
@@ -666,25 +684,6 @@ void DeclarationRegistrationHelper::registerDeclaration(Declaration& _declaratio
 
 	solAssert(_declaration.annotation().scope == m_currentScope, "");
 	solAssert(_declaration.annotation().contract == m_currentContract, "");
-}
-
-string DeclarationRegistrationHelper::currentCanonicalName() const
-{
-	string ret;
-	for (
-		ASTNode const* scope = m_currentScope;
-		scope != nullptr;
-		scope = m_scopes[scope]->enclosingNode()
-	)
-	{
-		if (auto decl = dynamic_cast<Declaration const*>(scope))
-		{
-			if (!ret.empty())
-				ret = "." + ret;
-			ret = decl->name() + ret;
-		}
-	}
-	return ret;
 }
 
 }
--- a/libsolidity/ast/AST.h
+++ b/libsolidity/ast/AST.h
@@ -750,6 +750,8 @@ public:
 
 	Type const* type() const override;
 
+	TypeDeclarationAnnotation& annotation() const override;
+
 	TypeName const* underlyingType() const { return m_underlyingType.get(); }
 	bool isVisibleViaContractTypeAccess() const override { return true; }
 
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -2555,7 +2555,7 @@ bool UserDefinedValueType::operator==(Type const& _other) const
 
 string UserDefinedValueType::toString(bool /* _short */) const
 {
-	return "user defined type " + definition().name();
+	return "user defined type " + *definition().annotation().canonicalName;
 }
 
 vector<tuple<string, Type const*>> UserDefinedValueType::makeStackItems() const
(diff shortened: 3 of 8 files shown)


libsolidity/ast/AST.h
```diff
@@ -730,7 +730,7 @@ class EnumValue: public Declaration
  * User defined value types, i.e., custom types, for example, `type MyInt is int`. Allows creating a
  * zero cost abstraction over value type with stricter type requirements.
  */
-class UserDefinedValueTypeDefinition: public Declaration
+class UserDefinedValueTypeDefinition: public Declaration, public ScopeOpener
```
hrkrshnn: The ScopeOpener is a little bit of a hack.

Without it, https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/NameAndTypeResolver.cpp#L680 this branch is never true and therefore always have empty names. (The value there is set in https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/NameAndTypeResolver.cpp#L615)

The other way is to make an exception for user defined value types inside `DeclarationRegistrationHelper::currentCanonicalName()`.

libsolidity/ast/Types.cpp
```diff
@@ -2555,7 +2555,7 @@ bool UserDefinedValueType::operator==(Type const& _other) const
 
 string UserDefinedValueType::toString(bool /* _short */) const
 {
-	return "user defined type " + definition().name();
+	return "user defined type " + *definition().annotation().canonicalName;
```
axic, line 2558: This is not fully related to this PR, but why does this have this free form text, when the other cases use strings which look more like code, such as `enum X`, `tuple(int256,int256)`, etc.

test/libsolidity/ASTJSON/userDefinedValueType.json
```diff
@@ -291,15 +291,15 @@
               "typeDescriptions":
               {
                 "typeIdentifier": "t_userDefinedValueType$_MyAddress_$18",
-                "typeString": "user defined type MyAddress"
+                "typeString": "user defined type C.MyAddress"
```
axic, line 294: Why is it only updated here and not in cases above where it still doesn't use the FQN?

libsolidity/ast/Types.cpp
hrkrshnn: What would be your suggestion? Just `type MyInt`?

axic: I wonder if we even need a special keyword, but `type MyInt` does sound good enough.

Mostly unsure about what is the use case for this, in the AST the `typeString` for mappings looks like valid Solidity syntax. If it is kind of supposed to be valid syntax, then this would break it.

hrkrshnn: This is currently used in error messages such as "Type user defined type L.T is not implicitly convertible to expected type user defined type C.T."

If we change to just `type MyInt`, the above error message would look a bit weird:

"Type type L.T is not implicitly convertible to expected type type C.T."

maybe `userdefined MyInt` is better?

test/libsolidity/ASTJSON/userDefinedValueType.json
hrkrshnn: The others are in global scope. But not sure what you mean.

libsolidity/ast/AST.h
chriseth: I don't think the hack is worth it, especially as `canonicalName` won't help with two user defined value types defined in different source files at the top level.

chriseth: Also it's weird that this requires ScopeOpener and not Scopable.

ekpyron: Why not just change this stuff? That's the only use of ``currentCanonicalName``, right? And anything that has ``TypeDeclarationAnnotation`` annotations, is a ``Declaration``? Then just remove ``currentCanonicalName`` and change 
```
	if (auto* annotation = dynamic_cast<TypeDeclarationAnnotation*>(&_node.annotation()))
		annotation->canonicalName = currentCanonicalName();
```
to sth like
```
	if (auto* annotation = dynamic_cast<TypeDeclarationAnnotation*>(&_node.annotation()))
    {
		annotation->canonicalName = dynamic_cast<Declaration&>(_node).name();
        // traverse and prepend parent scope names...
    }
```

But I only peeked at the links - might not work for reasons I didn't see.
But we should try to avoid making an AST node that doesn't open a scope a ``ScopeOpener``.

test/libsolidity/ASTJSON/userDefinedValueType.json
axic: FQN = Fully Qualified Name

libsolidity/ast/Types.cpp
axic: Yeah, `userdefined` is perhaps better. Why do we need to mark it specially though and not just by its name? Built in types can't be overridden.

libsolidity/ast/AST.h
chriseth: Yes, this should be properly fixed.

libsolidity/ast/Types.cpp
hrkrshnn: > Why do we need to mark it specially though and not just by its name

There were some errors where 
```solidity
contract C {
type MyInt is int;
}
contract D {
type MyInt is int;
}
contract D {
// The error message for this also need to mention the scope
C.MyInt x = D.MyInt.wrap(0);
}
```

libsolidity/analysis/NameAndTypeResolver.cpp
```diff
@@ -612,13 +612,38 @@ bool DeclarationRegistrationHelper::visitNode(ASTNode& _node)
 
 	if (auto* declaration = dynamic_cast<Declaration*>(&_node))
 		registerDeclaration(*declaration);
+
+	// Setting the canonicalname before ScopeOpener, so that m_currentScope is not same as the
```
chriseth: Not sure this comment is actually needed.

hrkrshnn: Can remove it.

libsolidity/analysis/NameAndTypeResolver.cpp
```diff
@@ -612,13 +612,38 @@ bool DeclarationRegistrationHelper::visitNode(ASTNode& _node)
...
+	// current _node
+	if (auto* annotation = dynamic_cast<TypeDeclarationAnnotation*>(&_node.annotation()))
+	{
+		string canonicalName;
+
+		if (auto decl = dynamic_cast<Declaration const*>(&_node))
```
chriseth: Can this really be false? If yes, I think we should do something, it would mean that the name of a type declaration does not contain its own name, but the name of a super scope...

libsolidity/ast/Types.cpp
chriseth: I think it's nice to know if it's a struct, a contract or a user-defined type.

test/libsolidity/ASTJSON/userDefinedValueType.json
hrkrshnn: The changes to ABIJson file was missing. Added it now.

chriseth (approved):

libsolidity/ast/Types.cpp
ekpyron: Didn't we say structs **are** user-defined types, so "user-defined type" is a bad designation for the user-defined-value-type things :-)?

PR #11960, opened 2021-09-14
Use StandardJson in the CLI

axic:
Part of #9583.


axic:
Only opened this to have a record of this code.

PR #11963, opened 2021-09-15
Improve abi decoding functions for arrays.

ekpyron:
Improves gas and decreases stack pressure while actually being nicer IMHO.
[noticed this when checking where the stack errors on abi decoding usually actually come from - variables changing in loops are the worst]


libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -1186,11 +1186,11 @@ string ABIFunctions::abiDecodingFunctionArrayAvailableLength(ArrayType const& _t
 					mstore(array, length)
 					dst := add(array, 0x20)
 				</dynamic>
-				let src := offset
-				if gt(add(src, mul(length, <stride>)), end) {
+				let srcEnd := add(offset, mul(length, <stride>))
+				if gt(srcEnd, end) {
 					<revertInvalidStride>()
```
ekpyron, line 1192: Actually... invalid stride? Why was the revert message here ever "invalid stride"? Am I missing something here?
This is really an out-of-bounds error, isn't it?

ekpyron: What does "invalid stride" even mean as a runtime error ;-)?

chriseth: true

chriseth: `length` is already checked to be non-gigantic, right?

chriseth: Ah, indirectly through the allocation.

ekpyron:
[pushed command line test updates]

chriseth (dismissed):
We should really have an optimizer stage that improves this.

ekpyron:
I just pushed some more gas cost updates - I didn't update the non-optimized run before.

chriseth (approved):

PR #11965, opened 2021-09-15
User defined value types changed toString()

hrkrshnn:
Came up in https://github.com/ethereum/solidity/pull/11957#discussion_r708412929

Depends on https://github.com/ethereum/solidity/pull/11957

### DIFF
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -2555,7 +2555,7 @@ bool UserDefinedValueType::operator==(Type const& _other) const
 
 string UserDefinedValueType::toString(bool /* _short */) const
 {
-	return "user defined type " + *definition().annotation().canonicalName;
+	return *definition().annotation().canonicalName;
 }
 
 vector<tuple<string, Type const*>> UserDefinedValueType::makeStackItems() const
--- a/test/libsolidity/ABIJson/user_defined_value_type.sol
+++ b/test/libsolidity/ABIJson/user_defined_value_type.sol
@@ -36,7 +36,7 @@ contract C {
 //     "outputs":
 //     [
 //       {
-//         "internalType": "user defined type C.MyAddress",
+//         "internalType": "C.MyAddress",
 //         "name": "",
 //         "type": "address"
 //       }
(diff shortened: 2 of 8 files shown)


libsolidity/ast/Types.cpp
```diff
@@ -2555,7 +2555,7 @@ bool UserDefinedValueType::operator==(Type const& _other) const
 
 string UserDefinedValueType::toString(bool /* _short */) const
 {
-	return "user defined type " + *definition().annotation().canonicalName;
+	return "userdefined " + *definition().annotation().canonicalName;
```
chriseth: What about `user-defined`?

hrkrshnn: Works for me.

ekpyron: Didn't you both argue that e.g. structs and contracts are also ``user-defined types`` and that this is **not** a good designation for the new user-defined-value-types?

hrkrshnn: Yeah, that is true. This is so conflicting now. What else then? `user-defined-value-type MyInt`?

ekpyron: I'd myself just make it ``MyInt``. It's the most basic form of a user-defined type, so to me it seems fine to only mark the other user-defined types specifically.
Just syntactically you say "struct S" and here you get "struct S" and you say "contract C" and here you get "contract C", so by that principle it would be sth like "type MyInt", which is rather redundant, so "MyInt" is fine ;-).

But I'm also fine with both "user-defined" and "user-defined-value-type", just saying that it's not really consistent.

chriseth: I'm also fine with using `type T` here.

hrkrshnn: Changing to `type T` would lead to bad error messages

> TypeError 7407: (86-103): Type type L.T is not implicitly convertible to expected type type C.T.

Maybe we salvage this by adding quotes?

> TypeError 7407: (86-103): Type "type L.T" is not implicitly convertible to expected type "type C.T."

Or the other option is make `toString()` just return only canonical name.

> TypeError 7407: (86-103): Type L.T is not implicitly convertible to expected type C.T.

chriseth:
Yeah, maybe we can just drop the prefix, since it is a value type.

chriseth:
Please double-check where `toString` is used - I think the library delegatecall abi uses it, but probably only for structs.

chriseth (approved):

Issue #11813, opened 2021-08-17
Try out include-what-you-use


cameel:
This is a part of #10269 but getting includes consistent also came up recently in #11812 as it might prevent bugs like #11811 and #11103. I'm creating a more focused issue to try out specifically [include-what-you-use](https://include-what-you-use.org) even if it does not solve the original problem as a whole.

---

The task here is to install [include-what-you-use](https://include-what-you-use.org) locally and run it on our codebase. Then submit changes in a PR so that we can discuss them and decide if it's good enough to add it as a PR check that runs automatically (it could run as a part of the [`chk_coding_style` job](https://github.com/ethereum/solidity/blob/v0.8.7/.circleci/config.yml#L320-L336)). Please also post all the commands you used for installing/building/running it.

The main criterion is whether it's reliable enough and does not cause too many false positives that would have to be ignored in some way or, worse, would make the check fail on perfectly valid PRs. Please also make sure that it works not only with our own headers but also with ones from STL and other libraries.

akbhuker:
hey, can I have this issue?
I'm quite new to this so I might need some guidance

cameel:
Sure, go ahead!

If you need help, you can always ask in this issue or on the `#solidity-dev` [Matrix](https://app.element.io/#/room/#ethereum_solidity-dev:gitter.im)/[Gitter](https://gitter.im/ethereum/solidity-dev) channel.

Sild:
Hi, I went around this issue

TLDR: it does't work from the box: solidity does not compile after iwyu updates (some of includes are broken I guess)

Reproduction snippet:

```
cd ~
apt install libclang-13-dev
git clone https://github.com/include-what-you-use/include-what-you-use.git
cd include-what-you-use
git fetch --all && git checkout clang_13
mkdir build && cd build
cmake -G "Unix Makefiles" -DCMAKE_C_COMPILER=/usr/bin/clang-13 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-13 -DCMAKE_PREFIX_PATH=/usr/lib/llvm-13/ ../
make
mv bin/include-what-you-use /usr/bin/
cd /solidity/build
cmake -DCMAKE_C_COMPILER=/usr/bin/clang-13 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-13 -DUSE_Z3=OFF -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE=include-what-you-use ../
# build is really slow with iwyu
[...]

PR #11273, opened 2021-04-20
Adds fmtlib as 3rdparty dependency for easier string composition.

christianparpart:
Adds fmtlib as 3rdparty dependency. It'll be part of C++20 anyways, and I feel like maybe other code paths (not just mine) will definitely benefit from it.

Issue #8860, opened 2020-05-06
External dependencies: json-cpp, fmtlib, ranges-v3, intx, z3, cvc4


ekpyron:
Situation:
- We have been talking about dependencies a while recently, but never came to a decision.
- It looks like we like to introduce dependencies on fmtlib and ranges-v3 and maybe intx. There also was the idea to replace json-cpp with nlohmann/json.
- Our repo contains a copy of evmc.
- For json-cpp we use the cmake ``ExternalProject_Add`` mechanism
- We also depend on boost and optionally on Z3 and CVC4. For those we use plain standard cmake ``find_package``.

Issues with ``ExternalProject_Add``:
- There is no way to override it and dynamically link against system json-cpp instead.
- ``make`` requires a network connection and starts downloading
- The latter complicates e.g. PPA scripts.
- We can't pre-pack dependencies to our CI docker images, increasing build time.

Apart from boost, all the other dependencies except json-cpp (including nlohmann/json as potential json-cpp replacement) are header-only or nearly header-only.

When considering to add new dependencies we could:
- Reuse the ``ExternalProject_Add`` mechanism despite its drawbacks.
- Use ``git submodules`` for the new dependencies (or all dependencies).
- Instead of ``submodules`` use ``subtree``s.
- Switch everything to plain standard cmake ``find_package``, while maintaining a better ``ìnstall_deps`` script to locally install dependencies in the build tree.
- Combine ``find_package`` with other mechanisms as fallback.

axic:
We discussed this countless times on gitter, but apparently no record of that here. 

I would say this issue **should not** block any other issues in https://github.com/ethereum/solidity/projects/40, in particular #3851 and #6900.

For those issues to proceed I think it is fine just to use `ExternalProject` as we do with jsoncpp.

Independently later we could clean up dependency handling. My personal preference is for git subtrees for reasonably sized projects (and intx, nlohmann-json, ranges-v3 all fit this bill).

christianparpart:
> My personal preference is for git subtrees for reasonably sized projects (and intx, nlohmann-json, ranges-v3 all fit this bill).

@axic the good thing about submodules is, you do not have to check them out and can still opt for system installed dependencies.

This ticket is also quite old, and I'm now at some ticket where fmtlib actually does make sense (solidity LSP server) for server side logging. I think I'd just use submodules for now until someone complains in the PR, as I think I can always change that at a later point.

[...]

PR #1542, opened 2017-01-04
Create source tarballs for releases.

chriseth:
Fixes https://github.com/ethereum/solidity/issues/1183

PR #11967, opened 2021-09-15
Basic nlohmann-json integration.

### LINKED ISSUE #6900: Consider replacing jsoncpp
- The repo looks like jsoncpp only receives minimal maintenance.
- The workaround of manually downloading jsoncpp in cmake is evil (we could of course attempt to change that in isolation, but there don't seem to be package updates for jsoncpp e.g. in recent ubuntu either making this a continuing hassle).
- jsoncpp is neither the fastest nor the smallest nor the most conformant option.
- There are a number of pretty much drop-in replacements available.
- We can easily move to a header-only library which pretty much avoids all dependency issues in perpetuity.

I'd suggest to consider replacing jsoncpp with either http://rapidjson.org/ or https://github.com/nlohmann/json (or any other of the available candidates).

Is there any reason for staying with jsoncpp other than the required work a move would be (which I wouldn't expect to actually be too bad)?

Closes #3557.

axic:
Fixes #6900.

A partial rebase of https://github.com/aarlt/solidity/commit/1329c25756f2b242d9af7860eda6f37a72d38dc0.


cmake/nlohmann-json.cmake
```diff
@@ -0,0 +1,14 @@
+include(ExternalProject)
+
+ExternalProject_Add(nlohmann-json
```
axic: Should just git import the file, shouldn't we?

libdevcore/JSON.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file JSON.cpp
+ * @author Alexander Arlt <alexander.arlt@arlt-labs.com>
+ * @date 2018
+ */
+
+#include "JSON.h"
```
stackenbotten, line 22: Coding style error

libdevcore/JSON.cpp
```diff
@@ -0,0 +1,82 @@
...
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e) {
+		/*
+		std::cout << "message: " << e.what() << '\n'
+				  << "exception id: " << e.id << '\n'
+				  << "byte position of error: " << e.byte << std::endl;
+		*/
+		if (_errs) {
```
stackenbotten, line 55: Coding style error

libdevcore/JSON.cpp
```diff
@@ -0,0 +1,82 @@
...
```
stackenbotten, line 75: Coding style error

libdevcore/JSON.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include "JSON.h"
+
+#include <sstream>
+#include <map>
+#include <memory>
+
+using namespace std;
+
+namespace dev {
```
stackenbotten, line 30: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 22: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 55: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 75: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 30: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 55: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 75: Coding style error

libdevcore/JSON.cpp
stackenbotten, line 30: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,46 @@ void removeNullMembersHelper(Json::Value& _json)
...
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try {
+		// parsing input with a syntax error
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e) {
+		// NOTE: e.id() gives the code and e.byte() gives the byte offset
+		if (_errs) {
```
stackenbotten, line 98: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,46 @@ void removeNullMembersHelper(Json::Value& _json)
...
+bool jsonParseStrict(string const& _input, Json& _json, string* _errs /* = nullptr */)
 {
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try {
+		// parsing input with a syntax error
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e) {
```
stackenbotten, line 96: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,46 @@ void removeNullMembersHelper(Json::Value& _json)
...
 {
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try {
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e) {
+		// NOTE: e.id() gives the code and e.byte() gives the byte offset
+		if (_errs) {
```
stackenbotten, line 98: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,46 @@ void removeNullMembersHelper(Json::Value& _json)
...
-bool jsonParseStrict(string const& _input, Json::Value& _json, string* _errs /* = nullptr */)
+bool jsonParseStrict(string const& _input, Json& _json, string* _errs /* = nullptr */)
 {
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try {
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e) {
```
stackenbotten, line 96: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,49 @@ void removeNullMembersHelper(Json::Value& _json)
...
+bool jsonParseStrict(string const& _input, Json& _json, string* _errs /* = nullptr */)
 {
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try
+	{
+		_json = Json::parse(_input);
+		return true;
+	}
+	catch (Json::parse_error &e)
```
stackenbotten, line 97: Coding style error

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,49 @@ void removeNullMembersHelper(Json::Value& _json)
...
 }
+#endif
 
 } // end anonymous namespace
 
-Json::Value removeNullMembers(Json::Value _json)
+Json removeNullMembers(Json _json)
 {
-	removeNullMembersHelper(_json);
+	// TODO: Support this.
```
axic, line 69: This needs to be fixed.

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,49 @@ void removeNullMembersHelper(Json::Value& _json)
...
-		settings["enableYAMLCompatibility"] = true;
-	}
-	else
-		settings["indentation"] = "";
-	StreamWriterBuilder writerBuilder(settings);
-	string result = print(_input, writerBuilder);
-	if (_format.format == JsonFormat::Pretty)
-		boost::replace_all(result, " \n", "\n");
-	return result;
+	// TODO: support the other features
```
axic: This needs to be fixed.

scripts/create_source_tarball.sh
```diff
@@ -31,7 +31,6 @@ REPO_ROOT="$(dirname "$0")"/..
     fi
     # Add dependencies
     mkdir -p "$SOLDIR/deps/downloads/" 2>/dev/null || true
-    wget -O "$SOLDIR/deps/downloads/jsoncpp-1.9.3.tar.gz" https://github.com/open-source-parsers/jsoncpp/archive/1.9.3.tar.gz
```
axic: Why does this needs jsoncpp, but not range-v3?

`scripts/release_ppa.sh` has both.

cmake/nlohmann-json.cmake
cameel: It's a huge file though (> 25k lines). I don't think that adding external code to the repo is much less of a hack than downloading the package in cmake config.

Maybe we should look into adopting some dependency management system to make using libraries less of a pain? I see that for example [Conan](https://docs.conan.io) and [Hunter](https://hunter.readthedocs.io) are getting some popularity in the last few years. There's also [vcpkg](https://github.com/microsoft/vcpkg) but it seems to be tied to MS stuff and VS in some ways even if it's ultimately meant to be cross-platform. Never used any of them but they look promising.

Tagging @ekpyron becuse he'll likely have an opinion on this :)

libsolidity/ast/AST.h
```diff
@@ -32,10 +32,9 @@
 #include <liblangutil/SourceLocation.h>
 #include <libevmasm/Instruction.h>
 #include <libsolutil/FixedHash.h>
+#include <libsolutil/JSON.h>
```
cameel: This seems unused in this file.

libevmasm/Assembly.h
```diff
@@ -28,12 +28,11 @@
 
 #include <libsolutil/Common.h>
 #include <libsolutil/Assertions.h>
+#include <libsolutil/JSON.h>
 #include <libsolutil/Keccak256.h>
 
 #include <libsolidity/interface/OptimiserSettings.h>
 
-#include <json/json.h>
```
cameel, line 36: Including the library via our own header to abstract it out a bit seems like a good idea but on the other hand I wonder how it would interact with `include-what-you-use` if we were to start using it (#11813).

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -130,33 +130,33 @@ string ASTJsonConverter::namePathToString(std::vector<ASTString> const& _namePat
...
 	{
-		Json::Value arguments(Json::arrayValue);
+		Json arguments(Json::array());
 		for (auto const& tp: _tps->types)
 			appendMove(arguments, typePointerToJson(tp));
 		return arguments;
 	}
 	else
-		return Json::nullValue;
+		return Json{};
```
cameel: So there's no explicit `null` value in this lib?

EDIT: I see in your tests that `nullptr` is treated as `null`. I guess we don't want to use it in that role though?

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -171,16 +171,17 @@ void ASTJsonConverter::appendExpressionAttributes(
...
-	tuple["isSlot"] = Json::Value(_info.second.suffix == "slot");
-	tuple["isOffset"] = Json::Value(_info.second.suffix == "offset");
+	tuple["isSlot"] = Json(_info.second.suffix == "slot");
+	tuple["isOffset"] = Json(_info.second.suffix == "offset");
 	if (!_info.second.suffix.empty())
-		tuple["suffix"] = Json::Value(_info.second.suffix);
-	tuple["valueSize"] = Json::Value(Json::LargestUInt(_info.second.valueSize));
+		tuple["suffix"] = Json(_info.second.suffix);
+	// TODO: assert?
+	tuple["valueSize"] = Json(static_cast<Json::number_integer_t>(_info.second.valueSize));
```
cameel: Yeah, I'd definitely assert. This can truncate the value if it's too big, which I don't think we'd be ok with happening.

Also, is the range of `Json::LargestUInt` and `Json::number_integer_t` the same?

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -72,11 +72,11 @@ map<string, ASTPointer<SourceUnit>> ASTJsonImporter::jsonToSourceUnit(map<string
 
 // =========== general creation functions ==============
 template <typename T, typename... Args>
-ASTPointer<T> ASTJsonImporter::createASTNode(Json::Value const& _node, Args&&... _args)
+ASTPointer<T> ASTJsonImporter::createASTNode(Json const& _node, Args&&... _args)
 {
-	astAssert(member(_node, "id").isInt64(), "'id'-field must be 64bit integer.");
+	astAssert(member(_node, "id").is_number_integer(), "'id'-field must be 64bit integer.");
```
cameel: [`integer_t`](https://nlohmann.github.io/json/api/basic_json/number_integer_t/) seems to be a template parameter. It defaults to `int64_t` but I'd use `static_assert` to document that it must be that eaxact type here.

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -915,129 +915,129 @@ ASTPointer<IndexRangeAccess> ASTJsonImporter::createIndexRangeAccess(Json::Value
...
 
 // ===== helper functions ==========
 
-Json::Value ASTJsonImporter::member(Json::Value const& _node, string const& _name)
+Json ASTJsonImporter::member(Json const& _node, string const& _name)
 {
-	if (!_node.isMember(_name))
-		return Json::nullValue;
+	if (_node.find(_name) == _node.end())
+		return Json{};
```
cameel: I think that [`contains()`](https://nlohmann.github.io/json/api/basic_json/contains/) would be a better replacement for `isMember()`.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1454,15 +1452,15 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
...
-			meta["sources"][s.first]["urls"].append("bzz-raw://" + toHex(s.second.swarmHash().asBytes()));
-			meta["sources"][s.first]["urls"].append(s.second.ipfsUrl());
+			meta["sources"][s.first]["urls"] = Json::array();
+			meta["sources"][s.first]["urls"].emplace_back("bzz-raw://" + toHex(s.second.swarmHash().asBytes()));
+			meta["sources"][s.first]["urls"].emplace_back(s.second.ipfsUrl());
 		}
 	}
 
-	static_assert(sizeof(m_optimiserSettings.expectedExecutionsPerDeployment) <= sizeof(Json::LargestUInt), "Invalid word size.");
-	solAssert(static_cast<Json::LargestUInt>(m_optimiserSettings.expectedExecutionsPerDeployment) < std::numeric_limits<Json::LargestUInt>::max(), "");
```
cameel, line 1526: This assert seems wrong. There should be no cast there. If we cast to `LargestUInt`, we'll always get a value lower than max for this type :)

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1454,15 +1452,15 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
...
-	meta["settings"]["optimizer"]["runs"] = Json::Value(Json::LargestUInt(m_optimiserSettings.expectedExecutionsPerDeployment));
+	static_assert(sizeof(m_optimiserSettings.expectedExecutionsPerDeployment) <= sizeof(Json::number_integer_t), "Invalid word size.");
```
cameel, line 1523: Wouldn't it be better to compare `std::numeric_limits<T>::max()` of both types here? A `sizeof()` check will still pass if `expectedExecutionsPerDeployment` is e.g. `uint64_t` while the limit is `int64_t`.

test/libsolutil/JSON.cpp
```diff
@@ -90,12 +86,12 @@ BOOST_AUTO_TEST_CASE(json_pretty_print)
...
-	"  \"2\": \"2\",\n"
-	"  \"3\":\n"
+	"  \"1\" : 1,\n"
+	"  \"2\" : \"2\",\n"
+	"  \"3\" : \n"
 	"  {\n"
-	"    \"3.1\": \"3.1\",\n"
-	"    \"3.2\": 2\n"
+	"    \"3.1\" : \"3.1\",\n"
+	"    \"3.2\" : 2\n"
```
cameel, line 94: Does the test fail without the extra spaces before `:`? Seems like a weird limitation.

test/libsolutil/JSON.cpp
```diff
@@ -135,31 +131,28 @@ BOOST_AUTO_TEST_CASE(parse_json_strict)
...
+	BOOST_CHECK(json.is_number());
+	BOOST_CHECK(jsonParseStrict("\"hello\"", json, &errors));
+	BOOST_CHECK(json.is_string());
+	BOOST_CHECK(jsonParseStrict("true", json, &errors));
+	BOOST_CHECK(json.is_boolean());
+	BOOST_CHECK(jsonParseStrict("null", json, &errors));
+	BOOST_CHECK(json.is_null());
 
 	// Single quotes are also disallowed by ECMA-404
 	BOOST_CHECK(!jsonParseStrict("'hello'", json, &errors));
```
cameel, line 158: So this part is non-compliant? Do we need it to be compliant?

test/libsolutil/JSON.cpp
```diff
@@ -172,15 +165,12 @@ BOOST_AUTO_TEST_CASE(parse_json_strict)
...
-	BOOST_CHECK(json.isArray());
-	BOOST_CHECK(json[0] == "\x80\xec\x80");
+	BOOST_CHECK(!jsonParseStrict("[ \"\x80\xec\x80\" ]", json, &errors));
+	BOOST_CHECK(errors == "[json.exception.parse_error.101] parse error at line 1, column 4: syntax error while parsing value - invalid string: ill-formed UTF-8 byte; last read: '\"?'");
 
 	// This would be valid more lenient parsers.
 	BOOST_CHECK(jsonParseStrict("[ \"\xF0\x9F\x98\x8A\" ]", json, &errors));
-	BOOST_CHECK(json.isArray());
+	BOOST_CHECK(json.is_array());
 	BOOST_CHECK(json[0] == "😊");
```
cameel, line 174: This too.

test/libsolidity/StandardCompiler.cpp
```diff
@@ -55,97 +63,98 @@ langutil::Error::Severity str2Severity(string const& _cat)
...
-				BOOST_TEST_REQUIRE(linkReferenceResult[fileName][libraryName][i]["start"].isUInt());
-			}
+// TODO: rewrite this properly
+//			for (int i = 0; i < static_cast<int>(linkReferenceResult.size()); ++i)
+//			{
+//				BOOST_TEST_REQUIRE(linkReferenceResult[fileName][libraryName][i].is_object());
+//				BOOST_TEST_REQUIRE(linkReferenceResult[fileName][libraryName][i].size() == 2);
+//				BOOST_TEST_REQUIRE(linkReferenceResult[fileName][libraryName][i]["length"].is_number_unsigned());
+//				BOOST_TEST_REQUIRE(linkReferenceResult[fileName][libraryName][i]["start"].is_number_unsigned());
+//			}
```
cameel, line 132: What's the problem with this one?

test/libsolidity/StandardCompiler.cpp
```diff
@@ -55,97 +63,98 @@ langutil::Error::Severity str2Severity(string const& _cat)
...
+	BOOST_TEST_REQUIRE(_contractResult.is_object());
+	BOOST_TEST_REQUIRE(_contractResult["evm"]["bytecode"].is_object());
 
-	Json::Value const& linkReferenceResult = _contractResult["evm"]["bytecode"]["linkReferences"];
-	BOOST_TEST_REQUIRE(linkReferenceResult.isObject());
+	Json const& linkReferenceResult = _contractResult["evm"]["bytecode"]["linkReferences"];
+	BOOST_TEST_REQUIRE(linkReferenceResult.is_object());
 
-	for (string const& fileName: linkReferenceResult.getMemberNames())
+	for (auto const& [fileName, _]: linkReferenceResult.items())
```
cameel, line 118: Would this work?
```suggestion
	for (string const& fileName: linkReferenceResult.items() | ranges::views::keys)
```

test/libsolidity/StandardCompiler.cpp
```diff
@@ -41,6 +41,14 @@ namespace solidity::frontend::test
 namespace
 {
 
+set<string> getMemberNames(Json const& _json)
+{
+	set<string> keys;
+	for (auto const& [key, _]: _json.items())
+		keys.insert(key);
+	return keys;
+}
```
cameel, line 50: Seems unused. Is it meant to replace the `.getMemberNames()` method?

test/libsolidity/SemanticTest.cpp
```diff
@@ -425,7 +425,7 @@ TestCase::TestResult SemanticTest::runTest(
 			{
 				soltestAssert(
 					m_allowNonExistingFunctions ||
-					m_compiler.methodIdentifiers(m_compiler.lastContractName(m_sources.mainSourceFile)).isMember(test.call().signature),
+					m_compiler.methodIdentifiers(m_compiler.lastContractName(m_sources.mainSourceFile)).count(test.call().signature),
```
cameel:
```suggestion
					m_compiler.methodIdentifiers(m_compiler.lastContractName(m_sources.mainSourceFile)).contains(test.call().signature),
```

Also in other files.

test/libsolidity/GasTest.cpp
```diff
@@ -123,14 +120,14 @@ TestCase::TestResult GasTest::run(ostream& _stream, string const& _linePrefix, b
...
-	Json::Value estimateGroups = compiler().gasEstimates(compiler().lastContractName());
+	Json estimateGroups = compiler().gasEstimates(compiler().lastContractName());
 	if (
 		m_expectations.size() == estimateGroups.size() &&
 		boost::all(m_expectations, [&](auto const& expectations) {
 		auto const& estimates = estimateGroups[expectations.first];
 		return estimates.size() == expectations.second.size() &&
 			boost::all(expectations.second, [&](auto const& entry) {
-				return entry.second == estimates[entry.first].asString();
+				return entry.second == estimates[entry.first].template get<string>();
```
cameel, line 130: Wouldn't it be better to just declare `estimates` as `Json const&` rather than `auto const&`?

solc/CommandLineInterface.cpp
```diff
@@ -463,23 +463,23 @@ bool CommandLineInterface::readInputFiles()
...
 	for (SourceCode const& sourceCode: m_fileReader.sourceCodes() | ranges::views::values)
 	{
-		Json::Value ast;
+		Json ast;
 		astAssert(jsonParseStrict(sourceCode, ast), "Input file could not be parsed to JSON");
-		astAssert(ast.isMember("sources"), "Invalid Format for import-JSON: Must have 'sources'-object");
+		astAssert(ast.contains("sources"), "Invalid Format for import-JSON: Must have 'sources'-object");
 
-		for (auto& src: ast["sources"].getMemberNames())
+		for (auto const& [src, _]: ast["sources"].items())
```
cameel, line 553:
```suggestion
		for (auto const& src: ast["sources"].items() | ranges::views::keys)
```

scripts/create_source_tarball.sh
cameel: Because it does not need either so it did not fail when we were adding `ranges-v3` and forgot to add it here.

If it's here though, the package gets included in sources and build does not rely on an internet connection. It might have been more important in the past when our cmake config was not downloading jsoncpp automatically - this script was created in #1542 and had the wget line from the very beginning while `jsoncpp.cmake` that downloads jsoncpp was only added in #2926.

Personally, I think that external libs should not be a part of our source package but if we do want that then both libs should be listed here. In that case we should also add a cmake option to disable the download - so that we get an error here if we add more libs in the future and forget to list one of them.

libsolutil/JSON.h
```diff
@@ -23,15 +23,17 @@
 
 #pragma once
 
-#include <json/json.h>
+#include <json/json.hpp>
 
 #include <string>
 
+using Json = nlohmann::json;
```
cameel, line 31: Do we really want this in a header?

libsolutil/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ set(sources
 )
 
 add_library(solutil ${sources})
-target_link_libraries(solutil PUBLIC jsoncpp Boost::boost Boost::filesystem Boost::system range-v3)
+target_link_libraries(solutil PUBLIC Boost::boost Boost::filesystem Boost::system range-v3)
```
cameel, line 47: So the new lib is header only. And it's a massive header. How does this affect compilation times?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1645,34 +1643,35 @@ bytes CompilerStack::createCBORMetadata(Contract const& _contract, bool _forIR)
...
-Json::Value gasToJson(GasEstimator::GasConsumption const& _gas)
+Json gasToJson(GasEstimator::GasConsumption const& _gas)
 {
 	if (_gas.isInfinite)
-		return Json::Value("infinite");
+		return Json("infinite");
 	else
-		return Json::Value(util::toString(_gas.value));
+		return Json(util::toString(_gas.value));
+//		return Json(_gas.value.str());
```
cameel: Commented-out code.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -49,65 +49,65 @@ using namespace solidity::langutil;
...
+Json formatError(
 	Error::Severity _severity,
 	string const& _type,
 	string const& _component,
 	string const& _message,
 	string const& _formattedMessage = "",
-	Json::Value const& _sourceLocation = Json::Value(),
-	Json::Value const& _secondarySourceLocation = Json::Value()
+	Json const& _sourceLocation ),
+	Json const& _secondarySourceLocation )
```
cameel:
```suggestion
	Json const& _sourceLocation,
	Json const& _secondarySourceLocation
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -146,13 +146,13 @@ Json::Value formatErrorWithException(
...
-map<string, set<string>> requestedContractNames(Json::Value const& _outputSelection)
+map<string, set<string>> requestedContractNames(Json const& _outputSelection)
 {
 	map<string, set<string>> contracts;
-	for (auto const& sourceName: _outputSelection.getMemberNames())
+	for (auto const& [sourceName, contractList]: _outputSelection.items())
 	{
 		string key = (sourceName == "*") ? "" : sourceName;
-		for (auto const& contractName: _outputSelection[sourceName].getMemberNames())
+		for (auto const& [contractName, _]: contractList.items())
```
cameel, line 158:
```suggestion
		for (string const& contractName: contractList.items() | ranges::views::keys)
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -208,13 +208,13 @@ bool isArtifactRequested(Json::Value const& _outputSelection, string const& _art
...
-bool isArtifactRequested(Json::Value const& _outputSelection, string const& _file, string const& _contract, string const& _artifact, bool _wildcardMatchesExperimental)
+bool isArtifactRequested(Json const& _outputSelection, string const& _file, string const& _contract, string const& _artifact, bool _wildcardMatchesExperimental)
 {
-	if (!_outputSelection.isObject())
+	if (!_outputSelection.is_object())
 		return false;
 
 	for (auto const& file: { _file, string("*") })
-		if (_outputSelection.isMember(file) && _outputSelection[file].isObject())
+		if (_outputSelection.find(file) != _outputSelection.end() && _outputSelection[file].is_object())
```
cameel:
```suggestion
		if (_outputSelection.contains(file) && _outputSelection[file].is_object())
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -557,32 +553,32 @@ std::optional<Json::Value> checkOutputSelection(Json::Value const& _outputSelect
...
 	}
 
-	if (_jsonInput.isMember("runs"))
+	if (_jsonInput.contains("runs"))
 	{
-		if (!_jsonInput["runs"].isUInt())
+		if (!_jsonInput["runs"].is_number_unsigned())
 			return formatFatalError("JSONError", "The \"runs\" setting must be an unsigned number.");
-		settings.expectedExecutionsPerDeployment = _jsonInput["runs"].asUInt();
+		settings.expectedExecutionsPerDeployment = _jsonInput["runs"].get<Json::number_unsigned_t>();
```
cameel: We should validate the range of the value or have a static assert that types are the same.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -740,67 +736,69 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
...
-				ret.smtLib2Responses[hash] = smtlib2Responses[hashString].asString();
+				ret.smtLib2Responses[hash] = response.get<string>();
 			}
 		}
 	}
 
-	Json::Value const& settings = _input.get("settings", Json::Value());
+//	if (!_input.contains("settings"))
+//		_input["settings"] = Json::object();
+	Json const& settings = _input["settings"];
```
cameel: So there's no `get()` with a default?

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1513,24 +1514,24 @@ string StandardCompiler::compile(string const& _input) noexcept
...
 		if (info.bytecodeOffset)
-			fun["entryPoint"] = Json::UInt64(*info.bytecodeOffset);
+			fun["entryPoint"] = Json::number_unsigned_t(*info.bytecodeOffset);
 		else
-			fun["entryPoint"] = Json::nullValue;
-		fun["parameterSlots"] = Json::UInt64(info.params);
-		fun["returnSlots"] = Json::UInt64(info.returns);
+			fun["entryPoint"] = Json{};
+		fun["parameterSlots"] = Json::number_unsigned_t(info.params);
+		fun["returnSlots"] = Json::number_unsigned_t(info.returns);
```
cameel, line 1580: These also needs some (static) asserts about sizes.

libsolutil/JSON.cpp
```diff
@@ -33,8 +33,8 @@
 using namespace std;
 
 static_assert(
-	(JSONCPP_VERSION_MAJOR == 1) && (JSONCPP_VERSION_MINOR == 9) && (JSONCPP_VERSION_PATCH == 3),
-	"Unexpected jsoncpp version: " JSONCPP_VERSION_STRING ". Expecting 1.9.3."
+	(NLOHMANN_JSON_VERSION_MAJOR == 3) && (NLOHMANN_JSON_VERSION_MINOR == 10) && (NLOHMANN_JSON_VERSION_PATCH == 2),
+	"Unexpected nlohmann-json version. Expecting 3.10.2."
 );
```
cameel: Do we actually need to enforce one specific version? Why not allow a range?

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,53 @@ void removeNullMembersHelper(Json::Value& _json)
...
 
-bool jsonParseStrict(string const& _input, Json::Value& _json, string* _errs /* = nullptr */)
+bool jsonParseStrict(string const& _input, Json& _json, string* _errs /* = nullptr */)
 {
-	static StrictModeCharReaderBuilder readerBuilder;
-	return parse(readerBuilder, _input, _json, _errs);
+	try
+	{
+		_json = Json::parse(_input);
+		return true;
```
cameel, line 100: Shouldn't we also set `_errs` to an empty string if there are no errors?

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,53 @@ void removeNullMembersHelper(Json::Value& _json)
...
+	// NOTE: -1 here means no new lines (it is also the default setting)
+	return _input.dump(
+		/* indent */ (_format.format == JsonFormat::Pretty) ? static_cast<int>(_format.indent) : -1,
+		/* indent_char */ ' ',
+		/* ensure_ascii */ true
+	);
 }
 
-bool jsonParseStrict(string const& _input, Json::Value& _json, string* _errs /* = nullptr */)
+bool jsonParseStrict(string const& _input, Json& _json, string* _errs /* = nullptr */)
```
cameel, line 94: Not necessarily in this PR but I think it would be better if this was returning either `optional<Json>` or maybe `variant<Json, string>`.

libsolidity/ast/AST.h
axic: The first commit started with reusing @aarlt's change. That basically includes `libsolutil/JSON.h` to implicitly include the nlohmann header.

libevmasm/Assembly.h
axic: Just commented a relevant piece here https://github.com/ethereum/solidity/pull/11967#discussion_r715829063

libsolidity/ast/ASTJsonConverter.cpp
axic: Yes, the commits are kind of organically developed, since I added those tests we can decide what style we want to use.

libsolidity/ast/ASTJsonImporter.cpp
axic: Yes, I found it after doing some of these and fixed most places, but probably missed some.

test/libsolidity/StandardCompiler.cpp
axic: It supposed to be used in one function, where it was messy doing it otherwise.

libsolidity/ast/AST.h
cameel: But I mean that `AST.h` does not seem to have anything JSON related (at least a quick search does not find anything named "json"). So I guess it's either not used at all or used indirectly by other files that include `AST.h`. I think we should have those files include `libsolutil/JSON.h` directly.

libsolidity/ast/ASTJsonConverter.cpp
cameel: I think `Json{}` or `{}` is fine. `nullptr` would look a bit like `null` but on the other hand I can't not see it as a pointer :) It's just interesting that the lib accepts it as `null`.

libsolidity/ast/ASTJsonImporter.cpp
cameel: I've seen quite a few, but for now pointed out only some of them. I can point out the rest in the second pass.

test/libsolidity/StandardCompiler.cpp
cameel: ok

libsolutil/JSON.cpp
axic: I thought about that. I kind of think there's still a second cleanup PR we could based on the learnings here, just to fix/improve things in general. This change is not specific to nlohmann-json.

libsolutil/JSON.cpp
axic: Not sure, just kept the same logic.

libsolidity/ast/ASTJsonConverter.cpp
axic: I think we actually want to review this whole mess with jsoncpp -- decide on a single number type and have assertions. Probably need a helper for that.

Then it is simple to keep using that with nlohmann-json.

libsolidity/interface/StandardCompiler.cpp
axic: Yes, that's one of the questions how to do that nicely.

axic:
@cameel any idea why the external project doesn't work on emscripten and arch?

cameel:
> @cameel any idea why the external project doesn't work on emscripten and arch?

It actually seems to fail everywhere (except for Windows for some reason). In case of emscripten it seems to be because it needs `emcmake` in cmake command. Both ranges and jsoncpp set `CMAKE_COMMAND` for that reason.

I don't know why other platforms fail but it might have something to do with you setting `CMAKE_ARGS` simply to `true`. These other libs have an extensive set of parameters for it.

By the way, maybe it would be better to use `FetchContent` instead as in #11273?

libsolutil/JSON.cpp
cameel: Yeah. Ok, leaving this comment open so that we can still find it when doing that second PR.

libsolidity/interface/StandardCompiler.cpp
cameel: We could just create a helper function for that.

axic:
Unrelated, but we also have this warning:
```
[  1%] No update step for 'range-v3-project'
[  1%] Built target evmc_loader
[  2%] No patch step for 'range-v3-project'
[  2%] Performing configure step for 'range-v3-project'
-- The CXX compiler identification is AppleClang 12.0.5.12050022
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning at cmake/ranges_env.cmake:77 (message):
  [range-v3 warning]: unknown build type, defaults to release!
  CMakeLists.txt:50 (include)


-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE) 
-- Found Git: /usr/bin/git (found version "2.30.1 (Apple Git-130)") 
```

Notice `[range-v3 warning]: unknown build type, defaults to release!`.

axic:
> By the way, maybe it would be better to use FetchContent instead as in #11273?

@cameel I actually did `FetchContent` in 5510abd, but it had the same problem.

I'd refrain from rebasing this before the problem is solved though.

cmake/nlohmann-json.cmake
ekpyron: **Please** don't pull in crazy package managers in the build system, things are bad enough as they are :-).
We should just use submodules for this stuff and be done with it. I know there have been complications with using submodules in the past, but IIUC that was due to developing both ends in parallel, which is much more of a hassle.
Pulling in fixed versions of external code that are occasionally updated with submodules is zero hassle.

ekpyron: IMHO the only viable alternative to git subtrees or git submodules (of which I'd prefer the latter for this), is just don't do anything at all, document the dependency, have users install it and cmake find it. Everything else is enormously bad and that goes *especially* for "dependency management systems". That's the worst you can do to anyone who ever wants to package your package and *that's* who one should be concerned about primarily - regular users rarely build anyways.

ekpyron: Just to rant a bit more about this :-D: https://github.com/ethereum/solc-bin/pull/101 is exactly what you get from dependency managers  - it's **so** much more convenient for everyone to just use a random dependency manager (which of course doesn't have like an API and interface of its own you first need to know about, if you ever want to override anything), that it's *totally* worth it to then require people to track down opaque dependency chains with 10 indirections without being able to influence any of it. That's *really* superior to the *extreme* hassle of an occasional ``git submodule update`` :-).

But yeah, sorry :-D. I really think the seeming convenience of dependency managers is an absolute fallacy - it exactly prevents you from hitting problems in cases in which you definitely *should* hit problems as in that you should update your stuff to work with proper up-to-date external dependencies. Even submodules are actually bad in that regard, but for some small header-only dependencies they may be excusable...

cameel: I think we should continue this discussion in the context of #8860. I quoted the comments there and will post my response in that issue in a moment.

libsolidity/interface/StandardCompiler.cpp
axic: There's actually `.value(key, default)` instead. Used that now in most places.

libsolutil/JSON.cpp
axic: Well we download a specific file too, so this is just a safety check. I think it is fine.

test/libsolidity/StandardCompiler.cpp
axic: Unfortunately `ranges::views` is not compatible with `items()`.

test/libsolidity/SemanticTest.cpp
axic: Done.

libsolutil/CMakeLists.txt
axic: Good question.

libsolidity/interface/StandardCompiler.cpp
axic: Done.

libsolidity/interface/ABI.cpp
```diff
@@ -96,79 +98,79 @@ Json::Value ABI::generate(ContractDefinition const& _contractDef)
...
+			errorJson["inputs"].emplace_back(
 				formatType(p->name(), *type, *p->annotation().type, false)
 			);
 		}
 		abi.emplace(move(errorJson));
 	}
 
-	Json::Value abiJson{Json::arrayValue};
+	Json abiJson{Json::array()};
 	for (auto& f: abi)
```
axic, line 143: Is the point of this to enforce the ordering?

libsolidity/interface/CompilerStack.cpp
axic: There seems to be a general mess regarding (large) numbers, both currently and with nlohmann.

libsolidity/interface/CompilerStack.cpp
axic: I think this meant to be kept as a TODO to double check stuff. Will replace it with an actual TODO.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -903,15 +898,18 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
...
 
-	Json::Value metadataSettings = settings.get("metadata", Json::Value());
+	Json metadataSettings = settings.value("metadata", Json::object());
 
 	if (auto result = checkMetadataKeys(metadataSettings))
 		return *result;
 
-	ret.metadataLiteralSources = metadataSettings.get("useLiteralContent", Json::Value(false)).asBool();
-	if (metadataSettings.isMember("bytecodeHash"))
+	ret.metadataLiteralSources =
```
axic, line 918: Could use `metadataSettings.value("useLiteralContent", false)`.

libsolidity/interface/StandardCompiler.cpp
axic: Are slots actually large values?

libsolutil/JSON.cpp
```diff
@@ -43,53 +43,9 @@ namespace solidity::util
...
-bool parse(Json::CharReaderBuilder& _builder, string const& _input, Json::Value& _json, string* _errs)
-{
-	unique_ptr<Json::CharReader> reader(_builder.newCharReader());
-	return reader->parse(_input.c_str(), _input.c_str() + _input.length(), &_json, _errs);
-}
-
+#if 0
 /// Takes a JSON value (@ _json) and removes all its members with value 'null' recursively.
-void removeNullMembersHelper(Json::Value& _json)
+void removeNullMembersHelper(Json& _json)
```
axic, line 48: TODO: need to check if this needed, and if yes, implement it.

libsolutil/JSON.h
axic: I don't think so, grandfathered in from @aarlt.

test/libsolidity/StandardCompiler.cpp
```diff
@@ -41,6 +41,14 @@ namespace solidity::frontend::test
 namespace
 {
 
+set<string> getMemberNames(Json const& _json)
```
axic, line 44: Ideally instead of this we would use range-v3 to get the keys from the `items()` iterator, but haven't found a nice way yet.

test/libsolidity/StandardCompiler.cpp
axic: I don't remember 😅

leonardoalt:
I guess this needs to be taken over?

axic:
> I guess this needs to be taken over?

I think this ran into some bug in one of the JSON output generators.

But also let's wait for https://github.com/nlohmann/json/issues/3130 to be fixed. (Well was just merged now.)

test/libsolutil/JSON.cpp
axic: This seems to be properly handled as failure, could add the `error` message check here though.

falbrechtskirchinger:
> > I guess this needs to be taken over?
> 
> I think this ran into some bug in one of the JSON output generators.
> 
> But also let's wait for [nlohmann/json#3130](https://github.com/nlohmann/json/issues/3130) to be fixed. (Well was just merged now.)

Let me know if there're any more issues. I've not explicitly tested with range-v3 but `std::ranges`. I expect 3.11.0 to be released soon (although it's not up to me).

libsolutil/JSON.h
Marenz: Implying that all headers should use the full notation?

cameel: Yeah, pretty much. I don't think we want to pollute the global namespace this way, do we?

libsolutil/CMakeLists.txt
falbrechtskirchinger: FYI, there's a forward declaration header (`json_fwd.hpp`) that can help with compile times. Since v3.11.x, there's also the added benefit of getting linker errors instead of incorrect behavior or crashes at runtime when inappropriately mixing different library versions or configurations.

Marenz:
I pushed a rebase and will now look & go through the comments here

libsolutil/JSON.h
Marenz: I suspect it will be very verbose to always write `nlohmann::json` instead of `Json`. It's also not exactly rolling of the tongue. 
Also, doesn't this conflict with the desire to abstract the library specific parts as mentioned here https://github.com/ethereum/solidity/pull/11967#discussion_r715772723

Marenz: And I guess it also means we'd need to use `using Json = nlohmann::json;` in all cpp files that use it then?

axic:
Rebased and upgraded to nlohmann-json 3.11.2 which has proper ranges-v3 support.

libsolidity/ast/AST.h
axic: Removed, compiles.

libsolidity/ast/ASTJsonImporter.cpp
axic: Added a `static_assert`.

libsolutil/JSON.cpp
```diff
@@ -104,46 +60,54 @@ void removeNullMembersHelper(Json::Value& _json)
...
-	StreamWriterBuilder writerBuilder(settings);
-	string result = print(_input, writerBuilder);
-	if (_format.format == JsonFormat::Pretty)
-		boost::replace_all(result, " \n", "\n");
-	return result;
+	// NOTE: -1 here means no new lines (it is also the default setting)
+	return _input.dump(
+		/* indent */ (_format.format == JsonFormat::Pretty) ? static_cast<int>(_format.indent) : -1,
+		/* indent_char */ ' ',
+		/* ensure_ascii */ true
```
aarlt, line 90: should be `false` here - we have some tests using unicode

axic: This means unicode stuff is escaped, which is the default behaviour of JSON expected by many implementations.

aarlt: I just noticed that some tests are not working because of that. But yeah, we could also rewrite the tests - but maybe we should not.

ekpyron:
Closing this as "postponed" for now. We'll keep the branch around as a basis for when we may pick this up again.

PR #11968, opened 2021-09-15
Refactor source comment parsing.

chriseth:


libyul/AsmParser.cpp
```diff
@@ -134,84 +134,97 @@ void Parser::fetchSourceLocationFromComment()
...
-				!boost::algorithm::ends_with(srcTagArgsMatch[4].str(), "\"") ||
-				boost::algorithm::ends_with(srcTagArgsMatch[4].str(), "\\\"")
-			))
-			{
-				m_errorReporter.syntaxError(
-					1544_error,
-					commentLocation,
-					"Invalid code snippet in source location mapping. Quote is not terminated."
-				);
-				return;
```
chriseth, line 181: Ah, this is a small difference: Previously, this fully aborted the parsing, while the other cases did not abort. I think it's fine to abort in all cases.

chriseth: Ok, one could also argue in the other direction and not abort so that we at least get valid data for the other tags. I'll change it.

libyul/AsmParser.cpp
```diff
@@ -119,99 +119,108 @@ langutil::Token Parser::advance()
...
+		m_errorReporter.syntaxError(
+			8387_error,
+			_commentLocation,
+			"Invalid values in source location mapping. Could not parse location specification."
+		);
+		return nullopt;
+	}
+
+	solAssert(match.size() == 5, "");
+	_arguments = _arguments.substr(static_cast<size_t>(match.position() + match.length()));
```
cameel: `_arguments` is really a confusing name for it when you cut out the actual arguments from it and keep the name :) It's more like `remainingComment`. Or maybe this:
```suggestion
	string_view tail = _argumentsAndTail.substr(static_cast<size_t>(match.position() + match.length()));
```

libyul/AsmParser.cpp
```diff
@@ -119,99 +119,108 @@ langutil::Token Parser::advance()
...
+			"Invalid code snippet in source location mapping. Quote is not terminated."
+		);
+		return {{_arguments, SourceLocation{}}};
+	}
+
+	optional<int> const sourceIndex = toInt(match[1].str());
+	optional<int> const start = toInt(match[2].str());
+	optional<int> const end = toInt(match[3].str());
+
+	m_debugDataOverride = DebugData::create();
```
cameel: This looks like a leftover from the old function. You seem to be returning the location now instead of putting it in `m_debugDataOverride` so I think this should be removed.

libyul/AsmParser.h
```diff
@@ -35,6 +35,7 @@
 #include <memory>
 #include <variant>
 #include <vector>
+#include <string>
```
cameel: The code you added does not even use `string`.
```suggestion
#include <string_view>
```

cameel (commented):
The new code seems correct so just a few remarks.

libyul/AsmParser.cpp
cameel: You code seems equivalent. It does not even break this test: https://github.com/ethereum/solidity/blob/febbd8deb31e3dafafc72e2e4704cd8b804cc46c/test/libyul/Parser.cpp#L761-L776

You technically do not abort the loop but you still return an empty source location and an empty string view so the loop terminates.

libyul/AsmParser.cpp
chriseth: I'll keep `_arguments` but introduce `tail`.

Marenz (approved):

cameel (approved):

PR #11971, opened 2021-09-15
Parse ast-id

chriseth:


libyul/AST.h
```diff
@@ -37,12 +38,22 @@ using Type = YulString;
...
+		langutil::SourceLocation _location = {},
+		std::optional<int64_t> _astID = {}
+	)
 	{
-		return std::make_shared<DebugData const>(_location);
+		return std::make_shared<DebugData const>(std::move(_location), std::move(_astID));
 	}
+
+	langutil::SourceLocation location;
+	/// ID in the source AST.
```
cameel: I'd emphasize that it's not the Yul source.
```suggestion
	/// ID in the Solidity source AST.
```

libyul/AsmParser.h
```diff
@@ -114,6 +114,12 @@ class Parser: public langutil::ParserBase
 		langutil::SourceLocation const& _commentLocation
 	);
 
+	std::optional<std::pair<std::string_view, std::optional<int>>>
+	parseASTIDComment(
```
cameel: Why not this? It's not that long.
```suggestion
	std::optional<std::pair<std::string_view, std::optional<int>>> parseASTIDComment(
```

libyul/AsmParser.cpp
```diff
@@ -222,6 +231,37 @@ optional<pair<string_view, SourceLocation>> Parser::parseSrcComment(
...
+		R"~~(^(\d+))~~",
+		regex_constants::ECMAScript | regex_constants::optimize
+	);
+	match_results<string_view::const_iterator> match;
+	optional<int> astID;
+	bool matched = regex_search(_arguments.cbegin(), _arguments.cend(), match, argRegex);
+	if (matched)
+	{
+		solAssert(match.size() == 2, "");
+		_arguments = _arguments.substr(static_cast<size_t>(match.position() + match.length()));
```
cameel:
```suggestion
		tail = _arguments.substr(static_cast<size_t>(match.position() + match.length()));
```

test/libyul/Parser.cpp
```diff
@@ -804,6 +804,67 @@ BOOST_AUTO_TEST_CASE(customSourceLocations_with_code_snippets_with_nested_locati
...
+			function f(x) -> y {}
+			mstore(1, 2)
+		}
+	)";
+	EVMDialectTyped const& dialect = EVMDialectTyped::instance(EVMVersion{});
+	shared_ptr<Block> result = parse(sourceText, dialect, reporter);
+	BOOST_REQUIRE(!!result);
+	BOOST_CHECK(result->debugData->astID == int64_t(7));
+	auto const& funDef = get<FunctionDefinition>(result->statements.at(0));
+	BOOST_CHECK(funDef.debugData->astID == int64_t(2));
```
cameel, line 824: Oh, looks like there are no dedicated `int64_t` literals in C++? Apparently the standard way to get such a literal is `INT64_C(7)`. But then I guess that's not much better than a cast.

libyul/AsmParser.cpp
```diff
@@ -222,6 +231,37 @@ optional<pair<string_view, SourceLocation>> Parser::parseSrcComment(
 	return {{tail, SourceLocation{}}};
 }
 
+optional<pair<string_view, optional<int>>> Parser::parseASTIDComment(
+	string_view _arguments,
+	langutil::SourceLocation const& _commentLocation
+)
+{
+	static regex const argRegex = regex(
+		R"~~(^(\d+))~~",
```
cameel: For consistency with `@src` I think this should require whitespace after a number:
```suggestion
		R"~~(^(\d+)(?:\s+|$))~~",
```

Also, we should have a test for this case.

test/libyul/Parser.cpp
```diff
@@ -804,6 +804,67 @@ BOOST_AUTO_TEST_CASE(customSourceLocations_with_code_snippets_with_nested_locati
...
+	BOOST_CHECK(funDef.parameters.at(0).debugData->astID == nullopt);
+	BOOST_CHECK(debugDataOf(result->statements.at(1))->astID == nullopt);
+}
+
+BOOST_AUTO_TEST_CASE(astid_invalid)
+{
+	ErrorList errorList;
+	ErrorReporter reporter(errorList);
+	auto const sourceText = R"(
+		/// @src -1:-1:-1 @ast-id abc @src 1:1:1
```
cameel, line 870: I'd add a test with more than one `@ast-id` in a single comment. The last value still wins in that case, right?

libyul/AsmParser.cpp
cameel: Also, we should have a test case with an ID larger than `int64_t`.

cameel (commented):
Looks fine. I see no big problems. Just the whitespace handling is a bit inconsistent with `@src` and I'd add tests for a few more corner cases.

libyul/AST.h
chriseth: I mean it doesn't have to be solidity...

libyul/AsmParser.h
```diff
@@ -108,8 +108,12 @@ class Parser: public langutil::ParserBase
...
 	void fetchDebugDataFromComment();
 
-	std::optional<std::pair<std::string_view, langutil::SourceLocation>>
-	parseSrcComment(
+	std::optional<std::pair<std::string_view, langutil::SourceLocation>>parseSrcComment(
+		std::string_view _arguments,
+		langutil::SourceLocation const& _commentLocation
+	);
+
+	std::optional<std::pair<std::string_view, std::optional<int>>>parseASTIDComment(
```
cameel:
```suggestion
	std::optional<std::pair<std::string_view, langutil::SourceLocation>> parseSrcComment(
		std::string_view _arguments,
		langutil::SourceLocation const& _commentLocation
	);

	std::optional<std::pair<std::string_view, std::optional<int>>> parseASTIDComment(
```

cameel (dismissed):

cameel (dismissed):

libyul/AsmParser.h
chriseth: argh :)

PR #11973, opened 2021-09-15
Fix command-line test filtering via script arguments

cameel:
Apparently #11933 broke support for specifying test name pattern in `cmdlineTests.sh`. This went undetected because we don't really use that feature in CI.

Originally patterns worked because we evaluated the whole argument list as the range of the `for` loop in side test dir. Now that we process arguments one by one, `selected_tests+=("$1")` just treats the pattern as a full test name specified verbatim.

The PR also adds a warning when one or more of the specified patterns match no tests.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -37,9 +37,11 @@ source "${REPO_ROOT}/scripts/common.sh"
 # shellcheck source=scripts/common_cmdline.sh
 source "${REPO_ROOT}/scripts/common_cmdline.sh"
 
+pushd "${REPO_ROOT}/test/cmdlineTests" > /dev/null
 autoupdate=false
 no_smt=false
 declare -a selected_tests
+declare -a patterns_with_no_matches
 while [[ $# -gt 0 ]]
 do
     case "$1" in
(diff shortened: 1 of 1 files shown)


test/cmdlineTests.sh
```diff
@@ -52,12 +54,26 @@ do
             shift
             ;;
         *)
-            selected_tests+=("$1")
+            matching_tests=$(find . -mindepth 1 -maxdepth 1 -type d -name "$1" | cut --characters 3- | sort)
```
Marenz, line 57: I might be wrong, but I thought this kind of `find` call for `-name` does only an exact match. Is that intended here?

cameel: It's intended. It does not do a regex match but still supports wildcards.

Actually the argument to `-name` should always be quoted because otherwise shell will expand `*` and `?`. Though it usually kinda works even without quoting as long as you are only searching the current directory.

Marenz: Ah right, you can just pass a wildcard there. good then :)

Marenz (approved):

PR #11974, opened 2021-09-15
New code transform (IR codegen only).

ekpyron:
Restriction of https://github.com/ethereum/solidity/pull/11493 to Yul IR code gen only.


libyul/AssemblyStack.h
```diff
@@ -124,9 +124,9 @@ class AssemblyStack: public langutil::CharStreamProvider
 private:
 	bool analyzeParsed();
 	bool analyzeParsed(yul::Object& _object);
-
+public:
```
chriseth: Is this needed or not? If yes, where is it needed?

test/libyul/evmCodeTransform/stackReuse/reuse_too_deep_slot.yul
```diff
@@ -33,78 +33,108 @@
...
-// POP
-// POP
-// POP
-// POP
-// POP
-// POP
+//     /* "":15:16   */
+//   0x07
+//     /* "":94:121   */
+//   verbatimbytecode_6042
```
axic, line 39: Why is this using verbatim and not `push1 0x42`?

test/libyul/evmCodeTransform/stackReuse/reuse_too_deep_slot.yul
```diff
@@ -33,78 +33,108 @@
...
+//   swap11
+//   swap10
+//   swap7
+//   swap10
+//   swap9
+//   swap8
+//   swap9
+//   sstore
+//     /* "":370:396   */
+//   verbatimbytecode_6043
```
axic, line 68: Same here.

test/libyul/objectCompiler/nested_optimizer.yul
```diff
@@ -20,22 +20,26 @@ object "a" {
...
 //   0x00
-//   dup1
 //     /* "source":35:50   */
+//   dup1
 //   calldataload
 //     /* "source":107:127   */
 //   sstore
+//     /* "source":20:131   */
+//   stop
 // stop
```
axic, line 30: Why are there two stops here?

test/libyul/evmCodeTransform/stackReuse/last_use_in_nested_block.yul
```diff
@@ -2,9 +2,11 @@
...
-// PUSH1 0x1
-// POP
+//     /* "":11:12   */
+//   0x00
+//     /* "":15:21   */
+//   pop
+//     /* "":33:34   */
+//   0x01
+//     /* "":0:36   */
+//   stop
```
axic, line 12: Hmm, is this a new feature that it won't try to clean up the stack? :)

test/libyul/evmCodeTransform/stackReuse/for_1.yul
```diff
@@ -2,18 +2,15 @@
 // ====
 // stackOptimization: true
 // ----
-// PUSH1 0x0
```
chriseth, line 5: Why is the format changed here?

chriseth: Is this test still valid at all? It was testing the old stack optimizer, right?

test/libyul/objectCompiler/long_object_name.yul
```diff
@@ -16,10 +16,14 @@ object "t" {
 //   0x00
 //     /* "source":23:147   */
 //   sstore
+//     /* "source":19:150   */
+//   stop
 // stop
 //
 // sub_0: assembly {
+//         /* "source":272:274   */
+//       stop
```
chriseth, line 25: Maybe we shouldn't always put a `stop` in the code?

chriseth (commented):
looks good apart from the comments!

libyul/AssemblyStack.h
ekpyron: I commented on it in the other PR - it's used for testing. The code transform tests used to also invoke assembling and disassembling which means that they didn't solely test what they were meant to test and they were extremely unreadable (evmasm is much easier on the eye than disassembled opcodes - especially tags are extreme effort to decipher otherwise).

test/libyul/objectCompiler/long_object_name.yul
ekpyron: I think we should - it doesn't really hurt and it's required in most cases that aren't merely a single block of straight control flow anyways.

test/libyul/objectCompiler/nested_optimizer.yul
ekpyron: The new code transform generates a stop at the end of the main block (instead of doing weirder things like jumping over functions), that's the new ``stop``. Not entirely sure where the one that was there before was coming from.

test/libyul/evmCodeTransform/stackReuse/reuse_too_deep_slot.yul
ekpyron: Because that's what's in the code?

test/libyul/evmCodeTransform/stackReuse/for_1.yul
ekpyron: It's code transform tests - I rather don't know why we assembled and disassembled stuff before yielding cryptic bytecodes instead of just actually testing what the code transform does, i.e. output evmasm.
And yes, this used to test the old stack optimizer - we could add a setting for still testing that instead, especially if we don't switch to the new code transform everywhere right away. But the idea was that it doesn't hurt to use the same tests for the new code transform.

test/libyul/evmCodeTransform/stackReuse/last_use_in_nested_block.yul
ekpyron: Rather a fluke of these weird test cases. Usually it will clean always clean up the stack (actually that's the first thing I'd intend to improve in this stuff - the new code transform has silly unneeded ``pop``s before reverting, etc. - might actually be a good part of the slight increase in code size).

Here it's just that the last operation in the block wants the ``1`` on stack, but doesn't consume it.  The exit layout of the outermost block is empty, but there's no need to actually produce it, i.e. actually pop.

test/libyul/evmCodeTransform/stackReuse/reuse_too_deep_slot.yul
axic: But why is it `verbatimbytecode` and not proper push instruction?

ekpyron: Why should it be a push instruction - the Yul code has a verbatim instruction :-)! Are you wondering since the old test setup actually assembled and disassembled instead? That has nothing to do with the code transform - in evmasm it's still verbatim.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -170,15 +170,35 @@ void OptimizedEVMCodeTransform::operator()(CFG::Assignment const& _assignment)
...
 	m_assembly(_assembly),
 	m_builtinContext(_builtinContext),
-	m_useNamedLabelsForFunctions(_useNamedLabelsForFunctions),
 	m_dfg(_dfg),
-	m_stackLayout(_stackLayout)
+	m_stackLayout(_stackLayout),
+	m_functionLabels([&](){
+		map<CFG::FunctionInfo const*, AbstractAssembly::LabelID> functionLabels;
+		set<YulString> assignedFunctionNames;
+		for(Scope::Function const* function: m_dfg.functions)
```
stackenbotten, line 184: Coding style error

stackenbotten, line 184: Coding style error

stackenbotten, line 184: Coding style error

libyul/AssemblyStack.h
ekpyron: Indeed also works via object compiler - fixed.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -170,15 +170,35 @@ void OptimizedEVMCodeTransform::operator()(CFG::Assignment const& _assignment)
...
+			if (_useNamedLabelsForFunctions == UseNamedLabels::YesAndForceUnique)
+				yulAssert(!nameAlreadySeen);
+			bool useNamedLabel = _useNamedLabelsForFunctions != UseNamedLabels::Never && !nameAlreadySeen;
+			functionLabels[&functionInfo] = useNamedLabel ?
+				m_assembly.namedLabel(
+					function->name.str(),
+					function->arguments.size(),
+					function->returns.size(),
+					functionInfo.debugData ? functionInfo.debugData->astID : nullopt
+					) :
```
chriseth:
```suggestion
				) :
```

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -192,14 +212,9 @@ void OptimizedEVMCodeTransform::assertLayoutCompatibility(Stack const& _currentS
...
-		m_functionLabels[&functionInfo] = m_useNamedLabelsForFunctions ? m_assembly.namedLabel(
-			functionInfo.function.name.str(),
-			functionInfo.function.arguments.size(),
-			functionInfo.function.returns.size(),
-			{}
-		) : m_assembly.newLabelId();
-	return m_functionLabels[&functionInfo];
+	if (AbstractAssembly::LabelID const* label = util::valueOrNullptr(m_functionLabels, &functionInfo))
+		return *label;
+	yulAssert(false, "");
```
chriseth: Why not just use `.at(...)`?

test/libsolidity/semanticTests/externalContracts/deposit_contract.sol
```diff
@@ -178,35 +178,35 @@ contract DepositContract is IDepositContract, ERC165 {
 // compileViaYul: also
 // ----
 // constructor()
-// gas irOptimized: 1657475
+// gas irOptimized: 1558013
 // gas legacy: 2580394
-// gas legacyOptimized: 1775403
+// gas legacyOptimized: 1776483
```
chriseth: Weird, why did this change?

chriseth (commented):
Only cosmetics left.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: Because that asserts in the STL, while this asserts where we know what happens...
Fine with using ``.at``, though, but are we fine with relying on STL exceptions and assertions these days in general :-)?

test/libsolidity/semanticTests/externalContracts/deposit_contract.sol
ekpyron: Good question, I hadn't noticed...

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: But well, it shouldn't ever actually fail anyways, so it doesn't matter.

chriseth: The only difference is that the exception is not as nicely formatted, but most of the time, the exception information does not help users anyway and we can easily hook it into a debugger. I would say: If it is more readable, prefer stl exceptions.

test/libsolidity/semanticTests/externalContracts/deposit_contract.sol
ekpyron: Just pushed it with the old value - if the tests pass now as well, this confirms that there's something broken with those...

chriseth:
Can you rebase the cosmetics?

ekpyron:
> Can you rebase the cosmetics?

Rebased

chriseth (approved):

PR #11975, opened 2021-09-15
Function debug info

chriseth:


chriseth:
I fear that the "gas estimation" test depends on whether or not we are in release mode or not :(

chriseth:
I think it's not feasibly to test internal gas costs, so I'll remove the test again. The FunctionDebugInfo in standard json output already shows that this feature is working in principle, so I think this is enough.

test/cmdlineTests/yul_source_locations_in_asm/output.json
```diff
@@ -1,10 +1,10 @@
 {"contracts":{"C":{"C":{"evm":{"assembly":"    /* \"C\":79:428  contract C... */
   mstore(0x40, 0xa0)
-  jumpi(tag_1, iszero(callvalue))
+  jumpi(tag_2, iszero(callvalue))
```
hrkrshnn: Why should tags in the tests change if only debug info is changed?

libyul/backends/evm/EVMCodeTransform.cpp
```diff
@@ -572,17 +576,30 @@ void CodeTransform::operator()(Block const& _block)
...
+	Scope::Function& scopeFunction = std::get<Scope::Function>(m_scope->identifiers.at(_function.name));
+	yulAssert(!m_context->functionEntryIDs.count(&scopeFunction), "");
+
+	optional<size_t> astID;
+	if (_function.debugData)
+		astID = _function.debugData->astID;
+
+	m_context->functionEntryIDs[&scopeFunction] =
+		m_useNamedLabelsForFunctions ?
+		m_assembly.namedLabel(
```
hrkrshnn, line 590: Is there a test that showcases this change?

libyul/backends/evm/EVMCodeTransform.cpp
```diff
@@ -563,6 +563,10 @@ void CodeTransform::operator()(Block const& _block)
 	Scope* originalScope = m_scope;
 	m_scope = m_info.scopes.at(&_block).get();
 
+	for (auto const& statement: _block.statements)
+		if (auto function = get_if<FunctionDefinition>(&statement))
+			createFunctionEntryID(*function);
```
hrkrshnn, line 568: Should the order of function definitions in the block matter? Keeping the same order as the source would be nice. This visit would change it for functions defined inside a nested block.

chriseth: Previously, it depended on the order in which functions were called, now it depends on the order in which they are defined (in a kind of breadth-first-search), I think this is better.

test/cmdlineTests/yul_source_locations_in_asm/output.json
chriseth: I think the reason is because of the changed visit order in https://github.com/ethereum/solidity/pull/11975#discussion_r712920688

libyul/backends/evm/EVMCodeTransform.cpp
chriseth: Yes, it allows us to create this structure: https://github.com/ethereum/solidity/pull/11975/files/a2fb5a1b470a4bd27a20da9780fd091ff03d1a0e#diff-f45c89f87fcc7cd29d5b5122cadfea72d13562131a9bd28f9f1a946265505bf2R6

libevmasm/LinkerObject.h
```diff
@@ -48,6 +48,7 @@ struct LinkerObject
 	struct FunctionDebugData
 	{
 		std::optional<size_t> bytecodeOffset;
+		std::optional<size_t> instructionIndex;
```
hrkrshnn, line 51: Wouldn't `tagIndex` be more appropriate?

libyul/backends/evm/EVMCodeTransform.cpp
hrkrshnn: Not sure if it matters, function hoister and function grouper would change the visit order in that case. Wouldn't it be nice if the order was the same? Maybe doing this in `operator()(FunctionDefinition const&...)` is more appropriate?

Also, where else are using the current order?

libevmasm/LinkerObject.h
chriseth: not sure - it's an index into the list of instructions - maybe `instructionOffset` to make it clear that it is similar to `bytecodeOffset`?

libyul/backends/evm/EVMCodeTransform.cpp
chriseth: It cannot be done in `operator()(FunctionDefinition` because we need the entry id to be available before the function can be called for the first time, that's the point of this change. I'm fine with the internal label numbers to change (note that this has zero effect on the generated bytecode) as an effect of the function hoister or grouper.

libevmasm/LinkerObject.h
hrkrshnn: > it's an index into the list of instructions

Ah, okay. I'm indifferent about `instructionOffset` v/s `instructionIndex`.

hrkrshnn (approved):

PR #11976, opened 2021-09-15
Some cleanup and testing for JSON

axic:
Found via #11967.

### DIFF
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -200,7 +200,7 @@ string Assembly::assemblyString(StringMap const& _sourceCodes) const
 
 Json::Value Assembly::createJsonValue(string _name, int _source, int _begin, int _end, string _value, string _jumpType)
 {
-	Json::Value value;
+	Json::Value value{Json::objectValue};
 	value["name"] = _name;
 	value["source"] = _source;
 	value["begin"] = _begin;
@@ -222,8 +222,9 @@ string Assembly::toStringInHex(u256 _value)
 Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) const
 {
 	Json::Value root;
+	root[".code"] = Json::arrayValue;
 
-	Json::Value& collection = root[".code"] = Json::arrayValue;
+	Json::Value& collection = root[".code"];
 	for (AssemblyItem const& i: m_items)
 	{
 		int sourceIndex = -1;
@@ -317,7 +318,8 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) 
 
 	if (!m_data.empty() || !m_subs.empty())
 	{
-		Json::Value& data = root[".data"] = Json::objectValue;
+		root[".data"] = Json::objectValue;
+		Json::Value& data = root[".data"];
 		for (auto const& i: m_data)
 			if (u256(i.first) >= m_subs.size())
 				data[toStringInHex((u256)i.first)] = toHex(i.second);
--- a/libsolidity/interface/Natspec.cpp
+++ b/libsolidity/interface/Natspec.cpp
@@ -37,7 +37,7 @@ using namespace solidity::frontend;
 
 Json::Value Natspec::userDocumentation(ContractDefinition const& _contractDef)
 {
-	Json::Value doc;
+	Json::Value doc{Json::objectValue};
 
 	doc["version"] = Json::Value(c_natspecVersion);
 	doc["kind"]    = Json::Value("user");
@@ -50,7 +50,7 @@ Json::Value Natspec::userDocumentation(ContractDefinition const& _contractDef)
 		if (!value.empty())
 		{
 			// add the constructor, only if we have any documentation to add
-			Json::Value user;
+			Json::Value user{Json::objectValue};
 			user["notice"] = Json::Value(value);
 			doc["methods"]["constructor"] = user;
 		}
@@ -90,7 +90,7 @@ Json::Value Natspec::userDocumentation(ContractDefinition const& _contractDef)
 		string value = extractDoc(error->annotation().docTags, "notice");
 		if (!value.empty())
 		{
-			Json::Value errorDoc;
+			Json::Value errorDoc{Json::objectValue};
 			errorDoc["notice"] = value;
 			doc["errors"][error->functionType(true)->externalSignature()].append(move(errorDoc));
 		}
@@ -225,7 +225,10 @@ Json::Value Natspec::extractCustomDoc(multimap<string, DocTag> const& _tags)
 	for (auto const& [tag, value]: _tags)
 		if (boost::starts_with(tag, "custom"))
 			concatenated[tag] += value.content;
-	Json::Value result;
+	// We do not want to create an object if there are no custom tags found.
+	if (concatenated.empty())
+		return Json::nullValue;
+	Json::Value result{Json::objectValue};
 	for (auto& [tag, value]: concatenated)
 		result[tag] = move(value);
 	return result;
--- a/libsolidity/interface/StandardCompiler.cpp
+++ b/libsolidity/interface/StandardCompiler.cpp
@@ -59,7 +59,7 @@ Json::Value formatError(
 	Json::Value const& _secondarySourceLocation = Json::Value()
 )
 {
-	Json::Value error = Json::objectValue;
+	Json::Value error{Json::objectValue};
 	error["type"] = _type;
 	error["component"] = _component;
 	error["severity"] = Error::formatErrorSeverityLowercase(_severity);
@@ -74,7 +74,7 @@ Json::Value formatError(
 
 Json::Value formatFatalError(string const& _type, string const& _message)
 {
-	Json::Value output = Json::objectValue;
+	Json::Value output{Json::objectValue};
 	output["errors"] = Json::arrayValue;
 	output["errors"].append(formatError(Error::Severity::Error, _type, "general", _message));
 	return output;
@@ -82,23 +82,22 @@ Json::Value formatFatalError(string const& _type, string const& _message)
 
 Json::Value formatSourceLocation(SourceLocation const* location)
 {
-	Json::Value sourceLocation;
-	if (location && location->sourceName)
-	{
-		sourceLocation["file"] = *location->sourceName;
-		sourceLocation["start"] = location->start;
-		sourceLocation["end"] = location->end;
-	}
+	if (!location || !location->sourceName)
+		return Json::nullValue;
 
+	Json::Value sourceLocation{Json::objectValue};
+	sourceLocation["file"] = *location->sourceName;
+	sourceLocation["start"] = location->start;
+	sourceLocation["end"] = location->end;
 	return sourceLocation;
 }
 
 Json::Value formatSecondarySourceLocation(SecondarySourceLocation const* _secondaryLocation)
 {
 	if (!_secondaryLocation)
-		return {};
+		return Json::nullValue;
 
-	Json::Value secondarySourceLocation = Json::arrayValue;
+	Json::Value secondarySourceLocation{Json::arrayValue};
 	for (auto const& location: _secondaryLocation->infos)
 	{
 		Json::Value msg = formatSourceLocation(&location.second);
@@ -330,7 +329,7 @@ bool isIRRequested(Json::Value const& _outputSelection)
 
 Json::Value formatLinkReferences(std::map<size_t, std::string> const& linkReferences)
 {
-	Json::Value ret(Json::objectValue);
+	Json::Value ret{Json::objectValue};
 
 	for (auto const& ref: linkReferences)
 	{
@@ -345,7 +344,7 @@ Json::Value formatLinkReferences(std::map<size_t, std::string> const& linkRefere
 		Json::Value fileObject = ret.get(file, Json::objectValue);
 		Json::Value libraryArray = fileObject.get(name, Json::arrayValue);
 
-		Json::Value entry = Json::objectValue;
+		Json::Value entry{Json::objectValue};
 		entry["start"] = Json::UInt(ref.first);
 		entry["length"] = 20;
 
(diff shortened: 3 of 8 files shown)


libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,22 +74,22 @@ Json::Value formatError(
...
 {
-	Json::Value sourceLocation;
-	if (location && location->sourceName)
-	{
-		sourceLocation["file"] = *location->sourceName;
-		sourceLocation["start"] = location->start;
-		sourceLocation["end"] = location->end;
-	}
+	if (!(location && location->sourceName))
+		return {};
```
axic: This way the function is clearer that we don't want to have any value.

axic: Also this commit should be squashed.

libsolidity/interface/Natspec.cpp
```diff
@@ -225,7 +225,7 @@ Json::Value Natspec::extractCustomDoc(multimap<string, DocTag> const& _tags)
 	for (auto const& [tag, value]: _tags)
 		if (boost::starts_with(tag, "custom"))
 			concatenated[tag] += value.content;
-	Json::Value result;
```
chriseth, line 228: This used to return `null` for the case that `_tags` is empty, right? Does this change anything in the output ("remove null values")?

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,17 +74,19 @@ Json::Value formatError(
...
-	Json::Value output = Json::objectValue;
+	Json::Value output{Json::objectValue};
 	output["errors"] = Json::arrayValue;
 	output["errors"].append(formatError(Error::Severity::Error, _type, "general", _message));
 	return output;
 }
 
 Json::Value formatSourceLocation(SourceLocation const* location)
 {
-	Json::Value sourceLocation;
```
chriseth, line 85: Same here, this used to return null.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,17 +74,19 @@ Json::Value formatError(
...
+	Json::Value sourceLocation{Json::objectValue};
+
 	if (location && location->sourceName)
 	{
+		sourceLocation = Json::objectValue;
```
chriseth: redundant

libsolidity/interface/StandardCompiler.cpp
chriseth: Can you make it return null explicitly?

axic: This last commit does that?

libsolidity/interface/Natspec.cpp
axic: Yes, I was wondering about this myself, whether I should address it or not. We can skip this or add an explicit condition for aborting early.

libsolidity/interface/StandardCompiler.cpp
axic: You are looking at the previous commit.

libsolidity/interface/StandardCompiler.cpp
axic: Do you mean something like `Json::nullValue`? That can be done, I just copied the logic from `formatSecondarySourceLocation`, but will update both then.

chriseth: For me, it was not obvious what the default value of Json::Value is.

libsolidity/interface/Natspec.cpp
chriseth: Does it change anything down the line?

cameel: I don't think so. Looking at this file, everything that that uses this function either uses `.empty()` to skip it the value (which works the same for `null` and for arrays/objects) or just adds more keys ensuring it's not empty.

libsolutil/JSON.h
```diff
@@ -38,8 +38,8 @@ struct JsonFormat
 {
 	enum Format
 	{
-		Compact,
-		Pretty
+		Compact, // No unnecessary whitespace (including new lines and indentation)
+		Pretty // Nicely indented, with new lines
```
cameel:
```suggestion
		Pretty, // Nicely indented, with new lines
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,22 +74,22 @@ Json::Value formatError(
...
-	{
-		sourceLocation["file"] = *location->sourceName;
-		sourceLocation["start"] = location->start;
-		sourceLocation["end"] = location->end;
-	}
+	if (!(location && location->sourceName))
+		return {};
 
+	Json::Value sourceLocation{Json::objectValue};
+	sourceLocation = Json::objectValue;
```
cameel: This line seems superfluous.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,22 +74,22 @@ Json::Value formatError(
...
 Json::Value formatSourceLocation(SourceLocation const* location)
 {
-	Json::Value sourceLocation;
-	if (location && location->sourceName)
-	{
-		sourceLocation["file"] = *location->sourceName;
-		sourceLocation["start"] = location->start;
-		sourceLocation["end"] = location->end;
-	}
+	if (!(location && location->sourceName))
```
cameel: I think it reads better with less nesting.
```suggestion
	if (!location || !location->sourceName)
```

cameel (dismissed):

libsolidity/interface/StandardCompiler.cpp
axic: Fixed.

libsolidity/interface/Natspec.cpp
axic: This is used by two places: `devDocumentation(ContractDefinition)` and `devDocumentation(map<string, DocTag>)`. The second used by the first. In cases there was no fields to be added, it does check for not inserting empty values. So I guess we can keep this one as it was, and look at improving this entire NatSpec generator at some later time.

libsolidity/interface/Natspec.cpp
```diff
@@ -225,7 +225,10 @@ Json::Value Natspec::extractCustomDoc(multimap<string, DocTag> const& _tags)
...
+	// We do not want to create an object if there are no custom tags found.
+	if (concatenated.empty())
```
axic, line 229: @chriseth @cameel as a response to [this](https://github.com/ethereum/solidity/pull/11976#discussion_r712325924) added this explicit code.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -74,31 +74,30 @@ Json::Value formatError(
...
 {
 	if (!_secondaryLocation)
-		return {};
+		return Json::nullValue;
 
-	Json::Value secondarySourceLocation = Json::arrayValue;
+	Json::Value secondarySourceLocation{Json::arrayValue};
 	for (auto const& location: _secondaryLocation->infos)
 	{
 		Json::Value msg = formatSourceLocation(&location.second);
```
axic, line 103: Should we have an assertion here that this returned non-empty JSON?

axic: Ah okay, I guess this can be empty as long as the message is filled out?

axic:
@chriseth @cameel how about this now?

test/libsolutil/JSON.cpp
```diff
@@ -33,6 +33,48 @@ namespace solidity::util::test
 
 BOOST_AUTO_TEST_SUITE(JsonTest, *boost::unit_test::label("nooptions"))
 
+BOOST_AUTO_TEST_CASE(json_types)
+{
+	auto check = [](Json::Value value, string const& expectation) {
+		BOOST_CHECK(jsonCompactPrint(value) == expectation);
+	};
+
+	Json::Value value;
```
axic, line 42: @chriseth added tests here to show what type corresponds to what output, `{} == Json::Value() == Json::nullValue`.

test/libsolutil/JSON.cpp
```diff
@@ -33,6 +33,48 @@ namespace solidity::util::test
...
+	check(value, "null");
+	check({}, "null");
+	check(Json::Value(), "null");
+	check(Json::nullValue, "null");
+	check(Json::objectValue, "{}");
+	check(Json::arrayValue, "[]");
+	check(Json::UInt(1), "1");
+	check(Json::UInt(-1), "4294967295");
+	check(Json::UInt64(1), "1");
+	check(Json::UInt64(-1), "18446744073709551615");
```
axic, line 60: Yes, we use this, and this is actually output which can't be properly parsed by JavaScript frameworks. It is used for gas estimation for example.

test/libsolutil/JSON.cpp
```diff
@@ -94,13 +143,28 @@ BOOST_AUTO_TEST_CASE(parse_json_strict)
 	BOOST_CHECK(json["3"]["3.1"] == "3.1");
 	BOOST_CHECK(json["3"]["3.2"] == 5);
 
-	// root element can only be object or array
+	// According to ECMA-404 object, array, number, string, true, false, null are allowed
+	// ... but JSONCPP disallows value types
```
axic, line 150: Compared it against the standards, and it seems libjsoncpp is not very compliant.

libevmasm/Assembly.cpp
```diff
@@ -200,7 +200,7 @@ string Assembly::assemblyString(StringMap const& _sourceCodes) const
 
 Json::Value Assembly::createJsonValue(string _name, int _source, int _begin, int _end, string _value, string _jumpType)
 {
-	Json::Value value;
+	Json::Value value{Json::objectValue};
```
hrkrshnn, line 203: What difference does this make?

chriseth: The default value is null

axic: @hrkrshnn the main point is explicitness. With the default value the type is determined by the first action, i.e. if it is accessed as a map, then the type becomes an object, if it has an `.append` first then it becomes an array, etc.

While in this example here it is clear what will happen, I think making it explicit (unless there is a reason not to) makes mistakes less possible. Mistakes could happen during refactoring where the definition and the first use get further apart.

libsolidity/interface/StandardCompiler.cpp
cameel: Oh, can it? Can it then be used to add a supplementary message in https://github.com/ethereum/solidity/issues/11987#issuecomment-923285509?

test/libsolutil/JSON.cpp
cameel: But that basically means that the range of integers supported in compliant JSON is less than we need it to support, isn't it? Do we want to switch to printing ints out of the supported range as strings or do we prefer to stay non-compliant in this regard?

test/libsolutil/JSON.cpp
```diff
@@ -94,15 +146,42 @@ BOOST_AUTO_TEST_CASE(parse_json_strict)
...
+	// a few control characters (\b, \f, \n, \r, \t)
+	//
+	// More lenient parsers allow hex escapes as long as they translate to a valid UTF-8 encoding.
+	//
+	// ... but JSONCPP allows any hex escapes
 	BOOST_CHECK(jsonParseStrict("[ \"\x80\xec\x80\" ]", json, &errors));
+	BOOST_CHECK(json.isArray());
 	BOOST_CHECK(json[0] == "\x80\xec\x80");
+
+	// This would be valid more lenient parsers.
```
cameel, line 181:
```suggestion
	// This would be valid for more lenient parsers.
```

test/libsolutil/JSON.cpp
```diff
@@ -94,15 +146,42 @@ BOOST_AUTO_TEST_CASE(parse_json_strict)
...
+	BOOST_CHECK(jsonParseStrict("[ \"\xF0\x9F\x98\x8A\" ]", json, &errors));
```
cameel, line 182: I'd add comments showing what these escapes are meant to decode into:
```suggestion
	BOOST_CHECK(jsonParseStrict("[ \"\x80\xec\x80\" ]", json, &errors)); // [ "ì" ]
	BOOST_CHECK(json.isArray());
	BOOST_CHECK(json[0] == "\x80\xec\x80");

	// This would be valid more lenient parsers.
	BOOST_CHECK(jsonParseStrict("[ \"\xF0\x9F\x98\x8A\" ]", json, &errors)); // [ "ð" ]
```

cameel (approved):
I have some remarks and small tweaks suggestions but overall this seems correct and already good enough to be merged.

test/libsolutil/JSON.cpp
axic: What is `"ð"` ?

The unicode is for the smiley emoji :)

axic:
@cameel I think I'll just merge this to get the ball rolling.

libsolidity/interface/StandardCompiler.cpp
axic: Since it was a pointer I think technically it can be message only, but I haven't validated it across the entire source code.

test/libsolutil/JSON.cpp
axic: [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) doesn't care about numbers:

>JSON is agnostic about the semantics of numbers. In any programming language, there can be a variety of number types of various capacities and complements, fixed or floating, binary or decimal. That can make interchange between different programming languages difficult. JSON instead offers only the representation of numbers that humans use: a sequence of digits. All programming languages know how to make sense of digit sequences even if they disagree on internal representations. That is enough to allow interchange.

However if we want to import JSON in Javascript, then it gets iffy if numbers are larger than `2^53-1`.

I would investigate this outside of this PR.

test/libsolutil/JSON.cpp
cameel: Oh :) That's what I got when I printed the string in Python. But now I realized that I should have used a bytestring and decoded it - if I do that I do get an emoji. `ð` is apparently `\xF0`.

All the more reason to have a comment that explains what it's meant to be :)

axic: The expectation has the actual emoji, and not an escape. Isn't that displayed correctly for you?

cameel: Ah, true. It's there. I focused on the first one which has escapes in both places :)

PR #11977, opened 2021-09-15
Clarify interpretation of literals.

acoglio:
This is based on discussions on Gitter.

This replaces the previous PR.

Closes #11924.


docs/yul.rst
```diff
@@ -174,18 +174,44 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
...
+- Integer constants in decimal or hexadecimal notation.
+
+- ASCII strings (e.g. ``"abc"``), which may contain hex escapes ``\xNN`` and Unicode escapes ``\uNNNN`` where ``N`` are hexadecimal digits.
+
+- Hex strings (e.g. ``hex"616263"``).
+
+In the EVM dialect of Yul, literals represent 256-bit words as follows:
+
+- Decimal or hexadecimal constants must be less than ``2**256``.
+  They represent the 256-bit word with that value as an unsigned integer.
```
chriseth:
```suggestion
  They represent the 256-bit word with that value as an unsigned integer in big-endian encoding.
```

docs/yul.rst
```diff
@@ -174,18 +174,44 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
...
+
+- An ASCII string is first viewed as a byte sequence, by viewing
+  a non-escape ASCII character as a single byte whose value is the ASCII code,
+  an escape ``\xNN`` as single byte with that value, and
+  an escape ``\uNNNN`` as the UTF-8 sequence of bytes for that code point.
+  The byte sequence must not exceed 32 bytes.
+  The byte sequence is padded with zeros on the right to reach 32 bytes in length;
+  in other words, the string is stored left-aligned.
+  The padded byte sequence represents a 256-bit word whose most significant 8 bits are the ones from the first byte,
```
chriseth, line 197: Isn't endianness only relevant for the translation between numbers and byte encoding?

docs/yul.rst
```diff
@@ -174,18 +174,44 @@ whitespace, i.e. there is no terminating ``;`` or newline required.
...
+  i.e. the bytes are interpreted in big endian form.
+  
+- A hex string is first viewed as a byte sequence, by viewing
+  each pair of contiguous hex digits as a byte.
+  The byte sequence must not exceed 32 bytes (i.e. 64 hex digits).
```
chriseth: Can we say that the rest is the same as above?

docs/yul.rst
acoglio: I pushed a commit that makes that change.

docs/yul.rst
acoglio: I pushed a commit that makes that change.

docs/yul.rst
acoglio: Given that a 256-bit word consists of 256 bits
```
b255 ... b0
```
(written left-to-right, but that doesn't matter), if we split that into 32 bytes
```
B31 = b255 ... b248
...
B0 = b7 ... b0
```
(each byte written left-to-right, but that doesn't matter), then we could store the bytes (in memory) in little endian
```
+----+-----+-----+
| B0 | ... | B31 |
+----+-----+-----+
```
or in big endian
```
+-----+-----+----+
| B31 | ... | B0 |
+-----+-----+----+
```

For instance, in the x86, words are stored in little endian, but they are independently interpreted as either unsigned or signed integers by different instructions -- similarly to EVM instructions.

The initial paragraph of https://en.wikipedia.org/wiki/Endianness describes it in terms of words as well, without reference to integers.

chriseth: In my opinion, the concept of a "most significant bit/byte" requires an interpretation as an integer - why would any byte of an (uninterpreted) byte sequence be more or less significant? But if mentioning this clarifies the concept for you, then it is probably also helpful for others, so we can leave it.

chriseth (dismissed):
Thanks a lot! I'll squash and merge.

PR #11983, opened 2021-09-16
Don't discard non-fatal errors in Yul mode in Standard JSON when followed by a fatal error

cameel:
In Yul mode `StandardCompiler` ignores any non-fatal errors collected so far when it encounters a fatal error. This results in a nonsensical message in some cases - for example when a path passed to `sources.urls` cannot be resolved by the import callback (a non-fatal error), it only prints a message saying that there were no input files (see the first commit).

### DIFF
--- a/libsolidity/interface/StandardCompiler.cpp
+++ b/libsolidity/interface/StandardCompiler.cpp
@@ -1312,16 +1312,49 @@ Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSetting
 
 Json::Value StandardCompiler::compileYul(InputsAndSettings _inputsAndSettings)
 {
-	if (_inputsAndSettings.sources.size() != 1)
-		return formatFatalError("JSONError", "Yul mode only supports exactly one input file.");
-	if (!_inputsAndSettings.smtLib2Responses.empty())
-		return formatFatalError("JSONError", "Yul mode does not support smtlib2responses.");
-	if (!_inputsAndSettings.remappings.empty())
-		return formatFatalError("JSONError", "Field \"settings.remappings\" cannot be used for Yul.");
-	if (_inputsAndSettings.revertStrings != RevertStrings::Default)
-		return formatFatalError("JSONError", "Field \"settings.debug.revertStrings\" cannot be used for Yul.");
-
 	Json::Value output = Json::objectValue;
+	output["errors"] = std::move(_inputsAndSettings.errors);
+
+	if (_inputsAndSettings.sources.size() != 1)
+	{
+		output["errors"].append(formatError(
+			Error::Severity::Error,
+			"JSONError",
+			"general",
+			"Yul mode only supports exactly one input file."
+		));
+		return output;
+	}
+	if (!_inputsAndSettings.smtLib2Responses.empty())
+	{
+		output["errors"].append(formatError(
+			Error::Severity::Error,
+			"JSONError",
+			"general",
+			"Yul mode does not support smtlib2responses."
+		));
+		return output;
+	}
+	if (!_inputsAndSettings.remappings.empty())
+	{
+		output["errors"].append(formatError(
+			Error::Severity::Error,
+			"JSONError",
+			"general",
+			"Field \"settings.remappings\" cannot be used for Yul."
+		));
+		return output;
+	}
+	if (_inputsAndSettings.revertStrings != RevertStrings::Default)
+	{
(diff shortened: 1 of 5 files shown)


libsolidity/interface/StandardCompiler.cpp
```diff
@@ -72,10 +72,12 @@ Json::Value formatError(
...
 }
 
-Json::Value formatFatalError(string const& _type, string const& _message)
+Json::Value formatFatalError(string const& _type, string const& _message, Json::Value _previousErrors = Json::arrayValue)
 {
+	solAssert(_previousErrors.isArray(), "");
+
 	Json::Value output = Json::objectValue;
-	output["errors"] = Json::arrayValue;
+	output["errors"] = _previousErrors;
```
chriseth: move?

test/cmdlineTests/standard_yul_single_file_via_urls/output.json
```diff
@@ -3,10 +3,10 @@
     [
```
chriseth, line 3: Why does this only have a single error now? Shouldn't it have the non-fatal followed by the fatal one?

chriseth:
Actually maybe approaching this in a very different way: if there are already errors in `InputsAndSettings`, does it make sense to process it further? I.e. we could just return an object with errors right at the beginning of that function if there are already errors.

test/cmdlineTests/standard_yul_single_file_via_urls/output.json
cameel: Fixed.

This was a leftover from my initial solution, where I made it not even run `compileYul()` if there were non-fatal errors. But then I realized that it's not good (they're non-fatal so we should continue). I forgot to update the test after that.

cameel:
> Actually maybe approaching this in a very different way: if there are already errors in `InputsAndSettings`, does it make sense to process it further? I.e. we could just return an object with errors right at the beginning of that function if there are already errors.

Haha, that's how I did it initially, but then I thought that I should not stop if these were not fatal so I rewrote it like it is now :)

Should I go back to my initial version? I think it was better for this particular error (the message that there wasn't exactly one input file is confusing) but in others situations it might be better to continue. User will get more feedback in one go that way.

libsolidity/interface/StandardCompiler.cpp
cameel: Done.

chriseth:
Yeah, maybe better to do it as you did it.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1313,15 +1315,32 @@ Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSetting
 Json::Value StandardCompiler::compileYul(InputsAndSettings _inputsAndSettings)
 {
 	if (_inputsAndSettings.sources.size() != 1)
-		return formatFatalError("JSONError", "Yul mode only supports exactly one input file.");
+		return formatFatalError(
```
chriseth: The main thing is that I find it weird to forward `_inputsAndSettings.errors` here. Maybe it would be better to set a local flag, move the

`output["errors"] = std::move(_inputsAndSettings.errors);`

to the start of this function and do `output["errors"] += formatFatalError(...)` here.

cameel: Why is it weird to forward the errors? After all `formatFatalError()` constructs full JSON output, just like `compileYul()` does so it makes sense to me that both would take over ownership of `_inputsAndSettings.errors` and add more stuff to it.

I can change it but I thought it was cleaner and more concise this way.

cameel: I mean, I would understand the problem if `formatFatalError()` was returning a single error as the name implies but it does not.

chriseth: it's just that it is easy to forget to supply the right error list - so I'm wondering if it would not be better to just return an error and append it at the call site to `output["errors"]`.

chriseth: Another option would be to factor out this checking of fatal errors into its own function

cameel: ok, so how about now. I now append to `output["errors"]` directly.

cameel:
Rebased on top of the pylint fix.

chriseth (approved):

Issue #11870, opened 2021-08-31
[soltest] Directive for skipping SMTChecker tests on specific platforms


cameel:
This came up in https://github.com/ethereum/solidity/pull/11861#discussion_r699457642.

## Abstract
If you search the repo for `Disabled because of Spacer nondeterminism`, you'll see that we have quite a few SMTChecker tests where assertions had to be disabled due to results not being deterministic between platforms. The workaround is to comment-out these assertions, which disables them on all platforms and weakens our test suite.

In most cases the results are different just on macOS and I think that it would be better to have a mechanism for skipping the whole test on the specific platforms where we're aware of the non-determinism.

## Specification
Add a soltest directive called `skipPlatforms` that accepts a list of values. Values can be: `macos`, `windows`, `linux`, `emscripten` (I don't think we need more granularity than that). When a test is executed, it should be skipped if the directive contains the name of the current platform.

Example usage:
```solidity
contract C {}
// ====
// SMTEngine: all
// skipPlatforms: macos, windows, linux
// ----
// Warning 6328: (1-5): CHC: Assertion violation might happen here.
// Warning 4661: (1-5): BMC: Assertion violation happens here.
```

PR #11985, opened 2021-09-16
[SMTChecker] Output values for block, msg and tx variables in cex

### LINKED ISSUE #10785: [SMTChecker] CHC Show tx variables in counterexample
`msg.value` is implemented in https://github.com/ethereum/solidity/pull/10828 when > 0, but it would be nice to also implement the others, when they're used.
An idea is, for every function in the transaction trace, if a `block`, `msg` or `tx` variable is used (can be detected syntactically), show its value in the transaction function call.

leonardoalt:
Fix https://github.com/ethereum/solidity/issues/10785
Depends on https://github.com/ethereum/solidity/pull/12036

This PR outputs values for block, msg and tx variables that are used in the outermost function of a transaction in the counterexample. `msg.value` is also added if the function is payable. These are not reported for internal functions and for external calls to `this`.


libsolidity/formal/CHC.cpp
```diff
@@ -1890,7 +1890,10 @@ optional<string> CHC::generateCounterexample(CHCSolverInterface::CexGraph const&
 			if (!pred->isConstructorSummary())
 				for (unsigned v: callGraph[node])
 					_dfs(node, v, depth + 1, _dfs);
-			calls.push_front(string(depth * 4, ' ') + pred->formatSummaryCall(nodeArgs(node), m_charStreamProvider));
+
+			bool appendTxVars = pred->isConstructorSummary() || pred->isFunctionSummary() || pred->isExternalCallUntrusted();
```
leonardoalt, line 1896: We only want to add the values for external calls to the contract.

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+	string txModel;
+
+	if (_appendTxVars)
 	{
-		bigint x(*v);
-		if (x > 0)
-			value = "{ value: " + *v + " }";
+		set<string> txVars;
+		if (isFunctionSummary() && programFunction()->isPayable())
+			txVars.insert("msg.value");
```
cameel: So the assumption is that `programFunction()` is not NULL when it's a normal function but it could be NULL if it's a constructor? I think there should be an assert against this:
```suggestion
		if (isFunctionSummary())
		{
			smtAssert(programFunction(), "");
			if (programFunction()->isPayable())
				txVars.insert("msg.value");
		}
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+		}
+
+		struct TxVarsVisitor: public ASTConstVisitor
+		{
+			bool visit(MemberAccess const& _memberAccess)
+			{
+				Expression const* memberExpr = SMTEncoder::innermostTuple(_memberAccess.expression());
+
+				auto const& exprType = memberExpr->annotation().type;
+				solAssert(exprType, "");
```
cameel, line 245: I see that we have a special assert type for SMTChecker. This code should be using it then :)
```suggestion
				smtAssert(exprType, "");
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
 	{
-		bigint x(*v);
-		if (x > 0)
-			value = "{ value: " + *v + " }";
+		set<string> txVars;
+		if (isFunctionSummary() && programFunction()->isPayable())
+			txVars.insert("msg.value");
+		else if (isConstructorSummary())
+		{
+			auto fun = programFunction();
```
cameel: I had to check what kind of thing `programFunction()` is actually returning (wasn't sure if plain `FunctionDefinition` or some wrapper and if it was a pointer or a reference). I think the code is easier to read with more types used explicitly.
```suggestion
			FunctionDefinition const* fun = programFunction();
```

leonardoalt:
This is ready for review.

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+				txVars.insert("msg.value");
+		}
+
+		struct TxVarsVisitor: public ASTConstVisitor
+		{
+			bool visit(MemberAccess const& _memberAccess)
+			{
+				Expression const* memberExpr = SMTEncoder::innermostTuple(_memberAccess.expression());
+
+				auto const& exprType = memberExpr->annotation().type;
```
cameel: I think you're creating a reference to a pointer here :)
```suggestion
				Type const* exprType = memberExpr->annotation().type;
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+				solAssert(exprType, "");
+				if (exprType->category() == Type::Category::Magic)
+				{
+					if (auto const* identifier = dynamic_cast<Identifier const*>(memberExpr))
+					{
+						auto const& name = identifier->name();
+						if (name == "block" || name == "msg" || name == "tx")
+							txVars.insert(name + "." + _memberAccess.memberName());
+					}
+				}
```
cameel:
```suggestion
				if (exprType->category() == Type::Category::Magic)
					if (auto const* identifier = dynamic_cast<Identifier const*>(memberExpr))
					{
						auto const& name = identifier->name();
						if (name == "block" || name == "msg" || name == "tx")
							txVars.insert(name + "." + _memberAccess.memberName());
					}
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+			{
+				Expression const* memberExpr = SMTEncoder::innermostTuple(_memberAccess.expression());
+
+				auto const& exprType = memberExpr->annotation().type;
+				solAssert(exprType, "");
+				if (exprType->category() == Type::Category::Magic)
+				{
+					if (auto const* identifier = dynamic_cast<Identifier const*>(memberExpr))
+					{
+						auto const& name = identifier->name();
```
cameel:
```suggestion
						ASTString const& name = identifier->name();
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+						if (name == "block" || name == "msg" || name == "tx")
+							txVars.insert(name + "." + _memberAccess.memberName());
```
cameel, line 251: Does this handle stuff like `(msg).value`? Seems to be valid syntax.

cameel, line 251: Shadowing `msg`, `tx` and `block` seems to be only a warning. How will this code deal with something like this?

```solidity
contract C {
    struct S {
        uint value;
        address origin;
        uint number;
    }
    function f() public payable {
        S memory msg = S(42, address(0), 666);
        S memory tx = S(42, address(0), 666);
        S memory block = S(42, address(0), 666);
        msg.value;
        tx.origin;
        block.number;
    }
}
```

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,64 @@ string Predicate::formatSummaryCall(
...
+			set<string> txVars;
+		} txVarsVisitor;
+
+		if (auto fun = programFunction())
+		{
+			fun->accept(txVarsVisitor);
+			txVars += txVarsVisitor.txVars;
+		}
+
+		auto txValues = readTxVars(_args.at(4));
```
cameel, line 267: Does 4 here mean the argument from function summary at position 4, i.e. `txData`?

test/libsolidity/SMTCheckerTest.cpp
```diff
@@ -66,6 +66,12 @@ SMTCheckerTest::SMTCheckerTest(string const& _filename): SyntaxTest(_filename, E
 		m_ignoreCex = true;
 	else
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid SMT counterexample choice."));
+
+	[[maybe_unused]] auto const& os = m_reader.stringSetting("SMTIgnoreOS", "none");
+#ifdef __APPLE__
+	if (os == "osx")
+		m_shouldRun = false;
+#endif
```
cameel: So this PR also partially implements #11870? :)

Why not add Linux and Windows as an option and extract it into a separate PR?

cameel (commented):
I haven't reviewed this fully yet but here's some general feedback.

libsolidity/formal/Predicate.cpp
cameel: Ah, I see it will only run for `Type::Category::Magic`. I'd still add a test that has this kind of shadowing to verify that.

cameel: Or how about this :)
```solidity
(true ? msg : msg).value;
```

cameel: Fortunately something like `[msg, tx, block][i].value` is an error though :)

test/libsolidity/SMTCheckerTest.cpp
leonardoalt: Yea could do that, I did it here mostly because it was quick and I wanted to see if it would make the tests pass.

libsolidity/formal/Predicate.cpp
leonardoalt: Yep, will add a comment.

libsolidity/formal/Predicate.cpp
leonardoalt: Ok!

leonardoalt: sigh...

libsolidity/formal/Predicate.cpp
leonardoalt: `smtAssert` is used only in `libsmtutil` I think. We use `solAssert` here because this is all part of `libsolidity`.

libsolidity/formal/Predicate.cpp
leonardoalt: If it's an actual constructor it should be the same as a normal function, but if it's a contract and that contract has a constructor this should also be true.

leonardoalt: Added assertion.

libsolidity/formal/Predicate.cpp
leonardoalt: Added test

leonardoalt: This case was actually buggy already in develop, nice find :)
Fixed in https://github.com/ethereum/solidity/pull/12036

leonardoalt:
Rebased and ready for review

Changelog.md
```diff
@@ -5,13 +5,13 @@ Language Features:
 
 Compiler Features:
 
+Compiler Features:
```
chriseth: duplication

leonardoalt: fixed

Changelog.md
```diff
@@ -23,7 +21,6 @@ Bugfixes:
  * AST: Export ``canonicalName`` for ``UserDefinedValueTypeDefinition`` and ``ContractDefinition``.
 
 
-
```
hrkrshnn: Some cleaning up needed.

test/libsolidity/smtCheckerTests/special/msg_vars_chc_internal.sol
```diff
@@ -21,8 +21,11 @@ contract C {
...
 		assert(sig == msg.sig); // should hold with CHC
 		assert(value == msg.value); // should hold with CHC
+
+		assert(msg.value == 10); // should fail
 	}
 }
 // ====
 // SMTEngine: chc
 // ----
+// Warning 6328: (621-644): CHC: Assertion violation happens here.\nCounterexample:\ndata = [38, 18, 31, 240], sender = 0, sig = 638722032, value = 0\n\nTransaction trace:\nC.constructor()\nState: data = [], sender = 0, sig = 0, value = 0\nC.f(){ msg.data: [38, 18, 31, 240], msg.sender: 0, msg.sig: 638722032, msg.value: 0 }\n    C.g() -- internal call
```
hrkrshnn: For data or sig, I think hex is a better choice. But fine either way.

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +215,67 @@ string Predicate::formatSummaryCall(
...
+
+		// Here we are interested in txData from the summary predicate.
+		auto txValues = readTxVars(_args.at(4));
+		vector<string> values;
+		for (auto const& _var: txVars)
+			if (auto v = txValues.at(_var))
+				values.push_back(_var + ": " + *v);
+
+		if (!values.empty())
+			txModel = "{ " + boost::algorithm::join(values, ", ") + " }";
```
hrkrshnn, line 274: I wonder if replacing this by the Json library is a good idea?

Changelog.md
leonardoalt: This is weird, I don't have it in my local diff

test/libsolidity/smtCheckerTests/special/msg_vars_chc_internal.sol
leonardoalt: I think so too. Let me see if I can do this in an easy way, otherwise I'd vote to do it in a different PR.

libsolidity/formal/Predicate.cpp
leonardoalt: which json library?

hrkrshnn: Just the regular `json-cpp` that we use. `txModel` would be `Json::value` and then you can write `txModel["msg.data"] = data`.

leonardoalt: But how does that help?

hrkrshnn: No need to a write custom translator to strings. The final output would be just the JSON as string.

leonardoalt: Oh I see what you mean, for the entire message?

hrkrshnn: Yes!

libsolidity/formal/Predicate.cpp
```diff
@@ -214,19 +219,79 @@ string Predicate::formatSummaryCall(
...
+		// We can either have a pretty version which has line breaks,
+		// or a compact version that does not but also has no spaces
+		// and has quotes.
+		// So we take the pretty version and remove line breaks,
+		// replace double spaces (which are left after removing line breaks)
+		// by single spaces, and remove quotes.
+		txModelStr = util::jsonPrettyPrint(txModel);
+		boost::replace_all(txModelStr, "\n", "");
+		boost::replace_all(txModelStr, "  ", " ");
+		boost::replace_all(txModelStr, "\"", "");
```
leonardoalt: @hrkrshnn this gives me what I want. Let me know if you prefer this or the previous version

leonardoalt: Actually it doesn't. The end of the string is then `name: value}`, whereas it should be `name: value }`. It's getting too hacky, I'm reverting

libsolidity/formal/Predicate.cpp
```diff
@@ -575,9 +595,9 @@ map<string, optional<string>> Predicate::readTxVars(smtutil::Expression const& _
 		{"block.number", TypeProvider::uint256()},
 		{"block.timestamp", TypeProvider::uint256()},
 		{"blockhash", TypeProvider::array(DataLocation::Memory, TypeProvider::uint256())},
-		{"msg.data", TypeProvider::bytesMemory()},
+		{"msg.data", TypeProvider::array(DataLocation::Memory, TypeProvider::fixedBytes(1))},
```
hrkrshnn: It is `calldata` in 
https://github.com/ethereum/solidity/blob/9d6eaa7a46425301ddb961a369b04f211d53691c/libsolidity/ast/Types.cpp#L3963

But not sure if that would be a problem here.

Maybe the `fixedBytes` argument can also be avoided?

Changelog.md
```diff
@@ -10,6 +10,8 @@ Bugfixes:
   * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
 
 
+
+
```
hrkrshnn: I think this commit can be squashed to the first one.

hrkrshnn (commented):
Looks good. Some small comments.

Changelog.md
leonardoalt: Yea I'm gonna squash everything after review

libsolidity/formal/Predicate.cpp
leonardoalt: Would that make the element type a `fixedBytes(1)` implicitly?

leonardoalt: We need that to report the type properly as hex.

hrkrshnn: Yes, `fixedBytes(1)` is implicit.

leonardoalt: Ok, changed and squashed.

hrkrshnn (approved):

PR #11986, opened 2021-09-16
Fix pylint warnings about the usage of f-strings

cameel:
Looks like we're now getting failures in [`chk_pylint`](https://app.circleci.com/pipelines/github/ethereum/solidity/19029/workflows/485209bf-fdcb-4e0d-bfe5-8679390a556f/jobs/845184) because pylint started complaining about `.format()` and `%` used for string formatting.

This PR fixes that by switching to f-string syntax.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1007,7 +1007,7 @@ jobs:
 
   b_bytecode_ems:
     docker:
-      - image: circleci/node:14
+      - image: circleci/node:16
     environment:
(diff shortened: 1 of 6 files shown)


cameel:
Looks like `b_bytecode_ems` is failing because its python interpreter is too old (3.5.3). Fortunately I can bump the base image from `circleci/node:14` to `circleci/node:16`, which has Python 3.7.3.

axic (approved):

PR #11990, opened 2021-09-17
[SMTChecker] Support user types

### LINKED ISSUE #11966: [SMTChecker] Support user defined types

leonardoalt:
Fix https://github.com/ethereum/solidity/issues/11966

Still need to add more tests, but code can be reviewed


test/libsolidity/smtCheckerTests/userTypes/user_abi_2.sol
```diff
@@ -0,0 +1,12 @@
...
+contract C {
+	function f(bytes memory data) public pure {
+		T x = abi.decode(data, (T));
+		T y = abi.decode(data, (T));
+		assert(T.unwrap(x) == T.unwrap(y)); // should hold
+		assert(T.unwrap(x) != T.unwrap(y)); // should fail
+	}
+}
+// ====
+// SMTEngine: all
```
hrkrshnn, line 12: The test updates are missing.

libsolidity/formal/SymbolicTypes.cpp
```diff
@@ -233,14 +233,18 @@ pair<bool, shared_ptr<SymbolicVariable>> newSymbolicVariable(
 	bool abstract = false;
 	shared_ptr<SymbolicVariable> var;
 	frontend::Type const* type = &_type;
-	if (!isSupportedTypeDeclaration(_type))
+
+	if (auto userType = dynamic_cast<UserDefinedValueType const*>(type))
+		type = &userType->underlyingType();
```
hrkrshnn: Maybe it's easier to implement this by returning `newSymbolicVariable(&userType->underlyingType(), _uniqueName, _context)`?

leonardoalt: I'm not sure that'd be easier. `newSymbolicVariable` is used in 4 places, so we'd need this logic in all those places. I'd prefer to centralize that here.

leonardoalt:
I *think* I managed to centralize the underlying unpacking in the fewest amount of places needed.

libsolidity/formal/Predicate.cpp
```diff
@@ -237,7 +237,7 @@ string Predicate::formatSummaryCall(
 	auto last = first + static_cast<int>(fun->parameters().size());
 	solAssert(first >= _args.begin() && first <= _args.end(), "");
 	solAssert(last >= _args.begin() && last <= _args.end(), "");
-	auto inTypes = FunctionType(*fun).parameterTypes();
+	auto inTypes = FunctionType(*fun).parameterTypesReplacingUserTypes();
```
hrkrshnn: I feel that this function `parameterTypesReplacingUserTypes` does not belong in `Types.h`, but as a helper function somewhere here. What do you think?

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -406,6 +406,9 @@ void SMTEncoder::endVisit(TupleExpression const& _tuple)
 	if (_tuple.annotation().type->category() == Type::Category::Function)
 		return;
 
+	if (_tuple.annotation().type->category() == Type::Category::TypeType)
```
hrkrshnn, line 409: Why is this needed?

libsolidity/formal/SymbolicTypes.cpp
hrkrshnn:
```suggestion
	if (auto userType = dynamic_cast<UserDefinedValueType const*>(type))
		return newSymbolicVariable(&userType->underlyingType(), _uniqueName, _context);
```
This is what I meant. The rest of the code can remain unchanged.

libsolidity/formal/Predicate.cpp
hrkrshnn: This would just be an applyMap or range-v3 transform using the function `underlyingType`.

libsolidity/formal/SymbolicTypes.cpp
leonardoalt: Ah I see

libsolidity/formal/SMTEncoder.cpp
leonardoalt: To avoid spurious unsupported warnings

libsolidity/formal/Predicate.cpp
leonardoalt: Ok!

libsolidity/formal/SymbolicTypes.cpp
leonardoalt: I left the other changes in this file just cus it looked better

hrkrshnn (approved):

chriseth:
Does this still need more tests (as written in the description) or is this fine to merge?

hrkrshnn:
I think this is fine to merge. The description is old. Will merge now.

PR #11235, opened 2021-04-10
Allow using stdin in cmdlineTests

cameel:
- If no `input.*` files are present in test directory, does not assume `input.sol` by default and does not pass any file name (not even `""`) to `solc` in that case.
- If a file called `stdin` is present and it's not a JSON test, puts its content on standard input of the compiler.

PR #11991, opened 2021-09-17
Fix handling of empty dirs and empty files in command-line tests

cameel:
After #11235, when `cmdlineTests.sh` finds an empty directory (which sometimes happens to me when editing tests and moving between branches), it assumes that it's a test with no empty files and empty stdin. This results in `solc` printing the command-line help, which is annoying and also probably confusing.

Also, `splitSources.py` assumes that the file it's processing is never empty.

This PR fixes these two problems.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -56,7 +56,8 @@ do
         *)
             matching_tests=$(find . -mindepth 1 -maxdepth 1 -type d -name "$1" | cut --characters 3- | sort)
 
-            if [[ $matching_tests == "" ]]; then
+            if [[ $matching_tests == "" ]]
+            then
                 patterns_with_no_matches+=("$1")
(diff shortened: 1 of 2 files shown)


test/cmdlineTests.sh
```diff
@@ -345,12 +345,17 @@ printTask "Running general commandline tests..."
 
         printTask " - ${tdir}"
 
+        if [[ $(ls -A "$tdir") == "" ]]; then
```
chriseth:
```suggestion
        if [[ $(ls -A "$tdir") == "" ]]
        then
```
(also below)

cameel: Changed.

I really hate how bulky the conditions seem to be in bash because of this though. Personally I strongly prefer version with `then` and `do` on the same line.

chriseth:
Can you squash the last commit?

cameel:
Squashed.

PR #11944, opened 2021-09-13
Debug info selection

cameel:
Fixes #11858.
~Depends on #11867.~ Merged.
~Weakly depends on #11993 - `chk_codingstyle` fails because of trailing whitespace without it.~ Merged
~Weakly depends on #11991 - I have one empty input file in cmdline tests, which makes `xxx_cli` jobs fail.~
Now it actually depends on #11991.

I could work around the dependency on #11991 and #11993 but they'll likely go in first anyway.

Still work in progress:
- [x] Make it affect `--asm` output as well.
- [x] Error when used in input modes that do not need it.
- [x] Standard JSON setting.
- [x] Command-line tests.
- [x] Documentation + changelog.

PR #11992, opened 2021-09-17
Rename existing debug info tests to match the new naming convention

cameel:
The naming I suggested in https://github.com/ethereum/solidity/pull/11903#discussion_r709454244 was not entirely correct because it's a Standard JSON test, but not a Yul test (it has `language: Solidity`). Also we're now starting to call these comments just "debug info" rather than "source location comments" so I'm renaming it to be consistent with names I used in #11944.

I'm submitting this as a separate PR because #11944 already touches too many files.


chriseth:
Needs rebase :(

chriseth (approved):

PR #11996, opened 2021-09-18
Isoltest validation

### LINKED ISSUE #11916: [isoltest] Command-line options are not validated
## Description
[CommonOptions::validate()](https://github.com/ethereum/solidity/blob/v0.8.7/test/Common.cpp#L112-L137) is responsible for validating command-line options passed to `soltest` and `isoltest`. It only runs in `soltest` though. In `isoltest` command-line parameters are not validated.

## Environment
- Compiler version: `0.8.8-develop.2021.9.9+commit.f6eecb8c.mod.Linux.g++`

## Steps to Reproduce

In `soltest` it works as expected:
```bash
soltest -- --enforce-gas-cost --evm-version spuriousDragon
```
```
Test setup error: /solidity/test/Common.cpp(127): Throw in function virtual void solidity::test::CommonOptions::validate() const
Dynamic exception type: boost::wrapexcept<solidity::test::ConfigException>
std::exception::what: Gas costs can only be enforced on latest evm version.
[solidity::util::tag_comment*] = Gas costs can only be enforced on latest evm version.
```
But for some reason it is not executed in `isoltest`:
```bash
isoltest --enforce-gas-cost --evm-version spuriousDragon
```
```
Running tests...
```

hawkess:
These are the current changes I've made that move defaults to the derived class constructor, explicitly define them for each option based on the values used to construct the `CommonOptions`/`IsolTestOptions` class and add `CommonOptions::validate` to the derived class's `validate`. 

This addresses the issue of hard coded overrides at the end of the parse chain, but with no way to negate a `true` default with a `po::bool_switch`, defaults essentially become permanent flags.

Fixes https://github.com/ethereum/solidity/issues/11916


Changelog.md
```diff
@@ -1,2254 +1,2328 @@
 ### 0.8.8 (unreleased)
 
 Language Features:
- * Inheritance: A function that overrides only a single interface function does not require the ``override`` specifier.
```
hrkrshnn: This file should be untouched. Could you please remove the changes in the PR? The last commit "changelog updated" should be removed.

test/tools/IsolTestOptions.h
```diff
@@ -33,8 +33,10 @@ struct IsolTestOptions: CommonOptions
 	bool noColor = false;
 	bool acceptUpdates = false;
 	std::string testFilter = std::string{};
+	std::string* editor = nullptr;
```
hrkrshnn:
```suggestion
	std::string const* editor = nullptr;
```
Maybe?

hrkrshnn (commented):
Thanks for the PR. Some quick comments.

Changelog.md
hawkess: Sorry. The Solidity documentation for contributing indicated that the changelog should be updated when making a PR; It has been removed.

test/tools/IsolTestOptions.h
hawkess: `program_options`  takes a reference to the resources to be modified in response to input on the command line so it can't be a `const std::string*` (or `std::string const*`) because what it points to is mutable. Did you mean make it a const pointer `std::string * const char`?

Changelog.md
hrkrshnn: No worries. We only change it when user facing things need to be updated. In this case, `isoltest` is only an internal tool and therefore doesn't need a changelog entry.

test/tools/IsolTestOptions.h
hrkrshnn: I see, this works, then.

hrkrshnn (commented):
The changes look good to me, but I'll wait for @cameel to see if this does everything from https://github.com/ethereum/solidity/issues/11916.

test/tools/IsolTestOptions.cpp
```diff
@@ -59,11 +59,20 @@ std::string editorPath()
 IsolTestOptions::IsolTestOptions(std::string* _editor):
 	CommonOptions(description)
 {
+	editor = _editor;
+	enforceViaYul = true;
+	enforceGasTest = (evmVersion() == langutil::EVMVersion{});
```
cameel, line 63: > This addresses the issue of hard coded overrides at the end of the parse chain, but with no way to negate a `true` default with a `po::bool_switch`, defaults essentially become permanent flags.

Let's just add `--no-enforce-via-yul` and `--no-enforce-gas-cost` for these.

test/tools/IsolTestOptions.cpp
```diff
@@ -59,11 +59,20 @@ std::string editorPath()
 IsolTestOptions::IsolTestOptions(std::string* _editor):
 	CommonOptions(description)
 {
+	editor = _editor;
```
cameel: The fact that this function gets a pointer to a string stored outside is really weird. I'd expect it to be stored in the struct.

My suggestion would be to make `editor` just a string and make the caller just take it out of the `IsolTestOptions` object.

test/tools/IsolTestOptions.cpp
```diff
@@ -59,11 +59,20 @@ std::string editorPath()
...
+	enforceViaYul = true;
+	enforceGasTest = (evmVersion() == langutil::EVMVersion{});
+	enforceGasTestMinValue = 100000;
```
cameel: Since this is a constructor, you should put these on the constructor initializer list rather than in the body.

Putting them directly on the struct fields would also be fine - that's how `CommonOptions` already does it after all.

test/tools/IsolTestOptions.h
cameel: Using a pointer from the outside is a weird thing to do though. Let's refactor it a bit to get rid of the pointer and just store the string in the struct instead.

test/tools/IsolTestOptions.h
```diff
@@ -33,8 +33,10 @@ struct IsolTestOptions: CommonOptions
...
 
-	IsolTestOptions(std::string* _editor);
+	explicit IsolTestOptions(std::string* _editor);
```
cameel: Nice.

test/Common.cpp
```diff
@@ -91,22 +91,27 @@ CommonOptions::CommonOptions(std::string _caption):
...
+		("ewasm", po::bool_switch(&ewasm)->default_value(ewasm), "tries to automatically find an ewasm vm and enable ewasm test-execution.")
+		("no-semantic-tests", po::bool_switch(&disableSemanticTests)->default_value(disableSemanticTests), "disable semantic tests")
+		("no-smt", po::bool_switch(&disableSMT)->default_value(disableSMT), "disable SMT checker")
+		("optimize", po::bool_switch(&optimize)->default_value(optimize), "enables optimization")
+		("enforce-via-yul", po::bool_switch(&enforceViaYul)->default_value(enforceViaYul), "Enforce compiling all tests via yul to see if additional tests can be activated.")
+		("enforce-compile-to-ewasm", po::bool_switch(&enforceCompileToEwasm)->default_value(enforceCompileToEwasm), "Enforce compiling all tests to Ewasm to see if additional tests can be activated.")
+		("enforce-gas-cost", po::bool_switch(&enforceGasTest)->default_value(enforceGasTest), "Enforce checking gas cost in semantic tests.")
+		("enforce-gas-cost-min-value", po::value(&enforceGasTestMinValue)->default_value(enforceGasTestMinValue), "Threshold value to enforce adding gas checks to a test.")
```
cameel, line 111: So without `default_value()` `program_options` just overwrites the original value of the variable? Does this mean that the default cost was being ignored so far?

cameel (commented):
Thanks for the PR!

Could you squash your commits a bit? Our convention is to have one commit per logical change, with no separate commits going back and forth, fixing typos, adding changes from review, etc. This makes reviewing commit by commit much easier.

test/tools/IsolTestOptions.cpp
hawkess: I agree that it is weird, but I didn't create the interface that led to that behavior. The `editor` string is used by `isoltest.cpp` for executing editor commands, so I would assume that is why it controls the resource and lifecycle. `IsolTestOptions` really only needs it to initialize the `options_description` object but since that can't be done in the constructor and still allow the option defaults to be modified, it's just passed through a helper variable right now rather than introducing a dependency to get it directly.

test/Common.cpp
hawkess: The default cost was previously being used because it was being directly assigned to at the end of the parse chain, it was actually user-defined cost that was being ignored. This change allows default values to be determined via initialization of the member variables at construction time and then propagated to the `options_description` object via the `default_value` calls so `program_options` controls all defaulting behavior and still accepts user input instead of direct assignment overwriting those values post-parse.

test/tools/IsolTestOptions.cpp
cameel: I know, it was here before your changes, I'm not saying that this is wrong on your part :)  I'm just saying that it's ok to refactor it instead of extending this design. I think it should be changed now because storing the pointer in the struct is worse than just passing it via parameter. The struct could easily outlive the string.

test/Common.cpp
cameel: > The default cost was previously being used because it was being directly assigned to at the end of the parse chain, it was actually user-defined cost that was being ignored.

That's only for `isoltest`. `soltest` was not overwriting it.

The change is fine. Even if it does not change behavior, it makes it more explicit, which is good. I'm just trying to figure out if the previous code was actually taking the defaults into account or just overwriting them.

test/tools/IsolTestOptions.cpp
hawkess: I can and will move the `editor` assignment to an initializer list, but the rest are members of the `CommonOptions` base class so, as far as I know, that is not valid syntax to initialize them on a list from the derived class. I did it this way to avoid having to pass a parameter for every single option to the base class constructor to modify defaults via the derived class; they can just be assigned in the derived class constructor after initialization by the base class and keep the interface simpler.

Changelog.md
cameel: Yeah, looks like the [Contributing docs > Workflow for pull requests](https://docs.soliditylang.org/en/latest/contributing.html#workflow-for-pull-requests) is a bit vague about that. It does not say that it's only for user-facing stuff.

```
New features and bugfixes should be added to the `Changelog.md` file: please follow the style of previous entries, when applicable.
```

We should clarify it.

test/tools/IsolTestOptions.cpp
cameel: Ah, right. Forgot that they're from the base class. It's fine then.

test/tools/IsolTestOptions.cpp
hawkess: I understand. However, `isoltest` needs the `edtior` resource independently of whether or not the idea of the `IsolTestOptions` class even existed.  How would you suggest avoiding the reverse problem if the resource is moved? `IsolTestOptions` is not necessarily created before the editor string is used by `isoltest`. An `IsolTestOptions` object cannot exist without a `TestTool` object so the resource `IsolTestOptions.editor` points to is guaranteed to exist and the non-static `IsolTestOptions` instance is stored as a member of `TestTool` alongside the static `editor` variable so the pointer it contains will necessarily be destroyed at least simultaneously to the resource it points to if not before.

test/Common.cpp
hawkess: Yes, `soltest` was always functioning as expected because the defaults were implicit via the default values the struct was initialized with, `default_value` just makes it explicit. Without it, `program_options` still does not modify the value of the variable if the flag is not specified, so if it was already initialized with what you wanted as the default, there is no actual change to behavior.

test/tools/IsolTestOptions.cpp
cameel: > `isoltest` needs the `edtior` resource independently of whether or not the idea of the `IsolTestOptions` class even existed.

I don't think it does. The static `editor` variable is used only by `TestTool::handleResponse()`, which is not a static function and could just take it from `m_options.editor` instead (if we make it a `string` rather than a pointer).

We don't even have to move it - I originally suggested that because I did not analyze the whole logic here and I thought that the owner of the `editor` variable is actually the one creating the options struct.

hawkess: I didn't fully follow the dataflow so I missed that as well. What you said makes sense and has been changed.

test/tools/IsolTestOptions.cpp
```diff
@@ -56,14 +56,22 @@ std::string editorPath()
 
 }
 
-IsolTestOptions::IsolTestOptions(std::string* _editor):
+IsolTestOptions::IsolTestOptions():
 	CommonOptions(description)
 {
+	enforceViaYul = true;
+	enforceGasTest = (evmVersion() == langutil::EVMVersion{});
+	enforceGasTestMinValue = 100000;
```
cameel: I just noticed that this value is the same as in `CommonOptions`. Could you remove it from here? The less places where defaults are hard-coded, the better.

test/Common.cpp
```diff
@@ -91,22 +91,27 @@ CommonOptions::CommonOptions(std::string _caption):
...
-		("show-messages", po::bool_switch(&showMessages), "enables message output")
-		("show-metadata", po::bool_switch(&showMetadata), "enables metadata output");
+		("ewasm", po::bool_switch(&ewasm)->default_value(ewasm), "tries to automatically find an ewasm vm and enable ewasm test-execution.")
+		("no-semantic-tests", po::bool_switch(&disableSemanticTests)->default_value(disableSemanticTests), "disable semantic tests")
+		("no-smt", po::bool_switch(&disableSMT)->default_value(disableSMT), "disable SMT checker")
+		("optimize", po::bool_switch(&optimize)->default_value(optimize), "enables optimization")
+		("enforce-via-yul", po::value<bool>(&enforceViaYul)->default_value(enforceViaYul)->implicit_value(true), "Enforce compiling all tests via yul to see if additional tests can be activated.")
+		("enforce-compile-to-ewasm", po::bool_switch(&enforceCompileToEwasm)->default_value(enforceCompileToEwasm), "Enforce compiling all tests to Ewasm to see if additional tests can be activated.")
+		("enforce-gas-cost", po::value<bool>(&enforceGasTest)->default_value(enforceGasTest)->implicit_value(true), "Enforce checking gas cost in semantic tests.")
```
cameel, line 110: These options show up like this in the command-line help now:
```
  --enforce-via-yul [=arg(=1)] (=1)
  --enforce-gas-cost [=arg(=1)] (=1)
                       Enforce checking gas cost in semantic tests.
  --enforce-gas-cost-min-value arg (=100000)
                       Threshold value to enforce adding gas checks to a test.
```

Is there a way to show something like `--enforce-via-yul [=arg(=true)] (=true)` (or even better `--enforce-via-yul (=true)`) instead? If not, I guess we can live with that but it being `1` makes it look a bit less like a simple switch and more like an option that actually requires some specific integer value.

cameel (commented):
Looks good now.

Actually, I'd prefer if it showed `true` rather than `1` on the help screen but if you can't find a way to do that, I can approve without it. @hrkrshnn what do you think about these boolean flags?

test/Common.cpp
hawkess: That is internal to the `program_options` implementation, unfortunately.

test/tools/IsolTestOptions.cpp
hawkess: Done.

cameel (approved):

hrkrshnn:
@cameel Since this is just an internal tool, I'm fine with the default value being shown as either 1 or true.

test/tools/IsolTestOptions.cpp
```diff
@@ -56,14 +56,21 @@ std::string editorPath()
 
 }
 
-IsolTestOptions::IsolTestOptions(std::string* _editor):
+IsolTestOptions::IsolTestOptions():
 	CommonOptions(description)
 {
+	enforceViaYul = true;
+	enforceGasTest = (evmVersion() == langutil::EVMVersion{});
+}
```
axic, line 64: What happened to `enforceGasTestMinValue = 100000;`?

hawkess: It is the same default value as in the base class, @cameel wanted to remove the redundancy.

cameel: Yeah, it was redundant. Hard-coding it in two places does not seem like a good idea and changing it to `enforceGasTestMinValue = enforceGasTestMinValue` would be pointless.

axic: Ok it was [here](https://github.com/ethereum/solidity/pull/11996#discussion_r712135031), I just didn't expand all the closed review comments.

Issue #9590, opened 2020-08-06
Output Yul ASTs


leonardoalt:
It would be helpful/essential for verification tools to have access to the Yul ASTs that are generated throughout compilation via Yul.

chriseth:
Related to https://github.com/ethereum/solidity/pull/9053

chriseth:
I'm wondering - is it really essential or would it also work to run solidity in yul mode and output the ast?

leonardoalt:
Which ast would that be? What I'm imagining is that all the Yul compilation steps could be exported:
- Solidity -> Yul
- Yul -> Optimized Yul
- Optimized Yul -> Dialect Yul (if applicable)

axic:
> Which ast would that be? What I'm imagining is that all the Yul compilation steps could be exported:
> 
> Solidity -> Yul

This is called `ir` in standard json, and displayed via `--ir` on cli.

> Yul -> Optimized Yul

This is displayed on cli via `--ir-optimized`.

> Optimized Yul -> Dialect Yul (if applicable)

This is not available yet.

chriseth:
I'm still not sure what this issue is about and what functionality we are really missing here.

Marenz:
As far as I understood, it's about the YulAST, being output as JSON just like our Solidity AST can be exported...

cameel:
This feature was also requested in #11043.

bendyarm:
I believe this issue has its roots in a discussion @acoglio and I had with @leonardoalt .

In order to verify the correctness of specific applications of Yul optimizers, you need the before and after Yul code.

Right now, the "standard JSON" form can contain "ir" and "irOptimized", but those are source code, not ASTs.  Yul surface syntax is unambiguous so that's not too bad.  It does require that we trust the prettyprinter and that we have a formal grammar and parser.  (We have those now.  E.g., the [ABNF grammar](https://github.com/acl2/acl2/blob/master/books/kestrel/yul/language/abnf-grammar-new.txt) could be useful to others.)

However, it would be higher assurance if, in addition to the before and after Yul code, the before and after internal ASTs (that are input and output to the optimizers) could be serialized to the JSON output structure, as @Marenz suggests.

cameel:
@bendyarm 
> the [ABNF grammar](https://github.com/acl2/acl2/blob/master/books/kestrel/yul/language/abnf-grammar-new.txt)

[...]

PR #11998, opened 2021-09-20
Store both Yul and Solidity locations in debug data

### LINKED ISSUE #11932: ICE when the file name in `@use-src` annotation does not match the name of the file passed to `--strict-assembly`
## Description
If you try to assemble a file using a different name than the one included in the `@use-src` annotation, the compiler crashes. 

There must be some additional factors involved because it only happens in specific circumstances. For example in the example below the problem goes away if I remove any of the annotation or if I do not use `calldataload(0)`.

## Steps to Reproduce
- `input.yul`
    ```yul
    /// @use-src 0:"input.sol"
    object "C" {
        code {
            /// @src 0:0:0
            calldataload(0)
        }
    }
    ```
```bash
solc --strict-assembly input.yul
```
```
Warning: Yul is still experimental. Please use the output with care.
terminate called after throwing an instance of 'boost::wrapexcept<solidity::yul::YulAssertion>'
  what():
Aborted (core dumped)
```
Note that we don't have a handler for `InternalCompilerError`s in assembly mode. #11732 should fix that.

The assertion failure happens at https://github.com/ethereum/solidity/blob/e28d00a76daa0747f00ffe47cea17862bca55771/libyul/AssemblyStack.cpp#L94

## Environment
- Compiler version: Happens both on 0.8.7 and on the latest code from `develop`.

cameel:
Fixes #11932.

- [x] ICE is gone but still not all errors from https://github.com/ethereum/solidity/issues/11932#issuecomment-922848821 have locations. This might still be using the original location in places where IR location should be used.
- [x] Make sure nothing is broken and maybe add some tests if the current ones did not detect all the breakage.
- [x] Some of the updated lines should be using the `locationOf()` helper. I only started adding it halfway through.

### DIFF
--- a/libsolidity/analysis/ControlFlowBuilder.cpp
+++ b/libsolidity/analysis/ControlFlowBuilder.cpp
@@ -409,6 +409,7 @@ bool ControlFlowBuilder::visit(InlineAssembly const& _inlineAssembly)
 void ControlFlowBuilder::visit(yul::Statement const& _statement)
 {
 	solAssert(m_currentNode && m_inlineAssembly, "");
+	solAssert(nativeLocationOf(_statement) == originLocationOf(_statement), "");
 	m_currentNode->location = langutil::SourceLocation::smallestCovering(m_currentNode->location, nativeLocationOf(_statement));
 	ASTWalker::visit(_statement);
 }
@@ -496,14 +497,15 @@ void ControlFlowBuilder::operator()(yul::Identifier const& _identifier)
 	solAssert(m_currentNode && m_inlineAssembly, "");
 	auto const& externalReferences = m_inlineAssembly->annotation().externalReferences;
 	if (externalReferences.count(&_identifier))
-	{
 		if (auto const* declaration = dynamic_cast<VariableDeclaration const*>(externalReferences.at(&_identifier).declaration))
+		{
+			solAssert(nativeLocationOf(_identifier) == originLocationOf(_identifier), "");
 			m_currentNode->variableOccurrences.emplace_back(
 				*declaration,
 				VariableOccurrence::Kind::Access,
 				nativeLocationOf(_identifier)
 			);
-	}
+		}
 }
 
 void ControlFlowBuilder::operator()(yul::Assignment const& _assignment)
@@ -514,11 +516,14 @@ void ControlFlowBuilder::operator()(yul::Assignment const& _assignment)
 	for (auto const& variable: _assignment.variableNames)
 		if (externalReferences.count(&variable))
 			if (auto const* declaration = dynamic_cast<VariableDeclaration const*>(externalReferences.at(&variable).declaration))
+			{
+				solAssert(nativeLocationOf(variable) == originLocationOf(variable), "");
 				m_currentNode->variableOccurrences.emplace_back(
 					*declaration,
 					VariableOccurrence::Kind::Assignment,
 					nativeLocationOf(variable)
 				);
+			}
 }
 
 void ControlFlowBuilder::operator()(yul::FunctionCall const& _functionCall)
--- a/libsolidity/analysis/ControlFlowBuilder.cpp
+++ b/libsolidity/analysis/ControlFlowBuilder.cpp
@@ -409,7 +409,7 @@ bool ControlFlowBuilder::visit(InlineAssembly const& _inlineAssembly)
 void ControlFlowBuilder::visit(yul::Statement const& _statement)
 {
 	solAssert(m_currentNode && m_inlineAssembly, "");
-	m_currentNode->location = langutil::SourceLocation::smallestCovering(m_currentNode->location, locationOf(_statement));
+	m_currentNode->location = langutil::SourceLocation::smallestCovering(m_currentNode->location, nativeLocationOf(_statement));
 	ASTWalker::visit(_statement);
 }
 
@@ -501,7 +501,7 @@ void ControlFlowBuilder::operator()(yul::Identifier const& _identifier)
 			m_currentNode->variableOccurrences.emplace_back(
 				*declaration,
 				VariableOccurrence::Kind::Access,
-				_identifier.debugData->location
+				nativeLocationOf(_identifier)
 			);
 	}
 }
@@ -517,7 +517,7 @@ void ControlFlowBuilder::operator()(yul::Assignment const& _assignment)
 				m_currentNode->variableOccurrences.emplace_back(
 					*declaration,
 					VariableOccurrence::Kind::Assignment,
-					variable.debugData->location
+					nativeLocationOf(variable)
 				);
 }
 
--- a/libsolidity/analysis/ReferencesResolver.cpp
+++ b/libsolidity/analysis/ReferencesResolver.cpp
@@ -201,9 +201,9 @@ bool ReferencesResolver::visit(Return const& _return)
 
 void ReferencesResolver::operator()(yul::FunctionDefinition const& _function)
 {
-	validateYulIdentifierName(_function.name, _function.debugData->location);
+	validateYulIdentifierName(_function.name, nativeLocationOf(_function));
 	for (yul::TypedName const& varName: _function.parameters + _function.returnVariables)
-		validateYulIdentifierName(varName.name, varName.debugData->location);
+		validateYulIdentifierName(varName.name, nativeLocationOf(varName));
 
 	bool wasInsideFunction = m_yulInsideFunction;
 	m_yulInsideFunction = true;
@@ -238,7 +238,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 	{
 		m_errorReporter.declarationError(
 			4718_error,
-			_identifier.debugData->location,
+			nativeLocationOf(_identifier),
 			"Multiple matching identifiers. Resolving overloaded identifiers is not supported."
 		);
 		return;
@@ -251,7 +251,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 		)
 			m_errorReporter.declarationError(
 				9467_error,
-				_identifier.debugData->location,
+				nativeLocationOf(_identifier),
 				"Identifier not found. Use \".slot\" and \".offset\" to access storage variables."
 			);
 		return;
@@ -261,7 +261,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 		{
 			m_errorReporter.declarationError(
 				6578_error,
-				_identifier.debugData->location,
+				nativeLocationOf(_identifier),
 				"Cannot access local Solidity variables from inside an inline assembly function."
 			);
 			return;
@@ -275,7 +275,7 @@ void ReferencesResolver::operator()(yul::VariableDeclaration const& _varDecl)
 {
 	for (auto const& identifier: _varDecl.variables)
 	{
-		validateYulIdentifierName(identifier.name, identifier.debugData->location);
+		validateYulIdentifierName(identifier.name, nativeLocationOf(identifier));
 
 
 		if (
@@ -289,7 +289,7 @@ void ReferencesResolver::operator()(yul::VariableDeclaration const& _varDecl)
 			if (!ssl.infos.empty())
 				m_errorReporter.declarationError(
 					3859_error,
-					identifier.debugData->location,
+					nativeLocationOf(identifier),
 					ssl,
 					"This declaration shadows a declaration outside the inline assembly block."
 				);
--- a/libsolidity/analysis/ReferencesResolver.cpp
+++ b/libsolidity/analysis/ReferencesResolver.cpp
@@ -201,9 +201,13 @@ bool ReferencesResolver::visit(Return const& _return)
 
 void ReferencesResolver::operator()(yul::FunctionDefinition const& _function)
 {
+	solAssert(nativeLocationOf(_function) == originLocationOf(_function), "");
 	validateYulIdentifierName(_function.name, nativeLocationOf(_function));
 	for (yul::TypedName const& varName: _function.parameters + _function.returnVariables)
+	{
+		solAssert(nativeLocationOf(varName) == originLocationOf(varName), "");
 		validateYulIdentifierName(varName.name, nativeLocationOf(varName));
+	}
 
 	bool wasInsideFunction = m_yulInsideFunction;
 	m_yulInsideFunction = true;
@@ -213,6 +217,8 @@ void ReferencesResolver::operator()(yul::FunctionDefinition const& _function)
 
 void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 {
+	solAssert(nativeLocationOf(_identifier) == originLocationOf(_identifier), "");
+
 	static set<string> suffixes{"slot", "offset", "length"};
 	string suffix;
 	for (string const& s: suffixes)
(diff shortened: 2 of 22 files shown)


libyul/AST.h
```diff
@@ -38,20 +38,34 @@ using Type = YulString;
...
+		return std::make_shared<DebugData const>(
+			std::move(_irLocation),
+			std::move(_originalLocation),
+			std::move(_astID)
+		);
 	}
 
-	langutil::SourceLocation location;
+	/// Location in the Yul code.
+	langutil::SourceLocation irLocation;
```
chriseth: I'm actually not sure we should call this `irLocation` - if you use yul in "native" mode, it is not really an IR.

libyul/AST.h
```diff
@@ -38,20 +38,34 @@ using Type = YulString;
...
+	/// Location in the original source that the Yul code was produced from.
+	/// Optional. Only present if the Yul source contains location annotations.
+	langutil::SourceLocation originalLocation;
```
chriseth: What about `originLocation?

libyul/AST.h
cameel: So `nativeLocation`?

libyul/AST.h
cameel: Sounds fine to me. I'd be even fine with `origin` but I'm not sure if it would then be clear enough that we're dealing with location.

libyul/AST.h
cameel: Or maybe just `yulLocation`?

chriseth:
I have the impression that AsmParser needs some more changes, but I might be wrong.

We have three modes for AsmParser:
1) native yul mode - irLocation and originalLocation should always point to the locations generated from the scanner ignoring the annotations
2) location override: This is used for snippets of yul code generated from solidity code in the old codegen. It probably makes sense for both to point at the solidity source location. We could also make the ir location point at the scanner location, but we probably don't have the actual source anymore at the point where we generate a potential parser error.
3) from comments: ir location points at scanner, original location is taken from comments

It might very well be that your code already accomplishes the above. If this is the case, it might be worth refactoring it at a later point in time to make the distinction clearer - especially if it is possible in 2) for the ir location to point at the scanner locations - in this case we can make it always point at the scanner locations.

cameel:
I'll take a deeper look at that tomorrow. For now in a few places I wasn't really sure which location to take but just went with it to get it to a state where it compiles, with the intention to do one more pass through the code after that. Judging by test failures, I must have chosen wrong in some places :)

libyul/AsmParser.h
```diff
@@ -99,9 +97,10 @@ class Parser: public langutil::ParserBase
...
 	langutil::SourceLocation currentLocation() const override
 	{
-		if (m_useSourceLocationFrom == UseSourceLocationFrom::Scanner)
-			return ParserBase::currentLocation();
-		return m_locationOverride;
+		if (m_useSourceLocationFrom == UseSourceLocationFrom::LocationOverride)
+			return m_locationOverride;
+
+		return ParserBase::currentLocation();
 	}
```
cameel, line 104: > ICE is gone but still not all errors from https://github.com/ethereum/solidity/issues/11932#issuecomment-922848821 have locations. This might still be using the original location in places where IR location should be used.

There was actually an extra bug here. Location override was overriding current location even when location from comments was selected. This resulted in empty location in errors reported from `AsmParser.cpp`. See `test/libyul/yulSyntaxTests/invalid/literals_on_stack_disallowed_with_debug_info.yul`.

libyul/AsmParser.cpp
```diff
@@ -76,11 +65,38 @@ std::shared_ptr<DebugData const> Parser::createDebugData() const
 	switch (m_useSourceLocationFrom)
 	{
 		case UseSourceLocationFrom::Scanner:
-			return DebugData::create(ParserBase::currentLocation());
+			return DebugData::create(ParserBase::currentLocation(), ParserBase::currentLocation());
 		case UseSourceLocationFrom::LocationOverride:
-			return DebugData::create(m_locationOverride);
+			return DebugData::create(m_locationOverride, m_locationOverride);
```
cameel, line 70: > 2. location override: This is used for snippets of yul code generated from solidity code in the old codegen. It probably makes sense for both to point at the solidity source location.

I think that `m_locationOverride` for both is the most consistent. If I use current location for `nativeLocation` instead, `Parser` will be reporting errors at `m_locationOverride` but `AsmAnalyzer` (which, I'm assuming gets locations from `nativeLocation` in the AST) will be be reporting them at the current location.

> We could also make the ir location point at the scanner location, but we probably don't have the actual source anymore at the point where we generate a potential parser error.

Errors are processed and reported from within `CompilerContext::appendInlineAssembly()` which still has the original snippet. To check that I introduced an error to one of the generated sources and there were no failing assertions and messages seemed to be pointing at the correct location in the snippet - even when `nativeLocation` was set to the current location rather than the override.

But there's still the problem with inconsistent location in errors I mentioned above. We can change that but we should then also change error messages to consistently use one kind of locations.

libsolidity/analysis/ControlFlowBuilder.cpp
```diff
@@ -409,7 +409,7 @@ bool ControlFlowBuilder::visit(InlineAssembly const& _inlineAssembly)
 void ControlFlowBuilder::visit(yul::Statement const& _statement)
 {
 	solAssert(m_currentNode && m_inlineAssembly, "");
-	m_currentNode->location = langutil::SourceLocation::smallestCovering(m_currentNode->location, locationOf(_statement));
+	m_currentNode->location = langutil::SourceLocation::smallestCovering(m_currentNode->location, nativeLocationOf(_statement));
```
cameel: Changing this to `originLocationOf()` does not break any tests so I think we don't have coverage for this in case where Yul source has debug info annotations.

libsolidity/analysis/ReferencesResolver.cpp
```diff
@@ -238,7 +238,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 	{
 		m_errorReporter.declarationError(
 			4718_error,
-			_identifier.debugData->location,
+			nativeLocationOf(_identifier),
 			"Multiple matching identifiers. Resolving overloaded identifiers is not supported."
 		);
```
cameel, line 249: Changing this to `originLocationOf()` does not break any tests so but I guess this is because we cannot have debug info annotations in inline assembly?

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -171,10 +171,10 @@ void ASTJsonConverter::appendExpressionAttributes(
 	_attributes += exprAttributes;
 }
 
-Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifier const* ,InlineAssemblyAnnotation::ExternalIdentifierInfo> _info) const
+Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifier const*, InlineAssemblyAnnotation::ExternalIdentifierInfo> _info) const
 {
 	Json::Value tuple(Json::objectValue);
-	tuple["src"] = sourceLocationToString(_info.first->debugData->location);
+	tuple["src"] = sourceLocationToString(nativeLocationOf(*_info.first));
```
cameel, line 177: Changing this to `originLocationOf()` does not break any tests so but I guess this is because we cannot have debug info annotations in inline assembly?

libyul/AST.h
cameel: I went with `nativeLocation`.

libsolidity/analysis/ControlFlowBuilder.cpp
chriseth: Just to check: `nativeLocationOf(_statement)` always returns a "proper" source location, no matter in which mode the parser was used, right?

What I'm a bit worried about is the possibility of the location being the same when we use `appendAssembly` in the old code generator and supply two different snippets (that would not have a different name) but two yul ast items have the same native locations inside these snippets.

IIRC, `location` is not only used for informational purposes, but to check equality in the source - is this the case here?

chriseth: Ah wait, this is only about inline assembly here (and not compiler-generated assembly for helper routines), so native location and origin location should be the same anyway. Maybe assert that?

libsolidity/analysis/ReferencesResolver.cpp
chriseth: Same here. Yes, debug info is invalid in inline assembly. You can add an assertion if you want.

libyul/AsmJsonConverter.cpp
```diff
@@ -33,23 +33,23 @@ namespace solidity::yul
 
 Json::Value AsmJsonConverter::operator()(Block const& _node) const
 {
```
chriseth, line 35: Do we want to store both locations in the asm ast?

libyul/AsmParser.cpp
chriseth: It probably doesn't matter too much, because these will all lead to internal compiler errors anyway.

libyul/AsmParser.cpp
```diff
@@ -76,11 +65,38 @@ std::shared_ptr<DebugData const> Parser::createDebugData() const
...
+) const
+{
+	switch (m_useSourceLocationFrom)
+	{
+		case UseSourceLocationFrom::Scanner:
+		{
+			DebugData updatedDebugData = *_debugData;
+			updatedDebugData.nativeLocation.end = _location.end;
+			updatedDebugData.originLocation.end = _location.end;
+			return make_shared<DebugData const>(updatedDebugData);
```
chriseth: move?

libyul/AsmParser.cpp
```diff
@@ -76,11 +65,38 @@ std::shared_ptr<DebugData const> Parser::createDebugData() const
...
+		}
+		case UseSourceLocationFrom::LocationOverride:
+			// Ignore the update. The location we're overriding with is not supposed to change
+			return _debugData;
+		case UseSourceLocationFrom::Comments:
+		{
+			DebugData updatedDebugData = *_debugData;
+			updatedDebugData.nativeLocation.end = _location.end;
+			return make_shared<DebugData const>(updatedDebugData);
```
chriseth: move?

libyul/AsmParser.h
```diff
@@ -119,6 +119,12 @@ class Parser: public langutil::ParserBase
 	/// Creates a DebugData object with the correct source location set.
 	std::shared_ptr<DebugData const> createDebugData() const;
 
+	[[nodiscard]]
+	std::shared_ptr<DebugData const> updateLocationEndFrom(
```
chriseth: Instead of a nodiscard, we could also change the first parameter to  be a non-const reference.

cameel: ok, I guess it makes sense given the name of the function.

libyul/AsmJsonConverter.cpp
cameel: Sounds like something that could be useful to tools. I think it should be properly introduced in a standalone PR though. We need to agree on the name (maybe `"origin"`?) and decide if we want other debug info here, like AST ID. Also needs some tests.

chriseth: I think this should be in the same PR - we are changing the value of "location" in the case of code that has `@use-src`.
origin sounds good.

libsolidity/analysis/ReferencesResolver.cpp
cameel: Not necessarily invalid. I think they should be the same.

Assertions added.

libsolidity/analysis/ControlFlowBuilder.cpp
cameel: > Just to check: `nativeLocationOf(_statement)` always returns a "proper" source location, no matter in which mode the parser was used, right?

Yes.

> IIRC, `location` is not only used for informational purposes, but to check equality in the source - is this the case here?

Yeah, it looks like it's using it to disambiguate the stuff it adds to `variableOccurrences`.

> What I'm a bit worried about is the possibility of the location being the same when we use `appendAssembly` in the old code generator and supply two different snippets (that would not have a different name) but two yul ast items have the same native locations inside these snippets.

I don't really understand this. We're not storing snippets in debug info or passing them around. Or do you mean the inline assembly snippets we compile with the location override? I'm also not sure what the consequences of these locations being the same would be.

> Ah wait, this is only about inline assembly here (and not compiler-generated assembly for helper routines), so native location and origin location should be the same anyway. Maybe assert that?

Ah, you're right. It's in `libsolidity` so it must be. Assertions added.

libyul/AsmJsonConverter.cpp
cameel: > we are changing the value of "location" in the case of code that has `@use-src`.

I wanted to say that we are not because there are no changes in test output, but you're right, there should have been some changes. So either something's wrong or, more likely, we do not have coverage for this.

Anyway, I'm out of time now so I'll be doing that change tomorrow. I'll do it in the same PR if I manage to before the release but I still think a separate PR would be more convenient - this one is pretty much a bugfix for the fact that these locations changed in 0.8.6 and is basically finished.

chriseth:
commandline tests are failing.

cameel:
Yeah, I'm working on that.

libyul/AsmJsonConverter.cpp
chriseth: Ok, I'm fine with either, but it seems that there is a problem in the ast export and re-import anyway...

libsolidity/analysis/ReferencesResolver.cpp
```diff
@@ -238,7 +244,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
...
```
cameel, line 249: The assertion I added for this is actually failing during AST export and reimport of `syntaxTests/inlineAssembly/storage_reference_old_shadow.sol`:

```solidity
contract C {
    uint[] x;
    fallback() external {
        uint y_slot = 2;
        uint y_offset = 3;
        uint[] storage y = x;
        assembly {
            pop(y_slot)
            pop(y_offset)
        }
        y[0] = 2;
    }
}
// ----
```

This is why CLI tests are failing. BTW the error handling for AST import is swallowing errors at multiple levels (both in solc and in the test script), which makes it harder to debug than it should be.

libyul/AsmJsonConverter.cpp
cameel: Right. I think that the problem is that the origin location is not preserved in AST nodes in inline assembly when you export and reimport the AST (https://github.com/ethereum/solidity/pull/11998/files#r713769635). So it's exactly what you pointed out but it's not enough to add it to JSON, we also need to import that information back.

I'm going to remove the assertions from this PR and add them in the one that extends the AST instead.

chriseth: Of course, anything you add to the json export also has to be added to the json import :)

chriseth (dismissed):

libyul/AsmJsonConverter.cpp
cameel: I knew there was a catch somewhere :P

Anyway, I'm going to actually push one more change. I'll make the importer use the `"src"` location for both native and origin location - I'm not not sure if adding `"origin"` actually makes sense since it has to include file index and I only have one when compiling inline assembly - and there it's the same as `"src"` anyway.

cameel: > So either something's wrong or, more likely, we do not have coverage for this.

And this turns out to be because without #9590 there's no way to output the Yul AST (other than for inline assembly or generated sources and we can't have debug info annotations there). So naturally it does not get tested.

chriseth (approved):

PR #11999, opened 2021-09-21
[docs] Separate "Visibility and getters" section for state variables and functions

nikitastupin:
Hi 👋 

In #11926 I suggested to separate visibility section for state variables and functions in order for them to be more clear. For now I'm moving towards creating two sections `Contracts -> Functions -> Visibility` and `Contracts -> State Variables -> Visibility` and removing `Contracts -> Visibility and Getters` section.

Any comments and questions are welcome!

### DIFF
--- a/docs/contracts/visibility-and-getters.rst
+++ b/docs/contracts/visibility-and-getters.rst
@@ -1,20 +1,41 @@
 .. index:: ! visibility, external, public, private, internal
 
+.. |visibility-caveat| replace:: Making something ``private`` or ``internal`` only prevents other contracts from reading or modifying the information, but it will still be visible to the whole world outside of the blockchain.
+
 .. _visibility-and-getters:
 
 **********************
 Visibility and Getters
 **********************
 
-Solidity knows two kinds of function calls: internal
-ones that do not create an actual EVM call (also called
-a "message call") and external
-ones that do. Because of that, there are four types of visibility for
-functions and state variables.
+State Variable Visibility
+=========================
 
-Functions have to be specified as being ``external``,
-``public``, ``internal`` or ``private``.
-For state variables, ``external`` is not possible.
+``public``
+    Public state variables differ from internal ones only in that the compiler automatically generates
+    :ref:`getter functions<getter-functions>` for them, which allows other contracts to read their values.
+    When used within the same contract, the external access (e.g. ``this.x``) invokes the getter
+    while internal access (e.g. ``x``) gets the variable value directly from storage.
+    Setter functions are not generated so other contracts cannot directly modify their values.
+
+``internal``
+    Internal state variables can only be accessed from within the contract they are defined in
+    and in derived contracts.
+    They cannot be accessed externally.
+    This is the default visibility level for state variables.
+
+``private``
+    Private state variables are like internal ones but they are not visible in derived contracts.
+
+.. warning::
+    |visibility-caveat|
+
+Function Visibility
+===================
+
+Solidity knows two kinds of function calls: external ones that do create an actual EVM message call and internal ones that do not.
+Furthermore, internal functions can be made inaccessible to derived contracts.
+This gives rise to four types of visibility for functions.
 
 ``external``
     External functions are part of the contract interface,
@@ -24,27 +45,19 @@ For state variables, ``external`` is not possible.
 
 ``public``
     Public functions are part of the contract interface
-    and can be either called internally or via
-    messages. For public state variables, an automatic getter
-    function (see below) is generated.
+    and can be either called internally or via message calls.
 
 ``internal``
-    Those functions and state variables can only be
-    accessed internally (i.e. from within the current contract
-    or contracts deriving from it), without using ``this``.
-    This is the default visibility level for state variables.
+    Internal functions can only be accessed from within the current contract
+    or contracts deriving from it.
+    They cannot be accessed externally.
+    Since they are not exposed to the outside through the contract's ABI, they can take parameters of internal types like mappings or storage references.
 
 ``private``
-    Private functions and state variables are only
-    visible for the contract they are defined in and not in
-    derived contracts.
+    Private functions are like internal ones but they are not visible in derived contracts.
 
-.. note::
-    Everything that is inside a contract is visible to
-    all observers external to the blockchain. Making something ``private``
-    only prevents other contracts from reading or modifying
-    the information, but it will still be visible to the
-    whole world outside of the blockchain.
+.. warning::
+    |visibility-caveat|
 
 The visibility specifier is given after the type for
 state variables and between parameter list and


docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
 Visibility and Getters
 **********************
 
-Solidity knows two kinds of function calls: internal
```
chriseth, line 9: I think we should not delete this paragraph, it explains what internal and external calls are and that is important for what follows.

nikitastupin: The reason why I decided to remove this paragraph is (1) because external / internal function calls are briefly described in visibility descriptions (e.g. in `external` function description) and (2) because of `Because of that, there are four types of visibility for functions and state variables.` sentence - it is not obvious why two call types produce four visibility types.

But looking with a fresh eye I agree that we should provide some note on external / internal function calls. It'll be useful for newcomers. I see at least two options here: (1) we can leave only the first sentence or (2) we can provide more information to explain how two call types produce four visibility types.

chriseth: Yeah, maybe the `because` can be changed to something like "connected to this".  I think there should at least be an introductory sentence that says something like there are four visibilities and they are connected to the way functions are called.

nikitastupin: Thank you for the reply 😃 I'll continue to work on the PR after I'll return back home from a vacation

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
+
+``private``
+    Private state variables are only visible for the contract
+    they are defined in and not in derived contracts.
 
-Functions have to be specified as being ``external``,
-``public``, ``internal`` or ``private``.
-For state variables, ``external`` is not possible.
+Functions Visibility
+====================
```
cameel:
```suggestion
Function Visibility
===================
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
-functions and state variables.
+.. note::
+    Everything that is inside a contract is visible to
+    all observers external to the blockchain. Making something ``private``
+    only prevents other contracts from reading or modifying
+    the information, but it will still be visible to the
+    whole world outside of the blockchain.
+
+State Variables Visibility
+==========================
```
cameel:
```suggestion
State Variable Visibility
=========================
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
+
+``public``
+    For public state variables, an automatic
+    :ref:`getter function<getter-functions>` is generated.
```
cameel:
```suggestion
    Public state variables differ from internal ones only in that the compiler automatically generates
    :ref:`getter functions<getter-functions>` for them, which allows other contracts to read their values.
    When used within the same contract, the external access (e.g. ``this.x``) invokes the getter
    while internal access (e.g. ``x``) gets the variable value directly from storage.
    Setter functions are not generated so other contracts cannot directly modify their values.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
+
+``internal``
+    Internal state variables can only be
+    accessed from within the current contract
+    or contracts deriving from it, without using ``this``.
+    This is the default visibility level for state variables.
```
cameel:
```suggestion
    Internal state variables can only be accessed from within the contract where they are defined in
    and in derived contracts.
    They cannot be accessed externally.
    This is the default visibility level for state variables.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
-ones that do not create an actual EVM call (also called
-a "message call") and external
-ones that do. Because of that, there are four types of visibility for
-functions and state variables.
+.. note::
+    Everything that is inside a contract is visible to
+    all observers external to the blockchain. Making something ``private``
+    only prevents other contracts from reading or modifying
+    the information, but it will still be visible to the
+    whole world outside of the blockchain.
```
cameel: It seems odd to me to put this before listing all the visibility options. I'd put it after `private` description in one of the sections below.

docs/contracts/visibility-and-getters.rst
```diff
@@ -6,15 +6,32 @@
...
+
+``internal``
+    Internal state variables can only be
+    accessed from within the current contract
+    or contracts deriving from it, without using ``this``.
+    This is the default visibility level for state variables.
+
+``private``
+    Private state variables are only visible for the contract
+    they are defined in and not in derived contracts.
```
cameel:
```suggestion
    Private state variables are only visible for the contract they are defined in and not in derived contracts.
    They cannot be accessed externally.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -24,27 +41,15 @@ For state variables, ``external`` is not possible.
...
-    function (see below) is generated.
+    and can be either called internally or via message calls.
 
 ``internal``
-    Those functions and state variables can only be
-    accessed internally (i.e. from within the current contract
-    or contracts deriving from it), without using ``this``.
-    This is the default visibility level for state variables.
+    Internal functions can only be accessed from within the current contract
+    or contracts deriving from it, without using ``this``.
```
cameel:
```suggestion
    or contracts deriving from it.
    They cannot be accessed externally.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -24,27 +41,15 @@ For state variables, ``external`` is not possible.
...
-    derived contracts.
-
-.. note::
-    Everything that is inside a contract is visible to
-    all observers external to the blockchain. Making something ``private``
-    only prevents other contracts from reading or modifying
-    the information, but it will still be visible to the
-    whole world outside of the blockchain.
+    Private functions are only visible for the contract
+    they are defined in and not in derived contracts.
```
cameel:
```suggestion
    Private functions are only visible for the contract they are defined in and not in derived contracts.
    They cannot be accessed externally.
```

docs/contracts/visibility-and-getters.rst
nikitastupin: I agree that putting the note after `private` description looks natural (after all, the note introduces the caveat so one should learn the concept first and then the caveat). However, the reason why I decided to put it before is the note refers to both state variables and functions. I'm afraid that if we put it after `private` description one may think that the note refers only to state variables or only to functions. What are your thoughts on this concern?

docs/contracts/visibility-and-getters.rst
nikitastupin: I've returned the paragraph and rephrased it a bit at 1fed766.

docs/contracts/visibility-and-getters.rst
chriseth: Maybe we can shrink this note a bit (i.e. only the first sentence) and just put it at the end of both "private" paragraphs?

docs/contracts/visibility-and-getters.rst
```diff
@@ -24,27 +46,16 @@ For state variables, ``external`` is not possible.
...
+    and can be either called internally or via message calls.
 
 ``internal``
-    Those functions and state variables can only be
-    accessed internally (i.e. from within the current contract
-    or contracts deriving from it), without using ``this``.
-    This is the default visibility level for state variables.
+    Internal functions can only be accessed from within the current contract
+    or contracts deriving from it.
+    They cannot be accessed externally.
```
chriseth, line 53: Maybe we can add here: "They can use internal types like mappings as parameters."?

docs/contracts/visibility-and-getters.rst
nikitastupin: Sound good, here are changes c5a5c076e

nikitastupin:
I think it's ready to be considered to merge 😃

docs/contracts/visibility-and-getters.rst
```diff
@@ -1,18 +1,13 @@
 .. index:: ! visibility, external, public, private, internal
 
+.. |visibility-caveat| replace:: Making something ``private`` or ``internal`` only prevents other contracts from reading or modifying the information, but it will still be visible to the whole world outside of the blockchain.
```
cameel, line 3: Nice. I didn't even know that RST allowed something like that!

docs/contracts/visibility-and-getters.rst
```diff
@@ -1,20 +1,40 @@
...
+.. warning::
+    |visibility-caveat|
 
-Functions have to be specified as being ``external``,
-``public``, ``internal`` or ``private``.
-For state variables, ``external`` is not possible.
+Function Visibility
+===================
+
+Solidity knows two kinds of function calls: external ones that do create an actual EVM message call and internal ones that do not. Furthermore, functions can be inherited or not inherited by derived contracts. This gives rise to four types of visibility for functions.
```
cameel: This line is a whole paragraph. I'd suggest wrapping at the end of each sentence.
```suggestion
Solidity knows two kinds of function calls: external ones that do create an actual EVM message call and internal ones that do not.
Furthermore, functions can be inherited or not inherited by derived contracts.
This gives rise to four types of visibility for functions.
```

cameel: Technically, private functions are still inherited (their code is present in the derived contract if they are used). They just can't be called directly from the derived contract.

Also the wording suggests to me that you can have 4 combinations (external inherited, external not inherited, internal inherited, internal not inherited) but it's not exactly like that.

How about this?
```suggestion
Solidity knows two kinds of function calls: external ones that do create an actual EVM message call and internal ones that do not.
Furthermore, internal functions can be made inaccessible to derived contracts.
This gives rise to four types of visibility for functions.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -1,20 +1,40 @@
...
+
+``public``
+    Public state variables differ from internal ones only in that the compiler automatically generates
+    :ref:`getter functions<getter-functions>` for them, which allows other contracts to read their values.
+    When used within the same contract, the external access (e.g. ``this.x``) invokes the getter
+    while internal access (e.g. ``x``) gets the variable value directly from storage.
+    Setter functions are not generated so other contracts cannot directly modify their values.
+
+``internal``
+    Internal state variables can only be accessed from within the contract where they are defined in
```
cameel:
```suggestion
    Internal state variables can only be accessed from within the contract they are defined in
```

docs/contracts/visibility-and-getters.rst
cameel: How about this?
```suggestion
    They cannot be accessed externally.
    Since they are not exposed to the outside through the contract's ABI, they can take parameters of internal types like mappings or storage references.
```

docs/contracts/visibility-and-getters.rst
```diff
@@ -24,27 +44,20 @@ For state variables, ``external`` is not possible.
...
-    derived contracts.
-
-.. note::
-    Everything that is inside a contract is visible to
-    all observers external to the blockchain. Making something ``private``
-    only prevents other contracts from reading or modifying
-    the information, but it will still be visible to the
-    whole world outside of the blockchain.
+    Private functions are only visible for the contract they are defined in and not in derived contracts.
+    They cannot be accessed externally.
```
cameel: Update to match https://github.com/ethereum/solidity/pull/11999/files#r744680506.
```suggestion
    They cannot be accessed externally and can use internal types for parameters.
```
Maybe it would be better to just say that they're like ``internal`` ones but not visible in derived contracts to avoid repeating stuff.

docs/contracts/visibility-and-getters.rst
cameel: We should probably also mention this for state variables. For example you cannot define this:
```solidity
contract C {
    function() internal public x;
}
```
but this is legal:
```solidity
contract C {
    function() external public y;
}
```
It's just that mappings do not have that problem because getters skip them and storage references are not allowed as storage variables.

cameel (commented):
I think this is pretty good already but I still have some change suggestions.

cameel:
Looks like `b_osx` is failing, but it's unrelated to the PR. Some problem with an expired SSL cert when downloading gdbm.

nikitastupin:
Thank you very much for the feedback and suggestions! I've incorporated almost all of them in 60651e6 and ff26239.

nikitastupin:
Hi there! Is there anything needed from my side to get this pull request merged? 🤔

cameel (approved):
Sorry to keep you waiting. I reviewed it now and looks fine the way it is.

Squashed the review fixes into one commit, rebased on current `develop` and pushed to your branch. If it passes tests, it'll get merged.

Thanks for your contribution (and your patience :))!

nikitastupin:
No worries! Thank you for your feedback and efforts to merge this 🙏 

Looking forward to make new contributions 🚀

PR #12002, opened 2021-09-21
Boolean LP Solver.

chriseth:
I extracted the lp solver from the fully-fledged smt solver interface, in the hope that we can merge this without it being actively used in the optimizer but nevertheless well tested.

### DIFF
--- a/libsolutil/BooleanLP.cpp
+++ b/libsolutil/BooleanLP.cpp
@@ -247,8 +247,9 @@ void BooleanLPSolver::addAssertion(Expression const& _expr)
 
 pair<CheckResult, vector<string>> BooleanLPSolver::check(vector<Expression> const& _expressionsToEvaluate)
 {
-	cout << "Solving boolean constraint system" << endl;
-	cout << toString() << endl;
+//	cout << "Solving boolean constraint system" << endl;
+//	cout << toString() << endl;
+//	cout << "--------------" << endl;
 
 	if (m_state.back().infeasible)
 		return make_pair(CheckResult::UNSATISFIABLE, vector<string>{});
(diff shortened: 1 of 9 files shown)


libsolutil/BooleanLP.cpp
```diff
@@ -0,0 +1,700 @@
...
+	return index;
+}
+
+void BooleanLPSolver::addBooleanEquality(Literal const& _left, smtutil::Expression const& _right)
+{
+	if (optional<Literal> right = parseLiteral(_right))
+	{
+		 // includes: not, <=, <, >=, >, boolean variables.
+		 // a = b <=> (-a \/ b) /\ (a \/ -b)
+		 Literal negLeft = negate(_left);
```
stackenbotten: Coding style error

libsolutil/BooleanLP.cpp
```diff
@@ -0,0 +1,700 @@
...
+		 Literal negRight = negate(*right);
```
stackenbotten: Coding style error

libsolutil/BooleanLP.cpp
```diff
@@ -0,0 +1,700 @@
...
+		 m_state.back().clauses.emplace_back(Clause{vector<Literal>{negLeft, *right}});
```
stackenbotten: Coding style error

libsolutil/BooleanLP.cpp
```diff
@@ -0,0 +1,700 @@
...
+		 m_state.back().clauses.emplace_back(Clause{vector<Literal>{_left, negRight}});
```
stackenbotten: Coding style error

libsolutil/LP.cpp
```diff
@@ -0,0 +1,746 @@
...
+
+
+bool Constraint::operator<(Constraint const& _other) const
+{
+	if (equality != _other.equality)
+		return equality < _other.equality;
+
+	for (size_t i = 0; i < max(data.size(), _other.data.size()); ++i)
+	{
+		 rational const& a = get(data, i);
```
stackenbotten: Coding style error

libsolutil/LP.cpp
```diff
@@ -0,0 +1,746 @@
...
+		 rational const& b = get(_other.data, i);
```
stackenbotten: Coding style error

libsolutil/LP.cpp
```diff
@@ -0,0 +1,746 @@
...
+		 if (a != b)
```
stackenbotten: Coding style error

libsolutil/LP.cpp
```diff
@@ -0,0 +1,746 @@
...
+	return false;
+}
+
+bool Constraint::operator==(Constraint const& _other) const
+{
+	if (equality != _other.equality)
+		return false;
+
+	for (size_t i = 0; i < max(data.size(), _other.data.size()); ++i)
+		 if (get(data, i) != get(_other.data, i))
```
stackenbotten: Coding style error

chriseth:
Re-structured the code a bit, still needs some cleanup and more tests.

chriseth:
We should check how disabling models in the LP solver affects performance.

libsolutil/BooleanLP.cpp
```diff
@@ -0,0 +1,723 @@
...
+	map<size_t, bool> assignments;
+	while (!clauses.empty() && !ranges::all_of(clauses, [](Clause const& _c) { return _c.literals.size() > 1; }))
+	{
+		// TODO this assumes that no clause contains more than one constraint
+		// TODO add an assertion for that.
+
+		for (Clause const& c: clauses)
+			if (c.literals.empty())
+				return {false, {}};
+			else if (c.literals.size() == 1)
```
msooseth, line 60: Modern CDCL solvers nowadays immediately clean & propagate 1-long clauses as soon as they are added to the solver. Hence, this cleaning is only needed once, when the clause is added. Then, the invariant always holds that clauses in the DB either (1) have at least 2 literals that are unassigned, or (2) the clause is satisfied, or (3) the clause is conflicting and the conflict is being handled. The only (4)th state is that the whole solver is in UNSAT state.

Of course this is not something that we must follow, it's just a convention in this space. The reason this convention came about is that it can become extremely complicated if the above explained invariant is not holding during various phases of the solving procedure.

chriseth: Yeah, that seems to be a useful invariant!
What is also interesting about this solver is the comment above: Any clause has at most one constraint. I currently don't fully understand why but it might have to do with the impossibility to negate them otherwise. Is that anything we could exploit?

msooseth: Sorry, I'm a bit confused regarding a clause containing only one constraint. In SAT literature (and this is a convention, again, so up to us how we follow it), we call a constraint made up of a disjunction of literals a "clause". So within the framework of this convention, clause=constraint, and it's a 1-to-1 relationship. There can be other types of constraints, such as an XOR constraint, a PB constraint, etc.

We usually don't group constraints together to a higher-level notion in a SAT solver. In an SMT solver sometimes a higher-level constraint is blasted into a set of independent constraints expressible in the language of the solver (e.g. PB constraints, or clauses, etc), and so a grouping may make sense. However, from the perspective of the SAT (or PB) solver, these constraints are independent, which gives flexibility (and also a certain form of "blindness" to the original problem).

NOTE: PB = Pseudo-Boolean constraint, i.e. "sum(a_i * x_i) >= N", where a_i are integers, x_i are literals (i.e. either 0 or 1), N is an integer, and both + and * are the standard addition/multiplication over (infinite) integers.

chriseth:
Superseded by #12711

PR #12003, opened 2021-09-21
[Trivial] Correct wrong error message referencing `.slot` and `.offset` when `.length` was used

Marenz:


Changelog.md
```diff
@@ -19,16 +19,17 @@ Compiler Features:
...
  * Name Resolver: Fix that when importing an aliased symbol using ``import {AliasedName} from "a.sol"`` it would use the original name of the symbol and not the aliased one.
+ * Opcode Optimizer: Prevent the optimizer from running multiple times to avoid potential bytecode differences for referenced code.
  * Parser: Properly check for multiple SPDX license identifiers next to each other and validate them.
+ * SMTChecker: Fix BMC's constraints regarding internal functions.
  * SMTChecker: Fix false negative caused by ``push`` on storage array references returned by internal functions.
  * SMTChecker: Fix false positive in external calls from constructors.
  * SMTChecker: Fix internal error on some multi-source uses of ``abi.*``, cryptographic functions and constants.
- * SMTChecker: Fix BMC's constraints regarding internal functions.
  * Standard JSON: Fix non-fatal errors in Yul mode being discarded if followed by a fatal error.
+ * Type Checker: Correct wrong error message in inline assembly complaining about ``.slot`` or ``.offset` not valid when actually ``.length` was used.
```
```suggestion
 * Type Checker: Correct wrong error message in inline assembly complaining about ``.slot`` or ``.offset`` not valid when actually ``.length` was used.
```

test/libsolidity/syntaxTests/inlineAssembly/invalid/constant_length_access.sol
```diff
@@ -0,0 +1,10 @@
+contract Test {
+	uint constant x = 2;
+	function f() public pure {
+		assembly {
+			let y := x.length
+		}
+	}
+}
+// ----
+// TypeError 4656: (91-99): State variables only support ".slot" and ".offset".
```
chriseth: This is better, but the message is still wrong: A constant variable is not really a state variable.

chriseth (changes requested):

chriseth (approved):

PR #12004, opened 2021-09-22
Disambiguate bytesRequired

chriseth:
There are three functions of the same name which do different things and it is easy to call the wrong one because one of them is a template.

### DIFF
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -56,7 +56,7 @@ AssemblyItem const& Assembly::append(AssemblyItem const& _i)
 	return m_items.back();
 }
 
-unsigned Assembly::bytesRequired(unsigned subTagSize) const
+unsigned Assembly::codeSize(unsigned subTagSize) const
 {
 	for (unsigned tagSize = subTagSize; true; ++tagSize)
 	{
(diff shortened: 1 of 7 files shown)


libevmasm/Assembly.cpp
```diff
@@ -56,7 +56,7 @@ AssemblyItem const& Assembly::append(AssemblyItem const& _i)
 	return m_items.back();
 }
 
-unsigned Assembly::bytesRequired(unsigned subTagSize) const
+unsigned Assembly::codeSize(unsigned subTagSize) const
```
christianparpart, line 59: Looking at the body of this function, I'd like to understand what *exactly* this body is doing.

It seems that `subTagSixe` looks like a byte offset into `bytecode` .  I am not sure why the condition-less for-loop will not never break out.

libevmasm/Assembly.cpp
```diff
@@ -66,7 +66,7 @@ unsigned Assembly::bytesRequired(unsigned subTagSize) const
 
 		for (AssemblyItem const& i: m_items)
 			ret += i.bytesRequired(tagSize);
-		if (util::bytesRequired(ret) <= tagSize)
+		if (util::numberEncodingSize(ret) <= tagSize)
```
christianparpart, line 69: (subjective: `util::requiredBytesForNumber(x)` would not require me to jump-to-definition to understand what this name means?)

christianparpart (commented):
with reviewing, only a view questions. :)

libevmasm/Assembly.cpp
chriseth: it is the number of bytes used for tags in the sub-assemblies.

hrkrshnn (approved):

Issue #11043, opened 2021-03-03
Compiling in Yul mode doesn't produce sources output


haltman-at:
## Description

When you use solc to compile Yul, you get `contracts` output but no `sources` output (with source IDs and ASTs).

Now, strictly speaking, one could say that, since Yul only supports compiling individual files (at least for now), source IDs aren't necessary.  But having this information would allow treating the compiler output in a more uniform manner without having to write special cases.

Also, it would be nice to have ASTs.  These exist when Yul is used in an assembly block inside Solidity, and in `generatedSources`, but we don't get them here.  It would be nice to have these for debugging purposes... e.g., Truffle Debugger relies on these to track variables.  It's been able to track Yul variables for some time now, but it wouldn't be able to do it for standalone Yul if there are no ASTs.

So, I would like to see `sources` added to the output when using Yul.

Context: I was working on finally adding [Yul support into Truffle](https://github.com/trufflesuite/truffle/issues/1906) and this came up as one of the problems I encountered.  There will be one more of these issues in a bit. :)

## Environment

- Compiler version: 0.8.1

axic:
Since we do prepare an AST of inline assembly for the Solidity AST output, technically it should be possible to create a similar JSON output for Yul. We are lacking it right now and given we have seen some reluctance changing the Solidity AST, I would be careful making this output well "designed".

As a side-note I am not sure it currently would cover Yul objects.

Do you depend on this for Yul support? Would it be okay if we say this feature is planned in the medium-term?

haltman-at:
Well, it depends on how much support you want and how much we're willing to special-case things to get it to work.

So we can distinguish between multiple things here: The presence of `sources` at all (along with `id`), and the presence of `ast` in those sources.  (So possibly I should split this into two separate issues?)

So, let's start with `sources` and `id`.  Is this *necessary* for support?  Hm... well, I thought we needed it to avoid having to mess up our code a bunch, but on checking I don't think that's the case and I think we can handle it without doing that.  It would still be nice but I think things should still work without it.  Just to check -- all versions of Yul include source maps, right?  If not I'll have to make some additional alterations.

As for how necessary are ASTs, well, this is where the question comes up of, what do you mean by support?  They're not necessary for compilation or most other purposes; they only really affect the debugger.  And (thanks to some recent changes made to support Vyper) they should no longer even be necessary for stepping through with the debugger.  However they will be necessary if you want to be able to see the values of assembly variables with the debugger.  That's basically what we'd need them for.

[...]

PR #12005, opened 2021-09-22
`"origin"` locations in Yul AST

cameel:
Originally requested in https://github.com/ethereum/solidity/pull/11998#discussion_r713163128.
Depends on #11998.
Related to #11043.

- [ ] Problem: the location does not always follow the format described by our docs on [Source Mappings](https://docs.soliditylang.org/en/latest/internals/source_mappings.html).

### DIFF
--- a/libyul/AsmJsonConverter.cpp
+++ b/libyul/AsmJsonConverter.cpp
@@ -28,13 +28,14 @@
 
 using namespace std;
 using namespace solidity::langutil;
+using namespace solidity::yul;
 
 namespace solidity::yul
 {
 
 Json::Value AsmJsonConverter::operator()(Block const& _node) const
 {
-	Json::Value ret = createAstNode(nativeLocationOf(_node), "YulBlock");
+	Json::Value ret = createAstNode(debugDataOf(_node).get(), "YulBlock");
 	ret["statements"] = vectorOfVariantsToJson(_node.statements);
 	return ret;
 }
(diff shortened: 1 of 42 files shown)


libyul/AsmJsonConverter.cpp
```diff
@@ -170,27 +172,55 @@ Json::Value AsmJsonConverter::operator()(ForLoop const& _node) const
...
-	ret["src"] = to_string(_location.start) + ":" + to_string(length) + ":" + (m_sourceIndex.has_value() ? to_string(m_sourceIndex.value()) : "-1");
+
+	ret["src"] = formatLocationWithFileIndex(_debugData->nativeLocation, m_sourceIndex);
+	if (_debugData->originLocation.isValid())
+	{
+		if (_debugData->nativeLocation.sourceName == _debugData->originLocation.sourceName)
+			ret["origin"] = formatLocationWithFileIndex(_debugData->originLocation, m_sourceIndex);
+		else
+			ret["origin"] = formatLocationWithFileName(_debugData->originLocation);
+	}
```
cameel, line 222: The problem with this PR is that we don't have indexes for the original Solidity files when compiling Yul.

This does work because currently we do not support outputting Yul AST (#9590) but will break when we implement it so I don't think this is a good solution.

chriseth:
I think we need to provide the source-name-to-index-mapping to the asm json converter in the same way we do it for the ASTJsonConverter - does this not work?

cameel:
But that will work only if we have access to Solidity sources. When we're importing inline assembly AST we do, but this is the case where `"origin"` is superfluous anyway because it matches `"src"`. `"origin"` is only useful if we allow importing AST of a pure-Yul file and in that case we don't.

leonardoalt:
> But that will work only if we have access to Solidity sources. When we're importing inline assembly AST we do, but this is the case where `"origin"` is superfluous anyway because it matches `"src"`. `"origin"` is only useful if we allow importing AST of a pure-Yul file and in that case we don't.

@cameel was this solved now?

cameel:
Not really. I think this PR does not really make sense without implementing Yul AST output first (#9590). These origin locations will never show up for inline assembly.

Originally I wanted to implement #9590 and then put this PR on top of that. But I got into other PRs and didn't really have time for this one. I could close it but @chriseth seems to think this should actually work. I didn't want to close it without clarifying that first.

cameel:
I'm going to close this. I definitely won't be working on this in the near future and like I said above, I'm not even sure if this PR even makes sense - it's only needed for Yul ASTs and we don't support printing those at this point.

PR #12007, opened 2021-09-22
`--include-path` option (revived)

cameel:
This is the same as #11848 but that PR got accidentally merged into its base branch.

> Implements the `solc` part of #11409.
> `solc-js` part is in https://github.com/ethereum/solc-js/pull/544.
> Depends on #11688.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,8 +7,9 @@ Language Features:
 
 
 Compiler Features:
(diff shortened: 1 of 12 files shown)


chriseth:
:zombie:

chriseth (approved):

PR #12013, opened 2021-09-23
Rational LP Solver.

chriseth:


libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+#include <optional>
+#include <stack>
+
+
+namespace solidity::util
+{
+
+using rational = boost::rational<bigint>;
+
+inline std::string toString(rational const& _value, size_t _paddedLength = 2)
```
leonardoalt: `solutil::formatNumberReadable` does something similar, might wanna check if it does what you want

libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+{
+	if (_data.size() <= _index)
+		_data.resize(_index + 1);
+	_data[_index] = std::move(_value);
+}
+
+inline std::vector<rational>& operator/=(std::vector<rational>& _data, rational const& _divisor)
+{
+	for (rational& x: _data)
+		if (x.numerator())
```
leonardoalt: Why the condition?

libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+			x *= _factor;
+	return _data;
+}
+
+inline std::vector<rational> operator-(std::vector<rational> const& _x, std::vector<rational> const& _y)
+{
+	std::vector<rational> result;
+	for (size_t i = 0; i < std::max(_x.size(), _y.size()); ++i)
+		result.emplace_back(get(_x, i) - get(_y, i));
```
leonardoalt: So you align the vectors left here?

libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+inline std::vector<rational>& operator-=(std::vector<rational>& _x, std::vector<rational> const& _y)
+{
+	solAssert(_x.size() == _y.size(), "");
+	for (size_t i = 0; i < _x.size(); ++i)
+		if (_y[i].numerator())
+			_x[i] -= _y[i];
+	return _x;
+}
+
+inline std::vector<rational> add(std::vector<rational> const& _x, std::vector<rational> const& _y)
```
leonardoalt: Why `add` and not `operator+` following the others?

libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+	return result;
+}
+
+inline bool isConstant(std::vector<rational> const& _x)
+{
+	return ranges::all_of(_x | ranges::views::tail, [](rational const& _v) { return _v == 0; });
+}
+
+/// Multiply two vectors where the first element of each vector is a constant factor.
+/// Only works if at most one of the vector has a nonzero element after the first.
```
leonardoalt:
```suggestion
/// Only works if at most one of the vectors has a nonzero element after the first.
```

libsolutil/RationalVectors.h
```diff
@@ -0,0 +1,162 @@
...
+)
+{
+	if (!_x || !_y)
+		return std::nullopt;
+	if (!isConstant(*_y))
+		swap(_x, _y);
+	if (!isConstant(*_y))
+		return std::nullopt;
+
+	rational factor = _y->front();
```
leonardoalt: Why do you only use the first element of `y`?

libsolutil/LP.h
```diff
@@ -0,0 +1,88 @@
...
+	bool operator==(Constraint const& _other) const;
+};
+
+/**
+ * State used when solving an LP problem.
+ */
+struct SolvingState
+{
+	/// Names of variables, the index zero should be left empty.
+	/// TODO can we change that?
```
leonardoalt: Use a wrapper class around vector that asserts the read/written index is never 0?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+	void addEQConstraint(vector<rational> _lhs, vector<rational> _rhs)
+	{
+		_lhs = _lhs - _rhs;
+		_lhs[0] = -_lhs[0];
+		m_solvingState.constraints.push_back({move(_lhs), true});
+	}
+
+	void addLowerBound(string _variable, rational _value)
+	{
+		size_t index = variableIndex(_variable);
```
leonardoalt, line 77: Move?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+		if (index >= m_solvingState.bounds.size())
+			m_solvingState.bounds.resize(index + 1);
+		m_solvingState.bounds.at(index)[0] = _value;
+	}
+
+	void addUpperBound(string _variable, rational _value)
+	{
+		size_t index = variableIndex(_variable);
```
leonardoalt, line 86: Move?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+		_lhs[0] = -_lhs[0];
+		m_solvingState.constraints.push_back({move(_lhs), true});
+	}
+
+	void addLowerBound(string _variable, rational _value)
+	{
+		size_t index = variableIndex(_variable);
+		if (index >= m_solvingState.bounds.size())
+			m_solvingState.bounds.resize(index + 1);
+		m_solvingState.bounds.at(index)[0] = _value;
```
leonardoalt: Move?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	void addUpperBound(string _variable, rational _value)
+	{
+		size_t index = variableIndex(_variable);
+		if (index >= m_solvingState.bounds.size())
+			m_solvingState.bounds.resize(index + 1);
+		m_solvingState.bounds.at(index)[1] = _value;
```
leonardoalt: Move?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	LPSolver m_solver;
+	SolvingState m_solvingState;
+};
+
+
+BOOST_FIXTURE_TEST_SUITE(LP, LPTestFramework, *boost::unit_test::label("nooptions"))
+
+BOOST_AUTO_TEST_CASE(basic)
```
leonardoalt, line 134: Does each test reset `m_solver` and `m_solvingState`?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,313 @@
...
+	addLowerBound("x", 200);
+	addUpperBound("x", 250);
+	addLowerBound("y", 2);
+	addUpperBound("y", 3);
+	feasible({{"x", 250}, {"y", 3}});
+
+	addLEConstraint(variable("y"), variable("x"));
+	feasible({{"x", 250}, {"y", 3}});
+
+	addEQConstraint(add(variable("y"), constant(231)), variable("x"));
```
leonardoalt: Could also overload the relation operators and do just `add(var("y") + 231 >= var("x")`

leonardoalt (commented):
Still need to look at LP.cpp itself.

libsolutil/RationalVectors.h
chriseth: Optimization: A lot of entries in the vector will be zero, but the rational class does not optimize for it.

libsolutil/RationalVectors.h
chriseth: No, alignment is not changed. `get` just returns zero if you access beyond the length.

libsolutil/RationalVectors.h
chriseth: Because `operator+` is already defined on vectors :(

libsolutil/RationalVectors.h
chriseth: This functions implements multiplications of linear expressions. The first element of the vector is constant, the others are factors of variables, i.e. `_x[0] + _x[1] * X1 + _x[2] * X2 + ...`. Since the result also has to be of the same form, this only works if at most one of `_x` and `_y` is not a constant.

If the condition is met, then we kno that at this point in the code, `_y` is a constant, so we can disregard any elements apart from its first.

test/libsolutil/LP.cpp
chriseth: It uses a const ref

test/libsolutil/LP.cpp
chriseth: It's test code...

test/libsolutil/LP.cpp
chriseth: It's test code...

test/libsolutil/LP.cpp
chriseth: It takes a const ref.

test/libsolutil/LP.cpp
chriseth: Yes, the fixture is newly constructed for each test case.

test/libsolutil/LP.cpp
chriseth: That is done in the next layer of the solver.

Aivolt1 (approved):

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	LinearExpression constant(rational _value)
+	{
+		return LinearExpression::factorForVariable(0, _value);
+	}
+
+	LinearExpression variable(string const& _name)
+	{
+		return LinearExpression::factorForVariable(variableIndex(_name), 1);
```
bshastry, line 51: Note to self: This is `1 * X_i` or `X_i`

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	/// Adds the constraint "_lhs <= _rhs".
+	void addLEConstraint(LinearExpression _lhs, LinearExpression _rhs)
+	{
+		_lhs -= _rhs;
+		_lhs[0] = -_lhs[0];
```
bshastry, line 58: Why is the LHS constant being negated?

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+		m_solvingState.constraints.push_back({move(_lhs), false});
```
bshastry: How does one interpret this constraint? Why is the linear expression `_lhs` false?

bshastry (commented):
Looks reasonable to me. What is missing is a constraint-input-format and its parser. The fuzzer would then generate inputs in the format that would be parsed and checked by the underlying solver.

test/libsolutil/LP.cpp
bshastry: Please ignore my previous question. Answer: It is false because the constraint may be of two types, namely, equality (`true`) and leq (`false`)

chriseth: The `false` just means "not an equality constraint" and thus a `<=` constraint.

test/libsolutil/LP.cpp
chriseth: We are constructing the constraint `_lhs <= _rhs` here. Inside generic expressions, the constant factor is treated in the same way as variables, but constraints are of the form `a_1 * x_1 + ... + a_n * x_n <= b`, so we negate the constant term in order to move it to the other side of the equation.

bshastry: Isn't the negation of `b` happening on the previous line i.e., `_lhs -= rhs;`?

chriseth: No, that takes the rhs over to the other side. We start out with:
```
a * x + b <= c * y + d
```
(``_lhs <= _rhs``)

and we want to end up with
```
a * x - c * y <= d - b
```
So you could say thayt `_lhs -= _rhs` pulls everything over to the left and then `_lhs[0] = -_lhs[0]` pulls the constant term back to the right.

bshastry: `_lhs -= _rhs` applies the `-=` operator on the constant (factor 0 in the linear expression convention) as well right? So I still don't understand the need for `_lhs[0] = -lhs[0]` since the constraint is an LHS only expression i.e., of the form 

```
a*X1 + b*X2 + c <= 0
```

bshastry: Ah I see the source of my confusion:

Linear expressions are of the form `a*X1 + b*X2 + c` i.e., factor[0] is constant
But constraints are of the form `a*X1 + b*X2 <= c` i.e., factor[0] is on the RHS not LHS

Could we please stick to one convention i.e., make constraints also `a*X1 + b*X2 + c <= 0`?

chriseth: @ekpyron also complained about that. Maybe one solution would be to use a different class for constraints, since you cannot multiply two constraints, for example.

libsolutil/LP.cpp
```diff
@@ -0,0 +1,806 @@
...
+	_tableau.objective.resize(columns);
+	_tableau.objective.resize(columns + rows, rational{-1});
+
+	// This sets the objective factors of the slack variables
+	// to zero (and thus selects a basic feasible solution).
+	selectLastVectorsAsBasis(_tableau);
+
+	LPResult result;
+	tie(result, _tableau) = simplexEq(move(_tableau));
+	solAssert(result == LPResult::Feasible || result == LPResult::Unbounded, "");
```
bshastry: This constraint input: https://gist.github.com/bshastry/f37681fea0447356d1dcf000d18413ed

throws here

bshastry: Minimised input: https://gist.github.com/bshastry/b3a67e44b40b69aaaa1540ab0c96c7ef

where each line is to be parsed as `constraintType, rhsConstant, coefficients...`

chriseth: oh wow! Of course the result can also be `Unknown` - simply when it takes too long to find the solution...

bshastry:
```suggestion
	solAssert(result != LPResult::Infeasible, "");
```

libsolutil/LP.cpp
```diff
@@ -0,0 +1,806 @@
...
+		tie(result, tableau) = simplexPhaseI(move(tableau));
+		if (result == LPResult::Infeasible || result == LPResult::Unknown)
+			return make_pair(result, vector<rational>{});
+		solAssert(result == LPResult::Feasible, "");
+	}
+	// We know that the system is satisfiable and we know a solution,
+	// but it might not be optimal.
+	LPResult result;
+	tie(result, tableau) = simplexEq(move(tableau));
+	solAssert(result == LPResult::Feasible || result == LPResult::Unbounded, "");
```
bshastry:
```suggestion
	solAssert(result != LPResult::Infeasible, "");
```

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+	addLEConstraint((x + y) + z, 2);
+	feasible({{"x", 2}, {"y", 0}, {"z", 0}});
+
+	addLEConstraint(constant(2), (x + y) + z);
+	feasible({{"x", 2}, {"y", 0}, {"z", 0}});
+
+	addEQConstraint(constant(2), (x + y) + z);
+	feasible({{"x", 2}, {"y", 0}, {"z", 0}});
+}
+
```
bshastry, line 322:
```suggestion

BOOST_AUTO_TEST_CASE(fuzzer)
{
        auto x = variable("x", -6);
        auto y = variable("y", -6);
        addEQConstraint(x + y, constant(8));
        sat();
}
```

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+				var + " = "s + ::toString(model.at(var)) + " (expected " + ::toString(value) + ")"
+			);
+	}
+
+	void infeasible()
+	{
+		auto [result, model] = m_solver.check(m_solvingState);
+		BOOST_CHECK(result == LPResult::Infeasible);
+	}
+
```
bshastry, line 112:
```suggestion

        void sat()
        {
                auto [result, model] = m_solver.check(m_solvingState);
                BOOST_REQUIRE(result == LPResult::Feasible);
        }
        
```

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,314 @@
...
+	LinearExpression constant(rational _value)
+	{
+		return LinearExpression::factorForVariable(0, _value);
+	}
+
+	LinearExpression variable(string const& _name)
+	{
+		return LinearExpression::factorForVariable(variableIndex(_name), 1);
+	}
+
```
bshastry, line 53:
```suggestion

	LinearExpression variable(string const& _name, rational _value)
	{
		return LinearExpression::factorForVariable(variableIndex(_name), _value);
	}
	
```

libsolutil/LP.cpp
```diff
@@ -0,0 +1,805 @@
...
+			}
+		}
+		solAssert(varsAdded == varsNeeded);
+	}
+
+	vector<LinearExpression> data;
+	for (Constraint& c: _constraints)
+		data.emplace_back(move(c.data));
+
+	return make_pair(move(data), varsNeeded < _constraints.size());
```
chriseth, line 143: Previously this was `varsNeeded > 0`, and I think this is the fix for the issue @bshastry found.

test/libsolutil/LP.cpp
chriseth: I think it's more intuitive to use `f * variable("x")` instead of `variable("x", f)`.

test/libsolutil/LP.cpp
chriseth: What about changing `feasible` above to make `_solution` optional?

test/libsolutil/LP.cpp
chriseth: I added `all_equality` which is the same test but with correct expectation (after the fix)

test/libsolutil/LP.cpp
chriseth: Actually I think it's better to always require a solution, then we can better check for any nondeterminism.

bshastry:
There is a new bug after applying the fix in https://github.com/ethereum/solidity/pull/12013#discussion_r808171996

Failing test case: https://github.com/ethereum/solidity/pull/12627/commits/aaf73eb50f1b7d912c02a44757d5f235a3d7ea25#diff-ddab3fa3f5b4c33bed92be047ce19440b55e4fd04d8680e1c19326c71f516514R151

libsolutil/LP.cpp
```diff
@@ -0,0 +1,805 @@
...
+	/// The factors of the objective function (first row of the tableau)
+	LinearExpression objective;
+	/// The tableau matrix (equational form).
+	std::vector<LinearExpression> data;
+};
+
+
+/// Adds slack variables to remove non-equality costraints from a set of constraints
+/// and returns the data part of the tableau / constraints.
+/// The second return variable is true if a the original input had any equality constraints.
```
bshastry:
```suggestion
/// The second return variable is true if the original input had any equality constraints.
```

test/libsolutil/LP.cpp
```diff
@@ -0,0 +1,322 @@
...
+BOOST_AUTO_TEST_CASE(equal_constant)
+{
+	auto x = variable("x");
+	auto y = variable("y");
+	addLEConstraint(x, y);
+	addEQConstraint(y, constant(5));
+	feasible({{"x", 5}, {"y", 5}});
+}
+
+BOOST_AUTO_TEST_CASE(equal_constant)
```
bshastry: Is it fine to have identically named boost unit tests?

libsolutil/LP.h
```diff
@@ -0,0 +1,192 @@
...
+ *  - data[1] * x_1 + data[2] * x_2 + ... <= data[0]  (equality == false)
+ *  - data[1] * x_1 + data[2] * x_2 + ...  = data[0]  (equality == true)
+ * The set and order of variables is implied.
+ */
+struct Constraint
+{
+	LinearExpression data;
+	bool equality = false;
+	/// Set of literals the conjunction of which implies this constraint.
+	std::set<size_t> reasons;
```
bshastry:
```suggestion
	std::set<size_t> reasons = {};
```

libsolutil/LP.cpp
```diff
@@ -0,0 +1,855 @@
...
+	return result;
+}
+
+/*
+string toString(LinearExpression const& _expr)
+{
+	vector<string> items;
+	for (auto&& multiplier: _expr)
+		 if (multiplier != 0)
+			 items.emplace_back(::toString(multiplier));
```
stackenbotten: Coding style error

libsolutil/LP.cpp
```diff
@@ -0,0 +1,855 @@
...
+		 else
```
stackenbotten: Coding style error

chriseth:
@bshastry https://github.com/ethereum/solidity/pull/12013/commits/46d8644da3cbaf8fd9ba7a753efd547e726ad3f3 should fix the reported error. The solver was cycling in the simplex routine, and we did not properly forward the "unknown" result.
The fact that it cycled might hit at a more serious underlying problem.

leonardoalt:
Closing in favor of https://github.com/ethereum/solidity/pull/12711

Issue #10358, opened 2020-11-19
Allow accessing stack slots of external function pointers in Solidity assembly


haltman-at:
## Abstract

Yul within Solidity is mostly able to work with Solidity variables.  Local variables mostly refer to an individual stack slot, which Yul is able to handle.  Storage variables are manipulable with `x.slot` and `x.offset`.  (I'll ignore `immutable` variables, which are their own problem.)  However there's the problem of using Solidity variables that take up two stack slots.  There are two sorts of these: calldata variables of dynamically-sized type, and external function pointers.  The former can now be used within assembly due to #10208, with the two stack slots accessible as `x.offset` and `x.length`.  However there's still no way to use the latter.

## Motivation

My suggestion is that, just as #10208 introduced `x.offset` and `x.length`, Solidity variables of `function external` type should have their two stack slots exposed in Yul via `x.address` and `x.selector`.

Note: I discussed this earlier in #8917.

## Specification

If `x` is a Solidity local variable of `function external` type, that is within scope within a give `assembly` (Yul) block, then `x.address` should refer to the lower of its two stack slots, and `x.selector` should refer to the upper of its two stack slots, in accordance with how `x` is stored on the stack.

## Backwards Compatibility

This change would be entirely backwards compatible.

axic:
Seems straightforward, at least for reading, but I'm curious: have you seen anyone use function pointers in the wild?

haltman-at:
I haven't myself, but @gnidan has told me of examples... unfortunately he seems to have lost them!  But hopefully we can find them again...

chriseth:
Implemented in https://github.com/ethereum/solidity/pull/12016

PR #12016, opened 2021-09-23
Add ``.address`` and ``.selector`` in inside assembly for external function pointers

Marenz:
So far read only.
assigning will happen in another PR.

refs #10358

#
#

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ### 0.8.10 (unreleased)
 
 Language Features:
+ * Inline Assembly: Support ``.address`` and ``.selector`` on external function pointers to access their address and function selector.
 
 
 Compiler Features:
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -136,6 +136,27 @@ evaluate to the address of the variable in calldata, not the value itself.
 The variable can also be assigned a new offset, but note that no validation to ensure that
 the variable will not point beyond ``calldatasize()`` is performed.
 
+For external function pointers the address and the function selector can be
+accessed using ``x.address`` and ``x.selector``.
+The selector consists of four right-aligned bytes.
+Both values are can be assigned to. For example:
+
+.. code-block:: solidity
+    :force:
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.10 <0.9.0;
+
+    contract C {
+        // Assigns a new selector and address to the return variable @fun
+        function combineToFunctionPointer(address newAddress, uint newSelector) public pure returns (function() external fun) {
+            assembly {
+                fun.selector := newSelector
+                fun.address  := newAddress
+            }
+        }
+    }
+
 For dynamic calldata arrays, you can access
 their calldata offset (in bytes) and length (number of elements) using ``x.offset`` and ``x.length``.
 Both expressions can also be assigned to, but as for the static case, no validation will be performed
--- a/libsolidity/analysis/ReferencesResolver.cpp
+++ b/libsolidity/analysis/ReferencesResolver.cpp
@@ -219,13 +219,15 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 {
 	solAssert(nativeLocationOf(_identifier) == originLocationOf(_identifier), "");
 
-	static set<string> suffixes{"slot", "offset", "length"};
+	static set<string> suffixes{"slot", "offset", "length", "address", "selector"};
 	string suffix;
 	for (string const& s: suffixes)
 		if (boost::algorithm::ends_with(_identifier.name.str(), "." + s))
 			suffix = s;
 
-	// Could also use `pathFromCurrentScope`, split by '.'
+	// Could also use `pathFromCurrentScope`, split by '.'.
+	// If we do that, suffix should only be set for when it has a special
+	// meaning, not for normal identifierPaths.
 	auto declarations = m_resolver.nameFromCurrentScope(_identifier.name.str());
 	if (!suffix.empty())
 	{
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -828,7 +828,7 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
 			if (!identifierInfo.suffix.empty())
 			{
 				string const& suffix = identifierInfo.suffix;
-				solAssert((set<string>{"offset", "slot", "length"}).count(suffix), "");
+				solAssert((set<string>{"offset", "slot", "length", "selector", "address"}).count(suffix), "");
 				if (!var->isConstant() && (var->isStateVariable() || var->type()->dataStoredIn(DataLocation::Storage)))
 				{
 					if (suffix != "slot" && suffix != "offset")
@@ -858,6 +858,19 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
 					if (suffix != "offset" && suffix != "length")
 					{
 						m_errorReporter.typeError(1536_error, nativeLocationOf(_identifier), "Calldata variables only support \".offset\" and \".length\".");
+						return false;
+					}
+				}
+				else if (auto const* fpType = dynamic_cast<FunctionTypePointer>(var->type()))
+				{
+					if (suffix != "selector" && suffix != "address")
+					{
+						m_errorReporter.typeError(9272_error, nativeLocationOf(_identifier), "Variables of type function pointer only support \".selector\" and \".address\".");
+						return false;
+					}
+					if (fpType->kind() != FunctionType::Kind::External)
+					{
+						m_errorReporter.typeError(8533_error, nativeLocationOf(_identifier), "Only Variables of type external function pointer support \".selector\" and \".address\".");
 						return false;
 					}
 				}
--- a/libsolidity/ast/ASTAnnotations.h
+++ b/libsolidity/ast/ASTAnnotations.h
@@ -211,7 +211,7 @@ struct InlineAssemblyAnnotation: StatementAnnotation
 	struct ExternalIdentifierInfo
 	{
 		Declaration const* declaration = nullptr;
-		/// Suffix used, one of "slot", "offset", "length" or empty.
+		/// Suffix used, one of "slot", "offset", "length", "address", "selector" or empty.
 		std::string suffix;
 		size_t valueSize = size_t(-1);
 	};
--- a/libsolidity/ast/ASTJsonConverter.cpp
+++ b/libsolidity/ast/ASTJsonConverter.cpp
@@ -31,6 +31,7 @@
 
 #include <libsolutil/JSON.h>
 #include <libsolutil/UTF8.h>
+#include <libsolutil/CommonData.h>
 
 #include <boost/algorithm/string/join.hpp>
 
@@ -178,9 +179,12 @@ Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifie
 	tuple["declaration"] = idOrNull(_info.second.declaration);
 	tuple["isSlot"] = Json::Value(_info.second.suffix == "slot");
 	tuple["isOffset"] = Json::Value(_info.second.suffix == "offset");
+
 	if (!_info.second.suffix.empty())
 		tuple["suffix"] = Json::Value(_info.second.suffix);
+
 	tuple["valueSize"] = Json::Value(Json::LargestUInt(_info.second.valueSize));
+
 	return tuple;
 }
 
--- a/libsolidity/codegen/ContractCompiler.cpp
+++ b/libsolidity/codegen/ContractCompiler.cpp
@@ -823,6 +823,16 @@ bool ContractCompiler::visit(InlineAssembly const& _inlineAssembly)
 							if (suffix == "length")
 								stackDiff--;
 						}
+						else if (
+							auto const* functionType = dynamic_cast<FunctionType const*>(variable->type());
+							functionType && functionType->kind() == FunctionType::Kind::External
+						)
+						{
+							solAssert(suffix == "selector" || suffix == "address", "");
+							solAssert(variable->type()->sizeOnStack() == 2, "");
+							if (suffix == "selector")
+								stackDiff--;
+						}
 						else
 							solAssert(false, "");
 					}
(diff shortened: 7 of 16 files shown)


Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 Language Features:
  * Inheritance: A function that overrides only a single interface function does not require the ``override`` specifier.
+ * Inline Assembly: Use ``.address`` and ``.selector`` on external function pointers to access their address and function selector.
```
```suggestion
 * Inline Assembly: Support ``.address`` and ``.selector`` on external function pointers to access their address and function selector.
```

docs/assembly.rst
```diff
@@ -136,6 +136,11 @@ evaluate to the address of the variable in calldata, not the value itself.
 The variable can also be assigned a new offset, but note that no validation to ensure that
 the variable will not point beyond ``calldatasize()`` is performed.
 
+For external function pointers the address and the function selector can be
+accessed using ``x.address`` and ``x.selector``.
+The selector will be left aligned and of type ``uint32``.
```
cameel: At first I read that as if the selector was being left somewhere :)
```suggestion
The selector will be left-aligned and of type ``uint32``.
```

docs/assembly.rst
```diff
@@ -136,6 +136,11 @@ evaluate to the address of the variable in calldata, not the value itself.
...
+Assigning is not possible.
```
cameel: How about this?
```suggestion
Both values are read-only and cannot be assigned.
```

docs/assembly.rst
cameel: If there's no unused space then isn't it technically both left- and right-aligned? Or not aligned at all? I'm not sure if it makes sense to speak of alignment of a 4-byte value in a 4-byte type.

libsolidity/analysis/ReferencesResolver.cpp
```diff
@@ -219,7 +219,7 @@ void ReferencesResolver::operator()(yul::Identifier const& _identifier)
 {
 	solAssert(nativeLocationOf(_identifier) == originLocationOf(_identifier), "");
 
-	static set<string> suffixes{"slot", "offset", "length"};
+	static set<string> suffixes{"slot", "offset", "length", "address", "selector"};
 	string suffix;
```
cameel, line 223:
```suggestion
	string const& suffix;
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -861,6 +861,24 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
 						return false;
 					}
 				}
+				else if (auto const* fpType = dynamic_cast<FunctionTypePointer>(var->type()))
```
cameel, line 864: This `else` does not have an `if`. The `{}` above is just a normal block.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -828,7 +828,7 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
 			if (!identifierInfo.suffix.empty())
 			{
 				string const& suffix = identifierInfo.suffix;
-				solAssert((set<string>{"offset", "slot", "length"}).count(suffix), "");
+				solAssert((set<string>{"offset", "slot", "length", "selector", "address"}).count(suffix), "");
 				if (!var->isConstant() && (var->isStateVariable() || var->type()->dataStoredIn(DataLocation::Storage)))
 				{
 					if (suffix != "slot" && suffix != "offset")
```
cameel, line 834: What about state variables of function types?

Can you add a test case with such a state variable?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -861,6 +861,24 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
 						return false;
 					}
 				}
+				else if (auto const* fpType = dynamic_cast<FunctionTypePointer>(var->type()))
+				{
+					if (suffix != "selector" && suffix != "address")
+					{
+						m_errorReporter.typeError(9272_error, _identifier.debugData->location, "Variables of type function pointer only support \".selector\" and \".address\".");
```
cameel: I think we never oficially refer to these as pointers.
```suggestion
						m_errorReporter.typeError(9272_error, _identifier.debugData->location, "Variables of function types only support \".selector\" and \".address\".");
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -861,6 +861,24 @@ bool TypeChecker::visit(InlineAssembly const& _inlineAssembly)
...
+						return false;
+					}
+					if (fpType->kind() != FunctionType::Kind::External)
+					{
+						m_errorReporter.typeError(8533_error, _identifier.debugData->location, "Only Variables of type external function pointer support \".selector\" and \".address\".");
+						return false;
+					}
+					if (_context == yul::IdentifierContext::LValue)
+					{
+						m_errorReporter.typeError(1259_error, _identifier.debugData->location, "\".selector\" and \".address\" for external function pointers cannot be assigned to.");
```
```suggestion
						m_errorReporter.typeError(1259_error, _identifier.debugData->location, "\".selector\" and \".address\" of external function pointers cannot be assigned to.");
```

libsolidity/ast/ASTAnnotations.h
```diff
@@ -211,7 +211,7 @@ struct InlineAssemblyAnnotation: StatementAnnotation
 	struct ExternalIdentifierInfo
 	{
 		Declaration const* declaration = nullptr;
-		/// Suffix used, one of "slot", "offset", "length" or empty.
+		/// Suffix used, one of "slot", "offset", "length", "address", "selector" or empty.
 		std::string suffix;
```
cameel, line 215: Why isn't this an enum? Maybe we could refactor it into one (in a separate PR of course).

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -178,6 +178,8 @@ Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifie
 	tuple["declaration"] = idOrNull(_info.second.declaration);
 	tuple["isSlot"] = Json::Value(_info.second.suffix == "slot");
 	tuple["isOffset"] = Json::Value(_info.second.suffix == "offset");
+	tuple["isSelector"] = Json::Value(_info.second.suffix == "selector");
+	tuple["isAddress"] = Json::Value(_info.second.suffix == "address");
```
cameel: This adds these values for all identifiers. Shouldn't we limit it only to variables of function types? Or do we intentionally want all these properties present for all identifiers? It seems like it would add a lot of noise though.

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -818,6 +819,19 @@ bool ContractCompiler::visit(InlineAssembly const& _inlineAssembly)
 							if (suffix == "length")
 								stackDiff--;
 						}
+						else if (
+							auto const* functionType = dynamic_cast<FunctionType const*>(variable->type());
+							functionType && functionType->kind() == FunctionType::Kind::External
+						)
```
cameel, line 829: This will not run for storage variables of function types.

test/libsolidity/semanticTests/inlineAssembly/external_function_pointer_selector.sol
```diff
@@ -0,0 +1,22 @@
...
+
+		assembly {
+			selectorValue := fp.selector
+		}
+
+		return uint32(bytes4(bytes32(selectorValue)));
+	}
+	function testSol() public returns (uint32) {
+		return uint32(this.testFunction.selector);
+	}
```
cameel: If `.selector` in Solidity is `uint32` (right-aligned in a 32-byte slot) why do we want it left-aligned in Yul? Wouldn't it be more consistent to have it right-aligned in both cases?

test/libsolidity/syntaxTests/inlineAssembly/invalid/external_function_pointer_selector_assignment.sol
```diff
@@ -0,0 +1,17 @@
...
+
+	function testYul() public {
+		function() external fp = this.testFunction;
+
+		assembly {
+			fp.selector := 0x12345678
+		}
+	}
+	function testSol() public {
+		this.testFunction.selector = bytes4(bytes32(uint256(0x1234568)));
```
cameel: So apparently we allow an explicit conversion from `uint32` to `bytes4` but not from an integer literal that fits in `uint32` to `bytes4`. @chriseth @hrkrshnn is that intentional?

docs/assembly.rst
cameel: Ah, it probably means alignment in the 32-byte slot. I think it should be clarified here. But then I do not see in what sense it's a `uint32` value - the type depends on the type of an underlying variable.

libsolidity/analysis/TypeChecker.cpp
cameel: Ah, sorry, it is an `if` :)

docs/assembly.rst
chriseth: There are no types in assembly, so I would propose to talk about four bytes that are left-aligned or something like that.

test/libsolidity/semanticTests/inlineAssembly/external_function_pointer_selector.sol
chriseth: it is not, it is a `bytes4`.

test/libsolidity/syntaxTests/inlineAssembly/invalid/external_function_pointer_selector_assignment.sol
chriseth: Yes, intentional - doing `bytes4(7)` would perform two steps at a time in my opinion.

chriseth:
If you plan to also allow assigning, I don't think it is worth arguing about the wording of "read only" inside the documentation. Maybe do both things in the same PR? Or at least keep this in draft mode and base the second one on top of it until both are ready.

libsolidity/analysis/TypeChecker.cpp
Marenz: As far as I understood that won't work. Will add a test case though!

libsolidity/codegen/ContractCompiler.cpp
Marenz: That is intended. The ticket specification was "If x is a Solidity local variable of function external type [..]"

libsolidity/ast/ASTJsonConverter.cpp
chriseth: Maybe we should find a more general solution here anyway? On the other hand, we now use `.selector` and `.address` instead of `_selector` / `_address`. I think it is still not unique since you might use an identifier path to access scopes.

Maybe it would suffice to say "the suffix is used to access a 'property' instead of a scoped variable"? Which property it is can be derived from the suffix itself.

Marenz:
> Maybe do both things in the same PR?

Sure. I just had this pretty much ready when you said we should allow assigning too, so I thought go ahead with that arleady.

libsolidity/analysis/ReferencesResolver.cpp
Marenz: It can't be a `const&` without more changes as we would need to assign it in the declaration.

Changelog.md
chriseth: Needs to be moved.

libsolidity/analysis/TypeChecker.cpp
chriseth: state variables of function type should only have `offset` and `slot`.

libsolidity/ast/ASTJsonConverter.cpp
chriseth: Can we fix this before we forget it?

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -830,6 +844,9 @@ bool ContractCompiler::visit(InlineAssembly const& _inlineAssembly)
 							errinfo_comment("Stack too deep, try removing local variables.")
 						);
 					_assembly.appendInstruction(dupInstruction(stackDiff));
+					if (leftShiftSelector)
```
chriseth: Ok, so it is right-aligned on the stack? Then we should maybe also just return it like that.

docs/assembly.rst
```diff
@@ -136,6 +136,34 @@ evaluate to the address of the variable in calldata, not the value itself.
...
+
+            assembly {
+                fp.selector := newSelector
+                fp.address  := newAddress
+            }
+
+            assert(fp.address == newAddress);
+            assert(fp.selector == bytes4(bytes32(newSelector)));
+        }
+    }
```
chriseth: What about (please check that it compiles)
```suggestion
    // Assigns a new selector and address to the local variable fp
    function combineToFunctionPointer(address newAddress, uint newSelector) pure returns (function() external fun)
            assembly {
                fun.selector := newSelector
                fun.address  := newAddress
            }
        }
    }
```

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -178,9 +178,20 @@ Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifie
...
 	if (!_info.second.suffix.empty())
 		tuple["suffix"] = Json::Value(_info.second.suffix);
+
 	tuple["valueSize"] = Json::Value(Json::LargestUInt(_info.second.valueSize));
+	if (
+		auto const* functionType = dynamic_cast<FunctionType const*>(_info.second.declaration->type());
+		functionType && functionType->kind() == FunctionType::Kind::External
+	)
+	{
+		tuple["isSelector"] = Json::Value(_info.second.suffix == "selector");
```
chriseth: Can we generalize this as I suggested, please?

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -178,9 +179,18 @@ Json::Value ASTJsonConverter::inlineAssemblyIdentifierToJson(pair<yul::Identifie
 	tuple["declaration"] = idOrNull(_info.second.declaration);
 	tuple["isSlot"] = Json::Value(_info.second.suffix == "slot");
 	tuple["isOffset"] = Json::Value(_info.second.suffix == "offset");
+
 	if (!_info.second.suffix.empty())
+	{
 		tuple["suffix"] = Json::Value(_info.second.suffix);
+
+		static auto specialSuffixes = std::set{"slot", "offset", "selector", "address"};
```
chriseth: Wait, do we actually need this? What does `suffix` contain if we have `.slot` without this actually being the slot of a state variable? Or is it always the slot of a state variable? Can we have
```
contract C { uint constant slot = 1; }
contarct D {
function f() public pure { assembly { let x := C.slot } }
}
```

Marenz: If we don't need it, we don't need any extra `specialAccess` attribute either, don't we? Because if there is a suffix it is special access, all the checks are done beforehand anyway...

chriseth (approved):

Issue #12017, opened 2021-09-23
Replace assertions with proper validations in AST import


cameel:
## Abstract
The [AST import code in `CommandLineInterface`](https://github.com/ethereum/solidity/blob/v0.8.7/solc/CommandLineInterface.cpp#L609-L626) is currently wrapped in an exception handler that catches all exceptions and prints an error message. This is most likely because the JSON input does not get validated before it's passed to `ASTJsonImporter` and the importer just asserts things that it expects not to happen.

These assertions should be replaced with proper validations.

The `--import-asm-json` option that will be introduced in #11807 will likely suffer from the same problem.

### Example
This is how you can trigger a validation error now:
```bash
echo "{}" | solc - --import-ast
```
```
Failed to import AST: Invalid Format for import-JSON: Must have 'sources'-object
```

## Motivation
The current handler suppresses all the extra debug information we include in internal compiler errors. This is what we want if these errors are actually validation errors but not for normal assertions. It means that our usual assertions result in a message looking like this:
```
Failed to import AST: 
```
The message is meant to be followed by error description but our assertions often have no description. This makes debugging harder.

## Specification
We should define a new exception type representing a validation error and use it in `ASTJsonImporter`. Then only catch this exception in `CommandLineInterface` and let other exceptions through.

Since the exception handler also covers `CompilerStack->analyze()`, it's likely that the assertions in `ASTJsonImporter` are not enough to satisfy its assumptions. As a part of this task we need to check what exactly these assumptions are and add additional validations against them in `ASTJsonImporter`.

## Backwards Compatibility
This is just a refactor and should not cause changes visible to the user other than in internal compiler errors - and we do not provide backwards-compatibility guarantees for these.

ryzheboka:
Hi, I would like to work on this issue. I'm new to contributing, can you please check my understanding of the needed changes? So I would need to add a new Exception to [Exceptions.h](https://github.com/ethereum/solidity/blob/develop/libsolutil/Exceptions.h). Then, I should throw this exception in ASTJsonImporter. ASRJsonImporter::jsonToSourceUnit is then used in CompilerStack::importASTs. This function would pass the Exception to CommandLineInterface. In CommandLineInterface, I need to catch Exceptions of the new type, instead of the type "Exception".

cameel:
[...]

PR #12019, opened 2021-09-23
Default messages in assertion macros

cameel:
This is an attempt to improve the situation in #12017 a little. It should also be an improvement for external tools because for these our ICEs are often puzzling when they have no message at all (see e.g. https://github.com/trufflesuite/truffle/issues/2849).

- Now if you use `""` in an assertion, it gets replaced with a generic message, like `"Solidity assertion failed"` or `"Unimplemented feature"`, depending on the assertion type.
- You can now skip the empty string in `solAssert()` and other assertions, making them more convenient.

### DIFF
--- a/libsolidity/codegen/YulUtilFunctions.cpp
+++ b/libsolidity/codegen/YulUtilFunctions.cpp
@@ -1217,7 +1217,7 @@ string YulUtilFunctions::extractByteArrayLengthFunction()
 std::string YulUtilFunctions::resizeArrayFunction(ArrayType const& _type)
 {
 	solAssert(_type.location() == DataLocation::Storage, "");
-	solUnimplementedAssert(_type.baseType()->storageBytes() <= 32, "...");
+	solUnimplementedAssert(_type.baseType()->storageBytes() <= 32);
 
 	if (_type.isByteArray())
 		return resizeDynamicByteArrayFunction(_type);
@@ -1259,7 +1259,7 @@ string YulUtilFunctions::cleanUpStorageArrayEndFunction(ArrayType const& _type)
 	solAssert(_type.location() == DataLocation::Storage, "");
 	solAssert(_type.baseType()->category() != Type::Category::Mapping, "");
 	solAssert(!_type.isByteArray(), "");
-	solUnimplementedAssert(_type.baseType()->storageBytes() <= 32, "");
+	solUnimplementedAssert(_type.baseType()->storageBytes() <= 32);
 
 	string functionName = "cleanup_storage_array_end_" + _type.identifier();
 	return m_functionCollector.createFunction(functionName, [&](vector<string>& _args, vector<string>&) {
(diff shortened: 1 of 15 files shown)


cameel:
Looks like MSVC does not like variadic macros very much...

libsolutil/Assertions.h
```diff
@@ -38,21 +40,41 @@ namespace solidity::util
...
 				::boost::throw_line(__LINE__) \
 			); \
 	} \
 	while (false)
 
+/// Assertion that throws an exception containing the given description if it is not met.
+/// Use it as assertThrow(1 == 1, ExceptionType, "Mathematics is wrong.");
+/// The second parameter must be an exception class (rather than an instance).
+#define assertThrow(_condition, _exceptionType, _description) \
+	assertThrowWithDefaultDescription(_condition, _exceptionType, _description, "Assertion failed")
```
cameel, line 80: Is `assertThrow()` really an assertion? In most cases I've seen it's just used as a convenience helper to throw a boost exception.

cameel:
> Looks like MSVC does not like variadic macros very much...

Looks like it requires extra `#ifdef`s unfortunately: [`BOOST_PP_OVERLOAD`](https://www.boost.org/doc/libs/1_77_0/libs/preprocessor/doc/ref/overload.html).

cameel:
By the way, is this something that should have a changelog entry? This might affect tools that try to detect some specific ICEs in the output and do some workarounds. We might want to at least inform them even if we do not guarantee any backwards-compatibility for these messages.

leonardoalt:
I don't think we should guarantee backwards compatibility in ICEs

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -1217,7 +1217,7 @@ string YulUtilFunctions::extractByteArrayLengthFunction()
 std::string YulUtilFunctions::resizeArrayFunction(ArrayType const& _type)
 {
 	solAssert(_type.location() == DataLocation::Storage, "");
-	solUnimplementedAssert(_type.baseType()->storageBytes() <= 32, "...");
```
leonardoalt, line 1220: lol

libsolutil/Assertions.h
```diff
@@ -38,21 +40,43 @@ namespace solidity::util
...
-/// Do NOT supply an exception object as the second parameter.
-#define assertThrow(_condition, _exceptionType, _description) \
+namespace assertions
+{
+
+inline std::string stringOrDefault(std::string _string, std::string _defaultString)
+{
+	// NOTE: Putting this in a function rather than directly in a macro prevents the string from
+	// being evaluated multiple times if it's not just a literal.
+	return (!_string.empty() ? _string : _defaultString);
```
leonardoalt, line 50: Are these properly moved?

cameel: True. They are not. Fixed.

leonardoalt (dismissed):

leonardoalt:
Needs rebase

chriseth (approved):

PR #12024, opened 2021-09-24
Require Boost 1.77 on Windows

cameel:
Originally requested in https://github.com/ethereum/solidity/pull/11688#discussion_r702995346. See also https://github.com/ethereum/solidity/pull/11688#issuecomment-888001567.
Depends on #11688.

In #11688 I introduced normalization, which triggers a bug in Boost on Windows (https://github.com/boostorg/filesystem/issues/201). The bug has been fixed in 1.77 but 1.76 and older versions are still affected. The bug can result in a crash when some relative and non-normalized paths are used with `--allow-paths`.

Since Windows has no official package manager, users are free to download and install any version so I think that increasing the minimum version just for that platform is reasonable.


chriseth (approved):

christianparpart:
sorry @cameel - i just noticed the review request by now :(

PR #12029, opened 2021-09-27
Update README.md

PavanAnanthSharma:
Using different versions of solidity for writing Hello World

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ A "Hello World" program in Solidity is of even less use than in other languages,
 
 ```solidity
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.6.0 <0.9.0;
+pragma solidity ^0.6.0 <=0.9.0;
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ A "Hello World" program in Solidity is of even less use than in other languages,
 
 ```solidity
-pragma solidity >=0.6.0 <0.9.0;
+pragma solidity ^0.6.0 <=0.9.0;
```
hrkrshnn, line 50: This is not the right use of `^`. Since `0.9.0` doesn't exist, the equality part in `<= 0.9.0` does not make sense. We usually bump the version (`0.9.0`) after its release.

hrkrshnn (changes requested):
Closing this. Feel free to reply if you have questions.

README.md
PavanAnanthSharma: G thanks

PR #12036, opened 2021-09-27
[SMTChecker] Fix magic access

leonardoalt:
Found in https://github.com/ethereum/solidity/pull/11985#discussion_r711226327
Depends on https://github.com/ethereum/solidity/pull/12037

The added test used to ICE.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@ Compiler Features:
 
 
 Bugfixes:
+  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
 
 
 
--- a/libsolidity/formal/SMTEncoder.cpp
+++ b/libsolidity/formal/SMTEncoder.cpp
@@ -1316,32 +1316,42 @@ bool SMTEncoder::visit(MemberAccess const& _memberAccess)
 			solAssert(name == "block" || name == "msg" || name == "tx", "");
 			defineExpr(_memberAccess, state().txMember(name + "." + _memberAccess.memberName()));
 		}
-		else if (auto magicType = dynamic_cast<MagicType const*>(exprType); magicType->kind() == MagicType::Kind::MetaType)
+		else if (auto magicType = dynamic_cast<MagicType const*>(exprType))
 		{
-			auto const& memberName = _memberAccess.memberName();
-			if (memberName == "min" || memberName == "max")
+			if (magicType->kind() == MagicType::Kind::Block)
+				defineExpr(_memberAccess, state().txMember("block." + _memberAccess.memberName()));
+			else if (magicType->kind() == MagicType::Kind::Message)
+				defineExpr(_memberAccess, state().txMember("msg." + _memberAccess.memberName()));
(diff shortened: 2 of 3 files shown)


libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1316,32 +1316,42 @@ bool SMTEncoder::visit(MemberAccess const& _memberAccess)
...
 		{
-			auto const& memberName = _memberAccess.memberName();
-			if (memberName == "min" || memberName == "max")
+			if (magicType->kind() == MagicType::Kind::Block)
+				defineExpr(_memberAccess, state().txMember("block." + _memberAccess.memberName()));
+			else if (magicType->kind() == MagicType::Kind::Message)
+				defineExpr(_memberAccess, state().txMember("msg." + _memberAccess.memberName()));
+			else if (magicType->kind() == MagicType::Kind::Transaction)
+				defineExpr(_memberAccess, state().txMember("tx." + _memberAccess.memberName()));
+			else if (magicType->kind() == MagicType::Kind::MetaType)
```
leonardoalt, line 1327: The block above was added, the rest just moved inside the other `if`.

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1419,12 +1429,12 @@ bool SMTEncoder::visit(MemberAccess const& _memberAccess)
...
-			7650_error,
-			_memberAccess.location(),
-			"Assertion checker does not yet support this expression."
-		);
+
+	m_errorReporter.warning(
+		7650_error,
+		_memberAccess.location(),
+		"Assertion checker does not yet support this expression."
+	);
```
leonardoalt, line 1437: This was moved outside the `else` to catch all cases that didn't return early. Didn't seem to affect our tests.

Changelog.md
```diff
@@ -1,3 +1,10 @@
+### 0.8.9 (unreleased)
+
+Bugfixes:
+
```
hrkrshnn, line 8:
```suggestion
```

libsolidity/formal/SMTEncoder.cpp
hrkrshnn, line 1327: We are missing the ABI kind, aren't we?

leonardoalt: Those are handled somewhere else

hrkrshnn: Wouldn't something like
```
contract C {
    function f() public {
        abi.encode;
    }
}
```
get into this branch and throw at the end?

leonardoalt: Let me see

leonardoalt: It would go in the `magic type` branch, so it wouldn't reach the assert(false). That's the same behavior as before actually, I just added the extra `msg`, `block`, `tx`.

chriseth (approved):

PR #12037, opened 2021-09-27
Add isoltest option to ignore OS

### LINKED ISSUE #11870: [soltest] Directive for skipping SMTChecker tests on specific platforms

leonardoalt:
Fix https://github.com/ethereum/solidity/issues/11870

### DIFF
--- a/test/libsolidity/SMTCheckerTest.cpp
+++ b/test/libsolidity/SMTCheckerTest.cpp
@@ -66,6 +66,21 @@ SMTCheckerTest::SMTCheckerTest(string const& _filename): SyntaxTest(_filename, E
 		m_ignoreCex = true;
 	else
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid SMT counterexample choice."));
+
+	auto const& ignoreOSSetting = m_reader.stringSetting("SMTIgnoreOS", "none");
+	for (string const& os: ignoreOSSetting | ranges::views::split(',') | ranges::to<vector<string>>())
+	{
+#ifdef __APPLE__
+		if (os == "macos")
+			m_shouldRun = false;
+#elif _WIN32
+		if (os == "windows")
+			m_shouldRun = false;
+#elif __linux__
+		if (os == "linux")
+			m_shouldRun = false;
+#endif
+	}
 }
 
 TestCase::TestResult SMTCheckerTest::run(ostream& _stream, string const& _linePrefix, bool _formatted)
--- a/test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
+++ b/test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
@@ -2,13 +2,17 @@ contract C {
 	function g(address payable i) public {
 		require(address(this).balance == 100);
 		i.call{value: 10}("");
-		// Disabled due to Spacer nondeterminism
-		//assert(address(this).balance == 90); // should hold
-		// Disabled due to Spacer nondeterminism
-		//assert(address(this).balance == 100); // should fail
+		assert(address(this).balance == 90); // should hold
+		assert(address(this).balance == 100); // should fail
 	}
 }
 // ====
 // SMTEngine: all
+// SMTIgnoreOS: macos
 // ----
 // Warning 9302: (96-117): Return value of low-level calls not used.
+// Warning 1218: (175-211): CHC: Error trying to invoke SMT solver.
+// Warning 6328: (121-156): CHC: Assertion violation might happen here.
+// Warning 6328: (175-211): CHC: Assertion violation might happen here.
+// Warning 4661: (121-156): BMC: Assertion violation happens here.
+// Warning 4661: (175-211): BMC: Assertion violation happens here.


test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
```diff
@@ -2,13 +2,17 @@ contract C {
...
+		assert(address(this).balance == 90); // should hold
+		assert(address(this).balance == 100); // should fail
 	}
 }
 // ====
 // SMTEngine: all
+// SMTIgnoreOS: macos
 // ----
 // Warning 9302: (96-117): Return value of low-level calls not used.
+// Warning 1218: (175-211): CHC: Error trying to invoke SMT solver.
```
hrkrshnn, line 14: Is this okay?

leonardoalt: Yea those are just z3 errors when generating counterexamples. What's important is that all OS agree.

hrkrshnn (dismissed):

hrkrshnn:
But maybe wait for the release to end before merging?

leonardoalt:
Yep! Didn't mean for it to go in before

hrkrshnn:
https://github.com/ethereum/solidity/pull/12038 would conflict with this.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
```diff
@@ -2,13 +2,17 @@ contract C {
 	function g(address payable i) public {
 		require(address(this).balance == 100);
 		i.call{value: 10}("");
-		// Disabled due to Spacer nondeterminism
-		//assert(address(this).balance == 90); // should hold
-		// Disabled due to Spacer nondeterminism
-		//assert(address(this).balance == 100); // should fail
+		assert(address(this).balance == 90); // should hold
+		assert(address(this).balance == 100); // should fail
```
christianparpart, line 6: While I do not disagree with this change. I wonder. Shouldn't we have a PR for one subject and another for something else? (bad wording maybe. jsut not having too much different things in one PR / commit).

leonardoalt: The code change is there exactly to allow re-enabling such tests.

test/libsolidity/SMTCheckerTest.cpp
```diff
@@ -66,6 +66,21 @@ SMTCheckerTest::SMTCheckerTest(string const& _filename): SyntaxTest(_filename, E
 		m_ignoreCex = true;
 	else
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid SMT counterexample choice."));
+
+	auto const& os = m_reader.stringSetting("SMTIgnoreOS", "none");
+	for (auto&& o: os | ranges::views::split(',') | ranges::to<vector<string>>())
```
cameel: It's a bit confusing to call the whole string `os`.
```suggestion
	string const& ignoreOSSetting = m_reader.stringSetting("SMTIgnoreOS", "none");
	for (string const& os: ignoreOSSetting | ranges::views::split(',') | ranges::to<vector<string>>())
```

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
cameel: Yeah, the whole point was that these had to be disabled becuse they were failing on macOS so uncommenting them shows that the new setting works.

cameel (dismissed):
I'd put it in `SyntaxTest` rather than make it SMT specific but, well, only SMTChecker needs it now so I guess we can extend it later if we need it elsewhere.

test/libsolidity/smtCheckerTests/external_calls/call_with_value_1.sol
leonardoalt: I'll still enable more tests in a different PR, but wanted to enable at least one back in this one to prove the point and not bloat it.

leonardoalt:
Fixed code

leonardoalt:
@cameel can you reapprove?

cameel (approved):

Issue #11987, opened 2021-09-16
Error message about disallowed argument location should explain why the location is not allowed


nventuro:
When compiling the following contract with solc v0.8.7:

```
pragma solidity ^0.8.0;

contract Test {
    struct MyStruct {
        uint256 a;
    }
    
    function foo(MyStruct storage bar) public;
}
```

we get the following error message:

```
TypeError: Data location must be "memory" or "calldata" for parameter in function, but "storage" was given.
    function foo(MyStruct storage bar) public;
                 ^------------------^
```

This is not particularly helpful as functions can actually receive `storage` arguments, but only if they are `internal` or `private`. The error message is not particularly insightful on this distinction, nor helps the developer achieve their task.

### Suggestion

A simple improvement could be:
```
Data location must be "memory" or "calldata" for parameter in "public" or "external" function
```

Even better, a suggestion could be appended at the end:

```
TypeError: Data location must be "memory" or "calldata" for parameter in "public" or "external" function, but 
"storage" was given.
    function foo(MyStruct storage bar) public;
                 ^------------------^

Use "internal" or "private" functions to pass storage pointers.
```

axic:
I think this would be as a nice improvement. Did you find similar other cases?

nventuro:
Hm I don't recall any off any of the top of my head - I just run into this the other day and was puzzled by the error message until I realized what was going on.

jhb10c:
Hey @cameel, @chriseth. I would like to work on this. I can update the error to the suggestion by @nventuro. If there are any specifications, could you all provide more info?

cameel:
Well, the version with adjusting the message is straightforward:
- find the message in the code
- tweak description
- run `isoltest` to update tests

A version with the extra suggestion would be harder. You can have an extra message if you specify a secondary location - but that's not what they were meant for and if would be weird if secondary location was pointing at the same thing. But maybe it would make sense if you made it point at the `public`/`external` keyword?

You could also issue two errors, one with "info" severity meant to contain just this extra suggestion - but there's the risk that the user will want to hide info messages.

Neither of these seems particularly nice so I I'd go with just tweaking the text unless you can figure out a nicer solution.

jhb10c:
Thanks for the instructions. I will update the error message for now and leave the extra suggestion for later. There are some similar cases. As such I will also update the error messages for external/public functions with the following data types: bytes; uint[]; structs; mapping(uint => uint); mapping(uint => uint)[].

[...]

PR #12040, opened 2021-09-28
Update error message for error 6651 (`Data location must be ...`)

jhb10c:
External and public functions are not allowed to have parameters whose data location is storage. The error message has been updated to reflect this fact. 

This was originally requested in  #[11987](https://github.com/ethereum/solidity/issues/11987)

### DIFF
--- a/libsolidity/analysis/DeclarationTypeChecker.cpp
+++ b/libsolidity/analysis/DeclarationTypeChecker.cpp
@@ -376,12 +376,17 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 			if (_variable.isConstructorParameter())
 				errorString += " for constructor parameter";
 			else if (_variable.isCallableOrCatchParameter())
+			{
+				string visibility;
+				if (auto const* callable = dynamic_cast<CallableDeclaration const*>(_variable.scope()))
+					visibility = Declaration::visibilityToString(callable->visibility());
 				errorString +=
-					" for " +
-					string(_variable.isReturnParameter() ? "return " : "") +
-					"parameter in" +
-					string(_variable.isExternalCallableParameter() ? " external" : "") +
-					" function";
+				" for " +
+				(_variable.isReturnParameter() ? "return "s : ""s) +
+				"parameter in" +
+				(visibility.empty() ? visibility : " "s + visibility) +
+				" function";
+			}
 			else
 				errorString += " for variable";
 		}
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,9 @@ Bugfixes:
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
 
 
+Important Bugfixes in Experimental Features:
+ * Yul IR Generator: Changes to function return variables referenced in modifier invocation arguments were not properly forwarded if there was more than one return variable.
+
 
 ### 0.8.9 (2021-09-29)
 
--- a/libsolidity/codegen/ir/IRGenerator.cpp
+++ b/libsolidity/codegen/ir/IRGenerator.cpp
@@ -438,7 +438,7 @@ string IRGenerator::generateModifier(
 		for (size_t i = 0; i < retParamsIn.size(); ++i)
 		{
 			retParams.emplace_back(m_context.newYulVariable());
-			assignRetParams += retParams.back() + " := " + retParamsIn[i] + "\n";
+			assignRetParams += retParams.at(i) + " := " + retParamsIn.at(i) + "\n";
 		}
 		t("retParams", joinHumanReadable(retParams));
 		t("assignRetParams", assignRetParams);
@@ -529,7 +529,7 @@ string IRGenerator::generateFunctionWithModifierInner(FunctionDefinition const& 
 		for (size_t i = 0; i < retParams.size(); ++i)
 		{
 			retParamsIn.emplace_back(m_context.newYulVariable());
-			assignRetParams += retParams.back() + " := " + retParamsIn[i] + "\n";
+			assignRetParams += retParams.at(i) + " := " + retParamsIn.at(i) + "\n";
 		}
 		vector<string> params = retParamsIn;
 		for (auto const& varDecl: _function.parameters())
--- a/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
+++ b/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
@@ -1,0 +1,25 @@
+interface Identity {
+    function selectorAndAppendValue(uint value) external pure returns (uint);
+}
+contract C {
+    Identity constant i = Identity(address(0x0004));
+    function testHighLevel() external pure returns (bool) {
+        // Should fail because `extcodesize(4) = 0`
+        i.selectorAndAppendValue(5);
+        return true;
+    }
+    function testLowLevel() external view returns (uint value) {
+        (bool success, bytes memory ret) =
+            address(4).staticcall(
+                abi.encodeWithSelector(Identity.selectorAndAppendValue.selector, uint(5))
+            );
+        value = abi.decode(ret, (uint));
+    }
+
+}
+// ====
+// compileViaYul: also
+// EVMVersion: >=constantinople
+// ----
+// testHighLevel() -> FAILURE
+// testLowLevel() -> 0xc76596d400000000000000000000000000000000000000000000000000000000
--- a/test/libsolidity/semanticTests/modifiers/function_modifier_return_reference.sol
+++ b/test/libsolidity/semanticTests/modifiers/function_modifier_return_reference.sol
@@ -1,0 +1,15 @@
+contract C {
+    modifier m1(uint value) {
+        _;
+    }
+    modifier m2(uint value) {
+        _;
+    }
+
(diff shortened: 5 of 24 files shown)


cameel:
Your PR does not actually change the the code that generates the message. It only changes the tests. You need to apply the change here:
https://github.com/ethereum/solidity/blob/d10e668f4f89878983fcaa46e7b88462ffa4380c/libsolidity/analysis/DeclarationTypeChecker.cpp#L370-L391

jhb10c:
That makes sense. 

I have added in the line below "string(_variable.isExternalCallableParameter() ? " external" : "")": string(_variable.isPublicCallableParameter() ? " public" : "") and updated the tests using isoltest. This time it should work.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -383,6 +383,7 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 					string(_variable.isReturnParameter() ? "return " : "") +
 					"parameter in" +
 					string(_variable.isExternalCallableParameter() ? " external" : "") +
+                    string(_variable.isPublicCallableParameter() ? " public" : "") +
```
jhb10c: This adds the word public in front function for error message 6651.

cameel: Your line is indented with spaces while the rest of the code is using tabs. This is probably why [`chk_coding_style`](https://circleci.com/gh/ethereum/solidity/860718) job is failing.

tied2token (dismissed):
It didn't seem to error for me

jhb10c:
@tied2token, I used spaces instead of tabs in the c++ code. This has been updated.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -383,6 +383,7 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 					string(_variable.isReturnParameter() ? "return " : "") +
 					"parameter in" +
 					string(_variable.isExternalCallableParameter() ? " external" : "") +
+					string(_variable.isPublicCallableParameter() ? " public" : "") +
```
cameel: This is correct but I think we should also be doing this for return parameters. I see a lot of messages in tests that still say "in function" and those are either for storage return parameters or for cases where a parameter has no location at all.

Actually, I see no reason not to make it simple and just always include the visibility in the message. You can get visibility like this (not tested but should work):
```c++
string visibility;
if (auto const* callable = dynamic_cast<CallableDeclaration const*>(_variable.scope()))
    visibility = Declaration::visibilityToString(callable->visibility());
```

cameel (commented):
Looks good already but we could do more. See my comment below.

Also, please squash your commits so that each one represents a single logical change (with no separate commits for PR fixes).

libsolidity/analysis/DeclarationTypeChecker.cpp
chriseth: Just a heads up: If you ask for the visibility of a constructor, you will get a failing assertion.

cameel: ok, yeah, this may need some exta ifs.

jhb10c: Gotcha, @cameel. The first part seems straight forward: update the message for storage return parameters and parameters where type is not defined by including the function type. 

An example of the second type would be like "Data location must be "storage", "memory" or "calldata" for return parameter in function, but none was given." . The function type should be mentioned in the message. 

Thanks for the heads up @chriseth. To check that the CallableDeclaration object is not a constructor, it should look something like:

`string visibility;
if (auto const* callable = dynamic_cast<FunctionDeclaration const*>(_variable.scope()))
{if (!*callable.isConstructor())
       visibility = Declaration::visibilityToString(callable->visibility());
.....
}`

Or would it be better to leave the type as a CallableDeclaration and check that if it is a FunctionDeclaration then check if it is a constructor or not?

cameel: Actually, I see that there's already a check for a constructor parameter in the branch above (`if (_variable.isConstructorParameter())`) so I think you do not have to worry about this case. It's already handled.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -378,12 +378,17 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
+							string visibility;
+							if (auto const* callable = dynamic_cast<CallableDeclaration const*>(_variable.scope()))
+								visibility = Declaration::visibilityToString(callable->visibility());
+							errorString +=
+							" for " +
+							string(_variable.isReturnParameter() ? "return " : "") +
+							"parameter in" +
+							string(visibility.empty() ? visibility : " " + visibility) +
+							" function";
+					}
```
cameel:
```suggestion
				string visibility;
				if (auto const* callable = dynamic_cast<CallableDeclaration const*>(_variable.scope()))
					visibility = Declaration::visibilityToString(callable->visibility());
				errorString +=
					" for " +
					string(_variable.isReturnParameter() ? "return " : "") +
					"parameter in" +
					string(visibility.empty() ? visibility : " " + visibility) +
					" function";
			}
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -378,12 +378,17 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
-					" function";
+			{
+							string visibility;
+							if (auto const* callable = dynamic_cast<CallableDeclaration const*>(_variable.scope()))
+								visibility = Declaration::visibilityToString(callable->visibility());
+							errorString +=
+							" for " +
+							string(_variable.isReturnParameter() ? "return " : "") +
+							"parameter in" +
+							string(visibility.empty() ? visibility : " " + visibility) +
```
cameel: Since we're touching this code, we could also change the string conversion to a more concise one:
```suggestion
							_variable.isReturnParameter() ? "return "s : ""s +
							"parameter in" +
							visibility.empty() ? visibility : " "s + visibility +
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -378,12 +378,17 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
+							" function";
```
cameel: We could use our `joinHumanReadable()` to make this more concise (needs `#include <libsolutil/StringUtils.h`). This removes the need to worry about including or omitting the spaces between components:
```suggestion
							string parameterKind = _variable.isReturnParameter() ? "return parameter"s : "parameter"s;
							errorString += " " + joinHumanReadable({"for", parameterKind, "in", visibility, "function"}, " ");
```

cameel (commented):
Overall the code seems correct. Only style needs some tweaking.

I see one more case where this message does not include visibility: arguments function pointers (see `function_type_argument_external.sol`, `function_type_return_external.sol` and `data_location_in_function_type_fail.sol`). We should fix them too but it's also fine to merge what we have here and do it in a separate PR in my opinion.

Please rebase on latest `develop` - it has fixes for the CI jobs that are failing in your PR. And we need to see that this passes all of them.

libsolidity/analysis/DeclarationTypeChecker.cpp
jhb10c: I tried using joinHumanReadable() but it could not find the function even with the header included. I have opted to for the second approach. 

There are still whitespace issues that I am fat fingering. ( I am missing how to obtain the correct styling completely....)

cameel: Maybe that's because this file does not use the namespace where the function is defined yet. Please try `util::joinHumanReadable()` or add `using namespace solidity::util;` at the top of the file.

cameel: The whitespace issue seems to be that you're indenting with spaces while we use tabs for .cpp files.

I see that the `chk_coding_style` job detects this but does not print any message about what is wrong. We should change that at some point.

leonardoalt:
Coding style still needs some fixing.

libsolidity/analysis/DeclarationTypeChecker.cpp
jhb10c: got you, let me try again real fast

jhb10c:
I'm a tech boomer. The whitespace on my IDE was set to space and not tab....

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -376,12 +376,17 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 			if (_variable.isConstructorParameter())
 				errorString += " for constructor parameter";
 			else if (_variable.isCallableOrCatchParameter())
+			{
+				string visibility;
```
chriseth, line 380: Maybe
```suggestion
				string functionVisibility;
```
?

jhb10c: Gotcha. I will update this and submit later tonight.

chriseth:
Closing due to inactivity. If you still want to work on in, feel free to re-open.

PR #12041, opened 2021-09-28
Update the snap to build on Ubuntu 20.04

come-maiz:


chriseth:
Thanks a lot for the PR! Any way we can verify this?

chriseth (approved):

come-maiz:
You can test it running `snapcraft`, which will use a multipass virtual machine, or with `snapcraft --use-lxd` which will use a container.
The setup for building this thing was weird, because it needed packages not in ubuntu 18.04, so we used launchpad instead of build.snapcraft.io

I'm seeing some errors on non amd64 architectures during the antlr compilation. We can avoid that using the package, so I'll try that next.

I think I missed an update on the command. I'm checking that too. The build for amd64 is stuck 👀 I'm not sure if I missed this part of my live 😄

PR #12044, opened 2021-09-28
UserDefinedValueType: test for reading from dirty storage slots

hrkrshnn:


test/libsolidity/semanticTests/userDefinedValueType/dirty_slot.sol
```diff
@@ -0,0 +1,37 @@
+type MyUInt16 is uint16;
+type MyBytes2 is bytes2;
+contract C {
+    MyUInt16 public a = MyUInt16.wrap(13);
+    MyBytes2 public b = MyBytes2.wrap(bytes2(uint16(1025)));
+    bytes2 public x;
+    function write_a() external {
+        uint max = type(uint).max;
```
chriseth: actually can you change this to some kind of irregular pattern, so we see what is actually read? (same with write_b)

hrkrshnn: Fixed this.

chriseth (approved):

PR #12045, opened 2021-09-28
Fix array copying check.

chriseth:


libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2416,7 +2416,7 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
 	return m_functionCollector.createFunction(functionName, [&]() {
 		if (_from.baseType()->isValueType())
 		{
-			solAssert(_from.baseType() == _to.baseType(), "");
+			solAssert(*_from.baseType() == *_to.baseType(), "");
```
hrkrshnn, line 2419: Can this assert be reproduced outside of using user defined value type?

chriseth: Maybe for enums, but I didn't check. My change only relaxed the check, so I think it's fine.

hrkrshnn (dismissed):

chriseth:
Updated (forced abi v2)

hrkrshnn (approved):

hrkrshnn:
What happens for abiv1?

chriseth:
v1 cannot handle the struct.

PR #12047, opened 2021-09-28
Properly handle fixed-byte-like types.

chriseth:


chriseth:
Actually, `FunctionType::leftAligned` also returns `true` for external functions - we need to check if that is a problem.

chriseth:
We should also check if signextend is properly called everywhere.

test/libsolidity/semanticTests/userDefinedValueType/storage_signed.sol
```diff
@@ -0,0 +1,35 @@
...
+// ====
+// compileViaYul: also
+// ----
+// a() -> -2
+// direct() -> -2
+// indirect() -> -2
+// toMemDirect() -> -2
+// toMemIndirect() -> -2
+// div() -> -1
+// viaasm() -> 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe
```
chriseth, line 35: This is still failing for the legacy codegen - not sure why.

chriseth: Ok, fixed now.

test/libsolidity/semanticTests/userDefinedValueType/immutable_signed.sol
```diff
@@ -0,0 +1,21 @@
...
+        MyBytes _b = b;
+        assembly { x := _a y := _b }
+    }
+    function g() internal pure returns (uint) { return 2; }
+}
+// ====
+// compileViaYul: also
+// ----
+// direct() -> -2, 0x6162000000000000000000000000000000000000000000000000000000000000
+// viaasm() -> 0xfffe, 0x6162000000000000000000000000000000000000000000000000000000000000
```
hrkrshnn: This is missing the `signextend`, right?

libsolidity/ast/Types.cpp
```diff
@@ -3071,10 +3072,7 @@ u256 FunctionType::storageSize() const
 
 bool FunctionType::leftAligned() const
 {
-	if (m_kind == Kind::External)
-		return true;
-	else
-		solAssert(false, "Alignment property of non-exportable function type requested.");
+	return m_kind == Kind::External;
```
leonardoalt, line 3075: Why was the assertion relaxed?

libsolidity/codegen/CompilerUtils.cpp
```diff
@@ -1552,10 +1552,13 @@ void CompilerUtils::storeStringData(bytesConstRef _data)
 unsigned CompilerUtils::loadFromMemoryHelper(Type const& _type, bool _fromCalldata, bool _padToWords)
 {
 	solAssert(_type.isValueType(), "");
+	Type const* type = &_type;
+	if (auto const* userDefined = dynamic_cast<UserDefinedValueType const*>(type))
+		type = &userDefined->underlyingType();
```
leonardoalt, line 1557: Now you have to do the same i had to do in the smtchecker ;)

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -192,7 +192,12 @@ size_t ContractCompiler::packIntoContractCreator(ContractDefinition const& _cont
 	auto const& immutables = contractType.immutableVariables();
 	// Push all immutable values on the stack.
 	for (auto const& immutable: immutables)
-		CompilerUtils(m_context).loadFromMemory(static_cast<unsigned>(m_context.immutableMemoryOffset(*immutable)), *immutable->annotation().type);
+		CompilerUtils(m_context).loadFromMemory(
+			static_cast<unsigned>(m_context.immutableMemoryOffset(*immutable)),
+			*immutable->annotation().type,
+			false,
+			true
```
leonardoalt: So this was buggy before?

libsolidity/codegen/LValue.cpp
```diff
@@ -159,7 +160,9 @@ void ImmutableItem::retrieveValue(SourceLocation const&, bool) const
 	if (m_context.runtimeContext())
 		CompilerUtils(m_context).loadFromMemory(
 			static_cast<unsigned>(m_context.immutableMemoryOffset(m_variable)),
-			*m_dataType
+			*m_dataType,
+			false,
+			true
```
leonardoalt: And this was buggy too?

leonardoalt (commented):
A few comments. Looks good in general besides my questions, but I can't really tell whether these are all the places that should be adjusted.

libsolidity/ast/Types.cpp
chriseth: In order to avoid the check whether or not the type is an external function type at the call site.

libsolidity/codegen/CompilerUtils.cpp
chriseth: yep! But only because not all properties that matter are part of the interface.

libsolidity/codegen/ContractCompiler.cpp
chriseth: Yes: It would load the packed version but not perform sign extension.

libsolidity/codegen/LValue.cpp
chriseth: Yes, same thing.

libsolidity/codegen/LValue.cpp
```diff
@@ -178,7 +181,7 @@ void ImmutableItem::storeValue(Type const& _sourceType, SourceLocation const&, b
 		utils.moveIntoStack(m_dataType->sizeOnStack());
 	else
 		utils.copyToStackTop(m_dataType->sizeOnStack() + 1, m_dataType->sizeOnStack());
-	utils.storeInMemoryDynamic(*m_dataType, false);
+	utils.storeInMemoryDynamic(*m_dataType);
```
chriseth: This changes from packed encoding to non-packed encoding.

leonardoalt:
Tests need update

chriseth:
I need to rebase this on top of the immutable bugfix for the tests to work properly.

hrkrshnn (dismissed):
LGTM. Added a test in the last commit.

hrkrshnn (approved):

PR #12049, opened 2021-09-28
Fix version check for docs

Marenz:
Versions using the syntax `>0.8.8` would cause an error in the CI if you are on
that version due to the removal of the `greater` check in c63fd0a1af

This adds that check back.

### DIFF
--- a/scripts/docs_version_pragma_check.sh
+++ b/scripts/docs_version_pragma_check.sh
@@ -124,7 +124,7 @@ function findMinimalVersion()
         fi
     done
 
-    if [[ "$version" == "" ]]
+    if [[ "$version" == "" && "$greater" == false ]]
     then
         printError "No release ${sign}${pragmaVersion} was listed in available releases!"
         exit 1


Marenz:
Regarding tests, I rebased https://github.com/ethereum/solidity/pull/12016 on this PR and now the pragma_version_check test succeeds

cameel:
The check was correct. It will use your local version if you have 0.8.9 in `CMakeLists.txt`.

I think that the problem is elsewhere. Looks like version in `CMakeLists.txt` has not been bumped to 0.8.9 in #12038, which is what we'd normally do (see e.g. #11788 or #11568).

scripts/docs_version_pragma_check.sh
```diff
@@ -124,7 +124,7 @@ function findMinimalVersion()
         fi
     done
 
-    if [[ "$version" == "" ]]
+    if [[ "$version" == "" && "$greater" == false ]]
```
cameel, line 127: This does not seem like a correct change. I'm not sure what happens if `version` is empty but the script was not meant to continue in that case. This basically allows anything after `>` in the pragma and ignores it unless it happens to match one of the released versions.

cameel: BTW, we should really rewrite this script in something sane like Python. Trying to read this with all kinds of variables being used as flags and implicit globals like `version` just hurts :( Every time I have get back to this I have to stare at the code for a good while before I understand what it does, even though it's not even that long.

PR #12050, opened 2021-09-28
Fix signed immutables bug.

chriseth:


docs/bugs.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "uid": "SOL-2021-3",
+        "name": "SignedImmutables",
+        "summary": "Immutable variables of signed integer type shorter than 256 bits can have non-cleaned higher order bits if they are accessed via inline assembly.",
```
hrkrshnn: But they cannot be directly accessed via inline assembly. Maybe we should rephrase it?

chriseth: This is the summary.... "if their value is accessed via inline assembly"?

bshastry:
```suggestion
        "summary": "Immutable variables of signed integer type shorter than 256 bits can have dirty higher order bits if they are accessed via inline assembly.",
```

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
...
+        "description": "When immutable variables of signed integer type shorter than 256 bits are read, their higher order bits will be zero. The correct operation would be to sign-extend the value, i.e. set the higher order bits to one if the sign bit is one. This sign-extension is performed by Solidity just prior to whern it matters, i.e. when a value is stored in memory, when it is compared or when a division is performed. Because of that, to our knowledge, the only way to access the value in its unclean state is by reading it through inline assembly.",
```
bshastry:
```suggestion
        "description": "When immutable variables of signed integer type shorter than 256 bits are read, their higher order bits were incorrectly set to zero. The correct operation would be to sign-extend the value, i.e. set the higher order bits to one if the sign bit is one. This sign-extension is performed by Solidity just prior to when it matters, i.e. when a value is stored in memory, when it is compared or when a division is performed. Because of that, to our knowledge, the only way to access the value in its unclean state is by reading it through inline assembly.",
```

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
...
+        "link": "https://blog.soliditylang.org/2021/09/29/signed-immutables-bug/",
```
bshastry, line 7: Just FYI: Link is dead.

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -439,7 +444,7 @@ void ContractCompiler::appendFunctionSelector(ContractDefinition const& _contrac
 	// retrieve the function signature hash from the calldata
 	if (!interfaceFunctions.empty())
 	{
-		CompilerUtils(m_context).loadFromMemory(0, IntegerType(CompilerUtils::dataStartOffset * 8), true);
+		CompilerUtils(m_context).loadFromMemory(0, IntegerType(CompilerUtils::dataStartOffset * 8), true, false);
```
hrkrshnn, line 447: Why was this change made?

hrkrshnn: Ah, I see the defaults are removed.

hrkrshnn (commented):
The changes look good. Will take one more look tomorrow.

docs/bugs.json
chriseth: Yes, it is not yet published.

docs/bugs.json
chriseth: Dirty is maybe the wrong word, because the will always be zero.

bshastry: `Non sign extended` maybe?

chriseth:
Updated the wording of the bug list entry a little.

docs/bugs.json
chriseth: I used "invalid" now.

hrkrshnn (approved):

PR #12054, opened 2021-09-29
Add one more test case.

chriseth:


test/libsolidity/semanticTests/userDefinedValueType/calldata_to_storage.sol
```diff
@@ -25,19 +25,19 @@ contract C {
 // ----
 // s() -> 0, 0, 0x00, 0
 // f((uint8,uint16,bytes2,uint8)): 1, 0xff, "ab", 15 ->
-// gas irOptimized: 110778
+// gas irOptimized: 44860
 // gas legacy: 112851
-// gas legacyOptimized: 110766
+// gas legacyOptimized: 44923
```
chriseth, line 30: This sounds like the gas costs were computed without the storage slot change?? How did it succeed in the PR?

hrkrshnn: Yeah, this looks too suspicious. The cost is almost half.

chriseth: Ah no it's fine - the test was added to develop before we fixed anything.

hrkrshnn: If that is the case, then I think the gas for legacy should also change.

PR #12060, opened 2021-09-29
Make evaluation of string literals more generic

acoglio:
This is based on a discussion on Gitter.

We could add, after the interpreter, an explanation of what the `str`, `dec`, and `hex` decoding functions are. In particular, we could add a reference to the Literals section (https://docs.soliditylang.org/en/latest/yul.html#literals) that explains how `str` works for the EVM dialect (from the previous PR I made). But we can do this in a separate PR.

### DIFF
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -714,13 +714,15 @@ We will use a destructuring notation for the AST nodes.
         L'[$parami] = vi and L'[$reti] = 0 for all i.
         Let G'', L'', mode = E(Gn, L', block)
         G'', Ln, L''[$ret1], ..., L''[$retm]
-    E(G, L, l: StringLiteral) = G, L, utf8EncodeLeftAligned(l),
-        where utf8EncodeLeftAligned performs a UTF-8 encoding of l
-        and aligns it left into 32 bytes
+    E(G, L, l: StringLiteral) = G, L, str(l),
+        where str is the string evaluation function,
+        which for the EVM dialect is defined in the section 'Literals' above
     E(G, L, n: HexNumber) = G, L, hex(n)
-        where hex is the hexadecimal decoding function
+        where hex is the hexadecimal evaluation function,
+        which turns a sequence of hexadecimal digits into their big endian value
     E(G, L, n: DecimalNumber) = G, L, dec(n),
-        where dec is the decimal decoding function
+        where dec is the decimal evaluation function,
+        which turns a sequence of decimal digits into their big endian value
 
 .. _opcodes:
 


chriseth (changes requested):
I think this should add a bit more what the string decoding function is. Also - but I think we already argued about something similar elsewhere - I think it should be "encoding" and not "decoding".

acoglio:
@chriseth I just added a commit that changes 'decoding' to 'encoding' as you suggested, for all three kinds of literals. I was actually thinking whether 'evaluation' may be a better term than both 'decoding' and 'encoding', i.e. `str` evaluates a string, `hex` evaluates a hexadecimal literal, and `dec` evaluates a decimal literal. What do you think?

(You also mention adding more detail. I agree, but I wanted to finalize the encoding/decoding/evaluation terminology first.)

chriseth:
Oh I'm so sorry! Now I see why you had "decoding" in the first place. The problem is that for strings, it is kind of both decoding and encoding (decoding: Interpret escape sequences and so on, encoding: left align / pad with zeros)

acoglio:
No problem! What do you think about my above proposal of using 'evaluation' instead?

leonardoalt:
ping @chriseth

chriseth:
I'm sorry, but I'm a bit lost here :)
I think that hex and dec count as "decoding" functions, but the string part is a bit more difficult because we do not really talk about how the source code is actually represented. I think before this can be merged we should expand what the string function actually does somewhere is the documentation and then maybe also briefly mention the other two functions.

acoglio:
@chriseth I just pushed a commit that adds some explanation of what the `str`, `hex`, and `dec` functions are.

The commit also calls them 'evaluation' functions, which in my opinion is the best term for them because they evaluate pieces of abstract syntax to semantic values. However, I'm happy to change the text if you prefer a different term.

I realize that the description of `str` is not perfect, making reference to a previous section and being only EVM-specific. Nonetheless, perhaps it is better than the current text, which refers to a `utf8EncodeLeftAligned` function.

(I guess that a dialect-independent definition of `str` may need to refer to a type for the literal.)

docs/yul.rst
```diff
@@ -714,13 +714,15 @@ We will use a destructuring notation for the AST nodes.
...
+        where str is the string evaluation function,
+        which for the EVM dialect is defined in the section 'Literals' above
     E(G, L, n: HexNumber) = G, L, hex(n)
-        where hex is the hexadecimal decoding function
+        where hex is the hexadecimal evaluation function,
+        which turns a sequence of hexadecimal digits into their big endian value
     E(G, L, n: DecimalNumber) = G, L, dec(n),
-        where dec is the decimal decoding function
+        where dec is the decimal evaluation function,
+        which turns a sequence of decimal digits into their little endian value
```
chriseth: Why do you say little endian here?

chriseth:
Looks good apart from the endianness question - can you squash it into a single commit please?

docs/yul.rst
acoglio: It's a mistake, sorry. Just fixed it.

acoglio:
> Looks good apart from the endianness question - can you squash it into a single commit please?

That got fixed, and yes, I'll squash.

acoglio:
I've squashed the commits.

chriseth (approved):
Thanks a lot and sorry for the long back and forth!

acoglio:
@chriseth It's all good. We all want good documentation, and that takes time and thought.

PR #12070, opened 2021-09-30
Refactor CLI output selection flags

cameel:
~Depends on #12069. In draft mode until that PR is merged.~ Merged.

A small refactor to bring the structs that represent sets of boolean flags for selecting compiler outputs more in line with what I did with `DebugInfoSelection` in #11944. Now they use maps of pointer-to-members to avoid enumerating fields by hand - this gets really repetitive and error prone with structs that have 15+ fields.

This is needed for the next PR which actually changes the validation logic.

### DIFF
--- a/solc/CommandLineParser.cpp
+++ b/solc/CommandLineParser.cpp
@@ -56,10 +56,8 @@ static string const g_strAllowPaths = "allow-paths";
 static string const g_strBasePath = "base-path";
 static string const g_strIncludePath = "include-path";
 static string const g_strAsm = "asm";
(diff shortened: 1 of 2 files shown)


cameel:
#12069 has been merged so this is now reviewable.

leonardoalt (approved):
I haven't tried the binary myself, but I'm assuming you did?

cameel:
Yeah, I did some manual testing in #12074, which is on top of it. I did not test every single flag but we do have test coverage for some of these flags (in command-line tests and some in the new tests for solc) and if they're not broken, I doubt the others are.

PR #11730, opened 2021-08-03
[CLI] Deduplicate optimizer initialization

cameel:
Related to #11720.
Also implements part of #11629.

There were multiple places that initialized `OptimiserSettings`: it's needed for validation in `CommandLineParser` and later in `CommandLineInterface` to actually initialize the compiler. In both cases the initialization was performed separately for the compiler and for the assembler. Now there's a single function that creates the struct based on `CommandLineOptions`.

This change would result in optimizer settings being parsed (but unused) in linker mode so I also made the compiler reject them in this mode and also in Standard JSON mode.

PR #12071, opened 2021-09-30
Validate selected compiler outputs on the CLI

cameel:
Implements part of #11629.
~Depends on #12070. In draft mode until that PR is merged.~ Merged.

This makes the compiler report errors when outputs that are unavailable in the current input mode are selected via CLI flags.

This was originally a part of a bigger PR that enables output selection in assembly mode but it actually also makes sense on its own and can be considered a separate feature/fix.


cameel:
#12070 merged so this is now reviewable.

cameel:
This change is technically breaking, but it could be argued that it's a bugfix (see https://github.com/ethereum/solidity/pull/11730#discussion_r706150636) if we consider missing CLI validations bugs rather than missing features.

solc/CommandLineParser.cpp
```diff
@@ -460,6 +468,47 @@ bool CommandLineParser::parseLibraryOption(string const& _input)
...
+	static auto outputSupported = [](InputMode _mode, string_view _outputName)
+	{
+		static set<string> const compilerModeOutputs =
+			CompilerOutputs::componentMap() |
+			ranges::views::keys |
+			ranges::to<set>();
+
+		switch (_mode)
+		{
+		case InputMode::Compiler:
```
Marenz, line 482: Ah, I thought we need explicit `[[fallthrough]];` but if this works, great!

Marenz (approved):

solc/CommandLineParser.cpp
cameel: We only need that if there's anything between the labels.

PR #10294, opened 2020-11-14
Fix bytecode replacement in command-line tests not detecting all bytecode segments between linker references

cameel:
A trivial issue I did not notice in #10225 because there is no test with more than two unresolved references.

The global regex matches whole link references on both sides of the bytecode wchich results in this (shortened for readability):
```
0123__$abcd$__4567__$efab$_8901__$cdef$__2345
```
being replaced with
```
<BYTECODE REMOVED>__$abcd$__<BYTECODE REMOVED>__$efab$_8901__$cdef$__<BYTECODE REMOVED>
```
rather than with:
```
<BYTECODE REMOVED>__$abcd$__<BYTECODE REMOVED>__$efab$_<BYTECODE REMOVED>__$cdef$__<BYTECODE REMOVED>
```

It leaves out every other piece of bytecode because `sed` does not find overlapping matches. The fix is to match only half of the link reference.

PR #12073, opened 2021-10-01
Remove linking to unqualified library name

### LINKED ISSUE #10298: [CLI] Unqualified library name in --libraries option matches any qualified link reference with the same library name
As reported in #10290, `LinkerObject` contains a [workaround for old-style link references in `LinkerObject`](https://github.com/ethereum/solidity/blob/v0.7.4/libevmasm/LinkerObject.cpp#L78-L85) that strips file names from the link references found in the code if an exact match does not work.

The consequence is that omitting the file name in library identifiers passed to the linker (e.g. `--library Log:0x123...`) results in the library file names being completely ignored. The same address will be used not only for bare `Log` without file name but also e.g. for `math.sol:Log` and `error.sol:Log`.

### Steps to reproduce
`math.sol`:
```solidity
library Log {
    function log10() external {}
}
```

`error.sol`:
```solidity
library Log {
    function print() external {}
}
```

`contract1.sol`:
```solidity
import "error.sol";

contract C {
    function foo() public {
        Log.print();
    }
}
```

`contract2.sol`:
```solidity
import "math.sol";

contract C {
    function foo() public {
        Log.log10();
    }
}
```
```
solc contract1.sol contract2.sol --libraries "Log:0x7777777777777777777777777777777777777777" --bin
```

Output
```
======= contract1.sol:C =======
Binary:
608060405234801561001057600080fd5b5060c78061001f6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c8063c298557814602d575b600080fd5b60336035565b005b7377777777777777777777777777777777777777776313bdfacd6040518163ffffffff1660e01b815260040160006040518083038186803b158015607857600080fd5b505af4158015608b573d6000803e3d6000fd5b5050505056fea2646970667358221220c0280f7ef9d1b2ba1184ec8e799afc473ea8df12e2b9e204ea4785cfdfade0d064736f6c63430007040033

======= contract2.sol:C =======
Binary:
608060405234801561001057600080fd5b5060c78061001f6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c8063c298557814602d575b600080fd5b60336035565b005b73777777777777777777777777777777777777777763452348166040518163ffffffff1660e01b815260040160006040518083038186803b158015607857600080fd5b505af4158015608b573d6000803e3d6000fd5b5050505056fea2646970667358221220d135b215d67b4d4fbfaf7676035f21cbd1f121e1fdac65f267663e51e10cce4764736f6c63430007040033

======= error.sol:Log =======
Binary:
6084610024600b82828239805160001a607314601757fe5b30600052607381538281f3fe730000000000000000000000000000000000000000301460806040526004361060335760003560e01c806313bdfacd146038575b600080fd5b818015604357600080fd5b50604a604c565b005b56fea264697066735822122059c366c03fc74e5ecfb1420b581cd2dc6fd828f60e3b226ed273c9bbf49952ad64736f6c63430007040033

======= math.sol:Log =======
Binary:
6084610024600b82828239805160001a607314601757fe5b30600052607381538281f3fe730000000000000000000000000000000000000000301460806040526004361060335760003560e01c806345234816146038575b600080fd5b818015604357600080fd5b50604a604c565b005b56fea26469706673582212206f5b0dd0a8ba813f39f4879e5e8b73ce5d62afb1e8e597a3a28afd42e713cca364736f6c63430007040033
```

Both references were resolved to address `0x7777777777777777777777777777777777777777` without the compiler complaining about it even though they are supposed to be different libraries.

soroosh-sdi:
fix https://github.com/ethereum/solidity/issues/10298


soroosh-sdi:
It is a draft version, only want to run all CI

cameel:
The overall concept looks good. From failing tests I think you must have just missed a few cases.

The case I described in #10298 should also be added as a command-line test.

chriseth:
Is this a breaking change?

cameel:
@chriseth Technically it is since this removes support for linking the old-style references but the compiler no longer generates them so I really doubt it will break anyone's workflow. You'll only ever see any breakage if you compile and link separately, using different compiler versions. Changelog says that we switched to new-style references in 0.5.0, which means that you'd have to be compiling a 0.4.x contract and then doing `solc --link` on 0.8.10+.

cameel:
@soroosh-sdi Is this ready for review now? I switched it to draft because that's what you said in https://github.com/ethereum/solidity/pull/12073#issuecomment-931932680 and you seemed to be still working on it. Please use the `ready for review` button if this is ready.

soroosh-sdi:
> @soroosh-sdi Is this ready for review now? I switched it to draft because that's what you said in [#12073 (comment)](https://github.com/ethereum/solidity/pull/12073#issuecomment-931932680) and you seemed to be still working on it. Please use the `ready for review` button if this is ready.

The only remains thing is the test you wanted.
I'll set it to ready when I pushed a new test case

soroosh-sdi:
This PR is ready now, but I see two unrelated failures in CI. @cameel

test/cmdlineTests/linking_qualified_library_name/output
```diff
@@ -0,0 +1,8 @@
+
+======= linking_qualified_library_name/contract1.sol:C =======
+Binary:
+<BYTECODE REMOVED>
```
soroosh-sdi, line 4: Do we have a way to test this address "0x7777777777777777777777777777777777777777" existed in the bytecode?

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -100,8 +100,23 @@ vector<solidity::frontend::test::FunctionCall> TestFileParser::parseFunctionCall
...
+									expect(Token::Colon);
+									lib += ':' + m_scanner.currentLiteral();
+									expect(Token::Identifier);
+								}
+							} else {
+								expect(Token::Colon);
+								lib += ':' + m_scanner.currentLiteral();
+								expect(Token::Identifier);
+							}
+							call.signature = lib;
```
soroosh-sdi: here I parse this two cases:
1 - `filename.extension:libraryName`
2 - `:libraryName`

EDIT:

1 - `"filename":librarayName`
2 - `:libraryName`
3 - `libraryName`

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -576,6 +591,9 @@ void TestFileParser::Scanner::scanNextToken()
 		case ':':
 			selectToken(Token::Colon);
 			break;
+		case '.':
+			selectToken(Token::Period);
+			break;
```
soroosh-sdi: we need this now because `filename.extension` is part of qualified library name.

cameel: That doesn't sound right. I mean, file names ([source unit names](https://docs.soliditylang.org/en/latest/path-resolution.html#virtual-filesystem) to be precise) can contain arbitrary characters so adding support for a dot is not enough to support them in general. I think this needs a different approach. Tokenizing the name as if it was composed of valid syntax is generally a bad idea because this way it might not be preserved verbatim.

I think we should require the file name to be quoted. Then you can just expect a string token with arbitrary content. As a special case you can allow the string to be omitted if it's empty (I'd also allow omitting the semicolon in that case - this way you won't even have to change most of the existing tests).

```solidity
// library: "a.sol":L
```

@chriseth: Could you take a look at this? Do you think that requiring a string here is a good idea or would you suggest a different approach?

test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -1565,7 +1565,7 @@ BOOST_AUTO_TEST_CASE(library_call_in_homestead)
 	ALSO_VIA_YUL(
 		DISABLE_EWASM_TESTRUN()
 		compileAndRun(sourceCode, 0, "Lib");
-		compileAndRun(sourceCode, 0, "Test", bytes(), map<string, h160>{{"Lib", m_contractAddress}});
+		compileAndRun(sourceCode, 0, "Test", bytes(), map<string, h160>{{":Lib", m_contractAddress}});
```
cameel, line 1568: I think that something like ~`:Lib`~ `Lib` should still be allowed. We should just assume that the file name is empty (and it should of course not match any other name).

test/cmdlineTests/linking_unqualified_library_name/exit
```diff
@@ -0,0 +1 @@
+0
```
cameel: If the exit code is zero, you can omit this file.

test/cmdlineTests/linking_qualified_library_name/output
cameel: Well, kinda. You can't check if it's exactly that address but the script that replaces bytecode with `<BYTECODE REMOVED>` takes care to preserve any unlinked library references. See #10294. If there's no unlinked reference here, you can assume that it works and it has been replaced as intended.

test/cmdlineTests/linking_qualified_library_name/args
```diff
@@ -0,0 +1 @@
+linking_qualified_library_name/contract1.sol --libraries linking_qualified_library_name/math.sol:Log:0x7777777777777777777777777777777777777777 --bin
```
cameel: I'd put `--bin` in front to make it more visible.
```suggestion
linking_qualified_library_name/contract1.sol  --bin --libraries linking_qualified_library_name/math.sol:Log:0x7777777777777777777777777777777777777777
```

test/cmdlineTests/linking_unqualified_library_name/error.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+
+library Log {
+    function print() external {}
+}
+
```
cameel: You have superfluous empty lines at the end of some input files.

cameel (commented):
Needs some adjustments in test parsing but overall it's pretty good.

Good job finding your way around the code. I see that you're dealing with it pretty well even though it turned out to be a bit more complex than I expected (I did not expect that we'd have to adjust the parser here) :)

cameel:
> This PR is ready now, but I see two unrelated failures in CI.

Yeah, don't worry about it :) I finally have a workaround for these: #12106. You can rebase on it if you want all tests to pass but hopefully it will go into `develop` tomorrow anyway. I see that all my PRs that I rebased on it are now green.

test/libsolidity/util/TestFileParser.cpp
soroosh-sdi: Yeah, that would be more accurate solution. I'm waiting to see what @chriseth is thinking about it.

soroosh-sdi:
> Needs some adjustments in test parsing but overall it's pretty good.
> 
> Good job finding your way around the code. I see that your dealing with it pretty well even though it turned out to be a bit more complex than I expected (I did not expect that we'd have to adjust the parser here) :)

Thanks, that part is the most interesting part of this task for me :)

test/libsolidity/SolidityEndToEndTest.cpp
soroosh-sdi: This comment is a little confusing to me, `:Lib` is allowed now but `Lib` causes an error.

test/libsolidity/util/TestFileParser.cpp
soroosh-sdi: > I think we should require the file name to be quoted. Then you can just expect a string token with arbitrary content. As a special case you can allow the string to be omitted if it's empty (I'd also allow omitting the semicolon in that case - this way you won't even have to change most of the existing tests).

I implemented your suggestion, now filename is a String; 
also, I treat `L` like  `:L`.

test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -1598,7 +1598,7 @@ BOOST_AUTO_TEST_CASE(library_call_protection)
 		ABI_CHECK(callContractFunction("np(Lib.S storage)", 0), encodeArgs());
 		ABI_CHECK(callContractFunction("v(Lib.S storage)", 0), encodeArgs(m_sender));
 		ABI_CHECK(callContractFunction("pu()"), encodeArgs(2));
-		compileAndRun(sourceCode, 0, "Test", bytes(), map<string, h160>{{"Lib", m_contractAddress}});
+		compileAndRun(sourceCode, 0, "Test", bytes(), map<string, h160>{{":Lib", m_contractAddress}});
```
soroosh-sdi, line 1601: Now, I can remove all these changes.
`matchLibrary()` treat `Lib` and `:Lib` in the same way.

test/libsolidity/util/TestFileParser.cpp
chriseth: What is the alternative? Is the file name optional or required? I think it is really important to also have it optional.

soroosh-sdi: The file name is required because we need an qualified library name.
It was optional and it caused this bug https://github.com/ethereum/solidity/issues/10298
@chriseth

soroosh-sdi: If the library is in the same file as the using contract, I made it optional user can provide unqualified library name too.

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -100,8 +100,23 @@ vector<solidity::frontend::test::FunctionCall> TestFileParser::parseFunctionCall
...
+								lib = m_scanner.currentLiteral();
+								expect(Token::String);
+								expect(Token::Colon);
+								lib += ':' + m_scanner.currentLiteral();
+								expect(Token::Identifier);
+							} else if (accept(Token::Colon, true))
+							{
+								lib = ':' + m_scanner.currentLiteral();
+								expect(Token::Identifier);
+							} else {
```
chriseth: Please check your coding style. Thanks!

soroosh-sdi: done

test/cmdlineTests/linking_unqualified_library_name/error.sol
soroosh-sdi: fixed

test/cmdlineTests/linking_qualified_library_name/args
soroosh-sdi: done

test/cmdlineTests/linking_unqualified_library_name/exit
soroosh-sdi: sure

test/libsolidity/util/TestFileParser.cpp
cameel: > It was optional and it caused this bug #10298

It's not really just that it was optional. It's because it was optional **and** when it was omitted it could match any other file name.

We don't have that problem if we treat it as an empty string when it's not provided. I would not do this on the command line but in tests it's a good default in my opinion.

test/libsolidity/SolidityEndToEndTest.cpp
cameel: Sorry, my mistake. I meant `Lib`.

I think that `Lib` should be equivalent to `:Lib`. If's not a problem as long as we assume that the file name is an empty string instead of assuming that it can match any other file name.

libevmasm/LinkerObject.cpp
```diff
@@ -75,13 +75,11 @@ LinkerObject::matchLibrary(
...
-		return nullptr;
-	it = _libraryAddresses.find(_linkRefName.substr(colon + 1));
-	if (it != _libraryAddresses.end())
-		return &it->second;
+	else if (_linkRefName.size() >= 2 && _linkRefName[0] == ':')
+	{
+		it = _libraryAddresses.find(_linkRefName.substr(1));
+		if (it != _libraryAddresses.end())
+			return &it->second;
+	}
```
cameel: I think this should be the other way around. i.e. if you get something without a colon in it, prepend a colon. So if you get `L`, look for `:L`.

Also, you should do that conversion before the first lookup instead of doing it only if the first lookup fails.

test/libsolidity/SolidityEndToEndTest.cpp
cameel: Yeah, let's remove them.

test/libsolidity/semanticTests/libraries/library_address_homestead.sol
```diff
@@ -15,7 +15,7 @@ contract C {
 // ====
 // compileViaYul: also
 // ----
-// library: L
+// library: :L
```
cameel: You should be able to remove these now too.

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -100,8 +100,25 @@ vector<solidity::frontend::test::FunctionCall> TestFileParser::parseFunctionCall
...
-							expect(Token::Identifier);
+							string lib;
+							if (accept(Token::String))
+							{
+								lib = m_scanner.currentLiteral();
+								expect(Token::String);
+								expect(Token::Colon);
+								lib += ':' + m_scanner.currentLiteral();
+								expect(Token::Identifier);
+							} else if (accept(Token::Colon, true))
```
cameel:
```suggestion
							}
							else if (accept(Token::Colon, true))
```

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -100,8 +100,25 @@ vector<solidity::frontend::test::FunctionCall> TestFileParser::parseFunctionCall
 						if (accept(Token::Library, true))
 						{
 							expect(Token::Colon);
-							call.signature = m_scanner.currentLiteral();
-							expect(Token::Identifier);
+							string lib;
```
cameel:
```suggestion
							string fullyQualifiedLibraryName;
```

test/libsolidity/SolidityEndToEndTest.cpp
cameel: Looks like you already implemented that. Great! It's just a matter or removing these changes then.

test/libsolidity/util/TestFileParser.cpp
cameel: @chriseth 
> What is the alternative? Is the file name optional or required? I think it is really important to also have it optional.

I can't see any great alternatives. One would be to introduce a new type of token to represent the file name. For example `@filename`:
```
// library: @file.sol:L
```
I think that using a string is simpler though and we do not need to worry about escaping the `@`.

libevmasm/LinkerObject.cpp
soroosh-sdi: you're suggesting that I add a `colon` on every unqualified library name came from command line parser. then I don't need any conversion here
@cameel

test/libsolidity/SolidityEndToEndTest.cpp
soroosh-sdi: done

libevmasm/LinkerObject.cpp
soroosh-sdi: I implemented what you said in this commit: https://github.com/ethereum/solidity/pull/12073/commits/35d1eaeea8704ac1b21dc66492876e6c02214cc5

test/libsolidity/semanticTests/libraries/library_address_homestead.sol
soroosh-sdi: done

test/libsolidity/util/TestFileParser.cpp
soroosh-sdi: done

test/libsolidity/util/TestFileParser.cpp
soroosh-sdi: done

libevmasm/LinkerObject.cpp
cameel: > you're suggesting that I add a `colon` on every unqualified library name came from command line parser

I assumed that you could still get a library name without colon here from tests but if not then removing the conversion here is fine too.

solc/CommandLineParser.cpp
```diff
@@ -464,6 +464,8 @@ bool CommandLineParser::parseLibraryOption(string const& _input)
 				serr() << "Invalid address for library \"" << libName << "\": " << addrString << endl;
 				return false;
 			}
+			if (libName.find(':') == string::npos)
+				libName = ":" + libName;
```
cameel: Let's make it an error. User won't be able to specify an empty name for a file containing a library on the command line anyway.

```suggestion
			if (libName.find(':') == string::npos)
			{
				serr() << "Expected a fully qualified library name but got \"" << libName << "\". ";
				serr() << "A fully qualified name must be prepended with the name of the source unit where the library is defined." << endl;
				return false;
			}
```

test/cmdlineTests/linking_strict_assembly_no_file_name_in_link_reference/input.yul
```diff
@@ -1,6 +1,6 @@
 object "a" {
     code {
-        let addr := linkersymbol("L")
+        let addr := linkersymbol(":L")
```
cameel: Please leave this one unchanged. It's specifically supposed to check what happens if there is no file name. It's ok if the test fails - your change is actually supposed to make it fail.

cameel: And I think we should test all four combinations:
1. `L` on the command-line + `linkersymbol("L")` in the file: should print a validation error saying that `L` is unqualified.
2. `L` on the command-line + `linkersymbol(":L")` in the file: should print a validation error saying that `L` is unqualified.
3. `:L` on the command-line + `linkersymbol("L")` in the file: should leave an unlinked reference.
4. `:L` on the command-line + `linkersymbol(":L")` in the file: should link correctly.

This tests covers (2). Please add new tests for the others. Naming could be something like `linking_strict_assembly_<qualified|unqualified>_library_<qualified|unqualified>_reference`.

test/libsolidity/SolidityEndToEndTest.cpp
cameel: I see they are still present. Was there a problem removing them?

libevmasm/LinkerObject.cpp
soroosh-sdi: yeah I get library name without colon from `SolidiyEndToEndTest.cpp`.
let's assume the `_libraryAddresss`  only contains qualified addresses; 
Now we only need to add colon in the command line parser.

cameel: Yeah, I think this should now not make any assumptions about colons. Overall, the way I see the system after this PR is this:
- ~In `--libraries` we always require the colon: `file.sol:L=0x...`. If you do not include a colon, you get a validation error. We could technically allow unqualified names (and make `L` equivalent to `:L` like you did) but empty file names are not possible on the command line so it's pointless and it's much more likely that the user just forgot to include the file name.~ In `--libraries` we allow both `L` and `:L` and consider them different. `L` should not match `:L` used in a `linkersymbol()`.
- In Standard JSON's `settings.libraries` option there's no way to omit the file name because there's no colon syntax (file and name are just separate keys) so there's no problem there.
    - EDIT: That actually makes it impossible to link Yul if it uses something like `linkersymbol("L")`. Maybe we should disallow unqualified names after all - but that's a separate issue so let's not deal with it here.
- In `linkersymbol()` we allow anything - we just not make any assumptions about what is in the name and therefore qualified and unqualified names are considered different. I.e. `linkersymbol("L")` and `linkersymbol(":L")` are different symbols. This is because this can be used in Yul and Yul might be just written by hand in which case there is no Solidity file  to specify. The `:` convention is only a thing at the Soldity level.
- In `LinkerObject` above we also do not make any assumptions - we support both unqualified symbols coming directly from Yul and qualified symbols from Solidity.
- In tests we want `:L` to be a shorthand for `L` because there we often use empty file names and it's all on the Solidity level so unqualified names won't occur.

solc/CommandLineParser.cpp
soroosh-sdi: I thought you wanted to treat both the following cases in same way:
- `:Lib`
- `Lib`

Also, after applying your suggestion I should add colon to the `semanticTests`.

test/libsolidity/SolidityEndToEndTest.cpp
soroosh-sdi: yeah, new we need them.
we assumed all library names in the map is fully qualified

solc/CommandLineParser.cpp
cameel: Well, it's a bit complicated. We want them to be the same in tests because we're already using `L` as a shorthand for `:L` there. But in other places, e.g. in `linkersymbol()` we consider them different. See https://github.com/ethereum/solidity/pull/12073#discussion_r728867808. There I summarized my reasoning and listed all the cases and explained where it's supposed to be different and where it's not.

test/libsolidity/SolidityEndToEndTest.cpp
cameel: ok. You're right. They are passed to `LinkerObject` and how it needs them qualified.

These don't matter all that much so it's ok to require a colon here in my opinion. The place where we really don't want colons are the `.sol` files with tests. We should add them somewhere in the class that gets the libraries from the parser but before `SolidityExecutionFramework::compileAndRun()` gets them.

libevmasm/LinkerObject.cpp
soroosh-sdi: > In linkersymbol() we allow anything - we just not make any assumptions about what is in the name and therefore qualified and unqualified names are considered different. I.e. linkersymbol("L") and linkersymbol(":L") are different symbols. This is because this can be used in Yul and Yul might be just written by hand in which case there is no Solidity file to specify. The : convention is only a thing at the Soldity level.

In this page of Yul documentation:
https://docs.soliditylang.org/en/v0.8.9/yul.html#linkersymbol

> Its first and only argument must be a string literal and represents the fully qualified library name used with the --libraries option.

test/cmdlineTests/linking_strict_assembly_no_file_name_in_link_reference/input.yul
soroosh-sdi: I added the tests

soroosh-sdi:
can you take a look at this PR? @cameel

libevmasm/LinkerObject.cpp
cameel: Good catch! That's my mistake in the docs :) `linkersymbol()` accepts a fully qualified library name but does not require it. It's just that [the new codegen will always insert a fully qualified one](https://github.com/ethereum/solidity/blob/v0.8.9/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L3217-L3221). If you look at how [`linkersymbol` builtin is implemented in libyul](https://github.com/ethereum/solidity/blob/v0.8.9/libyul/backends/evm/EVMDialect.cpp#L151-L159), you'll see that there's nothing there that actually requires a colon. It simply passes the whole string through several layers down to [`Assembly::newPushLibraryAddress`()](https://github.com/ethereum/solidity/blob/v0.8.9/libevmasm/Assembly.cpp#L358-L363) and that just computes a keccak256 hash.

But yeah, this means that we cannot really require the colon in `--libraries` in assembly mode. We can only do that in compiler mode so for simplicity it might be better to not require it at all. `L` on the command-line should be considered different from `:L` just like in `linkersymbol()`. I'll update my comment above to reflect that.

I'll submit a PR fixing the docs in a moment.

cameel: Here's the documentation update: #12144.

Also, I have created an issue about not being able to use unqualified references in Standard JSON: #12145. Maybe forbidding unqualified references in Yul is really the best way to solve this but it's a breaking change so in this PR we still have to support them.

test/cmdlineTests/linking_strict_assembly_no_file_name_in_link_reference/input.yul
cameel: Your comment (https://github.com/ethereum/solidity/pull/12073#discussion_r729068696) made me realize that we can't actually forbid the colon in assembly so unfortunately (1) and (2) need to be changed:
1. `L` on the command-line + `linkersymbol("L")` in the file: should link correctly.
2. `L` on the command-line + `linkersymbol(":L")` in the file: should leave an unlinked reference.
3. `:L` on the command-line + `linkersymbol("L")` in the file: should leave an unlinked reference.
4. `:L` on the command-line + `linkersymbol(":L")` in the file: should link correctly.

test/cmdlineTests/linking_strict_assembly_qualified_library_qualified_refrence/args
```diff
@@ -0,0 +1 @@
+--strict-assembly --libraries :L=0x1234567890123456789012345678901234567890
```
cameel: #11944 has been merged recently so now we can use `--debug-info none` to get a bit more compact assembly output. Please use it in all the new tests you added.
```suggestion
--strict-assembly --libraries :L=0x1234567890123456789012345678901234567890 --debug-info none
```

test/cmdlineTests/linking_strict_assembly_qualified_library_unqualified_refrence/output
```diff
@@ -0,0 +1,26 @@
+
+======= linking_strict_assembly_qualified_library_unqualified_refrence/input.yul (EVM) =======
```
cameel: You have a typo in the directory name: `refrence` -> `reference`.

test/libsolidity/util/TestFileParser.cpp
```diff
@@ -100,8 +100,26 @@ vector<solidity::frontend::test::FunctionCall> TestFileParser::parseFunctionCall
...
+								expect(Token::Identifier);
+							}
+							else if (accept(Token::Colon, true))
+							{
+								fullyQualifiedLibraryName = ':' + m_scanner.currentLiteral();
+								expect(Token::Identifier);
+							}
+							else
+							{
+								fullyQualifiedLibraryName = ":" + m_scanner.currentLiteral();
```
cameel: Like I said in https://github.com/ethereum/solidity/pull/12073/files#r728880975, the colon should really be added somewhere in the class that uses the parser, not in the parser itself. Otherwise it won't be possible to discern qualified and unqualified references if we need to in the future.

test/cmdlineTests/linking_strict_assembly_qualified_library_qualified_refrence/args
soroosh-sdi: I added that to new `linlking_strict_assembly_*` tests

test/cmdlineTests/linking_strict_assembly_qualified_library_unqualified_refrence/output
soroosh-sdi: oops

test/libsolidity/util/TestFileParser.cpp
soroosh-sdi: I moved this logic to `SematicTest.cpp`.

libevmasm/LinkerObject.cpp
soroosh-sdi: > In --libraries we allow both L and :L and consider them different. L should not match :L used in a linkersymbol().

command line parser adjusted to accept unqualified library name.

soroosh-sdi: > Here's the documentation update: #12144.

Thanks for the doc update.

> Also, I have created an issue about not being able to use unqualified references in Standard JSON: #12145. Maybe forbidding unqualified references in Yul is really the best way to solve this but it's a breaking change so in this PR we still have to support them.

I agree, it is more clear if we only accept qualified library names everywhere.

soroosh-sdi:
please review again @cameel

test/cmdlineTests.sh
```diff
@@ -474,11 +474,11 @@ rm -r "$SOLTMPDIR"
...
+echo '' | msg_on_error --no-stdout "$SOLC" - --link --libraries :a=0x90f20564390eAe531E810af625A22f51385Cd222
+echo '' | "$SOLC" - --link --libraries :a=0x80f20564390eAe531E810af625A22f51385Cd222 &>/dev/null && exit 1
 
 printTask "Testing long library names..."
-echo '' | msg_on_error --no-stdout "$SOLC" - --link --libraries aveeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerylonglibraryname=0x90f20564390eAe531E810af625A22f51385Cd222
+echo '' | msg_on_error --no-stdout "$SOLC" - --link --libraries :aveeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerylonglibraryname=0x90f20564390eAe531E810af625A22f51385Cd222
```
cameel: These changes should no longer be necessary now.

test/cmdlineTests/linking_unqualified_library_name/contract2.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-3.0
```
cameel, line 1: You can remove the empty `err` file from this test dir.

test/libsolidity/SemanticTest.cpp
```diff
@@ -383,7 +384,10 @@ TestCase::TestResult SemanticTest::runTest(
 			soltestAssert(
 				deploy(test.call().signature, 0, {}, libraries) && m_transactionSuccessful,
 				"Failed to deploy library " + test.call().signature);
-			libraries[test.call().signature] = m_contractAddress;
+			if (test.call().signature.find(':') == string::npos)
```
cameel, line 390: I'd add a comment to explain why we're doing it.

```suggestion
			# For convenience, in semantic tests we assume that an unqualified name like `L` is equivalent to one
			# with an empty source unit name (`:L`). This is fine because the compiler never uses unqualified
			# names in the Yul code it produces and does no allow `linkersymbol()` at all in inline assembly.
			if (test.call().signature.find(':') == string::npos)
```

cameel (commented):
It's good now. I'd approve already but the changelog entry is missing.

I also have three suggestions but that's just to make it a bit cleaner - the fix itself is already correct in my opinion.

test/cmdlineTests.sh
soroosh-sdi: you're right, I removed them

test/cmdlineTests/linking_unqualified_library_name/contract2.sol
soroosh-sdi: done

test/libsolidity/SemanticTest.cpp
soroosh-sdi: done

test/libsolidity/SemanticTest.cpp
```diff
@@ -383,7 +384,13 @@ TestCase::TestResult SemanticTest::runTest(
 			soltestAssert(
 				deploy(test.call().signature, 0, {}, libraries) && m_transactionSuccessful,
 				"Failed to deploy library " + test.call().signature);
-			libraries[test.call().signature] = m_contractAddress;
+			// For convenience, in semantic tests we assume that an unqualified name like `L` is equivalent to one
+			// with an empty source unit name (`:L`). This is fine because the compiler never uses unqualified
+			// names in the Yul code it produces and does no allow `linkersymbol()` at all in inline assembly.
```
cameel: Looks like I made a typo. Sorry for that.

```suggestion
			// names in the Yul code it produces and does not allow `linkersymbol()` at all in inline assembly.
```

cameel (commented):
What about the changelog entry? I mean [`Changelog.md`](https://github.com/ethereum/solidity/blob/develop/Changelog.md).

test/libsolidity/SemanticTest.cpp
soroosh-sdi: done

soroosh-sdi:
> What about the changelog entry? I mean [`Changelog.md`](https://github.com/ethereum/solidity/blob/develop/Changelog.md).

Just added that

Changelog.md
```diff
@@ -20,7 +20,7 @@ Bugfixes:
  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
  * TypeChecker: Fix internal error when using user defined value types in public library functions.
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
-
+ * Linking: Fix matching link reference name, like `a.sol:L`, with an unqualified library name, like `L`, from command line interface.
```
cameel: I'd be a bit more specific. Also, it applies to command-line interface:
```suggestion
 * Commandline Interface: When linking do not treat unqualified library names (e.g. ``L``) as matching the same name qualified with any source unit name (e.g. ``xyz.sol:L``).
```

cameel (dismissed):
I'd tweak that changelog entry a bit but other than that it's all good!

Changelog.md
soroosh-sdi: done

cameel (dismissed):

soroosh-sdi:
Could you please merge this PR?
@cameel 
EDIT: Oh, I see you requested a review from Chirseth. so I can wait.

cameel:
Yeah, I wanted at least one more person to have a look at it before we merge it, mainly because of the parser change and the stuff with `:L` vs `L` on the command line.

cameel:
The changelog entry I suggested turns out to be a bit hard to understand so it needs to be adjusted:
```
Commandline Interface: When linking only accept exact matches for library names passed to the ``--libraries`` option. Library names not prefixed with a file name used to match any library with that name.
```
After that this can be merged.

cameel (approved):

PR #12074, opened 2021-10-01
CLI output selection in assembler mode

### LINKED ISSUE #3870: [CLI] Do not ignore the output selection in assembler mode
I am using the `--assemble` mode for transforming assembly-syntax into binary representation through a tool.

This works good except for the fact that the output of the compiler is not really machine-readable as it just dumps everything to stdout and stderr.

It would be nice if `solc` would care about the `--bin` flag that is already present if run in assemble mode (`--assemble`).

This way, one would not have to use a regex over the data in stdout in order to retrieve the binary representation.

Current behaviour:
```
➜  /tmp solc --assemble --bin temp_2.asm

======= temp_2.asm (EVM) =======

Pretty printed source:
{
    0x40
    0x50
    add
    pop
}

Binary representation:
604060500150

Text representation:
    /* "temp_2.asm":3:7   */
  0x40
    /* "temp_2.asm":9:13   */
  0x50
    /* "temp_2.asm":15:18   */
  add
    /* "temp_2.asm":20:23   */
  pop
```

Desired behaviour:
```
➜  /tmp solc --assemble --bin temp_2.asm
604060500150
```

cameel:
Resolves #3870.
~Depends on #12071. In draft mode until that PR is merged.~ Merged.

This does not add any new outputs, just makes the ones that are already available respond to output selection flags so e.g. `--ir` is not available. We can 

Also, to make this non-breaking, all outputs are selected if no flags are specified, which is different from how the compiler mode behaves. If this goes through, I can create a separate PR to `breaking` branch to change that.


cameel:
#12071 merged so this is now reviewable.

cameel:

solc/CommandLineInterface.cpp
```diff
@@ -1013,34 +1013,54 @@ bool CommandLineInterface::assemble(yul::AssemblyStack::Language _language, yul:
...
 
 		if (_language != yul::AssemblyStack::Language::Ewasm && _targetMachine == yul::AssemblyStack::Machine::Ewasm)
 		{
 			stack.translate(yul::AssemblyStack::Language::Ewasm);
 			stack.optimize();
 
-			sout() << endl << "==========================" << endl;
-			sout() << endl << "Translated source:" << endl;
-			sout() << stack.print() << endl;
+			// TODO: This isn't ewasm but it's only present when we're doing Yul->EWASM translation.
```
leonardoalt: I guess this is not going to be done in this PR? Maybe create an issue so it doesn't get forgotten?

solc/CommandLineParser.cpp
```diff
@@ -497,6 +504,16 @@ bool CommandLineParser::parseOutputSelection()
 	for (auto&& [optionName, outputComponent]: CompilerOutputs::componentMap())
 		m_options.compiler.outputs.*outputComponent = (m_args.count(optionName) > 0);
 
+	if (m_options.input.mode == InputMode::Assembler && m_options.compiler.outputs == CompilerOutputs{})
+	{
+		// In assembly mode keep the default outputs enabled for backwards-compatibility.
+		// TODO: Remove this (must be done in a breaking release).
```
leonardoalt, line 487: Maybe also create issue?

leonardoalt (dismissed):

solc/CommandLineInterface.cpp
cameel: Will do.

solc/CommandLineParser.cpp
cameel: Will do.

solc/CommandLineInterface.cpp
cameel: Actually, I just realized that ewasm is still experimental. So maybe it's not breaking after all?

solc/CommandLineParser.cpp
cameel: Actually, I just realized that ewasm is still experimental. So maybe it's not breaking after all?

solc/CommandLineInterface.cpp
cameel: Actually, I just realized that ewasm is still experimental. So maybe it's not breaking after all?

solc/CommandLineParser.cpp
cameel: #12230.

solc/CommandLineInterface.cpp
cameel: Got a confirmation on the chat. It's not breaking.

In that case I'm adding an extra commit that adds an output called `--ewasm-ir` (I could not come up with a better name).

leonardoalt (dismissed):

leonardoalt:
wait, do the tests need update?

cameel:
I did update them but looks like I missed one change needed after a rebase. Apparently an extra `stop` instruction gets added after `sstore` due to #11974. There are other similar changes in that PR so it seems legit.

Should be fixed now.

leonardoalt (approved):

PR #12075, opened 2021-10-01
New Wikipedia feature implemented in Solidity docs!

SlashHash:
Hello everyone! Not many people knew about this (including me) but recently the Inuka team at the Wikimedia Foundation released Wikipedia previews for external websites, meaning that any website can implement this and have the Wikipedia page previews work on their website too! (Wikipedia page previews are the page previews that pop up when you hover over certain terms while browsing the Wikipedia on a computer, they also work on mobile but in a different manner).
Anyway, the point is: I think it would be a great addition to the Solidity docs, it would mean that the reader wouldn't have to open a 
  Wikipedia page every time they want to get a quick grasp of a concept or just to get a refresher, basically it has the same purpose as it does inside Wikipedia.
A few things to note are: 
1. I've never used reStructuredText before so the change in code I'm proposing may be totally or partially broken, so feel free to make changes if necessary.
2. I did test the script elements part of the code with developer tools override and it worked wonderfully.
3. My implementation doesn't make design changes to the Hyperlink, it is not the same as in the "Wikimedia's Demo" link below, the Hyperlinks stay the same, the only thing that changes is when you hover over (or click, on mobile) a Wikipedia Hyperlink a Wikipedia page preview is shown. 

I will now be leaving some relevant links to this PR below.

Wikimedia's blog post on the feature: https://diff.wikimedia.org/2021/07/29/new-wikipedia-preview-feature-provides-context-from-wikipedia-across-the-web/

Official GitHub repo for the feature being discussed: https://github.com/wikimedia/wikipedia-preview

Wikimedia's Demo: https://wikimedia.github.io/wikipedia-preview/demo/articles/english.html

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -4,3 +4,12 @@
     {{ super() }}
 	<a href="{{ pathto('genindex') }}">Keyword Index</a>
   {% endblock %}
+
+  {%- block extrahead %}
+	<script src="https://unpkg.com/wikipedia-preview@1.4.0/dist/wikipedia-preview.production.js"></script>
+	  <script>
+            wikipediaPreview.init({
+              detectLinks: true
+            })
+	  </script>
+  {% endblock %}


docs/_templates/layout.html
```diff
@@ -4,3 +4,12 @@
...
 	<a href="{{ pathto('genindex') }}">Keyword Index</a>
   {% endblock %}
+
+  {%- block extrahead %}
+	<script src="https://unpkg.com/wikipedia-preview@1.4.0/dist/wikipedia-preview.production.js"></script>
+	  <script>
+            wikipediaPreview.init({
+              detectLinks: true
+            })
+	  </script>
```
cameel, line 14: In Firefox I get a crash at `wikipedia-preview.production.js:3:134782` when I build docs locally and open them in the browser:

```
Uncaught TypeError: e is null
    v https://unpkg.com/wikipedia-preview@1.4.0/dist/wikipedia-preview.production.js:3
    Ye https://unpkg.com/wikipedia-preview@1.4.0/dist/wikipedia-preview.production.js:3
    <anonymous> file:///tmp/_build/html/smtchecker.html:29
```

docs/_templates/layout.html
```diff
@@ -4,3 +4,12 @@
     {{ super() }}
 	<a href="{{ pathto('genindex') }}">Keyword Index</a>
   {% endblock %}
+
+  {%- block extrahead %}
+	<script src="https://unpkg.com/wikipedia-preview@1.4.0/dist/wikipedia-preview.production.js"></script>
```
cameel, line 9: Not sure if it's a good idea to have the docs depend on external resources like this. It would be better if it could be downloaded when the docs get built and included in `html_js_files` in `docs/conf.py`.

At the very least this should have the [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) `integrity` attribute with a hash to ensure that the file that gets served is the one we built the docs with.

docs/_templates/layout.html
cameel, line 14: The `<script>` has no `async` or `defer` attributes so if a request to unpkg.com is slow for any reason (can be just a bad connection), it will block loading of any scripts that come after it. I think this should have `async` attribute and the `init()` call should be in some callback that fires when the script is already loaded.

docs/_templates/layout.html
```diff
@@ -4,3 +4,12 @@
...
+	  <script>
+            wikipediaPreview.init({
+              detectLinks: true
+            })
+	  </script>
+  {% endblock %}
```
cameel, line 15: There's a mix of tabs and spaces here.

cameel (commented):
Thanks for the PR.

Honestly, I'm not sure we really need this - we have like 20 wikipedia links total in the entirety of the docs and I think that docs being pretty static and not having too many JS bells and whistles is a good thing. That's just my own opinion though - let's see what others on the team think.

In the meantime, here are a few technical remarks on the change itself.

docs/_templates/layout.html
SlashHash: Hi, thank you for your feedback!
I'm sorry the code is really bad, front-end is not really what I usually do, I will admit though that the mix of tabs and spaces was on purpose haha, I just didn't think much of it but I guess it is pretty clumsy.
As for the testing, of course that error has to be fixed but if you just want to have a feel of how the page works with that script, this is how I did it: 
1. Open developer tools. 2. Go to Sources. 3. Click on the Overrides tab (probably collapsed) 4. Click on 'Enable Local Overrides' 5. Choose a folder. 6 Go back to the page tab on Sources. 7. Copy and paste the code in the html. 8. Do 'Ctrl + S'. 9. Press F5. 
I just want to say I will be doing all the necessary, but can we just be clear on whether this is going to actually be a implemented or not before I do so? I don't entirely know how the process works to be honest, if there needs to be other people's opinions for a decision or not. That's all for now, thanks again!

chriseth:
I think we should not pull in external scripts for such a tiny feature.

hrkrshnn:
Not sure if this is currently working. For example,  hovering over "C3 linearization" in https://860310-40892817-gh.circle-artifacts.com/0/docs-html/contracts.html isn't producing a preview.

docs/_templates/layout.html
SlashHash: I'm sorry, I misunderstood your response, only now am I realizing that you meant for the preview to be downloaded beforehand and not the code so it doesn't ever rely on external queries.
I'm up to code that, the only problem I see with that is that Wikipedia changes, and we may not always be up-to-date with the latest definitions, what I could do is a PR to update the previews (if there are any to be updated) every 30 days or so. Are you on board with that idea @cameel?

chriseth:
This is a nice feature, @SlashHash and thank you of making us aware of it, but I think the cost-benefit ratio is not good enough for this project. We have too few links to wikipedia and it does not justify pulling in an external dependency.

Issue #11877, opened 2021-09-01
Tests verifying that compilation with `--experimental-via-yul` produces same EVM assembly as two-stage compilation via Yul


cameel:
The bytecode from `solc --bin --experimental-via-ir` should be identical to the one obtained by getting the output of `solc --ir` and then assembling it with `solc --strict-assembly`. Add some command line-tests to verify that.

Depends on  #11858 - this issue adds a command-line flag to disable printing Solidity snippets in EVM assembly.
Related to #11876. Until it's done, `--experimental-via-ir` flag must be added to `solc --ir` in order to get identical metadata.
Related to #11494.
#11875 contains the commands you could use (though the `--asm` output needs some processing to extract just the EVM assembly).

cameel:
I added some tests in #12141. It does not test any multi-file contracts or files containing multiple contracts and @chriseth thinks it's important to include those as well so I'm going to do a followup PR to add them.

PR #12076, opened 2021-10-01
Error helpers + small refactors in `cmdlineTests.sh`

cameel:
A bunch of small tweaks to improve error reporting extracted out of my upcoming PR for #11877.

The biggest part is a helper Bash function that takes a command, and, if that command fails, prints stdout, stderr and exits. Doing this manually after every command adds a lot of boilerplate. I'll have 5 commands to handle so at this point it's more efficient to just write one reusable helper. It's also already useful in `cmdlineTests.sh` because there are a lot of places where we expect success and stderr gets sent to `/dev/null`. The helper will make debugging test failures there easier.

Other than that there are a few small refactors: `fail` helper, removal of `--force` from `rm`, unnecessary `set -e` in a few places and one missing `rm -r "$SOLTMPDIR"` (finally my `/tmp` won't be filled with garbage after running these tests). I have also enabled `set -o pipefail` - looks like it does not cause any spurious failures in this script.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -39,6 +39,74 @@ function fail()
     return 1
 }
 
+function msg_on_error()
+{
+    local error_message
+    local no_stdout=false
+    local no_stderr=false
+
+    while [[ $1 =~ ^-- ]]
+    do
+        case "$1" in
+            --msg)
+                error_message="$2"
+                shift
+                shift
+                ;;
+            --no-stdout)
+                no_stdout=true
+                shift
+                ;;
+            --no-stderr)
+                no_stderr=true
+                shift
+                ;;
+            --silent)
+                no_stdout=true
+                no_stderr=true
+                shift
+                ;;
+            *)
+                fail "Invalid option for msg_on_error: $1"
+                ;;
+        esac
+    done
+
+    local command=("$@")
+
+    local stdout_file stderr_file
+    stdout_file="$(mktemp -t cmdline_test_command_stdout_XXXXXX.txt)"
+    stderr_file="$(mktemp -t cmdline_test_command_stderr_XXXXXX.txt)"
+
+    if "${command[@]}" > "$stdout_file" 2> "$stderr_file"
+    then
+        [[ $no_stdout == "true" ]] || cat "$stdout_file"
+        [[ $no_stderr == "true" ]] || >&2 cat "$stderr_file"
+        rm "$stdout_file" "$stderr_file"
+        return 0
+    else
+        printError "Command failed: $SOLC ${command[*]}"
+        if [[ -s "$stdout_file" ]]
+        then
+            printError "stdout:"
+            >&2 cat "$stdout_file"
+        else
+            printError "stdout: <EMPTY>"
+        fi
+        if [[ -s "$stderr_file" ]]
+        then
+            printError "stderr:"
+            >&2 cat "$stderr_file"
+        else
+            printError "stderr: <EMPTY>"
+        fi
+
+        printError "$error_message"
+        rm "$stdout_file" "$stderr_file"
+        return 1
+    fi
+}
+
 safe_kill()
 {
     local PID=${1}
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -33,6 +33,12 @@ else
     function printLog() { echo "$(tput setaf 3)$1$(tput sgr0)"; }
 fi
 
+function fail()
+{
+    printError "$@"
+    return 1
+}
+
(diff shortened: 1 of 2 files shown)


test/cmdlineTests.sh
```diff
@@ -471,52 +469,50 @@ SOLTMPDIR=$(mktemp -d)
...
 
 printTask "Testing long library names..."
-echo '' | "$SOLC" - --link --libraries aveeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerylonglibraryname=0x90f20564390eAe531E810af625A22f51385Cd222 >/dev/null
+echo '' | msg_on_error --no-stdout "$SOLC" - --link --libraries aveeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerylonglibraryname=0x90f20564390eAe531E810af625A22f51385Cd222
 
 printTask "Testing linking itself..."
 SOLTMPDIR=$(mktemp -d)
 (
     cd "$SOLTMPDIR"
-    set -e
```
chriseth, line 488: Why can this be removed?

scripts/common.sh
```diff
@@ -33,6 +33,79 @@ else
     function printLog() { echo "$(tput setaf 3)$1$(tput sgr0)"; }
 fi
 
+function fail() {
```
leonardoalt:
```suggestion
function fail()
{
```

scripts/common.sh
```diff
@@ -33,6 +33,79 @@ else
...
+    printError "$@"
+    return 1
```
chriseth, line 39: is this essentially the same as `exit 1`?

leonardoalt, line 39: Why return instead of exit?

test/cmdlineTests.sh
cameel: Because this enables exiting on errors and we already do that at the top of the file. I suspect this is just leftover after a `set +e`/`set -e` pair that might have been here in the past.

scripts/common.sh
cameel: Same but better :) With `exit 1` you immediately exit from the script. With `return 1` you just return with a non-zero status giving the calling script control over what happens next. If the script has `set -e` (which all of our scripts either do or should be doing), it will exit by default but you can also silence the error with `|| true` or `set +e` and do some error handling if you need to.

Basically, this makes the error work more like throwing an exception which bubbles up the stack than a hard exit.

cameel: Answered in https://github.com/ethereum/solidity/pull/12076#discussion_r721323664 above.

test/cmdlineTests.sh
chriseth: My main question here is: Does the subshell `(...)` reset the flags or not?

cameel: Entering the subshell does not reset them. The `set -e` at the beginning of the script will get inherited by the subshells. Also, any `set` invocations done inside subshell will not affect the rest of the script.

`set -e` in this subshell has no effect.

leonardoalt (approved):

PR #12078, opened 2021-10-02
Snap: update cvc4 to 1.8

come-maiz:


come-maiz:
with this change we don't have to build antlr, which was failing on non amd64 architectures. Using the ubuntu packages is easier, and I hope will fix the errors.

cameel (dismissed):
Looks good.

Upgrading cvc is a good idea. I think we should generally use the latest one and looks like 1.8 is the latest. Not having to build antlr is a good thing too.

Coding style checks are failing in CI. Looks like you need to configure your editor to remove the trailing whitespace automatically.

cameel:
@leonardoalt looks like the [CVC4 repo](https://github.com/CVC4/CVC4-archived) has been archived back in May and they're releasing [CVC5](https://github.com/cvc5/cvc5) now. Do we support both? Should we suggest using CVC5 now in docs?

come-maiz:
Thanks for the review. I've fixed the error.

I don't think cvc5 has a stable release yet: https://github.com/cvc5/cvc5/releases

cameel (approved):

cameel:
Ah, ok then.

leonardoalt:
> @leonardoalt looks like the [CVC4 repo](https://github.com/CVC4/CVC4-archived) has been archived back in May and they're releasing [CVC5](https://github.com/cvc5/cvc5) now. Do we support both? Should we suggest using CVC5 now in docs?

Our CVC4Interface might not work with CVC5 because of a different ABI, so I'd say we just stay with CVC4 for now.

PR #12080, opened 2021-10-03
adding new example

phyBrackets:


chriseth (commented):
I generally like the idea of such a contract, and it is also really short!
If we call it "crowdfunding", though, maybe the mechanism could be redesigned to fund a specific project from the beginning, but the project is split into milestones where payout is only performed when a milestone is met (decided through voting). Plus, I think there should be a mechanism to refund every contributor individually at any time.

Also, why did you base the voting on contributions and not simply on amount contributed?

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+    constructor(uint _goal, uint _deadline) {
+        goal = _goal;
+        deadline = block.timestamp + _deadline;
+        admin = msg.sender;
+        minimumContribution = 100 wei;
+    }
+    
+   
+    modifier onlyAdmin() {
+        require(msg.sender == admin, "Only admin can execute this");
```
hrkrshnn, line 72: Could you replace all the revet strings with custom errors?

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+
+.. code-block:: solidity
+    :force:
+
+    // SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.8.5;
+ 
+contract CrowdFunding {
+    mapping(address => uint) public contributors;
+    address public admin;
```
hrkrshnn: Can be converted into immutable.

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+    uint public noOfContributors;
+    uint public minimumContribution;
+    uint public deadline; //timestamp
```
hrkrshnn, line 37: Maybe a good choice for user defined value types?

Something like `type Timestamp is uint256;`, then `Timestamp public deadline`. Maybe this can also be immutable? If not, usually timestamps are shorter than `256` bits so that it can be packed.

https://blog.soliditylang.org/2021/09/27/user-defined-value-types/

hrkrshnn (commented):
I generally like this example, I think rewriting this to use custom errors and user-defined value types would be great. This can be a showcase for the new features.

docs/examples/crowd-funding.rst
phyBrackets: I actually don't get this , that what is the benefit of this using here.

docs/examples/crowd-funding.rst
phyBrackets: I know using custom errors with if statements , but I don't know how to do that with modifiers. Could you share any resource link?

docs/examples/crowd-funding.rst
hrkrshnn: Good question. Better abstraction?

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+        contributors[msg.sender] += msg.value;
+        raisedAmount += msg.value;
+        
+        emit ContributeEvent(msg.sender, msg.value);
+    }
+    
+ 
+    function getBalance() public view returns(uint) {
+        return address(this).balance;
+    }
```
hrkrshnn, line 96: I think this is unnecessary, this is easy to obtain off chain. 

Also, can you change `public` to `external` whenever that is sufficient?

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+    
+ 
+    // a contributor can get a refund if goal was not reached within the deadline
+    function getRefund() public {
+        require(block.timestamp > deadline, "Deadline has not passed");
+        require(raisedAmount < goal, "The goal was met");
```
hrkrshnn, line 102: I think it's better to have an enum to represent the state, and the current state can be a state variable.

docs/examples/crowd-funding.rst
```diff
@@ -0,0 +1,153 @@
...
+        require(contributors[msg.sender] > 0);
+        
+        address payable recipient = payable(msg.sender);
+        uint value = contributors[msg.sender];
+        recipient.transfer(value);
+        // equivalent to:
+        // payable(msg.sender).transfer(contributors[msg.sender]);
```
hrkrshnn, line 109: Why not use this instead of the temporary local variable?

RusAlex:
Is not this recommended in the documentation to use all state changes
before the external contract calls or transfering ether calls.

To avoid a reentranc attacks.

Just my thoughts.

hrkrshnn:
It seems that https://github.com/ethereum/solidity/pull/12092 replaced this PR. Closing this. @phyBrackets Feel free to reply if that is not the case.

PR #12082, opened 2021-10-04
Control flow side effects of user defined functions

chriseth:


chriseth:
This is working now apart from recursive function calls. I'm actually wondering if it would be better to finish implementing `ControlFlowBuilder` on the full Yul AST and then doing reachability analysis there.

chriseth:
Ok actually I think ControlFlowBuilder cannot be really used, but essentially something that actually builds a graph.

stackenbotten:
There was an error when running `chk_coding_style` for commit `d908dd900d602299d8e5d859a2827733947f94f8`:
```
test/libyul/controlFlowSideEffects/recursion.yul:32:// c: 
test/libyul/controlFlowSideEffects/recursion.yul:33:// d: 
test/libyul/controlFlowSideEffects/recursion.yul:35:// x: 
test/libyul/controlFlowSideEffects/recursion.yul:36:// y: 
test/libyul/controlFlowSideEffects/recursion.yul:37:// z: 

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `6a6ed4e381d83b4ab998cabf2a4b14c2ce94091e`:
```
test/libyul/controlFlowSideEffects/recursion.yul:32:// c: 
test/libyul/controlFlowSideEffects/recursion.yul:33:// d: 
test/libyul/controlFlowSideEffects/recursion.yul:35:// x: 
test/libyul/controlFlowSideEffects/recursion.yul:36:// y: 
test/libyul/controlFlowSideEffects/recursion.yul:37:// z: 

```

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -230,7 +230,7 @@ void ControlFlowGraphBuilder::operator()(ExpressionStatement const& _exprStmt)
 	//       not only for builtins.
 	if (auto const* funCall = get_if<FunctionCall>(&_exprStmt.expression))
 		if (BuiltinFunction const* builtin = m_dialect.builtin(funCall->functionName.name))
-			if (builtin->controlFlowSideEffects.terminates)
```
chriseth, line 232: @ekpyron the above comment can now be implemented.

chriseth:
I think it might really make sense to cache the side-effects in the context (same is true for the other side-effects as well).
If optimizer steps are semantically-equivalent transformations, then it should be OK to compute the side-effects once per optimizer loop iteration.

stackenbotten:
There was an error when running `chk_coding_style` for commit `d5a397a680fdefabefe23289dfdb9ed0084f1a01`:
```
test/libyul/controlFlowSideEffects/recursion.yul:32:// c: 
test/libyul/controlFlowSideEffects/recursion.yul:33:// d: 
test/libyul/controlFlowSideEffects/recursion.yul:35:// x: 
test/libyul/controlFlowSideEffects/recursion.yul:36:// y: 
test/libyul/controlFlowSideEffects/recursion.yul:37:// z: 

```

libyul/ControlFlowSideEffects.h
```diff
@@ -19,21 +19,28 @@
 #pragma once
 
 #include <set>
+#include <stack>
+#include <optional>
```
hrkrshnn: All these includes are unnecessary, aren't they? Even `set`.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include <range/v3/view/reverse.hpp>
+
+using namespace std;
+using namespace solidity::yul;
+
+void ControlFlowBuilder::operator()(FunctionCall const& _functionCall)
+{
+	walkVector(_functionCall.arguments | ranges::views::reverse);
+	newConnectedNode();
+	m_currentNode->functionCall = {_functionCall.functionName.name};
```
hrkrshnn:
```suggestion
	m_currentNode->functionCall = _functionCall.functionName.name;
```
The braces are a bit misleading for the optional type.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,272 @@
...
+	}
+	m_currentNode = finalNode;
+}
+
+void ControlFlowBuilder::operator()(FunctionDefinition const& _function)
+{
+	ScopedSaveAndRestore currentNode(m_currentNode, nullptr);
+	ScopedSaveAndRestore leaveNode(m_leave, nullptr);
+	ScopedSaveAndRestore breakNode(m_break, nullptr);
+	ScopedSaveAndRestore continueNode(m_continue, nullptr);
```
hrkrshnn: Are these necessary?

libyul/ControlFlowSideEffectsCollector.h
```diff
@@ -0,0 +1,123 @@
...
+{
+	std::vector<ControlFlowNode const*> successors;
+	std::optional<YulString> functionCall;
+};
+
+struct FunctionFlow
+{
+	ControlFlowNode const* entry;
+	ControlFlowNode const* exit;
+};
```
hrkrshnn: This could use some documentation.

libyul/ControlFlowSideEffects.h
```diff
@@ -19,21 +19,28 @@
...
-	bool terminates = false;
-	/// If true, this code reverts all state changes in the transaction.
-	/// Whenever this is true, ``terminates`` has to be true as well.
-	bool reverts = false;
+	/// If true, the function contains at least one reachable branch that terminates successfully.
+	bool canTerminate = false;
+	/// If true, the function contains at least one reachable branch that reverts.
+	bool canRevert = false;
+	/// If true, the function has a regular outgoing control-flow.
+	bool canContinue = true;
```
hrkrshnn, line 41: I think this can be replaced by `willTerminate` and that sounds more appropriate.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,272 @@
...
+	while (it != end)
+	{
+		ControlFlowNode const* node = *it;
+		if (!node->functionCall || exitKnownReachable(*node->functionCall))
+		{
+			m_pendingNodes[_functionName].erase(it);
+			return node;
+		}
+		++it;
+	}
```
hrkrshnn: This can be replaced by `std::find`, followed by an erase.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+	m_functionFlows[_function.name] = move(flow);
+}
+
+void ControlFlowBuilder::operator()(ForLoop const& _for)
+{
+	(*this)(_for.pre);
+
+	ScopedSaveAndRestore scopedBreakNode(m_break, nullptr);
+	ScopedSaveAndRestore scopedContinueNode(m_continue, nullptr);
```
hrkrshnn, line 103: Shouldn't this be done before the visit to `pre`? I don't have an example to show why this is bad, but also don't see why it should be done after pre.

libyul/backends/evm/EVMDialect.cpp
```diff
@@ -57,8 +57,20 @@ pair<YulString, BuiltinFunctionForEVM> createEVMFunction(
 	f.parameters.resize(static_cast<size_t>(info.args));
 	f.returns.resize(static_cast<size_t>(info.ret));
 	f.sideEffects = EVMDialect::sideEffectsOfInstruction(_instruction);
-	f.controlFlowSideEffects.terminates = evmasm::SemanticInformation::terminatesControlFlow(_instruction);
-	f.controlFlowSideEffects.reverts = evmasm::SemanticInformation::reverts(_instruction);
+	if (evmasm::SemanticInformation::terminatesControlFlow(_instruction))
+	{
+		f.controlFlowSideEffects.canContinue = false;
```
hrkrshnn, line 62: About changing to `willTerminate` or even just `terminates`, it feels more appropriate here.

libyul/ControlFlowSideEffects.h
chriseth: I'm not sure. These three bools are essentially edges to special nodes in the control-flow graph, and I would prefer not to invert the meaning of one of them.

Also `canContinue` can be false due to the function being always recursive.

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: Maybe not if we assume the hoister, but better safe than sorry.

libyul/ControlFlowSideEffects.h
hrkrshnn: > Also canContinue can be false due to the function being always recursive.

This is not currently done, is it?

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: Like this? Would you prefer it?
```
	auto it = m_pendingNodes[_functionName].begin();
	auto end = m_pendingNodes[_functionName].end();
	it = find_if(it, end, [](ControlFlowNode const* node) {
		return !node->functionCall || exitKnownReachable(*node->functionCall);
	}
	if (it == end)
		return nullptr;
	else
	{
		ControlFlowNode const* node = *it;
		m_pendingNodes[_functionName].erase(it);
		return node;
	}
```

libyul/ControlFlowSideEffects.h
chriseth: It is

chriseth: https://github.com/ethereum/solidity/pull/12082/files#diff-206aa4eb462f2802b4a49004b969fa9ac56299142ff16692fc99eb8bfec42acbR182

libyul/ControlFlowSideEffectsCollector.cpp
hrkrshnn: Yes :) But it did look better when I thought about the change.

libyul/ControlFlowSideEffectsCollector.cpp
hrkrshnn: Yeah, okay. We could assert that `m_continue` and `m_break` are `nullptr` if we really want to be extra safe. Same with `m_leave` and `m_break`.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,269 @@
...
+	ScopedSaveAndRestore currentNode(m_currentNode, nullptr);
+	ScopedSaveAndRestore leaveNode(m_leave, nullptr);
+	ScopedSaveAndRestore breakNode(m_break, nullptr);
+	ScopedSaveAndRestore continueNode(m_continue, nullptr);
+
+	FunctionFlow flow;
+	flow.exit = newNode();
+	m_currentNode = newNode();
+	flow.entry = m_currentNode;
+	newConnectedNode();
```
hrkrshnn: I think this might be redundant.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+	(*this)(_for.post);
+	m_currentNode->successors.emplace_back(loopNode);
+
+	m_currentNode = breakNode;
+}
+
+void ControlFlowBuilder::operator()(Break const&)
+{
+	m_currentNode->successors.emplace_back(m_break);
```
hrkrshnn, line 132: Maybe a `yulAssert(m_break, "")` here? Similarly for `continue` and `leave`.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,269 @@
...
+void ControlFlowBuilder::operator()(Continue const&)
+{
+	m_currentNode->successors.emplace_back(m_continue);
+	m_currentNode = newNode();
+}
+
+void ControlFlowBuilder::operator()(Leave const&)
+{
+	m_currentNode->successors.emplace_back(m_leave);
+	m_currentNode = newNode();
```
hrkrshnn, line 147: I wonder if this is really needed. Effectively, edges coming from here doesn't matter, I think.

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: I think you are right.

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: I think we discussed this offline but if you don't disconnect the node, then `function f() { leave revert(0, 0) }` will have `can revert`. With dead code removal this might not matter, but I think it is still good to have.

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,271 @@
...
+{
+	std::list<ControlFlowNode const*>& nodes = m_pendingNodes[_functionName];
+	auto it = ranges::find_if(nodes, [this](ControlFlowNode const* _node) {
+		return !_node->functionCall || sideEffects(*_node->functionCall).canContinue;
+	});
+	if (it == nodes.end())
+		return nullptr;
+
+	ControlFlowNode const* node = *it;
+	m_pendingNodes[_functionName].erase(it);
```
hrkrshnn:
```suggestion
	nodes.erase(it);
```

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,271 @@
...
+		m_functionSideEffects[name] = {false, false, false};
+	}
+
+	// Process functions while we have progress. For now, we are only interested
+	// in `canContinue`.
+	bool progress = true;
+	while (progress)
+	{
+		progress = false;
+		for (auto const& fun: m_pendingNodes)
```
hrkrshnn: This may lead to undefined behaviour. The call `procesFunction(...)` can update `m_pendingNodes` in two different ways:

1. `nextProcessableNode(...)` can remove the current iterator; wouldn't that lead to undefined behaviour when the current one gets incremented?
2. `recordReachabilityAndQueue` can add to the beginning of `m_pendingNodes`. Why is it guaranteed that this node will get processed?

Would it be more appropriate to do something like `while(!m_pendingNodes.empty()) { node = m_pendingNodes.front(); ...}`

---

Also, otherwise could change to `m_pendingNodes | range::views::keys)`

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -0,0 +1,271 @@
...
+			else
+				calledSideEffects = &m_functionSideEffects.at(_function);
+
+			if (calledSideEffects->canTerminate)
+				sideEffects.canTerminate = true;
+			if (calledSideEffects->canRevert)
+				sideEffects.canRevert = true;
+
+			if (m_functionCalls.count(_function))
+				for (YulString callee: m_functionCalls.at(_function))
```
hrkrshnn: We could replace the `.count ` followed by `.at` by `valueOrDefault` from CommonData.h

libyul/ControlFlowSideEffectsCollector.h
```diff
@@ -0,0 +1,129 @@
...
+	/// and if it is a function call, records that `_functionName` calls
+	/// `*_node->functionCall`.
+	void recordReachabilityAndQueue(YulString _functionName, ControlFlowNode const* _node);
+
+	Dialect const& m_dialect;
+	ControlFlowBuilder m_cfgBuilder;
+	std::map<YulString, ControlFlowSideEffects> m_functionSideEffects;
+	std::map<YulString, std::list<ControlFlowNode const*>> m_pendingNodes;
+	std::map<YulString, std::set<ControlFlowNode const*>> m_processedNodes;
+	std::map<YulString, std::set<YulString>> m_functionCalls;
```
hrkrshnn, line 126: This could use documentation. Wasn't obvious that this is a set of reachable function calls inside a function.

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: I'm not sure I follow, I think this might be a misunderstanding on your part: `m_pendingNodes` is a map by function name - so there is one queue per function name. As long as we don't add or remove a function, the iterator here is not invalidated. I can split `recordReachabilityAndQueue` into two functions:
1) one that can also add a new function
2) one that asserts that the function is already there
and use the first one above in line 171 and the second one in line 232

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: I think we should either assert or save and restore. What would you prefer?

hrkrshnn: I would prefer the assert.

libyul/ControlFlowSideEffectsCollector.cpp
hrkrshnn: Thanks for the change. Was a misunderstanding.

hrkrshnn (dismissed):

hrkrshnn:
Some commits are left for squashing, can be merged afterwards.

hrkrshnn (approved):

PR #11371, opened 2021-05-10
Use a local instead of a temporary as an argument to valueOrDefault

bshastry:
Fixes https://github.com/ethereum/solidity/issues/11343

PR #10973, opened 2021-02-17
Function Call Graph v2

cameel:
This PR replaces #10332.

I fixed the problems I listed in https://github.com/ethereum/solidity/pull/10332#issuecomment-777617557 and I think I have implemented all the changes that actually affect what's in the graph. ~I'm still not done with some refactors suggested in the old PR like using a queue when visiting nodes adjusting some names or changing the way checks are done. I'm pushing what I have now as a draft but these changes are pretty small so I think I'll be done. The PR should at least already be correct and pass all tests.~ The PR is now ready.

PR #12091, opened 2021-10-05
Use side effects of user-defined functions in other optimizer steps.

chriseth:


chriseth:
I think it might really make sense to cache the side-effects in the context (same is true for the other side-effects as well).
If optimizer steps are semantically-equivalent transformations, then it should be OK to compute the side-effects once per optimizer loop iteration.

chriseth:
Performance does not seem to be as big an issue as I thought.

bshastry:
Fuzzer update:

There is a stack-use-after-scope here: https://github.com/ethereum/solidity/blob/863a0d3b9c4b180dc0b149def38384316df0ec2c/libyul/ControlFlowSideEffectsCollector.cpp#L215

(see https://github.com/ethereum/solidity/pull/10973#discussion_r580926054 for discussion on a similar issue)

I am not sure if YulString as an rvalue reference temporary is the issue. At least one of similar issues in the past could be fixed by replacing temporary with a value on the stack. (see https://github.com/ethereum/solidity/pull/11371)

chriseth:
Before merging this, some tests may need to be modified so that they are still purposeful.

chriseth:
Changed the tests to be meaningful, please double-check.

chriseth:
I think this could need some more tests in the relevant optimizer steps.

libyul/optimiser/Semantics.cpp
```diff
@@ -182,8 +182,19 @@ pair<TerminationFinder::ControlFlow, size_t> TerminationFinder::firstUncondition
...
+		holds_alternative<Assignment>(_statement) &&
+		containsNonContinuingFunctionCall(*std::get<Assignment>(_statement).value)
+	)
+		return ControlFlow::Terminate;
+	else if (
 		holds_alternative<ExpressionStatement>(_statement) &&
-		isTerminatingBuiltin(std::get<ExpressionStatement>(_statement))
+		containsNonContinuingFunctionCall(std::get<ExpressionStatement>(_statement).expression)
 	)
 		return ControlFlow::Terminate;
```
ekpyron, line 199: Not that we need it right away, but in general it makes sense to distinguish between terminating and reverting here, doesn't it?
```
function f() { return(0,0) }
function g() { revert(0,0) }
{
  switch calldataload(0)
  case 0 {
    sstore(0, 0) // needs to stay
    f()
  }
  case 1 {
    sstore(0, 0) // can be removed
    g()
  }
}
```

ekpyron (approved):
Assuming that the side effects are correct :-), this looks good!

I'm wondering if it's worth to catch this eventually:
```
for {} 1 {} {
  if g() { return(...) }
}
sstore(0, 1) // can be removed
```
but no need now in any case.

PR #12092, opened 2021-10-05
New updated example

phyBrackets:
https://github.com/ethereum/solidity/pull/12080 In addition to this , I update the things as much as i can!


leonardoalt:
Were the comments in https://github.com/ethereum/solidity/pull/12080 fixed? I think some were, mainly related to the user defined functions. Indentation is a bit off, which I also think makes the tests fail.

leonardoalt:
It looks like the comments from the other PR were also not applied here. I'm closing for now for lack of activity, please reopen if the changes are applied.

PR #12093, opened 2021-10-06
More strongly recommend to use the latest version

fulldecent:
The publish Solidity security policy should be helpful to people when deciding which version of Solidity to use. So a reference is added.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,8 +17,10 @@ With Solidity you can create contracts for uses such as voting, crowdfunding, bl
 and multi-signature wallets.
 
 When deploying contracts, you should use the latest released
-version of Solidity. This is because breaking changes as well as
-new features and bug fixes are introduced regularly. We currently use
+version of Solidity. Apart from exceptional cases, only the latest version receives
+`security fixes<https://github.com/ethereum/solidity/security/policy#supported-versions>`.
+Furthermore, breaking changes as well as
+new features are introduced regularly. We currently use
 a 0.x version number `to indicate this fast pace of change <https://semver.org/#spec-item-4>`_.
 
 .. warning::


hrkrshnn (dismissed):
Thanks. This is fine by me. I will wait to see what others in the team think about making the recommendation stronger.

docs/index.rst
```diff
@@ -21,10 +21,10 @@ user-defined types among other features.
 With Solidity you can create contracts for uses such as voting, crowdfunding, blind auctions,
 and multi-signature wallets.
 
-When deploying contracts, you should use the latest released
-version of Solidity. This is because breaking changes as well as
-new features and bug fixes are introduced regularly. We currently use
-a 0.x version number `to indicate this fast pace of change <https://semver.org/#spec-item-4>`_.
+When deploying contracts, you should only use the latest released version of Solidity.
```
chriseth: I think this reads (and has read before) a bit weird - maybe we should instead emphasize that you should use the latest version because only it is guaranteed to gets security fixes (sometimes we also fix older versions)?

fulldecent:
My goal here is to make this language as strong as the Solidity team will allow. My preference is:

> Use the latest released version of Solidity for your projects. DO NOT USE older versions as they have known problems and/or may not receive fixes.

chriseth:
Totally fine, @fulldecent  - my comment was mainly about restructuring the paragraph so that it is clearer why we only recommend the latest version.

cameel (dismissed):
I think the wording is ok now and the text does reflect our general policy.

cameel:
@fulldecent Can you squash your commits and rebase on the latest `develop` (which has a workaround for the failing `b_osx` job)?

docs/index.rst
```diff
@@ -21,10 +21,9 @@ user-defined types among other features.
 With Solidity you can create contracts for uses such as voting, crowdfunding, blind auctions,
 and multi-signature wallets.
 
-When deploying contracts, you should use the latest released
-version of Solidity. This is because breaking changes as well as
-new features and bug fixes are introduced regularly. We currently use
-a 0.x version number `to indicate this fast pace of change <https://semver.org/#spec-item-4>`_.
+Use the latest released version of Solidity for your projects. Do not use older versions as
+they have known problems and/or may `not receive fixes<https://github.com/ethereum/solidity/security/policy#supported-versions>`.
+We currently use a 0.x version number `to indicate this fast pace of change <https://semver.org/#spec-item-4>`_.
```
chriseth:
```suggestion
When deploying contracts, you should use the latest released
version of Solidity. Up to exceptional cases, only the latest version receives `security fixes<https://github.com/ethereum/solidity/security/policy#supported-versions>`.
Furthermore, breaking changes as well as
new features are introduced regularly. We currently use
a 0.x version number `to indicate this fast pace of change <https://semver.org/#spec-item-4>`_.
```

chriseth: The last line in the way it is written now does not really fit anymore.

chriseth (approved):

cameel (approved):

PR #11645, opened 2021-07-12
[Trivial] Provide a better error reporting for failed cmdline tests.

Marenz:

PR #12097, opened 2021-10-06
Less verbose errors from `compileFull()`

cameel:
Recently I've been getting oddly verbose output from command-line tests in case of some failures. Turns out that it's because #11645 made `compileFull()` print content of input files on failure - and in case of tests like `corion` the content is very long. This PR makes it print only file names.

I also noticed that shell globs used to find input files sometimes results in literal `./*/*` being passed to the compiler, which is apparently the reason why `--ignore-missing` flag is used. I fixed that by replacing the glob with `find`.

### DIFF
--- a/scripts/common_cmdline.sh
+++ b/scripts/common_cmdline.sh
@@ -20,7 +20,7 @@
 # ------------------------------------------------------------------------------
 
 YULARGS=(--strict-assembly)
-FULLARGS=(--optimize --ignore-missing --combined-json "abi,asm,ast,bin,bin-runtime,devdoc,hashes,metadata,opcodes,srcmap,srcmap-runtime,userdoc")
+FULLARGS=(--optimize --combined-json "abi,asm,ast,bin,bin-runtime,devdoc,hashes,metadata,opcodes,srcmap,srcmap-runtime,userdoc")
 OLDARGS=(--optimize --combined-json "abi,asm,ast,bin,bin-runtime,devdoc,interface,metadata,opcodes,srcmap,srcmap-runtime,userdoc")
 function compileFull()
 {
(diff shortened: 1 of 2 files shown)


chriseth (approved):

Marenz:
I found it very cumbersome to always having to manually find & output the file that failed, especially when it is a generated one in the tmp file (as does the docs tester).. so actually not so happy with that change..

I would have preferred that it maybe skips it if it's too long

cameel:
It prints the full path so it's just a matter of selecting it, typing `cat` and then pasting.

I have nothing against printing it when it's short but when the compiler throws an ICE in the `corion` test, the output takes multiple screens. At first I was confused as to what was happening and thought that something must have broken the script again. I didn't even realize that this was intentional and not a bug (even though I reviewed the original PR). Maybe you could add that back but with a check that the output is not longer than 10-15 lines?

Marenz:
> It prints the full path so it's just a matter of selecting it, typing cat and then pasting.

Sure, but imagine doing that, modifying one line, trying again, etc.. it doesn't go well if you do rapid changes & test them for an ongoing period... The best would maybe even be a direct offer to modify the file.. but that's more work for the docs tests..

> Maybe you could add that back but with a check that the output is not longer than 10-15 lines?

Sounds like a good idea!

cameel:
Actually, what is your use case here? Is it for debugging code snippets in docs or something else? I mean, it **is** less convenient but I find myself having to do that all the time in various situations, which is why I didn't think this one is particularly bad.

Marenz:
I didn't remember the exact reason at first, but it just came to me: In the docs files, multiple code samples are in the same file. The extracted test cases are all extra files with the same name and just a different hash (based on the specific code example). 
It wasn't always easy to find out which one of the examples was the one failing, thus I added the output.

PR #12100, opened 2021-10-07
Add `require(amount > 0)`

lukehutch:
Add missing check... this is a huge gaping security hole, and it's present in one of the very first code examples you show...

### DIFF
--- a/docs/introduction-to-smart-contracts.rst
+++ b/docs/introduction-to-smart-contracts.rst
@@ -118,6 +118,7 @@ registering with a username and password, all you need is an Ethereum keypair.
         // Sends an amount of existing coins
         // from any caller to an address
         function send(address receiver, uint amount) public {
+            require(amount > 0);
             if (amount > balances[msg.sender])
                 revert InsufficientBalance({
                     requested: amount,


docs/introduction-to-smart-contracts.rst
```diff
@@ -118,6 +118,7 @@ registering with a username and password, all you need is an Ethereum keypair.
         // Sends an amount of existing coins
         // from any caller to an address
         function send(address receiver, uint amount) public {
+            require(amount > 0);
```
hrkrshnn, line 121: `amount` is an unsigned integer. So it cannot be 'negative'. This excludes `amount = 0`. I don't think that is a good idea. For reference, from https://eips.ethereum.org/EIPS/eip-20:

> Note Transfers of 0 values MUST be treated as normal transfers and fire the Transfer event.

hrkrshnn (changes requested):
I think the change is unnecessary. Closing this for now. Feel free to reply if you have questions.

docs/introduction-to-smart-contracts.rst
lukehutch: I just realized this was a `uint`, and came here to close the PR. However, I agree that allowing `amount = 0` is not a great idea (though it is not a security problem).

PR #12106, opened 2021-10-07
Include OS and CPU architecture in CircleCI cache keys

cameel:
Recently [`b_osx`](https://app.circleci.com/pipelines/github/ethereum/solidity/19527/workflows/1cda61f8-788b-4fad-aa1a-a71e71696c04/jobs/865363) job has been failing in some of our PRs. Specifically external PRs that have been inactive for some time.

```
Failed to unarchive cache

Error untarring cache: Error extracting tarball /var/folders/1b/gl7yt7ds26vcyr1pkgld6l040000gn/T/cache3415993642 : usr/local/Homebrew/Library/Homebrew/shims/scm: Can't remove already-existing dir tar: Error exit delayed from previous errors. : exit status 1
```

["Error untarring cache" when running a macOS job](https://support.circleci.com/hc/en-us/articles/360056920051--Error-untarring-cache-when-running-a-macOS-job) suggests to include `{{ arch }}` in cache key. The reason for failures is probably that there has been an OS update after these PRs were last updated and the cache is simply no longer valid.

This PR adds `{{ arch }}` to cache keys on macOS and also on Windows just in case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -627,14 +627,14 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - dependencies-osx-{{ checksum ".circleci/osx_install_dependencies.sh" }}
+            - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
       # DO NOT EDIT between here and save_cache, but rather edit ./circleci/osx_install_dependencies.sh
       # WARNING! If you do edit anything here instead, remember to invalidate the cache manually.
       - run:
           name: Install build dependencies
           command: ./.circleci/osx_install_dependencies.sh
       - save_cache:
-          key: dependencies-osx-{{ checksum ".circleci/osx_install_dependencies.sh" }}
+          key: dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
           paths:
             - /usr/local/bin
             - /usr/local/sbin
@@ -663,7 +663,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - dependencies-osx-{{ checksum ".circleci/osx_install_dependencies.sh" }}
+            - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
       - attach_workspace:
           at: .
       - run: *run_soltest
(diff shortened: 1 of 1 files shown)


chriseth (dismissed):

chriseth:
Looks good for now.

cameel:
Looks like `{{ arch }}` did not help. It must be something else.

chriseth:
Wait but the build worked, did you forget to add arch to the test run?

cameel:
Hmm... I'm guessing that build went fine because the key was not present in the cache. The `t_` jobs restore it and that's what went wrong. Not sure why though. I'm trying to SSH into the machine but for some reason I get "permission denied".

cameel:
[`b_osx` rerun](https://app.circleci.com/pipelines/github/ethereum/solidity/19530/workflows/85104dc6-8eb5-4bf9-94f4-3a98ad9a00a0/jobs/865837) is failing too which confirms that there's nothing special about `t_` jobs here.

cameel:
Back to draft. I need to experiment with it a bit.

cameel:
We'll have to wait for CircleCI for a proper fix (I think this is a problem in `restore_cache` and not in our config) but I have a workaround. Removing `/usr/local/Homebrew/Library/Homebrew/shims/scm/` lets `restore_cache` finish without errors.

The dir contains only two files: `git` (some wrapper script over git) and `svn` (symlink to `git`). Removing them should be safe since they will be restored from cache anyway.

I'm leaving the `{{ arch }}` workaround in too. We probably do not need it but it does not break anything and this way we won't run into the problem described in the article.

cameel:
After digging deeper, I think that this is a bug in `restore_cache` that got triggered by a recent change in Homebrew. https://github.com/Homebrew/brew/pull/12170 replaced  the `/usr/local/Homebrew/Library/Homebrew/shims/scm/` directory with a symlink to `shared/` in the same parent dir. CircleCI has an older version of Homebrew already installed in the image but we install (and cache) a newer one. `restore_cache` runs `tar` to unpack the cache, which refuses to overwrite an existing dir with a symlink during unpacking. In this case the overwrite is legit (we want everything replaced with the cached content) but `restore_cache` does not take this scenario into account and just lets `tar` fail.

.circleci/config.yml
```diff
@@ -160,6 +160,17 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
+  - steps_restore_cache_homebrew_workaround: &steps_restore_cache_homebrew_workaround
+      steps:
+        - run:
+            # FIXME: For some reason restore_cache fails saying that it cannot remove the scm/ dir.
```
christianparpart, line 166: Do you mean the `rm -r` below? If so, what about `rm -rf`?

cameel: No, the `rm` **is** the workaround :) Works without `--force`.

It's `restore_cache` that can't remove the files and it's provided by CircleCI so we have to wait for a fix from them.

chriseth (approved):

PR #12115, opened 2021-10-11
[docs] Getters: add example using a complex getter

CrimsonGlory:
In the getters section of the documentation a simple example is given (with uint), then another simple example (with array), and then a quite complex example (called ``Complex``) that has a public mapping from uint to other mapping from bool to an array of a struct containing uint, a mapping, and dynamic and fixed size arrays of different types (uint and bytes)).

At least for me, it wasn't quite easy to grasp ``Complex`` at first (for example, it wasn't that intuitive what exactly was ``arg3``).

This PR adds a simple contract with loads some data and use the getter to retrieve it, so that ``Complex`` is easier to grasp.

### DIFF
--- a/docs/contracts/visibility-and-getters.rst
+++ b/docs/contracts/visibility-and-getters.rst
@@ -209,3 +209,39 @@ to select individual struct members or provide a key for the mapping:
         b = data[arg1][arg2][arg3].b;
         e = data[arg1][arg2][arg3].e;
     }
+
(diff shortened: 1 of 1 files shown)


chriseth:
Thanks for your pull request! While the complex example takes a while to understand and you creating code for it certainly helped in understanding it, I fear that just adding the example here does not really help others. Thanks a lot anyways!

PR #12118, opened 2021-10-11
Separate input modes for `--help`,  `--license` and `--version`

cameel:
Solves part of #11731.
Alternative approach for #11800.

To properly finish #11800 I need to remove the `exit()` from the code that prints license and version. Doing this complicates the control flow and I think that the right solution is to change that flow and actually treat `--help`, `--license`, and `--version` as separate input modes. This way we can properly move them to `CommandLineInterface` and only do actual parsing in `CommandLineParser`.


cameel:

leonardoalt (approved):

PR #12120, opened 2021-10-11
`solThrow()`

cameel:
Another smaller PR extracted from #11800.

This one adds a macro similar to `assertThrow()` but meant for actual error handling rather than for checking assertions. It can replace the more verbose `BOOST_THROW_EXCEPTION` in cases where the exception inherits from `util::Exception` and the only information we add to it is a message.

There's intentionally no variant without a message because I think that exceptions that interrupt the program should always have a message attached.


cameel:

chriseth (approved):

PR #12121, opened 2021-10-11
Extend using statement

### LINKED ISSUE #9211: Extend "using for"
This is a proposal to extend the "using for" directive in the following ways:

 - allow it at file level to be valid for all contracts and functions in that file
 - allow its first argument to be a free function
 - allow its first argument to be an imported module
 - allow its first argument to be `*`

The semantics are as follows:

The directive `using * for ...` matches all free functions in the current file (including non-scoped free functions imported from somewhere else), but it does not match functions in libraries or functions imported scoped into a module (`moduleName.functionName`).

The directive `using functionName for ...` matches the specified function. Here, `functionName` can be a path like `moduleName.functionName`.

The directive `using moduleName for ...` matches all free functions in the imported module (via `import "fileName" as moduleName`) including free functions imported from that module at file level. Here, `moduleName` can be a path like `moduleA.moduleB`.

Note that most of the new uses are incompatible with flatteners.

---
Update (2021-09-20):

Extend the `using A for B` statement everywhere to allow `A` to also be
 - a list of file-level functions in curly braces.
 - the name of a module
 - a single file-level function
 - `*`

If used at file-level, `B` cannot be `*`.

If used at file-level also allow `export using A for B`, when `B` is a type declared at file-level in the current file.

Clarification: In the above, all "items" can be paths (`A.B.f`) and don't have to be single identifiers.

hrkrshnn:
Fixes https://github.com/ethereum/solidity/issues/9211

### DIFF
--- a/docs/contracts/using-for.rst
+++ b/docs/contracts/using-for.rst
@@ -6,71 +6,89 @@
 Using For
 *********
 
-The directive ``using A for B;`` can be used to attach library
-functions (from the library ``A``) to any type (``B``)
-in the context of a contract.
+The directive ``using A for B;`` can be used to attach
+functions (``A``) as member functions to any type (``B``).
 These functions will receive the object they are called on
 as their first parameter (like the ``self`` variable in Python).
 
-The effect of ``using A for *;`` is that the functions from
-the library ``A`` are attached to *any* type.
+It is valid either at file level or inside a contract,
+at contract level.
 
-In both situations, *all* functions in the library are attached,
+The first part, ``A``, can be one of:
+
+- a list of file-level or library functions (``using {f, g, h, L.t} for uint;``) -
+  only those functions will be attached to the type.
+- the name of a library (``using L for uint;``) -
+  all functions (both public and internal ones) of the library are attached to the type
+
+At file level, the second part, ``B``, has to be an explicit type (without data location specifier).
+Inside contracts, you can also use ``using L for *;``,
+which has the effect that all functions of the library ``L``
+are attached to *all* types.
+
+If you specify a library, *all* functions in the library are attached,
 even those where the type of the first parameter does not
 match the type of the object. The type is checked at the
 point the function is called and function overload
 resolution is performed.
 
+If you use a list of functions (``using {f, g, h, L.t} for uint;``),
+then the type (``uint``) has to be implicitly convertible to the
+first parameter of each of these functions. This check is
+performed even if none of these functions are called.
+
 The ``using A for B;`` directive is active only within the current
-contract, including within all of its functions, and has no effect
-outside of the contract in which it is used. The directive
-may only be used inside a contract, not inside any of its functions.
+scope (either the contract or the current module/source unit),
+including within all of its functions, and has no effect
+outside of the contract or module in which it is used.
 
 Let us rewrite the set example from the
-:ref:`libraries` in this way:
+:ref:`libraries` section in this way, using file-level functions
+instead of library functions.
 
 .. code-block:: solidity
 
     // SPDX-License-Identifier: GPL-3.0
-    pragma solidity >=0.6.0 <0.9.0;
+    pragma solidity ^0.8.13;
 
-
-    // This is the same code as before, just without comments
     struct Data { mapping(uint => bool) flags; }
+    // Now we attach functions to the type.
+    // The attached functions can be used throughout the rest of the module.
+    // If you import the module, you have to
+    // repeat the using directive there, for example as
+    //   import "flags.sol" as Flags;
+    //   using {Flags.insert, Flags.remove, Flags.contains}
+    //     for Flags.Data;
+    using {insert, remove, contains} for Data;
 
-    library Set {
-        function insert(Data storage self, uint value)
-            public
-            returns (bool)
-        {
-            if (self.flags[value])
-                return false; // already there
-            self.flags[value] = true;
-            return true;
-        }
+    function insert(Data storage self, uint value)
+        returns (bool)
+    {
+        if (self.flags[value])
+            return false; // already there
+        self.flags[value] = true;
+        return true;
+    }
 
-        function remove(Data storage self, uint value)
-            public
-            returns (bool)
-        {
-            if (!self.flags[value])
-                return false; // not there
-            self.flags[value] = false;
-            return true;
-        }
+    function remove(Data storage self, uint value)
+        returns (bool)
+    {
+        if (!self.flags[value])
+            return false; // not there
+        self.flags[value] = false;
+        return true;
+    }
 
-        function contains(Data storage self, uint value)
-            public
-            view
-            returns (bool)
-        {
-            return self.flags[value];
-        }
+    function contains(Data storage self, uint value)
+        public
+        view
+        returns (bool)
+    {
+        return self.flags[value];
     }
 
 
     contract C {
-        using Set for Data; // this is the crucial change
         Data knownValues;
 
         function register(uint value) public {
@@ -82,12 +100,13 @@ Let us rewrite the set example from the
         }
     }
 
-It is also possible to extend elementary types in that way:
+It is also possible to extend built-in types in that way.
+In this example, we will use a library.
 
 .. code-block:: solidity
 
     // SPDX-License-Identifier: GPL-3.0
-    pragma solidity >=0.6.8 <0.9.0;
+    pragma solidity ^0.8.13;
 
     library Search {
         function indexOf(uint[] storage self, uint value)
@@ -100,9 +119,9 @@ It is also possible to extend elementary types in that way:
             return type(uint).max;
         }
     }
+    using Search for uint[];
 
     contract C {
-        using Search for uint[];
         uint[] data;
 
         function append(uint value) public {
--- a/docs/grammar/SolidityParser.g4
+++ b/docs/grammar/SolidityParser.g4
@@ -12,6 +12,7 @@ options { tokenVocab=SolidityLexer; }
 sourceUnit: (
 	pragmaDirective
 	| importDirective
+	| usingDirective
 	| contractDefinition
 	| interfaceDefinition
 	| libraryDefinition
@@ -311,10 +312,10 @@ errorDefinition:
 	Semicolon;
 
 /**
- * Using directive to bind library functions to types.
- * Can occur within contracts and libraries.
+ * Using directive to bind library functions and free functions to types.
+ * Can occur within contracts and libraries and at the file level.
  */
-usingDirective: Using identifierPath For (Mul | typeName) Semicolon;
+usingDirective: Using (identifierPath | (LBrace identifierPath (Comma identifierPath)* RBrace)) For (Mul | typeName) Semicolon;
 /**
  * A type name can be an elementary type, a function type, a mapping type, a user-defined type
  * (e.g. a contract or struct) or an array type.
--- a/libsolidity/analysis/DeclarationTypeChecker.cpp
+++ b/libsolidity/analysis/DeclarationTypeChecker.cpp
@@ -25,6 +25,7 @@
 #include <liblangutil/ErrorReporter.h>
 
 #include <libsolutil/Algorithms.h>
+#include <libsolutil/Visitor.h>
 
 #include <range/v3/view/transform.hpp>
 
@@ -451,12 +452,39 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 
 bool DeclarationTypeChecker::visit(UsingForDirective const& _usingFor)
 {
-	ContractDefinition const* library = dynamic_cast<ContractDefinition const*>(
-		_usingFor.libraryName().annotation().referencedDeclaration
-	);
+	if (_usingFor.usesBraces())
+	{
+		for (ASTPointer<IdentifierPath> const& function: _usingFor.functionsOrLibrary())
+			if (auto functionDefinition = dynamic_cast<FunctionDefinition const*>(function->annotation().referencedDeclaration))
+			{
+				if (!functionDefinition->isFree() && !(
+					dynamic_cast<ContractDefinition const*>(functionDefinition->scope()) &&
+					dynamic_cast<ContractDefinition const*>(functionDefinition->scope())->isLibrary()
+				))
+					m_errorReporter.typeError(
+						4167_error,
+						function->location(),
+						"Only file-level functions and library functions can be bound to a type in a \"using\" statement"
+					);
+			}
+			else
+				m_errorReporter.fatalTypeError(8187_error, function->location(), "Expected function name." );
+	}
+	else
+	{
+		ContractDefinition const* library = dynamic_cast<ContractDefinition const*>(
+			_usingFor.functionsOrLibrary().front()->annotation().referencedDeclaration
+		);
+		if (!library || !library->isLibrary())
+			m_errorReporter.fatalTypeError(
+				4357_error,
+				_usingFor.functionsOrLibrary().front()->location(),
+				"Library name expected. If you want to attach a function, use '{...}'."
+			);
+	}
 
-	if (!library || !library->isLibrary())
-		m_errorReporter.fatalTypeError(4357_error, _usingFor.libraryName().location(), "Library name expected.");
+	// We do not visit _usingFor.functions() because it will lead to an error since
+	// library names cannot be mentioned stand-alone.
 
 	if (_usingFor.typeName())
 		_usingFor.typeName()->accept(*this);
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -35,6 +35,7 @@
 #include <libsolutil/Algorithms.h>
 #include <libsolutil/StringUtils.h>
 #include <libsolutil/Views.h>
+#include <libsolutil/Visitor.h>
 
 #include <boost/algorithm/string/join.hpp>
 #include <boost/algorithm/string/predicate.hpp>
@@ -3626,12 +3627,67 @@ void TypeChecker::endVisit(Literal const& _literal)
 
 void TypeChecker::endVisit(UsingForDirective const& _usingFor)
 {
-	if (m_currentContract->isInterface())
-		m_errorReporter.typeError(
-			9088_error,
-			_usingFor.location(),
-			"The \"using for\" directive is not allowed inside interfaces."
+	if (!_usingFor.usesBraces())
+	{
+		solAssert(_usingFor.functionsOrLibrary().size() == 1);
+		ContractDefinition const* library = dynamic_cast<ContractDefinition const*>(
+				_usingFor.functionsOrLibrary().front()->annotation().referencedDeclaration
 		);
+		solAssert(library && library->isLibrary());
+		// No type checking for libraries
+		return;
+	}
+
+	if (!_usingFor.typeName())
+	{
+		solAssert(m_errorReporter.hasErrors());
+		return;
+	}
+
+	solAssert(_usingFor.typeName()->annotation().type);
+	Type const* normalizedType = TypeProvider::withLocationIfReference(
+		DataLocation::Storage,
+		_usingFor.typeName()->annotation().type
+	);
+	solAssert(normalizedType);
+
+	for (ASTPointer<IdentifierPath> const& path: _usingFor.functionsOrLibrary())
+	{
+		solAssert(path->annotation().referencedDeclaration);
+		FunctionDefinition const& functionDefinition =
+			dynamic_cast<FunctionDefinition const&>(*path->annotation().referencedDeclaration);
+
+		solAssert(functionDefinition.type());
+
+		if (functionDefinition.parameters().empty())
+			m_errorReporter.fatalTypeError(
+				4731_error,
+				path->location(),
+				"The function \"" + joinHumanReadable(path->path(), ".") + "\" " +
+				"does not have any parameters, and therefore cannot be bound to the type \"" +
+				(normalizedType ? normalizedType->toString(true) : "*") + "\"."
+			);
+
+		FunctionType const* functionType = dynamic_cast<FunctionType const&>(*functionDefinition.type()).asBoundFunction();
+		solAssert(functionType && functionType->selfType(), "");
+		BoolResult result = normalizedType->isImplicitlyConvertibleTo(
+			*TypeProvider::withLocationIfReference(DataLocation::Storage, functionType->selfType())
+		);
+		if (!result)
+			m_errorReporter.typeError(
+				3100_error,
+				path->location(),
+				"The function \"" + joinHumanReadable(path->path(), ".") + "\" "+
+				"cannot be bound to the type \"" + _usingFor.typeName()->annotation().type->toString() +
+				"\" because the type cannot be implicitly converted to the first argument" +
+				" of the function (\"" + functionType->selfType()->toString() + "\")" +
+				(
+					result.message().empty() ?
+					"." :
+					": " +  result.message()
+				)
+			);
+	}
 }
 
 void TypeChecker::checkErrorAndEventParameters(CallableDeclaration const& _callable)
--- a/libsolidity/ast/AST.h
+++ b/libsolidity/ast/AST.h
@@ -33,6 +33,7 @@
 #include <libevmasm/Instruction.h>
 #include <libsolutil/FixedHash.h>
 #include <libsolutil/LazyInit.h>
+#include <libsolutil/Visitor.h>
 
 #include <json/json.h>
 
@@ -630,9 +631,16 @@ private:
 };
 
 /**
- * `using LibraryName for uint` will attach all functions from the library LibraryName
- * to `uint` if the first parameter matches the type. `using LibraryName for *` attaches
- * the function to any matching type.
+ * Using for directive:
+ *
+ * 1. `using LibraryName for T` attaches all functions from the library `LibraryName` to the type `T`
+ * 2. `using LibraryName for *` attaches to all types.
+ * 3. `using {f1, f2, ..., fn} for T` attaches the functions `f1`, `f2`, ...,
+ *     `fn`, respectively to `T`.
+ *
+ * For version 3, T has to be implicitly convertible to the first parameter type of
+ * all functions, and this is checked at the point of the using statement. For versions 1 and
+ * 2, this check is only done when a function is called.
  */
 class UsingForDirective: public ASTNode
 {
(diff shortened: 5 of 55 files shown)


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3430,6 +3431,57 @@ void TypeChecker::endVisit(Literal const& _literal)
 
 void TypeChecker::endVisit(UsingForDirective const& _usingFor)
 {
+	auto checkFunctionDefinition = [&](FunctionDefinition const& _functionDefinition, SourceLocation _location)
```
```suggestion
	auto checkFunctionCompatibility = [&](FunctionDefinition const& _functionDefinition, SourceLocation _location)
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3430,6 +3431,57 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+				"The function \"" + _functionDefinition.name() + "\" "+
+				"cannot be bound to the type \"" + _usingFor.typeName()->annotation().type->toString() +
+				"\" because the type cannot be implicitly converted to the first argument" +
+				" of the function (\"" + functionType->selfType()->toString() + "\")."
+			);
+	};
+
+	UsingForDirective::LHS const& lhs = _usingFor.lhs();
+	std::visit(util::GenericVisitor{
+		[&](UsingForDirective::LibraryOrFunctionOrModule const& _libraryOrFunctionOrModule) {
```
chriseth: Maybe we could have a helper function in `UsingForDirective` that allows us to handle `using f for X` in the same way as `using {f} for X` if `f` is a function?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3430,6 +3431,57 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+			Declaration const* declaration = _libraryOrFunctionOrModule.name->annotation().referencedDeclaration;
+			if (auto functionDefinition = dynamic_cast<FunctionDefinition const*>(declaration))
+				checkFunctionDefinition(*functionDefinition, _libraryOrFunctionOrModule.name->location());
+		},
```
chriseth: If it's a library or module we don't check (which is fine I think, just wanted to confirm)?

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -453,12 +454,54 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
 
-	if (!library || !library->isLibrary())
-		m_errorReporter.fatalTypeError(4357_error, _usingFor.libraryName().location(), "Library name expected.");
+	auto checkFunctionDefinition = [&](FunctionDefinition const* _functionDefinition, SourceLocation _location)
+	{
+		if (!_functionDefinition->isFree())
+			m_errorReporter.typeError(
+				4167_error,
+				_location,
+				"Only free functions can be bound to a type in a \"using\" statement"
```
chriseth:
```suggestion
				"Functions need to be free in order to bind them explicitly to a type in a \"using\" statement"
```

chriseth: But mabe it's not really better :)

libsolidity/ast/Types.cpp
```diff
@@ -363,21 +388,68 @@ MemberList::MemberMap Type::boundFunctions(Type const& _type, ASTNode const& _sc
...
+					FunctionTypePointer fun = dynamic_cast<FunctionType const&>(*function.type()).asBoundFunction();
+					solAssert(fun && _type.isImplicitlyConvertibleTo(*fun->selfType()) , "");
+
+					addMember(&function, fun);
+				}
+			},
+			[&](UsingForDirective::Asterisk const&)
+			{
+				// TODO change this when using statement is allowed at file level.
+				ContractDefinition const& contract = dynamic_cast<ContractDefinition const&>(_scope);
```
chriseth: Shouldn't `*` only capture elements from the current scope? I would actually disallow except when used at file level.

chriseth:
Looks good in general. I think there is some code duplication and we should also check if everything works as expected when the name of a function is changed in an `import` statement.

libsolidity/ast/Types.cpp
hrkrshnn:
```solidity
library L {
    function f(S memory s) internal {}
}
struct S { uint x; }
contract C {
    using L for *;
    function g(S memory s) external {
        s.f();
    }
}
```
The following works. So I think capturing from the current source unit should be okay.

test/libsolidity/semanticTests/using/module_2.sol
```diff
@@ -0,0 +1,33 @@
...
+        return x.id();
+    }
+    function g(uint x) public pure returns (uint) {
+        return x.one_ext();
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// library: L
```
hrkrshnn: @cameel This fails currently, I think it is expecting `A.L`. However, isoltest seems to not support separating with dots.

test/libsolidity/syntaxTests/using/module_4.sol
```diff
@@ -0,0 +1,18 @@
...
+    return x;
+}
+
+==== Source: B ====
+import "A";
+
+contract C {
+    // This only attaches the free functions in the current source unit.
+    // Therefore, `id` is not accessible.
+    using * for uint;
```
hrkrshnn: @chriseth Is this how we want the semantics?

test/libsolidity/semanticTests/using/module_2.sol
cameel: ~True, but I think it was just never implemented (rather than broken by #12073). The thing that the PR changed is that you can now put the source unit name in front of the library name.~

~Should not be hard to implement though. Changes are likely needed in the same places as in #12073. The main thing was the [change in `TestFileParser.cpp`](https://github.com/soroosh-sdi/solidity/commit/816d8021e4f57b1b32eb3ce1fbfa421a585a0f0f#diff-ca8456220457a41b507d3968dbffa4ed8b777a91653fec0eedd4985fd9bf5606). The rest is mostly adjusting tests.~

test/libsolidity/syntaxTests/using/module_4.sol
chriseth: I would suggest that these things are only valid at file level.

chriseth: Especially inside contracts it is horribly difficult due to inheritance.

test/libsolidity/semanticTests/using/module_2.sol
cameel: Oh, wait. Even if you do `import "A" as M` and refer to library `L` as `M.L`, its fully qualified name does not become `M.L`. It's always qualified with the source unit name. So on the CLI you would use `A:L`. In soltest you need ~`"A".L`~ `"A":L` because source unit names can contain arbitrary characters and it was too hard to get the parser to allow that without something like quoting or a special prefix.

So no need to handle dots at all.

hrkrshnn: > So no need to handle dots at all.

What do you mean? Changing to "A".L is also unsupported by isoltest currently.

cameel: Sorry, it's a typo. I meant `"A":L`.

test/libsolidity/syntaxTests/using/module_4.sol
hrkrshnn: Disallowed this for contracts.

hrkrshnn:
There is an annoying problem here. One difference between libraries and free functions in using for statement is that, unlike `using LibraryName for T`, `using free_function for T` can be immediately type checked.

But this has issues for
```solidity
function sum(uint[] calldata arr) returns (uint partialSum) {
    for(uint i = 0; i < arr.length; i++) {
        partialSum += arr[i];
    }
}
contract C {
    using sum for uint[];
    function f(uint[] memory arr) external returns (uint) {
        arr.sum();
    }
}
```
The underlying type `uint[]` has to be treated as a storage type (the default location?). But at instantiation time, we need a conversion from `memory -> calldata`, which is not possible. There are two ways to proceed here:

If we really want `using X for T;` to be type checked at declaration time, then we should allow `using f for uint[] calldata;`
The other option is to individually type check during the calling, which doesn't look nice.

What do you think @chriseth

chriseth:
What about doing type checking right away for non-reference types and delaying it if there is at least one storage location that works?

hrkrshnn:
> What about doing type checking right away for non-reference types and delaying it if there is at least one storage location that works?

I could do that.

> it if there is at least one storage location that works?

This might be tricky, but I can try.

libsolidity/ast/AST.h
```diff
@@ -629,33 +631,71 @@ class InheritanceSpecifier: public ASTNode
...
+	};
+	/// `using {f1, f2, ..., fn} for T;`
+	struct FunctionList
+	{
+		std::vector<ASTPointer<IdentifierPath>> functions;
+	};
+	/// `using * for T;`
+	struct Asterisk {};
+
+	using LHS = std::variant<LibraryOrFunctionOrModule, FunctionList, Asterisk>;
```
chriseth: I think we should try to find a better name for this.

test/libsolidity/semanticTests/using/module_1.sol
```diff
@@ -0,0 +1,18 @@
+==== Source: A ====
+function id(uint x) pure returns (uint) {
+    return x;
+}
+
+==== Source: B ====
+import "A" as M;
+contract C {
+    using M for uint;
+	function f(uint x) public returns (uint) {
```
bshastry:
```suggestion
    function f(uint x) public returns (uint) {
```

docs/grammar/SolidityParser.g4
```diff
@@ -314,7 +314,7 @@ errorDefinition:
  * Using directive to bind library functions to types.
  * Can occur within contracts and libraries.
  */
-usingDirective: Using identifierPath For (Mul | typeName) Semicolon;
+usingDirective: Using (identifierPath | (LBrace identifierPath (Comma identifierPath)* RBrace) | Mul) For (Mul | typeName) Semicolon;
```
bshastry: `usingDirective` is missing from sourceUnit definitions i.e., this PR also enables `using * for T` at source unit level, right?

test/libsolidity/syntaxTests/using/module_2.sol
```diff
@@ -0,0 +1,14 @@
...
+function id(uint x) pure returns (uint) {
+    return x;
+}
+
+==== Source: B ====
+import {id as Id} from "A";
+
+contract C {
+    using Id for uint;
+	function f(uint x) public pure returns (uint) {
```
bshastry, line 11:
```suggestion
    function f(uint x) public pure returns (uint) {
```

test/libsolidity/syntaxTests/using/module_3.sol
```diff
@@ -0,0 +1,14 @@
...
+function id(uint x) pure returns (uint) {
+    return x;
+}
+
+==== Source: B ====
+import "A" as M;
+
+contract C {
+    using M.id for uint;
+	function f(uint x) public pure returns (uint) {
```
bshastry:
```suggestion
    function f(uint x) public pure returns (uint) {
```

test/libsolidity/semanticTests/using/module_1.sol
```diff
@@ -0,0 +1,18 @@
+==== Source: A ====
+function id(uint x) pure returns (uint) {
+    return x;
+}
+
+==== Source: B ====
+import "A" as M;
+contract C {
+    using M for uint;
```
bshastry: If the import statement above looks like `import "A.sol";` (`A.sol` being the file instead of simply `A`) and then we do `using A.sol for uint`, does the parser parse it as intended?

In other words, if no module name is defined, does it default the module name default to file name?

docs/grammar/SolidityParser.g4
chriseth: Right, maybe the is an argument for not having this logic in the parser :)

chriseth:
The current implementation performs type checking at the `using for` place when the overload is unique and not a module. Type checking is delayed to the call location if there are overloads.

chriseth:
The above is wrong: `using {f, g} for uint;` requires `f` and `g` to be unique.

test/libsolidity/semanticTests/using/module_1.sol
chriseth: No, it does not work like that at all. `import "A.sol";` imports all top-level elements from `"A.sol"` into the current scope, while `import "A.sol" as M;` creates a new top-level scope called `M` that you can reference.

chriseth: In the second syntax you need to access items as `M.f` while in the first you just use `f`.

chriseth:
Currently, `using {f} for x` fails if `f` is overloaded. Do we want to change that?

Also @cameel thinks that `using X ...` and `using {X} ...` should be fully equivalent, and I kind of agree.

chriseth:
Instead of `export using ...` we could use `using {f, h} for MyType sticky;`

chriseth:
Having thought about this, I believe a simpler specification / implementation might make sense for now:
- If `{...}` is used, it can only contain a list of functions (but they can be library functions)
- I would drop `using ModuleName for ...`
- `using LibraryName` is not allowed at file level (we can change that later)
- If `using { ... } for MyType` is used at file level and `MyType` (struct, enum or type) is defined in the same file at top level, the functions are available everywhere the type is available - if we can come up with a good keyword, this can still be changed

In summary, the change to how it was before this PR:
 - add `using {...} for ExplicitType` (no `*`)
 - allow `using` at file-level, but only in the `{...}`-form and make it auto-attach

We can even think about disallowing `using` at file-level on types that are not defined there.

chriseth:
We have to check what happens when a virtual function is referenced.

chriseth:
virtual functions are of course not a problem because they cannot be referenced: The functions have to be free or library functions and this is already checked.

chriseth:
Please review - the first commit is a more complicated version.

libsolidity/ast/AST.h
```diff
@@ -629,33 +630,47 @@ class InheritanceSpecifier: public ASTNode
...
+ *     `fn`, respectively to `T`.
+ *
+ * For version 3, T has to be implicitly convertible to the first parameter type of
+ * all functions, and this is checked at the point of the using statement. For versions 1 and
+ * 2, this check is only done when a function is called.
  */
 class UsingForDirective: public ASTNode
 {
 public:
+	struct Asterisk {};
```
ekpyron: Is this used?

ekpyron:
Do we have a test case for using a function multiple times in the list? ``using {f, f} for``?

test/libsolidity/syntaxTests/using/free_reference_type.sol
```diff
@@ -0,0 +1,12 @@
...
+}
+function g(uint[] storage x) pure returns (uint) {
+    return x[0];
+}
+function h(uint[] calldata x) pure returns (uint) {
+    return x[0];
+}
+using {f, g, h} for uint[];
+// ----
+// TypeError 2527: (131-135): Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
```
ekpyron: Is it the intention of this test to have this error?

test/libsolidity/semanticTests/using/imported_functions.sol
```diff
@@ -0,0 +1,19 @@
+==== Source: A ====
+function inc(uint x) pure returns (uint) {
+    return x + 1;
+}
+
+==== Source: B ====
+import {inc as f} from "A";
+import "A";
+using * for uint;
```
ekpyron: This is neither in the grammar nor in the documentation, is it?
Even more so, is this really something we expect to be used :-)?

docs/grammar/SolidityParser.g4
```diff
@@ -311,10 +311,10 @@ errorDefinition:
 	Semicolon;
 
 /**
- * Using directive to bind library functions to types.
- * Can occur within contracts and libraries.
+ * Using directive to bind library functions and free functions to types.
+ * Can occur within contracts and libraries and at the file level.
```
ekpyron, line 316: Doesn't it need to be added to ``sourceUnit`` then?

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -451,12 +452,39 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
+					dynamic_cast<ContractDefinition const*>(functionDefinition->scope())->isLibrary()
+				))
+					m_errorReporter.typeError(
+						4167_error,
+						function->location(),
+						"Only free functions and library functions can be bound to a type in a \"using\" statement"
+					);
+			}
+			else
+				m_errorReporter.fatalTypeError(8187_error, function->location(), "Expected function name." );
```
ekpyron, line 471: CI complains that this is untested.

test/libsolidity/syntaxTests/using/using_contract_err.sol
```diff
@@ -0,0 +1,13 @@
...
+}
+interface I {
+    function f(uint) external;
+}
+
+contract Test {
+    using C for uint;
+}
+// ----
+// TypeError 4357: (127-128): Library name expected. If you want to attach a function, use '{...}'.
```
ekpyron, line 13: ~~If we want to be perfectionistic, we can not emit the second part inside a contract... but also fine.~~ Nevermind

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3586,12 +3587,59 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+	}
+
+	for (ASTPointer<IdentifierPath> const& path: _usingFor.functionsOrLibrary())
+	{
+		solAssert(path->annotation().referencedDeclaration);
+		// No type checking if it is a library.
+		FunctionDefinition const* functionDefinition =
+			dynamic_cast<FunctionDefinition const*>(path->annotation().referencedDeclaration);
+		if (!functionDefinition)
+			continue;
```
ekpyron: I'd consider to instead do something like
```
if (!_usingFor.usesBraces())
{
  // assert that _usingFor.functionsOrLibrary() is a single library
  return;
}
```
Before the loop and then assert that ``functionDefinition`` is not ``nullptr`` here.

But also fine as is.

libsolidity/ast/Types.cpp
```diff
@@ -362,20 +381,28 @@ MemberList::MemberMap Type::boundFunctions(Type const& _type, ASTNode const& _sc
 			)
 		)
 			continue;
-		auto const& library = dynamic_cast<ContractDefinition const&>(
-			*ufd->libraryName().annotation().referencedDeclaration
-		);
-		for (FunctionDefinition const* function: library.definedFunctions())
+
```
ekpyron, line 385: Here I also wonder if checking ``ufd->usesBraces()`` and asserting accordingly would be more fool-proof.
But yeah - on the other hand "usesBraces" is a syntactic property and it makes sense not to consider it anymore later on - resp. again: also fine as is.

ekpyron (commented):
Looks good! But there seems to be some issue with the docs CI run. It could also use a rebase to fix unrelated CI issues I think.

libsolidity/analysis/TypeChecker.cpp
chriseth: Restructured it a bit.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3586,12 +3587,68 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+	Type const* normalizedType = TypeProvider::withLocationIfReference(
+		DataLocation::Storage,
+		_usingFor.typeName()->annotation().type
+	);
+	solAssert(normalizedType);
+
+	for (ASTPointer<IdentifierPath> const& path: _usingFor.functionsOrLibrary())
+	{
+		solAssert(path->annotation().referencedDeclaration);
+		// No type checking if it is a library.
```
ekpyron: The comment should be removed now - but apart from that the new version looks much nicer.

docs/contracts/using-for.rst
```diff
@@ -6,71 +6,89 @@
...
+
+If you specify a library, *all* functions in the library are attached,
 even those where the type of the first parameter does not
 match the type of the object. The type is checked at the
 point the function is called and function overload
 resolution is performed.
 
+If you use a list of functions (``using {f, g, h, L.t} for uint;``),
+then the type (``uint``) has to be implicitly convertible to the
+first parameter of each of these functions. This mean the check is
```
ekpyron:
```suggestion
first parameter of each of these functions. This check is
```
or otherwise ``means``

docs/contracts/using-for.rst
```diff
@@ -6,71 +6,89 @@
...
-contract, including within all of its functions, and has no effect
-outside of the contract in which it is used. The directive
-may only be used inside a contract, not inside any of its functions.
+scope (either the contract or the current module/source unit),
+including within all of its functions, and has no effect
+outside of the contract or module in which it is used.
 
 Let us rewrite the set example from the
-:ref:`libraries` in this way:
+:ref:`libraries` in this way, using file-level functions
```
ekpyron:
```suggestion
:ref:`libraries` section in this way, using file-level functions
```
Or otherwise the ``the`` before is weird, isn't it?

docs/contracts/using-for.rst
```diff
@@ -6,71 +6,89 @@
...
-    pragma solidity >=0.6.0 <0.9.0;
-
+    pragma solidity ^0.8.12;
 
-    // This is the same code as before, just without comments
     struct Data { mapping(uint => bool) flags; }
+    // Now we attach functions to the type.
+    // This is valid in the rest of the module.
+    // If you import the module, you have to
+    // repeat the statement there, potentially as
```
ekpyron:
```suggestion
    // repeat the using statement there, for example as
```

docs/contracts/using-for.rst
```diff
@@ -6,71 +6,89 @@
...
 
     // SPDX-License-Identifier: GPL-3.0
-    pragma solidity >=0.6.0 <0.9.0;
-
+    pragma solidity ^0.8.12;
 
-    // This is the same code as before, just without comments
     struct Data { mapping(uint => bool) flags; }
+    // Now we attach functions to the type.
+    // This is valid in the rest of the module.
```
ekpyron:
```suggestion
    // The attached functions can be used throughout the rest of the module.
```

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -310,10 +311,16 @@ bool ASTJsonConverter::visit(InheritanceSpecifier const& _node)
 
 bool ASTJsonConverter::visit(UsingForDirective const& _node)
 {
-	setJsonNode(_node, "UsingForDirective", {
-		make_pair("libraryName", toJson(_node.libraryName())),
+	std::vector<pair<string, Json::Value>> attributes = {
```
ekpyron:
```suggestion
	vector<pair<string, Json::Value>> attributes = {
```

libsolidity/ast/Types.cpp
```diff
@@ -330,30 +331,48 @@ Type const* Type::fullEncodingType(bool _inLibraryCall, bool _encoderV2, bool) c
...
 	// Normalise data location of type.
 	DataLocation typeLocation = DataLocation::Storage;
 	if (auto refType = dynamic_cast<ReferenceType const*>(&_type))
 		typeLocation = refType->location();
 
-	set<Declaration const*> seenFunctions;
+	set<pair<string, Declaration const*>> seenFunctions;
 	MemberList::MemberMap members;
 
+	auto addFunction = [&](FunctionDefinition const& _function, optional<string> _name = {})
```
ekpyron, line 353:
```suggestion
	auto addFunction = [&, seenFunctions = set<pair<string, Declaration const*>>{}](FunctionDefinition const& _function, optional<string> _name = {})
```
I'd actually consider this - I was wondering where it's actually used for a second, since it's defined outside.
But also fine keeping it as is.

ekpyron (commented):
Only more optional docs and style comments. Apart from maybe the outdated comment in https://github.com/ethereum/solidity/pull/12121/files#r801814258 I'd be fine with keeping as is.

ekpyron:
Oh, and it probably wouldn't hurt to have a test for importing a file that uses a global ``using``, checking that the attached functions are no longer available after import.

libsolidity/ast/Types.cpp
chriseth: I tried it, but it didn't work. I moved the variable closer to the function, though.

ekpyron: Ah, the lambda would probably have to be declared ``mutable`` to get a non-const capture. But yeah, doesn't matter.

libsolidity/parsing/Parser.cpp
```diff
@@ -962,7 +965,26 @@ ASTPointer<UsingForDirective> Parser::parseUsingDirective()
 	ASTNodeFactory nodeFactory(*this);
 
 	expectToken(Token::Using);
-	ASTPointer<IdentifierPath> library(parseIdentifierPath());
+
+	vector<ASTPointer<IdentifierPath>> functions;
+	bool const usesBraces = m_scanner->currentToken() == Token::LBrace;
+	if (usesBraces)
+	{
+		vector<ASTPointer<IdentifierPath>> functionList;
```
ekpyron: Not sure this helps much compared to just adding to ``functions`` directly...

ekpyron (dismissed):
Looks good to me now!

libsolidity/parsing/Parser.cpp
chriseth: Oh, I think I even changed that, but it seems the change got lost.

libsolidity/parsing/Parser.cpp
```diff
@@ -962,7 +965,26 @@ ASTPointer<UsingForDirective> Parser::parseUsingDirective()
...
+		advance();
```
chriseth: Wait - can't this just be moved inside the loop and the other one deleted?

ekpyron (dismissed):

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -310,10 +311,16 @@ bool ASTJsonConverter::visit(InheritanceSpecifier const& _node)
...
 bool ASTJsonConverter::visit(UsingForDirective const& _node)
 {
-	setJsonNode(_node, "UsingForDirective", {
-		make_pair("libraryName", toJson(_node.libraryName())),
+	vector<pair<string, Json::Value>> attributes = {
 		make_pair("typeName", _node.typeName() ? toJson(*_node.typeName()) : Json::nullValue)
-	});
+	};
+	if (_node.usesBraces())
+		attributes.emplace_back("functionList", toJson(_node.functionsOrLibrary()));
```
chriseth: To be compatible with future extensions, this should be an object.

chriseth:
When we have `using {..., f, ...} for T`, we should enforce that the first parameter of `f` is equal to T up to data location, because it is less confusing and we also will get fewer problems with operators.

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -311,10 +312,25 @@ bool ASTJsonConverter::visit(InheritanceSpecifier const& _node)
...
 		make_pair("typeName", _node.typeName() ? toJson(*_node.typeName()) : Json::nullValue)
-	});
+	};
+	if (_node.usesBraces())
+	{
+		Json::Value functionList;
+		for (auto const& function: _node.functionsOrLibrary())
+		{
+			Json::Value functionNode;
+			functionNode["function"] = toJson(*function);
```
chriseth, line 324: The `function` here is needed because of `using {add as +} for MyType;`.

docs/contracts/using-for.rst
```diff
@@ -82,12 +100,13 @@ Let us rewrite the set example from the
         }
     }
 
-It is also possible to extend elementary types in that way:
+It is also possible to extend elementary types in that way.
```
ekpyron: Strictly speaking, the array below is not an "elementary type", though, is it?
Not sure we need to change anything, though.

ekpyron (commented):
Maybe a test for
```
function f(uint256 x, uint256 y) pure {}
function f(uint256 x, uint256[] y) pure {}
using { f } for uint256;
```
?
We intentionally don't allow that, right?
And I guess mainly because it'd be annoying to implement? My guess would be that this not being possible might be rather annoying in the long run... this makes it impossible to add overloaded functions to a type without libraries, right?

chriseth:
I'm growing less and less fond of function overloading. Rust made a good decision to ban it altogether.

ekpyron:
> I'm growing less and less fond of function overloading. Rust made a good decision to ban it altogether.

Fair enough - I'd still expect this to be annoying as long as there generally is function overloading - but yeah, at least a test case would be nice in any case.

ekpyron (approved):

PR #12123, opened 2021-10-12
CodeGen: Only output switch when it has cases

### LINKED ISSUE #12068: Warning about switch with only a default case in compiler-generated IR of an empty contract
## Description
Generating IR for a trivial empty contract and then assembling that IR results in a warning: `"switch" statement with only a default case.` Compiler-generated IR should not trigger any warnings.

```bash
echo 'contract C {}' |
    solc - --ir |
    sed '/IR:/d' |
    solc - --strict-assembly
```
```
Warning: "switch" statement with only a default case.
  --> <stdin>:49:17:
   |
49 |                 switch selector
   |                 ^ (Relevant source part starts here and spans across multiple lines).
```

The part of the code that triggers this warning looks like this:
```yul
            if iszero(lt(calldatasize(), 4))
            {
                let selector := shift_right_224_unsigned(calldataload(0))
                switch selector

                default {}
            }
```

## Environment

- Compiler version: 0.8.10 (`develop`).

Marenz:
fixes #12068

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@ Bugfixes:
  * Commandline Interface: Report output selection options unsupported by the selected input mode instead of ignoring them.
  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
  * TypeChecker: Fix internal error when using user defined value types in public library functions.
+ * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
 
 
 
--- a/libsolidity/codegen/ir/IRGenerator.cpp
+++ b/libsolidity/codegen/ir/IRGenerator.cpp
@@ -944,7 +944,7 @@ string IRGenerator::callValueCheck()
 string IRGenerator::dispatchRoutine(ContractDefinition const& _contract)
 {
 	Whiskers t(R"X(
-		if iszero(lt(calldatasize(), 4))
+		<?+cases>if iszero(lt(calldatasize(), 4))
 		{
 			let selector := <shr224>(calldataload(0))
 			switch selector
@@ -962,8 +962,8 @@ string IRGenerator::dispatchRoutine(ContractDefinition const& _contract)
 			}
 			</cases>
 			default {}
-		}
-		if iszero(calldatasize()) { <receiveEther> }
+		}</+cases>
+		<?+receiveEther>if iszero(calldatasize()) { <receiveEther> }</+receiveEther>
 		<fallback>
 	)X");
 	t("shr224", m_utils.shiftRightFunction(224));
--- a/libsolutil/Whiskers.cpp
+++ b/libsolutil/Whiskers.cpp
@@ -189,11 +189,13 @@ string Whiskers::replace(
 			if (conditionName[0] == '+')
 			{
 				string tag = conditionName.substr(1);
-				assertThrow(
-					_parameters.count(tag),
-					WhiskersError, "Tag " + tag + " used as condition but was not set."
-				);
-				conditionValue = !_parameters.at(tag).empty();
+
+				if (_parameters.count(tag))
+					conditionValue = !_parameters.at(tag).empty();
+				else if (_listParameters.count(tag))
+					conditionValue = !_listParameters.at(tag).empty();
+				else
+					assertThrow(false, WhiskersError, "Tag " + tag + " used as condition but was not set.");
 			}
 			else
 			{
--- a/libsolutil/Whiskers.h
+++ b/libsolutil/Whiskers.h
@@ -60,12 +60,13 @@ DEV_SIMPLE_EXCEPTION(WhiskersError);
  *  - Condition parameter: <?name>...<!name>...</name>, where "<!name>" is optional
  *    replaced (and recursively expanded) by the first part if the condition is true
  *    and by the second (or empty string if missing) if the condition is false
- *  - Conditional string parameter: <?+name>...<!+name>...</+name>
- *    Works similar to a conditional parameter where the checked condition is
- *    that the regular (string) parameter called "name" is non-empty.
  *  - List parameter: <#list>...</list>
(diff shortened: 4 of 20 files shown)


libsolutil/Whiskers.h
```diff
@@ -60,9 +60,9 @@ DEV_SIMPLE_EXCEPTION(WhiskersError);
  *  - Condition parameter: <?name>...<!name>...</name>, where "<!name>" is optional
  *    replaced (and recursively expanded) by the first part if the condition is true
  *    and by the second (or empty string if missing) if the condition is false
- *  - Conditional string parameter: <?+name>...<!+name>...</+name>
```
chriseth, line 63: Maybe explain more explicitly that it works for strings and for list parameters? (and put it at the end)

libsolutil/Whiskers.cpp
```diff
@@ -189,11 +189,18 @@ string Whiskers::replace(
 			if (conditionName[0] == '+')
 			{
 				string tag = conditionName.substr(1);
-				assertThrow(
-					_parameters.count(tag),
-					WhiskersError, "Tag " + tag + " used as condition but was not set."
-				);
-				conditionValue = !_parameters.at(tag).empty();
+
```
chriseth, line 192: What about this?
```
if (_parameters.count(tag))
  conditionValue = !_parameters.at(tag).empty();
else if (_listParameters.count(tag))
  conditionValue = !_listParameters.at(tag).empty();
else
  assertThrow(...);
```

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -947,6 +947,7 @@ string IRGenerator::dispatchRoutine(ContractDefinition const& _contract)
 		if iszero(lt(calldatasize(), 4))
 		{
 			let selector := <shr224>(calldataload(0))
```
chriseth, line 949: You can also pull this inside the condition

chriseth: Or the whole first if statement...

Marenz: Yeah, I just had the same thought..

Marenz: The whole `if` with condition & body actually.

libsolutil/Whiskers.cpp
Marenz: I tried to avoid the repetition.. if you feel this is easier to read, okay by me

Marenz:
I also added that it won't output an `if` check if the body is empty anyway (in the selector part)

chriseth:
Oh can you add a changelog entry? "IR Codegen: " or something. We wanted to start adding changelog entries for ir codegen starting from the next version.

test/cmdlineTests/yul_optimizer_steps/output
```diff
@@ -31,16 +31,8 @@ object "C_7" {
             {
                 /// @src 0:80:112  "contract C..."
                 mstore(64, 128)
-                if iszero(lt(calldatasize(), 4))
-                {
-                    let selector := shift_right_unsigned(calldataload(0))
-                    pop(selector)
-                }
-                pop(iszero(calldatasize()))
```
Marenz, line 39: I am unsure why this `pop` existed and whether removing it is correct. @chriseth @cameel  could you validate this?

chriseth:
oh sorry, there already is one! But please make it clear that it's about the ir codegen.

test/cmdlineTests/yul_optimizer_steps_nested_brackets/output
```diff
@@ -138,7 +138,6 @@ object "C_6" {
                         return(memPos, sub(memEnd, memPos))
                     }
                 }
-                pop(iszero(calldatasize()))
```
Marenz, line 141: Same for this one

test/cmdlineTests/yul_optimizer_steps/output
cameel: Removing it looks fine. It seems to be an "optimized" version of `if iszero(calldatasize()) {  }`, which you can see if you compile this file without optimization.

This is a test for the optimizer sequence and it purposefully does not run all the steps so it not being optimized out was fine. I suspect that this `if` might be getting converted to a `switch` by the default sequence and then its only case might be getting removed due to not having any instructions. If that's the case then it would make total sense for it to disappear after your PR.

test/cmdlineTests/yul_optimizer_steps_nested_brackets/output
cameel: It's also a test for the optimizer sequence so probably the same reason as in https://github.com/ethereum/solidity/pull/12123#discussion_r727253382.

test/cmdlineTests/yul_optimizer_steps/output
Marenz: I actually suspected it was original a `switch` which was converted to an `if` because it only had one case (appart from the default).

Marenz: And well, it looks very much like our selector code, just as `if`, not `switch` (like we originally generated in the code gen)

Changelog.md
```diff
@@ -15,6 +15,7 @@ Bugfixes:
  * Commandline Interface: Report output selection options unsupported by the selected input mode instead of ignoring them.
  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
  * TypeChecker: Fix internal error when using user defined value types in public library functions.
+ * Yul IR Generator: Don't output empty switches/if-bodies for empty contracts.
```
```suggestion
 * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
```

chriseth (approved):

PR #12125, opened 2021-10-12
Don't return zero exit code when writing linked files to disk fails

cameel:
- Depends on #12124

A tiny bug I noticed while working on #11800. We're returning 0 even though the operation failed.


cameel:
Related question: why does `CommandLineInterface` not fail immediately when `createFile()` fails? Instead it has this weird `m_error` variable and checks it when it has finishes printing all the output and only then fails. Can we change it so that it fails right away?

chriseth (approved):

chriseth:
Feel free to change the error mechanism!

cameel:
Great. I added a commit changing that to #11800 then.

PR #12129, opened 2021-10-13
Enhance documentation in regards to modifiers and return

### LINKED ISSUE #12063: [Sol->Yul] Modifier changes to state variables gets reflected in function body in IR generated code but not in legacy
The following test passes via legacy but not via IR. Actually not sure what is defined behavior here

```
contract C0 {
  bool s0;
  constructor() { s0 = true; }
  modifier m0()
  {
    _;
    s0 = false;
    _;
  }
  function f0() external m0 payable returns(uint)
  {
    if (s0)
    {
      return (uint(1337));
    }
  }
}
// ====
// compileViaYul: also
// ----
// f0() -> 1337
```

Marenz:
fixes #12063

### DIFF
--- a/docs/ir/ir-breaking-changes.rst
+++ b/docs/ir/ir-breaking-changes.rst
@@ -1,3 +1,6 @@
+
+.. index: ir breaking changes
+
 *********************************
 Solidity IR-based Codegen Changes
 *********************************
@@ -35,13 +38,15 @@ hiding new and different behavior in existing code.
 
   We have the same behavior for implicit delete, for example when array of structs is shortened.
 
-- Function modifiers are implemented in a slightly different way regarding function parameters.
+- Function modifiers are implemented in a slightly different way regarding function parameters and return variables.
   This especially has an effect if the placeholder ``_;`` is evaluated multiple times in a modifier.
-  In the old code generator, each function parameter has a fixed slot on the stack. If the function
-  is run multiple times because ``_;`` is used multiple times or used in a loop, then a change to the
-  function parameter's value is visible in the next execution of the function.
+  In the old code generator, each function parameter and return variable has a fixed slot on the stack.
+  If the function is run multiple times because ``_;`` is used multiple times or used in a loop, then a
+  change to the function parameter's or return variable's value is visible in the next execution of the function.
   The new code generator implements modifiers using actual functions and passes function parameters on.
-  This means that multiple executions of a function will get the same values for the parameters.
+  This means that multiple evaluations of a function's body will get the same values for the parameters,
+  and the effect on return variables is that they are reset to their default (zero) value for each
+  execution.
 
   .. code-block:: solidity
 
@@ -56,6 +61,35 @@ hiding new and different behavior in existing code.
 
   If you execute ``f(0)`` in the old code generator, it will return ``2``, while
   it will return ``1`` when using the new code generator.
+
+  .. code-block:: solidity
+
+      // SPDX-License-Identifier: GPL-3.0
+      pragma solidity >=0.7.1 <0.9.0;
+
+      contract C {
+          bool active = true;
+          modifier mod()
+          {
+              _;
+              active = false;
+              _;
+          }
+          function foo() external mod() returns (uint ret)
+          {
+              if (active)
+                  ret = 1; // Same as ``return 1``
+          }
+      }
+
+  The function ``C.foo()`` returns the following values:
+
+  - Old code generator: ``1`` as the return variable is initialized to ``0`` only once before the first ``_;``
+    evaluation and then overwritten by the ``return 1;``. It is not initialized again for the second ``_;``
+    evaluation and ``foo()`` does not explicitly assign it either (due to ``active == false``), thus it keeps
+    its first value.
+  - New code generator: ``0`` as all parameters, including return parameters, will be re-initialized before
+    each ``_;`` evaluation.
 
 - The order of contract initialization has changed in case of inheritance.
 


docs/contracts/function-modifiers.rst
```diff
@@ -114,10 +114,36 @@ Their values can only be passed to them explicitly at the point of invocation.
...
+        modifier mod()
+        {
+            _;
+            active = false;
+            _;
+        }
+        function foo() external mod payable returns(uint)
+        {
+            if (active)
+                return 1337;
```
christianparpart: Maybe it's just me. But whenever I see numbers like 1337 or 42 it doesn't feel serious. Maybe just subjective.

docs/contracts/function-modifiers.rst
```diff
@@ -114,10 +114,36 @@ Their values can only be passed to them explicitly at the point of invocation.
 Explicit returns from a modifier or function body only leave the current
 modifier or function body. Return variables are assigned and
 control flow continues after the ``_`` in the preceding modifier.
+There is however an exception to this:
 
 .. warning::
-    In an earlier version of Solidity, ``return`` statements in functions
-    having modifiers behaved differently.
+    In the old EVM backend of Solidity, the code example below behaved differently.
```
chriseth: Please document this in "differences between the ir and legacy codegen".

Marenz: Additionally or only?

chriseth: As long as the ir backend is experimental, we should not explain the differences in the main documentation.

Marenz: Ah, I guess I'll just set a link here

docs/contracts/function-modifiers.rst
Marenz: I just copied that from the issue, I am innocent :P. I'll change it to 1 for you

Marenz:
I just pushed a small update, but I still need to get the test working, so still WIP.

docs/contracts/function-modifiers.rst
```diff
@@ -116,8 +116,8 @@ modifier or function body. Return variables are assigned and
 control flow continues after the ``_`` in the preceding modifier.
 
 .. warning::
-    In an earlier version of Solidity, ``return`` statements in functions
-    having modifiers behaved differently.
+    In the old code generator, ``return`` statements in functions
```
chriseth: I would just keep this unchanged.

docs/ir/ir-breaking-changes.rst
```diff
@@ -273,3 +273,39 @@ The function ``f(1)`` returns the following values:
 Note that, unlike the new code generator, the old code generator does not perform a cleanup after the bit-not assignment (``_a = ~_a``).
 This results in different values being assigned (within the inline assembly block) to return value ``_r1`` between the old and new code generators.
 However, both code generators perform a cleanup before the new value of ``_a`` is assigned to ``_r2``.
+
+.. _modifier_codegen_differences:
+
+Modifiers
```
chriseth: There is a section on modifiers already in line 38

chriseth: It does not talk about return parameters, I think we should add that they are reset to their zero value for each `_`.

docs/contracts/function-modifiers.rst
Marenz: I removed it again, but I find this warning very useless actually. It doesn't say what earlier version nor what differences nor where to look them up.
If earlier means the old code gen, that isn't really "old" as long as it is still default..

Marenz:
Ready now.

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,35 @@ hiding new and different behavior in existing code.
...
+          function foo() external mod payable returns(uint ret)
+          {
+              if (active)
+                  return 1;
+          }
+      }
+
+  The function ``C.foo()`` returns the following values:
+
+  - Old code generator: ``1`` due to the return variable never getting assigned and thus it never overwrites the first return value in ``mod()``
```
bshastry: Could you clarify what is meant by return variable being assigned and the distinction between return value and return variable?

As far as I understood, there are two placeholder statements in `mod` both of which invoke the body of `foo()`. The first invocation returns `1` but the return value (`1`) is not assigned. The second invocation in legacy (old) codegen returns `1` because the return value on the stack is not reset (?). However, IR resets it before the second invocation which is why it returns `0`.

bshastry (commented):
I'm still a bit confused by the docs. Added review comment.

docs/ir/ir-breaking-changes.rst
chriseth: The `return` statement in solidity is equivalent to an assignment to the return variables (even if they are not named) plus exiting the function. This means `_` does not return anything, it executes the function body, which assigns the return variables (which are not accessible directly in the modifier). In the new codegen, the second placeholder resets the values of the return variables while in the old codegen, they are only reset at the very beginning, not with each `_`.

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,35 @@ hiding new and different behavior in existing code.
   If you execute ``f(0)`` in the old code generator, it will return ``2``, while
   it will return ``1`` when using the new code generator.
 
+- In some cases modifiers also behave differently in regards to the return
```
chriseth: Can you try to combine parameters and return parameters? It's exactly the same effect.

docs/ir/ir-breaking-changes.rst
Marenz: I don't think that's exactly correct. They are also reset with the second `_`, but only if we assign something to the return variable (e.g. explicitly `return 0;` at the end of the function)

Marenz: Nevermind, you basically said the same, I just didn't understand it correctly at first

Marenz:
I moved & rephrased it to combined it with the function parameters. Is it better understandable now @bshastry ?

docs/ir/ir-breaking-changes.rst
```diff
@@ -37,9 +40,9 @@ hiding new and different behavior in existing code.
 
 - Function modifiers are implemented in a slightly different way regarding function parameters.
```
chriseth:
```suggestion
- Function modifiers are implemented in a slightly different way regarding function parameters and return variables.
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -37,9 +40,9 @@ hiding new and different behavior in existing code.
...
   This especially has an effect if the placeholder ``_;`` is evaluated multiple times in a modifier.
-  In the old code generator, each function parameter has a fixed slot on the stack. If the function
-  is run multiple times because ``_;`` is used multiple times or used in a loop, then a change to the
-  function parameter's value is visible in the next execution of the function.
+  In the old code generator, each function parameter and return parameter has a fixed slot on the stack.
+  If the function is run multiple times because ``_;`` is used multiple times or used in a loop, then a
+  change to the function (return-)parameter's value is visible in the next execution of the function.
```
chriseth:
```suggestion
  In the old code generator, each function parameter and return variable has a fixed slot on the stack.
  If the function is run multiple times because ``_;`` is used multiple times or used in a loop, then a
  change to the function parameter's or return variable's value is visible in the next execution of the function.
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -37,9 +40,9 @@ hiding new and different behavior in existing code.
...
   The new code generator implements modifiers using actual functions and passes function parameters on.
   This means that multiple executions of a function will get the same values for the parameters.
```
chriseth:
```suggestion
  This means that multiple executions of a function's body will get the same values for the parameters,
  and the effect on return variables is that they are reset to their default (zero) value for each
  execution.
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,32 @@ hiding new and different behavior in existing code.
...
   it will return ``1`` when using the new code generator.
 
+  .. code-block:: solidity
+
+      // SPDX-License-Identifier: GPL-3.0
+      pragma solidity >=0.7.1 <0.9.0;
+
+      contract C {
+          bool active;
+          constructor() { active = true; }
```
chriseth:
```suggestion
          bool active = true;
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,32 @@ hiding new and different behavior in existing code.
...
+          modifier mod()
+          {
+              _;
+              active = false;
+              _;
+          }
+          function foo() external mod payable returns(uint ret)
```
chriseth:
```suggestion
          function foo() external mod() returns (uint ret)
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,32 @@ hiding new and different behavior in existing code.
...
+          {
+              if (active)
+                  return 1;
+          }
+      }
+
+  The function ``C.foo()`` returns the following values:
+
+  - Old code generator: ``1`` as the return variable is initialized to ``0`` only once before the first call and then overwritten by the ``return 1;``. It's not initialized again for the second call and ``foo()`` doesn't explicitly assign it either (due to ``active == false``), thus it keeps its first value.
```
chriseth:
```suggestion
  - Old code generator: ``1`` as the return variable is initialized to ``0`` only once before the first call and
    then overwritten by the ``return 1;``. It is not initialized again for the second call and ``foo()`` does not
    explicitly assign it either (due to ``active == false``), thus it keeps its first value.
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,32 @@ hiding new and different behavior in existing code.
...
+          modifier mod()
+          {
+              _;
+              active = false;
+              _;
+          }
+          function foo() external mod payable returns(uint ret)
+          {
+              if (active)
+                  return 1;
```
chriseth: Actually I wonder if this example is easier to understand if we just replace this line by `ret = 1`.
Can you check if it is still valid in that case? We can add a comment that `return 1` would be have in the same way.

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +60,32 @@ hiding new and different behavior in existing code.
...
+          }
+      }
+
+  The function ``C.foo()`` returns the following values:
+
+  - Old code generator: ``1`` as the return variable is initialized to ``0`` only once before the first call and then overwritten by the ``return 1;``. It's not initialized again for the second call and ``foo()`` doesn't explicitly assign it either (due to ``active == false``), thus it keeps its first value.
+  - New code generator: ``0`` as all parameters, including return parameters, will be initialized again before each call.
```
chriseth: Here, but also in the rest of the changes: I would avoid using the term `call` but instead use `function body execution` or `evaluation of the function body` or `placeholder run` or something like that. Otherwise it could be confused with the outermost call, which is the only real function call.

docs/ir/ir-breaking-changes.rst
Marenz: I don't think a `,` before an `and` is correct?

chriseth: It is, it just changes the meaning. We can also use a `;` if you prefer that or just use `. The effect...`

Marenz: I just read up. You're right. I have learned something :)

docs/ir/ir-breaking-changes.rst
Marenz: Yep, still valid. Changed it.

docs/ir/ir-breaking-changes.rst
```diff
@@ -35,13 +38,15 @@ hiding new and different behavior in existing code.
...
 
-- Function modifiers are implemented in a slightly different way regarding function parameters.
+- Function modifiers are implemented in a slightly different way regarding function parameters and return variables.
   This especially has an effect if the placeholder ``_;`` is evaluated multiple times in a modifier.
-  In the old code generator, each function parameter has a fixed slot on the stack. If the function
-  is run multiple times because ``_;`` is used multiple times or used in a loop, then a change to the
-  function parameter's value is visible in the next execution of the function.
+  In the old code generator, each function parameter and return variable has a fixed slot on the stack.
+  If the function is run multiple times because ``_;`` is used multiple times or used in a loop, then a
+  change to the function parameter's or return variable's value is visible in the next execution of the function.
```
bshastry, line 45:
```suggestion
  change to the function parameter(s) or return variable(s) value is visible in the next execution of the function.
```

bshastry: Also `s/execution/invocation/g` may read better.

docs/ir/ir-breaking-changes.rst
```diff
@@ -35,13 +38,15 @@ hiding new and different behavior in existing code.
...
   The new code generator implements modifiers using actual functions and passes function parameters on.
```
bshastry, line 46:
```suggestion
  The new code generator implements modifiers using actual functions and passes function (return) parameters on.
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -35,13 +38,15 @@ hiding new and different behavior in existing code.
...
-  This means that multiple executions of a function will get the same values for the parameters.
+  This means that multiple executions of a function's body will get the same values for the parameters,
+  and the effect on return variables is that they are reset to their default (zero) value for each
```
bshastry, line 48: What about non-value type return parameters? E.g., bytes memory, struct memory etc.

docs/ir/ir-breaking-changes.rst
bshastry: May be evaluation is even better.

docs/ir/ir-breaking-changes.rst
```diff
@@ -35,13 +38,15 @@ hiding new and different behavior in existing code.
...
   This especially has an effect if the placeholder ``_;`` is evaluated multiple times in a modifier.
-  In the old code generator, each function parameter has a fixed slot on the stack. If the function
-  is run multiple times because ``_;`` is used multiple times or used in a loop, then a change to the
-  function parameter's value is visible in the next execution of the function.
+  In the old code generator, each function parameter and return variable has a fixed slot on the stack.
+  If the function is run multiple times because ``_;`` is used multiple times or used in a loop, then a
+  change to the function parameter's or return variable's value is visible in the next execution of the function.
   The new code generator implements modifiers using actual functions and passes function parameters on.
-  This means that multiple executions of a function will get the same values for the parameters.
+  This means that multiple executions of a function's body will get the same values for the parameters,
```
bshastry:
```suggestion
  This means that multiple evaluations of a function's body will get the same values for the parameters,
```

docs/ir/ir-breaking-changes.rst
```diff
@@ -57,6 +62,35 @@ hiding new and different behavior in existing code.
...
+          }
+      }
+
+  The function ``C.foo()`` returns the following values:
+
+  - Old code generator: ``1`` as the return variable is initialized to ``0`` only once before the first ``_;``
+    evaluation and then overwritten by the ``return 1;``. It is not initialized again for the second ``_;``
+    evaluation and ``foo()`` does not explicitly assign it either (due to ``active == false``), thus it keeps
+    its first value.
+  - New code generator: ``0`` as all parameters, including return parameters, will be initialized again before
```
bshastry:
```suggestion
  - New code generator: ``0`` as all parameters, including return parameters, will be re-initialized before
```

bshastry (commented):
There's definitely more clarity to the present text. Added a few minor comments.

docs/ir/ir-breaking-changes.rst
chriseth: Please let's not use `(s)`, everyone will understand that this is equally valid for exactly one parameter or variable.

bshastry: Please ignore the comment about (s). I misread the sentence.

Marenz:
I get the feeling in docu changes everyone can give feedback, compared to code :laughing:

docs/ir/ir-breaking-changes.rst
Marenz: I don't think return parameters are "passed" on, rather they are just properly initialized before each call..

docs/ir/ir-breaking-changes.rst
Marenz: I just tested and for a `bytes memory` it behaves the same way..

bshastry (approved):
LGTM

PR #12131, opened 2021-10-13
Update the workaround for restore_cache on macOS not to delete files completely

cameel:
#12106 fixed `b_osx` job in cases where there's something cached to restore but now it fails when we don't have the cache. That's because Homebrew needs the files we delete to be able to update itself.

I don't think there's a way to check if there's something to restore so this PR tries to fix it in a different way - it replaces the directory with a symlink manually so that it's still there but does not make tar refuse to overwrite it.

Closes #12134.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,16 +160,34 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
-  - steps_restore_cache_homebrew_workaround: &steps_restore_cache_homebrew_workaround
+  - steps_homebrew_workaround_before_restore_cache: &steps_homebrew_workaround_before_restore_cache
       steps:
         - run:
-            # FIXME: For some reason restore_cache fails saying that it cannot remove the scm/ dir.
-            # The directory contains only two files: `git` (wrapper script over git) and `svn` (symlink
-            # to `git`). This looks like a bug in restore_cache. Removing scm/ is a workaround.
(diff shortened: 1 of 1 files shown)


chriseth:
Why does it fail the build? :)

cameel:
Looks like downloading jsoncpp failed for some reason. Probably a random issue with the connection. I'll restart the job and see if it passes then.

```
CMake Error at jsoncpp-project-stamp/download-jsoncpp-project.cmake:170 (message):
  Each download failed!

    error: downloading 'https://github.com/open-source-parsers/jsoncpp/archive/1.9.3.tar.gz' failed
          status_code: 7
          status_string: "Couldn't connect to server"
          log:
```

chriseth (dismissed):

cameel:
ok, so `b_osx` passed but this was a run where there was some cached data to be restored. I need to also test the case where the cache gets rebuilt because this is where the problem was.

cameel:
Bad news:
```
error: 'Library/Homebrew/shims/scm/git' is beyond a symbolic link
fatal: Unable to process path Library/Homebrew/shims/scm/git
Error: Could not 'git stash' in /usr/local/Homebrew
```

Homebrew seems to do doing something very weird there. Looks like we'll need a wilder workaround.

cameel:
I'm trying a different workaround now: #12134.

cameel:
I tried a different thing - moving the dir and moving it back after restore - and it seems to work. It's faster than #12134 so let's go with this one.

chriseth (approved):

PR #12132, opened 2021-10-13
Use side-effects of user defined functions in evm code transform. 

chriseth:
I tried but failed. Assigning to @ekpyron 
~Depends on https://github.com/ethereum/solidity/pull/12759 and https://github.com/ethereum/solidity/pull/12795~


libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -229,7 +234,7 @@ void ControlFlowGraphBuilder::operator()(ExpressionStatement const& _exprStmt)
 	//       not only for builtins.
 	if (auto const* funCall = get_if<FunctionCall>(&_exprStmt.expression))
 		if (BuiltinFunction const* builtin = m_dialect.builtin(funCall->functionName.name))
-			if (builtin->controlFlowSideEffects.terminates)
+			if (builtin->controlFlowSideEffects.terminatesOrReverts())
```
ekpyron:
```suggestion
			if (!builtin->controlFlowSideEffects.canContinue)
```

ekpyron:
The remaining failing test is due to the new code transform allowing arbitrary inputs so far, but it's also perfectly fine to require disambiguated, hoisted, etc. code, so the failure is fine.

I just quickly ran the semantics tests on a merge of this with https://github.com/ethereum/solidity/pull/11974 and they all pass. And it does have effects: on affected tests I either saw a slight decrease by a few gas or a significant decrease by up to ~1% of gas.
Funnily, I also saw code size increase occasionally :-). Which pretty much goes along with my suspicion that good parts of the code size increase in the new code transform is due to needlessly cleaning up the stack before a ``revert(0,0)`` or the like (i.e. an array of senseless ``pop``s before). So once we make the new code transform clever about that, the code size concern might just vanish.

There is a paranoid assertion in ``OptimizedCodeTransform::operator()(CFG::BasicBlock const& _block)`` (L500+) that checks on blocks with ``CFG::BasicBlock::Terminated`` exits that the last operation was actually a terminating builtin... if we want to stay that paranoid, we'd need to pass in the side effects there, too, but it should also be fine to just remove that assertion.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -248,9 +241,11 @@ void ControlFlowGraphBuilder::operator()(Block const& _block)
 
 void ControlFlowGraphBuilder::operator()(If const& _if)
 {
+	StackSlot condition = std::visit(*this, *_if.condition);
 	auto& ifBranch = m_graph.makeBlock(debugDataOf(_if.body));
+	// TODO What is the correct debug data to use here?
```
chriseth: @ekpyron can you answer this question?

ekpyron: You mean for the ``afterIf`` block? That's returning to the scope containing the if, so I guess whatever ``debugDataOf(*m_currentBlock)`` *before* the ``if``, i.e. at the beginning of this function? But since that shouldn't change during the visit to the condition, it should be fine as is.

But yeah ``afterIf`` is conceptually just the continuation of the block of the containing scope and the debug data of that scope should be stored in the current node when entering the ``if``.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -490,14 +490,7 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
 			createStackLayout(debugDataOf(_functionReturn), exitStack);
 			m_assembly.appendJump(0, AbstractAssembly::JumpType::OutOfFunction);
 		},
-		[&](CFG::BasicBlock::Terminated const&)
-		{
-			// Assert that the last builtin call was in fact terminating.
```
ekpyron, line 497: Yeah - if we wanted to, we could of course drag along the entire logic into the code transform as well and keep validating things here - but I guess it's fine to trust that it's ok.

ekpyron (dismissed):
This just needs simple test updates, but is good to go otherwise, isn't it? So I'm approving modulo the tests.

No gas cost updates on this, though? Makes me wonder if this really just doesn't happen or if the gas expectations are just not properly updated and checked...

libyul/backends/evm/ControlFlowGraphBuilder.cpp
chriseth: Isn't the main change in this PR that `m_currentBlock` can actually change during the visit of the if condition? I'll just grab the debug data before the visit and then reuse it.

chriseth: I ended up just moving the `makeBlock` calls above the visit to the condition.

chriseth:
Ok this is weird - now there are gas changes...

test/libsolidity/semanticTests/array/function_array_cross_calls.sol
```diff
@@ -45,6 +45,6 @@ contract C {
 // compileViaYul: also
 // ----
 // test() -> 5, 6, 7
-// gas irOptimized: 290947
+// gas irOptimized: 294125
```
chriseth: This got more costly.

test/libsolidity/semanticTests/externalContracts/deposit_contract.sol
```diff
@@ -178,35 +178,35 @@ contract DepositContract is IDepositContract, ERC165 {
 // compileViaYul: also
 // ----
 // constructor()
-// gas irOptimized: 1558013
+// gas irOptimized: 1594911
 // gas legacy: 2580394
-// gas legacyOptimized: 1775403
+// gas legacyOptimized: 1776483
```
chriseth: I think this was wrong to begin with, I think we don't really check it, do we?

test/libsolidity/semanticTests/externalContracts/deposit_contract.sol
```diff
@@ -178,35 +178,35 @@ contract DepositContract is IDepositContract, ERC165 {
 // compileViaYul: also
 // ----
 // constructor()
-// gas irOptimized: 1558013
```
chriseth: This got longer - I will compare.

chriseth:
Ok, now I understand what @ekpyron was talking about in the last days.
```
let a, b, c, d, e, f, g = x();
if a { t() }
...
```
In the above, we have a rather deep stack at the point where we call `t`. If the analysis shows that `t` is always terminating, the block is marked as "terminating". The problem is now that the stack is cleaned up at the end of each block marked as "terminating". This results in an least `7` `pop` opcodes being introduced at the end of the if branch. Before this change, the if branch would just join again with the main branch and thus would have no cleanup at all.

So I actually think that the "allow garbage at the end of terminating branches" feature is really important to implement. The deposit contract contains the following code (highlights are from a diff):

image

chriseth:
Another optimization here could be that we should not push a return tag for a call to a terminating function.

test/libyul/evmCodeTransform/after_terminating_function.yul
```diff
@@ -0,0 +1,23 @@
...
+//     /* "":17:18   */
+//   0x02
+//     /* "":14:15   */
+//   0x01
+//     /* "":12:19   */
+//   jump(tag_1)
+//     /* "":21:59   */
+// tag_1:
+//   pop
+//   pop
```
ekpyron: Also need to allow filling in junk on function entries to get rid of these.

chriseth: The optimizer should remove them, though.

ekpyron:
This has more significant gains compared to https://github.com/ethereum/solidity/pull/12759

<details>
<pre>
bleeps
   ir-optimize-evm+yul
     bytecode_size :  -3.09 %
     method_gas :  -0.01 %
elementfi
   ir-optimize-evm+yul
     bytecode_size :  -5.65 %
     deployment_gas :  -5.17 %
     method_gas :  -2.02 %
ens
   ir-optimize-evm+yul
     bytecode_size :  -4.66 %
     deployment_gas :  -4.38 %
     method_gas :  -0.05 %
gnosis
   ir-optimize-evm+yul
     bytecode_size :  -3.36 %
     deployment_gas :  -3.34 %
     method_gas :  -0.02 %
pool-together
   ir-optimize-evm+yul
     bytecode_size :  -4.16 %
     deployment_gas :  -4.08 %
     method_gas :  -0.02 %
prb-math
   ir-optimize-evm+yul
     bytecode_size :  -1.61 %
     deployment_gas :  -1.58 %
trident
   ir-optimize-evm+yul
     bytecode_size :  -4.69 %
     deployment_gas :  -4.40 %
     method_gas :  -4.43 %
yield_liquidator
   ir-optimize-evm+yul
     bytecode_size :  -3.06 %
     deployment_gas :  -3.07 %
     method_gas :  -0.01 %
zeppelin
   ir-optimize-evm+yul
     bytecode_size :  -3.74 %
     deployment_gas :  -3.21 %
     method_gas :  -0.03 %
</pre>
</details>

In total relative to ``develop``:
<details>
<pre>
bleeps
   ir-optimize-evm+yul
     bytecode_size :  -4.02 %
     method_gas :  -0.01 %
elementfi
   ir-optimize-evm+yul
     bytecode_size :  -7.02 %
     deployment_gas :  -6.25 %
     method_gas :  -2.54 %
ens
   ir-optimize-evm+yul
     bytecode_size :  -6.44 %
     deployment_gas :  -5.93 %
     method_gas :  -0.03 %
gnosis
   ir-optimize-evm+yul
     bytecode_size :  -5.17 %
     deployment_gas :  -4.89 %
     method_gas :  -0.02 %
pool-together
   ir-optimize-evm+yul
     bytecode_size :  -5.79 %
     deployment_gas :  -5.49 %
     method_gas :  -0.02 %
prb-math
   ir-optimize-evm+yul
     bytecode_size :  -2.22 %
     deployment_gas :  -2.16 %
trident
   ir-optimize-evm+yul
     bytecode_size :  -6.01 %
     deployment_gas :  -5.55 %
     method_gas :  -5.00 %
yield_liquidator
   ir-optimize-evm+yul
     bytecode_size :  -4.17 %
     deployment_gas :  -3.97 %
     method_gas :  -0.01 %
zeppelin
   ir-optimize-evm+yul
     bytecode_size :  -5.59 %
     deployment_gas :  -4.80 %
     method_gas :  -0.04 %
</pre>
</details>

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -767,10 +767,61 @@ void StackLayoutGenerator::fillInJunk(CFG::BasicBlock const& _block)
 		createStackLayout(_source, _target, swap, dupOrPush, pop);
 		return opGas;
 	};
+
+	if (_functionInfo && !_functionInfo->canContinue && _block.allowsJunk())
+	{
+		// TODO: unify with the one below
```
ekpyron: If anyone wants to finish this while I'm off: this is pretty much the only thing left to do here.

leonardoalt: so I guess not so much and could be done soon? :D

leonardoalt:
Dependency was merged, so not so much left to do here

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: I added at least some validation back, even though it relies on ``functionCall->canContinue`` to be correct. But conceptually that's fine - the ``canContinue`` flag is part of the CFG and the code transform only guarantees correctness relative to the CFG...

ekpyron:
This is the output of the benchmark differ (https://github.com/ethereum/solidity/pull/12804) for the IR-OPTIMIZE-EVM+YUL run (the rest remains unaffected):

### `ir-optimize-evm+yul`
|           bleeps | **`-3.07% ✅`** |           `0%` | **`-0.01% ✅`** |
|        elementfi | **`-5.61% ✅`** |                |                |
|              ens | **`-4.62% ✅`** | **`-4.36% ✅`** | **`-0.08% ✅`** |
|           gnosis | **`-3.32% ✅`** | **`-3.23% ✅`** | **`-0.02% ✅`** |
|    pool-together | **`-4.14% ✅`** | **`-4.07% ✅`** | **`-0.02% ✅`** |
|         prb-math | **`-1.61% ✅`** | **`-1.58% ✅`** |           `0%` |
|          trident | **`-4.74% ✅`** | **`-4.49% ✅`** | **`-4.45% ✅`** |
| yield_liquidator | **`-3.04% ✅`** | **`-3.03% ✅`** | **`-0.01% ✅`** |
|         zeppelin | **`-3.65% ✅`** | **`-3.13% ✅`** | **`+0.02% ❌`** |

I would have expected more of a runtime difference - but it may be that this largely kicks in on reverting paths and those don't occur as much in the method gas benchmarks, but not entirely sure... and the code size difference alone is worthwhile anyways.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -69,35 +69,39 @@ void OptimizedEVMCodeTransform::operator()(CFG::FunctionCall const& _call)
...
 
 	// Emit code.
 	{
 		m_assembly.setSourceLocation(originLocationOf(_call));
 		m_assembly.appendJumpTo(
 			getFunctionLabel(_call.function),
-			static_cast<int>(_call.function.get().returns.size() - _call.function.get().arguments.size()) - 1,
-			AbstractAssembly::JumpType::IntoFunction
+			static_cast<int>(_call.function.get().returns.size() - _call.function.get().arguments.size()) - (_call.canContinue ? 1 : 0),
+			_call.canContinue ? AbstractAssembly::JumpType::IntoFunction : AbstractAssembly::JumpType::Ordinary
```
chriseth: Isn't it still an "into function", even if the function does not return?

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -69,35 +69,39 @@ void OptimizedEVMCodeTransform::operator()(CFG::FunctionCall const& _call)
...
 		);
-		m_assembly.appendLabel(m_returnLabels.at(&_call.functionCall.get()));
+		if (_call.canContinue)
+			m_assembly.appendLabel(m_returnLabels.at(&_call.functionCall.get()));
 	}
 
 	// Update stack.
 	{
 		// Remove arguments and return label from m_stack.
-		for (size_t i = 0; i < _call.function.get().arguments.size() + 1; ++i)
```
chriseth, line 102: Do we need to do anything here if the call cannot return?

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
chriseth: Ah, this is undone in the next commit.

Changelog.md
```diff
@@ -6,6 +6,7 @@ Language Features:
 Compiler Features:
  * Peephole Optimizer: Remove operations without side effects before simple terminations.
  * Assembly-Json: Export: Include source list in `sourceList` field.
+ * Yul EVM Code Transform: Generate more optimal code for user-defined functions that always terminate a transaction.
```
chriseth: I think it's important for tooling to explain here that we don't push a return label if the function always returns or reverts.

test/externalTests/zeppelin.sh
```diff
@@ -66,6 +66,10 @@ function zeppelin_test
     sed -i "s|it(\('reverts \)|it.skip(\1|g" math/SignedSafeMath.test.js
     sed -i "s|it(\('reverts \)|it.skip(\1|g" structs/EnumerableSet.behavior.js
     popd
+    # TODO: check
```
chriseth: todo

ekpyron: True - although it's a rather safe bet that it's just revert reason heuristics failing.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: In principle, we don't - but not doing anything will trigger some stack height consistency assertion later on. I could try to change those later assertions, but maybe easier to just always create stack-equilibrium here?

test/externalTests/zeppelin.sh
ekpyron: Just had it run again (https://app.circleci.com/pipelines/github/ethereum/solidity/23661/workflows/4202dc2e-60a5-4b55-af8d-c3dce6ac82e8/jobs/1038942) - the error is:
```
      -Transaction reverted and Hardhat couldn't infer the reason. Please report this to help us improve Hardhat.
      +TransparentUpgradeableProxy: admin cannot fallback to proxy target
```

So the test case indeed still reverts, but hardhat doesn't pick up the revert reason anymore - so I think we can safely say this is just testing heuristics failing. I'm actually rather surprised that it's only this one test that has this issue, I'd have expected more...

I'll disable the test again now in any case.

cameel:
Rebased on `develop`, fixed conflicts.

chriseth:
Why did you close this?

cameel:
I mistakenly deleted the wrong branch and github closed it. Reopened.

ekpyron:
Rebased again - my guess is that this has less of an effect after https://github.com/ethereum/solidity/pull/12731 now, but I'd say we should still do it.

cameel:
If you want to see the benchmark diff for this PR you should rebase on `develop` again. I just merged #13097 to fix the `c_ext_benchmark` job. Without the fix there's no baseline benchmark available on `develop` and diffing will fail.

ekpyron:
Still seems to have some positive effects, even though mainly on bytecode size apparently:

|------------------|---------------|----------------|------------|
|           bleeps |         -1.6% |             0% |     -0.01% |
|            brink |        -1.22% |                |            |
|           colony |            0% |                |            |
|        elementfi |        -3.87% |                |            |
|              ens |            0% |             0% |         0% |
|            euler |         -2.7% |         -2.64% |     -0.01% |
|           gnosis |        -2.55% |         -2.13% |     -0.02% |
|  perpetual-pools |        -1.55% |         -1.61% |     -0.05% |
|    pool-together |        -1.45% |         -1.55% |     -0.01% |
|         prb-math |            0% |            +0% |         0% |
|          trident |        -0.29% |         -0.56% |        -0% |
|          uniswap |            0% |            -0% |        +0% |
| yield_liquidator |        -0.47% |         -0.52% |        -0% |
|         zeppelin |        -0.07% |         -0.08% |        -0% |

libyul/backends/evm/ControlFlowGraph.h
```diff
@@ -210,10 +212,12 @@ struct CFG
 	{
 		std::shared_ptr<DebugData const> debugData;
 		Scope::Function const& function;
+		FunctionDefinition const& functionDefinition;
 		BasicBlock* entry = nullptr;
 		std::vector<VariableSlot> parameters;
 		std::vector<VariableSlot> returnVariables;
 		std::vector<BasicBlock*> exits;
+		bool canContinue = true;
```
cameel, line 220: `FunctionInfo` seems to share a lot of fields with `FunctionCall`. Can one just refer to the other?

Does `FunctionInfo` represent a function definition?

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -494,6 +490,7 @@ void ControlFlowGraphBuilder::registerFunction(FunctionDefinition const& _functi
 	bool inserted = m_graph.functionInfo.emplace(std::make_pair(&function, CFG::FunctionInfo{
 		_function.debugData,
 		function,
+		_function,
```
cameel: Too much function :P `_function` should have been called `_functionDefinition`.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -534,24 +534,34 @@ Stack const& ControlFlowGraphBuilder::visitFunctionCall(FunctionCall const& _cal
...
+		output = &m_currentBlock->operations.emplace_back(CFG::Operation{
 			// input
 			std::move(inputs),
 			// output
 			ranges::views::iota(0u, function.returns.size()) | ranges::views::transform([&](size_t _i) {
 				return TemporarySlot{_call, _i};
 			}) | ranges::to<Stack>,
 			// operation
-			CFG::FunctionCall{_call.debugData, function, _call}
+			CFG::FunctionCall{_call.debugData, function, _call, false, canContinue}
```
cameel:
```suggestion
			CFG::FunctionCall{_call.debugData, function, _call, /* recursive */ false, canContinue}
```

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -534,24 +534,34 @@ Stack const& ControlFlowGraphBuilder::visitFunctionCall(FunctionCall const& _cal
...
 		}).output;
 	}
+	if (!canContinue)
+	{
+		m_currentBlock->exit = CFG::BasicBlock::Terminated{};
+		m_currentBlock = &m_graph.makeBlock(debugDataOf(*m_currentBlock));
```
cameel, line 562: Why was this actually in `operator()(ExpressionStatement)` rather than in this function from the start? It does not look like it needed any specific info available only on expression statements and this place seems more natural.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -69,35 +69,39 @@ void OptimizedEVMCodeTransform::operator()(CFG::FunctionCall const& _call)
...
-			&m_stack.at(m_stack.size() - _call.functionCall.get().arguments.size() - 1)
-		);
-		yulAssert(returnLabelSlot && &returnLabelSlot->call.get() == &_call.functionCall.get(), "");
+		if (_call.canContinue)
+		{
+			auto const* returnLabelSlot = get_if<FunctionCallReturnLabelSlot>(
+				&m_stack.at(m_stack.size() - _call.functionCall.get().arguments.size() - 1)
+			);
+			yulAssert(returnLabelSlot && &returnLabelSlot->call.get() == &_call.functionCall.get(), "");
+		}
```
cameel, line 88: Can we assert that in the opposite case the slot is not there? Or can it actually happen in some corner case?

test/externalTests/zeppelin.sh
```diff
@@ -69,6 +69,9 @@ function zeppelin_test
     sed -i "s|it(\('reverts \)|it.skip(\1|g" math/SignedSafeMath.test.js
     sed -i "s|it(\('reverts \)|it.skip(\1|g" structs/EnumerableSet.behavior.js
     popd
+    pushd test/proxy/
+    sed -i "s|it(\('proxy admin cannot \)|it.skip(\1|g" transparent/TransparentUpgradeableProxy.behaviour.js
```
cameel: Looks like this one is just a single test so we can put in the full name. The `reverts` above was done like that because it covers multiple tests. I'd err on the side of being specific where possible.
```suggestion
    sed -i "s|it(\('proxy admin cannot call delegated functions')\)|it.skip(\1|g" transparent/TransparentUpgradeableProxy.behaviour.js
```

test/libsolidity/semanticTests/array/pop/byte_array_pop_long_storage_empty_garbage_ref.sol
```diff
@@ -15,7 +15,7 @@ contract c {
 }
 // ----
 // test() ->
-// gas irOptimized: 142639
+// gas irOptimized: 142475
```
cameel: I don't see any reverting paths here. Looking at IR, it's apparently because of the panic functions containing reverts but then why are so few cases affected? I'd expect these to be pretty much everywhere.

test/libsolidity/semanticTests/externalContracts/base64.sol
```diff
@@ -51,10 +51,10 @@ contract test {
 // encode_no_asm(bytes): 0x20, 5, "fooba" -> 0x20, 8, "Zm9vYmE="
 // encode_no_asm(bytes): 0x20, 6, "foobar" -> 0x20, 8, "Zm9vYmFy"
 // encode_inline_asm_large()
-// gas irOptimized: 1385042
+// gas irOptimized: 1368042
 // gas legacy: 1652033
 // gas legacyOptimized: 1201033
 // encode_no_asm_large()
-// gas irOptimized: 3335099
+// gas irOptimized: 3291092
```
cameel: I see that even with this PR and relaxed inlining the non-asm base64 implementation is still far behind OZ's [assembly implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2884).

test/libyul/evmCodeTransform/after_terminating_function.yul
```diff
@@ -0,0 +1,22 @@
+{
+	let b := f(1, 2)
+	function f(a, r) -> t { revert(0, 0) }
+	b := f(3, 4)
+}
+// ====
+// stackOptimization: true
+// ----
+//     /* "":17:18   */
```
cameel, line 9: Not related to the PR but I just noticed that you have these comments in all evm transform tests and they seem pretty useless. `AsmPrinter` now allows disabling them (#11944).

test/libyul/yulControlFlowGraph/function.yul
```diff
@@ -28,10 +28,9 @@
 // Block0 [label="\
 // i: [ RET[i] ] => [ TMP[i, 0] TMP[i, 1] ]\l\
 // Assignment(x, y): [ TMP[i, 0] TMP[i, 1] ] => [ x y ]\l\
-// h: [ RET[h] x ] => [ ]\l\
-// h: [ RET[h] y ] => [ ]\l\
+// h: [ x ] => [ ]\l\
 // "];
-// Block0Exit [label="MainExit"];
+// Block0Exit [label="Terminated"];
```
cameel, line 33: How well tested is Yul control flow graph? Do we have enough cases to be confident that terminating paths are detected correctly in corner cases? There's not much in `yulControlFlowGraph/`.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -767,10 +767,52 @@ void StackLayoutGenerator::fillInJunk(CFG::BasicBlock const& _block)
...
+			size_t bestNumJunk = getBestNumJunk(
+				entryLayout,
+				_block->operations.empty() ? blockInfo.exitLayout : m_layout.operationEntryLayout.at(&_block->operations.front())
+			);
+			if (bestNumJunk > 0)
+			{
+				addJunkRecursive(_block, bestNumJunk);
+				blockInfo.entryLayout = entryLayout;
+			}
+		}
```
cameel: Do I understand correctly that junk used to be added only for conditional jumps and now it can be added for every block? I don't see it reflected in tests. Can you add a case affected by this?

cameel (commented):
I don't see anything wrong in the PR but I'm also not sure if I'd be able to spot something that should be there but is missing.

test/libsolidity/semanticTests/array/pop/byte_array_pop_long_storage_empty_garbage_ref.sol
ekpyron: Funny thing, actually :-).
After relaxing the inlining heuristics, in a lot of cases inlining of revert- and panic-helpers kicks in - after which this PR does not help anymore, since the canonical user-defined functions that always revert that occur most often are gone.

However, in this case here, we have an assembly block that writes to memory and is not marked ``memory-safe``, so we do not perform aggressive inlining, the panic helper stays uninlined and the PR kicks in.

test/libsolidity/semanticTests/externalContracts/base64.sol
ekpyron: The main part is checked arithmetic on the index increments... if I put a large ``unchecked`` block in the body of ``NoAsmBase64.encode``, then the gas costs here reduce by 60% down to ``1911077``... which is still worse than the assembly case above, but at least in the same league...
Might be a nice case for adding a user-defined-value type for encapsulating unchecked arithmetic for those indices...

test/libyul/yulControlFlowGraph/function.yul
ekpyron: It's not exhaustively unit-tested, but generally still very well tested, since every semantics test is run through it...
Whether a user-defined function always reverts is just based on ``ControlFlowSideEffectsCollector``, though, and that has unit tests in ``test/libyul/controlFlowSideEffects/``.

libyul/backends/evm/ControlFlowGraph.h
ekpyron: Yes, ``FunctionInfo`` refers to a function definition - but calling it ``FunctionDefinition`` as well here would probably be even more confusing, wouldn't it? On the other hand, the same could be said for ``FunctionCall``, so maybe it'd indeed have been better to call it ``FunctionDefinition``...

I see two fields in common: ``function`` and ``canContinue`` - not sure there's that much need to unify them :-).
The original reason against just having ``FunctionCall`` just refer to ``FunctionInfo`` is that the original versions of the ``ControlFlowGraphBuilder`` did a straight-forward direct traversal of the Yul AST, in which function calls can refer to functions that are only defined later in the block (so I wouldn't have had the ``FunctionInfo`` available when constructing the ``FunctionCall``). Since then, however, we changed it to actually collect all functions defined in a block up front, so actually now we could indeed change this...

libyul/backends/evm/ControlFlowGraphBuilder.cpp
ekpyron: Fair enough :-). And this won't be overly invasive, so I'll change it right here in the PR.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
ekpyron: Before considering side effects of user-defined functions, the only expressions that could affect control flow were terminating opcodes, none of which have return values and thereby they only occured in ``ExpressionStatement``s - so it was easier to just consider control flow there.
The newly created (unreachable) current block here may actually be "ill-formed", since it basically continues from the middle of a function call - it was only later that we realized that we nonetheless don't need any more special treatment here, since we remove unreachable blocks from the CFG anyways, so we can just continue generation here as if no revert happened without that causing any issues later.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
ekpyron: Generally, the ``OptimizedEVMCodeTransform`` only ensures that each of the operations for which it generates code has everything it needs on stack, while the rest of the stack be arbitrary (giving the ``StackLayoutGenerator`` the freedom to do whatever it wants). Conceptually, the layout transition to the entry layout of the next operation would just remove the return label slot, if it was there (well, in the cases here, since the function will revert, there won't really be a "next operation", but still :-)).

So, in practice, there will never be a return label slot in the opposite case, so we *could* assert this here, but I'd still keep the ``OptimizedEVMCodeTransform`` agnostic to the stack layouts below the current operation it's looking at.

libyul/backends/evm/ControlFlowGraph.h
ekpyron: I'd tend towards not changing this in this same PR, though, in order not to complicate it further.

libyul/backends/evm/StackLayoutGenerator.cpp
ekpyron: This is basically just refactoring - I can't think of a case that would be affected by this change.
It's just a matter of whether to consider adding junk to the current block, when entering it, or to the exit just before exiting, the result will be the same. But doing it when entering will make things easier, once we for example add a switch exit for implementing jump tables.

That being said, this will now be run even when the jump to the current block is unconditional, in which case the entry layout should always be identical to the next required layout - I'll add a shortcut condition, if the layouts are already identical, in that case adding no junk will always be optimal anyways.

libyul/backends/evm/ControlFlowGraph.h
cameel: Yeah, I wouldn't touch it in the same PR.

But might still be worth changing eventually. Now you have two `canContinue` fields and this one has a default so you could easily forget to set it explicitly to the right value.

cameel: And as far as naming goes, maybe something like `FunctionDefinitionMetadata` or `FunctionDefinitionInfo` would have been clearer? But again, probably not something to change in this PR.

test/libsolidity/semanticTests/array/pop/byte_array_pop_long_storage_empty_garbage_ref.sol
cameel: Ah, right! These panic functions get inlined so there's no return label to remove :)

test/libsolidity/semanticTests/externalContracts/base64.sol
cameel: Or we could add unchecked operators :)

test/libyul/yulControlFlowGraph/function.yul
cameel: Semantic tests would not really detect the case where we wrongly think that the function can continue but it can't or vice-versa, right? It just affects this small optimization, which is optional. I'm just worried that it might not be working correctly in all corner cases and we're just not seeing it.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
cameel: ok, makes sense.

cameel (dismissed):
Looks fine to me and I think there's not much that could go horribly wrong here if it does pass all the tests :)

ekpyron:
[rebased again]

cameel (dismissed):
Looks like there's one gas expectation failing after rebase.

leonardoalt:
@ekpyron @cameel how about this PR?

ekpyron:
> @ekpyron @cameel how about this PR?

*Sigh*, I thought this was also already merged ages ago :-).
I'll rebase it and we can hope for another review :-).

Marenz:
I rebased this and updated the gas costs

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -517,14 +515,16 @@ Stack const& ControlFlowGraphBuilder::visitFunctionCall(FunctionCall const& _cal
 	yulAssert(m_scope, "");
 	yulAssert(m_currentBlock, "");
 
+	Stack const* output = nullptr;
+	bool canContinue = true;
```
aarlt, line 519: Correct here? Shadowing `ControlFlowGraphBuilder::canContinue`?

ekpyron: Not sure I follow :-). I don't see ``ControlFlowGraphBuilder::canContinue``...

aarlt: Hmm.. not sure what I've seen here.. I don't see anything strange now...

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -285,10 +289,10 @@ void ControlFlowGraphBuilder::operator()(Assignment const& _assignment)
 		return VariableSlot{lookupVariable(_var.name), _var.debugData};
 	}) | ranges::to<vector<VariableSlot>>;
 
-	yulAssert(m_currentBlock, "");
+	Stack input = visitAssignmentRightHandSide(*_assignment.value, assignedVariables.size());
+	yulAssert(m_currentBlock);
 	m_currentBlock->operations.emplace_back(CFG::Operation{
-		// input
-		visitAssignmentRightHandSide(*_assignment.value, assignedVariables.size()),
+		move(input),
```
stackenbotten, line 295: Coding style error

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -331,7 +324,8 @@ void ControlFlowGraphBuilder::operator()(If const& _if)
 {
 	auto& ifBranch = m_graph.makeBlock(debugDataOf(_if.body));
 	auto& afterIf = m_graph.makeBlock(debugDataOf(*m_currentBlock));
-	makeConditionalJump(debugDataOf(_if), std::visit(*this, *_if.condition), ifBranch, afterIf);
+	StackSlot condition = std::visit(*this, *_if.condition);
+	makeConditionalJump(debugDataOf(_if), move(condition), ifBranch, afterIf);
```
stackenbotten, line 328: Coding style error

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -349,8 +343,9 @@ void ControlFlowGraphBuilder::operator()(Switch const& _switch)
 	// Artificially generate:
 	// let <ghostVariable> := <switchExpression>
 	VariableSlot ghostVarSlot{ghostVar, debugDataOf(*_switch.expression)};
+	StackSlot expression = std::visit(*this, *_switch.expression);
 	m_currentBlock->operations.emplace_back(CFG::Operation{
-		Stack{std::visit(*this, *_switch.expression)},
+		Stack{move(expression)},
```
stackenbotten, line 348: Coding style error

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -430,7 +425,8 @@ void ControlFlowGraphBuilder::operator()(ForLoop const& _loop)
 	else
 	{
 		jump(debugDataOf(_loop.pre), loopCondition);
-		makeConditionalJump(debugDataOf(*_loop.condition), std::visit(*this, *_loop.condition), loopBody, afterLoop);
+		StackSlot condition = std::visit(*this, *_loop.condition);
+		makeConditionalJump(debugDataOf(*_loop.condition), move(condition), loopBody, afterLoop);
```
stackenbotten, line 429: Coding style error

ekpyron:
Unstaled and rebased this once more - this is mainly in dire need of reviews.

If anyone is up for this, ping me and I can walk you trough the PR to help in reviewing.

cameel (dismissed):
Needs another rebase.

chriseth:
Code looks good!

ekpyron (approved):
Approving based on @chriseth's approval above.

PR #12133, opened 2021-10-13
Prepare tests to be meaningful.

chriseth:
Preparation for #11352


chriseth:
This changes some tests so that the redundant store eliminator will not make them redundant.

cameel (approved):

PR #12134, opened 2021-10-13
Change the workaround for `restore_cache` on macOS to always perform Homebrew update

cameel:
Replaces #12131. This an alternative solution - instead of removing the symlink, we just update the Homebrew installed in the image. It is redundant if we already have it cached but at least should work no matter whether we have Homebrew cached or not.

Closes #12131.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,13 +163,19 @@ defaults:
   - steps_restore_cache_homebrew_workaround: &steps_restore_cache_homebrew_workaround
       steps:
         - run:
-            # FIXME: For some reason restore_cache fails saying that it cannot remove the scm/ dir.
-            # The directory contains only two files: `git` (wrapper script over git) and `svn` (symlink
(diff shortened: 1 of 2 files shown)


cameel:
I'm giving up for today. I have no idea why it fails even though the instructions I added are what `osx_install_dependencies.sh` runs normally anyway. Did the Homebrew update also break `osx_install_dependencies.sh` and we never noticed due to caching?

I'll go back to digging into it tomorrow.

cameel:
Oh, looks like it was a fluke. It does pass now.

The problem is though that not caching the homebrew update adds 12 min to every macOS run (so 24 min in total for both build and test). That's a bit much since the macOS run is already very slow. I think it might be worth it to fiddle with #12131 a bit more. Maybe removing the dir and restoring it after the data from cache is loaded would work?

cameel:
I managed to get #12131 to work and it's much faster so we can close this one.

Since both workarounds technically work, I'm keeping both open for now and adding `closes #<issue>` tags instead. This way the one we choose and merge will close the other one automatically.

PR #12139, opened 2021-10-14
Do not use named function labels if function names are not unique.

### LINKED ISSUE #12090: `Duplicate tag position` when compiling Yul code defining multiple functions with the same name
## Description

While updating dapptools to solc0.8.9 we noticed that the following yul program now produces a c++ exception when compiling with 0.8.9 (it was fine with 0.8.7).

```
> solc-0.8.9 --strict-assembly hi.yul
Warning: Yul is still experimental. Please use the output with care.

======= hi.yul (EVM) =======

Pretty printed source:
object "object" {
    code {
        calldatacopy(0, 0, 1024)
        {
            function z() -> x
            { x := y() }
            function y() -> x
            { x := z() }
        }
        {
            function z() -> x
            { x := y() }
            function y() -> x
            { x := z() }
        }
    }
}

Exception while assembling: /solidity/libevmasm/Assembly.cpp(720): Throw in function const solidity::evmasm::LinkerObject& solidity::evmasm::Assembly::assemble() const
Dynamic exception type: boost::wrapexcept<solidity::evmasm::AssemblyException>
std::exception::what: Duplicate tag position.
[solidity::util::tag_comment*] = Duplicate tag position.
```

## Environment

- Compiler version: 0.8.9
- Framework/IDE (e.g. Truffle or Remix): none
- EVM execution environment / backend / blockchain client: none
- Operating system: linux

### LINKED ISSUE #12149: `AssemblyException` in inline assembly when reusing the name of a recursive function
```
contract C {
  modifier m {
    _;
    assembly {
      function af0(i) -> a {
        if af0(0) {}
      }
    }
  }
  function f() external m m {}
}
```

throws

https://github.com/ethereum/solidity/blob/1e630fc5840bb12ad9bca2dcf1f6c534c9ba8d31/libsolidity/interface/CompilerStack.cpp#L1225

To reproduce

```
$ solc --experimental-via-ir --bin test.sol
```

### LINKED ISSUE #12158: ICE when defining multiple Yul functions with the same name but different arguments/returns in inline assembly
## Description
If a Yul function with the same name is defined again with different parameters and/or return parameters in a completely different scope in inline assembly, the compiler throws an ICE.

It happens only with the new code generator and I did not manage to reproduce it with optimizer enabled.

It's probably related to #12090, but this one can be triggered using inline assembly rather than pure Yul. The solution should also probably be different - the compiler can just give these functions different names internally instead of completely disabling named labels (which is the solution in #12139).

## Steps to Reproduce
`contract.sol`:
``` solidity
contract C {
    function f() external {
        assembly {
            {
                function x() -> a {}
            }
            {
                function x() {}
            }
        }
    }
}
```
```bash
solc --bin --experimental-via-ir contract.sol
```
```
Exception during compilation: /solidity/libevmasm/Assembly.cpp(350): Throw in function solidity::evmasm::AssemblyItem solidity::evmasm::Assembly::namedTag(const string&, size_t, size_t, std::optional<long unsigned int>)
Dynamic exception type: boost::wrapexcept<solidity::evmasm::AssemblyException>
std::exception::what:
```

The result is the same even if the functions are defined in separate assembly blocks, even if these blocks are inside different Solidity functions:
```solidity
contract C {
    function f() external {
        assembly {
            function x() -> a {}
        }
    }
    function g() external {
        assembly {
            function x() {}
        }
    }
}
```

Note that, unlike in #12090, just giving functions the same name is not enough to trigger this. For example this compiles fine:
```solidity
contract C {
    function f() external {
        assembly {
            function x() { mstore(0, 0) }
        }
    }
    function g() external {
        assembly {
            function x() { mstore(1, 1) }
        }
    }
}
```
## Environment
- Compiler version: 0.8.9

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12090
Fixes https://github.com/ethereum/solidity/issues/12149
Fixes https://github.com/ethereum/solidity/issues/12158

### DIFF
--- a/libyul/backends/evm/AsmCodeGen.cpp
+++ b/libyul/backends/evm/AsmCodeGen.cpp
@@ -52,7 +52,9 @@ void CodeGenerator::assemble(
 		builtinContext,
 		_optimizeStackAllocation,
 		_identifierAccessCodeGen,
-		_useNamedLabelsForFunctions
+			_useNamedLabelsForFunctions ?
+			CodeTransform::UseNamedLabels::YesAndForceUnique :
+			CodeTransform::UseNamedLabels::Never
 	);
 	transform(_parsedData);
 	if (!transform.stackErrors().empty())
--- a/libyul/backends/evm/EVMObjectCompiler.cpp
+++ b/libyul/backends/evm/EVMObjectCompiler.cpp
@@ -72,7 +72,7 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
 		context,
 		_optimize,
 		{},
-		true /* _useNamedLabelsForFunctions */
+		CodeTransform::UseNamedLabels::ForFirstFunctionOfEachName
 	};
 	transform(*_object.code);
 	if (!transform.stackErrors().empty())
--- a/test/cmdlineTests/inline_assembly_function_name_clash/args
+++ b/test/cmdlineTests/inline_assembly_function_name_clash/args
@@ -1,0 +1,1 @@
+--experimental-via-ir --combined-json function-debug-runtime --pretty-json --json-indent 4
--- a/test/cmdlineTests/inline_assembly_function_name_clash/input.sol
+++ b/test/cmdlineTests/inline_assembly_function_name_clash/input.sol
@@ -1,0 +1,18 @@
+contract C {
+  uint x;
+  modifier m() {
+    uint t;
+    assembly {
+      function f() -> x { x := 8 }
+      t := f()
+    }
+    x = t;
+    _;
+  }
+  function f() m m public returns (uint r) {
+    assembly { function f() -> x { x := 1 } r := f() }
+  }
+  function g() m m public returns (uint r) {
+    assembly { function f() -> x { x := 2 } r := f() }
+  }
+}
--- a/test/cmdlineTests/yul_function_name_clashes/args
+++ b/test/cmdlineTests/yul_function_name_clashes/args
@@ -1,0 +1,1 @@
+--strict-assembly --debug-info none
--- a/test/libsolidity/semanticTests/inlineAssembly/function_name_clash.sol
+++ b/test/libsolidity/semanticTests/inlineAssembly/function_name_clash.sol
@@ -1,0 +1,13 @@
+contract C {
+  function f() public pure returns (uint r) {
+    assembly { function f() -> x { x := 1 } r := f() }
+  }
+  function g() public pure returns (uint r) {
+    assembly { function f() -> x { x := 2 } r := f() }
+  }
+}
+// ====
+// compileViaYul: also
+// ----
+// f() -> 1
+// g() -> 2
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,7 @@ Bugfixes:
  * Commandline Interface: Don't return zero exit code when writing linked files to disk fails.
  * SMTChecker: Fix internal error in magic type access (``block``, ``msg``, ``tx``).
  * TypeChecker: Fix internal error when using user defined value types in public library functions.
+ * Yul Assembler: Fix internal error when function names are not unique.
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
 
 
--- a/libyul/backends/evm/EVMCodeTransform.cpp
+++ b/libyul/backends/evm/EVMCodeTransform.cpp
@@ -52,7 +52,7 @@ CodeTransform::CodeTransform(
 	EVMDialect const& _dialect,
 	BuiltinContext& _builtinContext,
 	ExternalIdentifierAccess::CodeGenerator _identifierAccessCodeGen,
-	bool _useNamedLabelsForFunctions,
+	UseNamedLabels _useNamedLabelsForFunctions,
 	shared_ptr<Context> _context,
 	vector<TypedName> _delayedReturnVariables,
 	optional<AbstractAssembly::LabelID> _functionExitLabel
@@ -405,7 +405,11 @@ void CodeTransform::operator()(FunctionDefinition const& _function)
 	if (!m_allowStackOpt)
 		subTransform.setupReturnVariablesAndFunctionExit();
 
+	subTransform.m_assignedNamedLabels = move(m_assignedNamedLabels);
+
 	subTransform(_function.body);
+
+	m_assignedNamedLabels = move(subTransform.m_assignedNamedLabels);
 
 	m_assembly.setSourceLocation(originLocationOf(_function));
 	if (!subTransform.m_stackErrors.empty())
@@ -585,8 +589,16 @@ void CodeTransform::createFunctionEntryID(FunctionDefinition const& _function)
 	if (_function.debugData)
 		astID = _function.debugData->astID;
 
+	bool nameAlreadySeen = !m_assignedNamedLabels.insert(_function.name).second;
+
+	if (m_useNamedLabelsForFunctions == UseNamedLabels::YesAndForceUnique)
+		yulAssert(!nameAlreadySeen);
+
 	m_context->functionEntryIDs[&scopeFunction] =
-		m_useNamedLabelsForFunctions ?
+		(
+			m_useNamedLabelsForFunctions != UseNamedLabels::Never &&
+			!nameAlreadySeen
+		) ?
 		m_assembly.namedLabel(
 			_function.name.str(),
 			_function.parameters.size(),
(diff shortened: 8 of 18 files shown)


libyul/optimiser/Disambiguator.h
```diff
@@ -55,7 +55,7 @@ class Disambiguator: public ASTCopier
 	void enterScope(Block const& _block) override;
 	void leaveScope(Block const& _block) override;
 	void enterFunction(FunctionDefinition const& _function) override;
-	void leaveFunction(FunctionDefinition const& _function) override;
+	void leaveFunction	(FunctionDefinition const& _function) override;
```
Marenz: stray space?

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -30,6 +30,30 @@
 using namespace solidity::yul;
 using namespace std;
 
+namespace
+{
+class AreFunctionNamesUnique: ASTWalker
```
Marenz: Just a minor observation: Didn't you prefer not to use `is` and terms like that (like `are`) for functions (or classes) that just check a true/false thing?

Marenz (dismissed):
LGTM, just a stray space

libyul/optimiser/Disambiguator.h
Marenz: TAB even

chriseth: I think that was my cat...

libyul/backends/evm/EVMObjectCompiler.cpp
chriseth: That is true!

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -30,6 +30,30 @@
...
+{
+public:
+	static bool check(Block& _block)
```
cameel:
```suggestion
	static bool check(Block const& _block)
```

libyul/backends/evm/EVMObjectCompiler.cpp
cameel: Maybe something like `FunctionNameUniquenessChecker` would be better? I think that nouns read better as class names.

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -72,7 +100,7 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
 		context,
 		_optimize,
 		{},
-		true /* _useNamedLabelsForFunctions */
+		useNamedLabelsForFunctions
```
cameel: Is this meant to be just a temporary workaround or a long-term solution? I'd imagine that if debuggers start depending on these labels then having them disappear completely when you change something in your code - even for functions with unique names - might be quite annoying.

test/cmdlineTests/yul_function_name_clashes/args
```diff
@@ -0,0 +1 @@
+--strict-assembly
```
cameel: I don't think the annotations add much in this particular test. Might actually be more readable without them. Maybe it would be better to disable them?
```suggestion
--strict-assembly --debug-info none
```

libyul/backends/evm/EVMObjectCompiler.cpp
chriseth: If you compile via yul, it should not be possible to have non-unique functions (we can actually add an assert for that).

So this is only for the case where you compile yul directly. We could also try to exclude just these functions, but that could be weird as well.

cameel: I think it would be better to exclude just these functions or have some convention where duplicates get suffixes or something.

But yeah, depends on how much it's used in practice. Maybe it's not worth the effort if not many people are debugging asm generated directly from Yul.

cameel:
Does this PR fix also a similar case I found in inline assembly https://github.com/ethereum/solidity/issues/12090#issuecomment-945192513?

EDIT: Reported as a separate issue (#12158).

test/cmdlineTests/inline_assembly_function_name_clash/args
```diff
@@ -0,0 +1 @@
+--experimental-via-ir --combined-json function-debug-runtime
```
cameel: Output would be more readable if it was not all on a single line:
```suggestion
--experimental-via-ir --combined-json function-debug-runtime --pretty-json --json-indent 4
```

test/cmdlineTests/inline_assembly_function_name_clash/input.sol
```diff
@@ -0,0 +1,13 @@
+contract C {
+  function f() public pure returns (uint r) {
+    assembly { function f() -> x { x := 1 } r := f() }
+  }
+  function g() public pure returns (uint r) {
+    assembly { function f() -> x { x := 2 } r := f() }
+  }
+}
```
cameel:
```suggestion
contract C {
    function f() public pure returns (uint r) {
        assembly { function f() -> x { x := 1 } r := f() }
    }
    function g() public pure returns (uint r) {
        assembly { function f() -> x { x := 2 } r := f() }
    }
}
```

Also in `function_name_clash.sol` below.

test/libsolidity/semanticTests/inlineAssembly/function_name_clash.sol
```diff
@@ -0,0 +1,13 @@
+contract C {
+  function f() public pure returns (uint r) {
+    assembly { function f() -> x { x := 1 } r := f() }
+  }
+  function g() public pure returns (uint r) {
+    assembly { function f() -> x { x := 2 } r := f() }
+  }
+}
```
cameel, line 8: Does this PR also fix #12158 or only #12149? If it solves #12158, I'd add the test case from there because it triggers a different assertion.

libyul/backends/evm/AsmCodeGen.cpp
```diff
@@ -52,7 +52,9 @@ void CodeGenerator::assemble(
 		builtinContext,
 		_optimizeStackAllocation,
 		_identifierAccessCodeGen,
-		_useNamedLabelsForFunctions
+			_useNamedLabelsForFunctions ?
+			CodeTransform::UseNamedLabels::YesAndForceUnique :
+			CodeTransform::UseNamedLabels::Never
```
cameel, line 57: This is used only for inline assembly, right? It never gets just raw Yul? From the name I'd expect it to be more generic.

test/libsolidity/semanticTests/inlineAssembly/function_name_clash.sol
cameel, line 8: I don't really understand why this example works now. I don't see anything in the PR that would make the function name actually unique, only the change to `YesAndForceUnique` so that should crash, right?

chriseth: The example never failed, unfortunately. When we run semantic tests via yul, we never switch off the optimizer (due to the stack issues).

chriseth: But [this](https://github.com/ethereum/solidity/pull/12139/files#diff-9c19c4e7d8a1f3f4179decda6f6ff45da88a9c6a781b805ed05e8796218d3dccR1) used to fail, and it works now not by making the function names unique but rather not fail if they are not unique (because it is run in full yul mode).

chriseth:
Incorporated https://github.com/ethereum/solidity/issues/12149 into the test.

libyul/backends/evm/AsmCodeGen.cpp
cameel: Answer: 
> IIRC it's "never" for inline assembly and "yes and force unique" for compiler-context "system inline assembly"
> so whenever we use m_context.appendInlineAssembly(...)
> so for example for the abi functions, I believe

For inline assembly it goes through `Never` and then (if it's IR codegen) goes through `ForFirstFunctionOfEachName` when it's assembled as Yul.
Pure Yul  input only goes through the latter.

test/libsolidity/semanticTests/inlineAssembly/function_name_clash.sol
cameel: Answer: only the Yul generated by the compiler goes though `YesAndForceUnique` path. Code from inline assembly blocks does not.

It's fine then.

cameel (dismissed):
There are still some open comments that could be applied to improve the command-line test output but overall the fix seems correct and does the job even in the current form so I'm approving.

chriseth:
Rebased, simplified commandline tests and updated the "fixes".

cameel (approved):

PR #12141, opened 2021-10-14
Tests for the equivalence of one- and two-stage compilation via IR

cameel:
Partially solves #11877.
~Depends on #12074 (draft until that PR is merged).~ Merged.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -308,6 +308,61 @@ function test_solc_assembly_output
     fi
 }
 
+function test_via_ir_equivalence()
+{
+    (( $# <= 2 )) || fail "This function accepts at most two arguments."
+
+    if [[ $2 != --optimize ]] && [[ $2 != "" ]]
+    then
+        fail "The second argument must be --optimize if present."
+    fi
+
+    local solidity_code="$1"
+    local optimize_flag="$2"
+
+    local optimizer_flags=()
+    [[ $optimize_flag == "" ]] || optimizer_flags+=("$optimize_flag")
+
+    local ir_output
+    ir_output=$(
+        echo "$solidity_code" |
+        msg_on_error --no-stderr "$SOLC" - --ir-optimized --debug-info location "${optimizer_flags[@]}" |
+        sed '/^Optimized IR:$/d'
+    )
+
+    local asm_output_two_stage asm_output_via_ir
+    asm_output_two_stage=$(
+        echo "$ir_output" |
+        msg_on_error --no-stderr "$SOLC" - --strict-assembly --asm "${optimizer_flags[@]}" |
+        sed '/^======= <stdin>/d' |
+        sed '/^Text representation:$/d'
+    )
+    asm_output_via_ir=$(
+        echo "$solidity_code" |
+        msg_on_error --no-stderr "$SOLC" - --experimental-via-ir --asm --debug-info location "${optimizer_flags[@]}" |
+        sed '/^======= <stdin>/d' |
+        sed '/^EVM assembly:$/d'
+    )
+
+    diff_values "$asm_output_two_stage" "$asm_output_via_ir" --ignore-space-change --ignore-blank-lines
+
+    local bin_output_two_stage bin_output_via_ir
+    bin_output_two_stage=$(
+        echo "$ir_output" |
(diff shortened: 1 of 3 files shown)


test/cmdlineTests.sh
```diff
@@ -530,6 +585,36 @@ printTask "Testing assemble, yul, strict-assembly and optimize..."
...
+    printTask " - Smoke test"
+    test_via_ir_equivalence "contract C {}"
+
+    printTask " - Smoke test (optimized)"
+    test_via_ir_equivalence "contract C {}" --optimize
+
+    externalContracts=(
+        deposit_contract.sol
+        FixedFeeRegistrar.sol
+        _stringutils/stringutils.sol
```
cameel, line 602: There are more in `semanticTests/externalContracts/` but these are the only ones that aren't either a multi-file test (can't compile them without splitting first) or a file containing multiple contracts (the `--ir-optimized` output needs splitting in that case).

The deposit contract takes a bit long to compile. All the compilations in this whole subshell take ~30 seconds total to run on my machine and that contract is a big part of that.

cameel:
#12074 merged. This is now reviewable.

ekpyron (dismissed):
Needs rebase, but looks good!

cameel:
@ekpyron Are we fine with the extra time it adds to CLI tests?

test/cmdlineTests.sh
chriseth: I'm actually more worried about there not being multi-file or multi-contract tests.

ekpyron:
> @ekpyron Are we fine with the extra time it adds to CLI tests?

30 seconds, is it? That's hardly noticable compared to the total runtime, isn't it :-D?

ekpyron:
Or you can move them to the end - to be honest I'm hardly ever patient enough to wait for the AST import tests locally anyways :-).

test/cmdlineTests.sh
cameel: That's doable for multi-file but I'd have to restructure the functions a bit and pass some extra paths as arguments. It would be easiest with a map and that's a real pain in Bash so I'd probably just do it in Python. Should I do that? We probably have enough single-file contracts that I didn't think it was worth spending time on.

For multi-contract it's because in Yul mode we only support compiling one file at a time so I'd have to manually split the IR output, assemble and then combine again in the right order before diffing. If we really want that, I'd rather just add a multi-file mode in the CLI than cobble it up in Bash.

cameel:
> 30 seconds, is it?

These things add up quickly and CLI tests are still much faster than a full soltest run. I'm cautious about making it longer without a good reason. Something on the order of ~1 min is a relatively big change.

> Or you can move them to the end

It's already far down, behind all the individual cases and also a few other things.

I mean, recently @Marenz even suggested refactoring the CLI test script to be able to select specific tests from the ones that run after individual cases so it does matter somewhat.

chriseth:
This can also just be a specialized test, maybe even its own script, but I think it is important to run it on a multi-file and multi-contract example.

cameel:
Ok, I'll refactor it for multi-file and multi-contract in that case.

Do we still want to merge it as is to have something for now or only once it covers these cases?

test/cmdlineTests.sh
```diff
@@ -533,6 +588,36 @@ printTask "Testing assemble, yul, strict-assembly and optimize..."
     test_solc_assembly_output "{ let x := 0 }" "{ { } }" "--strict-assembly --optimize"
 )
 
+printTask "Testing the eqivalence of --experimental-via-ir equivalence and a two-stage compilation..."
```
Marenz: This sentence phrasing has problems :D

Marenz (changes requested):

test/cmdlineTests.sh
cameel: True, not sure how the second `equivalence` got in there :)

cameel: Fixed.

Marenz (approved):

Marenz:
I suppose only merge after @chriseth responded positively to your question ^^

cameel:
I'm going to merge this to avoid keeping it open and clogging our PR queue and do the refactors in a separate PR.

PR #12142, opened 2021-10-14
Assert and stack traces for Bash scripts

cameel:
While working on #12141 I was really missing a quick way to assert stuff and the lack of a proper stack traces when a script fails makes debugging really annoying. This PR adds a rudimentary helper that can print a stack trace and a function that we can get an assert-like failure.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -92,7 +92,7 @@ echo "Using solc binary at ${SOLC}"
 INTERACTIVE=true
 if ! tty -s || [ "$CI" ]
 then
-    INTERACTIVE=""
+    INTERACTIVE=false
 fi
 
 # extend stack size in case we run via ASAN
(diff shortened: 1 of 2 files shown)


test/cmdlineTests.sh
```diff
@@ -142,12 +142,13 @@ function ask_expectation_update
...
                     u*) update_expectation "$newExpectation" "$expectationFile"; break;;
                     s*) return;;
-                    q*) exit 1;;
+                    q*) fail;;
                 esac
             done
         fi
     else
-        exit 1
+        [[ $INTERACTIVE == false ]] || assertFail
```
cameel, line 150: Initially I implemented a full assert that you would invoke like this:
```bash
assert "[[ $INTERACTIVE == false ]]"
```
It could print the condition if you did not provide a message, which was nice.

The problem is that it required `eval` to check the condition and also that quoting was tricky - you had to quote the whole condition with `""` (otherwise the variables would be evaluated inside the assert rather than in the current function) and then use `''` over any variable used inside that could evaluate to an empty string.

In the end I thought that just a simple `assertFail` that prints a stack trace is enough. It still clearly indicates that the condition you check is an assertion and condition evaluation has no quirks (at least not more than usually in Bash). Also, the condition itself can be seen in the stack trace.

Marenz (approved):

PR #12021, opened 2021-09-24
Do not whitelist remapping targets

soroosh-sdi:
first try to solve #12014.

Signed-off-by: soroosh-sdi <soroosh.sardari@gmail.com>

PR #12150, opened 2021-10-15
Merge `develop` into `breaking`

cameel:
Update of the `breaking` branch.
Needed for tests to pass in #12021.

### `Changelog.md`
```markdown
<<<<<<< HEAD
### 0.9.0 (unreleased)

Breaking changes:
 * Disallow ``.pop()`` on arrays containing nested mappings.
 * Disallow ``delete`` on types that contain nested mappings.


### 0.8.8 (unreleased)
=======
### 0.8.10 (unreleased)
>>>>>>> develop
```

### `libsolidity/analysis/ReferenceResolver.cpp`:
```c++
void ReferencesResolver::operator()(yul::VariableDeclaration const& _varDecl)
{
	for (auto const& identifier: _varDecl.variables)
<<<<<<< HEAD
		validateYulIdentifierName(identifier.name, identifier.debugData->location);
=======
	{
		solAssert(nativeLocationOf(identifier) == originLocationOf(identifier), "");
		validateYulIdentifierName(identifier.name, nativeLocationOf(identifier));

		if (
			auto declarations = m_resolver.nameFromCurrentScope(identifier.name.str());
			!declarations.empty()
		)
		{
			SecondarySourceLocation ssl;
			for (auto const* decl: declarations)
				ssl.append("The shadowed declaration is here:", decl->location());
			if (!ssl.infos.empty())
					3859_error,
					nativeLocationOf(identifier),
					ssl,
					"This declaration shadows a declaration outside the inline assembly block."
				);
		}
	}
>>>>>>> develop

	if (_varDecl.value)
		visit(*_varDecl.value);
}
```


leonardoalt (dismissed):

leonardoalt:
Maybe actually better to wait for https://github.com/ethereum/solidity/pull/12151

cameel:
Does not matter all that much. Having #12151 would be nice because it's a proper fix and makes `b_osx` job complete faster but `develop` already has a working workaround so even without that PR this update is enough to have PRs based on `breaking` pass.

cameel:
Closing in favor of #12164, which has the full CI fix and no conflicts to resolve.

PR #11267, opened 2021-04-19
Force Homebrew update to work around CircleCI auto-update issue

chriseth:

PR #12151, opened 2021-10-15
Remove CircleCI+Homebrew workarounds

cameel:
The response from CircleCI regarding the workarounds we had to do in #12106/#12131 made me realize that maybe we can just remove the `brew update` that causes this problem in the first place. We added it in #11267 because of the breakage caused by Boost binaries moving away from bintray - a newer version of Homebrew was needed to resolve it. Looks like CircleCI updated Homebrew in their images so `brew update` is no longer needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,35 +160,6 @@ defaults:
         - store_test_results: *store_test_results
         - store_artifacts: *artifacts_test_results
 
-  - steps_homebrew_workaround_before_restore_cache: &steps_homebrew_workaround_before_restore_cache
-      steps:
-        - run:
-            # Homebrew has recently replaced one of its dirs with a symlink to another.
-            # This makes tar used by restore_cache fail - the image contains an older version of
-            # Homebrew installed where it's a directory and tar refuses to overwrite a dir with a link.
-            # Removing the directory manually is a workaround that works only in case we do have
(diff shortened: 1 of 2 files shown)


docs/yul.rst
```diff
@@ -959,10 +959,13 @@ to ``loadimmutable("name")`` in the runtime code.
...
-The function ``linkersymbol("fq_library_name")`` is a placeholder for an address literal to be
-substituted by the linker. Its first and only argument must be a string literal and represents the
-fully qualified library name used with the ``--libraries`` option.
+The function ``linkersymbol("library_id")`` is a placeholder for an address literal to be substituted
+by the linker.
+Its first and only argument must be a string literal and uniquely represents the address to be inserted.
+Identifiers can be arbitrary but when the compiler produces Yul code from Solidity sources,
+it uses a library name qualified with the name of the source unit that defines that library.
+To link the code with a particular library address, the same identifier must be provided to the
+``--libraries`` option on the command line.
```
leonardoalt: Why is this part of this PR?

cameel: Oh, looks like I started the PR on top of #12144 rather than `develop`. Fixed.

leonardoalt (approved):

chriseth:
Nice!

PR #12155, opened 2021-10-16
Use range-v3 loops: delete boost includes

vowchick:
An addition to issue #10738
couldn't change boost::for_each to
std::for_each, doesn't compile.

- Build successful
- ./scripts/test.sh successful


cameel:
There's `ranges::for_each()`, which might work but actually I don't see much point in using it over a normal range loop:
```c++
for (Type& value: container)
    foo(value);
```
This is already pretty compact and does not hide the loop so I'd recommend using it instead of `for_each()`.

cameel (approved):

Issue #12119, opened 2021-10-11
[CLI] Warn when a relative path specified on the CLI ends up being absolute in metadata


cameel:
## Abstract
Now that we have #11408 and #11409 implemented, the compiler always normalizes source paths specified on the command line and only after that attempts to make them relative. If this is not possible, the path remains absolute but it's not apparent unless you inspect the metadata. Many users are probably not even aware of the issue due to this but would want to fix if if they were. I think that it would be helpful if the compiler warned about it.

#11410 would be another solution to this problem but it's not backwards-compatible and cannot be ignored like a warning - if one is just casually using the compiler to try out something without the intention to actually deploy the contract, the paths might not matter at all.

## Example
It's very easy to run into this problem:
```bash
echo 'contract C {}' > ../test.sol
solc --metadata ../test.sol
```
```
======= /tmp/test.sol:C =======
Metadata:
{"compiler":{"version":"0.8.9+commit.e5eed63a"},"language":"Solidity","output":{"abi":[],"devdoc":{"kind":"dev","methods":{},"version":1},"userdoc":{"kind":"user","methods":{},"version":1}},"settings":{"compilationTarget":{"/tmp/test.sol":"C"},"evmVersion":"london","libraries":{},"metadata":{"bytecodeHash":"ipfs"},"optimizer":{"enabled":false,"runs":200},"remappings":[]},"sources":{"/tmp/test.sol":{"keccak256":"0xbe86d3681a198587296ad6d4a834606197e1a8f8944922c501631b04e21eeba2","urls":["bzz-raw://7ef32f5b5f3b1ebe447f46969c01ce0e9190d0b90241cbc228255866326be8be","dweb:/ipfs/QmZRepHCmeSSTfxZvKU5KS7pebBYT4khWhTnqh6GzBFU3M"]}},"version":1}
```

The right way to compile such contract is to use `--base-path`:
```bash
echo 'contract C {}' > ../test.sol
solc --metadata ../test.sol --base-path ..
```
```
======= test.sol:C =======
Metadata:
[...]

Issue #12011, opened 2021-09-22
Support `--pretty-json` and `--json-indent` for all CLI options


cameel:
In  #11583 we made `--pretty-print` affect Standard JSON and also added the `--json-indent` option. There are still some command-line options that ignore these:
- `--asm-json`
- `--ast-compact-json`
- `--abi`
- `--userdoc`
- `--devdoc`
- `--storage-layout`
- `--metadata`

Some of them always pretty-print the JSON but with hard-coded indent size, some do not pretty-print at all. Change them all so that by default the JSON output is compact and changes when `--pretty-print`/`json-indent` is used.

We might want to exempt `--metadata` from this since we embed its hash in the bytecode and a pretty-printed version would not match it. But maybe that's ok - it's meant to be reproducible and being able to easily inspect it is also important. And in any case `--combined-json metadata` does get affected by `--pretty-print` so we already have this problem.

After changing it, we'll need to add pretty printing options to command-line tests to keep them readable. Alternatively, we could implement #7665 first.

## Backwards Compatibility
I don't think that we guarantee stable JSON formatting between versions so I'd say this is not breaking.

Changing the formatting for `--metadata` might affect tools that hash its output but also use `--pretty-json` to get nice output from `--combined-json` at the same time. I think it's a pretty obscure case though.

akali:
Creater PR for this issue, skipped metadata as it prints json in a bit different way

Issue #10275, opened 2020-11-12
[CLI] Allow for printing output without headings (such as bytecode)


axic:
(This issue is splitting up #8847.)

> This is more of a personal preference / nit, but in builds with only one sourcefile and contract, it could be nice to skip the explaining text and filename for outputs, such the output can be used directly in other contexts or other tools, as in:
> ```
> echo "contract A {uint public x;}" | solc --bin-runtime - > A.txt | evm --code $(<A.txt)
> ```

EDIT by @cameel:
One addition to the task (moved from #4617):

> Consistently print the header with contract name in all cases (i.e. the `======= test.sol:C =======` part). Currently `--ir`, `--ir-optimized`, `--ewasm` do not print it when used on their own.

Doing this makes piping IR output more annoying so we should really to it after we have an option to disable frames.

KukretiShubham:
Can I be assigned for this?

cameel:
My suggestion would be to call the new option `--machine-readable`.

The primary goal is to make the compiler produce only the single selected artifact on stdout, without any frames, headers or any other adornment so that it can be piped to another command. The stderr should be unaffected.

### Example
For example when you use `solc test.sol --bin`, currently you get this:
```
======= test.sol:C =======
Binary:
6080604052348015600f57600080fd5b50603f80601d6000396000f3fe6080604052600080fdfea2646970667358221220b742f1d41505dd2912b87c6f730d28a4d3d4929fc5cac1df9820a78ad007a18464736f6c63430008090033
```

With `solc test.sol --bin --machine-readable` it should look like this instead:
```
6080604052348015600f57600080fd5b50603f80601d6000396000f3fe6080604052600080fdfea2646970667358221220b742f1d41505dd2912b87c6f730d28a4d3d4929fc5cac1df9820a78ad007a18464736f6c63430008090033
```

### Specification
- The option should affect the following other options:
    - All the options from the "Output Components" group (see `solc --help`)
    - `--gas`
    - `--combined-json` (but it's already printed without any frames so it's good as is)
- It cannot be used together with with:
    - `--standard-json` (technically we could allow it though because it does not need any frames)
    - `--output-dir`
    - `--link` (in this mode the output is always written to disk)
- Doing it for assembly mode will require #3870. I'd say let's implement it for other modes and just keep this issue open until that PR is merged and we can add it.

### Open questions
There are also some open questions we need to figure out:
[...]

Issue #9601, opened 2020-08-11
solidity::util::formatNumberReadable signed integer support


leonardoalt:
Currently that function does not support signed values, causing, for example:
`Warning 3944: (110-115): Underflow (resulting value less than -57896044618658097711785492504343953926634992332820282019728792003956564819968) happens here`.
That number is actually `-2**255`, so it could be better formatted.

axic:
Probably the same issue: https://github.com/ethereum/solidity/pull/9479#discussion_r464505293

Neyromancer:
Hi, i will try to work on it.

Neyromancer:
Hi, just to clarify. Should the number `-57896044618658097711785492504343953926634992332820282019728792003956564819968 `
be transformed to `-2**255 `
or it can also be represented as 
`-0x80 * 2**248`?
Asking because the number `57896044618658097711785492504343953926634992332820282019728792003956564819968` is now transformed into `0x80 * 2**248`.

cameel:
In my opinion, if that's the current behavior in the positive case then doing a similar thing in the negative one is fine as well. It's still some improvement.

But, it would be best to just change it in both cases to `-2**255`/`2**255`. Unless there's some technical reason preventing this - I haven't really looked at the current implementation yet so there might be something I'm not aware of.

Also note that after implementing it you should update expectations in tests because that's where these numbers are used currently. But don't do it manually - you can just run `isoltest` and accept the changes it proposes (see [Writing and Running Syntax Tests](https://docs.soliditylang.org/en/latest/contributing.html#writing-and-running-syntax-tests) in the docs in case you haven't yet).

Neyromancer:
Actually, i think it is technically possible to move from `0x80 * 2 ** 248` to` 2 ** 255` so i will try to implement it.

I just was not sure whether to follow the current implementation or to do it in the way it was written in the task, i.e. which number representation to take as a basiс representation. But i understand now.
By the way thank you for the link:)

Neyromancer:
if this problem does not block anything. i am still working on it. it took a bit more than expected but guess it is only due to personal reasons.

cameel:
There's no hurry. It would be pretty helpful, especially in cases like https://github.com/ethereum/solidity/pull/10776#discussion_r559728114 but fixing this is not a priority right now so you can work on it at your own pace.

cameel:
@Neyromancer Are you still working on it? Do you need any help?

Ruko97:
[...]

PR #12157, opened 2021-10-17
Add wrong output selection artfact error

### LINKED ISSUE #12153: Output selection is not validated in Standard JSON
## Description
The compiler does not validate the values specified in the output type array inside `settings.outputSelection` in Standard JSON. It does not report an error if you specify an output that does not exist. This makes it harder to notice a problem if you misspell it or e.g. use something like `evm.metadata` instead of `metadata`.

## Steps to Reproduce
```json
{
    "language": "Solidity",
    "sources": {"contract.sol": {"content": "contract C {}"}},
    "settings": {
        "outputSelection": {"*": { "*": ["xyz", "evm.metadata", "ZZZZZZZZZZZZZZZZZZZZZZZZZZ", ""]}}
    }
}
```
This compiles without errors even though none of the specified inputs exists.

NOTE: This is about validating the array (i.e. the `["xyz", "evm.metadata", "ZZZZZZZZZZZZZZZZZZZZZZZZZZ", ""]` part in the above), not file or contract names.

akali:
Closes #12153

### DIFF
--- a/libsolidity/interface/StandardCompiler.cpp
+++ b/libsolidity/interface/StandardCompiler.cpp
@@ -580,7 +580,7 @@ std::optional<Json::Value> checkOutputSelection(Json::Value const& _outputSelect
 						"." +
 						contractName +
 						"." +
-						artifact + 
+						artifact +
 						"\" must be a valid artifact"
 					)
 				}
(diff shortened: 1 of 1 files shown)


cameel:
Hello @akali. Thanks for the PR, but unfortunately I've been too quick in labeling this as `good first issue`. Turns out that the issue still needs some discussion due to how it would affect tools (especially Truffle). This will take a while because right now many people are at a conference in Lisbon.

In the meantime, maybe you'd like to try some other issue that can be implemented right away? Here are some suggestions:
-  #12119
- #12011
- #10275
- #9601

akali:
No worries, I'll try to work on another one

PR #12164, opened 2021-10-19
Merge `develop` into `breaking`

cameel:
Replaces #12150.

Weirdly, this PR does not need any conflict resolution even though #12150 did. I could not figure out why but I guess there's not much to mess up in a straight PR from `develop` to `breaking` so if anything, the problem must have been in #12150 and not this one.

Now the full CI fix is in as requested in https://github.com/ethereum/solidity/pull/12150#issuecomment-945557620.


cameel:
> I could not figure out why but I guess there's not much to mess up in a straight PR from develop to breaking so if anything, the problem must have been in #12150 and not this one.

I somehow missed 8bcbe946c65159c88929a1f7845bb9d6217def68 which already resolved these conflicts. This means that in #12150 I must have forgotten to update my local copy of `breaking` before merge.

Issue #9237, opened 2020-06-19
Incorporating devtools' tests suites as part of the integration tests


alcuadrado:
Hi,

I'm creating this issue to propose incorporating parts of the different devtools' test suites as integration tests for solc. We believe this would make the devtools ecosystem much more stable.

Here's the rationale behind this idea.

The current situation has multiple problems:

- Small changes in the compilation output break the devtools without any previous notice
    - This is due to changes in the AST format
    - Also happened a few times in how source maps are emitted
        - Solidity 0.6.3 is a good example
        - Solidity 0.6.9 also break multiple things while reverting the change of 0.6.3
- Devtools have to implement complex and fragile heuristics to keep offering the same functionality to their users
    - If they don't do this, users blame them for breaking
    - The more they have to introduce, the more complex they are to maintain
    - These can break with any minor release, so tool developers live in a constant fear of their tools breaking at any time
- Each version of solc has to be **supported forever** by the different devtools
    - The reason for this is that once a project deploys an immutable contract compile with a given version of solc, they can't change it on-chain
        - Most projects use multiple versions of Solidity because of this
            - They use newer versions for new contracts
            - They compile already-deployed contracts with the same version used for deployment
            - They run their test combining the output of multiple compilations
                - Tools need to support this usecase, which is extremely common in the defi ecosystem
- The happiness of users with Solidity & Ethereum as a development platform is highly influenced by devtools being stable. Despite this, the Solidity project doesn't currently have an early feedback process on how their actions could affect tools.
    - They only get to know this after the tools break, and if tool developers get to report it
        - Given how often this happens, not everything gets reported.

Proposals for improving this situation

- Incorporate the relevant parts of the main tools' tests suites, as part of Solidity integration testing
    - This has been done in Web3.js recently and helped improved the stability of the project a lot. Some very subtle bugs were discovered, that weren't detected by Web3.js' tests, but broke some consumer's test suites
    - Candidates for this are
        - Truffle
        - Remix
        - Buidler EVM
        - Brownie
            - It does some very advanced things with Solidity, so it can provide coverage to parts of the compiler than other tools won't.
        - Tenderly
            - If the relevant parts are open source. Are they, @BencicAndrej ?
    - This will provide immediate feedback on how changes in Solidity affect the different tools
        - It will also help identify which of them would break, making it easier to start conversations on how to better approach each change
        - This may mean creating a dedicated communication channel for direct consumers of solc. The current channels are too broad for tool developers to keep up to date

[...]

PR #12165, opened 2021-10-20
[CI] Run hardhat tests as a PR check

### LINKED ISSUE #10854: Running Buidler's test suite in our CI
From https://github.com/ethereum/solidity/issues/9237#issuecomment-678411675:

> This is now ready on our side. All you need to do to run our test suite with a custom compiler is:
> 
> ```shell
> git clone git@github.com:nomiclabs/hardhat.git
> cd hardhat
> git checkout development
> HARDHAT_TESTS_SOLC_VERSION="0.7.0" HARDHAT_TESTS_SOLC_PATH="/some/path/soljson-v0.7.0+commit.9e61f92b.js" scripts/run-tests-with-custom-solc.sh
> ```
> 
> (Note the version and path in the last command)
> 
> This will install all the necessary dependencies and run the relevant tests. The only requirement is node.js. Any of the currently actively maintained versions would do it.
> 
> If the tests fail, the script will exit with a non-zero status number.

**EDIT**: I slightly modified the quote above to use the updated snippet from https://github.com/ethereum/solidity/issues/9237#issuecomment-769030695.

We should hook it up to run in our CI. To run either on every PR (if it takes no more than 15 min) or in the nightly run (if it takes more).

The setup should be similar to what we do in [test/externalTests/](https://github.com/ethereum/solidity/tree/develop/test/externalTests) but there's no need to fork the repo in `solidity-external-test` - it should check out Buidler's repo directly.

These tests should be running also on the `breaking` branch.

cameel:
Fixes  #10854.
~Depends on https://github.com/nomiclabs/hardhat/pull/1962.~ Fixed.
~Depends on https://github.com/nomiclabs/hardhat/issues/1969.~ Fixed.
~Depends on https://github.com/nomiclabs/hardhat/pull/1974.~ Fixed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -840,6 +840,32 @@ jobs:
             npm --version
             test/externalTests/solc-js/solc-js.sh /tmp/workspace/soljson.js $(cat /tmp/workspace/version.txt)
 
+  t_ems_ext_hardhat:
+    docker:
+      - image: circleci/node
+    environment:
+      TERM: xterm
+      HARDHAT_TESTS_SOLC_PATH: /tmp/workspace/soljson.js
+    steps:
+      - checkout
+      - attach_workspace:
+          at: /tmp/workspace
+      - run: git clone --depth 1 https://github.com/nomiclabs/hardhat.git
+      - run:
+          name: Install dependencies
+          command: |
+            cd hardhat
+            yarn
+      - run:
+          name: Run hardhat-core test suite
+          command: |
+            HARDHAT_TESTS_SOLC_VERSION=$(scripts/get_version.sh)
+            export HARDHAT_TESTS_SOLC_VERSION
+
+            # NOTE: This is expected to work without running `yarn build` first.
+            cd hardhat/packages/hardhat-core
+            yarn test
+
   t_ems_ext:
     parameters:
       project:
@@ -1110,6 +1136,7 @@ workflows:
       # Emscripten build and tests that take 15 minutes or less
       - b_ems: *workflow_trigger_on_tags
       - t_ems_solcjs: *workflow_emscripten
+      - t_ems_ext_hardhat: *workflow_emscripten
 
       - t_ems_ext:
           <<: *workflow_emscripten


cameel:
There are two open hardhat PRs that need to be merged first:
- nomiclabs/hardhat#1962 fixes two failing tests on 0.8.5.
- nomiclabs/hardhat#1969 lets us use an absolute path to the soljson.js executable which removes the need for a workaround.

Also, I made it a PR check because it's pretty fast (runs in under a minute) but given that this depends on Hardhat's `master` branch, maybe it would be better to run it only nightly to avoid accidental breakage blocking our PRs?

cameel:
All dependent issues have been resolved on Hardhat side and now `t_ems_ext_hardhat` succeeds so this is now ready for reviews.

.circleci/config.yml
```diff
@@ -840,6 +840,32 @@ jobs:
...
+            yarn
+      - run:
+          name: Run hardhat-core test suite
+          command: |
+            HARDHAT_TESTS_SOLC_VERSION=$(scripts/get_version.sh)
+            export HARDHAT_TESTS_SOLC_VERSION
+
+            # NOTE: This is expected to work without running `yarn build` first.
+            cd hardhat/packages/hardhat-core
+            yarn test
```
hrkrshnn, line 867: The comment https://github.com/ethereum/solidity/issues/9237#issuecomment-769030695 mentioned running `scripts/run-tests-with-custom-solc.sh`. Would `yarn test` also work?

Marenz (approved):
LGTM

.circleci/config.yml
Marenz: Are you asking if `yarn test` is enough or that it should be used?

cameel: It works. I took these commands from that script. The commands are simple enough that I think that it's better to run them directly. It also makes it possible to have separate steps in CI for more readable output.

hrkrshnn: Ah, just looked at the script. Nice.

hrkrshnn (approved):

PR #12166, opened 2021-10-20
Disallowed --metadata-literal, --model-checker-show-unproved, --model-checker-div-mod-no-slacks outside of compiler mode

Midhun07:
Partially solves #11629.

Disallowed following options outside of compiler mode:  
1. --metadata-literal
2. --model-checker-show-unproved
3. --model-checker-div-mod-no-slacks 
4. . --via-ir
5. --metadata-hash=swarm

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
 
 
 ### 0.8.15 (2022-06-15)
--- a/solc/CommandLineParser.cpp
+++ b/solc/CommandLineParser.cpp
@@ -911,12 +911,26 @@ void CommandLineParser::processArgs()
 		// TODO: This should eventually contain all options.
 		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
 		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
-		{g_strViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}}
+		{g_strViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataLiteral, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataHash, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerShowUnproved, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerDivModNoSlacks, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerEngine, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerInvariants, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerSolvers, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerTimeout, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerContracts, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerTargets, {InputMode::Compiler, InputMode::CompilerWithASTImport}}
 	};
 	vector<string> invalidOptionsForCurrentInputMode;
 	for (auto const& [optionName, inputModes]: validOptionInputModeCombinations)
 	{
-		if (m_args.count(optionName) > 0 && inputModes.count(m_options.input.mode) == 0)
+		if (
+			m_args.count(optionName) > 0 &&
+			inputModes.count(m_options.input.mode) == 0 &&
+			!m_args[optionName].defaulted()
+		)
 			invalidOptionsForCurrentInputMode.push_back(optionName);
 	}
 
--- a/test/solc/CommandLineParser.cpp
+++ b/test/solc/CommandLineParser.cpp
@@ -281,20 +281,6 @@ BOOST_AUTO_TEST_CASE(assembly_mode_options)
 			"--libraries="
 				"dir1/file1.sol:L=0x1234567890123456789012345678901234567890,"
 				"dir2/file2.sol:L=0x1111122222333334444455555666667777788888",
-			"--metadata-hash=swarm",       // Ignored in assembly mode
-			"--metadata-literal",          // Ignored in assembly mode
-			"--model-checker-contracts="   // Ignored in assembly mode
-				"contract1.yul:A,"
-				"contract2.yul:B",
-			"--model-checker-div-mod-no-slacks", // Ignored in assembly mode
-			"--model-checker-engine=bmc",  // Ignored in assembly mode
-			"--model-checker-invariants=contract,reentrancy",  // Ignored in assembly mode
-			"--model-checker-show-unproved", // Ignored in assembly mode
-			"--model-checker-solvers=z3,smtlib2", // Ignored in assembly mode
-			"--model-checker-targets="     // Ignored in assembly mode
-				"underflow,"
-				"divByZero",
-			"--model-checker-timeout=5",   // Ignored in assembly mode
 			"--asm",
 			"--bin",
 			"--ir-optimized",
@@ -378,20 +364,6 @@ BOOST_AUTO_TEST_CASE(standard_json_mode_options)
 			"dir2/file2.sol:L=0x1111122222333334444455555666667777788888",
 		"--gas",                           // Accepted but has no effect in Standard JSON mode
 		"--combined-json=abi,bin",         // Accepted but has no effect in Standard JSON mode
-		"--metadata-hash=swarm",           // Ignored in Standard JSON mode
-		"--metadata-literal",              // Ignored in Standard JSON mode
-		"--model-checker-contracts="       // Ignored in Standard JSON mode
-			"contract1.yul:A,"
-			"contract2.yul:B",
-		"--model-checker-div-mod-no-slacks", // Ignored in Standard JSON mode
-		"--model-checker-engine=bmc",      // Ignored in Standard JSON mode
-		"--model-checker-invariants=contract,reentrancy",      // Ignored in Standard JSON mode
-		"--model-checker-show-unproved",      // Ignored in Standard JSON mode
-		"--model-checker-solvers=z3,smtlib2", // Ignored in Standard JSON mode
-		"--model-checker-targets="         // Ignored in Standard JSON mode
-			"underflow,"
-			"divByZero",
-		"--model-checker-timeout=5",       // Ignored in Standard JSON mode
 	};
 
 	CommandLineOptions expectedOptions;
@@ -424,16 +396,30 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
 		// TODO: This should eventually contain all options.
 		{"--error-recovery", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
 		{"--experimental-via-ir", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
-		{"--via-ir", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}}
+		{"--via-ir", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--metadata-literal", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--metadata-hash=swarm", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-show-unproved", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-div-mod-no-slacks", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-engine=bmc", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-invariants=contract,reentrancy", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-solvers=z3,smtlib2", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-timeout=5", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-contracts=contract1.yul:A,contract2.yul:B", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--model-checker-targets=underflow,divByZero", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}}
 	};
 
 	for (auto const& [optionName, inputModes]: invalidOptionInputModeCombinations)
 		for (string const& inputMode: inputModes)
 		{
 			stringstream serr;
+			size_t separatorPosition = optionName.find("=");
+			string optionNameWithoutValue = optionName.substr(0, separatorPosition);
+			soltestAssert(!optionNameWithoutValue.empty());
+
 			vector<string> commandLine = {"solc", optionName, "file", inputMode};
 
-			string expectedMessage = "The following options are not supported in the current input mode: " + optionName;
+			string expectedMessage = "The following options are not supported in the current input mode: " + optionNameWithoutValue;
 			auto hasCorrectMessage = [&](CommandLineValidationError const& _exception) { return _exception.what() == expectedMessage; };
 
 			BOOST_CHECK_EXCEPTION(parseCommandLine(commandLine), CommandLineValidationError, hasCorrectMessage);


Midhun07:
@cameel Added some more fixes please check.

Changelog.md
```diff
@@ -17,6 +17,7 @@ Bugfixes:
  * Commandline Interface: Fix extra newline character being appended to sources passed through standard input, affecting their hashes.
  * Commandline Interface: Report output selection options unsupported by the selected input mode instead of ignoring them.
  * Commandline Interface: Don't return zero exit code when writing linked files to disk fails.
+ * Commandline Interface: Disallow following options outside of the compiler mode:``--error-recovery``, ``--metadata-literal``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``.
```
```suggestion
 * Commandline Interface: Disallow the following options outside of the compiler mode: ``--error-recovery``, ``--metadata-literal``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``.
```

solc/CommandLineParser.cpp
```diff
@@ -898,13 +898,24 @@ bool CommandLineParser::processArgs()
...
-	if (
-		m_args.count(g_strExperimentalViaIR) > 0 &&
-		m_options.input.mode != InputMode::Compiler &&
-		m_options.input.mode != InputMode::CompilerWithASTImport
-	)
+	map<string, set<InputMode>> validOptionInputModeCombinations = {
+		// TODO: This should eventually contain all options.
+		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataLiteral, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
```
cameel, line 915: You could also take care of `--metadata-hash` option in the same PR. It's closely related and used in the same modes (i.e. compiler only).

solc/CommandLineParser.cpp
```diff
@@ -898,13 +898,24 @@ bool CommandLineParser::processArgs()
...
+		{g_strModelCheckerShowUnproved, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerDivModNoSlacks, {InputMode::Compiler, InputMode::CompilerWithASTImport}}
```
cameel: Same here. All the `--model-checker-` options are closely related.

cameel (commented):
Overall this looks good but please rebase on `develop` (I have just merged #11909). I thought you'll start a new branch on top of the one from #11909.

Also, changes are pretty simple at this point, so you could cover more options in one go. Especially the ones that are related.

hrkrshnn:
@Midhun07 Thanks for the PR. Could you please follow up with the changes that cameel requested?

solc/CommandLineParser.cpp
Midhun07: The reason why i left this option out is because it has a value assignment to it. So do we need to check that or simply checking if option is enabled or not is enough?

solc/CommandLineParser.cpp
Midhun07: Same here.

Midhun07:
@hrkrshnn @cameel  I have added changelog change. Other changes I need some help as in I just want to know how to check the values of the options. Like for example: `metadata_hash=swarm` How to check this option? Simple checking like other options is not working. Please help.

solc/CommandLineParser.cpp
cameel: It's enough to check if the option is present (with `count()`). The check applies to the option as a whole, not specific values.

cameel:
> Simple checking like other options is not working. Please help.

For most options `m_args.count(optionName) > 0` should work. For options that have default values and are always present because of this, you must check `!m_args[optionaName].defaulted()` instead.

Midhun07:
@cameel My problem is that the message `The following options are not supported in the current input mode: --metadata-hash` is the error produced but the test compares that message against the message: `The following options are not supported in the current input mode: --metadata-hash=swarm`.  Is there any way to append `=swarm` apart from the usual string append? I mean something like `joinOptionNames`.

cameel:
Well, you can do it in several ways:
- Split on `=` and remove everything after it when comparing.
- Have a more complex data structure. I.e. under `--metadata-hash` store both the allowed modes and a version with parameter value (`--metadata-hash=swarm`). Pass one into the command, use the other for comparison.
- Have two separate lists, one for options with values (and include the value there), one for options without.

cameel:
I mean, you don't have to stick to the current stucture at all cost. You need to adjust it for this to work.

hrkrshnn:
@Midhun07 Could you please rebase into develop instead of the merge? That is our workflow.

Midhun07:
@hrkrshnn I have rebased my branch on top of `ethereum:develop`. Not sure though if it had worked properly. A little help would be much appreciated. Some 14 new commits have crept in. Not sure if this is the right way.

cameel:
@Midhun07 I think you must have forgotten to pull in the latest commits into your local `develop` branch before rebasing.

cameel:
Actually looks like it's something else. I'd have to look closer to see what actually happened but it was definitely now what should have happened :)  In any case it's the result of the branch not being properly rebased. If something like this happens and you're not sure what's wrong it's usually a good idea to visualize the branches. E.g. with `gitk --all` or `git log --all --graph`. The problem is usually easy to see.

Anyway, I rebased it properly and pushed just now.

solc/CommandLineParser.cpp
```diff
@@ -910,6 +910,10 @@ bool CommandLineParser::processArgs()
 		// TODO: This should eventually contain all options.
 		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
 		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataLiteral, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerShowUnproved, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerDivModNoSlacks, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataHash, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
```
cameel, line 916: I'd put `g_strMetadataHash` next to `g_strMetadataLiteral`.

test/solc/CommandLineParser.cpp
```diff
@@ -420,18 +412,24 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
 	};
 
 	for (auto const& [optionName, inputModes]: invalidOptionInputModeCombinations)
 		for (string const& inputMode: inputModes)
 		{
 			stringstream sout, serr;
-			vector<string> commandLine = {"solc", optionName, "file", inputMode};
+			size_t pos = optionName.find("=");
+			string optionNameWithoutValue = (pos != string::npos) ? optionName.substr(0, pos) : optionName;
+			vector<string> commandLine = {"solc", optionNameWithoutValue, "file", inputMode};
```
cameel: Isn't it supposed to have a value on the command line?
```suggestion
			vector<string> commandLine = {"solc", optionName, "file", inputMode};
```

test/solc/CommandLineParser.cpp
```diff
@@ -420,18 +412,24 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+		{"--model-checker-div-mod-no-slacks", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
+		{"--metadata-hash=swarm", {"--assemble", "--yul", "--strict-assembly", "--standard-json", "--link"}},
 	};
 
 	for (auto const& [optionName, inputModes]: invalidOptionInputModeCombinations)
 		for (string const& inputMode: inputModes)
 		{
 			stringstream sout, serr;
-			vector<string> commandLine = {"solc", optionName, "file", inputMode};
+			size_t pos = optionName.find("=");
```
```suggestion
			size_t separatorPosition = optionName.find("=");
```

test/solc/CommandLineParser.cpp
```diff
@@ -420,18 +412,24 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+			string optionNameWithoutValue = (pos != string::npos) ? optionName.substr(0, pos) : optionName;
```
cameel: `npos` in `substr()` represents the end of the string so you don't have to have a special case for it.
```suggestion
			string_view optionNameWithoutValue = optionName.substr(0, optionName.find("="));
```

Also, you can use `string_view` here to avoid copying the string.

cameel: I'd add an assert against the name being empty here:
```c++
soltestAssert(!optionNameWithoutValue.empty());
```

Just to make sure that if someone by mistake uses `=` at the beginning of an option name on the list above, the code just stops instead of doing something weird (and potentially not even failing).

test/solc/CommandLineParser.cpp
Midhun07: If we give the option as it is, it will throw an error as we are checking if the option is added or not in the `CommandLineParser`. That is why changed it to this.

cameel: Not sure I understand. Are you saying that e.g. `solc --standard-json --metadata-hash=swarm` will report an error and `solc --standard-json --metadata-hash` won't? Isn't the error for `solc --standard-json --metadata-hash=swarm` exactly what should happen?

`solc  --standard-json --metadata-hash` does report an error but it's not the one we expect:
```
the required argument for option '--metadata-hash' is missing
```

Midhun07: No that is not what I meant. In `CommandLineParser.cpp` we are checking if the option exists or not. There the string is `--metadata-hash` and not `--metadata-hash=swarm`. So in `test/CommandlineParser.cpp` we should be passing only `--metadata-hash` so that the test passes.

Midhun07: I'm also asking if this is the right way to test this case. Since the option itself is invalid we need to check its value so I made it like this.

cameel: If I understand correctly, you're worried that `CommandLineParser` will see `--metadata-hash=swarm` as a single option named `metadata-hash=swarm` rather than `metadata-hash` option with a value of `swarm`? That should not be the case. boost::program_options handles both `--metadata-hash swarm` and `--metadata-hash=swarm` in the same way. So if you call `solc --metadata-hash=swarm`, `m_args.count("metadata-hash")` should still give you 1.

solc/CommandLineParser.cpp
cameel: So what about this? Can you add all model checker options? I think we should do them all in one go.

cameel:
This needed a rebase and had some small issues (tabs in the middle of a line in a few places, not all options listed in the changelog. Also we have just released 0.8.10 so the changelog entry should go to 0.8.11 now.

I fixed these and pushed my changes to your branch.

Not all comments above are resolved so I cannot approve yet. Please take a look at them and let me know if any of them is problematic.

dzhalaevd (dismissed):

test/solc/CommandLineParser.cpp
Midhun07: This is removed.

Midhun07:
Hi @cameel, I have been trying to implement the suggestions. The `string_view` isn't working fine. I'm not able to reconstruct the original string back. It is printing garbage. Here is what I'm trying to do.

`string_view optionNameWithoutValue = optionName.substr(0, optionName.find("="));`
this is to get the address and the length of the substring. And then to reconstruct it:
`string(optionNameWithoutValue.begin(), optionNameWithoutValue.length())`.

I have tried this on a template and there it was working fine. For some reason it is not working here. Is there any issue with this reconstruction? Also I have been trying to debug using `soltest.sh --debug`. But the working of the test cases is little different from a normal execution i think. I'm not able to print the value of the variables.

cameel:
@Midhun07 Try `string(optionNameWithoutValue)`.

But I'm really surprised that the operator for adding a `string_view` to `string` is not available in C++. Apparently it's by design: [Why is there no support for concatenating std::string and std::string_view?](https://stackoverflow.com/questions/44636549/why-is-there-no-support-for-concatenating-stdstring-and-stdstring-view/47735624#47735624).

This makes this change pointless so, well, let's go back to `string`.

> Also I have been trying to debug using `soltest.sh --debug`. But the working of the test cases is little different from a normal execution i think. I'm not able to print the value of the variables.

It's debuggable just as `solc`. I just checked and I can actually inspect variables. The only ones I can't are the "fake" ones that come from destructuring (`optionName` and `inputModes`). Maybe you were trying to inspect these?

leonardoalt:
This PR has been stale for a while. Maybe take over?

cameel:
@nishant-sachdeva 
> Slight doubt regarding git. 
> I took over the PR https://github.com/Midhun07/solidity/tree/fix_invalid_options_mode_combinations.
> Have rebased i and  pushed changes to it
> But that is not being reflected in the PR created from this branch at https://github.com/ethereum/solidity/pull/12166
 >
> So, the question is, could I be missing a step here? Any inputs would be great!

Are you sure you're pushing to the right remote? This is an extrnal PR so you need to add the contributor's fork as a remote first:
```bash
git remote add Midhun07 git@github.com:Midhun07/solidity.git
```
Then you can do `git push Midhun07 fix_invalid_options_mode_combinations --force-with-lease`

cameel:
By the way I see that you did manage to push here. But something went very wrong because there's a ton of extra commits. Is the branch properly rebased on `develop`? Also make sure it was rebased, not merged. There should be no merge commits.

solc/CommandLineParser.cpp
```diff
@@ -911,7 +911,14 @@ void CommandLineParser::processArgs()
...
 		{g_strErrorRecovery, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
 		{g_strExperimentalViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+<<<<<<< HEAD
 		{g_strViaIR, {InputMode::Compiler, InputMode::CompilerWithASTImport}}
+=======
+		{g_strMetadataLiteral, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strMetadataHash, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerShowUnproved, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerDivModNoSlacks, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+>>>>>>> Disallowed --metadata-literal, --model-checker-show-unproved, --model-checker-div-mod-no-slacks ----metadata-hash=swarm outside of compiler mode
```
cameel: Looks like some unresolved conflicts were committed.

@nishant-sachdeva Will you be continuing this? At least this should be cleaned up but I see you unassigned yourself.

test/solc/CommandLineParser.cpp
nishant-sachdeva: Marking this as resolved, since in a later comment, it's been advised to abandon the use of `string_view`

test/solc/CommandLineParser.cpp
```diff
@@ -424,19 +416,28 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+			string optionNameWithoutValue = optionName.substr(0, separatorPosition);
+			soltestAssert(!optionNameWithoutValue.empty());
+
 			vector<string> commandLine = {"solc", optionName, "file", inputMode};
 
 			string expectedMessage = "The following options are not supported in the current input mode: " + optionName;
 			auto hasCorrectMessage = [&](CommandLineValidationError const& _exception) { return _exception.what() == expectedMessage; };
 
 			BOOST_CHECK_EXCEPTION(parseCommandLine(commandLine), CommandLineValidationError, hasCorrectMessage);
+			BOOST_TEST(serr.str() == "The following options are not supported in the current input mode: " + optionNameWithoutValue + "\n");
```
cameel: @nishant-sachdeva 
> Can't quite figure out how to fix the Boost tests that are failing.
> Would love some pointers on that

`BOOST_TEST` here is failing because stderr is empty. And it's right. `parseCommandLine()` is supposed to throw an exception, not print to stderr and `BOOST_CHECK_EXCEPTION()` already correctly checks for that. The `BOOST_TEST` line should not be here at all.

Also, the error message you're getting is a bit confusing so let's look at it closer:
```
message: check serr.str() == "The following options are not supported in the current input mode: " + optionNameWithoutValue + "\n" has failed [ != The following options are not supported in the current input mode: --error-recovery
]
```
What it's saying is that `""` is not equal to `"The following options are not supported in the current input mode: --error-recovery"`. It's just that there are no quotes around the strings so it looks as if the empty string were not there and it were comparing with `"The following options are not supported in the current input mode: " + optionNameWithoutValue + "\n"` instead.

solc/CommandLineParser.cpp
cameel: @nishant-sachdeva 
> Since some of the model-checker options have different values for the flags, and we may not need to disallow all of the values, is there some compiled list / documentation which mentions what options are allowed for which compiler mode ?
> Or do have to manually go through the code and check that ?

If you want the values allowed on the CLI see `solc --help` or [this part of `CommandLineParser::optionsDescription()`](https://github.com/ethereum/solidity/blob/v0.8.14/solc/CommandLineParser.cpp#L785-L834).

Same info for Standard JSON is under [Compiler Input and Output JSON Description > Input description](https://docs.soliditylang.org/en/latest/using-the-compiler.html#input-description). Look at the `modelChecker` field.

And [SMTChecker and Formal Verification](https://docs.soliditylang.org/en/latest/smtchecker.html) has an in-depth explanation of what these options do.

test/solc/CommandLineParser.cpp
nishant-sachdeva: I see. 
I'll remove the BOOST_TEST line and push changes.

solc/CommandLineParser.cpp
cameel: @nishant-sachdeva This one is still unresolved. Can you add all the model checker options?

Changelog.md
```diff
@@ -13,6 +13,7 @@ Compiler Features:
 Bugfixes:
  * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
  * Yul Optimizer: Do not remove ``returndatacopy`` in cases in which it might perform out-of-bounds reads that unconditionally revert as out-of-gas. Previously, any ``returndatacopy`` that wrote to memory that was never read from was removed without accounting for the out-of-bounds condition.
+ * Commandline Interface: Disallow the following options outside of the compiler mode: ``--error-recovery``,``--experimental-via-ir``,``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``.
```
cameel: `--experimental-via-ir` and `--error-recovery` were already covered in previous PRs.
```suggestion
 * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``, ``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``.
```

test/solc/CommandLineParser.cpp
cameel: @nishant-sachdeva I think this still needs to be changed.

solc/CommandLineParser.cpp
nishant-sachdeva: Pushed code to resolve this comment.

test/solc/CommandLineParser.cpp
nishant-sachdeva: Added this change in a later push to the code.

Changelog.md
```diff
@@ -19,6 +19,7 @@ Bugfixes:
  * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
  * Common Subexpression Eliminator: Process assembly items in chunks with maximum size of 2000. It helps to avoid extremely time-consuming searches during code optimization.
  * Yul Optimizer: Do not remove ``returndatacopy`` in cases in which it might perform out-of-bounds reads that unconditionally revert as out-of-gas. Previously, any ``returndatacopy`` that wrote to memory that was never read from was removed without accounting for the out-of-bounds condition.
+ * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
```
cameel: Needs to be moved to 0.8.16 now.

solc/CommandLineParser.cpp
```diff
@@ -911,12 +911,25 @@ void CommandLineParser::processArgs()
...
+		{g_strModelCheckerContracts, {InputMode::Compiler, InputMode::CompilerWithASTImport}},
+		{g_strModelCheckerTargets, {InputMode::Compiler, InputMode::CompilerWithASTImport}}
 	};
 	vector<string> invalidOptionsForCurrentInputMode;
 	for (auto const& [optionName, inputModes]: validOptionInputModeCombinations)
 	{
-		if (m_args.count(optionName) > 0 && inputModes.count(m_options.input.mode) == 0)
+		if (
+			(m_args.count(optionName) > 0 && inputModes.count(m_options.input.mode) == 0) &&
+			(!m_args[optionName].defaulted())
```
```suggestion
			m_args.count(optionName) > 0 &&
			inputModes.count(m_options.input.mode) == 0 &&
			!m_args[optionName].defaulted()
```

cameel (dismissed):
Looks good now apart from some minor style issues. I'm approving because I won't be available later but we do still need to move the changelog entry before we merge.

ekpyron (approved):
Reapproving on a casual look only, based on @cameel's earlier approval.

PR #12167, opened 2021-10-20
[SMTChecker] Merge all entry points for a target

leonardoalt:
Previously, for every verification target (an assertion for example), we created N queries where N is the number of entry points (public/external functions + constructor) that could syntactically reach the assertion.
The N queries were very similar, where the only difference was the entry point.
Say we have functions F1, F2 and F3. The 3 queries would be
```
for each F:
    add rule (interface && F && constraints => error) and query for error
```

In the new version we have
```
for each F:
    add rule (interface && F && constraints => error)
query for error
```

Each query is a bit lighter in the previous version, but the queries are very similar so it's a lot of redundancy to keep asking almost the same question.

When we use z3's C++ interface this isn't really a problem because the CHC system is already in memory, and we just add the new rules above one at a time.
However, when we use the smtlib2 interface, the queries are individually dumped, so each smt2 file has a big redundant part. One example is the erc777 openzeppelin contract that I'm verifying:
in the previous version it creates 7 queries of 117MB, whereas in the new version it's one query of 121MB.
The solver before took 11.7 minutes to solve all 7 queries, and now takes 85s to solve the one query.

### DIFF
--- a/test/libsolidity/smtCheckerTests/inheritance/base_contract_assertion_fail_1.sol
+++ b/test/libsolidity/smtCheckerTests/inheritance/base_contract_assertion_fail_1.sol
@@ -15,5 +15,5 @@ contract C is B {
 // ====
 // SMTEngine: all
 // ----
-// Warning 6328: (52-66): CHC: Assertion violation happens here.\nCounterexample:\ny = 0, x = 1\n\nTransaction trace:\nC.constructor()\nState: y = 0, x = 0\nC.g()\n    B.f() -- internal call\nState: y = 0, x = 1\nB.f()
+// Warning 6328: (52-66): CHC: Assertion violation happens here.\nCounterexample:\ny = 0, x = 1\n\nTransaction trace:\nC.constructor()\nState: y = 0, x = 0\nC.g()\n    B.f() -- internal call
 // Info 1180: Contract invariant(s) for :B:\n(x <= 0)\n
--- a/test/libsolidity/smtCheckerTests/operators/conditional_assignment_5.sol
+++ b/test/libsolidity/smtCheckerTests/operators/conditional_assignment_5.sol
@@ -25,4 +25,6 @@ contract C {
 // SMTIgnoreCex: yes
 // ----
 // Warning 2072: (249-255): Unused local variable.
-// Warning 6328: (271-295): CHC: Assertion violation happens here.
+// Warning 1218: (271-295): CHC: Error trying to invoke SMT solver.
+// Warning 6328: (271-295): CHC: Assertion violation might happen here.
+// Warning 4661: (271-295): BMC: Assertion violation happens here.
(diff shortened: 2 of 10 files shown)


test/libsolidity/smtCheckerTests/operators/conditional_assignment_5.sol
```diff
@@ -25,4 +25,6 @@ contract C {
 // SMTIgnoreCex: yes
 // ----
 // Warning 2072: (249-255): Unused local variable.
-// Warning 6328: (271-295): CHC: Assertion violation happens here.
+// Warning 1218: (271-295): CHC: Error trying to invoke SMT solver.
```
blishko, line 28: Any idea why this error popped up? The contract seems relatively simple and the solver should be able to solve it.

leonardoalt: It has an external call to unknown code, so maybe it just got stuck within the rlimit.

test/libsolidity/smtCheckerTests/inheritance/base_contract_assertion_fail_1.sol
```diff
@@ -15,4 +15,4 @@ contract C is B {
 // ====
 // SMTEngine: all
 // ----
-// Warning 6328: (52-66): CHC: Assertion violation happens here.\nCounterexample:\ny = 0, x = 1\n\nTransaction trace:\nC.constructor()\nState: y = 0, x = 0\nC.g()\n    B.f() -- internal call\nState: y = 0, x = 1\nB.f()
+// Warning 6328: (52-66): CHC: Assertion violation happens here.\nCounterexample:\ny = 0, x = 1\n\nTransaction trace:\nC.constructor()\nState: y = 0, x = 0\nC.g()\n    B.f() -- internal call
```
blishko, line 18: What is the information that got lost here?
What does 
```
State: y = 0, x = 1 
B.f()
```
mean?

blishko (commented):
I think this change is great!
I tried to went through the encoding on a paper and I agree that it does what you wrote in the description.
This is very useful (especially for safe contracts) because it reduces the number of issued queries significantly.

I am just wondering about the two changes in the tests.
I agree that the second one could be just the solver not able to find the counterexample in the given time (although I would like to double-check that).
I don't know what the change in the second test represents.

test/libsolidity/smtCheckerTests/inheritance/base_contract_assertion_fail_1.sol
leonardoalt: It's just a different counterexample. The previous one used to perform an extra call to `B.f()`, and the new one doesn't but can still break the assertion.

blishko: I see. The new counterexample seems to be even better, because the old one was ignoring the fact, that the assertion would be triggered already in the first call to `f`.
So this is an improvement.

blishko (dismissed):

leonardoalt:
Just rebased.
@hrkrshnn how about this one?
@blishko already approved it from the CHC side

ekpyron (approved):

PR #12173, opened 2021-10-21
[CI] Base image templates

cameel:
Implements part of #11846.
~Depends on #12182 (draft until it's merged).~ Merged.

This is a refactor I did while trying to enable gitter notifications for all nightly jobs. The fact that steps are often shared by being included in the `test_` templates made adding notifications only for jobs that do not run on PRs a bit complicated.

This PR replaces them with a consistent set of `base_` templates that define the base docker image, resource class and some default environment variables and steps are now always provided explicitly in the job. It also renames jobs so that it's easier to see what they run - `soltest` or `soltest_all` is now always included in the name and the python job names now follow the same naming convention as other jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -491,7 +491,7 @@ jobs:
       - checkout
       - run: *run_docs_pragma_min_version
 
-  t_pyscripts_ubu:
+  t_ubu_pyscripts:
     <<: *base_ubuntu2004
     steps:
       - checkout
@@ -499,7 +499,7 @@ jobs:
           name: Python unit tests
           command: python3 test/pyscriptTests.py
 
-  t_pyscripts_win:
+  t_win_pyscripts:
     <<: *base_win_powershell
     steps:
       - run: git config --global core.autocrlf false
@@ -767,7 +767,7 @@ jobs:
           path: docs/_build/html/
           destination: docs-html
 
-  t_ubu_soltest: &t_ubu_soltest
+  t_ubu_soltest_all: &t_ubu_soltest_all
     <<: *base_ubuntu2004
     parallelism: 6
     <<: *steps_soltest_all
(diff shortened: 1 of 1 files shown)


cameel:
#12182 has been merged so this is now reviewable.

cameel:
All tests passed here (https://app.circleci.com/pipelines/github/ethereum/solidity/19979/workflows/82b5e272-f459-4732-ab33-373217d9d88a). Not sure why some are still listed as pending on the list above.

ekpyron (dismissed):

cameel:
Looks like there's some glitch here. Github is still reporting the jobs as pending. Rerunning them all on CircleCI did not help. I have reported the issue to CircleCI.

I'll try pushing the branch again to see if that helps.

cameel:
Uh, some jobs are stuck even after the push. Let's see if rebase on `develop` (which has new commits now) will help. If not, I may need to just create a new PR with the same content.

cameel:
Still stuck...

cameel:
I got a response from CircleCI. Looks like this is because I renamed some jobs but the ones with the old names are still marked as required. For example `t_pyscripts_ubu` is now called `t_ubu_pyscripts`. Both are on the list but the former is pending because it does not exist in the workflow on the CircleCI side.

In that case I need help from someone with access to settings to switch the required checks to the new ones. Here's how I renamed them:
| old name                | new name                     |
|-------------------------|------------------------------|
| `t_pyscripts_ubu`       | `t_ubu_pyscripts `           |
| `t_pyscripts_win`       | `t_win_pyscripts `           |
| `t_ubu_release_soltest` |  `t_ubu_release_soltest_all` |
| `t_ubu_soltest`         | `t_ubu_soltest_all`          |
| `t_win`                 | `t_win_soltest`              |
| `t_win_release`         | `t_win_release_soltest`      |

@chriseth can you help?

PR #12175, opened 2021-10-21
Define namespace for C++ 17

dzhalaevd:
Since the definition of a nested namespace is an extension of C++17, we need to define each namespace separately.
Also made the code more beautiful

### DIFF
--- a/libsolidity/interface/DebugSettings.h
+++ b/libsolidity/interface/DebugSettings.h
@@ -22,18 +22,17 @@
 #pragma once
 
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


cameel (changes requested):
I'm sorry but I don't think we're going to accept this PR. `namespace solidity::frontend` does match our preferred style and does not need changing. Other changes in this PR also do not seem like an improvement and on top of that the code is just broken (`revertStringsFromString()` got copied into a `switch` statement without any regard for the code around it).

PR #12176, opened 2021-10-21
Refactor code

dzhalaevd:


cameel (changes requested):
Again, I'm going to close this PR because it's not the kind of refactor we would accept. The code no longer matches the style used in the rest of the project and that's the only change introduced by the PR. Please note the failing `chk_coding_style` CI job.

PR #12177, opened 2021-10-21
Include actual contract size in the code size warning

### LINKED ISSUE #12171: Contract size warning missing a crucial bit of information required in order to fix it
## Abstract

When a contract exceeds the maximum size (24K), it would be very helpful to know the actual size of that specific contract.

## Motivation

Allowing the developer to perform a more efficient process of reducing the "rogue" contract.

## Specification

In https://github.com/ethereum/solidity/blob/develop/libsolidity/interface/CompilerStack.cpp#L1241-L1255, change this:
```
"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
"This contract may not be deployable on mainnet. "
"Consider enabling the optimizer (with a low \"runs\" value!), "
"turning off revert strings, or using libraries."
```
To I guess something like this:
```
(
    string("Contract code size = ") +
    std:to_string(compiledContract.runtimeObject.bytecode.size()) +
    " > 24576 bytes (a limit introduced in Spurious Dragon). "
    "This contract may not be deployable on mainnet. "
    "Consider enabling the optimizer (with a low \"runs\" value!), "
    "turning off revert strings, or using libraries."
).c_str()
```

Thanks :)

dzhalaevd:
Contract size warning missing a crucial bit of information required in order to fix it
Fixes #12171

### DIFF
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -83,6 +83,7 @@
 #include <utility>
 #include <map>
 #include <limits>
+#include <string>
 
 using namespace std;
 using namespace solidity;
@@ -1248,7 +1249,9 @@ void CompilerStack::assemble(
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
+			"Contract code size is "s +
+			to_string(compiledContract.runtimeObject.bytecode.size()) +
+			" bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
 			"This contract may not be deployable on mainnet. "
 			"Consider enabling the optimizer (with a low \"runs\" value!), "
 			"turning off revert strings, or using libraries."
--- a/test/libsolidity/SyntaxTest.cpp
+++ b/test/libsolidity/SyntaxTest.cpp
@@ -77,6 +77,8 @@ void SyntaxTest::setupCompiler()
 		OptimiserSettings::full() :
 		OptimiserSettings::minimal()
 	);
+	compiler().setMetadataFormat(CompilerStack::MetadataFormat::NoMetadata);
+	compiler().setMetadataHash(CompilerStack::MetadataHash::None);
 }
 
 void SyntaxTest::parseAndAnalyze()
--- a/test/libsolidity/syntaxTests/bytecode_too_large_abiencoder_v1.sol
+++ b/test/libsolidity/syntaxTests/bytecode_too_large_abiencoder_v1.sol
@@ -1,4 +1,4 @@
-pragma abicoder v2;
+pragma abicoder v1;
 
 contract test {
     function f() public pure returns (string memory ret) {
@@ -9,4 +9,4 @@ contract test {
 // ====
 // EVMVersion: >byzantium
 // ----
-// Warning 5574: (21-27154): Contract code size is 27199 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
+// Warning 5574: (21-27154): Contract code size is 27209 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
--- a/test/libsolidity/syntaxTests/bytecode_too_large_byzantium.sol
+++ b/test/libsolidity/syntaxTests/bytecode_too_large_byzantium.sol
@@ -1,5 +1,3 @@
-pragma abicoder v2;
-
 contract test {
     function f() public pure returns (string memory ret) {
         // 27000 bytes long data
@@ -7,6 +5,6 @@ contract test {
     }
 }
 // ====
-// EVMVersion: >byzantium
+// EVMVersion: =byzantium
 // ----
-// Warning 5574: (21-27154): Contract code size is 27199 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
+// Warning 5574: (0-27133): Contract code size is 27227 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
--- a/test/libsolidity/syntaxTests/bytecode_too_large_homestead.sol
+++ b/test/libsolidity/syntaxTests/bytecode_too_large_homestead.sol
@@ -1,5 +1,3 @@
-pragma abicoder v2;
-
 contract test {
     function f() public pure returns (string memory ret) {
         // 27000 bytes long data
@@ -7,6 +5,5 @@ contract test {
     }
 }
 // ====
-// EVMVersion: >byzantium
+// EVMVersion: =homestead
 // ----
-// Warning 5574: (21-27154): Contract code size is 27199 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.


dzhalaevd:
why do we use the "namespace std"? It's bad practice :3, right?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1245,14 +1246,19 @@ void CompilerStack::assemble(
...
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
-			"This contract may not be deployable on mainnet. "
-			"Consider enabling the optimizer (with a low \"runs\" value!), "
-			"turning off revert strings, or using libraries."
+			(
+				string("Contract code size exceeds ") +
+				"24576 bytes (a limit introduced in Spurious Dragon). "
```
cameel: The part that actually prints the contract size is missing here.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1245,14 +1246,19 @@ void CompilerStack::assemble(
...
+	{
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
-			"This contract may not be deployable on mainnet. "
-			"Consider enabling the optimizer (with a low \"runs\" value!), "
-			"turning off revert strings, or using libraries."
+			(
+				string("Contract code size exceeds ") +
```
```suggestion
				"Contract code size exceeds "s +
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1245,14 +1246,19 @@ void CompilerStack::assemble(
 		m_evmVersion >= langutil::EVMVersion::spuriousDragon() &&
 		compiledContract.runtimeObject.bytecode.size() > 0x6000
 	)
+	{
```
cameel: This brace is not necessary because the `if` contains only one instruction. Please remove it.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1568,8 +1574,8 @@ class MetadataCBOREncoder
 	bytes serialise() const
 	{
 		size_t size = m_data.size() + 1;
-		solAssert(size <= 0xffff, "Metadata too large.");
-		solAssert(m_entryCount <= 0x1f, "Too many map entries.");
+		solAssert(size <= 0xffffU, "Metadata too large.");
+		solAssert(m_entryCount <= 0x1fU, "Too many map entries.");
```
cameel: This seems completely unrelated to the issue and I don't really see a point of this change. I think it should be removed from the PR.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1245,14 +1246,19 @@ void CompilerStack::assemble(
...
-			"This contract may not be deployable on mainnet. "
-			"Consider enabling the optimizer (with a low \"runs\" value!), "
-			"turning off revert strings, or using libraries."
+			(
+				string("Contract code size exceeds ") +
+				"24576 bytes (a limit introduced in Spurious Dragon). "
+				"This contract may not be deployable on mainnet. "
+				"Consider enabling the optimizer (with a low \"runs\" value!), "
+				"turning off revert strings, or using libraries."
+				).c_str()
```
cameel: I don't think calling `c_str()` is necessary here. `ErrorReporter` works with `string` just fine.

cameel (dismissed):
I think this is meant to fix #12171? Please include `Fixes #12171` in your PR description so that the issue gets closed automatically once we merge the PR.

Also, please use more meaningful descriptions for your commits, branches and the PR itself. `Contract`, `Fixed` or `Update` do not explain anything and everyone needs to guess what you mean.

Overall the PR needs some more work. In particular it does not yet solve the problem from the issue - the actual size is not included in the message.

cameel:
> why do we use the "namespace std"? It's bad practice :3, right?

Depends on who you ask :) It's a matter of style so any answer you'll get is just an opinion. In any case in this project our convention is to use it in .cpp files.

dzhalaevd:
> Overall the PR needs some more work. In particular it does not yet solve the problem from the issue - the actual size is not included in the message.

@cameel thank you for feedback!
So, the problem is that the tests give an error if I specify the actual size

Here's a warning ->
 
**Expected result:**
```
⠀⠀Warning 5574: (0-27133): Contract code size exceeds 24576 bytes (a limit introduced in Spurious 
⠀⠀Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low 
⠀⠀"runs" value!), turning off revert strings, or using libraries.
```
**Obtained result:**
```
⠀⠀Warning 5574: (0-27133): Contract code size = 27286 < 24576 bytes (a limit introduced in Spurious 
⠀⠀Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low 
⠀⠀"runs" value!), turning off revert strings, or using libraries.
```

Is that what you meant?

cameel:
You are getting an error because tests still expect the old message. You need to update the test expectations. Please run `isoltest` with the `--accept-updates` options - it will update expectations in .sol files automatically. If there are any in .cpp files (boost tests), you'll need to update these manually though.

dzhalaevd:
> You are getting an error because tests still expect the old message. You need to update the test expectations. Please run `isoltest` with the `--accept-updates` options - it will update expectations in .sol files automatically. If there are any in .cpp files (boost tests), you'll need to update these manually though.

In isoltest.cpp I see boost test, and which tests should I change manually and where? It turns out I will need to make a commit or what? It's just that I haven't worked with CI tests and with ordinary too

cameel:
> It's just that I haven't worked with CI tests and with ordinary too

Do you have a local build of the compiler and can you run tests locally? If not please take a look at [Contributing](https://docs.soliditylang.org/en/latest/contributing.html) and [Building from source](https://docs.soliditylang.org/en/latest/installing-solidity.html#building-from-source).

Once you have that, you'll find `isoltest` in `build/test/tools/`. It's a program that executes most of our tests case and if you run it with the `--accept-updates` option, it will update error messages in tests automatically. 

We run the same tests in CI on various platforms - these are the `xxx_soltest` jobs you can see on the list in this PR. If it passes for you locally it should generally also pass in CI unless you're doing something that might be platform specific.

The `chk_coding_style` is just a style check. It's currently failing because you have some whitespace at the end of one of the lines (please check the diff). I'd generally recommend to set up your editor to automatically remove trailing whitespace. This makes diffs much easier to read because it prevents you from accidentally leaving small meaningless changes all over the place.

> It turns out I will need to make a commit or what?

Well, you can just amend the existing commit. But yes, please include the changes made by `isoltest` in your commits. Otherwise they won't be seen by our CI.

> In isoltest.cpp I see boost test, and which tests should I change manually and where?

I mentioned boost tests because some messages might also be there but I suspect that just running `isoltest` might be enough here. Please try it and if some tests are still failing after that, we'll see what else need to be updated.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1244,15 +1245,16 @@ void CompilerStack::assemble(
 	if (
 		m_evmVersion >= langutil::EVMVersion::spuriousDragon() &&
 		compiledContract.runtimeObject.bytecode.size() > 0x6000
-	)
+	   )
```
cameel: This should be at the same level as the `if`, not `(`.
```suggestion
	)
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1244,15 +1245,16 @@ void CompilerStack::assemble(
...
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
+			string("Contract code size is ") +
+			to_string(compiledContract.runtimeObject.bytecode.size()) +
+			" and he exceeds  24576 bytes (a limit introduced in Spurious Dragon). "
 			"This contract may not be deployable on mainnet. "
 			"Consider enabling the optimizer (with a low \"runs\" value!), "
-			"turning off revert strings, or using libraries."
-		);
+			"turning off revert strings, or using libraries.");
```
```suggestion
			"turning off revert strings, or using libraries."
		);
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1244,15 +1245,16 @@ void CompilerStack::assemble(
...
 		compiledContract.runtimeObject.bytecode.size() > 0x6000
-	)
+	   )	
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
+			string("Contract code size is ") +
+			to_string(compiledContract.runtimeObject.bytecode.size()) +
+			" and he exceeds  24576 bytes (a limit introduced in Spurious Dragon). "
```
```suggestion
			" and exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1244,15 +1245,16 @@ void CompilerStack::assemble(
 	if (
 		m_evmVersion >= langutil::EVMVersion::spuriousDragon() &&
 		compiledContract.runtimeObject.bytecode.size() > 0x6000
-	)
+	   )	
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
+			string("Contract code size is ") +
```
cameel: This is a more compact syntax for creating a string from a literal:
```suggestion
			"Contract code size is "s +
```

dzhalaevd:
> Do you have a local build of the compiler and can you run tests locally? If not please take a look at [Contributing](https://docs.soliditylang.org/en/latest/contributing.html) and [Building from source](https://docs.soliditylang.org/en/latest/installing-solidity.html#building-from-source).
> 
> Once you have that, you'll find `isoltest` in `build/test/tools/`. It's a program that executes most of our tests case and if you run it with the `--accept-updates` option, it will update error messages in tests automatically.

I did everything as it was in the documentation and also executed the command ```isoltest.cpp --accept-updates```, but nothing has been updated

as I understood correctly, all tests are written on ```.sol``` and are located in ```test/libsolidity/syntaxtests/bytecode_too_large.sol```
that is, I can change it without doing any update. I just spent 3 hours sorting out the documentation, first I had an error with truffle, then with Cmake, etc. And I use two IDE Visual Studio 2022 and Clion (in the documentation we used Visual Studio 2019 and later versions and there was nothing written about ```CLion``` and in `CLion` I have infinite Indexing module)


`test/libsolidity/syntaxtests/bytecode_too_large.sol`
```Solidity
contract test {
    function f() public pure returns (string memory ret) {
        // 27000 bytes long data
       ret = "";
    }
}
//
// ====
// EVMVersion: >=spuriousDragon
// ----
// Warning 5574: (0-27133): Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
```
That is, we just manually rewrite the Warning to which we need?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1248,7 +1249,9 @@ void CompilerStack::assemble(
 		m_errorReporter.warning(
 			5574_error,
 			_contract.location(),
-			"Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
+			"Contract code size is "s +
+			to_string(compiledContract.runtimeObject.bytecode.size()) +
+			" and exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
```
hrkrshnn: I'll add " bytes " here.

hrkrshnn (dismissed):
I squashed the commits, slightly modified the error string and updated the tests.


@DavidRomanovizc Thanks for the PR!

hrkrshnn:
@DavidRomanovizc 

> I did everything as it was in the documentation and also executed the command isoltest.cpp --accept-updates, but nothing has been updated

Are you sure it's `isoltest` and not `isoltest.cpp`? What happens when you run just `./isoltest` inside `build/test/tools/`?

dzhalaevd:
@cameel can you merge my pull requests, please

cameel:
Not yet because CI checks fail. The expected bytecode size in the message seems to be different than in @hrkrshnn's update. I wonder if it's just a mistake or if the bytecode built locally really ends up having a slightly different size.

Another problem is that not all runs of this test return the same size. The is executed by `soltest_all.sh`, which runs it for several different EVM versions: https://github.com/ethereum/solidity/blob/3774955dcc902e870f9fcaa3927fb27015fb2619/.circleci/soltest_all.sh#L31

Looks like in the fourth run the result is different. You need to check which evm version this is and use the `EVMVersion` directive in isoltest to make the test not run on that version: https://github.com/ethereum/solidity/blob/3774955dcc902e870f9fcaa3927fb27015fb2619/test/libsolidity/syntaxTests/bytecode_too_large.sol#L8

Then also create a copy of the test for that particular version with different expectations so that this version does not go untested.

hrkrshnn:
@DavidRomanovizc Could you please remove / squash the last commits?

About the test update. The issue seems to be that the bytecode size is different for different versions of EVM and perhaps settings as well, as cameel mentioned. If you want, you could try changing `evmVersion >= spuriousDragon` to just `evmVersion = london`. Maybe changing to ` >= constantinopole` might just work (my guess is that the difference comes from the shift opcodes). I can also take a look later today, if it's difficult for you.

dzhalaevd:
@hrkrshnn, I tried to change the evm version, but the tests still failed with some other errors
I still don't understand what the problem lies. 

>Could you please remove / squash the last commits?

when I deleting commit, I get an error:

```remote: Permission to ethereum/solidity.git denied to DavidRomanovizc. fatal: unable to access```

File: test/libsolidity/syntaxTests/bytecode_too_large.sol [hunk 11]
```diff
@@ -5,6 +5,6 @@ contract test {
     }
 }
 // ====
-// EVMVersion: >=spuriousDragon
+// EVMVersion: >= constantinopole
```
chriseth: Can you try
```suggestion
// EVMVersion: =constantinopole
```
?

dzhalaevd: @chriseth, I tried it, but tests still didn't pass

chriseth:
I think we also have to restrict this to "only one codegen and only without optimizer" or something. Maybe someone from the team can take this over?

dzhalaevd:
> I think we also have to restrict this to "only one codegen and only without optimizer" or something. Maybe someone from the team can take this over?

Сan you explain in more detail, please, what exactly to restrict?

dzhalaevd:
It is normal that the number of ```bytecode_too_large.sol``` file is _27286_, the program expects _27286_, but for some reason this value changes on the next test? Please can you check the test results ```ci/circleci: t_ubu_release_soltest``` and the file itself
it just didn't pass the only test, and for some reason other inputs were submitted to the input

cameel:
I have squashed the commits and adjusted tests. I'm not sure if they will pass though. There's something weird going on - locally I get numbers completely different from the ones I've seen in the failing tests. I checked with various EVM versions, with/without `--optimize` and with/without `--abiencoderv1` and I still can't match these numbers. Well, let's see if it passes with the numbers I'm getting locally. If not, then we'll see what else we can do.

> Сan you explain in more detail, please, what exactly to restrict?

We have various directives that can be used in tests. For example one that controls whether the compilation goes through the old or the new code generator (`compileViaYul`). You'd have to have a look through various test files to see what they are.

Honestly, I didn't expect that you'd need them for this task.

> It is normal that the number of `bytecode_too_large.sol` file is 27286, the program expects 27286, but for some reason this value changes on the next test? 

As I said in https://github.com/ethereum/solidity/pull/12177#issuecomment-950284843, we use `soltest_all.sh` which does multiple test runs with different options. In particular with different EVM versions (and the `EVMVersion` directive in test just decides whether to run the test on that version or skip it) but also with/without optimization and on ABI encoder v1/v2. Changing these can affect the size of produced bytecode because it makes the generator produce slightly different output.

The size should not differ between `t_ubu_release_soltest` and `t_ubu_soltest` though so I think there might be some other factor we're not taking into account,

cameel:
Looks like in syntax tests we do not remove metadata from bytecode. Metadata contains compiler version and this differs between local builds and CI. I added a commit that disables the metadata in all syntax tests. I'm not sure we want to go that far though.

An alternative solution would be to add a new directive that would let us disable this only in specific tests.

cameel:
Looks like it passed tests now. Except for the Windows ones but these failures are unrelated - CircleCI seems to be having some problems with Windows VMs right now.

chriseth (approved):

dzhalaevd:
@cameel my PR is ready to merge? May you do make merge :)

cameel:
Well, yeah. We discussed the metadata change today and disabling it should be fine so I guess I can merge it as is.

PR #10363, opened 2020-11-20
[BREAKING] Remove the legacy AST JSON output

axic:
Fixes #9512.

This can be squashed prior to merge.

PR #12179, opened 2021-10-22
Set handlers use unified jsonPrint with prettifier if needed

akali:
Partially addresses #12011 (all except for `--metadata`).


cameel (commented):
Thanks for the PR! This looks good overall but tests in CI are failing. You need to update tests so that the output matches expectations.
- in `t_ubu_cli` the test named `ast_compact_json_with_base_path` is failing because `--ast-compact-json` used to pretty-print the output by default. Now it no longer does. You need to add `--pretty-print` to its `args` file to perserve the current output. There might be more tests that need a similar change. You can run the CLI tests locally with `commandLineTests.sh`
- Lots of test cases fail in `t_ubu_soltest`. Same reason as above. To fix them you need to modify the test case runners to pass the right options to the functions that generate JSON. For example this needs to be done in `ABIJsonTest`: https://github.com/ethereum/solidity/blob/ef21e43fa36be8ff58725133aa5bb6e74bd06baf/test/libsolidity/ABIJsonTest.cpp#L64
     This is of course not the only place.
- Please add new CLI tests that cover both pretty-printed and normal output for these options so that we can see that the change works. Take a look at #11639 if you want to see how to add such tests.
- Finally, since this does not cover `--metadata`, please remove `Closes #12011` from PR description. We don't want to close the issue before all options are addressed. But doing the metadata separately is definitely a good idea since this one might require some discussion in the team while the other options should not be controversial at all.

cameel:
Do you need any help with adding/updating these tests?

dallonasnes:
I'd be happy to update the tests if that's still blocking merge

cameel:
Sure. Looks like @akali is not working on it so it's fine to take over.

dallonasnes:
Is there a way to disable tests that require SMT/Horn solver in `test/cmdLineTests.sh`?

cameel:
It supports the `--no-smt` option just like `soltest`/`isoltest`.
You can also give it a pattern with shell wildcards to run only the tests with matching names.

dallonasnes:
How do you input the pattern? Here's what I'm trying to run, but it doesn't like the arg format
```
./cmdlineTests.sh ast_compact_json_with_base_path
```

cameel:
That should work if it's the full name of the test dir. If not, you have to add some wildcards:
```bash
./cmdlineTests.sh "*ast_compact_json_with_base_path*"
```

dallonasnes:
```
test git:(fix-tests) $ ./cmdlineTests.sh "*ast_compact_json_with_base_path*"
cut: illegal option -- -
usage: cut -b list [-n] [file ...]
       cut -c list [file ...]
       cut -f list [-s] [-d delim] [file ...]

**Comment by cameel** [comment]
Looks like you're having a problem with this line:
https://github.com/ethereum/solidity/blob/e0c85c6f58310e61dbc6f153b6228e6240e63f8a/test/cmdlineTests.sh#L57

Are you by chance on macOS? It's probably that `cut` on your platform supports only short options. Replacing `cut --characters` with `cut -c` should fix that. If that works for you, please submit a PR with the change.

This probably slipped though because we never really use this feature in CI - we always run all tests - and locally it worked for me on Linux so I never noticed.

**Comment by dallonasnes** [comment]
Yup I'm on macOS. I'll try that out, thanks

**Comment by dallonasnes** [comment]
The -c fix worked for me. PR #12348

**Comment by cameel** [comment]
@dallonasnes Are you still interested in working on this?

**Comment by dallonasnes** [comment]
Hi, better if someone else takes it. For tests to pass, it'll be small changes in many dozens of files, hadn't realized it was that broad of a change.

**Comment by cameel** [comment]
@dallonasnes Test input must be updated, yes, but if you mean test output/expectations, we of course have auto-update mode so it does not have to be done manually. I.e. there's the `--accept-updates` option for `isoltest` and `--update` for `cmdlineTests.sh`.

**Comment by wechman** [comment]
I take over this PR.

**Comment by wechman** [comment]
> @dallonasnes Test input must be updated, yes, but if you mean test output/expectations, we of course have auto-update mode so it does not have to be done manually. I.e. there's the `--accept-updates` option for `isoltest` and `--update` for `cmdlineTests.sh`.

@cameel, I guess this is not the best way to go. We want to keep pretty JSONs in the test expectations and modify only the test runners, right? Please check a fixup commit I delivered and let me know it is fine.

**Comment by cameel** [comment]
@wechman Ah, you're right! Looks like we were already pretty-printing by default in all cases so we can actually keep the expectations and just adjust the options. I misremembered that and thought that we needed to update expectation in some cases too but apparently not.

This still needs some tests for the compact version though.

**Comment by wechman** [comment]
@cameel, I added compact JSON output tests to the fixup commit.  But, I noticed surprising result in ast_pretty_json_with_base_path test.  There is pretty-printed JSON with following first line: "JSON AST (compact format):" in the test output. I am not sure if I the information about "compact format" is required to be part of the output, is it? Unless, I should not mix "--ast-compact-json" with "--pretty-json" in the test scenarios?

**Comment by cameel** [comment]
The `compact` part in `--ast-compact-json` is not related to pretty-printing. There used to be an option called `--ast-json` and it was removed in #10363. The new format is just a bit more dense. It's unfortunate that it's now easily confused with compact printing.

Maybe we could remove the "compact format" part from the header, though changing the output is technically a breaking change. Personally, I'd just leave it as is, it does correspond to the option name.

Anyway, the output of `--ast-compact-json` should be affected by `--pretty-print` just like the other options.

File: test/cmdlineTests/ast_compact_json_with_base_path/output [hunk 1]
```diff
@@ -2,92 +2,6 @@ JSON AST (compact format):
...
-      "nodeType": "ContractDefinition",
-      "nodes": [],
-      "scope": 6,
-      "src": "60:13:0",
-      "usedErrors": []
-    }
-  ],
-  "src": "36:38:0"
-}
+{"absolutePath":"ast_compact_json_with_base_path/c.sol","exportedSymbols":{"C":[5]},"id":6,"license":"GPL-3.0","nodeType":"SourceUnit","nodes":[{"id":4,"literals":["solidity",">=","0.0"],"nodeType":"PragmaDirective","src":"36:22:0"},{"abstract":false,"baseContracts":[],"canonicalName":"C","contractDependencies":[],"contractKind":"contract","fullyImplemented":true,"id":5,"linearizedBaseContracts":[5],"name":"C","nameLocation":"69:1:0","nodeType":"ContractDefinition","nodes":[],"scope":6,"src":"60:13:0","usedErrors":[]}],"src":"36:38:0"}
```
cameel: Like I said in my comment, this was working correctly. You should be able to decide if you want the output of `--ast-compact-json` pretty-printed or not.

test/cmdlineTests/ast_pretty_json_with_base_path/args
```diff
@@ -0,0 +1 @@
+--ast-compact-json --pretty-json --base-path . --allow-paths .
```
cameel: To make test names consistent and easier to understand when you know the option names, I'd make them somewhat correspond to these names:
- I'd use `ast_compact_json_with_base_path` for this one because it was primarily meant to test it with `--base-path`. Pretty printing is in the default state and I think it's fine not to create a separate test for that.
- I'd use `ast_compact_json_no_pretty_json` for the non-pretty-printing test because we want it specifically to test that feature. Base path is not relevant here so I'd remove it from the name and `--base-path` from the command line.

test/cmdlineTests/compact_json_combined/args
```diff
@@ -0,0 +1 @@
+--combined-json abi
```
cameel: Let's call this one `combined_json_no_pretty_print`. That's how I'd expect the name to look like if I were looking for tests covering `--combined-json`.

test/cmdlineTests/compact_json_standard/input.json
```diff
@@ -0,0 +1,10 @@
+{
```
cameel: We usually prefix tests for Standard JSON with `standard_`. I'd use `standard_no_pretty_print` here.

solc/CommandLineInterface.cpp
```diff
@@ -317,7 +317,7 @@ void CommandLineInterface::handleABI(string const& _contract)
 	if (!m_options.compiler.outputs.abi)
 		return;
 
-	string data = jsonCompactPrint(removeNullMembers(m_compiler->contractABI(_contract)));
+	string data = jsonPrint(removeNullMembers(m_compiler->contractABI(_contract)), m_options.formatting.json);
```
cameel, line 320: Can you make sure that we have tests covering both pretty printing and not pretty printing for all of these?
- `--asm-json`
- `--abi`
- `--userdoc`
- `--devdoc`
- `--storage-layout`

This PR does not touch any tests covering them so I think they're either covered only in one of these cases or even not covered at all. It's enough to cover only the non-default case just to make sure it works.

cameel (dismissed):
I added some suggestions about test naming.

The remaining issues are that `--pretty-print` is meant to affect `--ast-compact-json` (so that needs to be changed back) and that some of the options covered by the PR appear not to be fully covered.

solc/CommandLineInterface.cpp
wechman: Good point, especially since scenario with "--asm-json" showed up change that was still missing.

test/cmdlineTests/ast_compact_json_with_base_path/output
wechman: ok

wechman:
> The `compact` part in `--ast-compact-json` is not related to pretty-printing. There used to be an option called `--ast-json` and it was removed in #10363. The new format is just a bit more dense. It's unfortunate that it's now easily confused with compact printing.
> 
> Maybe we could remove the "compact format" part from the header, though changing the output is technically a breaking change. Personally, I'd just leave it as is, it does correspond to the option name.
> 
> Anyway, the output of `--ast-compact-json` should be affected by `--pretty-print` just like the other options.

Yes, it is easy to confuse compact JSON with compact printing.  But, with your explanation, it all makes sense. Thanks for that!

About header, I agree we can keep it as it is now.

test/cmdlineTests/ast_compact_json_no_pretty_json/input.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+
+import "./c.sol";
```
cameel: I'd remove `c.sol` to simplify the test. It was relevant in the test with base path but for pretty printing we don't need it.

test/cmdlineTests/combined_json_with_devdoc/c.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+
+contract C {}
```
cameel: I'd just put this directly in `input.sol` to simplify the test.

Also, would be nice to have some actual dev doc here. For example:
```suggestion
/// @dev This is devdoc.
contract C {}

/// And this is a notice.
contract D {}
```

test/cmdlineTests/combined_json_with_userdoc/c.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+
+contract C {}
```
cameel: Here again, I'd remove `c.sol` and put the contract in `input.sol`. Then add some natspec:
```suggestion
/// @notice Description for users.
contract C {}

/// @dev Description for developers.
contract D {}
```

test/cmdlineTests/storage_layout_user_defined_no_pretty_print/err
```diff
@@ -0,0 +1,2 @@
+Warning: Source file does not specify required compiler version!
+--> storage_layout_user_defined_no_pretty_print/input.sol
```
cameel: Just add the version pragma to `input.sol` and we can get rid of this file.

test/cmdlineTests/storage_layout_user_defined_no_pretty_print/input.sol
```diff
@@ -0,0 +1,16 @@
...
+    MyInt128 b;
+    // slot 1
+    MyInt128 c;
+    MyInt8 d;
+    MyInt8 e;
+    MyInt8 f;
+    MyInt8 g;
+    // slot 2
+    MyInt8 h;
+}
```
cameel: This test is just for JSON printing so it can be very simple:
```suggestion
contract C {}
```
We only really care that it prints the output from `--storage-layout` and that that output is not prettified so no point having much more here.

cameel (commented):
In terms of functionality all is in order here :)
I have some simplification suggestions and after that it's good to merge.

wechman:
@cameel Thank you all your comments! I updated code accordingly. I also squashed commits and rebased them on the newest develop.

cameel (dismissed):
Looks good.

I just noticed that it's missing a changelog entry though. This is a change visible to users.

wechman:
@cameel Changelog updated.

Changelog.md
```diff
@@ -5,6 +5,7 @@ Language Features:
 
 Compiler Features:
  * Peephole Optimizer: Remove operations without side effects before simple terminations.
+ * Commandline Interface: option ``--pretty-json`` works also with following options: ``--abi``, ``--asm-json``, ``--ast-compact-json``, ``--devdoc``, ``--metadata``, ``--storage-layout``, ``--userdoc``.
```
cameel: ``--metadata`` is not affected, is it?

```suggestion
 * Commandline Interface: option ``--pretty-json`` works also with the following options: ``--abi``, ``--asm-json``, ``--ast-compact-json``, ``--devdoc``, ``--storage-layout``, ``--userdoc``.
```

wechman: good point, it is not

cameel (dismissed):

leonardoalt:
@wechman not sure if you just rebased it, but cli tests failing

cameel:
Yeah, needs a rebase and updating CLI test expectations but other than that it's done.

wechman:
@leonardoalt @cameel I have just rebased. Also, I have updated expectations in one command line test case.

cameel (approved):

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12180, opened 2021-10-22
Count backwards instead of forwards

SiliconA-Z:


cameel:
I'm sorry but I don't think this is a meaningful change. There was nothing wrong about that line before it was rewritten and the change also makes it use `auto`, which is completely unnecessary because the type is simple.

PR #12181, opened 2021-10-22
[CI] Automate preparation of release binaries for publishing (via artifacts)

### LINKED ISSUE #9933: Automate publishing of release binaries
Extracted from #9258. Best implemented after we're done with #9932.

Currently release binaries are built on Travis and CircleCI and then manually uploaded to github release page and `solc-bin`. We probably don't want to automate it all the way through - we need a human in the loop to be able to verify that everything is ok and possibly intervene - but parts of this process could still be streamlined. For example the creation of a draft release and submission of a PR with binaries to `solc-bin`.

It will be much easier if we decide to move the builds to Github actions in #9932. The downside of doing this way is that it would not be exactly the same build that is extensively tested on CircleCI.

This was extensively discussed with @ekpyron and @chriseth in #9258 and on Gitter but we did not come to a firm decision yet. We only decided that it's not a priority and can wait until we're done with more important tasks related to the binary release process overhaul.

cameel:
Closes #9933.
~Depends on #12173 (draft until it's merged).~ Merged.

It's not complete automation of the publishing process as stated in #9933 but it at least takes care of the most tedious and error prone part of the process so I think it's good enough.

The PR adds:
- A script for downloading artifacts representing the binaries we use in releases.
    - Surprisingly there's no command for easy downloading of artifacts inside a CircleCI job and it requires using their API (see [Access uploaded artifact URL in job](https://support.circleci.com/hc/en-us/articles/360045457592-Access-uploaded-artifact-URL-in-job)). Fortunately I already had a script for doing something similar (https://github.com/ethereum/solc-bin/pull/41) so I repurposed it for this.
- The `c_release_binaries` job that uses the script to gather artifacts from jobs that build release binaries, puts them in two directories (one for github release, one for solc-bin) and names files in each directory according to our naming conventions. Then attaches the results as its own artifacts.
- A step that posts a notification to gitter when it succeeds.
- ~A small fix for missing `*workflow_trigger_on_tags` in some of the existing jobs.~ Merged separately.

The new job is supposed to run only on tagged commits. Note that I can only really test it on a PR. On a tagged commit some things will be different (CircleCI env vars, version string, etc.). We'll have to test that in practice once the PR is merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1401,6 +1401,32 @@ jobs:
           path: all-bytecode-reports.zip
       - gitter_notify_failure_unless_pr
 
+  b_release_binaries:
+    <<: *base_ubuntu2004
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: |
+            apt-get update
+            apt-get install --assume-yes --no-install-recommends jq
+      - run:
+          name: Download artifacts of dependent jobs
+          command: |
+            mkdir github/
+            cd github/
+            ../scripts/solc-bin/download-circleci-binaries.sh "$CIRCLE_WORKFLOW_ID"
+      - store_artifacts:
+          path: github/
+      - run:
+          name: Rename binaries to solc-bin naming convention
+          command: |
+            mv github/ solc-bin/
+            cd solc-bin/
+            ../scripts/solc-bin/rename-circleci-binaries-for-solc-bin.sh
+      - store_artifacts:
+          path: solc-bin/
+
 workflows:
   version: 2
 
@@ -1518,6 +1544,20 @@ workflows:
             - b_bytecode_win
             - b_bytecode_osx
             - b_bytecode_ems
+
+      # Final artifacts
+      - b_release_binaries:
+          filters:
+            tags:
+              # We only need release binaries from tagged commits
+              only: /^v.*/
+            branches:
+              ignore: /.*/
+          requires:
+            - b_ubu_static
+            - b_osx
+            - b_win_release
+            - b_ems
 
   nightly:
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ commands:
     parameters:
       event:
         type: enum
-        enum: ["failure", "success"]
+        enum: ["failure", "success", "release"]
       condition:
         type: string
     steps:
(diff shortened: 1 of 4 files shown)


cameel:
This passed CI. Here's a successful run of [`b_release_binaries`](https://app.circleci.com/pipelines/github/ethereum/solidity/19930/workflows/f63cdca9-3ae7-4420-873d-30f4c59eca1c/jobs/880945) and [a message posted to gitter](https://gitter.im/ethereum/solidity-dev?at=6172efabfb8ca0572bd8bc69) (slightly misformatted because the variable containing tag name is not defined on a PR).

Now I'm removing the temporary commits - the job will only run on tags from now on.

cameel:
This is now reviewable too.

.circleci/config.yml
```diff
@@ -1113,6 +1124,33 @@ jobs:
...
+          name: Install dependencies
+          command: |
+            apt-get update
+            apt-get install --assume-yes --no-install-recommends jq
+      - run:
+          name: Download artifacts of dependent jobs
+          command: |
+            mkdir github/
+            cd github/
+            ../scripts/solc-bin/download-circleci-binaries.sh "$CIRCLE_WORKFLOW_ID"
```
ekpyron, line 1429: Wouldn't it be easier, cleaner and safer to rename the binaries in the respective builds already and attach them to a common workspace similar to what we do for the bytecode run?

leonardoalt:
Waiting for answer from @cameel

.circleci/config.yml
cameel: Not necessarily easier but might be doable. The biggest issue is that I need to get all binaries in the workspace but under different paths. You get an error if a job depends on multiple jobs adding a workspace file under the same path. Doing it via artifacts does not have this problem.

When I started this PR I actually did not understand our workspace layout all that well. It's often mounted at different locations in similar jobs and not in a consistent way. It was really confusing to me and I tried to clean that up in #12450. Unfortunately this resulted in a conflict between binaries on Linux and macOS now that the paths are consistent.

But after working on it I think that the current gnarly layout would work without a conflict. So yeah, I'm going to try to redo it like that.

Another option I was thinking of was rewriting the Bash script here in Python. In #12818 I already have a Python module with helpers for CircleCI and Github API so this would replace the big script in this PR with something much smaller. But if the workspace solution works, maybe it's better.

cameel:
Answered. Basically, I need to decide which way to go with this and update it one way (workspace) or another (Python-based script).

.circleci/config.yml
ekpyron: Yeah, I know the problem with path conflicts in merging workspaces... but I thought since we have the same problem with the bytecode comparison runs, it may just work or at least not be too hard to achieve, but not entirely sure...

cameel: Yeah, you must be right. `t_bytecode_compare` does attach the workspace so it should work. It's just that I did not realize that at first and thought that trying to do it via workspace would be very messy and require renaming binaries. I only see now that it shouldn't :)

cameel: We should really clean up the workspace layout though. Once we do this, we **will** have a conflict. My current idea to sidestep that is to name each binary after the platform and use symlinks in #12450.

cameel:
Closing in favor of #12929.

PR #12182, opened 2021-10-22
[CI] Reusable command for gitter notifications

cameel:
This is a simple refactor, extracted into a separate PR because I need it in more than one other PR.

It converts the existing notification step templates into a parameterized and reusable command. This reduces duplication and makes it easier to add new notifications (which is needed in #12181).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,8 +43,8 @@ commands:
       - run:
           name: "Gitter notification"
           command: |
-            [[ "<< parameters.event >>" == "failure" ]] && message=" ❌ Nightly job **${CIRCLE_JOB}** failed on **${CIRCLE_BRANCH}**. Please see ${CIRCLE_BUILD_URL} for details."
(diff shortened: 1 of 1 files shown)


hrkrshnn:
@cameel Was this tested by any chance? Looks good. Can approve if this was tested. We can also test in prod :)

Marenz:
Yeah, I think the easiest is to test it in production, as it explicitly excludes PRs. Though of course you could add a test-commit that removes that exception so it runs on PRs as well.

Marenz (approved):
LGTM. I am okay with production testing, but a simple commit that removes the PR exclusion might also do the trick to test it

cameel:
@hrkrshnn I did test it but not very extensively (I didn't want to spam the channel and also the variables are going to be different in non-PR runs anyway so it's not foolproof). I tested it together with #12181. Take a look at https://github.com/ethereum/solidity/pull/12181#issuecomment-949876968.

cameel:
@Marenz I did that in #12181 (it was originally a part of that PR) and when it worked I removed the debug commits.

hrkrshnn (approved):

PR #12184, opened 2021-10-24
Fix typo compiler->constructor.

abliss:
I think this might be an error in the docs? If not, maybe some more details or examples should be added, because I don't understand what external state the solidity *compiler* might be accessing that's relevant here.


hrkrshnn (approved):
Thanks @abliss

Issue #10745, opened 2021-01-12
OpenZeppelin external tests should be using Buidler/Hardhat instead of Truffle


cameel:
Continuing the discussion from #10672.

OpenZeppelin uses Buidler as its build system while our `test/externalTests/zeppelin.sh` attempts to use Truffle to build it. It appears to work but only because the whole thing does not fail too easily and has multiple failsafes where default things get used/downloaded when not available:
- We launch Truffle via `npx truffle compile`. `npx` just installs it if it's not available instead of failing.
- `prepare` command in OpenZeppelin's `package.json` compiles the contracts using Buidler anyway.
- The Truffle config that `test/externalTests/common.sh` creates is completely ignored. Tests run on Buidler instead (which is good) but they're just using the compiler binary it downloads (rather the one built in the CI).

`test/externalTests/zeppelin.sh` / `test/externalTests/common.sh` need to be fixed to stop trying to use Truffle for this test and to make Buidler use the right binary.

cameel:
It's now possible to use a custom binary with Hardhat: https://github.com/fvictorio/hardhat-examples/blob/master/custom-solc/README.md

PR #10556, opened 2020-12-09
README describing the workflow around external tests and their repositories

cameel:
Somewhat related to #10429 and all the chaos around updating branches in `solidity-external-tests`.

I decided to write down the recommended workflow since it's not all that obvious what can/should/must be done when updating these tests. Different people do it slightly differently and this way sometimes get in the way of each other. The current text describes how I think it should be done but it's also an opportunity to discuss it. I'll apply any changes that result from the discussion.

I'm not sure if a README in an obscure directory is the best place for it but I think it's not relevant enough to end-users to put it in the main docs. Someone tasked with updating these tests will probably manage to find it.

PR #12187, opened 2021-10-25
Run external tests directly on upstream repo where possible

cameel:
In preparation for #10745 I'm updating external tests to run on the latest upstream code.

Where possible this actually switches the tests to use upstream directly rather than our forks. We do not update them very often and @ekpyron in the past mentioned that maintaining forks is not necessarily preferable (https://github.com/ethereum/solidity/pull/10556#discussion_r540174098). I think that this will give us better feedback though it could also result in more frequent breakage unfortunately.

- OpenZeppelin: already on 0.8.x. Can be used directly.
- ENS: Can be used directly.
- Colony: still on 0.7.x. Requires changes from our fork. I rebased them on the latest `develop` though ([updated `develop_080` branch](https://github.com/solidity-external-tests/colonyNetwork/tree/develop_080)).
- Gnosis: looks like v2 is now the main branch and also it has been upgraded to 0.7.x. They switched from Truffle to Hardhat though so the update will be a bit more complex. I'll do it in a separate PR.

### DIFF
--- a/test/externalTests/README.md
+++ b/test/externalTests/README.md
@@ -13,16 +13,17 @@ these projects *can* be upgraded at all.
 ### Recommended workflow
 
 #### Adding a new external project
-1. Create a fork of the upstream repository in https://github.com/solidity-external-tests/. If the
-    project consists of several repositories, fork them all.
-2. Remove all the branches except for main one (`master`, `develop`, `main`, etc). This branch is
-    going to be always kept up to date with the upstream repository and should not contain any extra
-    commits.
+1. If the upstream code cannot be compiled without modifications, create a fork of the repository
+   in https://github.com/solidity-external-tests/.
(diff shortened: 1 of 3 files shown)


cameel:
This should now pass all tests and is ready for review.

cameel:
Looks like [Colony external tests actually fail](https://app.circleci.com/pipelines/github/ethereum/solidity/20015/workflows/b4999e7a-3843-4e9f-a6c9-31e891c5162e/jobs/884228), though it's not visible here because we only run them nightly. They run out of gas during a migration. The limit is by default at ~6 mln but even increasing that to 100 mln does not help.

I'm going to revert the `develop_080` branch to the previous position and work on a fix separately.

In any case, this does not really affect this PR because it contains no Colony-related changes. They were only in our forked repos.

leonardoalt (approved):
Nice!

PR #12191, opened 2021-10-26
Make external test scripts more flexible

cameel:
Another refactor in preparation for fixing #10745.
~Depends on #12173 (draft until that PR is merged).~ Merged.
Depends on #12259.

Currently the external test scripts are all made with Truffle in mind and are a bit hard to customize both if they need some special steps and when they use a different framework. This PR "unrolls" the top-level functions and puts their content directly in each test script so that it can be adjusted more easily. It also does small general refactors:
- Adds `set -e` to the scripts so that failures interrupt them.
- Changes arguments of some functions or adds defaults so that the calls are a bit less verbose.
- Removes some global variables in favor of locals+parameters.

### DIFF
--- a/test/externalTests/colony.sh
+++ b/test/externalTests/colony.sh
@@ -41,8 +41,7 @@ function colony_test
     setup_solcjs "$DIR" "$SOLJSON"
     download_project "$repo" "$branch" "$DIR"
 
-    replace_version_pragmas
-    force_truffle_solc_modules "$SOLJSON"
+    neutralize_package_json_hooks
     force_truffle_compiler_settings "$config_file" "${DIR}/solc" "$min_optimizer_level"
     yarn
     git submodule update --init
--- a/test/externalTests/colony.sh
+++ b/test/externalTests/colony.sh
@@ -52,7 +52,7 @@ function colony_test
     cd ..
 
     replace_version_pragmas
-    force_truffle_solc_modules "$SOLJSON"
+    force_solc_modules "${DIR}/solc"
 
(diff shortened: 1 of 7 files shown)


cameel:
Added two more commits from work on Gnosis external tests. Now the scripts just neutralize any `prepublish`/`prepare` hooks instead of doing gymnastics to make `npm install` succeed trying to compile contracts. As a nice side effect this should make them slightly faster - less stuff need to be repeated twice.

cameel:
This now depends on #12173. The PRs themselves are completely independent but I have another PR (#12214) that needs both.

cameel:
#12173 has been merged so this is now reviewable.

scripts/ci/docker_upgrade.sh
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 set -e
 
-function error() {
+function error {
```
axic: The others in this commit put the `{` on a new line.

axic:
I think the `set -e` and function syntax change could be just merged separately, as is, as that doesn't require a lot of review.

Reviewed most of it, except the two major commits (fda1ab5ce1af068d308d62478d0da47b6f1694db and be6a85be14fb9fd644eafe1b8f90649b7ce8be69) which require more thorough look.

cameel:
Extracted the minor refactors into a separate PR: #12259. This PR now depends on it.

ekpyron (approved):
I won't claim that I can be sure that it all works, but the changes do look reasonable and I trust @cameel has tested it, so I'll go on a leap and approve this.

PR #12192, opened 2021-10-26
Use Hardhat and the right binary for OpenZeppelin external tests

### LINKED ISSUE #10745: OpenZeppelin external tests should be using Buidler/Hardhat instead of Truffle

cameel:
Fixes #10745.
~Depends on #12191 (draft until that PR is merged).~ Merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1223,7 +1223,6 @@ workflows:
           name: t_ems_compile_ext_zeppelin
           project: zeppelin
           compile_only: 1
-          nodejs_version: '14'
       - t_ems_ext:
           <<: *workflow_emscripten
           name: t_ems_compile_ext_ens
@@ -1250,7 +1249,8 @@ workflows:
           <<: *workflow_emscripten
           name: t_ems_test_ext_zeppelin
           project: zeppelin
-          nodejs_version: '14'
+          # NOTE: Tests crash on nodejs 17: "Error: error:0308010C:digital envelope routines::unsupported"
+          nodejs_version: '16'
       - t_ems_ext:
           <<: *workflow_emscripten
           name: t_ems_test_ext_ens
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -946,8 +946,8 @@ jobs:
         type: integer
         default: 0
       nodejs_version:
-        type: integer
-        default: 14
+        type: string
+        default: latest
     docker:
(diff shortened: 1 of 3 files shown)


leonardoalt:
Do we still need this PR after https://github.com/ethereum/solidity/pull/12187?
Why is this PR needed?

cameel:
Yeah, it's still relevant. It's a separate thing. The issue is that OZ has hardhat config but we're trying (and unfortunately succeeding) to run truffle - which then executes tests with the wrong binary.

The URL change overlaps with #12187 and this PR could really be on top of it but I'm trying to reduce long dependency chains between PRs as much as possible so they can be reviewed independently.

I need the URL change in both because I'm not sure if the HH version used in our external fork already supports using a custom binary (most likely not). The upstream one definitely does.

cameel:
Dependencies merged. This is now ready for review.

.circleci/config.yml
```diff
@@ -939,8 +939,8 @@ jobs:
         type: integer
         default: 0
       nodejs_version:
-        type: integer
-        default: 14
+        type: string
+        default: latest
```
axic, line 950: Which other task uses this with default?

cameel: In this PR only `t_ems_compile_ext_zeppelin`. But I have a bunch of other PRs updating external tests. Eventually most of the `t_ems_compile_XYZ` jobs will be using the default because they work fine on nodejs 17. And I would use it for `t_ems_test_XYZ` jobs too but most of them now use Hardhat and it does not support node 17 yet (https://github.com/nomiclabs/hardhat/issues/1988).

chriseth (approved):

PR #12193, opened 2021-10-26
Fix crash when using @use-src while compiling yul -> ewasm

### LINKED ISSUE #12065: Crash when `@use-src` is used with `--machine ewasm` in assembly mode
## Description
`@use-src` annotations cause a crash in `AsmPrinter::formatSourceLocation()` when Yul->Ewasm translation is requested.

## Steps to Reproduce
- `input.yul`
    ```yul
    /// @use-src 0:"test.sol"
    object "C" {
        code {}
    }
    ```
```bash
solc input.yul --strict-assembly --yul-dialect evm --machine ewasm
```
```
======= input.yul (Ewasm) =======

Pretty printed source:
/// @use-src 0:"test.sol"
object "C" {
    code { }
}


==========================

Translated source:
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

The assertion failure happens at https://github.com/ethereum/solidity/blob/0e7e936fe6d49afeec45362cbd228b635833d983/libyul/AsmPrinter.cpp#L273

## Environment
- Compiler version: Happens both on 0.8.8/0.8.9 and on the latest code from `develop`.

Marenz:
fixes #12065

I am not really sure this is the correct way to fix this, but at least all tests run..

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to ewasm.
 
 
 ### 0.8.10 (2021-11-09)
--- a/libyul/backends/wasm/EVMToEwasmTranslator.cpp
+++ b/libyul/backends/wasm/EVMToEwasmTranslator.cpp
@@ -137,7 +137,11 @@ void EVMToEwasmTranslator::parsePolyfill()
 			string(solidity::yul::wasm::polyfill::Logical) +
 			string(solidity::yul::wasm::polyfill::Memory) +
 		"}", "");
-	m_polyfill = Parser(errorReporter, WasmDialect::instance()).parse(charStream);
+
+	// Passing an empty SourceLocation() here is a workaround to prevent a crash
+	// when compiling from yul->ewasm. We're stripping nativeLocation and
+	// originLocation from the AST (but we only really need to strip nativeLocation)
+	m_polyfill = Parser(errorReporter, WasmDialect::instance(), langutil::SourceLocation()).parse(charStream);
 	if (!errors.empty())
 	{
 		string message;
--- a/test/cmdlineTests/yul_to_wasm_source_location_crash/args
+++ b/test/cmdlineTests/yul_to_wasm_source_location_crash/args
@@ -1,0 +1,1 @@
+--strict-assembly --yul-dialect evm --machine ewasm --optimize --ewasm-ir
--- a/test/cmdlineTests/yul_to_wasm_source_location_crash/output
+++ b/test/cmdlineTests/yul_to_wasm_source_location_crash/output
@@ -1,0 +1,34 @@
+
+======= yul_to_wasm_source_location_crash/input.yul (Ewasm) =======
+
+==========================
+
+Translated source:
+/// @use-src 0:"test.sol"
+object "C" {
+    code {
+        function main()
+        {
+            let hi := i64.shl(i64.extend_i32_u(bswap32(i32.wrap_i64(0))), 32)
+            let y := i64.or(hi, i64.extend_i32_u(bswap32(i32.wrap_i64(i64.shr_u(0, 32)))))
+            i64.store(0:i32, y)
+            i64.store(i32.add(0:i32, 8:i32), y)
+            i64.store(i32.add(0:i32, 16:i32), y)
+            i64.store(i32.add(0:i32, 24:i32), y)
+            i64.store(32:i32, y)
+            i64.store(i32.add(32:i32, 8:i32), y)
+            i64.store(i32.add(32:i32, 16:i32), y)
+            i64.store(i32.add(32:i32, 24:i32), y)
+            eth.storageStore(0:i32, 32:i32)
+        }
+        function bswap16(x:i32) -> y:i32
+        {
+            y := i32.or(i32.and(i32.shl(x, 8:i32), 0xff00:i32), i32.and(i32.shr_u(x, 8:i32), 0xff:i32))
+        }
+        function bswap32(x:i32) -> y:i32
+        {
+            let hi:i32 := i32.shl(bswap16(x), 16:i32)
+            y := i32.or(hi, bswap16(i32.shr_u(x, 16:i32)))
+        }
+    }
+}
--- a/test/cmdlineTests/yul_to_wasm_source_location_crash/err
+++ b/test/cmdlineTests/yul_to_wasm_source_location_crash/err
@@ -1,0 +1,1 @@
+Warning: Yul is still experimental. Please use the output with care.
--- a/test/cmdlineTests/yul_to_wasm_source_location_crash/input.yul
+++ b/test/cmdlineTests/yul_to_wasm_source_location_crash/input.yul
@@ -1,0 +1,4 @@
+/// @use-src 0:"test.sol"
+object "C" {
+    code { sstore(0,0) }
+}


libyul/AsmPrinter.cpp
```diff
@@ -274,7 +274,7 @@ string AsmPrinter::formatSourceLocation(
 
 	string sourceIndex = "-1";
 	string solidityCodeSnippet = "";
-	if (_location.sourceName)
+	if (_location.sourceName && !_location.sourceName->empty())
```
Marenz: I traced back that the sourceName is `""` and added like that in the parser itself.. not sure if I should fix that rather than adding the check here..

hrkrshnn: Do you have a stack trace?

Marenz: I'd need to reproduce it now, but the problem was basically that sourceName is `""` and that in `_nameToSourceIndex` only `test.sol` is defined, so the `at()` in line 279 throws.

Changelog.md
```diff
@@ -15,16 +15,17 @@ Compiler Features:
 
 
 Bugfixes:
+ * Code Generator: Fix a crash when using ``@use-scr`` and compiling from yul to ewasm
  * Code Generator: Fix constructor source mappings for immutables.
+ * Commandline Interface: Disallow ``--error-recovery`` option outside of the compiler mode.
```
chriseth: Please sort the changelog in its own pr

libyul/AsmPrinter.cpp
chriseth: I would say empty source name should not be a problem. The question is why `_nameToSourceIndex` is not properly filled.

On the other hand: If this is more complicated, I would not spend too much time on it since we are not doing any wasm-related stuff currently.

Marenz: I think it is correctly filled with the file name specified in `@use-src`.. or should be the behavior be different here?

Marenz: Or did you mean that `""` should be in the _nameToSourceIndex? In what situation would it contain an empty file name?

Changelog.md
```diff
@@ -15,6 +15,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Code Generator: Fix a crash when using ``@use-scr`` and compiling from yul to ewasm
```
```suggestion
 * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to Ewasm
```

libyul/AsmPrinter.cpp
chriseth: I'm just saying that the empty file name is just as valid as any other file name and should not be treated differently. Not sure if this actually applies here.

test/cmdlineTests/yul_to_wasm_source_location_crash/output
```diff
@@ -0,0 +1,5190 @@
+
```
chriseth, line 1: Please not.

leonardoalt:
Tests are failing.

libyul/backends/wasm/EVMToEwasmTranslator.cpp
```diff
@@ -137,7 +137,7 @@ void EVMToEwasmTranslator::parsePolyfill()
 			string(solidity::yul::wasm::polyfill::Logical) +
 			string(solidity::yul::wasm::polyfill::Memory) +
 		"}", "");
-	m_polyfill = Parser(errorReporter, WasmDialect::instance()).parse(charStream);
+	m_polyfill = Parser(errorReporter, WasmDialect::instance(), langutil::SourceLocation()).parse(charStream);
```
axic, line 144: Hmm, why do we have `Parser` with a default value (without the explicit arg) if it causes crashes?

chriseth: It only causes a crash because we use it on a source file that is not known to the later asm printer.
The default value means "use the source locations as they are reported by the scanner". What we do here is combining regular source locations with the ones reported from the parser here, which are pointing to an on-the-fly generated source file.

test/cmdlineTests/yul_to_wasm_source_location_crash/output
```diff
@@ -0,0 +1,5190 @@
...
+object "C" {
+    code { }
+}
+
+
+==========================
+
+Translated source:
+/// @use-src 0:"test.sol"
+object "C" {
```
ekpyron, line 8: What a monster :-).

Maybe use

```
/// @use-src 0:"test.sol"
object "C" {
    code { sstore(0,0) }
}
```

as test case instead and run it with
``--strict-assembly --yul-dialect evm --machine ewasm --optimize``

:-)?

That also triggered the bug and is fixed with the PR, but it doesn't require 5K of test expectation ;-).

cameel: We have #12074 now so you can use output selection to make this smaller. `--ir-optimized` or `--bin` should be enough in this case.

Marenz: i'll test with the new command parameters and see if they also trigger the problem

cameel: I think they should. For the compiler we actually skip generating the stuff for which output is not required but assembler is more straightforward. Output selection just makes it not print some stuff but it still does all the work. Maybe that's actually something we should change though.

Marenz: Both parameters (together and on their own) make it not crash.

Marenz: Adding `sstore(0,0)` makes it crash again though.

cameel: Then it sounds like it might be because of the optimizer?

Option selection should not disable anything unless some actual work is being done inside functions like `print()` or `toHex()`. https://github.com/ethereum/solidity/blob/v0.8.10/solc/CommandLineInterface.cpp#L1036-L1091

Marenz: I think I misspelled something, I am currently testing again

Marenz:
```
crash:
	with sstore(0,0):
		--strict-assembly --yul-dialect evm --machine ewasm
		--strict-assembly --yul-dialect evm --machine ewasm --optimize
	without sstore(0,0):
		--strict-assembly --yul-dialect evm --machine ewasm
No crash:
	with and without sstore(0,0):
		--strict-assembly --yul-dialect evm --machine ewasm --ir-optimized
		--strict-assembly --yul-dialect evm --machine ewasm --ir-optimized --bin
	with no sstore(0,0):
		--strict-assembly --yul-dialect evm --machine ewasm --optimize
```

output wise I guess it's best to go with
```
	with sstore(0,0):

		--strict-assembly --yul-dialect evm --machine ewasm --optimize
```

cameel: `--strict-assembly --yul-dialect evm --machine ewasm --ewasm-ir` also reproduces the crash for me on 0.8.10 (without `sstore` if `--optimize` is not used).

I guess the thing that triggers the crash is called from within `AssemblyStack::print()` after all.

test/cmdlineTests/yul_to_wasm_source_location_crash/args
```diff
@@ -0,0 +1 @@
+--strict-assembly --yul-dialect evm --machine ewasm --optimize
```
cameel:
```suggestion
--strict-assembly --yul-dialect evm --machine ewasm --optimize --ewasm-ir
```

Changelog.md
```diff
@@ -7,6 +7,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to ewasm
```
```suggestion
 * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to ewasm.
```

libyul/backends/wasm/EVMToEwasmTranslator.cpp
cameel: It's a bit more complicated. After #11998 we can discern between the native locations and those grabbed from annotations. In normal yul compilation it does not cause a crash even though the the original source is not available. So why does it crash here?

I think that the problem is actually that the `nativeLocation`s in the generated AST are wrong. `EVMToEwasmTranslator` [copies the whole debug info](https://github.com/ethereum/solidity/blob/v0.8.10/libyul/backends/wasm/EVMToEwasmTranslator.cpp#L96), when it translates the AST. This means that the new AST gets the native location pointing at the original assembly input file rather than the translated source it's supposed to point at.

The problem is that the translated assembly does not actually exist until we generate it from the AST. At the point where the debug info is being copied, we only have the (incomplete) AST and not source we could point at. I think the only thing we can do is to use empty `nativeLocation` in the translated AST. Not having native location means that we can't use it in errors reported from the analysis of that AST. But I don't think we even analyze it - it's generated by the compiler after all so it's not supposed to have errors - those should have been caught during parsing of the original assembly input. It's very likely that the only code that actually tries to access the native location is the `AsmPrinter` (which is where this crash happens).

The location override is another way to get rid of native locations from the AST but a very heavy-handed one - it also discards the origin location. Currently it does not matter much, because I don't think we make use of it when generating ewasm. It will matter once we decide we want to include debug info in ewasm though.

Overall, I think this is an acceptable workaround as long as work on ewasm is stalled. I'd just add a comment explaining that this is just a workaround and that it effectively strips both `originLocation` and `nativeLocation` from the AST even though only `nativeLocation` really has to be discarded.

cameel (commented):
After looking at this closely, this is not a real fix but just a workaround. It strips origin locations from translated AST so we can't use them in Ewasm. But since Ewasm does not use debug info yet and work on it is stalled, I think it's acceptable. I'd just add some comments explaining the situation. See my comment below for details.

Other than that, just two minor tweaks and it's good to go.

chriseth (dismissed):

chriseth (approved):

PR #12195, opened 2021-10-26
Run GnosisSafe external tests with Hardhat and directly on upstream

cameel:
~Depends on #12197.~ Merged.

This is related to #12187, but Gnosis required a lot more changes to switch to the upstream branch due to a switch to Hardhat (with TypeScript config which differs from external test for OZ).

We still need our fork of `gnosis/mock-contract` and also OpenZeppelin needed to be force-updated to 4.0.0 but I still managed to get everything to compile and pass tests on 0.8.9 even though technically the project is still at 0.7.3.

Gnosis v2 appears to have been merged into the `main` branch so I'm removing separate script and CI jobs for it.


cameel:
Upstream tests are failing now:
```
  1) CompatibilityFallbackHandler
       isValidSignature(bytes,bytes)
         should revert if called directly:
     AssertionError: Expected transaction to be reverted with function call to a non-contract account, but other exception was thrown: Error: Transaction reverted: function returned an unexpected amount of data
  

  2) CompatibilityFallbackHandler
       isValidSignature(bytes32,bytes)
         should revert if called directly:
     AssertionError: Expected transaction to be reverted with function call to a non-contract account, but other exception was thrown: Error: Transaction reverted: function returned an unexpected amount of data
```

Given that this is rebased on `develop` now - I wonder if this is a temporary upstream problem or if it's actually caused by something we added in 0.8.10.

cameel:
Rerunning to make sure it's reproducible.

chriseth:
"but other exception was thrown" does not sound too bad, but it looks like it is caused by removing the extcodesize check. I hope that they only check about the exact place the revert happens.

chriseth:
Ok, I'm pretty sure hardhat tries to infer the reason for a revert based on the trace and this mechanism of course does not work any more: https://github.com/nomiclabs/hardhat/blob/6ace4efd4de68bd876330861185b9271dba6ddcf/packages/hardhat-core/src/internal/hardhat-network/stack-traces/mapped-inlined-internal-functions-heuristics.ts#L98

chriseth:
Created an issue on the hardhat repository: https://github.com/nomiclabs/hardhat/issues/2051

cameel:
This now depends on #12197 instead of #12192. I have also moved some changes from here to #12197.

cameel:
Since I don't know when Hardhat will update their heuristics for 0.8.10, I have added a workaround to skip the two failing tests for now.

cameel:
This PR removes `t_native_compile_ext_gnosis` and `t_native_test_ext_gnosis_v2` jobs and both are currently required so we'll need to adjust repo settings after this is merged.

cameel:
[`t_native_test_ext_gnosis`](https://app.circleci.com/pipelines/github/ethereum/solidity/21994/workflows/ea1665c5-9987-4b44-8969-ed96209c34cf/jobs/964519) is now failing with the same error that we've seen in Bleeps recently (https://github.com/ethereum/solidity/pull/12611):

```
  256 passing (27s)
  6 pending
  1 failing

  1) GnosisSafe
       requiredTxGas
         should revert without reason if tx fails:
     AssertionError: Expected transaction to be reverted with Transaction reverted without a reason, but other exception was thrown: Error: invalid sender address, address is not EOA (EIP-3607) (vm hf=arrowGlacier -> block number=16 hash=0x9ad912c5f4e39aabee81e59cf166136dc819b7d225278fb69c3809db61b31848 hf=arrowGlacier baseFeePerGas=0 txs=1 uncles=0 -> tx type=2 hash=0x82941352e8e016735b1e8f21ccdd1b14e931b6b77026b5ab6df0e75d1ec3f8e5 nonce=1 value=0 signed=true hf=arrowGlacier maxFeePerGas=0 maxPriorityFeePerGas=0)
```

In this case it's not reproducible outside of the external test though. With this all tests pass:
```bash
git clone https://github.com/gnosis/safe-contracts.git
cd safe-contracts/
npm install
npm test
```
This means that maybe the issue is in the `gnosis-mock` dependency, which is our fork with modifications for 0.8.x. I need to try updating it.

ekpyron (dismissed):

cameel:
@chriseth Can you adjust required ext tests so that we can merge this now that it's been approved?

PR #12197, opened 2021-10-27
External tests for sushiswap/trident

cameel:
This adds an external test run for [trident](https://github.com/sushiswap/trident).

Unfortunately there is a problem: it uses mixed Solidity versions. `contracts/flat/BentoBoxV1Flat.sol` wants 0.6.12 and requires some minor changes to compile on 0.8.9 (some conversions and constructor visibility. We need a workaround for that or an upstream PR to make the contracts compatible with both versions.


cameel:
- Switched this to a fork with fixed conversions: [solidity-external-tests/trident, branch `master_080`](https://github.com/solidity-external-tests/trident/tree/master_080). If this works I will submit it as a PR to sushiswap.
- Had to add workaround for other 0.6.12 contracts present in `node_modules/`. Fortunately they seem unsued.
- Now I can run tests but many fail with `ProviderError: Must be authenticated!`. I think I need to find a way to run only the subset that does not require an external node.

cameel:
This now works, though not without workarounds:
- Still on our fork until my 0.8.x compatibility fixes are merged upstream (https://github.com/sushiswap/trident/pull/282).
- Some tests fail and had to be disabled (https://github.com/sushiswap/trident/issues/283).
- Can't force latest hardhat because tests only pass with the exact dependency versions from `yarn.lock` (https://github.com/sushiswap/trident/issues/284).

cameel:
Since #12195 has the unresolved issue with Hardhat tests failing due to the changes in the extcodesize check, I swapped it with this PR so that it can go in first. Now this one is directly on top of  #12192 and #12195 is on top of this PR.

cameel:
This is now reviewable. The workarounds are unfortunately still there but they're not that bad and we can remove them once stuff gets fixed upstream.

cameel:
I got no feedback in https://github.com/sushiswap/trident/pull/282 so far so it looks like it might not get merged in the near future. I adjusted the test script to solve the problem by patching upstream repo. This way we can track project's `master` branch and not rely on our fork that needs to be manually updated.

test/externalTests/trident.sh
```diff
@@ -0,0 +1,96 @@
...
+    replace_version_pragmas
+    force_solc_modules "${DIR}/solc"
+
+    # BentoBoxV1Flat.sol requires a few small tweaks to compile on 0.8.x.
+    # TODO: Remove once https://github.com/sushiswap/trident/pull/282 gets merged.
+    sed -i 's|uint128(-1)|type(uint128).max|g' contracts/flat/BentoBoxV1Flat.sol
+    sed -i 's|uint64(-1)|type(uint64).max|g' contracts/flat/BentoBoxV1Flat.sol
+    sed -i 's|uint32(-1)|type(uint32).max|g' contracts/flat/BentoBoxV1Flat.sol
+    sed -i 's|IERC20(0)|IERC20(address(0))|g' contracts/flat/BentoBoxV1Flat.sol
+    sed -i 's|IStrategy(0)|IStrategy(address(0))|g' contracts/flat/BentoBoxV1Flat.sol
```
hrkrshnn, line 85: Oh wow!

hrkrshnn: Mudit said bentobox is a legacy deployed contract, and they probably won't change it in the repo.

I just checked that there are no contract dependencies for Bento. So the project will still compile without this file, however, it's needed for tests. Fine with the sed.

hrkrshnn (dismissed):

test/externalTests/trident.sh
cameel: Yeah, unfortunately we really do want to run tests, not just compile. Especially now that #12441 requires a full test run to gather gas usage for the benchmark.

hrkrshnn (approved):

PR #12198, opened 2021-10-27
Switch ENS external test to ens-contracts repo

### LINKED ISSUE #8557: [Test] add ens contracts to external tests
https://github.com/ensdomains/ethregistrar

cameel:
Fixes #8557.
~Depends on #12192 (draft until that PR is merged).~ Merged.

[ens-contracts](https://github.com/ensdomains/ens-contracts) seems to be the new consolidated repo for all ENS contracts. The project uses Hardhat now and is compatible with Solidity >= 0.8.4.

The test also now uses the upstream directly rather than our fork in solidity-external-tests.

### DIFF
--- a/test/externalTests/common.sh
+++ b/test/externalTests/common.sh
@@ -117,6 +117,19 @@ function neutralize_package_json_hooks
     sed -i 's|"prepare": *".*"|"prepare": ""|g' package.json
 }
 
+function neutralize_packaged_contracts
+{
+    # Frameworks will build contracts from any package that contains a configuration file.
+    # This is both unnecessary (any files imported from these packages will get compiled again as a
+    # part of the main project anyway) and trips up our version check because it won't use our
+    # custom compiler binary.
+    printLog "Removing framework config and artifacts from npm packages..."
+    find node_modules/ -type f '(' -name 'hardhat.config.*' -o -name 'truffle-config.*' ')' -delete
+
+    # Some npm packages also come packaged with pre-built artifacts.
+    find node_modules/ -path '*artifacts/build-info/*.json' -delete
+}
+
 function force_solc_modules
 {
     local custom_solcjs_path="${1:-solc/}"
@@ -216,8 +229,12 @@ function hardhat_verify_compiler_version
     local full_solc_version="$2"
 
     printLog "Verify that the correct version (${solc_version}/${full_solc_version}) of the compiler was used to compile the contracts..."
-    grep '"solcVersion": "'"${solc_version}"'"' --with-filename artifacts/build-info/*.json || fail "Wrong compiler version detected."
-    grep '"solcLongVersion": "'"${full_solc_version}"'"' --with-filename artifacts/build-info/*.json || fail "Wrong compiler version detected."
+    local build_info_files
+    build_info_files=$(find . -path '*artifacts/build-info/*.json')
+    for build_info_file in $build_info_files; do
+        grep '"solcVersion": "'"${solc_version}"'"' --with-filename "$build_info_file" || fail "Wrong compiler version detected in ${build_info_file}."
+        grep '"solcLongVersion": "'"${full_solc_version}"'"' --with-filename "$build_info_file" || fail "Wrong compiler version detected in ${build_info_file}."
+    done
 }
 
 function truffle_clean
--- a/test/externalTests/ens.sh
+++ b/test/externalTests/ens.sh
@@ -28,22 +28,23 @@ verify_input "$@"
 BINARY_TYPE="$1"
 BINARY_PATH="$2"
 
-function compile_fn { npx truffle compile; }
-function test_fn { npm run test; }
+function compile_fn { yarn build; }
+function test_fn { yarn test; }
 
 function ens_test
 {
-    local repo="https://github.com/ensdomains/ens.git"
-    local branch=master
-    local config_file="truffle.js"
+    local repo="https://github.com/ensdomains/ens-contracts.git"
+    local branch="v0.0.8"  # The project is in flux right now and master might be too unstable for us
+    local config_file="hardhat.config.js"
 
-    local compile_only_presets=()
+    local compile_only_presets=(
+        legacy-no-optimize        # Compiles but tests fail to deploy GovernorCompatibilityBravo (code too large).
+    )
     local settings_presets=(
         "${compile_only_presets[@]}"
-        #ir-no-optimize           # "YulException: Variable var_ttl_236 is 1 slot(s) too deep inside the stack."
-        #ir-optimize-evm-only     # "YulException: Variable var_ttl_236 is 1 slot(s) too deep inside the stack."
-        ir-optimize-evm+yul
-        legacy-no-optimize
+        #ir-no-optimize           # Compilation fails with "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
+        #ir-optimize-evm-only     # Compilation fails with "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
+        #ir-optimize-evm+yul      # Compilation fails with "YulException: Variable _5 is 1 too deep in the stack [ _5 usr$i usr$h _7 usr$scratch usr$k usr$f _4 usr$len usr$j_2 RET _2 _1 var_data_mpos usr$totallen usr$x _12 ]"
         legacy-optimize-evm-only
         legacy-optimize-evm+yul
     )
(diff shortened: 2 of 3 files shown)


cameel:
This requires a bit more work. I got Hardhat to compile the project with the custom compiler but `@ensdomains` in `node_modules/` still gets compiled using 0.8.4 (even though I'm replacing all the copies of `solc-js` installed by npm).

leonardoalt:
Is this needed after https://github.com/ethereum/solidity/pull/12187 ?

cameel:
Yes. Both update the repo URL but #12187 is just a trivial switch from our fork to the upstream (which just works, without any other changes needed so it can be merged right away) while this PR switches to a different repo and requires more work. Also depends on other PRs. ENS has moved all of their contracts into a single repo one and switched from Truffle to Hardhat.

I still need to fix one problem here - some dependency is being compiled with older solc, triggering our compiler version check.

test/externalTests/ens.sh
```diff
@@ -27,34 +27,32 @@ source test/externalTests/common.sh
...
-function compile_fn { npx truffle compile; }
-function test_fn { npm run test; }
+function compile_fn { yarn build; }
+function test_fn { yarn test; }
 
 function ens_test
 {
-    local repo="https://github.com/ensdomains/ens.git"
+    local repo="https://github.com/ensdomains/ens-contracts.git"
     local branch=master
```
axic: Not sure about this, I think it is heavily in flux, so a tagged version would be better.

cameel: Oh. You mean an upstream tag or a full-blown fork in solidity-external-tests?

axic: Upstream tag, or even just commit hash.

cameel: Done.

cameel: Looks like the tag I used (`v0.0.5`) is too ancient and lots of tests fail. Switching to a specific commit.

cameel: Looks like `git clone` does not work with specific commits. The only thing I could do is to clone the whole repo (without `--depth 1`) and then check out the commit. I'd need to add a separate function to handle that case. I can do it but it's a bit of a hassle. How unstable is their `master` really?

cameel:
This is now reviewable.

test/externalTests/ens.sh
cameel: So what do we do about it? Can it stay at `master` after all?

test/externalTests/common.sh
```diff
@@ -117,6 +117,19 @@ function neutralize_package_json_hooks
     sed -i 's|"prepare": *".*"|"prepare": ""|g' package.json
 }
 
+function neutralize_packaged_contracts
```
bshastry, line 120: What does `neutralize` mean in this context? I'm wondering if the name could be more readable?

test/externalTests/common.sh
```diff
@@ -216,8 +229,12 @@ function hardhat_verify_compiler_version
...
 
     printLog "Verify that the correct version (${solc_version}/${full_solc_version}) of the compiler was used to compile the contracts..."
-    grep '"solcVersion": "'"${solc_version}"'"' --with-filename artifacts/build-info/*.json || fail "Wrong compiler version detected."
-    grep '"solcLongVersion": "'"${full_solc_version}"'"' --with-filename artifacts/build-info/*.json || fail "Wrong compiler version detected."
+    local build_info_files
+    build_info_files=$(find . -path '*artifacts/build-info/*.json')
+    for build_info_file in $build_info_files; do
+        grep '"solcVersion": "'"${solc_version}"'"' --with-filename "$build_info_file" || fail "Wrong compiler version detected in ${build_info_file}."
+        grep '"solcLongVersion": "'"${full_solc_version}"'"' --with-filename "$build_info_file" || fail "Wrong compiler version detected in ${build_info_file}."
+    done
```
bshastry, line 237: Nice, this looks like better debugging support.

test/externalTests/ens.sh
```diff
@@ -28,22 +28,23 @@ verify_input "$@"
 BINARY_TYPE="$1"
 BINARY_PATH="$2"
 
-function compile_fn { npx truffle compile; }
-function test_fn { npm run test; }
+function compile_fn { yarn build; }
+function test_fn { yarn test; }
```
bshastry, line 32: Why yarn instead of npm?

test/externalTests/ens.sh
```diff
@@ -28,22 +28,23 @@ verify_input "$@"
...
 
 function ens_test
 {
-    local repo="https://github.com/ensdomains/ens.git"
+    local repo="https://github.com/ensdomains/ens-contracts.git"
     local branch=master
-    local config_file="truffle.js"
+    local config_file="hardhat.config.js"
```
bshastry, line 38: Why hardhat instead of truffle?

test/externalTests/ens.sh
```diff
@@ -28,22 +28,23 @@ verify_input "$@"
...
 
-    local compile_only_presets=()
+    local compile_only_presets=(
+        legacy-no-optimize        # Compiles but tests fail to deploy GovernorCompatibilityBravo (code too large).
+    )
```
bshastry, line 42: Why is it okay to tolerate test failures due to large code size (presumably due to no optimization)?

test/externalTests/ens.sh
```diff
@@ -28,22 +28,23 @@ verify_input "$@"
...
     local settings_presets=(
         "${compile_only_presets[@]}"
-        #ir-no-optimize           # "YulException: Variable var_ttl_236 is 1 slot(s) too deep inside the stack."
-        #ir-optimize-evm-only     # "YulException: Variable var_ttl_236 is 1 slot(s) too deep inside the stack."
-        ir-optimize-evm+yul
-        legacy-no-optimize
+        #ir-no-optimize           # "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
+        #ir-optimize-evm-only     # "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
+        #ir-optimize-evm+yul      # "YulException: Variable _5 is 1 too deep in the stack [ _5 usr$i usr$h _7 usr$scratch usr$k usr$f _4 usr$len usr$j_2 RET _2 _1 var_data_mpos usr$totallen usr$x _12 ]"
```
bshastry: How is this to be read? `#<setting> # Description`? Or is it `#<setting>` causes `#YulException:...`?

test/externalTests/common.sh
cameel: "Make them not affect the compilation of contracts in the external project". In this particular case it's that some contracts installed with npm come with a framework config file and/or already compiled JSON artifacts with bytecode.This may or may not affect the contracts in the project but I'm just not taking chances that Hardhat would reuse their already compiled bytecode. We have a check that all build artifacts in the directory tree must come from our selected compiler binary and this trips up that check.

I could call it something like `remove_nested_framework_config_and_artifacts` but the current name conveys my intention a bit better IMO. I.e. if there's anything else needed to make these contracts not interfere (like removing their `package.lock` or something), it's something I'd add to this function and then the name would no longer be precise.

test/externalTests/ens.sh
cameel: I suspect that either would work but I thought it would be better to follow what the project does. In this case the original ENS repo had instructions using `npm` in the README and the new one tells you to use yarn (even has `yarn.lock`).

test/externalTests/ens.sh
cameel: Because they switched to Hardhat.

It's not the same repo really. The [ens](https://github.com/ensdomains/ens) repo was only ENS and was using Truffle. The [ens-contracts](https://github.com/ensdomains/ens-contracts) has contracts from the original ENS repo and from some other repos and has config written from scratch for Hardhat.

test/externalTests/ens.sh
cameel: The latter. It means that with this particular configuration preset the project cannot be compiled (causes the compilation error in the comment). I'll add more text to make it clearer.

test/externalTests/ens.sh
cameel: In this case it's really because I don't want to spend time on tweaking their test/ganache config for this to pass and the this particular config preset is not all that important for us anyway - we skip it in many tests due to stack-related errors. I'd prioritize it more if it was the only preset that could work or if it was the `legacy-optimize-evm+yul` one.

test/externalTests/ens.sh
cameel: Done.

cameel:
Ready for another round or review.

test/externalTests/common.sh
bshastry: Sounds good. I'd still reccommend something like `clean_packaged_contracts` that references Makefile's clean command. However, I'd be fine if you would want to go with the current name.

bshastry:
> Ready for another round or review.

lint of solcjs is failing though.

test/externalTests/common.sh
cameel: I'd rather stay with the current one because I think it's better that it sounds more like workaround than something you'd do normally. If the name makes you uneasy, it works as intended :)

cameel:
Failing job was unrelated, fixed in #12446. I just rebased to make tests pass.

bshastry (dismissed):
LGTM. What is missing consensus is whether to test the `master` branch (unstable), a specific commit etc. So I'd wait for what @axic has to say before merging this PR.

cameel:
Fair enough. I'd be inclined to just track their `master` and see how often it breaks. We can always switch to a fork if it happens too often.

Tracking a specific commit will require some slight adjustments because it cannot be done after `git clone --depth 1`. Would be best if all of these projects had sensible tags but the ones where they would be most helpful are ones that are missing them :)

Also, ENS is not the only project having this problem. Trident is also developing quite rapidly.

test/externalTests/ens.sh
bshastry: There's a new upstream tag dated yesterday: https://github.com/ensdomains/ens-contracts/tags

cameel: Oh, nice. I even checked the repo today but I missed it because I only looked at releases.

In that case I'll switch to that and then let's get this merged :)

bshastry (approved):

PR #12199, opened 2021-10-27
Expose "parsingAndImporting" setting to user

### LINKED ISSUE #11756: `absolutePath` should be available for imports in the AST when using `stopAfter: parsing`
## Abstract
Tools should be using `absolutePath` from `ImportDirective` AST node for more reliable import resolution but it's not available when `stopAfter: parsing` option is used - even though that's when they need it the most.

## Motivation
One of the uses case for `stopAfter: parsing` is to make it possible for tools to easily scan source files for imports. The raw path used in the import statement however is of limited utility - many tools are actually interested in a resolved path that can be used as a key in `sources` in Standard JSON. This resolution must account for [relative imports](https://docs.soliditylang.org/en/latest/path-resolution.html#relative-imports) and [remappings](https://docs.soliditylang.org/en/latest/path-resolution.html#import-remapping) and current compiler logic for doing that is full of unintuitive corner cases so it's easy to get it wrong. Doing it wrong, however, results in the compiler not being able to find the file matching the import where it expects it to be.

The compiler does make the resolved path available in the AST but it's not there yet when `stopAfter: parsing` setting is used.

## Specification
Either perform import resolution earlier so that `absolutePath` is already in the AST when `stopAfter: parsing` is used or introduce a new value for `stopAfter` (`parsingAndImportResolution`?) that still stops before analysis but after import resolution.

## Backwards Compatibility
Making a previously unavailable AST key available is fully backwards-compatible.

#### Input
```json
{
    "language": "Solidity",
    "sources": {
        "/project/../C.sol": {"content": "import \"../L.sol\";"}
    },
    "settings": {
        "stopAfter": "parsing",
        "remappings": [":/project/=/lib/"],
        "outputSelection": {"*": {"": ["ast"]}}
    }
}
```
#### AST from the output
```json
{
    "sources": {
        "/project/../C.sol": {
            "ast": {
                "absolutePath": "/project/../C.sol",
                "id": 2,
                "nodeType": "SourceUnit",
                "nodes": [
                    {
                        "file": "../L.sol",
                        "id": 1,
                        "nameLocation": "-1:-1:-1",
                        "nodeType": "ImportDirective",
                        "src": "0:18:0",
                        "symbolAliases": [],
                        "unitAlias": ""
                    }
                ],
                "src": "0:18:0"
            },
            "id": 0
        }
    }
}
```

#### Input
```json
{
    "language": "Solidity",
    "sources": {
        "/project/../C.sol": {"content": "import \"../L.sol\";"},
        "/lib/L.sol": {"content": ""}
    },
    "settings": {
        "remappings": [":/project/=/lib/"],
        "outputSelection": {"*": {"": ["ast"]}}
    }
}
```
#### AST from the output
```json
{
    "sources": {
        "/lib/L.sol": {
            "ast": {
                "absolutePath": "/lib/L.sol",
                "exportedSymbols": {},
                "id": 1,
                "nodeType": "SourceUnit",
                "nodes": [],
                "src": "0:0:0"
            },
            "id": 0
        },
        "/project/../C.sol": {
            "ast": {
                "absolutePath": "/project/../C.sol",
                "exportedSymbols": {},
                "id": 3,
                "nodeType": "SourceUnit",
                "nodes": [
                    {
                        "absolutePath": "/lib/L.sol",
                        "file": "../L.sol",
                        "id": 2,
                        "nameLocation": "-1:-1:-1",
                        "nodeType": "ImportDirective",
                        "scope": 3,
                        "sourceUnit": 1,
                        "src": "0:18:1",
                        "symbolAliases": [],
                        "unitAlias": ""
                    }
                ],
                "src": "0:18:1"
            },
            "id": 1
        }
    }
}
```

Marenz:
fixes #11756

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ Compiler Features:
  * Commandline Interface: Accept nested brackets in step sequences passed to ``--yul-optimizations``.
  * Commandline Interface: Add ``--debug-info`` option for selecting how much extra debug information should be included in the produced EVM assembly and Yul code.
  * Commandline Interface: Use different colors when printing errors, warnings and infos.
+ * JSON AST: Set absolute paths of imports earlier, in the ``parsing`` stage.
  * SMTChecker: Output values for ``block.*``, ``msg.*`` and ``tx.*`` variables that are present in the called functions.
  * SMTChecker: Report contract invariants and reentrancy properties. This can be enabled via the CLI option ``--model-checker-invariants`` or the Standard JSON option ``settings.modelChecker.invariants``.
  * Standard JSON: Accept nested brackets in step sequences passed to ``settings.optimizer.details.yulDetails.optimizerSteps``.
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -350,8 +350,22 @@ bool CompilerStack::parse()
 		else
 		{
 			source.ast->annotation().path = path;
+
+			for (auto const& import: ASTNode::filteredNodes<ImportDirective>(source.ast->nodes()))
+			{
+				solAssert(!import->path().empty(), "Import path cannot be empty.");
+
+				// The current value of `path` is the absolute path as seen from this source file.
+				// We first have to apply remappings before we can store the actual absolute path
+				// as seen globally.
+				import->annotation().absolutePath = applyRemapping(util::absolutePath(
+					import->path(),
+					path
+				), path);
+			}
+
 			if (m_stopAfter >= ParsedAndImported)
-				for (auto const& newSource: loadMissingSources(*source.ast, path))
+				for (auto const& newSource: loadMissingSources(*source.ast))
 				{
 					string const& newPath = newSource.first;
 					string const& newContents = newSource.second;
@@ -1091,7 +1105,7 @@ string const& CompilerStack::Source::ipfsUrl() const
 	return ipfsUrlCached;
 }
 
-StringMap CompilerStack::loadMissingSources(SourceUnit const& _ast, std::string const& _sourcePath)
+StringMap CompilerStack::loadMissingSources(SourceUnit const& _ast)
 {
 	solAssert(m_stackState < ParsedAndImported, "");
 	StringMap newSources;
@@ -1100,14 +1114,8 @@ StringMap CompilerStack::loadMissingSources(SourceUnit const& _ast, std::string 
 		for (auto const& node: _ast.nodes())
 			if (ImportDirective const* import = dynamic_cast<ImportDirective*>(node.get()))
 			{
-				solAssert(!import->path().empty(), "Import path cannot be empty.");
+				string const& importPath = *import->annotation().absolutePath;
 
-				string importPath = util::absolutePath(import->path(), _sourcePath);
-				// The current value of `path` is the absolute path as seen from this source file.
-				// We first have to apply remappings before we can store the actual absolute path
-				// as seen globally.
-				importPath = applyRemapping(importPath, _sourcePath);
-				import->annotation().absolutePath = importPath;
 				if (m_sources.count(importPath) || newSources.count(importPath))
 					continue;
 
--- a/test/cmdlineTests/standard_parsing_import_absolute_paths/output.json
+++ b/test/cmdlineTests/standard_parsing_import_absolute_paths/output.json
@@ -1,0 +1,67 @@
+{
+    "sources":
+    {
+        "/lib/L.sol":
+        {
+            "ast":
+            {
+                "absolutePath": "/lib/L.sol",
+                "id": 2,
+                "license": "GPL-2.0",
+                "nodeType": "SourceUnit",
+                "nodes":
+                [
+                    {
+                        "id": 1,
+                        "literals":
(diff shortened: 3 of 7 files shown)


chriseth:
I don't understand - why do we need a new "stop after" setting? I thought the idea would be to just fill the absolute path during parsing and before actually trying to load the file?

Marenz:
The ticket specified 

> Either perform import resolution earlier so that absolutePath is already in the AST when stopAfter: parsing is used or introduce a new value for stopAfter (parsingAndImportResolution?) that still stops before analysis but after import resolution.

and I figured we have that internal step already so just exposing it would be the easiest solution.. and I figured it makes sense to expose more steps as necessary..

chriseth:
I think that, unless it is complicated to do, we should expose more useful data earlier.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1102,12 +1116,8 @@ StringMap CompilerStack::loadMissingSources(SourceUnit const& _ast, std::string
 			{
 				solAssert(!import->path().empty(), "Import path cannot be empty.");
```
chriseth: Can you check for `import->annotation().absolutePath` instead now?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -350,8 +350,22 @@ bool CompilerStack::parse()
 		else
 		{
 			source.ast->annotation().path = path;
+
+			for (auto const& node: source.ast->nodes())
```
chriseth: You can use `filteredNodes` here

libsolidity/interface/CompilerStack.cpp
```diff
@@ -350,8 +350,22 @@ bool CompilerStack::parse()
...
+				if (ImportDirective const* import = dynamic_cast<ImportDirective*>(node.get()))
+				{
+					solAssert(!import->path().empty(), "Import path cannot be empty.");
+
+					string importPath = util::absolutePath(import->path(), path);
+					// The current value of `path` is the absolute path as seen from this source file.
+					// We first have to apply remappings before we can store the actual absolute path
+					// as seen globally.
+					importPath = applyRemapping(importPath, path);
+					import->annotation().absolutePath = importPath;
```
chriseth:
```suggestion
					import->annotation().absolutePath = applyRemapping(importPath, path);
```

Changelog.md
```diff
@@ -8,6 +8,7 @@ Compiler Features:
  * Commandline Interface: Accept nested brackets in step sequences passed to ``--yul-optimizations``.
  * Commandline Interface: Add ``--debug-info`` option for selecting how much extra debug information should be included in the produced EVM assembly and Yul code.
  * Commandline Interface: Use different colors when printing errors, warnings and infos.
+ * General: Absolute paths of imports are now set in the ``parsing`` stage.
```
chriseth: We could use "now" in every entry, so I think it is better to not use it at all.

chriseth: Oh and as category, maybe "JSON AST"? Because that is what is relevant to the user reading the changelog.

Marenz: The option is available using the command line as well though?

chriseth: Sure, but the json ast is a generic thing that is available to all ways to access the compiler.

leonardoalt:
Tests failing

libsolidity/interface/CompilerStack.cpp
Marenz: It turns out: No. The path can actually be empty after we resolved it. What was previously "." becomes empty. But we still have the assert pre-resolve stage as it was originally...

libsolidity/interface/CompilerStack.cpp
```diff
@@ -350,8 +350,26 @@ bool CompilerStack::parse()
...
+					// as seen globally.
+					import->annotation().absolutePath =
+						applyRemapping(
+							util::absolutePath(
+								import->path(),
+								path
+							),
+							path
+						);
+				}
```
chriseth:
```suggestion
			for (auto const& import: ASTNode::filteredNodes<ImportDirective>(source.ast->nodes()))
			{
				solAssert(!import->path().empty(), "Import path cannot be empty.");

				// The current value of `path` is the absolute path as seen from this source file.
				// We first have to apply remappings before we can store the actual absolute path
				// as seen globally.
				import->annotation().absolutePath = applyRemapping(util::absolutePath(
					import->path(),
					path
				), path);
			}
```

test/cmdlineTests/standard_parsing_import_absolute_paths/output.json
```diff
@@ -0,0 +1,78 @@
...
+                "nodes": [],
+                "src": "0:0:0"
+            },
+            "id": 0
+        },
+        "/project/../C.sol":
+        {
+            "ast":
+            {
+                "absolutePath": "/project/../C.sol",
```
chriseth, line 34: Interesting!

chriseth: Ah ok, it was named like that in the input json.

chriseth (dismissed):

chriseth (approved):

PR #12200, opened 2021-10-27
Refactor ASTJSON tests to allow easier addition of variations.

Marenz:
I did all the work, wanting to add a variation (parsingAndImporting) and when I was almost done I noticed that there are no imports in the ASTJSON tests because they are individual .sol files, not .json inputs.
But the work was almost done and the result is a nicer class, so I am submiting this now :)

### DIFF
--- a/test/libsolidity/ASTJSONTest.cpp
+++ b/test/libsolidity/ASTJSONTest.cpp
@@ -16,17 +16,20 @@
 */
 // SPDX-License-Identifier: GPL-3.0
 
-#include <boost/algorithm/string/replace.hpp>
-#include <test/libsolidity/ASTJSONTest.h>
-#include <test/Common.h>
-#include <libsolutil/AnsiColorized.h>
 #include <liblangutil/SourceReferenceFormatter.h>
 #include <libsolidity/ast/ASTJsonConverter.h>
-#include <libsolidity/interface/CompilerStack.h>
+#include <libsolutil/AnsiColorized.h>
+#include <libsolutil/CommonIO.h>
+
+#include <test/Common.h>
+#include <test/libsolidity/ASTJSONTest.h>
+
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/predicate.hpp>
-#include <boost/throw_exception.hpp>
+#include <boost/algorithm/string/replace.hpp>
 #include <boost/test/unit_test.hpp>
+#include <boost/throw_exception.hpp>
+
 #include <fstream>
 #include <memory>
 #include <stdexcept>
@@ -72,8 +75,13 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 	if (!boost::algorithm::ends_with(_filename, ".sol"))
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid test contract file name: \"" + _filename + "\"."));
 
-	m_astFilename = _filename.substr(0, _filename.size() - 4) + ".json";
-	m_astParseOnlyFilename = _filename.substr(0, _filename.size() - 4) + "_parseOnly.json";
+	string_view baseName = _filename;
+	baseName.remove_suffix(4);
+
+	m_variants = {
+		TestVariant(baseName, CompilerStack::State::Parsed),
+		TestVariant(baseName, CompilerStack::State::AnalysisPerformed),
+	};
 
 	ifstream file(_filename);
 	if (!file)
@@ -102,26 +110,13 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 	}
 
 	m_sources.emplace_back(sourceName.empty() ? "a" : sourceName, source);
-
 	file.close();
-	file.open(m_astFilename);
-	if (file)
+
+	for (TestVariant& variant: m_variants)
 	{
-		string line;
-		while (getline(file, line))
-			m_expectation += line + "\n";
+		variant.expectation = readFileAsString(variant.astFilename());
+		boost::replace_all(variant.expectation, "\r\n", "\n");
 	}
-
-	file.close();
-	file.open(m_astParseOnlyFilename);
-	if (file)
-	{
-		string line;
-		while (getline(file, line))
-			m_expectationParseOnly += line + "\n";
-	}
-
-	file.close();
 }
 
 TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefix, bool const _formatted)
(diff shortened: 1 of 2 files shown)


test/libsolidity/ASTJSONTest.cpp
```diff
@@ -135,97 +133,87 @@ TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefi
...
+			c,
+			_stream,
+			_linePrefix,
+			_formatted
+		);
 
-		SourceReferenceFormatter{_stream, c, _formatted, false}
-			.printErrorInformation(c.errors());
-		return TestResult::FatalError;
+		continueOuterLoop:;
```
Marenz: Ops, this needs to be removed

test/libsolidity/ASTJSONTest.h
```diff
@@ -37,6 +38,28 @@ namespace solidity::frontend::test
 class ASTJSONTest: public TestCase
 {
 public:
+	struct TestVariant
+	{
+		TestVariant(std::string _baseName, std::string _variation, CompilerStack::State _stopAfter):
+			variation(_variation),
```
cameel:
```suggestion
		TestVariant(std::string_view _baseName, std::string _variation, CompilerStack::State _stopAfter):
			variation(std::move(_variation)),
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -72,8 +74,11 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 	if (!boost::algorithm::ends_with(_filename, ".sol"))
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid test contract file name: \"" + _filename + "\"."));
 
-	m_astFilename = _filename.substr(0, _filename.size() - 4) + ".json";
-	m_astParseOnlyFilename = _filename.substr(0, _filename.size() - 4) + "_parseOnly.json";
+	string const baseName = _filename.substr(0, _filename.size() - 4);
```
cameel:
```suggestion
	string_view baseName = _filename;
	baseName.remove_suffix(4);
```

test/libsolidity/ASTJSONTest.h
```diff
@@ -64,12 +85,9 @@ class ASTJSONTest: public TestCase
 		std::string const& _variation
 	) const;
 
+	std::vector<TestVariant> m_variations;
```
cameel: Why not `m_variants`?
```suggestion
	std::vector<TestVariant> m_variants;
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -102,26 +107,19 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
...
+		{
+			string line;
+			while (getline(file, line))
+				variation.expectation += line + "\n";
+		}
+		file.close();
 	}
-
-	file.close();
 }
```
cameel: We have a helper for reading files from disk. It's `readFileAsString()`, defined in `libsolutil/CommonIO.h`. You could use it here:
```c++
for (TestVariant& variation: m_variations)
	if (boost::filesystem::exists(variation.astFilename))
		variation.expectation = readFileAsString(variation.astFilename);
```

test/libsolidity/ASTJSONTest.h
```diff
@@ -37,6 +38,28 @@ namespace solidity::frontend::test
...
+				_baseName +
+				(_variation.empty() ? "" : "_") +
+				_variation +
+				".json"
+			)
+		{}
+
+		std::string variation;
+		CompilerStack::State stopAfter;
+		std::string expectation;
```
cameel: I'd make it optional. The corresponding AST file might not exist.
```suggestion
		std::optional<std::string> expectation;
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -135,97 +133,85 @@ TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefi
...
+				continue;
+
+			// Empty Expectations means we expect failure
+			if (variation.expectation.empty())
+			{
+				if (resultsMatch)
+					continue;
+				else
+					return TestResult::Failure;
+			}
```
cameel: The comment here seems wrong. I think it's not that "we expect failure" - it will actually succeed if the first variant succeeded. It looks like we just skip the second variant if expectations are empty.

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -135,97 +133,85 @@ TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefi
...
-	c.setSources(sources);
-	c.setEVMVersion(solidity::test::CommonOptions::get().evmVersion());
-	if (!c.parse())
-	{
-		// Empty Expectations means we expect failure
-		if (m_expectation.empty())
-			return resultsMatch ? TestResult::Success : TestResult::Failure;
+		if (!c.parseAndAnalyze(variation.stopAfter))
+		{
+			// Ignore analysis errors, we only want to export.
```
cameel: This was a bit confusing until I checked what happens inside `analyze()`. So if the stack just exits in the middle of analysis, we treat it as a failure but if it reaches the end and only leaves some errors in `m_errorReporter` then we assume that these are not fatal and we can still try to run subsequent variants?
```suggestion
			// Ignore non-fatal analysis errors, we only want to export.
```

test/libsolidity/ASTJSONTest.h
```diff
@@ -37,6 +38,28 @@ namespace solidity::frontend::test
...
+			stopAfter(_stopAfter),
+			astFilename(
+				_baseName +
+				(_variation.empty() ? "" : "_") +
+				_variation +
+				".json"
+			)
+		{}
+
+		std::string variation;
```
cameel:
```suggestion
		std::string name;
```

cameel: Actually, we could get rid of this member and just create a function that returns the name based on `CompilerStack::State` (it could have a static map).

Same with `astFilename` - it could be a function and we could just store `_baseName` instead.

test/libsolidity/ASTJSONTest.h
```diff
@@ -37,6 +38,28 @@ namespace solidity::frontend::test
...
+		CompilerStack::State stopAfter;
+		std::string expectation;
+		std::string astFilename;
+		std::string result;
+		bool success = false;
+
```
cameel: This seems unused.

test/libsolidity/ASTJSONTest.cpp
Marenz: Well, we also would skip the first variant just the same..

test/libsolidity/ASTJSONTest.cpp
Marenz: The previous code basically ignored any errors coming from the call to `c.analyse()`. Here I tried to unify it while keeping that logic. So if we fail and we reached a stage after `Parsing` then we can safely assume no fatal parsing errors happened.

test/libsolidity/ASTJSONTest.cpp
cameel: Sure, I just mean the comment. The code seems fine.

test/libsolidity/ASTJSONTest.h
Marenz: I don't think that case exists in practice

test/libsolidity/ASTJSONTest.cpp
Marenz: Well, in other words though it means that we ignore the result if the expectation is empty.. and we are in the if-branch where a failure happened, to me that sounds pretty much like the comment?

test/libsolidity/ASTJSONTest.h
cameel: Then I'd throw an error when the file does not exist (actually, if you use `readFileAsString()` as I suggested you can just drop the `exists()` check and the function will throw).

And then the `empty()` checks (in particular the one I commented on in https://github.com/ethereum/solidity/pull/12200#discussion_r737758783) can be completely removed (replaced with asserts).

test/libsolidity/ASTJSONTest.cpp
cameel: To me "we expect failure" sounds as if the test should fail if this operation does not fail and succeed if it does fail.

Marenz: That is a fair point. In light that we don't actually have any tests where this is used it might be better to just remove it

leonardoalt:
Tests failing

Changelog.md
```diff
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+ * JSON AST: Set absolute paths of imports earlier, in the ``parsing`` stage.
```
cameel: I think this change belonged to the other PR? This one is only a refactor.

cameel:
This only needs minor tweaks so I just added them myself and pushed to the branch to avoid another round of review. With that I can approve it already.

All changes are in fixups. Please squash if you think they're fine and I'll reapprove.

cameel (dismissed):

Changelog.md
cameel: I just pushed a fix myself.

test/libsolidity/ASTJSONTest.cpp
cameel: Change added as a fixup.

test/libsolidity/ASTJSONTest.cpp
cameel: I saw on the chat you had some problems with this. Apparently due to git converting line endings on Windows?

I looked into making the helper more robust but actually the `\r\n -> \n` conversion is just a one liner in Boost. I have pushed a fixup that adds `readFileAsString()` + the conversion.

cameel: By the way, I think we could extract some more file-related helpers from this file. Stuff like printing indented output for example. I think I've seen that in some other places and it's unnecessarily verbose for something that could just be a one-liner.

Not in this PR of course but something to keep in mind in when we do refactors in these test classes this next time.

cameel (approved):

PR #12205, opened 2021-10-27
Skip extcodesize check if return data is expected.

### LINKED ISSUE #12204: Remove the extcodesize check for external calls that expect returndata
Currently, a high-level call `C.f(...)` does an `extcodesize` check before the `call`  and reverts if the address has empty code. However, this check can be avoided for function calls that are expected to return data. Because such calls are followed up by abi decoding the return data, which has a check for `returndatasize` is being at least a non-zero number. So for empty contracts, they would always revert in the end.

**Proposal**: remove the `extcodesize` check for high-level calls, if such functions have return variables.

Note: because of EIP-2929, this saves 100 gas.

Example

```solidity
interface I {
    function f() external;
    function g() external returns (bool);
}
contract C {
    I immutable i = I(address(0x1234));
    function test() external {
        // Skip extcodesize check in this case.
        // Becuase when ABI decoding the return data of the call
        // it checks if the return data is at least 32 (?) bytes.
        // Note: it does not matter if `i.g()` is assigned to a value
        // the return data is always checked.
        bool b = i.g();
        // Do not skip extcodesize check in this case.
        i.f();
    }
}
```

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12204

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@ Language Features:
 
 
 Compiler Features:
+ * Code Generator: Skip existence check for external contract if return data is expected. In this case, the ABI decoder will revert if the contract does not exist.
  * Commandline Interface: Accept nested brackets in step sequences passed to ``--yul-optimizations``.
  * Commandline Interface: Add ``--debug-info`` option for selecting how much extra debug information should be included in the produced EVM assembly and Yul code.
  * Commandline Interface: Use different colors when printing errors, warnings and infos.
--- a/docs/control-structures.rst
+++ b/docs/control-structures.rst
@@ -114,9 +114,18 @@ otherwise, the ``value`` option would not be available.
 Due to the fact that the EVM considers a call to a non-existing contract to
 always succeed, Solidity uses the ``extcodesize`` opcode to check that
 the contract that is about to be called actually exists (it contains code)
-and causes an exception if it does not.
+and causes an exception if it does not. This check is skipped if the return
+data will be decoded after the call and thus the ABI decoder will catch the
+case of a non-existing contract.
+
 Note that this check is not performed in case of :ref:`low-level calls <address_related>` which
 operate on addresses rather than contract instances.
+
+.. note::
+    Be careful when using high-level calls to
+    :ref:`precompiled contracts <precompiledContracts>`,
+    since the compiler considers them non-existing according to the
+    above logic even though they execute code and can return data.
 
 Function calls also cause exceptions if the called contract itself
 throws an exception or goes out of gas.
--- a/docs/introduction-to-smart-contracts.rst
+++ b/docs/introduction-to-smart-contracts.rst
@@ -565,3 +565,24 @@ contracts, the Ether is forever lost.
 If you want to deactivate your contracts, you should instead **disable** them
 by changing some internal state which causes all functions to revert. This
 makes it impossible to use the contract, as it returns Ether immediately.
+
+
+.. index:: ! precompiled contracts, ! precompiles, ! contract;precompiled
+
+.. _precompiledContracts:
+
+Precompiled Contracts
+=====================
+
+There is a small set of contract addresses that are special:
+The address range between ``1`` and (including) ``8`` contains
+"precompiled contracts" that can be called as any other contract
+but their behaviour (and their gas consumption) is not defined
+by EVM code stored at that address (they do not contain code)
+but instead is implemented in the EVM execution environment itself.
+
+Different EVM-compatible chains might use a different set of
+precompiled contracts. It might also be possible that new
+precompiled contracts are added to the Ethereum main chain in the future,
+but you can reasonabyly expect them to always be in the range between
+``1`` and ``0xffff`` (inclusive).
--- a/libsolidity/codegen/ExpressionCompiler.cpp
+++ b/libsolidity/codegen/ExpressionCompiler.cpp
@@ -2630,9 +2630,21 @@ void ExpressionCompiler::appendExternalFunctionCall(
 	// Check the target contract exists (has code) for non-low-level calls.
 	if (funKind == FunctionType::Kind::External || funKind == FunctionType::Kind::DelegateCall)
 	{
-		m_context << Instruction::DUP1 << Instruction::EXTCODESIZE << Instruction::ISZERO;
-		m_context.appendConditionalRevert(false, "Target contract does not contain code");
-		existenceChecked = true;
+		size_t encodedHeadSize = 0;
+		for (auto const& t: returnTypes)
+			encodedHeadSize += t->decodingType()->calldataHeadSize();
+		// We do not need to check extcodesize if we expect return data, since if there is no
+		// code, the call will return empty data and the ABI decoder will revert.
+		if (
+			encodedHeadSize == 0 ||
+			!haveReturndatacopy ||
+			m_context.revertStrings() >= RevertStrings::Debug
+		)
+		{
+			m_context << Instruction::DUP1 << Instruction::EXTCODESIZE << Instruction::ISZERO;
+			m_context.appendConditionalRevert(false, "Target contract does not contain code");
+			existenceChecked = true;
+		}
 	}
 
 	if (_functionType.gasSet())
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
@@ -2451,8 +2451,10 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 			appendCode() << "mstore(add(" << m_utils.allocateUnboundedFunction() << "() , " << to_string(returnInfo.estimatedReturnSize) << "), 0)\n";
 	}
 
-	Whiskers templ(R"(if iszero(extcodesize(<address>)) { <revertNoCode>() }
-
+	Whiskers templ(R"(
+		<?checkExtcodesize>
+			if iszero(extcodesize(<address>)) { <revertNoCode>() }
+		</checkExtcodesize>
 		// storage for arguments and returned data
 		let <pos> := <allocateUnbounded>()
 		mstore(<pos>, <shl28>(<funSel>))
@@ -2477,6 +2479,18 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		}
 	)");
 	templ("revertNoCode", m_utils.revertReasonIfDebugFunction("Target contract does not contain code"));
+
+	// We do not need to check extcodesize if we expect return data: If there is no
+	// code, the call will return empty data and the ABI decoder will revert.
+	size_t encodedHeadSize = 0;
+	for (auto const& t: returnInfo.returnTypes)
+		encodedHeadSize += t->decodingType()->calldataHeadSize();
+	bool const checkExtcodesize =
+		encodedHeadSize == 0 ||
+		!m_context.evmVersion().supportsReturndata() ||
+		m_context.revertStrings() >= RevertStrings::Debug;
+	templ("checkExtcodesize", checkExtcodesize);
+
 	templ("pos", m_context.newYulVariable());
 	templ("end", m_context.newYulVariable());
 	if (_functionCall.annotation().tryCall)
@@ -2532,6 +2546,8 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		u256 gasNeededByCaller = evmasm::GasCosts::callGas(m_context.evmVersion()) + 10;
 		if (funType.valueSet())
 			gasNeededByCaller += evmasm::GasCosts::callValueTransferGas;
+		if (!checkExtcodesize)
+			gasNeededByCaller += evmasm::GasCosts::callNewAccountGas; // we never know
 		templ("gas", "sub(gas(), " + formatNumber(gasNeededByCaller) + ")");
 	}
 	// Order is important here, STATICCALL might overlap with DELEGATECALL.
--- a/test/cmdlineTests/standard_debug_info_in_yul_location/output.json
+++ b/test/cmdlineTests/standard_debug_info_in_yul_location/output.json
@@ -526,7 +526,6 @@ object \"C_54\" {
                 let expr_46_address := convert_t_contract$_C_$54_to_t_address(expr_45_address)
                 let expr_46_functionSelector := 0x26121ff0
                 /// @src 0:410:418  \"this.f()\"
-                if iszero(extcodesize(expr_46_address)) { revert_error_0cc013b6b3b6beabea4e3a74a6d380f0df81852ca99887912475e1f66b2a2c20() }
 
                 // storage for arguments and returned data
                 let _10 := allocate_unbounded()
@@ -640,7 +639,7 @@ object \"C_54\" {
                     case 0x26121ff0 {
                         if callvalue() { revert(_1, _1) }
                         abi_decode(calldatasize())
-                        let ret := /** @src 0:286:305  \"constVar + immutVar\" */ checked_add_int256_566(/** @src 0:297:305  \"immutVar\" */ loadimmutable(\"8\"))
+                        let ret := /** @src 0:286:305  \"constVar + immutVar\" */ checked_add_int256_556(/** @src 0:297:305  \"immutVar\" */ loadimmutable(\"8\"))
                         /// @src 0:79:435  \"contract C...\"
                         let memPos := mload(64)
                         return(memPos, sub(abi_encode_int256(memPos, ret), memPos))
@@ -664,7 +663,7 @@ object \"C_54\" {
                         if callvalue() { revert(_1, _1) }
                         abi_decode(calldatasize())
                         let memPos_3 := mload(64)
-                        return(memPos_3, sub(abi_encode_int256_565(memPos_3), memPos_3))
+                        return(memPos_3, sub(abi_encode_int256_555(memPos_3), memPos_3))
                     }
                 }
                 revert(0, 0)
@@ -673,7 +672,7 @@ object \"C_54\" {
             {
                 if slt(add(dataEnd, not(3)), 0) { revert(0, 0) }
             }
-            function abi_encode_int256_565(headStart) -> tail
+            function abi_encode_int256_555(headStart) -> tail
             {
                 tail := add(headStart, 32)
                 mstore(headStart, /** @src 0:124:126  \"41\" */ 0x29)
@@ -690,7 +689,7 @@ object \"C_54\" {
                 mstore(4, 0x11)
                 revert(0, 0x24)
             }
-            function checked_add_int256_566(y) -> sum
+            function checked_add_int256_556(y) -> sum
             {
                 if and(1, sgt(y, sub(shl(255, 1), 42))) { panic_error_0x11() }
                 sum := add(/** @src 0:124:126  \"41\" */ 0x29, /** @src 0:79:435  \"contract C...\" */ y)
@@ -711,13 +710,6 @@ object \"C_54\" {
                 if eq(_3, sub(shl(255, 1), 1)) { panic_error_0x11() }
                 let ret := add(_3, 1)
                 sstore(_2, ret)
-                /// @src 0:410:418  \"this.f()\"
-                if iszero(extcodesize(/** @src 0:410:414  \"this\" */ address()))
-                /// @src 0:410:418  \"this.f()\"
-                {
-                    /// @src 0:79:435  \"contract C...\"
-                    revert(_2, _2)
-                }
                 /// @src 0:410:418  \"this.f()\"
                 let _4 := /** @src 0:79:435  \"contract C...\" */ mload(64)
                 /// @src 0:410:418  \"this.f()\"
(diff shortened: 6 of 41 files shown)


Changelog.md
```diff
@@ -5,6 +5,7 @@ Language Features:
 
 
 Compiler Features:
+ * Code Generator: Skip existence check for external calls if return data is expected. In this case, the ABI decoder will revert if the contract does not exist.
```
```suggestion
 * Code Generator: Skip existence check for external contract if return data is expected. In this case, the ABI decoder will revert if the contract does not exist.
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2532,6 +2546,8 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		u256 gasNeededByCaller = evmasm::GasCosts::callGas(m_context.evmVersion()) + 10;
 		if (funType.valueSet())
 			gasNeededByCaller += evmasm::GasCosts::callValueTransferGas;
+		if (!checkExtcodesize)
+			gasNeededByCaller += evmasm::GasCosts::callNewAccountGas; // we never know
```
hrkrshnn, line 2550: This used to happen for all external calls in the old codegen. What is the consequence of missing this in the new codegen?

hrkrshnn (commented):
Can you add the following tests:
- A test which calls some function that expects return data of an empty contract.
- A test which calls some function without return data of an empty contract.
- A test with debug revert strings enabled on an empty address.

Maybe a test for a function that returns a string?

Generally looks good. I'll think this through tomorrow if there are any unintended consequences.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
chriseth: This function in new codegen is more specialized than the function in the old codegen. For example, this function here does not handle low-level calls. Because of that, previously, we would always do a code size check. If you call a contract with a value transfer that does not exist, the call opcode is more expensive: You have to pay for the account creation. That is what this number is for. Pre-tangerine-whistle, you could not try to send more gas with the call than was available after subtracting the gas for the call opcode. Now, you can try to send an arbitrary amount of gas, and the EVM will just forward as much gas to the call as is still available.

test/cmdlineTests/standard_debug_info_in_yul_location/output.json
```diff
@@ -1551,13 +1542,6 @@ object \"D_72\" {
...
-                if iszero(extcodesize(/** @src 0:410:414  \"this\" */ address()))
-                /// @src 0:410:418  \"this.f()\"
-                {
-                    /// @src 1:91:166  \"contract D is C(3)...\"
-                    revert(_2, _2)
-                }
-                /// @src 0:410:418  \"this.f()\"
                 let _4 := /** @src 1:91:166  \"contract D is C(3)...\" */ mload(64)
                 /// @src 0:410:418  \"this.f()\"
                 mstore(_4, /** @src 1:91:166  \"contract D is C(3)...\" */ shl(228, 0x026121ff))
```
hrkrshnn, line 1549: What happened in this test case?

chriseth: Not sure what you mean, some extcodesize checks are removed, what is the problem?

hrkrshnn: Ah. Github is not properly rending this diff. The render is all red to me and looks like a large amount of code is removed.  Seems like they use red in `.json` files?

But, yeah, this is fine.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2630,9 +2630,21 @@ void ExpressionCompiler::appendExternalFunctionCall(
 	// Check the target contract exists (has code) for non-low-level calls.
 	if (funKind == FunctionType::Kind::External || funKind == FunctionType::Kind::DelegateCall)
 	{
-		m_context << Instruction::DUP1 << Instruction::EXTCODESIZE << Instruction::ISZERO;
-		m_context.appendConditionalRevert(false, "Target contract does not contain code");
-		existenceChecked = true;
+		size_t encodedSize = 0;
+		for (auto const& t: returnTypes)
+			encodedSize += t->decodingType()->calldataHeadSize();
```
cameel: This is only the head size, right? Seems enough for this check but I'd use a more specific name to make sure no one tries to use it as actual total encoded size by mistake in the future.
```suggestion
		size_t encodedHeadSize = 0;
		for (auto const& t: returnTypes)
			encodedHeadSize += t->decodingType()->calldataHeadSize();
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2630,9 +2630,21 @@ void ExpressionCompiler::appendExternalFunctionCall(
...
+		// We do not need to check extcodesize if we expect return data, since if there is no
+		// code, the call will return empty data and the ABI decoder will revert.
+		if (
+			encodedSize == 0 ||
+			!haveReturndatacopy ||
+			m_context.revertStrings() >= RevertStrings::Debug
```
cameel, line 2641: Why do we actually care about message being different here? Is it because getting a message about this not being a contract is potentially more informative than failing due to empty returndata? And is the assumption that if someone wants revert strings then he does not really care about the cost of this extra check?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2477,6 +2479,18 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		}
 	)");
 	templ("revertNoCode", m_utils.revertReasonIfDebugFunction("Target contract does not contain code"));
+
+	// We do not need to check extcodesize if we expect return data: If there is no
+	// code, the call will return empty data and the ABI decoder will revert.
+	size_t encodedSize = 0;
+	for (auto const& t: returnInfo.returnTypes)
+		encodedSize += t->decodingType()->calldataHeadSize();
```
```suggestion
	size_t encodedHeadSize = 0;
	for (auto const& t: returnInfo.returnTypes)
		encodedHeadSize += t->decodingType()->calldataHeadSize();
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: Do we have a test covering this?

test/libsolidity/semanticTests/functionCall/failed_create.sol
```diff
@@ -18,17 +18,17 @@ contract C {
 // compileViaYul: also
 // ----
 // constructor(), 20 wei
-// gas irOptimized: 219233
-// gas legacy: 288299
-// gas legacyOptimized: 177933
+// gas irOptimized: 212915
+// gas legacy: 294569
+// gas legacyOptimized: 174699
```
cameel: Why did legacy gas usage increase here?

This is the only place in the whole PR where it increased.

test/libsolidity/semanticTests/functionCall/failed_create.sol
```diff
@@ -18,17 +18,17 @@ contract C {
...
 // x() -> 1
 // f(uint256): 20 -> FAILURE
 // x() -> 1
 // stack(uint256): 1023 -> FAILURE
-// gas irOptimized: 349023
-// gas legacy: 535367
-// gas legacyOptimized: 354656
+// gas irOptimized: 278786
+// gas legacy: 483942
+// gas legacyOptimized: 298807
```
cameel: The fact that these values dropped must mean that we were doing the extcodesize check even when the contract calls itself. In that case maybe unconditionally removing this check for `this` would be another optimization we could do?

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity >= 0.6.0;
```
cameel: Why 0.6.0? Do we need this pragma in these tests?

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
```diff
@@ -0,0 +1,38 @@
...
+
+// This tests skipping the extcodesize check.
+
+interface I {
+    function a() external pure;
+    function b() external;
+    function c() external payable;
+    function x() external returns (uint);
+    function y() external returns (string memory);
+}
```
cameel, line 9: Do we have this check for libraries too? Or just ordinary contracts?

libsolidity/codegen/ExpressionCompiler.cpp
chriseth: It's an optimization and the optimization should be transparent to the semantics (apart from gas costs). We can optimize in some cases but not in all. If you want debug strings, we cannot optimize.

hrkrshnn: > Why do we actually care about message being different here?

We could allow this in theory, but this would be a breaking change.

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
chriseth: Ah sorry, just copied it over from somewhere.

test/libsolidity/semanticTests/functionCall/failed_create.sol
chriseth: calling an external function on `this` is extremely uncommon and it's quite a hassle to detect it at compile-time.

test/libsolidity/semanticTests/functionCall/failed_create.sol
chriseth: This is weird - looking at the asm diff, there is an extcodesize check removed and labels are renamed, I don't see how the code could get longer.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
chriseth: Do we run our tests on ancient EVM versions?

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
chriseth: Yes, it is done for "external" and "delegatecall"

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting_debugstrings.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity >= 0.6.0;
```
cameel:
```suggestion
```

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
cameel: Then I think an extra test with a library might be a good idea.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: Yeah we have a `soltest` run on on `homestead`.

test/libsolidity/semanticTests/functionCall/failed_create.sol
cameel: ok, fine then.

hrkrshnn:
This would break for pre compiles
```solidity
interface Identity {
    function selectorAndAppendValue(uint value) external pure returns (uint);
}
contract C {
    Identity constant i = Identity(address(0x0004));
    function test() external pure returns (bool) {
        i.selectorAndAppendValue(5);
        return true;
    }
}
// ====
// compileViaYul: also
// ----
// test() -> true
```

Changelog.md
```diff
@@ -5,6 +5,7 @@ Language Features:
 
 
 Compiler Features:
+ * Code Generator: Skip existence check for external contract if return data is expected. In this case, the ABI decoder will revert if the contract does not exist.
```
hrkrshnn, line 8: Do we want to add the precompile changes as bugfix?

docs/control-structures.rst
```diff
@@ -114,10 +114,18 @@ otherwise, the ``value`` option would not be available.
 Due to the fact that the EVM considers a call to a non-existing contract to
 always succeed, Solidity uses the ``extcodesize`` opcode to check that
 the contract that is about to be called actually exists (it contains code)
-and causes an exception if it does not.
+and causes an exception if it does not. This check is skipped if return
```
hrkrshnn:
```suggestion
and causes an exception if it does not. This check is skipped if the return
```

docs/control-structures.rst
```diff
@@ -114,10 +114,18 @@ otherwise, the ``value`` option would not be available.
...
+data will be decoded after the call and thus the ABI decoder will catch the
+case of a non-existing contract.
+
 Note that this check is not performed in case of :ref:`low-level calls <address_related>` which
 operate on addresses rather than contract instances.
 
+.. note::
+    Be careful when using high-level calls to precompiled contracts,
+    since the compiler considers them non-existing according to the
+    above logic even though the execute code and can return data.
```
hrkrshnn:
```suggestion
    above logic even though they can execute code and can return data.
```

docs/control-structures.rst
```diff
@@ -114,10 +114,18 @@ otherwise, the ``value`` option would not be available.
...
-and causes an exception if it does not.
+and causes an exception if it does not. This check is skipped if return
+data will be decoded after the call and thus the ABI decoder will catch the
+case of a non-existing contract.
+
 Note that this check is not performed in case of :ref:`low-level calls <address_related>` which
 operate on addresses rather than contract instances.
 
+.. note::
+    Be careful when using high-level calls to precompiled contracts,
```
cameel: I wanted to suggest that we could link to the page or section about the precompiled contracts but looks like we do not really have a dedicated one. The closest I could find was the [Units and Globally Available Variables > Mathematical and Cryptographic Functions](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions) section but it's not only about precompiles. I think we should really have some place that explains what they are in general.

Changelog.md
chriseth: I thought about it but concluded that it's just two weird edge cases that are solved differently now, I'm not sure anymore we can call it a bugfix...

docs/control-structures.rst
chriseth: I think it should be in the "introduction to EVM" section.

leonardoalt:
Tests failing

docs/control-structures.rst
chriseth: Maybe the "introduction to EVM" is too high-level. I have to think about it further.

test/libsolidity/semanticTests/functionCall/failed_create.sol
cameel: I see that the gas here is still at `294569` but it's no longer a change. Was part of this extracted into a separate PR that already went in?

docs/introduction-to-smart-contracts.rst
```diff
@@ -565,3 +565,21 @@ contracts, the Ether is forever lost.
 If you want to deactivate your contracts, you should instead **disable** them
 by changing some internal state which causes all functions to revert. This
 makes it impossible to use the contract, as it returns Ether immediately.
+
+
+.. index:: ! precompiled contracts, precompiles, contract;precompiled
```
cameel: This is the main entry for all these keywords so they should all have exclamation marks.
```suggestion
.. index:: ! precompiled contracts, ! precompiles, ! contract;precompiled
```

docs/introduction-to-smart-contracts.rst
```diff
@@ -565,3 +565,21 @@ contracts, the Ether is forever lost.
...
+
+.. _precompiledContracts:
+
+Precompiled Contracts
+=====================
+
+There is a small set of contract addresses that are special:
+The address range between `1` and (including) `8` contain
```
cameel:
```suggestion
The address range between ``1`` and (including) ``8`` contains
```

docs/introduction-to-smart-contracts.rst
```diff
@@ -565,3 +565,21 @@ contracts, the Ether is forever lost.
...
+"precompiled contracts" that can be called as any other contract
+but their behaviour (and their gas consumption) is not defined
+by EVM code stored at that address (they do not contain code)
+but instead is implemented in the EVM execution environment itself.
+
+It could be that further precompiled contracts can be added in the future,
+but their addresses will always be small.
```
cameel:
```suggestion
but their addresses will always be small integers.
```

docs/introduction-to-smart-contracts.rst
cameel: Maybe it would be useful to list them? Or at least link somewhere where people can find a list?

test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
```diff
@@ -18,8 +18,8 @@ contract C {
 
 }
 // ====
-// compileViaYul: also
 // EVMVersion: >=constantinople
+// compileViaYul: also
 // ----
-// testHighLevel() -> FAILURE
+// testHighLevel() -> true
```
cameel, line 32: The "Should fail" comment above needs to be adjusted.

test/libsolidity/semanticTests/functionCall/failed_create.sol
cameel: Ah, no, it's still a change. I was looking at some partial diff rather than the whole thing.

So anyway, do we have an explanation for why this happens even though the PR only removes code?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: So I guess we'll be dropping `homestead` support?

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting.sol
cameel: What about this test?

test/libsolidity/semanticTests/functionCall/external_call_at_construction_time.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity >= 0.6.0;
```
cameel:
```suggestion
```

docs/control-structures.rst
cameel: Looks great!

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
ekpyron: Opened https://github.com/ethereum/solidity/issues/12246 for that question.

ekpyron (commented):
Code looks good, but haven't looked through docs changes. So "soft-approval" modulo @cameel's comments on the docs.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
chriseth: Why do you think we should drop homestead support, @cameel?

chriseth: Or I mean why do you infer we are dropping homestead support from the comments above?

cameel: Sorry, that wasn't really based only on the comments above. There was some discussion about it (on the chat I think?) and there it was suggested.

If we're not dropping it in the near future then I think we should really have some tests covering this case on some older EVM. Do we have any?

docs/introduction-to-smart-contracts.rst
chriseth: This implies that addresses are integers. Maybe `small (as integers).`?

docs/introduction-to-smart-contracts.rst
chriseth: I thought about it, but I think it goes a bit too much into detail for this introductory section. The solidity documentation is not meant as an evm reference.

cameel: Fair enough. Personally, I think it would be better to link to them anyway but I can see the logic behind not doing that.

docs/introduction-to-smart-contracts.rst
cameel: I see that you just added the range, which makes my comment moot :)

test/libsolidity/semanticTests/functionCall/external_call_to_nonexisting_debugstrings.sol
cameel: This one is still there.

cameel (dismissed):
I see no more actual issues so I'm approving.

It would be good to include a test with libraries (https://github.com/ethereum/solidity/pull/12205#discussion_r738865060) or `homestead` (https://github.com/ethereum/solidity/pull/12205#discussion_r738404078) but I guess the first might be already covered somewhere and `homestead` is ancient so I'm also fine with not adding them if you think they're not worth the effort.

chriseth:
Acutally since the semantic test does not specify an EVM version, it will also run on homestead.

chriseth:
Rebased and squsahed.

cameel (approved):

cindylhiam01-jpg (commented):
const Department = await ethers.getContractFactory("Department");
const dept = await Department.deploy(
  "DBM",                                // _code
  "Department of Budget and Management", // _name
  "Office of the Secretary",            // _mainAgencyName
  DBC_ADDRESS,                          // _dbc
  MAIN_AGENCY_MANAGER                   // mainAgencyOwner
);

PR #10742, opened 2021-01-11
[Yul Optimizer] CommonSwitchPrefixMover

ekpyron:
Depends on https://github.com/ethereum/solidity/pull/10739
Part of https://github.com/ethereum/solidity/issues/6726
So far it neither checks for reverting code paths or does the suggested split in https://github.com/ethereum/solidity/issues/6726 in case not all, but only some cases share the prefix.
Not sure how likely it is to happen in this ways and thus how useful this is...

PR #12206, opened 2021-10-27
Extract external function part

chriseth:
This refactors the dispatch routine: It extracts a function that does the abi decoding and encoding.
The point of this is that for the extracted function, we can determine that it will call `return(...)` and thus optimize the memory usage inside that function. The hope is that the internal part is inlined and thus the memory optimization also extends to this.

Before this change, we would need to inline into the dispatch routine in order to get memory optimization.

Gas numbers do not look that promising, but maybe we have to wait for the #11352 ?


chriseth:
The code looks much nicer, though: https://gist.github.com/chriseth/05075e7706c8ea3f1ac628a28be486f7/revisions

And looking at the functions, I hope the #11352 will do a ton of optimization there. Why do we have to repeat the `keccak` calls all the time?

chriseth:
Maybe we can ignore this after we have "always inline".

libsolidity/codegen/ir/Common.cpp
```diff
@@ -41,6 +41,15 @@ YulArity YulArity::fromType(FunctionType const& _functionType)
...
 
+string IRNames::externalFunctionPart(Declaration const& _functionOrVarDecl)
+{
+	if (auto const* function = dynamic_cast<FunctionDefinition const*>(&_functionOrVarDecl))
+		solAssert(!function->isConstructor());
+
+	return "external_fun_" + _functionOrVarDecl.name() + "_" + to_string(_functionOrVarDecl.id());
+}
+
+
```
Marenz: superfluous blankline

Marenz (changes requested):
minor formatting, LGTM otherwise

libsolidity/codegen/ir/Common.h
```diff
@@ -49,6 +49,7 @@ struct YulArity
 
 struct IRNames
 {
+	static std::string externalFunctionPart(Declaration const& _functionOrVardecl);
```
cameel: "part" sounds non-specific enough that I didn't really know what to expect from the PR name alone. Maybe we could have a more specific term to refer to it? How about something like "ABI wrapper"?

```suggestion
	static std::string externalFunctionABIWrapper(Declaration const& _functionOrVardecl);
```

Alternative ideas: `externalFunctionGlue`, `externalFunctionAdapter`.

cameel:
I rebased this on `develop` to have the Yield test pass so that I can get benchmark report and compare it with `develop`.

While at it I also removed the extra line pointed out by @Marenz.

test/libsolidity/semanticTests/constructor/constructor_arguments_external.sol
```diff
@@ -19,7 +19,7 @@ contract Main {
 // compileViaYul: also
 // ----
 // constructor(): "abc", true
-// gas irOptimized: 112563
+// gas irOptimized: 107175
```
cameel, line 22: This is interesting. Why does this change affect the constructor cost at all? And by that much?

cameel (dismissed):
In terms of correctness this looks perfectly fine.

In terms of gas usage results are a bit mixed but there are a lot of cases where gas goes down and when it goes up the difference is not dramatic. I'm going to check the benchmark report to see how it impacts real-life projects on average.

I also have one naming suggestion.

Overall, I'm approving already because, I don't see any big blockers here.

IngridMi (dismissed):
Tank you 👋

cameel:
Here's the diff of the benchmark summary. For clarity I removed the non-optimized runs. I also removed results for zeppelin and euler because these projects had new commits in the meantime so some of the differences there come from code changes. Perhaps we should switch to hard-coded commits in external tests for benchmarks to stay comparable longer.

Overall there are both positive and negative changes in gas cost but either way they're pretty small. In case of PRBMath the bytecode grew by 2% but in the absolute terms it's just ~500 GAS. Pool Together has the biggest difference in runtime gas usage but it's still a decrease of less than 0.1%.

Apparently in many cases the code produced by legacy codegen was affected. I wonder why - the PR does not seem to be touching anything it uses.

<details>

[summarized-benchmarks-develop.json.txt](https://github.com/ethereum/solidity/files/8111181/summarized-benchmarks-develop.json.txt)
[summarized-benchmarks-extractExternalFunctionPart.json.txt](https://github.com/ethereum/solidity/files/8111183/summarized-benchmarks-extractExternalFunctionPart.json.txt)

```diff
--- summarized-benchmarks-develop.json  2022-02-18 18:03:35.626236065 +0100
+++ summarized-benchmarks-accc5fa.json  2022-02-18 18:03:26.409651647 +0100
@@ -1,9 +1,9 @@
 {
     "bleeps": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 132165,
+            "bytecode_size": 132868,
             "deployment_gas": 0,
-            "method_gas": 39289935,
+            "method_gas": 39289198,
             "version": "bb90cd0"
         },
         "legacy-optimize-evm+yul": {
@@ -29,29 +29,29 @@
     },
     "elementfi": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 599714,
+            "bytecode_size": 597240,
             "deployment_gas": null,
             "method_gas": null,
             "version": "87f8b5e"
         },
         "legacy-optimize-evm+yul": {
             "bytecode_size": 536668,
-            "deployment_gas": 40951128,
-            "method_gas": 104956892,
+            "deployment_gas": 40951104,
+            "method_gas": 104956863,
             "version": "87f8b5e"
         },
     },
     "ens": {
         "legacy-optimize-evm+yul": {
             "bytecode_size": 156937,
-            "deployment_gas": 30073789,
+            "deployment_gas": 30073777,
             "method_gas": 105365362,
             "version": "v0.0.8"
         },
     },
     "gnosis": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 56453,
+            "bytecode_size": 56069,
             "deployment_gas": null,
             "method_gas": null,
             "version": "ea09294"
@@ -65,7 +65,7 @@
     },
     "gnosis2": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 46753,
+            "bytecode_size": 46676,
             "deployment_gas": null,
             "method_gas": null,
             "version": "c4af526"
@@ -80,28 +80,28 @@
     "perpetual-pools": {
         "legacy-optimize-evm+yul": {
             "bytecode_size": 198380,
-            "deployment_gas": 40722937,
-            "method_gas": 807592065,
+            "deployment_gas": 40722962,
+            "method_gas": 807593036,
             "version": "c7636d5"
         },
     },
     "pool-together": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 205294,
-            "deployment_gas": 35803068,
-            "method_gas": 64203843,
+            "bytecode_size": 204861,
+            "deployment_gas": 35774770,
+            "method_gas": 64155656,
             "version": "eae976e"
         },
         "legacy-optimize-evm+yul": {
             "bytecode_size": 204356,
-            "deployment_gas": 37064055,
-            "method_gas": 64253572,
+            "deployment_gas": 37064031,
+            "method_gas": 64253650,
             "version": "eae976e"
         },
     },
     "prb-math": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 37877,
+            "bytecode_size": 38363,
             "deployment_gas": null,
             "method_gas": null,
             "version": "e2b59d8"
@@ -116,29 +116,29 @@
     "trident": {
         "legacy-optimize-evm+yul": {
             "bytecode_size": 335968,
-            "deployment_gas": 29186831,
-            "method_gas": 449651995,
+            "deployment_gas": 29186843,
+            "method_gas": 449651536,
             "version": "0cab5ae"
         },
     },
     "uniswap": {
         "legacy-optimize-evm+yul": {
             "bytecode_size": 178918,
-            "deployment_gas": 20660190,
-            "method_gas": 2833811509,
+            "deployment_gas": 20660178,
+            "method_gas": 2833811608,
             "version": "bd8805c"
         },
     },
     "yield_liquidator": {
         "ir-optimize-evm+yul": {
-            "bytecode_size": 89326,
-            "deployment_gas": 15499886,
-            "method_gas": 2951392,
+            "bytecode_size": 89152,
+            "deployment_gas": 15451227,
+            "method_gas": 2947797,
             "version": "74f969a"
         },
         "legacy-optimize-evm+yul": {
             "bytecode_size": 88852,
-            "deployment_gas": 15499043,
+            "deployment_gas": 15499055,
             "method_gas": 2933980,
             "version": "74f969a"
         },
```

</details>

Marenz (dismissed):

test/libsolidity/semanticTests/constructor/constructor_arguments_external.sol
chriseth: I think it just reduces the deployed code size because it could e.g. find some functions to be equal.

chriseth:
ABIWrapper sounds good - I'll do the change later.

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -735,6 +735,44 @@ string IRGenerator::generateGetter(VariableDeclaration const& _varDecl)
 	});
 }
 
+void IRGenerator::generateExternalFunction(ContractDefinition const& _contract, FunctionType const& _functionType)
+{
+	string functionName = IRNames::externalFunctionABIWrapper(_functionType.declaration());
```
cameel, line 740: Just noticed that this new function is not returning `functionName` like `generateFunction()` is. If it was, you would not have to call `externalFunctionABIWrapper()` second time in `dispatchRoutine()`.

test/cmdlineTests/standard_debug_info_in_yul_location/output.json
```diff
@@ -172,48 +172,28 @@ object \"C_54\" {
...
                     // stateVar()
 
-                    if callvalue() { revert_error_ca66f745a3ce8ff40e2ccaf1ad45db7774001b90d25810abd9040049be7bf4bb() }
-                    abi_decode_tuple_(4, calldatasize())
-                    let ret_0 :=  getter_fun_stateVar_10()
-                    let memPos := allocate_unbounded()
-                    let memEnd := abi_encode_tuple_t_int256__to_t_int256__fromStack(memPos , ret_0)
-                    return(memPos, sub(memEnd, memPos))
+                    external_fun_stateVar_10()
                 }
```
cameel, line 183: This probably does not matter given that we ended up not merging #10742 but I just realized that this would prevent the optimization from that PR (unless this gets inlined back most of the time?).

cameel: Also one of the problems with #10742 was that `switches` with **all** identical cases are apparently rare. But if we removed the `default` here (relying on the fact that cases always end up terminating the execution), it would actually be able to extract the first 2 lines from each case as the common prefix (3 if we moved `memPos` one line up).

cameel (dismissed):
I have one more suggestion but it's not a blocker for merging this so I'm approving already.

test/cmdlineTests/standard_debug_info_in_yul_location/output.json
chriseth: Yeah, I think moving that part out is just too difficult in general. If we have the more specialized dispatch routine, we could group functions into payable and non-payable and then move the check out or something.

libsolidity/codegen/ir/IRGenerator.cpp
chriseth: sounds good, will change it!

cameel (approved):

test/cmdlineTests/standard_debug_info_in_yul_location/output.json
cameel: Could we just move these two lines outside permanently, i.e. could codegen just put them before the switch?

cameel:
Here's the gas diff formatted using the experimental version of my benchmark diffing script:
### `ir-optimize-evm+yul`
|           bleeps | **`+0.53% ❌`** |           `0%` |          `-0%` |
|        elementfi | **`-0.41% ✅`** |                |                |
|              ens |                |                |                |
|           gnosis | **`-0.68% ✅`** |                |                |
|          gnosis2 | **`-0.16% ✅`** |                |                |
|    pool-together | **`-0.21% ✅`** | **`-0.08% ✅`** | **`-0.08% ✅`** |
|         prb-math | **`+1.28% ❌`** |                |                |
|          trident |                |                |                |
| yield_liquidator | **`-0.19% ✅`** | **`-0.31% ✅`** | **`-0.12% ✅`** |

### `legacy-optimize-evm+yul`
|          project | bytecode_size | deployment_gas |     method_gas |
|:----------------:|--------------:|---------------:|---------------:|
|           bleeps |          `0%` |           `0%` |           `0%` |
|           colony |          `0%` |                |                |
|        elementfi |          `0%` |          `+0%` |           `0%` |
|              ens |          `0%` |          `-0%` |           `0%` |
|            euler |          `!V` |           `!V` |           `!V` |
|           gnosis |          `0%` |                |                |
|          gnosis2 |          `0%` |                |                |
|  perpetual-pools |          `0%` |          `+0%` | **`+0.02% ❌`** |
|    pool-together |          `0%` |          `+0%` |          `+0%` |
|         prb-math |          `0%` |           `0%` |           `0%` |
|          trident |          `0%` |          `+0%` |          `-0%` |
|          uniswap |          `0%` |          `-0%` |          `-0%` |
| yield_liquidator |          `0%` |           `0%` |           `0%` |
|         zeppelin |          `!V` |           `!V` |           `!V` 


`!V` = version mismatch

IngridMi (approved):

PR #12209, opened 2021-10-27
Update expectations in tests for inline assembly with name clashes on `breaking`

cameel:
On `breaking` inline assembly with name clashes is disallowed so the tests added in #12139 cause syntax errors.

This PR replaces command-line/semantic tests with syntax tests and updates their expectations.


cameel:
There seems to be a problem with Windows VMs on CircleCI.

chriseth (approved):

PR #12212, opened 2021-10-28
Fix assignment of return variables from modifiers.

### LINKED ISSUE #12061: [Sol->Yul] Function parameters not restored after modifier invocation
The following test case passes via IR and legacy (expectation is to fail via IR because output should be zero).

```
pragma solidity >= 0.0.0;
contract C0 {
  modifier m0(uint i1)
  {
        _;
  }
  function IRReturnsTwo() external m0(o0++) m0(o0++) returns (uint o0)
  {
  }
}
// ====
// compileViaYul: also
// ----
// IRReturnsTwo() -> 2
```

The following test passes on IR.

```
pragma solidity >= 0.0.0;
contract C0 {
  modifier m0(uint i1)
  {
        _;
  }
  function IRReturnsZero() external m0(o0++) m0(o0++) returns (uint o0, uint o1)
  {
  }
}
// ====
// compileViaYul: true
// ----
// IRReturnsZero() -> 0, 0
```

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12061

### DIFF
--- a/libsolidity/codegen/ir/IRGenerator.cpp
+++ b/libsolidity/codegen/ir/IRGenerator.cpp
@@ -438,7 +438,7 @@ string IRGenerator::generateModifier(
 		for (size_t i = 0; i < retParamsIn.size(); ++i)
 		{
 			retParams.emplace_back(m_context.newYulVariable());
-			assignRetParams += retParams.back() + " := " + retParamsIn[i] + "\n";
+			assignRetParams += retParams.at(i) + " := " + retParamsIn.at(i) + "\n";
 		}
 		t("retParams", joinHumanReadable(retParams));
 		t("assignRetParams", assignRetParams);
@@ -529,7 +529,7 @@ string IRGenerator::generateFunctionWithModifierInner(FunctionDefinition const& 
 		for (size_t i = 0; i < retParams.size(); ++i)
 		{
 			retParamsIn.emplace_back(m_context.newYulVariable());
-			assignRetParams += retParams.back() + " := " + retParamsIn[i] + "\n";
+			assignRetParams += retParams.at(i) + " := " + retParamsIn.at(i) + "\n";
 		}
 		vector<string> params = retParamsIn;
 		for (auto const& varDecl: _function.parameters())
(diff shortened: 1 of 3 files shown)


libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -529,7 +529,7 @@ string IRGenerator::generateFunctionWithModifierInner(FunctionDefinition const&
 		for (size_t i = 0; i < retParams.size(); ++i)
 		{
 			retParamsIn.emplace_back(m_context.newYulVariable());
-			assignRetParams += retParams.back() + " := " + retParamsIn[i] + "\n";
+			assignRetParams += retParams[i] + " := " + retParamsIn[i] + "\n";
```
hrkrshnn: Was this just a typo?

hrkrshnn: What about line 441 in the same file? I guess the bug was because it got copied from line 441?

chriseth: But there it is correct, right?

bshastry: Line 441 lgtm.

bshastry (dismissed):
LGTM

libsolidity/codegen/ir/IRGenerator.cpp
hrkrshnn:
```suggestion
			assignRetParams += retParams[i] + " := " + retParamsIn.at(i) + "\n";
```

or like in line 441, but complementary.

```suggestion
			assignRetParams += retParams[i] + " := " + retParamsIn.back() + "\n";
```

hrkrshnn (dismissed):
Looks good. One optional suggestion.

libsolidity/codegen/ir/IRGenerator.cpp
chriseth: What about using `retParams.at(i) + " := " + retParamsIn.at(i) + "\n";` in both places?

hrkrshnn (approved):

PR #12219, opened 2021-11-01
Semantic test for high level call to precompiles.

hrkrshnn:
Because of the extcodesize check, the high level call will fail. Even though the low level call can
succeed and return data.

Related to https://github.com/ethereum/solidity/pull/12205

### DIFF
--- a/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
+++ b/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
@@ -1,0 +1,25 @@
+interface Identity {
+    function selectorAndAppendValue(uint value) external pure returns (uint);
+}
+contract C {
+    Identity constant i = Identity(address(0x0004));
+    function testHighLevel() external pure returns (bool) {
+        // Should fail because `extcodesize(4) = 0`
+        i.selectorAndAppendValue(5);
+        return true;
+    }
+    function testLowLevel() external view returns (uint value) {
+        (bool success, bytes memory ret) =
+            address(4).staticcall(
+                abi.encodeWithSelector(Identity.selectorAndAppendValue.selector, uint(5))
+            );
+        value = abi.decode(ret, (uint));
+    }
+
+}
+// ====
+// compileViaYul: also
+// EVMVersion: >=constantinople
+// ----
+// testHighLevel() -> FAILURE
+// testLowLevel() -> 0xc76596d400000000000000000000000000000000000000000000000000000000


test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
```diff
@@ -0,0 +1,15 @@
+interface Identity {
+    function selectorAndAppendValue(uint value) external pure returns (uint);
+}
+contract C {
+    Identity constant i = Identity(address(0x0004));
+    function test() external pure returns (bool) {
+        // Should fail because `extcodesize(4) = 0`
+        i.selectorAndAppendValue(5);
```
bshastry, line 8: Just curious: How do precompiles work? Are they implemented in the client (host) which is why extcodesize is zero?

hrkrshnn: > Are they implemented in the client (host)

Yes.

> which is why extcodesize is zero?

Not sure why they decided to go with zero. @axic Maybe you can tell us why they went with it?

test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
```diff
@@ -0,0 +1,24 @@
...
+        return true;
+    }
+    function testLowLevel() external view returns (uint decoded) {
+        (bool success, bytes memory ret) =
+            address(4).staticcall(
+                abi.encodeWithSelector(Identity.selectorAndAppendValue.selector, uint(5))
+            );
+        decoded = abi.decode(ret, (uint));
```
hrkrshnn: We need something like `abi.decodePacked` to properly decode this. I think the return value is [4-byte selector, 32-byte padded value 5].

test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
chriseth: If you say extcodesize is nonzero, it implies that there is some code, but they do not have code that is accessible via extcodecopy.

chriseth (approved):

test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol
axic: It is actually 0 as far as I can tell.

The reliable way to tell if something will execute or not is extcodehash: https://eips.ethereum.org/EIPS/eip-1052

chriseth: Oh that is actually a good insight, thanks!

chriseth: But wait, don't we also want to exclude existing accounts with zero code?

PR #12221, opened 2021-11-02
Non-interactive mode for yulopti.

ekpyron:
Background: make it possible to easily run isolated optimizer steps to aid in developing automated verification tools.

Not sure how much we want to refine this, the PR is just a quickly written version that seems to work.

### DIFF
--- a/test/tools/yulopti.cpp
+++ b/test/tools/yulopti.cpp
@@ -76,58 +76,66 @@ namespace po = boost::program_options;
 class YulOpti
 {
 public:
-	void printErrors()
+	static void printErrors(CharStream const& _charStream, ErrorList const& _errors)
 	{
 		SourceReferenceFormatter{
 			cerr,
-			SingletonCharStreamProvider(*m_charStream),
+			SingletonCharStreamProvider(_charStream),
 			true,
 			false
-		}.printErrorInformation(m_errors);
+		}.printErrorInformation(_errors);
 	}
 
-	bool parse(string const& _input)
+	void parse(string const& _input)
 	{
-		ErrorReporter errorReporter(m_errors);
-		m_charStream = make_shared<CharStream>(_input, "");
-		m_ast = yul::Parser(errorReporter, m_dialect).parse(*m_charStream);
-		if (!m_ast || !errorReporter.errors().empty())
+		ErrorList errors;
+		ErrorReporter errorReporter(errors);
+		CharStream _charStream(_input, "");
+		try
 		{
-			cerr << "Error parsing source." << endl;
-			printErrors();
-			return false;
+			m_ast = yul::Parser(errorReporter, m_dialect).parse(_charStream);
+			if (!m_ast || !errorReporter.errors().empty())
+			{
+				cerr << "Error parsing source." << endl;
+				printErrors(_charStream, errors);
+				throw std::runtime_error("Could not parse source.");
+			}
+			m_analysisInfo = make_unique<yul::AsmAnalysisInfo>();
+			AsmAnalyzer analyzer(
+				*m_analysisInfo,
+				errorReporter,
+				m_dialect
+			);
+			if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
+			{
+				cerr << "Error analyzing source." << endl;
+				printErrors(_charStream, errors);
+				throw std::runtime_error("Could not analyze source.");
+			}
 		}
-		m_analysisInfo = make_shared<yul::AsmAnalysisInfo>();
-		AsmAnalyzer analyzer(
-			*m_analysisInfo,
-			errorReporter,
-			m_dialect
-		);
-		if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
+		catch(...)
 		{
-			cerr << "Error analyzing source." << endl;
-			printErrors();
-			return false;
+			cerr << "Fatal error during parsing: " << endl;
+			printErrors(_charStream, errors);
+			throw;
 		}
-		return true;
 	}
 
 	void printUsageBanner(
-		map<char, string> const& _optimizationSteps,
 		map<char, string> const& _extraOptions,
 		size_t _columns
 	)
 	{
 		yulAssert(_columns > 0);
-
+		auto const& optimiserSteps = OptimiserSuite::stepAbbreviationToNameMap();
 		auto hasShorterString = [](auto const& a, auto const& b) { return a.second.size() < b.second.size(); };
 		size_t longestDescriptionLength = std::max(
-			max_element(_optimizationSteps.begin(), _optimizationSteps.end(), hasShorterString)->second.size(),
+			max_element(optimiserSteps.begin(), optimiserSteps.end(), hasShorterString)->second.size(),
 			max_element(_extraOptions.begin(), _extraOptions.end(), hasShorterString)->second.size()
 		);
 
 		vector<string> overlappingAbbreviations =
-			ranges::views::set_intersection(_extraOptions | ranges::views::keys, _optimizationSteps | ranges::views::keys) |
+			ranges::views::set_intersection(_extraOptions | ranges::views::keys, optimiserSteps | ranges::views::keys) |
 			ranges::views::transform([](char _abbreviation){ return string(1, _abbreviation); }) |
 			ranges::to<vector>();
 
@@ -141,7 +149,7 @@ public:
 		);
 
 		vector<tuple<char, string>> sortedOptions =
-			ranges::views::concat(_optimizationSteps, _extraOptions) |
+			ranges::views::concat(optimiserSteps, _extraOptions) |
 			ranges::to<vector<tuple<char, string>>>() |
 			ranges::actions::sort([](tuple<char, string> const& _a, tuple<char, string> const& _b) {
 				return (
@@ -162,54 +170,28 @@ public:
 		}
 	}
 
-	int runSteps(string source, string steps)
+	void disambiguate()
 	{
-		if (!parse(source))
-			return 1;
-
-		set<YulString> reservedIdentifiers;
 		*m_ast = std::get<yul::Block>(Disambiguator(m_dialect, *m_analysisInfo)(*m_ast));
 		m_analysisInfo.reset();
-		m_nameDispenser = make_shared<NameDispenser>(m_dialect, *m_ast, reservedIdentifiers);
-
-		OptimiserStepContext context{
-			m_dialect,
-			*m_nameDispenser,
-			reservedIdentifiers,
-			solidity::frontend::OptimiserSettings::standard().expectedExecutionsPerDeployment
-		};
-
-		map<char, string> const& abbreviationMap = OptimiserSuite::stepAbbreviationToNameMap();
-		for (char stepAbbreviation: steps)
-			if (auto abbreviationAndName = util::valueOrNullptr(abbreviationMap, stepAbbreviation))
-				OptimiserSuite::allSteps().at(*abbreviationAndName)->run(context, *m_ast);
-			else
-			{
-				cerr << "Unknown optimizer step." << endl;
-				return 1;
-			}
-		cout << AsmPrinter{m_dialect}(*m_ast) << endl;
-		return 0;
+		m_nameDispenser.reset(*m_ast);
 	}
 
-	void runInteractive(string source)
+	void runSteps(string _source, string _steps)
 	{
-		bool disambiguated = false;
+		parse(_source);
+		disambiguate();
+		OptimiserSuite{m_context}.runSequence(_steps, *m_ast);
+		cout << AsmPrinter{m_dialect}(*m_ast) << endl;
+	}
+
+	void runInteractive(string _source, bool _disambiguated = false)
+	{
+		bool disambiguated = _disambiguated;
 		while (true)
 		{
-			cout << "----------------------" << endl;
-			cout << source << endl;
-			if (!parse(source))
-				return;
-			set<YulString> reservedIdentifiers;
-			if (!disambiguated)
-			{
-				*m_ast = std::get<yul::Block>(Disambiguator(m_dialect, *m_analysisInfo)(*m_ast));
-				m_analysisInfo.reset();
-				m_nameDispenser = make_shared<NameDispenser>(m_dialect, *m_ast, reservedIdentifiers);
-				disambiguated = true;
-			}
-			map<char, string> const& abbreviationMap = OptimiserSuite::stepAbbreviationToNameMap();
+			parse(_source);
+			disambiguated = disambiguated || (disambiguate(), true);
 			map<char, string> const& extraOptions = {
 				// QUIT starts with a non-letter character on purpose to get it to show up on top of the list
 				{'#', ">>> QUIT <<<"},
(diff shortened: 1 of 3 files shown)


test/tools/yulopti.cpp
```diff
@@ -162,6 +162,36 @@ class YulOpti
 		}
 	}
 
+	int runSteps(string source, string steps)
+	{
+		if (!parse(source))
+			return 1;
+
+		set<YulString> reservedIdentifiers;
+		*m_ast = std::get<yul::Block>(Disambiguator(m_dialect, *m_analysisInfo)(*m_ast));
```
chriseth, line 175: Can you try to reduce code duplication? I mean in the end, can't we even re-use the `runInteractive` function:

```
void runInteractive(string source, optional<string> _steps)
{
    // ...
    char option;
    if (_steps)
    {
        if (_steps->empty()) return;
        option = _steps.front();
        _steps = steps.substr(1);
    }
    else
        option = static_cast<char>(readStandardInputChar());
    // ...

}
```
Or don't you want the intermediate outputs to be printed?

ekpyron: I want to neither print the input nor the intermediate outputs nor the option descriptions - it's meant for non-interactive runs, i.e. the result should be easy to parse and if I want intermediate outputs, I can just run the steps in separate invocations...

Also the disambiguation logic differs slightly, since I don't need to do it over and over again, since I don't allow the var name cleaner... i.e. one global context is enough and I don't need to recreate one each step...

I also don't need to re-parse between all steps...

But yeah, it could still be merged together... but then it's a lot of conditionals replacing a bit of duplication...

ekpyron: I could probably do the opposite, i.e. move all logic for running a single step into a function and call it from ``runInteractive`` each time...

test/tools/yulopti.cpp
```diff
@@ -162,6 +162,36 @@ class YulOpti
...
+
+		map<char, string> const& abbreviationMap = OptimiserSuite::stepAbbreviationToNameMap();
+		for (char stepAbbreviation: steps)
+			if (auto abbreviationAndName = util::valueOrNullptr(abbreviationMap, stepAbbreviation))
+				OptimiserSuite::allSteps().at(*abbreviationAndName)->run(context, *m_ast);
+			else
+			{
+				cerr << "Unknown optimizer step." << endl;
+				return 1;
+			}
```
cameel: Since there's no need to stop after every step, this could easily be replaced with a single call to [`OptimiserSuite::runSequence()`](https://github.com/ethereum/solidity/blob/a7b137829f5df503a1f14f1b15c6049a9d70767b/libyul/optimiser/Suite.h#L78). This way we'd get support for (potentially nested) brackets for free.

The function requires the sequence to be validated with `OptimiserSuite::validateSequence()` - I'd check that in `main()` just before the call to `YulOpti::runSteps()`.

test/tools/yulopti.cpp
```diff
@@ -256,6 +286,11 @@ Allowed options)",
 			po::value<string>(),
 			"input file"
 		)
+		(
+			"steps",
+			po::value<string>(),
+			"steps to execute non-interactively"
+		)
```
cameel: Maybe it would be useful for non-interactive execution to be an independent option? That way you could start with a sequence and then continue interactively on top of that.

It would make invocation a bit more verbose though.

test/tools/yulopti.cpp
cameel: I think that at least the part between `parse()` and context creation looks like it could be extracted into a function. And the rest could be replaced with a call to `OptimiserSuite::runSequence()` as I described above.

ekpyron:
Ah, *sigh*, I mainly wanted to have something quickly to be able to get back to people who would need this :-).
But I guess I can refactor ``OptimiserSuite`` (e.g. ``runSequence`` currently is public, but the constructor is private, so there's still no way to run it, etc.) and do this properly in the process...

test/tools/yulopti.cpp
ekpyron: Not without changing the entire encapsulation and ownership mechanisms of ``OptimiserSuite`` :-).

cameel:
Oh, true, forgot that it's not public.  Still, `runSequence()` can pretty easily be made into a static function and made public. The only state it needs are really `m_context` and `m_debug` used by the other overload. It's just a matter of passing these via parameters.

It's not a huge change and I think that not having two separate pieces of code executing these sequences is worth it. And people might want the ability to run bracketed and nested sequences anyway (we do have them in our default sequence after all).

test/tools/yulopti.cpp
cameel: Ok, you have a point :) But it's only about these two functions, not the whole `OptimiserSuite` and I think that exposing them (and making them static) would not really be all that invasive.

ekpyron: You mean the *only* two functions that ``OptimiserSuite`` has, really ;-)? I'll push some refactoring in a moment, but it wasn't without reason that I wanted to keep this simple at first :-).

ekpyron:
Happier now :-)? We could spend a week to make this nicer, but I'd really just like to have *anything* working fast ;-).

test/tools/yulopti.cpp
```diff
@@ -162,85 +162,97 @@ class YulOpti
...
-	shared_ptr<AsmAnalysisInfo> m_analysisInfo;
-	shared_ptr<NameDispenser> m_nameDispenser;
+	unique_ptr<AsmAnalysisInfo> m_analysisInfo;
+	set<YulString> const m_reservedIdentifiers = {};
+	NameDispenser m_nameDispenser{m_dialect, m_reservedIdentifiers};
+	OptimiserStepContext m_context{
+		m_dialect,
+		m_nameDispenser,
+		m_reservedIdentifiers,
+		solidity::frontend::OptimiserSettings::standard().expectedExecutionsPerDeployment
```
cameel, line 255: Just a question: I just noticed that we have it hard-coded. Isn't this parameter kinda important when playing with the optimizer? Maybe we should have a prameter for this (not in this PR of course)?

test/tools/yulopti.cpp
```diff
@@ -278,7 +301,10 @@ Allowed options)",
 	string input;
 	try
 	{
-		input = readFileAsString(arguments["input-file"].as<string>());
+		if (arguments.count("input-file"))
+			input = readFileAsString(arguments["input-file"].as<string>());
+		else if (nonInteractive)
+			input = readUntilEnd(cin);
```
cameel: I think it would be better to require `-` to use stdin. That would be more consistent with how `solc` works.

I'm fine with it staying like this in this PR in the interest of getting it through faster though :) Could be changed separately if you agree it's a good idea.

test/tools/yulopti.cpp
```diff
@@ -291,8 +317,33 @@ Allowed options)",
...
+		if (!nonInteractive)
+			cout << input << endl;
+		if (arguments.count("steps"))
+		{
+			string sequence = arguments["steps"].as<string>();
+			try
+			{
+				OptimiserSuite::validateSequence(sequence);
+			}
+			catch(std::exception const& _exception)
```
cameel: `OptimizerException` is the only exception that can really be considered a validation error. Anything else is just a random failure. Especially, right now it silences our asserts.

```suggestion
			catch (OptimizerException const& _exception)
```

test/tools/yulopti.cpp
```diff
@@ -162,85 +162,97 @@ class YulOpti
...
-			auto abbreviationAndName = abbreviationMap.find(option);
-			if (abbreviationAndName != abbreviationMap.end())
+			try
 			{
-				OptimiserStep const& step = *OptimiserSuite::allSteps().at(abbreviationAndName->second);
-				step.run(context, *m_ast);
+				switch (option)
+				{
+					case 4:
+					case '#':
```
cameel, line 213: Nice :)

test/tools/yulopti.cpp
```diff
@@ -162,85 +162,97 @@ class YulOpti
...
-				break;
-			case ';':
+			catch (exception const& _exception)
 			{
-				Object obj;
-				obj.code = m_ast;
-				StackCompressor::run(m_dialect, obj, true, 16);
-				break;
+				cerr << "Exception during optimiser step: " << _exception.what() << endl;
 			}
```
cameel: I think we want more diagnostic info for assertions and random exceptions:
```suggestion
			catch (OptimizerException const& _exception)
			{
				cerr << "Exception during optimiser step: " << _exception.what() << endl;
			}
			catch (exception const& _exception)
			{
				cerr << "Unexpected exception during optimiser step:" << endl;
				cerr << boost::diagnostic_information(_exception) << endl;
			}
			catch (...)
			{
				cerr << "Unexpected exception during optimiser step:" << endl;
				cerr << boost::current_exception_diagnostic_information() << endl;
			}
```

Actually, maybe we can even skip the `exception` case (the `(...)` block should cover it) but I haven't checked that so to be sure I'd add both.

cameel (dismissed):
Looks good.

I have some suggestions about stdin handling and exceptions but I'd only consider the ones about exceptions to be important (and I provided suggestions for them that you can just commit and squash) so in the interest of getting it through quickly, I'm approving already.

test/tools/yulopti.cpp
ekpyron: Yep, when I stumbled across this now, I also thought, it should probably be configurable - it would be a bit annoying to do in interactive mode (i.e. allow changing it between steps on the fly), but at least a command line option might be nice. But yeah, we can do it separately.

test/tools/yulopti.cpp
ekpyron: Yeah, that's indeed better, not sure why it didn't cross my mind :-), I'll do it here.

test/tools/yulopti.cpp
cameel: It's a good easy task for an external contributor. I'll create an issue.

test/tools/yulopti.cpp
ekpyron: I'm not sure - there's also ``Unhandled syntactic element in the abbreviation sequence``, which throws a ``YulAssertion`` and to me it seems preferable to just catch and print it here than to crash...

test/tools/yulopti.cpp
ekpyron: I can go for the ``(...)`` case only - but the main reason for doing this is to catch assertions due to unmet requirements, which are "expected", but not ``OptimizerExceptions``, but failing asserts. But yeah, I can just use ``boost::current_exception_diagnostic_information()`` in an catch-all for it.

ekpyron: E.g. things like ``Need to run the function grouper before the stack compressor.`` (a ``YulAssertion``) - stuff like that happens easily and it seems nicer to me to just report it and give you the chance to run the proper step up front rather than quitting or crashing.

test/tools/yulopti.cpp
cameel: This one is an assert on purpose. It's no a runtime error. It can only happen if you add a new special char to [`NonStepAbbreviations`](https://github.com/ethereum/solidity/blob/a06cb0929ff264af483d914edaff2b346e165ee8/libyul/optimiser/Suite.h#L54) and then forget to handle it here.

test/tools/yulopti.cpp
cameel: I'm not suggesting quitting here. Just printing more info than a simple `what()` will give you. Especially for asserts you often get only a generic message (and until #12019 they didn't even have that) so full info might be more helpful in a general case.

For the few cases where there is enough info in the message, maybe we should just replace them with `OptimizerException` (or a subclass)? That could also be a `good first issue`.

ekpyron:
Updated. I also noticed that most parser errors are fatal and throw, so to properly report that, we need to catch in ``parse``, too (and the exception does not provide useful info, but it rather remains in the error list).

It also doesn't ignore ``--help`` anymore now and has proper exit codes on usage errors and more similar minor changes.

test/tools/yulopti.cpp
```diff
@@ -278,7 +323,22 @@ Allowed options)",
 	string input;
 	try
 	{
-		input = readFileAsString(arguments["input-file"].as<string>());
+		if (arguments.count("input-file"))
```
cameel, line 310: I think it's missing an `else` branch that would report an error if the file is not specified.

Looks like this case wasn't being handled correctly before your PR either. Technically, it will now run as if the file was empty so I guess it's bad but I think an error would be better. Especially for non-interactive use where you might not notice and wonder why you're getting empty output :)

test/tools/yulopti.cpp
```diff
@@ -278,7 +323,22 @@ Allowed options)",
...
+		{
+			string filename = arguments["input-file"].as<string>();
+			if (filename == "-")
+			{
+				if (!arguments.count("steps"))
+				{
+					cerr << options;
+					return 1;
+				}
```
cameel: Actually, why not allow stdin in interactive mode too? Combined with #10275 this could allow us to pipe the compiler output directly into the tool and then work on it :)

test/tools/yulopti.cpp
```diff
@@ -162,90 +172,114 @@ class YulOpti
...
+			disambiguate();
+			OptimiserSuite{m_context}.runSequence(_steps, *m_ast);
+			cout << AsmPrinter{m_dialect}(*m_ast) << endl;
+		}
+		catch (...)
+		{
+			cerr << endl << "Exception:" << endl;
+			cerr << boost::current_exception_diagnostic_information() << endl;
+			return false;
+		}
```
cameel: I'd put this handler over the whole `main()` instead.

Other than the loop (which must not exit on errors), I think that nothing in this file should have a `(...)` handler. There should be only a single top-level handler that catches anything not caught lower, prints a message and exits.

cameel (dismissed):
Again, I have some change suggestions but none of it is critical so I'm approving in case you just want to merge it as is.

test/tools/yulopti.cpp
ekpyron: We also won't get rid of catching in ``parse`` at least, because that needs the error reporter, if it's supposed to report anything meaningful...

test/tools/yulopti.cpp
ekpyron: It will fall through below in another ``arguments.count("input-file")`` and print the usage message... not that that's nice, but the effect is reasonable :-)... but I can also add an ``else`` here and do the same right away.

test/tools/yulopti.cpp
ekpyron: How would you tell at which point the input file ends and interaction input begins in stdin :-)?
I forced non-interactive mode, because that doesn't work trivially at least (we could of course wait for the ``}`` closing the main yul block that is expected in input anyways, then skip all whitespace and expect interactive input from there... but if we want anything like that we can do it separately :-)).

test/tools/yulopti.cpp
cameel: True. I mean, I think `parse()` should catch it, print the errors from error reporter and then rethrow it to the top-level handler but that won't let us get rid of the extra handler unfortunately.

test/tools/yulopti.cpp
cameel: Ah, you got me :) Seems obvious but somehow I didn't think of that. We can't read input from the user and pipe to stdin at the same time.

test/tools/yulopti.cpp
cameel: Ah, right, it's this case that prints help if you don't give you anything. I never liked when programs do that but I can understand why. Personally I'd prefer to have an error but if anyone really wants the help screen in that case then I would not object that much, it's still reasonable in a way.

ekpyron: Well, what it currently does is print the help screen but with failure exit-code (while ``--help`` prints it and indicates success) - maybe that's reasonable :-).

test/tools/yulopti.cpp
```diff
@@ -76,58 +76,66 @@ namespace po = boost::program_options;
...
+			AsmAnalyzer analyzer(
+				*m_analysisInfo,
+				errorReporter,
+				m_dialect
+			);
+			if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
+			{
+				cerr << "Error analyzing source." << endl;
+				printErrors(_charStream, errors);
+				throw std::runtime_error("Could not analyze source.");
```
cameel, line 113: ~Why not rethrow the original? At least that's what I meant when I said that I would rethrow it.~

ekpyron:
Ok, I globalized the exception handling and while doing so eliminated some status returns :-).

test/tools/yulopti.cpp
cameel: Ah, disregard this :) It's not inside the `catch` block.

ekpyron: It's not an exception handler - there is no original exception here :-).

ekpyron: In the ``catch(...)`` block below I do rethrow, but here I have nothing to throw, so I make up something new to throw :-).
I guess throwing nothing would also be fine, though, since the error was already printed and we have a catch-all handler in main...

cameel (dismissed):

test/tools/yulopti.cpp
cameel: sounds like a good compromise :)

test/tools/yulopti.cpp
ekpyron: Actually, I had to check what happens if you ``throw;`` without active exception :-). Could have imagined both: termination and something like throwing void - but it in fact aborts, which makes more sense - so I need to throw *something* anyways, and then a ``runtime_error`` is as good as anything I guess.

ekpyron: > Ah, disregard this :) It's not inside the `catch` block.

Only seeing that now, but yeah :-).

cameel: Right, sorry. I commented too fast, before realizing that I'm not looking at the `catch` block. I edited my comment but you also responded pretty fast so you probably caught the original comment :)

ekpyron:
Rebased in hopes of it fixing an hopefully random download error in external tests.

cameel (approved):

cameel:
Was it a download error or a test error? #12187 was merged today so we will start getting errors if the upstream of these tests has breakage in its main branch. At least for OZ and ENS (and Gnosis after the respective PR gets merged).

ekpyron:
> Was it a download error or a test error? #12187 was merged today so we will start getting errors if the upstream of these tests has breakage in its main branch. At least for OZ and ENS (and Gnosis after the respective PR gets merged).

I didn't pay that much attention and don't see the run anymore... I *think* it was one of the ``t_ems_*_ext_zeppelin`` runs and it complained about not being able to fetch some 0.6 soljson.js (whyever it would want to do that)

cameel:
[`t_ems_compile_ext_zeppelin`](https://app.circleci.com/pipelines/github/ethereum/solidity/20241/workflows/4a45d672-d084-4a67-a574-d5f4022312dc/jobs/893678). It does seem like a download failure:
```
Error BDLR503: Couldn't download compiler version 0.6.12. Checksum verification failed. Please check your connection or use local version 0.6.8
```

The fact that it tries to download the compiler is probably due to #10745 - it builds with hardhat but we only configure truffle. #12192 fixes that for good but restarting the test seems to have helped here.

PR #12224, opened 2021-11-03
Added the BigSur 11.6 in the supported os list

dbrans:
See https://github.com/ethereum/solidity/commit/553d3ed7142c692268a2669f82a2fa33b118c0e3 for a comparable change.


cameel (approved):
Thanks!

Please note that we're going to remove this script (#9582) in the near future. The few lines it runs for macOS are simple enough that I'd recommend to just run them manually. Since the script is still here though (and I see that we even still recommend running it in the docs), I'm going to accept this PR.

PR #12226, opened 2021-11-03
Control flow side effects on non-disambiguated source.

chriseth:


libyul/FunctionReferenceResolver.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+	(*this)(_ast);
+	yulAssert(m_scopes.empty());
+}
+
+void FunctionReferenceResolver::operator()(FunctionCall const& _functionCall)
+{
+	for (auto&& scope: m_scopes | ranges::views::reverse)
+		if (scope.count(_functionCall.functionName.name))
+			m_functionReferences[&_functionCall] = scope.at(_functionCall.functionName.name);
```
ekpyron:
```suggestion
		if (auto** functionDefinition = util::valueOrNullptr(scope, _functionAll))
			m_functionReferences[&_functionCall] = *functionDefinition;
```

test/libyul/ControlFlowSideEffectsTest.cpp
```diff
@@ -61,19 +62,20 @@ TestCase::TestResult ControlFlowSideEffectsTest::run(ostream& _stream, string co
...
+	std::vector<std::string> result;
+	for (auto&& [function, effects]: sideEffects.functionSideEffects())
+	{
+		string effectStr = toString(effects);
+		result.emplace_back(
+			function->name.str() + (effectStr.empty() ? ":" : ": " + effectStr) + "\n"
+		);
+	}
+	sort(result.begin(), result.end());
+	m_obtainedResult = util::joinHumanReadable(result, "");
```
ekpyron:
```suggestion
struct FunctionCollector: ASTWalker
{
	void operator()(FunctionDefinition const& _fun) override
	{
		funs.emplace_back(_fun);
		ASTWalker::operator()(_fun);
	}
	vector<FunctionDefinition const*> functions;
};
FunctionCollector functionCollector;
functionCollector(*obj.code);
for (FunctionDefinition const* fun: functionCollector.functions)
{
	string effectStr = toString(sideEffects.functionSideEffects().at(fun));
	  m_obtainedResult += fun->name.str() + (effectStr.empty() ? ":" : ": " + effectStr) + "\n";
}
```

Can't we do it like this :-)? That's not *that* much more code and this way the expectations are actually meaningful in that I can tell which ``b`` is which and don't sort them by side effects...

ekpyron: Or even
```
struct: ASTWalker
{
	ControlFlowSideEffectsCollector sideEffects;
	string& obtainedResult;
	void operator()(FunctionDefinition const& _fun) override
	{
		string effectStr = toString(sideEffects.functionSideEffects().at(fun));
		m_obtainedResult += fun->name.str() + (effectStr.empty() ? ":" : ": " + effectStr) + "\n";
		ASTWalker::operator()(_fun);
	}
	vector<FunctionDefinition const*> functions;
} createResult {
	ControlFlowSideEffectsCollector{
		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion()),
		*obj.code
	},
	obtainedResult;
}(*obj.code);
```

ekpyron: We could even generalize this with
```C++
template <typename Node, typename Visitor>
struct ForEach: std::conditional_t<std::is_const_v<Node>, ASTWalker, ASTModifier> {
	Visitor visitor;
	template<typename Callable>
	ForEach(Callable&& _visitor): visitor(std::forward<Callable>(_visitor)) {}
	void operator()(Node& _node) override
	{
		visitor(_node);
		std::conditional_t<std::is_const_v<Node>, ASTWalker, ASTModifier>::operator()(_node);
	}
	using std::conditional_t<std::is_const_v<Node>, ASTWalker, ASTModifier>::operator();
};
template<typename Node, typename Entry, typename Visitor>
void forEach(Entry&& _entry, Visitor&& visitor)
{
	ForEach<Node, Visitor>(std::forward<Visitor>(visitor))(std::forward<Entry>(_entry));
//      Might need to use this, to also support ``visitor`` as an actual reference to an object with call operators and not an rvalue:
//	ForEach<Node, std::decay_t<Visitor>>(std::forward<Visitor>(visitor))(std::forward<Entry>(_entry));
}
```

and then just use that:
```C++
	forEach<FunctionDefinition const>(*obj.code, [&](auto const& _fun) {
		string effectStr = toString(sideEffects.functionSideEffects().at(&_fun));
		m_obtainedResult += _fun.name.str() + (effectStr.empty() ? ":" : ": " + effectStr) + "\n";
	});
```
might help elsewhere...

chriseth: True!

libyul/ControlFlowSideEffectsCollector.cpp
```diff
@@ -190,57 +195,64 @@ ControlFlowSideEffectsCollector::ControlFlowSideEffectsCollector(
...
+	// Now it is sufficient to handle the reachable function calls (`m_functionCalls`),
+	// we do not have to consider the control-flow graph anymore.
+	for (auto&& [function, calls]: m_functionCalls)
 	{
-		ControlFlowSideEffects& sideEffects = m_functionSideEffects[functionName];
-		auto _visit = [&, visited = std::set<YulString>{}](YulString _function, auto&& _recurse) mutable {
-			if (sideEffects.canTerminate && sideEffects.canRevert)
+		yulAssert(function);
+		ControlFlowSideEffects& functionSideEffects = m_functionSideEffects[function];
+		auto _visit = [&, visited = std::set<FunctionDefinition const*>{}](FunctionDefinition const& _function, auto&& _recurse) mutable {
```
ekpyron, line 204: A recursive lambda with mutable captures :-)? I always got complaints when I did something like that ;-P. But I like it :-).

ekpyron (approved):
I also re-examined the algorithm in the constructor of ``ControlFlowSideEffectsCollector`` - It takes a while to think through it and I'm still not sure if this couldn't be done in an easier, while similarly efficient way, but I haven't found anything wrong with it, so I think it should work.

libyul/ControlFlowSideEffectsCollector.cpp
chriseth: I just copied it from you.

PR #12229, opened 2021-11-03
Fix util::valueOrDefault.

ekpyron:


libsolutil/CommonData.h
```diff
@@ -267,17 +267,21 @@ template<
...
 {
 	auto it = _map.find(_key);
 	static_assert(
 		std::is_same_v<AllowCopyType, detail::allow_copy> ||
-		std::is_reference_v<decltype((it == _map.end()) ? _defaultValue : it->second)>,
+		std::is_reference_v<decltype((it == _map.end()) ? std::forward<ValueType>(_defaultValue) : it->second)>,
 		"valueOrDefault does not allow copies by default. Pass allow_copy as additional argument, if you want to allow copies."
 	);
-	return (it == _map.end()) ? _defaultValue : it->second;
+	return (it == _map.end()) ? std::forward<ValueType>(_defaultValue) : it->second;
```
ekpyron, line 285: The forwarding was definitely missing - without it the ``_defaultValue`` always counts as lvalue and ``is_reference_v`` returns ``true`` (if no conversion needs to happen) and it breaks for temporaries...

libsolutil/CommonData.h
```diff
@@ -267,17 +267,21 @@ template<
 	typename MapType,
 	typename KeyType,
 	typename ValueType = std::decay_t<decltype(std::declval<MapType>().find(std::declval<KeyType>())->second)> const&,
-	typename AllowCopyType = void*
+	typename AllowCopyType = std::conditional_t<std::is_pod_v<ValueType> || std::is_pointer_v<ValueType>, detail::allow_copy, void*>
```
ekpyron, line 270: Defaulting to allow copies for PODs and pointers is probably reasonable.

ekpyron:
This probably helps, but I haven't verified if it always does what it should yet.

libsolutil/CommonData.h
```diff
@@ -267,17 +267,21 @@ template<
...
 >
-decltype(auto) valueOrDefault(MapType&& _map, KeyType const& _key, ValueType&& _defaultValue = {}, AllowCopyType = nullptr)
+decltype(auto) valueOrDefault(
+	MapType&& _map,
+	KeyType const& _key,
+	ValueType&& _defaultValue = {},
+	AllowCopyType = {})
```
bshastry: Just curious: What does an empty initializer list as a default for a type mean?

libsolutil/CommonData.h
bshastry: What does `std::forward<ValueType>(_defaultValue)` do, if `_defaultValue` is a temporary? I am guessing it returns a temporary of `ValueType`?

ekpyron: Yeah - the result is that ``std::forward<ValueType>(_defaultValue)`` will not be an lvalue reference, so the entire expression ``(it == _map.end()) ? std::forward<ValueType>(_defaultValue) : it->second`` will have to copy the value in the map, if it's there...

libsolutil/CommonData.h
ekpyron: That's not the type, but the argument ``AllowCopyType _allowCopy = {}`` just without mentioning the name :-).

bshastry: Ah I see! It may be good to name all parameters just for consistency :-)

ekpyron: It would trigger an "argument is unused" error, though :-). It's this weird meta-programming trick, in which just the type of the argument is really relevant, not its value - and with the our warning-as-errors settings I can't still give it a name :-).

bshastry: Oh I see! Please ignore my comment in that case :-)

chriseth:
```suggestion
	AllowCopyType = {}
)
```

chriseth (changes requested):
Looks good apart from the coding style

chriseth (approved):

Issue #11689, opened 2021-07-24
Upgrade emscripten to 2.0.26+


axic:
Keeping in line with regular upgrades, last we upgraded in 0.8.1 in #10747:
> Update the soljson.js build to emscripten 2.0.12 and boost 1.75.0.

Emscripten is at 2.0.26 now, the [interesting changes for us](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md):
2.0.26:
> - Undefined data symbols (in static executables) are no longer silently ignored at link time. The previous behaviour (which was to silently give all undefined data symbols address zero, which could lead to bugs) can be enabled by passing either -Wl,--allow-undefined or -Wl,--unresolved-symbols=ignore-all.
> - The alignment of long double, which is a 128-bit floating-point value implemented in software, is reduced from 16 to 8. The lower alignment allows max_align_t to properly match the alignment we use for malloc, which is 8 (raising malloc's alignment to achieve correctness the other way would come with a performance regression). (#10072)

And a lot of other asyncify imporvements/changes, though we do not use that feature yet.

2.0.25:
> - **By default (unless EXIT_RUNTIME=1 is specified) emscripten programs running under node will no longer call process.exit() on exit().** Instead they will simply unwind the stack and return to the event loop, much like they do on the web. In many cases the node process will then exit naturally if there is nothing keeping the event loop going. Note for users of node + pthreads: Because of the way that threads are implemented under node multi-threaded programs now require EXIT_RUNTIME=1 (or call emscripten_force_exit) in order to actually bring down the process.
> - **Drop support for node versions older than v5.10.0.** We now assume the existence of Buffer.from which was added in v5.10.0. If it turns out there is still a need to support these older node versions we can add a polyfil under LEGACY_VM_SUPPORT (#14447).

2.0.24:
> - CMake projects (those that either use emcmake or use Emscripten.cmake directly) are new configured to install (by default) directly into the emscripten sysroot. This means that running cmake --install (or running the install target, via make install for example) will install resources into the sysroot such that they can later be found and used by find_path, find_file, find_package, etc. Previously the default was to attempt to install into the host system (e.g /usr/local) which is almost always not desirable. Folks that were previously using CMAKE_INSTALL_PREFIX to build their own secondary sysroot may be able to simplify their build system by removing this completely and relying on the new default.
[...]

PR #12233, opened 2021-11-04
Pass emscripten linker options only when linking.

ekpyron:
In preparation of https://github.com/ethereum/solidity/issues/11689

Emscripten 2.0.32 will raise warnings on this which are treated as error by us (see https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md):
- Emscripten will now warn when linker-only flags are specified in compile-only (-c) mode. Just like with clang itself, this warning can be disabled using the flag: -Wno-unused-command-line-argument.

Will need to look at the soljson.js CI artifact to confirm that the options are correctly honoured and work using ``CMAKE_EXE_LINKER_FLAGS``.

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -111,39 +111,41 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
 			# http://stackoverflow.com/questions/21617158/how-to-silence-unused-command-line-argument-error-with-clang-without-disabling-i
 			add_compile_options(-Qunused-arguments)
 		elseif(EMSCRIPTEN)
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --memory-init-file 0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --memory-init-file 0")
 			# Leave only exported symbols as public and aggressively remove others
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdata-sections -ffunction-sections -fvisibility=hidden")
 			# Optimisation level
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
 			# Re-enable exception catching (optimisations above -O1 disable it)
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s DISABLE_EXCEPTION_CATCHING=0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s DISABLE_EXCEPTION_CATCHING=0")
 			# Remove any code related to exit (such as atexit)
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s EXIT_RUNTIME=0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s EXIT_RUNTIME=0")
 			# Remove any code related to filesystem access
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s FILESYSTEM=0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s FILESYSTEM=0")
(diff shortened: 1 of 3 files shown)


stackenbotten:
There was an error when running `chk_coding_style` for commit `22888b876892e81987163ef4d4865a240f0261d7`:
```
Changelog.md:34:* Pass linker-only Emscripten options only when linking. 

```

ekpyron:
The resulting ``soljson.js`` does contain base64-encoded wasm, which means the options do get through correctly.

ekpyron:
As for the set of options: those are exactly the options that a build with emscripten 2.0.33 complained about.

axic (dismissed):
Looks good to me, but more approvals should be sought.

cmake/EthCompilerSettings.cmake
```diff
@@ -111,39 +111,39 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
...
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s DYNAMIC_EXECUTION=0")
 			# Disable greedy exception catcher
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s NODEJS_CATCH_EXIT=0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s NODEJS_CATCH_EXIT=0")
 			# Abort if linking results in any undefined symbols
 			# Note: this is on by default in the CMake Emscripten module which we aren't using
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s ERROR_ON_UNDEFINED_SYMBOLS=1")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s ERROR_ON_UNDEFINED_SYMBOLS=1")
 			# Disallow deprecated emscripten build options.
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s STRICT=1")
```
axic, line 136: Actually how about this one?

chriseth (dismissed):

cmake/EthCompilerSettings.cmake
```diff
@@ -111,39 +111,39 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
...
 			add_compile_options(-Qunused-arguments)
 		elseif(EMSCRIPTEN)
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --memory-init-file 0")
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --memory-init-file 0")
 			# Leave only exported symbols as public and aggressively remove others
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdata-sections -ffunction-sections -fvisibility=hidden")
 			# Optimisation level
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
 			# Re-enable exception catching (optimisations above -O1 disable it)
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s DISABLE_EXCEPTION_CATCHING=0")
```
axic, line 120: And this?

ekpyron: Apparently ``CMAKE_CXX_FLAGS`` are still passed to the linker as well by emscripten - but I added this and ``-s STRICT`` to both now, just to be sure.

ekpyron:
For the record: the options are documented in https://github.com/emscripten-core/emscripten/blob/main/src/settings.js

chriseth (dismissed):

axic (dismissed):

cameel (dismissed):
Looks fine as far as I can tell. Options seem fine at a glance, though if there are any subtleties involved, I probably would not spot them without digging into the docs.

I checked the build log and there are no warnings. I also downloaded the binary and tried to use it locally with solc-js and it works so at least it's not broken in an obvious way.

ekpyron:
Damn it, now I had to rebase due to a Changelog conflict due to the entry in https://github.com/ethereum/solidity/pull/12236 (which was probably superfluous anyways) :-)... Can I trouble you guys for one more re-approval :-)?

axic (approved):

cameel (approved):

PR #12236, opened 2021-11-04
Remove old compatibility patch to soljson.js.

### LINKED ISSUE #12235: Remove emscripten workarounds.
Looking into https://github.com/ethereum/solidity/issues/11689 I just saw:
```
# Patch soljson.js for backwards compatibility.
# TODO: remove this with 0.7.
# "viiiii" encodes the signature of the callback function.
sed -i -e 's/addFunction(func,sig){/addFunction(func,sig){sig=sig||"viiiii";/' libsolc/soljson.js
```
in ``solidity/scripts/ci/build_emscripten.sh.``.
We should try not to forget this for 0.9 at least.

ekpyron:
Fixes #12235

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -31,6 +31,10 @@ Bugfixes:
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
 
 
+Build System:
+ * Remove obsolete compatibility workaround for emscripten builds.
+
+
 Important Bugfixes in Experimental Features:
  * Yul IR Generator: Changes to function return variables referenced in modifier invocation arguments were not properly forwarded if there was more than one return variable.
 
--- a/scripts/ci/build_emscripten.sh
+++ b/scripts/ci/build_emscripten.sh
@@ -68,10 +68,6 @@ emcmake cmake \
 	-DTESTS=0 \
   ..
 make soljson
-# Patch soljson.js for backwards compatibility.
-# TODO: remove this with 0.7.
-# "viiiii" encodes the signature of the callback function.
-sed -i -e 's/addFunction(func,sig){/addFunction(func,sig){sig=sig||"viiiii";/' libsolc/soljson.js
 
 cd ..
 mkdir -p upload


axic (approved):
Looks good to me, but should get more approvals.

Changelog.md
```diff
@@ -31,6 +31,10 @@ Bugfixes:
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
 
 
+Build System:
+ * Remove obsolete compatibility workaround for emscripten builds.
```
chriseth, line 35: Not sure we need this...

chriseth (approved):

Changelog.md
ekpyron: Yeah, I was wondering whether to add it... but if unexpectedly anyone tries to use soljson.js without solc-js and expects this hack, maybe nice to at least have some mention of it...

cameel:
The hack is no longer needed due to https://github.com/ethereum/solc-js/pull/470, right?

Might be worth making an announcement about this. Some tools and frameworks may be still using old solc-js because it generally works and they download the binaries on their own. That was the case with Truffle until very recently. See https://github.com/trufflesuite/truffle/issues/4134. They've been on 0.6.0 and even now [they're only on 0.6.9](https://github.com/trufflesuite/truffle/blob/92c195085c843130e561e682d754679c3078df26/packages/compile-solidity/package.json#L36) which is just barely enough to avoid this problem.

I see that [Waffle is still on 0.6.3](https://github.com/EthWorks/Waffle/blob/master/waffle-compiler/package.json#L49) so we should give them a heads up - might be a good idea to submit an issue in their repo.

cameel:
We discussed it on the chat and started wondering what versions are used in general. Here's the list I compiled. Reposting it here so that it's easier to find and reference later:
- Truffle - 0.6.9
- Waffle - 0.6.3
- Hardhat - [0.7.3](https://github.com/nomiclabs/hardhat/blob/633b8bf1b0e866553b23641adfcdb740a39399a2/packages/hardhat-core/package.json#L134)
- vscode-solidity: [0.8.6](https://github.com/juanfranblanco/vscode-solidity/blob/d3a0feecfdc8a06a08c78e52c2e9a56ac520e971/package.json#L65)
- Brownie: uses native binaries
- dapp-tools: uses native binaries
- pocketh: [0.5.8](https://github.com/ajsantander/pocketh/blob/df6ac1a583e02750e0030e5bb98050ca35598d86/package.json#L40)
- ethers.js: [0.7.1](https://github.com/ethers-io/ethers.js/blob/b1458989761c11bf626591706aa4ce98dae2d6a9/package.json#L100)
- embark: [0.6.1](https://github.com/embarklabs/embark/blob/c7a5c4c36880dd254917e9a1060788ed7763f943/packages/plugins/solidity/package.json#L58)
- [saddle](https://github.com/compound-finance/saddle): does not seem to have a specific dependency in `package.json`
- [crytic](https://github.com/crytic/crytic-compile): written in Python, probably uses native binaries.

PR #12237, opened 2021-11-04
Remove install_deps script.

ekpyron:
Fixes #12237.

Just throwing this out there, s.t. we hopefully finally get rid of this script eventually.
We may want to extend the documentation a bit - however, actually all required information is in there already, so maybe it's just fine and we can revisit, if we notice people complaining about not being able to setup a build?


ekpyron:
An option would be to add subsections with the concrete commands to install the dependencies for a selected set of distributions (e.g. debian, ubuntu, redhat, archlinux, osx) to subsections in the docs - that's still better than asking users to run an opaque and unmaintained script...

axic (approved):
Checked that the documentation mentions cmake/boost, and it does. The install script only added cmake, boost, and z3/cvc4 in some cases.

I think we wanted to get rid of this for a while, because we can't reasonably maintain it, but I'm fine either way.

Please wait for other reviews.

chriseth (approved):

cameel:
This fixes #12237.

I wonder if github will close it if I add it to the description now?

cameel:
Apparently it won't.

PR #12240, opened 2021-11-04
Introduce forEach for yul ast nodes.

chriseth:


libyul/optimiser/ASTWalker.h
```diff
@@ -102,4 +102,32 @@ class ASTModifier
...
 
+template <
+	typename Node,
+	typename Visitor,
+	typename Base = std::conditional_t<std::is_const_v<Node>, ASTWalker, ASTModifier>
+>
+struct ForEach: Base
+{
+	template<typename Callable>
+	ForEach(Callable&& _visitor): visitor(std::forward<Callable>(_visitor)) {}
```
chriseth, line 115: Why do we need the constructor?

libyul/optimiser/FunctionDefinitionCollector.h
```diff
@@ -31,14 +31,10 @@ namespace solidity::yul
  *
  * Prerequisite: Disambiguator
  */
-class FunctionDefinitionCollector: ASTWalker
+class FunctionDefinitionCollector
```
chriseth: Maybe this whole thing can now be a function in Utility.h?

libyul/optimiser/ASTWalker.h
ekpyron: Since the base classes have virtual functions, aggregate initialization won't work.. or sth like that :-)...
I templated it to be sure that the forwarding works correctly.

But in general, the class itself is not meant for general use - I might even put it into ``namespace detail`` to clarify that ``forEach`` is the proper interface and this is merely a technical helper...

chriseth: moved it into detail.

libyul/optimiser/NameCollector.h
```diff
@@ -130,4 +116,12 @@ class AssignmentsSinceContinue: public ASTWalker
 	std::set<YulString> m_names;
 };
 
+/// @returns the names of all variables that are assigned to inside @a _code.
+/// (ignores variable declarations)
+std::set<YulString> assignedVariableNames(Block const& _code);
+
+/// @returns all function definitions anywhere in the AST.
+/// Requires disambiguated source.
+std::map<YulString, FunctionDefinition const*> allFunctionDefinitions(Block const& _block);
```
ekpyron, line 125: ``static``?

ekpyron (approved):

libyul/optimiser/ASTWalker.h
```diff
@@ -102,4 +102,35 @@ class ASTModifier
...
+	Visitor visitor;
+};
+}
+
+/// Helper function that traverses the AST and calls the visitor for each
+/// node of a specific type.
+template<typename Node, typename Entry, typename Visitor>
+void forEach(Entry&& _entry, Visitor&& _visitor)
+{
+	detail::ForEach<Node, std::decay_t<Visitor>>{std::forward<Visitor>(_visitor)}(std::forward<Entry>(_entry));
```
ekpyron, line 133: Actually, looking at this once more, my comment about ``std::decay_t`` was wrong...
If ``Visitor`` resolves to a reference type, then that reference is sure to be alive during the body of ``forEach`` - so it is actually *good* if ``detail::ForEach`` has a reference member and the constructor does not need to copy. I'll quickly check if this is indeed what will happen and will probably create a followup PR to remove it.

PR #12241, opened 2021-11-04
Update emscripten Dockerfile.

ekpyron:
Let's see if this works.

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.emscripten
+++ b/scripts/docker/buildpack-deps/Dockerfile.emscripten
@@ -26,10 +26,14 @@
 # contains a Makefile in the docker/ subdirectory that can be used to create the
 # required base image using:
 #
-#   make version=2.0.12 build
+#   make version=2.0.33 build
 #
-FROM emscripten/emsdk:2.0.12 AS base
-LABEL version="6"
+# Note that emscripten is supposed to automatically install to $(em-config CACHE)/sysroot, but
+# apparently this currently breaks due to conflicting compatibility headers.
+# Using $(em-config CACHE)/sysroot/usr seems to work, though, and still has cmake find the
+# dependencies automatically.
+FROM emscripten/emsdk:2.0.33 AS base
+LABEL version="7"
 
 ADD emscripten.jam /usr/src
 RUN set -ex; \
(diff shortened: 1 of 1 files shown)


cameel:
Looks good to me and checks passed. Are we merging it?

ekpyron:
@cameel The process is usually:
- Open a draft with this to build a docker image - that worked here, but the test runs here still use the old docker image - so the docker build itself performed one build of soljson.js, but no tests and the other CI test runs do not yet use it.
- Open a second PR that updates the hash of the docker image to the one posted by the bot. I'm about to do that :-). The CI runs in *that* PR then accurately judge the change.
- If CI passes in the second PR, merge both.

So bear with me ;-).

cameel:
Ah, ok, I know the process. I just thought there was something more holding it back. Like an unsolved problem or something.

ekpyron:
> Ah, ok, I know the process. I just thought there was something more holding it back. Like an unsolved problem or something.

No, I was just in a call when the build passed :-).

cameel (approved):
In that case I'm going to approve it already since there doesn't seem to be anything wrong with it.

scripts/docker/buildpack-deps/Dockerfile.emscripten
```diff
@@ -39,8 +43,8 @@ RUN set -ex; \
 	mkdir build; \
 	cd build; \
 	emcmake cmake \
+		-DCMAKE_INSTALL_PREFIX=$(em-config CACHE)/sysroot/usr \
```
axic, line 46: Can you pull this out into a config variable so that the two copies are kept in sync?

ekpyron: I probably could... but do you think that warrants rebuilding everything :-)?
It's a bug that it's required here anyways, so I'm pretty sure we'll need to change this on the next update anyways.

ekpyron: I actually thought about adding a comment about removing it here next time (or at least trying to), but wasn't sure how nicely that would play with the line break escape, so I added the one on top.

ekpyron: Also there's little danger updating this incorrectly - if it's not done right, the builds will fail.

axic: I'm not too set on it.

axic (approved):

PR #12243, opened 2021-11-05
Minor fix to forEach on Yul AST nodes.

ekpyron:
I commented that we might need ``decay_t`` to make this work for lvalue references to callable objects as ``_visitor``, thinking that the member of ``detail::ForEach`` couldn't be a reference - but actually, thinking through lifetimes, it *should* be a reference in this case and the current version with ``decay_t`` would needlessly copy.

Not that it matters too much - in all likelyhood it will always be a lambda passed by value anyways, but still :-).

Or well, one might e.g. try to pass a reference to the same mutable generic lambda to ``forEach<Assignment const>`` and ``forEach<VariableDeclaration const>`` and would get into trouble, since the lambda closure would be copied and not passed by reference... so better to fix.

And actually, we can *always* use a reference and don't need to forward: no matter what you pass into ``forEach``, it will live through the entire lifetime of the body of ``forEach``, so no need to ever copy or move it anywhere, but we can just keep it there and reference it.

I tested it and this way it works with zero-copy and zero-move no matter what you pass to it.

### DIFF
--- a/libyul/optimiser/ASTWalker.h
+++ b/libyul/optimiser/ASTWalker.h
@@ -111,8 +111,7 @@ template <
 >
 struct ForEach: Base
 {
-	template<typename Callable>
-	ForEach(Callable&& _visitor): visitor(std::forward<Callable>(_visitor)) {}
+	ForEach(Visitor& _visitor): visitor(_visitor) {}
 
 	using Base::operator();
 	void operator()(Node& _node) override
@@ -121,7 +120,7 @@ struct ForEach: Base
 		Base::operator()(_node);
 	}
 
-	Visitor visitor;
+	Visitor& visitor;
 };
 }
 
@@ -130,7 +129,7 @@ struct ForEach: Base
 template<typename Node, typename Entry, typename Visitor>
 void forEach(Entry&& _entry, Visitor&& _visitor)
 {
-	detail::ForEach<Node, std::decay_t<Visitor>>{std::forward<Visitor>(_visitor)}(std::forward<Entry>(_entry));
+	detail::ForEach<Node, Visitor&>{_visitor}(_entry);
 }
 
 }


libyul/optimiser/ASTWalker.h
```diff
@@ -130,7 +129,7 @@ struct ForEach: Base
 template<typename Node, typename Entry, typename Visitor>
 void forEach(Entry&& _entry, Visitor&& _visitor)
 {
-	detail::ForEach<Node, std::decay_t<Visitor>>{std::forward<Visitor>(_visitor)}(std::forward<Entry>(_entry));
+	detail::ForEach<Node, Visitor&>{_visitor}(std::forward<Entry>(_entry));
```
ekpyron: Could actually skip the ref now, since ``detail::ForEach`` makes it a reference anyways... but ref to ref is just ref, so it doesn't matter and maybe clearer to be explicit about it on both ends.

libyul/optimiser/ASTWalker.h
```diff
@@ -130,7 +129,7 @@ struct ForEach: Base
 template<typename Node, typename Entry, typename Visitor>
 void forEach(Entry&& _entry, Visitor&& _visitor)
 {
-	detail::ForEach<Node, std::decay_t<Visitor>>{std::forward<Visitor>(_visitor)}(std::forward<Entry>(_entry));
+	detail::ForEach<Node, Visitor&>{_visitor}(_entry);
```
ekpyron, line 132: Actually, we also don't need to forward ``_entry`` - we know that it will always be passed either by reference or const reference anyways.

ekpyron:
Yeah, looking at the result after a few simplifications, I must say my original version was overcomplicated and a bit convoluted :-). The result in this PR is much cleaner and simpler and will be functionally equivalent, respectively, actually even superior.

chriseth (approved):

Issue #11940, opened 2021-09-13
Snapcraft build has been broken since 0.5.16


lightyear15:
## Abstract
https://snapcraft.io/build
Auto-build and publish new solc version on the snapcraft app store.

## Motivation
solc on snapcraft is at version 0.5.16. Last time it was built was January 2020.

## Specification
As explained in the link attached in the Abstract section, a snapcraft app can be connected to a github repo, so new builds are triggered every new commit.

cameel:
Looks like the snapcraft build is already automated. It's just broken. According to the discussion in #2475 it's tied to https://launchpad.net/solidity. In [~elopio/+snap/solc](https://launchpad.net/~elopio/+snap/solc) I see that builds were running (and failing) until a year ago. Then they stopped. [The build from 2020-09-28](https://launchpad.net/~elopio/+snap/solc/+build/1131834/+files/buildlog_snap_ubuntu_bionic_amd64_solc_BUILDING.txt.gz) has some build errors related to Z3:

```
[  9%] Building CXX object libsmtutil/CMakeFiles/smtutil.dir/Z3CHCInterface.cpp.o
[01m[K/build/solc/parts/solidity/src/libsmtutil/Z3Interface.cpp:[m[K In member function ‘[01m[Kz3::expr solidity::smtutil::Z3Interface::toZ3Expr(const solidity::smtutil::Expression&)[m[K’:
[01m[K/build/solc/parts/solidity/src/libsmtutil/Z3Interface.cpp:201:15:[m[K [01;31m[Kerror: [m[K‘[01m[Kint2bv[m[K’ is not a member of ‘[01m[Kz3[m[K’
    return z3::[01;31m[Kint2bv[m[K(size, arguments[0]);
               [01;31m[K^~~~~~[m[K
[01m[K/build/solc/parts/solidity/src/libsmtutil/Z3Interface.cpp:207:15:[m[K [01;31m[Kerror: [m[K‘[01m[Kbv2int[m[K’ is not a member of ‘[01m[Kz3[m[K’
    return z3::[01;31m[Kbv2int[m[K(arguments[0], intSort->isSigned);
               [01;31m[K^~~~~~[m[K
[01m[K/build/solc/parts/solidity/src/libsmtutil/Z3Interface.cpp:207:15:[m[K [01;36m[Knote: [m[Ksuggested alternative: ‘[01m[Kis_int[m[K’
[...]

PR #12247, opened 2021-11-05
Remove more outdated install_deps scripts.

ekpyron:
``scripts/install_deps.ps1`` we do need in CI, but the fact that these other two pull in a rather old boost version seems to indicate that we don't maintain them either, so I'd remove them as well...

I'm not sure what's the state of https://github.com/ethereum/solidity/issues/11940 and if the (apparently forever broken) snap builds even actually use ``snap/snapcraft.yaml``, so I'd tend to remove that as well actually...


ekpyron:
Following https://github.com/ethereum/solidity/issues/9582#issuecomment-962062026 I'm also removing ``install_static_z3.sh`` - it's outdated as well.

cameel (approved):

PR #12249, opened 2021-11-06
is this not 0.8.x instead of 0.x ?

phyBrackets:
https://docs.soliditylang.org/en/v0.8.9/#:~:text=We%20currently%20use%20a%200.x%20version%20number%20to%20indicate%20this%20fast%20pace%20of%20change.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,7 +21,7 @@ version of Solidity. Apart from exceptional cases, only the latest version recei
 `security fixes <https://github.com/ethereum/solidity/security/policy#supported-versions>`_.
 Furthermore, breaking changes as well as
 new features are introduced regularly. We currently use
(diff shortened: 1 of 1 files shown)


cameel (commented):
No, it was meant to be exactly what it says. It refers to the leading zero in the version number. Have you read the text that the link leads to?

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

cameel:
Well, technically we could change `0.x` to `0.y.z` if you think that makes the point clearer.

phyBrackets:
> Well, technically we could change `0.x` to `0.y.z` if you think that makes the point clearer.

aah yes , this makes more sense .

cameel:
In that case feel free to reopen the PR with that specific change.

PR #12250, opened 2021-11-07
0.y.z makes more sense than 0.x for the current version.

phyBrackets:
https://docs.soliditylang.org/en/v0.8.9/#:~:text=We%20currently%20use%20a%200.x%20version%20number%20to%20indicate%20this%20fast%20pace%20of%20change.


axic (approved):
I think this makes sense.

dzhalaevd (approved):

PR #12252, opened 2021-11-07
Include picosha2 in the license header

axic:


ekpyron (approved):
Actually, should these contain links to the original sources? Not necessary, but I'd find it annoying if I ever wanted to look any such project up that there *aren't* links in here (even if they may fall out of date eventually).

chriseth (approved):

axic:
@ekpyron I would put that link into the file and not in the license header.

ekpyron:
Why :-)? I'd rather actually *not* put the files in the license header :-D.
But it doesn't matter too much anyways :-).

PR #10240, opened 2020-11-09
Immutables with literal values are pure.

chriseth:
Fixes https://github.com/ethereum/solidity/issues/9554

PR #12256, opened 2021-11-08
Clarify ``pure``.

chriseth:


ekpyron (approved):
Well, it's an improvement...
I'm still wondering about ``msg.data``... for external calls, that's fine, but internal pure functions are really not pure anymore, if they read ``msg.data``...

In any case, we should then disallow ``codecopy`` and check if there's anything else that still violates this.

chriseth:
Agreed! Can you create an issue and schedule for 0.9.x?

ekpyron:
> Agreed! Can you create an issue and schedule for 0.9.x?

Do we have any way to actually schedule anything for 0.9.x? Maybe we should, but I wouldn't see any. Created https://github.com/ethereum/solidity/issues/12260 in any case.

docs/contracts/functions.rst
```diff
@@ -222,6 +222,9 @@ Pure Functions
 --------------
 
 Functions can be declared ``pure`` in which case they promise not to read from or modify the state.
+In particular, it should be possible to evaluate a ``pure`` function at compile-time given
+only its inputs and ``msg.data``, but without any knowledge of the current blockchain state.
+This means that reading from ``immutable`` variables can be a non-pure operation.
```
axic, line 227: Hmm, should we disallow `immutable` in `pure`? :)

axic: Ok, it seems my review was stuck from a few hours ago ^^

But I see that now @ekpyron opened an issue for this.

ekpyron: immutables are already view, I think that's what prompted the change here

axic: Well than the text here is confusing as it suggests they are still allowed in pure.

ekpyron: It does say it's a "non-pure operation", not sure how that's unclear :-)!

axic: It says "**can be** a non-pure". So in some cases it "can be pure"?

ekpyron: You mean it should have said ``is`` not ``can be`` :-)? Maybe, yeah...

ekpyron: Actually, whenever reading an immutable **can be*** pure, it's actually a constant... except that we don't properly evaluate those :-D.

chriseth: We do: https://github.com/ethereum/solidity/pull/10240

Issue #12260, opened 2021-11-08
Disallow ``codecopy`` in pure functions (and check for other cases)


ekpyron:
Related to https://github.com/ethereum/solidity/issues/8153 and following https://github.com/ethereum/solidity/pull/12256 we should make sure we actually are as strict as we want to be for ``pure`` functions.

In particular, we should disallow ``codecopy`` with 0.9.0, but we should double-check if there is other cases that we should strengthen as well.

chriseth:
Isn't disallowing access to `msg.data` also part of this issue?

ekpyron:
I mean, I can also just do it right away, maybe that's safest for not missing it again.

ekpyron:
> Isn't disallowing access to `msg.data` also part of this issue?

If we agree that we should, then yes, absolutely :-)! I'm all for that, but I wasn't sure we had consensus about that.

ekpyron:
Hm... I'm just looking through the instruction list... and hit ``CALLDATALOAD`` :-)...
That one we can't really disallow I guess... but if we allow it, we might as well keep ``msg.data``...

The problem is still that "externally pure" is different from "internally pure". An external function call I can compile-time evaluate, even if it involves msg.data and ``calldataload`` - but an internal one I cannot...

axic:
Well basically `external pure` can have different rules, than `private/public/internal`. We had another issue for tracking the memory-mutability of pure, that should be also revived.

ekpyron:
Hm, yeah... I actually thought we couldn't just use different rules due to ``public`` being both internal and external, but it actually makes perfect sense to apply the stricter ``internal`` rules to ``public``...

Anyways, I pushed #12261 for the obvious cases I saw that shouldn't be ``pure``, i.e. ``codecopy`` and ``codesize`` - the rest we can do, once we decide if to split external and internal analysis or what else to do.

chriseth:
What about internal functions that have `calldata` parameters? Can they use inline assembly and thus calldataload to access their parameters?

ekpyron:
Yeah - which is a problem. If we allow loading *something* from calldata, we effectively allow loading *anything* from calldata and thus might as well keep ``msg.data`` pure...
Then again, I'm not sure if we should ever even try to actually compile-time-evaluate inline assembly anyways...

[...]

PR #12261, opened 2021-11-08
[BREAKING] Disallow codecopy and codesize in pure functions.

ekpyron:
Obvious part of https://github.com/ethereum/solidity/issues/12260.

Not entirely sure about ``msg.data`` and ``calldataload``/``calldatasize``, so I'll keep them as is for now and the issue open.

### DIFF
--- a/libevmasm/SemanticInformation.cpp
+++ b/libevmasm/SemanticInformation.cpp
@@ -351,6 +351,8 @@ bool SemanticInformation::invalidInPureFunctions(Instruction _instruction)
 	case Instruction::ORIGIN:
 	case Instruction::CALLER:
 	case Instruction::CALLVALUE:
+	case Instruction::CODESIZE:
+	case Instruction::CODECOPY:
 	case Instruction::CHAINID:
 	case Instruction::BASEFEE:
 	case Instruction::GAS:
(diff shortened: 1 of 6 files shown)


libevmasm/SemanticInformation.cpp
```diff
@@ -351,6 +351,8 @@ bool SemanticInformation::invalidInPureFunctions(Instruction _instruction)
 	case Instruction::ORIGIN:
```
ekpyron, line 351: This list could use a sort by the way... it's neither alphabetic nor opcode value... but I didn't do it here, because that'll only create conflicts later :-).

Marenz (approved):

chriseth (commented):
Does this not need documentation updates?

ekpyron:
> Does this not need documentation updates?

Didn't you just update the documentation towards this end in https://github.com/ethereum/solidity/pull/12256 ;-)? This just actually does what that one said... although maybe I missed an explicit mention of this being pure somewhere...

chriseth:
Yeah, we don't list the opcodes, but shouldn't we at least list those that are not totally clear? Especially since going from "immutables" to "codecopy" is a (albeit small) mental step.

dzhalaevd (approved):

chriseth (approved):

PR #11000, opened 2021-02-23
Fix isDynamicallyEncoded() for structs containing internal types

cameel:
Fixes  #10516.

This PR changes the implementation of `isDynamicallyEncoded()` for structs to just use the member types directly instead of their `interfaceType()`. This does not seem to be causing any problems but I don't know why `interfaceType()` was used there in the first place.

Issue #10516, opened 2020-12-07
ICE in `StructType::isDynamicallyEncoded()` when a function has a `memory` struct argument with an internal type inside


agroce:
## Description

This contract:
```solidity
contract C {
    struct S {
    	function () e;
    }
    function f(S[2] memory s1) public {
    }
}
```

produces:

```
Internal compiler error during compilation:
/home/user/solidity/libsolidity/ast/Types.cpp(2321): Throw in function virtual bool solidity::frontend::StructType::isDynamicallyEncoded() const
Dynamic exception type: boost::exception_detail::clone_impl<solidity::langutil::InternalCompilerError>
std::exception::what: 
[solidity::util::tag_comment*] = 
```

when compiled with solc.

Looks like all other deaths in isDynamicallyEncoded were fixed, so this is a new variant (the structure looks different to me at a glance).

On master, using AFL fuzzing.  Another discovery using https://github.com/agroce/afl-compiler-fuzzer.

## Environment

- Compiler version:  0.7.6-develop.2020.11.30+commit.7e7a42c6.Linux.clang
- Target EVM version (as per compiler settings): N/A
- Framework/IDE (e.g. Truffle or Remix): N/A
- EVM execution environment / backend / blockchain client: N/A
- Operating system: Ubuntu 18.04 in docker

## Steps to Reproduce

Above shows pretty clearly, I think.

cameel:
The crash happens because `TypeChecker::visit(FunctionDefinition)` does not stop when it detects `TypeError 4103: Internal type is not allowed for public or external functions.`. It continues to visit function parameters and in case of an array, it ends up calling `isDynamicallySized()` on the base type - which is not allowed when the type is an internal one.

If we change the type to a struct (rather than an array of structs) it does not crash. Apparently `isDynamicallySized()` does not get called in that case. It also does not crash if I change the example so that there's a single struct that contains an array of function pointers.

So the question is: does `isDynamicallySized()` really need to use an external type? I see that `interfaceType()` is used by it only in case of structs. For example we just use the base type directly in case of arrays. I looked at what `interfaceType()` does an it looks like it either returns an unchanged type, creates a copy of the type as a `memory` type or returns an error (if there's no external type). In my opinion none of these things can affect whether the type is dynamically sized or not.

So my first attempt at fixing this would be to simply use the base type directly. I'm going to submit a PR that does this.

If this is not a good solution then another one would be to just make it a fatal error and stop the analysis before we analyze the parameters.

[...]

PR #12266, opened 2021-11-10
ICE in StructType::isDynamicallyEncoded() when a function has a calldata struct argument with an internal type inside :: Issue 11610

### LINKED ISSUE #11610: ICE in `StructType::isDynamicallyEncoded()` when a function has a `calldata` struct argument with an internal type inside
## Environment

- Compiler version: Version: 0.8.7-develop.2021.7.4+commit.f6cb933f.Linux.clang
- Target EVM version (as per compiler settings):
- Framework/IDE (e.g. Truffle or Remix):
- Operating system: Linux 5.4.0-77-generic x86_64 GNU/Linux

#### Test program: small.sol
```solidity
contract C {
  struct S {
    function() a;
  }
  function f(S[2] calldata) {}
}
```
#### Compiler Command
```bash
solc-trunk-asan_ubsan --optimize small.sol
```
#### Output
```
Internal compiler error during compilation:
/solidity/libsolidity/ast/Types.cpp(2149): Throw in function virtual bool solidity::frontend::StructType::isDynamicallyEncoded() const
std::exception::what: 
```

nishant-sachdeva:
TypeChecker.cpp :: Replced typeError with fataTypeError in lines 390 and 406. 
Test file added in test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
Relevant line added to Changelog.md

Closes #11610

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,7 @@ Bugfixes:
  * TypeChecker: Improved error message for constant variables with (nested) mapping types.
  * Yul Assembler: Fix internal error when function names are not unique.
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
+ * ICE in StructType::isDynamicallyEncoded() when a function has a calldata struct argument with an internal type inside
 
 
 Build System:
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -387,7 +387,7 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
 			_var.referenceLocation() == VariableDeclaration::Location::Storage &&
 			!m_currentContract->abstract()
 		)
-			m_errorReporter.typeError(
+			m_errorReporter.fatalTypeError(
 				3644_error,
 				_var.location(),
 				"This parameter has a type that can only be used internally. "
@@ -403,7 +403,7 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
 				solAssert(!message.empty(), "Expected detailed error message!");
 				if (_function.isConstructor())
 					message += " You can make the contract abstract to avoid this problem.";
-				m_errorReporter.typeError(4103_error, _var.location(), message);
+				m_errorReporter.fatalTypeError(4103_error, _var.location(), message);
 			}
 			else if (
 				!useABICoderV2() &&
--- a/test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
+++ b/test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
@@ -1,0 +1,9 @@
+contract C {
+  struct S {
+    function() a;
+  }
+  function f(S[2] calldata) {}
+}
+// ----
+// SyntaxError 4937: (50-78): No visibility specified. Did you intend to add "public"?
+// TypeError 4103: (61-74): Internal type is not allowed for public or external functions.


Changelog.md
```diff
@@ -29,6 +29,7 @@ Bugfixes:
  * TypeChecker: Improved error message for constant variables with (nested) mapping types.
  * Yul Assembler: Fix internal error when function names are not unique.
  * Yul IR Generator: Do not output empty switches/if-bodies for empty contracts.
+ * ICE in StructType::isDynamicallyEncoded() when a function has a calldata struct argument with an internal type inside
```
cameel, line 32: Please rebase on latest `develop`. We have just relased 0.8.10 yesterday so this should be in the 0.8.11 section :)

test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
```diff
@@ -0,0 +1,9 @@
+contract C {
+  struct S {
+    function() a;
+  }
+  function f(S[2] calldata) {}
+}
```
cameel, line 6:
```suggestion
contract C {
    struct S {
        function() a;
    }
    function f(S[2] calldata) {}
}
```

test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
```diff
@@ -0,0 +1,9 @@
+contract C {
```
cameel, line 1: Unless the test is really unique in some way, I think it would be better to find a subdirectory where it fits best instead of dropping it directly into `syntaxTests/`.

I'd place it next to the test from #10516.

Changelog.md
cameel, line 32: Does it happen for `calldata` structs or only for arrays of structs? Can you add a test case covering single structs (unless there's one already, please check)?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -387,7 +387,7 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
 			_var.referenceLocation() == VariableDeclaration::Location::Storage &&
 			!m_currentContract->abstract()
 		)
-			m_errorReporter.typeError(
+			m_errorReporter.fatalTypeError(
```
cameel, line 390: This generally seems fine to me but please check if it shouldn't rather be fixed in the same place as #11000 instead.

cameel (commented):
Congrats on the first PR!

I added some remarks for specific places in code in the comments below. Here are some general ones:
- I'd avoid making PRs from `develop` branch in your fork. It just makes it harder to update it or make multiple PRs. Please use named branches instead.
- Please use some sensible title for the PR (and a similar name for the branch). We generally don't know all issue numbers by heart so it's hard to say what it's about without looking it up :P In branch names the issue number also would make it annoying to type on the CLI because you have to type the number before you get autocompletion. Also no need to say "Draft" in the title.

cameel:
About your question from the call: you need to update test expectations (`isoltest` can do it automatically), which will remove the error from some tests but after doing that please make sure that every test still makes sense. If stopping early makes it ineffective at testing what it was supposed to (i.e. it should produce a specific error message that no longer shows up), it may have to be adjusted.

test/libsolidity/syntaxTests/calldata_struct_argument_with_internal_type_inside.sol
```diff
@@ -0,0 +1,9 @@
...
+  struct S {
+    function() a;
+  }
+  function f(S[2] calldata) {}
+}
+// ----
+// SyntaxError 4937: (50-78): No visibility specified. Did you intend to add "public"?
```
chriseth, line 8: Please make the function public (`function <functionName>() public ...`) so that this error goes away - it is not relevant for the test.

Changelog.md
nishant-sachdeva: @cameel  You've mentioned in #10516  that this works for a struct but not for an array of structs. 
This is because isDynamicallyEncoded() does not get called in the first case case.

However, I've added a test case to check for a single struct. Will push in the next commit.

PR #12267, opened 2021-11-10
Output searched locations on import failure.

### LINKED ISSUE #11042: Import errors should include full path of the file that is being imported
## Description
When using relative paths the error message often (but not always) shows only a relative path to the file. When using path remapping and base path together with various flavors of relative imports (see #11036) it might not be obvious where the compiler is looking for the file. I think the compiler should display the final, resolved filesystem path instead of (or in addition to) the relative one.

### Example 1
`/tmp/a/b/contract.sol`
```solidity
import "relative.sol";
```
```bash
cd /tmp/c/d/
solc /tmp/a/b/contract.sol --allow-paths .
```
Here the compiler will look for the file in `/tmp/c/d/relative.sol`, not `/tmp/a/b/relative.sol` as some users might (wrongly) expect. It only says `relative.sol` though:
```
Error: Source "relative.sol" not found: File not found.
 --> /tmp/a/b/contract.sol:1:1:
  |
1 | import "relative.sol";
  | ^^
```

### Example 2
`/tmp/a/b/contract.sol`
```solidity
import "/tmp/absolute.sol";
```
```bash
cd /tmp/c/d/
solc /tmp/a/b/contract.sol /tmp=/contracts --base-path /home
```

Here the compiler says that the missing import is `/contracts/absolute.sol` but omits the fact that it prepends base path to it and actually expects the file at `/home/contracts/absolute.sol`:
```
Error: Source "/contracts/absolute.sol" not found: File not found.
 --> /tmp/a/b/contract.sol:1:1:
  |
1 | import "/tmp/absolute.sol";
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The fact that the base path is prepended to an absolute path is of course a bug (reported in #11036) but showing the full path would at least allow the user to realize why his code does not compile.

## Environment

- Compiler version: 0.8.1

joshuatarkwski:
Fixes #11042

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1054,8 +1054,9 @@ jobs:
     docker:
       - image: circleci/node:<<parameters.nodejs_version>>
     resource_class: <<parameters.resource_class>>
-    # NOTE: Each external test does 3 separate compile&test runs
-    parallelism: 3
+    # NOTE: Each external test runs up to 6 independent settings presets. If parallelism is higher than
+    # actual preset count, some runs will exit immediately. If it's lower, some runs will get more than one preset.
+    parallelism: 6
     environment:
       TERM: xterm
       COMPILE_ONLY: <<parameters.compile_only>>
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -211,6 +211,7 @@ function safe_kill
 
 function circleci_select_steps
 {
+    # We expect multiple lines in $all_steps, one step per line
     local all_steps="$1"
     (( $# == 1 )) || assertFail
 
@@ -222,11 +223,26 @@ function circleci_select_steps
     fi
 }
 
+function circleci_select_steps_multiarg
+{
+    # We expect multiple arguments, one step per argument.
+    circleci_select_steps "$(printf '%s\n' "$@")"
+}
+
 function circleci_step_selected
 {
     local selected_steps="$1"
     local step="$2"
+    (( $# == 2 )) || assertFail
     [[ $step != *" "* ]] || assertFail "Step names must not contain spaces."
 
     [[ " $selected_steps " == *" $step "* ]] || return 1
+}
+
+function first_word
+{
+    local words="$1"
+    (( $# == 1 )) || assertFail
+
+    echo "$words" | cut -d " " -f 1
 }
--- a/test/externalTests/colony.sh
+++ b/test/externalTests/colony.sh
@@ -36,20 +36,25 @@ function colony_test
     local repo="https://github.com/solidity-external-tests/colonyNetwork.git"
     local branch=develop_080
     local config_file="truffle.js"
-    # On levels 1 and 2 it compiles but tests run out of gas
-    local min_optimizer_level=3
-    local max_optimizer_level=3
+    local settings_presets=(
+        #ir-no-optimize            # Compiles but tests run out of gas
+        #ir-optimize-evm-only      # Compiles but tests run out of gas
+        ir-optimize-evm+yul
+        #legacy-no-optimize        # Compiles but tests run out of gas
+        #legacy-optimize-evm-only  # Compiles but tests run out of gas
+        legacy-optimize-evm+yul
+    )
 
-    local selected_optimizer_levels
-    selected_optimizer_levels=$(circleci_select_steps "$(seq "$min_optimizer_level" "$max_optimizer_level")")
-    print_optimizer_levels_or_exit "$selected_optimizer_levels"
+    local selected_optimizer_presets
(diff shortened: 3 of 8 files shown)


libsolidity/interface/FileReader.cpp
```diff
@@ -113,19 +113,23 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
 		vector<reference_wrapper<boost::filesystem::path>> prefixes = {m_basePath};
 		prefixes += (m_includePaths | ranges::to<vector<reference_wrapper<boost::filesystem::path>>>);
 
+		vector<boost::filesystem::path> searchedPaths;
```
cameel: Now that we have this list, `candidates` could be converted to a list of references to `searchedPaths`. No need to store full copies.

libsolidity/interface/FileReader.cpp
```diff
@@ -113,19 +113,23 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
 
 		auto pathToQuotedString = [](boost::filesystem::path const& _path){ return "\"" + _path.string() + "\""; };
 
 		if (candidates.empty())
-			return ReadCallback::Result{false, "File not found."};
-
+			return ReadCallback::Result{
+				false,
+				"File not found. Searched the following locations: " +
+				joinHumanReadable(searchedPaths | ranges::views::transform(pathToQuotedString), ", ")
```
cameel:
```suggestion
				joinHumanReadable(searchedPaths | ranges::views::transform(pathToQuotedString), ", ") +
				"."
```

libsolidity/interface/FileReader.cpp
```diff
@@ -113,19 +113,23 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
...
+			searchedPaths.push_back(canonicalPath);
 			if (boost::filesystem::exists(canonicalPath))
 				candidates.push_back(std::move(canonicalPath));
 		}
 
 		auto pathToQuotedString = [](boost::filesystem::path const& _path){ return "\"" + _path.string() + "\""; };
 
 		if (candidates.empty())
-			return ReadCallback::Result{false, "File not found."};
-
```
cameel: The extra whitespace was here for readability. No need to remove it.

cameel:
I'd recommend not submitting PRs from `develop` in your fork. It makes it harder to pull changes from the main repo and you can't use it for more than one PR at a time anyway.

cameel:
I think that the `"File outside of allowed directories."` and `"Not a valid file."` should print the canonical path too. For allowed directories we could list them as well.

libsolidity/interface/FileReader.cpp
```diff
@@ -113,19 +113,23 @@ ReadCallback::Result FileReader::readFile(string const& _kind, string const& _so
 		vector<reference_wrapper<boost::filesystem::path>> prefixes = {m_basePath};
 		prefixes += (m_includePaths | ranges::to<vector<reference_wrapper<boost::filesystem::path>>>);
 
+		vector<boost::filesystem::path> searchedPaths;
 		for (auto const& prefix: prefixes)
 		{
 			boost::filesystem::path canonicalPath = normalizeCLIPathForVFS(prefix / strippedSourceUnitName, SymlinkResolution::Enabled);
-
+			searchedPaths.push_back(canonicalPath);
```
cameel: This adds the whole path, including file name, not just the directory. The directory is just `prefix`. But if you want to print the prefixes, you don't even need `searchedPaths`. You can just use `prefixes`.

cameel:
> Merge branch 'develop' of https://github.com/joshuatarkwski/solidity into develop

Just a note on our PR workflow: [we're generally using a rebase workflow](https://docs.soliditylang.org/en/latest/contributing.html#workflow-for-pull-requests). Instead of merging `develop` into your branch, please rebase it on `develop` instead.

libsolidity/interface/FileReader.cpp
chriseth: `searchedPaths` also has `normalizeCLIPathForVFS` - can we just ignore it (should we)?

cameel: Right. Depends on whether we prefer to print paths as the user specified them on the CLI or the normalized versions we actually use. I guess the latter might be more helpful in case the path had `.`, `..`, `~` in it or had symlinks and the user expected them to be resolved (they're not).

So yeah, in that case `searchedPaths` is fine but we could also just add a lambda that converts them when the message is generated instead of storing them.

In any case the source unit names should not be included.

PR #12202, opened 2021-10-27
Merge `develop` into `breaking`

cameel:
The main reason for updating the branch is that I want to see if the `t_ems_ext_hardhat` job added in #12165 will pass on `breaking` too.

PR #12269, opened 2021-11-10
Fix gitter notifications for non-PR runs in CI

cameel:


cameel:
I think that the problem is just another consequence of https://github.com/ethereum/solidity/pull/12202#issuecomment-952950518. I still have a support request open with CircleCI about that one so I sent them an update but fixing this apparently requires some architectural changes and the overall impact on users is pretty low so it's far behind other more important things in their backlog. We probably will have to live with the workaround for some time. I think it's not that bad as long as we don't have many long-lived non-PR branches.

Marenz (approved):

PR #12274, opened 2021-11-11
Re-use knowledge about keccak calls.

chriseth:
Solves the second part of https://github.com/ethereum/solidity/issues/12211


chriseth:
This is essentially a common subexpression eliminator, but for non-movable functions. Maybe we can extend this in some more generic way? Of course the idea here is also that we have specialized knowledge about which mstore opcode destroys which keccak call.

libyul/optimiser/DataFlowAnalyzer.h
```diff
@@ -125,7 +125,8 @@ class DataFlowAnalyzer: public ASTModifier
 	/// i.e. `_otherStorage` and `_otherMemory` cannot have additional changes.
```
cameel:
```suggestion
	/// i.e. `_olderStorage` and `_olderMemory` cannot have additional changes.
```

libyul/optimiser/DataFlowAnalyzer.h
```diff
@@ -171,6 +176,8 @@ class DataFlowAnalyzer: public ASTModifier
 
 	std::unordered_map<YulString, YulString> m_storage;
 	std::unordered_map<YulString, YulString> m_memory;
+	/// If m_keccak[s, l] = y then y := keccak256(s, l) occurs in the code.
+	std::map<std::pair<YulString, YulString>, YulString> m_keccak;
```
cameel: What happens if something like this occurs in the code?
```solidity
x := keccak256(a, b)
y := keccak256(a, b)
```

Shouldn't key and value be reversed in this map?

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -310,6 +323,12 @@ void DataFlowAnalyzer::clearValues(set<YulString> _variables)
 	});
 	cxx20::erase_if(m_storage, eraseCondition);
 	cxx20::erase_if(m_memory, eraseCondition);
+	cxx20::erase_if(m_keccak, [&_variables](auto&& _item) {
+		return
+			_variables.count(_item.first.first) ||
+			_variables.count(_item.first.second) ||
+			_variables.count(_item.second);
+	});
```
cameel: Why not use `mapTuple()` here as well?

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -345,16 +367,26 @@ void DataFlowAnalyzer::clearKnowledgeIfInvalidated(Expression const& _expr)
...
-	unordered_map<YulString, YulString> const& _olderMemory
+	unordered_map<YulString, YulString> const& _olderMemory,
+	map<pair<YulString, YulString>, YulString> const& _olderKeccak
 )
 {
 	joinKnowledgeHelper(m_storage, _olderStorage);
 	joinKnowledgeHelper(m_memory, _olderMemory);
+
+	// TODO is this correct?
+	cxx20::erase_if(m_keccak, mapTuple([&_olderKeccak](auto&& key, auto&& currentValue){
```
cameel:
```suggestion
	cxx20::erase_if(m_keccak, mapTuple([&_olderKeccak](auto&& key, auto&& currentValue) {
```

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -310,6 +323,12 @@ void DataFlowAnalyzer::clearValues(set<YulString> _variables)
 	});
 	cxx20::erase_if(m_storage, eraseCondition);
 	cxx20::erase_if(m_memory, eraseCondition);
```
cameel: This is not a part of this PR but just a question to understand this better: are the strings in `m_storage` and `m_memory` guaranteed to be identifiers? This condition will only invalidate a straight `let b := a` assignment and not for example something like `let b := f(a)` so if they could be more complex expressions, the analyzer would probably not work correctly.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -253,6 +261,9 @@ void DataFlowAnalyzer::handleAssignment(set<YulString> const& _variables, Expres
 			m_memory.erase(name);
 			// assignment to slot contents denoted by "name"
 			cxx20::erase_if(m_memory, mapTuple([&name](auto&& /* key */, auto&& value) { return value == name; }));
+			cxx20::erase_if(m_keccak, [&name](auto&& _item) {
+				return _item.first.first == name || _item.first.second == name || _item.second == name;
+			});
```
cameel: Again, why not `mapTuple()`?

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -345,16 +367,26 @@ void DataFlowAnalyzer::clearKnowledgeIfInvalidated(Expression const& _expr)
...
 )
 {
 	joinKnowledgeHelper(m_storage, _olderStorage);
 	joinKnowledgeHelper(m_memory, _olderMemory);
+
+	// TODO is this correct?
+	cxx20::erase_if(m_keccak, mapTuple([&_olderKeccak](auto&& key, auto&& currentValue){
+		YulString const* oldValue = valueOrNullptr(_olderKeccak, key);
+		return !oldValue || *oldValue != currentValue;
+	}));
```
cameel: So, after `joinKnowledge` `m_keccak` will contain only the keys whose values remained unchanged and are present in both old and new state? Basically, `if` branches and `switch` cases can invalidate parts of state but cannot add anything that would be still useful outside of them?

In case of `if` it sounds about right because there's no `else` in Yul but for `switch` I think it could be more permissive. I.e. if we have the same keccak calculation in all cases (including the default one), we should keep it.

Though it probably does not matter much if even the switch prefix mover (#10742) turned out to be not very useful in practice.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -135,8 +141,9 @@ void DataFlowAnalyzer::operator()(Switch& _switch)
 	{
 		unordered_map<YulString, YulString> storage = m_storage;
 		unordered_map<YulString, YulString> memory = m_memory;
+		map<pair<YulString, YulString>, YulString> keccak = m_keccak;
 		(*this)(_case.body);
-		joinKnowledge(storage, memory);
+		joinKnowledge(storage, memory, keccak);
```
cameel: This is suprising. It does not preserve the precondition stated in the `joinKnowledge()` docstring:

```
/// This only works if the current state is a direct successor of the older point,
/// i.e. `_otherStorage` and `_otherMemory` cannot have additional changes.
```

The state after each case is applied on top of the state after the previous case passed through join, not on top of the initial state from before the `switch`. Which I guess is fine for keccak seeing how `joinKnowledge()` is implemented now  but not sure about memory and storage (I haven't checked how the join helper works for them). Maybe the docstring should be updated to say that the state returned from `joinKnowledge()` is guaranteed to be reusable that way?

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -77,6 +80,8 @@ void DataFlowAnalyzer::operator()(ExpressionStatement& _statement)
 		cxx20::erase_if(m_memory, mapTuple([&](auto&& key, auto&& /* value */) {
 			return !m_knowledgeBase.knownToBeDifferentByAtLeast32(vars->first, key);
 		}));
+		// TODO erase keccak knowledge, but in a more clever way
+		m_keccak = {};
```
cameel: Ouch, so writing anything anywhere in memory just discards all the info about keccak? I'd expect it to limit the usefulness of this step a lot. Basically it will not help much beyond just situations where there are keccak calls just next to each other?

libyul/optimiser/DataFlowAnalyzer.h
cameel: Weird, this is actually a situation from your test case and this does not seem to be a problem. Not sure why.

test/libyul/yulOptimizerTests/loadResolver/keccak_reuse_basic.yul
```diff
@@ -0,0 +1,17 @@
+{
+    let x := calldataload(0)
+    let a := keccak256(0, x)
+    sstore(a, 2)
+    let t := mload(2)
+    let b := keccak256(0, x)
+    sstore(b, 3)
+}
```
cameel, line 8: I'd add a few more tests to be able to see if the parts changed in this PR actually work correctly:
- One with `mstore` between keccak calls.
- One like the example above but with `a` ressigned to a different value after the first `sstore`.
- keccak used in an expression (e.g. `add(keccak256(a, b), keccak256(a, b))`).
- keccak used in an expression that invalidates the knowledge about keccak calls (e.g. `mstore(keccak256(a, b), keccak256(a, b))`).
- Same keccak call in all cases of a `switch`.
- Same keccak call followed by `mstore` in all cases of a `switch`.
- A call to a function containing something that invalidates memory in between same keccak calls.
- Code that contains `msize`.

libyul/optimiser/DataFlowAnalyzer.h
cameel: Class docstring should be updated to mention keccak. Looks like [`DataflowAnalyzer`](https://docs.soliditylang.org/en/latest/internals/optimizer.html#dataflowanalyzer) is also in the docs so I'd update that too.

libyul/optimiser/LoadResolver.cpp
```diff
@@ -64,6 +65,22 @@ void LoadResolver::visit(Expression& _e)
...
+		if (
+			!m_containsMSize &&
+			toEVMInstruction(m_dialect, funCall->functionName.name) == evmasm::Instruction::KECCAK256
+		)
+		{
+			Identifier const* start = get_if<Identifier>(&funCall->arguments.at(0));
+			Identifier const* length = get_if<Identifier>(&funCall->arguments.at(1));
+			if (start && length)
+				if (auto value = util::valueOrNullptr(m_keccak, make_pair(start->name, length->name)))
+					if (inScope(*value))
```
cameel, line 72: Oh, so `DataflowAnalyzer` does not clear the variables after exiting a scope/block? Is this for a situation this?
```solidity
{
    let a := 42;
    let b := 42;
    sstore(0, keccak256(a, b))
}
{
    let x := 42;
    let y := 42;
    sstore(0, keccak256(x, y))
}
```
If not, can you add a test that actually fails this condition?

cameel:
> This is essentially a common subexpression eliminator, but for non-movable functions. Maybe we can extend this in some more generic way? Of course the idea here is also that we have specialized knowledge about which mstore opcode destroys which keccak call.

Is the only side-effect of keccak that it writes to some scratch space in memory during computation? Do we have other builtins that are almost "side-effect free but not quite" this way? If so, we could maybe generalize it to all of them.

libyul/optimiser/DataFlowAnalyzer.cpp
chriseth: I think we should have a call about this.

libyul/optimiser/DataFlowAnalyzer.cpp
chriseth: Yes, that's why there is the TODO.

chriseth:
No, keccak only reads from memory. Keccak is side-effect free, but its value depends on the memory contents.

hrkrshnn:
Will rebase this.

leonardoalt:
So is this just waiting for review?

ekpyron:
Replacing ``stale`` by ``takeover`` for now here, let's see if that helps.

libyul/optimiser/LoadResolver.cpp
chriseth: The problem is in the value:
```
let a := 0
let b := 32
{
  let c := keccak(a, b)
}
let d := keccak(a, b)
```
We cannot replace the RHS of the last line by `c` because it is not in scope. The DataFlowAnalyzer might erase the knowledge, but I think this is good to be here, just to be sure.

test/libyul/yulOptimizerTests/loadResolver/keccak_reuse_basic.yul
chriseth: Couldn't find a good example for the switch case, but added tests for most of these scenarios.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -86,6 +89,8 @@ void DataFlowAnalyzer::operator()(ExpressionStatement& _statement)
 			cxx20::erase_if(m_state.memory, mapTuple([&](auto&& key, auto&& /* value */) {
 				return !m_knowledgeBase.knownToBeDifferentByAtLeast32(vars->first, key);
 			}));
+			// TODO erase keccak knowledge, but in a more clever way
```
chriseth, line 90: Will be done in a future PR when we optimize the knowledge base.

chriseth:
The code
```
struct S { uint8 a; uint8 b; }
contract C {
    S s;
    function f(S calldata _s) external { s = _s; }
}
```
causes
```
Unknown exception during compilation: Dynamic exception type: std::out_of_range
std::exception::what: vector
```
on emscripten (with optimizer).

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -368,7 +394,10 @@ void DataFlowAnalyzer::clearKnowledgeIfInvalidated(Expression const& _expr)
 	if (sideEffects.invalidatesStorage())
 		m_state.environment.storage.clear();
 	if (sideEffects.invalidatesMemory())
+	{
 		m_state.environment.memory.clear();
+		m_state.environment.	keccak.clear();
```
chriseth:
```suggestion
		m_state.environment.keccak.clear();
```

chriseth:
Fund the bug. Reference tracking would have saved me 2 hours. ;)

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -416,12 +445,26 @@ std::optional<YulString> DataFlowAnalyzer::isSimpleLoad(
 	return {};
 }
 
+optional<pair<YulString, YulString>> DataFlowAnalyzer::isKeccak(Expression const& _expression) const
+{
+	if (FunctionCall const* funCall = get_if<FunctionCall>(&_expression))
+		if (toEVMInstruction(m_dialect, funCall->functionName.name) == evmasm::Instruction::KECCAK256)
```
chriseth: Will change this to use `Dialect::hashFunction` as well.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -314,7 +331,7 @@ void DataFlowAnalyzer::clearValues(set<YulString> _variables)
 	// let a := 1
 	// let b := a
 	// let c := b
-	// let a := 2
+	// a := 2
```
chriseth, line 332: This was wrong all the time :D

ekpyron (commented):
Code looks good; haven't verified test coverage yet; at least needs a Changelog entry.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -237,6 +239,14 @@ optional<YulString> DataFlowAnalyzer::memoryValue(YulString _key) const
 		return nullopt;
 }
 
+optional<YulString> DataFlowAnalyzer::keccakValue(YulString _start, YulString _length) const
+{
+	if (YulString const* value = util::valueOrNullptr(m_state.environment.keccak, make_pair(_start, _length)))
```
nikola-matic:
```suggestion
	if (YulString const* value = valueOrNullptr(m_state.environment.keccak, make_pair(_start, _length)))
```
There's also `valueOrDefault` in `CommonData.h` (if this is even where `util::valueOrNullptr` is even implemented that could potentially simplify this function by just having
```
return valueOrDefault(m_state.environment.keccak, make_pair(_start, _length), nullopt);
```
Note that I haven't actually looked into this at great detail, so I may be wrong :)

chriseth: It at least doesn't work right away, because the vale is a YulString and in the positive case we would have to convert it to `option<YulString>` first.

nikola-matic: Ah yes, true. In that case, just get rid of the `util::` qualifier for consistency.

chriseth:
Removed util prefix.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -223,15 +225,23 @@ void DataFlowAnalyzer::operator()(Block& _block)
 
 optional<YulString> DataFlowAnalyzer::storageValue(YulString _key) const
 {
-	if (YulString const* value = util::valueOrNullptr(m_state.environment.storage, _key))
+	if (YulString const* value = valueOrNullptr(m_state.environment.storage, _key))
```
ekpyron, line 228: Probably makes sense to add something like ``valueOrNullopt`` to do just this as well (the main reason it's not there is that it has to copy, which may not always be desirable, but we can still have it for cases for which copying doesn't matter).

But no point in delaying this PR for that.

chriseth: Sounds like rust, then :D

test/libyul/yulOptimizerTests/loadResolver/keccak_reuse_mstore.yul
```diff
@@ -0,0 +1,21 @@
+{
+    let x := calldataload(0)
+    let a := keccak256(0x200, x)
+    sstore(a, 2)
+    // will disable loading for now, might improve later
+    mstore(0, 1)
+    let b := keccak256(0x20, x)
```
ekpyron, line 7: Is this intentionally ``0x20`` instead of ``0x200``.

chriseth: Probbaly not, will fix.

test/libyul/yulOptimizerTests/loadResolver/keccak_reuse_expr_mstore.yul
```diff
@@ -0,0 +1,28 @@
+{
+    let a := calldataload(0)
+    sstore(f(keccak256(0, a)), keccak256(0, a))
+    sstore(f(keccak256(0, a)), keccak256(0, a))
```
ekpyron, line 4: Shouldn't we also test the opposite case, i.e.
``sstore(keccak256(0, a), f(keccak256(0, a)))``
?

test/libyul/yulOptimizerTests/loadResolver/keccak_reuse_reassigned_branch.yul
```diff
@@ -0,0 +1,30 @@
...
+    let x := calldataload(0)
+    let y := calldataload(1)
+    let a := keccak256(x, y)
+    if calldataload(2) {
+        a := 8
+    }
+    let b := keccak256(x, y)
+    sstore(b, 2)
+    if calldataload(3) {
+        x := 8
```
ekpyron, line 11: wouldn't hurt to do the same thing with only ``y`` instead of only ``x``, but yeah, also not crucial...

ekpyron (commented):
Added a few more tweaks to the test cases myself.

Now the only thing left I see missing is a Changelog entry :-). And squashing. And maybe doesn't hurt to check if it's worth a quick mention in the internal optimiser docs.

chriseth:
Also wouldn't hurt to add a bit of fuzzing here...

Changelog.md
```diff
@@ -11,6 +11,7 @@ Compiler Features:
  * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
  * Yul EVM Code Transform: Generate more optimal code for user-defined functions that always terminate a transaction. No return labels will be pushed for calls to functions that always terminate.
  * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
+ * Yul Optimizer: Eliminate `keccak256`` calls if the value was already calculated by a previous call and can be reused.
```
```suggestion
 * Yul Optimizer: Eliminate ``keccak256`` calls if the value was already calculated by a previous call and can be reused.
```

nikola-matic: Ugh, damn Clion keeps adding double quotes, and then I have to delete one every time.

ekpyron: We should consistently define whether using double or single ones in general - it's a bit of a mess right now, there's even single entries that have both :-). Didn't this even come up before already :-)?

nikola-matic: Yeah, it did. For `rst` we use double, for `md` we use single, except in `Changelog.md` where we use double, except in cases where we forgot to use double and then used single. Quite simple really.

ekpyron (approved):

ekpyron:
I see the internal optimiser docs about the data flow analyzer are vague enough, s.t. not mentioning this, doesn't make things any worse - we could improve them eventually, but don't need to bother with it here.

Changelog.md
ekpyron: Luckily, we don't keep a record of such glorious mechanisms :-). Or wait, I see that we indeed do...

**Use double backticks in RST (``` ``x`` ```). Prefer single backticks in Markdown (`` `x` ``),**
but note that double backticks are valid too and we use them in some cases for legacy reasons.

So yeah, use either one or both - extremely helpful that we cleared that up :-D.

cameel: Should we change that to "always use single backticks in Markdown" and convert all existing .md files?

PR #12276, opened 2021-11-12
Yul interpreter: Changes required before merging Redundant store eliminator

bshastry:
Required to fuzz #11352

### DIFF
--- a/test/libyul/EwasmTranslationTest.cpp
+++ b/test/libyul/EwasmTranslationTest.cpp
@@ -108,13 +108,18 @@ string EwasmTranslationTest::interpret()
 	state.maxExprNesting = 64;
 	try
 	{
-		Interpreter::run(state, WasmDialect{}, *m_object->code);
+		Interpreter::run(
+			state,
+			WasmDialect{},
+			*m_object->code,
+			/*disableMemoryTracing=*/false
+		);
 	}
 	catch (InterpreterTerminatedGeneric const&)
 	{
 	}
 
 	stringstream result;
-	state.dumpTraceAndState(result);
+	state.dumpTraceAndState(result, false);
 	return result.str();
 }
--- a/test/libyul/YulInterpreterTest.cpp
+++ b/test/libyul/YulInterpreterTest.cpp
@@ -94,13 +94,18 @@ string YulInterpreterTest::interpret()
 	state.maxExprNesting = 64;
 	try
 	{
-		Interpreter::run(state, EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}), *m_ast);
+		Interpreter::run(
+			state,
+			EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}),
+			*m_ast,
+			/*disableMemoryTracing=*/false
+		);
 	}
 	catch (InterpreterTerminatedGeneric const&)
 	{
 	}
 
 	stringstream result;
-	state.dumpTraceAndState(result);
+	state.dumpTraceAndState(result, false);
 	return result.str();
 }
--- a/test/tools/ossfuzz/strictasm_diff_ossfuzz.cpp
+++ b/test/tools/ossfuzz/strictasm_diff_ossfuzz.cpp
@@ -81,10 +81,15 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t const* _data, size_t _size)
 
 	ostringstream os1;
 	ostringstream os2;
+	// Disable memory tracing to avoid false positive reports
+	// such as unused write to memory e.g.,
+	// { mstore(0, 1) }
+	// that would be removed by the redundant store eliminator.
 	yulFuzzerUtil::TerminationReason termReason = yulFuzzerUtil::interpret(
 		os1,
 		stack.parserResult()->code,
-		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion())
+		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion()),
+		/*disableMemoryTracing=*/true
 	);
 	if (yulFuzzerUtil::resourceLimitsExceeded(termReason))
 		return 0;
@@ -93,7 +98,8 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t const* _data, size_t _size)
 	termReason = yulFuzzerUtil::interpret(
 		os2,
 		stack.parserResult()->code,
-		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion())
+		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion()),
+		/*disableMemoryTracing=*/true
 	);
 
 	if (yulFuzzerUtil::resourceLimitsExceeded(termReason))
--- a/test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
+++ b/test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
@@ -27,6 +27,7 @@
 #include <libyul/AST.h>
 
 #include <libevmasm/Instruction.h>
+#include <libevmasm/SemanticInformation.h>
 
 #include <libsolutil/Keccak256.h>
 #include <libsolutil/Numeric.h>
@@ -35,6 +36,7 @@
 
 using namespace std;
 using namespace solidity;
+using namespace solidity::evmasm;
 using namespace solidity::yul;
 using namespace solidity::yul::test;
 
@@ -99,6 +101,7 @@ u256 EVMInstructionInterpreter::eval(
 	switch (_instruction)
 	{
 	case Instruction::STOP:
+		logTrace(_instruction);
 		BOOST_THROW_EXCEPTION(ExplicitlyTerminated());
 	// --------------- arithmetic ---------------
 	case Instruction::ADD:
@@ -204,6 +207,7 @@ u256 EVMInstructionInterpreter::eval(
 	case Instruction::CALLDATASIZE:
 		return m_state.calldata.size();
 	case Instruction::CALLDATACOPY:
+		logTrace(_instruction, arg);
 		if (accessMemory(arg[0], arg[2]))
 			copyZeroExtended(
 				m_state.memory, m_state.calldata,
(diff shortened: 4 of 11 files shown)


test/libyul/YulInterpreterTest.cpp
```diff
@@ -94,13 +94,13 @@ string YulInterpreterTest::interpret()
 	state.maxExprNesting = 64;
 	try
 	{
-		Interpreter::run(state, EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}), *m_ast);
+		Interpreter::run(state, EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}), *m_ast, false);
```
bshastry: Memory tracing is enabled (last argument = `false`) for Yul interpreter tests. This may be changed to `true` if tests pass only without memory tracing.

test/libyul/EwasmTranslationTest.cpp
```diff
@@ -108,13 +108,13 @@ string EwasmTranslationTest::interpret()
 	state.maxExprNesting = 64;
 	try
 	{
-		Interpreter::run(state, WasmDialect{}, *m_object->code);
+		Interpreter::run(state, WasmDialect{}, *m_object->code, false);
```
bshastry: Memory tracing is enabled (last argument = `false`) for Ewasm->Yul tests. This may be changed to `true` if tests pass only without memory tracing.

test/libyul/EwasmTranslationTest.cpp
```diff
@@ -108,13 +108,13 @@ string EwasmTranslationTest::interpret()
...
 	}
 	catch (InterpreterTerminatedGeneric const&)
 	{
 	}
 
 	stringstream result;
-	state.dumpTraceAndState(result);
+	state.dumpTraceAndState(result, false);
```
bshastry, line 123: Memory tracing enabled.

test/tools/ossfuzz/strictasm_diff_ossfuzz.cpp
```diff
@@ -84,7 +84,8 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t const* _data, size_t _size)
 	yulFuzzerUtil::TerminationReason termReason = yulFuzzerUtil::interpret(
 		os1,
 		stack.parserResult()->code,
-		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion())
+		EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion()),
+		true
```
bshastry: Disable memory tracing in the fuzzer harness since it will lead to false positives like

```
{
   mstore(0, 1) // will be removed by RSE but if memory tracing is enabled, a trace MSTORE(0,1) and memory contents at address zero will appear in the unoptimized run but not in the optimized run
}

File: test/tools/yulInterpreter/EVMInstructionInterpreter.cpp [hunk 5]
```diff
@@ -204,6 +205,7 @@ u256 EVMInstructionInterpreter::eval(
 	case Instruction::CALLDATASIZE:
 		return m_state.calldata.size();
 	case Instruction::CALLDATACOPY:
+		logTrace(_instruction, true, arg);
```
bshastry: `logTrace(instruction, true, arg)` means interpret instruction `I` whose args are `A...` and tell the tracer that this instruction may write to memory.

If the interpreter is configured to disable memory tracing, this information (whether an instruction writes to memory or not) is useful in deciding whether to append the instruction to execution trace or not.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -99,6 +99,7 @@ u256 EVMInstructionInterpreter::eval(
 	switch (_instruction)
 	{
 	case Instruction::STOP:
+		logTrace(_instruction, false);
```
bshastry: For instructions that do not write to memory such as `stop()`, the second argument is `false`

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -476,7 +485,7 @@ bool EVMInstructionInterpreter::accessMemory(u256 const& _offset, u256 const& _s
 	{
 		u256 newSize = (_offset + _size + 0x1f) & ~u256(0x1f);
 		m_state.msize = max(m_state.msize, newSize);
-		return _size <= 0xffff;
+		return _size <= 0xffff && _offset <= u256(numeric_limits<size_t>::max());
```
bshastry: This is a bug fix change: Since we only interpret memory offsets modulo `numeric_limits<size_t>::max()`, not having this change would mean, we interpret

```
{
  let sizeTMaxPlusOne := 0x10000000000
  extcodecopy(someValidAddress, sizeTMaxPlusOne, 0, 32)
  return(0, 32)
}
```

as a write to memory address zero. The subsequent `return` then returns what was written by extcodecopy.

chriseth:
Can you summarize the change? Do we at least compare the data in storage at the end?

bshastry:
> Can you summarize the change? Do we at least compare the data in storage at the end?

- Storage is compared as before (no changes there).
- Memory changes are not recorded if the flag added in this PR is set (disableMemoryTracing = true)
- Memory dumps are not compared if the flag added in this PR is set (disableMemoryTracing = true)

test/tools/ossfuzz/strictasm_diff_ossfuzz.cpp
chriseth: Shouldn't this be a comment in the code instead? :)

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
chriseth: I'm wondering - shouldn't we use some utility to infer this fact from the opcode itself?

bshastry: Sounds reasonable to me. More concretely, would it be a static look up table per opcode or an annotation to the opcode itself?

chriseth: I think `SemanticInformation::memory` would be suitable. It is a slight problem since we cannot test this function here anymore, but I guess it's better to use constants here all the time.

chriseth: I think `SemanticInformation::memory` would be suitable. It is a slight problem since we cannot test this function here anymore, but I guess it's better to use constants here all the time.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
chriseth: Can you extract this into its own PR?

test/tools/ossfuzz/strictasm_diff_ossfuzz.cpp
bshastry: Added comment.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
bshastry: Updated to make use of `SemanticInformation::memory`. If there are any mistaken assumptions, the fuzzer would report a bug which we could later investigate.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
bshastry: Sure, found a mistake on this line in the process :-)

Created: https://github.com/ethereum/solidity/pull/12484

hrkrshnn:
> Storage is compared as before (no changes there).

@bshastry How is that done? I have a feeling that might not be enough for something like

```solidity
{
    let x := calldataload(0)
    let y := sload(x)
    // both of these can be removed
    sstore(x, y)
    sstore(x, y)

    let a := x
    let b := mload(a)
    // both of these can be removed
    mstore(a, b)
    mstore(a, b)
}
// ----
// step: equalStoreEliminator
//
// {
//     let x := calldataload(0)
//     let y := sload(x)
//     let a := x
//     let b := mload(a)
// }
```

bshastry:
> > Storage is compared as before (no changes there).
> 
> @bshastry How is that done? I have a feeling that might not be enough for something like
> 
> ```solidity
> {
>     let x := calldataload(0)
>     let y := sload(x)
>     // both of these can be removed
>     sstore(x, y)
>     sstore(x, y)
> 
>     let a := x
>     let b := mload(a)
>     // both of these can be removed
>     mstore(a, b)
>     mstore(a, b)
> }
> // ----
> // step: equalStoreEliminator
> //
> // {
> //     let x := calldataload(0)
> //     let y := sload(x)
> //     let a := x
> //     let b := mload(a)
> // }
> ```

Essentially the following

- sstore(a, b): an entry in map is created but not an instruction trace `SSTORE(a, b)`
- mstore(a, b) is neither recorded in state nor trace

Since storage writes are always preserved in map, they may be compared pre/post optimisation. OTOH, memory may not be compared.

test/libyul/YulInterpreterTest.cpp
```diff
@@ -94,13 +94,18 @@ string YulInterpreterTest::interpret()
 	state.maxExprNesting = 64;
 	try
 	{
-		Interpreter::run(state, EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}), *m_ast);
+		Interpreter::run(
+			state,
+			EVMDialect::strictAssemblyForEVMObjects(langutil::EVMVersion{}),
+			*m_ast,
+			/*disableMemoryTracing=*/false
```
chriseth, line 101: Maybe avoid the negation and call it `enableMemoryTracing`?

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -339,12 +344,18 @@ u256 EVMInstructionInterpreter::eval(
 	case Instruction::REVERT:
 		accessMemory(arg[0], arg[1]);
 		logTrace(_instruction, arg);
+		m_state.storage.clear();
```
chriseth, line 347: How "correct" is this interpreter? Shouldn't it only clear the changes to storage? Should we extract this change into its own PR?

chriseth:
Could be merged like this.

chriseth (approved):

PR #12277, opened 2021-11-12
Relocating download Section in docs

### LINKED ISSUE #12238: [DOCS] Make Download Section Easier To Find
## Problem

People are not aware that there is the option to download the Solidity docs and we've had several requests about this recently. The download option is so hidden in the flyout menu (some need to scroll to see it), since the list of versions has grown quite long.

Example:
image

## Solution

Place the download options at a more obvious / easier visible place in the menu.

## Related Discussion

https://forum.soliditylang.org/t/please-can-i-get-the-documentation-in-downloadable-form/615/7

nishant-sachdeva:
1. Modified index.rst to include a Hint section for Downloads
2. Modified _templates/versions.html to place downloads section ahead of Versions section

Closes #12238

### DIFF
--- a/docs/_templates/versions.html
+++ b/docs/_templates/versions.html
@@ -13,13 +13,13 @@
     </span>
     <div class="rst-other-versions">
         <dl>
-            <dt>{{ _('Versions') }}</dt> {% for slug, url in versions %}
-            <dd><a href="{{ url }}">{{ slug }}</a></dd>
+            <dt>{{ _('Downloads') }}</dt> {% for type, url in downloads %}
+            <dd><a href="{{ url }}">{{ type }}</a></dd>
             {% endfor %}
         </dl>
         <dl>
-            <dt>{{ _('Downloads') }}</dt> {% for type, url in downloads %}
-            <dd><a href="{{ url }}">{{ type }}</a></dd>
+            <dt>{{ _('Versions') }}</dt> {% for slug, url in versions %}
+            <dd><a href="{{ url }}">{{ slug }}</a></dd>
             {% endfor %}
         </dl>
         <dl>
--- a/docs/_templates/versions.html
+++ b/docs/_templates/versions.html
@@ -1,7 +1,11 @@
 {# Add rst-badge after rst-versions for small badge style. #}
 <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="versions">
     <span class="rst-current-version" data-toggle="rst-current-version">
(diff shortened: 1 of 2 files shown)


dzhalaevd (dismissed):

chriseth:
Maybe we can use the area currently occupied by `<span class="fa fa-book fa-element"> RTD </span>` to have a simple "pdf" link?

docs/index.rst
```diff
@@ -31,6 +31,12 @@ a 0.y.z version number `to indicate this fast pace of change <https://semver.org
 Ideas for improving Solidity or this documentation are always welcome,
 read our :doc:`contributors guide <contributing>` for more details.
 
+.. Hint::
+
+  You can download this documentation as PDF, HTML or Epub by clicking on the versions
+  flyout menu in the left bottom corner and selecting the preferred download format.
```
cameel:
```suggestion
  flyout menu in the bottom left corner and selecting the preferred download format.
```

cameel:
> Maybe we can use the area currently occupied by `<span class="fa fa-book fa-element"> RTD </span>` to have a simple "pdf" link?

I agree. The `RTD` is just a piece of static text that is not needed for anything and we already credit RTD in the footer.

docs/_templates/versions.html
```diff
@@ -1,7 +1,7 @@
 {# Add rst-badge after rst-versions for small badge style. #}
 <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="versions">
     <span class="rst-current-version" data-toggle="rst-current-version">
-    <span class="fa fa-book fa-element"> RTD </span>
+    <a href="https://docs.soliditylang.org/_/downloads/en/latest/pdf/">PDF</a>
```
cameel: How about this?
```suggestion
    {% for type, url in downloads %}
        {% if type == 'PDF' %}
            <a href="{{ url }}">{{ type }}</a>
        {% endif %}
    {% endfor %}
```
Otherwise it will always display the link `latest`, regardless of which version of the docs you visit. I think it might be unexpected. It will also have a link to `docs.soliditylang.org` even if you visit from https://solidity.readthedocs.org (though you could fix that one also by just using `{{ PRODUCTION_DOMAIN }}`).

nishant-sachdeva: Thanks for the suggestion! Committed the same in the latest push.

cameel (approved):

PR #12282, opened 2021-11-15
Fix gas cost enforcement for constructors and make `--enforce-gas-cost-min-value` actually work

### LINKED ISSUE #11600: [isoltest] Bug in auto update for gas in constructor case
Take the following semantic test:

```solidity
contract C {
    uint public x = 1;
}
// ====
// compileViaYul: also
// ----
// constructor()
// gas irOptimized: 1
// gas legacy: 1
// gas legacyOptimized: 1
// x() -> 1
// gas irOptimized: 1
// gas legacy: 1
// gas legacyOptimized: 1
```

Running isoltest will automatically update the gas costs for `x()`, however the constructor call doesn't seem to get updated.

(Or only updated for `legacy`, which was the case https://github.com/ethereum/solidity/pull/11598/files#r661433855)

### LINKED ISSUE #12225: Fix gas expectations in the semantics tests
In https://github.com/ethereum/solidity/pull/11974 I noticed that the tests seem to pass no matter what ``legacyOptimized`` gas costs were used for the deposit contract.
The expectation only got handled correctly with a value for ``--enforce-gas-cost-min-value 1`` in ``isoltest`` and also ``soltest`` doesn't seem to fail on wrong expectation.

cameel:
Fixes #12225.
Fixes #11600.

There were two problems in `SemanticTest`:
- In case of a constructor call, the cause of failure was determined correctly (`m_gasCostFailure = true`) but the whole test was not marked as failed (`success = false`).
- The `--enforce-gas-cost-min-value` check was incorrectly bundled with another check meant prevent `ir` cost from being added to the expectations if it was not there originally. As a result it was enforced only if the cost it was checking was not yet listed in the file.

The PR also includes two tiny refactors - renaming a file with a typo and renaming `m_gas` in `ExecutionFramework` to make it clear how it differs from `m_gasUsed` (it's the gas available at the beginning of the transaction).


test/ExecutionFramework.cpp
```diff
@@ -182,15 +182,15 @@ void ExecutionFramework::sendMessage(bytes const& _data, bool _isCreation, u256
 		message.kind = EVMC_CALL;
 		message.destination = EVMHost::convertToEVMC(m_contractAddress);
 	}
-	message.gas = m_gas.convert_to<int64_t>();
+	message.gas = InitialGas.convert_to<int64_t>();
```
chriseth, line 185: Constants should be either full caps or just regular variables. UpperCamelCase is reserved for types.

chriseth (approved):
I'll approve and merge anyway because the comment is kind of minor.

test/ExecutionFramework.cpp
cameel: So how about full caps? Can we start using that everywhere? The `m_` was especially misleading in this case but I often find myself wishing we had some convention for constants to easily distinguish them from normal variables.

chriseth: I think the coding style guide says `c_` for constants.

cameel: Unfortunately it does not. Should that be preferred then? You mentioned that convention before but I don't remember seeing it actually used anywhere in the code base.

chriseth: Ah ok, we had that once but removed it because we don't really do it for local variables.

PR #12283, opened 2021-11-15
Graceful error handling in soltest/isoltest + improved soltestAssert()

### LINKED ISSUE #11917: [soltest] Command-line validation errors should not look like an ICE
Currently validation errors are reported as if they were internal compiler errors which makes it look like as if a bug in the compiler was triggered. A nicely formatted message should be printed instead.

For example the error caused by using `--testpath` with an empty value looks like this:
```bash
soltest -- --testpath
```
```
Test setup error: Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::wrapexcept<boost::program_options::invalid_command_line_syntax>
std::exception::what: the required argument for option '--testpath' is missing
```

cameel:
Fixes #11917.

While working on #12282 I hit an assertion in `ByteUtils` but it looked like something completely different. That's because `soltestAssert()` throws `std::runtime_error` rather than some special exception type reserved for assertions. I also noticed that in #12019 I did not cover `soltestAssert()`. This PR refactors the macro to be in line with `solAssert()`, i.e. allow skipping the message, provide a default message and use a dedicated exception type.

I have also added exception handlers that intercept normal errors (like command-line validation errors) and print them in a way that does not make you think you hit a bug.

Also, `--help` no longer returns an error code. It's normal and expected behavior. A similar change was made for `solc` in #12118.

I recommend reviewing with the option to ignore whitespace changes enabled - changes are actually small but some blocks had to be reindented due to the added `try`/`catch`.

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -204,15 +204,18 @@ int registerTests(
 	return numTestsAdded;
 }
 
-void initializeOptions()
+bool initializeOptions()
 {
 	auto const& suite = boost::unit_test::framework::master_test_suite();
 
 	auto options = std::make_unique<solidity::test::CommonOptions>();
-	solAssert(options->parse(suite.argc, suite.argv), "Failed to parse options!");
+	bool shouldContinue = options->parse(suite.argc, suite.argv);
+	if (!shouldContinue)
+		return false;
 	options->validate();
 
 	solidity::test::CommonOptions::setSingleton(std::move(options));
+	return true;
 }
 
 }
@@ -228,7 +231,9 @@ test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] )
 	master_test_suite_t& master = framework::master_test_suite();
 	master.p_name.value = "SolidityTests";
 
-	initializeOptions();
+	bool shouldContinue = initializeOptions();
+	if (!shouldContinue)
+		exit(0);
 
 	if (!solidity::test::loadVMs(solidity::test::CommonOptions::get()))
 		exit(1);
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -235,10 +235,10 @@ test_suite* init_unit_test_suite(int /*argc*/, char* /*argv*/[])
 	{
 		bool shouldContinue = initializeOptions();
 		if (!shouldContinue)
-			exit(0);
+			exit(EXIT_SUCCESS);
 
 		if (!solidity::test::loadVMs(solidity::test::CommonOptions::get()))
-			exit(1);
+			exit(EXIT_FAILURE);
 
 		if (solidity::test::CommonOptions::get().disableSemanticTests)
 			cout << endl << "--- SKIPPING ALL SEMANTICS TESTS ---" << endl << endl;
@@ -298,12 +298,12 @@ test_suite* init_unit_test_suite(int /*argc*/, char* /*argv*/[])
 	catch (solidity::test::ConfigException const& exception)
 	{
 		cerr << exception.what() << endl;
-		exit(1);
+		exit(EXIT_FAILURE);
 	}
 	catch (std::runtime_error const& exception)
 	{
 		cerr << exception.what() << endl;
-		exit(1);
+		exit(EXIT_FAILURE);
 	}
 
 	return nullptr;
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -222,75 +222,88 @@ bool initializeOptions()
 
 // TODO: Prototype -- why isn't this declared in the boost headers?
 // TODO: replace this with a (global) fixture.
-test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] );
+test_suite* init_unit_test_suite(int /*argc*/, char* /*argv*/[]);
 
-test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] )
+test_suite* init_unit_test_suite(int /*argc*/, char* /*argv*/[])
 {
 	using namespace solidity::test;
 
 	master_test_suite_t& master = framework::master_test_suite();
 	master.p_name.value = "SolidityTests";
 
-	bool shouldContinue = initializeOptions();
-	if (!shouldContinue)
-		exit(0);
-
-	if (!solidity::test::loadVMs(solidity::test::CommonOptions::get()))
-		exit(1);
-
-	if (solidity::test::CommonOptions::get().disableSemanticTests)
-		cout << endl << "--- SKIPPING ALL SEMANTICS TESTS ---" << endl << endl;
-
-	if (!solidity::test::CommonOptions::get().enforceGasTest)
-		cout << endl << "WARNING :: Gas Cost Expectations are not being enforced" << endl << endl;
-
-	Batcher batcher(CommonOptions::get().selectedBatch, CommonOptions::get().batches);
-	if (CommonOptions::get().batches > 1)
-		cout << "Batch " << CommonOptions::get().selectedBatch << " out of " << CommonOptions::get().batches << endl;
-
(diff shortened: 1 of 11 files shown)


test/boostTest.cpp
```diff
@@ -162,71 +162,89 @@ int registerTests(
...
-			master,
-			options.testPath / ts.path,
-			ts.subpath,
-			options.enforceViaYul,
-			options.enforceCompileToEwasm,
-			ts.labels,
-			ts.testCaseCreator
-		) > 0, std::string("no ") + ts.title + " tests found");
+		cerr << exception.what() << endl;
+		exit(1);
```
christianparpart: I think instead of numerical values (unless they do encoding something non-standard), the process exit code [should use](https://stackoverflow.com/a/8868139/386670) `EXIT_FAILURE` and `EXIT_SUCCESS`.

test/boostTest.cpp
```diff
@@ -162,71 +162,89 @@ int registerTests(
...
 	}
-
-	if (solidity::test::CommonOptions::get().disableSemanticTests)
+	catch (std::runtime_error const& exception)
```
christianparpart, line 303: Why `runtime_error` and not `exception`?

test/tools/isoltest.cpp
```diff
@@ -422,8 +422,9 @@ int main(int argc, char const *argv[])
 		{
 			auto options = std::make_unique<solidity::test::IsolTestOptions>();
 
-			if (!options->parse(argc, argv))
-				return -1;
+			bool shouldContinue = options->parse(argc, argv);
+			if (!shouldContinue)
+				return 0;
```
christianparpart: from -1 to 0, hmm.. so EXIT_SUCCESS. I think you expect the return value to not never reflect any error but whether or not the program should be terminated (in success). I am not sure that is so self-explanatory :)

test/tools/isoltest.cpp
```diff
@@ -480,9 +481,20 @@ int main(int argc, char const *argv[])
 
 		return global_stats ? 0 : 1;
 	}
-	catch (std::exception const& _exception)
+	catch (boost::program_options::error const& exception)
```
christianparpart, line 503: This is actually having `std::logic_error` as base which is based on `std::exception`. Why not simply catching `std::exception` here?

Not sure what other types of catches we should expect in the default catch.

christianparpart (changes requested):

test/boostTest.cpp
cameel: Because the exceptions we throw from assertions inherit from `exception` and I don't want to lump them together. `runtime_error` seems to be used for reporting validation errors and if we catch one, `isoltest` should just print the message and exit. This is normal and can happen e.g. if you just use wrong arguments, file is missing, etc. But if we get an assertion or another unexpected exception that should have been handled earlier we have a bug - we want to defer to the top-level exception handler, which will just print as much debug info as it can and exit.

test/tools/isoltest.cpp
cameel: I'm not sure I understand. It used to return -1 if you used `--help`. I think this is a case that should result in success and not an error.

Good suggestion about `EXIT_SUCCESS` instead of `0`. I'll change that.

test/tools/isoltest.cpp
cameel: I want to catch specifically `program_options::error` because I known it can be thrown in case of a command-line validation error. I do not expect `logic_error` here so if I get one I consider it a bug, to be handled the same way as an exception that should have been handled earlier and wasn't or an assertion failure.

To be precise, this specific catch branch could be placed just around the `options->parse()` call because that's the only place where we expect a CLI validation error from `boost::program_options` but I thought it would be better to place it next to the handler for `runtime_error`. They should really be handled the same way (print a nice message + exit).

And there are no other catches needed. I don't think there are any other types that can reach this handler in a normal flow and all unexpected ones and assertions are handled by `catch (...)`.

Marenz (dismissed):

cameel:
Rebased on `develop` to have the CI pass.

leonardoalt:
Has conflicts

leonardoalt:
Still has conflicts.

cameel:
Just needs a rebase (and another look from @christianparpart :)). Otherwise it should be fine.

test/libsolidity/util/SoltestErrors.h
```diff
@@ -15,20 +15,35 @@
...
+struct InternalSoltestError: virtual util::Exception {};
+
+#if !BOOST_PP_VARIADICS_MSVC
+#define soltestAssert(...) BOOST_PP_OVERLOAD(soltestAssert_,__VA_ARGS__)(__VA_ARGS__)
+#else
+#define soltestAssert(...) BOOST_PP_CAT(BOOST_PP_OVERLOAD(soltestAssert_,__VA_ARGS__)(__VA_ARGS__),BOOST_PP_EMPTY())
+#endif
+
+#define soltestAssert_1(CONDITION) \
+	soltestAssert_2(CONDITION, "")
```
christianparpart:
```suggestion
	soltestAssert_2((CONDITION), "")
```

test/libsolidity/util/SoltestErrors.h
```diff
@@ -15,20 +15,35 @@
...
 
+#define soltestAssert_2(CONDITION, DESCRIPTION) \
+	assertThrowWithDefaultDescription( \
+		CONDITION, \
+		::solidity::frontend::test::InternalSoltestError, \
+		DESCRIPTION, \
```
christianparpart:
```suggestion
		(CONDITION), \
		::solidity::frontend::test::InternalSoltestError, \
		(DESCRIPTION), \
```

test/tools/isoltest.cpp
```diff
@@ -480,9 +481,20 @@ int main(int argc, char const *argv[])
...
 	{
-		cerr << _exception.what() << endl;
+		cerr << exception.what() << endl;
+		return 1;
```
christianparpart:
```suggestion
		return EXIT_FAILURE;
```

test/tools/isoltest.cpp
```diff
@@ -480,9 +481,20 @@ int main(int argc, char const *argv[])
...
+	}
+	catch (std::runtime_error const& exception)
+	{
+		cerr << exception.what() << endl;
+		return 1;
```
christianparpart:
```suggestion
		return EXIT_FAILURE;
```

test/tools/isoltest.cpp
```diff
@@ -480,9 +481,20 @@ int main(int argc, char const *argv[])
...
+	}
+	catch (...)
+	{
+		cerr << "Uncaught exception" << endl;
```
christianparpart:
```suggestion
		cerr << "Uncaught exception." << endl;
```
Actually. You just caught it. So probably you want to write:
```suggestion
		cerr << "Unhandled exception caught." << endl;
```

christianparpart (changes requested):
LGTM, apart from the very minors. :)

test/libsolidity/util/SoltestErrors.h
cameel: Oh, that's a really good point! Fixed now and I also added a new commit that fixes this in the assert macros I edited earlier.

test/tools/isoltest.cpp
cameel: Done.

test/tools/isoltest.cpp
cameel: Well, it really means that it was not caught until now but you have a point :) For the user seeing this message this might not be clear. I changed the message to your second version.

cameel:
Ready for another round of review.

christianparpart (approved):
great. thanks. :-)

Issue #7197, opened 2019-08-07
[isoltest] Formatting should support strings longer than 32 bytes


erak:
The following test
```
contract C {
    function e() public {
        revert("Transaction failed. Transaction failed.");
    }
}
// ====
// EVMVersion: >homestead
// ----
// e() ->
```
does not handle strings longer than 32 bytes correctly:
```
// e() -> FAILURE, hex"08c379a0", 0x20, 39, "Transaction failed. Transaction "
```

chriseth:
Is this only an issue with error strings or with strings in general?

erak:
This applies to all strings (error strings aren't handled differently):
```
contract C {
    function s() public pure returns (string memory) {
        return "A string that is longer than 32 bytes.";
    }
}
// ----
// s() -> 0x20, 38, "A string that is longer than 32 bytes."
```
Which leads to an exception since the expected value can not be encoded. 

Leaving the expected values empty in order to let the auto-update routine fill them in, results in a broken value:
```
// s() -> 0x20, 38, "A string that is longer than 32 "
```

mijovic:
I think this is solved now

axic:
Aren't those strings still split into 32-byte chunks, e.g. "A string..", "is long...".

mijovic:
> Aren't those strings still split into 32-byte chunks, e.g. "A string..", "is long...".

Yes, they are. But I was thinking back than it wasn't working, so maybe issue is not worth fixing.

PR #12284, opened 2021-11-15
Fix formatting for strings with control characters in isoltest expectations

cameel:
Found while working on #12282.

An autoupdate of gas expectations in `semanticTests/externalContracts/strings.sol` resulted in an assertion failure in `BytesUtils::alignLeft()` because the string was longer than 32 bytes. I think this is the problem from #7197.
```
Exception during test: /solidity/test/libsolidity/util/BytesUtils.cpp(41): Throw in function static solidity::bytes solidity::frontend::test::BytesUtils::alignLeft(solidity::bytes)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: 
```

The root cause though was that isoltest reformatted the expectations in a weird way. It read this:
```
// utf8len(string): 0x20, 16, "\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83" -> 4 # Input: "😃😃😃😃" #
```
but spewed out this:
```
// utf8len(string): 0x20, 16, "\x00000000000000000000000000000000000000000000000000000000000000f0\x000000000000000000000000000000000000000000000000000000000000009f\x0000000000000000000000000000000000000000000000000000000000000098\x0000000000000000000000000000000000000000000000000000000000000083\x00000000000000000000000000000000000000000000000000000000000000f0\x000000000000000000000000000000000000000000000000000000000000009f\x0000000000000000000000000000000000000000000000000000000000000098\x0000000000000000000000000000000000000000000000000000000000000083\x00000000000000000000000000000000000000000000000000000000000000f0\x000000000000000000000000000000000000000000000000000000000000009f\x0000000000000000000000000000000000000000000000000000000000000098\x0000000000000000000000000000000000000000000000000000000000000083\x00000000000000000000000000000000000000000000000000000000000000f0\x000000000000000000000000000000000000000000000000000000000000009f\x0000000000000000000000000000000000000000000000000000000000000098\x0000000000000000000000000000000000000000000000000000000000000083" -> 4 # Input: "😃😃😃😃" #
```

Apparently calling `toHex(static_cast<uint8_t>(c))` results in a call to the overload for `bytes` rather than `uint8_t`. Which is bizarre because while `toHex()` has optional arguments, the type is an exact match so there's no ambiguity. I wonder if it's some obscure quirk of the language or a compiler bug.

### DIFF
--- a/test/libsolidity/semanticTests/isoltestTesting/format_raw_string_with_control_chars.sol
+++ b/test/libsolidity/semanticTests/isoltestTesting/format_raw_string_with_control_chars.sol
@@ -1,0 +1,12 @@
+contract C {
+    function f(string memory s) external pure returns (string memory) {
+        return s;
+    }
+}
+// NOTE: The test is here to illustrate the problem with formatting control chars in strings in
+// test expectations but unfortunately it can only be triggered manually. It does not test anything
+// unless you introduce a difference in expectations to force isoltest to reformat them.
+// ====
+// compileViaYul: also
+// ----
+// f(string): 0x20, 16, "\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83" -> 0x20, 16, "\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83" # Input/Output: "😃😃😃😃" #
--- a/test/libsolidity/util/BytesUtils.cpp
+++ b/test/libsolidity/util/BytesUtils.cpp
@@ -219,7 +219,7 @@ string BytesUtils::formatString(bytes const& _bytes, size_t _cutOff)
 				if (isprint(v))
 					os << v;
 				else
-					os << "\\x" << toHex(v);
+					os << "\\x" << toHex(v, HexCase::Lower);
 		}
 	}
 	os << "\"";
--- a/test/libsolidity/util/TestFunctionCall.cpp
+++ b/test/libsolidity/util/TestFunctionCall.cpp
@@ -322,7 +322,9 @@ string TestFunctionCall::formatRawParameters(
 			if (param.format.newline)
 				os << endl << _linePrefix << "// ";
 			for (auto const c: param.rawString)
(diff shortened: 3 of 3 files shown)


test/libsolidity/semanticTests/isoltestTesting/format_raw_string_with_control_chars.sol
```diff
@@ -0,0 +1,12 @@
...
+        return s;
+    }
+}
+// NOTE: The test is here to illustrate the problem with formatting control chars in strings in
+// test expectations but unfortunately it can only be triggered manually. It does not test anything
+// unless you introduce a difference in expectations to force isoltest to reformat them.
+// ====
+// compileViaYul: also
+// ----
+// f(string): 0x20, 16, "\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83" -> 0x20, 16, "\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83\xf0\x9f\x98\x83" # Input/Output: "😃😃😃😃" #
```
cameel, line 12: Not sure if it's worth keeping this test. To trigger it you have to change expectations, e.g. replace `16` with `15`.

`strings.sol` actually has a higher chance of triggering it by accident because it has gas expectations, which change often.

test/libsolidity/util/BytesUtils.cpp
```diff
@@ -219,7 +219,7 @@ string BytesUtils::formatString(bytes const& _bytes, size_t _cutOff)
 				if (isprint(v))
 					os << v;
 				else
-					os << "\\x" << toHex(v);
+					os << "\\x" << toHex(v, HexCase::Lower);
```
cameel, line 222: I think this is the only other place where we call `toHex()` with a single char argument (BTW, because of `auto`, I'm not 100% sure what exact type is `v` here, I'd have to check how `bytes` is defined).

I did not manage to trigger this in tests but I changed it too just in case.

chriseth (approved):

Issue #12057, opened 2021-09-29
[Sol->Yul] Invalid IR generated during copy of dynamically sized storage arrays of function type


bshastry:
```
contract C {
  function() external [] s0;
  function() external [] s1;
  constructor() {
    s0 = s1;
  }
  receive() external payable {}
}
```

throws

https://github.com/ethereum/solidity/blob/e5eed63a3e83d698d8657309fd371248945a1cda/libsolidity/codegen/ir/IRGenerator.cpp#L107

To repro, do

```
$ solc --experimental-via-ir test.sol
```

nishant-sachdeva:
Okay, so I put the test mentioned here in a .sol file of it's own and ran the following commands 

1. solc --ir test.sol 2> only_ir.txt

2. solc --strict-assembly  only_ir.txt 2> error.txt

I got the following error description
`Error: Function "convert_t_function_external_nonpayable$__$returns$__$_to_t_function_external_nonpayable$__$returns$__$" expects 2 arguments but got 1.
   --> only_ir.txt:236:38:`

`Error: Variable count mismatch for declaration of "itemValue": 1 variables and 2 values.
   --> only_ir.txt:236:21:`

`Error: Function "prepare_store_t_function_external_nonpayable$__$returns$__$" expects 2 arguments but got 1.
   --> only_ir.txt:239:34:`


![Screenshot from 2021-11-24 23-36-19](https://user-images.githubusercontent.com/32475507/143291970-23255aaf-967f-4828-822b-75af7678dc28.png)

cameel:
Yeah, that's the issue. Yul code generated by the compiler should not have syntax errors.

To find the cause I'd start looking around the code that processes assignments in the IR codegen: https://github.com/ethereum/solidity/blob/71f8576b86e7c2f9161683bf2c6c7129f624f182/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L434-L489

It adds a conversion for the value being assigned. We have a helper that generates the conversion function you see in the error messages and that function apparently gets too few arguments. My guess would be that it's because external function pointers take one slot in storage but two slots on the stack and the code that inserts a call to it does not take that into account: https://github.com/ethereum/solidity/blob/71f8576b86e7c2f9161683bf2c6c7129f624f182/libsolidity/ast/Types.cpp#L3103-L3151

[...]

Issue #11826, opened 2021-08-20
Getters returning structs can not be defined in interfaces


axic:
```solidity
interface I {
    struct S {
        uint a;
    }

    function s() external returns (S memory);
}

contract C is I {
    S public override s;
}
```

The function `s` can not be defined in a way this works. The problem lies in the impossibility to define a working location.

`S storage` is not allowed in interfaces, but that could be one option:
```
Error: Data location must be "memory" or "calldata" for return parameter in function, but "storage" was given.
 --> struct-interface.sol:6:36:
  |
6 |     function s() external returns (S storage);
  |                                    ^^^^^^^^^
```

`S memory` is not allowed by overriding, but that could be equally correct:
```
Error: Overriding public state variable return types differ.
  --> struct-interface.sol:10:5:
   |
10 |     S public override s;
   |     ^^^^^^^^^^^^^^^^^^^
Note: Overridden public state variable is here:
 --> struct-interface.sol:6:5:
  |
6 |     function s() external returns (S memory);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

axic:
Oops, forgot that struct getters return the members, and not the struct itself. That is super unintuitive.

leonardoalt:
> Oops, forgot that struct getters return the members, and not the struct itself. That is super unintuitive.

Don't forget that if the struct has arrays those are ignored.

axic:
```
Error: Overriding public state variable return types differ.
  --> struct-interface.sol:10:5:
   |
10 |     S public override s;
   |     ^^^^^^^^^^^^^^^^^^^
```

Could also consider displaying the return types in this error.

ekpyron:
> Oops, forgot that struct getters return the members, and not the struct itself. That is super unintuitive.

Can we change that at *some* point :-)?

cameel:
Even weirder that this goes only one level deep. You can have a getter returning a struct if it's nested inside another struct.

For example this compiles just fine and both the function and the getter return `S`:
```solidity
interface I {
    struct S {
        uint a;
    }
    struct WrappedS {
        S s;
    }

    function s() external returns (S memory);
}

contract C is I {
    WrappedS public override s;
}
```

Anyway, you should update issue the description to say what we want to do about it. Do we want to make getters return structs instead of tuples (breaking change I guess?). Or just make the message more explicit?

abhijitgawai:
[...]

PR #12287, opened 2021-11-16
Implement typechecked abi.encodeCall()

### LINKED ISSUE #11849: Easy type-safe encoding a method call.
Decision: Add `abi.encodeCall(f, (arg1, arg2))`
Check that `abi.encodeCall(f, ())` also works.

---

In modern solidity code, it is very common to pass encoded function to other solidity contract for multiple purposes (delayed operation, adding access-control, meta-transactions, etc)

something similar to
```solidity
someContract.executeLater(target,
    abi.encodeWithSelector(MyContract.myMethod.selector, param1, uint(param2)) ); 
```

however, using encodeWithSelector is very error-prone: there is no type checking, and no tell if the method signature is modified.

### Suggestion:
add a syntactic sugar encoder, so that the above can be written like this:
```solidity
someContract.executeLater(target,
    MyContract.myMethod.encode(param1, uint(param2)) ); 
```

The output of such "method call" is just like of encodeWithSelector - a `bytes memory`
The only difference is clearer clearer syntax and type-checking.

Note that it is a "method call" of a contract type, not an instance: just like encodeWithAbi, there is no contract instance involved.

Marenz:
fixes #11849


test/libsolidity/semanticTests/abiencodedecode/abi_encode_checked.sol
```diff
@@ -0,0 +1,49 @@
...
+        s.b = "Lorem ipsum dolor sit ethereum........";
+        s.c = 0x1234;
+        return abi.encodeWithSignature(s.b, type(uint).max, s, uint(3));
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// f0() -> 0x20, 0x24, -34435155370463444793260793355178157075203752403645521721995013737368954863616, 26959946667150639794667015087019630673637144422540572481103610249216
+// f0Checked() -> 0x20, 0x24, -34435155370463444793260793355178157075203752403645521721995013737368954863616, 26959946667150639794667015087019630673637144422540572481103610249216
```
Marenz: This currently results in `f0Checked() -> 0x20, 0` for the old backend.

libsolidity/ast/Types.h
```diff
@@ -1237,6 +1237,7 @@ class FunctionType: public Type
 		ABIEncode,
 		ABIEncodePacked,
 		ABIEncodeWithSelector,
+		ABIEncodeChecked,
```
chriseth: Why don't you call it `ABIEncodeCall`?

Marenz: I thought the unique new thing about this one was that it checks the types.. thus `Checked` but I don't mind how exactly we call it...

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1236,22 +1236,40 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
-					argumentTypes.push_back(arguments[i]->annotation().type);
+				solAssert(arguments.size() == 2, "");
+
+				auto const tuple = dynamic_cast<TupleExpression const*>(arguments[1].get());
+
+				// Account for tuples with one component which become that component
+				if (!tuple)
+					argumentTypes.emplace_back(arguments[1]->annotation().type);
+				else
+					for (auto component: tuple->components())
```
chriseth: `argumentTypes = tupel->annotation().type.components()`?

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1308,10 +1326,15 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 						dataOnStack = TypeProvider::fixedBytes(32);
 					}
 				}
-				else
+				else if (function.kind() == FunctionType::Kind::ABIEncodeChecked)
 				{
-					solAssert(function.kind() == FunctionType::Kind::ABIEncodeWithSelector, "");
+					auto const functionPtr = dynamic_cast<FunctionTypePointer>(arguments[0]->annotation().type);
+					solAssert(functionPtr, "");
+					m_context << util::selectorFromSignature(functionPtr->externalSignature());
```
chriseth: Is this correctly left or right aligned?

libsolidity/codegen/ExpressionCompiler.cpp
Marenz: That is indeed better!

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2083,76 @@ void TypeChecker::typeCheckABIEncodeFunctions(
 	}
 }
 
+#include <typeindex>
+#include <typeinfo>
+void TypeChecker::typeCheckABIEncodeCheckedfunction(FunctionCall const& _functionCall)
```
aarlt: Maybe the function name is better readable if renamed to `typeCheckABIEncodeCheckedFunction`

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2083,76 @@ void TypeChecker::typeCheckABIEncodeFunctions(
 	}
 }
 
+#include <typeindex>
```
aarlt: I guess these includes will be moved to the top of the file, right?

Changelog.md
```diff
@@ -1,6 +1,7 @@
 ### 0.8.11 (unreleased)
 
 Language Features:
+ * General: New function ``abi.encodeCall(FunctionPointer functionPointer, (...)) returns (bytes memory)`` that performs a full type-check.
```
chriseth: Maybe
```suggestion
 * General: New function ``abi.encodeCall(functionPointer, (arg1, arg2, ...))`` that type-checks the arguments and returns the ABI-encoded function call data.
```

docs/units-and-global-variables.rst
```diff
@@ -135,6 +135,7 @@ ABI Encoding and Decoding Functions
 - ``abi.encode(...) returns (bytes memory)``: ABI-encodes the given arguments
 - ``abi.encodePacked(...) returns (bytes memory)``: Performs :ref:`packed encoding <abi_packed_mode>` of the given arguments. Note that packed encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: ABI-encodes the given arguments starting from the second and prepends the given four-byte selector
+- ``abi.encodeCall(FunctionPointer functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature.
```
chriseth:
```suggestion
- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature.
```
The problem is that we cannot really write down the type because it is different for every function ,but at least the ABI calls the type `function`

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2507,9 +2581,14 @@ bool TypeChecker::visit(FunctionCall const& _functionCall)
 		case FunctionType::Kind::ABIEncodePacked:
 		case FunctionType::Kind::ABIEncodeWithSelector:
 		case FunctionType::Kind::ABIEncodeWithSignature:
+		case FunctionType::Kind::ABIEncodeCall:
 		{
 			typeCheckABIEncodeFunctions(_functionCall, functionType);
 			returnTypes = functionType->returnParameterTypes();
+
+			if (functionType->kind() == FunctionType::Kind::ABIEncodeCall)
+				typeCheckABIEncodeCallFunction(_functionCall);
```
chriseth: Wouldn't it be nicer to put this call inside `typeCheckABIEncodeFunctions`?

Marenz: True, it's kind of a subset of the other if you go by the naming

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+			_functionCall.location(),
+			"Expected first argument to be a function pointer, not \"" +
+			type(*arguments.front())->canonicalName() +
+			"\"."
+		);
+		return;
+	}
+
+	// There are no checks we can do if all types are accepted
+	if (functionPointerType->takesArbitraryParameters())
```
chriseth: I think we should ensure / assert that it does not take arbitrary parameters.
Can you add a type check that the function kind is `FunctionType::Kind::External`?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+		m_errorReporter.typeError(
+			6219_error,
+			_functionCall.location(),
+			"Expected two arguments: a function pointer followed by a tuple."
+		);
+		return;
+	}
+
+	auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front()));
+	auto const argumentTuple = dynamic_cast<TupleType const*>(type(*arguments[1]));
```
chriseth: Can you move this closer to where it is first used?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+
+	std::vector<Type const*> givenArguments;
+
+	// Tuples with only one component become that component
+	if (!argumentTuple)
+		givenArguments.push_back(type(*arguments[1]));
+	else
+		givenArguments = argumentTuple->components();
+
+	size_t const numArguments = functionPointerType->parameterTypes().size();
```
```suggestion
	size_t const numParameters = functionPointerType->parameterTypes().size();
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+		m_errorReporter.typeError(
+			7788_error,
+			_functionCall.location(),
+			"Expected " + to_string(numArguments) + " instead of " + to_string(givenArguments.size()) + " components for the tuple parameter."
+		);
+		return;
+	}
+
+	for (size_t i = 0; i < numArguments; i++)
+		if (!givenArguments[i]->isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]))
```
chriseth: This function returns a `BoolResult` that might contain further explanations about why the conversion is not possible.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+			m_errorReporter.typeError(
+				5407_error,
+				_functionCall.location(),
+				"Parameter mismatch: Expected \"" +
```
chriseth: Should say "implicit conversion not possible" or something like that.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,75 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+	size_t const numArguments = functionPointerType->parameterTypes().size();
+
+	if (givenArguments.size() != numArguments)
+	{
+		m_errorReporter.typeError(
+			7788_error,
+			_functionCall.location(),
+			"Expected " + to_string(numArguments) + " instead of " + to_string(givenArguments.size()) + " components for the tuple parameter."
+		);
+		return;
```
chriseth: You could still continue and check the first `min(numArguments, givenArguments.size())` arguments.

chriseth:
Does this need a change in ViewPureChecker.cpp?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,57 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
-		for (size_t i = 0; i < arguments.size(); ++i)
+		string selector;
+		vector<ASTPointer<Expression const>> argumentsEncodeFunction;
+
+		if (functionType->kind() == FunctionType::Kind::ABIEncodeCall)
+		{
+			solAssert(arguments.size() == 2, "");
+			auto const tuple = dynamic_cast<TupleExpression const*>(arguments[1].get());
+
+			// Account for tuples with one component which become that component
```
chriseth, line 1126: I think we might have a helper for this at some point.

chriseth: Actually are you sure you need this special case?
`type(*arguments[1])` will already handle it properly, and I don't see why `IRVariable(*argument)` should have a problem.

Marenz: I am not sure what you're suggesting? That I use/add the tuple (or normal type) directly without iterating over the components in case of a tuple?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,112 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+	{
+		SecondarySourceLocation ssl{};
+
+		if (functionPointerType->hasDeclaration())
+			ssl.append("Function is declared here:", functionPointerType->declaration().location());
+		m_errorReporter.typeError(
+			3509_error,
+			arguments[0]->location(),
+			ssl,
+			"Function must be \"public\" or \"external\". Did you forget to precede \"this.\"?"
```
chriseth: Do you think it is a common error to miss `this.`? I.e. is it common to encode calls to the current contract?
Also I think `prefix` could be better than `precede`.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,112 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+		);
+	}
+
+	// There are no checks we can do if all types are accepted
+	if (functionPointerType->takesArbitraryParameters())
```
chriseth: Do you have an example here this is true after the check above?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,112 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+
+	for (size_t i = 0; i < numParameters; i++)
+	{
+		Type const& argType = *type(*getArg(i));
+		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]);
+		if (!result)
+			m_errorReporter.typeError(
+				5407_error,
+				getArg(i)->location(),
+				"Cannot perform an implicit conversion for component at position " +
```
chriseth: In all the other error messages we write "cannot be implicitly converted.."

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,112 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+
+	// Get i'th component if it's a tuple or just the argument itself if it's not.
+	auto getArg = [&](size_t _i) -> ASTPointer<Expression const>
+	{
+		solAssert(_i < numParameters, "");
+
+		if (!argumentTuple)
+			return arguments[1];
+
+		return argumentTuple->components()[_i];
```
chriseth: Maybe this is a bit simpler?
```suggestion
		if (argumentTuple)
			return argumentTuple->components().at(_i);
		else
			return arguments.at(1);
```

chriseth: If you want to add an assertion, then what I would suggest is adding `solAssert(_i == 0)` in the case where we do not have an argument tuple.

libsolidity/analysis/TypeChecker.cpp
Marenz: > it common to encode calls to the current contract?

Hmm I naively used my own experience where I _did_ forget `this.` in the tests, but you're right, the common case is usually external contracts, not the current one.

I like `prefix` more, too. Or I just remove it.

libsolidity/analysis/TypeChecker.cpp
Marenz: Yes, in one of the test cases:
```
	function fail6() public returns(bytes memory) {
		return abi.encodeCall(abi.encodeCall, (1, 2, 3, "test"));
	}
// TypeError 3509: (619-633): Function must be "public" or "external". Did you forget to precede "this."?
// TypeError 6716: (619-633): Function takes arbitrary parameters. No type-checking possible.
```

However, that's only possible because I don't bail out of the function in the check above this one..

chriseth: Yeah, I think it is better to have a proper assertion here, this looks rather dangerous.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,100 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+	{
+		SecondarySourceLocation ssl{};
+
+		if (functionPointerType->hasDeclaration())
+			ssl.append("Function is declared here:", functionPointerType->declaration().location());
+		m_errorReporter.typeError(
+			3509_error,
+			arguments[0]->location(),
+			ssl,
+			"Function must be \"public\" or \"external\". Did you forget to prefix \"this.\"?"
```
chriseth: I still think the suggestion could be confusing. We can check if `functionPointerType->declaration().scope()` is the same as `m_currentContract` at least.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,100 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+		);
+		return;
+	}
+
+	solAssert(!functionPointerType->takesArbitraryParameters(), "No checks possible for arbitrary parameters.");
+
+	// Tuples with only one component become that component
+	auto const argumentTuple = std::dynamic_pointer_cast<TupleExpression const>(arguments[1]);
+	size_t const numArguments = argumentTuple ? argumentTuple->components().size() : 1;
+	size_t const numParameters = min(numArguments, functionPointerType->parameterTypes().size());
```
chriseth: Can you move this down to where it is first used?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,100 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+			ssl,
+			"Function must be \"public\" or \"external\". Did you forget to prefix \"this.\"?"
+		);
+		return;
+	}
+
+	solAssert(!functionPointerType->takesArbitraryParameters(), "No checks possible for arbitrary parameters.");
+
+	// Tuples with only one component become that component
+	auto const argumentTuple = std::dynamic_pointer_cast<TupleExpression const>(arguments[1]);
```
chriseth: You could also use
```
	vector<ASTPointer<Expression>> arguments;
	if (auto const* argumentTuple = dynamic_cast<TupleExpression coonst*>(arguments[1].get()))
		arguments = argumentTuple->components();
	else
		arguments = {arguments[1]};
```
And then you can use `arguments.size()` instead of `numArguments` and `getArg` below is just `arguments.at(i)`.

chriseth:
Can you add a test for `return abi.encodeCall(this.f, [1, 2]);` I.e. check that it only accepts tuples and not inline arrays?

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call.sol
```diff
@@ -0,0 +1,31 @@
+pragma abicoder v2;
+contract C {
+	function f(uint256 p) public {}
+
+	function f0() public pure returns (bytes memory) {
+		return abi.encodeWithSignature("f(uint256)", 1);
```
chriseth: The test expectation below is unreadable. Can you add a function `test()` that checks if the keccak hashes of the two byte arrays are equal?

chriseth (commented):
Looks really good already, just some more test and cleanup suggestions.

libsolidity/analysis/TypeChecker.cpp
Marenz: I think I tried something like that, but the problem ended up being that `components()` returns an array without any const, where as `_functionCall.arguments()` is an array of const expressions. 

I can't assign/add const `arguments[1]` to `arguments` and if I make `arguments` contain const expressions then I can't assign `components()` to `arguments`..

```
	// Tuples with only one component become that component
	vector<ASTPointer<Expression>> callArguments;

	if (auto const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get()))
		callArguments = argumentTuple->components(); // failure here if you change `callArguments` to have Expression const as element type, but below works
	else
		callArguments.push_back(arguments[1]); // failure here, const to no-const conversion
```

chriseth:
Can you also add a test with a function pointer stored in a local variable and one stored in storage?

libsolidity/analysis/TypeChecker.cpp
chriseth: You can use const pointers and then convert the non-const to the const-version using `| ranges::to<decltype(callArguments)>;` or something.
another way is to use `{argumentTuple->components().begin(), argumentTuple->components().end()}`

docs/units-and-global-variables.rst
```diff
@@ -135,6 +135,7 @@ ABI Encoding and Decoding Functions
 - ``abi.encode(...) returns (bytes memory)``: ABI-encodes the given arguments
 - ``abi.encodePacked(...) returns (bytes memory)``: Performs :ref:`packed encoding <abi_packed_mode>` of the given arguments. Note that packed encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: ABI-encodes the given arguments starting from the second and prepends the given four-byte selector
+- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature.
```
chriseth: Maybe you could add ```equivalent to `bytes.concat(functionPointer.selector, abi.encode(...))` ```?

chriseth: Also there is another file called `cheatsheet.rst` that lists these functions as well.

chriseth: Sorry, it's of course `encodeWithSelector`

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1236,22 +1236,39 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
 			bool const hasSelectorOrSignature =
 				function.kind() == FunctionType::Kind::ABIEncodeWithSelector ||
+				function.kind() == FunctionType::Kind::ABIEncodeCall ||
 				function.kind() == FunctionType::Kind::ABIEncodeWithSignature;
 
 			TypePointers argumentTypes;
-			TypePointers targetTypes;
-			for (unsigned i = 0; i < arguments.size(); ++i)
+
+			ASTNode::listAccept(arguments, *this);
```
chriseth, line 1250: Doesn't this put the function pointer on the stack?

Marenz:
This is currently failing as there is a problem with the extraction of the selector in the code gen when it's a function pointer. I am working on it

Changelog.md
```diff
@@ -1,6 +1,7 @@
 ### 0.8.11 (unreleased)
 
 Language Features:
+ * General: New function ``abi.encodeCall(functionPointer, (arg1, arg2, ...))`` that type-checks the arguments and returns the ABI-encoded function call data.
```
cameel: Maybe we can call it a "built-in" function for clarity?
```suggestion
 * General: New built-in function ``abi.encodeCall(functionPointer, (arg1, arg2, ...))`` that type-checks the arguments and returns the ABI-encoded function call data.
```

docs/cheatsheet.rst
```diff
@@ -80,6 +80,8 @@ Global Variables
   the given arguments. Note that this encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: :ref:`ABI <ABI>`-encodes
   the given arguments starting from the second and prepends the given four-byte selector
+- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the
+  tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to `bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))`
```
cameel:
```suggestion
  tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```

docs/units-and-global-variables.rst
```diff
@@ -136,6 +136,7 @@ ABI Encoding and Decoding Functions
 - ``abi.encodePacked(...) returns (bytes memory)``: Performs :ref:`packed encoding <abi_packed_mode>` of the given arguments. Note that packed encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: ABI-encodes the given arguments starting from the second and prepends the given four-byte selector
 - ``abi.encodeWithSignature(string memory signature, ...) returns (bytes memory)``: Equivalent to ``abi.encodeWithSelector(bytes4(keccak256(bytes(signature))), ...)``
+- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to `bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))`
```
cameel:
```suggestion
- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,97 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+	if (functionPointerType->kind() != FunctionType::Kind::External)
+	{
+		string msg = "Function must be \"public\" or \"external\".";
+		SecondarySourceLocation ssl{};
+
+		if (functionPointerType->hasDeclaration())
+		{
+			ssl.append("Function is declared here:", functionPointerType->declaration().location());
+			if (functionPointerType->declaration().scope() == m_currentContract)
+				msg += " Did you forget to prefix \"this.\"?";
```
cameel, line 2127:
```suggestion
				msg += " Did you forget to prefix it with \"this.\"?";
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,97 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+		}
+
+		m_errorReporter.typeError(3509_error, arguments[0]->location(), ssl, msg);
+		return;
+	}
+
+	solAssert(!functionPointerType->takesArbitraryParameters(), "No checks possible for arbitrary parameters.");
```
cameel: I think this would be a better explanation of why we assume the assertion will never fail:
```suggestion
	solAssert(!functionPointerType->takesArbitraryParameters(), "Function must have fixed parameters.");
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,97 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+
+	// Tuples with only one component become that component
+	vector<ASTPointer<Expression const>> callArguments;
+
+	if (
+		auto const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get());
+		argumentTuple && !argumentTuple->isInlineArray()
+	)
```
cameel: I think this would be more readable:
```suggestion
	auto const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get());
	if (argumentTuple && !argumentTuple->isInlineArray())
```

cameel: What about empty tuples? If they're not possible here, can you assert that `arguments` is not empty?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,97 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+			7788_error,
+			_functionCall.location(),
+			"Expected " +
+			to_string(functionPointerType->parameterTypes().size()) +
+			" instead of " +
+			to_string(callArguments.size()) +
+			" components for the tuple parameter."
+		);
+
+	size_t const numParameters = min(callArguments.size(), functionPointerType->parameterTypes().size());
```
cameel, line 2171: You're only using `min()` to make the error non-fatal and be able to continue, right? I think I'd add a comment about that.

cameel (commented):
Only managed to review about half of it so far. Here are some initial remarks. Mostly text corrections.

libsolidity/analysis/TypeChecker.cpp
Marenz: Hmm strictly speaking I think they should even be possible. but I should add a test case for that

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1236,22 +1236,39 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+
+			if (function.kind() == FunctionType::Kind::ABIEncodeCall)
 			{
-				arguments[i]->accept(*this);
-				// Do not keep the selector as part of the ABI encoded args
-				if (!hasSelectorOrSignature || i > 0)
-					argumentTypes.push_back(arguments[i]->annotation().type);
+				solAssert(arguments.size() == 2, "");
```
cameel: The message in assertion macros is now optional (#12019).
```suggestion
				solAssert(arguments.size() == 2);
```

libsolidity/codegen/ExpressionCompiler.cpp
cameel: I think the pushing to the stack is due to this part, right?
```c++
				if (!hasSelectorOrSignature || i > 0)
					argumentTypes.push_back(arguments[i]->annotation().type);
```
That's still preserved below.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2078,6 +2086,97 @@ void TypeChecker::typeCheckABIEncodeFunctions(
...
+	}
+
+	solAssert(!functionPointerType->takesArbitraryParameters(), "No checks possible for arbitrary parameters.");
+
+	// Tuples with only one component become that component
+	vector<ASTPointer<Expression const>> callArguments;
+
+	if (
+		auto const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get());
+		argumentTuple && !argumentTuple->isInlineArray()
```
cameel: Wow, I had no idea that `[1, 2, 3]` counts as a tuple in our type system.

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+// ----
+// TypeError 5407: (209-215): Cannot implicitly convert component at position 0 from "literal_string "test"" to "int256".
+// TypeError 7788: (280-310): Expected 1 instead of 2 components for the tuple parameter.
+// TypeError 7788: (373-399): Expected 1 instead of 0 components for the tuple parameter.
+// TypeError 6219: (462-484): Expected two arguments: a function pointer followed by a tuple.
+// TypeError 5511: (562-563): Expected first argument to be a function pointer, not "int_const 1".
+// TypeError 3509: (650-664): Function must be "public" or "external".
+// TypeError 7788: (747-778): Expected 2 instead of 1 components for the tuple parameter.
```
cameel: I think that this message will be confusing to users. I don't think we actually allow `[]` in any context where a tuple is expected so users won't see `[1, 2]` as a 2-component tuple. I think you should catch the case of inline arrays separately in the type checker and issue a message tailored to that.

Alternatively - maybe we could actually support inline arrays here?

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1236,22 +1236,39 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+
+				auto const tupleType = dynamic_cast<TupleType const*>(arguments[1]->annotation().type);
+
+				// Account for tuples with one component which become that component
+				if (!tupleType)
+					argumentTypes.emplace_back(arguments[1]->annotation().type);
+				else
+					argumentTypes = tupleType->components();
```
cameel: Can you assert that the tuple is not an inline array? Or would that case actually work if we allowed it in the type checker?

Same in the IR codegen.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1299,53 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
-						utils().fetchFreeMemoryPointer();
-						// stack: <memory pointer> <selector> <free mem ptr>
-						utils().packedEncode(TypePointers{selectorType}, TypePointers());
-						utils().toSizeAfterFreeMemoryPointer();
-						m_context << Instruction::KECCAK256;
-						// stack: <memory pointer> <hash>
-
-						dataOnStack = TypeProvider::fixedBytes(32);
+						m_context << Instruction::SWAP1 << Instruction::POP;
+						/// need to store it as bytes4
```
cameel:
```suggestion
						// need to store it as bytes4
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1299,53 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+						utils().leftShiftNumberOnStack(224);
 					}
```
cameel: Regarding our discussion on the call, it works like this: `this.f` does not need the `SWAP1; POP` sequence because the only thing at the top of the stack is the selector we have just calculated. There's no contract address there so no need to drop it. If you have a function pointer already on the stack though, the address is there and we need to get rid of it to get the same stack structure in both cases.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,60 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
 			functionType->kind() == FunctionType::Kind::ABIEncodeWithSelector ||
+			functionType->kind() == FunctionType::Kind::ABIEncodeCall ||
 			functionType->kind() == FunctionType::Kind::ABIEncodeWithSignature;
 
 		TypePointers argumentTypes;
 		TypePointers targetTypes;
 		vector<string> argumentVars;
-		for (size_t i = 0; i < arguments.size(); ++i)
+		string selector;
+		vector<ASTPointer<Expression const>> argumentsEncodeFunction;
```
```suggestion
		vector<ASTPointer<Expression const>> argumentsOfEncodeFunction;
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: Yeah. But if that's the case then what about nested tuples? Does `type()` handle them correctly?

I think it's worth giving it a try anyway to see if semantic tests still pass.

libsolidity/formal/SymbolicState.cpp
```diff
@@ -241,6 +241,7 @@ void SymbolicState::buildABIFunctions(set<FunctionCall const*> const& _abiFuncti
 			outTypes = returnTypes;
 			if (
 				t->kind() == FunctionType::Kind::ABIEncodeWithSelector ||
+				t->kind() == FunctionType::Kind::ABIEncodeCall ||
```
cameel: Doesn't this need special handling to treat tuple as a list of args rather than a single arg?

Maybe you could add a single SMT checker test that contains this function and check if it behaves reasonably?

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
```diff
@@ -0,0 +1,75 @@
...
+	}
+	function test3() public returns (bool) {
+		return keccak256(fPointerCall()) == keccak256(f0());
+	}
+	function test4() public returns (bool) {
+		return keccak256(fLocalPointerCall()) == keccak256(f0());
+	}
+	function test5() public returns (bool) {
+		return keccak256(fReturnedFunctionPointer()) == keccak256(f0());
+	}
```
cameel: I think that putting all the checks in one function would make these tests a bit easier to process. It would make it apparent that these are just boolean checks and not functions that are meant to be called by other functions.
```suggestion
	function assertConsistentSelectors() public {
		assert(keccak256(f0()) == keccak256(f0Call()));
		assert(keccak256(f1()) == keccak256(f1Call()));
		assert(keccak256(f1Call()) == keccak256(f1s()));
		assert(keccak256(fPointerCall()) == keccak256(f0()));
		assert(keccak256(fLocalPointerCall()) == keccak256(f0()));
		assert(keccak256(fReturnedFunctionPointer()) == keccak256(f0()));
	}
}
```

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call.sol
```diff
@@ -0,0 +1,44 @@
...
+		return keccak256(f1()) == keccak256(f1Call());
+	}
+	function test2() public returns (bool) {
+		return keccak256(f1Call()) == keccak256(f1s());
+	}
+
+	function f0() public pure returns (bytes memory) {
+		return abi.encodeWithSignature("f(uint256)", 1);
+	}
+	function f0Call() public view returns (bytes memory) {
```
cameel: You're testing multiple things here. I think names should be a bit more verbose to make it clear what is being tested. How about this?
- `f0` -> `fSignatureFromLiteral`,
- `f0Call` -> `fCall`,
- `f1` -> `fSignatureFromMemory`,
- `f1Call` -> it's identical with `f0Call`, you could drop it,
- `f1s` -> `fSignatureFromStorage`.

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
```diff
@@ -0,0 +1,75 @@
+pragma abicoder v2;
```
cameel: This test seems to be testing everything `abi_encode_call.sol` does and more. Do you actually need `abi_encode_call.sol`?

Also, I'd give it a more specific name. E.g. `abi_encode_call_is_consistent_with_abi_encode_with_signature.sol`.

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
```diff
@@ -0,0 +1,75 @@
...
+	}
+	string xstor;
+	function f1s() public returns (bytes memory) {
+		xstor = "f(uint256,string)";
+		return abi.encodeWithSignature(xstor, 1, "123");
+	}
+	function fPointerCall() public returns(bytes memory) {
+		fPointer = this.f;
+		return abi.encodeCall(fPointer, (1, "123"));
+	}
```
cameel: Can you add a case where the pointer is a local variable or a parameter? This can actually matter in some cases (see e.g. #12057).

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
```diff
@@ -0,0 +1,75 @@
...
+	function fLocalPointerCall() public returns(bytes memory) {
+		function(uint256, string memory) external localFunctionPointer = this.f;
+		return abi.encodeCall(localFunctionPointer, (1, "123"));
+	}
+	function getfunctionPointer() public returns (function(uint256, string memory) external) {
```
cameel:
```suggestion
	function getFunctionPointer() public returns (function(uint256, string memory) external) {
```

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
```diff
@@ -0,0 +1,45 @@
...
+	}
+	function success() public returns(bytes memory) {
+		return abi.encodeCall(this.f, (1));
+	}
+	function success2() public returns(bytes memory) {
+		return abi.encodeCall(this.f, 1);
+	}
+	function success3() public returns(bytes memory) {
+		return abi.encodeCall(this.f2, (1, "test"));
+	}
```
cameel: These also could have more meaningful names to explain what they test. Here's how I would name them:
- `fail()` -> `failFunctionArgsWrongType()`
- `fail2()` -> `failFunctionArgsTooMany()`
- `fail3()` -> `failFunctionArgsTooFew()`
- `fail4()` -> `failFunctionPtrMissing()`
- `fail5()` -> `failFunctionPtrWrongType()`
- `fail6()` -> `failFunctionPtrBuiltin()`
- `fail7()` -> `failFunctionArgsArrayLiteral()`
- `success()` -> `successFunctionArgsIntLiteralTuple()`
- `success2()` -> `successFunctionArgsIntLiteral()`
- `success3()` -> `successFunctionArgsLiteralTuple()`

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
```diff
@@ -0,0 +1,45 @@
+
+contract C {
+	function f(int a) public {}
```
cameel: I'd put this test in `syntaxTests/specialFunctions/encodeCall.sol`. That's where tests for other `abi.*` functions seem to be.

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
```diff
@@ -0,0 +1,45 @@
...
+	function f2(int a, string memory b) public {}
+	function f3(int a, int b) public {}
+
+	function fail() public returns(bytes memory) {
```
cameel: More syntax test ideas:
- Function pointer from:
    - [x] Internal function
    - [x] Variable of internal function type
    - [x] Library function
    - [x] Bound library function
    - [x] Interface function
- Function arguments:
   - [x] `()`
   - [x] `[]`
   - [x] `(x,)`
   - [x] `((x, y), (z, w))`
   - [ ] Result of a function call that returns multiple values
   - [ ] Value returned by a getter of a public state variable of struct type (it unpacks the struct into individual components, https://github.com/ethereum/solidity/issues/11826#issuecomment-902835510)

libsolidity/analysis/TypeChecker.cpp
Marenz: But why should they not be possible?

libsolidity/analysis/TypeChecker.cpp
Marenz: Yeah, was a surprise to me, too

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
Marenz: We support passing only a single non-tuple parameter which will be treated as a tuple with one component, in this case it would treat it as exactly that, a tuple with one array parameter...

libsolidity/analysis/TypeChecker.cpp
cameel: I mean that there could be some check earlier and we won't even reach this point if there's an empty tuple used.

libsolidity/codegen/ExpressionCompiler.cpp
Marenz: Hmm I wouldn't assert, just handle the case as a single parameter, like it is for non-tuple parameters..

test/libsolidity/syntaxTests/abiEncoder/encodeCallMismatch.sol
cameel: Ah, I got the message backwareds :)

But I still think the message could be improved. Can we say just `Expected 2-component tuple` when the input is just a single parameter (or a 1-component tuple)?

libsolidity/codegen/ExpressionCompiler.cpp
cameel: That's even better :)

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
Marenz: If we collect them all in one function though, we can't easily see which one failed if/when one fails..

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
Marenz: Just a few lines below exactly that is done?

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
cameel: True.

So how about this instead?
```solidity
	function assertConsistentSelectors() public returns (bool, bool, bool, bool, bool, bool) {
		return (
                    keccak256(f0()) == keccak256(f0Call()),
                    keccak256(f1()) == keccak256(f1Call()),
                    keccak256(f1Call()) == keccak256(f1s()),
                    keccak256(fPointerCall()) == keccak256(f0()),
                    keccak256(fLocalPointerCall()) == keccak256(f0()),
                    keccak256(fReturnedFunctionPointer()) == keccak256(f0())
                )
	}
```

I mean, the thing that bothers me here the most is that at a glance it's not clear that the `testX()` functions are the test itself and not part of the logic you're testing. So maybe just putting the `f` functions at the top would help a bit.

They could also have better names but I thought that putting them all in one function would sidesteps that problem more elegantly :)

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call1.sol
cameel: Not sure how I missed that ;)

libsolidity/analysis/TypeChecker.cpp
chriseth: Never trust the syntax, always use the type system.

libsolidity/codegen/ExpressionCompiler.cpp
chriseth:
```suggestion
				// Account for tuples with one component which become that component
				if (auto const* tupleType = dynamic_cast<TupleType const*>(arguments[1]->annotation().type))
					argumentTypes = tupleType->components();
				else
					argumentTypes.emplace_back(arguments[1]->annotation().type);
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1299,53 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
-					// hash the signature
-					if (auto const* stringType = dynamic_cast<StringLiteralType const*>(selectorType))
-					{
-						m_context << util::selectorFromSignature(stringType->value());
-						dataOnStack = TypeProvider::fixedBytes(4);
-					}
+					auto const functionPtr = dynamic_cast<FunctionTypePointer>(arguments[0]->annotation().type);
+					solAssert(functionPtr, "");
+
+					if (functionPtr->hasDeclaration())
```
chriseth: Does the `accept()` call above not put the funciton pointer on the stack as a value? I.e. I think you can remove the "if" part and only keep the "else" case.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1299,53 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
 					{
-						utils().fetchFreeMemoryPointer();
-						// stack: <memory pointer> <selector> <free mem ptr>
-						utils().packedEncode(TypePointers{selectorType}, TypePointers());
-						utils().toSizeAfterFreeMemoryPointer();
-						m_context << Instruction::KECCAK256;
-						// stack: <memory pointer> <hash>
-
-						dataOnStack = TypeProvider::fixedBytes(32);
+						m_context << Instruction::SWAP1 << Instruction::POP;
```
chriseth: Maybe add an assertion `functionPtr->sizeOnStack() == 2`

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1299,53 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+						/// need to store it as bytes4
+						utils().leftShiftNumberOnStack(224);
```
chriseth: Instead of shifting, you could set the `dataOnStack` to `uint32` and the conversion below will do the rest.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
chriseth: Yes, nested tuples should be handled properly.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,60 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
+			argumentTypes.emplace_back(&type(*argument));
+			targetTypes.emplace_back(type(*argument).fullEncodingType(false, true, isPacked));
+			argumentVars += IRVariable(*argument).stackSlots();
+		}
+
+		if (functionType->kind() == FunctionType::Kind::ABIEncodeCall)
+		{
+			auto const functionPtr = dynamic_cast<FunctionTypePointer>(&type(*arguments[0]));
+			solAssert(functionPtr, "");
+			if (functionPtr->hasDeclaration())
```
chriseth: I also don't see why we need this here. `IRVariable(*arguments[0])` should handle both cases.

chriseth:
Please add a test that uses `abi.encodeCall` together with `address.call` and use some more complex types (structs with dynamically-sized elements, user-defined value types etc.).

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
Marenz: It's still not pushed chris, I am still working on cameels comments :)

Marenz: Nevermind, your comment is valid non-the-less

Marenz: Though, as a response: Shouldn't we try to do as much as we can at compile-time and avoid runtime when we can? If we have the information already without going into the stack, shouldn't we use it?

Marenz: In this particular case it probably just saves as a conversation from 32 -> 4

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1282,35 +1304,54 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+				)
 				{
-					// hash the signature
-					if (auto const* stringType = dynamic_cast<StringLiteralType const*>(selectorType))
+					m_context << util::selectorFromSignature(functionPtr->externalSignature());
+					dataOnStack = TypeProvider::fixedBytes(4);
+				}
+				else
+				{
+					// stack: <selector> <memory pointer>
```
chriseth: This comment is wrong ATM

Marenz: Then the comment in line 1303 is also wrong. Why do you think it's wrong?

libsolidity/analysis/TypeChecker.cpp
Marenz: I mean, prefix is also a verb? I think both is valid.

libsolidity/codegen/ExpressionCompiler.cpp
chriseth: Because in case of `abi.encodeCall`, the stack contains a full function pointer and not just a selector.

libsolidity/analysis/TypeChecker.cpp
chriseth: I also like just prefix

libsolidity/codegen/ExpressionCompiler.cpp
Marenz: hmm I see. Then it was previously also wrong for the case of `EncodeWithSignature`, it just got one more meaning now. 
We can correct that by calling it .. well, I struggle to come up with a generic enough name. `selector-ish`?

chriseth: `selector/functionPointer/signature`

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1252,13 +1252,16 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+				solAssert(functionPtr);
 
+				auto const tupleExpression = dynamic_pointer_cast<TupleExpression const>(arguments[1]);
 				auto const tupleType = dynamic_cast<TupleType const*>(arguments[1]->annotation().type);
 
+				solAssert(!tupleType || tupleExpression, "Tuple type without tuple expression?!");
+
 				// Account for tuples with one component which become that component
-				if (!tupleType)
+				if (!tupleType || tupleExpression->isInlineArray())
```
chriseth: TupleType is only for tuples. So if the tuple expression is an inline array. its type will not be a tuple type.

chriseth:
This also needs a change in `void ViewPureChecker::endVisit(MemberAccess const& _memberAccess)`

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,65 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
 		{
-			// ignore selector
-			if (hasSelectorOrSignature && i == 0)
-				continue;
-			argumentTypes.emplace_back(&type(*arguments[i]));
-			targetTypes.emplace_back(type(*arguments[i]).fullEncodingType(false, true, isPacked));
-			argumentVars += IRVariable(*arguments[i]).stackSlots();
+			solAssert(arguments.size() == 2, "");
+			auto const tupleExpression = dynamic_pointer_cast<TupleExpression const>(arguments[1]);
+			auto const tupleType = dynamic_cast<TupleExpression const*>(arguments[1].get());
```
chriseth: I don't understand, this `dynamic_cast` should be redundant. Why do you need the two variables?
Also please don't name it `tupleType` - it is not a type, it is an expression.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,65 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
+
+			solAssert(!tupleType || tupleExpression, "Tuple type without tuple expression?!");
+			// Account for tuples with one component which become that component
+			if (!tupleType || tupleExpression->isInlineArray())
+				argumentsOfEncodeFunction.push_back(arguments[1]);
+			else
+				for (auto component: tupleExpression->components())
```
chriseth, line 1134: Did you test this in case the argument is not a tuple expression? This should result in a null pointer dereference.

libsolidity/codegen/ExpressionCompiler.cpp
Marenz: I am marking this as resolved as the updates made it obsolete

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1104,29 +1104,65 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
-			argumentTypes.emplace_back(&type(*arguments[i]));
-			targetTypes.emplace_back(type(*arguments[i]).fullEncodingType(false, true, isPacked));
-			argumentVars += IRVariable(*arguments[i]).stackSlots();
+			solAssert(arguments.size() == 2, "");
+			auto const tupleExpression = dynamic_pointer_cast<TupleExpression const>(arguments[1]);
+			auto const tupleType = dynamic_cast<TupleExpression const*>(arguments[1].get());
+
+			solAssert(!tupleType || tupleExpression, "Tuple type without tuple expression?!");
+			// Account for tuples with one component which become that component
+			if (!tupleType || tupleExpression->isInlineArray())
```
chriseth, line 1131: If `tupleTYpe` is null, then `tupleExpression` should be as well.

chriseth: Maybe shorter: `if (arguments[1].type->category() != Type::Category::Tuple)`

libsolidity/analysis/TypeChecker.cpp
Marenz: Well, we'd still need to validate that the target function also has zero arguments, not sure the extra complexity is worth the shortcut

chriseth (commented):
Code looks good now!
So the only things left are tests and potentially the SMT checker.

libsolidity/analysis/TypeChecker.cpp
cameel: Both are valid but I think that the meaning is different:
- `prefix "this."` = `add a prefix to "this."`. I.e. `this.x` -> `prefix.this.x`.
- `prefix it with "this."` = `add "this." as a prefix to it`. I.e. `x` -> `this.x`.

I mean, it's will probably be understood in context but technically the first version is wrong here.

cameel: ok, maybe it's my interpretation that's wrong after all: https://www.merriam-webster.com/dictionary/prefix:

> 1. \ (ˌ)prē-​ˈfiks: to fix or appoint beforehand
> 2. \ ˈprē-​ˌfiks, prē-​ˈfiks \ [partly from prefix entry 2] : to place in front
>     especially : to add as a prefix prefix a syllable to a word 
>     // prefix a syllable to a word

 Weird. I always understood "prefix something" as "prefix something (with)" = "add a prefix to something".

libsolidity/analysis/ViewPureChecker.cpp
```diff
@@ -367,6 +367,7 @@ void ViewPureChecker::endVisit(MemberAccess const& _memberAccess)
 			{MagicType::Kind::ABI, "encode"},
 			{MagicType::Kind::ABI, "encodePacked"},
 			{MagicType::Kind::ABI, "encodeWithSelector"},
+			{MagicType::Kind::ABI, "encodeWithCall"},
```
cameel:
```suggestion
			{MagicType::Kind::ABI, "encodeCall"},
```

I see that @chriseth commented on it but his comment appeared only on the commit and not in the PR (https://github.com/ethereum/solidity/commit/361e0b64ae989950db93ccb6161fd971ea0baa5d#r61910809).

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call.sol
```diff
@@ -0,0 +1,51 @@
...
+			5,
+			"TestName"
+		)));
+
+		return success;
+	}
+}
+
+// ====
+// compileViaYul: also
```
chriseth, line 49:
```suggestion
// compileViaYul: also
// EVMVersion: >=byzantium
```

chriseth: It might also help to remove `bytes memory data` above

Marenz:
So, SMT now says the function pointer parameter is not supported or so and only prints that as a warning. 
Is this good enough for now? I suspect the issue is maybe not with my addition in that case

chriseth:
Sounds good!

docs/cheatsheet.rst
```diff
@@ -80,6 +80,8 @@ Global Variables
   the given arguments. Note that this encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: :ref:`ABI <ABI>`-encodes
   the given arguments starting from the second and prepends the given four-byte selector
+- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the
+  tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```
chriseth:
```suggestion
  tuple. Performs a full type-check, ensuring the types match the function signature. Result equals ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```

docs/units-and-global-variables.rst
```diff
@@ -136,6 +136,7 @@ ABI Encoding and Decoding Functions
 - ``abi.encodePacked(...) returns (bytes memory)``: Performs :ref:`packed encoding <abi_packed_mode>` of the given arguments. Note that packed encoding can be ambiguous!
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: ABI-encodes the given arguments starting from the second and prepends the given four-byte selector
 - ``abi.encodeWithSignature(string memory signature, ...) returns (bytes memory)``: Equivalent to ``abi.encodeWithSelector(bytes4(keccak256(bytes(signature))), ...)``
+- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature. Equivalent to ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```
chriseth:
```suggestion
- ``abi.encodeCall(function functionPointer, (...)) returns (bytes memory)``: ABI-encodes a call to ``functionPointer`` with the arguments found in the tuple. Performs a full type-check, ensuring the types match the function signature. Result equals ``bytes.concat(functionPointer.selector, abi.encodeWithSelector(...))``
```

docs/cheatsheet.rst
chriseth: erm, shouldn't we rather use `abi.encodeWithSelector(functionPointer.selector, ...)`?

Marenz: You mean without the concat? Yeah, probably true

chriseth (approved):

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple.sol
```diff
@@ -0,0 +1,26 @@
...
+		assert(b1.length == b3.length); // should fail
+	}
+}
+// ====
+// SMTEngine: all
+// SMTIgnoreCex: yes
+// ----
+// Warning 6031: (233-249): Internal error: Expression undefined for SMT solver.
+// Warning 6031: (298-314): Internal error: Expression undefined for SMT solver.
+// Warning 6031: (398-414): Internal error: Expression undefined for SMT solver.
```
leonardoalt, line 20: That's not nice... I've been working on removing this kind of stuff, so it would be nice if this was fixed before merging.
I think this has to do with shortcutting `bool SMTEncoder::visit(FunctionCall const& _funCall)`, like it does with `ABIDecode`.

libsolidity/formal/SymbolicState.cpp
```diff
@@ -236,6 +236,15 @@ void SymbolicState::buildABIFunctions(set<FunctionCall const*> const& _abiFuncti
...
 				solAssert(false, "Unexpected argument of abi.decode");
 		}
+		else if (t->kind() == FunctionType::Kind::ABIEncodeCall)
+		{
+			// abi.encodeCall : (functionPointer, tuple_of_args_or_one_non_tuple_arg(arguments)) -> bytes
+			solAssert(args.size() == 2, "Unexpected number of arguments for abi.encodeCall");
+
+			outTypes.emplace_back(TypeProvider::bytesMemory());
+			inTypes.emplace_back(args.at(0)->annotation().type);
+			inTypes.emplace_back(args.at(1)->annotation().type);
```
leonardoalt, line 246: Isn't this the type of the tuple though?

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple.sol
leonardoalt: Actually no, now I think this is because of my other comment, since it's trying to get the smt expression of the tuple argument (still need to investigate more)

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple.sol
```diff
@@ -0,0 +1,26 @@
+contract C {
+	function callMeMaybe(uint a, uint b, uint[] memory c) external {}
+
+	function abiEncodeSimple(uint x, uint y, uint z, uint[] memory a, uint[] memory b) public view {
+		require(x == y);
+		bytes memory b1 = abi.encodeCall(this.callMeMaybe, (x, z, a));
+		bytes memory b2 = abi.encodeCall(this.callMeMaybe, (y, z, a));
+		assert(b1.length == b2.length);
```
leonardoalt, line 8: This assertion should hold because `x == y`. *This* might then be because it's using the tuple of arguments and not the arguments

leonardoalt: When I test it locally without the array being part of `callMeMaybe`, the solver finds a counterexample for it (that's why I wrote the above).

libsolidity/formal/SymbolicState.cpp
Marenz: Yep

Issue #6482, opened 2019-04-05
Explicit exporting of symbols


frangio:
Currently Solidity exports all symbols in a file, including those that were `import`ed from other files. Has there been discussion around changing this to an explicit exports model? It would be nice to be able to control what gets exported, to modularize an implementation while keeping that an implementation detail.

chriseth:
Which syntax would you like to suggest?

frangio:
I would go with a subset of JavaScript's syntax.

1) Exporting a symbol that is being defined.
```
export contract Foo { }
```

2) Exporting previously defined or imported symbols.
```
export { Bar, Qux };
```

Symbols can be contracts, interfaces, or libraries.

chriseth:
What about backwards compatibility? Export everything if no `export` keyword is used in the whole file? What about exporting symbols imported from other files?

frangio:
I think this should be a breaking change eventually: nothing should be exported unless explicitly requested. However, to ease transition, the semantics you describe are a good idea. Implicit exports could be deprecated some time after, and then removed in the next breaking version bump.

Exporting imported symbols is covered in part 2 of the syntax I suggested above.

chriseth:
Sounds good! Of course this has to be a breaking change. I'm adding it to the 0.6.0 discussion.

chriseth:
Especially with functions and other symbols at file-level, I think we should reconsider this.

I really like the `export { symbolA, symbolB }` syntax with the following semantics:

 - if there is no `export` statement, export all symbols can be imported into other files (and will be with a "simple" import statement) - this is the current behaviour
 - mulitiple `export` statements are disallowed
 - if `export {A, B, C}` is specified, only the specified symbols are accessible from outside, regardless of the way the import is specified. Whether the specified symbols are declared in the file itself or in turn imported is irrelevant.

This mechanism allows people to cleanly specified the external interface of a file at a single place.

frangio:
What is the argument for disallowing multiple `export` statements?

chriseth:
@frangio I don't see a big benefit and the downside is that you might not see the second export statement and think that the identifiers in the first is all that is exported.

chriseth:
Discussion: Main question here is if we want to have a statement that lists exports or if we want to mark individual declarations to be exported. Another question is if imports should be re-exported by default or not. And then of course there is the question if this feature is needed at all.

frangio:
[...]

PR #12288, opened 2021-11-16
Using global

chriseth:
The alternative syntax for this is
```
struct X { ... } using {f, mul as *};
type MyFloat is uint using {f};
```
The disadvantage is that you can only have one `using` statement per type and you cannot "mix" e.g. multiple modules, i.e. you cannot do `export using Modulename for MyFloat; export using {f, g, h} for MyFloat;`

We can of course change the `using` statement so that `using {f, M} for MyFloat;` is equivalent to `using f for MyFloat; using M for MyFloat` - currently modules and libraries cannot be used inside the braces.

### DIFF
--- a/docs/grammar/SolidityLexer.g4
+++ b/docs/grammar/SolidityLexer.g4
@@ -47,6 +47,7 @@ FixedBytes:
 	'bytes25' | 'bytes26' | 'bytes27' | 'bytes28' | 'bytes29' | 'bytes30' | 'bytes31' | 'bytes32';
 For: 'for';
 Function: 'function';
+Global: 'global'; // not a real keyword
 Hex: 'hex';
 If: 'if';
 Immutable: 'immutable';
--- a/libsolidity/ast/ASTJsonImporter.cpp
+++ b/libsolidity/ast/ASTJsonImporter.cpp
@@ -359,7 +359,8 @@ ASTPointer<UsingForDirective> ASTJsonImporter::createUsingForDirective(Json::Val
 		_node,
 		move(functions),
 		!_node.isMember("libraryName"),
-		_node["typeName"].isNull() ? nullptr  : convertJsonToASTNode<TypeName>(_node["typeName"])
+		_node["typeName"].isNull() ? nullptr  : convertJsonToASTNode<TypeName>(_node["typeName"]),
+		memberAsBool(_node, "global")
 	);
 }
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Language Features:
  * General: Allow annotating inline assembly as memory-safe to allow optimizations and stack limit evasion that rely on respecting Solidity's memory model.
  * General: ``using M for Type;`` is allowed at file level and ``M`` can now also be a brace-enclosed list of free functions or library functions.
+ * General: ``using ... for T global;`` is allowed at file level where the user-defined type ``T`` has been defined, resulting in the effect of the statement being available everywhere ``T`` is available.
 
 
 Compiler Features:
--- a/docs/contracts/using-for.rst
+++ b/docs/contracts/using-for.rst
@@ -42,6 +42,13 @@ scope (either the contract or the current module/source unit),
 including within all of its functions, and has no effect
 outside of the contract or module in which it is used.
 
+When the directive is used at file level and applied to a
+user-defined type which was defined at file level in the same file,
+the word ``global`` can be added at the end. This will have the
+effect that the functions are attached to the type everywhere
+the type is available (including other files), not only in the
+scope of the using statement.
+
 Let us rewrite the set example from the
 :ref:`libraries` section in this way, using file-level functions
 instead of library functions.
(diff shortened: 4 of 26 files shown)


chriseth:
This is the "export using" statement and actually much easier to implement than I thought.

chriseth:
What about `using {...} for MyType attached;` or something like that?

chriseth:
This is ready for review, the name of the keyword is still up for debate.

docs/grammar/SolidityLexer.g4
```diff
@@ -47,6 +47,7 @@ FixedBytes:
 	'bytes25' | 'bytes26' | 'bytes27' | 'bytes28' | 'bytes29' | 'bytes30' | 'bytes31' | 'bytes32';
 For: 'for';
 Function: 'function';
+Global: 'global'; // not a real keyword
```
ekpyron, line 50: This needs to be added to ``identifier`` in the parser part now, like ``From``, ``Error`` and ``Revert``, doesn't it :-)?
We should probably add a test case for using ``global`` as identifier, that would need this :-).
And an issue for making it a keyword in 0.9... ~~although I'm not entirely sure if ``export using`` wasn't actually nicer :-).~~

ekpyron: But I guess this also depends on https://github.com/ethereum/solidity/issues/6482, resp. if we want anything like that... if so, it may be nice to work towards syntax that can be reused... EDIT: [note from chat: it really isn't exporting, but an effect on the type, so ``export`` doesn't really fit here]

chriseth:
Idea for test:
file A: define type `X`, function `f` and using... global
file B: import `X` from A and define a function `g` that returns `X`
file C: import only `g` from B and use `g().f()`.

It should work even though neither `f` nor `X` are available in `C` by name.

ekpyron:
Maybe
```
==== Source: A ====
using {f} for S global;
using {g} for S;
struct S { uint x; }
function gen() pure returns (S memory) {}
function f(S memory _x) pure { _x.g(); }
function g(S memory _x) pure { }
==== Source: B ====
import "A";
function test() pure
{
    gen().f(); // fine
    gen().g(); // some error
}
```
or something similar to check global and non-global using together?

In general this looks good, but may need a few more tests.

ekpyron:
If I see it correctly, the ``global`` thing still needs to be exported to and imported from the json AST.

ekpyron (commented):
Looks good in general apart from the missing json ast part - but I also need another look after a rebase.

ekpyron:
AST Json tests need updating. Looking at the rest regardless.

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -359,7 +359,8 @@ ASTPointer<UsingForDirective> ASTJsonImporter::createUsingForDirective(Json::Val
 		_node,
 		move(functions),
 		!_node.isMember("libraryName"),
-		_node["typeName"].isNull() ? nullptr  : convertJsonToASTNode<TypeName>(_node["typeName"])
+		_node["typeName"].isNull() ? nullptr  : convertJsonToASTNode<TypeName>(_node["typeName"]),
+		_node.isMember("global") && _node["global"].asBool()
```
ekpyron:
```suggestion
		_node.isMember("global") && memberAsBool(_node, "global")
```
?
Or only ``memberAsBool`` or is the idea to default to ``false`` instead of raising an error?

ekpyron:
Just for the sake of completeness, it could have some tests for ``global`` as identifier or type name and ``type global is uint; using { f } for global; using { g } for global global;`` - but on the other hand all that should work just fine...

libsolidity/ast/ASTJsonImporter.cpp
chriseth: The idea is to default to false.

ekpyron: ``_node.isMember("global") && memberAsBool(_node, "global")`` would still be more graceful I guess, if ``global`` is there and not a boolean, but yeah, fine either way.

chriseth:
Added the test.

ekpyron:
I'm mainly wondering if we need more tests at this point...
But even weirdest stuff like this should just work:
```
==== Source: A ====
import {T as U} from "A";
import "A" as X;

type T is uint;
function f(T) pure {}
using { f } for X.X.U global;
using { f } for T global;

function g() pure {
        U t;
        t.f();
}
```
So maybe coverage is already fine...

libsolidity/ast/ASTJsonImporter.cpp
ekpyron: Now it *won't* default to ``false`` anymore, though, will it :-)? But yeah, it also doesn't have to, so also fine.

ekpyron (dismissed):
Looks good to me - should be squashed before merging, though.

chriseth:
Added another semantic test and squashed.

ekpyron (approved):

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove (commented):
Thanks

Sticknmove (commented):
Thanks

Issue #12253, opened 2021-11-08
Move test cases from SolidityEndToEnd to semanticTests


axic:
There is the `test/libsolidity/SolidityEndToEndTest.cpp` file, which contains a lot of semantic tests.  It is a large monolithic single C++ file, which takes a long time to compile and is hard to maintain.

We want to move as many test cases out to individual files under `test/libsolidity/semanticTests`.  They should be grouped logically, by following the grouping in SolidityEndToEnd.

Have a look at existing test case in both places to see how they operate.

(This was previously designed/discussed in #3486, but creating a new issue to be more helpful.)

PR #12289, opened 2021-11-16
trial test cases extracted from SolidityEndToEndTest.cpp into .sol files

nishant-sachdeva:
Test cases have been extracted from SolidityEndtoEnd.cpp

Partially solves issue #12253


chriseth:
You accidentally included some files that should not be there (evmone etc) - can you please remove them?

test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -1650,37 +1606,38 @@ BOOST_AUTO_TEST_CASE(bytes_from_calldata_to_memory)
...
-		ABI_CHECK(callContractFunction("forward()"), encodeArgs(true));
-		ABI_CHECK(callContractFunction("val()"), encodeArgs(8));
-		ABI_CHECK(callContractFunction("clear()"), encodeArgs(true));
-		ABI_CHECK(callContractFunction("val()"), encodeArgs(8));
-		ABI_CHECK(callContractFunction("forward()"), encodeArgs(true));
-		ABI_CHECK(callContractFunction("val()"), encodeArgs(0x80));
-	);
-}
+// // extractable
+// BOOST_AUTO_TEST_CASE(call_forward_bytes)
```
chriseth: Why is this commented out?

nishant-sachdeva: Updates : 
1. Removed the libevmone.so* files. 
2. The commented regions are only there for this draft PR. I was looking to use this as a safety net in case I needed to revert any particular test case made, since I extracted 7 test cases in the first commit itself. An alternate to this should have been to extract one test case per commit. I'll be following that from now on.

cameel: If you want to revert a part of the commit you can always do an interactive rebase, set the commit to `edit` and amend the commit.

test/libsolidity/semanticTests/call_forward_bytes.sol
```diff
@@ -0,0 +1,26 @@
...
+    constructor() { rec = new receiver(); }
+    fallback() external { savedData = msg.data; }
+    function forward() public returns (bool) { address(rec).call(savedData); return true; }
+    function clear() public returns (bool) { delete savedData; return true; }
+    function val() public returns (uint) { return rec.received(); }
+    receiver rec;
+    bytes savedData;
+}
+// ====
+// compileViaYul: also
```
cameel: The original test had `DISABLE_EWASM_TESTRUN()`, which I think is equivalent to `// compileToEwasm: false`.

But if the test still passes without it then great.

test/libsolidity/semanticTests/call_forward_bytes.sol
```diff
@@ -0,0 +1,26 @@
...
+// ----
+// () ->
+// recv(): 7 ->
+// val() -> 0
+// forward() -> true
+// val() -> 0x80
+// clear() -> true
+// val() -> 0x80
+// forward() -> true
+// val() -> 0x80
```
cameel: Would be best to configure your editor to always keep a newline at the end. Otherwise it annoyingly changes back and forth in diffs when people edit the same file.

test/libsolidity/semanticTests/call_forward_bytes.sol
```diff
@@ -0,0 +1,26 @@
+contract receiver {
```
cameel: I'd avoid adding more tests directly in `semanticTests/`. `semanticTests/fallback/` might be a better place for this one.

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
```diff
@@ -0,0 +1,13 @@
...
+    function f() public {
+        sender = Lib.m();
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// library: Lib
+// f() ->
+// sender() -> 0x1212121212121212121212121212120000000012
```
cameel, line 15: @chriseth Do we have a helper for library addresses or do we always just hard-code them?

test/libsolidity/semanticTests/libraries/library_staticcall_delegatecal.sol
```diff
@@ -0,0 +1,20 @@
+library Lib {
```
cameel: Typo in test file name: `delegatecal` -> `delegatecall`.

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
```diff
@@ -0,0 +1,13 @@
+library Lib { function m() public returns (address) { return msg.sender; } }
```
cameel, line 1: Interesting that this case does not need to set EVM version even though the name suggests that it should. Maybe it's testing stuff that would not work before homestead rather than after it?

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
```diff
@@ -0,0 +1,28 @@
+contract receiver {
```
cameel: This test should probably go into `semanticTests/array/copying/`.

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
```diff
@@ -0,0 +1,28 @@
...
+    }
+    function val() public returns (uint) { return rec.received(); }
+    receiver rec;
+    bytes savedData1;
+    bytes savedData2;
+}
+// ====
+// compileViaYul: also
+// ----
+// val() -> 0
```
cameel: This is missing the `recv()` call from the original test. I guess it's because you could not call it on the `receiver` contract?

@chriseth Do we have support for that? `TestFileParser` does not seem to have code for handling something like `receiver.recv()`.

test/libsolidity/semanticTests/calldata/copy_from_calldata_removes_bytes_data.sol
```diff
@@ -0,0 +1,10 @@
+contract c {
+    function set() public returns (bool) { data = msg.data; return true; }
+    fallback() external { data = msg.data; }
+    bytes data;
+}
+// ====
+// compileViaYul: also
+// ----
+// set(): 1, 2, 3, 4, 5 -> true
+// gas legacy: 177656
```
cameel: The original test calls the `fallback` here.

cameel (commented):
Here's some feedback from me. Since this is still draft and some stuff seems to be missing (not all expectations are there) I only reviewed part of it.

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
chriseth: I think this is the hardcoded msg.sender - isn't it?

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
chriseth: This tests a new feature of homestead: delegatecall. We should restrict the evm version to `>= homestead`.

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
chriseth: A vaguely remember that there is a feature that allows isoltest to call a non-existing function.

chriseth:
Can you maybe isolate out the test cases that do not have any review comments into their own PR so it can be merged quicker?

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
nishant-sachdeva: Added the recv() call to the test case. Indeed, the test case fails to compile. 

Error message is : 
`Exception during test: /solidity/test/libsolidity/SemanticTest.cpp(433): Throw in function solidity::frontend::test::TestCase::TestResult solidity::frontend::test::SemanticTest::runTest(std::ostream&, const string&, bool, bool, bool)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: The function recv() is not known to the compiler`

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
nishant-sachdeva: Moved to array/copying

test/libsolidity/semanticTests/libraries/library_staticcall_delegatecal.sol
nishant-sachdeva: Thanks for pointing this out. 
Has been corrected.

test/libsolidity/semanticTests/call_forward_bytes.sol
nishant-sachdeva: Changed the compileToEwasm to false

But the test case was failing before also. 
`Exception during test: /solidity/test/libsolidity/SemanticTest.cpp(433): Throw in function solidity::frontend::test::TestCase::TestResult solidity::frontend::test::SemanticTest::runTest(std::ostream&, const string&, bool, bool, bool)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: The function recv() is not known to the compiler`

test/libsolidity/semanticTests/call_forward_bytes.sol
nishant-sachdeva: Moved test case to semanticTests/fallback

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
nishant-sachdeva: Is this okay ? Can we add this to our PR , or is there some change needed ?

test/libsolidity/semanticTests/call_forward_bytes.sol
nishant-sachdeva: Thanks for pointing out! Will take care of the same in the future

test/libsolidity/semanticTests/call_forward_bytes.sol
nishant-sachdeva: Update : Fixed

test/libsolidity/semanticTests/libraries/library_call_in_homestead.sol
nishant-sachdeva: Added  `EVMVersion: >=homestead` to the test case file

test/libsolidity/semanticTests/various/copying_bytes_multiassign.sol
nishant-sachdeva: Figured that the primary role of the recv() function call was to invoke the fallback() function within the contract. So, we replaced recv() with a call to the constructor ( since the parameters are incorrect, the fallback() function will be invoked )

test/libsolidity/semanticTests/calldata/copy_from_calldata_removes_bytes_data.sol
nishant-sachdeva: Update : Added relevant code to the test case to make it call the fallback()

chriseth:
It seems you still have failing tests here.

nishant-sachdeva:
The following tests cases run on my local system using isoltest. But CI/CD on github shows them to be failing. 

1. library_function_external.sol ( works on local but output mismatch on Github )
2. copy_from_calldata_remove_bytes_data.sol ( works on local but Compile Error on Github )

test/libsolidity/semanticTests/calldata/copy_from_calldata_removes_bytes_data.sol
```diff
@@ -0,0 +1,16 @@
...
+    function set() public returns (bool) { data = msg.data; return true; }
+    function checkIfDataIsEmpty() public returns (bool) { return data.length == 0; }
+    function sendMessage() public returns (bool, bytes memory) { bytes memory emptyData; return address(this).call(emptyData);}
+    fallback() external { data = msg.data; }
+    bytes data;
+}
+// ====
+// compileToEwasm: false
+// compileViaYul: also
+// ----
```
chriseth, line 12: ``// EVMVersion: >=byzantium``

test/libsolidity/semanticTests/fallback/call_forward_bytes.sol
```diff
@@ -0,0 +1,26 @@
...
+    function clear() public returns (bool) { delete savedData; return true; }
+    function val() public returns (uint) { return rec.received(); }
+    receiver rec;
+    bytes savedData;
+}
+// ====
+// compileToEwasm : false
+// compileViaYul: also
+// ----
+// (): 7 ->
```
chriseth:
```suggestion
// recv(uint256): 7 ->
```

test/libsolidity/semanticTests/fallback/call_forward_bytes.sol
```diff
@@ -0,0 +1,26 @@
...
+// val() -> 0
+// forward() -> true
+// val() -> 0x80
```
chriseth: Used to be 8

test/libsolidity/semanticTests/fallback/call_forward_bytes.sol
nishant-sachdeva: We get this error, when we use recv(uint256) instead of () . 
`semanticTests/fallback/call_forward_bytes.sol: Exception during test: /solidity/test/libsolidity/SemanticTest.cpp(436): Throw in function solidity::frontend::test::TestCase::TestResult solidity::frontend::test::SemanticTest::runTest(std::ostream&, const string&, bool, bool, bool)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: The function recv(uint256) is not known to the compiler
`

We figured that, calling recv() should only invoke the fallback function of the sender contract.  Adding a call to `(): 7 -> ` would achieve that as well. And in both cases the next call to `val()` should return 0.

nishant-sachdeva: The other option was to find a way to call recv() function from the receiver contract. 
I've added a call to rec.recv(7) inside the constructor in the latest push.
Will that be okay ?

chriseth: You can prevent the error by adding `// allowNonExistingFunctions: true` to the settings of the test. Please don't change the test from how it was.

nishant-sachdeva: Reverted the change. Added the  `//allowNonExistingFunctions: true` line to the settings of the test

test/libsolidity/semanticTests/fallback/call_forward_bytes.sol
nishant-sachdeva: From what I've understood from the logic of the test code, it should be 0 only.

The call to recv() will make way for a call to the fallback. That will merely store the savedData.
After that, the call to val() should be returning a 0 only, since that contract has not been touched by the control flow of the code yet

chriseth (approved):

PR #11350, opened 2021-05-05
Language Server

christianparpart:
This PR contains the Language Server mode for solc. Closes #7763.

Patch for LSP TCP mode saved [here](https://gist.github.com/christianparpart/e300b2c67353811fe1141f641633213c).

PR #11403, opened 2021-05-17
factor out some StandardCompiler' input JSON parsing/configuration…

christianparpart:
factor out some parsing/configuration API to be able to use them at other places too.

This refactoring is needed in order to share some code bits with the LSP PR to avoid code duplication.

PR #12304, opened 2021-11-22
LSP related refactors

christianparpart:
This is the first 6 commits of the LSP PR that are **not** the LSP but some heads-up work.

Required by #11350

NOTE: the failing CLI tests I have fixed but have to wait until tonight to push it (wrong VM :D)

### DIFF
--- a/liblangutil/CharStream.cpp
+++ b/liblangutil/CharStream.cpp
@@ -144,3 +144,37 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
 
 	return cut;
 }
+
+optional<int> CharStream::translateLineColumnToPosition(int _line, int _column) const
+{
+	return translateLineColumnToPosition(m_source, _line, _column);
+}
+
+optional<int> CharStream::translateLineColumnToPosition(std::string const& _text, int _line, int _column)
+{
+	size_t offset = 0;
+
+	while (_line > 0)
+	{
+		offset = _text.find('\n', offset);
+		if (offset == _text.npos)
+			return nullopt;
+		offset++; // Skip linefeed.
+		_line--;
+	}
+
+	auto const nextEndOfLineOffset = _text.find('\n', offset);
+	if (nextEndOfLineOffset != string::npos)
+	{
+		if (offset + static_cast<size_t>(_column) < nextEndOfLineOffset)
+			return static_cast<int>(offset) + _column; // Column fitting current non-last line.
+		else
+			return nullopt; // Column-overflow in current non-last line.
+	}
+
+	if (offset + static_cast<size_t>(_column) < _text.size())
+		return static_cast<int>(offset) + _column; // Column fitting into last line
+
+	return nullopt;
+}
+
--- a/liblangutil/CharStream.h
+++ b/liblangutil/CharStream.h
@@ -51,6 +51,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 #include <string>
 #include <tuple>
 #include <utility>
@@ -97,8 +98,18 @@ public:
 	/// Functions that help pretty-printing parse errors
 	/// Do only use in error cases, they are quite expensive.
 	std::string lineAtPosition(int _position) const;
+
+	/// @returns (1-based) line and column that matches to the byte offset @p _position.
 	std::tuple<int, int> translatePositionToLineColumn(int _position) const;
 	///@}
+
+	/// Translates a line:column to the absolute position.
+	/// Parameters @p _line and @p _column are 0-based indices.
+	std::optional<int> translateLineColumnToPosition(int _line, int _column) const;
+
+	/// Translates a line:column to the absolute position for the given input text.
+	/// Parameters @p _line and @p _column are 0-based indices.
+	static std::optional<int> translateLineColumnToPosition(std::string const& _text, int _line, int _column);
 
 	/// Tests whether or not given octet sequence is present at the current position in stream.
 	/// @returns true if the sequence could be found, false otherwise.
--- a/libsolidity/ast/ASTUtils.cpp
+++ b/libsolidity/ast/ASTUtils.cpp
@@ -17,12 +17,47 @@
 // SPDX-License-Identifier: GPL-3.0
 
 #include <libsolidity/ast/AST.h>
+#include <libsolidity/ast/ASTVisitor.h>
 #include <libsolidity/ast/ASTUtils.h>
 
 #include <libsolutil/Algorithms.h>
 
 namespace solidity::frontend
 {
+
+namespace
+{
+
+class ASTNodeLocator: public ASTConstVisitor
+{
+public:
+	explicit ASTNodeLocator(int _pos): m_offsetInFile{_pos}, m_innermostMatch{nullptr} {}
+
+	bool visitNode(ASTNode const& _node) override
+	{
+		// In the AST parent location always covers the whole child location.
+		// The parent is visited first so to get the innermost node we simply
+		// take the last one that still contains the offset.
+
+		if (!_node.location().containsOffset(m_offsetInFile))
+			return false;
+
+		m_innermostMatch = &_node;
+		return true;
+	}
+
+	int const m_offsetInFile;
+	ASTNode const* m_innermostMatch;
+};
+
+}
+
+ASTNode const* locateInnermostASTNode(int _pos, SourceUnit const& _sourceUnit)
+{
+	ASTNodeLocator locator{_pos};
+	_sourceUnit.accept(locator);
+	return locator.m_innermostMatch;
+}
 
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl)
 {
--- a/libsolidity/ast/ASTUtils.h
+++ b/libsolidity/ast/ASTUtils.h
@@ -21,9 +21,11 @@
 namespace solidity::frontend
 {
 
-class VariableDeclaration;
+class ASTNode;
 class Declaration;
 class Expression;
+class SourceUnit;
+class VariableDeclaration;
 
 /// Find the topmost referenced constant variable declaration when the given variable
 /// declaration value is an identifier. Works only for constant variable declarations.
@@ -32,5 +34,8 @@ VariableDeclaration const* rootConstVariableDeclaration(VariableDeclaration cons
 
 /// Returns true if the constant variable declaration is recursive.
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
+
+/// Teturns the innermost AST node that covers the given location or nullptr if not found.
+ASTNode const* locateInnermostASTNode(int _pos, SourceUnit const& _sourceUnit);
 
 }
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -1447,7 +1447,7 @@ CompilerStack::Source const& CompilerStack::source(string const& _sourceName) co
 {
 	auto it = m_sources.find(_sourceName);
 	if (it == m_sources.end())
-		solThrow(CompilerError, "Given source file not found.");
+		solThrow(CompilerError, "Given source file not found: " + _sourceName);
 
 	return it->second;
 }
--- a/libsolidity/interface/StandardCompiler.cpp
+++ b/libsolidity/interface/StandardCompiler.cpp
@@ -1065,12 +1065,15 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
 	return { std::move(ret) };
 }
 
-Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSettings _inputsAndSettings)
+unique_ptr<CompilerStack> StandardCompiler::createCompilerStack(
+	ReadCallback::Callback _readFile,
+	StandardCompiler::InputsAndSettings const& _inputsAndSettings
+)
 {
-	CompilerStack compilerStack(m_readFile);
+	auto compilerStackPtr = make_unique<CompilerStack>(_readFile);
+	CompilerStack& compilerStack = *compilerStackPtr;
 
-	StringMap sourceList = std::move(_inputsAndSettings.sources);
-	compilerStack.setSources(sourceList);
+	compilerStack.setSources(_inputsAndSettings.sources);
 	for (auto const& smtLib2Response: _inputsAndSettings.smtLib2Responses)
 		compilerStack.addSMTLib2Response(smtLib2Response.first, smtLib2Response.second);
 	compilerStack.setViaIR(_inputsAndSettings.viaIR);
@@ -1090,6 +1093,14 @@ Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSetting
 	compilerStack.enableEvmBytecodeGeneration(isEvmBytecodeRequested(_inputsAndSettings.outputSelection));
 	compilerStack.enableIRGeneration(isIRRequested(_inputsAndSettings.outputSelection));
 	compilerStack.enableEwasmGeneration(isEwasmRequested(_inputsAndSettings.outputSelection));
+
+	return compilerStackPtr;
+}
+
+Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSettings _inputsAndSettings)
+{
+	auto compilerStackPtr = createCompilerStack(m_readFile, _inputsAndSettings);
+	auto& compilerStack = *compilerStackPtr;
 
 	Json::Value errors = std::move(_inputsAndSettings.errors);
 
@@ -1294,7 +1305,7 @@ Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSetting
 		// EVM
 		Json::Value evmData(Json::objectValue);
 		if (compilationSuccess && isArtifactRequested(_inputsAndSettings.outputSelection, file, name, "evm.assembly", wildcardMatchesExperimental))
-			evmData["assembly"] = compilerStack.assemblyString(contractName, sourceList);
+			evmData["assembly"] = compilerStack.assemblyString(contractName, _inputsAndSettings.sources);
 		if (compilationSuccess && isArtifactRequested(_inputsAndSettings.outputSelection, file, name, "evm.legacyAssembly", wildcardMatchesExperimental))
 			evmData["legacyAssembly"] = compilerStack.assemblyJSON(contractName);
 		if (isArtifactRequested(_inputsAndSettings.outputSelection, file, name, "evm.methodIdentifiers", wildcardMatchesExperimental))
--- a/libsolidity/interface/StandardCompiler.h
+++ b/libsolidity/interface/StandardCompiler.h
@@ -67,7 +67,6 @@ public:
 		std::map<std::string, evmasm::LinkerObject::FunctionDebugData> const& _debugInfo
 	);
 
-private:
 	struct InputsAndSettings
 	{
 		std::string language;
@@ -89,6 +88,12 @@ private:
 		bool viaIR = false;
 	};
 
+	static std::unique_ptr<CompilerStack> createCompilerStack(
+		ReadCallback::Callback _readFile,
+		StandardCompiler::InputsAndSettings const& _inputsAndSettings
+	);
+
+private:
 	/// Parses the input json (and potentially invokes the read callback) and either returns
 	/// it in condensed form or an error as a json object.
 	std::variant<InputsAndSettings, Json::Value> parseInput(Json::Value const& _input);
--- a/test/liblangutil/CharStream.cpp
+++ b/test/liblangutil/CharStream.cpp
@@ -48,6 +48,62 @@ BOOST_AUTO_TEST_CASE(test_fail)
 	);
 }
 
+#define testLineColumnToPositionTranslationFail(_line, _column, _text) \
+	do                                                                       \
+	{                                                                        \
+		auto const source = std::make_shared<CharStream>((_text), "source"); \
+		auto const actualPosition =                                          \
+			source->translateLineColumnToPosition((_line), (_column));       \
+		BOOST_CHECK_EQUAL(actualPosition.value_or(-1), -1);                  \
+	}                                                                        \
+	while (0)
+
+#define testLineColumnToPositionTranslation(_line, _column, _expectedPosition, _text) \
+	do                                                                       \
+	{                                                                        \
+		auto const source = std::make_shared<CharStream>((_text), "source"); \
+		auto const actualPosition =                                          \
+			source->translateLineColumnToPosition((_line), (_column));       \
+		BOOST_CHECK(actualPosition.has_value());                             \
+		BOOST_CHECK_EQUAL(*actualPosition, (_expectedPosition));             \
+	}                                                                        \
+	while (0)
+
+BOOST_AUTO_TEST_CASE(translateLineColumnToPosition)
+{
+	testLineColumnToPositionTranslationFail(0, 0, "");
+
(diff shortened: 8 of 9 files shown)


liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,31 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
...
+	return translateLineColumnToPosition(m_source, _line, _column);
+}
+
+optional<int> CharStream::translateLineColumnToPosition(std::string const& _text, int _line, int _column)
+{
+	size_t offset = 0;
+
+	while (_line > 0)
+	{
+		if (auto const found = _text.find('\n', offset); found != _text.npos)
```
cameel: Why not keep the assignment out of the `if` for better readability?
Also, found sounds like a boolean variable but it isn't one.
```suggestion
		size_t const nextNewlinePosition = _text.find('\n', offset);
		if (nextNewlinePosition != _text.npos)
```

liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,31 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
...
+			_line--;
+		}
+		else
+			return nullopt;
+	}
+
+	if (offset + size_t(_column) >= _text.size())
+		return nullopt;
+
+	return int(offset) + _column;
```
cameel:
```suggestion
	if (offset + static_cast<size_t>(_column) >= _text.size())
		return nullopt;

	return static_cast<int>(offset) + _column;
```

liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,31 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
...
+
+// @p _line and @p _column should be 1-based, because it's not an index but a line/column number.
+optional<int> CharStream::translateLineColumnToPosition(int _line, int _column) const
+{
+	return translateLineColumnToPosition(m_source, _line, _column);
+}
+
+optional<int> CharStream::translateLineColumnToPosition(std::string const& _text, int _line, int _column)
+{
+	size_t offset = 0;
```
cameel, line 155: You could just use `nextNewlinePosition + 1` instead of this var.

liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,31 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
 
 	return cut;
 }
+
+// @p _line and @p _column should be 1-based, because it's not an index but a line/column number.
```
cameel: This should probably be a part of the docstring in the header.

It should also say what happens then column is greater than line length or line is past the end of the text. I see that for example it can actually point at a char on the next line if the column exceeds line length, which is not obvious.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1447,7 +1447,7 @@ CompilerStack::Source const& CompilerStack::source(string const& _sourceName) co
 {
 	auto it = m_sources.find(_sourceName);
 	if (it == m_sources.end())
-		solThrow(CompilerError, "Given source file not found.");
+		solThrow(CompilerError, "Given source file not found. " + _sourceName);
```
```suggestion
		solThrow(CompilerError, "Given source file not found: " + _sourceName);
```

libsolidity/interface/StandardCompiler.h
```diff
@@ -89,6 +88,10 @@ class StandardCompiler
 		bool viaIR = false;
 	};
 
+	/// Configures @p _compilerStack according to the input settings in @p _inputsAndSettings.
+	static void configure(InputsAndSettings const& _inputsAndSettings, CompilerStack& _compilerStack);
```
cameel: Reposting my comment from https://github.com/ethereum/solidity/pull/11403#discussion_r633572409:

> Why not make it a factory method? Do you actually need to be able to reconfigure an existing instance of `CompilerStack` in LSP?
> ```suggestion
> 	static CompilerStack makeCompilerStack(InputsAndSettings const& _inputsAndSettings, ReadCallback::Callback _readCallback);
> ```
>
> Also, it might be better to have it as a factory method on `CompilerStack` instead. I think it makes more sense than > exposing it as an internal implementation detail of `StandardCompiler`.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -756,153 +883,9 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
 	}
 
 	Json::Value const& settings = _input.get("settings", Json::Value());
-
-	if (auto result = checkSettingsKeys(settings))
+	if (auto result = parseInputSettings(settings, ret))
```
cameel:
```suggestion
	if (optional<Json::Value> result = parseInputSettings(settings, ret))
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1065,31 +1048,33 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
...
+	_compilerStack.setLibraries(_inputsAndSettings.libraries);
+	_compilerStack.useMetadataLiteralSources(_inputsAndSettings.metadataLiteralSources);
+	_compilerStack.setMetadataHash(_inputsAndSettings.metadataHash);
+	_compilerStack.setRequestedContractNames(requestedContractNames(_inputsAndSettings.outputSelection));
+	_compilerStack.setModelCheckerSettings(_inputsAndSettings.modelCheckerSettings);
+
+	_compilerStack.enableEvmBytecodeGeneration(isEvmBytecodeRequested(_inputsAndSettings.outputSelection));
+	_compilerStack.enableIRGeneration(isIRRequested(_inputsAndSettings.outputSelection));
+	_compilerStack.enableEwasmGeneration(isEwasmRequested(_inputsAndSettings.outputSelection));
+}
```
cameel: This function currently ignores `InputsAndSettings.language`. It was fine when the code was a part of `compileSolidity()` but now it's more generic, we should at least assert that the language is Solidity.

Or, better, it should create either `CompilerStack` or `AssemblyStack` based on the language. Or we should have two separate functions for that (e.g. `configureForSoldity()` and `conigureForYul()`), which might be more convenient - we'd  avoid having to return a variant that way.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -623,8 +623,135 @@ std::variant<OptimiserSettings, Json::Value> parseOptimizerSettings(Json::Value
 	return { std::move(settings) };
 }
 
+/// Only parses the input settings and fills the InputsAndSettings with it.
+/// If this fails, it does return a JSON error response, otherwise nullopt on success.
+optional<Json::Value> parseInputSettings(Json::Value _settings, StandardCompiler::InputsAndSettings& _output)
```
cameel: @chriseth's comments from the old PR that are still relevant (https://github.com/ethereum/solidity/pull/11403#discussion_r638950553):

> Please try to avoid output parameters at all costs. Really. At all costs. There is no need for this in C++17.

> This is a rather big function - maybe it's better to have this as a static member in the class?

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -623,8 +623,135 @@ std::variant<OptimiserSettings, Json::Value> parseOptimizerSettings(Json::Value
...
+		std::optional<langutil::EVMVersion> version = langutil::EVMVersion::fromString(_settings["evmVersion"].asString());
+		if (!version)
+			return formatFatalError("JSONError", "Invalid EVM version requested.");
+		_output.evmVersion = *version;
+	}
+
+	if (_settings.isMember("debug"))
+	{
+		if (auto result = checkKeys(_settings["debug"], {"revertStrings"}, "settings.debug"))
+			return *result;
```
cameel: This part is out of date and should be redone on top of current `develop`. In the current form it reverts everything related to debug info that was added in #11944. Note for example that in this diff `{"revertStrings", "debugInfo"}` gets reverted back to `{"revertStrings"}`. Same with the whole `if (settings["debug"].isMember("debugInfo"))` condition that should be just below.

libsolidity/ast/ASTUtils.h
```diff
@@ -33,4 +35,7 @@ VariableDeclaration const* rootConstVariableDeclaration(VariableDeclaration cons
 /// Returns true if the constant variable declaration is recursive.
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
 
+/// Locates closest AST node match within a SourceUnit given its absolute file offset or nullptr if none matched.
+ASTNode const* locateASTNode(int _pos, SourceUnit const& _sourceUnit);
```
cameel: The names here could be a bit more self-explanatory:
```suggestion
ASTNode const* closestASTNode(int _positionInFile, SourceUnit const& _sourceUnit);
```

libsolidity/ast/ASTUtils.cpp
```diff
@@ -17,13 +17,45 @@
...
+		{
+			m_closestMatch = &_node;
+			return true;
+		}
+		return false;
+	}
+
+	explicit ASTNodeLocator(int _pos): m_pos{_pos}, m_closestMatch{nullptr} {}
+
+	int const m_pos;
```
cameel:
```suggestion
	int const m_positionInFile;
```

libsolidity/ast/ASTUtils.h
cameel: It should mention that if there are multiple matches it just takes the one it comes across first.

EDIT: Actually the docstring is wrong. It does not take the node closest to a given position. It takes any node that overlaps with that position. If the node is close but does not cover the exact location, you won't get it.

cameel: But actually, maybe it would be better to have it look at all matches and choose the shortest one? For example I think you do not want to get the whole function definition as a result when you're only pointing at a local variable.

cameel (changes requested):

libsolidity/ast/ASTUtils.h
cameel: Actually, the docstring is wrong, it does not locate the closest node, just any node covering that location, so maybe this would be better:

```suggestion
ASTNode const* anyASTNodeAt(int _positionInFile, SourceUnit const& _sourceUnit);
```

liblangutil/CharStream.cpp
chriseth: is it "should be" or "are"? Please clarify.

libsolidity/ast/ASTUtils.cpp
```diff
@@ -17,13 +17,45 @@
...
+	{
+		if (_node.location().containsOffset(m_pos))
+		{
+			m_closestMatch = &_node;
+			return true;
+		}
+		return false;
+	}
+
+	explicit ASTNodeLocator(int _pos): m_pos{_pos}, m_closestMatch{nullptr} {}
```
chriseth: constructor should be at the top

libsolidity/ast/ASTUtils.h
chriseth: It returns the innermost ast node that covers the location.

liblangutil/CharStream.cpp
chriseth, line 155: Does this work if the file starts with `\n`?

liblangutil/CharStream.cpp
christianparpart: > Why not keep the assignment out of the if for better readability?

Ugh, "readability" is sometimes a little subjective. I can move it out though. `found` identifier wasn't the most intuitive choice. `foundIndex` would probably be most precise - not sure.

liblangutil/CharStream.cpp
christianparpart: i agree

libsolidity/interface/StandardCompiler.h
christianparpart: @cameel this is a bit tricky. I generally agree that configuration is best placed at object construction, but in here in case, I simply moved the function body of that configure() function out of the `compileSolidity` function, which is invoked by `compile()` which may or may NOT contain settings.

I've moved this configuration step out because in the LSP PR we also need it (without code duplication). We could do that in a second step (different PR only taking care of further improving the CompilerStack API) what do you think?

libsolidity/interface/StandardCompiler.cpp
christianparpart: @cameel Here I kept the same coding style as 5 lines below. I'm fine with both ways.

libsolidity/interface/StandardCompiler.cpp
christianparpart: @cameel sooo many possibilities. :D

libsolidity/ast/ASTUtils.h
christianparpart: If the position is pointing at a variable then you won't get the whole function AST node back but the `Identifier` AST node of that variable. Mind, our AST currently is not **that** precise. There is room for improvement, such as we don't store all identifiers as AST node `Identifier` but instead just as `std::string` - that is why some while ago I introduced `nameLocation()` along with `location()` to get access to the name's location. I'd like to have that changed to a real AST node at some point.

> But actually, maybe it would be better to have it look at all matches and choose the shortest one? 

I think that depends on the use-case. If one needs more fine grained control over finding AST nodes, maybe there should be a specialized finder then. Currently this API suffices the current use I think.

libsolidity/ast/ASTUtils.cpp
christianparpart: @cameel i'm not sure if names like this is a little bit over the top for such a small struct. If at all, I'd prefer an actual file offset type over var names of type int for file offsets (e.g. `cpp struct FileOffset { int value; };`. I like to box primitive types in other projects for better readability / code quality / less accidental mix-ups.

liblangutil/CharStream.cpp
christianparpart: If a `_text` starts with `\n` then `_text.find('\n', 0)` returns `0`, so `offset` gets set to `0+1=1` and it continues again in finding `'\n`, this time at `1`. So it seems to work for leading `\n`'s too.

liblangutil/CharStream.cpp
cameel: For me what makes this hard to read is putting two instructions inside the `if` condition. I often miss the condition behind the variable declaration. I think that this pattern is only readable if the condition is just checking if the variable's value is true/false. And even then I'm not a big fan of assignments inside `if`s.

liblangutil/CharStream.cpp
cameel: "should be". This is a comment in the .cpp file. I think it should be part of the actual docstring in the header.

liblangutil/CharStream.cpp
christianparpart: I don't follow. `offset` actually does reflect the offset of the next line's starting position to look for another linefeed. But I still don't know what exactly you point is? :)

libsolidity/interface/StandardCompiler.cpp
cameel: I'd change both but it was just a tiny suggestion because I saw the line was modified in the PR anyway. It's not crucial.

libsolidity/interface/StandardCompiler.cpp
cameel: Let's at least add an assert :)

BTW, I see that this code disappeared from the PR. Have you moved it to a separate one?

libsolidity/ast/ASTUtils.h
cameel: If it does then it's really not clear why. I don't see anything that would choose specifically an innermost match:

```c++
	bool visitNode(ASTNode const& _node) override
	{
		if (!_node.location().containsOffset(m_offsetInFile))
			return false;
		m_innermostMatch = &_node;
		return true;
	}
```

Is this because of the visiting order? If so, this at least needs a comment.

libsolidity/ast/ASTUtils.cpp
cameel: A dedicated type would be helpful but I think that good, informative names are still needed. It's not a big difference in a small struct but even then I don't think it's worth saving a few keystrokes. With `positionInFile` I immediately understand it when I see it and can move on quickly while with `pos` I have to stop for a moment and look for context clues. Especially given that you see this struct before you encounter the function that uses it so you do not have that context yet.

BTW, the coding style says not to use abbreviations for names (and I think it's good rule).

libsolidity/ast/ASTUtils.h
cameel: > If the position is pointing at a variable then you won't get the whole function AST node back but the Identifier AST node of that variable.

But how does it do that? Is it because of visiting order? If so, it's not obvious at all.

christianparpart: inner-most as in recursively descending the AST until `containsOffset` fails. If it starts to fail you know you have reached the inner-most AST node (in `m_innermostMatch` that still contains the input file offset. What text would you like to see as comment here?

liblangutil/CharStream.cpp
cameel: IMO the code would be simpler without the extra variable. It wasn't immediately clear to me what the relation between `found` and `offset` was and I thought that having just one variable would solve this.

Just a suggestion though. If you really prefer it the way it is I guess, it's not a big deal.

libsolidity/ast/ASTUtils.h
cameel: Looks like I missed the `!` and misinterpreted the code. I thought that you're actually taking the first one you find. Does not help that between `return false` and `return true` it's not that obvious which one makes the visitor stop (maybe we should have constants for them, e.g. `AST_CONTINUE` and `AST_STOP`?).

How about this:
``` c++
// In the AST parent location always covers the whole child location. Parent is visited first so to
// get the inntermost node we simply take the last one that still contains the offset.
if (!_node.location().containsOffset(m_offsetInFile))
	return false;
...
```

liblangutil/CharStream.h
```diff
@@ -97,9 +98,19 @@ class CharStream
 	/// Functions that help pretty-printing parse errors
 	/// Do only use in error cases, they are quite expensive.
 	std::string lineAtPosition(int _position) const;
+
+	/// @returns (1-based) line and column that matches to the byte offset @p _position.
 	std::tuple<int, int> translatePositionToLineColumn(int _position) const;
 	///@}
 
+	/// Translates a line:column to the absolute position.
```
chriseth, line 106: Is this also 1-based for both line and column?

liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,31 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
...
+	while (_line > 0)
+	{
+		auto const index = _text.find('\n', offset);
+		if (index != _text.npos)
+		{
+			offset = index + 1;
+			_line--;
+		}
+		else
+			return nullopt;
```
chriseth, line 161: This algorithm always returns nullopt if the source contains a single line without `\n` at the end. I don't think this is how it should be.

libsolidity/ast/ASTUtils.h
christianparpart: > Does not help that between return false and return true it's not that obvious which one makes the visitor stop

@cameel That's not a problem of this PR but rather how we stop the visitors in general. All visitor implementations are doing the same return-logic. And we (a few at least) agreed on `bool` not being the best choice as return type for that exact reason. I'm all in for changing that (in another PR) for our Visitor API in general.

cameel: Answered in https://github.com/ethereum/solidity/pull/12304#discussion_r755250156.

liblangutil/CharStream.h
christianparpart: This is actually documented one line below. I probably just pushed this up and you checked earlier. Sorry :)

libsolidity/ast/ASTUtils.h
cameel: Yeah, I did not mean that we should change that now, just saying that this is another factor that makes it easier to misinterpret the code, which is why it could use some comments.

liblangutil/CharStream.cpp
christianparpart: added tests including for that case.

test/liblangutil/CharStream.cpp
```diff
@@ -48,6 +48,62 @@ BOOST_AUTO_TEST_CASE(test_fail)
...
+		auto const actualPosition =                                          \
+			source->translateLineColumnToPosition((_line), (_column));       \
+		BOOST_CHECK(actualPosition.has_value());                             \
+		BOOST_CHECK_EQUAL(*actualPosition, (_expectedPosition));             \
+	}                                                                        \
+	while (0)
+
+BOOST_AUTO_TEST_CASE(translateLineColumnToPosition)
+{
+	testLineColumnToPositionTranslationFail(0, 0, "");
```
chriseth, line 74: Is this what we want to have?

christianparpart: I think you're right. we should not want that.

EDIT: Sorry, I did overlook the "Fail". Yes, we want that, because `(0, 0)` reflects the first comumn in the first line, but there is no single character in the first line, so it's actually an overflow. In the end, I think it's up to us, because at least as far as LSP is concerned, it won't hurt either way.

chriseth: I disagree. If you run `contentChanges` on an empty file, there should be a way to add text to it (apart from changing the whole file). Can you please check if the "line/colmn" thing is defined in the lsp specification somewhere?

christianparpart: ok. I understand. I agree. :)

christianparpart: quote from the spec:
> A position inside a document (see Position definition below) is expressed as a zero-based line and character offset. The offsets are based on a UTF-16 string representation.

with that example:
> So a string of the form a𐐀b the character offset of the character a is 0, the character offset of 𐐀 is 1 and the character offset of b is 3 since 𐐀 is represented using two code units in UTF-16.

I am not sure how urgent multi-codepoint graphemes are with respect to cursor positioning or if that can wait until after initial release. We would need a UTF-8 <-> UTF-16 conversion helper.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1065,12 +1065,15 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
...
 }
 
-Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSettings _inputsAndSettings)
+unique_ptr<CompilerStack> StandardCompiler::createCompilerStack(
+	ReadCallback::Callback _readFile,
+	StandardCompiler::InputsAndSettings const& _inputsAndSettings
+)
 {
-	CompilerStack compilerStack(m_readFile);
+	auto compilerStackPtr = make_unique<CompilerStack>(_readFile);
```
chriseth, line 1073: CompilerStack has a check that prevents two instances to be alive at the same time, so I don't think this function is a good idea. I know you had this as a first version, but we need a way to apply the settings to the compiler stack instead.

The StandardCompiler was not meant to be usend in a long-lived process..

chriseth: Why do we need the StandardCompiler at all for the LSP? The only two thing in the settings that are not related to bytecode generation is remappings and sources and I'm not sure if that warrants such a refactoring...

test/liblangutil/CharStream.cpp
chriseth: So it has nothing about how newlines are interpreted and which columns are valid?

chriseth:
Do we still need this?

PR #12310, opened 2021-11-23
Rename RedundantAssignEliminator to UnusedAssignEliminator.

chriseth:


libyul/optimiser/Suite.cpp
```diff
@@ -260,7 +260,7 @@ map<string, char> const& OptimiserSuite::stepNameToAbbreviationMap()
 		{LoadResolver::name,                  'L'},
 		{LoopInvariantCodeMotion::name,       'M'},
 		{ReasoningBasedSimplifier::name,      'R'},
-		{RedundantAssignEliminator::name,     'r'},
+		{UnusedAssignEliminator::name,        'r'},
```
chriseth, line 263: I would still keep this at `r`

cameel: Yeah, I think it's better to keep it. Change would require finding it when used in step sequences which is error-prone. Also, `u` is already taken and `U` will probably be taken by the other eliminator...

cameel (dismissed):
I only skimmed over it but looks like it's really only renaming + some docs so as long as it passes tests, it should be fine.

Marenz:
The chromosome seems to have changed, the r position was moved..
```
flcCUnDvejsxIOoighFTLMRmVat_r_pud != 
flcCUnDvejsxIOoighFTLMR_r_mVatpud
```

Marenz:
I pushed a small test update to fix the failing CI

PR #12312, opened 2021-11-23
Bump z3 version to 4.8.13 in our docker files

leonardoalt:
Depends on https://github.com/ethereum/solidity/pull/12317

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.emscripten
+++ b/scripts/docker/buildpack-deps/Dockerfile.emscripten
@@ -33,12 +33,12 @@
 # Using $(em-config CACHE)/sysroot/usr seems to work, though, and still has cmake find the
 # dependencies automatically.
 FROM emscripten/emsdk:2.0.33 AS base
-LABEL version="7"
+LABEL version="8"
 
 ADD emscripten.jam /usr/src
 RUN set -ex; \
 	cd /usr/src; \
-	git clone https://github.com/Z3Prover/z3.git -b z3-4.8.12 --depth 1 ; \
+	git clone https://github.com/Z3Prover/z3.git -b z3-4.8.13 --depth 1 ; \
 	cd z3; \
 	mkdir build; \
 	cd build; \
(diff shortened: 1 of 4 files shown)


cameel (dismissed):

leonardoalt:
waittt

cameel:
Oh, right. We now have the 4.8.12 check in CMakeLists.txt that needs to be updated as well.

leonardoalt:
We still need to wait for the PPA update to finish the builds before running the docker updates

leonardoalt:
This PR will only update the docker images. We'll use those hashes in another PR together with the test updates, and I'll update cmake stuff there as well.

Marenz:
Arch test failing

leonardoalt:
We need to fix the intx thing to build the final docker image that is missing here

leonardoalt:
@bshastry is this the final version of the ossfuzz build?

bshastry:
> @bshastry is this the final version of the ossfuzz build?

Yes. Turns out Hera (which caused the build failure) is redundant for fuzzing. Trimmed the dockerfile :-)

leonardoalt:
Ok cool! The PRs can be merged then.

leonardoalt:
@bshastry but do we want to keep the second last commit, regarding the clang warning flag?

leonardoalt:
Depends on https://github.com/ethereum/solidity/pull/12317

axic:
> Depends on #12317

Isn't it the reverse? #12317 depends on this, but this can be freely merged without issues.

bshastry:
> @bshastry but do we want to keep the second last commit, regarding the clang warning flag?

No, that is not necessary now. Only the latest commit (removing build steps for Hera) is relevant.

leonardoalt:
> > Depends on #12317
> 
> Isn't it the reverse? #12317 depends on this, but this can be freely merged without issues.

Well the other one depends on this one running,  just to update the docker images, but not necessarily to be merged. But yea at this point they're rather independent, unless we decide to remove the second last commit and re run the docker builds.

chriseth (approved):

PR #12317, opened 2021-11-23
Update smtchecker tests to z3 4.8.13

leonardoalt:
Let's see how much nondeterminism we get...


test/libsolidity/smtCheckerTests/array_members/push_overflow_2_safe_no_overflow_assumption.sol
```diff
@@ -13,4 +13,3 @@ contract C {
 // SMTEngine: all
 // SMTIgnoreOS: macos
 // ----
-// Info 1180: Contract invariant(s) for :C:\n(x.length >= 0)\n
```
cameel, line 16: Is that non-determinism too?

test/libsolidity/smtCheckerTests/types/address_staticcall.sol
```diff
@@ -10,14 +10,16 @@ contract C
 		assert(success);
 		assert(x == 0);
 		assert(map[0] == 0);
-		assert(localMap[0] == 0);
+		// Disabled because of Spacer's seg fault
+		//assert(localMap[0] == 0);
```
cameel, line 14: Is that on macOS again or on Linux too?

test/libsolidity/smtCheckerTests/array_members/push_overflow_2_safe_no_overflow_assumption.sol
leonardoalt: If it happens on all OSs it should be fine.

test/libsolidity/smtCheckerTests/types/address_staticcall.sol
leonardoalt: Also on Linux, the current state of this PR is what I got locally.

leonardoalt:
I think I resolved the current nondeterminism issues.
There are still 3 things to do on this PR:
- [x] Fix the z3 hash for the OSX build
- [x] Fix the intx thing so that the last docker build in https://github.com/ethereum/solidity/pull/12312 goes through, and use that final hash to update the 16.04 image used by CI here
- [x] Fix the remaining nondeterminism problems from the OSX build

test/libsolidity/smtCheckerTests/types/address_staticcall.sol
cameel: Is that maybe this? https://github.com/Z3Prover/z3/issues/4690

leonardoalt: Yea I guess

leonardoalt:
Ok I think this can be merged

chriseth (approved):

PR #12323, opened 2021-11-24
Fix file reader.

chriseth:


libsolidity/interface/FileReader.h
```diff
@@ -48,7 +48,7 @@ class FileReader
 	/// that will be used when requesting files from this file reader instance.
 	explicit FileReader(
 		boost::filesystem::path _basePath = {},
-		std::vector<boost::filesystem::path> const& _includePaths = {},
+		std::vector<boost::filesystem::path> _includePaths = {},
```
chriseth, line 51: Cannot have a `const&` to a temporary.

cameel: Wait, isn't this only true for non-const references? I thought const ref would extend the life of a temporary?

In any case I'd rather just remove the defaults if this is problematic. I don't think we really need them practice.
```suggestion
		std::vector<boost::filesystem::path> const& _includePaths,
```

chriseth: Actually I think you are right. I was worried about a place in the code where this parameter is passed from a freshly created and empty local variable.

PR #12340, opened 2021-11-29
Report unused library references

### LINKED ISSUE #12008: Reporting unused library references
## Abstract
Currently the compiler does not report an error if you supply library addresses that do not match any library referenced in the compiled sources. I think the compiler should warn user about with a message at 'info' severity.

## Motivation
The current behavior is convenient for tools, which can simply pass in the list of all deployed libraries, without worrying whether they are actually used by the contracts being compiled or not. This is the case especially for incremental compilation where only a subset of contracts is being recompiled.

On the other hand this means that supplying a bad name does not result in an error. This does not make it easy for users to determine when they are doing it wrong.

## Specification
When a list of libraries is passed to the compiler using either `--libraries` option (on the CLI) or in `settings.libraries` (in Standard JSON), the compiler should check if all of them are present in the sources. Any that do not should be reported. The message should have 'info' severity and the compilation/linking should still succeed.

In assembly and linker modes the source of the library is not available. In that case the error should be based on whether the library address was actually used to replace a link reference in the bytecode.

### Example 1
- `contract.sol`:
    ```solidity
    library L {
        function f() public {}
    }
    contract C {
        function f() public { L.f(); }
    }
    ```
When compiled with:
```bash
solc contract.sol \
    --libraries "contract.sol:X=0x1111111111111111111111111111111111111111" \
    --libraries "test.sol:L=0x2222222222222222222222222222222222222222" \
    --libraries "contract.sol:L=0x3333333333333333333333333333333333333333"
```
this should result in the following messages:
```
Unused link reference: 'contract.sol:X'. Library not found.
Unused link reference: 'test.sol:L'. Library not found.
```
The compiler should continue despite the messages.

### Example 2
- `contract.sol`:
    ```solidity
    contract C {}
    ```
this should result in the following messages:
```
Unused link reference: 'contract.sol:L'. Library not found.
Unused link reference: 'contract.sol:X'. Library not found.
Unused link reference: 'test.sol:L'. Library not found.
```

### Example 3
This example should report the same messages as example 1. `contract.sol:L` should **not** be reported as unused despite not being referenced by `C` because the library does exist:
- `contract.sol`:
    ```solidity
    library L {}
    contract C {}
    ```

### Example 4
- `contract.yul`:
    ```yul
    {
        mstore(0x100, linkersymbol("contract.sol:L"))
    }
    ```
When compiled with:
```bash
solc --strict-assembly contract.yul \
    --libraries "contract.sol:X=0x1111111111111111111111111111111111111111" \
    --libraries "test.sol:L=0x2222222222222222222222222222222222222222" \
    --libraries "contract.sol:L=0x3333333333333333333333333333333333333333"
```
this should result in the following messages:
```
Unused link reference: 'contract.sol:X'. Library not found.
Unused link reference: 'test.sol:L'. Library not found.
```

### Example 5
- `contract.bin`:
    ```
    608060405234801561001057600080fd5b5060c78061001f6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c806326121ff014602d575b600080fd5b60336035565b005b73__$a83509b47300fb2872e107211bec0983bd$__6326121ff06040518163ffffffff1660e01b815260040160006040518083038186803b158015607857600080fd5b505af4158015608b573d6000803e3d6000fd5b5050505056fea2646970667358221220c2f72d183b0908ff6e63a38057ad4b11b727aad99e02b5e656f123d4fa91e36864736f6c63430008070033

    // $a83509b47300fb2872e107211bec0983bd$ -> contract.sol:L
    ```

When linked with:
```bash
solc --link contract.bin \
    --libraries "contract.sol:X=0x1111111111111111111111111111111111111111" \
    --libraries "test.sol:L=0x2222222222222222222222222222222222222222" \
    --libraries "contract.sol:L=0x3333333333333333333333333333333333333333"
```
this should result in the following messages:
```
Unused link reference: 'contract.sol:X'. Library not found.
Unused link reference: 'test.sol:L'. Library not found.
```

## Backwards Compatibility
The change is backwards-compatible because reporting a message at an info level does not interrupt the compilation.

mejsiej:
Closes #12008


chriseth:
Thank you for your pull request! I think the architecture could be a bit improved here:

1. LinkerObject is too low-level and should not print anything on stdin / stdout. I think it would be better to change the link function to return the set of used libraries, for examples. Alternatively, you could add a new function on LinkerObject that returns the set of referenced libraries and is called before link() is called. This way, you could also detect references to unknown libraries

2. If you have two different contracts and compile them at the same time, they might reference different sets of libraries. The compiler should only report if a library was supplied but not used by any of the contracts. This means that this mechanism has to be put inside the commandline interface.

mejsiej:
Hey @chriseth, can you take a look at it? I moved the reporting to the CommandLineInterface as you requested.

libevmasm/LinkerObject.h
```diff
@@ -61,7 +61,7 @@ struct LinkerObject
 	void append(LinkerObject const& _other);
 
 	/// Links the given libraries by replacing their uses in the code and removes them from the references.
-	void link(std::map<std::string, util::h160> const& _libraryAddresses);
+	std::list<std::string> link(std::map<std::string, util::h160> const& _libraryAddresses);
```
chriseth: Wouldn't a `set` be better here?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -696,18 +696,32 @@ bool CompilerStack::compile(State _stopAfter)
...
+	std::list<std::string> libNames;
+
 	for (auto& contract: m_contracts)
 	{
-		contract.second.object.link(m_libraries);
+		for (auto& library: m_libraries)
+			for (auto& con: m_contracts)
+				if (library.first == con.first.substr(con.first.find_last_of('/') + 1))
+				{
+					contract.second.object.bytecode.push_back(static_cast<uint8_t>(solidity::evmasm::Instruction::PUSH20));
```
chriseth: What are you doing here?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -630,7 +630,7 @@ bool CompilerStack::isRequestedContract(ContractDefinition const& _contract) con
 	return false;
 }
 
-bool CompilerStack::compile(State _stopAfter)
+bool CompilerStack::compile(State _stopAfter, bool performLinking)
```
chriseth, line 633: I think you should either never link or use the `_stopAfter` parameter and add a linking stage.

libsolidity/interface/CompilerStack.cpp
mejsiej: I am checking if there are any matching libraries in m_contracts and if there are any, I am adding them to the object's linkReferences. In compiler mode, before that change linkReferences where always empty and it did not even enter the linking process in LinkerObject::link().

libsolidity/interface/CompilerStack.cpp
mejsiej: performLinking is set to true in CompilerStack.h. 
It's only there so I can avoid linking in only one specific case, which is call to that function in compiler mode, 
and do it later in CommandLineinterface to get the used libraries.

libevmasm/LinkerObject.h
mejsiej: I will try it out.

libsolidity/interface/CompilerStack.cpp
chriseth: And why are you doing this? Note that you are adding a push instruction to the end of the bytecode after the assembly stage - I don't think this will do any good.

mejsiej: I added the push instruction because I saw it was used in LinkerObject::assemble() when setting the linkReferences, and thought it was necessary. I will delete it.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -696,18 +696,31 @@ bool CompilerStack::compile(State _stopAfter)
...
+	std::set<std::string> libNames;
+
 	for (auto& contract: m_contracts)
 	{
-		contract.second.object.link(m_libraries);
+		for (auto& library: m_libraries)
+			for (auto& con: m_contracts)
+				if (library.first == con.first.substr(con.first.find_last_of('/') + 1))
+				{
+					contract.second.object.linkReferences[contract.second.object.bytecode.size()] = library.first;
```
chriseth: Even just with this line - why would you modify every contract's link references just because of the fact that we compiled it together with an unrelated library?

mejsiej: Oh yeah, you are right, it's useless. I will remove that foreach loop.

solc/CommandLineInterface.cpp
```diff
@@ -728,7 +728,14 @@ bool CommandLineInterface::compile()
 			m_compiler->setParserErrorRecovery(m_options.input.errorRecovery);
 		}
 
-		bool successful = m_compiler->compile(m_options.output.stopAfter);
+		bool successful = m_compiler->compile(m_options.output.stopAfter, false);
+		if (successful)
+		{
+			std::set<std::string> usedLibs = m_compiler->link();
+			for (auto lib: m_options.linker.libraries)
```
nemerle: You should use `const auto&` here since a plain `auto` will copy the underlying string value.

mejsiej: Got it!

test/tools/ossfuzz/solProto.proto
```diff
@@ -1,18 +1,18 @@
 /*
-	This file is part of solidity.
+    This file is part of solidity.
 
-	solidity is free software: you can redistribute it and/or modify
```
chriseth, line 4: Please revert these changes (rebasing might help)

mejsiej: Alright!

leonardoalt:
What's the current state here? Tests are failing.

cameel:
Seems to need another round of review and a rebase.

cameel:
Closing this for now. Like I said in https://github.com/ethereum/solidity/pull/12341#issuecomment-1176524665, feel free to comment if you're actually still interested in getting reviews here or reopen if you want to continue working on it.

PR #12341, opened 2021-11-29
Error when a library reference is something other than a library

### LINKED ISSUE #12009: Error when a library reference is something other than a library
## Abstract
The compiler does not complain if you use a contract or an interface in places where a fully qualified library name is expected. Such a reference can never work and should be reported as an error.

### Example
```bash
echo 'contract C {}' | solc - --libraries "<stdin>:C=0x1111111111111111111111111111111111111111"
```
This does not produce an error even though `<stdin>:C` refers to a contract and not a library.

## Specification
Each fully qualified library name used in `--libraries`/`settings.libraries` that actually exists and refers to something else than a library should be reported as an error. For example in the above:
```
Invalid link reference: '<stdin>:C'. Not a library.
```

If possible, this should work that way not only for contracts but also for other names matching the reference. E.g. structs, enums, errors, events.

## Backwards Compatibility
The code that previously compiled fine with such references will no longer compile so it's a breaking change.

mejsiej:
Closes #12009

Based on #12340


cameel:
Closing this for now. Given the silence on other PRs I suspect you're no longer interested in working on them (which is understandable - sorry for the lackluster reviews but unfortunately our capacity to review extra PRs was and still is a bit limited). 

You're welcome to reopen if you decide to resume working on it. Or, in this case, post a comment if you're still interested in getting a review - this one was actually not reviewed at all so far.

PR #12345, opened 2021-11-29
Use LineColumn type.

chriseth:
This is a refactor that documents the LineColumn type and uses it in one additional function.

### DIFF
--- a/liblangutil/SourceLocation.h
+++ b/liblangutil/SourceLocation.h
@@ -119,4 +119,23 @@ SourceLocation parseSourceLocation(
 /// Stream output for Location (used e.g. in boost exceptions).
 std::ostream& operator<<(std::ostream& _out, SourceLocation const& _location);
 
+
+/**
+ * Alternative, line-column-based representation for source locations.
+ * Both line and column are zero-based.
+ * If used as a range, the second location is considered exclusive.
+ * Negative values are invalid.
+ */
+struct LineColumn
+{
+	/// Line value, can be between zero and number of `\n` characters in the source file.
+	int line = -1;
+	/// Column value, can be between zero and number of characters in the line (inclusive).
+	int column = -1;
+
+	LineColumn() = default;
+	explicit LineColumn(int _line, int _column): line(_line), column(_column) {}
+};
+
+
 }
--- a/liblangutil/CharStream.cpp
+++ b/liblangutil/CharStream.cpp
@@ -100,7 +100,7 @@ string CharStream::lineAtPosition(int _position) const
 	return line;
 }
 
-tuple<int, int> CharStream::translatePositionToLineColumn(int _position) const
+LineColumn CharStream::translatePositionToLineColumn(int _position) const
 {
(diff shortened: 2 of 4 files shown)


liblangutil/SourceLocation.h
```diff
@@ -119,4 +119,24 @@ SourceLocation parseSourceLocation(
...
+ */
+struct LineColumn
+{
+	/// Line value, can be between zero and number of `\n` characters in the source file.
+	int line = {-1};
+	/// Column value, can be between zero and number of characters in the line (inclusive).
+	int column = {-1};
+
+	LineColumn() = default;
+	explicit LineColumn(std::tuple<int, int> const& _t): line{std::get<0>(_t)}, column{std::get<1>(_t)} {}
```
chriseth: I think this is actually not needed anymore.

liblangutil/SourceLocation.h
```diff
@@ -119,4 +119,23 @@ SourceLocation parseSourceLocation(
...
+ */
+struct LineColumn
+{
+	/// Line value, can be between zero and number of `\n` characters in the source file.
+	int line = {-1};
+	/// Column value, can be between zero and number of characters in the line (inclusive).
+	int column = {-1};
+
+	LineColumn() = default;
+	explicit LineColumn(int _line, int _column): line(_line), column(_column) {}
```
cameel, line 137: Does it even need a constructor now that arguments match struct fields and are no longer a tuple?

liblangutil/SourceLocation.h
```diff
@@ -119,4 +119,23 @@ SourceLocation parseSourceLocation(
...
+ * Both line and column are zero-based.
+ * If used as a range, the second location is considered exclusive.
+ * Negative values are invalid.
+ */
+struct LineColumn
+{
+	/// Line value, can be between zero and number of `\n` characters in the source file.
+	int line = {-1};
+	/// Column value, can be between zero and number of characters in the line (inclusive).
+	int column = {-1};
```
cameel: It's just an `int`. Why use an initializer list?
```suggestion
	/// Line value, can be between zero and number of `\n` characters in the source file.
	int line = -1;
	/// Column value, can be between zero and number of characters in the line (inclusive).
	int column = -1;
```

cameel (dismissed):
I have two minor remarks but looks good anyway.

liblangutil/SourceLocation.h
cameel: Yeah, I think it's not needed.

cameel:
By the way, nice PR number :)

chriseth:
Haha, I was actually just wondering today if this number has already turned up :)

liblangutil/SourceLocation.h
chriseth: Yes, but can we have default values and prevent construction from just a single value?

cameel: True, that's a good argument for keeping the constructor.

cameel (approved):

PR #12352, opened 2021-12-01
LineColumn handling for CharStream.

chriseth:


test/liblangutil/CharStream.cpp
```diff
@@ -28,6 +28,25 @@
 
 #include <boost/test/unit_test.hpp>
 
+using namespace std;
+
+namespace std
+{
+template<typename T>
+std::ostream& operator<<(std::ostream& _ostream, std::optional<T> const& _value)
```
christianparpart: I think you've done this because Boost.Test makes use of it for failure outputs?

christianparpart (dismissed):

test/liblangutil/CharStream.cpp
cameel: I think that adding stuff to `std` is undefined behavior though, which is why I've usually been doing it this way instead:
https://github.com/ethereum/solidity/blob/a7d89a92434cea438f3b212fe6360a8d2a304011/test/solc/CommandLineInterface.cpp#L50-L108

cameel: Also, it would be nice to have this definition in `test/Common.h` or some file it includes so that you don't have to define it every time locally.

test/liblangutil/CharStream.cpp
```diff
@@ -28,6 +28,25 @@
...
+		_ostream << *_value;
+	else
+		_ostream << "[nullopt]";
+	return _ostream;
+}
+inline std::ostream& operator<<(std::ostream& _ostream, std::nullopt_t const&)
+{
+	return _ostream << "[nullopt]";
+}
+}
```
cameel: Given that we do not indent namespaces, I think it's a nice convention to comment where they end:

```suggestion

} //namespace std
```

Also, some whitespace after opening and before ending brace would be nice for readability.

liblangutil/CharStream.cpp
```diff
@@ -144,3 +144,32 @@ string CharStream::singleLineSnippet(string const& _sourceCode, SourceLocation c
...
+	if (_input.line < 0)
+		return nullopt;
+
+	size_t offset = 0;
+	for (int i = 0; i < _input.line; i++)
+	{
+		offset = _text.find('\n', offset);
+		if (offset == _text.npos)
+			return nullopt;
+		offset++; // Skip linefeed.
```
cameel, line 164: Can we actually assume that the file does not have Windows CFLR line endings?

cameel (commented):
Looks fine in general but I just realized that we completely ignore CRLF. Can we actually do that?

liblangutil/CharStream.cpp
chriseth: For `\r\n` we are fine, because column numbers higher than the number of characters are truncated to the max value at least for LSP.

cameel (approved):

test/liblangutil/CharStream.cpp
christianparpart: anonymous namespace is not visible across compilation units though. I think `ostream& operator<<(ostream&, T)` should simply be in the same namespace as `T` to be found by ADL (argument dependent lookup), isn't it?

PR #12353, opened 2021-12-01
Accessors for exceptions.

chriseth:


test/libsolidity/SyntaxTest.cpp
```diff
@@ -120,7 +120,7 @@ void SyntaxTest::filterObtainedErrors()
 	{
 		int locationStart = -1, locationEnd = -1;
 		string sourceName;
-		if (auto location = boost::get_error_info<errinfo_sourceLocation>(*currentError))
+		if (auto location = currentError->sourceLocation())
```
christianparpart: In other places (above) the explicit type is used. I don't mind either, But I think we tend towards explicit typing, do we?

christianparpart: Ah, and it's returning a pointer, I think it should be `auto*` then, right?

cameel (approved):

PR #12354, opened 2021-12-01
Function to read a number of bytes from an input stream.

chriseth:


libsolutil/CommonIO.cpp
```diff
@@ -79,6 +79,16 @@ string solidity::util::readUntilEnd(istream& _stdin)
 	return ss.str();
 }
 
+string solidity::util::readBytes(istream& _input, size_t _length)
+{
+	string ret;
+	ret.resize(_length);
+	_input.read(ret.data(), static_cast<streamsize>(_length));
+	if (!_input)
```
cameel: There are a few different failure conditions in iostreams and I always have to look them up to make sure I get this right. I'd add a comment explaining what's happening. Initially I thought it was wrong because you do not check `eofbit` but looks like `read()` sets `failbit` too on EOF so you do not have to.

I'd also check `fail()` explicitly so that it's not necessary to remember if `operator bool()` translates to `good()` or `!fail()` (they're not opposites of each other).

```suggestion
	// read() sets failbit in addition to eofbit if it reads fewer characters due to EOF
	if (_input.fail())
```

libsolutil/CommonIO.cpp
```diff
@@ -79,6 +79,16 @@ string solidity::util::readUntilEnd(istream& _stdin)
 	return ss.str();
 }
 
+string solidity::util::readBytes(istream& _input, size_t _length)
+{
+	string ret;
```
cameel: Why not `output`?
```suggestion
	string output;
```

libsolutil/CommonIO.cpp
```diff
@@ -79,6 +79,16 @@ string solidity::util::readUntilEnd(istream& _stdin)
...
+	ret.resize(_length);
+	_input.read(ret.data(), static_cast<streamsize>(_length));
+	if (!_input)
+		ret.resize(static_cast<size_t>(_input.gcount()));
```
cameel: Can you add a test case in `solidity/test/libsolutil/CommonIO.cpp` to make sure this actually works correctly if the input is shorter than `_length`?

cameel (commented):
Looks like it works but I have some suggestions regarding tests and readability.

christianparpart (dismissed):

libsolutil/CommonIO.cpp
```diff
@@ -79,6 +79,17 @@ string solidity::util::readUntilEnd(istream& _stdin)
 	return ss.str();
 }
 
+string solidity::util::readBytes(istream& _input, size_t _length)
+{
+	string output;
+	output.resize(_length);
+	_input.read(output.data(), static_cast<streamsize>(_length));
+	// If read() reads fewer bytes it sets falibit.
```
cameel: Typo. Also I think EOF should be mentioned in some way to make it understandable.
```suggestion
	// If read() reads fewer bytes it sets failbit in addition to eofbit.
```

cameel (dismissed):
Looks fine overall so I'm already approving but the comment has a typo :)

cameel (approved):

PR #12358, opened 2021-12-01
Fix ICE when unsafe targets are solved more than once and the cex is !=

leonardoalt:
The fix here is to actually not reset which targets were proven unsafe/safe. This is fine because the CHC and ModelChecker objects are never reset (from CompilerStack). This potentially also saves redundant queries, like in the added example. This test in isoltest didn't fail before this PR, but it did ICE when running the CLI and standard JSON. I ran it locally and it doesn't happen anymore.

### DIFF
--- a/test/libsolidity/smtCheckerTests/imports/ExtCall.sol
+++ b/test/libsolidity/smtCheckerTests/imports/ExtCall.sol
@@ -1,0 +1,42 @@
+==== Source: ExtCall.sol ====
+interface Unknown {
+    function callme() external;
+}
+
+contract ExtCall {
+    uint x;
+
+    bool lock;
+    modifier mutex {
+        require(!lock);
+        lock = true;
+        _;
+        lock = false;
+    }
+
+    function setX(uint y) mutex public {
+        x = y;
+    }
+
+    function xMut(Unknown u) public {
+        uint x_prev = x;
+        u.callme();
+        assert(x_prev == x);
+    }
+}
+==== Source: ExtCall.t.sol ====
+import "ExtCall.sol";
+
+contract ExtCallTest {
+    ExtCall call;
+
+    function setUp() public {
+        call = new ExtCall();
+    }
+}
+// ====
+// SMTEngine: all
+// SMTIgnoreCex: yes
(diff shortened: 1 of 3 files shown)


test/libsolidity/smtCheckerTests/imports/ExtCall.sol
```diff
@@ -0,0 +1,48 @@
+==== Source: ExtCall.sol ====
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.6;
```
cameel: Superfluous pragma/comment. soltest will insert that automatically.

test/libsolidity/smtCheckerTests/imports/ExtCall.sol
```diff
@@ -0,0 +1,48 @@
...
+    function setUp() public {
+        call = new ExtCall();
+    }
+}
+// ====
+// SMTEngine: all
+// SMTIgnoreCex: yes
+// ----
+// Warning 6328: (ExtCall.sol:360-379): CHC: Assertion violation happens here.
+// Warning 4588: (ExtCall.t.sol:180-193): Assertion checker does not yet implement this type of function call.
```
cameel: This does not produce an ICE for me. Checked on both `develop` and 0.8.10 release binary. I remember a similar situation in https://github.com/ethereum/solidity/pull/11852#discussion_r697551335 but there I forgot to enable the SMTChecker. Here I tried with:
```bash
solc ExtCall.t.sol --model-checker-engine all
```
```
Warning: CHC: Assertion violation happens here.
Counterexample:
x = 1, lock = false
u = 0
x_prev = 0

Transaction trace:
ExtCall.constructor()
State: x = 0, lock = false
ExtCall.xMut(0)
    u.callme() -- untrusted external call, synthesized as:
        ExtCall.setX(1) -- reentrant call
  --> /tmp/ExtCall.t.sol:26:3:
   |
26 |            assert(x_prev == x);
   |            ^^^^^^^^^^^^^^^^^^^
```

cameel (approved):

PR #12361, opened 2021-12-02
Changed the message of the error thrown when the structure has all it's members omitted

### LINKED ISSUE #12302: Confusing error message when a getter would return zero values and the compiler refuses to generate it
Original title: "unusual behavior with mapping within struct within library".

## Description

When a library contains a struct type whose _only_ member is a `mapping`, and a contract contains a _public_ member of that struct type, then compilation fails with error `Internal or recursive type is not allowed for public state variables.`. But if you add a _further_ field in the struct (of type `uint256` for example), then compilation **does not fail**. This looks like a bug.

## Environment

- Compiler version: 0.8.10
- Target EVM version (as per compiler settings): compiler default (London)
- Operating system: MacOS

## Steps to Reproduce
`Lib.sol`
```solidity
pragma solidity 0.8.10;
library Lib {
    struct MyStruct {
        // uint256 myInt;
        mapping(uint256 => uint256) myMap;
    }
}
```
`A.sol`
```solidity
pragma solidity 0.8.10;
import './Lib.sol';
contract A {
    Lib.MyStruct public myStruct;
}
```
```
TypeError: Internal or recursive type is not allowed for public state variables.
 --> contracts/A.sol:6:5:
  |
6 |     Lib.MyStruct public myStruct;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
BUT IF YOU UNCOMMENT the commented line, then it compiles.

mejsiej:
Closes #12302

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -592,7 +592,14 @@ bool TypeChecker::visit(VariableDeclaration const& _variable)
 				);
 		}
 		if (!getter.interfaceFunctionType())
-			m_errorReporter.typeError(6744_error, _variable.location(), "Internal or recursive type is not allowed for public state variables.");
+		{
+			solAssert(getter.returnParameterNames().size() == getter.returnParameterTypes().size());
+			solAssert(getter.parameterNames().size() == getter.parameterTypes().size());
+			if (getter.returnParameterTypes().empty() && getter.parameterTypes().empty())
+				m_errorReporter.typeError(5359_error, _variable.location(), "The struct has all its members omitted, therefore the getter cannot return any values.");
+			else
+				m_errorReporter.typeError(6744_error, _variable.location(), "Internal or recursive type is not allowed for public state variables.");
+		}
 	}
 
 	bool isStructMemberDeclaration = dynamic_cast<StructDefinition const*>(_variable.scope()) != nullptr;
--- a/test/libsolidity/syntaxTests/nameAndTypeResolution/070_struct_accessor_one_array_only.sol
+++ b/test/libsolidity/syntaxTests/nameAndTypeResolution/070_struct_accessor_one_array_only.sol
@@ -3,4 +3,4 @@ contract test {
     Data public data;
 }
 // ----
-// TypeError 6744: (58-74): Internal or recursive type is not allowed for public state variables.
+// TypeError 5359: (58-74): The struct has all its members omitted, therefore the getter cannot return any values.


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -592,7 +592,12 @@ bool TypeChecker::visit(VariableDeclaration const& _variable)
 				);
 		}
 		if (!getter.interfaceFunctionType())
-			m_errorReporter.typeError(6744_error, _variable.location(), "Internal or recursive type is not allowed for public state variables.");
+		{
+			if (getter.returnParameterNames().empty() && getter.returnParameterTypes().empty() && getter.parameterNames().empty() && getter.parameterTypes().empty())
+				m_errorReporter.typeError(5359_error, _variable.location(), "The structure has all it's members omitted, therefore the getter cannot return any values.");
```
cameel: I think you can assume that there are as many names as there are types:
```suggestion
			solAssert(getter.returnParameterNames().size() == getter.returnParameterTypes().size());
			solAssert(getter.parameterNames().size() == getter.parameterTypes().size());
			if (getter.returnParameterTypes().empty() && getter.parameterTypes().empty())
				m_errorReporter.typeError(5359_error, _variable.location(), "The structure has all it's members omitted, therefore the getter cannot return any values.");
```
Or might be `<=` - not sure if in case like `function f(uint x, uint)` you'll get less names or just `""` as names. Please see if tests will pass with one of these.

```suggestion
				m_errorReporter.typeError(5359_error, _variable.location(), "The struct has all its members omitted, therefore the getter cannot return any values.");
```

cameel: Adding this message changes the expected output in some syntax tests. Please run `isoltest` with the `--update` option locally and commit the changes.

Also note that the `chk_errorcodes` is failing because it ensures that we have at least one case covering each error. There's no test for the newly added error 5359. Running `isoltest` will probably change that but it seems we only have one test case for such getters and I think we need more to test this properly. Could you add some more cases, covering various situations that can result in such a getter?
- Mapping in the struct.
- Dynamic arrays other than `bytes`/`string` in the struct.
- Multiple fields being omitted (rather than just one).
- Struct containing a struct that has all its fields omitted (see https://github.com/ethereum/solidity/issues/11826#issuecomment-902835510).
    ```solidity
    struct T { mapping(uint => uint) m; }
    struct S { T t; }
    ```
    or
    ```solidity
    struct U { mapping(uint => uint) m; }
    struct T { U u; }
    struct S { T t; }
    ```
- A struct that contains two fields: one that won't be omitted (e.g. `uint`) and another that is also a struct and would have all of its members omitted.
    ```solidity
    struct U { mapping(uint => uint) m; }
    struct T { U u; }
    struct S { uint i; T t; }
    ```
- Nested dynamic array. Surprisingly, this does not fail. Apparently the rule for dynamic arrays is not recursive.
    ```solidity
    struct T { uint[] u; }
    struct S { T t; }
    ```

cameel (dismissed):

libsolidity/analysis/TypeChecker.cpp
mejsiej: Yeah, I can try.

chriseth (approved):
Looks good now, thanks a lot!

PR #12364, opened 2021-12-03
Removed the initiation of default outputs in assembler mode

### LINKED ISSUE #12230: [CLI] Do not select any outputs by default in assembler mode
#12074 added support for CLI output options (like `--bin` or `--ewasm`) in assembler mode. If no outputs are requested though, the compiler chooses the default ones to keep the behavior backwards-compatible. We should change this so that nothing is printed by default (this is how the compiler mode behaves) but it needs to be done in a breaking release.

mejsiej:
Closes #12230

### DIFF
--- a/solc/CommandLineParser.cpp
+++ b/solc/CommandLineParser.cpp
@@ -481,17 +481,6 @@ bool CommandLineParser::parseOutputSelection()
 	for (auto&& [optionName, outputComponent]: CompilerOutputs::componentMap())
 		m_options.compiler.outputs.*outputComponent = (m_args.count(optionName) > 0);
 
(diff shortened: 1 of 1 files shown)


cameel (changes requested):
Thanks for the PR! There are a few issues here:
- This is a breaking change (this is why it was not done in the original PR even though it's relatively simple). As such it must be based on the `breaking` branch rather than on `develop`.
- It's a user-facing change so it needs a changelog entry.
- Command-line tests are failing and need to be updated. That's because some of them actually depend on this default. You need to modify their `args` files to actually request inputs they need. See the failing CI jobs.

mejsiej:
Alright! I will resolve them asap.

PR #12366, opened 2021-12-05
Fix several three consecutive backquotes in document

minaminao:
I found several ``` and fixed them to ``.

### DIFF
--- a/docs/cheatsheet.rst
+++ b/docs/cheatsheet.rst
@@ -81,7 +81,7 @@ Global Variables
 - ``abi.encodeWithSelector(bytes4 selector, ...) returns (bytes memory)``: :ref:`ABI <ABI>`-encodes
   the given arguments starting from the second and prepends the given four-byte selector
 - ``abi.encodeWithSignature(string memory signature, ...) returns (bytes memory)``: Equivalent
-  to ``abi.encodeWithSelector(bytes4(keccak256(bytes(signature)), ...)```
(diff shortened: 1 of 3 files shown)


chriseth (approved):
Nice, thanks!

cameel (approved):

cameel:
Looks like CI checks don't run on this PR.

[Why aren't pull requests triggering jobs on my organization?](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-)
> Sometimes you'll have a user who submits a pull request to your repository from a fork, but no pipeline will be triggered with the pull request. This can happen when the user is following the project fork on their personal account rather than the project itself on CircleCI.
>
> This will cause the jobs to trigger under the user's personal account. If the user is following a fork of the repository on CircleCI, we will only build on that fork and not the parent, so the parent’s PR will not get status updates. 
>
> In these cases have the user unfollow their fork of the project on CircleCI and follow the source project instead. This will trigger their jobs to run under the organization when they submit pull requests.

So we need to either have @minaminao [log into CircleCi](https://circleci.com/vcs-authorize/) and follow the project or just force merge the PR.

minaminao:
I looked at the CircleCI dashboard and it seems that I can't follow `ethereum/solidity` because I don't belong to the `ethereum` organization.

PR #12370, opened 2021-12-06
Removed the initiation of default outputs in assembler mode

### LINKED ISSUE #12230: [CLI] Do not select any outputs by default in assembler mode

mejsiej:
~Depends on #12376.~ Merged.
Closes #12230


Changelog.md
```diff
@@ -5,6 +5,7 @@ Breaking changes:
  * Disallow ``delete`` on types that contain nested mappings.
  * Inline Assembly: Consider functions, function parameters and return variables for shadowing checks.
  * Commandline Interface: Remapping targets are not automatically added to allowed paths.
+ * Commandline Interface: Removed the selection of outputs by default in assembler mode.
```
cameel: How about this?
```suggestion
 * Commandline Interface: Assembler mode no longer enables all outputs by default.
```

test/cmdlineTests/linking_strict_assembly/args
```diff
@@ -1 +1 @@
---strict-assembly --libraries contract/test.sol:L=0x1234567890123456789012345678901234567890
+--strict-assembly --bin --asm --ewasm --ewasm-ir --ir-optimized --libraries contract/test.sol:L=0x1234567890123456789012345678901234567890
```
cameel: In case of all the linker tests we can reduce this to `--bin`. All other outputs are not affected.
```suggestion
--strict-assembly --bin --libraries contract/test.sol:L=0x1234567890123456789012345678901234567890
```

test/cmdlineTests/strict_asm_optimizer_steps/args
```diff
@@ -1 +1 @@
---strict-assembly --optimize --yul-optimizations dhfoDgvulfnTUtnIf
+--strict-assembly --optimize --bin --asm --ewasm --ewasm-ir --ir-optimized --yul-optimizations dhfoDgvulfnTUtnIf
```
cameel: Here only the IR output is relevant.
```suggestion
--strict-assembly --optimize --ir-optimized --yul-optimizations dhfoDgvulfnTUtnIf
```

test/cmdlineTests/yul_function_name_clashes/args
```diff
@@ -1 +1 @@
---strict-assembly --debug-info none
\ No newline at end of file
+--strict-assembly --debug-info none --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel: The "name clashes" tests do not produce any ewasm code so ewasm output is superfluous..
```suggestion
--strict-assembly --debug-info none --bin --asm --ir-optimized
```

test/cmdlineTests/wasm_to_wasm_function_returning_multiple_values/args
```diff
@@ -1 +1 @@
---yul --yul-dialect ewasm --machine ewasm
+--yul --yul-dialect ewasm --machine ewasm --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel: Currently you can only get output from either `--asm` or `--ewasm`, never both. `--asm` is generally not needed in any tests that use `ewasm` dialect and I'd leave only in cases where we actually want to verify that (e.g. in "debug info" tests). 
```suggestion
--yul --yul-dialect ewasm --machine ewasm --bin --ewasm --ewasm-ir --ir-optimized
```

test/cmdlineTests/yul_optimize_runs/args
```diff
@@ -1 +1 @@
---yul --yul-dialect evm --optimize --optimize-runs 10000
+--yul --yul-dialect evm --optimize --bin --asm --ewasm --ewasm-ir --ir-optimized --optimize-runs 10000
```
cameel:
```suggestion
--yul --yul-dialect evm --optimize --bin --asm --ir-optimized --optimize-runs 10000
```

test/cmdlineTests/yul_verbatim/args
```diff
@@ -1 +1 @@
---strict-assembly
+--strict-assembly --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel:
```suggestion
--strict-assembly --bin --asm --ir-optimized
```

test/cmdlineTests/yul_verbatim_msize/args
```diff
@@ -1 +1 @@
---strict-assembly --optimize
\ No newline at end of file
+--strict-assembly --bin --asm --ewasm --ewasm-ir --ir-optimized --optimize
```
cameel:
```suggestion
--strict-assembly --bin --asm --ir-optimized --optimize
```

cameel (changes requested):
1) The changes look good overall. I'd adjust the outputs in some of the tests though. Not all are needed.
2) Looks like we need to update the `breaking` branch before this can be merged. It still uses Z3 0.8.12 so `b_archlinux` will not pass.
3) It's possible to change the base branch without creating a new PR. You could have just rebased the branch from #12364 on `breaking` and then in the PR clicked "Edit" next to the title - this would show a dropdown where you can select a different base branch.

test/cmdlineTests/linking_strict_assembly/args
mejsiej: Alright, but this also means I have to change the expected output in the tests.

mejsiej:
I have implemented the changes you have requested @cameel.

test/cmdlineTests/yul_function_name_clashes_different_params/args
```diff
@@ -1 +1 @@
---strict-assembly
+--strict-assembly --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel:
```suggestion
--strict-assembly --bin --asm --ir-optimized
```

test/cmdlineTests/wasm_to_wasm_memory_instructions_alignment/args
```diff
@@ -1 +1 @@
---yul --yul-dialect ewasm --machine ewasm
+--yul --yul-dialect ewasm --machine ewasm --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel:
```suggestion
--yul --yul-dialect ewasm --machine ewasm --bin --ewasm --ewasm-ir --ir-optimized
```

test/cmdlineTests/object_compiler/args
```diff
@@ -1 +1 @@
---strict-assembly --optimize
+--strict-assembly --optimize --bin --asm --ewasm --ewasm-ir --ir-optimized
```
cameel:
```suggestion
--strict-assembly --optimize --bin --asm --ir-optimized
```

cameel (approved):
There are 3 more places where args can be removed. Other than that it looks good to merge so I'm approving already.

> Alright, but this also means I have to change the expected output in the tests.

Yeah, that's the whole point. The whole output is very vebose and the new output selection options allow us to limit it to what is relevant in each test.

Note that this is blocked by #12376 though. When @chriseth merges it, you'll need to rebase on `breaking` again so that all CI jobs can pass.

mejsiej:
Alright! Will do it as soon as it gets merged.

cameel:
Oh, one more thing: could you squash your commits? Our convention is to have each commit represent one complete change, without subsequent commits adding stuff and then removing it. Any review fixes should be squashed into the commits they modify before the PR is merged.

mejsiej:
Of course, I will do it when I get home.

cameel:
I have just reviewed #12378 and it reminded me that we need one more thing here: print a message when the output is empty. We currently only have that for compiler mode. Fortunately the situation in assembler mode is simpler than for compilation because we can only have one input file and the file always represents one contract. So we only need a message saying that assembling was successful but no outputs were selected.

mejsiej:
I added the printing of this message.

scripts/common_cmdline.sh
```diff
@@ -70,6 +70,7 @@ function compileFull
     local errors; errors=$(grep -v -E \
         -e 'Warning: This is a pre-release compiler version|Warning: Experimental features are turned on|pragma experimental ABIEncoderV2|^ +--> |^ +\||^[0-9]+ +\| ' \
         -e 'Warning: Yul is still experimental. Please use the output with care.' \
+        -e 'Assembler run successful, no output requested.' \
```
cameel:
```suggestion
        -e '^Assembler run successful, no output requested.$' \
```

cameel (commented):
Thanks! Just one small correction to make the regex less susceptible to false-positives.

scripts/common_cmdline.sh
cameel: Actually, I'd add that to all of those excluded lines.

mejsiej: Got it, added that to every line there.

cameel:
#12376 has been merged today so tests should pass if you rebase on `breaking` now.

mejsiej:
Alright!

mejsiej:
Hey @cameel, how should I do it? When I try to rebase it shows that it will also add 122 commits there and I can't find a way to squash them.

cameel:
That does not sound right. When I try it locally it only tries to rebase the one commit you have here. Is your local `breaking` up to date? And are you sure you have the same `dont-select-any-outputs` you have here checked out and not some older version that was still on `develop`?

Your `breaking` should be at commit 0bbf58ec5e7fc920423e4b4ba4fbfae8612e3af2. The version of `dont-select-any-outputs` that I see is at a3422536dd049df388e05b269b462c878093221d.

mejsiej:
Everything is like you said, this and the breaking branch are exactly like this.

cameel:
But are the local labels you're using at these commits too? They could be out of sync with remote ones. Can you check if the commit ID printed by `git rev-parse HEAD` is what it should be when you check out these branches?

Also, which commands do you use to rebase it? I'm using these:
```bash
git checkout mejsiej/dont-select-any-outputs
git rebase origin/breaking --interactive
```
where:
- `origin` = `git@github.com:ethereum/solidity.git`
- `mejsiej` = `git@github.com:mejsiej/solidity.git`

mejsiej:
Ok, it is now rebased!

cameel:
I think you have pushed the wrong branch. It looks like one of your other PRs.

mejsiej:
No, the problem was that I used the version of CommandLineInterface.cpp from my branch and not breaking. Just changed it right now.

mejsiej:
All of the tests have passed.

cameel (approved):

Issue #11879, opened 2021-09-01
Array literals should be convertible to both dynamically-sized and statically-sized arrays


chriseth:
We could implement this using a new type for array literals and the codegen could store the elements in memory while reserving a memory slot for the length (and just not use it for the static case).

chriseth:
This could also solve the weird type deduction for array literals: Currently, `[1,2,3]` is a `uint8[3]` and `uint[3] t = [1,2,3];` is invalid. If we do not assign a "proper" array type to the array literal expression right away, but rather keep it more similar to how tuple types currently work, this should be much easier to use.

ekpyron:
We don't need to (nor want to?) allow naming these types, do we?
Then it would only be usable in contexts in which we could deduce whether it will end up dynamically or statically sized, so code generation would already know and be able to handle the cases separately without reserving slots and the like.

chriseth:
That is true, but we would need to pass that through the type checker, which is not always easy to do.

ekpyron:
At one point or another, the type checker needs to check if the array literal expression is convertible to the type it'll end up in and can just annotate that then, can't it?
It'd even consistently work for plain ``[1,2,3];`` or even for ``[f(), g(), h()];``, which wouldn't be annotated by any concrete array type, meaning that nothing needs to be stored anywhere and the subexpressions can just be evaluated and the results discarded in codegen...

chriseth:
`((([1,2,3])))` - it can be arbitrarily nested.

chriseth:
Oh and there also is `[[1,2], [4,5]]`

ekpyron:
Ok, maybe it's nontrivial :-). But should still be possible, shouldn't it :-)...?

And those nested single-component tuples are quite annoying everywhere, maybe we should just get rid of that once and for all :-D. But that's a different topic.

chriseth:
Implementation note:
Should introduce a new type similar to `TupleType` (maybe that one can be re-used with an additional boolean) which has one separate type per component. In `TypeChecker::visit(TupleExpression`, most of the checks can be removed. The type of the expression should be the new inline array type, where each component is just the type of the component.

Convertibility of this new type should be to statically-sized and dynamically-sized array type where each component type is convertible and the size matches.

Code generation should store each component in memory (and probably already convert to the target type).

Since we already need to convert to the target type anyway, we can maybe use @ekpyron 's idea of forwarding the target conversion type.

So it might be that this gets a bit more complicated.

[...]

PR #12372, opened 2021-12-06
Added sameType check for fromType and toType in YulUtilFunctions.cpp and relevant tests in semanticTests

nishant-sachdeva:
Solves #12057

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,7 @@ Bugfixes:
  * SMTChecker: Fix internal error when an unsafe target is solved more than once and the counterexample messages are different.
  * SMTChecker: Fix soundness of assigned storage/memory local pointers that were not erasing enough knowledge.
  * Fix internal error when a function has a calldata struct argument with an internal type inside.
-
+ * IR Generator: Fix IR syntax error when copying storage arrays of functions.
 
 
 ### 0.8.10 (2021-11-09)
--- a/libsolidity/codegen/YulUtilFunctions.cpp
+++ b/libsolidity/codegen/YulUtilFunctions.cpp
@@ -2057,7 +2057,7 @@ string YulUtilFunctions::copyValueArrayStorageToStorageFunction(ArrayType const&
 			solAssert(!_fromType.isValueType(), "");
 		templ("functionName", functionName);
 		templ("resizeArray", resizeArrayFunction(_toType));
-		templ("arrayLength",arrayLengthFunction(_fromType));
+		templ("arrayLength", arrayLengthFunction(_fromType));
 		templ("panic", panicFunction(PanicCode::ResourceError));
 		templ("srcDataLocation", arrayDataAreaFunction(_fromType));
 		templ("dstDataLocation", arrayDataAreaFunction(_toType));
@@ -2065,7 +2065,14 @@ string YulUtilFunctions::copyValueArrayStorageToStorageFunction(ArrayType const&
 		unsigned itemsPerSlot = 32 / _toType.storageStride();
 		templ("itemsPerSlot", to_string(itemsPerSlot));
 		templ("multipleItemsPerSlotDst", itemsPerSlot > 1);
-		bool sameType = _fromType.baseType() == _toType.baseType();
+		bool sameType = *_fromType.baseType() == *_toType.baseType();
+		if (auto functionType = dynamic_cast<FunctionType const*>(_fromType.baseType()))
+		{
+			solAssert(functionType->equalExcludingStateMutability(
+				dynamic_cast<FunctionType const&>(*_toType.baseType())
+			));
+			sameType = true;
+		}
 		templ("sameType", sameType);
 		if (sameType)
 		{
--- a/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
+++ b/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
@@ -1,0 +1,43 @@
+contract C {
+    function testFunction1() public {}
+    function testFunction2() public {}
+    function testFunction3() public {}
+
+
+    function() external [3] externalArray0;
+    function() external [3] externalArray1 = [
+        this.testFunction1,
+        this.testFunction2,
+        this.testFunction3
+    ];
+
+    function copyExternalStorageArrayOfFunctionType() external returns (bool) {
+        assert(keccak256(abi.encode(externalArray0)) != keccak256(abi.encode(externalArray1)));
+        externalArray0 = externalArray1;
+        return keccak256(abi.encode(externalArray0)) == keccak256(abi.encode(externalArray1));
+    }
+
+    function() internal [3] internalArray0;
+    function() internal [3] internalArray1 = [
+        testFunction1,
+        testFunction2,
+        testFunction3
+    ];
+
+    function copyInternalArrayOfFunctionType() external returns (bool) {
+        internalArray0 = internalArray1;
+        assert(internalArray0.length == 3);
+
+        return
+        internalArray0.length == internalArray1.length &&
+        internalArray0[0] == internalArray1[0] &&
+        internalArray0[1] == internalArray1[1] &&
+        internalArray0[2] == internalArray1[2];
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// copyExternalStorageArrayOfFunctionType() -> true
+// gas legacy: 103412
+// copyInternalArrayOfFunctionType() -> true
--- a/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic_different_mutability.sol
+++ b/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic_different_mutability.sol
@@ -1,0 +1,45 @@
+contract C {
+    function testFunction1() public {}
+    function testFunction2() public view {}
+    function testFunction3() public pure {}
+
+    function() external [3] externalArray0;
+    function() external [3] externalArray1 = [
+        this.testFunction1,
+        this.testFunction2,
+        this.testFunction3
+    ];
+
+    function copyExternalStorageArraysOfFunctionType() external returns (bool)
+    {
+        assert(keccak256(abi.encodePacked(externalArray0)) != keccak256(abi.encodePacked(externalArray1)));
+        externalArray0 = externalArray1;
+        return keccak256(abi.encodePacked(externalArray0)) == keccak256(abi.encodePacked(externalArray1));
+    }
+
+    function() internal [3] internalArray0;
+    function() internal [3] internalArray1 = [
+        testFunction1,
+        testFunction2,
+        testFunction3
+    ];
+
+    function copyInternalArrayOfFunctionType() external returns (bool)
+    {
+        internalArray0 = internalArray1;
+        assert(internalArray0.length == 3);
+
+        return
+        internalArray0.length == internalArray1.length &&
+        internalArray0[0] == internalArray1[0] &&
+        internalArray0[1] == internalArray1[1] &&
+        internalArray0[2] == internalArray1[2];
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// copyExternalStorageArraysOfFunctionType() -> true
+// gas legacy: 103398
+// copyInternalArrayOfFunctionType() -> true
+// gas legacy: 104178
--- a/test/libsolidity/semanticTests/array/copying/copy_function_internal_storage_array.sol
+++ b/test/libsolidity/semanticTests/array/copying/copy_function_internal_storage_array.sol
@@ -18,6 +18,6 @@ contract C {
 // compileViaYul: also
 // ----
 // test() -> 7
-// gas irOptimized: 126552
+// gas irOptimized: 124080
 // gas legacy: 205196
 // gas legacyOptimized: 204987
--- a/test/libsolidity/syntaxTests/array/copy_storage_arrays_of_function_type.sol
+++ b/test/libsolidity/syntaxTests/array/copy_storage_arrays_of_function_type.sol
@@ -1,0 +1,9 @@
+contract C {
+    function() external [] s0;
+    function() external view [] s1;
+    function copyStorageArrayOfFunctionType() public {
+        s1 = s0;
+    }
+}
+// ----
+// TypeError 7407: (148-150): Type function () external[] storage ref is not implicitly convertible to expected type function () view external[] storage ref.


Changelog.md
```diff
@@ -9,6 +9,7 @@ Compiler Features:
 Bugfixes:
  * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to ewasm.
  * SMTChecker: Fix internal error when an unsafe target is solved more than once and the counterexample messages are different.
+ * Fix Invalid IR generation error when copying dynamically sized storage arrays
```
```suggestion
 * IR Generator: Fix IR syntax error when copying storage arrays of functions.
```
(want to make clear that it was not "wrong IR" but "IR that does not compile", which is a huge difference)

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,12 @@
...
+  function() external [] s1;
+  constructor() {
+    s0 = s1;
+  }
+}
+// ====
+// compileToEwasm: also
+// compileViaYul: also
+// ----
+// () -> FAILURE
```
chriseth: "FAILURE" does not look good.
Can you move it to a function and add some code to the function that actually copies data and then call the function in the test?

chriseth: Same for the other test.

nishant-sachdeva: I'm not sure if it completely addressed the concern here, but I've moved the code to a function , and called it in the test.
Pushed the same in the latest commit

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol~Stashed changes
```diff
@@ -0,0 +1,16 @@
+contract C {
```
hrkrshnn: This file seems to be included by accident?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2057,15 +2057,22 @@ string YulUtilFunctions::copyValueArrayStorageToStorageFunction(ArrayType const&
...
 		unsigned itemsPerSlot = 32 / _toType.storageStride();
 		templ("itemsPerSlot", to_string(itemsPerSlot));
 		templ("multipleItemsPerSlotDst", itemsPerSlot > 1);
-		bool sameType = _fromType.baseType() == _toType.baseType();
+		bool sameType = *_fromType.baseType() == *_toType.baseType();
+		if (auto functionType = dynamic_cast<FunctionType const*>(_fromType.baseType()))
+		{
+			solAssert(functionType->equalExcludingStateMutability(
+				dynamic_cast<FunctionType const&>(*_toType.baseType())
+			));
```
hrkrshnn, line 2073: Can't this be made stricter, i.e., we should assert that `toType`'s state mutability is greater than or equal to that of `fromType`? So a `view` function can be assigned to a function without state mutability, but not the other way around.

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
```diff
@@ -0,0 +1,27 @@
+contract C {
+  function() external [] s0;
+  function() external view [] s1;
+  function copyStorageArrayOfFunctionType() public {
+    s0 = s1;
```
hrkrshnn: Can you add a test where the assignment is the other way around? That should result in an error.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol~Stashed changes
nishant-sachdeva: Thanks for pointing that out! Removed it in the latest push.

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
nishant-sachdeva: Indeed yes. It gives an error. 
I ran the command solc --ir test.sol and the error I get is 

`Type function () external[] storage ref is not implicitly convertible to expected type function () view external[] storage ref.`

Is this error expected? or do we need to make some change in this behavior?

Changelog.md
```diff
@@ -10,7 +10,7 @@ Bugfixes:
  * Code Generator: Fix a crash when using ``@use-src`` and compiling from Yul to ewasm.
  * SMTChecker: Fix internal error when an unsafe target is solved more than once and the counterexample messages are different.
  * Fix internal error when a function has a calldata struct argument with an internal type inside.
-
+ * IR Generator : Fix IR syntax error when copying storage arrays of functions.
```
```suggestion
 * IR Generator: Fix IR syntax error when copying storage arrays of functions.
```

libsolidity/codegen/YulUtilFunctions.cpp
chriseth: I don't think this falls into the responsibility of the code generator. If we at some time allow to have such a type conversion (for example through explicit conversion), the code generator should not fail. What we care about here is that the data representation is the same.

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
chriseth: Yes, it is expected. You have to move the changed test into the syntaxTests directory. Inside the semanticTests, everything has to compile without error.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
...
+}
+contract copyDynamicArrays {
+  C public testContract;
+
+  constructor() {
+    testContract = new C();
+  }
+
+  function tryCopyingDynamicallySizedStorageArray() public returns (uint ret) {
+    try testContract.copyDynamicallySizedStorageArray() {
```
chriseth: If you don't expect an error, you don't have to use `try`

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
+contract C {
+  function() external [] s0;
+  function() external [] s1;
+  function copyDynamicallySizedStorageArray() public {
+    s0 = s1;
```
chriseth: Please store something inside the array before copying and also check that the copy contains it afterwards.

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
nishant-sachdeva: Okay! I'll add a third test covering this behavior in the Syntax Tests folder.

libsolidity/codegen/YulUtilFunctions.cpp
cameel: Yeah, the only situation where the data representation can be different is if the function kind is different (external vs internal for example) and the `equalExcludingStateMutability()` assert already takes care of that so I think the current asserts are enough.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+}
+contract copyDynamicArrays {
+  C public testContract;
+
+  constructor() {
+    testContract = new C();
+  }
```
cameel: I think we should stick to our own style guide that [recommends 4-space indents](https://docs.soliditylang.org/en/latest/style-guide.html#indentation). The two space ones are really annoying, especially with deeper nesting when you want to see which brace ends which statement.

```suggestion
contract copyDynamicArrays {
    C public testContract;

    constructor() {
        testContract = new C();
    }
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
+contract C {
+  function() external [] s0;
+  function() external [] s1;
+  function copyDynamicallySizedStorageArray() public {
+    s0 = s1;
+  }
+
+}
+contract copyDynamicArrays {
```
cameel: This name sounds like a function, not a contract.
```suggestion
contract DynamicArrayCopy {
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
cameel: As long as it's not relevant to the test, I'd make it more concise:
```suggestion
  C public testContract = new C();
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
...
+  C public testContract;
+
+  constructor() {
+    testContract = new C();
+  }
+
+  function tryCopyingDynamicallySizedStorageArray() public returns (uint ret) {
```
cameel:
```suggestion
  function tryCopyingDynamicallySizedStorageArray() public returns (uint) {
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
cameel: Yeah. In case of error this will revert and you'll see `FAILURE` instead of result.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
...
+      return 1;
+    } catch {
+      return 0;
+    }
+  }
+}
+// ====
+// compileViaYul: also
+// ----
+// tryCopyingDynamicallySizedStorageArray() -> 1
```
cameel: Actually, this whole contract seems superfluous. An instance of the contract gets created automatically and you can just call the function on it in expectations:
```suggestion
// ====
// compileViaYul: also
// ----
// copyDynamicallySizedStorageArray() ->
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,27 @@
+contract C {
+  function() external [] s0;
+  function() external [] s1;
+  function copyDynamicallySizedStorageArray() public {
+    s0 = s1;
+  }
```
cameel: Can you also add the same test for internal functions? Can be in the same contract.

test/libsolidity/semanticTests/array/copying/copy_function_storage_array.sol
```diff
@@ -18,6 +18,6 @@ contract C {
 // compileViaYul: also
 // ----
 // test() -> 7
-// gas irOptimized: 126552
+// gas irOptimized: 124080
```
cameel, line 21: This looks bad. Gas usage changed so this was generating code, just differently. Smells like a codegen error to me. I think we should investigate (compare at least the `--ir` output before and after).

@chriseth What do you think? Results are still correct so could it really have been generating code that works incorrectly?

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
cameel: Also check if the test is not already there.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
cameel: Overall most of my comments about this test apply to the other one too.

test/libsolidity/semanticTests/array/copying/copy_function_storage_array.sol
chriseth: The main fix in this PR is activating a special case that was previously always (or mostly) disabled because of an incorrect check. This reduces gas costs obviously.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,21 @@
+contract C {
+    function testFunction1() public {}
+    function testFunction2() public {}
+    function testFunction3() public {}
+
+
+    function() external [] externalArray0 = new function() external[] (3);
+    function() external [] externalArray1 = [this.testFunction1, this.testFunction2, this
+    .testFunction3];
```
cameel: This is the way we usually wrap parameters, lists, etc. and I think it's more readable than wrapping it at a random spot in the expression:
```suggestion
    function() external [] externalArray1 = [
        this.testFunction1,
        this.testFunction2,
        this.testFunction3
    ];
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
nishant-sachdeva: In the PR , for external functions, we're trying two cases
1. external V external
2. external V external view

Should we do both corresponding cases for internal functions as well ? or just internal V internal would be okay ?

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
cameel: By the way, I think it' won't pass tests. We don't support dynamic array literals yet (#11879). To make it work you'll probably need to use a statically-sized array:
```suggestion
    function() external[3] externalArray1 = [this.testFunction1, this.testFunction2, this
    .testFunction3];
```
Or if you specifically want a dynamic array, you have to initialize the elements in the constructor.

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,21 @@
...
+
+    function copyExternalStorageArrayOfFunctionType() external returns (bool)
+    {
```
cameel: For some reason, we only use this style in C++ :) Not really sure why. In Solidity code this should look like this:
```suggestion
    function copyExternalStorageArrayOfFunctionType() external returns (bool) {
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,21 @@
...
+        externalArray0 = externalArray1;
+        return keccak256(abi.encodePacked(externalArray0)) == keccak256(abi.encodePacked(externalArray1));
```
cameel: Now that I think of it, we probably should use `encode()` most of the time. `encodePacked()` will work too but we were thinking of deprecating it at some point so using `encode()` is better unless we have a good reason not to.
```suggestion
        return keccak256(abi.encode(externalArray0)) == keccak256(abi.encode(externalArray1));
```

cameel: Might be a good idea to also compare them before assingment to make sure they were really different.
```suggestion
        assert(keccak256(abi.encodePacked(externalArray0)) != keccak256(abi.encodePacked(externalArray1)));
        externalArray0 = externalArray1;
        return keccak256(abi.encodePacked(externalArray0)) == keccak256(abi.encodePacked(externalArray1));
```

test/libsolidity/semanticTests/array/copying/copy_storage_arrays_of_function_types.sol
```diff
@@ -0,0 +1,20 @@
+contract C {
+    function testFunction1() external view {}
+    function testFunction2() external view {}
+    function testFunction3() external view {}
```
cameel: Let's make the test even more interesting:
```suggestion
    function testFunction1() external {}
    function testFunction2() external view {}
    function testFunction3() external pure {}
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,21 @@
+contract C {
```
cameel: The fact that the array contains function pointers is the key point here so it should be included in the name. Looking at naming in this test dir, I'd call your new test cases something like
- `copy_dynamically_sized_storage_array.sol` -> `array_of_function_external_storage_to_storage_dynamic.sol`
- `copy_storage_arrays_of_function_types.sol` -> `array_of_function_external_storage_to_storage_dynamic_different_mutability.sol`

The existing `copy_function_storage_array.sol` should also really have `internal` somewhere in its name.

test/libsolidity/semanticTests/array/copying/copy_function_storage_array.sol
cameel: @chriseth But are we 100% sure that the code without this special case was correct? Was that just a bunch of extra conversion functions that would disappear when inlined?

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,36 @@
...
+        return keccak256(abi.encodePacked(externalArray0)) == keccak256(abi.encodePacked(externalArray1));
+    }
+
+    function() internal [] internalArray0 = new function() internal[] (3);
+    function() internal [] internalArray1 = [testFunction1, testFunction2, testFunction3];
+
+    function copyInternalArrayOfFunctionType() external returns (bool)
+    {
+        internalArray0 = internalArray1;
+        for(uint i = 0; i<internalArray0.length; i++)
```
cameel:
```suggestion
        for (uint i = 0; i < internalArray0.length; i++)
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
```diff
@@ -0,0 +1,36 @@
...
+        {
+            if (internalArray0[i] != internalArray1[i])
+                return false;
+        }
+        return true;
```
cameel: This will not detect the case where the arrays are different lengths but have equal initial elements :)

The array is short and has hard-coded length so it might be simpler to just compare elements manually here:
```suggestion
        assert(internalArray1.length == 3);
        return
            internalArray0.length == 3 &&
            internalArray0[0] == internalArray1[0] &&
            internalArray0[1] == internalArray1[1] &&
            internalArray0[2] == internalArray1[2];
```

test/libsolidity/semanticTests/array/copying/copy_dynamically_sized_storage_array.sol
cameel: I'd do both unless we already have a test for it somewhere. Better too many than too few as long as the tests are concise (the way they are now seems fine +/- my comments).

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
```diff
@@ -0,0 +1,43 @@
...
+        internalArray0.length == internalArray1.length &&
+        internalArray0[0] == internalArray1[0] &&
+        internalArray0[1] == internalArray1[1] &&
+        internalArray0[2] == internalArray1[2];
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// copyExternalStorageArrayOfFunctionType() -> false
```
chriseth: Shouldn't this return true?

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic_different_mutability.sol
```diff
@@ -0,0 +1,45 @@
...
+        internalArray0.length == internalArray1.length &&
+        internalArray0[0] == internalArray1[0] &&
+        internalArray0[1] == internalArray1[1] &&
+        internalArray0[2] == internalArray1[2];
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// copyExternalStorageArraysOfFunctionType() -> false
```
chriseth: Shouldn't this return true?

test/libsolidity/semanticTests/array/copying/copy_function_internal_storage_array.sol
```diff
@@ -18,6 +18,6 @@ contract C {
 // compileViaYul: also
 // ----
 // test() -> 7
-// gas irOptimized: 126552
+// gas irOptimized: 124080
```
chriseth, line 21: @cameel I checked the assembly. The difference is that the old version copies element by element and the new version copies slot by slot, as expected.

chriseth (approved):

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
```diff
@@ -0,0 +1,43 @@
...
+    function testFunction2() public {}
+    function testFunction3() public {}
+
+
+    function() external [3] externalArray0;
+    function() external [3] externalArray1 = [
+        this.testFunction1,
+        this.testFunction2,
+        this.testFunction3
+    ];
```
cameel, line 12: I think we should test both static and dynamic arrays but the dynamic ones are more important - that's what the original issue was about.

Also the test name is wrong now - it says `dynamic` but uses static arrays.

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
```diff
@@ -0,0 +1,43 @@
...
+
+    function copyInternalArrayOfFunctionType() external returns (bool) {
+        internalArray0 = internalArray1;
+        assert(internalArray0.length == 3);
+
+        return
+        internalArray0.length == internalArray1.length &&
+        internalArray0[0] == internalArray1[0] &&
+        internalArray0[1] == internalArray1[1] &&
+        internalArray0[2] == internalArray1[2];
```
cameel, line 35:
```suggestion
        return
            internalArray0.length == internalArray1.length &&
            internalArray0[0] == internalArray1[0] &&
            internalArray0[1] == internalArray1[1] &&
            internalArray0[2] == internalArray1[2];
```

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
nishant-sachdeva: This PR had already been merged. So, I've added the change in #12417

PR #12373, opened 2021-12-06
Explanation about operators.

chriseth:


docs/types/operators.rst
```diff
@@ -1,7 +1,34 @@
...
+For example, you can compute ``y = x + z``, where ``x`` is a ``uint8`` and ``z`` has
+the type ``int32``. In these cases, the following mechanism will be used to determine
+the type in which the operation is computed (this is important in case of overflow)
+and the type of the operator result:
+
+1. If the type of the right operand can be implicitly converted to the type of the left
+   operand, use the type of the left operand,
+2. if the type of the left operand can be implicitly converted to the type of the right
+   operand, use the type of the right operand,
+3. otherwise, the operation is not allowed.
```
cameel, line 16: According to this rule `1 + 0.99` and `1 == 0.99` should both compile or both be an error. Only the first one works though.

Also, `1 == 1.0` does compile which means that `+` can implicitly convert `int_const` to `rational_const` but `==` only considers the opposite conversion.

docs/types/operators.rst
```diff
@@ -1,7 +1,34 @@
-.. index:: assignment, ! delete, lvalue
+.. index:: operator
```
cameel:
```suggestion
.. index:: ! operator
```

docs/types/operators.rst
```diff
@@ -1,7 +1,34 @@
...
+Binary operators can be applied even if the two operands do not have the same type.
+For example, you can compute ``y = x + z``, where ``x`` is a ``uint8`` and ``z`` has
+the type ``int32``. In these cases, the following mechanism will be used to determine
+the type in which the operation is computed (this is important in case of overflow)
+and the type of the operator result:
+
+1. If the type of the right operand can be implicitly converted to the type of the left
+   operand, use the type of the left operand,
+2. if the type of the left operand can be implicitly converted to the type of the right
+   operand, use the type of the right operand,
```
cameel, line 15: Maybe we should mention that operator `=` is a special case and the second rule does not apply to it?

docs/types/operators.rst
chriseth: We could say that these are the generic rules, but they may be different for some types - check the documentation for the types to be sure.

docs/types/operators.rst
chriseth: Sure, `=` does not even perform an operation.

chriseth: I mean, boolean operators are also special. Maybe limit this to "arithmetic and bitwise operators"?

docs/types/operators.rst
cameel: We could, but it would be nice to have consistent rules for this. Is there a good reason for comparison between int and fixed-point not to work?

BTW, I don't see this quirk documented for the fixed-point type. It only says it provides the `==` operator.

docs/types/operators.rst
cameel: Sounds good.

cameel: > Sure, `=` does not even perform an operation.

You can use it in an expression though and it returns a value so technically it's still a binary operator, right?

docs/types/operators.rst
chriseth: There is no such thing as `int_const`. `1.0` and `1` are absolutely identical.

docs/types/operators.rst
cameel: What about this one? Isn't this the main section about operators?

docs/types/operators.rst
```diff
@@ -1,7 +1,35 @@
...
-Operators Involving LValues
-===========================
+Operators
+=========
+
+Arithmetic and bit operators can be applied even if the two operands do not have the same type.
+For example, you can compute ``y = x + z``, where ``x`` is a ``uint8`` and ``z`` has
+the type ``int32``. In these cases, the following mechanism will be used to determine
+the type in which the operation is computed (this is important in case of overflow)
+and the type of the operator result:
```
cameel:
```suggestion
and the type of the operator's result:
```

docs/types/operators.rst
cameel: Oh, true.

Anyway, I still think this inconsistency should be documented somewhere.

docs/types/operators.rst
```diff
@@ -1,7 +1,35 @@
...
+2. if the type of the left operand can be implicitly converted to the type of the right
+   operand, use the type of the right operand,
+3. otherwise, the operation is not allowed.
+
+In case one of the operands is a :ref:`literal number <rational_literals>` it is first converted to its
+"mobile type", which is the smallest type that can hold the value
+(unsigned types of the same bit-width are considered "smaller" than the signed types).
+If both are literal numbers, the operation is computed with arbitrary precision.
+
+The operator result type is the same as the type the operation is performed in,
```
cameel:
```suggestion
The operator's result type is the same as the type the operation is performed in,
```

docs/types/operators.rst
```diff
@@ -1,7 +1,35 @@
...
+except for comparison operators where the result is always ``bool``.
+
+The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
+left operand for the operation and the result.
+
+.. index:: assignment, lvalue, compound operators
```
cameel:
```suggestion
.. index:: assignment, lvalue, ! compound operators
```

cameel (dismissed):
Just a few minor tweaks needed so I'm already approving.

PR #12378, opened 2021-12-07
Report lack of contracts to compile and uncompiled empty interfaces / abstract contracts.

### LINKED ISSUE #4617: [CLI] Solc should report if there was no contract to compile
Follow-up of https://github.com/ethereum/solidity/pull/4525 :

After that PR, solc properly reports why it does not output anything. There is still one edge, case, though: If you compile a file that does not contain any contracts, it still does not output anything. We should report something like 'input did not contain any contracts'. We should also check what happens if contracts are present but are abstract or interfaces.

mejsiej:
Closes #4617.

Implemented changes:
-  If the input does not have any contracts to compile the compiler shows a message "Input did not contain any contracts to compile."

- If the input has contracts to compile but some of them are empty abstract contracts or empty interfaces then the compiler reports that they have not been compiled because of that.

### DIFF
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -1097,7 +1097,10 @@ void CommandLineInterface::outputCompilationResults()
 {
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
-	if (!m_options.compiler.outputs.signatureHashes && !m_options.compiler.outputs.irOptimized)
+	if (!m_options.compiler.outputs.signatureHashes &&
+		!m_options.compiler.outputs.astCompactJson &&
+		!m_options.compiler.combinedJsonRequests.has_value()
+	)
 	{
 		size_t contractsWithEmptyBytecode = 0;
 		for (string const& contract: m_compiler->contractNames())
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -1098,14 +1098,14 @@ void CommandLineInterface::outputCompilationResults()
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
 	set<string> uncompiledContracts;
-	for (auto const& contract: m_compiler->contractNames())
+	for (string const& contract: m_compiler->contractNames())
 		if (objectWithLinkRefsHex(m_compiler->object(contract)).empty())
 			uncompiledContracts.insert(contract.substr(contract.find_last_of('/') + 1));
 
 	if (uncompiledContracts.size() == m_compiler->contractNames().size())
 		sout() << "Input did not contain any contracts to compile." << endl;
 	else
-		for (auto const& uncompiledContract: uncompiledContracts)
+		for (string const& uncompiledContract: uncompiledContracts)
 		{
 			sout() << "'" << uncompiledContract << "' ";
 			sout() << "was not compiled because it is an empty interface or empty abstract contract." << endl;
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -1101,6 +1101,7 @@ void CommandLineInterface::outputCompilationResults()
 		!m_options.compiler.outputs.ir &&
 		!m_options.compiler.outputs.irOptimized &&
 		!m_options.compiler.outputs.astCompactJson &&
+		!m_options.modelChecker.initialize &&
 		!m_options.compiler.combinedJsonRequests.has_value()
 	)
 	{
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -1098,21 +1098,18 @@ void CommandLineInterface::outputCompilationResults()
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
 	std::set<std::string> uncompiledContracts;
-	for (const auto& contract: m_compiler->contractNames())
+	for (auto const& contract: m_compiler->contractNames())
 		if (objectWithLinkRefsHex(m_compiler->object(contract)).empty())
 			uncompiledContracts.insert(contract.substr(contract.find_last_of('/') + 1));
 
-	if (!uncompiledContracts.empty())
-		if (uncompiledContracts.size() == m_compiler->contractNames().size())
-			sout() << "Input did not contain any contracts to compile." << endl;
-		else
-			for (const auto& uncompiledContract: uncompiledContracts)
-			{
-				sout() << "'" << uncompiledContract << "' ";
-				sout() << "was not compiled because it is an empty interface or empty abstract contract." << endl;
-			}
-	else if (uncompiledContracts.size() == m_compiler->contractNames().size())
+	if (uncompiledContracts.size() == m_compiler->contractNames().size())
 		sout() << "Input did not contain any contracts to compile." << endl;
+	else
+		for (auto const& uncompiledContract: uncompiledContracts)
+		{
+			sout() << "'" << uncompiledContract << "' ";
+			sout() << "was not compiled because it is an empty interface or empty abstract contract." << endl;
+		}
 
 
 	handleCombinedJSON();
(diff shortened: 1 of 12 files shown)


solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,21 @@ void CommandLineInterface::outputCompilationResults()
 {
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
+	std::set<std::string> uncompiledContracts;
```
cameel: Our convention is to use `using namespace std;` in .cpp files and drop the `std::` prefix there.
```suggestion
	set<string> uncompiledContracts;
```

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,21 @@ void CommandLineInterface::outputCompilationResults()
...
+	for (auto const& contract: m_compiler->contractNames())
+		if (objectWithLinkRefsHex(m_compiler->object(contract)).empty())
+			uncompiledContracts.insert(contract.substr(contract.find_last_of('/') + 1));
+
+	if (uncompiledContracts.size() == m_compiler->contractNames().size())
+		sout() << "Input did not contain any contracts to compile." << endl;
+	else
+		for (auto const& uncompiledContract: uncompiledContracts)
```
cameel: The type is very short and readable here so no point in hiding it behind `auto`:
```suggestion
	for (string const& contract: m_compiler->contractNames())
		if (objectWithLinkRefsHex(m_compiler->object(contract)).empty())
			uncompiledContracts.insert(contract.substr(contract.find_last_of('/') + 1));

	if (uncompiledContracts.size() == m_compiler->contractNames().size())
		sout() << "Input did not contain any contracts to compile." << endl;
	else
		for (string const& uncompiledContract: uncompiledContracts)
```

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,21 @@ void CommandLineInterface::outputCompilationResults()
 {
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
+	std::set<std::string> uncompiledContracts;
+	for (auto const& contract: m_compiler->contractNames())
+		if (objectWithLinkRefsHex(m_compiler->object(contract)).empty())
```
cameel: `object()` requires the `CompilerStack` state to be `CompilationSuccessful`. Can you make sure that this code works with `--stop-after=parsing`, which stops at the analysis phase and does not actually perform compilation? We should add a test case for this (unless we already have one).

cameel: `objectWithLinkRefsHex()` does extra processing you don't need (hex-encoding, listing link references).
```suggestion
		if (m_compiler->object(contract).bytecode.empty())
```

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,21 @@ void CommandLineInterface::outputCompilationResults()
...
+			uncompiledContracts.insert(contract.substr(contract.find_last_of('/') + 1));
```
cameel: Are you trying to extract the contract name here? Fully-qualified names in `CompilerStack` have the form `<source unit name>:<contract name>` so you should split on the last `:` (`<contract name>` cannot contain `:`).

Also, since contract names are not necessarily unique, I think we should list the file name too.

solc/CommandLineInterface.cpp
cameel: I'd be more explicit about what is being kept on this list.
```suggestion
	std::set<std::string> contractsWithEmptyBytecode;
```

solc/CommandLineInterface.cpp
cameel: I don't think we need the `else` here. Having interfaces and abstract contracts is **very** common. We don't want to always list them. We only want to prompt the user when all contracts in the project are like this and the `if` already covers that.

I'd also adjust the message to be clearer:
```suggestion
	if (uncompiledContracts.size() == m_compiler->contractNames().size())
	{
		sout() << "Input does not contain any deployable contracts. ";
		sout() << "Note that interface and abstract contract definitions do not produce bytecode." << endl;
	}
```

Alternatively, you could have an `if` that prints that there were no contracts if `m_compiler->contractNames().empty()` and that there were no deployable contracts otherwise (with the note about interfaces/abstract contracts).

You could also say how many non-deployable contracts were actually found but that's just extra.

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,21 @@ void CommandLineInterface::outputCompilationResults()
 {
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
```
cameel, line 1098: Note that this function processes a lot of option combinations. Some of them do not actually result in full compilation being performed. For example the `--stop-after` I already mentioned (though I'm not 100% certain that one actually enters here) but also `--hashes`, which is why command-line tests fail. You might need to add a check that `CompilerStack` state is `CompilationSuccessful` but you should also make sure this does not exclude cases where we actually want the message to show up.

cameel (changes requested):
This needs some changes. Also tests fail, some due to problems I listed, some because they just need their expectations updated.

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,26 @@ void CommandLineInterface::outputCompilationResults()
...
 
+	if (!m_options.compiler.outputs.signatureHashes &&
+		!m_options.compiler.outputs.ir &&
+		!m_options.compiler.outputs.irOptimized &&
+		!m_options.compiler.outputs.astCompactJson &&
+		!m_options.modelChecker.initialize &&
+		!m_options.compiler.combinedJsonRequests.has_value()
+	)
```
cameel, line 1106:
```suggestion
	if (
		!m_options.compiler.outputs.signatureHashes &&
		!m_options.compiler.outputs.ir &&
		!m_options.compiler.outputs.irOptimized &&
		!m_options.compiler.outputs.astCompactJson &&
		!m_options.modelChecker.initialize &&
		!m_options.compiler.combinedJsonRequests.has_value()
	)
```

solc/CommandLineInterface.cpp
```diff
@@ -1097,6 +1097,26 @@ void CommandLineInterface::outputCompilationResults()
 {
 	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
 
+	if (!m_options.compiler.outputs.signatureHashes &&
+		!m_options.compiler.outputs.ir &&
+		!m_options.compiler.outputs.irOptimized &&
+		!m_options.compiler.outputs.astCompactJson &&
+		!m_options.modelChecker.initialize &&
```
cameel, line 1104: I think that it should still be possible to enter this `if` when using model checker, `--hashes` or IR output. You can still request for example `--bin` together with all these and it can result in the compiler telling you `Compiler run successful, no output requested.`, which is not true.

 Consider this example:
```solidity
// SPDX-License-Identifier: GPL-2.0
pragma solidity >=0.0;

function f() pure {
    assert(true == true);
}
```
```bash
solc test.sol --model-checker-engine all --bin
```

mejsiej: Alright, I will let them get there and update the expected outputs in cmdlineTests.

cameel:
@mejsiej I think we need to change the approach here. See https://github.com/ethereum/solidity/issues/4617#issuecomment-991337025 for details.

I originally took the issue at a face value but while checking your code I noticed that compiler actually never exits without printing something. The issue was outdated and needed some adjustment. We still need it, just not exactly the way the description said.

So currently the PR solves points (1) and (2) listed in my comment. I think we also need (3) and (4). And in case of (2) the compiler already prints sensible output so I think it's better to leave it as it was. We might want to change that behavior a little so I reported #12398 but that issue will require a design decision and might turn out to be a breaking change so I'd keep it out of this PR.

solc/CommandLineInterface.cpp
cameel: Actually, please see https://github.com/ethereum/solidity/pull/12378#issuecomment-991356098 first.

leonardoalt:
@mejsiej would you like to apply the latest comments?
@cameel or do you want to take over?

leonardoalt:
Bump question

cameel:
Not really, I'm already working on too many things at once. I really need to finish ext tests first. I can review it but with the amount of PRs we're getting I unfortunately didn't manage to get back to this one yet.

cameel:
@mejsiej if you want to resume working on it later, feel free to reopen. For the time being I'm closing this since we're flooded with PRs and this not a critical roadmap feature (it's just a "nice to have") so it does not make much sense for us to take over and finish it ourselves.

PR #12380, opened 2021-12-07
External tests with native compiler binary

cameel:
Currently external tests can only use a `soljson.js` binary. This makes compilation slower and also makes running them with a locally compiled binary more cumbersome (you have to compile emscripten locally in addition to your normal build). Both Hardhat and Truffle support using a native binary and this PR extends our external test scripts to allow choosing between them.

This will be especially important for the (soon to be added) test runs with compilation via IR, where the compilation is slower overall and the difference in external tests is very noticeable.

The PR also temporarily adds an extra CI run with a native binary for every existing emscripten one. We should really choose one or the other. Native would be better by default with a fallback to emscripten in cases where it breaks (e.g. Colony: https://github.com/trufflesuite/truffle/issues/4499). For now I need to see which run succeed and also compare the running time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -396,6 +396,11 @@ defaults:
       requires:
         - b_ubu_release
 
+  - workflow_ubuntu2004_static: &workflow_ubuntu2004_static
+      <<: *workflow_trigger_on_tags
+      requires:
+        - b_ubu_static
+
   - workflow_archlinux: &workflow_archlinux
       <<: *workflow_trigger_on_tags
       requires:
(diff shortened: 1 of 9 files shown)


cameel:
This PR renames some of the jobs so it needs repo settings to be adjusted:
- `t_ems_compile_ext_gnosis` -> `t_native_compile_ext_gnosis`
- `t_ems_test_ext_ens` -> `t_native_test_ext_ens`
- `t_ems_test_ext_gnosis_v2` -> `t_native_test_ext_gnosis_v2`
- `t_ems_test_ext_zeppelin` -> `t_native_test_ext_zeppelin`

I'd also make `t_ems_test_ext_gnosis_v2` and `t_ems_compile_ext_gnosis` not required because my upcoming PRs will remove these jobs anyway. Only `t_ems_test_ext_gnosis` will remain.

chriseth (approved):

PR #12381, opened 2021-12-08
Improve description of break/continue restrictions.

acoglio:
This is as discussed on Element.

### DIFF
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -541,10 +541,24 @@ evaluate to zero values.
 
 In all other situations, expressions have to evaluate to exactly one value.
 
-The ``continue`` and ``break`` statements can only be used inside loop bodies
-and have to be in the same function as the loop (or both have to be at the
-top level). The ``continue`` and ``break`` statements cannot be used
-in other parts of a loop, not even when it is scoped inside a second loop's body.
+A ``continue`` or ``break`` statement can only be used inside the body of a for-loop, as follows.
+Consider the innermost loop that contains the statement.
+The loop and the statement must be in the same function, or both must be at the top level.
+The statement must be in the loop's body block;
+it cannot be in the loop's initialization block or update block.
+It is worth emphasizing that this restriction applies just
+to the innermost loop that contains the ``continue`` or ``break`` statement:
+this innermost loop, and therefore the ``continue`` or ``break`` statement,
+may appear anywhere in an outer loop, possibly in an outer loop's initialization block or update block.
+For example, the following is legal,
+because the ``break`` occurs in the body block of the inner loop,
+despite also occurring in the update block of the outer loop:
+
+.. code-block:: yul
+
+    for {} true { for {} true {} { break } }
+    {
+    }
 
 The condition part of the for-loop has to evaluate to exactly one value.
 


docs/yul.rst
```diff
@@ -541,10 +541,11 @@ evaluate to zero values.
...
 In all other situations, expressions have to evaluate to exactly one value.
 
-The ``continue`` and ``break`` statements can only be used inside loop bodies
-and have to be in the same function as the loop (or both have to be at the
-top level). The ``continue`` and ``break`` statements cannot be used
-in other parts of a loop, not even when it is scoped inside a second loop's body.
+A ``continue`` or ``break`` statement can only be used inside the body of a for-loop, as follows.
+Consider the innermost loop that contains the statement.
+The loop and the statement must be in the same function, or both must be at the top level.
+The statement must in the loop's body block;
```
cameel:
```suggestion
The statement must be in the loop's body block;
```

docs/yul.rst
```diff
@@ -541,10 +541,11 @@ evaluate to zero values.
...
+it cannot be in the loop's initialization block or update block.
```
cameel, line 548: I'd make this statement even more general:
```suggestion
it cannot be in a loop's initialization block or update block.
```

cameel (commented):
Looks good to me. Just two tiny corrections.

docs/yul.rst
acoglio: Done, thanks, and sorry for the error.

docs/yul.rst
acoglio: I think that 'the' is important here, because it refers to the innermost loop defined by the sentence
```
Consider the innermost loop that contains the statement.
```
The break or continue must be in the innermost loop's body block, not in the innermost loop's initialization block or update block. The whole loop, and therefore the break or continue in question, could be anywhere in some outer loop, but that does not affect this requirement. We discussed this on Gitter, including this example:
```
for Init1 Test1 { ... for Init2 Test2 Update2 { ... break ... } ... } Body1
```
This is legal, even though the break occurs in a loop's update block, namely in the outer loop's update block. But the important loop is the inner one, and the break correctly occurs in the body block of the inner loop.

If 'the' is not clear enough, we could rephrase the last sentence as either
```
The statement must be in that loop's body block;
it cannot be in that loop's initialization block or update block.
```
or, even more explicitly,
```
The statement must be in that innermost loop's body block;
it cannot be in that innermost loop's initialization block or update block.
```

cameel: Oh, you got me with that example :) I didn't think of that, which is why I thought that it would be true to say that `break` could never appear in any loop's update block. Just shows that this clarification is indeed very necessary. I think I'd even go as far as to mention this example explicitly because currently the article is the only piece of text hinting at it.

cameel (commented):
This looks fine now (unless you want to also add that example).

Could you squash your commits?

docs/yul.rst
acoglio: I think that it's worth adding the example and explanation, so I've just pushed a commit with that. It everything looks good, I'll squash the commits and update this PR.

docs/yul.rst
```diff
@@ -541,10 +541,22 @@ evaluate to zero values.
...
+to the innermost loop that contains the ``continue`` or ``break`` statement:
+this innermost loop, and therefore the ``continue`` or ``break`` statement,
+may appear anywhere in an outer loop, possibly in an outer loop's initialization block or update block.
+For example, the following is legal as far as the ``break`` is concerned,
+because it occurs in the body block of the inner loop,
+despite also occurring in the update block of the outer loop:
+
+.. code-block:: yul
+
+    for Init1 Test1 { ... for Init2 Test2 Update2 { ... break ... } ... } Body1
```
cameel: CI is failing because this does not pass highlighter's syntax checks in Sphinx and also because we have scripts that extract every snippet marked as Yul or Solidity and check if it compiles.

You have 3 options:
- Don't set highlighting language to Yul. It's just pseudo-code after all.
- Force the highlighter to shut up using `:force:`
    ```rst
    .. code-block:: yul
        :force:
    
        for Init1 Test1 { ... for Init2 Test2 Update2 { ... break ... } ... } Body1
    ```
- Use a real code snippet
    ```rst
    .. code-block:: yul

        for {} true { for {} true {} { break } }
        {
        }
    ```

acoglio: Ah, I didn't realize that -- sorry. I like your suggestion to use real Yul code, because I think it makes the example even clearer. Thanks!

acoglio:
I'll squash the commits after the CIs pass.

cameel (approved):

chriseth (approved):

PR #12387, opened 2021-12-08
[CI] Adjust job parallelism to actually match available resources

cameel:
I have noticed that `parallelism` setting for our jobs does not actually match how parallel they actually are:
- `t_ubu_codecov` and `t_ubu_asan_soltest` run `soltest.sh` (not `soltest_all.sh`) so I think they do not take advantage of `parallelism: 6` at all.
- `t_ubu_soltest_all` and `t_ubu_release_soltest_all` actually do 15 separate soltest runs so `parallelism: 6` seems low. I think that increasing `parallelism` would decrease their runtime from 30 min to something more reasonable.

Still a draft because I need to see how much this actually speeds things up and potentially adjust the numbers. I set `soltest_all` parallelism to `15` (2.5 times more) but some runs are longer than others so I suspect it won't give us a 2.5x speed up. On the other hand maybe there's no harm in having it at 29 because I think that we only pay for the time each parallel job actually takes. I would have to look at our credit use to be able to verify that though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -695,7 +695,6 @@ jobs:
 
   t_ubu_codecov:
     <<: *base_ubuntu2004
-    parallelism: 6
     environment:
       EVM: << pipeline.parameters.evm-version >>
       OPTIMIZE: 1
@@ -878,7 +877,7 @@ jobs:
 
   t_ubu_soltest_all: &t_ubu_soltest_all
     <<: *base_ubuntu2004
-    parallelism: 6
+    parallelism: 15 # 7 EVM versions, each with/without optimization + 1 ABIv1/@nooptions run
     <<: *steps_soltest_all
 
   t_archlinux_soltest: &t_archlinux_soltest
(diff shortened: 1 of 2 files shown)


cameel:
This change decreased `soltest_all` runtime by half. It now takes 15-18 min compared to 30-33 min before. No results for `codecov` yet - still running (it normally takes ~1h).

There are actually 2 groups of runs: optimized that take ~5 min and non-optimized that take ~15 min. So it looks like 15 min is the lowest we can get without #12388.

cameel:
- `t_ubu_asan_soltest`: no significant difference, still at ~15 min.
- `t_ubu_codecov`: 1:06h and 1:10h compared to 1:07h in the today's nightly run.

I thinks it's safe to assume that these do not need parallelism. In that case I'm removing it and also changing `soltest_all` to 15.

cameel:
This is now ready for review.

chriseth (approved):

PR #12389, opened 2021-12-09
Remove redundant typo

gitpushOmnik:


docs/brand-guide.rst
```diff
@@ -59,7 +59,7 @@ Under the following terms:
 
 - **Attribution** — You must give appropriate credit, provide a link to
   the license, and indicate if changes were made. You may do so in any
-  reasonable manner, but not in any way that suggests the the Solidity
+  reasonable manner, but not in any way that suggests the Solidity
```
cameel: I think it was actually meant to be `that`:
```suggestion
  reasonable manner, but not in any way that suggests that the Solidity
```

PR #12394, opened 2021-12-10
Make optimizer level 2 in external tests actually different from level 3

cameel:
`{enabled: true}` is equivalent to `{enabled: true, details: {yul: true}}` in optimizer settings so currently level 2 is the same as 3. It just makes the test ineffective but does not break it, which is why it must have gone unnoticed when we made Yul optimizer setting dependent on the general optimizer setting.

This PR makes level 2 effective again. It also corrects minimum level for the gnosis test because fails on the real level 2. I also added comments explaining in each case why certain levels are skipped.

There are also two tiny unrelated corrections:
- Both gnosis tests used the same name. This does not make anything fail but affects the name of the temporary dir where the output is stored. With the same name it's hard to distinguish them.
- ENS test was writing config to `truffle-config.js` rather than the `truffle.js` present in the repo. It did not break the test because `truffle-config.js` takes precedence and `truffle.sj` does not contain anything critical but it was triggering a warning from Truffle.

### DIFF
--- a/test/externalTests/colony.sh
+++ b/test/externalTests/colony.sh
@@ -35,6 +35,7 @@ function colony_test
     local repo="https://github.com/solidity-external-tests/colonyNetwork.git"
     local branch=develop_080
     local config_file="truffle.js"
+    # On levels 1 and 2 it compiles but tests run out of gas
     local min_optimizer_level=3
     local max_optimizer_level=3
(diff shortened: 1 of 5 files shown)


chriseth (approved):
Nice, good catch!

PR #12396, opened 2021-12-10
External tests via IR

cameel:
~Depends on #12394.~

This PR refactors external tests to replace optimizer "levels" with named presets that represent a full set of settings. This lets us have create separate IR and legacy runs without complicating the mechanism for parallelizing tests in CI.

It also inserts `time` command into the function that executes compilation so that we can compare the running time of IR and legacy compilation.


cameel:
This is now rebased on `develop` and no longer depends on anything.

chriseth (approved):

PR #12397, opened 2021-12-10
Enable external test presets that compile but fail tests

cameel:
~Depends on #12396.~ Merged.

There are several cases where we have to disable a run with specific settings because it does not pass tests but it still compiles. Such a preset is skipped even in `COMPILE_ONLY` mode. This PR adjust the testing mechanism to make it possible to skip only the JS tests without skipping compilation.


cameel:
Base PR merged so this is now ready for review.

chriseth (approved):

Issue #12251, opened 2021-11-07
"UNLICENSED" (recommended for closed source in docs) is not a valid SPDX identifier, and the closest valid license (UNLICENSE) is the exact opposite. 


looksnap:
from: https://docs.soliditylang.org/en/latest/layout-of-source-files.html?highlight=unlicensed#spdx-license-identifier

> If you do not want to specify a license or if the source code is not open-source, please use the special value `UNLICENSED`.

Except that, "Unlicensed" is not a valid SPDX identifier (find the list here: https://spdx.org/licenses/), and the closest valid license is the "Unlicense" (https://spdx.org/licenses/Unlicense.html), which is: "This is free and unencumbered software released into the public domain."

looksnap:
And if you mean to say that "UNLICENSED" is a special value for Solidity specifically, perhaps you should change this special value to "None" or something that is less ambiguous. Etherscan uses "None" for no license, and "Unlicense" for public domain, and I've seen a couple contracts that use "UNLICENSED" in the SPDX, but chose "Unlicense" from the Etherscan dropdown.

chriseth:
It is not a valid SPDX identifier on purpose. Using `UNLICENSED` here specifically says that you are not granted a license to use the source code in this form. It cannot be a valid SPDX identifier, since those are all licenses.

We decided to use `UNLICENSED` after the recommendation from npm: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license

axic:
> We decided to use UNLICENSED after the recommendation from npm: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license

Perhaps we want to document this?

tied2token:
I don't know if I can iam a trustee of the ietf and they just granted me
the ability to use any of there's

looksnap:
Ok well, the most common place for the license to appear is in Etherscan and Bscscan, and you can find multiple verified contracts where the SPDX identifier is "UNLICENSED" but the License selected from the Etherscan/Bscscan dropdown is "Unlicense". Which one takes precedence?  

This indicates confused usage, regardless of whether that was the npm recommendation (which is a more generic recommendation anyways). Given the confusion in this specific area (solidity + smart contracts), I suggest the npm usage to not be followed.

At the very least, it should be well documented.

caffeinum:
I asked a question on StackOverflow, too, before finding this issue.

[...]

PR #12399, opened 2021-12-11
docs: note on invalid SPDX `UNLICENSED`

caffeinum:
#12251

### DIFF
--- a/docs/layout-of-source-files.rst
+++ b/docs/layout-of-source-files.rst
@@ -28,6 +28,9 @@ it does include the supplied string in the :ref:`bytecode metadata <metadata>`.
 If you do not want to specify a license or if the source code is
 not open-source, please use the special value ``UNLICENSED``.
 
+.. note::
+    There might be a confusion between ``UNLICENSED`` (no usage allowed, not present in SPDX license list) and ``UNLICENSE`` (grants all rights to everyone). Check out `this discussion<https://github.com/ethereum/solidity/issues/12251>`_.
+
 Supplying this comment of course does not free you from other
 obligations related to licensing like having to mention
 a specific license header in each source file or the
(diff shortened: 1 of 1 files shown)


docs/layout-of-source-files.rst
```diff
@@ -28,6 +28,9 @@ it does include the supplied string in the :ref:`bytecode metadata <metadata>`.
 If you do not want to specify a license or if the source code is
 not open-source, please use the special value ``UNLICENSED``.
 
+.. note::
+    There might be a confusion between ``UNLICENSED`` (no usage allowed, not present in SPDX license list) and ``UNLICENSE`` (grants all rights to everyone). Check out `this discussion<https://github.com/ethereum/solidity/issues/12251>`_.
```
Enxhi-io: What listen e is that for ? You see I’m been hacked and I have no clue what they have done and changed and deleted. Do you want to give me a hand and I’ll make it worth your while ?

docs/layout-of-source-files.rst
```diff
@@ -28,6 +28,9 @@ it does include the supplied string in the :ref:`bytecode metadata <metadata>`.
 If you do not want to specify a license or if the source code is
 not open-source, please use the special value ``UNLICENSED``.
 
+.. note::
```
chriseth, line 31: I would not make this stand out as a note, because then it becomes more prominent than the paragraph above. What about just saying that ``UNLICENSED`` is different than "the UNLICENSE" and that we follow the recommendation by npm (and link `https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license` instead of the github issue)

leonardoalt:
Closing in favor of https://github.com/ethereum/solidity/pull/12669 which has the comments' fix, keeping @caffeinum as the author.

PR #12402, opened 2021-12-13
Clarify FileReader interface.

chriseth:


libsolidity/interface/FileReader.h
```diff
@@ -61,18 +61,17 @@ class FileReader
 	void allowDirectory(boost::filesystem::path _path);
 	FileSystemPathSet const& allowedDirectories() const noexcept { return m_allowedDirectories; }
 
-	StringMap const& sourceCodes() const noexcept { return m_sourceCodes; }
-
-	/// Retrieves the source code for a given source unit name.
-	SourceCode const& sourceCode(SourceUnitName const& _sourceUnitName) const { return m_sourceCodes.at(_sourceUnitName); }
```
chriseth, line 67: This was actually not used at all.

libsolidity/interface/FileReader.h
```diff
@@ -61,18 +61,17 @@ class FileReader
...
 	/// Does not enforce @a allowedDirectories().
-	void setSources(StringMap _sources);
+	void setSourceUnits(StringMap _sources);
 
-	/// Adds the source code under a source unit name created by normalizing the file path.
+	/// Adds the source code under a source unit name created by normalizing the file path
+	/// or changes an existing source.
 	/// Does not enforce @a allowedDirectories().
-	void setSource(boost::filesystem::path const& _path, SourceCode _source);
+	void addOrUpdateFile(boost::filesystem::path const& _path, SourceCode _source);
```
cameel, line 74: How about this?
```suggestion
	void loadSourceFromPath(boost::filesystem::path const& _path, SourceCode _source);
```

libsolidity/interface/FileReader.h
```diff
@@ -61,18 +61,17 @@ class FileReader
 	void allowDirectory(boost::filesystem::path _path);
 	FileSystemPathSet const& allowedDirectories() const noexcept { return m_allowedDirectories; }
 
-	StringMap const& sourceCodes() const noexcept { return m_sourceCodes; }
-
-	/// Retrieves the source code for a given source unit name.
-	SourceCode const& sourceCode(SourceUnitName const& _sourceUnitName) const { return m_sourceCodes.at(_sourceUnitName); }
+	/// @returns all sources by their internal source unit names.
+	StringMap const& sourceUnits() const noexcept { return m_sourceCodes; }
```
cameel, line 65: I really like this name.

libsolidity/interface/FileReader.h
```diff
@@ -61,18 +61,17 @@ class FileReader
...
 
 	/// Resets all sources to the given map of source unit name to source codes.
 	/// Does not enforce @a allowedDirectories().
-	void setSources(StringMap _sources);
+	void setSourceUnits(StringMap _sources);
```
cameel, line 69: I think we should remove this method and add this instead:
```c++
void updateSourceUnit(std::string const& _sourceUnitName, SourceCode _source);
```

solc/CommandLineInterface.cpp
```diff
@@ -528,7 +528,7 @@ map<string, Json::Value> CommandLineInterface::parseAstFromInput()
 		}
 	}
 
-	m_fileReader.setSources(tmpSources);
+	m_fileReader.setSourceUnits(tmpSources);
```
cameel, line 531: This is the place I talking about on the call. It's the only reason we needed `setSources()` at all. Looks like it do need the ability to set by source name after all.

But I think that we should just have a special constructor for this case and just create a fresh FileReader instance. It should not be possible to manipulate by source unit name once `FileReader` already exists.

libsolidity/interface/FileReader.h
cameel: It used to be in the past, back when we were dumping CLI names straight into the VFS, without translation:
https://github.com/ethereum/solidity/blob/257672e56f5ba9d88e443100f95da879ad7b9dbb/solc/CommandLineInterface.cpp#L628-L636

Now we have `setStdin()` for that.

libsolidity/interface/FileReader.h
chriseth: This sounds like it would read from the filesystem.

libsolidity/interface/FileReader.h
chriseth: You have to decide, would you prefer `update` based on the file name or on the source unit name?

solc/CommandLineInterface.cpp
chriseth: If we allow setting source units by their internal names, I don't see why this function would not be allowed as well.

chriseth: The problem is that `m_compiler` contains a reference to the file reader callback.

libsolidity/interface/FileReader.h
cameel: So `addSourceFromPath()`. I mean, I'd emphasize that the path is being translated here into a source unit name. We could also go for `addSourceFromCLI()`

libsolidity/interface/FileReader.h
cameel: I think update should work in both cases. Adding a new entry should only work based on file name.

solc/CommandLineInterface.cpp
cameel: > The problem is that `m_compiler` contains a reference to the file reader callback.

Yes, but file reader callback does not have to use this function. It can add stuff directly to `m_sources`.

> If we allow setting source units by their internal names, I don't see why this function would not be allowed as well.

I think we should not allow setting by source unit name. Only updating.

The reason is that if you insert a new source unit name, you have to ensure that it's relative to base path. It's error prone. I think you should have to go through the CLI -> source unit name translation because it's safer that way.

libsolidity/interface/FileReader.h
chriseth: I will add it to the docs. I'm not sure we should emphasize the "CLI" aspect here. The only thing to consider here is that it is a path that is assumed to exist in the host filesystem.

libsolidity/interface/FileReader.h
chriseth: I won't do the refactor with the constructor as it currently really does not work.

solc/CommandLineInterface.cpp
chriseth: Sure, but these are two distict ways we use the file reader. Maybe we could actually turn these into two different file readers.

libsolidity/interface/FileReader.h
chriseth: Ah, it already is part of the docstring.

solc/CommandLineInterface.cpp
cameel: I really think we're constructing `FileReaders` to early, before we have all data for them. That's one thing I wanted to eventually refactor in the CLI (make it `optional` and initialize it once, only when I have all the data for it) and now there's the same thing with LSP where you're creating the adapter before you know the base path.

But anyway, I agree with https://github.com/ethereum/solidity/pull/12402#discussion_r767865777, that's a bit out of scope of this PR so we can do it some other time. For now it looks like `setSourceUnits()` unfortunately has to stay.

cameel (approved):

PR #12404, opened 2021-12-13
[SMTChecker] Fix soundness of struct storage/memory pointers

leonardoalt:
The bug here is that not enough knowledge was being erased for soundness when struct pointers were involved.

The newly added tests were all buggy, since the smtchecker was saying the assertions were all safe.

### DIFF
--- a/libsolidity/formal/SMTEncoder.cpp
+++ b/libsolidity/formal/SMTEncoder.cpp
@@ -32,6 +32,8 @@
 
 #include <liblangutil/CharStreamProvider.h>
 
+#include <libsolutil/Algorithms.h>
+
 #include <range/v3/view.hpp>
 
 #include <limits>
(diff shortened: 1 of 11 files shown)


libsolidity/formal/SMTEncoder.cpp
```diff
@@ -2371,28 +2373,25 @@ void SMTEncoder::resetReferences(Type const* _type)
...
+			foundSame = true;
+			bfs.abort();
 		}
-	}
-	return false;
+		if (auto const* mapType = dynamic_cast<MappingType const*>(_type))
+			_addChild(mapType->valueType());
+		else if (auto const* arrayType = dynamic_cast<ArrayType const*>(_type))
+			_addChild(arrayType->baseType());
+		else if (auto const* structType = dynamic_cast<StructType const*>(_type))
```
chriseth, line 2389: Is TupleType an option or do you only consider nameable types?

leonardoalt: Only consider reference types here

chriseth (approved):

PR #12408, opened 2021-12-15
Update circleci mac xcode version to 13.2

sbillig:
The fe github ci is failing to build solidity v0.8.10 (via solc-rust) on macos, using a newer version of xcode than is specified in the circleci config in this repo (xcode 12.x vs xcode 11.0).

https://github.com/ethereum/fe/runs/4529012071?check_suite_focus=true

The problem seems to be that clang 12.x (apple's fork at least) emits a warning here:
```
solidity/libyul/backends/evm/StackLayoutGenerator.cpp:250:19: error: loop variable '[slot, idealPosition]' is always a copy because the range of type 'ranges::zip_view<
[...snip...]
does not return a reference [-Werror,-Wrange-loop-analysis]

          for (auto const& [slot, idealPosition]: ranges::zip_view(_post, layout))
```

which makes the build fail due to the `-Werror` flag.

I'm starting this pr by just updating the version of xcode specified in the circleci config in order to reproduce the issue. If the build fails on circleci, I'm happy make the tiny code change to fix the warning. I'm also inclined to add an option to the cmake config to allow `-Werror` to be disabled (by downstream users like solc-rust) to avoid this sort of issue in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,7 @@ defaults:
 
   - base_osx: &base_osx
       macos:
-        xcode: "11.0.0"
+        xcode: "13.2.0"
       environment:
         TERM: xterm
         MAKEFLAGS: -j 5


cameel:
Thanks for the PR! If the compiler warns about it, we definitely want to fix it, especially if the warning is correct (and looks like in this case it is). Looks like CI passed here though. Maybe it's not just about the Xcode version? Is there anything else in your CI environment that's different? Maybe you're already on CircleCI's Big Sur preview or something?

Bumping Xcode version sounds fine to me.
@chriseth Do we want to keep the on the older version too to ensure backwards compatibility or are we fine with building only on one version?

> I'm also inclined to add an option to the cmake config to allow `-Werror` to be disabled (by downstream users like solc-rust) to avoid this sort of issue in the future.

We don't really mind you submitting PRs/issues about such problems - we want them fixed just as much as you do :) So it's more of a question of how often do they slip through and how much of a nuisance having to wait for us is for you. I think this is the first one I've seen so I'd prefer not to do that for now - better to try to keep our CI setups close enough that we can detect these problems before they hit you.

sbillig:
@cameel I changed the xcode version to 12.4 to match that of our ci (gh workflow's default mac environment), and it is now indeed failing. I had guessed that clang 13 would warn on the code as well, but that's apparently not the case.

(For posterity's sake: https://app.circleci.com/pipelines/github/ethereum/solidity/21129/workflows/a04b69ce-0af7-45c1-aabd-ad670b9d7195/jobs/928278 )

(I also temporarily removed the 'restore cache' step, as that was failing for some reason.)

I haven't been able to reproduce this on linux, with clang 12.0.1-8build1. (xcode 12.4 is allegedly "Apple clang version 12.0.0 (clang-1200.0.32.29)"). Perhaps this particular warning behavior was short lived in the clang codebase, and isn't worth worrying about? FWIW, removing the `&` in question seems fine.

>> I'm also inclined to add an option to the cmake config to allow -Werror to be disabled (by downstream users like solc-rust) to avoid this sort of issue in the future.

> We don't really mind you submitting PRs/issues about such problems - we want them fixed just as much as you do :) So it's more of a question of how often do they slip through and how much of a nuisance having to wait for us is for you. I think this is the first one I've seen so I'd prefer not to do that for now - better to try to keep our CI setups close enough that we can detect these problems before they hit you.

I'm happy to submit issues or prs, of course. The trouble is that there's no easy way for us to unbreak our CI builds when a new compiler version introduces a new warning in solc code as it existed at some git tag (v0.8.10 in this case).

What I was trying to accomplish was adding support for the basefee opcode to fe, but what I ended up doing was fiddling with cmake options in the solc-rust build script trying (and failing) to disable either this warning or -Werror, and rebuilding solc over and over again for several hours trying to figure out what's going on, and I'm no closer to my original (simple) goal.

For now, I'll just roll back to an earlier tag that doesn't have this exact line of code, or find a gh ci mac environment with a different clang version.

cameel:
Thanks for reproducing this. We fixed the warning in #12414.

We decided that we're going to try to stay on the latest version of Xcode in CI. In this PR let's bump it to 13.2 in that case. Also, feel free to ping us on Gitter/Matrix in case we fall behind and get out of sync with version used in your CI. I'll try to keep it updated, but to be honest, we usually bump these versions only when something actually requires the update, there's a security issue, etc.

> I'm happy to submit issues or prs, of course. The trouble is that there's no easy way for us to unbreak our CI builds when a new compiler version introduces a new warning in solc code as it existed at some git tag (v0.8.10 in this case).

Fair enough. I can see how it can be a problem. In that case we're fine with adding a flag to control that behavior.

sbillig:
> Thanks for reproducing this. We fixed the warning in #12414.

Great, thanks!

> We decided that we're going to try to stay on the latest version of Xcode in CI. In this PR let's bump it to 13.2 in that case. Also, feel free to ping us on Gitter/Matrix in case we fall behind and get out of sync with version used in your CI. I'll try to keep it updated, but to be honest, we usually bump these versions only when something actually requires the update, there's a security issue, etc.

:+1: It's now back to 13.2

> > I'm happy to submit issues or prs, of course. The trouble is that there's no easy way for us to unbreak our CI builds when a new compiler version introduces a new warning in solc code as it existed at some git tag (v0.8.10 in this case).
> 
> Fair enough. I can see how it can be a problem. In that case we're fine with adding a flag to control that behavior.

Ok, cool. I'll add the flag in a future pr. We'll keep -Werror enabled for routine ci builds, and only disable it if needed to unstick things (and we'll create an issue here).

cameel (approved):

PR #12412, opened 2021-12-15
Labels in issue templates

cameel:
https://github.com/ethereum/solidity-blog/pull/159#discussion_r769602005 by @christianparpart gave me and idea that we could automatically apply labels in our issue templates.

The PR also does some minor cleanup in templates:
- Replaces tabs with spaces
- Removes hard wrapping to make the instructions look more concise. Text in Github's issue editor is always soft wrapped. It also will never show up in the preview and even if it did, Github does not remove hard wrapping when rendering issue comments as would normally happen in markdown.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,7 @@
 ---
 name: Bug Report
 about: Bug reports about the Solidity Compiler.
+labels: ["bug :bug:"]
 ---
 
 <!--## Prerequisites
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,13 +9,12 @@ labels: ["bug :bug:"]
 - First, many thanks for taking part in the community. We really appreciate that.
 - We realize there is a lot of information requested here. We ask only that you do your best to provide as much information as possible so we can better help you.
 - Support questions are better asked in one of the following locations:
-	- [Solidity chat](https://gitter.im/ethereum/solidity)
-	- [Stack Overflow](https://ethereum.stackexchange.com/)
+    - [Solidity chat](https://gitter.im/ethereum/solidity)
+    - [Stack Overflow](https://ethereum.stackexchange.com/)
 - Ensure the issue isn't already reported.
 - The issue should be reproducible with the latest solidity version; however, this isn't a hard requirement and being reproducible with an older version is sufficient.
 
 *Delete the above section and the instructions in the sections below before submitting*
-
 -->
 
 ## Description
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,20 +1,20 @@
...
 ---
 
 <!--## Prerequisites
 
 - First, many thanks for taking part in the community. We really appreciate that.
 - We realize there is a lot of information requested here. We ask only that you do your best to provide as much information as possible so we can better help you.
 - Support questions are better asked in one of the following locations:
-	- [Solidity chat](https://gitter.im/ethereum/solidity)
-	- [Stack Overflow](https://ethereum.stackexchange.com/)
+    - [Solidity chat](https://gitter.im/ethereum/solidity)
```
christianparpart, line 12: Do we want to reference the linked Matrix chat here?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,20 +1,20 @@
 ---
 name: Bug Report
 about: Bug reports about the Solidity Compiler.
+labels: ["bug :bug:"]
```
chriseth, line 4: Currently (at least that's how I have handled it), the label served the purpose of "confirmed bug". What do we want to do about that? Just keep the label and use the project columns to track confirmed / not confirmed?

.github/ISSUE_TEMPLATE/bug_report.md
cameel: Personally, I always give people the gitter link because it just drops you directly on the channel without having to go through any nag screens about downloading the client.

Maybe we should list both?

.github/ISSUE_TEMPLATE/bug_report.md
cameel: Yeah, I had doubts about this myself because of that system. Maybe we should actually change it? I think it would be useful to have this stuff immediately categorized and a `confirmed` label would have more obvious meaning - the current system is a bit too implicit.

But I can remove it from the template if you prefer to stay with the current system.

chriseth: let's use the columns in the project to mark it as confirmed.

chriseth (approved):

.github/ISSUE_TEMPLATE/bug_report.md
chriseth: That is true, so let's stay with gitter as long as it still works.

.github/ISSUE_TEMPLATE/bug_report.md
cameel: ok, so not in `New issues` means that it's confirmed?

chriseth: yes

Issue #8969, opened 2020-05-18
Implement project board automation


chriseth:
Implement automation (maybe through github actions) for the following:
 - newly opened issues should be added to the "Solidity" project board under "new issues"
 - stale issues that are not in the icebox should get a "ping" comment

mijovic:
For adding newly opened issues to "new issues", I think there is already automation which moves all new issues in "Solidity" project to "new issues". We just need to find a way to move newly created issues to "Solidity" project

cameel:
Looks like github docs have some ready to use examples for these:
- [Moving assigned issues on project boards](https://docs.github.com/en/actions/managing-issues-and-pull-requests/moving-assigned-issues-on-project-boards) + [Adding labels to issues](https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues)
- [Closing inactive issues](https://docs.github.com/en/actions/managing-issues-and-pull-requests/closing-inactive-issues) + [Commenting on an issue when a label is added](https://docs.github.com/en/actions/managing-issues-and-pull-requests/commenting-on-an-issue-when-a-label-is-added)

Looks like they actually rely on third-party github actions.

Also one more idea for what we could do: assigning triage issues (and maybe PRs to review) to random team members. The assigned person has to either deal with it or give it to someone else, but has to do *something*.

cameel:
Another idea: adding PRs labeled `takeover` to the `Takeover column` on the Focus Board.

cameel:
@r0qs We talked with @ekpyron today about next steps here.
- One would be to finalize #13499 but do it only for PRs for now.
- Second, to implement the idea with adding new issues to the board. Since permissions are a problem we'll do it like this:
    - Rename `New Issues` column on the [old board](https://github.com/ethereum/solidity/projects/43) to `Triage` and start adding new issues there for now until we figure out if we can do it for the focus board.
    - Clear the `New Issues` column. As it is now it's not very useful.

cameel:
I cleared the column an renamed it (actually deleted it and created a new one).

Also, I have an idea for how we can work around the board not being able to handle solc-js issues. We could just add notes with links. Not great but better than not being able to manage them at all.

cameel:
As for new boards, looks like PAT seems to be the only way forward and we do want to move on to them if only because github is very likely to deprecate old ones at some point. So once devops creates a custom account for us, we could move on with that.

r0qs:
[...]

PR #12413, opened 2021-12-15
Links to issue templates that automatically apply the `Solidity` project

cameel:
Partially addresses #8969.

Apparently it's not possible to have github select the project automatically in issue templates but it is possible to do so through the `project` query parameter in new issue URLs. And the `URL`s can be customized via [`config.yml`](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser). This PR creates this config and assigns new issues by default to the `Solidity` project.

It also removes the `general.md` template, which seems unused.

Unfortunately I have no way to test this so if we want to try this, we'll have to just merge it and see what happens. If things go bad we can revert. **A few things to check after it's merged**:
1. Currently issues in the [template chooser](https://github.com/ethereum/solidity/issues/new/choose) have descriptions taken from the `about` property in the templates. You can specify `about` in `config.yml` too. I did not do it to avoid duplication and I hope github will just take it from the templates.
2. You can specify labels in the URLs via `labels` parameter. I did not specify it and I'm hoping that github will also take these from the issue templates (once we have them there: #12412).
3. The `Report a security vulnerability` entry on the list seems special. I did not specify it because docs don't mention it. We need to make sure it does not disappear.
4. I'm setting `blank_issues_enabled` to `false` so that users always have to use one of these templates. I'm not sure if this will just remove the blank issue link from template chooser or completely forbid creating blank ones. In case it forbids them we might want to actually start using the `general.md` template for this.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Bug Report
+    url: https://github.com/ethereum/solidity/issues/new?template=bug_report.md&projects=ethereum/solidity/43
+  - name: Documentation Issue
+    url: https://github.com/ethereum/solidity/issues/new?template=documentation_issue.md&projects=ethereum/solidity/43
+  - name: Feature Request
+    url: https://github.com/ethereum/solidity/issues/new?template=feature_request.md&projects=ethereum/solidity/43


christianparpart (approved):
nice :)

chriseth:
It did not work. I would also guess that externals cannot assign issues to projects, to be honest.

I think the "blank" did work, though.

cameel:
Too bad. I wonder if it might be because I did not include the `about` field? Because it's not that the new links do not work, the old links are just still there in the chooser. Github just ignored the ones in the config.

When go to the new link directly, I do see `Solidity` selected in the project drop down: https://github.com/ethereum/solidity/issues/new?template=bug_report.md&projects=ethereum/solidity/43. Unfortunately I can't check how it is for someone without permissions in the project - github requires login so I'd have to create a dummy account to test that.

Also, we have an answer to (4) - you can still sidestep the form and create a blank issue by going to https://github.com/ethereum/solidity/issues/new/

PR #12417, opened 2021-12-16
Added sameType check for fromType and toType in YulUtilFunctions.cpp and relevant tests in semanticTests

nishant-sachdeva:
Further solves #12057  

Improving upon #12372

### DIFF
--- a/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
+++ b/test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
@@ -4,12 +4,27 @@ contract C {
     function testFunction3() public {}
 
 
-    function() external [3] externalArray0;
-    function() external [3] externalArray1 = [
-        this.testFunction1,
-        this.testFunction2,
-        this.testFunction3
-    ];
+    function() external [] externalArray0;
+    function() external [] externalArray1;
+
+    function() internal [] internalArray0;
+    function() internal [] internalArray1;
+
+    constructor() {
+        externalArray0 = new function() external[] (3);
+        externalArray1 = [
+            this.testFunction1,
+            this.testFunction2,
+            this.testFunction3
+        ];
+
+        internalArray0 = new function() internal[] (3);
+        internalArray1 = [
+            testFunction1,
+            testFunction2,
+            testFunction3
+        ];
+    }
 
     function copyExternalStorageArrayOfFunctionType() external returns (bool) {
         assert(keccak256(abi.encode(externalArray0)) != keccak256(abi.encode(externalArray1)));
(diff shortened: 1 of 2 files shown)


test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
```diff
@@ -4,8 +4,8 @@ contract C {
     function testFunction3() public {}
 
 
-    function() external [3] externalArray0;
-    function() external [3] externalArray1 = [
+    function() external [] externalArray0 = new function() external[] (3);
+    function() external [] externalArray1 = [
         this.testFunction1,
         this.testFunction2,
         this.testFunction3
```
cameel: I'm actually really surprised that this works. I suggested changing it to a static array in the original PR because I thought we never allow assigning array literals to dynamic arrays but turns out that we do allow that for storage. Probably because we're not assigning a reference but just copying the content.

cameel (commented):
Looks ok, but can you also change this uf you're adjusting these tests: https://github.com/ethereum/solidity/pull/12372#discussion_r769964255

cameel:
You also need to adjust gas expectations.

test/libsolidity/semanticTests/array/copying/array_of_function_external_storage_to_storage_dynamic.sol
nishant-sachdeva: Would it be better if I add the assignment part into a constructor ?

cameel: No, it's fine. I suggested that in the original PR only because I thought it was not possible to do it like this. But I wasn't sure - I assumed that if I'm wrong you'll notice that the CI passes and stay with that version or at least comment something :)

nishant-sachdeva: Ohh, well, for precaution's sake, I already merged changes where I shifted the assignments to a Constructor. Hope that's fine :)

cameel: It's fine. I liked the constructor-less version better but it's not like it's a big difference. It can stay as is.

What about https://github.com/ethereum/solidity/pull/12372#discussion_r769964255? It's a tiny change but that's how we format conditions in `return` and is an easy detail to change.

nishant-sachdeva: Added the changes in the latest push

cameel (approved):
Should be fine as long as tests pass.

PR #12418, opened 2021-12-16
LSP for solcjs

chriseth:
Open questions:

**Interface**
- [ ] do we want to have a generic, input/output function that has a context (probably yes)?
- [ ] do we want to have a generic "set read callback" function?
- [ ] do we actually need to "start the lsp"?
- [ ] how to handle the fact that the file read callback has to be synchronous?

**Implementation**
- [ ] we need to redesign the FileRepository so that it can get a file read callback from outside. How should it handle things like base path and include paths? Not at all?

---

Some ideas / solutions from a call with @axic:
- we stay with `lsp_start` and `lsp_send_receive` for now.
- we should add two new kinds to the read callback:
  - `LookupPath`: gets a compiler-internal path and returns an absolute file url (or other url)
  - `ReadFileAt`: gets a url returned by `LookupPath` and returns its contents
 this mechanism allows the LSP server to store the internal / url mapping and thus report source locations correctly in all circumstances.
- the solcjs wrapper should have `lsp: { start(readCallback), abort() }`, where `start(readCallback)` returns a function that wraps `send_receive`.
- the include path mechanism has to be implemented by the lsp client plugin (this also has the benefit that the plugin can e.g. detect the framework the project uses and handle special paths properly)

What is still unsolved how to make this work nicely in a context where we cannot do synchronous file reads. The `LookupPath` callback would have to store the path and return an error. When the `send_receive` calls returns, we can read the file and once we have it, we somehow need to re-trigger the compiler analysis so that it asks for the file again. How to trigger the compiler is the main question.

### DIFF
--- a/libsolc/libsolc.cpp
+++ b/libsolc/libsolc.cpp
@@ -24,6 +24,8 @@
 #include <libsolc/libsolc.h>
 #include <libsolidity/interface/StandardCompiler.h>
 #include <libsolidity/interface/Version.h>
+#include <libsolidity/lsp/LanguageServer.h>
+#include <libsolidity/lsp/Transport.h>
 #include <libyul/YulString.h>
 
 #include <cstdlib>
@@ -112,6 +114,9 @@ string compile(string _input, CStyleReadFileCallback _readCallback, void* _readC
 	return compiler.compile(move(_input));
 }
 
+unique_ptr<lsp::MockTransport> languageServerTransport;
+unique_ptr<lsp::LanguageServer> languageServer;
+
 }
 
 extern "C"
@@ -130,6 +135,56 @@ extern char const* solidity_version() noexcept
 extern char* solidity_compile(char const* _input, CStyleReadFileCallback _readCallback, void* _readContext) noexcept
 {
 	return solidityAllocations.emplace_back(compile(_input, _readCallback, _readContext)).data();
+}
+
+extern int solidity_lsp_start(CStyleReadFileCallback /*TODO(pr) _readCallback*/, void* /*_readContext*/) noexcept
+{
+	if (languageServer || languageServerTransport)
+		return -1;
+
+	languageServerTransport = make_unique<lsp::MockTransport>();
+	languageServer = make_unique<lsp::LanguageServer>(*languageServerTransport);
+	return 0;
+}
+
+extern int solidity_lsp_send(char const* _input) noexcept
+{
+	if (!languageServer)
+		return -1;
+
+	if (languageServer->isRunning())
+		return -2;
+
+	solAssert(languageServerTransport, "");
+
+	std::string errors{};
+	Json::Value jsonMessage{};
+	if (!jsonParseStrict(_input, jsonMessage, &errors))
+		return -3;
+
+	languageServerTransport->appendInput(jsonMessage);
+	languageServer->runIteration();
+	return 0;
+}
+
+extern char const* solidity_try_receive() noexcept
+{
+	if (!languageServerTransport)
+		return "";
+
+	optional<Json::Value> messageJson = languageServerTransport->popOutput();
+	if (!messageJson)
+		return "";
+
+	return solidityAllocations.emplace_back(jsonPrettyPrint(messageJson.value())).data();
+}
+
+extern char const* solidity_lsp_send_receive(char const* _input) noexcept
+{
+	if (solidity_lsp_send(_input) < 0)
+		return "";
+
+	return solidity_try_receive();
 }
 
 extern char* solidity_alloc(size_t _size) noexcept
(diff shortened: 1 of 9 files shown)


chriseth:
@axic you wanted to comment here

chriseth:
One way forward is to have another set of callbacks that translate between client path and source unit name. The problem is that the client sends files to the server based on file paths (We could just work inside the compiler's virtual file system, but I don't think this is how LSP is meant to be used.). Now these file paths need to be translated to internal source unit names. For that, the server needs to know what the base path and the include directories are - which could be a config option.
But then, on import, the compiler calls the callback with an internal source unit name. This means the client needs to know about the mappings anyway.
Just a thought....

chriseth:
OK, now I see the reason: The read callback returns the contents of an included file, but if we "jump to definition", for example, we need to provide the location with the correct client path, and we can only guess this client path, so it has to be provided by the read callback.

libsolc/libsolc.h
```diff
@@ -90,6 +90,21 @@ void solidity_free(char* _data) SOLC_NOEXCEPT;
...
+
+/// Switch into LSP mode. Can be undone using solidity_reset().
+void solidity_lsp_start(CStyleReadFileCallback _readCallback, void* _readContext) SOLC_NOEXCEPT;
+/// Send one or more JSON-RPC messages to the LSP (including the HTTP headers), expecting a response.
+/// If the input is empty, just checks for a pending response.
+/// @returns JSON-RPC message (inculding HTTP headers), can be empty (or nullptr).
+/// If the result is not null, it has to be freed by the caller using solidity_free.
+///
+/// This can cause the callback provided in solidity_lsp to be invoked.
+/// Should only be called after having called solidity_lsp.
```
cameel: This refers to `solidity_lsp_start`, right?
```suggestion
/// This can cause the callback provided in solidity_lsp_start to be invoked.
/// Should only be called after having called solidity_lsp_start.
```

libsolc/libsolc.cpp
```diff
@@ -156,5 +178,7 @@ extern void solidity_reset() noexcept
 	// can be freed here.
 	yul::YulStringRepository::reset();
 	solidityAllocations.clear();
+	languageServer.reset();
+	languageServerTrasport.reset();
```
cameel:
```suggestion
	languageServerTransport.reset();
```

cameel:
> What is still unsolved how to make this work nicely in a context where we cannot do synchronous file reads. The `LookupPath` callback would have to store the path and return an error. When the `send_receive` calls returns, we can read the file and once we have it, we somehow need to re-trigger the compiler analysis so that it asks for the file again. How to trigger the compiler is the main question.

I don't fully understand this requirement. Why would you not be able to do a synchronous read? Is that because the file comes from the IDE and the IDE will queue the operation to process it only after its message to the LSP is handled or something?

Also, the LSP itself is synchronous and only runs in response to calls from the wrapper, right? This would mean that you cannot really do anything after `send_receive` returns.

chriseth:
You need to do synchronous reads for imports: If a file imports another, the compiler needs to request the contents from the IDE (the compiler has no filesystem access). The IDE is unlikely to answer this request synchronously. This means it can only answer "file not found". Even if we extend the return code with something like "has to be read later", there is no way in the protocol where the client can tell the server to "try again". The assumption is that the server is constantly compiling everything and reacting to every change.
So since everything is client-driven in the solcjs context (as you say), we might add a custom extension to tell the server to "try again". In the context of an event server, this would just be a "tick" that is sent to all the long-running processes.

cameel:
> The IDE is unlikely to answer this request synchronously.

So this is the part I find weird. Does this really happen in practice with the editors we tested LSP with or is it just a concern that it could happen?

> So since everything is client-driven in the solcjs context (as you say), we might add a custom extension to tell the server to "try again".

We could do it like this but it just feels that if we have to, we're trying to do it wrong. Do other implementations have to deal with something like this?

We have the file content already cached in `FileRepository`. How about we give the callback access to this?

chriseth:
It is a problem with solc-js, but in general, sync filesystem access is very bad practice in js. It might be that the extension is extracted into a worker, then it would not be as bad. If not, it can fully stall the IDE.

We don't know about any implementation that does not use a proper async rpc interface, that's the problem.

If the content is already in FileRepository, then it is of course not an issue. The problem is that a file can be both imported and "closed", and the IDE cannot send the content of closed files to the server, since the idea in LSP is that the server has access to the file system.

leonardoalt:
This draft seems to be stale. What's the status? Is it going to be done?

axic:
Here's an idea: https://github.com/cameel/yul-lsp/issues/16

PR #12419, opened 2021-12-16
Move `about` field from issue templates to `config.yml`

cameel:
Follow-up to #12413.

The config gets picked up but the URLs don't show up in the template chooser. Let's see if adding `about` fields in the config fixes that.


chriseth (approved):

cameel:
It worked - URLs were replaced - but also the link for reporting vulnerability is gone.

For the record, here's how template chooser looked before this change:
template-chooser-before

PR #12429, opened 2021-12-20
[SMTChecker] abi.encodeCall

leonardoalt:
Going with the small fix, the better fix would also solve the false positive on test `_2`.


test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple_1.sol
```diff
@@ -0,0 +1,19 @@
+contract C {
+	function callMeMaybe(uint a, uint b) external {}
```
axic, line 2: No.

leonardoalt: just copied from @Marenz 's original test

test/libsolidity/smtCheckerTests/out_of_bounds/array_1.sol
```diff
@@ -18,7 +18,7 @@ contract C {
 }
 // ====
 // SMTEngine: all
-// SMTIgnoreOS: macos
+// SMTIgnoreOS: macos,linux
```
hrkrshnn: So basically ignore always? Do we run this on windows?

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple_2.sol
```diff
@@ -0,0 +1,19 @@
+contract C {
+	function callMeMaybe(uint a, uint b) external {}
+
+	function abiEncodeSimple(uint x, uint y, uint z) public view {
+		require(x == y);
+		bytes memory b1 = abi.encodeCall(this.callMeMaybe, (x, z));
+		bytes memory b2 = abi.encodeCall(this.callMeMaybe, (y, z));
+		assert(b1.length == b2.length); // should hold but fails for now because function pointers like the above do not have a concrete value
```
hrkrshnn: Are you planning on adding a constraint about the length?

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple_1.sol
```diff
@@ -0,0 +1,19 @@
...
+
+	function abiEncodeSimple(uint x, uint y, uint z) public view {
+		require(x == y);
+		function (uint, uint) external f = this.callMeMaybe;
+		bytes memory b1 = abi.encodeCall(f, (x, z));
+		bytes memory b2 = abi.encodeCall(f, (y, z));
+		assert(b1.length == b2.length); // should hold
+
+		bytes memory b3 = abi.encodeCall(this.callMeMaybe, (3, z));
+		assert(b1.length == b3.length); // should fail
```
hrkrshnn: This is also a false positive, right?

hrkrshnn (commented):
There are merge conflicts.

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple_1.sol
leonardoalt: ah yea, right. It's true that b1 == b3 is not necessarily true, but the length should be the same

test/libsolidity/smtCheckerTests/abi/abi_encode_call_simple_2.sol
leonardoalt: I use the lengths in these tests because I can't just do `assert(b1 == b2)`, so I'm kind of testing that b1 and b2 are the same by testing that there lengths are the same. We don't compute the lenghts precisely, so they can be only equal here for sure if the whole thing is the same.

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1313,11 +1313,16 @@ void SMTEncoder::endVisit(Return const& _return)
 
 bool SMTEncoder::visit(MemberAccess const& _memberAccess)
 {
+	createExpr(_memberAccess);
+
 	auto const& accessType = _memberAccess.annotation().type;
 	if (accessType->category() == Type::Category::Function)
+	{
+		auto const* functionType = dynamic_cast<FunctionType const*>(accessType);
+		if (functionType->hasDeclaration())
```
stackenbotten: Coding style error

leonardoalt:
@ekpyron this is also ready for review

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1313,17 +1313,23 @@ void SMTEncoder::endVisit(Return const& _return)
 
 bool SMTEncoder::visit(MemberAccess const& _memberAccess)
 {
-	auto const& accessType = _memberAccess.annotation().type;
-	if (accessType->category() == Type::Category::Function)
-		return true;
-
 	createExpr(_memberAccess);
 
+	auto const& accessType = _memberAccess.annotation().type;
```
ekpyron:
```suggestion
	auto const* accessType = _memberAccess.annotation().type;
```

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -1313,17 +1313,23 @@ void SMTEncoder::endVisit(Return const& _return)
...
+
 	Expression const* memberExpr = innermostTuple(_memberAccess.expression());
 
 	auto const& exprType = memberExpr->annotation().type;
 	solAssert(exprType, "");
-	if (exprType->category() == Type::Category::Magic)
+
+	if (accessType->category() == Type::Category::Function)
+	{
+		auto const* functionType = dynamic_cast<FunctionType const*>(accessType);
```
ekpyron:
```suggestion
	if (auto const* functionType = dynamic_cast<FunctionType const*>(accessType))
	{
```

ekpyron:
What happens for this?
```
abstract contract D {
	function no(uint a, uint b) external virtual;
	function f(function(uint, uint) external) external pure virtual returns (uint256);
}

contract C {
	function abiEncodeSimple(D d1, D d2) public pure {
		assert(d1.f(d1.no) == d1.f(d2.no)); // should fail
	}
}
```

Or put differently: the only case in which the effects of a member-access expression of a function type is *solely* determined by the function by the function signature, is in the first argument of ``abi.encodeCall``, isn't it? Anywhere else the address in there can have an effect... So maybe it makes sense to do this more selectively?

leonardoalt:
> 

Output:

```Warning: CHC: Assertion violation happens here.
Counterexample:

d1 = 0
d2 = 0

Transaction trace:
C.constructor()
C.abiEncodeSimple(0, 0)
    d1.f(d1.no) -- untrusted external call
    d1.f(d2.no) -- untrusted external call
 --> a.sol:8:3:
  |
8 | 		assert(d1.f(d1.no) == d1.f(d2.no)); // should fail
  | 		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Not sure I get exactly what you mean? The function expression itself is represented by the signature, but if you *call* that function pointer, its value will be nondeterministic if we can't infer what the function actually is.

ekpyron (commented):
Code looks good (and more future-proof than before ;-)) now! But tests seem to need fixing.

ekpyron:
This should not affect any tests that don't involve ``abi.encodeCall``... did you try reverting all those test changes after changing the PR to only do this specifically for ``abi.encodeCall`` instead of for all member accesses of function type?

ekpyron (approved):

PR #12430, opened 2021-12-20
LSP goto definition

christianparpart:
Implements `textDocument/definition` and adds test for it.

- [x] type symbol -> type definition (contract, enum type, enum value type, ...)
- [x] variable name -> var decl
- [x] return variable name `x` -> jump to `x` in `returns (uint x)`
- [x] `import "file.sol"` -> jump to top of imported file "file.sol"
- [x] `import {x} from "file.sol"`, hover `x` -> jump to `x` in "file.sol"
- [x] `import {x as y} from "file.sol"`, having `y` hovered in file -> jump `x` in "file.sol"
- [x] function call name -> jump to function definition
- [x] function parameter name `x` to jump to function parameter def `x`
- [x] "Can you check how virtual lookup works in this case? Is it at least relative to the location of the expression or will it always go to the base?"
- ...?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ Compiler Features:
 * JSON-AST: Added selector field for errors and events.
  * Peephole Optimizer: Optimize comparisons in front of conditional jumps and conditional jumps across a single unconditional jump.
  * Yul Optimizer: Remove ``sstore`` and ``mstore`` operations that are never read from.
+ * LSP: Implements goto-definition.
 
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
--- a/libsolidity/ast/ASTUtils.cpp
+++ b/libsolidity/ast/ASTUtils.cpp
@@ -18,11 +18,34 @@
 
 #include <libsolidity/ast/AST.h>
 #include <libsolidity/ast/ASTUtils.h>
+#include <libsolidity/ast/ASTVisitor.h>
 
 #include <libsolutil/Algorithms.h>
 
 namespace solidity::frontend
 {
+
+ASTNode const* locateInnermostASTNode(int _offsetInFile, SourceUnit const& _sourceUnit)
+{
+	ASTNode const* innermostMatch = nullptr;
+	auto locator = SimpleASTVisitor(
+		[&](ASTNode const& _node) -> bool
+		{
+			// In the AST parent location always covers the whole child location.
+			// The parent is visited first so to get the innermost node we simply
+			// take the last one that still contains the offset.
+
+			if (!_node.location().containsOffset(_offsetInFile))
+				return false;
+
+			innermostMatch = &_node;
+			return true;
+		},
+		[](ASTNode const&) {}
+	);
+	_sourceUnit.accept(locator);
+	return innermostMatch;
+}
 
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl)
 {
--- a/libsolidity/ast/ASTUtils.h
+++ b/libsolidity/ast/ASTUtils.h
@@ -21,9 +21,11 @@
 namespace solidity::frontend
 {
 
-class VariableDeclaration;
+class ASTNode;
 class Declaration;
 class Expression;
+class SourceUnit;
+class VariableDeclaration;
 
 /// Find the topmost referenced constant variable declaration when the given variable
 /// declaration value is an identifier. Works only for constant variable declarations.
@@ -32,5 +34,8 @@ VariableDeclaration const* rootConstVariableDeclaration(VariableDeclaration cons
 
 /// Returns true if the constant variable declaration is recursive.
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
+
+/// Returns the innermost AST node that covers the given location or nullptr if not found.
+ASTNode const* locateInnermostASTNode(int _offsetInFile, SourceUnit const& _sourceUnit);
 
 }
--- a/libsolidity/lsp/GotoDefinition.cpp
+++ b/libsolidity/lsp/GotoDefinition.cpp
@@ -1,0 +1,66 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+#include <libsolidity/lsp/GotoDefinition.h>
+#include <libsolidity/lsp/Transport.h> // for RequestError
+#include <libsolidity/lsp/Utils.h>
+#include <libsolidity/ast/AST.h>
+#include <libsolidity/ast/ASTUtils.h>
+
+#include <fmt/format.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+using namespace solidity::frontend;
+using namespace solidity::langutil;
+using namespace solidity::lsp;
+using namespace std;
+
+void GotoDefinition::operator()(MessageID _id, Json::Value const& _args)
+{
+	auto const [sourceUnitName, lineColumn] = extractSourceUnitNameAndLineColumn(_args);
+
+	ASTNode const* sourceNode = m_server.astNodeAtSourceLocation(sourceUnitName, lineColumn);
+
+	vector<SourceLocation> locations;
+	if (auto const* expression = dynamic_cast<Expression const*>(sourceNode))
+	{
+		// Handles all expressions that can have one or more declaration annotation.
+		if (auto const* declaration = referencedDeclaration(expression))
+			if (auto location = declarationLocation(declaration))
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* identifierPath = dynamic_cast<IdentifierPath const*>(sourceNode))
+	{
+		if (auto const* declaration = identifierPath->annotation().referencedDeclaration)
+			if (auto location = declarationLocation(declaration))
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* importDirective = dynamic_cast<ImportDirective const*>(sourceNode))
+	{
+		auto const& path = *importDirective->annotation().absolutePath;
+		if (fileRepository().sourceUnits().count(path))
+			locations.emplace_back(SourceLocation{0, 0, make_shared<string const>(path)});
+	}
+
+	Json::Value reply = Json::arrayValue;
+	for (SourceLocation const& location: locations)
+		reply.append(toJson(location));
+	client().reply(_id, reply);
+}
--- a/libsolidity/lsp/GotoDefinition.h
+++ b/libsolidity/lsp/GotoDefinition.h
@@ -1,0 +1,31 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+#include <libsolidity/lsp/HandlerBase.h>
+
+namespace solidity::lsp
+{
+
+class GotoDefinition: public HandlerBase
+{
+public:
+	explicit GotoDefinition(LanguageServer& _server): HandlerBase(_server) {}
+
+	void operator()(MessageID, Json::Value const&);
+};
+
+}
--- a/libsolidity/lsp/HandlerBase.cpp
+++ b/libsolidity/lsp/HandlerBase.cpp
@@ -1,0 +1,63 @@
+#include <libsolidity/lsp/HandlerBase.h>
+#include <libsolidity/lsp/LanguageServer.h>
+#include <libsolidity/lsp/Utils.h>
+#include <libsolidity/ast/AST.h>
+
+#include <liblangutil/Exceptions.h>
+
+using namespace std;
+
+namespace solidity::lsp
+{
+
+using namespace langutil;
+
+Json::Value HandlerBase::toRange(SourceLocation const& _location) const
+{
+	if (!_location.hasText())
+		return toJsonRange({}, {});
+
+	solAssert(_location.sourceName, "");
+	langutil::CharStream const& stream = charStreamProvider().charStream(*_location.sourceName);
+	LineColumn start = stream.translatePositionToLineColumn(_location.start);
+	LineColumn end = stream.translatePositionToLineColumn(_location.end);
+	return toJsonRange(start, end);
+}
+
+Json::Value HandlerBase::toJson(SourceLocation const& _location) const
+{
+	solAssert(_location.sourceName);
+	Json::Value item = Json::objectValue;
+	item["uri"] = fileRepository().sourceUnitNameToClientPath(*_location.sourceName);
+	item["range"] = toRange(_location);
+	return item;
+}
+
+optional<SourceLocation> HandlerBase::parsePosition(string const& _sourceUnitName, Json::Value const& _position) const
+{
+	if (!fileRepository().sourceUnits().count(_sourceUnitName))
+		return nullopt;
+
+	if (optional<LineColumn> lineColumn = parseLineColumn(_position))
+		if (optional<int> const offset = CharStream::translateLineColumnToPosition(
+			fileRepository().sourceUnits().at(_sourceUnitName),
+			*lineColumn
+		))
+			return SourceLocation{*offset, *offset, make_shared<string>(_sourceUnitName)};
+	return nullopt;
+}
+
+optional<SourceLocation> HandlerBase::parseRange(string const& _sourceUnitName, Json::Value const& _range) const
+{
+	if (!_range.isObject())
+		return nullopt;
+	optional<SourceLocation> start = parsePosition(_sourceUnitName, _range["start"]);
+	optional<SourceLocation> end = parsePosition(_sourceUnitName, _range["end"]);
+	if (!start || !end)
+		return nullopt;
+	solAssert(*start->sourceName == *end->sourceName);
+	start->end = end->end;
+	return start;
+}
+
+}
--- a/libsolidity/lsp/HandlerBase.cpp
+++ b/libsolidity/lsp/HandlerBase.cpp
@@ -1,3 +1,22 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+#include <libsolutil/Exceptions.h>
+
 #include <libsolidity/lsp/HandlerBase.h>
 #include <libsolidity/lsp/LanguageServer.h>
 #include <libsolidity/lsp/Utils.h>
@@ -5,12 +24,12 @@
 
 #include <liblangutil/Exceptions.h>
 
+#include <fmt/format.h>
+
+using namespace solidity::langutil;
+using namespace solidity::lsp;
+using namespace solidity::util;
 using namespace std;
-
-namespace solidity::lsp
-{
-
-using namespace langutil;
 
 Json::Value HandlerBase::toRange(SourceLocation const& _location) const
 {
@@ -33,31 +52,27 @@ Json::Value HandlerBase::toJson(SourceLocation const& _location) const
 	return item;
 }
 
-optional<SourceLocation> HandlerBase::parsePosition(string const& _sourceUnitName, Json::Value const& _position) const
+pair<string, LineColumn> HandlerBase::extractSourceUnitNameAndLineColumn(Json::Value const& _args) const
 {
-	if (!fileRepository().sourceUnits().count(_sourceUnitName))
-		return nullopt;
+	string const uri = _args["textDocument"]["uri"].asString();
+	string const sourceUnitName = fileRepository().clientPathToSourceUnitName(uri);
+	if (!fileRepository().sourceUnits().count(sourceUnitName))
+		BOOST_THROW_EXCEPTION(
+			RequestError(ErrorCode::RequestFailed) <<
+			errinfo_comment("Unknown file: " + uri)
+		);
 
-	if (optional<LineColumn> lineColumn = parseLineColumn(_position))
-		if (optional<int> const offset = CharStream::translateLineColumnToPosition(
-			fileRepository().sourceUnits().at(_sourceUnitName),
-			*lineColumn
-		))
-			return SourceLocation{*offset, *offset, make_shared<string>(_sourceUnitName)};
-	return nullopt;
-}
+	auto const lineColumn = parseLineColumn(_args["position"]);
+	if (!lineColumn)
+		BOOST_THROW_EXCEPTION(
+			RequestError(ErrorCode::RequestFailed) <<
+			errinfo_comment(fmt::format(
+				"Unknown position {line}:{column} in file: {file}",
+				fmt::arg("line", lineColumn.value().line),
+				fmt::arg("column", lineColumn.value().column),
+				fmt::arg("file", sourceUnitName)
+			))
+		);
 
-optional<SourceLocation> HandlerBase::parseRange(string const& _sourceUnitName, Json::Value const& _range) const
-{
-	if (!_range.isObject())
-		return nullopt;
-	optional<SourceLocation> start = parsePosition(_sourceUnitName, _range["start"]);
-	optional<SourceLocation> end = parsePosition(_sourceUnitName, _range["end"]);
-	if (!start || !end)
-		return nullopt;
-	solAssert(*start->sourceName == *end->sourceName);
-	start->end = end->end;
-	return start;
-}
-
+	return {sourceUnitName, *lineColumn};
 }
--- a/libsolidity/lsp/HandlerBase.h
+++ b/libsolidity/lsp/HandlerBase.h
@@ -1,3 +1,20 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
 #pragma once
 
 #include <libsolidity/lsp/FileRepository.h>
@@ -24,22 +41,15 @@ public:
 	Json::Value toRange(langutil::SourceLocation const& _location) const;
 	Json::Value toJson(langutil::SourceLocation const& _location) const;
 
-	std::optional<langutil::SourceLocation> parsePosition(
-		std::string const& _sourceUnitName,
-		Json::Value const& _position
-	) const;
+	/// @returns source unit name and the line column position as extracted
+	/// from the JSON-RPC parameters.
+	std::pair<std::string, langutil::LineColumn> extractSourceUnitNameAndLineColumn(Json::Value const& _params) const;
 
-	/// @returns the source location given a source unit name and an LSP Range object,
-	/// or nullopt on failure.
-	std::optional<langutil::SourceLocation> parseRange(
-		std::string const& _sourceUnitName,
-		Json::Value const& _range
-	) const;
+	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.charStreamProvider(); }
+	FileRepository const& fileRepository() const noexcept { return m_server.fileRepository(); }
+	Transport& client() const noexcept { return m_server.client(); }
 
-	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.charStreamProvider(); };
-	FileRepository const& fileRepository() const noexcept { return m_server.fileRepository(); };
-	Transport& client() const noexcept { return m_server.client(); };
-
+protected:
 	LanguageServer& m_server;
 };
 
--- a/libsolidity/lsp/HandlerBase.h
+++ b/libsolidity/lsp/HandlerBase.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include <libsolidity/lsp/FileRepository.h>
+#include <libsolidity/lsp/LanguageServer.h>
+
+#include <liblangutil/SourceLocation.h>
+#include <liblangutil/CharStreamProvider.h>
+
+#include <optional>
+
+namespace solidity::lsp
+{
+
+class Transport;
+
+/**
+ * Helper base class for implementing handlers.
+ */
+class HandlerBase
+{
+public:
+	explicit HandlerBase(LanguageServer& _server): m_server{_server} {}
+
+	Json::Value toRange(langutil::SourceLocation const& _location) const;
+	Json::Value toJson(langutil::SourceLocation const& _location) const;
+
+	std::optional<langutil::SourceLocation> parsePosition(
+		std::string const& _sourceUnitName,
+		Json::Value const& _position
+	) const;
+
+	/// @returns the source location given a source unit name and an LSP Range object,
+	/// or nullopt on failure.
+	std::optional<langutil::SourceLocation> parseRange(
+		std::string const& _sourceUnitName,
+		Json::Value const& _range
+	) const;
+
+	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.charStreamProvider(); };
+	FileRepository const& fileRepository() const noexcept { return m_server.fileRepository(); };
+	Transport& client() const noexcept { return m_server.client(); };
+
+	LanguageServer& m_server;
+};
+
+}
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -21,6 +21,9 @@
 #include <libsolidity/interface/ReadFile.h>
 #include <libsolidity/interface/StandardCompiler.h>
 #include <libsolidity/lsp/LanguageServer.h>
+#include <libsolidity/lsp/HandlerBase.h>
+#include <libsolidity/lsp/Utils.h>
+
 
 #include <liblangutil/SourceReferenceExtractor.h>
 #include <liblangutil/CharStream.h>
@@ -47,31 +50,6 @@ using namespace solidity::frontend;
 
 namespace
 {
-
-Json::Value toJson(LineColumn _pos)
-{
-	Json::Value json = Json::objectValue;
-	json["line"] = max(_pos.line, 0);
-	json["character"] = max(_pos.column, 0);
-
-	return json;
-}
-
-Json::Value toJsonRange(LineColumn const& _start, LineColumn const& _end)
-{
-	Json::Value json;
-	json["start"] = toJson(_start);
-	json["end"] = toJson(_end);
-	return json;
-}
-
-optional<LineColumn> parseLineColumn(Json::Value const& _lineColumn)
-{
-	if (_lineColumn.isObject() && _lineColumn["line"].isInt() && _lineColumn["character"].isInt())
-		return LineColumn{_lineColumn["line"].asInt(), _lineColumn["character"].asInt()};
-	else
-		return nullopt;
-}
 
 int toDiagnosticSeverity(Error::Type _errorType)
 {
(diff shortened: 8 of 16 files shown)


libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+
+	optional<int> sourcePos = m_compilerStack.charStream(_sourceUnitName)
+		.translateLineColumnToPosition(_filePos);
+	if (!sourcePos.has_value())
+		return nullptr;
+
+	return locateInnermostASTNode(*sourcePos, m_compilerStack.ast(_sourceUnitName));
+}
+
+void LanguageServer::handleGotoDefinition(MessageID _id, Json::Value const& _args)
```
chriseth: I think it would be better to extract some of the functionality to their own files - the original PR contained a gigantic LanguageServer.cpp file

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+{
+	string const uri = _args["textDocument"]["uri"].asString();
+	string const sourceUnitName = m_fileRepository.clientPathToSourceUnitName(uri);
+	if (!m_fileRepository.sourceUnits().count(sourceUnitName))
+	{
+		m_client.error(_id, ErrorCode::RequestFailed, "Unknown file: " + uri);
```
chriseth: Kamil suggested to refactor this to use exceptions for error handling (at least for input validation) - do you think this would be a good idea?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+			)
+		);
+		return;
+	}
+
+	ASTNode const* sourceNode = requestASTNode(sourceUnitName, lineColumn.value());
+	vector<SourceLocation> locations;
+	if (auto const* identifier = dynamic_cast<Identifier const*>(sourceNode))
+	{
+		for (auto const* declaration: allAnnotatedDeclarations(identifier))
```
chriseth: Why do you return the candidates if we have resolved them to a single declaration?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+				fmt::arg("column", lineColumn.value().column),
+				fmt::arg("file", sourceUnitName)
+			)
+		);
+		return;
+	}
+
+	ASTNode const* sourceNode = requestASTNode(sourceUnitName, lineColumn.value());
+	vector<SourceLocation> locations;
+	if (auto const* identifier = dynamic_cast<Identifier const*>(sourceNode))
```
chriseth: If you first check if it is an Expression you can group Identifier, IdentifierPath and MemberAccess, work with Declarations instead of SourceLocations and reduce quite some code duplicaiton.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+	}
+	else if (auto const* memberAccess = dynamic_cast<MemberAccess const*>(sourceNode))
+	{
+		auto const location = declarationPosition(memberAccess->annotation().referencedDeclaration);
+		if (location.has_value())
+			locations.emplace_back(location.value());
+	}
+	else if (auto const* importDirective = dynamic_cast<ImportDirective const*>(sourceNode))
+	{
+		auto const& path = *importDirective->annotation().absolutePath;
```
chriseth: Can you also support "go to definition" when clicking on the imported item (i.e. `import x from "bla"` - you click on 'x' and jump to that one directly instead of just the file)

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -400,3 +426,89 @@ void LanguageServer::handleTextDocumentDidClose(MessageID _id, Json::Value const
...
+			locations.emplace_back(SourceLocation{0, 0, make_shared<string const>(path)});
+	}
+	else if (auto const* declaration = dynamic_cast<Declaration const*>(sourceNode))
+	{
+		if (auto location = declarationPosition(declaration); location.has_value())
+			locations.emplace_back(move(location.value()));
+	}
+	else if (sourceNode)
+	{
+		m_client.error(_id, ErrorCode::InternalError, fmt::format("Could not infer def of {}", typeid(*sourceNode).name()));
```
chriseth: Shouldn't we just return an empty array in this case? `null` also seems to be an option.

libsolidity/lsp/LanguageServer.cpp
chriseth: Can you also support TypeName? We could pull the function `declaration` that is defined by some sub-classes of Type to the top level class and use this function to jump to.

libsolidity/lsp/LanguageServer.cpp
christianparpart: @chriseth yes. But I want to log that case somehow, and we do not have a logging facility anymore. I wonder what might be the best that you'd be willing to merge. ;-)

chriseth: `ofstream("/tmp/log.txt", ofstream::app) << "My message." << endl;`

libsolidity/lsp/LanguageServer.cpp
christianparpart: @chriseth I am not against using exceptions, but IMHO explicit error handling is more readable and maintainable.

I personally prefer to use explicit error handling for everything that is recoverable and exceptions for unrecoverable errors.

So here in this case, if we want to stick with exceptions I'd do it by introducing a new exception type `lsp::Error` that accepts a `lsp::ErrorCode` and a text string.

But then, what if you want to continue the code path for other logic so throwing would not be good? (e.g. `checkServerInitialized()`). 
EDIT: I found two (not severe) bugs my looking for cases. Interesting.

stackenbotten:
There was an error when running `chk_coding_style` for commit `7d7440df2cde49cb8919f59133a10ebbf82d5a45`:
```
Coding style error:
libsolidity/lsp/Transport.h:64:    const char* what() const noexcept override { return m_message.c_str(); }
libsolidity/lsp/Transport.h:64:    const char* what() const noexcept override { return m_message.c_str(); }

```

libsolidity/lsp/LanguageServer.cpp
christianparpart: It's iterating over reference declaration (first) and then all candidate declarations. Or was that in case of identifiers an impossible case? 

EDIT: If I remember correctly, your point last time was that this also then lists all overloads in case there are any.

libsolidity/lsp/LanguageServer.cpp
christianparpart: This is however not being sent to the editor client. I will rewrite this line however.

stackenbotten:
There was an error when running `chk_coding_style` for commit `e716c48899c4f8b8d40808f833fbd6722ede0c7f`:
```
Coding style error:
libsolidity/lsp/Transport.h:64:    char const* what() const noexcept override { return m_message.c_str(); }

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `6706f73f200d488d1a474bda9934af8a4ed016fc`:
```
Coding style error:
libsolidity/lsp/Transport.h:64:    char const* what() const noexcept override { return m_message.c_str(); }

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `160ba482fb4d08d11ef6de203b49f631b9e6f35c`:
```
Coding style error:
libsolidity/lsp/Transport.h:64:    char const* what() const noexcept override { return m_message.c_str(); }

```

libsolidity/lsp/LanguageServer.cpp
chriseth: I agree with you, and I think most errors are unrecoverable. For now, all such errors should lead to bugs which should be reported. We can think about recovering some errors later, but for now to speed up development, I think we should just send a protocol error for most of them. I'm actually not sure if you can have an error response and a non-error response for the same request.

libsolidity/lsp/Transport.h
```diff
@@ -45,6 +47,28 @@ enum class ErrorCode
...
+		m_id{_id},
+		m_code{_code},
+		m_message{_message}
+	{
+	}
+
+	MessageID id() const noexcept { return m_id; }
+	ErrorCode code() const noexcept { return m_code; }
+	std::string const& message() const noexcept { return m_message; }
+    char const* what() const noexcept override { return m_message.c_str(); }
```
chriseth: Do you actually need to implement this? Don't we have a clever base class?

chriseth: Also please do this in a separate pr

libsolidity/ast/ASTUtils.cpp
```diff
@@ -18,12 +18,47 @@
...
 namespace solidity::frontend
 {
 
+namespace
+{
+
+class ASTNodeLocator: public ASTConstVisitor
+{
+public:
+	explicit ASTNodeLocator(int _pos): m_offsetInFile{_pos}, m_innermostMatch{nullptr} {}
```
chriseth: `pos` is way too generic for this variable name. What about `_sourceOffset`?

libsolidity/ast/ASTUtils.cpp
```diff
@@ -18,12 +18,47 @@
...
+		// The parent is visited first so to get the innermost node we simply
+		// take the last one that still contains the offset.
+
+		if (!_node.location().containsOffset(m_offsetInFile))
+			return false;
+
+		m_innermostMatch = &_node;
+		return true;
+	}
+
```
chriseth: private

chriseth: I know it has to be read by `locateInnermostASTNode`, but I think mutable members should only be public if it is OK to modify them at any time.

libsolidity/lsp/LanguageServer.h
```diff
@@ -107,4 +109,22 @@ class LanguageServer
 	Json::Value m_settingsObject;
 };
 
+// Helper base class for implementing handlers.
```
chriseth: Please use our style

libsolidity/lsp/LanguageServer.h
```diff
@@ -107,4 +109,22 @@ class LanguageServer
...
+class Handler
+{
+public:
+	explicit Handler(LanguageServer& _server):
+		m_server{_server},
+		m_charStreamProvider{_server.charStreamProvider()},
```
chriseth: If w estore the server, why do we need all these const refs?

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,54 @@
...
+std::optional<langutil::LineColumn> parseLineColumn(Json::Value const& _lineColumn);
+Json::Value toJson(langutil::LineColumn _pos);
+Json::Value toJsonRange(langutil::LineColumn const& _start, langutil::LineColumn const& _end);
+
+Json::Value toRange(
+	langutil::CharStreamProvider const& _charStreamProvider,
+	langutil::SourceLocation const& _location
+);
+
+Json::Value toJson(
```
chriseth: I think it could be very cumbersome to pass `charStreamProvider` along all the way. Can't this be a nice util function inside the Handler base class?

libsolidity/lsp/Transport.h
christianparpart: Done in #12482.

libsolidity/lsp/LanguageServer.h
christianparpart: For quicker access. I can drop them.

libsolidity/lsp/LanguageServer.h
christianparpart: Note to myself: I think you mean to name it `HandlerBase`.

libsolidity/lsp/LanguageServer.cpp
christianparpart: all done in the top mentioned base-PR.

libsolidity/lsp/LanguageServer.cpp
christianparpart: Maybe I misunderstood that, but `IdentifierPath` is not a base of `Expression`, but I tried to group the others however in front.

christianparpart: @chriseth I think this is the only comment left that I'm not sure I really understand. So jumping from any type symbol to their type definition is already handled. What case am I missing?

libsolidity/ast/AST_accept.h
```diff
@@ -58,13 +59,19 @@ void PragmaDirective::accept(ASTConstVisitor& _visitor) const
 
 void ImportDirective::accept(ASTVisitor& _visitor)
 {
-	_visitor.visit(*this);
+	if (_visitor.visit(*this))
+		for (SymbolAlias const& symbolAlias: symbolAliases())
+			if (symbolAlias.symbol)
+				symbolAlias.symbol->accept(_visitor);
```
chriseth: I think this is a correct change, but we have to cross-check how this impacts the other code - we essentially have to take a look at all `visit(Identifier` and `endVisit(Identifier` functions to see that this is fine.

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <libsolidity/ast/AST.h>
+#include <libsolidity/ast/ASTUtils.h>
+
+#include <fmt/format.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+using std::make_shared;
```
chriseth: Did we even run into problems recently where we had a name clash? I really don't like having to modify the `using` statements here whenever the code below changes. If all `move`s are gone, do you really think to remember removing this statement below?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,79 @@
...
+using std::move;
+using std::string;
+using std::vector;
+
+using namespace solidity::frontend;
+using namespace solidity::langutil;
+
+namespace solidity::lsp
```
chriseth: We usually only put the declarations inside the namespace and use `using` for the definitions.

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,79 @@
...
+			RequestError(ErrorCode::RequestFailed) <<
+			errinfo_comment(fmt::format(
+				"Unknown position {line}:{column} in file: {file}",
+				fmt::arg("line", lineColumn.value().line),
+				fmt::arg("column", lineColumn.value().column),
+				fmt::arg("file", sourceUnitName)
+			))
+		);
+
+	ASTNode const* sourceNode = m_server.requestASTNode(sourceUnitName, lineColumn.value());
```
chriseth: Shouldn't all this go to a helper in `Handler`?

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+vector<Declaration const*> allAnnotatedDeclarations(Expression const* _expression)
+{
+	vector<Declaration const*> output;
+
+	if (auto const* identifier = dynamic_cast<Identifier const*>(_expression))
+	{
+		output.push_back(identifier->annotation().referencedDeclaration);
+		output += identifier->annotation().candidateDeclarations;
```
chriseth: I think it is wrong and dangerous to add the candidates here.

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+	json["start"] = toJson(_start);
+	json["end"] = toJson(_end);
+	return json;
+}
+
+vector<Declaration const*> allAnnotatedDeclarations(Expression const* _expression)
+{
+	vector<Declaration const*> output;
+
+	if (auto const* identifier = dynamic_cast<Identifier const*>(_expression))
```
chriseth, line 62: Can you check how virtual lookup works in this case? Is it at least relative to the location of the expression or will it always go to the base?

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+	}
+	else if (auto const* memberAccess = dynamic_cast<MemberAccess const*>(_expression))
+	{
+		output.push_back(memberAccess->annotation().referencedDeclaration);
+	}
+
+	return output;
+}
+
+optional<SourceLocation> declarationPosition(Declaration const* _declaration)
```
chriseth: Why not `declarationLocation`?

chriseth: Also why do you return the name location? Is this how it's supposed to work?

chriseth:
General question I only asked in the meeting today: How de we handle virtual function calls? Is it fine to just jump to the definition that the call would resolve to from the current context?

libsolidity/ast/AST_accept.h
christianparpart: Resolving, as this is done in #12494.

libsolidity/lsp/GotoDefinition.cpp
christianparpart: good idea.

libsolidity/lsp/Utils.cpp
christianparpart: i've renamed the function.

With regards to `nameLocation`. If you want to goto a definition of a symbol, the editor/IDE will highlight the jumped-to area. If the whole location (i.e. with function body) is returned here, it'll all be highlighted. Which would be okay as it's only highlighted for a moment (in my editor), but the cursor is placed at the beginning, which would be the `f` of `function hello() ....` for example instead of the `h` of that `function hello() ...f`. Looking at other LSPs, it seems common to jump to the name of the definition, that's why we jump to the name location - if present - but fall back to the bigger-range location otherwise.

libsolidity/lsp/Utils.cpp
christianparpart: IIUC, I've been including a test case for that, which is exactly like clangd jumps. If you've meant something else, we can talk about ASAP then.

Assawal:
Sent from my Huawei phone-------- Original message --------From: Christian Parpart ***@***.***>Date: Mon, 31 Jan 2022, 16:07To: ethereum/solidity ***@***.***>Cc: Subscribed ***@***.***>Subject: Re: [ethereum/solidity] LSP goto definition (PR #12430)
@christianparpart commented on this pull request.



In libsolidity/lsp/Utils.cpp:
> +	return json;
+}
+
+Json::Value toJsonRange(LineColumn const& _start, LineColumn const& _end)
+{
+	Json::Value json;
+	json["start"] = toJson(_start);
+	json["end"] = toJson(_end);
+	return json;
+}
+
+vector<Declaration const*> allAnnotatedDeclarations(Expression const* _expression)
+{
+	vector<Declaration const*> output;
+
+	if (auto const* identifier = dynamic_cast<Identifier const*>(_expression))

IIUC, I've been including a test case for that, which is exactly like clangd jumps. If you've meant something else, we can talk about ASAP then.

libsolidity/ast/ASTUtils.h
```diff
@@ -33,4 +35,6 @@ VariableDeclaration const* rootConstVariableDeclaration(VariableDeclaration cons
 /// Returns true if the constant variable declaration is recursive.
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
 
+/// Teturns the innermost AST node that covers the given location or nullptr if not found.
```
Marenz: Typo `Teturns` -> `Returns`

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,78 @@
+#include <liblangutil/CharStreamProvider.h>
```
Marenz, line 19: Missing GPL header

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
```
Marenz, line 19: Missing GPL header

test/lsp.py
```diff
@@ -361,6 +361,52 @@ def expect_diagnostic(
...
+        obj: dict,
+        uri: str,
+        lineNo: int,
+        startEndColumns: Tuple[int, int]
+    ):
+        """
+        obj is an JSON object containing two keys:
+            - 'uri': a string of the document URI
+            - 'range': the location range, two child objects 'start' and 'end',
+                        each having a 'line' and 'cahracter' integer value.
```
Marenz: Typo `cahracter` -> `character`

Marenz (changes requested):
LGTM, just a few minor details.
I assume the whole thing will be merged to only a few commits eventually?

christianparpart:
> LGTM, just a few minor details. I assume the whole thing will be merged to only a few commits eventually?

Applied.

scripts/pylintrc
```diff
@@ -28,6 +28,7 @@ disable=
     too-many-arguments,
     too-many-branches,
     too-many-instance-attributes,
+    too-many-lines,
```
chriseth: Are you sure we want this global setting?

libsolidity/ast/ASTUtils.cpp
```diff
@@ -18,12 +18,50 @@
...
 
 namespace solidity::frontend
 {
 
+namespace
+{
+
+class ASTNodeLocator: public ASTConstVisitor
+{
+public:
```
chriseth: This will be used in a very very specific way. Does it make sense to change the public interface to only consist of a single static function that takes the offset and returns the ast node pointer?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+	ASTNode const* sourceNode = m_server.requestASTNode(sourceUnitName, lineColumn);
+	lspDebug(fmt::format("GotoDefinition: {}", typeid(*sourceNode).name()));
+
+	vector<SourceLocation> locations;
+	if (auto const* expression = dynamic_cast<Expression const*>(sourceNode))
+	{
+		// Handles all expressions that can have one or more declaration annotation.
+		for (auto const* declaration: allAnnotatedDeclarations(expression))
+			if (auto location = declarationLocation(declaration); location.has_value())
```
chriseth:
```suggestion
			if (auto location = declarationLocation(declaration))
```

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,54 @@
...
+namespace solidity::lsp
+{
+
+class FileRepository;
+
+std::optional<langutil::LineColumn> parseLineColumn(Json::Value const& _lineColumn);
+Json::Value toJson(langutil::LineColumn _pos);
+Json::Value toJsonRange(langutil::LineColumn const& _start, langutil::LineColumn const& _end);
+
+std::vector<frontend::Declaration const*> allAnnotatedDeclarations(frontend::Expression const* _expression);
```
chriseth: Please add documentation

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+Json::Value toJsonRange(LineColumn const& _start, LineColumn const& _end)
+{
+	Json::Value json;
+	json["start"] = toJson(_start);
+	json["end"] = toJson(_end);
+	return json;
+}
+
+vector<Declaration const*> allAnnotatedDeclarations(Expression const* _expression)
```
chriseth: In which case can this vector be larger than one?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -403,3 +351,20 @@ void LanguageServer::handleTextDocumentDidClose(Json::Value const& _args)
 
 	compileAndUpdateDiagnostics();
 }
+
+ASTNode const* LanguageServer::requestASTNode(std::string const& _sourceUnitName, LineColumn const& _filePos)
```
chriseth:
```suggestion
ASTNode const* LanguageServer::astNodeAtSourceLocation(std::string const& _sourceUnitName, LineColumn const& _filePos)
```
?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -403,3 +351,20 @@ void LanguageServer::handleTextDocumentDidClose(Json::Value const& _args)
...
+{
+	if (m_compilerStack.state() < CompilerStack::AnalysisPerformed)
+		return nullptr;
+
+	if (!m_fileRepository.sourceUnits().count(_sourceUnitName))
+		return nullptr;
+
+	optional<int> sourcePos = m_compilerStack.charStream(_sourceUnitName)
+		.translateLineColumnToPosition(_filePos);
```
chriseth:
```suggestion
	optional<int> sourcePos = m_compilerStack.charStream(_sourceUnitName).translateLineColumnToPosition(_filePos);
```

libsolidity/ast/ASTUtils.cpp
```diff
@@ -18,12 +18,50 @@
...
+	ASTNode const* innermostMatch() const noexcept { return m_innermostMatch; }
+
+private:
+	int const m_offsetInFile;
+	ASTNode const* m_innermostMatch;
+};
+
+}
+
+ASTNode const* locateInnermostASTNode(int _sourceOffset, SourceUnit const& _sourceUnit)
```
chriseth: Ah sorry, here is the function - shouldn't this be a static function in the class?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,55 @@
...
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* identifierPath = dynamic_cast<IdentifierPath const*>(sourceNode))
+	{
+		if (auto const* declaration = identifierPath->annotation().referencedDeclaration)
+			if (auto location = declarationLocation(declaration); location.has_value())
```
chriseth:
```suggestion
			if (auto location = declarationLocation(declaration))
```

libsolidity/lsp/HandlerBase.h
```diff
@@ -0,0 +1,67 @@
...
+
+	/// @returns source unit name and the line column position as extracted
+	/// from the JSON-RPC parameters.
+	std::pair<std::string, langutil::LineColumn> extractSourceUnitNameAndLineColumn(Json::Value const& _params) const;
+
+	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.charStreamProvider(); }
+	FileRepository const& fileRepository() const noexcept { return m_server.fileRepository(); }
+	Transport& client() const noexcept { return m_server.client(); }
+
+	LanguageServer& m_server;
```
chriseth, line 53: public?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -97,65 +77,31 @@ LanguageServer::LanguageServer(Transport& _transport):
...
-		return nullopt;
-
-	if (optional<LineColumn> lineColumn = parseLineColumn(_position))
-		if (optional<int> const offset = CharStream::translateLineColumnToPosition(
-			m_fileRepository.sourceUnits().at(_sourceUnitName),
-			*lineColumn
-		))
-			return SourceLocation{*offset, *offset, make_shared<string>(_sourceUnitName)};
-	return nullopt;
+	return HandlerBase{*this}.parseRange(_sourceUnitName, _range);
```
chriseth: It looks really weird that the server constructs a HandlerBase. Do these helper functions really need to go into `HandlerBase`? What about a helper class?

libsolidity/lsp/HandlerBase.cpp
```diff
@@ -0,0 +1,105 @@
...
+				fmt::arg("line", lineColumn.value().line),
+				fmt::arg("column", lineColumn.value().column),
+				fmt::arg("file", sourceUnitName)
+			))
+		);
+
+	return {sourceUnitName, *lineColumn};
+}
+
+optional<SourceLocation> HandlerBase::parseRange(string const& _sourceUnitName, Json::Value const& _range) const
```
chriseth: This is not even used in any handler yet.

scripts/pylintrc
christianparpart: At least until the the test suite has been refactored by @Marenz. I did not want to refactor this file just to eliminate this warning. We could create a ticket to remember to let go of `too-many-lines` in pylintrc again. What do you think?

chriseth: I think you can annotate individual files and even functions.

libsolidity/ast/ASTUtils.cpp
christianparpart: I've deleted this and rewrote `locateInnermostASTNode` to use `SimpleASTVisitor` instead. That should be even better?

libsolidity/lsp/Utils.cpp
christianparpart: I was just arguing on that on in the documentation I wrote. Basically I wanted to reserve the possibility that the resolver could not resolve the underlying type but found multiple candidates. Then the candidates should be reported IMHO.

libsolidity/ast/ASTUtils.cpp
christianparpart: I've eliminated the class.

libsolidity/lsp/HandlerBase.cpp
christianparpart: until then, moved back.

libsolidity/lsp/LanguageServer.cpp
christianparpart: moved into Utils.h now. As free functions however, as only one parameter needed to be added (FileRepository const&)

scripts/pylintrc
christianparpart: didn't think of being able to do that on a global file level. done.

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,55 @@
+#include <libsolidity/lsp/GotoDefinition.h>
```
Marenz, line 18: no gpl header

libsolidity/lsp/GotoDefinition.h
```diff
@@ -0,0 +1,14 @@
+#include <libsolidity/lsp/HandlerBase.h>
```
Marenz, line 18: no gpl header

Marenz (changes requested):

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,82 @@
...
+
+/// Extracts all resolved annotations of the given expression AST node.
+///
+/// This may for example by the type declaration of an identifier,
+/// or the type declaration of a structured member identifier.
+///
+/// The returned vector is empty if the expression was not resolved (maybe due to syntax/semantic error),
+/// or one if resolved.
+/// In the future, it may be possible to return more than one type declaration,
+/// if the type resolver did find more than one possible candidate declaration.
```
chriseth: Wouldn't it be better to return  nothing instead of multiple candidates? If we return multiple candidates, the result will be wrong, won't it?

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,82 @@
...
+/// or nullopt on failure.
+std::optional<langutil::SourceLocation> parseRange(
+	FileRepository const& _fileRepository,
+	std::string const& _sourceUnitName,
+	Json::Value const& _range
+);
+
+/// Extracts all resolved annotations of the given expression AST node.
+///
+/// This may for example by the type declaration of an identifier,
```
chriseth:
```suggestion
/// This may for example be the type declaration of an identifier,
```

libsolidity/lsp/Utils.h
```diff
@@ -0,0 +1,82 @@
...
+/// or the type declaration of a structured member identifier.
+///
+/// The returned vector is empty if the expression was not resolved (maybe due to syntax/semantic error),
+/// or one if resolved.
+/// In the future, it may be possible to return more than one type declaration,
+/// if the type resolver did find more than one possible candidate declaration.
+std::vector<frontend::Declaration const*> allAnnotatedDeclarations(frontend::Expression const* _expression);
```
chriseth: Also why "annotated" - shouldn't it be "referenced"?

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+	vector<Declaration const*> output;
+
+	if (auto const* identifier = dynamic_cast<Identifier const*>(_expression))
+	{
+		Declaration const* referencedDeclaration = identifier->annotation().referencedDeclaration;
+		if (referencedDeclaration)
+		{
+			lspDebug(fmt::format("referenced declaration: {} {}", typeid(*referencedDeclaration).name(), referencedDeclaration->name()));
```
chriseth: still needed?

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+			output.push_back(referencedDeclaration);
+		}
+	}
+	else if (auto const* memberAccess = dynamic_cast<MemberAccess const*>(_expression))
+	{
```
chriseth: not needed

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+Json::Value toJsonRange(LineColumn const& _start, LineColumn const& _end)
+{
+	Json::Value json;
+	json["start"] = toJson(_start);
+	json["end"] = toJson(_end);
+	return json;
+}
+
+vector<Declaration const*> allAnnotatedDeclarations(Expression const* _expression)
```
chriseth: Can _expression be nullptr? If not, please use a reference.

libsolidity/lsp/Utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+optional<LineColumn> parseLineColumn(Json::Value const& _lineColumn)
+{
+	if (_lineColumn.isObject() && _lineColumn["line"].isInt() && _lineColumn["character"].isInt())
+		return LineColumn{_lineColumn["line"].asInt(), _lineColumn["character"].asInt()};
+	else
+		return nullopt;
+}
+
+Json::Value toJson(LineColumn _pos)
```
chriseth:
```suggestion
Json::Value toJson(LineColumn const& _pos)
```

libsolidity/ast/ASTUtils.h
```diff
@@ -33,4 +35,6 @@ VariableDeclaration const* rootConstVariableDeclaration(VariableDeclaration cons
 /// Returns true if the constant variable declaration is recursive.
 bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
 
+/// Returns the innermost AST node that covers the given location or nullptr if not found.
+ASTNode const* locateInnermostASTNode(int _pos, SourceUnit const& _sourceUnit);
```
chriseth:
```suggestion
ASTNode const* locateInnermostASTNode(int _sourceOffset, SourceUnit const& _sourceUnit);
```

chriseth: actually it's _offsetInFile in the cpp file

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,72 @@
...
+using namespace solidity::langutil;
+using namespace solidity::lsp;
+using namespace std;
+
+void GotoDefinition::operator()(MessageID _id, Json::Value const& _args)
+{
+	auto const [sourceUnitName, lineColumn] = extractSourceUnitNameAndLineColumn(_args);
+
+	ASTNode const* sourceNode = m_server.astNodeAtSourceLocation(sourceUnitName, lineColumn);
+	lspDebug(fmt::format("GotoDefinition: {}", typeid(*sourceNode).name()));
```
chriseth: still needed?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,72 @@
...
+			if (auto location = declarationLocation(declaration))
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* importDirective = dynamic_cast<ImportDirective const*>(sourceNode))
+	{
+		auto const& path = *importDirective->annotation().absolutePath;
+		if (fileRepository().sourceUnits().count(path))
+			locations.emplace_back(SourceLocation{0, 0, make_shared<string const>(path)});
+	}
+	else if (auto const* declaration = dynamic_cast<Declaration const*>(sourceNode))
```
chriseth: What is the use-case here?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,72 @@
...
+			if (auto location = declarationLocation(declaration))
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* identifierPath = dynamic_cast<IdentifierPath const*>(sourceNode))
+	{
+		if (auto const* declaration = identifierPath->annotation().referencedDeclaration)
+			if (auto location = declarationLocation(declaration))
+				locations.emplace_back(move(location.value()));
+	}
+	else if (auto const* importDirective = dynamic_cast<ImportDirective const*>(sourceNode))
```
chriseth, line 55: Did you (manually) test that this also works with `import {x as y} from "abc.sol";` and `import "abc.sol" as X;`? In the first case, clicking on `x` should go to the definition of `x` inside of `abc.sol`.

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -0,0 +1,72 @@
...
+	{
+		auto const& path = *importDirective->annotation().absolutePath;
+		if (fileRepository().sourceUnits().count(path))
+			locations.emplace_back(SourceLocation{0, 0, make_shared<string const>(path)});
+	}
+	else if (auto const* declaration = dynamic_cast<Declaration const*>(sourceNode))
+	{
+		if (auto location = declarationLocation(declaration))
+			locations.emplace_back(move(location.value()));
+	}
```
chriseth: I think TypeName or at least UserDefinedTypeName is missing.

Please check, but it might be that we need a `declaration()` for the Type class. I have that in a pending PR, can look it up if you need it.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -403,3 +346,21 @@ void LanguageServer::handleTextDocumentDidClose(Json::Value const& _args)
...
+		return nullptr;
+
+	optional<int> sourcePos =
+		m_compilerStack.charStream(_sourceUnitName).translateLineColumnToPosition(_filePos);
+
+	if (!sourcePos)
+		return nullptr;
+
+	return locateInnermostASTNode(*sourcePos, m_compilerStack.ast(_sourceUnitName));
+}
```
chriseth:
```suggestion
	if (optional<int> sourcePos = m_compilerStack.charStream(_sourceUnitName).translateLineColumnToPosition(_filePos))
		return locateInnermostASTNode(*sourcePos, m_compilerStack.ast(_sourceUnitName));
	else
		return nullptr;
}
```

chriseth:
Please also check that a "goto definition" on a struct constructor call goes to the struct.

libsolidity/lsp/Utils.h
christianparpart: I think so.

libsolidity/lsp/Utils.h
christianparpart: good point!

libsolidity/lsp/Utils.cpp
christianparpart: Signature changed. nullptr now expresses that it could not be resolved.

test/lsp.py
```diff
@@ -744,6 +790,139 @@ def test_textDocument_didChange_at_eol(self, solc: JsonRpcProcess) -> None:
...
+            expected_lineNo=48,
+            expected_startEndColumns=(14, 15),
+            description="local variable declaration"
+        )
+
+        # User defined type
+        self.expect_goto_definition_location(
+            solc=solc,
+            document_uri=FILE_URI,
+            document_position=(58, 8), # symbol `Price` in `Price p ...`
```
chriseth, line 920: Interesting! Can you tell me the code path that is taken here?

chriseth: Oh, it's via identifire path! then everything is fine!

libsolidity/lsp/Utils.cpp
christianparpart: I think passing them by value is better, `LineColumn` also only holds two ints, which - by x64 ABI - are simply passed by assigning the member values to a register (for the first N parameters). Passing via cref requires dereferencing. I've removed the cref in the other occasions. If you dislike that, I can revert back. :-)

libsolidity/lsp/GotoDefinition.cpp
christianparpart: verified. no. removed.

Changelog.md
```diff
@@ -21,7 +21,7 @@ Compiler Features:
  * Commandline Interface: Event and error signatures are also returned when using ``--hashes``.
  * Yul Optimizer: Remove ``mstore`` and ``sstore`` operations if the slot already contains the same value.
  * Yul: Emit immutable references for pure yul code when requested.
-
+ * LSP: Implements goto-definition.
```
chriseth: Needs to move.

libsolidity/lsp/Utils.cpp
chriseth: I think we shouldn't do optimizations based on the size of a struct. It might change and people could be confused why you are passing it by value - as I am confused here.

libsolidity/lsp/GotoDefinition.cpp
christianparpart: I've added tests now.

Changelog.md
```diff
@@ -5,7 +5,8 @@ Language Features:
 
 
 Compiler Features:
-* JSON-AST: Added selector field for errors and events.
+ * LSP: Implements goto-definition.
+ * JSON-AST: Added selector field for errors and events.
```
chriseth: Does this belong here?

Changelog.md
```diff
@@ -23,7 +24,6 @@ Compiler Features:
  * Yul: Emit immutable references for pure yul code when requested.
 
 
-
```
chriseth: please fix

Marenz (dismissed):
my previous comments are resolved :)
Pending chrises now.

chriseth:
It also needs squashing.

libsolidity/lsp/GotoDefinition.cpp
christianparpart: as discussed, probably not needed anymore.

Changelog.md
christianparpart: LSP related items have been under "Compiler Features" category before too. I think (for now) that is okay?

chriseth (approved):

PR #12431, opened 2021-12-20
Safer constructors for FunctionType

chriseth:


libsolidity/ast/Types.h
```diff
@@ -1247,6 +1247,38 @@ class FunctionType: public Type
 		/// Cannot be called.
 		Declaration,
 	};
+	struct Options
+	{
+		/// true if the function takes an arbitrary number of arguments of arbitrary types
```
cameel:
```suggestion
		/// true iff the function takes an arbitrary number of arguments of arbitrary types
```

libsolidity/ast/Types.h
```diff
@@ -1247,6 +1247,38 @@ class FunctionType: public Type
...
+		bool gasSet = false;
+		/// true iff the value to be sent is on the stack
+		bool valueSet = false;
+		/// iff the salt value (for create2) to be used is on the stack
+		bool saltSet = false;
+		/// true iff the function is called as arg1.fun(arg2, ..., argn).
+		/// This is achieved through the "using for" directive.
+		bool bound = false;
+
+		static Options arbitraryParametersFactory()
```
cameel: How about this?
```suggestion
		static Options withArbitraryParameters()
```

libsolidity/ast/Types.h
```diff
@@ -1263,18 +1295,21 @@ class FunctionType: public Type
 		strings const& _parameterTypes,
 		strings const& _returnParameterTypes,
 		Kind _kind,
-		bool _arbitraryParameters = false,
-		StateMutability _stateMutability = StateMutability::NonPayable
+		StateMutability _stateMutability = StateMutability::NonPayable,
+		Options _options = Options{false, false, false, false, false}
```
cameel, line 1299: Would this not work?
```suggestion
		Options _options = Options{}
```

libsolidity/ast/TypeProvider.cpp
```diff
@@ -445,13 +445,18 @@ FunctionType const* TypeProvider::function(
...
 	strings const& _returnParameterTypes,
 	FunctionType::Kind _kind,
-	bool _arbitraryParameters,
-	StateMutability _stateMutability
+	StateMutability _stateMutability,
+	FunctionType::Options _options
 )
 {
+	// Can only use this constructor for "arbitraryParameters".
+	solAssert(!_options.valueSet && !_options.gasSet && !_options.bound);
```
cameel: Isn't salt missing here?
```suggestion
	solAssert(!_options.valueSet && !_options.gasSet && !options.saltSet && !_options.bound);
```

libsolidity/ast/Types.h
```diff
@@ -1247,6 +1247,38 @@ class FunctionType: public Type
...
+		/// This is achieved through the "using for" directive.
+		bool bound = false;
+
+		static Options arbitraryParametersFactory()
+		{
+			Options result;
+			result.arbitraryParameters = true;
+			return result;
+		}
+		static Options ofFunction(FunctionType const& _type)
```
cameel: After seeing it used, I think this would be slightly more readable:
```suggestion
		static Options fromFunctionType(FunctionType const& _type)
```

cameel (commented):
I think you may have missed the salt parameter in one of the assertions (see below). Other than that it seems fine.

libsolidity/ast/Types.h
chriseth: No, didn't work for me.

ekpyron (approved):

libsolidity/ast/Types.h
ekpyron: weird, though... don't see why it *wouldn't* work...

PR #12433, opened 2021-12-20
Remove constexpr.

chriseth:
Some compiler complained that the assertion contains a non-constexpr operation, and I think it is right.


cameel (commented):
So maybe we could replace this with `static_assert()`?

chriseth:
`static_assert` does not work it seems.

cameel (approved):
Weird. Even though it has a constant parameter?

Well, anyway, it's not like this `constexpr` matters all that much here so I'm going to approve the PR.

PR #12434, opened 2021-12-20
Output searched locations on import failure.

chriseth:
Revival of https://github.com/ethereum/solidity/pull/12267

### DIFF
--- a/test/solc/CommandLineInterfaceAllowPaths.cpp
+++ b/test/solc/CommandLineInterfaceAllowPaths.cpp
@@ -100,7 +100,7 @@ ImportCheck checkImport(
 		return ImportCheck::OK();
 
 	static regex const sourceNotFoundErrorRegex{
-		R"(^Error \(6275\): Source ".+" not found: (.*)\.\n)"
+		R"(^Error \(6275\): Source "[^"]+" not found: (.*)\.\n)"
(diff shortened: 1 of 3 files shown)


cameel (commented):
CLI tests need updating.

Also, I think it would be a good idea to also print allowed paths in the same way when you get a message that your path is not allowed.

ekpyron (dismissed):

ekpyron:
Needs CLI test updates.

test/solc/CommandLineInterfaceAllowPaths.cpp
```diff
@@ -100,7 +100,7 @@ ImportCheck checkImport(
 		return ImportCheck::OK();
 
 	static regex const sourceNotFoundErrorRegex{
-		R"(^Error \(6275\): Source ".+" not found: (.*)\.\n)"
+		R"(^Error \(6275\): Source "[^"]+" not found: (.*)\.\n)"
```
ekpyron, line 103: I wonder if we should expect that somebody may be parsing this just like we did here and therefore at least generously add a changelog entry for this...

ekpyron (approved):

test/solc/CommandLineInterfaceAllowPaths.cpp
chriseth: The problem was the greedy match to begin with.

PR #12437, opened 2021-12-20
abi.encodeCall for declarations.

### LINKED ISSUE #12435: abi.encodeCall of contractName.functionName doesn't work
## Description
The documentation of encodeCall shows:
```solidity
abi.encodeCall(contractName.functionName, (arg1, arg2, arg3)) 
```
https://blog.soliditylang.org/2021/12/20/solidity-0.8.11-release-announcement/

However when trying to make a POC this actually doesn't work.
It only works with a function pointer, which could be deduced from the rest of the documentation.
Note: encodeWithSelector works with both contractName.functionName and a function pointer.
See example below
Suggestion: Update the documentation and/or implement this feature

## Environment
Solidity 0.8.11 in Remix

## Steps to Reproduce

```solidity
// SPDX-License-Identifier: GPL-3.0-or-later
pragma solidity =0.8.11;
import "hardhat/console.sol";  

contract OtherContract {
   function hi(uint n) public  {
     console.log("hi",n);
   }
}

contract Account {
   OtherContract c=new OtherContract();
   
  constructor() {
    (bool success1, bytes memory data1) = address(c).call(abi.encodeWithSelector(OtherContract.hi.selector,1));
    (bool success2, bytes memory data2) = address(c).call(abi.encodeWithSelector(c.hi.selector,2));
    (bool success3, bytes memory data3) = address(c).call(abi.encodeCall(c.hi,(3)));
    //(bool success4, bytes memory data4) = address(c).call(abi.encodeCall(OtherContract.hi,(4))); // TypeError: Function must be public or external
  }
}
```

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12435

TODO:
 - [x] changelog entry
 - [x] general tests
 - [x] test to check if this works properly if referencing a base contract function.

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -2122,16 +2122,35 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 		return;
 	}
 
-	if (functionPointerType->kind() != FunctionType::Kind::External)
+	if (
+		functionPointerType->kind() != FunctionType::Kind::External &&
+		functionPointerType->kind() != FunctionType::Kind::Declaration
+	)
 	{
-		string msg = "Function must be \"public\" or \"external\".";
+		string msg = "Expected regular external function type, or external view on public function.";
+		if (functionPointerType->kind() == FunctionType::Kind::Internal)
+			msg += " Provided internal function.";
+		else if (functionPointerType->kind() == FunctionType::Kind::DelegateCall)
+			msg += " Cannot use library functions for abi.encodeCall.";
+		else if (functionPointerType->kind() == FunctionType::Kind::Creation)
+			msg += " Provided creation function.";
+		else
+			msg += " Cannot use special function.";
 		SecondarySourceLocation ssl{};
 
 		if (functionPointerType->hasDeclaration())
 		{
 			ssl.append("Function is declared here:", functionPointerType->declaration().location());
-			if (functionPointerType->declaration().scope() == m_currentContract)
+			if (
+				functionPointerType->declaration().visibility() == Visibility::Public &&
+				functionPointerType->declaration().scope() == m_currentContract
+			)
 				msg += " Did you forget to prefix \"this.\"?";
+			else if (contains(
+				m_currentContract->annotation().linearizedBaseContracts,
+				functionPointerType->declaration().scope()
+			) && functionPointerType->declaration().scope() != m_currentContract)
+				msg += " Functions from base contracts have to be external.";
 		}
 
 		m_errorReporter.typeError(3509_error, arguments[0]->location(), ssl, msg);
--- a/test/libsolidity/syntaxTests/specialFunctions/encodeCall.sol
+++ b/test/libsolidity/syntaxTests/specialFunctions/encodeCall.sol
@@ -2,81 +2,47 @@ interface I {
 	function fExternal(uint256 p, string memory t) external;
 }
 
-library L {
-	function fExternal(uint256 p, string memory t) external {}
+contract Other {
+	function fExternal(uint) external pure {}
+	function fPublic(uint) public pure {}
+	function fInternal(uint) internal pure {}
 }
 
-contract C {
-	using L for uint256;
+library L {
+	function fExternal(uint256 p, string memory t) external {}
+	function fInternal(uint256 p, string memory t) internal {}
+}
 
+contract Base {
+	function baseFunctionExternal(uint) external pure {}
+}
+
+contract C is Base {
 	function f(int a) public {}
 	function f2(int a, string memory b) public {}
-	function f3(int a, int b) public {}
 	function f4() public {}
-	function fInternal(uint256 p, string memory t) internal {}
 
-	function failFunctionArgsWrongType() public returns(bytes memory) {
-		return abi.encodeCall(this.f, ("test"));
-	}
-	function failFunctionArgsTooMany() public returns(bytes memory) {
-		return abi.encodeCall(this.f, (1, 2));
-	}
-	function failFunctionArgsTooFew0() public returns(bytes memory) {
-		return abi.encodeCall(this.f, ());
-	}
-	function failFunctionArgsTooFew1() public returns(bytes memory) {
-		return abi.encodeCall(this.f);
-	}
-	function failFunctionPtrMissing() public returns(bytes memory) {
-		return abi.encodeCall(1, this.f);
-	}
-	function failFunctionPtrWrongType() public returns(bytes memory) {
-		return abi.encodeCall(abi.encodeCall, (1, 2, 3, "test"));
-	}
-	function failFunctionInternal() public returns(bytes memory) {
-		return abi.encodeCall(fInternal, (1, "123"));
-	}
-	function failFunctionInternalFromVariable() public returns(bytes memory) {
-		function(uint256, string memory) internal localFunctionPointer = fInternal;
-		return abi.encodeCall(localFunctionPointer, (1, "123"));
-	}
-	function failFunctionArgsArrayLiteral() public returns(bytes memory) {
-		return abi.encodeCall(this.f3, [1, 2]);
-	}
-	function failLibraryPointerCall() public returns (bytes memory) {
-		return abi.encodeCall(L.fExternal, (1, "123"));
-	}
-	function failBoundLibraryPointerCall() public returns (bytes memory) {
-		uint256 x = 1;
-		return abi.encodeCall(x.fExternal, (1, "123"));
-	}
-	function failInterfacePointerCall() public returns (bytes memory) {
-		return abi.encodeCall(I.fExternal, (1, "123"));
-	}
-	function successFunctionArgsIntLiteralTuple() public returns(bytes memory) {
+	function successFunctionArgsIntLiteralTuple() public view returns(bytes memory) {
 		return abi.encodeCall(this.f, (1));
 	}
-	function successFunctionArgsIntLiteral() public returns(bytes memory) {
+	function successFunctionArgsIntLiteral() public view returns(bytes memory) {
 		return abi.encodeCall(this.f, 1);
 	}
-	function successFunctionArgsLiteralTuple() public returns(bytes memory) {
+	function successFunctionArgsLiteralTuple() public view returns(bytes memory) {
 		return abi.encodeCall(this.f2, (1, "test"));
 	}
-	function successFunctionArgsEmptyTuple() public returns(bytes memory) {
+	function successFunctionArgsEmptyTuple() public view returns(bytes memory) {
 		return abi.encodeCall(this.f4, ());
+	}
+	function viaDeclaration() public pure returns (bytes memory) {
+		return bytes.concat(
+			abi.encodeCall(Other.fExternal, (1)),
+			abi.encodeCall(Other.fPublic, (1)),
+			abi.encodeCall(I.fExternal, (1, "123"))
+		);
+	}
+	function viaBaseDeclaration() public pure returns (bytes memory) {
+		return abi.encodeCall(Base.baseFunctionExternal, (1));
 	}
 }
 // ----
-// TypeError 5407: (486-494): Cannot implicitly convert component at position 0 from "literal_string "test"" to "int256".
-// TypeError 7788: (576-606): Expected 1 instead of 2 components for the tuple parameter.
-// TypeError 7788: (687-713): Expected 1 instead of 0 components for the tuple parameter.
-// TypeError 6219: (794-816): Expected two arguments: a function pointer followed by a tuple.
-// TypeError 5511: (911-912): Expected first argument to be a function pointer, not "int_const 1".
-// TypeError 3509: (1018-1032): Function must be "public" or "external".
-// TypeError 3509: (1145-1154): Function must be "public" or "external". Did you forget to prefix "this."?
-// TypeError 3509: (1350-1370): Function must be "public" or "external".
-// TypeError 7515: (1469-1500): Expected a tuple with 2 components instead of a single non-tuple parameter.
-// TypeError 5407: (1493-1499): Cannot implicitly convert component at position 0 from "uint8[2]" to "int256".
-// TypeError 3509: (1596-1607): Function must be "public" or "external".
-// TypeError 3509: (1738-1749): Function must be "public" or "external".
-// TypeError 3509: (1860-1871): Function must be "public" or "external".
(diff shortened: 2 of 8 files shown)


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2132,6 +2135,11 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 			ssl.append("Function is declared here:", functionPointerType->declaration().location());
 			if (functionPointerType->declaration().scope() == m_currentContract)
 				msg += " Did you forget to prefix \"this.\"?";
+			else if (contains(
+				m_currentContract->annotation().linearizedBaseContracts,
+				functionPointerType->declaration().scope()
+			))
+				msg += " Functions from base contracts have to be external.";
```
cameel, line 2153: This will make the user think you can't use a function at with `abi.encodeCall()` when it's public. Make we should suggest assigning it to a function pointer or using a contract variable instead?

test/libsolidity/semanticTests/abiencodedecode/abi_encodecall_declaration.sol
```diff
@@ -0,0 +1,49 @@
...
+
+        (success, result) = t.staticcall(abi.encodeCall(Base.a, 100));
+        require(success && result.length == 32);
+        r += abi.decode(result, (uint));
+        require(r == 114);
+
+        (success, result) = t.staticcall(abi.encodeCall(this.a, 1000));
+        require(success && result.length == 32);
+        r += abi.decode(result, (uint));
+        require(r == 1115);
```
cameel: I don't think `+=` and `==` work on `bytes`.

Also, why add to `r` instead of just overwriting it? Would make the logic simpler. It took me a moment to see why you're not expecting 1000 here but I'm still not sure why it's 1115.

test/libsolidity/semanticTests/abiencodedecode/abi_encodecall_declaration.sol
```diff
@@ -0,0 +1,49 @@
...
+contract Base {
+    function a(uint x) external pure returns (uint) { return x + 1; }
+}
+
+contract C is Base {
+	function test() public view returns (uint r) {
+        address t = address(this);
+        bool success;
+        bytes memory result;
+        (success, result) = t.staticcall(abi.encodeCall(X.a, 1));
```
cameel: I think it's much easier to read without helper variables, especially when their names are just single letters.
```suggestion
        (success, result) = address(this).staticcall(abi.encodeCall(X.a, 1));
```

test/libsolidity/semanticTests/abiencodedecode/abi_encodecall_declaration.sol
```diff
@@ -0,0 +1,49 @@
+pragma abicoder v2;
```
cameel: This test should be called `abi_encode_call_declaration.sol` to match others in the same directory.

test/libsolidity/syntaxTests/specialFunctions/encodeCall.sol
```diff
@@ -2,81 +2,49 @@ interface I {
...
+contract Base {
+	function baseFunctionExternal(uint) external pure {}
+}
 
+contract C is Base {
 	function f(int a) public {}
 	function f2(int a, string memory b) public {}
 	function f3(int a, int b) public {}
 	function f4() public {}
 	function fInternal(uint256 p, string memory t) internal {}
```
cameel: These seem unused now.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_args.sol
```diff
@@ -0,0 +1,52 @@
...
+
+library L {
+	function fExternal(uint256 p, string memory t) external {}
+	function fInternal(uint256 p, string memory t) internal {}
+}
+
+contract Base {
+	function baseFunction(uint) internal pure {}
+	function baseFunctionExternal(uint) external pure {}
+}
```
cameel: This seems unused.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	function baseFunctionPublic(uint) public pure {}
+}
+
+contract C is Base {
+	using L for uint256;
+
+	function f(int a) public {}
+	function f2(int a, string memory b) public {}
+	function f3(int a, int b) public {}
+	function f4() public {}
```
cameel: This seems unused.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	function fInternal(uint) internal pure {}
+}
+
+library L {
+	function fExternal(uint256 p, string memory t) external {}
+	function fInternal(uint256 p, string memory t) internal {}
+}
+
+contract Base {
+	function baseFunction(uint) internal pure {}
```
cameel:
```suggestion
	function baseFunctionInternal(uint) internal pure {}
```

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	function baseFunctionPublic(uint) public pure {}
+}
+
+contract C is Base {
+	using L for uint256;
+
+	function f(int a) public {}
```
cameel:
```suggestion
	function fPublic(int a) public {}
```

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	function failLibraryPointerCall() public {
+		abi.encodeCall(L.fInternal, (1, "123"));
+		abi.encodeCall(L.fExternal, (1, "123"));
+	}
+	function failBoundLibraryPointerCall() public returns (bytes memory) {
+		uint256 x = 1;
+		return abi.encodeCall(x.fExternal, (1, "123"));
+	}
+	function viaBaseDeclaration() public pure returns (bytes memory) {
+		return abi.encodeCall(C.f, (2));
```
cameel: The error here is
```
Function must be "public" or "external". Did you forget to prefix "this."?
```
I think it would be better to get the same one you get when you use a public function from base class because it better explains why this does not work.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+		return abi.encodeCall(localFunctionPointer, (1, "123"));
+	}
+	function failLibraryPointerCall() public {
+		abi.encodeCall(L.fInternal, (1, "123"));
+		abi.encodeCall(L.fExternal, (1, "123"));
+	}
+	function failBoundLibraryPointerCall() public returns (bytes memory) {
+		uint256 x = 1;
+		return abi.encodeCall(x.fExternal, (1, "123"));
+	}
```
cameel, line 49: All of these give this error currently:
```
Function must be "public" or "external".
```
But two of these are already external. I think that we should have a more specific error message for library functions.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+contract Other {
+	function fExternal(uint) external pure {}
+	function fPublic(uint) public pure {}
+	function fInternal(uint) internal pure {}
+}
+
+library L {
+	function fExternal(uint256 p, string memory t) external {}
+	function fInternal(uint256 p, string memory t) internal {}
+}
```
cameel, line 14: Can you also add a free function to the test?

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	}
+	function failFunctionPtrWrongType() public returns(bytes memory) {
+		return abi.encodeCall(abi.encodeCall, (1, 2, 3, "test"));
+	}
+	function failFunctionInternal() public returns(bytes memory) {
+		return abi.encodeCall(fInternal, (1, "123"));
+	}
+	function failFunctionInternalFromVariable() public returns(bytes memory) {
+		function(uint256, string memory) internal localFunctionPointer = fInternal;
+		return abi.encodeCall(localFunctionPointer, (1, "123"));
```
cameel, line 40: The error here is:
```
Error: Function must be "public" or "external". Did you forget to prefix "this."?
```
I think we should not add this part about `this` when it's a pointer stored in a variable.

test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol
```diff
@@ -0,0 +1,72 @@
...
+	}
+	function viaBaseDeclaration() public pure returns (bytes memory) {
+		return abi.encodeCall(C.f, (2));
+	}
+	function viaBaseDeclaration() public pure returns (bytes memory) {
+		return bytes.concat(
+			abi.encodeCall(Base.baseFunctionPublic, (1)),
+			abi.encodeCall(Base.baseFunction, (1))
+		);
+	}
```
cameel: Both functions are called `viaBaseDeclaration()`. That's an error (and not one we're supposed to test here :)).

cameel (changes requested):

cameel:
> Fixes #12435

Before we can use that here we also need to change the blog post because that's a part of the issue.

libsolidity/analysis/TypeChecker.cpp
chriseth: Ok, this is actually true. But if it works if you assign it to an external function pointer, then it should also work without, shouldn't it? Maybe we can perform an implicit conversion or something.

cameel: Yeah, that would be even better. I thought about suggesting that but I wasn't sure how complicated it would be.

But I just checked to make sure this workaround would even work - and turns out it does not. We don't have a contract address on a function pointer referenced like that so you can't directly assign it to a function pointer:
```solidity
contract Base {
    function baseFunctionPublic(uint) public pure {}
}

contract C is Base {
    function viaBaseDeclaration() public pure  {
        function(uint) external pure fPtr = Base.baseFunctionPublic;
        abi.encodeCall(fPtr, (1));
    }
}
```
```
Error: Type function (uint256) pure is not implicitly convertible to expected type function (uint256) pure external. Special functions can not be converted to function types.
 --> test.sol:7:9:
  |
7 |         function(uint) external pure fPtr = Base.baseFunctionPublic;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^```

**Comment by bshastry** [comment]
Is this legal? Or is it rightly illegal?

```
contract C {
  function f() external { abi.encodeCall(C.f, ()); } // Error: Function must be "public" or "external".
}
```

Edit:

Funnily enough, the following already works. I suppose this is because `this.g` is not referencing the declaration but a pointer to an externally callable function.

```
contract C {
  function g() public { abi.encodeCall(this.g, ()); }
}
```

However, encoding a call to a public function declaration does not work but selector on the said declaration works:

```
contract C {
  function g() public {}
  function h() external { abi.encodeCall(C.g, ()); } // Error: Function must be "public" or "external".
}

contract D {
  function i() external { abi.encodeCall(C.g, ()); } // Error: Function must be "public" or "external".
  function j() external { C.g.selector; } // Works
}
```

**Comment by cameel** [comment]
Oh, we should definitely have this as a test.

I think it should be legal because you don't really have to call the function right away. You could just store it somewhere and call later. Whether such a construct would be actually useful in practice is another matter though :)

File: test/libsolidity/semanticTests/abiencodedecode/abi_encodecall_declaration.sol (hunk 2 above)
Comment (chriseth) [reply test/libsolidity/semanticTests/abiencodedecode/abi_encodecall_declaration.sol]: This is just to catch errors that could cancel each other out, and to detect where something went wrong.
Also note that `r` is `uint` and not `bytes`.

File: test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol (hunk 13 above)
Comment (chriseth) [reply test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol:40]: Oh interesting that it still has "hasDeclaration" set to true...

Comment (chriseth) [reply test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol:40]: Ah sorry, I think the error you quoted is from line 34, i.e. `abi.encodeCall(abi.encodeCall`

**Comment by chriseth** [comment]
Ok, I think this is as good as it gets WRT error messages.

**Comment by chriseth** [comment]
The example by @bshastry is working now, added a test for it.

File: test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol (hunk 11 above)
Comment (bshastry) [reply test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_funType.sol:49]: Would just like to add that the following compiles without an error

```
library L0 { function f(uint) public {} }
library L1 { function g() public { abi.encodeWithSelector(L0.f.selector, 1); } }
```

Also

```
library L {
  function f() external {}
  function g() internal { L.f.selector; }
}
```

Both of the above tests error via `abi.encodeCall` so there is an asymmetry to it i.e., users may still make use of the type unsafe `abi.encodeWithSelector` but not yet make use of the type safe `abi.encodeCall` alternative

libsolidity/analysis/TypeChecker.cpp
bshastry: Why are base contract public functions not supported by `abi.encodeCall`?

Marenz (changes requested):
PR looks good to me, though CI is still failing

libsolidity/analysis/TypeChecker.cpp
chriseth: `BaseContract.functionName` is the way to call the base implementation of a potentially virtual function, so it is an internal access.

chriseth:
Restricted the test to `>=byzantium` - so this hopefully works now.

Marenz (approved):

134dd3v:
Any idea when the new version that includes the fix'd be available?

cameel:
Can't give you a firm date on this but we usually release once per 1-2 months (when there's enough new stuff accumulated to be worth a release) - unless there's a security issue or a critical bug. I think you can expect this to be released somewhere in February.

PR #12441, opened 2021-12-20
Benchmarking external tests

cameel:
~Depends on #12440.~ Merged.
~Depends on #12532.~ Merged.
~Depends on #12629.~ Merged.

This PR adds a mechanism for gathering a set of metrics from all external tests and combining them into an easily diffable JSON report.

Currently it gathers the following information:
1. Gas usage as reported by the [eth-gas-reporter](https://github.com/cgewecke/eth-gas-reporter) plugin for Hardhat and Truffle. The plugin watches all function calls and contract deployments during test execution, measures used gas and presents results in an RST table.
2. Bytecode size extracted from Truffle and Hardhat artifacts with `jq`.

The biggest part (report creation) is already implemented but there are still a few small things to iron out:
- [x] Make sure the plugin works and actually produces reports in all external projects.
    - [ ] ~Colony~ (only runs on nigthly)
    - [ ] ~Gnosis Safe~ (will work after #12195)
    - [x] OpenZeppelin
    - [x] ENS
- [x] Skipping the report in compile-only runs.
- [x] Attaching reports as artifacts in CI.
- [x] Collector job to gather all artifacts and combine them into a single JSON file.
- [x] Simple command to process the combined JSON report into a short summary containing only totals.


cameel:
This is pretty much done now.

Benchmark results for Gnosis are still incomplete, probably because its tests are not based on mocha but they will work after the switch to Hardhat in #12195 so there's no point fixing that here. For Colony I'm not sure if they're complete but it only runs on nightly so I'm leaving it as is for now. We should update it to the latest version first anyway.

When reviewing check the artifacts of the `c_ext_benchmarks` job. This is the primary output we'll use for benchmarking.

test/externalTests/bleeps.sh
```diff
@@ -24,6 +24,8 @@ set -e
 source scripts/common.sh
 source test/externalTests/common.sh
 
+REPO_ROOT=$(realpath "$(dirname "$0")/../..")
+
```
ekpyron, line 28: Is this used? Same in colony.sh. Ah, nevermind, now I see the comment in ``common.sh``...
But why exactly don't you just compute it there? It's not the same everywhere?

ekpyron:
I had a casual look through this at least - since the result is informational and doesn't affect anything critically, we could probably merge it without a line-by-line review :-).
I'm mostly worried about how stable the construction will be, i.e. how often it'll break and how much maintenance it will require...

test/externalTests/bleeps.sh
cameel: Originally I put it in `common.sh` but IIRC `dirname "$0"` was empty because it's a sourced script. Just moving it here was a quick solution, though I could try harder to make it work in `common.sh`. It must be possible somehow.

cameel:
> I'm mostly worried about how stable the construction will be, i.e. how often it'll break and how much maintenance it will require...

It was fairly easy to write parsing for the table in Python and it's covered with unit tests so hopefully it won't be too bad. I also don't think the output format is fluctuating a lot.

I'd prefer being able to get a bit of JSON directly out of the gas reporter plugin but it does not seem possible without patching the plugin. I feel much more comfortable in Python than JS so writing a parsing script was a much faster way for me to deal with this problem. I actually envisioned the script as a much shorter thing but in the end with proper formatting, functions, data types, etc. it ended up longer. It's still conceptually a fairly simple thing that just runs a bunch of regexes over a table, line by line. To be honest I think I spent more time on the `jq` commands that later merge that JSON than on the script itself (surprisingly, `jq` does not have any nice and simple syntax for merging of nested dicts). If I had to do merging again I'd probably just do it in Python.

Also, this is just one specific gas reporting plugin (though it works with a lot of projects due to how popular Hardhat+Truffle are). If it turns out to be hard to maintain, I could try to replace it with something with JSON output. Or I could submit a patch upstream to add JSON output directly to the plugin. It's much more effort compared to what I have here though so I think there's much loss in merging it now and replacing it with something better later.

cameel:
Rebased on the Uniswap PR (#12532) since that's likely to get merged soon and that way I can already include it here..

cameel:
I updated the code for new external tests. This should now pass CI.

I also fixed an issue with benchmark results being missing for PRBMath and Bleeps. It looks like running tests via `mocha` suppresses the output from `hardhat-gas-reporter`. I had to change it so that tests are executed directly with `hardhat test`.

Unfortunately results seem incomplete in both cases. In Bleeps the table has no deployment costs while in PRBMath method call costs are missing. I'm not sure why so I asked for help in https://github.com/paulrberg/prb-math/issues/70.

In any case, I think we should merge it even in the current state. Benchmark results are just extra information and jobs are on purpose designed in such a way that not having them does not make them fail.

cameel:
Regarding no calls in PRBMath - mystery solved. All the methods called in tests are `pure` or `view` so there's no actual transaction happening and and `eth_call` does not report gas usage.

[There are plans to make hardhat-gas-reporter inject `eth_estimageGas` instead](https://github.com/cgewecke/hardhat-gas-reporter/issues/92) and once that happens, we should automatically start seeing gas values for these methods in our benchmarks.

cameel:
As for Bleeps, the issue is caused by the use of hardhat-deploy plugin. It deploys the contracts before tests are executed and hardhat-gas-reporter does not monitor these calls. The project actually has a separate command to report deployment gas. There's an issue to fix that in hardhat-gas-reporter though: https://github.com/cgewecke/hardhat-gas-reporter/issues/86. So this is another thing that should just work once it's fixed upstream.

ekpyron (approved):
I vote for just merging this and seeing how it performs.
I by far have not done a complete line-by-line review, but since
(1) all of this is purely informative and non-critical, yet also (2) the information
will be quite helpful and useful, I think it may be reasonable to merge
and see how well it works without a full rigorous review.

PR #12443, opened 2021-12-21
[pylint] Use multiple processes when possible

cameel:
I noticed that the default is for pylint to use only one core. This PR enables auto-detection.

It does not change that much about running time since it's pretty fast already (IMO it's not even worth changing the resource class in CI) but if we can enable it, why not. One unexpected advantage is that when running locally, output appears faster, while it's still working.

### DIFF
--- a/scripts/pylintrc
+++ b/scripts/pylintrc
@@ -1,4 +1,10 @@
 # vim:et:ts=4
+[MASTER]
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=0
+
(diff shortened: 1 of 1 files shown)


cameel:
Weird. Looks like the autodetection does not work well in CI because it just hangs.

cameel:
Let's see if rebasing makes it pass. But if not, I have no more ideas and I'll probably just close it.

leonardoalt:
> Let's see if rebasing makes it pass. But if not, I have no more ideas and I'll probably just close it.

I didn't. Based on this I'll close it.

PR #12444, opened 2021-12-21
[pylint] Disable the opinionated too-many-xyz warnings

cameel:
Followup for https://github.com/ethereum/solidity/pull/11350#discussion_r760497667.

We decided we'd prefer these warnings disabled globally since they're not helpful - we're always disabling them locally anyway.


chriseth (dismissed):
:shrug: I find it kind of useful to be honest, but maybe not too important for typed parameters

cameel:
Yeah, n some cases the code might be better off refactored to have fewer parameters but I just don't find it important enough to have linter enforce it. But if you think it's better as is, I'd also be fine with having to keep disabling it on a case-by-case basis.

PR #12453, opened 2021-12-23
Split content-holding from file reader.

chriseth:
This is a refactoring towards allowing custom file read callbacks for the LSP which is needed by solc-js.

I'm not yet sure if this actually goes in the right direction.


leonardoalt:
How would you decide whether this is going in the right direction?

PR #12456, opened 2021-12-24
added warning if isoltest gas-cost-expectations are not enforced

### LINKED ISSUE #12386: [isoltest] Warn if the gas cost expectations are not being enforced
Currently `isoltest` automatically enables `--enforce-gas-cost` option but only if other options conflicting with it are not used (i.e. `--abiencoderv1` or `--evm-version` other than the latest). It's easy to miss that, especially if someone is not yet familiar with all the quirks of the tool. Isoltest should print a message if it's not actually enforcing gas costs.

nishant-sachdeva:
Added preliminary code to generate a warning. 

Issues with the code : Not sure how to generate a new error code

Closes #12386

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -191,6 +191,9 @@ test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] )
 	if (solidity::test::CommonOptions::get().disableSemanticTests)
 		cout << endl << "--- SKIPPING ALL SEMANTICS TESTS ---" << endl << endl;
 
+	if (!solidity::test::CommonOptions::get().enforceGasTest)
+		cout << endl << "WARNING :: Gas Cost Expectations are not being enforced" << endl << endl;
+
 	// Include the interactive tests in the automatic tests as well
 	for (auto const& ts: g_interactiveTestsuites)
 	{
--- a/test/tools/isoltest.cpp
+++ b/test/tools/isoltest.cpp
@@ -437,6 +437,9 @@ int main(int argc, char const *argv[])
 		if (options.disableSemanticTests)
 			cout << endl << "--- SKIPPING ALL SEMANTICS TESTS ---" << endl << endl;
 
+		if (!options.enforceGasTest)
+			cout << "WARNING :: Gas Cost Expectations are not being enforced" << endl << endl;
+
 		TestStats global_stats{0, 0};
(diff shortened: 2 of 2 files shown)


chriseth:
I think it's better to put this message at the place where the `parse` function is called. We already have a message "SKIPPING ALL SEMANTIC TESTS" - I think it would be good to place it next to that. Note that there are two files that have this message.

Changelog.md
```diff
@@ -5,6 +5,7 @@ Language Features:
 
 
 Compiler Features:
+* Generate warnings on isoltest when Gas Cost Expectations are not being enforced.
```
chriseth: This only concerns the tests and thus should not have a changelog entry.

test/tools/IsolTestOptions.cpp
```diff
@@ -84,7 +84,6 @@ bool IsolTestOptions::parse(int _argc, char const* const* _argv)
 	}
 
 	enforceGasTest = enforceGasTest || (evmVersion() == langutil::EVMVersion{} && !useABIEncoderV1);
-
```
chriseth: Can you undo this change?

Changelog.md
nishant-sachdeva: Removed in the latest push

chriseth (dismissed):

test/tools/IsolTestOptions.cpp
nishant-sachdeva: Fixed in the latest push.

chriseth:
Weird, I'm not seeing the warning in circleCI - will double check.

test/boostTest.cpp
```diff
@@ -191,6 +191,9 @@ test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] )
 	if (solidity::test::CommonOptions::get().disableSemanticTests)
 		cout << endl << "--- SKIPPING ALL SEMANTICS TESTS ---" << endl << endl;
 
+	if (solidity::test::CommonOptions::get().enforceGasTest)
```
chriseth:
```suggestion
	if (!solidity::test::CommonOptions::get().enforceGasTest)
```

nishant-sachdeva: Fixed in the latest push

chriseth (approved):

PR #12461, opened 2021-12-28
[trivial] Remove unnecessary checks in modifier resolve function

Marenz:


libsolidity/ast/AST.cpp
```diff
@@ -514,20 +514,16 @@ ModifierDefinition const& ModifierDefinition::resolveVirtual(
...
 	// If we are not doing super-lookup and the modifier is not virtual, we can stop here.
-	if (_searchStart == nullptr && !virtualSemantics())
+	if (!virtualSemantics())
 		return *this;
 
 	solAssert(!dynamic_cast<ContractDefinition const&>(*scope()).isLibrary(), "");
 
 	for (ContractDefinition const* c: _mostDerivedContract.annotation().linearizedBaseContracts)
-	{
-		if (_searchStart != nullptr && c != _searchStart)
```
bshastry, line 524: I am wondering what the intention of this check was prior to this PR.

- `_searchStart` will always be nullptr here (because of previous assert plus following assignment)
- `c != _searchStart` is essentially `c != nullptr` (this does not make sense)

libsolidity/ast/AST.cpp
```diff
@@ -514,20 +514,16 @@ ModifierDefinition const& ModifierDefinition::resolveVirtual(
 	solAssert(_searchStart == nullptr, "Used super in connection with modifiers.");
 
 	// If we are not doing super-lookup and the modifier is not virtual, we can stop here.
```
chriseth: The comment talks about super lookup. Why can we ignore it? Can you modify the comment and add an assertion?

chriseth:
Why is this trivial?

libsolidity/ast/AST.cpp
chriseth: Ah, sorry, did not see the assert above. Can you modify the comment, please?

libsolidity/ast/AST.cpp
chriseth: I think the assert was added after this code was written.

Marenz:
> Why is this trivial?

I meant with that it's small and easy to review

Marenz:
Updated comments

chriseth (approved):

PR #12462, opened 2021-12-29
[docs] fix issue with delegating votes to wallets who cannot vote.

semaraugusto:
The voting example on the docs allows addresses who cannot vote yet to be delegated to. This makes it so that they lose out on their own vote. This change makes sure only people who can vote are delegated to, which fixes the problem

(due credits where credits are due: https://github.com/miaortizma pointed out that someone could delegate to addresses who couldn't vote)

### DIFF
--- a/docs/examples/voting.rst
+++ b/docs/examples/voting.rst
@@ -133,7 +133,7 @@ of votes.
             Voter storage delegate_ = voters[to];
 
             // Voters cannot delegate to wallets that cannot vote.
-            require(delegate_.weight >= 1)
+            require(delegate_.weight >= 1);
             sender.voted = true;
             sender.delegate = to;
             if (delegate_.voted) {
--- a/docs/examples/voting.rst
+++ b/docs/examples/voting.rst
@@ -130,9 +130,12 @@ of votes.
 
             // Since `sender` is a reference, this
             // modifies `voters[msg.sender].voted`
+            Voter storage delegate_ = voters[to];
+
+            // Voters cannot delegate to wallets that cannot vote.
+            require(delegate_.weight >= 1)
             sender.voted = true;
             sender.delegate = to;
-            Voter storage delegate_ = voters[to];
             if (delegate_.voted) {
                 // If the delegate already voted,
                 // directly add to the number of votes


docs/examples/voting.rst
```diff
@@ -130,9 +130,12 @@ of votes.
 
             // Since `sender` is a reference, this
             // modifies `voters[msg.sender].voted`
+            Voter storage delegate_ = voters[to];
+
+            // Voters cannot delegate to wallets that cannot vote.
+            require(delegate_.weight >= 1)
```
bshastry:
```suggestion
            require(delegate_.weight >= 1);
```

bshastry (commented):
Syntax error that is failing CI tests.

chriseth:
Why do you think this is a problem? If you delegate to someone who does not have the right to vote yet, the delegation will allow them to vote.

semaraugusto:
yes, delegation will allow the address to vote, but it won't ever have it's own vote. I feel like this should not be the case, so either we only allow delegations to happen to addresses that can already vote or there could be a "phase 1" and a "phase 2", where phase 1 allows all allowed participants and phase 2 is when the actual voting begin

chriseth:
The code is built so that it does not require these phases - already existing delegations are followed and forwarded. I don't see why we need the complication to check if someone has a nonzero weight before they receive a delegation.

semaraugusto:
if voter A has right to vote and delegates to voter B who cannot, the chairman is know prohibited from giving that wallet the right to vote. That way voter B cannot get his own vote if someone has delegated to him

chriseth:
Ah, that is an interesting case! So should we rather change the `giveRightToVote` to add `1` to `weight` unless the person has already voted?

semaraugusto:
Well the require on the giveRightsToVoter would not allow that. Removing it whilst changing it to add one to the voter would make it so that chairman can give infinite votes to the voter.

I also implemented another version of the code where there's two structs as follows. I didn't commit it because I really like how simple the example is, so I thought that adding the require would be the cleaner option to keep the doc example "as-is"
```
    struct Voter{
        bool can_vote;
        bool voted; // preventing chairperson from cheating
        address delegate;
    }

    // Delegators accumulate votes sent to them
    struct Delegate{
        uint32 weight;
        uint32 proposal_id;
        bool voted;
    }
```

the more important changes that this enforces are
```
    function giveRightToVote(address[] memory voters_addr) external onlyChairperson {
        for(uint i=0; i < voters_addr.length; i++) {
            address voter = voters_addr[i];
            require(
                !voters[voter].can_vote,
                "The voter is already allowed."
            );
            require(
                !voters[voter].voted,
                "The voter has already voted."
            );
            voters[voter].can_vote = true;
        }
    }

    function delegate(address to) external {
        require(to != msg.sender, "Self-delegation is disallowed.");

        Voter storage sender = voters[msg.sender];
        require(sender.can_vote, "You already voted.");

        while (voters[to].delegate != address(0)) {
            to = voters[to].delegate;

            // We found a loop in the delegation, not allowed.
            require(to != msg.sender, "Found loop in delegation.");
        }

        sender.can_vote = false;
        sender.voted = true;
        sender.delegate = to;

        Delegate storage delegate_ = delegates[to];

        if (delegate_.voted) {
            // If the delegate already voted,
            // directly add to the number of votes
            proposals[delegate_.proposal_id].voteCount += 1;
        } else {
            if (delegate_.weight == 0) {
                delegate_.weight = 1; // delegates have their own votes
            }
            // If the delegate did no vote yet,
            // add to her weight.
            delegate_.weight += 1;
        }
    }

    function vote(uint32 proposal) external {
        Voter storage voter = voters[msg.sender];
        require(voter.can_vote, "Has no right to vote");

        Delegate storage sender = delegates[msg.sender];
        require(!sender.voted, "Already voted.");

        voter.can_vote = false;
        voter.voted = true;

        // Allows for a random voter to become a delegate himself if he wishes to vote directly
        if (sender.weight == 0) {
            sender.weight = 1;
        }

        sender.voted = true;
        sender.proposal_id = proposal;

        // If `proposal` is out of the range of the array,
        // this will throw automatically and revert all
        // changes.
        proposals[proposal].voteCount += sender.weight;
    }

```

dzhalaevd (approved):

leonardoalt:
@chriseth is the current PR fine then? I think it's ok, but you had concerns so I'll wait.

chriseth (approved):

PR #12465, opened 2021-12-29
Improved code generation for modifiers.

### LINKED ISSUE #6584: Have modifiers JUMP rather than inlining
The code of a modifier is currently copied into the function to which it is applied. Instead, I suggest to treat the modifier more like an internal function and to use a `jump` instead.

This has some benefits:

1. Reduces codesize of contracts through code reuse.
2. Some static analysis can be made more efficient (e.g. lemmas can be made for internal functions much more easily than inlined code).
3. Modifiers would no longer contribute to "stack too deep" errors.

chriseth:
fixes https://github.com/ethereum/solidity/issues/6584

This changes modifier invocations to use function calls to the modifier with a dynamic return jump, but only if the modifier only has a single `_;`, that is the last statement and it does not have a `return` statement.

TODO:
- [ ] some more tests (open for suggestions)
- [ ] codegen for yul

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 ### 0.8.12 (unreleased)
 
 Compiler Features:
-
+ * Modifiers: Use internal function calls and thus prevent code duplication for modifiers that have a single `_;` at the end.
 
 
 Bugfixes:
--- a/libsolidity/analysis/SyntaxChecker.cpp
+++ b/libsolidity/analysis/SyntaxChecker.cpp
@@ -179,6 +179,7 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
 bool SyntaxChecker::visit(ModifierDefinition const&)
 {
 	m_placeholderFound = false;
+	m_modifierIsSimple = true;
 	return true;
 }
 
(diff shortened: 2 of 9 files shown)


hrkrshnn:
I wonder if we should only do this for the new codegen. The increased complexity for the legacy codegen for this optimization is not worth it, IMO.

chriseth:
I actually think that it's more complicated to implement it in the new codegen :)

Having said that, we really should check that it provides a good improvement in "real" contracts.

frangio:
Does this unconditionally use an internal function call or does it get inlined if the modifier is used only once? (Similar to how I understand internal functions are optimized.)

cameel:
It's done at the code generation stage and the optimizer runs after that so looks like it's subject to all the optimizations, including inlining. Whether it'll actually get inlined depends on the decisions inliner will make and a lot can be improved there but that's a separate matter.

leonardoalt:
Does it make sense to do this for the old codegen still?

leonardoalt:
Closing due to inactivity + it's done only for the old codegen. Probably makes sense to do it directly on the new gen.

PR #12468, opened 2021-12-30
Always allow full filesystem access to LSP.

chriseth:
Maybe we need a different solution for Windows? Something like a flag to just disable the check inside the file reader?

### DIFF
--- a/libsolidity/lsp/FileRepository.cpp
+++ b/libsolidity/lsp/FileRepository.cpp
@@ -17,48 +17,130 @@
 // SPDX-License-Identifier: GPL-3.0
 
 #include <libsolidity/lsp/FileRepository.h>
+#include <libsolidity/lsp/Utils.h>
+
+#include <libsolutil/StringUtils.h>
+#include <libsolutil/CommonIO.h>
+
+#include <range/v3/range/conversion.hpp>
+#include <range/v3/view/transform.hpp>
+
+#include <regex>
 
 using namespace std;
 using namespace solidity;
 using namespace solidity::lsp;
+using namespace solidity::frontend;
 
-namespace
-{
+using solidity::util::readFileAsString;
+using solidity::util::joinHumanReadable;
 
-string stripFilePrefix(string const& _path)
+FileRepository::FileRepository(boost::filesystem::path _basePath): m_basePath(std::move(_basePath))
 {
-	if (_path.find("file://") == 0)
-		return _path.substr(7);
-	else
-		return _path;
 }
 
-}
-
-string FileRepository::sourceUnitNameToClientPath(string const& _sourceUnitName) const
+string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
 {
-	if (m_sourceUnitNamesToClientPaths.count(_sourceUnitName))
-		return m_sourceUnitNamesToClientPaths.at(_sourceUnitName);
+	regex const windowsDriveLetterPath("^[a-zA-Z]:/");
+
+	if (m_sourceUnitNamesToUri.count(_sourceUnitName))
+		return m_sourceUnitNamesToUri.at(_sourceUnitName);
 	else if (_sourceUnitName.find("file://") == 0)
 		return _sourceUnitName;
+	else if (regex_search(_sourceUnitName, windowsDriveLetterPath))
+		return "file:///" + _sourceUnitName;
+	else if (_sourceUnitName.find("/") == 0)
+		return "file://" + _sourceUnitName;
 	else
-		return "file://" + (m_fileReader.basePath() / _sourceUnitName).generic_string();
+		return "file://" + m_basePath.generic_string() + "/" + _sourceUnitName;
 }
 
-string FileRepository::clientPathToSourceUnitName(string const& _path) const
+string FileRepository::uriToSourceUnitName(string const& _path) const
 {
-	return m_fileReader.cliPathToSourceUnitName(stripFilePrefix(_path));
+	return stripFileUriSchemePrefix(_path);
 }
 
-map<string, string> const& FileRepository::sourceUnits() const
-{
-	return m_fileReader.sourceUnits();
-}
-
-void FileRepository::setSourceByClientPath(string const& _uri, string _text)
+void FileRepository::setSourceByUri(string const& _uri, string _source)
 {
 	// This is needed for uris outside the base path. It can lead to collisions,
 	// but we need to mostly rewrite this in a future version anyway.
-	m_sourceUnitNamesToClientPaths.emplace(clientPathToSourceUnitName(_uri), _uri);
-	m_fileReader.addOrUpdateFile(stripFilePrefix(_uri), move(_text));
+	auto sourceUnitName = uriToSourceUnitName(_uri);
+	m_sourceUnitNamesToUri.emplace(sourceUnitName, _uri);
+	m_sourceCodes[sourceUnitName] = std::move(_source);
 }
+
+frontend::ReadCallback::Result FileRepository::readFile(string const& _kind, string const& _sourceUnitName)
+{
+	solAssert(
+		_kind == ReadCallback::kindString(ReadCallback::Kind::ReadFile),
+		"ReadFile callback used as callback kind " + _kind
+	);
+
+	try
+	{
+		// File was read already. Use local store.
+		if (m_sourceCodes.count(_sourceUnitName))
+			return ReadCallback::Result{true, m_sourceCodes.at(_sourceUnitName)};
+
+		string const strippedSourceUnitName = stripFileUriSchemePrefix(_sourceUnitName);
+
+		if (
+			boost::filesystem::path(strippedSourceUnitName).has_root_path() &&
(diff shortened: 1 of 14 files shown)


ekpyron:
@christianparpart For CI to pass this will need a rebase to current ``develop``.

libsolidity/lsp/FileRepository.h
```diff
@@ -29,7 +29,10 @@ class FileRepository
 {
 public:
 	explicit FileRepository(boost::filesystem::path const& _basePath):
-		m_fileReader(_basePath) {}
+		m_fileReader(_basePath)
+	{
+		m_fileReader.allowDirectory(boost::filesystem::path("/"));
```
chriseth: Does this work on windows?

Aivolt1 (dismissed):

libsolidity/lsp/FileRepository.h
christianparpart: Yes, with my changes added to this PR: All tests pass (locally).

test/lsp.py
```diff
@@ -180,6 +177,10 @@ def __init__(self):
         args = create_cli_parser().parse_args()
         self.solc_path = args.solc_path
         self.project_root_dir = os.path.realpath(args.project_root_dir) + "/test/libsolidity/lsp"
+        if os.name == 'nt':
```
chriseth: What about using `as_uri()` from `pathlib`?

test/lsp.py
```diff
@@ -60,10 +60,7 @@ def receive_message(self) -> Union[None, dict]:
                 # server quit
                 return None
             line = line.decode("utf-8")
-            if not line.endswith("\r\n"):
```
chriseth, line 146: Why did you remove this check?

libsolidity/lsp/Transport.cpp
```diff
@@ -107,8 +107,15 @@ void IOStreamTransport::send(Json::Value _json, MessageID _id)
 
 	string const jsonString = solidity::util::jsonCompactPrint(_json);
 
-	m_output << "Content-Length: " << jsonString.size() << "\r\n";
-	m_output << "\r\n";
+#if defined(_WIN32)
```
chriseth: I don't think this is a proper solution. I saw some posts mentioning freopen and others `setmode`: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-170&viewFallbackFrom=vs-2019

libsolidity/lsp/FileRepository.cpp
```diff
@@ -42,7 +42,16 @@ string FileRepository::sourceUnitNameToClientPath(string const& _sourceUnitName)
 	else if (_sourceUnitName.find("file://") == 0)
 		return _sourceUnitName;
 	else
-		return "file://" + (m_fileReader.basePath() / _sourceUnitName).generic_string();
+	{
+		auto basePath = m_fileReader.basePath();
+#if defined(_WIN32)
+		// Ensure basePath does contain Windows drive letter to fully qualify the path.
+		if (!basePath.has_root_name())
+			basePath = (boost::filesystem::current_path().root_path() / basePath).normalize();
```
chriseth: I don't think we should use the current path.

test/lsp.py
christianparpart: I did not know about that. will be trying.

test/lsp.py
christianparpart: the rstrip does also remove CRLF but more relaxed, such that it would have allowed a simple LF too. I've reverted it though.

libsolidity/lsp/Transport.cpp
```diff
@@ -75,67 +70,148 @@ optional<Json::Value> IOStreamTransport::receive()
...
-		if (delimiterPos == string::npos)
-			return nullopt;
+std::string StdioTransport::readBytes(size_t _byteCount)
+{
+	std::string buffer;
+	buffer.resize(_byteCount);
+	auto const n = fread(buffer.data(), 1, _byteCount, stdin);
+	if (n < _byteCount)
+		buffer.resize(n);
+	lspDebug(fmt::format("StdioTransport.readBytes({};{};{}): \"{}\"",
```
chriseth: coding style

libsolidity/lsp/Transport.cpp
```diff
@@ -75,67 +70,148 @@ optional<Json::Value> IOStreamTransport::receive()
...
-			boost::trim_copy(value)
-		).second)
-			return nullopt;
-	}
-	return {move(headers)};
+std::string StdioTransport::getline()
+{
+	char* lineBuffer = nullptr;
+	size_t numBytesRead = 0;
+	auto const getlineResult = ::getline(&lineBuffer, &numBytesRead, stdin);
```
chriseth: Can you use an explicit type please?

libsolidity/lsp/Transport.cpp
```diff
@@ -75,67 +70,148 @@ optional<Json::Value> IOStreamTransport::receive()
...
+	auto const ensureFreedBuffer = ScopeGuard{ [&]() { free(lineBuffer); }};
```
chriseth:
```suggestion
	ScopeGuard const ensureFreedBuffer{[&]() { free(lineBuffer); }};
```

libsolidity/lsp/Transport.cpp
```diff
@@ -75,67 +70,148 @@ optional<Json::Value> IOStreamTransport::receive()
...
+	if (getlineResult > 0)
+		return std::string(lineBuffer, static_cast<size_t>(getlineResult));
+	else
+		return {};
+}
+
+void StdioTransport::writeBytes(std::string_view _data)
+{
+	auto const bytesWritten = fwrite(_data.data(), 1, _data.size(), stdout);
```
chriseth, line 213: Please use an explicit type

leonardoalt:
Is this PR waiting for fixes or reviews? @chriseth @christianparpart

christianparpart:
> Is this PR waiting for fixes or reviews? @chriseth @christianparpart

yes, please. :)

libsolidity/lsp/FileRepository.cpp
```diff
@@ -17,48 +17,129 @@
...
+
+	try
+	{
+		// File was read already. Use local store.
+		if (m_sourceCodes.count(_sourceUnitName))
+			return ReadCallback::Result{true, m_sourceCodes.at(_sourceUnitName)};
+
+		string const strippedSourceUnitName = stripFileUriSchemePrefix(_sourceUnitName);
+
+		if (boost::filesystem::path(strippedSourceUnitName).has_root_path() &&
```
Marenz: `if (` should be on its own

Marenz (changes requested):
Found a tiny styling problem, but pylint also demands fixes.

libsolidity/lsp/Transport.cpp
```diff
@@ -75,79 +69,156 @@ optional<Json::Value> IOStreamTransport::receive()
...
+		if (boost::trim_copy(line).empty())
+			break;
+
+		auto const delimiterPos = line.find(':');
+		if (delimiterPos == string::npos)
+			return nullopt;
+
+		auto const name = boost::to_lower_copy(line.substr(0, delimiterPos));
+		auto const value = line.substr(delimiterPos + 1);
+		if (!headers.emplace(
```
Marenz: `if (` should be on its own

test/lsp.py
```diff
@@ -1,26 +1,57 @@
...
+    """
+    Gets a single character from stdin without line-buffering.
+    """
+    if os.name == 'nt':
+        # pragma pylint: disable=import-error
+        return msvcrt.getch().decode("utf-8")
+    else:
+        # Turn off user input buffering so we get the input immediately,
+        # not only after a line break
+        tty.setcbreak(sys.stdin.fileno())
```
Marenz: This doesn't need to be set every time we want a char. Once in the life-time of the program is enough.

Marenz (changes requested):

libsolidity/lsp/Transport.cpp
christianparpart: This is just moved code, not new one. I change change however. :)

test/lsp.py
christianparpart: Exactly, but that way I kept the logic more closely together without over-complifying something simple.

Marenz (dismissed):

Marenz (approved):

libsolidity/lsp/Transport.h
```diff
@@ -119,27 +143,34 @@ class IOStreamTransport: public Transport
...
+	void writeBytes(std::string_view _data) override;
+	void flushOutput() override;
 
 private:
 	std::istream& m_input;
 	std::ostream& m_output;
 };
 
+/**
+ * Standard I/O transport Layer utilizing stdin/stdout for communication.
```
chriseth, line 160: The changes in these files do not look like they are about allowing full filesystem access. Please put them in their own PR in the future.

christianparpart: I was considering that. Okay.

PR #12470, opened 2021-12-30
Equality operator allowed for external function types

### LINKED ISSUE #12411: Equality operator for external functions
## Abstract
Currently the equality operator only works with internal function types. It should be allowed also for external function types.

## Motivation
The way you compare function pointers today is not intuitive. For external ones you can only compare serialized values (e.g. `keccak256(abi.encodePacked(f))`) and using `==` and `!=` is disallowed. For internal ones the opposite is true. It would be much more intuitive to have one method that works uniformly with both.

The main reasons not to allow easy comparison with `==` for some types are that (1) it could potentially hide a very costly operation (e.g. when comparing arrays) or (2) it would be ambiguous (e.g. shallow comparison of references vs deep comparison of content). I think that none of these applies here. External function pointers have fixed size and the address+selector combination is enough to uniquely identify an external function without ambiguity.

Serialization also makes the comparison unnecessarily costly. The `keccak256()` + `abi.encode()` combination allocates memory and this allocation has been notoriously hard to optimize out so far (#12335).

### Example of current behavior
```solidity
contract C {
    function () external externalStorage;

    function comparePtr() public {
        function () external externalLocal1;
        function () external externalLocal2;

        externalLocal1 == externalLocal2;
        externalLocal1 != externalLocal2;

        externalLocal1 == externalStorage;
        externalStorage != externalLocal2;
    }
}
```
```
Error: Operator == not compatible with types function () external and function () external
 --> test.sol:8:9:
  |
8 |         externalLocal1 == externalLocal2;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Operator != not compatible with types function () external and function () external
 --> test.sol:9:9:
  |
9 |         externalLocal1 != externalLocal2;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Operator == not compatible with types function () external and function () external
  --> test.sol:11:9:
   |
11 |         externalLocal1 == externalStorage;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Operator != not compatible with types function () external and function () external
  --> test.sol:12:9:
   |
12 |         externalStorage != externalLocal2;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Specification
`f == g` and `f != g` should not cause a compilation error if `f` and `g` are external functions (or variables of an external function type) with the same signature. The comparison should always return `false` if they belong to different contracts. If should return `true` if and only if both the address and the selector are identical.

## Backwards Compatibility
The change is fully backwards-compatible because such comparisons are currently disallowed and result in a compiler error.

nishant-sachdeva:
Closes #12411

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 Language Features:
  * General: Support ``ContractName.functionName`` for ``abi.encodeCall``, in addition to external function pointers.
+ * General: Add equality-comparison operators for external function types.
 
 
 Compiler Features:
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -3018,8 +3018,18 @@ TypeResult FunctionType::binaryOperatorResult(Token _operator, Type const* _othe
 	if (_other->category() != category() || !(_operator == Token::Equal || _operator == Token::NotEqual))
 		return nullptr;
 	FunctionType const& other = dynamic_cast<FunctionType const&>(*_other);
-	if (kind() == Kind::Internal && other.kind() == Kind::Internal && sizeOnStack() == 1 && other.sizeOnStack() == 1)
+	if (kind() == Kind::Internal && sizeOnStack() == 1 && other.kind() == Kind::Internal && other.sizeOnStack() == 1)
 		return commonType(this, _other);
+	else if (
+		kind() == Kind::External &&
+		sizeOnStack() == 2 &&
+		!bound() &&
+		other.kind() == Kind::External &&
+		other.sizeOnStack() == 2 &&
+		!other.bound()
+	)
+		return commonType(this, _other);
+
 	return nullptr;
 }
 
--- a/libsolidity/codegen/ExpressionCompiler.cpp
+++ b/libsolidity/codegen/ExpressionCompiler.cpp
@@ -2278,12 +2278,29 @@ void ExpressionCompiler::appendAndOrOperatorCode(BinaryOperation const& _binaryO
 
 void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const& _type)
 {
-	solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
 	if (_operator == Token::Equal || _operator == Token::NotEqual)
 	{
-		if (FunctionType const* funType = dynamic_cast<decltype(funType)>(&_type))
+		FunctionType const* functionType = dynamic_cast<decltype(functionType)>(&_type);
+		if (functionType && functionType->kind() == FunctionType::Kind::External)
 		{
-			if (funType->kind() == FunctionType::Kind::Internal)
+			solUnimplementedAssert(functionType->sizeOnStack() == 2, "");
+			m_context << Instruction::SWAP3;
+
+			m_context << ((u256(1) << 160) - 1) << Instruction::AND;
+			m_context << Instruction::SWAP1;
+			m_context << ((u256(1) << 160) - 1) << Instruction::AND;
+			m_context << Instruction::EQ;
+			m_context << Instruction::SWAP2;
+			m_context << ((u256(1) << 32) - 1) << Instruction::AND;
+			m_context << Instruction::SWAP1;
+			m_context << ((u256(1) << 32) - 1) << Instruction::AND;
+			m_context << Instruction::EQ;
+			m_context << Instruction::AND;
+		}
+		else
+		{
+			solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
+			if (functionType && functionType->kind() == FunctionType::Kind::Internal)
 			{
 				// We have to remove the upper bits (construction time value) because they might
 				// be "unknown" in one of the operands and not in the other.
@@ -2291,13 +2308,14 @@ void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const& 
 				m_context << Instruction::SWAP1;
 				m_context << ((u256(1) << 32) - 1) << Instruction::AND;
 			}
+			m_context << Instruction::EQ;
 		}
-		m_context << Instruction::EQ;
 		if (_operator == Token::NotEqual)
 			m_context << Instruction::ISZERO;
 	}
 	else
 	{
+		solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
 		bool isSigned = false;
 		if (auto type = dynamic_cast<IntegerType const*>(&_type))
 			isSigned = type->isSigned();
--- a/libsolidity/codegen/YulUtilFunctions.cpp
+++ b/libsolidity/codegen/YulUtilFunctions.cpp
@@ -29,6 +29,7 @@
 #include <libsolutil/FunctionSelector.h>
 #include <libsolutil/Whiskers.h>
 #include <libsolutil/StringUtils.h>
+#include <libsolidity/ast/TypeProvider.h>
 
 using namespace std;
 using namespace solidity;
@@ -4545,6 +4546,34 @@ string YulUtilFunctions::externalCodeFunction()
 		)")
 		("functionName", functionName)
 		("allocateArray", allocateMemoryArrayFunction(*TypeProvider::bytesMemory()))
+		.render();
+	});
+}
+
+std::string YulUtilFunctions::externalFunctionPointersEqualFunction()
+{
+	std::string const functionName = "externalFunctionPointersEqualFunction";
+	return m_functionCollector.createFunction(functionName, [&]() {
+		return util::Whiskers(R"(
+			function <functionName>(
+				leftAddress,
+				leftSelector,
+				rightAddress,
+				rightSelector
+			) -> result {
+				result := and(
+					eq(
+						<addressCleanUpFunction>(leftAddress), <addressCleanUpFunction>(rightAddress)
+					),
+					eq(
+						<selectorCleanUpFunction>(leftSelector), <selectorCleanUpFunction>(rightSelector)
+					)
+				)
+			}
+		)")
+		("functionName", functionName)
+		("addressCleanUpFunction", cleanupFunction(*TypeProvider::address()))
+		("selectorCleanUpFunction", cleanupFunction(*TypeProvider::uint(32)))
 		.render();
 	});
 }
--- a/libsolidity/codegen/YulUtilFunctions.h
+++ b/libsolidity/codegen/YulUtilFunctions.h
@@ -522,6 +522,9 @@ public:
 	/// Signature: (address) -> mpos
 	std::string externalCodeFunction();
 
+	/// @return the name of a function that that checks if two external functions pointers are equal or not
+	std::string externalFunctionPointersEqualFunction();
+
 private:
 /// @returns the name of a function that copies a struct from calldata or memory to storage
 	/// signature: (slot, value) ->
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
@@ -799,13 +799,8 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
 
 	if (TokenTraits::isCompareOp(op))
 	{
-		if (auto type = dynamic_cast<FunctionType const*>(commonType))
-		{
-			solAssert(op == Token::Equal || op == Token::NotEqual, "Invalid function pointer comparison!");
-			solAssert(type->kind() != FunctionType::Kind::External, "External function comparison not allowed!");
-		}
-
 		solAssert(commonType->isValueType(), "");
+
 		bool isSigned = false;
 		if (auto type = dynamic_cast<IntegerType const*>(commonType))
 			isSigned = type->isSigned();
@@ -813,8 +808,25 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
 		string args = expressionAsType(_binOp.leftExpression(), *commonType, true);
 		args += ", " + expressionAsType(_binOp.rightExpression(), *commonType, true);
 
+		auto functionType = dynamic_cast<FunctionType const*>(commonType);
+		solAssert(functionType ? (op == Token::Equal || op == Token::NotEqual) : true, "Invalid function pointer comparison!");
+
 		string expr;
-		if (op == Token::Equal)
+
+		if (functionType && functionType->kind() ==  FunctionType::Kind::External)
+		{
+			solUnimplementedAssert(functionType->sizeOnStack() == 2, "");
+			expr = m_utils.externalFunctionPointersEqualFunction() +
+				"(" +
+				IRVariable{_binOp.leftExpression()}.part("address").name() + "," +
+				IRVariable{_binOp.leftExpression()}.part("functionSelector").name() + "," +
+				IRVariable{_binOp.rightExpression()}.part("address").name() + "," +
+				IRVariable{_binOp.rightExpression()}.part("functionSelector").name() +
+				")";
+			if (op == Token::NotEqual)
+				expr = "iszero(" + expr + ")";
+		}
+		else if (op == Token::Equal)
 			expr = "eq(" + move(args) + ")";
 		else if (op == Token::NotEqual)
 			expr = "iszero(eq(" + move(args) + "))";
--- a/test/libsolidity/semanticTests/functionTypes/comparison_operator_for_external_function_cleans_dirty_bits.sol
+++ b/test/libsolidity/semanticTests/functionTypes/comparison_operator_for_external_function_cleans_dirty_bits.sol
@@ -1,0 +1,17 @@
+contract C {
+  function g() external {}
+  function comparison_operators_for_external_function_pointers_with_dirty_bits() external returns (bool) {
+        function() external g_ptr_dirty = this.g;
+        assembly {
+                g_ptr_dirty.address := or(g_ptr_dirty.address, shl(160, sub(0,1)))
+                g_ptr_dirty.selector := or(g_ptr_dirty.selector, shl(32, sub(0,1)))
+        }
+        function() external g_ptr = this.g;
+        return g_ptr == g_ptr_dirty;
+  }
+}
+// ====
+// compileViaYul: also
+// EVMVersion: >=constantinople
+// ----
+// comparison_operators_for_external_function_pointers_with_dirty_bits() -> true
--- a/test/libsolidity/semanticTests/functionTypes/comparison_operators_for_external_functions.sol
+++ b/test/libsolidity/semanticTests/functionTypes/comparison_operators_for_external_functions.sol
@@ -1,0 +1,81 @@
+contract C {
+    function f() external {}
+    function g() external {}
+    function h() pure external {}
+    function i() view external {}
+
+    function comparison_operators_for_external_functions() public returns (bool) {
+        assert(
+            this.f != this.g &&
+            this.f != this.h &&
+            this.f != this.i &&
+
+            this.g != this.h &&
+            this.g != this.i &&
+
+            this.h != this.i &&
+
+            this.f == this.f &&
+            this.g == this.g &&
+            this.h == this.h &&
+            this.i == this.i
+        );
+        return true;
+    }
+
+    function comparison_operators_for_local_external_function_pointers() public returns (bool) {
+        function () external f_local = this.f;
+        function () external g_local = this.g;
+        function () external pure h_local = this.h;
+        function () external view i_local = this.i;
+
+        assert(
+            f_local == this.f &&
+            g_local == this.g &&
+            h_local == this.h &&
+            i_local == this.i &&
+
+            f_local != this.g &&
+            f_local != this.h &&
+            f_local != this.i &&
+
+            g_local != this.f &&
+            g_local != this.h &&
+            g_local != this.i &&
+
+            h_local != this.f &&
+            h_local != this.g &&
+            h_local != this.i &&
+
+            i_local != this.f &&
+            i_local != this.g &&
+            i_local != this.h
+        );
+
+        assert(
+            f_local == f_local &&
+            f_local != g_local &&
+            f_local != h_local &&
+            f_local != i_local
+        );
+
+        assert(
+            g_local == g_local &&
+            g_local != h_local &&
+            g_local != i_local
+        );
+
+        assert(
+            h_local == h_local &&
+            i_local == i_local &&
+            h_local != i_local
+        );
+
+        return true;
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// comparison_operators_for_external_functions() -> true
+// comparison_operators_for_local_external_function_pointers() -> true
--- a/test/libsolidity/syntaxTests/functionTypes/comparison_operator_for_external_functions_with_extra_gas_slots.sol
+++ b/test/libsolidity/syntaxTests/functionTypes/comparison_operator_for_external_functions_with_extra_gas_slots.sol
@@ -1,0 +1,12 @@
+contract C {
+    function external_test_function() external {}
+    function comparison_operator_for_external_function_with_extra_slots() external returns (bool) {
+        return (
+            (this.external_test_function{gas: 4} == this.external_test_function) &&
+            (this.external_test_function{gas: 4} == this.external_test_function{gas: 4})
+        );
+    }
+}
+// ----
+// TypeError 2271: (193-259): Operator == not compatible with types function () external and function () external
+// TypeError 2271: (277-351): Operator == not compatible with types function () external and function () external
--- a/test/libsolidity/syntaxTests/functionTypes/comparison_operators_between_internal_and_external_function_pointers.sol
+++ b/test/libsolidity/syntaxTests/functionTypes/comparison_operators_between_internal_and_external_function_pointers.sol
@@ -1,0 +1,23 @@
+contract C {
+    function external_test_function() external {}
+    function internal_test_function() internal {}
+
+    function comparison_operator_between_internal_and_external_function_pointers() external returns (bool) {
+        function () external external_function_pointer_local = this.external_test_function;
+        function () internal internal_function_pointer_local = internal_test_function;
+
+        assert(
+            this.external_test_function == external_function_pointer_local &&
+            internal_function_pointer_local == internal_test_function
+        );
+        assert(
+            internal_function_pointer_local != external_function_pointer_local &&
+            internal_test_function != this.external_test_function
+        );
+
+        return true;
+    }
+}
+// ----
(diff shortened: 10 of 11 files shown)


chriseth:
Looks good so far! Please also experiment a bit with comparing function pointers of non-identical type, i.e. compare view, pure, payable and non-payable function pointers to different function pointers.

libsolidity/ast/Types.cpp
```diff
@@ -3019,8 +3019,18 @@ TypeResult FunctionType::binaryOperatorResult(Token _operator, Type const* _othe
 	if (_other->category() != category() || !(_operator == Token::Equal || _operator == Token::NotEqual))
 		return nullptr;
 	FunctionType const& other = dynamic_cast<FunctionType const&>(*_other);
-	if (kind() == Kind::Internal && other.kind() == Kind::Internal && sizeOnStack() == 1 && other.sizeOnStack() == 1)
+	if (kind() == Kind::Internal && sizeOnStack() == 1 && other.kind() == Kind::Internal && other.sizeOnStack() == 1)
 		return commonType(this, _other);
```
hrkrshnn, line 3022: Why this change?

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2275,7 +2275,6 @@ void ExpressionCompiler::appendAndOrOperatorCode(BinaryOperation const& _binaryO
 
 void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const& _type)
 {
-	solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
```
hrkrshnn, line 2281: Might still be useful to keep this assert, everywhere except for external function types.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -800,9 +800,17 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
 	if (TokenTraits::isCompareOp(op))
 	{
 		if (auto type = dynamic_cast<FunctionType const*>(commonType))
+			// what's this common Type class ?
```
hrkrshnn: https://github.com/ethereum/solidity/blob/6849774bd758d0e984ed676eb49ed47c38aec0e7/libsolidity/ast/ASTAnnotations.h#L309-L314

hrkrshnn: https://github.com/ethereum/solidity/blob/6849774bd758d0e984ed676eb49ed47c38aec0e7/libsolidity/analysis/TypeChecker.cpp#L1704

Where it gets set.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -800,9 +800,17 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
...
+			// with this CommonType, they're not allowing ==, != operators
+			// but they're being dealt with later on
 		{
 			solAssert(op == Token::Equal || op == Token::NotEqual, "Invalid function pointer comparison!");
+			// looks like ==, != operators are not allowed on commonType function pointers
+			// will have to see if external function pointers are a type of commonType pointers only
 			solAssert(type->kind() != FunctionType::Kind::External, "External function comparison not allowed!");
+			// we want to enable comparison operators for External Functions
+			// So, this solAssert block needs to go
```
hrkrshnn: Yes.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -800,9 +800,17 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
...
+			// but another question :: Why are they not allowing any ==, != comaprisons to go beyond this point
```
hrkrshnn: Effectively, this code expects that the stack slots for the type to be just `1`. For external functions, it would be 2.

https://github.com/ethereum/solidity/blob/6849774bd758d0e984ed676eb49ed47c38aec0e7/libsolidity/ast/Types.cpp#L3116-L3121

You use the `.part(...)` to access the stack type. For an example, see

https://github.com/ethereum/solidity/blob/6849774bd758d0e984ed676eb49ed47c38aec0e7/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L1746

Two external function types are equal if both the parts / stack slots are the same.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2288,6 +2287,27 @@ void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const&
...
+				m_context << Instruction::SWAP3;
+				// now the stack looks like <fun2_selector> <fun1_selector> <fun2_address> <fun1_address>
+				m_context << Instruction::EQ;
+				// now the stack looks like <fun2_selector> <fun1_selector> <TRUE/FALSE>;
+				m_context << Instruction::SWAP2;
+				// now the stack looks like <TRUE/FALSE> <fun1_selector> <fun2_selector>;
+				m_context << Instruction::EQ;
+				// now the stack looks like <TRUE/FALSE> <TRUE/FALSE>
+				m_context << Instruction::AND;
+				// now the stack looks lke <TRUE/FALSE>
```
hrkrshnn: Looks right. Didn't verify how the stack looks at the beginning.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2288,6 +2287,27 @@ void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const&
...
+
+				// TO DO :: Figure out what the left shift operations are for
+				// Check and see if we need these operations in our case as well.
```
hrkrshnn: Even though it takes two stack slots, the entire data can be fit into a single slot. It's a selector `4` bytes and address `20` bytes--24 bytes in total. A stack slot is 32 bytes. That's why, often, you see some shifts related to function types.

libsolidity/ast/Types.cpp
nishant-sachdeva: Simply a design choice. Prior to this , the order was such that a test was done for both function pointers, followed by other tests in similar fashion. I felt the change to all tests for a single pointer, and then all tests for the next pointer would make it more readable. 

And I wanted to keep this consistent across both the if() and else if() block. So made this change. 
I can revert it if needed

libsolidity/codegen/ExpressionCompiler.cpp
nishant-sachdeva: That makes sense. But will we need to perform this shift here ? 
I have a feeling that if the (32 + 8) bytes of unoccupied slot space ( assuming there are two stack slots allocated for the selector and address ) is garbage values then we'll need the shift. If , however, they've been initialized to 0 ( or null etc) , then we could do without the shift operator as well.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2288,6 +2287,27 @@ void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const&
...
 				m_context << ((u256(1) << 32) - 1) << Instruction::AND;
 			}
+			else if (funType->kind() == FunctionType::Kind::External)
+			{
+				// TO DO::
+				// The stack will look like <fun1_addr> <fun1_selector> <fun2_addr> <fun2_selector>
+				// In order to compare you need to swap fun2_selector with fun1_addr to move fun1_addr up.
+				// You can use swap3 for that.
+				m_context << Instruction::SWAP3;
+				// now the stack looks like <fun2_selector> <fun1_selector> <fun2_address> <fun1_address>
```
chriseth: Can you reduce the number of these comments a bit?

nishant-sachdeva: Yes definitely. These are merely for my reference till the main code gets done. I'll remove them

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -4548,3 +4548,27 @@ string YulUtilFunctions::externalCodeFunction()
...
+	// to be considered template code for now. Not fully sure if the YulSyntax used here is correct
+	std::string expr = "";
+	return expr;
+	// here we shall write code that returns our function
+	string const functionName = "checkExternalFunctionsEqual";
+
+	return m_functionCollector.createFunction(functionName, [&]() {
+		return util::Whiskers(R"(
+			function <functionName>(leftAddress, leftSelector, rightAddress, rightSelector) -> result {
+				// result := and(isequal(leftAddress, rightAddress) , isequal(leftSelector, rightSelector));
```
hrkrshnn:
```suggestion
				// result := and(eq(leftAddress, rightAddress) , eq(leftSelector, rightSelector));
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2291,6 +2294,20 @@ void ExpressionCompiler::appendCompareOperatorCode(Token _operator, Type const&
...
+
+				// Assumption :: Stack looks like <fun1_addr> <fun1_selector> <fun2_addr> <fun2_selector>
+				// TO DO :: Check if we need to add some verification regarding the same
+
+				m_context << Instruction::SWAP3;
+				m_context << Instruction::EQ;
+				m_context << Instruction::SWAP2;
+				m_context << Instruction::EQ;
+				m_context << Instruction::AND;
+			}
```
ekpyron: Won't this fall through and emit an additional ``EQ`` below?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -4548,3 +4548,27 @@ string YulUtilFunctions::externalCodeFunction()
 		.render();
 	});
 }
+
+std::string YulutilFunctions::checkExternalFunctionsEqual(
+		AddressType const leftAddr, 
+		IntegerType const* leftSelector, 
+		AddressType const rightAddr, 
+		IntegerType const* rightSelector
```
hrkrshnn: This may be unnecessary.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -799,22 +799,24 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
...
 		args += ", " + expressionAsType(_binOp.rightExpression(), *commonType, true);
 
+		auto functionType = dynamic_cast<FunctionType const*>(commonType);
+		solAssert(functionType && (op == Token::Equal || op == Token::NotEqual), "Invalid function pointer comparison!");
+
 		string expr;
-		if (op == Token::Equal)
+
+		if (functionType->kind() ==  FunctionType::Kind::External) {}
+			expr = checkExternalFunctionsEqual( /*leftAddress, rightAddress, leftSelector, rightSelector*/);
```
hrkrshnn: Something like 
```cpp
expr = checkExternalFunctionsEqual(...) << "(" <<
    IRVariable{leftExpression}.part("address").name()  << ", " <<
    IRVariable{rightExpression}.part("address").name() << ", " << 
    IRVariable{leftExpression}.part("functionSelector").name() << ", " << 
    .... << ")";
```

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2278,23 +2278,36 @@ void ExpressionCompiler::appendAndOrOperatorCode(BinaryOperation const& _binaryO
...
 {
-	solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
+	// solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
+	// "Might still be useful to keep this assert, everywhere except for external function types." --Comment from issue
+	// TO DO :: check for the implications of this comment. And see if and where to we retain this solAssert
+
 	if (_operator == Token::Equal || _operator == Token::NotEqual)
 	{
-		if (FunctionType const* funType = dynamic_cast<decltype(funType)>(&_type))
+		if (_operator == Token::Equal || _operator == Token::NotEqual)
```
ekpyron: This ``if`` is redundant, isn't it?
Apart from that, it looks better now, yes.

test/libsolidity/semanticTests/functionTypes/comparison_operators_for_external_functions.sol
```diff
@@ -0,0 +1,18 @@
...
+        externalLocal1 != externalLocal2;
+
+        externalLocal1 == externalStorage;
+        externalStorage != externalLocal2;
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// comparePtr() ->
```
ekpyron: Might need some adjustment, but maybe something along the lines of the following? Might be nicer to split into several cases, though.

```suggestion
contract C {
    function f() external {}
    function g() external {}

    function check() public returns(bool) {
        assert(f != g && f == f && g == g);
        function () external f_local = f;
        function () external g_local = g;
        assert(f != g_local && f_local != g && f_local == f && g_local == g);
        assert(f_local == f_local && f_local != g_local);
        assert(g_local != f_local && g_local == g_local);
        return true;
    }
}
// ====
// compileViaYul: also
// ----
// check() -> 1
```

libsolidity/codegen/ExpressionCompiler.cpp
nishant-sachdeva: Indeed yes, it's redundant, I've removed it in the latest push. 
Thanks for pointing it out!

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -522,6 +522,9 @@ class YulUtilFunctions
 	/// Signature: (address) -> mpos
 	std::string externalCodeFunction();
 
+	/// @return Yul Code to check if two external functions pointers are equal or not
+	std::string checkExternalFunctionsEqual();
```
ekpyron: I guess the canonical naming scheme for these would mean this should rather be called something like
```suggestion
	std::string externalFunctionPointerEqualFunction();
```
Not that that's overly nice, but we should at least stick to the schema of ``whateverFunction``

ekpyron:
As far as I can see, this mainly needs better tests now.

test/libsolidity/semanticTests/functionTypes/comparison_operators_for_external_functions.sol
```diff
@@ -0,0 +1,77 @@
+contract C {
+    function () external f;
+    function () external g;
+    function () pure external h;
+    function () view external i;
```
ekpyron: Those should be proper functions, not function-pointer state variables, i.e.
```
function f() external {}
...
```

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 Language Features:
  * General: Support ``ContractName.functionName`` for ``abi.encodeCall``, in addition to external function pointers.
+ * Comparison operators allowed for external function types.
```
Marenz: should be present tense, not past tense (`allowed`)

test/libsolidity/semanticTests/functionTypes/comparison_operators_for_external_functions.sol
```diff
@@ -0,0 +1,77 @@
...
+            h_local != i_local
+        );
+
+        return true;
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// check_comparison_operators() -> FAILURE, hex"4e487b71", 0x01
```
ekpyron:
```suggestion
// check_comparison_operators() -> true
```
CI now seems to agree that the test passes :-).

ekpyron:
Another set of tests that we should add is to test function types with additional slots like ``gas``, e.g.:
```
contract C {
        function f() external {}
        function test() external returns (bool) {
                return this.f{gas: 4} == this.f;
        }
}
```
We don't necessarily need to properly support such comparisons just yet, but we should at least make sure that they either work properly or raise an appropriate syntax error.

test/libsolidity/syntaxTests/functionTypes/comparison_operator_for_external_functions_with_extra_gas_slots.sol
```diff
@@ -0,0 +1,8 @@
+contract C {
+    function external_test_function() external {}
+    function comparison_operator_for_external_function_with_extra_slots() external returns (bool) {
+        return this.external_test_function{gas: 4} == this.external_test_function;
```
ekpyron: We could also test that ``this.external_test_function{gas: 4} == this.external_test_function{gas: 4}`` yields another error.
Since the error is not fatal you could just add it to the same test.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2278,21 +2278,27 @@ void ExpressionCompiler::appendAndOrOperatorCode(BinaryOperation const& _binaryO
...
+		}
+		else
+		{
+			if (funType && funType->kind() == FunctionType::Kind::Internal)
 			{
-				// We have to remove the upper bits (construction time value) because they might
-				// be "unknown" in one of the operands and not in the other.
 				m_context << ((u256(1) << 32) - 1) << Instruction::AND;
 				m_context << Instruction::SWAP1;
 				m_context << ((u256(1) << 32) - 1) << Instruction::AND;
```
ekpyron, line 2309: For taking care of the dirty higher-order bits you should be able to do something similar to this here, both for the selector and the address parts. The selector is 4 bytes, i.e. you can mask with ``((u256(1) << 32) - 1)``, the address part is 160 bits, i.e. you can mask with ``((u256(1) << 160) - 1)``.
I also just noticed that you removed the comment above the masking of internal function types, was that intentional?

test/libsolidity/semanticTests/functionTypes/comparison_operator_for_external_function_cleans_dirty_bits.sol
```diff
@@ -0,0 +1,15 @@
+contract C {
+  function g() external {}
+  function comparison_operators_for_external_function_pointers_with_dirty_bits() external returns (bool) {
+        function() external g_ptr_dirty = this.g;
+        assembly {
+                g_ptr_dirty.address := or(g_ptr_dirty.address, shl(160, sub(0,1)))
```
ekpyron, line 6: By the way: we should also probably test this for ``.selector`` (with a shift of ``32`` instead of ``160`` in that case).

test/libsolidity/syntaxTests/functionTypes/comparison_operator_for_external_functions_with_extra_gas_slots.sol
nishant-sachdeva: Added in the latest push

test/libsolidity/semanticTests/functionTypes/comparison_operator_for_external_function_cleans_dirty_bits.sol
nishant-sachdeva: Added in the latest push

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -4548,3 +4548,34 @@ string YulUtilFunctions::externalCodeFunction()
 		.render();
 	});
 }
+
+std::string YulUtilFunctions::externalFunctionPointersEqualFunction(
+	std::string addressCleanUpFunction,
+	std::string selectorCleanUpFunction
```
ekpyron: Don't pass these in as strings - just use ``cleanupFunction(...)`` with the appropriate type below directly - the type will always be the same, no reason to extract it specifically from ``.part(...)``.

ekpyron (changes requested):

ekpyron:
The antlr grammar test failure is probably actually a bug in the antlr grammar - maybe I'll take a look at that myself tomorrow to fix it. The rest of the test failures as far as I can see are due to a mismatching source location for one of the errors, so running ``isoltest`` locally and updating the tests should fix that.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -4548,3 +4548,34 @@ string YulUtilFunctions::externalCodeFunction()
...
+					),
+					eq(
+						<selectorCleanUpFunction>(leftSelector), <selectorCleanUpFunction>(rightSelector)
+					)
+				)
+			}
+		)")
+		("functionName", functionName)
+		("addressCleanUpFunction", addressCleanUpFunction)
+		("selectorCleanUpFunction", selectorCleanUpFunction)
```
ekpyron: To clarify: I meant you can just remove the arguments and use this intead:
```suggestion
		("addressCleanUpFunction", cleanupFunction(*TypeProvider::address()))
		("selectorCleanUpFunction", cleanupFunction(*TypeProvider::uint(32)))
```

test/libsolidity/semanticTests/functionTypes/comparison_operator_for_external_function_cleans_dirty_bits.sol
```diff
@@ -0,0 +1,16 @@
...
+                g_ptr_dirty.selector := or(g_ptr_dirty.selector, shl(32, sub(0,1)))
+        }
+        function() external g_ptr = this.g;
+        return g_ptr == g_ptr_dirty;
+  }
+}
+// ====
+// compileViaYul: also
```
ekpyron, line 14:
```suggestion
// compileViaYul: also
// EVMVersion: >=constantinople
```
This should fix the remaining test failure in CI. Shift opcodes like ``shl`` are not available in ancient EVM versions, so we can't run this test on them. But since the cleanup code is the same across EVM version only checking for newer EVM versions is just fine for this.

ekpyron (commented):
https://github.com/ethereum/solidity/pull/12545 is merged, so this can be rebased.

libsolidity/codegen/ExpressionCompiler.cpp
ekpyron: What about this?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -799,22 +799,33 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
...
 		args += ", " + expressionAsType(_binOp.rightExpression(), *commonType, true);
 
+		auto functionType = dynamic_cast<FunctionType const*>(commonType);
+		solAssert(functionType ? (op == Token::Equal || op == Token::NotEqual) : true, "Invalid function pointer comparison!");
+
 		string expr;
-		if (op == Token::Equal)
+
+		if (functionType && functionType->kind() ==  FunctionType::Kind::External)
+		{
```
ekpyron, line 817: We should probably add another safeguard here like an
```
solUnimplementedAssert(functionType->sizeOnStack() == 2, "");
```

ekpyron (commented):
Looks good, but we should still add, resp. add back, some assertions about the stack size of the types in both old and new code generation. (i.e. https://github.com/ethereum/solidity/pull/12470#discussion_r777238929 and https://github.com/ethereum/solidity/pull/12470#discussion_r786750176)

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -799,22 +799,33 @@ bool IRGeneratorForStatements::visit(BinaryOperation const& _binOp)
...
+			expr = m_utils.externalFunctionPointersEqualFunction() +
+				"(" +
+				IRVariable{_binOp.leftExpression()}.part("address").name() + "," +
+				IRVariable{_binOp.leftExpression()}.part("functionSelector").name() + "," +
+				IRVariable{_binOp.rightExpression()}.part("address").name() + "," +
+				IRVariable{_binOp.rightExpression()}.part("functionSelector").name() +
+				")";
```
ekpyron, line 825: Looking at this once more, it may also have been nice to build this as
```suggestion
			expr = m_utils.externalFunctionPointersEqualFunction(functionType) +
				"(" +
				IRVariable{_binOp.leftExpression()}.commaSeparatedList() +
				"," +
				IRVariable{_binOp.rightExpression()}.commaSeparatedList() +
				")";
```
and then have ``externalFunctionPointersEqualFunction`` assert that the function type has only two stack slots...

But I'd say we can keep things like they are for now and change this, if we ever want to compare multi-slot function types (with additional gas slots, etc. - which I'm not sure we will, since it's probably not particularly useful).

So instead a simple assertion for the type having ``sizeOnStack() == 2`` will also be fine.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -2278,26 +2278,44 @@ void ExpressionCompiler::appendAndOrOperatorCode(BinaryOperation const& _binaryO
...
+			m_context << Instruction::SWAP1;
+			m_context << ((u256(1) << 32) - 1) << Instruction::AND;
+			m_context << Instruction::EQ;
+			m_context << Instruction::AND;
+		}
+		else
+		{
+			if (functionType && functionType->kind() == FunctionType::Kind::Internal)
 			{
+				solAssert(_type.sizeOnStack() == 1, "Comparison of multi-slot types.");
```
ekpyron: This should be outside the inner ``if`` shouldn't it?

nishant-sachdeva: Indeed yes. Changed in the latest push to the PR

ekpyron (dismissed):
In the Yul codegen part, we could probably have spent some more time unifying this with the other cases (e.g. ``args`` is unused in the function type path now - we might have defined auxiliary variables containing cleaned up values in all cases and then defined a tuple comparison function as util function instead of the specialized external function pointer comparison function)...
But at a quick glance I see that that would require some more refactoring of the ``IRVariable`` mechanism, so I'd say we stay with the current solution for now :-).

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 Language Features:
  * General: Support ``ContractName.functionName`` for ``abi.encodeCall``, in addition to external function pointers.
+ * Adds comparison operators for external function types.
```
```suggestion
 * General: Add equality-comparison operators for external function types.
```
If we want to be pedantic :-). ``<`` is also a comparison operator and it's not implemented for external function types here...

ekpyron: Maybe also ``General:`` - but looking through the changelog we're not consistent about that anyways...

nishant-sachdeva: Changed in the latest push.

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 Language Features:
  * General: Support ``ContractName.functionName`` for ``abi.encodeCall``, in addition to external function pointers.
+ * General: Adds equality-comparison operators for external function types.
```
```suggestion
 * General: Add equality-comparison operators for external function types.
```

Since you now quickly changed it once, maybe also loose the ``s`` to be consistent with the style of the other entries, then we can finally merge :-).

ekpyron (approved):

PR #12471, opened 2021-12-30
Properly generated control flows for inherited modifiers

### LINKED ISSUE #11483: Overridden modifiers are ignored in control flow analysis resulting in missing uninitialized storage error.
Example:
```
contract A {
        function f() mod internal returns (uint[] storage) {
        }
        modifier mod() virtual {
                revert();
                _;
        }
}
contract B is A {
        modifier mod() override { _; }
        function g() public {
                f()[0] = 42;
        }
}
```

The analysis only visits ``f`` in ``A``, in which the modifiers reverts, the function never returns, hence there is no uninitialized storage return error.
In ``B``, however, since the modifier is overridden, ``f`` can return, but is not visited again, so there is no error and the uninitialized storage return can actually happen.

Marenz:
fixes #11483

What is done:
* Controlflows are built for modifiers
* They are properly analyzed for pruning
* lots of small changes all over
* functionflows are no longer a vector -> simpler code and unnecessary
* Prune modifiers
* Analyze modifiers in the later step
* Only build flows for modifiers that are actually used

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
 
 
 Solc-Js:
--- a/libsolidity/analysis/ControlFlowBuilder.cpp
+++ b/libsolidity/analysis/ControlFlowBuilder.cpp
@@ -25,19 +25,21 @@ using namespace solidity::langutil;
 using namespace solidity::frontend;
 using namespace std;
 
-ControlFlowBuilder::ControlFlowBuilder(CFG::NodeContainer& _nodeContainer, FunctionFlow const& _functionFlow):
+ControlFlowBuilder::ControlFlowBuilder(CFG::NodeContainer& _nodeContainer, FunctionFlow const& _functionFlow, ContractDefinition const* _contract):
 	m_nodeContainer(_nodeContainer),
 	m_currentNode(_functionFlow.entry),
 	m_returnNode(_functionFlow.exit),
 	m_revertNode(_functionFlow.revert),
-	m_transactionReturnNode(_functionFlow.transactionReturn)
+	m_transactionReturnNode(_functionFlow.transactionReturn),
+	m_contract(_contract)
 {
 }
 
 
 unique_ptr<FunctionFlow> ControlFlowBuilder::createFunctionFlow(
 	CFG::NodeContainer& _nodeContainer,
-	FunctionDefinition const& _function
+	FunctionDefinition const& _function,
+	ContractDefinition const* _contract
 )
 {
 	auto functionFlow = make_unique<FunctionFlow>();
@@ -45,7 +47,7 @@ unique_ptr<FunctionFlow> ControlFlowBuilder::createFunctionFlow(
 	functionFlow->exit = _nodeContainer.newNode();
 	functionFlow->revert = _nodeContainer.newNode();
 	functionFlow->transactionReturn = _nodeContainer.newNode();
-	ControlFlowBuilder builder(_nodeContainer, *functionFlow);
+	ControlFlowBuilder builder(_nodeContainer, *functionFlow, _contract);
 	builder.appendControlFlow(_function);
 
 	return functionFlow;
@@ -297,7 +299,8 @@ bool ControlFlowBuilder::visit(FunctionCall const& _functionCall)
 				_functionCall.expression().accept(*this);
 				ASTNode::listAccept(_functionCall.arguments(), *this);
 
-				m_currentNode->functionCalls.emplace_back(&_functionCall);
+				solAssert(!m_currentNode->functionCall);
+				m_currentNode->functionCall = &_functionCall;
 
 				auto nextNode = newLabel();
 
@@ -321,8 +324,20 @@ bool ControlFlowBuilder::visit(ModifierInvocation const& _modifierInvocation)
 	auto modifierDefinition = dynamic_cast<ModifierDefinition const*>(
 		_modifierInvocation.name().annotation().referencedDeclaration
 	);
-	if (!modifierDefinition) return false;
-	if (!modifierDefinition->isImplemented()) return false;
+
+	if (!modifierDefinition)
+		return false;
+
+	VirtualLookup const& requiredLookup = *_modifierInvocation.name().annotation().requiredLookup;
+
+	if (requiredLookup == VirtualLookup::Virtual)
+		modifierDefinition = &modifierDefinition->resolveVirtual(*m_contract);
+	else
+		solAssert(requiredLookup == VirtualLookup::Static);
+
+	if (!modifierDefinition->isImplemented())
+		return false;
+
 	solAssert(!!m_returnNode, "");
 
 	m_placeholderEntry = newLabel();
@@ -355,8 +370,8 @@ bool ControlFlowBuilder::visit(FunctionDefinition const& _functionDefinition)
 
 	}
 
-	for (auto const& modifier: _functionDefinition.modifiers())
-		appendControlFlow(*modifier);
+	for (auto const& modifierInvocation: _functionDefinition.modifiers())
+		appendControlFlow(*modifierInvocation);
 
 	appendControlFlow(_functionDefinition.body());
 
--- a/libsolidity/analysis/ControlFlowGraph.cpp
+++ b/libsolidity/analysis/ControlFlowGraph.cpp
@@ -44,7 +44,7 @@ bool CFG::visit(ContractDefinition const& _contract)
 		for (FunctionDefinition const* function: contract->definedFunctions())
 			if (function->isImplemented())
 				m_functionControlFlow[{&_contract, function}] =
-					ControlFlowBuilder::createFunctionFlow(m_nodeContainer, *function);
+					ControlFlowBuilder::createFunctionFlow(m_nodeContainer, *function, &_contract);
 
 	return true;
 }
--- a/libsolidity/analysis/ControlFlowGraph.h
+++ b/libsolidity/analysis/ControlFlowGraph.h
@@ -98,8 +98,8 @@ struct CFGNode
 	std::vector<CFGNode*> entries;
 	/// Exit nodes. All CFG nodes to which control flow may continue after this node.
 	std::vector<CFGNode*> exits;
-	/// Function calls done by this node
-	std::vector<FunctionCall const*> functionCalls;
+	/// Function call done by this node
+	FunctionCall const* functionCall = nullptr;
 
 	/// Variable occurrences in the node.
 	std::vector<VariableOccurrence> variableOccurrences;
--- a/libsolidity/analysis/ControlFlowRevertPruner.cpp
+++ b/libsolidity/analysis/ControlFlowRevertPruner.cpp
@@ -81,27 +81,27 @@ void ControlFlowRevertPruner::findRevertStates()
 				if (_node == functionFlow.exit)
 					foundExit = true;
 
-				for (auto const* functionCall: _node->functionCalls)
+				if (auto const* functionCall = _node->functionCall)
 				{
 					auto const* resolvedFunction = ASTNode::resolveFunctionCall(*functionCall, item.contract);
 
-					if (resolvedFunction == nullptr || !resolvedFunction->isImplemented())
-						continue;
-
-					CFG::FunctionContractTuple calledFunctionTuple{
-						findScopeContract(*resolvedFunction, item.contract),
-						resolvedFunction
-					};
-					switch (m_functions.at(calledFunctionTuple))
+					if (resolvedFunction && resolvedFunction->isImplemented())
 					{
-						case RevertState::Unknown:
-							wakeUp[calledFunctionTuple].insert(item);
-							foundUnknown = true;
-							return;
-						case RevertState::AllPathsRevert:
-							return;
-						case RevertState::HasNonRevertingPath:
-							break;
+						CFG::FunctionContractTuple calledFunctionTuple{
+							findScopeContract(*resolvedFunction, item.contract),
+							resolvedFunction
+						};
+						switch (m_functions.at(calledFunctionTuple))
+						{
+							case RevertState::Unknown:
+								wakeUp[calledFunctionTuple].insert(item);
+								foundUnknown = true;
+								return;
+							case RevertState::AllPathsRevert:
+								return;
+							case RevertState::HasNonRevertingPath:
+								break;
+						}
 					}
 				}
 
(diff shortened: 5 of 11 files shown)


libsolidity/analysis/ControlFlowGraph.h
```diff
@@ -98,8 +98,17 @@ struct CFGNode
 	std::vector<CFGNode*> entries;
 	/// Exit nodes. All CFG nodes to which control flow may continue after this node.
 	std::vector<CFGNode*> exits;
-	/// Function calls done by this node
-	std::vector<FunctionCall const*> functionCalls;
+	/// Modifier this node invokes
+	ModifierInvocation const* modifierInvocation;
```
Marenz: should init to `nullptr`

libsolidity/analysis/ControlFlowRevertPruner.cpp
```diff
@@ -32,9 +32,9 @@ namespace
 /// Find the right scope for the called function: When calling a base function,
 /// we keep the most derived, but we use the called contract in case it is a
 /// library function or nullptr for a free function.
```
Marenz: comment needs update

libsolidity/analysis/ControlFlowRevertPruner.cpp
```diff
@@ -49,59 +49,99 @@ ContractDefinition const* findScopeContract(FunctionDefinition const& _function,
...
 {
 	for (auto& [pair, flow]: m_cfg.allFunctionFlows())
-		m_functions[pair] = RevertState::Unknown;
+		m_callables[pair] = RevertState::Unknown;
 
 	findRevertStates();
 	modifyFunctionFlows();
 }
 
+ModifierDefinition const* resolveModifierInvocation(ModifierInvocation const& _invocation, ContractDefinition const& _contract);
```
Marenz: Added because the compiler complained about a missing declaration..

libsolidity/analysis/ControlFlowGraph.h
Marenz: Maybe use a union/variant?

libsolidity/analysis/ControlFlowRevertPruner.cpp
```diff
@@ -110,18 +150,23 @@ void ControlFlowRevertPruner::findRevertStates()
 			}
 		);
 
-		auto& revertState = m_functions[item];
+		auto& revertState = m_callables[item];
 
 		if (foundExit)
-			revertState = RevertState::HasNonRevertingPath;
+		{
+			if (foundPlaceholder)
```
Marenz: Add explaining comment

libsolidity/analysis/ControlFlowGraph.h
Marenz: What do you think @chriseth

libsolidity/analysis/ControlFlowGraph.cpp
```diff
@@ -41,17 +41,44 @@ bool CFG::visit(FunctionDefinition const& _function)
...
 		for (FunctionDefinition const* function: contract->definedFunctions())
 			if (function->isImplemented())
+			{
+				std::set<ModifierInvocation const*> usedModifierInvocations;
+
 				m_functionControlFlow[{&_contract, function}] =
-					ControlFlowBuilder::createFunctionFlow(m_nodeContainer, *function);
+					ControlFlowBuilder::createFunctionFlow(m_nodeContainer, *function, &usedModifierInvocations);
+
+	 			for (auto const* modifierInvocation: usedModifierInvocations)
```
stackenbotten: Coding style error

libsolidity/analysis/ControlFlowBuilder.cpp
```diff
@@ -318,24 +320,10 @@ bool ControlFlowBuilder::visit(ModifierInvocation const& _modifierInvocation)
...
-	appendControlFlow(*modifierDefinition);
-	connect(m_currentNode, m_returnNode);
+	modifierInvocations.insert(&_modifierInvocation);
 
-	m_currentNode = m_placeholderEntry;
-	m_returnNode = m_placeholderExit;
-
-	m_placeholderEntry = nullptr;
-	m_placeholderExit = nullptr;
+	m_currentNode->modifierInvocation = &_modifierInvocation;
```
Marenz: Previously, modifiers were analyzed and the control flow became part of the function, but inheritance was ignored.
Now we just remember the invocation and create for each their own flow.

libsolidity/analysis/ControlFlowGraph.h
```diff
@@ -98,8 +98,17 @@ struct CFGNode
...
-	/// Function calls done by this node
-	std::vector<FunctionCall const*> functionCalls;
+	/// Modifier this node invokes
+	ModifierInvocation const* modifierInvocation = nullptr;
+	/// Function call done by this node
+	FunctionCall const* functionCall = nullptr;
+
+	// TODO possible alternative: have a vector of placeholder nodes in
+	// struct FunctionFlow. Advantages: similar logic/structure, not sure what
+	// else yet.
```
Marenz: remove TODO

libsolidity/analysis/ControlFlowRevertPruner.cpp
```diff
@@ -29,79 +29,128 @@ namespace solidity::frontend
...
+			solAssert(requiredLookup == VirtualLookup::Static, "");
+			return modifier;
+		}
+	}
+
+	return nullptr;
+}
+
+
+CallableDeclaration const* resolveAndValidateCalls(CFGNode const& _node, ContractDefinition const* _contract)
```
Marenz: add comment

test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol
```diff
@@ -18,3 +18,6 @@ contract C {
 
 }
 // ----
+// Warning 5740: (247-275): Unreachable code.
+// Warning 5740: (328-332): Unreachable code.
+// Warning 5740: (348-366): Unreachable code.
```
ekpyron: What exactly is unreachable here and why?

test/libsolidity/syntaxTests/controlFlow/storageReturn/modifier_fine.sol
```diff
@@ -9,12 +9,15 @@ contract C {
...
-    function g(bool flag) revertIfNoReturn() ifFlag(flag) internal view returns(S storage) {
+    function g(bool flag) callAndRevert() ifFlag(flag) internal view returns(S storage) {
         return s;
     }
 
 }
 // ----
+// Warning 5740: (260-295): Unreachable code.
+// Warning 5740: (345-349): Unreachable code.
+// Warning 5740: (384-409): Unreachable code.
```
ekpyron: Also here, I'm not quite sure what should become unreachable in this test.

test/libsolidity/syntaxTests/controlFlow/storageReturn/modifier_err.sol
```diff
@@ -18,5 +18,4 @@ contract C {
     }
 }
 // ----
-// TypeError 3464: (249-258): This variable is of storage pointer type and can be returned without prior assignment, which would lead to undefined behaviour.
-// TypeError 3464: (367-376): This variable is of storage pointer type and can be returned without prior assignment, which would lead to undefined behaviour.
```
ekpyron, line 22: ``f(true)`` *does* return an uninitialized storage pointer and this error is vital - so there is something off in the analysis here.

ekpyron (changes requested):

test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol
Marenz:
```
Warning: Unreachable code.
  --> test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol:12:60:
   |
12 |     function f(bool flag) revertIfNoReturn() internal view {
   |                                                            ^ (Relevant source part starts here and spans across multiple lines).

Warning: Unreachable code.
  --> test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol:15:53:
   |
15 |     function g(bool flag) revertIfNoReturn() ifFlag(flag) internal view {
   |                                                     ^^^^

Warning: Unreachable code.
  --> test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol:15:73:
   |
15 |     function g(bool flag) revertIfNoReturn() ifFlag(flag) internal view {
   |                                                                         ^ (Relevant source part starts here and spans across multiple lines).


```

Marenz: Basically, because the mod always reverts, nothing in the function is reachable. Though I realize it's not strictly correct because we do call the function, we just revert() after. If the revert undoes everything the function did, it is as if it can never actually run.. but I am going to look into this properly

ekpyron: Yeah, I thought so... it has me somewhat worried about the correctness of the analysis, so yeah: probably worth looking into. Generally I'd expect the analysis of a function to start with the modifier invocations and just replace each placeholder with the inner modifier or function flow during analysis, but I also haven't really seen that logic in the code.

libsolidity/analysis/ControlFlowBuilder.cpp
```diff
@@ -297,7 +299,7 @@ bool ControlFlowBuilder::visit(FunctionCall const& _functionCall)
 				_functionCall.expression().accept(*this);
 				ASTNode::listAccept(_functionCall.arguments(), *this);
 
-				m_currentNode->functionCalls.emplace_back(&_functionCall);
+				m_currentNode->functionCall = &_functionCall;
```
ekpyron, line 303:
```suggestion
				yulAssert(!m_currentNode->functionCall, "");``
				m_currentNode->functionCall = &_functionCall;
```
maybe :-)? Then I don't need to check if it's really only ever at most one :-).

Marenz: Maybe not `yul...` but yeah, I'll add an assert

ekpyron: Ok, it creates a new node afterwards right below and it's never set anywhere else, so it's fine... but still wouldn't hurt to assert.

ekpyron: > Maybe not `yul...` but yeah, I'll add an assert

Habit :-).

libsolidity/analysis/ControlFlowBuilder.cpp
```diff
@@ -321,8 +324,20 @@ bool ControlFlowBuilder::visit(ModifierInvocation const& _modifierInvocation)
 	auto modifierDefinition = dynamic_cast<ModifierDefinition const*>(
 		_modifierInvocation.name().annotation().referencedDeclaration
 	);
-	if (!modifierDefinition) return false;
-	if (!modifierDefinition->isImplemented()) return false;
+
+	if (!modifierDefinition)
+		return false;
```
ekpyron, line 329: Is this really still right? I would guess that there may be no modifier definition at the invocation, but after virtual lookup you'll still get one...

```
abstract contract C {
        modifier mod() virtual;
        function f() mod() public {
        }
}

contract D is C {
        modifier mod() override {
                revert();
                _;
        }
}
```
My guess would be that this would report "unreachable code" for ``f`` relative to ``D``, but only if you remove this check here (respectively move it down to after the virtual lookup).

libsolidity/analysis/ControlFlowBuilder.cpp
```diff
@@ -321,8 +324,20 @@ bool ControlFlowBuilder::visit(ModifierInvocation const& _modifierInvocation)
...
+
+	VirtualLookup const& requiredLookup = *_modifierInvocation.name().annotation().requiredLookup;
+
+	if (requiredLookup == VirtualLookup::Virtual)
+		modifierDefinition = &modifierDefinition->resolveVirtual(*m_contract);
```
ekpyron, line 334: Is this actually guaranteed to never return ``nullptr`` :-)?

libsolidity/analysis/ControlFlowBuilder.cpp
ekpyron: Ah, no, only the definition allows for the virtual lookup anyways apparently, so nevermind...

test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol
```diff
@@ -1,5 +1,5 @@
 contract C {
```
ekpyron, line 1: I'm not quite sure what this test case is supposed to test actually... it doesn't have a "local storage variable" as suggested by the path... but well, that has little to do with this PR...

test/libsolidity/syntaxTests/controlFlow/modifiers/modifier_override.sol
```diff
@@ -0,0 +1,17 @@
...
+	modifier mod() virtual {
+		revert();
+		_;
+	}
+}
+contract B is A {
+	modifier mod() override { _; }
+	function g() public {
+		f()[0] = 42;
+	}
```
ekpyron, line 13: This part is actually irrelevant for the test and could be removed... it's a demonstration of this being bad without an error, but for the test case, it's a bit distracting. But also fine with keeping it.

ekpyron (dismissed):
I think this is fine to merge now.

Changelog.md
```diff
@@ -9,6 +9,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Control Flow Graph: Fix modifiers not being analyzed
```
chriseth: This covers way much more than the bug.

chriseth:
Ah nice, very happy to see that we actually already had the loop, so there is no impact at all on runtime!

Changelog.md
chriseth: What about something like `Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.` - does it currently do more than these two?

Marenz: Nope that about covers it. The rest is just refactoring

test/libsolidity/syntaxTests/controlFlow/localStorageVariables/modifier_declaration_fine.sol
Marenz: I think it's from a time when modifiers behaved differently

ekpyron (approved):
Reapproving

PR #12477, opened 2022-01-03
Added specificity to data location.

0xpranay:
Added the specificity that bytes1[] and bytes differ because of padding only in memory data location. Added extra sentence that they are similar when used in storage data location.

### DIFF
--- a/docs/types/reference-types.rst
+++ b/docs/types/reference-types.rst
@@ -153,7 +153,8 @@ third-party string libraries. You can also compare two strings by their keccak25
 concatenate two strings using ``bytes.concat(bytes(s1), bytes(s2))``.
 
 You should use ``bytes`` over ``bytes1[]`` because it is cheaper,
-since ``bytes1[]`` adds 31 padding bytes between the elements. As a general rule,
+since using ``bytes1[]`` in ``memory`` adds 31 padding bytes between the elements. Note that in ``storage``, the
+padding is absent due to tight packing, see :ref:`bytes and string <bytes-and-string>`. As a general rule,
 use ``bytes`` for arbitrary-length raw byte data and ``string`` for arbitrary-length
 string (UTF-8) data. If you can limit the length to a certain number of bytes,
 always use one of the value types ``bytes1`` to ``bytes32`` because they are much cheaper.


docs/types/reference-types.rst
```diff
@@ -153,7 +153,8 @@ third-party string libraries. You can also compare two strings by their keccak25
 concatenate two strings using ``bytes.concat(bytes(s1), bytes(s2))``.
 
 You should use ``bytes`` over ``bytes1[]`` because it is cheaper,
-since ``bytes1[]`` adds 31 padding bytes between the elements. As a general rule,
+since using ``bytes1[]`` in ``memory`` adds 31 padding bytes between the elements. Note that in ``storage`` the
```
hrkrshnn:
```suggestion
since using ``bytes1[]`` in ``memory`` adds 31 byte padding bytes between the elements. Note that in ``storage`` the
```

docs/types/reference-types.rst
```diff
@@ -153,7 +153,8 @@ third-party string libraries. You can also compare two strings by their keccak25
...
+padding is absent due to tight packing and hence both ``bytes1[]`` and ``bytes`` occupy the same space in ``storage``. As a general rule,
```
hrkrshnn: >  hence both ``bytes1[]`` and ``bytes`` occupy the same space in ``storage``

With some caveats, `bytes` would get stored in place if the length is less than 31.

hrkrshnn: Maybe you can remove the "...hence both" line, or add the additional information about in place encoding.

0xpranay: I'm not sure I follow. Do you refer to fact that if bytes' size is at-most 31 bytes, it would be stored in place with the array(length * 2) slot?

hrkrshnn: Yes, see https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html?highlight=stringlibytes-and-string#bytes-and-string.

0xpranay: Yes, I now follow and will make changes.

0xpranay: I have made the changes and added more info about the storage encoding.

docs/types/reference-types.rst
```diff
@@ -153,7 +153,11 @@ third-party string libraries. You can also compare two strings by their keccak25
 concatenate two strings using ``bytes.concat(bytes(s1), bytes(s2))``.
 
 You should use ``bytes`` over ``bytes1[]`` because it is cheaper,
-since ``bytes1[]`` adds 31 padding bytes between the elements. As a general rule,
+since using ``bytes1[]`` in ``memory`` adds 31 byte padding bytes between the elements. Note that in ``storage`` the
+padding is absent due to tight packing. While ``bytes1[]`` is stored in ``storage`` based on general array packing rules,
+the storage encoding of ``bytes`` depends on the length. If the length of ``bytes`` is at most 31 bytes, it is stored adjacent to the array
+length slot ``p``. If the length is more than 32 bytes, ``bytes`` is treated in the same way as an array and it's data is stored at
+``keccak256(p)``. As a general rule,
```
hrkrshnn:
```suggestion
padding is absent due to tight packing, see :ref:`bytes and string <bytes-and-string>`. As a general rule,
```

Sorry, I think it's best not to repeat and just link the section here.

hrkrshnn:
Looks good. I'll squash the commits and merge.

0xpranay:
Okay, thanks for the help!

hrkrshnn (dismissed):

docs/types/reference-types.rst
0xpranay: Okay that's cool too. I initially wanted to add a link, but I thought that maintainers would want to encourage more comprehension and less references.

Nevertheless linking is better in my opinion too.

hrkrshnn (approved):

docs/types/reference-types.rst
hrkrshnn: I think it's generally nice to avoid some duplication.

hrkrshnn:
@0xpranay Thanks for the PR!

PR #12481, opened 2022-01-03
Clarify which functions are added.

### LINKED ISSUE #12474: Add Documentation: Only internal functions from a library that are actually used by a contract are added to the contract bytecode
## Page

https://docs.soliditylang.org/en/v0.8.11/contracts.html#libraries

## Abstract

It is useful for developers to know that only internal functions from a Solidity library that are actually used by a contract are added to the contract bytecode.  With this knowledge developers can be free to create larger Solidity libraries with internal functions without fear that they are bloating the bytecode size of their contracts.

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12474


ekpyron (approved):

mudgen:
Thanks sir!

PR #12482, opened 2022-01-03
LSP: Introduce RequestError exception for easier handling

christianparpart:
Required by #12430.

### DIFF
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -38,6 +38,7 @@
 #include <string>
 
 using namespace std;
+using namespace std::string_literals;
 using namespace std::placeholders;
 
 using namespace solidity::lsp;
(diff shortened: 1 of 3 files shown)


libsolidity/lsp/LanguageServer.cpp
```diff
@@ -268,24 +268,17 @@ bool LanguageServer::run()
...
 }
 
 void LanguageServer::handleInitialize(MessageID _id, Json::Value const& _args)
 {
 	if (m_state != State::Started)
-	{
-		m_client.error(_id, ErrorCode::RequestFailed, "Initialize called at the wrong time.");
-		return;
-	}
+		throw HandlerError(_id, ErrorCode::RequestFailed, "Initialize called at the wrong time.");
```
chriseth: I think we can remove `_id` here and fill it out at the `catch`-site

chriseth: Also isn't it rather `RequestError`?

chriseth: And finally: Why not use `BOOST_THROW_EXCEPTION`?

chriseth: (or create a new `assert` macro)

libsolidity/lsp/Transport.h
```diff
@@ -45,6 +47,29 @@ enum class ErrorCode
 	RequestFailed = -32803
 };
 
+// Error exception used to bail out on errors in the LSP function-call handlers.
```
chriseth: Please document classes using `/**\n*\n*/\n`

libsolidity/lsp/Transport.h
```diff
@@ -45,6 +47,29 @@ enum class ErrorCode
...
+		m_code{_code},
+		m_message{std::move(_message)}
+	{
+	}
+
+	MessageID id() const noexcept { return m_id; }
+	ErrorCode code() const noexcept { return m_code; }
+	std::string const& message() const noexcept { return m_message; }
+
+	char const* what() const noexcept override { return m_message.c_str(); }
```
chriseth: Why not re-use the mechanisms provided by `util::Exception`?

libsolidity/lsp/LanguageServer.cpp
christianparpart: yes I had an issue in finding a good name for it. `RequestError` I thought about too. I'll pick it. wrt `_id`, yes, thx. I didn't use BOOST_THROW_EXCEPTION because it felt overkill, but i can use it.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -260,6 +260,10 @@ bool LanguageServer::run()
 			else
 				m_client.error({}, ErrorCode::ParseError, "\"method\" has to be a string.");
 		}
+		catch (RequestError const& error)
+		{
+			m_client.error(id, error.code(), boost::current_exception_diagnostic_information());
```
chriseth: AFAIK, `diagnostic_information` contains everything (line number, etc) and is rather difficult to understand for regular users. Shouldn't we just use `error.comment() ? *error.comment() : string{}`?

chriseth (commented):
Looks good apart from the minor comment. A future change could also introduce a solAssert-style macro that makes this easier to use.

chriseth (approved):

PR #12487, opened 2022-01-03
Improve Error Reporting of SemVer Parser

vinayman:


liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +275,30 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
 		{
 			c = currentChar();
 			if (v * 10 < v || v * 10 + static_cast<unsigned>(c - '0') < v * 10)
-				BOOST_THROW_EXCEPTION(SemVerError());
+			{
+				BOOST_THROW_EXCEPTION(
+					SemVerError(
+						1465_error,
```
chriseth: You cannot re-use the same error id in multiple locations, just create new (random) error IDs.

liblangutil/SemVerHandler.cpp
```diff
@@ -171,12 +171,22 @@ optional<SemVerMatchExpression> SemVerMatchExpressionParser::parse()
...
+					)
+				);
+			}
 			nextToken();
 		}
 	}
-	catch (SemVerError const&)
+	catch (SemVerError const& e)
 	{
+		BOOST_THROW_EXCEPTION(e);
```
chriseth: I don't think it is good to throw an exception here. Instead, you could return an `std::variant<SemVerError, SemVerMatchExpression>` here and handle it accordingly in the parser to create the error.

Issue #12488, opened 2022-01-04
Check Randomness in Phaser Tests


ekpyron:
The phaser tests use partly fixed partly time-based seeds, so the choice of tests will affect behaviour.
@cameel wanted to look into them on a per-case basis.

I'd suggest to give all of them a fixed seed by default and add an soltest option with a nightly run that changes that behaviour to random seeds (and maybe higher sample sizes to avoid failures then, since runtime will be less of a concern).

PR #12489, opened 2022-01-04
Use fixed seed for flaky phaser test for now.

ekpyron:
Quick fix to be superseded by a more proper solution with https://github.com/ethereum/solidity/issues/12488 eventually.


ekpyron:
Just passed locally with that seed for me at least, so maybe worth a try.

chriseth (approved):

Issue #11690, opened 2021-07-24
Release version 1.0.0


fulldecent:
This is a project management issue.

Not a milestone issue to track a bunch of features that are needed for version 1.

---

Let's release version 1 today.

The same exact code as the latest release. Just renamed as version 1.0.0. Or take the next release that would have been 0.x.y and call that 1.0.0.

---

It's time.

We are not breaking things every release.

There are many contracts deployed already, responsible for hundreds of billions of USD worth of things.

Core developers here are responsible and deliver quality releases reliably. The community is engaged. They are accessible and available at public events. When they *aren't* at public events, people ask how they are doing. Thank you.

This is what a >= 1.0.0 project feels like.

---

There is no risk.

Currently when experimental or breaking features are introduced they are kept behind experimental flags or they are documented.

When new versions of Solidity are released, all support for old versions is dropped. [As a general rule, only the latest release gets security updates.](https://github.com/ethereum/solidity/security/policy#supported-versions)

Releasing version 1.0.0 does not change any of this.

---

There is a huge benefit.

When people release a library or some other code, they want to say "this works on Solidity x.y.z and anything later that is compatible". This is much easier after you release 1.0.0 because you can use Semantic Versioning to compare compatible versions. (This fails for < 1.0.0.)

Two specific problems are: people are hardcoding 0.8.0 version. People should not be required to hardcode Solidity versions to guarantee stability. Another example is Hardhat, it looks great but it doesn't support the latest version of Solidity and produces a warning. Hardhat developers should not need to do anything when a new compatible version of Solidity is released. We definitely don't want tools telling contract developers *not* to use the latest version of Solidity.

---

So again, thank you to everyone here. We did it! This software is amazing. We love you and your work. Let us allow official guarantees of feature stability by releasing 1.0.0 today.

axic:
We are indeed making breaking releases (and many major features we want are still not implemented). Just remember both 0.7.x and 0.8.x came out in 2020/2021, and we plan 0.9.x this year.  Going to 1.0.0, while still following semver, would mean we'd need to release 2.0.0 for a breaking release.

axic:
To give two concrete reasons:
1) the IR,
2) standard library.

**The IR**
The IR (aka new codegen via Yul) is something we started to work on around 2017. The work is progressing much better this year and hopefully it can be released before 2022. We have already found a number of issues how the old codegen worked, and the IR works in these cases differently. This is basically a breaking change.

[...]

PR #12491, opened 2022-01-05
Write down version policy

fulldecent:
Related to https://github.com/ethereum/solidity/issues/11690

---

This PR briefly documents the Solidity / SemVer policy.

### DIFF
--- a/docs/installing-solidity.rst
+++ b/docs/installing-solidity.rst
@@ -9,11 +9,22 @@ Installing the Solidity Compiler
 Versioning
 ==========
 
-Solidity versions follow `semantic versioning <https://semver.org>`_ and in addition to
-releases, **nightly development builds** are also made available.  The nightly builds
-are not guaranteed to be working and despite best efforts they might contain undocumented
-and/or broken changes. We recommend using the latest release. Package installers below
-will use the latest release.
+Solidity versions follow `Semantic Versioning <https://semver.org>`_. In
+addition, patch level releases with major release 0 (i.e. 0.x.y) will not
+contain breaking changes. That means code that compiles with version 0.x.y
+can be expected to compile with 0.x.z where z > y.
+
+In addition to releases, we provide **nightly development builds** with the
+intention of making it easy for developers to try out upcoming features and
+provide early feedback. Note, however, that while the nightly builds are usually
+very stable, they contain bleeding-edge code from the development branch and are
+not guaranteed to be always working. Despite our best efforts, they might
+contain undocumented and/or broken changes that will not become a part of an
+actual release. They are not meant for production use.
+
+When deploying contracts, you should use the latest released version of Solidity. This
+is because breaking changes, as well as new features and bug fixes are introduced regularly.
+We currently use a 0.x version number [to indicate this fast pace of change](https://semver.org/#spec-item-4).
 
 Remix
 =====
@@ -551,10 +562,10 @@ of the current nightly build, but without the ``prerelease`` specifier.
 
 Example:
 
-0. The 0.4.0 release is made.
-1. The nightly build has a version of 0.4.1 from now on.
-2. Non-breaking changes are introduced --> no change in version.
-3. A breaking change is introduced --> version is bumped to 0.5.0.
-4. The 0.5.0 release is made.
+1. The 0.4.0 release is made.
+2. The nightly build has a version of 0.4.1 from now on.
+3. Non-breaking changes are introduced --> no change in version.
+4. A breaking change is introduced --> version is bumped to 0.5.0.
+5. The 0.5.0 release is made.
 
 This behaviour works well with the  :ref:`version pragma <version_pragma>`.


fulldecent:
Ping people interested in #11690: @axic @cameel @0xGrapefruit @jwdonahue

docs/installing-solidity.rst
```diff
@@ -9,11 +9,18 @@ Installing the Solidity Compiler
 Versioning
 ==========
 
-Solidity versions follow `semantic versioning <https://semver.org>`_ and in addition to
-releases, **nightly development builds** are also made available.  The nightly builds
-are not guaranteed to be working and despite best efforts they might contain undocumented
-and/or broken changes. We recommend using the latest release. Package installers below
-will use the latest release.
+Solidity versions mostly follow `Semantic Versioning <https://semver.org>`_.  We make
```
chriseth: Why `mostly`?

chriseth: Ah ok, because of the below. Maybe we should use something like "npm's flavour of semver"?

docs/installing-solidity.rst
```diff
@@ -9,11 +9,18 @@ Installing the Solidity Compiler
...
+an assurance in addition to SemVer that we will not introduce breaking changes into a patch
+level release. That means if you are using 0.x.y, then you can expect everything to work
+the same or better with 0.x.z with z > y.
```
chriseth: What about something like this?
```suggestion
Solidity versions follow `Semantic Versioning <https://semver.org>`_. We follow npm's
extension of Semantic Versioning in that before the 1.0.0 release, patch level releases
do not contain breaking changes. That means code that compiles with version 0.x.y
can be expected to compile with .x.z with z > y as well.
```

docs/installing-solidity.rst
```diff
@@ -9,11 +9,18 @@ Installing the Solidity Compiler
...
+
+The nightly builds are not guaranteed to be working and despite best efforts they might
+contain undocumented and/or broken changes. We recommend using the latest release. Package
+installers below will use the latest release.
+
+When deploying contracts, you should use the latest released version of Solidity. This
```
chriseth, line 25: Now this paragraph reads a bit unmotivated :(

docs/installing-solidity.rst
```diff
@@ -9,11 +9,18 @@ Installing the Solidity Compiler
...
-releases, **nightly development builds** are also made available.  The nightly builds
-are not guaranteed to be working and despite best efforts they might contain undocumented
-and/or broken changes. We recommend using the latest release. Package installers below
-will use the latest release.
+Solidity versions mostly follow `Semantic Versioning <https://semver.org>`_.  We make
+an assurance in addition to SemVer that we will not introduce breaking changes into a patch
+level release. That means if you are using 0.x.y, then you can expect everything to work
+the same or better with 0.x.z with z > y.
+
+The nightly builds are not guaranteed to be working and despite best efforts they might
```
chriseth: I think it would be nice to frame this a bit more positively - nightly builds are usually very stable and we want to encourage people to try them out to get feedback as early as possible.

cameel: It's actually pretty much unchanged from how it was before. It was just moved :) But I agree that it could be framed better. How about this:

```rst
in addition to releases, we provide **nightly development builds** with the intention of making it easy
for developers to try out upcoming features and provide early feedback. Note, however, that while
the nightly builds are usually very stable, they contain bleeding-edge code from the development branch
and are not guaranteed to be always working. Despite our best efforts, they might contain undocumented
and/or broken changes that will not become a part of an actual release. They are not meant for production
use.
```

docs/installing-solidity.rst
fulldecent: The official notes from NPM on SemVer are at https://docs.npmjs.com/about-semantic-versioning

They no longer recommend doing 0.x releases.

docs/installing-solidity.rst
fulldecent: Thank you, added. But also want to remove the npm note because [citation needed]

fulldecent: Also updated with 39c8087d49876cfcfe5f10f02cda71e46795bd7e

docs/installing-solidity.rst
fulldecent: Thank you @cameel  added at f71b3de3d93598f534f15f660235c4bd3c78b5bd.

Looks great

docs/installing-solidity.rst
fulldecent: Addressed below

docs/installing-solidity.rst
fulldecent: This should be fixed with update above at f71b3de3d93598f534f15f660235c4bd3c78b5bd.

Yes?

fulldecent:
@chriseth @cameel I think all items are addressed now. Please lmk

fulldecent:
Hello @chriseth @cameel 

This is rebased onto of HEAD and already squashed. Good to merge! What do you think?

ekpyron:
Unfortunately, we've had some CI issues on ``develop`` for a while, which seems to affect this PR - Chris is on vacation right now, so he can't force-merge over CI failures - so if it's not too much trouble, you could rebase this PR again - by now current ``develop`` should have fixed CI, so then I could approve and merge! (I cannot rebase the PR myself, since I don't have write access to your fork)

fulldecent:
@ekpyron done // checks have passed

ekpyron (approved):

PR #12494, opened 2022-01-05
Adds missing accepts to ImportDirective's SymbolAlias::symbol.

christianparpart:
Found out by accident while working on goto-definition and wondering why I could not catch that AST node's `Identifier` via AST Visitor API.

Required by #12430

### ASTConstVisitor

- [x] ConstantEvaluator: looks at Identifier, but doesn't look like it's having a negative impact
- [x] ControlFlowBuilder: only looks at Variable identifiers and adds them to the current node's variable occurrences
- [x] DeclarationTypeChecker: n/a
- [x] DocStringAnalyzer: n/a
- [x] DocStringTagParser: n/a
- [x] FuncionCallGraph: performs additional checks on VarDecl's and CallableDecl's
- [x] ImmutableValidator: performs additional checks on VarDecl and CallableDecl
- [x] PostTypeChecker: cannot enter its if-condition when traversing into ImportDirective
- [x] Scoper: n/a
- [x] StaticAnalyzer: cannot enter its if-condition as it's not inside function
- [x] SyntaxChecker: n/a
- [x] ASTJsonConverter: should now contain **more** information in the exported JSON (should have a changelog entry then?)
- [x] ContractCompiler: n/a
- [x] ExpressionCompiler: does work with Identifer but is only invoked on `Expression`'s
- [x] IRGeneratorForStatements: same, but is not invoked on `ImportStatement` AST nodes
- [x] formal/Predicate: n/a
- [x] formal/VariableUsage: performs additional checks iff it's a variable that is written to (cannot happen?)

### ASTConstVisitor: unclear uses

- [x] ReferenceResolver: finds decl(s) for the identifier. It does seem to work okay.
- [x] TypeChecker: seems to work. another eye would be helpful here as it's a big handler :-)
- [x] ViewPureChecker: ?
- [x] formal/SMTEncoder: ?

### ASTVisitor

- [x] DeclarationRegistrationHelper: n/a
- [x] test/FirstExpressionExtractor: only checks expressions

### DIFF
--- a/libsolidity/analysis/NameAndTypeResolver.cpp
+++ b/libsolidity/analysis/NameAndTypeResolver.cpp
@@ -571,7 +571,8 @@ bool DeclarationRegistrationHelper::visit(ImportDirective& _import)
 	if (!m_scopes[importee])
 		m_scopes[importee] = make_shared<DeclarationContainer>(nullptr, m_scopes[nullptr].get());
 	m_scopes[&_import] = m_scopes[importee];
-	return ASTVisitor::visit(_import);
+	ASTVisitor::visit(_import);
+	return false; // Do not recurse into child nodes (Identifier for symbolAliases)
 }
 
 bool DeclarationRegistrationHelper::visit(ContractDefinition& _contract)
--- a/libsolidity/ast/AST_accept.h
+++ b/libsolidity/ast/AST_accept.h
@@ -58,13 +58,19 @@ void PragmaDirective::accept(ASTConstVisitor& _visitor) const
 
 void ImportDirective::accept(ASTVisitor& _visitor)
 {
-	_visitor.visit(*this);
+	if (_visitor.visit(*this))
+		for (SymbolAlias const& symbolAlias: symbolAliases())
+			if (symbolAlias.symbol)
+				symbolAlias.symbol->accept(_visitor);
 	_visitor.endVisit(*this);
 }
 
 void ImportDirective::accept(ASTConstVisitor& _visitor) const
 {
-	_visitor.visit(*this);
+	if (_visitor.visit(*this))
+		for (SymbolAlias const& symbolAlias: symbolAliases())
+			if (symbolAlias.symbol)
+				symbolAlias.symbol->accept(_visitor);
 	_visitor.endVisit(*this);
 }
 


chriseth:
As I said on the other PR: Did you check for every single visitor what the consequences of this change are?

christianparpart:
> As I said on the other PR: Did you check for every single visitor what the consequences of this change are?

I'm looking into it right now. I also hope that CI will be green though.

chriseth:
`ReferencesResolver::visit(Identifier const& _identifier)` looks suspicious. It calls `nameFromCurrentScope` which would be wrong for the import statement. Can you tell how `annotation().referencedDeclaration` is assigned for the import?

christianparpart:
> ReferencesResolver::visit(Identifier const& _identifier) looks suspicious. It calls nameFromCurrentScope which would be wrong for the import statement. Can you tell how annotation().referencedDeclaration is assigned for the import?

For ReferenceResolver (just like for SMTEncoder), both did not specially handle ImportDirective before, so I have both overridden and made them not recurse into it. I think that's the right solution, IMHO.

EDIT: Same for ViewPureChecker, and TypeChecker.

libsolidity/analysis/ViewPureChecker.h
```diff
@@ -63,6 +63,9 @@ class ViewPureChecker: private ASTConstVisitor
 	void endVisit(FunctionCall const& _functionCall) override;
 	void endVisit(InlineAssembly const& _inlineAssembly) override;
 
+	/// Do not handle nor recurse into ImportDirective.
```
chriseth: I think we don't need that

libsolidity/analysis/ReferencesResolver.h
```diff
@@ -85,6 +85,9 @@ class ReferencesResolver: private ASTConstVisitor, private yul::ASTWalker
 	bool visit(InlineAssembly const& _inlineAssembly) override;
 	bool visit(Return const& _return) override;
 
+	/// Do not handle nor recurse into ImportDirective.
+	bool visit(ImportDirective const&) override { return false; }
```
chriseth: Do you get an error if we don't have this? I hope we do.

libsolidity/analysis/TypeChecker.h
```diff
@@ -159,6 +159,9 @@ class TypeChecker: private ASTConstVisitor
 	void endVisit(Literal const& _literal) override;
 	void endVisit(UsingForDirective const& _usingForDirective) override;
 
+	/// Do not handle nor recurse into ImportDirective.
```
chriseth: I think this is not needed.

chriseth: Please try to import a library and rename it.

libsolidity/formal/SMTEncoder.h
```diff
@@ -175,6 +175,9 @@ class SMTEncoder: public ASTConstVisitor
 	void endVisit(TryCatchClause const&) override {}
 	bool visit(TryStatement const&) override { return false; }
 
+	/// Do not handle nor recurse into ImportDirective.
```
chriseth: I'm pretty sure this is not needed.

libsolidity/analysis/ReferencesResolver.h
chriseth: So I think this is the only one that we might need, because the resolving works differently for imports.
Please check that we get an error for:
```
function f() pure {}
import {f as g} from "emptyfile.sol"; // where this is an empty file
```

christianparpart: No we don't. This is more a fail-safe to make sure we won't enter by accident.

NameAndTypeResolver seems to be the only user of ReferenceResolver and it's passing a SourceUnit, so this makes sure Identifier from ImportDirective isn't handled (without this I did not get any error though).

libsolidity/analysis/TypeChecker.h
christianparpart: in both cases, I think you meant the comments, or the override at all?

chriseth: I meant the full change (both lines) is not needed.

libsolidity/analysis/ReferencesResolver.h
chriseth: Can you please clarify what "no we don't" refers to? Can you add a relevant test?

Marenz: The above code causes an error with and without that visit

```
Error: Declaration "f" not found in "emptyfile.sol" (referenced as "emptyfile.sol").
 --> test.sol:3:1:
  |
3 | import {f as g} from "emptyfile.sol"; // where this is an empty file
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


```

libsolidity/analysis/TypeChecker.h
Marenz:
```
import {A as B} from "test1.sol";
```
with `A` being an empty library

Does not cause an error, both with and without the visit here.

libsolidity/analysis/ReferencesResolver.h
Marenz: So, for this particular code example, what is happening is that the error I posted is triggered in `performImports` which is happeneing before `resolveNamesAndTypes` is called which internally calls the ReferencesResolver and thus the visit is not happening at all.

I am now checking with a modified example

Marenz: In a modified example where `f` exists as contract in `emptyfile.sol`, it does indeed set the wrong referencedDeclaration

Marenz: And no error is caused either.

Marenz: No, I was wrong. It is setting the correct referencedDeclaration when it does visit it in `ReferencesResolver::visit(Identifier..)`

Marenz: To be clear: `f` is set to the contract found in `emptyfile.sol`

I would continue the analysis now, but it's getting too late now

Marenz: And it is correct because the `ImportDirective` gets the scope of the file it is importing. So can visit as normal here and can remove the `return false;` here.

ekpyron: It would probably be wise to add the cases you investigated here as test cases.

Marenz: I am pretty sure we already have tons of tests for that simple case I investigated, I mainly traced it to make sure it doesn't just work by accident

ekpyron (approved):
Ok, I think we can be reasonably sure that this is fine.

Issue #7261, opened 2019-08-14
Rethink (or maybe even remove?) ``super``


ekpyron:
Given the inheritance changes in 0.6.0 I think we should check where and how ``super`` fits into that and whether we should keep, remove or change it.
E.g. in
```
contract A { function f() virtual {} }
contract B { function f() virtual {} }
contract C is A B { function f() override(A, B) {
  super.f();
} }
```
using ``super`` would defeat the purpose of the whole construction with overrides and unambiguous inheritance...

chriseth:
With the explicit overrides, the burden of checking that inheritance is correct is shifted to the derived contract, but there are still cases where `super` is useful, and these are the ones where we have a shared base:
```
contract I {
  function set() public virtual {}
}
contract A is I {
  uint a;
  function set() public virtual override { a = 1; super.set(); a = 2; }
}
contract B is I {
  uint b;
  function set() public virtual override { b = 1; super.set(); b = 2; }
}
contract X is A, B {
  function set() public override(A, B) { super.set(); }
}
```

These contracts are still valid according to our new rules and all functions will be called in sequence. Without `super`, the derived contract has to call the base functions, but it will not be possible anymore, to call the super function "in between".

chriseth:
Having said that, not sure if there is a use-case for that :)

ekpyron:
You did mean both ``A is I`` and ``B is I``, right? Yeah, this is a point, but to me this always seemed like a really weird construct...

chriseth:
Yes, edited.

Marenz:
Wouldn't `I.set()` also need `virtual` here?

chriseth:
Yes, edited.

ekpyron:
Maybe we should bring this up in the user group as well to get some input on whether this remaining super construction has actual use cases?

ekpyron:
@cameel commented on gitter: ``Maybe making it not look that "constant" would help?``.

For which examples to consider would be ``super()`` or ``super(this)``.

EDIT: ``super(this)`` would make it look like it will become an external call, though, unfortunately...

ekpyron:
In any case: the existence of issues like https://github.com/ethereum/solidity/issues/8354 also indicates that the meaning of ``super`` isn't entirely clear...

cameel:
After our discussion, I think that the syntax could help a little but won't solve the problem on its own. The feature just has complex behavior that needs to be understood.

[...]

PR #12499, opened 2022-01-07
Updated a mistake in inheritance.rst

0xpranay:
Reviewers can you guys take a look here, the doc was explaining that `Base2` contract was calling `Base1` using `super` which does not make any sense as they aren't related in the inheritance graph going up.

### DIFF
--- a/docs/contracts/inheritance.rst
+++ b/docs/contracts/inheritance.rst
@@ -187,7 +187,7 @@ explicitly in the final override, but this function will bypass
         function destroy() public override(Base1, Base2) { super.destroy(); }
     }
 
-If ``Base2`` calls a function of ``super``, it does not simply
+If ``Final`` calls a function of ``super``, it does not simply
 call this function on one of its base contracts.  Rather, it
 calls this function on the next base contract in the final
 inheritance graph, so it will call ``Base1.destroy()`` (note that


ekpyron:
At a quick look, the docs seem correct to me. This is exactly the point of the explanation: ``super`` does *not* refer to the next immediate base contract, but it refers to the next contract in the C3 linearization of the inheritance graph of the most derived contract.

So if the call originated in ``Final``, then ``Base2.destroy()`` will indeed call ``Base1.destroy()``, even though ``Base1`` is not a base of ``Base2``.

I actually already suggested removing ``super`` (https://github.com/ethereum/solidity/issues/7261) because if one is not used to C3 linearizations this is indeed rather confusing, but there are cases that cannot be expressed without this behaviour.

0xpranay:
Hey, thanks for checking them. I've learnt more on C3 Linearisation and now understood the concept. I stand corrected. But I think that for people coming from languages without support for multiple inheritance will have a hard time understanding this. I'm preparing a guide/notes for beginners while I'm learning too so I'm thoroughly checking the docs and stumbled upon this.

If it is okay, I feel linking to this article [Diamond Inheritance](https://forum.openzeppelin.com/t/solidity-diamond-inheritance/2694) will help learners understand clearly what's happening exactly. Thanks!

PR #12500, opened 2022-01-07
Explaining payable(msg.sender)

safwen1:
At this point of the documentation a new Solidity learner will not understand this line without further explanation:
if (!payable(msg.sender).send(amount)) {
It should explain how msg.sender is of type "address" and not "address payable" so it cannot send or receive Ether. Therefore it must be explicitly converted to payable.

### DIFF
--- a/docs/examples/blind-auction.rst
+++ b/docs/examples/blind-auction.rst
@@ -121,6 +121,9 @@ to receive their money - contracts cannot activate themselves.
                 // before `send` returns.
                 pendingReturns[msg.sender] = 0;
 
+                // msg.sender is not of type `address payable` and must be
+                // explicitly converted using `payable(msg.sender)` in order
+                // use the member function `send()`.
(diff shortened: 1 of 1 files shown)


hrkrshnn (dismissed):

docs/examples/blind-auction.rst
```diff
@@ -121,6 +121,9 @@ to receive their money - contracts cannot activate themselves.
                 // before `send` returns.
                 pendingReturns[msg.sender] = 0;
 
+                // msg.sender is not of type payable and must be explicitly 
+                // converted using payable(msg.sender) in order to use the 
+                // function send()
```
hrkrshnn:
```suggestion
                // msg.sender is not of type `address payable` and must be 
                // explicitly converted using `payable(msg.sender)` in order 
                // use the member function `send()`.
```

hrkrshnn (dismissed):

hrkrshnn (approved):

PR #2475, opened 2017-06-28
Add the packaging metadata to build the solidity snap

come-maiz:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

PR #12507, opened 2022-01-10
Refine documentation about externally maintained and unmaintained linux packages.

ekpyron:
Maybe not uncontroversial, but:

I don't think we should have this in the repo, if none of us actively maintains it and the external support cycle is as long as it seems to be. Is there a particular reason why this has to be in our repo? Can't people still maintain snap packages externally, if they want to, without having this in our repo?

### DIFF
--- a/docs/installing-solidity.rst
+++ b/docs/installing-solidity.rst
@@ -118,8 +118,17 @@ The nightly version can be installed using these commands:
     sudo apt-get update
     sudo apt-get install solc
 
-We are also releasing a `snap package <https://snapcraft.io/>`_, which is
-installable in all the `supported Linux distros <https://snapcraft.io/docs/core/install>`_. To
+Furthermore, some Linux distributions provide their own packages. These packages are not directly
+maintained by us, but usually kept up-to-date by the respective package maintainers.
+
+For example, Arch Linux has packages for the latest development version:
+
+.. code-block:: bash
+
+    pacman -S solidity
+
+There is also a `snap package <https://snapcraft.io/solc>`_, however, it is **currently unmaintained**.
+It is installable in all the `supported Linux distros <https://snapcraft.io/docs/core/install>`_. To
 install the latest stable version of solc:
 
 .. code-block:: bash
@@ -139,18 +148,6 @@ with the most recent changes, please use the following:
     but it comes with limitations, like accessing only the files in your ``/home`` and ``/media`` directories.
     For more information, go to `Demystifying Snap Confinement <https://snapcraft.io/blog/demystifying-snap-confinement>`_.
 
-Arch Linux also has packages, albeit limited to the latest development version:
-
-.. code-block:: bash
-
-    pacman -S solidity
-
-Gentoo Linux has an `Ethereum overlay <https://overlays.gentoo.org/#ethereum>`_ that contains a Solidity package.
-After the overlay is setup, ``solc`` can be installed in x86_64 architectures by:
-
-.. code-block:: bash
-
-    emerge dev-lang/solidity
 
 macOS Packages
 ==============


ekpyron:
If we agree to remove this, we should also remove it from the "installing the solidity compiler" docs.

docs/installing-solidity.rst
```diff
@@ -118,27 +118,6 @@ The nightly version can be installed using these commands:
...
-
-    sudo snap install solc --edge
-
-.. note::
-
-    The ``solc`` snap uses strict confinement. This is the most secure mode for snap packages
-    but it comes with limitations, like accessing only the files in your ``/home`` and ``/media`` directories.
-    For more information, go to `Demystifying Snap Confinement <https://snapcraft.io/blog/demystifying-snap-confinement>`_.
-
 Arch Linux also has packages, albeit limited to the latest development version:
```
ekpyron: Actually, I also think we should make it clear that we don't maintain the Archlinux and Gentoo ones ourselves...

cameel:
Pinging @elopio so that he can chime in.

From #11940 looks like the current problem is not that it's broken (it's now up to date in the `edge` channel) but, as I understand it, it's done for the code from `develop` and we need to release 0.8.12 before it can hit `beta` and then `stable`. Also, there was pretty much no feedback from any interested users.

I agree that it looks like having package configuration not maintained by us in the repo did not turn out great and is likely not the way forward for other packages. But in case snap in particular it looks like things are getting under control and it'll start working with 0.8.12 so I wouldn't throw it away now. I'd vote for keeping it as long as it stays actively maintained by @elopio.

> Is there a particular reason why this has to be in our repo? Can't people still maintain snap packages externally, if they want to, without having this in our repo?

From the original PR that added it (https://github.com/ethereum/solidity/pull/2475#issuecomment-313155334):
>> The continuous delivery in build.snapcraft.io and launchpad.net assume that the file is in `{root}/snapcraft.yaml`, or in `{root}/snap/snapcraft.yaml`. We can put it anywhere else, but we would have to use travis or something like that for CD, which is a little more work.
>>
>> So doable, of course, just not as comfortable. It's your repo, so your decision. Let me know if I should move it.

ekpyron:
Alright then, we can wait and if it works out again keep it in here. Closing this for the time being then.

ekpyron:
We should probably still mention in the docs that the snap, archlinux and gentoo packages are externally maintained.

come-maiz:
I'm not using solidity or snapcraft on my job anymore, so being an active maintainer sounds a little far from my current interests. I still like both projects so I can take a look every now at the automated pipelines and then and check if things get stuck. I won't spend a lot of time testing this though. If you don't want to maintain it, instead of "externally maintained" I would call it "unmaintained". "Maintained by the robot uprising syndicate" :)

cameel:
I think it would be a waste to just drop this completely. Maybe we could encourage some people from the community to step in and help maintain this? I mean, it does not necessarily have to sit inside the main repo but would still nice to have it somewhere. People clearly want these packages and work on snap one is mostly done.

This situation looks a bit like the one with translations where we do have some people willing to contribute but not enough so quality is a bit lacking. We could create a separate repo (or a whole org though that might be an overkill) for third-party packages and set up some rules on contribution. Then promote it a bit and only consider a package official if it reaches some level of quality and has at least two active maintainers.

@franzihei What's your opinion on this? Do you think it would be worth the effort?

franzihei:
As discussed in the chat, I think it's best to for now, not remove it entirely yet but rather say it's "unmaintained" and also add the info to relevant others (e.g. Arch) that they are externally maintained (if they are).

Then, if nobody shows interest in snap in the coming months we can still consider removing it from the docs completely. If it's the other way around and it turns out people have interest we can try to help find/coordinate external maintainers.

docs/installing-solidity.rst
```diff
@@ -118,8 +118,24 @@ The nightly version can be installed using these commands:
...
+Furthermore, some Linux distributions provide their own packages. These packages are not directly
+maintained by us, but usually kept up-to-date by the respective package maintainers.
+
+Arch Linux has packages for the latest development version:
+
+.. code-block:: bash
+
+    pacman -S solidity
+
+Gentoo Linux has an `Ethereum overlay <https://overlays.gentoo.org/#ethereum>`_ that contains a Solidity package.
```
ekpyron: https://github.com/coolparadox/ethereum-gentoo-overlay/tree/master/dev-lang/solidity actually doesn't look particularly up-to-date :-)... I'll just keep it in here for now, but we should probably do something about it as well (i.e. either poke the maintainers or mention that it's falling out of date here or remove it)...

docs/installing-solidity.rst
```diff
@@ -118,8 +118,24 @@ The nightly version can be installed using these commands:
...
+After the overlay is setup, ``solc`` can be installed in x86_64 architectures by:
```
cameel:
```suggestion
After the overlay is setup, ``solc`` can be installed on x86_64 architectures by:
```

docs/installing-solidity.rst
cameel: Maybe it would be best to say that it's unmaintained just like the snap? I'd either do that or just drop it.

docs/installing-solidity.rst
```diff
@@ -118,8 +118,24 @@ The nightly version can be installed using these commands:
...
+
+.. code-block:: bash
+
+    emerge dev-lang/solidity
+
+There is also a `snap package <https://snapcraft.io/>`_, however, it is **currently unmaintaned**.
```
cameel:
```suggestion
There is also a `snap package <https://snapcraft.io/solc>`_, however, it is **currently unmaintained**.
```

cameel (dismissed):
I have some tweak suggestions but overall it's fine so I'm approving already.

come-maiz:
> Then, if nobody shows interest in snap in the coming months we can still consider removing it from the docs completely. If it's the other way around and it turns out people have interest we can try to help find/coordinate external maintainers.

Makes sense.
When I was working on this thing, we did a bunch of calls for testing. They helped a lot to get new contributors and have fun with them.
https://forum.snapcraft.io/t/call-for-testing-ipfs/97

docs/installing-solidity.rst
ekpyron: Hm, yeah... I think I'll drop it and create an issue in the repo there, letting them know that we'll add it back, if they update it - does that sound reasonable :-)?

docs/installing-solidity.rst
ekpyron: doh

docs/installing-solidity.rst
ekpyron: https://github.com/coolparadox/ethereum-gentoo-overlay/issues/1

cameel (approved):

docs/installing-solidity.rst
cameel: Sounds good.

PR #12508, opened 2022-01-10
Patch 3

demaupeoucorentin:


ekpyron:
I guess this was supposed to be a PR to https://github.com/Karocyt/solidity-fr instead?

franzihei:
Hey there! This looks like a French translation of the docs? Frist of all, thank you for contributing to this! :)
We have created a new process and org for community translations, and, as @ekpyron already pointed out, the Solidity repo is anyways not the right place for translations. 

There are two options to go from here:
- You can either create a PR to update the translation in https://github.com/Karocyt/solidity-fr
- Or, preferably, you can create a fr-french repo in the [newly created Solidity docs translations GitHub org](https://github.com/solidity-docs), which would also allow more people to see and contribute to the French translation. Here's also more context on the [translation process in general](https://github.com/solidity-docs/translation-guide).

Lmk if you have any questions!

demaupeoucorentin:
Thanks for your comments, I just have mistaken with the correct repo 😅

Sorry for the spent time,
Corentin

PR #12509, opened 2022-01-10
Pin Trident external test to a specific commit to work around failing tests on master

cameel:
#12197 did not pass the trident test but github auto-merged it anyway because that check is not marked as required.

The failing job was [`t_native_test_ext_trident`](https://app.circleci.com/pipelines/github/ethereum/solidity/21505/workflows/b0a5e25d-5bb4-419f-b306-772c9e44ae21/jobs/942648). One of the reasons it fails is that the file I need to ignore to work around https://github.com/sushiswap/trident/issues/283 was moved. ~This PR fixes that specifically.~

I'm not sure if it's a complete fix yet because the failing job had some other failures I'm not seeing locally and they look more serious:
```
1) Constant Product Pool Old
       #mint
         Add liquidity in 16 different ways before swap fees:

      AssertionError: Expected "46000000000000000000" to be equal 0
      + expected - actual

       {
      -  "_hex": "0x00"
      +  "_hex": "0x027e60d44813f80000"
         "_isBigNumber": true
       }
```
The PR is a draft until I ensure that this one is fixed too.

**EDIT**: It's still failing. To work around this I'm switching to an earlier upstream commit that should still work. This required adding support for checking out specific commits in ext tests.

I'm also removing the change for the moved file from this PR because it has not yet been moved in this upstream revision.

### DIFF
--- a/test/externalTests/common.sh
+++ b/test/externalTests/common.sh
@@ -77,12 +77,24 @@ function setup_solc
 function download_project
 {
     local repo="$1"
-    local solcjs_branch="$2"
-    local test_dir="$3"
+    local ref_type="$2"
+    local solcjs_ref="$3"
+    local test_dir="$4"
 
-    printLog "Cloning $solcjs_branch of $repo..."
-    git clone --depth 1 "$repo" -b "$solcjs_branch" "$test_dir/ext"
-    cd ext
+    [[ $ref_type == commit || $ref_type == branch || $ref_type == tag ]] || assertFail
+
+    printLog "Cloning ${ref_type} ${solcjs_ref} of ${repo}..."
+    if [[ $ref_type == commit ]]; then
+        mkdir ext
+        cd ext
+        git init
+        git remote add origin "$repo"
+        git fetch --depth 1 origin "$solcjs_ref"
+        git reset --hard FETCH_HEAD
+    else
+        git clone --depth 1 "$repo" -b "$solcjs_ref" "$test_dir/ext"
+        cd ext
+    fi
     echo "Current commit hash: $(git rev-parse HEAD)"
 }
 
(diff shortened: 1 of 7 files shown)


cameel:
The test finally passed after reverting to a version from 2021-12-16 (i.e. last commit in `master` before the new year). This should be enough to fix `develop` and I'll prepare a proper fix in a separate PR.

ekpyron (approved):

test/externalTests/common.sh
```diff
@@ -77,12 +77,24 @@ function setup_solc
...
+    [[ $ref_type == commit || $ref_type == branch || $ref_type == tag ]] || assertFail
+
+    printLog "Cloning ${ref_type} ${solcjs_ref} of ${repo}..."
+    if [[ $ref_type == commit ]]; then
+        mkdir ext
+        cd ext
+        git init
+        git remote add origin "$repo"
+        git fetch --depth 1 origin "$solcjs_ref"
+        git reset --hard FETCH_HEAD
```
ekpyron, line 93: Is that really the easiest way to fetch a commit? Can't you just also use ``-b`` with commit hashes or sth like that?

ekpyron: But also fine otherwise.

ekpyron: Apparently, this is really the way to do this... I now remember looking into this myself at some point being rather surprised by there not being an easy way to do this...

cameel: You can't, that's the problem. There's no straightforward way to get a shallow clone with just a single commit by its hash.

The only way to simplify it I see is by just doing a normal clone and checking out the commit/branch/tag we want. Most of these repos seem relatively small so maybe it would not matter all that much - npm/yarn is downloading orders of magnitude more data.

This is not *that* much more complicated than a clone though (just normal git commands, not  reaching into any internals) and more future-proof in case some project has a huge monorepo for everything so personally I'd keep it. But I can also use a non-shallow clone if you think it would be better.

ekpyron: It's fine, it's just weird that there's no way to do this more concisely :-).

cameel:
Problem reported in https://github.com/sushiswap/trident/issues/303.

PR #12511, opened 2022-01-10
Deterministic test order in Trident external test

cameel:
Follow-up to #12509. 
~Depends on #12512.~ Merged.

This PR ensures that tests run in the same order on different machines - by sorting them. This is unfortunately not a fix for https://github.com/sushiswap/trident/issues/303 by itself - the sorted order happens to be one where tests fail - but I don't want to make it more complicated by piling on more workarounds. It would be much better to get the problem fixed upstream.

### DIFF
--- a/test/externalTests/trident.sh
+++ b/test/externalTests/trident.sh
@@ -58,9 +58,8 @@ function trident_test
         legacy-optimize-evm+yul
     )
 
-    local selected_optimizer_presets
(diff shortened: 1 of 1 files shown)


cameel:
Tests do not pass due to breakage on `develop`. Fix: #12512.

hrkrshnn (approved):

PR #12516, opened 2022-01-11
Document address members code and codehash. Fixes #12495

### LINKED ISSUE #12495: Document the address.code and address.codeHash properties
## Page

https://docs.soliditylang.org/en/v0.8.11/types.html?highlight=address#address

## Abstract

Features address.code and address.codeHash should be documented here

## Pull request

None yet

fulldecent:


docs/types/value-types.rst
```diff
@@ -337,6 +337,10 @@ on ``call``.
 
 .. _contract_types:
 
+* ``code`` and ``codehash``
+
+You can query the deployed code for any smart contract. Use ``code`` to get the EVM bytecode as a string, which might be empty. Use ``codehash`` get the Keccak-256 hash of that code.
+
```
ekpyron: Looks good, but I guess the right place for it would be a few lines up - probably right before the "All contracts can be converted to ``address`` type..." note.

fulldecent: Good note. Thank you, fixed at 5be61b2ea362b3d84ddd3892dc791fea28e4a768

docs/types/value-types.rst
```diff
@@ -333,6 +333,10 @@ on ``call``.
     All contracts can be converted to ``address`` type, so it is possible to query the balance of the
     current contract using ``address(this).balance``.
```
ekpyron: I actually meant moving it up above this one as well:-).
The first note "It is best to avoid..." still refers to the "* ``call``..." bullet point, but the last note is general and refers to address members in general, so I guess it should come after the new bullet point.

If you move it up once more and squash to a single commit, I'll merge!

fulldecent: Yes, that makes sense.

Done with 78f0be56c3ed7a6c60e39871cdca877644a67afc

ekpyron (approved):

PR #12519, opened 2022-01-12
Update docker images to Z3 4.8.14.

ekpyron:


CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ endif()
 
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
-option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
+option(STRICT_Z3_VERSION "Use the latest version of Z3" OFF)
```
ekpyron: We need to remember to revert this with actually updating the docker image hashes.

leonardoalt: You can also just update the required z3 version

ekpyron: That I'd also do in the next PR.
This PR has to pass its own CI runs relative to the *old* docker images, so it has to build and run against *4.8.13* - but it also has to *build* against 4.8.14, since the docker builds build the compiler once as a "test"... so I guess flipping this here temporarily is the easiest way...

ekpyron: Ah, damn it... I forgot that the archlinux test run will fail here as well :-).
Well, then we'll have to muddle our way through merging this without "override force merge" anyways. I should probably also have disabled the smt tests for the archlinux run earlier, but I can also update this PR later and manually cancel the docker builds.

scripts/deps-ppa/static_z3.sh
```diff
@@ -25,9 +25,9 @@ set -ev
 keyid=70D110489D66E2F6
 email=builds@ethereum.org
 packagename=z3-static
-version=4.8.13
+version=4.8.14
 
-DISTRIBUTIONS="focal groovy hirsute impish"
+DISTRIBUTIONS="focal hirsute impish jammy"
```
ekpyron, line 30: FYI: groovy is EOL and jammy is upcoming (22.04), but already accepts PPA builds.

ekpyron:
@leonardoalt Counterapprove on this one :-)?

leonardoalt (approved):

PR #12520, opened 2022-01-12
Fix ICE when a variable declaration before a referenced struct

### LINKED ISSUE #12505: [TypeChecker] Failed assertion for forward declared struct constant
```
S constant x;struct S{x y;}
```

throws

https://github.com/ethereum/solidity/blob/eee30b6ead0e2917eb75ab410d571a18b57c24ae/libsolidity/ast/Types.cpp#L2194

Marenz:
fixes #12505

### DIFF
--- a/libsolidity/analysis/DeclarationTypeChecker.cpp
+++ b/libsolidity/analysis/DeclarationTypeChecker.cpp
@@ -437,16 +437,14 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
 		type = TypeProvider::withLocation(ref, typeLoc, isPointer);
 	}
 
-	if (
-		_variable.isConstant() &&
-		!dynamic_cast<UserDefinedValueType const*>(type) &&
-		type->containsNestedMapping()
-	)
-		m_errorReporter.fatalDeclarationError(
-			3530_error,
-			_variable.location(),
-			"The type contains a (nested) mapping and therefore cannot be a constant."
-		);
+	if (_variable.isConstant() && !type->isValueType())
+	{
+		bool allowed = false;
+		if (auto arrayType = dynamic_cast<ArrayType const*>(type))
+			allowed = arrayType->isByteArray();
+		if (!allowed)
+			m_errorReporter.fatalTypeError(9259_error, _variable.location(), "Only constants of value type and byte array type are implemented.");
+	}
 
 	_variable.annotation().type = type;
 }
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ Compiler Features:
 Bugfixes:
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the parent contract contains immutable variables.
+ * TypeChecker: Fix ICE when a constant variable declaration forward references a struct.
 
 
 Solc-Js:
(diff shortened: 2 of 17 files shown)


libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -437,16 +437,22 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
+			bool allowed = false;
+			if (auto arrayType = dynamic_cast<ArrayType const*>(type))
+				allowed = arrayType->isByteArray();
+			if (!allowed)
+				m_errorReporter.fatalTypeError(9259_error, _variable.location(), "Constants of non-value type not yet implemented.");
+		}
+
+		solAssert(
+			!dynamic_cast<UserDefinedValueType const*>(type) &&
+			type->containsNestedMapping()
```
ekpyron: This doesn't look right - you meant this negated, right?

ekpyron (changes requested):

libsolidity/analysis/DeclarationTypeChecker.cpp
Marenz: Oh. Haha, yeah. Famous last minute changes. Before it would complain that the error code is in no test, so I figured make it an assert if it can't ever happen anyway.

ekpyron: Or rather ``solAssert(dynamic_cast<UserDefinedValueType const*>(type) || !type->containsNestedMapping());``?
Not sure we need it at all, though, since we assert that it's a value type above anyways.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -437,16 +437,22 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
-			"The type contains a (nested) mapping and therefore cannot be a constant."
+	if (_variable.isConstant())
+	{
+		if (!type->isValueType())
+		{
+			bool allowed = false;
+			if (auto arrayType = dynamic_cast<ArrayType const*>(type))
+				allowed = arrayType->isByteArray();
+			if (!allowed)
+				m_errorReporter.fatalTypeError(9259_error, _variable.location(), "Constants of non-value type not yet implemented.");
```
ekpyron: Actually, since we allow byte arrays, the error message is (and has been) wrong :-). But I guess we can keep it.
You can also change it to sth like "Only constants of value types and byte array types are implemented." or so.

Marenz:
Gah, I should update the tests when I modify the error message.,

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -437,16 +437,14 @@ void DeclarationTypeChecker::endVisit(VariableDeclaration const& _variable)
...
-			_variable.location(),
-			"The type contains a (nested) mapping and therefore cannot be a constant."
-		);
+	if (_variable.isConstant() && !type->isValueType())
+	{
+		bool allowed = false;
+		if (auto arrayType = dynamic_cast<ArrayType const*>(type))
+			allowed = arrayType->isByteArray();
+		if (!allowed)
+			m_errorReporter.fatalTypeError(9259_error, _variable.location(), "Only constants of value type and byte arrays are implemented.");
```
ekpyron: If we want to be pedantic this should probably either be
``Only constants of value and byte array type...``, resp. ``Only constants of value type and byte array type...``
or
``Only constants of value type and byte array constants...``

ekpyron (commented):
I'd say this is fine, so mainly blocked by the Z3 update.

ekpyron (approved):

PR #12523, opened 2022-01-12
Fix wrong error with immutables when base contract c'tor uses return

### LINKED ISSUE #12379: Early return in constructor forbidden if child contract has immutable fields
## Description

The constructor fails to compile if it contains an `return` statement and another contract containing immutable fields inherits from it.

## Environment

- Compiler version: 0.8.10
- Target EVM version (as per compiler settings):  London
- EVM execution environment / backend / blockchain client: JS EVM
- Operating system: web

## Steps to Reproduce
This fails:
```solidity
contract Parent {
    constructor() {
        return;
    }
}

contract Child is Parent {
    uint public immutable baked = 123;
}
```
```
TypeError: Construction control flow ends without initializing all immutable state variables.
  --> tests/remix_tests_test.sol:10:9:
   |
10 |         return;
   |         ^^^^^^^
Note: Not initialized: 
  --> tests/remix_tests_test.sol:15:5:
   |
15 |     uint public immutable baked = 123;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
It's **not** that an early `return` stops child constructors from being initialized. When e.g. a slot variable is initialized in a child contract, it works fine.
<details> <summary> <b> The test example to play with in Remix </b></summary>

```solidity
pragma solidity 0.8.10;
import "remix_tests.sol"; 

contract Parent {
    constructor() {
        return;
    }
}

contract Child is Parent {
    uint public immutable baked = 123;
    uint public stored = 456;
}

contract test {
    function checkSuccess() public {
        Child child = new Child();
        Assert.equal(child.baked(), 123, '');
        Assert.equal(child.stored(), 456, '');
    }
}
```
</details>

Marenz:
fixes #12379

### DIFF
--- a/libsolidity/analysis/ImmutableValidator.cpp
+++ b/libsolidity/analysis/ImmutableValidator.cpp
@@ -29,7 +29,7 @@ void ImmutableValidator::analyze()
 {
 	m_inCreationContext = true;
 
-	auto linearizedContracts = m_currentContract.annotation().linearizedBaseContracts | ranges::views::reverse;
+	auto linearizedContracts = m_mostDerivedContract.annotation().linearizedBaseContracts | ranges::views::reverse;
 
 	for (ContractDefinition const* contract: linearizedContracts)
 		for (VariableDeclaration const* stateVar: contract->stateVariables())
@@ -62,7 +62,7 @@ void ImmutableValidator::analyze()
 			visitCallableIfNew(*modDef);
 	}
 
-	checkAllVariablesInitialized(m_currentContract.location());
+	checkAllVariablesInitialized(m_mostDerivedContract.location());
 }
 
 bool ImmutableValidator::visit(Assignment const& _assignment)
@@ -137,7 +137,7 @@ void ImmutableValidator::endVisit(IdentifierPath const& _identifierPath)
 	if (auto const callableDef = dynamic_cast<CallableDeclaration const*>(_identifierPath.annotation().referencedDeclaration))
 		visitCallableIfNew(
 			*_identifierPath.annotation().requiredLookup == VirtualLookup::Virtual ?
-			callableDef->resolveVirtual(m_currentContract) :
+			callableDef->resolveVirtual(m_mostDerivedContract) :
 			*callableDef
 		);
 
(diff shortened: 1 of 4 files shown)


libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -29,7 +29,7 @@ void ImmutableValidator::analyze()
 {
 	m_inCreationContext = true;
 
-	auto linearizedContracts = m_currentContract.annotation().linearizedBaseContracts | ranges::views::reverse;
+	auto linearizedContracts = m_mostDerivedContract.annotation().linearizedBaseContracts | ranges::views::reverse;
 
 	for (ContractDefinition const* contract: linearizedContracts)
```
ekpyron, line 34: Why do we need to look at the base contracts at all here :-)?

ekpyron: To me it seems like what we currently do analysis-wise, pretty much effectively looks at each constructor in the hierarchy individually anyways - so why not only do just that?

ekpyron: Ah, I guess this is required for the ``read before they are initialized`` checks and virtual lookup. That makes sense.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -161,14 +161,19 @@ void ImmutableValidator::endVisit(Return const& _return)
 bool ImmutableValidator::analyseCallable(CallableDeclaration const& _callableDeclaration)
 {
 	FunctionDefinition const* prevConstructor = m_currentConstructor;
+	ContractDefinition const* prevConstructorContract = m_currentConstructorContract;
```
ekpyron:
```suggestion
	ScopedSaveAndRestore constructorGuard{m_currentConstructorContract, nullptr};
```
and remove the manual restore. Same for the line above...

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -253,7 +259,8 @@ void ImmutableValidator::analyseVariableReference(VariableDeclaration const& _va
 
 void ImmutableValidator::checkAllVariablesInitialized(solidity::langutil::SourceLocation const& _location)
 {
-	for (ContractDefinition const* contract: m_currentContract.annotation().linearizedBaseContracts)
+	for (ContractDefinition const* contract: m_mostDerivedContract.annotation().linearizedBaseContracts | ranges::views::reverse)
+	{
```
ekpyron, line 258: But now here: why do we look at all the contracts in the hierarchy? What would we miss if we only looked at the current/most derived contract alone?

Each contract in the hierarchy will be visited as the "most derived" one eventually, won't it?
And each one has to initialize exactly those immutables in the contract that it is defined it, no more, no less, right?
So I guess we could just visit only that contract relative to those state variables here each time?

ekpyron (commented):
To me it seems like this could generally be simplified, but maybe I'm missing something.
It needs a rebase on top of the Z3 changes anyways, I'll have another look afterwards.

libsolidity/analysis/ImmutableValidator.cpp
ekpyron: Or put differently:
```
contract C {
        uint256 immutable x;
        constructor() {
        }
}
contract D is C {
}
```
currently produces two errors, one relative to ``C``, one relative to ``D``. The one relative to ``D`` is redundant: the issue can only be fixed in ``C`` anyways and ``D`` cannot be compiled without ``C`` anyways.

ekpyron (approved):
Ok, approving this, since it fixes the issue and I don't see how it could cause harm - but we might want to look into simplifying this eventually :-)!

Issue #5956, opened 2019-02-07
Add new external tests


chriseth:
| Project                                                                             | compiler                 | framework                                            | Remarks    |
|-------------------------------------------------------------------------------------|--------------------------|------------------------------------------------------|------------|
| [iExec](https://github.com/iExecBlockchainComputing/PoCo)                           | 0.6.x                    | Truffle                                              |            |
| [Polymath](https://github.com/PolymathNetwork/polymath-core)                        | 0.5.8                    | Truffle                                              |            |
| [erc20-meta-token](https://github.com/horizon-games/ERC20-meta-wrapper)             | 0.7.4                    |                                                      |            |
| [Augur](https://github.com/AugurProject/augur/tree/master/packages/augur-core)      | 0.5.15                   | [eth-tester](https://github.com/ethereum/eth-tester) |            |
| [Kyber](https://github.com/KyberNetwork/smart-contracts)                            | 0.6.6                    | Truffle                                              |            |
| [Veil](https://github.com/veilco/veil-contracts)                                    | 0.4.24                   | Truffle                                              | dead?      |
| [Compound](https://github.com/compound-finance/compound-protocol)                   | ^0.5.16                  | Saddle                                               |            |
| [0x](https://github.com/0xProject/protocol)                                         | various 0.5.x and 0.6.x  | Truffle                                              |            |
| [makerdao/dss](https://github.com/makerdao/dss)                                     | >= 0.5.12                | dapptools                                            |            |
| [Trust Token](https://github.com/trusttoken/smart-contracts)                        | 0.6.10                   | Hardhat                                              |            |
| [delphy](https://github.com/DelphyProject/delphy-contracts)                         | ^0.4.11                  | Truffle                                              | dead?      |
| [gitcoin governance](https://github.com/gitcoinco/governance/tree/main/contracts)   | 0.6.12                   | Brownie                                              |            |
| [StandardBounties](https://github.com/Bounties-Network/StandardBounties)            | 0.5.12                   | Truffle                                              | dead?      |
| [centre-tokens](https://github.com/centrehq/centre-tokens)                          | 0.6.12                   | Truffle                                              |            |
| ~[trident](https://github.com/sushiswap/trident)~                                   | 0.8.x + 0.6.12           | Hardhat                                              | PR: #12197 |
| [solmate](https://github.com/Rari-Capital/solmate)                                  | 0.8.10                   | dapptools                                            |            |
[...]

PR #12530, opened 2022-01-14
External test for Yield Protocol Liquidator v2

cameel:
Part of #5956.

This adds an external test for [Yield Protocol Liquidator v2](https://github.com/yieldprotocol/yield-liquidator-v2).

This one was also straightforward. It needed the `allowUnlimitedContractSize` setting too. This is why one commit is shared with #12529. I put it in both to make the PRs independent. It will disappear when one is merged and the other rebased on top of it.

 I could easily add a similar test for [Yield Protocol Vault V2](https://github.com/yieldprotocol/vault-v2) but I'm not sure if we want it. The code will likely by pretty similar and it's actually a dependency of the liquidator. Let me know if you think I should add it.


hrkrshnn (dismissed):

ekpyron:
FYI: (probably trivial) merge conflicts to be resolved after I merged the first one

cameel:
Yeah, these ext test PRs unfortunately all have trivial conflicts with each other because they all add a new job at the end of CI config.

I could have created them on top of each other or put all of them in a single PR but I wasn't sure which ones will pass review quickly so I decided to do independent PRs instead :)

Now that they have all been accepted, I'll put them on top of each other for easier merge once they get reaccepted.

ekpyron (approved):

Issue #12350, opened 2021-11-30
Solc fails compiling code with inline assembly when locale is set to Turkish


Shungy:
## Description

Solc does not compile code with any inline assembly if the Linux system locale is set to Turkish.

## Environment

- Compiler version: any solc linux-amd64
- Target EVM version (as per compiler settings): any
- Framework/IDE (e.g. Truffle or Remix): n/a
- EVM execution environment / backend / blockchain client: n/a
- Operating system: linux

## Steps to Reproduce

Optional steps to get the same environment:

1. `docker run -it --rm ubuntu bash`
2. `apt update -y`
3. `apt install -y locales-all curl`

Then to replicate:

1. `export LANG=tr_TR.UTF-8`
2. `curl -O 'https://raw.githubusercontent.com/OpenZeppelin/openzeppelin-contracts/master/contracts/utils/Address.sol'`
3. `curl -O 'https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.10+commit.fc410830'`
4. `chmod +x solc-linux*`
5. `./solc-linux-amd64-v0.8.10+commit.fc410830 --bin Address.sol` # compilation fails

You can then change `LANG` back to `en_US.UTF-8` and the compilation will succeed.

chriseth:
Maybe related to the function `setDefaultOrCLocale` in main.cpp

chriseth:
We should ensure that the same calls are made i libsolc.cpp

Shungy:
I think I figured out the exact reason. Turkish capitalization rules are `i to İ`  and `I to ı`. I have realized that the both the inline assembly code I have tested have function names with letter `i`. So `extcodesize` is changed to `EXTCODESİZE` (notice dotted capital İ). This explains "function not found error".

cameel:
Wow. I knew about this capitalization weirdness with Turkish "i" but I didn't even realize that it means that plain ASCII characters can be capitalized into ones from the Unicode set :) Maybe we should just randomly set system locale to Turkish in one of our CI runs just to catch problems like this :)

In any case, we do attempt to set the locale to `C` to avoid such problems and this bug means that somehow it's not enough. And once we fix it, we should recommend frameworks to set it on their side as a workaround because old binaries will still be affected.

@chriseth Isn't `libsolc` just statically compiled into the binary? Why would we have to set the locale there if `solc` already does that?

chriseth:
libsolc has its own main function

chriseth:
[...]

PR #12531, opened 2022-01-14
External test for Bleeps

cameel:
Part of #5956.
**Depends on #12530. Don't merge until that PR is merged.**

This adds an external test for [Bleeps](https://github.com/wighawag/bleeps).

The project has an unusual structure compared to the others we have integrated so far (it's a monorepo with `pnpm` as the recommended package manager) so it was a bit tricky to get just the part we need to build and tests to run. The `main` branch also seems to be broken right now (https://github.com/wighawag/bleeps/issues/1). Fortunately there's a tag close to the top commit (`bleeps_migrations`) so I'm using that for now.

Note that the extra parameters my PR adds to `neutralize_package_json_hooks` are not used. I originally needed them for the `common-lib` library in the repo but in the and I was able to drop that. They may come in handy later though so I want to keep that change anyway.


ekpyron:
FYI: (probably trivial) merge conflicts to be resolved after I merged the first one

cameel:
Conflicts solved. As I said in, https://github.com/ethereum/solidity/pull/12530#issuecomment-1013492792, I'm going to put this PR on top of #12530 now.

ekpyron:
Wait, why does the commit of [#12350](https://github.com/ethereum/solidity/pull/12530) still show up here after that's merged? Ah, I guess that's what happens, if you don't also switch the base branch on github? Anyways, the test run seems to fail, so I'll have another look next week.

cameel:
Yeah, sorry looks like I did forget to switch the base branch :)

It's fails because the maintainer fixed https://github.com/wighawag/bleeps/issues/1 and I pushed a commit to switch it to `main`, hoping that it now passes tests. I think it would if not for the fact that the project now contains one 0.4.19 contract that won't compile with 0.8.11 :(

So back to the tag we go. I'll revert it to the previous version.

cameel:
Fixed. It should pass tests now.

hrkrshnn (approved):

Issue #12528, opened 2022-01-14
ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different: "No return param name given"


cameel:
## Description
When a virtual function has a documented return parameter, overriding it with one that returns a struct instead triggers an internal compiler error.

Such an override is valid (see https://github.com/ethereum/solidity/issues/11826#issuecomment-902835510) but apparently we don't have any tests covering it with Natspec.

The error happens also if the return value is unnamed.

It does not happen when you give the getter the type of the field (rather than the whole struct).

I found this while trying to compile [Uniswap v3](https://github.com/Uniswap/v3-core) with 0.8.11. Apart from missing explicit conversions this was the only issue preventing successful compilation.

## How to reproduce
```solidity
interface I {
    /// @return abc A value
    function p() external view returns (uint abc);
}

struct S {
    uint xyz;
}

contract C is I {
    S public override p;
}
```
```
Internal compiler error:
/solidity/libsolidity/interface/Natspec.cpp(201): Throw in function static Json::Value solidity::frontend::Natspec::extractReturnParameterDocs(const std::multimap<std::__cxx11::basic_string<char>, solidity::frontend::DocTag>&, const std::vector<std::__cxx11::basic_string<char> >&)
Dynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>
std::exception::what: No return param name given: xyz
[solidity::util::tag_comment*] = No return param name given: xyz
```

## Environment

- Compiler version: 0.8.11

bshastry:
May be a duplicate of https://github.com/ethereum/solidity/issues/11514

cameel:
It might but the repro is not completely the same. Your case has no name in `@return`. Given that #9947 was fixed and the fix did not cover #11514, I think it's better to keep both open until we're sure both are fixed.

Rav3nPL:
Seems like it is there since 0.8.0, tried compile Uniswap v3 contracts every version from 0.8.0 up to .11

cameel:
@Rav3nPL This is going to be fixed soon, but if you need a workaround, simply rename the `_liqiudity` return value to `liquidity`. Or just remove the whole line:
```bash
sed -i '/@return _liquidity/d' contracts/interfaces/pool/IUniswapV3PoolState.sol
```

Rav3nPL:
I see PR is waiting, just wanted to point it was there for some time.
Yes, I found that nasty underscore causing this.

Issue #12343, opened 2021-11-29
sushiswap/trident: `StackTooDeepError` in the new Yul->EVM transform


cameel:
I'm trying to compile [Trident](https://github.com/sushiswap/trident) via IR to use it for benchmarking the optimizer. Unfortunately it runs into "stack too deep" errors both with and without the optimizer. The project compiles fine with the legacy codegen (also both with and without the optimizer).

If the legacy codegen has no problem with it then I guess it should be possible to improve the Yul->EVM transform to handle it as well. If not,  we should at least have some specific workarounds to recommend since it's a real-life project and has to be compilable one way or another once the IR codegen becomes the default.

Another problem is that the old EVM transform runs out of memory trying to compile this project. I'm not sure if the new transform replaced it completely or if it can still be used in some situations. If it can, we should to track down the problem and fix it.

### Repro
```bash
git clone https://github.com/sushiswap/trident
cd trident/
yarn install
find . test -name '*.sol' -type f -print0 | xargs -0 sed -i -E -e 's/pragma solidity [^;]+;/pragma solidity >=0.0;/'

solc \
    --bin \
    --optimize \
    --experimental-via-ir \
    --metadata-literal \
    --base-path . \
    --include-path node_modules/ \
    $(find contracts/ -name '*.sol' -type f | grep -v "contracts/flat/BentoBoxV1Flat.sol") 
```

Note: I'm excluding `BentoBoxV1Flat.sol` because it can't be compiled with 0.8.x (https://github.com/sushiswap/trident/pull/282). The problem is reproducible without it too.

#### New EVM transform (0.8.10)
If you run the above with sol 0.8.10, the compiler crashes with a `StackTooDeepError`:
```
Uncaught exception:
/solidity/libyul/backends/evm/EVMObjectCompiler.cpp(77): Throw in function void solidity::yul::EVMObjectCompiler::run(solidity::yul::Object&, bool)
Dynamic exception type: boost::wrapexcept<solidity::yul::StackTooDeepError>
[...]

PR #12532, opened 2022-01-14
External test for Uniswap v3

cameel:
Part of #5956.
~Depends on #12531. Don't merge until that PR is merged.~ Merged.

This adds an external test for [Uniswap v3 core](https://github.com/Uniswap/v3-core). Still work in progress - I got it to compile but tests do not pass.

Hooking up Uniswap required quite a few workarounds:
- It's still on 0.7.6 so there were some missing explicit conversions. I patched them all with `sed` but I need to review them more carefully to make sure they keep the code working as intended. Some tests might be failing because of that.
- The code crashes the compiler due to a bug in Natspec. Reported in #12528.
- The config file exports settings without giving them a named variable. Because of this I cannot easily change settings. I had to add a helper for giving the variable a name (I've seen this in ElementFi too so It'll be useful).
- Tests run into the contract size limit just like in #12529 and #12530 and so this is another PR that has a copy of the commit that adds a helper for `allowUnlimitedContractSize`.
- This is yet another project that cannot be compiled via IR at all (#12343).

Even with these workarounds, over 150 tests fail for various reasons. Some due to checked math, some on unexpected values, some on JS errors (accessing `undefined`), which might be the symptom of Hardhat's heuristic not being prepared for IR.

The PR has some value even in the current form though so we might want to merge it anyway. Thanks to #12397 I can mark presets with failing tests as compile-only so they won't fail in CI. Then we could merge it as is and work on fixing the tests later.


cameel:
Looking closer at failing tests - most of the ones where we get unexpected values are checking gas usage for exact values. I think we could ignore those. There are overall just a few kinds of errors so fixing them might actually not be all that hard.

cameel:
Uniswap actually has an automatic update mechanism for these gas expectations (https://github.com/Uniswap/v3-core/issues/473):
```bash
UPDATE_SNAPSHOTS=1 yarn test
```

I'm going to add it to the test.

hrkrshnn:
It might be possible that wrapping everything in unchecked would make the tests pass. I believe their math libraries rely on wrapping behaviour.

cameel:
Right, but there seems to be a lot that need to be wrapped. Someone even already did that and the diff looks pretty large: https://github.com/Uniswap/v3-core/pull/505.

I wonder it it makes sense to use that PR here. It's not in the official repo and even the submitter says that it's not intended to be merged.

cameel:
Uniswap test is working!

- With so many patches needed I decided to create a fork: https://github.com/solidity-external-tests/uniswap-v3-core
    - My patches are in [`main_080`](https://github.com/solidity-external-tests/uniswap-v3-core/tree/main_080).
    - I submitted part of the branch in a PR to Uniswap: https://github.com/Uniswap/v3-core/pull/527. It contains only explicit conversions and even then only those that don't make tests fail. Other changes unfortunately break backwards-compatibility with 0.7.6 so I don't think they would accept them.
        - One conversion unfortunately affects code generation (adds an extra `SIGNEXTEND` that is not getting optimized out) so it breaks tests on 0.7.6.
        - The only other thing needed were the `unchecked` blocks. Lots of `unchecked` blocks.
- I am using `UPDATE_SNAPSHOT=1` when running tests to silence failures due to different gas expectations.
- Some test expectations work only with older Hardhat. On the other hand we need newer Hardhat to handle Panics. Version 2.4.3 seems to be the sweet spot.
- Even with that there were issues with missing methods on newer `@ethereumjs/tx`. Had to force downgrade it to 3.1.3.

ekpyron (dismissed):
Looks good, but again a (probably trivial) conflict with the last one I merged.

cameel:
Yeah. Same thing as the last time - all of these PRs are appending to the end of the list of tests concurrently :) Resolved.

I also noticed that I forgot to remove the compile-only run for Uniswap. Removed it now. It's superfluous now that the run that includes tests is working.

cameel:
Looks like the test is failing. It's because of the `??=` I used in `common.sh` in another ext test PR that's now merged. The operator is only available in Typescript 4 but Uniswap is on 3.7.3 and upgrading it breaks tests. I fixed it by replacing the operator with `??` which was already available in Typescript 3.7.

ekpyron (approved):

Issue #11670, opened 2021-07-15
[CI] Optimize our CircleCI machine sizes


cameel:
In our CI config we only use `medium` and `xlarge` machines. There's probably does not give us the best cost/performance ratio. We should try optimizing it a bit.

See [list of all available machine types](https://circleci.com/pricing/#compute-options-table).

On the call with CircleCI people we've been recommended to try the larger Windows and macOS machines to speed up CI runs. Currently we're using `medium` instances and there are larger ones available. Not sure if that's because larger ones weren't originally available or if we just decided smaller ones are most cost-effective. We should at least try them out and compare performance and cost (if we haven't already)

Also, we have some jobs that do not require much processing and only make use of one core. For example the ones I'm adding in #12181 and #12165 but I'm sure there are more. These can very likely be switched to `small` without affecting the total runtime too much.

Finally, we should make sure that all the linux jobs that use `xlarge` actually do benefit from it.

chriseth:
Please check first if we are using all the cores on the machine. I think the intent back then was that windows and macos failures are rare and thus it is not a big deal if the tests take a bit longer to run.

It would somehow be nice if we could select the machine based on the type of test run: For PRs and releases, it might be important to run quickly, for regular develop builds not so much. Also for PRs, we could default to "slow run" unless specified otherwise.

chriseth:
If we switch from medium to large, we get double the amount of cores and double the RAM for triple the price. Not sure if that is worth it.

cameel:
OK. If that's the pricing then it's probably not worth it. I assumed their pricing would be more linear :)

cameel:
> If we switch from medium to large, we get double the amount of cores and double the RAM for triple the price. Not sure if that is worth it.

That's true for Windows but looks like for macOS the cost/performance scaling is more linear (double cores+RAM = double price). So its seems worth trying.

Anyway, after working with CI today I realized that our machine allocation in general could be better optimized. We never use `small` even though there are some jobs that are pretty light and only use one core. I have updated the title and description to make this issue more general.

cameel:
### Observations
- Most jobs don't use all CPU cores. We can safely give them smaller machines.
- I found only two cases where a bigger machine significantly speeds things up:
    1. **Arch Linux**: We do not set `MAKEFLAGS` so the build does not even use the 2 cores it has in `medium`. Even with that a `large` machine makes it significantly faster.
[...]

PR #12536, opened 2022-01-14
[CI] Use `large` resource class for macOS and use `-j` without spaces

### LINKED ISSUE #12422: Parallelize macOS build in CI
In https://github.com/ethereum/solidity/issues/11670#issuecomment-954297346 I discovered that our [macOS build](https://github.com/ethereum/solidity/blob/v0.8.10/.circleci/config.yml#L673-L707) job in CI takes ~20 min regardless of how many cores the machine has. Apparently the `-j` flag we put in `MAKEFLAGS` variable is getting ignored. I think that the build config generated by CMake on macOS is still `make`-based so I'm not sure why this is the case.

This needs to be investigated. Once we know how to parallelize it, we should adjust the number of threads/processes and machine resource class in CI for the best cost/benefit ratio.

cameel:
Fixes #12422.

This is an experiment to see if it's the space in `-j 5` that prevents the option from taking effect on macOS.

EDIT: Looks like it works!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -358,7 +358,15 @@ defaults:
         xcode: "13.2.0"
       environment:
         TERM: xterm
-        MAKEFLAGS: -j 5
+        MAKEFLAGS: -j5
+
+  - base_osx_large: &base_osx_large
+      macos:
+        xcode: "13.2.0"
+      resource_class: large
+      environment:
+        TERM: xterm
+        MAKEFLAGS: -j10
 
   - base_ems_large: &base_ems_large
       docker:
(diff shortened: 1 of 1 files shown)


willianpaixao:
Well, it did take half of the time (from 22m to 11m), but you also increased the machine, so we can't know for sure it was the space.

cameel:
It must have been. Take a look at https://github.com/ethereum/solidity/issues/11670#issuecomment-954297346. When I was previously experimenting with it, increasing the machine size did nothing on its own. I was getting ~20 min both on `medium` and on `large`.

Thanks! Looks like this solves the issue for us.

ekpyron (approved):
Really :-)? Well, alright then :-).

cameel:
Yeah :) The worst thing is not even the inconsistent parameter syntax but that it just ignores it instead of failing.

willianpaixao:
Oh, right, that's awesome then.
As for the  inconsistent parameter syntax, how about removing the space from all the other jobs?
The Mandalorian would say, "this is the way".

cameel:
Since `-j` is not the only option that works like this, I suspect we have both variants scattered all over the codebase so I don't think it's that important. But if you want to submit a PR to change this just in the CI config, I don't mind either so feel free, it would be accepted :)

PR #12537, opened 2022-01-15
feat: add sphinx-intl to facilitate internacionalization

willianpaixao:
Hello,

this PR introduces the [sphinx-intl](https://sphinx-intl.readthedocs.io/en/master/) library and its usage in the `scripts/docs.sh` script.
The idea is to facilitate translation of the documentation from the source, the Sphinx framework.

Note, to archive the "full translation experience", end-to-end, there are more steps to be made, such as translating the generated files and committing back to a branch (not necessarily `develop`). But the library usage already can be used by RTD, just will contain the same content as the original English. 😅 

### Reference
- [Manage Translations for Sphinx projects](https://docs.readthedocs.io/en/stable/guides/manage-translations-sphinx.html)
- [Sphinx docs on Internationalization](https://www.sphinx-doc.org/en/master/usage/advanced/intl.html)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -297,3 +297,11 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+# -- Options for gettext builder ------------------------------------------
+
+# Set document text domain to markup/code. Recommended by sphinx-intl.
+gettext_compact = False
+
+# Generates uuid information for version tracking in message catalogs.
+gettext_uuid = True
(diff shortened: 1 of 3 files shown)


cameel:
Unfortunately, this does not fit into our current translation workflow. We don't really use the gettext mechanism and https://github.com/solidity-docs is not set up for that. If anyone really wanted to use gettext for some reason despite this, they can add `sphinx-intl` in their fork. This would only be of any use here if we wanted to store the `.po` files in the main repo and host them together with the English docs, which currently we're not planning to. We do not want to burden translators with having to be always perfectly in lockstep with our releases, so it makes more sense to keep translations independent.

Personally, I have nothing against gettext-based translations but I think that in our case the current system fits potential contributors better. It does not depend on any external tools and is very simple for anyone used to code. Also, with gettext AFAIK you get the text split into individual snippets without context and it would be hard to translate whole pages of related text that way.

PR #12542, opened 2022-01-17
External test for PRBMath

cameel:
Part of #5956.

This adds an external test for [PRBMath](https://github.com/paulrberg/prb-math). We already have a copy of an older version of this contract in [`test/libsolidity/semanticTests/externalContracts/`](https://github.com/ethereum/solidity/tree/develop/test/libsolidity/semanticTests/externalContracts) but the external test will track upstream directly and will also be useful for benchmarking.

`ir-optimize-evm+yul` compiles but does not pass tests due to https://github.com/nomiclabs/hardhat/issues/2115.

This one was pretty easy and did not need any special workarounds.


test/externalTests/prb-math.sh
```diff
@@ -0,0 +1,77 @@
...
+    setup_solc "$DIR" "$BINARY_TYPE" "$BINARY_PATH"
+    download_project "$repo" "$ref_type" "$ref" "$DIR"
+
+    cp .env.example .env
+
+    neutralize_package_lock
+    neutralize_package_json_hooks
+    force_hardhat_compiler_binary "$config_file" "$BINARY_TYPE" "$BINARY_PATH"
+    force_hardhat_compiler_settings "$config_file" "$(first_word "$SELECTED_PRESETS")" "$config_var"
+    yarn install --no-immutable
```
hrkrshnn: Looking at the test: maybe this should go before the hardhat steps?

cameel: It should not matter. All these commands just create/modify/delete files without running yarn or hardhat. `yarn install` command is the first thing that actually might make use of them.

I don't fully understand why it's failing. There are no nested workspaces here and on my machine locally it works fine with just `yarn install`. I'm removing the lock file so it should just create one and fill it during `install` but for some reason it does not work that way in CI.

cameel: Apparently there's globally installed yarn v1 in the image but when I run it in the script, I get yarn v3. I wonder if it autoupdates or something? In any case, I think the simplest solution might be to try npm before digging deeper into it :)

cameel: Turns out that there's a [yarn 3.1.0 binary stored in the repository](https://github.com/paulrberg/prb-math/blob/main/.yarn/releases/yarn-3.1.0.cjs) and that you can force yarn v1 to use it via [.yarnrc.yml](https://github.com/paulrberg/prb-math/blob/main/.yarnrc.yml). Removing the rc file gives me back yarn v1 and makes the test finally pass.

hrkrshnn (approved):

PR #12544, opened 2022-01-17
Natspec: Fix ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different.

### LINKED ISSUE #12528: ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different: "No return param name given"

Marenz:
fixes #12528

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ Compiler Features:
 Bugfixes:
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the parent contract contains immutable variables.
+ * Natspec: Fix ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different.
  * TypeChecker: Fix ICE when a constant variable declaration forward references a struct.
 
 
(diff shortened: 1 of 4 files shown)


ekpyron (commented):
Looks good at a quick glance - solc-js test failing is a fluke and should work on retrigger. I'll have a closer look tomorrow and will probably approve then.

ekpyron (approved):
Yeah... I see the problem you mentioned in the chat... but I guess the solution is good enough :-).

PR #12545, opened 2022-01-17
Allow builtins in yul identifier paths in antlr grammar.

ekpyron:
Came up in https://github.com/ethereum/solidity/pull/12470

### DIFF
--- a/docs/grammar/SolidityParser.g4
+++ b/docs/grammar/SolidityParser.g4
@@ -564,7 +564,7 @@ yulFunctionDefinition:
  * While only identifiers without dots can be declared within inline assembly,
  * paths containing dots can refer to declarations outside the inline assembly block.
  */
-yulPath: YulIdentifier (YulPeriod YulIdentifier)*;
+yulPath: YulIdentifier (YulPeriod (YulIdentifier | YulEVMBuiltin))*;
 /**
  * A call to a function with return values can only occur as right-hand side of an assignment or
  * a variable declaration.
--- a/test/libsolidity/syntaxTests/inlineAssembly/assignment_to_function_pointer.sol
+++ b/test/libsolidity/syntaxTests/inlineAssembly/assignment_to_function_pointer.sol
@@ -1,9 +1,9 @@
 contract C {
-  function f() public pure {
-    function() external g;
-    assembly {
-      g.address := 0x42
-      g.selector := 0x23
+    function f() public pure {
+        function() external g;
+        assembly {
+            g.address := 0x42
+            g.selector := 0x23
+        }
     }
-  }
 }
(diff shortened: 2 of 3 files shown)


docs/grammar/SolidityParser.g4
```diff
@@ -564,7 +564,7 @@ yulFunctionDefinition:
  * While only identifiers without dots can be declared within inline assembly,
  * paths containing dots can refer to declarations outside the inline assembly block.
  */
-yulPath: YulIdentifier (YulPeriod YulIdentifier)*;
+yulPath: YulIdentifier (YulPeriod (YulIdentifier | YulEVMBuiltin))*;
```
ekpyron, line 567: Strictly speaking probably ``address.test`` would in theory be a valid ``yulPath`` as well, but I'd say this hack is good enough for all intents and purposes.

ekpyron:
If we want, we can think about whether we can do this "properly", i.e. have the solidity yul version just parse identifiers containing dots - but since this won't really happen too much in practice, since we reserved identifiers with dots for high-level-language specific constructions and we're unlikely to do more weird stuff like with ``functionPointer.address``, I'd say we should at least temporarily merge this, s.t. we can proceed with https://github.com/ethereum/solidity/pull/12470.

test/libsolidity/syntaxTests/inlineAssembly/assignment_to_function_pointer.sol
```diff
@@ -0,0 +1,9 @@
+contract C {
+  function f() public pure {
+    function() external g;
+    assembly {
+      g.address := 0x42
+      g.selector := 0x23
+    }
+  }
+}
```
cameel:
```suggestion
contract C {
    function f() public pure {
        function() external g;
        assembly {
            g.address := 0x42
            g.selector := 0x23
        }
    }
}
```

cameel (dismissed):
Just one cosmetic tweak needed so I'm approving already.

By the way. How does this affect recent efforts to specify Yul formally? Was that in any way dependent on our grammar definitions or done from scratch?

cameel (approved):

ekpyron:
> By the way. How does this affect recent efforts to specify Yul formally? Was that in any way dependent on our grammar definitions or done from scratch?

I wouldn't be aware of the grammar being used for that (or actually for anything apart from our docs and tests for that matter).

cameel:
Ok, good. Looking at the article that was posted last week ([Securing Warp: A formal specification of the Yul IR](https://medium.com/nethermind-eth/securing-warp-a-formal-specification-of-the-yul-ir-85bb3bf51c62)), they just [treat Yul identifiers as strings and don't have a "yul path" like we do in our grammar](https://github.com/NethermindEth/Yul-Specification/blob/main/yul_ast.lean).

ekpyron:
Ah, that one you meant - that's a rather bold simplification of Yul anyways :-).

PR #12548, opened 2022-01-17
Add event and error identifiers to --hashes

joshieDo:
Adds event and error identifiers when calling `solc --hashes`

Output example:
```
======= D.sol:D =======
Function signatures:
c6b29b88: bigOops()
6b630cb2: smallOops()

Error signatures:
44c3d84d: BigError()
1e71f739: SmallError(uint256)

Event signatures:
57050ab73f6b9ebdd9f76b8d4997793f48cf956e965ee070551b9ca0bb71584e: Event()
510e730eb6600b4c67d51768c6996795863364461fee983d92d5e461f209c7cf: Event(uint256)
```


libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,34 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
 }
 
+Json::Value CompilerStack::errorIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value errorIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceErrors()) {
+		string signature = it->functionType(true)->externalSignature();
```
hrkrshnn:
```suggestion
	for (auto const& it: contractDefinition(_contractName).interfaceErrors()) 
	{
		string signature = it->functionType(true)->externalSignature();
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,34 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
+}
+
+Json::Value CompilerStack::eventIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value eventIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceEvents()) {
+		string signature = it->functionType(true)->externalSignature();
```
hrkrshnn:
```suggestion
	for (auto const& it: contractDefinition(_contractName).interfaceEvents()) 
	{
		string signature = it->functionType(true)->externalSignature();
```

hrkrshnn (commented):
Thanks for the PR. Looks very good!

Could you add a test similar to https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/abiencoderv2_no_warning or simply modify the contract to include an event and error?

The test output can be then interactively updated by running `./test/cmdlineTests.sh --no-smt` (assuming that you have built the `solc` binary at `build/solc/`).

Left two minor coding style suggestions.

libsolidity/interface/CompilerStack.cpp
joshieDo: For some reason, `./scripts/check_style.sh` didn't catch it. Is that normal?

joshieDo:
> Thanks for the PR. Looks very good!
> 
> Could you add a test similar to https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/abiencoderv2_no_warning or simply modify the contract to include an event and error?
> 
> The test output can be then interactively updated by running `./test/cmdlineTests.sh --no-smt` (assuming that you have built the `solc` binary at `build/solc/`).
> 
> Left two minor coding style suggestions.

Ended up adding to the existing one, along with the coding style changes.

libsolidity/interface/CompilerStack.cpp
hrkrshnn: @joshieDo Yeah, the check style is just a basic bash script. Not surprised.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,36 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
+Json::Value CompilerStack::eventIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value eventIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceEvents())
+	{
+		string signature = it->functionType(true)->externalSignature();
+		eventIdentifiers[signature] = toHex(toCompactBigEndian(selectorFromSignature32(signature), 4));
```
hrkrshnn: There is an issue here. There are two kinds of events, the regular events and anonymous events. For the regular event, the `topic` for the `log*` opcode is the full 32 byte `keccak` hash of the selector. For anonymous events, this is skipped. So there are two things to be changed here:

1. Skip the signature for anonymous events. Check for `EventDefinition::isAnonymous()`.
2. For the regular ones, the signature should be the full 32 bytes. Here's an example from the codebase, where it is computed

https://github.com/ethereum/solidity/blob/a07b3ec70f91fdbf35929d5d474e91e739f792e8/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L998

hrkrshnn (changes requested):

hrkrshnn:
@joshieDo We were wondering what made you write this PR? How are you using this information?

joshieDo:
Initially, I was dealing with some test framework which was comparing error selectors and giving me a comparison log in the format of `0xXXXXXXXX != 0xYYYYYYYY`. I tried to find a straightforward way to get the error signatures like `--hashes` but couldn't.

In the past, I also had a similar experience, when subscribing to event topics, where I wanted a straightforward way like `--hashes`. (but totally forgot that it's 32 bytes when making this change)

Ended up forking and adding this change, and wondered if other people might be interested in it.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,36 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
 	return methodIdentifiers;
 }
 
+Json::Value CompilerStack::errorIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value errorIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceErrors())
```
cameel: This is not an iterator.
```suggestion
	for (ErrorDefinition const* error: contractDefinition(_contractName).interfaceErrors())
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,36 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
+	return errorIdentifiers;
+}
+
+Json::Value CompilerStack::eventIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value eventIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceEvents())
```
```suggestion
	for (EventDefinition const* event: contractDefinition(_contractName).interfaceEvents())
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,39 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
 
+Json::Value CompilerStack::errorIdentifiers(string const& _contractName) const
+{
+	if (m_stackState < AnalysisPerformed)
+		solThrow(CompilerError, "Analysis was not successful.");
+
+	Json::Value errorIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceErrors())
+	{
+		string signature = it->functionType(true)->externalSignature();
```
cameel: `true` means that you are requesting an internal function type. Does this actually work with `true`?
```suggestion
		string signature = it->functionType(false)->externalSignature();
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,39 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
...
+
+	Json::Value eventIdentifiers(Json::objectValue);
+	for (auto const& it: contractDefinition(_contractName).interfaceEvents())
+	{
+		if (!it->isAnonymous())
+		{
+			string signature = it->functionType(true)->externalSignature();
+			eventIdentifiers[signature] = toHex(u256(h256::Arith(keccak256(signature))));
+		}
+	}
```
```suggestion
	for (auto const& it: contractDefinition(_contractName).interfaceEvents())
		if (!it->isAnonymous())
		{
			string signature = it->functionType(true)->externalSignature();
			eventIdentifiers[signature] = toHex(u256(h256::Arith(keccak256(signature))));
		}
```

solc/CommandLineInterface.cpp
```diff
@@ -271,14 +271,30 @@ void CommandLineInterface::handleSignatureHashes(string const& _contract)
 		return;
 
 	Json::Value methodIdentifiers = m_compiler->methodIdentifiers(_contract);
-	string out;
+	string out = "Function signatures:\n";
 	for (auto const& name: methodIdentifiers.getMemberNames())
 		out += methodIdentifiers[name].asString() + ": " + name + "\n";
 
+	Json::Value errorIdentifiers = m_compiler->errorIdentifiers(_contract);
+	if (errorIdentifiers.size() >0)
```
```suggestion
	if (errorIdentifiers.size() > 0)
```
or, better
```suggestion
	if (!errorIdentifiers.empty())
```

test/cmdlineTests/abiencoderv2_no_warning/input.sol
```diff
@@ -4,6 +4,8 @@ pragma abicoder v2;
 
 contract C {
     struct S { uint x; }
+    event ev(uint y);
+    error err(uint z, uint w);
     function f(S memory) public pure {
     }
```
cameel: @hrkrshnn 
> Could you add a test similar to https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/abiencoderv2_no_warning or simply modify the contract to include an event and error?

I think we should really have a dedicated test for `--hashes`.

cameel (changes requested):
Two more things:
- This needs an entry in the [Changelog](https://github.com/ethereum/solidity/blob/develop/Changelog.md).
- We also have `--combined-json hashes`. Adding this to the CLI only will make them inconsistent. On the other hand, we're planning to deprecate `--combined-json`, so I'm not sure if we want to add it there. What's your opinion @hrkrshnn, @ekpyron?

test/cmdlineTests/abiencoderv2_no_warning/input.sol
hrkrshnn: Maybe this test is not named properly, but the `args` file only has `--hashes`!

cameel: True! It's still meant to test something different though: 0973ae751af1c6a010fafe193c5e099c09eb8e7a. I think it could have been done using another other option and it might change in the future due.

We could use a name that implies that it tests both, but at this point, I'd rather have two tests even if they're very similar.

libsolidity/interface/CompilerStack.cpp
joshieDo: It does work, <del>but after taking a deeper look, I should probably change it for `type()`?</del>
edit: Seems to me, that I do have to call it with `true`. I'm assuming it was done like this, since some things like `signatures` are shared.

https://github.com/ethereum/solidity/blob/a07b3ec70f91fdbf35929d5d474e91e739f792e8/libsolidity/ast/AST.cpp#L533-L544

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1024,6 +1025,37 @@ Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
 	return methodIdentifiers;
 }
 
+Json::Value CompilerStack::errorIdentifiers(string const& _contractName) const
```
axic: I think we should change `methodIdentifiers` to return a single large JSON and display them that way on the CLI, and pass it down as a new field in standard json.

cameel: Good point (I'd rename it from `methodIdentifiers()` to something more genric though).

Just to make things clear - you don't mean literally printing the JSON on the CLI but just passing that JSON to `CommandLineInterface` and then printing it the same way we do now, right?

Also, what's your opinion on https://github.com/ethereum/solidity/pull/12548#pullrequestreview-855403715? Does it make sense to add it to `--combined-json` or should we keep it as is since it's going to be removed anyway?

> pass it down as a new field in standard json.

I guess it would be fine to do it in a separate PR. I don't think we have this info in Standard JSON even for method identifiers right now.

axic: > Just to make things clear - you don't mean literally printing the JSON on the CLI but just passing that JSON to CommandLineInterface and then printing it the same way we do now, right?

No, it still needs to be displayed nicely, but that can be done by iteration the two-level JSON.

> Does it make sense to add it to --combined-json or should we keep it as is since it's going to be removed anyway?

Please do not touch `--combined-json`, it is due to be removed next breaking release.

>  I don't think we have this info in Standard JSON even for method identifiers right now.

I think we have it.

chriseth:
Please ensure that this does not only include the defined errors and events, but also the used ones, e.g. errors or events defined at file level.

ekpyron:
> Please ensure that this does not only include the defined errors and events, but also the used ones, e.g. errors or events defined at file level.

@joshieDo Do you want to take care of this as well or should one of us take over for this last part?

joshieDo:
> @joshieDo Do you want to take care of this as well or should one of us take over for this last part?

I'm not sure where to start on that, and since I'm also tight on time nowadays, I'd really appreciate if you could take over the last part. Thank you!

ekpyron:
> > @joshieDo Do you want to take care of this as well or should one of us take over for this last part?
> 
> I'm not sure where to start on that, and since I'm also tight on time nowadays, I'd really appreciate if you could take over the last part. Thank you!

Sure, no worries - thank you very much for the PR thus far, we'll take it from here!

Marenz:
file level events are not yet supported, but added a check for file level errors.
Rebased & merged into one commit

libsolidity/interface/CompilerStack.cpp
cameel: > I think we have it.

Right, we do. It's `evm.methodIdentifiers`. Not sure how I missed that.

In that case @Marenz I think we should add `evm.eventIdentifiers` and `evm.errorIdentifiers` too. Either in this PR or in a separate one (either is fine with me).

Changelog.md
```diff
@@ -7,7 +7,7 @@ Language Features:
 
 Compiler Features:
  * Yul Optimizer: Remove ``mstore`` and ``sstore`` operations if the slot already contains the same value.
-
+ * Commandline Interface: Event and error signatures are also returned when using `--hashes`.
```
```suggestion
 * Commandline Interface: Event and error signatures are also returned when using ``--hashes``.
```

Single quotes work too but for some reason we're always using double ones in the changelog and I'd stick to that.

libsolidity/interface/CompilerStack.cpp
cameel: > I think we should change `methodIdentifiers` to return a single large JSON and display them that way on the CLI, and pass it down as a new field in standard json.

What do you think of this one @Marenz?

test/cmdlineTests/hashes/err
```diff
@@ -0,0 +1 @@
+
```
cameel: This file is empty so it can be removed.

test/cmdlineTests/hashes/input.sol
```diff
@@ -0,0 +1,17 @@
...
+
+    event ev(uint y);
+    event anon_ev(uint y) anonymous;
+
+    error err(uint z, uint w);
+
+    function f(S memory) public pure {
+        revert fileLevelError(3);
+    }
+}
```
cameel: I think it would be a good idea to have a library here with both events/errors used and not used by a contract. Maybe also an interface.

I'd also add a free function and an internal one to make sure they do not appear in output.

cameel (commented):
Looks fine overall but there's an unanswered question about Standard JSON and I thing that the test could include a few more things to make sure they work.

test/cmdlineTests/hashes/input.sol
Marenz: so both events/errors used, but not used by a contract.. used by what then?

cameel: By nothing. The point is that they should not show up on the list if they are not used.

Marenz:
hmm not really sure why the externals are failing

cameel:
I think this is actual breakage caused by the PR :) In Hardhat the function that fails is [`correctSelectors()`](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/stack-traces/compiler-to-model.ts#L603-L637) - I think the PR must have changed the `evm.methodIdentifiers` output in some way. In Trident I think that Hardhat just stopped recognizing the error that the test expects. Not sure about OpenZeppelin but it's also because it fails for a different reason so I think that Hardhat heuristics might have misdetected it as `invalid opcode`.

By the way, if our own tests passed this through, they're clearly too weak. Please try to add something that would fail in this case.

chriseth (changes requested):
Since we are creating a new output element, we should include used events and not only defined ones. Please either modify `interfaceEvents` to add a bool flag to indicate whether to add the used events or not (or create a new function for that).

libsolidity/ast/AST.cpp
```diff
@@ -213,11 +213,22 @@ vector<EventDefinition const*> const& ContractDefinition::interfaceEvents() cons
...
+vector<EventDefinition const*> const ContractDefinition::usedInterfaceEvents() const
+{
+	solAssert(annotation().creationCallGraph.set(), "");
+
+	std::vector<EventDefinition const*> events;
+
+	events += (*annotation().creationCallGraph)->emittedEvents;
+	events += (*annotation().deployedCallGraph)->emittedEvents;
+
+	return events;
```
chriseth: Why not
```
return
    ...->emittedEvents +
    ...->emittedEvents;
```
?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1013,15 +1016,34 @@ Json::Value const& CompilerStack::natspecDev(Contract const& _contract) const
 	return _contract.devDocumentation.init([&]{ return Natspec::devDocumentation(*_contract.contract); });
 }
 
-Json::Value CompilerStack::methodIdentifiers(string const& _contractName) const
+Json::Value CompilerStack::contractIdentifiers(string const& _contractName) const
```
chriseth: Can we choose a less generic name for this method? It does not return all identifiers of the contract...

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1013,15 +1016,34 @@ Json::Value const& CompilerStack::natspecDev(Contract const& _contract) const
...
-	return methodIdentifiers;
+		contractIdentifiers["methods"][it.second->externalSignature()] = it.first.hex();
+	for (ErrorDefinition const* error: contractDefinition(_contractName).interfaceErrors())
+	{
+		string signature = error->functionType(true)->externalSignature();
+		contractIdentifiers["errors"][signature] = toHex(toCompactBigEndian(selectorFromSignature32(signature), 4));
+	}
+
+	for (EventDefinition const* event: ranges::concat_view(
+		contractDefinition(_contractName).interfaceEvents(),
```
chriseth: Now this is not in sync with `interfaceErrors`, the two functions sound similar but they return totally different items.
Maybe either keep just a single function with a parameter or rename `interfaceEvents` to something like `definedInterfaceEvents`.

libsolidity/interface/CompilerStack.cpp
Marenz: How about `visibleIdentifiers()`? Or `exposedIndentifiers()`?

Marenz: Well or I guess `interfaceIdentifiers`?

libsolidity/ast/AST.cpp
Marenz: emittedEvents is a `set`, so we can't just return the concatination

chriseth: ah, but then it's even more important to first add the sets and then turn it into a vector.

libsolidity/interface/CompilerStack.cpp
chriseth: `interfaceMethodsErrorsAndEvents`?

libsolidity/interface/CompilerStack.cpp
chriseth: In a PR to breaking, I changed the `interfaceEvents` function to be identical to `interfaceErrors`. Maybe we should just add that function as `interfaceEventsBreaking` or something instead of concatenating the vectors here again? Also note that most events will be in both vectors, I think that's confusing.

Marenz: Why would that be confusing? It is a given that some of the used events will also be defined events, that shouldn't be surprising to anyone. 

As we're assigning all the events to only their key, any duplicates are filtered out here anyway..

Having it identical to `interfaceErrors` means we can't control anymore if we want to include the used events or not.. if that is fine with the assumed usage of those functions, sure, we can do that, too...

chriseth (approved):

PR #12551, opened 2022-01-18
Add test for fixed natspec ICE

### LINKED ISSUE #11514: [NatSpec] ICE in extractReturnParameterDocs due to lack of return parameter name
```
interface IThing {
  ///@return
  function value(uint) external returns (uint128,uint128);
}

contract g is IThing {
  struct Value {
    uint128 x;
    uint128 A;
  }
  mapping(uint=>Value) public override value;
}
```

throws

https://github.com/ethereum/solidity/blob/98e7b61a378467116bd5dfdd8d1455a60ef26d2e/libsolidity/interface/Natspec.cpp#L201

To reproduce

```
$ solc test.sol
```

Marenz:
fixes #11514

### DIFF
--- a/test/libsolidity/SolidityNatspecJSON.cpp
+++ b/test/libsolidity/SolidityNatspecJSON.cpp
@@ -2534,6 +2534,55 @@ BOOST_AUTO_TEST_CASE(dev_struct_getter_override)
 	checkNatspec(sourceCode, "Thing", natspec2, false);
 }
 
+BOOST_AUTO_TEST_CASE(dev_struct_getter_override_no_return_name)
+{
+	char const *sourceCode = R"(
+		interface IThing {
+			///@return
+			function value(uint) external returns (uint128,uint128);
+		}
+
+		contract Thing is IThing {
+			struct Value {
+				uint128 x;
+				uint128 A;
+			}
+			mapping(uint=>Value) public override value;
+		}
+	)";
+
+	char const *natspec = R"ABCDEF({
+		"methods":
+		{
+			"value(uint256)":
+			{
+				"returns":
+				{
+					"_0": ""
+				}
+			}
+		}
+	})ABCDEF";
+
+	char const *natspec2 = R"ABCDEF({
+		"methods": {},
+		"stateVariables":
+		{
+			"value":
+			{
+				"return": "x ",
+				"returns":
+				{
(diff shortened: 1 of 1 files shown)


ekpyron (approved):

test/libsolidity/SolidityNatspecJSON.cpp
```diff
@@ -2534,6 +2534,55 @@ BOOST_AUTO_TEST_CASE(dev_struct_getter_override)
...
+		}
+	})ABCDEF";
+
+	char const *natspec2 = R"ABCDEF({
+		"methods": {},
+		"stateVariables":
+		{
+			"value":
+			{
+				"return": "x ",
```
ekpyron, line 2573: ``"x "``? Where's that coming from? Seems a bit weird.

Marenz: From the value struct. It assumed it's only one parameter and the names were different so it adjusted it

ekpyron: And what about ``A``?
I'm wondering whether everything is sane here or whether we should check if this is really what we want, resp. whether there's some issue hidden here.

Marenz: Basically, the amount of (documented) parameters is inherited and the base documentation only documented one (I mean, not really, but an empty return is understood as the first parameter with an empty descrption)

Marenz: I wouldn't dive too deep into this because I think the whole structure of natspec needs a more .. structured rewrite anyway

ekpyron: I'm not entirely convinced :-). Where is the space coming from then :-)? Is there a similar case that already does something equally weird on ``develop`` :-)?

ekpyron: I mean, this PR is only the test, so we can merge it of course, but we should more properly look into this further before the next release, so I tend to keep it open as a reminder...

Marenz: Usually it would be `"return": "x description of parameter",` but as the description is missing, it is just a space. That's all there is to it

PR #12553, opened 2022-01-18
Fix .push() not considering external functions

### LINKED ISSUE #12546: [Sol->Yul] ICE due to invalid IR when compiling push on dynamic storage array of one of it's own elements
```
contract C {
  struct St1 {
    function () external el;
  }
  St1[] s;
  function f() public {
    s.push(s[0]);
  }
}
```

throws

https://github.com/ethereum/solidity/blob/07d1bc5f37ab071ec8a33d98f663462ce83c38ea/libsolidity/codegen/ir/IRGenerator.cpp#L112

Repro

```
$ solc --ir test.sol
```

Marenz:
fixes #12546

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Bugfixes:
  * Antlr Grammar: Allow builtin names in ``yulPath`` to support ``.address`` in function pointers.
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the parent contract contains immutable variables.
+ * IR Generator: Fix IR syntax error when copying storage arrays of structs containing functions.
  * Natspec: Fix ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different.
  * TypeChecker: Fix ICE when a constant variable declaration forward references a struct.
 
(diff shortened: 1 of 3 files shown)


ekpyron:
It's a bit silly to split the function pointer into two slots only to then recombine them to a single one instead of just ``sload`` and ``sstore`` the same value...  but it's also an advantage to keep the code generic and simple, so it may even better to just keep it like this - I'm wondering how well this optimizes :-).

ekpyron:
I'm also wondering if we should be more paranoid about testing this in various corner cases...

ekpyron (commented):
Well, at least a Changelog entry would be good, apart from that I may approve it as is :-).

Changelog.md
```diff
@@ -12,6 +12,7 @@ Compiler Features:
 Bugfixes:
  * Antlr Grammar: Allow builtin names in ``yulPath`` to support ``.address`` in function pointers.
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
+ * IR Generator: Fix IR syntax error when copying storage arrays of structs containing functions.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the parent contract contains immutable variables.
```
ekpyron: Do we sort this in case-sensitive or case-insensitive manner :-D?

ekpyron (dismissed):

Changelog.md
Marenz: well, I used `:sort` in vim :> I can change it

ekpyron: I'm not saying we should order it differently, just that I'm not sure, if we have a fixed convention for this :-).

ekpyron (approved):

Issue #12558, opened 2022-01-19
ICE when copying storage structs to memory arrays via IR


cameel:
## Description
Found while working on the external test for [Pool Together v4](https://github.com/pooltogether/v4-core). The contract does not compile via IR at all due to an Internal Compiler Error. With legacy codegen the compilation succeeds.

## Steps to Reproduce
Here's a distilled repro for the ICE:
```solidity
struct S {
    uint x;
}

contract C {
    S sStorage;

    function f() external {
        S[] memory sMemory;

        sMemory[0] = sStorage;
    }
}
```
```
Internal compiler error:
/solidity/libsolidity/codegen/ir/IRGeneratorForStatements.cpp(2962): Throw in function solidity::frontend::IRGeneratorForStatements::writeToLValue(const solidity::frontend::IRLValue&, const solidity::frontend::IRVariable&)::<lambda(const solidity::frontend::IRLValue::Memory&)>
Dynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>
std::exception::what: Solidity assertion failed
[solidity::util::tag_comment*] = Solidity assertion failed
```

It only happens when the target struct is a part of an array (does not matter whether dynamic or static). Does not happen when the target is just a single struct.

**EDIT**: When this is fixed, remember to re-enable the `ir-optimize-evm+yul` preset in the `pool-together` external test (which is now in `develop`): https://github.com/ethereum/solidity/blob/93f7d5f8f0be28d84158a6ccde7fd7d2050598d6/test/externalTests/pool-together.sh#L48

## Environment

- Compiler version: 0.8.11

ekpyron:
This will happen for any reference type, e.g. the following will also trigger it:
```
contract C {
    uint[] sStorage;

    function f() external {
        uint[][] memory sMemory;
        sMemory[0] = sStorage;
    }
}
```

bshastry:
Co incidentally what I have been trying to get the fuzzer to generate :-)

ekpyron:
Another one - apparently also doesn't have to be storage :-D...
```
contract C {
    function f(uint[] calldata a) external {
        uint[][] memory m;
        m[0] = a;
    }
}
```

For some reason we *only* took care of memory-to-memory there :-).

ekpyron:
Actually, I think this is even the same issue as https://github.com/ethereum/solidity/issues/12220

cameel:
[...]

PR #12560, opened 2022-01-19
External test for PoolTogether v4

cameel:
Part of #5956.

This adds an external test for [PoolTogether V4](https://github.com/pooltogether/v4-core). The contract uncovered a bug in our IR pipeline (#12558) and the optimized viaIR preset can't be enabled until that's fixed.

Other than that it was pretty easy to add. Just required removal of artifacts for a few contracts that come via npm already compiled.


ekpyron (approved):
Doesn't bode well for the state of the IR :-).

PR #12562, opened 2022-01-19
External test for Brink

cameel:
Part of #5956.

This adds an external test for [Brink](https://github.com/brinktrade/brink-core).
- Needed manual installation of `chai` due to https://github.com/brinktrade/brink-core/issues/48
- Tests are very picky about the bytecode.
    - On 0.8.10 they pass only with the exact settings the project uses. I.e. runs of 800 and disabled metadata.
    - On 0.8.11 tests do not pass at all.
    - `SNAPSHOT_UPDATE=1` makes some of them pass but not all. There are some hard-coded values in [`constants.js`](https://github.com/brinktrade/brink-core/blob/master/constants.js) which probably need to be calculated specifically for the bytecode a given version of the compiler produces.

As is, this could be used to test compilation but tests won't pass. Alternatively, we could try to add something that recalculates constants specifically for our compiler binary. Not sure how complex that would be.


leonardoalt:
@cameel so looks like we won't use this for now?

cameel:
Tests won't pass and (I'm not sure it's worth the effort to try to get them to) but it's still usable as compile-only test.

I'll change it to compile-only and make it reviewable.

cameel:
Converted to compile-only and ready to review.

test/externalTests/brink.sh
```diff
@@ -0,0 +1,83 @@
...
+
+function brink_test
+{
+    local repo="https://github.com/brinktrade/brink-core"
+    local ref_type=branch
+    local ref=master
+    local config_file="hardhat.config.js"
+    local config_var=""
+    local extra_settings="metadata: {bytecodeHash: 'none'}"
+    local extra_optimizer_settings="runs: 800"
```
ekpyron, line 45: What happens without this?

ekpyron (approved):

test/externalTests/brink.sh
cameel: It's needed for tests to pass-  they seem to depend on the exact bytecode on 0.8.10. With this tests on 0.8.10 pass.

The thing is, I can't get them to pass on any other versions so I gave up and they're disabled. So I could remove this since it's now compile only. But if we keep it, we won't have to rediscover that if we ever decide to try to get these tests to run.

PR #12566, opened 2022-01-20
Merge remote-tracking branch 'origin/develop' into breaking

Marenz:


ekpyron (approved):

ekpyron:
Nothing I can do here :-).
You can try pushing another branch and merging that, maybe then it randomly works...

cameel:
Oh, I know what this is and I have bad news :(
I renamed this job in #12173 back in October. The job can't run because it does not exist but apparently it's still marked as required in settings of the `breaking` branch. It was done quite a while ago and I thought that this was already on `breaking` by now. Apparently not. The bad news is that there's no way to disable the requirement other than in settings. Well, you could work around it by adding back a no-op job with that name to CI config but that's ugly. How badly do you need this merged before Chris is back? :)

ekpyron:
We could just work on a copy of the ``breaking`` branch like ``breaking-tmp`` for the time being and merge that into breaking once the settings are changed...

Marenz:
Could we add a mocking-job that just finishes successfully?

cameel:
I'm fine with that if you really want it that way but I think that the solution @ekpyron proposed would be cleaner, i.e. would not leave any workaround commits in `breaking` long term :)

Marenz:
obsolete

PR #12568, opened 2022-01-20
Disallow comparison operators on contracts

### LINKED ISSUE #11700: Reconsider operators on ContractType
```solidity
contract Test {}
contract Derived is Test {}
function f(Test a, Test b) {
    a < b;
    a == b;
    a > b;
    a <= b;
    a >= b;
}

function g(Derived a, Test b) {
    a < b;
    a == b;
    a > b;
    a <= b;
    a >= b;
}
```
Currently, the above code compiles. I think the comparison operators does not belong with the ContractType. Even the equality operator is questionable. Disallowing all these operators is especially important to be consistent with how operators are disallowed by default in user defined types: https://github.com/ethereum/solidity/issues/11531.

Marenz:
fixes #11700
depends on #12566

### DIFF
--- a/test/libsolidity/syntaxTests/errors/contract_no_comparison.sol
+++ b/test/libsolidity/syntaxTests/errors/contract_no_comparison.sol
@@ -1,0 +1,21 @@
+contract C {
+	function f(string calldata data) external pure {
+		C c;
+		assert(this == this);
+
+		assert(this == c);
+		assert(this != c);
+		assert(this >= c);
+		assert(this <= c);
+		assert(this < c);
+		assert(this > c);
+	}
+}
+// ----
+// TypeError 2271: (79-91): Operator == not compatible with types contract C and contract C. No arithmetic or comparison operations are allowed on contract types. Consider converting to "address".
+// TypeError 2271: (104-113): Operator == not compatible with types contract C and contract C. No arithmetic or comparison operations are allowed on contract types. Consider converting to "address".
+// TypeError 2271: (125-134): Operator != not compatible with types contract C and contract C. No arithmetic or comparison operations are allowed on contract types. Consider converting to "address".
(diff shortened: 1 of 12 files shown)


ekpyron (commented):
(Changed the base branch, so that this becomes readable)

Besides the issue in updating breaking, the tests need adjusting.

test/libsolidity/syntaxTests/errors/contract_no_comparison.sol
```diff
@@ -0,0 +1,7 @@
+contract C {
+	function f(string calldata data) external pure {
+		bool test = this == this;
```
hrkrshnn: Can you add a couple more operations? Perhaps also for `C`.

ekpyron (approved):

ekpyron:
[rebased this using github's rebase button]

ekpyron:
This will need forking of the external test repos.

leonardoalt:
> This will need forking of the external test repos.

Bump this

cameel:
Since this is a bit stale anyway, maybe it would be a good idea to first submit some upstream PRs changing those comparisons? With a bit of luck, we'll be able to merge it without having to switch to forks. Or we'll at least reduce their number.

Marenz:
I changed it to a draft until our external tests are ready for this. Updated the issue accordingly as well.

leonardoalt:
any updates?

Marenz:
The current state here is that the external tests need to be updated;
This is open for take over at the moment.

cameel:
Looks like we don't have resources to pursue this now and the PR itself is not that complicated so I'd vote for closing this now and getting back to this later, when we can actually spend some time on getting upstream projects to adopt the change.

ekpyron:
Given the amount of external test failures, i.e. the extent of the real-world usage of these comparisons, I'm closing the PR. Actually, I'm even considering closing the issue, but I'll leave it open for the time being.

PR #12571, opened 2022-01-21
Force-enable stack-to-memory to see how it affects external tests.

ekpyron:


test/externalTests/ens.sh
```diff
@@ -46,7 +46,7 @@ function ens_test
         "${compile_only_presets[@]}"
         #ir-no-optimize           # Compilation fails with "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
         #ir-optimize-evm-only     # Compilation fails with "YulException: Variable var__945 is 1 slot(s) too deep inside the stack."
-        #ir-optimize-evm+yul      # Compilation fails with "YulException: Variable _5 is 1 too deep in the stack [ _5 usr$i usr$h _7 usr$scratch usr$k usr$f _4 usr$len usr$j_2 RET _2 _1 var_data_mpos usr$totallen usr$x _12 ]"
+        ir-optimize-evm+yul      # Compilation fails with "YulException: Variable _5 is 1 too deep in the stack [ _5 usr$i usr$h _7 usr$scratch usr$k usr$f _4 usr$len usr$j_2 RET _2 _1 var_data_mpos usr$totallen usr$x _12 ]"
```
ekpyron, line 49: Works: https://app.circleci.com/pipelines/github/ethereum/solidity/21737/workflows/3588ee0e-de5a-4b29-81ce-fe17cff72fbb/jobs/952234

test/externalTests/perpetual-pools.sh
```diff
@@ -45,7 +45,7 @@ function perpetual_pools_test
         "${compile_only_presets[@]}"
         #ir-no-optimize            # Compilation fails with "YulException: Variable var_amount_527 is 9 slot(s) too deep inside the stack."
         #ir-optimize-evm-only      # Compilation fails with "YulException: Variable var_amount_527 is 9 slot(s) too deep inside the stack."
-        #ir-optimize-evm+yul       # Compilation fails with "YulException: Variable expr_mpos is 1 too deep in the stack"
+        ir-optimize-evm+yul       # Compilation fails with "YulException: Variable expr_mpos is 1 too deep in the stack"
```
ekpyron, line 48: On a casual look, seems to compile but has testing issues https://app.circleci.com/pipelines/github/ethereum/solidity/21737/workflows/3588ee0e-de5a-4b29-81ce-fe17cff72fbb/jobs/952230

test/externalTests/trident.sh
```diff
@@ -53,7 +53,7 @@ function trident_test
         "${compile_only_presets[@]}"
         #ir-no-optimize            # Compilation fails with: "YulException: Variable var_amount_165 is 9 slot(s) too deep inside the stack."
         #ir-optimize-evm-only      # Compilation fails with: "YulException: Variable var_amount_165 is 9 slot(s) too deep inside the stack."
-        #ir-optimize-evm+yul       # Compilation fails with: "YulException: Cannot swap Variable var_nearestTick with Variable _4: too deep in the stack by 4 slots"
+        ir-optimize-evm+yul       # Compilation fails with: "YulException: Cannot swap Variable var_nearestTick with Variable _4: too deep in the stack by 4 slots"
```
ekpyron, line 56: Works https://app.circleci.com/jobs/github/ethereum/solidity/952233?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

test/externalTests/euler.sh
```diff
@@ -44,7 +44,7 @@ function euler_test
         "${compile_only_presets[@]}"
         #ir-no-optimize           # Compilation fails with "YulException: Variable var_utilisation_307 is 6 slot(s) too deep inside the stack."
         #ir-optimize-evm-only     # Compilation fails with "YulException: Variable var_utilisation_307 is 6 slot(s) too deep inside the stack."
-        #ir-optimize-evm+yul      # Compilation fails with "YulException: Variable var_status_mpos is 3 too deep in the stack"
+        ir-optimize-evm+yul      # Compilation fails with "YulException: Variable var_status_mpos is 3 too deep in the stack"
```
ekpyron, line 47: Looks like it compiles, but fails its tests: https://app.circleci.com/pipelines/github/ethereum/solidity/21737/workflows/3588ee0e-de5a-4b29-81ce-fe17cff72fbb/jobs/952228

Might be worth looking into. Could either be unrelated, or they have stack-to-memory incompatible inline assembly or a proper bug in the code transform.

ekpyron: @cameel You said something about euler tests breaking - but it's probably not what I'm seeing here, right? This looks like it fails specifically for the ``ir-optimize-evm+yul`` run, right?

cameel: No, the one I saw was completely different: https://app.circleci.com/pipelines/github/ethereum/solidity/21732/workflows/4b8b15e2-8d05-4a00-be6e-bd449ebf01dc/jobs/952012

```
Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/tmp/ext-test-Euler-7owwrh/ext/test/activateMarket.js)
```
```
Error: the string "expected error \"e/no-uniswap-pool-avail\" but no error was thrown" was thrown, throw an Error :)
```
```
Error: call revert exception (method="getPool(address,address,uint24)", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.5.0)
```

It was also a bunch of different errors and only a handful of tests failed. It went away after rerunning. Yours seem to be breaking tons of tests, all with the same error.

cameel: > This looks like it fails specifically for the `ir-optimize-evm+yul` run, right?

Yeah, it's only that one. I'm looking into it.

From the stack trace looks like the revert is coming from [`revertBytes()`](https://github.com/euler-xyz/euler-contracts/blob/70cfa81ef67a76c44ac1e28249cadecd34f00a13/contracts/Base.sol#L87-L92). This is just a helper through. The relevant code seems to be [this part in `getNewMarketParameters()`](https://github.com/euler-xyz/euler-contracts/blob/70cfa81ef67a76c44ac1e28249cadecd34f00a13/contracts/modules/RiskManager.sol#L71-L93) instead even though I don't see it directly listed in the trace (maybe line numbers are off?). It's the only place that throws that error message.

Now the question is why it reverts. To figure it out we'll likely need to extract this part of code and play with it. Unfortunately it looks like it has external calls to other deployed contracts so it might be a bit inconvenient to replicate other than just by launching Hardhat tests.

ekpyron: Yeah... the assembly in here
https://github.com/euler-xyz/euler-contracts/blob/70cfa81ef67a76c44ac1e28249cadecd34f00a13/contracts/Euler.sol#L49
and here
https://github.com/euler-xyz/euler-contracts/blob/70cfa81ef67a76c44ac1e28249cadecd34f00a13/contracts/Proxy.sol#L18
are actually stack-to-memory incompatible - so setting it to forced-enabled is random luck, maybe it's that.

ekpyron: And given that, it may not be worth it to investigate further for the time being.

cameel:
While looking at tests I gathered some timing stats for compilation:
| Project         | ir-optimize-evm+yul | legacy-optimize-evm+yul |
|-----------------|---------------------|-------------------------|
| ens             |   52s               | 14s                     |
| euler           |   59s               | 15s                     |
| trident         |  101s               | 39s                     |
| perpetual-pools |  103s               | 28s                     |

The value is the `user` time as reported by `time` utility.

test/externalTests/euler.sh
cameel: Oh, so they do what we used to do, i.e. assume that they can just write all over memory when they're about to revert? Would exposing the memory allocation helpers in assembly be a solution for porting cases like these to the new codegen?

Also, I think this is generally going to be a problem for people when porting. They might often not realize where the problem is coming from. Is there something we can do at compilation time to detect this and report an error?

test/externalTests/perpetual-pools.sh
cameel: This looks like https://github.com/nomiclabs/hardhat/issues/2115. I.e. a problem with Hardhat's heuristics not being compatible with IR yet. I got this in two other cases (OpenZeppelin and PRBMath) and the only thing we can do for now is to skip tests.

You can put it in the `compile_only_presets` array so that it at least gets compiled. Here's how I'm doing it for OZ: https://github.com/ethereum/solidity/blob/76540b338e0b354a1d55e33062c151a59fb67b50/test/externalTests/zeppelin.sh#L42-L44

ekpyron:
Yeah, I was actually wondering about that :-). Not sure if that's better or worse than I expected - it's not ideal in any case :-).

test/externalTests/euler.sh
ekpyron: I mean - that's why this stuff doesn't "just compile" right now - we disable the stack-to-memory stuff as soon as we cannot be sure that people don't just write to memory arbitrarily - currently rather crude by globally disabling it once there is any inline assembly anywhere whatsoever.

The plan is to allow people to annotate assembly as "memory safe" and then still do it - but we also need to be more fine-grained, i.e. *some* unsafe inline assembly in *some* paths should not prevent us from doing stack-to-memory anywhere else.

In this PR I just disabled all such safeguards to see what would happen - and for the others it seems fine - here it *might* be that the two snippets actually cause problems, but it may also be something completely unrelated...

The cases both seem proxy-related - in those cases people probably often avoid and want to avoid proper allocations to avoid the overhead...

cameel:
The interesting thing is that for example for ENS the whole compilation+test takes almost the same amount of time in both cases (only 6s difference). This probably means that faster execution compensates for longer compilation time.

In case of Trident IR it's different though - compilation+test takes ~1 min longer which is exactly the extra compilation overhead.

test/externalTests/euler.sh
ekpyron: Scrolling through it again, it looks to me like all errors have the Proxy fallback and ``Euler.dispatch`` in their trace and those exactly contain such unsafe inline assembly - so it's probably that. FYI there's only a problem if a stack error occurs *while* you're writing to unsafe memory locations - it's a bit surprising that this happens in these cases, but apparently it does...

cameel: Ah, right. In that case I think we really need to report it better regardless of how/if it gets fixed properly. If it fails due to "stack to deep" **and** stack-to-memory was disabled due to inline assembly, there should be a hint for the user that this might be the cause. Maybe we actually should have an INFO message every time this stuff gets disabled, regardless of "stack-too-deep".

ekpyron: Yep - although there's not much point in that until we support annotating inline assembly as "memory safe" (until we have that, we could just suggest in bulk not to use inline assembly at all :-)).
But that we do need soon.

What I'd also actually be curious about is the gas differences between ``ir-optimize-evm+yul`` amd ``legacy-optimize-evm+yul`` on all the ones that work both ways in this PR... but it can wait until we have the benchmark PR merged.

ekpyron:
> The interesting thing is that for example for ENS the whole compilation+test takes almost the same amount of time in both cases (only 6s difference). This probably means that faster execution compensates for longer compilation time.
> 
> In case of Trident IR it's different though - compilation+test takes ~1 min longer which is exactly the extra compilation overhead.

I almost can't imagine how this is possible - we may save a bit of gas, but so much as to that execution time compensates half a minute compile time? Maybe, but somehow seems a bit unlikely to me :-D.

cameel:
Looks like you're right. It was probably just a fluctuation in npm package installation time or something. I checked times between two CI runs of #12441, for those jobs that work both on IR and legacy, and I while I did see similar pattern for yield-liquidator in [today's run](https://app.circleci.com/pipelines/github/ethereum/solidity/21742/workflows/c87245cf-2188-44f0-93f4-7f6156601a18):

| Project          | ir (compilation) | ir (compilation+test) | legacy (compilation) | legacy (compilation+test) |
|------------------|------------------|-----------------------|----------------------|---------------------------|
| gnosis           | 0:17             | 3:22                  | 0:06                 | 3:14                      |
| yield-liquidator | 0:33             | 2:27                  | 0:21                 | 2:39                      |
| bleeps           | 2:10             | 7:24                  | 0:44                 | 5:20                      |

it was not present in an [earlier run from 2 days ago](https://app.circleci.com/pipelines/github/ethereum/solidity/21719/workflows/d6667709-78e3-4a89-9301-ae399fcb97eb):
| Project          | ir (compilation) | ir (compilation+test) | legacy (compilation) | legacy (compilation+test) |
|------------------|------------------|-----------------------|----------------------|---------------------------|
| gnosis           | 0:16             | 3:14                  | 0:06                 | 2:48                      |
| yield-liquidator | 0:42             | 2:40                  | 0:18                 | 2:11                      |
| bleeps           | 2:18             | 8:27                  | 0:36                 | 4:39                      |

Overall, other factors seem to have much bigger effect on total running time.

test/externalTests/euler.sh
cameel: > FYI there's only a problem if a stack error occurs while you're writing to unsafe memory locations - it's a bit surprising that this happens in these cases, but apparently it does...

Shows that we really do need more of these external projects. Some important things surfaced this week (this and the hole in IR coverage for array copying) so I'm wondering if it might actually be worthwhile to add even more of them. I was going to stop after I finish the ones I have half-working now (ElementFi, PRB Math, Gnosis GP2) and maybe Rarible and Brink if I can get them to work at all. Maybe I should continue instead?

ekpyron: I said it's surprising, since I didn't think these particular projects were likely to do something like this, but generally it's not surprising at all that cases like that occur :-). So it's really only the array copy thing.

So I think it's fine to stop after the half-working ones for the time being at least :-). The more of them it is the more likely it becomes that some will break, so maybe better to monitor the ones we have now or which will be added shortly for a while.

ekpyron:
Closing this for now, since it was only meant for checking what happens anyways.

PR #12574, opened 2022-01-21
External test for ElementFi

cameel:
Part of #5956.

This adds an external test for [ElementFi](https://github.com/element-fi/elf-contracts).
- The contracts are on 0.8.x but the compilation pulls in Balancer which is still on 0.7.x. Lots of patching needed due to that.
    - I actually have to explicitly run `scripts/load-balancer-contracts.sh` script to pull in Balancer.
- Tests via IR fail due to https://github.com/nomiclabs/hardhat/issues/2115.
-  It needs `allowUnlimitedContractSize`. Otherwise one test via IR was failing due to too high deployment cost.


cameel:
~The interesting thing about this one is that the compilation via IR reproducibly takes ca. 20 min on my machine, while in CI on `medium` (2 cores, 4 GB RAM) it's just ca. 3 min. In both cases it runs single-threaded of course.~

Ah, wait. Recently I've been using a debug build instead of a release binary due to Uniswap crashing on #12528. That's why it's so slow. Mystery solved :)

hrkrshnn:
Looks good. Needs a rebase.

cameel:
Rebased.

hrkrshnn (approved):

PR #5815, opened 2019-01-17
Strict abi decoder (validate incoming data instead of cleaning it)

axic:
Fixes #5844.

This shows the changes discussed in #3303. We can discuss it here and close this PR if it turns out to be a bad direction.

Todo:
- [x] merge bool change from #3303 
- [x] add validation case for integer type
- [x] update tests

It fixes the following currently:
- enums do not generate a revert function on encoding
- ~~decoding address/contract will not generate code with "invalid()" if revert was requested (this can easily be pulled into a separate PR)~~ [moved to #5843]
- decoding dirty FixedBytes generates a revert/abort

PR #12575, opened 2022-01-22
External test for Gnosis Protocol v2

cameel:
Part of #5956.
~Depends on #12868.~ Merged.
~Depends on  #12580.~ Merged.

This adds an external test for [Gnosis Protocol v2](https://github.com/gnosis/gp-v2-contracts).
- Only minimal patching required for it to work on 0.8.x. This includes one file in OpenZeppelin.
    - OZ can't easily be upgraded to 4.0 because the project imports some files which were either moved or removed.
- Needed to delete pre-built artifacts that get installed with `@gnosis.pm/safe-contracts` npm package.
- Tests via IR fail due to https://github.com/nomiclabs/hardhat/issues/2115.
- It needs `allowUnlimitedContractSize`. Otherwise one test via IR was failing due to too high deployment cost.

Still TODO:
- [x] One of the E2E tests (`contractOrdersWithGnosisSafe.test.ts`) depends on being able to import the pre-built artifacts from Gnosis Safe. We always remove all such artifacts in external tests to ensure that we're dealing only with the output of our locally built compiler binary. The problem is that the npm package does not come with Hardhat config so it's not that easy to rebuild in place.
    - The solution might be to just remove the package and instead clone the repo and build it.


test/externalTests/gp2.sh
```diff
@@ -0,0 +1,105 @@
...
+    sed -i 's|uint256(-1)|type(uint256).max|g' src/contracts/GPv2Settlement.sol
+    sed -i 's|return msg\.sender;|return payable(msg.sender);|g' node_modules/@openzeppelin/contracts/utils/Context.sol
+    perl -i -0pe \
+        "s/uint256 (executedBuyAmount = \(-tokenDeltas\[trade.buyTokenIndex\]\)\n\s+.toUint256\(\);)/uint256 executedBuyAmount; unchecked \{\1\}/g" \
+        src/contracts/GPv2Settlement.sol
+
+    # This test is not supposed to work. The compiler is supposed to enforce zero padding since
+    # at least 0.5.8 (see https://github.com/ethereum/solidity/pull/5815). For some reason the
+    # test worked on 0.7.6 but no longer works on 0.8.x.
+    sed -i 's|it(\("invalid EVM transaction encoding does not change order hash" \)|it.skip(\1|g' test/GPv2Signing.test.ts
```
cameel: I don't know why this test works for them on 0.7.6. For now I just know that #5815 added checks against padding being non-zero and that PR was merged in 0.5.8. Was it buggy before 0.8.0 or something?

leonardoalt:
@cameel does this need any updates after the dependency PR got merged?

cameel:
Same as for #12580, this still needs some work. In this case I managed to resolve issues in all presets except for `legacy-no-optimize` last week. The dependency was only to get test results not influenced by breakage in unrelated external tests.

cameel:
This is now ready for review. I had to make `legacy-no-optimize` compile-only because, despite telling Hardhat not to enforce contract size limit, a contract doing `new GPv2VaultRelayer` fails.

ekpyron (dismissed):

cameel:
Rebased on current `develop` to make sure it didn't break in the meantime due to some change in dependencies.

Also, turns out that the upstream repo is now archived and the code was migrated to https://github.com/cowprotocol/contracts :) It still works though so I'd like to merge this anyway and I switch to the new repo in a separate PR.

ekpyron (approved):

PR #12580, opened 2022-01-24
External test for Chainlink

cameel:
Part of #5956.

This adds an external test for [Chainlink](https://github.com/smartcontractkit/chainlink).

Making it work with the latest compiler required patching the source and also disabling quite a few tests failing either due to Hardhat heuristics or because of hard-coded gas checks. I originally thought that the repo contains independent contracts for each breaking release (it has separate dirs called `v0.4`, `v0.6`, `v0.7`, `v0.8`) and would be easy to integrate. Unfortunately it looks like newer contracts depend on older ones, compiled with older compilers. This all had to be made to work together.

The ported contract is in [`develop_080` branch in our fork](https://github.com/solidity-external-tests/chainlink/tree/develop_080). Please take a look at the fork as a part of this review. There are 23 extra commits, each fixing a specific incompatibility across all contracts.


test/externalTests/chainlink.sh
```diff
@@ -0,0 +1,149 @@
...
+BINARY_PATH="$2"
+
+function compile_fn { yarn compile; }
+function test_fn { yarn test; }
+
+function chainlink_test
+{
+    local repo="https://github.com/smartcontractkit/chainlink"
+    local ref_type=branch
+    local ref=develop
```
axic: Shouldn't this use some release if we make so many changes?

cameel: Yeah. I think I'm actually going to fork it and point that at a branch in that fork. It's just too much patching. Fortunately it's mostly their 0.4.x token but it does require a lot fo simple tweaks.

leonardoalt:
@cameel does this need any updates after the dependency PR got merged?

cameel:
No, this PR still needs some work. Specifically, patching for tests to pass on 0.8.x. I was working on it last week but was interrupted by the ext test breakage - the dependency was only to get test results not influenced by that breakage.

cameel:
I'm finally done here. It should be passing all tests now so I'm marking it as ready to review.

ekpyron (dismissed):

cameel:
Rebased on current `develop` to make sure it didn't break in the meantime due to some change in dependencies.

ekpyron (approved):

PR #12583, opened 2022-01-25
Compatibility fixes for solc-js TypeScript migration

cameel:
This PR adjusts the command we run from solc-js and adds `npm run build` required to compile the TypeScript code. It should be merged after https://github.com/ethereum/solc-js/pull/566.

### DIFF
--- a/scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh
+++ b/scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh
@@ -228,6 +228,7 @@ mkdir -p "${OUTPUTDIR}"/bin
 echo "Prepare solc-js."
 cd /root/solc-js
 npm install >/dev/null 2>&1
+npm run build >/dev/null 2>&1
 
 echo "Install semver helper."
 npm install -g semver >/dev/null 2>&1
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,7 @@ Bugfixes:
 
 Solc-Js:
  * The wrapper now requires at least nodejs v10.
+ * The code has been ported to TypeScript.
 
 
 ### 0.8.11 (2021-12-20)
--- a/ReleaseChecklist.md
+++ b/ReleaseChecklist.md
@@ -66,7 +66,7 @@
 ### Release solc-js
  - [ ] Wait until solc-bin was properly deployed. You can test this via remix - a test run through remix is advisable anyway.
  - [ ] Increment the version number, create a pull request for that, merge it after tests succeeded.
- - [ ] Run ``node verifyVersion.js && npm publish`` in the updated ``solc-js`` repository.
+ - [ ] Run ``npm run updateBinary && npm publish`` in the updated ``solc-js`` repository.
  - [ ] Create a tag using ``git tag --annotate v$VERSION`` and push it with ``git push --tags``.
 
 ### Post-release
--- a/scripts/bytecodecompare/prepare_report.js
+++ b/scripts/bytecodecompare/prepare_report.js
@@ -2,7 +2,7 @@
 const process = require('process')
 const fs = require('fs')
 
-const compiler = require('./solc-js/wrapper.js')(require('./solc-js/soljson.js'))
+const compiler = require('solc')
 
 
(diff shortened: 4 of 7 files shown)


ekpyron (dismissed):

cameel:
I found two more places where we do not run `solcjs` but we do `npm install` on solc-js. I think they now need `npm run build` too. PR updated.

scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh
```diff
@@ -228,6 +228,7 @@ mkdir -p "${OUTPUTDIR}"/bin
 echo "Prepare solc-js."
 cd /root/solc-js
 npm install >/dev/null 2>&1
+npm run build >/dev/null 2>&1
```
cameel, line 231: Not sure about this one and these scripts are not actively tested but I'm adding it just in case.

ekpyron (dismissed):

ekpyron:
Tests are still failing.
Also: does the release checklist need adjustment or is it fine?

cameel:
> Also: does the release checklist need adjustment or is it fine?

Right, it does. You cannot run `verifyVersion.js` directly. You need `npx ts-node verifyVersion.ts` instead.
But I think we should replace it with `npm run updateBinary`. It does this but also downloads the binary. I think we should have changed it to that even earlier, back when we made it not download a new binary automatically.

cameel:
- Checklist updated.
- Added changelog entry for TypeScript migration.
- Bytecode comparison was failing because the JS snippet was importing specific .js files directly instead of doing it via the public interface. I changed it completely so that now the package is installed via npm from a local dir and we can just import `solc`.
- solc-js and Colony external tests were failing because now `soljson.js` is actually under `dist/`. `npm run build` copies it there from the old location but only if we put it in that old location before running the command. That was not the case.
    - It's only Colony test that failed because it's the only one that's still on solc-js rather than native binary. I think we should keep one working like that just to have this code path somewhat tested.

ekpyron:
Still something off in the tests apparently...

cameel:
Pushing a fix for bytecode comparison (turns out I still need to run `npm install` and `npm run build` inside the directory I install a local package from).

Still working on the other fail.

cameel:
The test failure in solc-js external test is actually a bug in solc-js itself. The failing test is buggy but gets skipped due to missing Z3. It does not get skipped here :) We need to merge https://github.com/ethereum/solc-js/pull/592 to fix it.

cameel:
The problem with Colony is harder. I think we may have actually broken backwards-compatibility with the Typescript change.
It fails with `TypeError: soljson.cwrap is not a function` in Trufle's [`Local.load()`](https://github.com/trufflesuite/truffle/blob/v5.4.30/packages/compile-solidity/src/compilerSupplier/loadingStrategies/Local.ts#L7-L25).

Truffle does `soljson = originalRequire(compilerPath)`, where `compilerPath` is actually a path to the directory containing solc-js checkout (because that's what I'm giving it in the test). In JS that's fine and `soljson.compile` function is present. Now, however, it's under `soljson.default.compile` due to `default export` in our `wrapper.js` and `soljson.compile` is `undefined`.

From what I'm reading, not having `default` in there was either some kind of hack or it's something that changed by design in TypeScript. In any case, we probably need to change how we export stuff because otherwise JS tools will be able to import it without changes on their side.

cameel:
@stephensli What do you think about https://github.com/ethereum/solidity/pull/12583#issuecomment-1021430071? Is that actually a breaking change or is there some easy workaround?

axic:
So maybe we do need to do those manual function exports, so they are named.

stephen-slm:
https://github.com/ethereum/solc-js/pull/593

This is the way to get around, it should now function as expected, if you could use my changes to just verify that first, but this should be it. 

instead of exporting the default as the object, we rexport the entire content.

ekpyron:
Force-pushed to retrigger tests after merging https://github.com/ethereum/solc-js/pull/593
 [by the way a bit annoying that circle CI doesn't easily let you retrigger the entire bytecode comparison workflow]

ekpyron (approved):

PR #12584, opened 2022-01-25
Code generators needed fixing of the cleanup process during typecasting

### LINKED ISSUE #12535: [Logging] Indexed log topic differs between legacy and IR if explicitly downcast
```
contract C {
  event ev0(bytes1 indexed);
  constructor() {
    emit ev0(bytes1(bytes16(0x31313131313131313131313131313131)));
  }
}
// ====
// compileViaYul: also
// ----
// constructor() ->
// ~ emit ev0(bytes1): #"1"
```

Legacy truncates indexed log topic, while IR leaves it untruncated. This would lead to a divergence of blockchain state depending on which backend is used.

To reproduce

```
$ cp test.sol test/libsolidity/semanticTests
$ isoltest -t semanticTests/test
  Running via Yul:
  Expected result:
  // constructor() ->
  // ~ emit ev0(bytes1): #"1"

  Obtained result:
  // constructor() ->
  // ~ emit ev0(bytes1): #"1111111111111111"
  
  Attention: Updates on the test will apply the detected format displayed.
  
  Note that the test passed without Yul. 
```

Note that the divergence is maintained with/without optimizer so I would guess this may be a bug related to truncation (or lack thereof) in Sol->Yul.

nishant-sachdeva:
Typecasting process for "bytes" and "integers" has problems in the Old and New Code Generators

1. In the New Code generator, for the bytes type, it was sending the "from" type for cleanup instead of the "to" type.
2. Errors in the old code generator are yet to be identified

Closes #12535

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,8 @@ Bugfixes:
  * Code Generator: Fix ICE when doing an explicit conversion from ``string calldata`` to ``bytes``.
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the parent contract contains immutable variables.
+ * IR Generator: Add missing cleanup during the conversion of fixed bytes types to smaller fixed bytes types.
+ * IR Generator: Add missing cleanup for indexed event arguments of value type.
  * IR Generator: Fix IR syntax error when copying storage arrays of structs containing functions.
  * Natspec: Fix ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different.
  * TypeChecker: Fix ICE when a constant variable declaration forward references a struct.
--- a/libsolidity/codegen/YulUtilFunctions.cpp
+++ b/libsolidity/codegen/YulUtilFunctions.cpp
@@ -3400,11 +3400,11 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 						.render();
 			else
 			{
-				// clear for conversion to longer bytes
 				solAssert(toCategory == Type::Category::FixedBytes, "Invalid type conversion requested.");
+				FixedBytesType const& to = dynamic_cast<FixedBytesType const&>(_to);
 				body =
 					Whiskers("converted := <clean>(value)")
-					("clean", cleanupFunction(from))
+					("clean", cleanupFunction((to.numBytes() <= from.numBytes()) ? to : from))
 					.render();
 			}
 			break;
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
@@ -1035,7 +1035,10 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
 						")\n";
 				}
 				else
-					indexedArgs.emplace_back(convert(arg, *paramTypes[i]));
+				{
+					solAssert(parameterTypes[i]->sizeOnStack() == 1, "");
+					indexedArgs.emplace_back(convert(arg, *paramTypes[i], true));
+				}
 			}
 			else
 			{
@@ -2724,14 +2727,14 @@ void IRGeneratorForStatements::assignInternalFunctionIDIfNotCalledDirectly(
 	m_context.addToInternalDispatch(_referencedFunction);
 }
 
-IRVariable IRGeneratorForStatements::convert(IRVariable const& _from, Type const& _to)
+IRVariable IRGeneratorForStatements::convert(IRVariable const& _from, Type const& _to, bool _forceCleanup)
 {
-	if (_from.type() == _to)
+	if (_from.type() == _to && !_forceCleanup)
 		return _from;
 	else
 	{
 		IRVariable converted(m_context.newYulVariable(), _to);
-		define(converted, _from);
+		define(converted, _from, _forceCleanup);
 		return converted;
 	}
 }
@@ -2763,10 +2766,10 @@ void IRGeneratorForStatements::declare(IRVariable const& _var)
 		appendCode() << "let " << _var.commaSeparatedList() << "\n";
 }
 
-void IRGeneratorForStatements::declareAssign(IRVariable const& _lhs, IRVariable const& _rhs, bool _declare)
+void IRGeneratorForStatements::declareAssign(IRVariable const& _lhs, IRVariable const& _rhs, bool _declare, bool _forceCleanup)
 {
 	string output;
-	if (_lhs.type() == _rhs.type())
+	if (_lhs.type() == _rhs.type() && !_forceCleanup)
 		for (auto const& [stackItemName, stackItemType]: _lhs.type().stackItems())
 			if (stackItemType)
 				declareAssign(_lhs.part(stackItemName), _rhs.part(stackItemName), _declare);
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.h
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.h
@@ -86,7 +86,11 @@ public:
 	IRVariable evaluateExpression(Expression const& _expression, Type const& _to);
 
 	/// Defines @a _var using the value of @a _value while performing type conversions, if required.
-	void define(IRVariable const& _var, IRVariable const& _value) { declareAssign(_var, _value, true); }
+	/// If @a _forceCleanup is set to true, it also cleans the value of the variable after the conversion.
+	void define(IRVariable const& _var, IRVariable const& _value, bool _forceCleanup = false)
+	{
+		declareAssign(_var, _value, true, _forceCleanup);
+	}
 
 	/// @returns the name of a function that computes the value of the given constant
 	/// and also generates the function.
@@ -162,8 +166,8 @@ private:
 	);
 
 	/// Generates the required conversion code and @returns an IRVariable referring to the value of @a _variable
-	/// converted to type @a _to.
-	IRVariable convert(IRVariable const& _variable, Type const& _to);
+	/// If @a _forceCleanup is set to true, it also cleans the value of the variable after the conversion.
+	IRVariable convert(IRVariable const& _variable, Type const& _to, bool _forceCleanup = false);
 
 	/// @returns a Yul expression representing the current value of @a _expression,
 	/// converted to type @a _to if it does not yet have that type.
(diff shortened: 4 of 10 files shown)


libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3404,7 +3404,7 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 				solAssert(toCategory == Type::Category::FixedBytes, "Invalid type conversion requested.");
 				body =
 					Whiskers("converted := <clean>(value)")
-					("clean", cleanupFunction(from))
+					("clean", cleanupFunction(_to))
```
ekpyron: What this should really do is probably to use the cleanup function of the smaller of the types ``from`` and ``_to``.

Can you verify that we have tests for the following and they don't break:
```
ontract C {
    function g() external returns (bytes32 y) {
        bytes1 x;
        assembly { x := sub(0,1) }
        y = x;
    }

}
// ====
// compileViaYul: also
// ----
// g() -> 0xff00000000000000000000000000000000000000000000000000000000000000
```
?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3387,6 +3387,7 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 		case Type::Category::FixedBytes:
 		{
 			FixedBytesType const& from = dynamic_cast<FixedBytesType const&>(_from);
+			FixedBytesType const& to = dynamic_cast<FixedBytesType const&>(_to);
```
ekpyron: You can only do this cast after the ``solAssert(toCategory == Type::Category::FixedBytes, "Invalid type conversion requested.");`` below, otherwise it will throw depending on what ``_to`` is.

ekpyron:
The idea was not to generally *always* call the conversion functions at *all* times, but selectively enforce the conversion when emitting events, i.e. in particular in IRGeneratorForStatements.cpp:L1038 only :-).

ekpyron:
Actually, looking at IRGeneratorForStatements.cpp:L1038 a bit more closely - you can probably get away with asserting that ``paramTypes[i]`` has ``sizeOnStack() == 1`` and then use ``expressionAsType(arg, *paramTypes[i], true)`` instead of using ``convert(arg, *paramTypes[i])`` (and thus you can leave ``convert`` and ``declareAssign`` unchanged).

test/libsolidity/semanticTests/cleanup/cleanup_bytes_types_shortening.sol
```diff
@@ -12,6 +12,6 @@ contract C {
 }
 // ====
 // compileToEwasm: also
-// compileViaYul: also
+// compileViaYul: true
```
ekpyron: You can revert this change now and the test will still pass, won't it?

nishant-sachdeva: Nope, the test does not pass if I revert this change.
New Code and OId Code Generators give conflicting outputs

ekpyron: Ah, right, now I see the diff in the expectation correctly.
Then please split this test into two tests, one for old and one for new codegen and adjust the comment in the test case accordingly.

test/libsolidity/semanticTests/cleanup/cleanup_bytes_types_shortening_newCodeGen.sol
```diff
@@ -0,0 +1,17 @@
+contract C {
+    function f() public pure returns (bytes32 r) {
+        bytes4 x = 0xffffffff;
+        bytes2 y = bytes2(x);
+        assembly {
+            r := y
+        }
+        // At this point, r stores four bytes and y stores 2 bytes
+        // y is properly cleaned before the equality check
+        require(y == bytes2(0xffff));
```
ekpyron: These three lines can actually be removed - the fact that `y` is already cleaned is apparent in the return value.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3402,9 +3402,10 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 			{
 				// clear for conversion to longer bytes
 				solAssert(toCategory == Type::Category::FixedBytes, "Invalid type conversion requested.");
+				FixedBytesType const& to = dynamic_cast<FixedBytesType const&>(_to);
 				body =
 					Whiskers("converted := <clean>(value)")
-					("clean", cleanupFunction(from))
+					("clean", cleanupFunction((to.numBytes() <= from.numBytes())? to : from))
```
ekpyron:
```suggestion
					("clean", cleanupFunction((to.numBytes() <= from.numBytes()) ? to : from))
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1035,7 +1035,10 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
 						")\n";
 				}
 				else
-					indexedArgs.emplace_back(convert(arg, *paramTypes[i]));
+				{
+					solAssert((*parameterTypes[i]).sizeOnStack() == 1, "Size of indexed argument is not 1.");
```
ekpyron:
```suggestion
					solAssert(parameterTypes[i]->sizeOnStack() == 1, "");
```

ekpyron (commented):
Apart from some minor style suggestions, this mainly needs a Changelog entry

Changelog.md
```diff
@@ -19,6 +19,7 @@ Bugfixes:
  * IR Generator: Fix IR syntax error when copying storage arrays of structs containing functions.
  * Natspec: Fix ICE when overriding a struct getter with a Natspec-documented return value and the name in the struct is different.
  * TypeChecker: Fix ICE when a constant variable declaration forward references a struct.
+ * Code Generator: Fix differences in variable cleanup procedured of Old and New Code Generators.
```
```suggestion
 * IR Generator: Add missing cleanup during the conversion of fixed bytes types to smaller fixed bytes types.
 * IR Generator: Add missing cleanup for indexed event arguments of value type.
```

ekpyron: Also note that the Changelog is sorted alphabetically.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -3402,9 +3402,10 @@ string YulUtilFunctions::conversionFunction(Type const& _from, Type const& _to)
 			{
 				// clear for conversion to longer bytes
```
ekpyron: This comment also no longer applies - it clears for conversion to longer and shorter bytes types.

test/libsolidity/semanticTests/cleanup/cleanup_bytes_types_shortening_OldCodeGen.sol
```diff
@@ -0,0 +1,13 @@
+contract C {
+    function f() public pure returns (bytes32 r) {
+        bytes4 x = 0xffffffff;
+        bytes2 y = bytes2(x);
+        assembly {
+            r := y
+        }
```
ekpyron: I only meant to remove the
```
        // At this point, r and y both store four bytes, but
        // y is properly cleaned before the equality check
        require(y == bytes2(0xffff));
```
part in the new code gen version. In the test for old code gen we should keep that.

test/libsolidity/semanticTests/cleanup/cleanup_bytes_types_shortening_newCodeGen.sol
```diff
@@ -5,13 +5,13 @@ contract C {
         assembly {
             r := y
         }
-        // At this point, r and y both store four bytes, but
+        // At this point, r stores four bytes and y stores 2 bytes
         // y is properly cleaned before the equality check
         require(y == bytes2(0xffff));
```
ekpyron: Haha, sorry for this apparently being confusing :-). But the idea is to keep the *original*, *unmodified* comment and ``require`` in the ``OldCodeGen`` file and *not* have those three linkes in the ``newCodeGen`` file :-).

ekpyron: In the new code gen test, the output of ``r`` already shows that the conversion properly did the cleanup.
In the old code gen test, the output shows that the conversion did *not* do any cleanup. That's why in the *old* codegen file, we want to make sure that the comparison still works as expected.
Clear now :-)?

Changelog.md
```diff
@@ -16,6 +16,8 @@ Bugfixes:
  * Code Generator: Fix ICE when doing an explicit conversion from ``string calldata`` to ``bytes``.
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
```
Aivolt1, line 17: Approved

Aivolt1 (dismissed):

libsolidity/codegen/ir/IRGeneratorForStatements.h
```diff
@@ -86,7 +86,10 @@ class IRGeneratorForStatements: public IRGeneratorForStatementsBase
 	IRVariable evaluateExpression(Expression const& _expression, Type const& _to);
 
 	/// Defines @a _var using the value of @a _value while performing type conversions, if required.
-	void define(IRVariable const& _var, IRVariable const& _value) { declareAssign(_var, _value, true); }
+	void define(IRVariable const& _var, IRVariable const& _value, bool _forceCleanup = false)
```
ekpyron: One last thing ;-). It might be nice to document ``_forceCleanup`` in the comment here (and also for ``convert``), similar to the comment at ``expressionAsType``.

Aivolt1 (approved):

test/libsolidity/semanticTests/userDefinedValueType/erc20.sol
```diff
@@ -115,7 +115,7 @@ contract ERC20 {
 // ----
 // constructor()
 // ~ emit Transfer(address,address,uint256): #0x00, #0x1212121212121212121212121212120000000012, 0x14
-// gas irOptimized: 442239
+// gas irOptimized: 447831
```
ekpyron, line 118: We should (separately) investigate the increase in gas cost here and check if it can be avoided. But for now I'll approve and merge this PR.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -1035,7 +1035,10 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
 						")\n";
 				}
 				else
-					indexedArgs.emplace_back(convert(arg, *paramTypes[i]));
+				{
+					solAssert(parameterTypes[i]->sizeOnStack() == 1, "");
```
ekpyron, line 1039: Actually, since we now ended up not using ``expressionAsType``, we don't really need this assertion anymore - so we could remove it again. I'll approve nonetheless.

ekpyron (approved):

PR #12586, opened 2022-01-26
Code transform fuzzer: Filter log statements and remove gas comparison.

bshastry:
The New Yul code transform fuzzer would fail if the gas consumed by the call to a function optimized via the new code transform is greater than the gas consumed by the call to a function optimized via the legacy optimizer by `1000` gas  (arbitrary threshold). This created a problem for Yul code such as 

```
{
  addmod(i1, i2, keccak256(i3, i4))
}
```

because the new Yul code transformed code would always evaluate the keccak call first in corner cases (and fail the gas compare assertion).

The second problem that the fuzzer had was that log statements would lead to false positives because

```
{
  log1(i1, i2, i3)
}
```

would log different values depending on the content of `i1` and `i2`.

This PR fixes these two problems
  - by disabling gas comparison
  - by disabling log statement generation

### DIFF
--- a/test/tools/ossfuzz/StackReuseCodegenFuzzer.cpp
+++ b/test/tools/ossfuzz/StackReuseCodegenFuzzer.cpp
@@ -71,11 +71,6 @@ DEFINE_PROTO_FUZZER(Program const& _input)
 		of.write(yul_source.data(), static_cast<streamsize>(yul_source.size()));
 	}
 
-	// Do not proceed with tests that are too large. 1200 is an arbitrary
-	// threshold.
-	if (yul_source.size() > 1200)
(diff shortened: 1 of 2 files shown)


ekpyron:
I'm not quite sure I understand - why is ``log`` a problem? Shouldn't the values always be the same?

ekpyron:
Ah, ``log*`` reads from memory and we don't have a guaranteed 1:1 match of memory contents... yeah, makes sense...

ekpyron (approved):

ekpyron:
I'm not entirely clear on the issue with the gas threshold, but dropping it should nonetheless be fine.

Issue #11969, opened 2021-09-15
Operators for user defined value types.


hrkrshnn:
This came up in https://github.com/ethereum/solidity/issues/9211#issuecomment-911705547

**Proposal**:

```solidity
type MyInt is int;
using {add as +} for UncheckedInt;
function add(MyInt a, MyInt b) view returns(MyInt c) { ... }
// MyInt.wrap(5) + MyInt.wrap(10) will now call add(...)
```

---

### Allowing mixed types

```solidity
using {addInt as +} for UncheckedInt;
function addInt(MyInt a, int b) view returns(MyInt c) { ... }
// MyInt.wrap(5) + 10 will now call addInt(...)
```

Useful for fixed points (`Fixed * uint`), duration type system (`Timestamp + Duration = Timestamp`, `Timestamp + Timestamp` is error), elliptic curve multiplication (`Point * uint`), etc.  Also can make code less verbose: see in example how we can write `MyInt.wrap(5) + 10` instead of `MyInt.wrap(5) + MyInt.wrap(10)`.

Problems: 
1. associativity of operators can be problematic (see https://github.com/ethereum/solidity/issues/9211#issuecomment-918392653 and  https://github.com/ethereum/solidity/issues/9211#issuecomment-915922517). May actually be solved by enforcing parenthesis for expressions that can be vague.
2. @ekpyron argues that this will cause issues with generics in the future.

### Disallowing mixed types

All types will have the type: `'a -> 'a -> 'a`, i.e., only operate on the same type.

To allow applications mentioned before, allow defining user defined operators.
```solidity
infix function <*>(UFixed18 a, uint b) returns (uint) { ... }
```

Problems: might complicate our parser, we'll need to assign left / right binding to operators, precedence etc. May be easier for the language to introduce new operators like `+', *'` or force fixed delimiters for new operators.

#### Allowing implicit conversions

If we only allow operators of type  `'a -> 'a -> 'a`, one way to do `Fixed * uint` would be to have a mechanism that allows implicit `uint -> Fixed` conversion. Thus, this would bind to the `* :: Fixed -> Fixed -> Fixed`. This is similar to how the compiler already implements arithmetic: for example `uint(5) + uint8(20)`.

Problem: the duration type system may not work in this case. Assuming we defined `Duration + Duration = Duration`, this allows `Timestamp + Duration = Duration`. **TODO**.

### Checked and Unchecked operator

If we allow custom operators, then should `unchecked` block affect it?

[...]

PR #12588, opened 2022-01-26
Update `IterableMapping` example to indicate that the calculations will overflow on 0.8.x

CryptoKiddies:
added a caveat in the comment above `iterate_start` as a notice, since fixing the issue will require a reworking of the library

### DIFF
--- a/docs/types/mapping-types.rst
+++ b/docs/types/mapping-types.rst
@@ -165,7 +165,8 @@ the ``sum`` function iterates over to sum all the values.
         function contains(itmap storage self, uint key) internal view returns (bool) {
             return self.data[key].keyIndex > 0;
         }
-
+        
+        // won't work with compiler 0.8 overflow checks
         function iterate_start(itmap storage self) internal view returns (uint keyIndex) {
             return iterate_next(self, type(uint).max);
         }


docs/types/mapping-types.rst
```diff
@@ -165,7 +165,8 @@ the ``sum`` function iterates over to sum all the values.
         function contains(itmap storage self, uint key) internal view returns (bool) {
             return self.data[key].keyIndex > 0;
         }
-
+        
+        // won't work with compiler 0.8 overflow checks
```
ekpyron, line 169: That's a good spot, thank you very much :-). Rather than adding this comment we should probably fix the example to work properly with 0.8, though...

ekpyron: On a very quick glance, a possible implementation would probably be something like
```
function iterate_start(itmap storage self) internal view returns (uint keyIndex) {
  if (0 < self.keys.length && self.keys[0].deleted)
            return iterate_next(self, 0);
  else
            return 0;
}
```
But might be worth spending some more thought on it.

chriseth:
If we want to further improve this, we should turn the library into a set of file-level functions and change the iterator from `uint` to a user-defined value type.

leonardoalt:
> If we want to further improve this, we should turn the library into a set of file-level functions and change the iterator from `uint` to a user-defined value type.

@CryptoKiddies do you think you could apply this suggestion? Thanks!

docs/types/mapping-types.rst
CryptoKiddies: I thought about this as well, but there's that efficiency hit. Hate those one time checks! I'll offer up something this week when I have time.

CryptoKiddies:
Incidentally, I'd rename `<IndexValue.keyindex>` to `<IndexValue.keyIndexWithOffset>` to document the `+1` offset which makes this logic work

CryptoKiddies:
> @CryptoKiddies do you think you could apply this suggestion? Thanks!

Don't think I can proceed with the solution in mind until operators for user-defined value types is implemented: https://github.com/ethereum/solidity/issues/11969

leonardoalt:
> Don't think I can proceed with the solution in mind until operators for user-defined value types is implemented: #11969

True, you'd need a bunch of wrap/unwrap or a bunch of new internal functions right now.

chriseth:
Sorry, I went ahead and implemented it here using user-defined value types: https://github.com/ethereum/solidity/pull/12721

ekpyron:
Fixed in https://github.com/ethereum/solidity/pull/12721

CryptoKiddies:
Looking forward to user-defined value types support. I'll push that example refinement in then, if @chriseth doesn't beat me to it!

PR #12589, opened 2022-01-27
Emit immutable references for pure yul code

### LINKED ISSUE #11044: Compiling Yul doesn't produce immutableReferences even when setimmutable/loadimmutable is used
## Description

Compiling Yul that uses `linkersymbol` will produce link references in the compiler output, but compiling Yul that uses `setimmutable`/`loadimmutable` doesn't produce immutable references in the compiler output.

So, for why this is relevant -- this goes along with #11043 as one of the issues I encountered when trying to add [Yul support into Truffle](https://github.com/trufflesuite/truffle/issues/1906).

Why is this an issue?  It's for performing bytecode matching; there are a number of contexts where we're given a piece of externally-obtained bytecode and we want to know what contract class it is (so we can get an appropriate source mpa, etc).   In order to do this, we need to know which parts of any given compiled bytecode may possibly vary, so that when we attempt to match other bytecode against it, we don't get a false negative.  To do this we rely on both link references and immutable references (and knowledge of the delegatecall guard) to say, hey, these parts of the bytecode are allowed to vary.  So, the lack of immutable references would be a problem in that we wouldn't be able to recognize instances of Yul contracts that made use of `setimmutable`/`loadimmutable`.

So, this would be useful to have.  There is the problem that they couldn't be keyed by AST ID like with Solidity, since Yul ASTs don't have node IDs, but maybe they could just use the immutable name or something (which would be unambiguous in this context).  This probably wouldn't be used for decoding in the debugger or anything, but we'd want to have some way of knowing which places in the code might vary for bytecode matching.

(I mean technically since it's Yul *anywhere* in the code might vary, so obviously this couldn't be done with full reliability, but we'd still like to handle the cases that we can.)

## Environment

- Compiler version: 0.8.1

## Steps to Reproduce

```yul
object "YulTest" {
  code {
    let size := datasize("runtime")
    datacopy(0, dataoffset("runtime"), size)
    setimmutable(0, "test", 1)
    return(0, size)
  }
  object "runtime" {
    code {
      mstore(0, loadimmutable("test"))
      return(0, 0x20)
    }
  }
}
```

Marenz:
fixes #11044

### DIFF
--- a/libsolidity/interface/StandardCompiler.cpp
+++ b/libsolidity/interface/StandardCompiler.cpp
@@ -1461,36 +1461,36 @@ Json::Value StandardCompiler::compileYul(InputsAndSettings _inputsAndSettings)
 	stack.optimize();
 
 	MachineAssemblyObject object;
-	MachineAssemblyObject runtimeObject;
-	tie(object, runtimeObject) = stack.assembleWithDeployed();
+	MachineAssemblyObject deployedObject;
+	tie(object, deployedObject) = stack.assembleWithDeployed();
 
 	if (object.bytecode)
 		object.bytecode->link(_inputsAndSettings.libraries);
-	if (runtimeObject.bytecode)
-		runtimeObject.bytecode->link(_inputsAndSettings.libraries);
+	if (deployedObject.bytecode)
+		deployedObject.bytecode->link(_inputsAndSettings.libraries);
 
-	for (string const& objectKind: vector<string>{"bytecode", "deployedBytecode"})
+	for (auto&& [kind, isDeployed]: {make_pair("bytecode"s, false), make_pair("deployedBytecode"s, true)})
 		if (isArtifactRequested(
 			_inputsAndSettings.outputSelection,
 			sourceName,
 			contractName,
-			evmObjectComponents(objectKind),
+			evmObjectComponents(kind),
 			wildcardMatchesExperimental
 		))
 		{
-			MachineAssemblyObject const& o = objectKind == "bytecode" ? object : runtimeObject;
+			MachineAssemblyObject const& o = isDeployed ? deployedObject : object;
 			if (o.bytecode)
-				output["contracts"][sourceName][contractName]["evm"][objectKind] =
+				output["contracts"][sourceName][contractName]["evm"][kind] =
 					collectEVMObject(
 						*o.bytecode,
 						o.sourceMappings.get(),
 						Json::arrayValue,
-						false,
-						[&](string const& _element) { return isArtifactRequested(
+						isDeployed,
+						[&, kind = kind](string const& _element) { return isArtifactRequested(
 							_inputsAndSettings.outputSelection,
 							sourceName,
 							contractName,
-							"evm." + objectKind + "." + _element,
+							"evm." + kind + "." + _element,
 							wildcardMatchesExperimental
 						); }
 					);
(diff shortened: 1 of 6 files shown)


ekpyron (commented):
Hm :-). This actually looks like this really may be all there is to it - but may be good if somebody else had a look at this to confirm that I'm not missing anything here :-)!

ekpyron:
This is a case in which ``--pretty-json`` in the ``args`` file of the test would definitely be nice.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1485,7 +1485,7 @@ Json::Value StandardCompiler::compileYul(InputsAndSettings _inputsAndSettings)
 						*o.bytecode,
 						o.sourceMappings.get(),
 						Json::arrayValue,
-						false,
+						objectKind == "deployedBytecode",
```
chriseth: We have a similar switch above. Can you create a local bool flag that does the string comparison only once, or iterate over `{true, false}`?

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1469,28 +1469,29 @@ Json::Value StandardCompiler::compileYul(InputsAndSettings _inputsAndSettings)
 	if (runtimeObject.bytecode)
 		runtimeObject.bytecode->link(_inputsAndSettings.libraries);
 
-	for (string const& objectKind: vector<string>{"bytecode", "deployedBytecode"})
+	struct ObjectKind { string const kind; bool isByteCode; };
+	for (auto const& objectKind: vector<ObjectKind>{{"bytecode", true}, {"deployedBytecode", false}})
```
```suggestion
	for (auto&& [kind, isByteCode]: {make_pair("bytecode", true), make_pair("deployedBytecode", false)})
```
That should work, shouldn't it :-D? If we want to be fanciful at least :-).

ekpyron: Fun quiz question: what's the type of the expression after the colon in that suggestion :-)...

ekpyron: By the way - it may be nicer to invert the ``bool`` and call it ``runtimeObject``.

Marenz: I don't see the argument for the inverting, it is checked for both, true and false at different places ;)
I do like the make_pair solution though. I suppose it is a pair again?

Marenz: ah, it seems to be a tuple?

ekpyron: I think the entire bracketed expression will be a ``std::initializer_list<std::pair<const char*, bool>>`` in the end - but I'd also need to recheck to be sure :-).

ekpyron: About inverting and renaming the flag: the main reason for the suggestion is that both the ``bytecode`` and the ``deployedBytecode``... well are byte code, so ``isByteCode`` is a weird name. And the most natural choice I found for renaming is the name of the argument of ``collectEVMObject``, which is ``runtimeObject`` (but is inverted)...

ekpyron: Or ``isRuntimeObject``, otherwise it conflicts with the ``runtimeObject`` we already have :-).

chriseth:
build error

chriseth:
The "interesting question" above is actually even more interesting: Since `kind` is not a variable, it cannot be captured by the lambda, which causes a build error on clang but not on gcc.

chriseth (approved):

PR #12593, opened 2022-01-28
Reduce isByteArary in scope

### LINKED ISSUE #12087: Properly allow `string.concat`
This may have been allowed unintentionally:

https://github.com/ethereum/solidity/blob/9d6eaa7a46425301ddb961a369b04f211d53691c/libsolidity/ast/Types.cpp#L3817-L3821

https://github.com/ethereum/solidity/blob/9d6eaa7a46425301ddb961a369b04f211d53691c/libsolidity/ast/Types.h#L841

https://github.com/ethereum/solidity/blob/9d6eaa7a46425301ddb961a369b04f211d53691c/libsolidity/ast/Types.h#L866

--- 

Anyway, `string.concat(...)` behaves exactly like `bytes.concat`, i.e., it expects `byte` argument and not string argument. It also returns `bytes memory` and not `string memory`.

Suggestions:

- [ ] Make it return `string memory`
- [ ] Make it accept `string` instead of `bytes`.

nishant-sachdeva:
Supposed to have two commits :
1. Change occurences of isByteArray() to isByteArrayOrString(). 
2. The idea is to, in a second commit, replace such occurences of isByteArrayOrString() which are required to return True only for Bytes type with a new isByteArray() function.

### DIFF
--- a/libsolidity/ast/Types.h
+++ b/libsolidity/ast/Types.h
@@ -837,6 +837,8 @@ public:
 
 	BoolResult validForLocation(DataLocation _loc) const override;
 
+	/// @returns true if this is a byte array.
+	bool isByteArray() const { return m_arrayKind == ArrayKind::Bytes; }
 	/// @returns true if this is a byte array or a string
 	bool isByteArrayOrString() const { return m_arrayKind != ArrayKind::Ordinary; }
 	/// @returns true if this is a string
--- a/libsolidity/ast/Types.h
+++ b/libsolidity/ast/Types.h
@@ -838,7 +838,7 @@ public:
 	BoolResult validForLocation(DataLocation _loc) const override;
 
 	/// @returns true if this is a byte array or a string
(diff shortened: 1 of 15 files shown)


libsolidity/ast/Types.h
```diff
@@ -838,7 +838,9 @@ class ArrayType: public ReferenceType
 	BoolResult validForLocation(DataLocation _loc) const override;
 
 	/// @returns true if this is a byte array or a string
```
ekpyron: The comment needs changing.

ekpyron (approved):
Looks good! I haven't checked each case for existing bugs, but this will at least make future mistakes less likely, so I'd say we merge this for now and move back to the actual ``string.concat`` issue!

chriseth:
We don't have a test run on the first commit (at least I cannot see it). Can you split this into two PRs just to be extra safe?

chriseth:
Types.h:872 has a comment `/// String is interpreted as a subtype of Bytes.` - this is now incorrect, can you change it please?

Issue #11412, opened 2021-05-19
Detect platform-dependent paths in default file loader (path spec v3)


cameel:
Related to #6608 and https://github.com/nomiclabs/hardhat/issues/538.
Part of a set of issues that replaces #11138.

## Abstract
Currently the default file loader interprets the source unit names it receives as platform-dependent paths. For example on case-insensitive filesystems it resolves file names in a case-insensitive way and on Windows it recognizes backslashes as valid path separators. This may lead to users unwittingly creating contracts that compile only on their platform.

The file loader should prevent this by recognizing and reporting situations that may result in such problems.

## Motivation
Here's an example of an import that will result in completely different files being expected on different platforms if the default file loader is involved:
```solidity
import "C:\\project\\lib\\token.sol"; // source unit name: C:\project\lib\token.sol
```

On Windows `C:` is seen as the root directory and backslashes as directory separators. On Linux it's all just one weird long file name.

## Specification
The default file loader should report a problem when:
- The canonical path of the file to be loaded matches the requested path only in a case-insensitive way.
- The underlying system allows backslashes as directory separators and the requested path contains one or more such slashes.
    - Or, better, when a source unit name contains backslashes at all.
- The requested path does not start with a forward slash but despite this is recognized as an absolute path.
- (Optional) The path contains characters known to be disallowed in paths on some platforms. E.g. `<` or `>` on Windows or `\0` characters on Linux.

If any of these are detected, at least warn the user about consequences. We should also consider disallowing them completely. That's what Hardhat did already: https://github.com/nomiclabs/hardhat/issues/538#issuecomment-843170204.

Note that this should be done at the file loader level. All of these are perfectly fine as long as they're only used in compiler's virtual filesystem. It's only when the actual filesystem is involved that they can cause problems.

## Backwards Compatibility
It's only a warning so it's fully backwards-compatible. Making it an error would make the change backwards-incompatible for any contract that uses such platform-dependent paths.

chriseth:
Maybe we should just disallow `\\` in import statements (or warn about them)?

chriseth:
About the case, I'm not so sure, it could create problems with recompilation on a case-insensitive file system.

cameel:
> About the case, I'm not so sure, it could create problems with recompilation on a case-insensitive file system.

[...]

PR #12598, opened 2022-01-30
Treat root path in normalizeCLIPathForVFS as case insensitive on Windows

a3d4:
`FileReader::normalizeCLIPathForVFS` strips root name if the path and working directory are on the same drive. Root comparison should perhaps be case insensitive on Windows.

### DIFF
--- a/libsolidity/interface/FileReader.cpp
+++ b/libsolidity/interface/FileReader.cpp
@@ -269,7 +269,8 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
 	if (!isUNCPath(normalizedPath))
 	{
 		boost::filesystem::path workingDirRootPath = canonicalWorkDir.root_path();
-		if (normalizedRootPath == workingDirRootPath)
+		// Ignore drive letter case on Windows (C:\ <=> c:\).
+		if (boost::filesystem::equivalent(normalizedRootPath, workingDirRootPath))
 			normalizedRootPath = "/";
 	}
 
--- a/test/libsolidity/interface/FileReader.cpp
+++ b/test/libsolidity/interface/FileReader.cpp
@@ -25,6 +25,7 @@
 #include <test/TemporaryDirectory.h>
 #include <test/libsolidity/util/SoltestErrors.h>
 
+#include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/test/unit_test.hpp>
 
@@ -192,8 +193,8 @@ BOOST_AUTO_TEST_CASE(normalizeCLIPathForVFS_root_name_only)
 
 #if defined(_WIN32)
 		boost::filesystem::path driveLetter = boost::filesystem::current_path().root_name();
-		solAssert(!driveLetter.empty(), "");
-		solAssert(driveLetter.is_relative(), "");
+		soltestAssert(!driveLetter.empty(), "");
+		soltestAssert(driveLetter.is_relative(), "");
 
 		BOOST_CHECK_EQUAL(FileReader::normalizeCLIPathForVFS(driveLetter, resolveSymlinks), expectedWorkDir);
 #endif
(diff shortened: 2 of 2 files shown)


test/FilesystemUtils.cpp
```diff
@@ -81,5 +81,6 @@ bool solidity::test::createSymlinkIfSupportedByFilesystem(
 		BOOST_THROW_EXCEPTION(runtime_error(
 			"Failed to create a symbolic link: \"" + _linkName.string() + "\""
 			" -> " + _targetPath.string() + "\"."
+			" " + symlinkCreationError.message() + "."
```
a3d4: When I run tests on my Windows system as a regular user, link creation fails. It works if I run tests as an administrator.

Showing "_A required privilege is not held by the client._" might help to figure out why tests are failing.

a3d4:
@cameel Hi Kamil, I wanted to make you a reviewer but do not seem to have sufficient rights (at least there is no gear icon next to **Reviewers**).

libsolidity/interface/FileReader.cpp
```diff
@@ -269,7 +269,8 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
 	if (!isUNCPath(normalizedPath))
 	{
 		boost::filesystem::path workingDirRootPath = canonicalWorkDir.root_path();
-		if (normalizedRootPath == workingDirRootPath)
+		// Ignore drive letter case on Windows (C:\ <=> c:\).
+		if (boost::filesystem::equivalent(normalizedRootPath, workingDirRootPath))
 			normalizedRootPath = "/";
```
cameel, line 274: The thing is, we completely ignore case-insensitivity on case-insensitive systems right now and drive letters are not an exception. You'll for example get into problems if you have different case in `--base-path` and the paths you use in file paths. And this is not specific to case - you can have problems using chars not allowed on some filesystems or stuff that depends on unicode equivalence.

Originally I wanted the normalization to include using the canonical capitalization (both NTFS on Windows and HFS on macOS are case-preserving and the case-insensitivity is a layer above that) but unfortunately Boost does not provide anything that would let me check what it is so I dropped the idea. I think that something like #11412 might be a better and more general way to approach the problem.

Case normalization specifically for root path is *probably* ok (the only two cases with named roots I'm aware of are drive letters and UNC paths) but I'd rather not add such specific hacks unless this is actually impacting some practical case. Can you run into it by running a `solc` command that has consistent capitalization? Or by just running tests? Or does it require doing something unusual?

test/FilesystemUtils.cpp
cameel: Interesting. Looks like symlinks on Windows are admin-only feature. And only in [Win 10 MS added some "developer mode"](https://www.atmosera.com/news/non-admin-users-can-now-create-symlinks-windows-10/) that lets you create them as a regular user.

This change looks good but I think we should also mention in the "contributing" docs that our test suite requires privileges for link creation.

a3d4: See 07ed09a255c2d96a853d6cd90cda0ac08d5e1500.

libsolidity/interface/FileReader.cpp
a3d4: Perhaps I should explain how I came here.

On my Windows system I have `soltest` on `D:` and the temporary folder on `C:`. That leads to failures like:

```
error: in "CommandLineInterfaceTest/cli_ignore_missing_some_files_exist": check result.reader.sourceUnits() == expectedSources has failed [{
C:/Users/ . . . . /input1.sol: ,
} != {
/Users/ . . . . /input1.sol: ,
}]
```

I can workaround it by changing `TMP` like `set TMP=D:\something`, but not like `set TMP=d:\something` (small `d`).

Hence the proposal. But if it creates more troubles than fixes (and the path normalization code already looks fragile), we can just extend the "contributing" docs correspondingly.

What do you think?

docs/contributing.rst
```diff
@@ -94,6 +94,13 @@ dependencies (`evmone <https://github.com/ethereum/evmone/releases>`_,
 On macOS some of the testing scripts expect GNU coreutils to be installed.
 This can be easiest accomplished using Homebrew: ``brew install coreutils``.
 
+On Windows systems make sure that you have a privilege to create symlinks,
+otherwise several tests may fail.
+Administrators should have that privilege, but you may also
+`grant it to other users <https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links#policy-management>`_.
+or
+`eneble Developer Mode <https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development>`_.
```
cameel:
```suggestion
`enable Developer Mode <https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development>`_.
```

docs/contributing.rst
```diff
@@ -94,6 +94,13 @@ dependencies (`evmone <https://github.com/ethereum/evmone/releases>`_,
 On macOS some of the testing scripts expect GNU coreutils to be installed.
 This can be easiest accomplished using Homebrew: ``brew install coreutils``.
 
+On Windows systems make sure that you have a privilege to create symlinks,
+otherwise several tests may fail.
+Administrators should have that privilege, but you may also
+`grant it to other users <https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links#policy-management>`_.
```
cameel:
```suggestion
`grant it to other users <https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links#policy-management>`_
```

test/FilesystemUtils.cpp
cameel: That looks good. Could you put the docs and this change in the same commit? They're closely related.

libsolidity/interface/FileReader.cpp
cameel: Thanks for the use case. That actually helps a lot. It's only a special case but looks like it could be common enough to be worth having a workaround for it.

> On my Windows system I have `soltest` on `D:` and the temporary folder on `C:`. That leads to failures

That seems to actually be a mistake in these tests. They assume that temp is always on the same drive and that's true in our CI but wrong in general. I need to fix them.

cameel (changes requested):
After seeing how it can happen, I'm fine with adding this special case after all.

Just needs a few tiny corrections and we can merge it.

a3d4:
@cameel 
I created #12632 for symlink Windows permissions and left other changes here.

cameel (dismissed):
LGTM. Just needs a rebase on latest `develop` which has fixes/workarounds for the failing CI jobs.

a3d4:
I also replaced `solAssert` with `soltestAssert`.

cameel (approved):

PR #12607, opened 2022-01-31
Fix changelog for the bugfix about immutables with return in base constructor

chriseth:
@Marenz @ekpyron please check if this change is correct.


chriseth:
Also please turn test/libsolidity/syntaxTests/immutable/base_ctor_return_no_immutables.sol  into a semantic test if there is not already one. We need to check that the immutable is properly initialized.

cameel (approved):

cameel:
Looking at the test case in #12523, it's correct:
https://github.com/ethereum/solidity/blob/2c4c8264e4428ff45cfd5f8f9b5458f1044ba8aa/test/libsolidity/syntaxTests/immutable/base_ctor_return_no_immutables.sol#L1-L10

ekpyron (approved):
Yeah, looks like that's what it should have said.

Marenz (approved):
That is indeed what it should have said

PR #11628, opened 2021-07-07
Add override exception for interface functions.

chriseth:
Fixes #8281

PR #12616, opened 2022-02-02
Fix missing override errors when indirectly inheriting interfaces.

### LINKED ISSUE #12615: Function override inconsistency bug
## Description

Inconsistent `override` identifier necessity behavior. We can have a typo in the method name and the compiler will not give us a clue we have 2 methods instead of 1.

## Environment

- Compiler version: 0.8.11
- Framework/IDE (e.g. Truffle or Remix): Remix, Hardhat

## Steps to Reproduce

Using `contract Ext is IExt, Base {}` in the following code unexpectedly produces no errors but changing inheritance order to the following `contract Ext is Base, IExt {}` leads to an expected error of missing `override` identifier in `Impl::foo()`.

```solidity

pragma solidity 0.8.11;

interface IBase {
    function foo() external view;
}

contract Base is IBase {
    function foo() public virtual view {}
}

interface IExt is IBase {}

contract Ext is IExt, Base {}

contract Impl is Ext {
    function foo() public view {}
}
```

ekpyron:
Fixes https://github.com/ethereum/solidity/issues/12615

### DIFF
--- a/libsolidity/analysis/OverrideChecker.cpp
+++ b/libsolidity/analysis/OverrideChecker.cpp
@@ -897,10 +897,11 @@ OverrideChecker::OverrideProxyBySignatureMultiSet const& OverrideChecker::inheri
 				if (var->isPublic())
 					functionsInBase.emplace(OverrideProxy{var});
 
-			for (OverrideProxy const& func: inheritedFunctions(*base))
-				functionsInBase.insert(func);
-
 			result += functionsInBase;
+
+			for (OverrideProxy const& func: inheritedFunctions(*base))
+				if (!functionsInBase.count(func))
+					result.insert(func);
 		}
 
 		m_inheritedFunctions[&_contract] = result;
--- a/test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
@@ -1,0 +1,16 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base is IBase {
+    function foo() public virtual view {}
+}
+
+interface IExt is IBase {}
+
+contract Ext is IExt, Base {}
+
+contract Impl is Ext {
+    function foo() public view override (IBase, Base) {}
+}
+// ----
--- a/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine.sol
@@ -1,0 +1,18 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base1 is IBase { function foo() public virtual view {} }
+contract Base2 is IBase { function foo() public virtual view {} }
+
+interface IExt1a is IBase {}
+interface IExt1b is IBase {}
+interface IExt2a is IBase {}
+interface IExt2b is IBase {}
+
+contract Ext1 is IExt1a, IExt1b, Base1 {}
+contract Ext2 is IExt2a, IExt2b, Base2 {}
+
+contract Impl is Ext1, Ext2 {
+    function foo() public view override (IBase, Base1, Base2) {}
+}
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,7 @@ Bugfixes:
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
  * General: ``string.concat`` now properly takes strings as arguments and returns ``string memory``. It was accidentally introduced as a copy of ``bytes.concat`` before.
  * Immutables: Fix wrong error when the constructor of a base contract uses ``return`` and the derived contract contains immutable variables.
+ * Inheritance: Consider functions in all ancestors during override analysis.
  * IR Generator: Add missing cleanup during the conversion of fixed bytes types to smaller fixed bytes types.
  * IR Generator: Add missing cleanup for indexed event arguments of value type.
  * IR Generator: Fix internal error when copying reference types in calldata and storage to struct or array members in memory.
--- a/test/libsolidity/syntaxTests/inheritance/override/correct_choice_for_base_function.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/correct_choice_for_base_function.sol
@@ -1,0 +1,17 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base is IBase {
+    function foo() public virtual view {}
+}
+
+interface IExt is IBase {}
+
+contract Ext is IExt, Base {}
+
+contract T { function foo() public virtual view {} }
+
+contract Impl is Ext, T {
+    function foo() public view override(IBase, Base, T) {}
+}
--- a/test/libsolidity/syntaxTests/inheritance/override/correct_choice_for_base_function_abstract_contract.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/correct_choice_for_base_function_abstract_contract.sol
@@ -1,0 +1,17 @@
+abstract contract IBase {
+    function foo() external view virtual;
+}
+
+contract Base is IBase {
+    function foo() public virtual override view {}
+}
+
+abstract contract IExt is IBase {}
+
+contract Ext is IExt, Base {}
+
+contract T { function foo() public virtual view {} }
+
+contract Impl is Ext, T {
+    function foo() public view override(IBase, Base, T) {}
+}
--- a/test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_err.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_err.sol
@@ -1,0 +1,18 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base is IBase {
+    function foo() public virtual view {}
+}
+
+interface IExt is IBase {}
+
+contract Ext is IExt, Base {}
+
+contract Impl is Ext {
+    function foo() public view {}
+}
+// ----
+// TypeError 9456: (211-240): Overriding function is missing "override" specifier.
+// TypeError 4327: (211-240): Function needs to specify overridden contracts "Base" and "IBase".
--- a/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_error.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_error.sol
@@ -1,0 +1,22 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base1 is IBase { function foo() public virtual view {} }
+contract Base2 is IBase { function foo() public virtual view {} }
+
+interface IExt1a is IBase {}
+interface IExt1b is IBase {}
+interface IExt2a is IBase {}
+interface IExt2b is IBase {}
+
+contract Ext1 is IExt1a, IExt1b, Base1 {}
+contract Ext2 is IExt2a, IExt2b, Base2 {}
+
+contract Impl is Ext1, Ext2 {
+    function foo() public view {}
+}
+// ----
+// TypeError 9456: (424-453): Overriding function is missing "override" specifier.
+// TypeError 9456: (424-453): Overriding function is missing "override" specifier.
+// TypeError 4327: (424-453): Function needs to specify overridden contracts "Base1", "Base2" and "IBase".
--- a/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine_abstract.sol
+++ b/test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine_abstract.sol
@@ -1,0 +1,18 @@
+interface IBase {
+    function foo() external view;
+}
+
+contract Base1 is IBase { function foo() public virtual view {} }
+contract Base2 is IBase { function foo() public virtual view {} }
+
+interface IExt1a is IBase {}
+abstract contract IExt1b is IBase {}
+abstract contract IExt2a is IBase {}
+interface IExt2b is IBase {}
+
+contract Ext1 is IExt1a, IExt1b, Base1 {}
+contract Ext2 is IExt2a, IExt2b, Base2 {}
+
+contract Impl is Ext1, Ext2 {
+    function foo() public view override (IBase, Base1, Base2) {}
+}


cameel:
So the linearized order in the snippet from #12615 looks something like this:
```
Impl -> Ext -> Base -> IExt -> IBase
```
I'd expect the bug to be triggered by `IExt` jumping in front of `Base` in the linearized order but surprisingly in this snippet it's the opposite - the override error disappears only when `Base` is before `IExt` in C3 order.

Can you add an extra test case where there are multiple layers of interfaces and concrete contracts mixed in the linearizer order? For example something like this:
```solidity
interface IBase {
    function foo() external view;
}

contract Base1 is IBase { function foo() public virtual view {} }
contract Base2 is IBase { function foo() public virtual view {} }

interface IExt1a is IBase {}
interface IExt1b is IBase {}
interface IExt2a is IBase {}
interface IExt2b is IBase {}

contract Ext1 is IExt1a, IExt1b, Base1 {}
contract Ext2 is IExt2a, IExt2b, Base2 {}

contract Impl is Ext1, Ext2 {
    function foo() public view {}
}
```
The order here should look like this:
```
Impl -> Ext2 -> Base2 -> IExt2b -> IExt2a -> Ext1 -> Base1 -> IExt1b -> IExt1a -> IBase
```

cameel:
I'd also add a test with abstract contract - that case is broken too. I think your fix should cover it but it would be better to have a test checking that.

test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
```diff
@@ -0,0 +1,16 @@
...
+contract Base is IBase {
+    function foo() public virtual view {}
+}
+
+interface IExt is IBase {}
+
+contract Ext is IExt, Base {}
+
+contract Impl is Ext {
+    function foo() public view override (IBase, Base) {}
```
ekpyron, line 14: I rechecked now - and this is actually not a bug, it's actually how we specified this...
But nowadays, I'd actually change the spec and have this only require a single ``override`` here (resp. only require it to override ``Base``).

``Ext`` is fine even *without* overriding, since we have a unique choice of an implementation of ``foo``. That doesn't change in ``Impl``, so it's rather weird that we require *both* overrides there...

test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine.sol
```diff
@@ -0,0 +1,18 @@
...
+interface IExt1a is IBase {}
+interface IExt1b is IBase {}
+interface IExt2a is IBase {}
+interface IExt2b is IBase {}
+
+contract Ext1 is IExt1a, IExt1b, Base1 {}
+contract Ext2 is IExt2a, IExt2b, Base2 {}
+
+contract Impl is Ext1, Ext2 {
+    function foo() public view override (IBase, Base1, Base2) {}
```
ekpyron, line 17: Similarly here I actually don't think it's reasonable for us to require ``IBase`` in the list.

test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
cameel: I agree. There's only one `foo()` not coming from an interface here so a simple `override` should be enough. I think this should be reported as a separate issue/bug.

test/libsolidity/syntaxTests/inheritance/override/override_multi_layered_fine.sol
cameel: Yeah, I agree.

cameel (approved):

cameel:
So do we consider this a full fix for #12615? I see you did not include that in the description so the issue won't be automatically closed.

ekpyron:
> So do we consider this a full fix for #12615? I see you did not include that in the description so the issue won't be automatically closed.

Yeah, I'd say this fixes #12615.

Are we 100% sure that the change won't have *any* unwanted side-effects :-)? I'd hope so, but I'd feel better if somebody confirmed explicitly once more that it's probably fine :-). Since #12615 slipped through in the first place, I don't entirely trust our test coverage on this, though... and the override checker logic is a tad intransparent :-). On the other hand zeppelin compiles with it and it probably contains an example of most inheritance permutations :-).

cameel:
Well, I thought the fix looked reasonable but now you gave me some doubts :)

The change is that now `inheritedFunctions()` will list a function at most once for each direct base, and before it would list it as many times as it occurred in the whole inheritance tree of that direct base? It sounds like what it's supposed to be doing but honestly, I don't see a direct relation to how it makes the override error appear. If anything, I'd expect it to disappear in some cases instead. But the fact that it passes the new tests and external ones suggests that it can't be that bad. We might miss some corner cases but it at least shows that it does not break anything obvious.

I agree that we need more coverage for overrides. The question is - how much and how to be sure we have enough? It's not like we'll be able to generate all possible combinations. @bshastry is there any chance the fuzzer could help us here?

ekpyron:
I don't think we can really fuzz this :-). There's no reference for correctness whatsoever.

The insidious part of the change is that ``functionsInBase`` is a set and ``result`` is a multiset...

So before the change, the inner loop over ``inheritedFunctions`` potentially ranged over multiple copies of functions, but only one would end up in ``functionsInBase``. 

So in the example:
```
interface IBase {
    function foo() external view;
}

contract Base is IBase {
    function foo() public virtual view {}
}

interface IExt is IBase {}

contract Ext is IExt, Base {}

contract Impl is Ext {
    function foo() public view {}
}
```

in ``Impl`` when looking at functions inherited from ``Ext``, it'd only consider one base function ``foo`` - and it happened to be the one in the interface. And since we don't require ``override`` for interface functions it didn't complain.

With the change I propagate *all* function with the signature from ``Ext`` to ``Impl``. Which means I need to override both in ``Impl`` - which is a bit weird, but I think that's how our spec works right now.

In a similar case:

```
interface IBase {
    function foo() external view;
}

contract Base is IBase {
    function foo() public virtual view {}
}

interface IExt is IBase {}

contract Ext is IExt, Base { function foo() public /* maybe some override */ {} }

contract Impl is Ext {
    function foo() public view {}
}
```

When again in ``Impl`` looking at functions inherited from ``Ext``, I *only* want to consider ``Ext.foo`` itself. Previously that was guaranteed by the ``functionsInBase`` set already containing ``foo``, so inserting the functions from ``IExt`` and ``Base`` did nothing. After the change, I do it explicitly by checking ``!functionsInBase.count(func)`` before adding to ``result``.

So I think this makes sense at least for these cases, but I get a headache if I try to convince myself that there is no other cases in which this does something weird :-D.

I would in general say that we should try and simplify the entire mechanism - especially if we decide to go for https://github.com/ethereum/solidity/pull/12616#discussion_r797985635. If possible at least :-). I actually hope that the proposed refinement of the spec with https://github.com/ethereum/solidity/pull/12616#discussion_r797985635 (that is each direct base can contribute at most one override requirement) will make all of this easier...

ekpyron:
The annoying part is that it'll be a challenge to do the change https://github.com/ethereum/solidity/pull/12616#discussion_r797985635 in a non-breaking way :-). If we want to do that and the change to be non-breaking, then we need to allow to specify additional bases beyond the required ones in the override list - but we shouldn't allow arbitrary ones, but only ones that are bases of the ones required after the change I guess :-).

bshastry:
> @bshastry is there any chance the fuzzer could help us here?

The fuzzer could synthesize arbitrary inheritance patterns but I don't see how it would have flagged the underlying issue as a bug. Afaiu, the user noticed the bug because one of the two inheritance patterns produced an error while the other didn't? I mean we could write a fuzzer that checks that `contract A is B, C` and `contract A is C, B` both produce errors or both produce no errors but I think that is not what we really want here.

test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
cameel: **EDIT**: *This is actually a different situation than the one we have in the PR. We might still want to relax the requirement but that's a separate matter. See my comment later in the thread.*

I just remembered that this is probably coming from how I specified this originally: [Should overriding function declarations require the `override` keyword?](https://forum.soliditylang.org/t/should-overriding-function-declarations-require-the-override-keyword/347). And we were in favor of not relaxing this in the `override(A, B)` case. I.e. this is exactly what we wanted:

```solidity
interface IA {
    function foo() external;
}
interface IB {
    function foo() external;
}
contract C is IA, IB {
    function foo() external override(IA, IB) {}
}
```

Whether that was a good decision... given how we're ourselves confused by it now I'd say probably not :) Just dropping the requirement entirely would be more consistent. It does have one advantage though - you only have to treat interfaces specially in the single `override` case. In all other cases all inherited contracts count.

cameel: **EDIT**: *This is actually a different situation than the one we have in the PR. We might still want to relax the requirement but that's a separate matter. See my comment later in the thread.*

Actually, this is not the whole story. If we did include the `override(A, B)` case for interfaces it would hide this problem but the problem itself is really what we have been discussing recently on the forum, just for interfaces. I.e. you need to include `IBase` because you're inheriting from it directly and not only through `Base`. The situation could be simplified to this:
```solidity
interface IBase {
    function foo() external view;
}
contract Base is IBase {
    function foo() public virtual view override {}
}
contract Ext is IBase, Base {
    function foo() public view virtual override(IBase, Base) {}
}
```
This requires `override(IBase, Base)` now but it did on 0.8.0 as well.

cameel:
> The annoying part is that it'll be a challenge to do the change [#12616 (comment)](https://github.com/ethereum/solidity/pull/12616#discussion_r797985635) in a non-breaking way :-). If we want to do that and the change to be non-breaking, then we need to allow to specify additional bases beyond the required ones in the override list - but we shouldn't allow arbitrary ones, but only ones that are bases of the ones required after the change I guess :-).

I don't think it has to be that complicated. We just compute the contracts you must list the same way we do now. Then, when we compare with what you actually listed, we ignore missing contracts if they're interfaces.

test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
cameel: After digging deeper into what's happening here, I think this is actually a bug. The message says that you need `override(Base, IBase)` here and IMO this would be correct in `Ext` but not in `Impl`. You should never have to specify more contracts than you directly inherit from. Each one you inherit from has to already have settled on a single implementation of `foo()` and you're overriding that specific implementation. I don't think it can just forward that decision down the inheritance hierarchy. At least not without making the intermediate contract `abstract`.

This is also how it worked before we relaxed the requirements for interfaces in #11628. I.e. it only required `override` (without specifying bases) on 0.8.0.

libsolidity/analysis/OverrideChecker.cpp
```diff
@@ -897,10 +897,11 @@ OverrideChecker::OverrideProxyBySignatureMultiSet const& OverrideChecker::inheri
...
 					functionsInBase.emplace(OverrideProxy{var});
 
-			for (OverrideProxy const& func: inheritedFunctions(*base))
-				functionsInBase.insert(func);
-
 			result += functionsInBase;
+
+			for (OverrideProxy const& func: inheritedFunctions(*base))
+				if (!functionsInBase.count(func))
+					result.insert(func);
```
cameel, line 904: > The insidious part of the change is that ``functionsInBase`` is a set and ``result`` is a multiset...
>
> So before the change, the inner loop over ``inheritedFunctions`` potentially ranged over multiple copies of functions, but only one would end up in ``functionsInBase``. 

> in ``Impl`` when looking at functions inherited from ``Ext``, it'd only consider one base function ``foo`` - and it happened to be the one in the interface. And since we don't require ``override`` for interface functions it didn't complain.
>
> With the change I propagate *all* function with the signature from ``Ext`` to ``Impl``. Which means I need to override both in ``Impl`` - which is a bit weird, but I think that's how our spec works right now.

Thanks for this explanation. So I actually interpreted this wrong the first time I looked at it. Now it makes sense. I did a deeper dive through the code now and I think I finally understand well what is happening and why.

I think that this lets through only a single function on purpose. If you inherit from `A`, `B` and `C`, each of them must have already settled on a single definition for `foo()`. So the code above should never have to insert more than one function into `result`.

The bug was introduced in #11628, specifically by https://github.com/ethereum/solidity/commit/f7916f2940d3266a28c001a08a474f0c6eb3df01#diff-980e91f229ba8ff7a7ed8705977e094851571ba47f544a50a9722371e8086dc4L521-R522. It wrongly assumes that it always gets that single, reconciled `foo()` from a parent contract that you would end up calling if you referred to `foo()` inside that parent contract.

So to fix this cleanly, `inheritedFunctions()` has to satisfy this assumption. Basically, it needs to perform the same lookup logic you get when you call `foo()`. It should use `CallableDeclaration::resolveVirtual()`. The question is - can we do it here? That function might depend on something that's not yet available at this analysis stage.

If we can't do that, a viable workaround would be just to return a non-interface function if one is available.

cameel: As for testing this, I think that we just need to cover variants of a "bottleneck" in the inheritance hierarchy:
```
A  B  C
 \ | /
   X
   |
   Y
```
The problem should really exist only if `X` does not have its own version of the function and there are multiple versions in its ancestors.

cameel (changes requested):
I analyzed this again and I think this actually needs changes. See my comments.

ekpyron:
@cameel I agree with what you want to do - that's basically https://github.com/ethereum/solidity/pull/12616#discussion_r797985635.
But the current behaviour is *not* a bug - the current behaviour is how we specified inheritance and overrides.
And changing it, unfortunately, is breaking. In fact the first commit of this PR does something along those lines (even though not properly), but breaks zeppelin.

So while I agree that that's what *should* be happening, I think we have to stick to the fix in this PR for now, since that's the specified behaviour that's actually in use...

The only way out of the "nicer" solution being breaking would be to allow additional bases in the override specifier to keep the change backwards compatible... we should consider that, but we should fix the underlying issue asap anyways, so I still think we should merge this as is.

libsolidity/analysis/OverrideChecker.cpp
ekpyron: The bug was already in the initial implementation of overrides and is older than https://github.com/ethereum/solidity/pull/11628 - https://github.com/ethereum/solidity/pull/11628 only made it more visible, since you end up *not* having to override stuff due to this - but in situations like https://github.com/ethereum/solidity/pull/12616#discussion_r799823373, if ``X`` does not implement ``f``, but ``A``, ``B`` and ``C`` do, we've always only considered ``A.f`` for the override logic, which has always been wrong.

ekpyron: Or wait a second, that's also not entirely true - let me recheck...

ekpyron:
As for the specification - we have the following in https://docs.soliditylang.org/en/develop/contracts.html#function-overriding:

> For multiple inheritance, the most derived base contracts that define the same function must be specified explicitly after the override keyword. In other words, you have to specify all base contracts that define the same function and have not yet been overridden by another base contract (on some path through the inheritance graph).

Actually, reading this again now, it does say ``define`` and not ``declare``... so in that sense, we could get away with calling this a (then even more serious) bug and implement https://github.com/ethereum/solidity/pull/12616#discussion_r797985635 right away... the problem is that it *does* break zeppelin, though :-).

libsolidity/analysis/OverrideChecker.cpp
ekpyron:
```
abstract contract I { function f() external virtual; }
abstract contract A is I {
}
abstract contract B is I {
        function f() public pure virtual override {}
}
contract X is A, B {
}

contract Y is X {
        function f() public pure override(B) {}
}
```
This for example fails on 0.8.0 - it wants you to use ``override(I)`` instead of ``B``. Which is wrong :-). And once you extend it to more than one base this error becomes visible... that can only happen, if you have a common base in which the function is unimplemented and then multiple paths without mention of the function and one path that defines it... that's the situation in which we ended up *not* requiring an ``override`` in ``X``... unfortunately, we did not clearly define what should happen in ``Y`` in this situation - respectively, we implemented it incorrectly...

ekpyron: This is the relevant test case in zeppelin (when solving this differently), let me have another look at that:
https://app.circleci.com/pipelines/github/ethereum/solidity/21931/workflows/31931f3e-fc90-4ebe-8670-f57eeb69d6a7/jobs/961852

ekpyron: Reducing the case of ``state()`` in there to only the relevant parts, then it's this situation:
```
interface IG {
	function f() external pure virtual;
}

abstract contract G is IG {
	function f() public pure virtual override {}
}
abstract contract GS is IG {
}

abstract contract GTL is IG {
	function f() public pure virtual override {}
}

abstract contract GVC is G {
}

abstract contract GCB is G {
	
}

contract MD is GCB, GS, GTL, GVC {
	function f() public pure override (IG, G, GTL) {}
}
```

We *do* currently require ``IG`` in the override list. The same happens if you change ``IG`` to an abstract contract.
The situation is a bit different than I thought. I think the third override requirement is due to ``GVC`` and ``GCB`` both bringing in an *implemented* version of ``f``... however I'm not quite sure why it is ``IG`` of all things the compiler asks to additionally override here...

test/libsolidity/syntaxTests/inheritance/override/interface_and_base_override_fine.sol
ekpyron: Hm... In 0.8.0, it required to override ``IBase``, but *not* ``Base``... that was already wrong, but just less visibly wrong..

libsolidity/analysis/OverrideChecker.cpp
ekpyron: Ah, ok and ``GS`` is the one bringing in the ``override (IG)`` requirement...

ekpyron: Ok, maybe zeppelin breaking was actually only a bug in my preliminary fix in the first commit, not an issue with implementing https://github.com/ethereum/solidity/pull/12616#discussion_r797985635...

ekpyron: > If we can't do that, a viable workaround would be just to return a non-interface function if one is available.

That's in any case what the first commit in this PR does, isn't it? And that commit breaks zeppelin... I'm still trying to wrap my head around this entirely, though :-).

ekpyron:
Notes from call with @chriseth: we tend towards implementing the comment https://github.com/ethereum/solidity/pull/12616#discussion_r797985635, i.e. not require all bases in override lists, but only those that actually implement the function - but we want to keep things backwards-compatible, so we want to *allow* specifying additional bases (in case they are bases of the required contracts).

We also need to adjust the specification in the docs accordingly.

libsolidity/analysis/OverrideChecker.cpp
cameel: ok, after rethinking it, I agree with your analysis.

> if `X` does not implement `f`, but `A`, `B` and `C` do, we've always only considered `A.f` for the override logic, which has always been wrong.

True. I somehow disregarded the case of `X` being abstract but that's the case that happens in OZ. In such a case `f` is not yet resolved to a single definition in `X` so we have more than one definition coming through the same contract. Forwarding only one is wrong in such a situation.

> That's in any case what the first commit in this PR does, isn't it?

Right. And you're right that it's wrong. I now think that the current solution in this PR is correct. It includes either the copy of `f` from current contract or all copies of `f` from ancestors that have at least one path to the current contract on which they're not overridden.

cameel (commented):
> not require all bases in override lists, but only those that actually implement the function - but we want to keep things backwards-compatible, so we want to *allow* specifying additional bases

Yeah. this seems to be the best way forward.

chriseth (approved):

chriseth:
The failing test dos not look like it is caused by this change.

k06a:
Great, should we expect 0.8.12 soon?

ekpyron:
> Great, should we expect 0.8.12 soon?

Probably some time this week, yes.
However, we left https://github.com/ethereum/solidity/issues/12665 open, so the result will not be that nice until we fix that one as well (i.e. in your example in https://github.com/ethereum/solidity/issues/12615 you will now not only need an override, but ``override (IBase, Base)`` which is a bit silly, but which is how we spec'ed things so far unfortunately - so we plan to relax this, but this will be some more internal effort, so we won't manage for 0.8.12...)...

k06a:
Glad to hear you’re gonna fix this annoying issue too!

PR #12620, opened 2022-02-03
Memory-safety annotation for inline assembly.

ekpyron:
We can also properly make inline assembly ``StructurallyDocumented``, etc, but then we need to double-check that existing natspec-style comments don't cause any errors to keep this fully non-breaking.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ### 0.8.13 (unreleased)
 
 Language Features:
+ * General: Allow annotating inline assembly as memory-safe to allow optimizations and stack limit evasion that rely on respecting Solidity's memory model.
 
 
 Compiler Features:
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -228,6 +228,11 @@ of their block is reached.
 Conventions in Solidity
 -----------------------
 
+.. _assembly-typed-variables:
+
+Values of Typed Variables
+=========================
+
 In contrast to EVM assembly, Solidity has types which are narrower than 256 bits,
 e.g. ``uint24``. For efficiency, most arithmetic operations ignore the fact that
 types can be shorter than 256
@@ -236,6 +241,11 @@ i.e., shortly before they are written to memory or before comparisons are perfor
 This means that if you access such a variable
 from within inline assembly, you might have to manually clean the higher-order bits
 first.
+
+.. _assembly-memory-management:
+
+Memory Management
+=================
 
 Solidity manages memory in the following way. There is a "free memory pointer"
 at position ``0x40`` in memory. If you want to allocate memory, use the memory
@@ -268,3 +278,89 @@ first slot of the array and followed by the array elements.
     Statically-sized memory arrays do not have a length field, but it might be added later
     to allow better convertibility between statically- and dynamically-sized arrays, so
     do not rely on this.
+
+Memory Safety
+=============
+
+Without the use of inline assembly, the compiler can rely on memory to remain in a well-defined
+state at all times. This is especially relevant for :ref:`the new code generation pipeline via Yul IR <ir-breaking-changes>`:
+this code generation path can move local variables from stack to memory to avoid stack-too-deep errors and
+perform additional memory optimizations, if it can rely on certain assumptions about memory use.
+
+While we recommend to always respect Solidity's memory model, inline assembly allows you to use memory
+in an incompatible way. Therefore, moving stack variables to memory and additional memory optimizations are,
+by default, disabled in the presence of any inline assembly block that contains a memory operation or assigns
+to solidity variables in memory.
+
+However, you can specifically annotate an assembly block to indicate that it in fact respects Solidity's memory
+model as follows:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+        ...
+    }
+
+In particular, a memory-safe assembly block may only access the following memory ranges:
+
+- Memory allocated by yourself using a mechanism like the ``allocate`` function described above.
+- Memory allocated by Solidity, e.g. memory within the bounds of a memory array you reference.
+- The scratch space between memory offset 0 and 64 mentioned above.
+- Temporary memory that is located *after* the value of the free memory pointer at the beginning of the assembly block,
+  i.e. memory that is "allocated" at the free memory pointer without updating the free memory pointer.
+
+Furthermore, if the assembly block assigns to Solidity variables in memory, you need to assure that accesses to
+the Solidity variables only access these memory ranges.
+
+Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block
+reverts or terminates. As an example, the following assembly snippet is not memory safe:
+
+.. code-block:: solidity
+
+    assembly {
+      returndatacopy(0, 0, returndatasize())
+      revert(0, returndatasize())
+    }
+
+But the following is:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+      let p := mload(0x40)
+      returndatacopy(p, 0, returndatasize())
+      revert(p, returndatasize())
+    }
+
+Note that you do not need to update the free memory pointer if there is no following allocation,
+but you can only use memory starting from the current offset given by the free memory pointer.
+
+If the memory operations use a length of zero, it is also fine to just use any offset (not only if it falls into the scratch space):
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+      revert(0, 0)
+    }
+
+Note that not only memory operations in inline assembly itself can be memory-unsafe, but also assignments to
+solidity variables of reference type in memory. For example the following is not memory-safe:
+
+.. code-block:: solidity
+
+    bytes memory x;
+    assembly {
+      x := 0x40
+    }
+    x[0x20] = 0x42;
+
+Inline assembly that neither involves any operations that access memory nor assigns to any solidity variables
+in memory is automatically considered memory-safe and does not need to be annotated.
+
+.. warning::
+    It is your responsibility to make sure that the assembly actually satisfies the memory model. If you annotate
+    an assembly block as memory-safe, but violate one of the memory assumptions, this **will** lead to incorrect and
+    undefined behaviour that cannot easily be discovered by testing.
--- a/libsolidity/analysis/DocStringTagParser.cpp
+++ b/libsolidity/analysis/DocStringTagParser.cpp
@@ -30,6 +30,7 @@
 #include <liblangutil/Common.h>
 
 #include <range/v3/algorithm/any_of.hpp>
+#include <range/v3/view/filter.hpp>
 
 #include <boost/algorithm/string.hpp>
 
@@ -158,6 +159,71 @@ bool DocStringTagParser::visit(EventDefinition const& _event)
 bool DocStringTagParser::visit(ErrorDefinition const& _error)
 {
 	handleCallable(_error, _error, _error.annotation());
+
+	return true;
+}
+
+bool DocStringTagParser::visit(InlineAssembly const& _assembly)
+{
+	if (!_assembly.documentation())
+		return true;
+	StructuredDocumentation documentation{-1, _assembly.location(), _assembly.documentation()};
+	ErrorList errors;
+	ErrorReporter errorReporter{errors};
+	auto docTags = DocStringParser{documentation, errorReporter}.parse();
+
+	if (!errors.empty())
+	{
+		SecondarySourceLocation ssl;
+		for (auto const& error: errors)
+			if (error->comment())
+				ssl.append(
+					*error->comment(),
+					_assembly.location()
+				);
+		m_errorReporter.warning(
+			7828_error,
+			_assembly.location(),
+			"Inline assembly has invalid NatSpec documentation.",
+			ssl
+		);
+	}
+
+	for (auto const& [tagName, tagValue]: docTags)
+	{
+		if (tagName == "solidity")
+		{
+			vector<string> values;
+			boost::split(values, tagValue.content, isWhiteSpace);
+
+			set<string> valuesSeen;
+			set<string> duplicates;
+			for (auto const& value: values | ranges::views::filter(not_fn(&string::empty)))
+				if (valuesSeen.insert(value).second)
+				{
+					if (value == "memory-safe-assembly")
+						_assembly.annotation().memorySafe = true;
+					else
+						m_errorReporter.warning(
+							8787_error,
+							_assembly.location(),
+							"Unexpected value for @solidity tag in inline assembly: " + value
+						);
+				}
+				else if (duplicates.insert(value).second)
+					m_errorReporter.warning(
+						4377_error,
+						_assembly.location(),
+						"Value for @solidity tag in inline assembly specified multiple times: " + value
+					);
+		}
+		else
+			m_errorReporter.warning(
+				6269_error,
+				_assembly.location(),
+				"Unexpected NatSpec tag \"" + tagName + "\" with value \"" + tagValue.content + "\" in inline assembly."
+			);
+	}
 
 	return true;
 }
--- a/libsolidity/analysis/DocStringTagParser.cpp
+++ b/libsolidity/analysis/DocStringTagParser.cpp
@@ -202,7 +202,7 @@ bool DocStringTagParser::visit(InlineAssembly const& _assembly)
 				if (valuesSeen.insert(value).second)
 				{
 					if (value == "memory-safe-assembly")
-						_assembly.annotation().memorySafe = true;
+						_assembly.annotation().markedMemorySafe = true;
 					else
 						m_errorReporter.warning(
 							8787_error,
--- a/libsolidity/codegen/ir/IRGenerationContext.h
+++ b/libsolidity/codegen/ir/IRGenerationContext.h
@@ -160,8 +160,8 @@ public:
 
 	std::set<ContractDefinition const*, ASTNode::CompareByID>& subObjectsCreated() { return m_subObjects; }
 
-	bool inlineAssemblySeen() const { return m_inlineAssemblySeen; }
-	void setInlineAssemblySeen() { m_inlineAssemblySeen = true; }
+	bool memoryUnsafeInlineAssemblySeen() const { return m_memoryUnsafeInlineAssemblySeen; }
+	void setMemoryUnsafeInlineAssemblySeen() { m_memoryUnsafeInlineAssemblySeen = true; }
 
 	/// @returns the runtime ID to be used for the function in the dispatch routine
 	/// and for internal function pointers.
@@ -202,8 +202,8 @@ private:
 	/// Whether to use checked or wrapping arithmetic.
 	Arithmetic m_arithmetic = Arithmetic::Checked;
 
-	/// Flag indicating whether any inline assembly block was seen.
-	bool m_inlineAssemblySeen = false;
+	/// Flag indicating whether any memory-unsafe inline assembly block was seen.
+	bool m_memoryUnsafeInlineAssemblySeen = false;
 
 	/// Function definitions queued for code generation. They're the Solidity functions whose calls
 	/// were discovered by the IR generator during AST traversal.
--- a/libyul/backends/evm/EVMObjectCompiler.cpp
+++ b/libyul/backends/evm/EVMObjectCompiler.cpp
@@ -25,6 +25,8 @@
 #include <libyul/backends/evm/EVMDialect.h>
 #include <libyul/backends/evm/OptimizedEVMCodeTransform.h>
 
+#include <libyul/optimiser/FunctionCallFinder.h>
+
 #include <libyul/Object.h>
 #include <libyul/Exceptions.h>
 
@@ -74,7 +76,22 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
 			OptimizedEVMCodeTransform::UseNamedLabels::ForFirstFunctionOfEachName
 		);
 		if (!stackErrors.empty())
-			BOOST_THROW_EXCEPTION(stackErrors.front());
+		{
+			vector<FunctionCall*> memoryGuardCalls = FunctionCallFinder::run(
+				*_object.code,
+				"memoryguard"_yulstring
+			);
+			auto stackError = stackErrors.front();
+			string msg = stackError.comment() ? *stackError.comment() : "";
+			if (memoryGuardCalls.empty())
+				msg += "\nNo memoryguard was present. "
+					"Consider using memory-safe assembly only and annotating it via "
+					"\"/// @solidity memory-safe-assembly\".";
+			else
+				msg += "\nmemoryguard was present.";
+			stackError << util::errinfo_comment(msg);
+			BOOST_THROW_EXCEPTION(stackError);
+		}
 	}
 	else
 	{
--- a/test/libsolidity/memoryGuardTests/free_function.sol
+++ b/test/libsolidity/memoryGuardTests/free_function.sol
@@ -1,0 +1,29 @@
+function safe() pure returns (uint256 x) {
+    assembly { x := 42 }
+    /// @solidity memory-safe-assembly
+    assembly { mstore(0, 0) }
+}
+function unsafe() pure returns (uint256 x) {
+    assembly { pop(mload(0)) }
+}
+contract C {
+    constructor() {
+        unsafe();
+    }
+    function f() public pure {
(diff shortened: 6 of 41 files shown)


libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -590,7 +590,8 @@ bool ASTJsonConverter::visit(InlineAssembly const& _node)
 	setJsonNode(_node, "InlineAssembly", {
 		make_pair("AST", Json::Value(yul::AsmJsonConverter(sourceIndexFromLocation(_node.location()))(_node.operations()))),
 		make_pair("externalReferences", std::move(externalReferencesJson)),
-		make_pair("evmVersion", dynamic_cast<solidity::yul::EVMDialect const&>(_node.dialect()).evmVersion().name())
+		make_pair("evmVersion", dynamic_cast<solidity::yul::EVMDialect const&>(_node.dialect()).evmVersion().name()),
+		make_pair("memorySafe", _node.annotation().memorySafe)
```
ekpyron: Actually, I need to check, if I can be sure that this always has been correctly assigned before exporting.

chriseth:
Can you also define what exactly memory safety means?

ekpyron:
> Can you also define what exactly memory safety means?

Sure, this needs extensive documentation before I undraft it :-).
Mainly pushed it already to see if it works for the external test runs. For trident it does - but ENS still fails, I'm just trying to figure out whether that's because it pulls in dependencies with assembly that I didn't patch or because of something else.

ekpyron:
Eventually we should also add better reports for all of this...
Ideally, I'd report an info message whenever any variables are moved to memory and if compilation fails due to stack too deep, but could not run stack-to-memory due to inline assembly, it should properly suggest adding those annotations... Ideally even with source locations :-).

But that's a bit of a hassle, since codegen so far is not meant to report stuff like that and we may need to store the source locations somewhere, etc., so I'd keep that separate for now...

test/libsolidity/ASTJSON/assembly/inline_assembly_memory_safe_parseOnly.json
```diff
@@ -0,0 +1,144 @@
...
+                      "nodeType": "YulFunctionDefinition",
+                      "src": "119:64:1"
+                    }
+                  ]
+                },
+                "documentation": "@solidity memory-safe-assembly",
+                "evmVersion": %EVMVERSION%,
+                "externalReferences": [],
+                "id": 5,
+                "memorySafe": false,
```
ekpyron: This is a problem...

ekpyron: On the other hand: since the documentation node is in here, maybe I don't need to add this to the JSON AST specifically at all and it will hopefully just rerun the doc string parser on import, I'll check.

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -74,7 +76,22 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
...
+		{
+			vector<FunctionCall*> memoryGuardCalls = FunctionCallFinder::run(
+				*_object.code,
+				"memoryguard"_yulstring
+			);
+			auto stackError = stackErrors.front();
+			string msg = stackError.comment() ? *stackError.comment() : "";
+			if (memoryGuardCalls.empty())
+				msg += "\nNo memoryguard was present. "
+					   "Consider using memory-safe assembly only and annotating it via "
```
stackenbotten, line 88: Coding style error

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -74,7 +76,22 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
...
+					   "\"/// @solidity memory-safe-assembly\".";
```
stackenbotten, line 89: Coding style error

libyul/backends/evm/EVMObjectCompiler.cpp
stackenbotten, line 88: Coding style error

libyul/backends/evm/EVMObjectCompiler.cpp
stackenbotten, line 89: Coding style error

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +162,56 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
 	return true;
 }
 
+bool DocStringTagParser::visit(InlineAssembly const& _assembly)
+{
+	if (!_assembly.documentation())
+		return true;
+	StructuredDocumentation documentation{-1, _assembly.location(), _assembly.documentation()};
```
ekpyron, line 170: The ``-1`` node id is a bit creepy...

ekpyron:
I just added a first rough draft for documenting this.

I'm wondering how restrictive we should define this - technically it's valid for stack-to-memory to access any memory that is known to be past the initial value of the free memory pointer...
Also we need to decide whether we want to specifically add the requirement for memory accesses to be bounded, i.e. the assumption for the redundant store eliminator (either as only ever allocating bounded amounts of memory, if we stay with only accessing allocated memory, or more freely as not accessing memory from unbounded offsets).

ekpyron:
Test failures are just pending https://github.com/ethereum/solidity/pull/12624 I think

cameel:
#12624 is ready.

docs/assembly.rst
```diff
@@ -268,3 +278,36 @@ first slot of the array and followed by the array elements.
...
+    /// @solidity memory-safe-assembly
+    assembly {
+        ...
+    }
+
+In particular, a memory-safe assembly block has to follow the adhere to the following restriction:
+All accessed memory (read or written) was properly allocated. This can either be memory allocated by yourself
+using a mechanism like the ``allocate`` function described above, or memory allocated by Solidity,
+e.g. memory within the bounds of a memory array you reference. The only exception to this is the scratch space
+between memory offset 0 and 64 mentioned above.
```
ekpyron: For this we still need to discuss what our actual requirements should really be, so I expect this to still change - for the rest of the docs, a review would be good already, though :-).

Changelog.md
```diff
@@ -3,6 +3,7 @@
 Language Features:
  * General: Support ``ContractName.functionName`` for ``abi.encodeCall``, in addition to external function pointers.
  * General: Add equality-comparison operators for external function types.
+ * General: Allow annotating inline assembly as memory-safe to allow optimizations and stack limit evasion that rely on respecting Solidity's memory model.
```
chriseth, line 4: This only applies to "via ir", does it?

docs/assembly.rst
```diff
@@ -268,3 +278,36 @@ first slot of the array and followed by the array elements.
...
+Memory Safety
+=============
+
+Without the use of inline assembly, the compiler can rely on memory to remain in a well-defined
+state at all times. This is especially relevant for :ref:`the new code generation pipeline via Yul IR <ir-breaking-changes>`:
+this code generation path can move local variables from stack to memory to avoid stack-too-deep errors and
+perform additional memory optimizations, if it can rely on certain assumptions about memory use.
+
+While we recommend to always respect Solidity's memory model, it is possible to use memory
+in an incompatible way from inline assembly. Therefore, moving stack variables to memory and additional
```
chriseth:
```suggestion
While we recommend to always respect Solidity's memory model, inline assembly allows you to use memory
in an incompatible way. Therefore, moving stack variables to memory and additional
```

docs/assembly.rst
```diff
@@ -268,3 +278,36 @@ first slot of the array and followed by the array elements.
...
+model as follows:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+        ...
+    }
+
+In particular, a memory-safe assembly block has to follow the adhere to the following restriction:
```
chriseth:
```suggestion
In particular, a memory-safe assembly block has to adhere to the following restriction:
```

Changelog.md
ekpyron: In practice at least... in principle, we could use the information for old codegen as well, but we probably won't.

docs/assembly.rst
```diff
@@ -268,3 +278,36 @@ first slot of the array and followed by the array elements.
...
+All accessed memory (read or written) was properly allocated. This can either be memory allocated by yourself
+using a mechanism like the ``allocate`` function described above, or memory allocated by Solidity,
+e.g. memory within the bounds of a memory array you reference. The only exception to this is the scratch space
+between memory offset 0 and 64 mentioned above.
+
```
chriseth, line 362: Maybe add: Since this is mainly about the optimizer, it does not matter that there is no Solidity code following your assembly block. As an example, the following assembly snippet is not memory safe:
```
assembly {
  returndatacopy(0, returndatasize())
  revert(0, returndatasize())
}
```
But the following is:
```
/// @solidity memory-safe-assembly
assembly {
  let p := mload(0x60)
  returndatacopy(0x60, returndatasize())
  revert(0x60, returndatasize())
}
```
(note that you do not need to update the free memory pointer if there is no following allocation, but can only use memory starting with where the free memory pointer points to)

If the memory operations use a length of zero, it is also fine to just use any offset (not only if it falls into the scratch space):
```
/// @solidity memory-safe-assembly
assembly {
  revert(0, 0)
}
```

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +162,56 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+	if (!errors.empty())
+	{
+		SecondarySourceLocation ssl;
+		for (auto const& error: errors)
+			if (error->comment())
+				ssl.append(
+					*error->comment(),
+					error->sourceLocation() ? *error->sourceLocation() : _assembly.location()
+				);
+		m_errorReporter.warning(
```
chriseth, line 184: Is this a breaking change?

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +162,56 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+	auto docTags = DocStringParser{documentation, errorReporter}.parse();
+
+	if (!errors.empty())
+	{
+		SecondarySourceLocation ssl;
+		for (auto const& error: errors)
+			if (error->comment())
+				ssl.append(
+					*error->comment(),
+					error->sourceLocation() ? *error->sourceLocation() : _assembly.location()
```
chriseth: Could this leak the `-1` node id?

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +162,56 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+			"Inline assembly has invalid natspec documentation.",
+			ssl
+		);
+	}
+
+	for (auto const& [tagName, tagValue]: docTags)
+	{
+		if (tagName == "solidity")
+		{
+			if (tagValue.content == "memory-safe-assembly")
```
chriseth: Do we already want to split this into words?

libsolidity/codegen/ir/IRGenerationContext.h
```diff
@@ -160,8 +160,8 @@ class IRGenerationContext
 
 	std::set<ContractDefinition const*, ASTNode::CompareByID>& subObjectsCreated() { return m_subObjects; }
 
-	bool inlineAssemblySeen() const { return m_inlineAssemblySeen; }
-	void setInlineAssemblySeen() { m_inlineAssemblySeen = true; }
+	bool memoryUnsafeInlineAssemblySeen() const { return m_memoryUnsafeInlineAssemblySeen; }
```
chriseth, line 163: Would be nice to have a type that can be set false->true but not true->false.

chriseth (commented):
Looks good!
Would it be much work to run a scanner over the inline assembly to see if it contains any memory operations?

libsolidity/analysis/DocStringTagParser.cpp
ekpyron: The main point of doing all this locally with an artificial AST node with ID -1 and all that is to be able to translate the errors to warnings here, s.t. it's not breaking...

chriseth: ok!

ekpyron: That's also implementation-wise the main thing I'm wondering about. We could also promote ``InlineAssembly`` to properly inherit ``StructurallyDocumented``, but then we need case distinctions in the "normal" natspec parsing logic to prevent it from becoming breaking on pain of stuff like this being proper errors instead of mere warnings.

ekpyron:
> Looks good! Would it be much work to run a scanner over the inline assembly to see if it contains any memory operations?

Memory operations themselves we can rather easily detect in the ``AsmAnalysis`` and store in an annotation during type checking...
I'm rather worried about cases like this:
```
function f() {
    bytes memory x;
    assembly {
        x := 64
    }
    x[0x20] = 42;
}
```
So it's not only memory operations, but also lvalue references to variables of reference type in memory... that we could probably also detect in ``AsmAnalysis``... and I guess that's everthing that can go wrong :-)?
This may actually also be worth a note in the docs...

chriseth:
Oh wow, this is true - haven't thought about that! What could be fine is: no memory operations and only access to value types.

ekpyron:
Hm... I wonder if I should add a nicer way to test this than command line tests - they are quite verbose and to check for all combinations of memory opcodes and memory reference variable assignments, we'll need quite a few tests...

libsolidity/analysis/DocStringTagParser.cpp
ekpyron: Not sure what you mean exactly... make it ``memory safe assembly`` or ``assembly: memory-safe`` or something? Or just split at the hyphens here in case we want to add more of them?

chriseth: Split the tag value at whitespace so that we get an array. Loop over the array. Create a warning for any unknown value and handle the value `memory-safe-assembly` as here.

test/libsolidity/memoryGuardTests/free_function.sol
```diff
@@ -0,0 +1,29 @@
+function safe() pure returns (uint256 x) {
+    assembly { x := 42 }
+    /// @solidity memory-safe-assembly
+    assembly { mstore(0, 0) }
+}
+function unsafe() pure returns (uint256 x) {
+    assembly { pop(mload(0)) }
```
ekpyron, line 7: Not sure reading actually needs to be unsafe...

test/libsolidity/semanticTests/externalContracts/prbmath_unsigned.sol
```diff
@@ -50,7 +50,7 @@ contract test {
 // compileViaYul: also
 // ----
 // constructor()
-// gas irOptimized: 1778342
+// gas irOptimized: 1778102
```
ekpyron: This is rather unexpected... probably a fluke due to the ``memoryguard`` preventing some rematerialization...

libsolidity/analysis/DocStringTagParser.cpp
ekpyron: It does emit a error location with ``-1`` node id for ``@param`` related errors... not sure if that actually bleeds through anywhere then, though... but maybe safer to just generally use ``_assembly.location()``.

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +162,80 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+
+	if (!errors.empty())
+	{
+		SecondarySourceLocation ssl;
+		for (auto const& error: errors)
+			if (error->comment())
+				ssl.append(
+					*error->comment(),
+					_assembly.location()
+				);
```
ekpyron, line 183: Alternatives would be: use an empty source location - or not emit any of this at all and just leave it at the simple warning without reporting the ``DocStringParser`` errors as secondary messages.

ekpyron:
FYI: neither ENS nor trident are "implicitly memory-safe", so both still need the annotations.

docs/assembly.rst
```diff
@@ -268,3 +278,83 @@ first slot of the array and followed by the array elements.
...
+=============
+
+Without the use of inline assembly, the compiler can rely on memory to remain in a well-defined
+state at all times. This is especially relevant for :ref:`the new code generation pipeline via Yul IR <ir-breaking-changes>`:
+this code generation path can move local variables from stack to memory to avoid stack-too-deep errors and
+perform additional memory optimizations, if it can rely on certain assumptions about memory use.
+
+While we recommend to always respect Solidity's memory model, inline assembly allows you to use memory
+in an incompatible way. Therefore, moving stack variables to memory and additional memory optimizations are,
+by default, disabled in the presence of any inline assembly block.
```
chriseth: Not true anymore

docs/assembly.rst
```diff
@@ -268,3 +278,83 @@ first slot of the array and followed by the array elements.
...
+model as follows:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+        ...
+    }
+
+In particular, a memory-safe assembly block has to adhere to the following restriction:
```
chriseth: Do you think we could hilight this a little?

docs/assembly.rst
```diff
@@ -268,3 +278,83 @@ first slot of the array and followed by the array elements.
...
+All accessed memory (read or written) was properly allocated. This can either be memory allocated by yourself
```
chriseth: Do we want to say that it's fine to read or write to positions >= `mload(0x60)` without actually updating the pointer?

chriseth: Ah, we actually say it below, so I would prefer this to be part of the specification.

docs/assembly.rst
```diff
@@ -268,3 +278,83 @@ first slot of the array and followed by the array elements.
...
+using a mechanism like the ``allocate`` function described above, or memory allocated by Solidity,
+e.g. memory within the bounds of a memory array you reference. The only exception to this is the scratch space
+between memory offset 0 and 64 mentioned above.
+
+Since this is mainly about the optimizer, it does not matter that there is no Solidity code following your assembly block.
```
chriseth: What I wanted to say is that it does not matter that the code reverts/terminates. Do you think this distinction is important?

docs/assembly.rst
```diff
@@ -268,3 +278,83 @@ first slot of the array and followed by the array elements.
...
+
+But the following is:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+      let p := mload(0x40)
+      returndatacopy(0x40, returndatasize())
+      revert(0x40, returndatasize())
```
chriseth:
```suggestion
      returndatacopy(p, returndatasize())
      revert(p, returndatasize())
```

docs/assembly.rst
ekpyron: I just took this directly from your suggestion - but yes, it's probably clearer to say that it doesn't matter if the code revers or terminates.

docs/assembly.rst
ekpyron: I'm not an expert for rst :-). I can try to find something, though.

docs/assembly.rst
```diff
@@ -268,3 +278,86 @@ first slot of the array and followed by the array elements.
...
+    }
+
+In particular, a memory-safe assembly block may only access the following memory ranges:
+
+- Memory allocated by yourself using a mechanism like the ``allocate`` function described above.
+- Memory allocated by Solidity, e.g. memory within the bounds of a memory array you reference.
+- The scratch space between memory offset 0 and 64 mentioned above.
+- Temporary memory that is located *after* the value of the free memory pointer at the beginning of the assembly block,
+  i.e. memory that is "allocated" at the free memory pointer without updating the free memory pointer.
+
```
chriseth, line 312: plus, if it writes to solidity memory variables, it can only assign allocated pointers or something like that.

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +163,71 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+	for (auto const& [tagName, tagValue]: docTags)
+	{
+		if (tagName == "solidity")
+		{
+			vector<string> values;
+			boost::split(values, tagValue.content, isWhiteSpace);
+
+			set<string> valuesSeen;
+			set<string> duplicates;
+			for (auto const& value: values | ranges::views::filter(not_fn(&string::empty)))
```
ekpyron, line 201: Instead of the ``boost::split`` I could also split on the fly using ranges...
```c++
constexpr auto splitString = [](auto _pred) {
		return ranges::views::split_when(move(_pred)) | ranges::views::transform([](auto&& _rng){
			return string_view(&*begin(_rng), static_cast<string_view::size_type>(ranges::distance(_rng)));
		});
};
for (auto value: tagValue.content | splitString(isWhiteSpace) | ranges::views::filter(not_fn(&string_view::empty)))
```
But maybe that's over the top...

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +163,71 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+						4377_error,
+						_assembly.location(),
+						"Value for @solidity tag in inline assembly specified multiple times: " + value
+					);
+		}
+		else
+			m_errorReporter.warning(
+				6269_error,
+				_assembly.location(),
+				"Unexpected natspec tag in inline assembly: " + tagName
```
chriseth:
```suggestion
				"Unexpected NatSpec tag in inline assembly: " + tagName
```

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -162,6 +163,71 @@ bool DocStringTagParser::visit(ErrorDefinition const& _error)
...
+		for (auto const& error: errors)
+			if (error->comment())
+				ssl.append(
+					*error->comment(),
+					_assembly.location()
+				);
+		m_errorReporter.warning(
+			7828_error,
+			_assembly.location(),
+			"Inline assembly has invalid natspec documentation.",
```
chriseth:
```suggestion
			"Inline assembly has invalid NatSpec documentation.",
```

test/libsolidity/memoryGuardTests/free_function.sol
chriseth: We can relax it later, but I don't see why you would read if you don't write...

chriseth (dismissed):

ekpyron:
Updated the warnings. We should probably be clear that we can also merge the other one, before merging this, though. Although it doesn't matter as long as we manage for the same release I guess.

test/libsolidity/syntaxTests/inlineAssembly/natspec_multi.sol
```diff
@@ -0,0 +1,19 @@
+function f() pure {
+    /// @unrelated
+
+    /// @before
+    ///
+    /// @solidity a   memory-safe-assembly b    c
+    ///           d
```
ekpyron: Oh... I'm just realizing that apparently the natspec parser reads this as a tag "before" with value
``@solidity a   memory-safe-assembly b    c``...
That's not ideal...
But maybe just fine, if I also output the value in the error message.

ekpyron: Not entirely sure, if this is "normal", though...

test/libsolidity/syntaxTests/inlineAssembly/natspec_multi_swallowed.sol
```diff
@@ -0,0 +1,19 @@
...
+    ///           d
+    /// @after bogus-value
+    assembly {}
+    /// @solidity memory-safe-assembly a a a
+    ///           memory-safe-assembly
+    assembly {}
+}
+// ----
+// Warning 6269: (177-188): Unexpected NatSpec tag "after" with value "bogus-value" in inline assembly.
+// Warning 6269: (177-188): Unexpected NatSpec tag "before" with value "@solidity a   memory-safe-assembly b    c           d" in inline assembly.
```
ekpyron, line 16: See https://github.com/ethereum/solidity/issues/12732
But if we report the full value here, I'd say it's still ok for this purpose and we can fix it with https://github.com/ethereum/solidity/issues/12732...

Cryptomooner111 (approved):

Cryptomooner111:
True

PR #12624, opened 2022-02-03
Adjust solc-js installation to match the new release process

cameel:
Related to #12583.

This should fix breakage in [`t_ems_compile_ext_colony`](https://app.circleci.com/pipelines/github/ethereum/solidity/21953/workflows/7f2b69f9-342e-4e85-8684-074134661d8c/jobs/962750) and [`b_bytecode_ems`](https://app.circleci.com/pipelines/github/ethereum/solidity/21953/workflows/7f2b69f9-342e-4e85-8684-074134661d8c/jobs/962749) that came up now that we merged https://github.com/ethereum/solc-js/pull/596.

Our publishing process changed so that now the content of `dist/` is the actual package. S
- ~I adjusted the bytecode compare script so that it now installs from the tarball and gets whatever structure we actually package. This will make it work no matter whether the published content is at the root, in `dist/` or anywhere else.~
    - **EDIT**: This breaks bytecode comparison because `npm run build:tarball` includes the command to download the release binary. Perhaps we should change it so that downloading the binary is a separate step. For now I switched to the solution with hard-coding `dist/` subdir.
- In external tests I'm hard-coding `dist/`. Using the tarball would complicate them too much IMO. This will unfortunately break once we change the structure of solc-js again but the adjustment will be simple.
- The PR check script for solc-bin also needed some adjustments (tested with https://github.com/ethereum/solc-bin/pull/109):
    - I added the `dist/` subdir.
    - I changed it so that installs the package instead of using the repo checkout directly (`prepare-report.js` now requires this).

### DIFF
--- a/scripts/bytecodecompare/storebytecode.sh
+++ b/scripts/bytecodecompare/storebytecode.sh
@@ -52,7 +52,7 @@ TMPDIR=$(mktemp -d)
         popd
 
         cp "$REPO_ROOT/scripts/bytecodecompare/prepare_report.js" .
-        npm install solc-js/
+        npm install ./solc-js/dist
 
         echo "Running the compiler..."
         # shellcheck disable=SC2035
--- a/scripts/solc-bin/bytecode_reports_for_modified_binaries.sh
+++ b/scripts/solc-bin/bytecode_reports_for_modified_binaries.sh
@@ -149,11 +149,11 @@ for binary_name in $platform_binaries; do
         if [[ $platform == emscripten-wasm32 ]] || [[ $platform == emscripten-asmjs ]]; then
             ln -sf "${solc_bin_dir}/${platform}/${binary_name}" "${solcjs_dir}/soljson.js"
             ln -sf "${solc_bin_dir}/${platform}/${binary_name}" "${solcjs_dir}/dist/soljson.js"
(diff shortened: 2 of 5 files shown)


ekpyron (approved):
I haven't followed the latest solc-js, typescript and ``dist/`` developments, but this PR still does not look like the current state of solc-js is particularly backwards compatible :-).

I'll approve and merge it nonetheless to fix CI for now...

cameel:
The thing is, it's compatible only as a published package - because only the content of the `dist/` directory gets published. And that's only the kind of compatibility that counts for us. It's not compatible if you're using the repo directly.

The fixes here are basically a mix of switching from using the repo to using the package and adding `dist/` in places where we can't easily do that.

These are actually changes that I would expect to be needed already in #12583 and I don't fully understand why they weren't. The switch to `dist/` already happended there. My theory is that having `"files": ["dist/abi.js", "dist/index.js", ...]` in `package.json` was what allowed the module loader to find these files despite this. Now when https://github.com/ethereum/solc-js/pull/596 replaced it with `"files": ["abi.js", "index.js", ...]` we got the breakage.

ekpyron:
I wonder if the separation into a dedicated build directory (i.e. ``dist/``) is really worth all the hassle it's causing :-).

PR #12625, opened 2022-02-03
Docs: fix slot calculation for mappings of strings/byte arrays

### LINKED ISSUE #12612: Doc: slot calculation for mappings of strings/byte arrays
## Page

https://docs.soliditylang.org/en/v0.8.11/internals/layout_in_storage.html#mappings-and-dynamic-arrays

## Abstract

> The value corresponding to a mapping key k is located at keccak256(h(k) . p) where . is concatenation and h is a function that is applied to the key depending on its type:
>
> - for value types, h pads the value to 32 bytes in the same way as when storing the value in memory.
>
> - for strings and byte arrays, h computes the keccak256 hash of the unpadded data.

I think the second bullet point here is incorrect. For strings and byte arrays, h is just the unpadded data and not its hash.

## Pull request

I can make a PR once I get a confirmation

shuklaayush:
FIxes #12612

### DIFF
--- a/docs/internals/layout_in_storage.rst
+++ b/docs/internals/layout_in_storage.rst
@@ -90,7 +90,7 @@ The value corresponding to a mapping key ``k`` is located at ``keccak256(h(k) . 
 where ``.`` is concatenation and ``h`` is a function that is applied to the key depending on its type:
 
 - for value types, ``h`` pads the value to 32 bytes in the same way as when storing the value in memory.
-- for strings and byte arrays, ``h`` computes the ``keccak256`` hash of the unpadded data.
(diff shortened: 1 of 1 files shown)


ekpyron (dismissed):

cameel:
Can you rebase on top of `develop`? It has fixes/workarounds for the CI jobs that failed here.

cameel:
Looks like we have some new CI breakage. Rebasing on #12633 (or waiting until it's in `develop` and then rebasing on `develop`) should fix the problem. Sorry for the extra complication - it's an unexpected upstream change :)

cameel (approved):

PR #12627, opened 2022-02-04
Fuzzer for LP Solver

bshastry:


test/tools/ossfuzz/lpsolver/ConstraintGenerator.cpp
```diff
@@ -0,0 +1,42 @@
...
+string ConstraintGenerator::generate()
+{
+	string constraint;
+	for (int i = 0; i < numConstraints(); i++)
+	{
+		// First entry is always constraint type. If it is equal to "1", it is an equality constraint
+		// otherwise an less-than-equal constraint.
+		constraint += to_string(zeroOrOne());
+		for (int j = 0; j < numFactors(); j++)
+			if (bernoulliDist(s_piecewiseConstantProb))
```
chriseth, line 35: Why this distinction?

test/tools/ossfuzz/LPSolverFuzzer.cpp
```diff
@@ -0,0 +1,110 @@
...
+		constraints.emplace_back(constraint);
+	}
+
+	if (!validConstraints(constraints))
+		return 0;
+	else
+	{
+		// TODO: Z3 on constraints provided by fuzzer interface and comparing its outcome
+		// with LP solver.
+		FuzzerSolverInterface solverWithoutModels(/*supportModels=*/false);
```
chriseth: I don't think you really need the FuzzerSolverInterface, your constraints already have the right structure to go directly into the solver state.

bshastry: Constraints parsed from fuzzer input are of the form `vector<pair<bool, vector<int>>>` but constraints in the solving state are a `vector<LinearExpression>`. This interface essentially facilitates the translation of the former to the latter.

chriseth: While that is true, the class has tons of other functions that you don't use and probably also don't need.

test/tools/ossfuzz/lpsolver/ConstraintGenerator.cpp
bshastry: Since the random integer generator (used for generating constant term and factors for variables in linear expressions) uses uniform random distribution over the range [INT_MIN, INT_MAX], the probability of zero is roughly `1/(INT_MAX - INT_MIN)` which is minuscule. If it weren't for this distinction, linear expressions that do not feature certain variables would be very rare because by convention the integer vector `{i0, i1, i2,...,in}` translates to

```
i1*x1 + i2*x2 + ... + in*xn <= i0
```

If `ik` is rarely zero, then it features almost always in a constraint which we may not really desire.

test/tools/ossfuzz/LPSolverFuzzer.cpp
bshastry: Do you have any specific functions in mind? I just reviewed that class again and found that I use almost all functions defined in that class.

Informal call graph:

Fuzzer calls `addConstraints` and `checkResult`

`addConstraints` -> `addConstraint`

`addConstraint` -> `addLEConstraint`, `addEQConstraint`

`addLEConstraint` translates constraint of the form `pair<bool, vector<int>>` to `LinearExpression`

and so on.

test/tools/ossfuzz/lpsolver/ConstraintGenerator.cpp
bshastry: Having said that the argument for `1` and `-1` in the piecewise range is weaker. Basically, I am not sure if they are special enough to be corner cases but I thought they wouldn't hurt :)

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
```diff
@@ -0,0 +1,136 @@
...
+	for (auto f: _factors)
+	{
+		if (first)
+		{
+			first = false;
+			lexp += constant(f);
+		}
+		else
+			lexp += variable(f, "x" + to_string(count++));
+	}
```
chriseth: You could change this to
```suggestion
	LinearExpression lexp;
	lexp.resize(_factors.size());
	for (auto&& [index, value]: _factors | ranges::views::enumerate)
		lexp[index] = rational{value};
```
and then you can delete variable and constant and the construction is much cheaper because you don't add vectors all the time.

test/tools/ossfuzz/LPSolverFuzzer.cpp
chriseth: Ah now I see how it's done! Ok, I was thinking more along the lines of https://github.com/ethereum/solidity/pull/12627/files#r801517125

bshastry: Oh wow, thanks for the suggestion. That got rid of multiple intermediate helper functions and also the unused function`zero()` :-)

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
```diff
@@ -0,0 +1,103 @@
...
+	m_solver(_supportModels)
+{
+	m_solvingState.variableNames.emplace_back("");
+}
+
+void FuzzerSolverInterface::addLEConstraint(LinearExpression _lhs)
+{
+	// Move constant to RHS
+	if (_lhs[0])
+		_lhs[0] = -_lhs[0];
```
chriseth: I mean this is also not really needed in your case, is it?

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	addLEConstraint(linearExpression(_factors));
+}
+
+void FuzzerSolverInterface::addConstraint(pair<bool, vector<int>> _constraint)
+{
+	if (_constraint.first)
+		addEQConstraint(_constraint.second);
+	else
+		addLEConstraint(_constraint.second);
```
chriseth:
```suggestion
	m_solvingState.constraints.push_back({linearExpression(move(_lhs)), _constraint.first});
```

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
bshastry: Right. This was a semantic convention to know how to read a fuzzer input while debugging i.e.,

```
1, 23, 2, 1
```

would be read as `2*x1 + x2 = 23` instead of `2*x1 + x2 = -23` i.e., not having to negate fuzzer generated input to derive the constraint.

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
bshastry: Fixed it.

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
bshastry: Fixed it by adjusting `linearExpression` to follow the first convention. Gets rid of the intermediate functions.

test/tools/ossfuzz/lpsolver/ConstraintGenerator.h
```diff
@@ -0,0 +1,86 @@
...
+	bool bernoulliDist(double _truthProbability)
+	{
+		return Bernoulli(_truthProbability)(*prng);
+	}
+
+
+	std::shared_ptr<RandomEngine> prng;
+
+	static constexpr int s_minFactors = 2;
+	static constexpr int s_maxFactors = 10;
```
chriseth: I think these number should be way higher. If we want to find bugs in the caching mechanism we either have to invoke the solver multiple times or use more variables / factors.

test/tools/ossfuzz/lpsolver/ConstraintGenerator.h
```diff
@@ -0,0 +1,86 @@
...
+	/// @returns random number of constraints.
+	int numConstraints()
+	{
+		return Distribution(s_minConstraints, s_maxConstraints)(*prng);
+	}
+
+	/// @returns an integer chosen uniformly at random.
+	int randomInteger()
+	{
+		return Distribution(std::numeric_limits<int>::min(), std::numeric_limits<int>::max())(*prng);
```
chriseth: Maybe we should actually limit this to +- 100 or something. Again about the caching: If we use smaller numbers, then the likelihood of repetitions is higher.

test/tools/ossfuzz/lpsolver/ConstraintGenerator.h
bshastry: So, say a 100x100 matrix as the upper bound? I just noticed that it cuts down execution speed by about 6x but of course priority should be finding bugs

test/tools/ossfuzz/lpsolver/ConstraintGenerator.h
bshastry: This was a good suggestion. Found a bug already in the PR: https://github.com/ethereum/solidity/pull/12013/files#r801989737

test/tools/ossfuzz/lpsolver/FuzzerSolverInterface.cpp
```diff
@@ -0,0 +1,151 @@
...
+)
+{
+	bool isEquality = _constraint.first;
+	expr reduce(m_z3Ctx);
+	unsigned count = 0;
+	for (auto&& [index, value]: _constraint.second | ranges::views::enumerate)
+	{
+		if (index != 0)
+			if (value != 0)
+				reduce = reduce + expr(value * _symbolicVars[count++]);
```
chriseth: shouldn't this be
```suggestion
				reduce = reduce + expr(value * _symbolicVars[index]);
```
?

leonardoalt:
@chriseth is this still gonna be used? Do we need a new one for the Combined Solver instead?

chriseth:
I completely re-wrote the lp solver, so we still need something like that. Maybe we should just implement smtlib input.

cameel:
Closing since we won't be merging the other solver PRs anyway.

PR #12628, opened 2022-02-04
Added support for FunctionType::Kind::StringConcat 

### LINKED ISSUE #12087: Properly allow `string.concat`

nishant-sachdeva:
Also adds functions for function string.concat.

Closes #12087


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2213,7 +2213,11 @@ void TypeChecker::typeCheckBytesConcatFunction(
 )
 {
```
ekpyron, line 2242: Yeah - in here you now need to adjust the type-checking of the arguments for ``string.concat``.
There actually may be little in common here between ``bytes.concat`` and ``string.concat``. ``string.concat`` should only allow arguments that are implicitly convertible to ``string memory``. You can case-distinguish in this function or it may even call for adding a new function ``typeCheckStringConcatFunction`` to do that.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1102,6 +1102,7 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 			break;
 		}
 		case FunctionType::Kind::BytesConcat:
+		case FunctionType::Kind::StringConcat:
```
ekpyron, line 1104: Here you will need to case-distinguish in the ``for (auto const& argument: arguments)`` loop below.
If it's ``BytesConcat`` you keep doing what it does now - if it's ``StringConcat``, you only do the equivalent of the ``else`` case, i.e. something like
```
					solAssert(argument->annotation().type->isImplicitlyConvertibleTo(*TypeProvider::stringMemory()), "");
					targetTypes.emplace_back(TypeProvider::stringMemory());
```
I expect the actual implementation after the loop will work out-of-the-box - we can confirm that in semantics tests.

test/libsolidity/syntaxTests/string/string_concat_with_parameters_as_bytes.sol
```diff
@@ -0,0 +1,11 @@
+contract C {
```
ekpyron: Would be good to have some more test cases, e.g. for fixed bytes types and number literals as arguments.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2207,13 +2207,43 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+
+	for (shared_ptr<Expression const> const& argument: _functionCall.arguments())
+	{
+		Type const* argumentType = type(*argument);
+		bool notConvertibleToString = !argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory());
+		bool numberLiteral = (dynamic_cast<RationalNumberType const*>(argumentType) != nullptr);
+
+		if (notConvertibleToString || numberLiteral)
+			m_errorReporter.typeError(
+				0000_error, // I'm not how to get new error codes for this, but we'll get to it in time
```
ekpyron: ``scripts/error_codes.py --fix``

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -1101,6 +1101,44 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+				solAssert(argument->annotation().type, "");
+				argumentTypes.emplace_back(argument->annotation().type);
+				if (argument->annotation().type->category() == Type::Category::FixedBytes)
+					targetTypes.emplace_back(argument->annotation().type);
+				else if (
+					auto const* literalType = dynamic_cast<StringLiteralType const*>(argument->annotation().type);
+					literalType && !literalType->value().empty() && literalType->value().size() <= 32
+				)
+					targetTypes.emplace_back(TypeProvider::fixedBytes(static_cast<unsigned>(literalType->value().size())));
+					// looks like it's converting stringLiteral type to FixedBytes. Probably should not do that for string.concat?
```
ekpyron: All of this should be removed shouldn't it?

Also I think it would be less code duplication, if you just case distinguished with another ``if (function.kind() == FunctionType::Kind::StringConcat)`` here, but apart from that used a shared
```
case FunctionType::Kind::StringConcat:
case FunctionType::Kind::BytesConcat:
```

libsolidity/analysis/TypeChecker.cpp
chriseth: You can also just come up with a random number yourself.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2474,6 +2474,54 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
 		}
 	});
 }
+string YulUtilFunctions::stringConcatFunction(vector<Type const*> const& _argumentTypes)
```
chriseth: I don't think we need a new function on the yul level, since it is identical to bytes concat.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2207,13 +2207,43 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+	solAssert(_functionType->kind() == FunctionType::Kind::StringConcat, "");
+	solAssert(_functionCall.names().empty(), "");
+
+	typeCheckFunctionGeneralChecks(_functionCall, _functionType);
+
+	for (shared_ptr<Expression const> const& argument: _functionCall.arguments())
+	{
+		Type const* argumentType = type(*argument);
+		bool notConvertibleToString = !argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory());
+		bool numberLiteral = (dynamic_cast<RationalNumberType const*>(argumentType) != nullptr);
```
chriseth: I don't think we need this special case. Some number literals are convertible to `bytes32`, but we excluded that already. Can you add a test with `string.concat(0)` to verify?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2207,13 +2207,43 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+
 void TypeChecker::typeCheckBytesConcatFunction(
 	FunctionCall const& _functionCall,
 	FunctionType const* _functionType
 )
 {
 	solAssert(_functionType, "");
-	solAssert(_functionType->kind() == FunctionType::Kind::BytesConcat, "");
+	solAssert(
+		_functionType->kind() == FunctionType::Kind::BytesConcat , "");
```
```suggestion
	solAssert(_functionType->kind() == FunctionType::Kind::BytesConcat, "");
```

libsolidity/ast/Types.cpp
```diff
@@ -3811,9 +3812,22 @@ MemberList::MemberMap TypeType::nativeMembers(ASTNode const* _currentScope) cons
 			)
 		);
 	}
+	else if (auto const* arrayType = dynamic_cast<ArrayType const*>(m_actualType);
```
chriseth:
```suggestion
	else if (
		auto const* arrayType = dynamic_cast<ArrayType const*>(m_actualType);
```
(If you have a line break inside a pair of parentheses, there also needs to be a line break right after the opening parenthesis and right before the closing one)

libsolidity/codegen/YulUtilFunctions.cpp
chriseth: Maybe more specifically: The code here is almost the same as in bytesConcat except for the type checking. Maybe you could extract that from both and then turn the part starting at `createFunction` into a third more low-level function?
Alternatively, you could add a parameter and combine the type checking part.

chriseth (changes requested):

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -314,7 +314,12 @@ class YulUtilFunctions
 
 	/// @returns the name of a function that does concatenation of variadic number of bytes
 	/// or fixed bytes
-	std::string bytesConcatFunction(std::vector<Type const*> const& _argumentTypes);
+	/// or Strings
+	std::string concatBytesOrStringFunctionCreate(
+		std::string functionName,
```
chriseth: I think you should derive the name from the function type kind inside the function

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -314,7 +314,12 @@ class YulUtilFunctions
 
 	/// @returns the name of a function that does concatenation of variadic number of bytes
 	/// or fixed bytes
-	std::string bytesConcatFunction(std::vector<Type const*> const& _argumentTypes);
+	/// or Strings
+	std::string concatBytesOrStringFunctionCreate(
```
chriseth:
```suggestion
	std::string bytesOrStringConcatFunction(
```

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,20 +2475,29 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
...
 }
 
-string YulUtilFunctions::bytesConcatFunction(vector<Type const*> const& _argumentTypes)
+string YulUtilFunctions::concatBytesOrStringFunctionCreate(
+	std::string functionName,
+	vector<Type const*> const& _argumentTypes,
+	FunctionType::Kind const& functionTypeKind
+)
 {
-	string functionName = "bytes_concat";
```
chriseth, line 2480: Please add:
`solAssert(functionTypeKind == FunctionType::Kind::BytesConcat || functionTypeKind == FunctionType::Kind::StringConcat);`

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,20 +2475,29 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
 	});
 }
 
-string YulUtilFunctions::bytesConcatFunction(vector<Type const*> const& _argumentTypes)
+string YulUtilFunctions::concatBytesOrStringFunctionCreate(
+	std::string functionName,
+	vector<Type const*> const& _argumentTypes,
+	FunctionType::Kind const& functionTypeKind
```
chriseth:
```suggestion
	FunctionType::Kind functionTypeKind
```
(it's a simple value type, so no need to use a const&)

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,20 +2475,29 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
...
+
 	for (Type const* argumentType: _argumentTypes)
 	{
-		solAssert(
-			argumentType->isImplicitlyConvertibleTo(*TypeProvider::bytesMemory()) ||
-			argumentType->isImplicitlyConvertibleTo(*TypeProvider::fixedBytes(32)),
-			""
-		);
+		if (functionTypeKind == FunctionType::Kind::StringConcat)
+			solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory()), "");
```
chriseth:
```suggestion
			solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory()));
```
(we added an overload recently)

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,20 +2475,29 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
...
+		else if (functionTypeKind == FunctionType::Kind::BytesConcat)
+		{
```
chriseth: No need for braces

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2497,14 +2506,21 @@ string YulUtilFunctions::bytesConcatFunction(vector<Type const*> const& _argumen
 		else
 		{
 			solAssert(!dynamic_cast<RationalNumberType const*>(argumentType), "");
-			solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::bytesMemory()), "");
-			targetTypes.emplace_back(TypeProvider::bytesMemory());
+			if (functionTypeKind == FunctionType::Kind::StringConcat)
+			{
+				solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory()), "");
```
chriseth: Don't we already check that above?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,36 +2475,45 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
...
+				targetTypes.emplace_back(argumentType);
 		else if (
 			auto const* literalType = dynamic_cast<StringLiteralType const*>(argumentType);
 			literalType && !literalType->value().empty() && literalType->value().size() <= 32
 		)
 			targetTypes.emplace_back(TypeProvider::fixedBytes(static_cast<unsigned>(literalType->value().size())));
 		else
 		{
-			solAssert(!dynamic_cast<RationalNumberType const*>(argumentType), "");
-			solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::bytesMemory()), "");
```
ekpyron, line 2500: Better to keep some assertion towards this end, just to make sure.
E.g.
```c++
auto targetType = functionTypeKind == FunctionType::Kind::StringConcat ? TypeProvider::stringMemory()  : TypeProvider::bytesMemory();
solAssert(argumentType->isImplicitlyConvertibleTo(*targetType), "");
targetTypes.emplace_back(targetType);
```

ekpyron: Ah, ok, that's done above already, right? Then nevermind :-).

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2207,13 +2207,42 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+}
+
 void TypeChecker::typeCheckBytesConcatFunction(
 	FunctionCall const& _functionCall,
 	FunctionType const* _functionType
 )
 {
 	solAssert(_functionType, "");
-	solAssert(_functionType->kind() == FunctionType::Kind::BytesConcat, "");
+	solAssert(_functionType->kind() == FunctionType::Kind::BytesConcat , "");
```
ekpyron:
```suggestion
	solAssert(_functionType->kind() == FunctionType::Kind::BytesConcat, "");
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2207,13 +2207,42 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+	solAssert(_functionType->kind() == FunctionType::Kind::StringConcat, "");
+	solAssert(_functionCall.names().empty(), "");
+
+	typeCheckFunctionGeneralChecks(_functionCall, _functionType);
+
+	for (shared_ptr<Expression const> const& argument: _functionCall.arguments())
+	{
+		Type const* argumentType = type(*argument);
+		bool notConvertibleToString = !argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory());
+		bool numberLiteral = (dynamic_cast<RationalNumberType const*>(argumentType) != nullptr);
```
ekpyron: What about this? As @chriseth says in https://github.com/ethereum/solidity/pull/12628/files#r801353987, I also think this is unnecessary, since ``RationalNumberType`` is not implicitly convertible to ``string memory``, so ``notConvertibleToString`` will always be true, if ``numberLiteral`` is true. So to be clear: you can remove this line and the ``|| numberLiteral`` part from the condition below and this should not affect tests or alter behaviour.

libsolidity/ast/Types.cpp
```diff
@@ -3813,7 +3814,21 @@ MemberList::MemberMap TypeType::nativeMembers(ASTNode const* _currentScope) cons
 	}
 	else if (
 		auto const* arrayType = dynamic_cast<ArrayType const*>(m_actualType);
-		arrayType && arrayType->isByteArrayOrString()
+		arrayType && arrayType->isString()
+	)
```
ekpyron: No need to do the dynamic cast twice - 
```
else if (auto const* arrayType = dynamic_cast<ArrayType const*>(m_actualType))
{
   if (arrayType->isString())
           ...
   else if (arrayType->isBytes())
           ...
}
```
should do. Or you might even use ``isByteArrayAndString`` and just switch out function type and return parameter in a single ``members.emplace_back``.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -2475,36 +2475,45 @@ string YulUtilFunctions::copyArrayFromStorageToMemoryFunction(ArrayType const& _
...
+			solAssert(argumentType->isImplicitlyConvertibleTo(*TypeProvider::stringMemory()));
+		else if (functionTypeKind == FunctionType::Kind::BytesConcat)
+			solAssert(
+				argumentType->isImplicitlyConvertibleTo(*TypeProvider::bytesMemory()) ||
+				argumentType->isImplicitlyConvertibleTo(*TypeProvider::fixedBytes(32))
+			);
+
 		if (argumentType->category() == Type::Category::FixedBytes)
-			targetTypes.emplace_back(argumentType);
+				targetTypes.emplace_back(argumentType);
```
ekpyron:
```suggestion
			targetTypes.emplace_back(argumentType);
```
That's too much indentation, isn't it?

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -314,7 +314,11 @@ class YulUtilFunctions
 
 	/// @returns the name of a function that does concatenation of variadic number of bytes
 	/// or fixed bytes
-	std::string bytesConcatFunction(std::vector<Type const*> const& _argumentTypes);
+	/// or Strings
```
ekpyron: Can you instead extend the line above by something like
``, if @a functionTypeKind is FunctionType::Kind::BytesConcat, or of strings, if @a functionTypeKind is FunctionType::Kind::StringConcat.``
?

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -314,7 +314,11 @@ class YulUtilFunctions
...
+	std::string bytesOrStringConcatFunction(
+		std::vector<Type const*> const& _argumentTypes,
+		FunctionType::Kind functionTypeKind
```
ekpyron:
```suggestion
		FunctionType::Kind _functionTypeKind
```

ekpyron (changes requested):
Apart from the immediate comments: can you make a note of the tests that produce different results when compiled to ewasm and create a new issue for investigating those cases?

ekpyron:
I need to recheck everything once more, but for now this needs a changelog entry and documentation updates.

Changelog.md
```diff
@@ -16,6 +16,7 @@ Bugfixes:
  * Code Generator: Fix ICE when accessing the members of external functions occupying more than two stack slots.
  * Code Generator: Fix ICE when doing an explicit conversion from ``string calldata`` to ``bytes``.
  * Control Flow Graph: Perform proper virtual lookup for modifiers for uninitialized variable and unreachable code analysis.
+ * General: Fixed signature of string.concat to work for Strings instead of Bytes ( as was the case earlier ).
```
ekpyron: Maybe something like this?
```suggestion
 * General: ``string.concat`` now properly takes strings as arguments and returns ``string memory``. It was accidentally introduced as a copy of ``bytes.concat`` before.
```

docs/types/reference-types.rst
```diff
@@ -165,6 +165,32 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
     and not the individual characters.
 
+.. index:: ! string-concat
+
+.. _string-concat:
+
+``string.concat`` function
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can concatenate a variable number of ``string`` variables using ``string.concat``.
```
chriseth: Wouldn't it be better to only have one section for both?

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
     that you are accessing the low-level bytes of the UTF-8 representation,
     and not the individual characters.
 
-.. index:: ! bytes-concat
```
chriseth, line 168: Please keep the index. You can also add multiple items here.

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+.. index:: ! concat-functions
 
-.. _bytes-concat:
+.. _string-concat:
 
-``bytes.concat`` function
-^^^^^^^^^^^^^^^^^^^^^^^^^
+``<bytes/string>.concat`` function
```
chriseth: What about "The functions ``bytes.concat`` and ``string.concat``"

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Using the string.concat function you can concatenate a variable number of ``string`` variables using ``string.concat``.
```
chriseth:
```suggestion
You can concatenate a arbitrary number of ``string`` values using ``string.concat``.
```

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+The function returns a single ``string memory`` array that contains the contents of the arguments without padding.
+If you want to use parameters of other types, you need to convert them to ``string`` first.
 
-You can concatenate a variable number of ``bytes`` or ``bytes1 ... bytes32`` using ``bytes.concat``.
+Analogously, the bytes.concat function can concatenate a variable number of ``bytes`` or ``bytes1 ... bytes32`` using ``bytes.concat``.
```
chriseth:
```suggestion
Analogously, the ``bytes.concat`` function can concatenate a variable number of ``bytes`` or ``bytes1 ... bytes32`` values.
```

chriseth: Actually I think for both bytes.concat and string.concat you can also use array slices

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+    pragma solidity ^0.8.12;
 
     contract C {
-        bytes s = "Storage";
-        function f(bytes calldata c, string memory m, bytes16 b) public view {
-            bytes memory a = bytes.concat(s, c, c[:2], "Literal", bytes(m), b);
-            assert((s.length + c.length + 2 + 7 + bytes(m).length + 16) == a.length);
+        string s = "Storage";
+        function f(bytes calldata c, string memory m, string memory b) public view {
+            string memory concat_string = string.concat(s, string(c), "Literal", m , b);
```
chriseth:
```suggestion
            string memory concat_string = string.concat(s, string(c), "Literal", m, b);
```

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+
 .. code-block:: solidity
 
     // SPDX-License-Identifier: GPL-3.0
-    pragma solidity ^0.8.4;
+    pragma solidity ^0.8.12;
 
     contract C {
-        bytes s = "Storage";
-        function f(bytes calldata c, string memory m, bytes16 b) public view {
```
chriseth, line 186: Can you have at least one `bytes16` (or other fixed bytes) variable in the example?

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+            assert((bytes(s).length + c.length + 7 + bytes(m).length + bytes(b).length) == bytes(concat_string).length);
+
+            bytes memory concat_bytes = bytes.concat(bytes(s), c, c[:2], "Literal", bytes(m), bytes(b));
+            assert((bytes(s).length + c.length + 2 + 7 + bytes(m).length + bytes(b).length) == concat_bytes.length);
         }
     }
 
-If you call ``bytes.concat`` without arguments it will return an empty ``bytes`` array.
+If you call ``string.concat`` without arguments it will return an empty ``string`` array.
+Whereas, If you call ``bytes.concat`` without arguments it will return an empty ``bytes`` array.
```
chriseth:
```suggestion
If you call ``string.concat`` or ``bytes.concat`` without arguments they return an empty array.
```

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
 
-.. index:: ! bytes-concat
+.. index:: ! bytes-concat, ! string-concat
 
 .. _bytes-concat:
+.. _string-concat:
 
-``bytes.concat`` function
-^^^^^^^^^^^^^^^^^^^^^^^^^
+The functions ``bytes-concat`` and ``string-concat``
```
chriseth: Why not bytes.concat and string.concat?

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can concatenate a variable number of ``bytes`` or ``bytes1 ... bytes32`` using ``bytes.concat``.
-The function returns a single ``bytes memory`` array that contains the contents of the arguments without padding.
+You can concatenate a arbitrary number of ``string`` values using ``string.concat``.
```
chriseth:
```suggestion
You can concatenate an arbitrary number of ``string`` values using ``string.concat``.
```

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
+The function returns a single ``string memory`` array/array-slices that contains the contents of the arguments without padding.
```
chriseth: The arguments can be strings or string slices, not the return value.

chriseth: Maybe just add to the previous sencence: You can use anything that can be implicitly converted to `string memory`, which also includes string slices and string literals.

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
 .. code-block:: solidity
 
     // SPDX-License-Identifier: GPL-3.0
-    pragma solidity ^0.8.4;
+    pragma solidity ^0.8.12;
 
     contract C {
-        bytes s = "Storage";
+        string s = "Storage";
         function f(bytes calldata c, string memory m, bytes16 b) public view {
```
chriseth: Maybe call the variables `bc`, `sm` and `b`?

docs/types/reference-types.rst
```diff
@@ -165,31 +165,40 @@ always use one of the value types ``bytes1`` to ``bytes32`` because they are muc
...
-^^^^^^^^^^^^^^^^^^^^^^^^^
+The functions ``bytes.concat`` and ``string.concat``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can concatenate a variable number of ``bytes`` or ``bytes1 ... bytes32`` using ``bytes.concat``.
+You can concatenate an arbitrary number of ``string`` values/array-slices using ``string.concat``.
+The function returns a single ``string memory`` array that contains the contents of the arguments without padding.
+If you want to use parameters of other types, you need to convert them to ``string`` first.
+
+Analogously, the ``bytes.concat`` function can concatenate an arbitrary number of ``bytes`` or ``bytes1 ... bytes32`` values/array-slices.
```
chriseth:
```suggestion
You can concatenate an arbitrary number of ``string`` values using ``string.concat``.
The function returns a single ``string memory`` array that contains the contents of the arguments without padding.
If you want to use parameters of other types that are not implicitly convertible to ``string``, you need to convert them to ``string`` first.

Analogously, the ``bytes.concat`` function can concatenate an arbitrary number of ``bytes`` or ``bytes1 ... bytes32`` values.
```

Since we also don't mention string literals, maybe adding "not implicitly convertible" is enough.

chriseth (approved):

test/libsolidity/semanticTests/strings/concat/string_concat_2_args.sol
```diff
@@ -0,0 +1,14 @@
+contract C {
+    function f(string memory a, string memory b) public returns (string memory) {
+        return string.concat(a, b);
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// f(string,string): 0x40, 0x80, 32, "abcdabcdabcdabcdabcdabcdabcdabcd", 5, "bcdef" -> 0x20, 0x25, 0x6162636461626364616263646162636461626364616263646162636461626364, 44502269928904312298000709931354278973409164155382318144318241583783949107200
```
ekpyron, line 9: The expectation formatting here is unfortunate - we may want to file an ``isoltest`` issue about that, if there is not already... for ``bytes memory`` it works better.
Nothing we need to do in this PR, though.

ekpyron (approved):

PR #12633, opened 2022-02-04
Always run Hardhat on node.js 16

cameel:
[Hardhat has just added a check against unsupported node.js versions](https://github.com/nomiclabs/hardhat/issues/2120). Turns out the core tests worked just fine on 17 though and we've been doing just that. Now `t_ems_ext_hardhat` job fails. This PR switches it to node.js 16.

There are also two tiny refactors:
- `base_node_latest_small` renamed to `base_node_small`. This better reflects the fact that version is just not specified and meshes better with jobs where I override the version.
- The notes about failures on node.js 17 are no longer true. Since it's officially not supported, I just removed them altogether since it's not something to fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1137,6 +1137,8 @@ jobs:
 
   t_ems_ext_hardhat:
     <<: *base_node_small
+    docker:
+      - image: circleci/node:16
     environment:
       TERM: xterm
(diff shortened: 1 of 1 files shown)


chriseth:
Can you remove the merge commit?

cameel:
Removed. It came from my botched merge of #12634/#12642.

chriseth (approved):

PR #11426, opened 2021-05-21
Move contracts/FixedFeeRegistrar to semanticTests/externalContracts

axic:
Depends on #11423.

PR #12634, opened 2022-02-05
Remove a reference to unused SolidityFixedFeeRegistrar

a3d4:
This fixes `soltest` on Windows in Debug configuration.

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -57,7 +57,7 @@ void removeTestSuite(std::string const& _name)
 {
 	master_test_suite_t& master = framework::master_test_suite();
 	auto id = master.get(_name);
-	assert(id != INV_TEST_UNIT_ID);
+	soltestAssert(id != INV_TEST_UNIT_ID, "Removing non-existent test suite!");
 	master.remove(id);
 }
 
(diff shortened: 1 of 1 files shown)


test/boostTest.cpp
```diff
@@ -279,7 +279,6 @@ test_suite* init_unit_test_suite( int /*argc*/, char* /*argv*/[] )
 			"ABIDecoderTest",
 			"ABIEncoderTest",
 			"SolidityAuctionRegistrar",
-			"SolidityFixedFeeRegistrar",
```
cameel, line 282: We still have it under `externalTests/FixedFeeRegistrar`. Not sure why these suites are disabled though (i.e. failures, slowdowns, or maybe they already run separately?) so I'm not sure if we should update the path or remove it.

BTW, what's the problem with it on Windows? What are we missing that it does not fail in CI?

cameel: Ah, wait these reference the suites from .cpp files and in #11426 @axic moved this one to a .sol file. So regardless of why these suites are ignored, `SolidityFixedFeeRegistrar` is indeed not needed here.

cameel (approved):

test/boostTest.cpp
a3d4: >what's the problem with it on Windows

The problem was in Debug configuration (Windows should not really be relevant indeed), because of the failing `assert`.

Maybe I should put it a way around, and say that the problem was that tests were succeeding in non-Debug configurations because `assert` was disabled there :-).

cameel: Oh, really? We should definitely have plain asserts enabled, even if we generally avoid them. That can be done separately though: #12641.

cameel:
I mistakenly merged it before merging the base PR so it went into the wrong branch. Reopened as #12642.

PR #12640, opened 2022-02-07
Re-enable preset for poolTogether.

chriseth:


cameel (approved):
If it passes, it's good :)

cameel:
Ah, right. The hardhat test needs my fix from #12633.

Issue #12621, opened 2022-02-03
AST JSON tests do not clear AST for parse-only test cases.


ekpyron:
For example ``test/libsolidity/ASTJSON/assembly/switch.json`` and ``ASTJSON/assembly/nested_functions.sol`` can have arbitrary contents and the tests still pass.
My guess is that's because it's tests with syntax errors and the tests only consider the ``parseOnly`` cases. I'm not sure it's intentional for the cases, though, or whether they just broke at some point unnoticed. In any case this should be fixed.

Furthermore, it's not so nice that the AST JSON tests throw, if the ``.json`` files are missing - it should rather just create them.

chriseth:
The logic for these tests are in `test/libsolidity/ASTJSONTest.h` / `.cpp`

You should be able to run only these tests using `isoltest -t *ASTJ*` or something similar.

PR #12653, opened 2022-02-10
Added errorSelector field for custom errors

### LINKED ISSUE #11387: Add selector for custom errors in compiler output
## Abstract

Add the selector of a custom error to the `sources.contract.ast` entry of the compiler output, similar to how functions have a `functionSelector` property.

## Motivation

To detect which custom error was thrown, you need to inspect the return data of a revert and check the first four bytes to see which signature they match. If a tool wants to use the compiler output to detect this automatically (like Hardhat does), right now you'll need to manually build the signature and hash it. For simple types this is not that hard, but if the signature includes nested data with arrays and or structs, things get way harder. 

## Specification

I'm not sure about this part. I guess `errorSelector` would work, analog to the `functionSelector` one. I also wonder if it would make sense to have an `errorIdentifiers` entry, similar to the current `methodIdentifiers`.

Also, and maybe unrelated to this issue: the same thing happens with events. Since as far as I know custom errors are somewhat modeled after events, maybe it would make sense to include them as part of this change. We don't need that right now, but I think they should be in the output too. (If that would make more sense as a separate issue, please let me know and I'll submit it).

## Backwards Compatibility

I think adding a new property in the compiler output is backwards compatible.

nishant-sachdeva:
Closes #11387

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,10 +4,9 @@ Language Features:
 
 
 Compiler Features:
-
+* JSON-AST: Added selector field for errors and events.
 
 Bugfixes:
-
 
 
 ### 0.8.12 (2022-02-16)
--- a/libsolidity/ast/ASTJsonConverter.cpp
+++ b/libsolidity/ast/ASTJsonConverter.cpp
@@ -32,6 +32,7 @@
 #include <libsolutil/JSON.h>
 #include <libsolutil/UTF8.h>
 #include <libsolutil/CommonData.h>
+#include <libsolutil/Keccak256.h>
 
 #include <boost/algorithm/string/join.hpp>
 
@@ -493,24 +494,36 @@ bool ASTJsonConverter::visit(ModifierInvocation const& _node)
 bool ASTJsonConverter::visit(EventDefinition const& _node)
 {
 	m_inEvent = true;
-	setJsonNode(_node, "EventDefinition", {
+	std::vector<pair<string, Json::Value>> _attributes = {
 		make_pair("name", _node.name()),
 		make_pair("nameLocation", sourceLocationToString(_node.nameLocation())),
 		make_pair("documentation", _node.documentation() ? toJson(*_node.documentation()) : Json::nullValue),
 		make_pair("parameters", toJson(_node.parameterList())),
 		make_pair("anonymous", _node.isAnonymous())
-	});
+	};
+	if (m_stackState >= CompilerStack::State::AnalysisPerformed)
+			_attributes.emplace_back(
+				make_pair(
+					"eventSelector",
+					toHex(u256(h256::Arith(util::keccak256(_node.functionType(true)->externalSignature()))))
+				));
+
+	setJsonNode(_node, "EventDefinition", std::move(_attributes));
 	return false;
 }
 
 bool ASTJsonConverter::visit(ErrorDefinition const& _node)
 {
-	setJsonNode(_node, "ErrorDefinition", {
+	std::vector<pair<string, Json::Value>> _attributes = {
 		make_pair("name", _node.name()),
 		make_pair("nameLocation", sourceLocationToString(_node.nameLocation())),
 		make_pair("documentation", _node.documentation() ? toJson(*_node.documentation()) : Json::nullValue),
 		make_pair("parameters", toJson(_node.parameterList()))
-	});
+	};
+	if (m_stackState >= CompilerStack::State::AnalysisPerformed)
+		_attributes.emplace_back(make_pair("errorSelector", _node.functionType(true)->externalIdentifierHex()));
+
+	setJsonNode(_node, "ErrorDefinition", std::move(_attributes));
 	return false;
 }
 
--- a/test/libsolidity/ASTJSON/documentation.json
+++ b/test/libsolidity/ASTJSON/documentation.json
@@ -162,6 +162,7 @@
             "src": "69:26:3",
             "text": "Some comment on Evt."
           },
+          "eventSelector": "a69007916fc1145953e5a7032d7c3eab4b8e2f33ec59b0f71e732904eeede3a4",
           "id": 12,
           "name": "Evt",
           "nameLocation": "102:3:3",
--- a/test/libsolidity/ASTJSON/used_errors.json
+++ b/test/libsolidity/ASTJSON/used_errors.json
@@ -20,6 +20,7 @@
   "nodes":
   [
     {
+      "errorSelector": "c1599bd9",
       "id": 2,
       "name": "X",
       "nameLocation": "6:1:1",
@@ -127,6 +128,7 @@
       "nodes":
       [
         {
+          "errorSelector": "2bc80f3a",
           "id": 11,
           "name": "T",
           "nameLocation": "63:1:1",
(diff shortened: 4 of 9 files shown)


libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -404,7 +404,8 @@ bool ASTJsonConverter::visit(FunctionDefinition const& _node)
 		make_pair("modifiers", toJson(_node.modifiers())),
 		make_pair("body", _node.isImplemented() ? toJson(_node.body()) : Json::nullValue),
 		make_pair("implemented", _node.isImplemented()),
-		make_pair("scope", idOrNull(_node.scope()))
+		make_pair("scope", idOrNull(_node.scope())),
+		make_pair("errorSelector", _node.functionType(true)->externalIdentifierHex())
```
chriseth: This should go to `ErrorDefinition`

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -404,7 +404,7 @@ bool ASTJsonConverter::visit(FunctionDefinition const& _node)
 		make_pair("modifiers", toJson(_node.modifiers())),
 		make_pair("body", _node.isImplemented() ? toJson(_node.body()) : Json::nullValue),
 		make_pair("implemented", _node.isImplemented()),
-		make_pair("scope", idOrNull(_node.scope()))
+		make_pair("scope", idOrNull(_node.scope())),
```
ekpyron:
```suggestion
		make_pair("scope", idOrNull(_node.scope()))
```

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -498,7 +498,8 @@ bool ASTJsonConverter::visit(EventDefinition const& _node)
 		make_pair("nameLocation", sourceLocationToString(_node.nameLocation())),
 		make_pair("documentation", _node.documentation() ? toJson(*_node.documentation()) : Json::nullValue),
 		make_pair("parameters", toJson(_node.parameterList())),
-		make_pair("anonymous", _node.isAnonymous())
+		make_pair("anonymous", _node.isAnonymous()),
+		make_pair("eventSelector", _node.functionType(true)->externalIdentifierHex())
```
ekpyron: Is this really what we want to output? Compare with https://github.com/ethereum/solidity/pull/12548.

test/libsolidity/ASTJSON/used_errors.json
```diff
@@ -1,231 +0,0 @@
-{
```
ekpyron: This file being removed happens due to the test file containing analysis errors - I recently reported this as https://github.com/ethereum/solidity/issues/12621
But in this case here: is it intentional that there are analysis errors in this test? Probably not (I haven't looked at the input myself)? Then we should fix them.

ekpyron (commented):
Some comments - also needs a changelog entry.

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -493,24 +493,32 @@ bool ASTJsonConverter::visit(ModifierInvocation const& _node)
...
-	setJsonNode(_node, "EventDefinition", {
+	std::vector<pair<string, Json::Value>> _attributes = {
 		make_pair("name", _node.name()),
 		make_pair("nameLocation", sourceLocationToString(_node.nameLocation())),
 		make_pair("documentation", _node.documentation() ? toJson(*_node.documentation()) : Json::nullValue),
 		make_pair("parameters", toJson(_node.parameterList())),
 		make_pair("anonymous", _node.isAnonymous())
-	});
+	};
+	if (m_stackState > CompilerStack::State::ParsedAndImported)
```
chriseth: Can you use `>=` for comparison? We need type checking here, don't we?

chriseth: I'm not sure if `AnalysisPerformed` is acutally enough as well, because it is set even if there were errors.

test/libsolidity/ASTJSON/documentation.json
```diff
@@ -162,6 +162,7 @@
             "src": "69:26:3",
             "text": "Some comment on Evt."
           },
+          "eventSelector": "a6900791",
```
chriseth: This is too short (just to keep an open note)

libsolidity/ast/ASTJsonConverter.cpp
```diff
@@ -493,24 +493,32 @@ bool ASTJsonConverter::visit(ModifierInvocation const& _node)
...
+	std::vector<pair<string, Json::Value>> _attributes = {
 		make_pair("name", _node.name()),
 		make_pair("nameLocation", sourceLocationToString(_node.nameLocation())),
 		make_pair("documentation", _node.documentation() ? toJson(*_node.documentation()) : Json::nullValue),
 		make_pair("parameters", toJson(_node.parameterList())),
 		make_pair("anonymous", _node.isAnonymous())
-	});
+	};
+	if (m_stackState >= CompilerStack::State::AnalysisPerformed)
+			_attributes.emplace_back(make_pair("eventSelector", _node.functionType(true)->externalIdentifierHex()));
```
ekpyron:
```suggestion
			_attributes.emplace_back(make_pair("eventSelector", toHex(u256(h256::Arith(keccak256(_node.functionType(true)->externalSignature())))));
```
Something like this.

test/libsolidity/ASTJSON/used_errors.json
nishant-sachdeva: I restored the contents of the file after recent code changes ( as an experiment ) and the tests  ran perfectly. Have added in the latest push to the PR

Marenz (changes requested):
I don't see any test case result that includes `errorSelector`. We should have at least one test for that

chriseth:
Please add a changelog entry.

Changelog.md
```diff
@@ -2,12 +2,14 @@
 
 Language Features:
 
-
 Compiler Features:
-
+* General: Added selector field in the Error and Event Description.
```
```suggestion
* JSON-AST: Added selector field for errors and events.
```

Changelog.md
```diff
@@ -2,12 +2,14 @@
...
 
 Bugfixes:
 
+Solc-Js:
```
chriseth: Please remove.

chriseth:
Can you please add a test where you have an event that uses internal types like an internal function? For example `event E(function() internal);` - this should generate the AST json, but without an error selector.

Changelog.md
```diff
@@ -2,14 +2,12 @@
 
 Language Features:
 
-
```
chriseth, line 5: Please keep the double empty lines.

test/libsolidity/ASTJSON/used_errors.sol
```diff
@@ -2,7 +2,7 @@ error X();
 function f() { revert X(); }
 contract C {
     error T();
-    function h() public pure { f(); }
+    function h() public { f(); }
```
chriseth, line 5: Why did you remove the pure?

chriseth (changes requested):
Please don't forget to add the suggested test case.

test/libsolidity/ASTJSON/used_errors.sol
nishant-sachdeva: The code in this .sol file was giving errors on compilation.  Removing pure fixed them

chriseth (approved):

PR #12362, opened 2021-12-02
Using with operators

chriseth:
Fixes #11969. Part of #13718.
~Depends on #13660.~ Merged.
~Depends on #13685.~ Merged.
~Depends on #13693.~ Merged.
~Depends on #13739.~ Merged.
Depends on #13741.

PR #12656, opened 2022-02-11
User-defined literal suffixes.

chriseth:
Part of #13718.
~Depends on #14066.~ Merged.
~Depends on #14087.~ Merged.
~Depends on #14104.~ Merged.

## Design
[Literal suffixes - design decisions](https://notes.ethereum.org/62ZGU8OHQw-KimH7inZ-4w)

## TODO
Here's a summary of what's left to do here. Overall not much concrete stuff. Just tons of small things.

### Concrete things to do
- [x] Fix failing command line tests.
- [x] Fix failing external tests.
- [x] Require functions to be declared as a suffix to be recognized as one. Syntax:
    ```solidity
    function km(uint meters) pure suffix returns (uint) {}
    ```
    `suffix` becomes a built-in modifier, usable only on file-level functions but we do not make it a keyword.
- [x] Disallow suffixes with zero or more than one return value.
- [x] Disallow suffixes with storage and calldata return values.
- [x] Disallow signed exponent for decimal suffixes.
- [x] Disallow non-pure functions as suffixes.
    - ~Or fix: `ViewPureChecker` does not complain about mutable/`view` suffixes used inside `view`/`pure` functions.~
- [x] Disallow suffix definitions with parameter types that no literal can be implicitly converted to.
- [x] Suffixed literals should be marked as function calls in the AST.
    - Use a `FunctionCall` node with `isSuffixCall` annotation for suffixed literals (but not for denominations!).
    - ~Replace `isSuffixCall` with a new `CallKind`.~ *It does not look like a good idea after all. We won't be doing it.*
- [x] Include the suffix and new annotation in the AST exporter and importer.
- [x] Allow suffix overloading.
    - [x] Disallow suffix overloading.
        - Done as a follow-up PR #14105.
- [x] Fix: ICE when an integer or decimal with zero fractional part is decomposed into mantissa and exponent.
- [x] String literal handling in the IR codegen.
- [x] Fix: Suffixed values are not disallowed as array sizes even though they're not constants.
- [x] Fix: Suffixes not detected as function calls by control flow graph (no "unreachable code" warning).
- [x] More detailed error message for exponent/mantissa out of range.
- (Optional) ~When no suffix definition found, check if there are any non-suffix definitions and list them in error message.~
- [x] Include suffix definition as a secondary location in errors.
- [x] Warning against missing suffix support in SMTChecker (if needed).
- [x] Grammar update.
- [x] Documentation.
- [x] Changelog.

### Things to check and maybe fix/implement
There are some things that might be already partially or fully working:
- [x] Research syntax variants for marking function declaration usable as a suffix.
- [x] Make sure types other than `uint` in suffixes for rational literals are allowed (both single arg and with mantissa/exponent).
- [x] Make sure our analysis (in particular the function call graph and control flow graph) follows suffixes like other function calls. Where possible, add test coverage.
- [x] Make sure that the `MemberAccess` case is always handled, add test coverage (i.e. suffixes with dots)
- [x] There are a lot of TODOs in the code that need to be reviewed

### Extra test cases
- Syntax tests
    - [x] Suffixing negative values.
    - [x] Suffix overloading.
    - [x] Range of mantissa and exponent, especially with suffixes accepting integer types other than `uint`.
    - [x] Correct parsing of suffixes in ternary operator expressions without parentheses.
    - Invalid cases
        - [x] Suffixed literals are not constants (should not be usable in constant definitions, array declarations, etc.).
        - [x] Suffixing array literals, tuple literals, structs, enums.
        - [x] Suffixing errors and events.
        - [x] Suffix returning a struct containing a mapping.
    - Overloading and shadowing
        - [x] Overloading between `suffix` and non-`suffix` functions.
        - [x] Functions shadowing suffixes and vice-versa.
- Semantic tests
    - [x] Correct return values for all kinds of literals (integers, booleans, decimals, addresses, strings, bytes).
    - [x] Type of literal different from the return type.
    - [x] Suffix use in various places (different types of calls, `new`, indexing, initialization, modifiers, `abi.encode`, etc.)
    - [x] Decomposition into mantissa and exponent.
    - [x] Checked/unchecked suffixes
    - [x] Suffix shadowing a built-in
    - [x] Suffixes making pure/view external calls
    - [x] Suffix return parameter cleanup
    - [x] Recursive suffix application
    - [x] Overloaded suffixes.
    - [x] Values at the edges of supported ranges.
    - [x] Suffixes called `szabo` and `finney` (deprecated denominations).
    - [x] Arbitrary return types.
    - [x] Calling a function pointer returned by a suffix.
- [x] SMTChecker tests


test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
```diff
@@ -0,0 +1,21 @@
+type Length is uint;
+
+function km(uint x) pure returns (Length) { return Length.wrap(x * 1000); }
+
+struct Float {
+    uint mantissa;
+    int exponent;
+}
+
+function f(uint mat, int exp) pure returns (Float memory) {
```
hrkrshnn: I wonder if this should be mantissa and exponent, or numerator and denominator.

But yeah, this example looks nice!

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
```diff
@@ -0,0 +1,21 @@
...
+  return Float(mat, exp);
+}
+contract C {
+  Length public length = 1000 km;
+  Float public factor = 1.23 f;
```
axic: So is the dots delimiting the arguments in terms of user defined types?

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
ekpyron: The problem is that for converting numerator and denominator to mantissa and exponent, in general you need a higher precision... if we had arbitrary-precision compile-time-constant expression evaluation we could easily make it numerator/denominator or simply ``rational`` - but mantissa + exponent is the significantly quicker route to get fixed point literals... (see also https://notes.ethereum.org/@solidity/HJXUx2Jkq)

chriseth: My main point is that the input is really in mantissa + exponent notation and not in numerator+denominator notation. You can apply this function only to literals (`1.34`, `-3.4e20`)  and not to rational expressions.

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
chriseth: Sorry, I don't understand your question. There is only one dot, and `1.23 f` is only related to user defined value types in that `f` returns one, but you don't need this concept for the example.

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
hrkrshnn: Not sure if we should introduce mantissa and exponent. The notation gives the impression of floating points, and not fixed. And more importantly, I don't think users are used to writing lengthy decimals: `1.15792089237316195423570985008687907853269984665640564039457584007913129639935 f` is just weird, while `(2**256 - 1) / 10**76` feels more appropriate. The former is harder to test outside of solidity as well. Most users are not used to multi precision floats, but arbitrary precision integers are lot more common.

ekpyron: Rational literals *are* floating point in that sense - they literally do have a point at floating position :-).
And ``(2**256 - 1) / 10**76`` is not a user-defined literal... that would rather fall under an either explicit or implicit conversion from a rational type (once we had one).

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
Marenz: I believe the question is whether `1` and `23` will be the arguments for `f(uint mat, int exp)`. 
If so, the example is maybe a bit confusing because `1.23` looks very much like the decimal `1.23` and not like mantissa/exponent...

chriseth: `1.23` is the same as `123e-2`. This line will result in `f(123, -2)` being called.

Marenz: Aah, is that simply the mechanism of how the suffixes work with decimals then?

chriseth: Yes.

test/libsolidity/semanticTests/literal_suffixes/intermediate.sol
hrkrshnn: @chriseth How would you write this to return a user defined value type?

Here's one way, but it needs branching.

```solidity
function f18(uint mantissa, int exponent) returns (Fixed18 memory) {
    if (exponent >= 0) {
        return Fixed18.wrap(mantissa * 10 ** exponent * 10**18);
    }
    else {
        return Fixed18.wrap(mantissa * 10**18 / 10 ** (-exponent));
    }
}
```

chriseth: As discussed on the call, the exponent is always negative (because otherwise, the integer suffix function would be called). 
Because of that, there is no need for branching.
Since signed types for `**` are unsupported, it might be better to use "mantissa + digits", i.e. negate the exponent:
```solidity
function f18(uint mantissa, int8 digits) returns (Fixed18) {
  return Fixed18.wrap(mantissa * 10 ** (18 - digits);
}
```
Even if we had a branch, this should all be fully inlined/specialized because of the constants everywhere.

leonardoalt:
Is this waiting for reviews or code changes?

chriseth:
Both.

cameel:
One thing to note about the `1 f` notation (i.e. without `_` between the literal and the suffix) is that we'd have to either add `finney` and `szabo` back to the parser or they'd be now allowed as custom literal suffixes. Someone could define them though not with the same exact semantics (i.e. no arbitrary-precision arithmetic).

chriseth:
I wrote some example code without `_` and it looked really weird, so I would say we should require the `_` both for defining the function and for using it.

cameel:
@chriseth Actually we need to talk about syntax again because there is a problem with `_`. We use it as a number separator and this makes `_` suffixes ambiguous with hexadecimals. For example:
```solidity
    0x1_000_fff
```
Is this the number `0x1000fff` or `0x1000` with a suffix called `fff`?

cameel:
EDIT: I moved the TODO list to PR description.

aarlt:
user-defined “post-fix operators” vs. user-defined literal suffixes:
https://github.com/ethereum/solidity/pull/12362#discussion_r953938583

chriseth:
I really like using `using` to tie functions to literal suffixes. In my opinion, the following problems still need to be ironed out:

1. the literal suffixes have to be usable for fractional literals. Those do not have a type. We added a mechanism to call the function with to arguments. A potential solution could be to either only use the using statement to declare a function being a literal suffix without specifying the type (`usign f as suffix;`) or have some special words to denote the different kind of literals (`using f as string suffix; using f as int suffix; using f as fractional suffix;`)
2. Suffixes usually have short names and thus easily clash with other identifiers. Because of that (and also to avoid confusion), it might be useful to disallow referencing (and thus calling) such a function in a regular context, i.e. disallow `m(1)`, but only allow `1 m`. In this example, just writing `m` would result in an "identifier not found" error. The added complication here is that if the effect of `using f as int suffix;` is that `f` is not available as stand-alone identifier, how does the reference to `f` work inside the `using` statement? Maybe we can make an exception to that rule for `using` statements....

aarlt:
> I really like using `using` to tie functions to literal suffixes. In my opinion, the following problems still need to be ironed out:
> 
> 1. the literal suffixes have to be usable for fractional literals. Those do not have a type. We added a mechanism to call the function with to arguments. A potential solution could be to either only use the using statement to declare a function being a literal suffix without specifying the type (`usign f as suffix;`) or have some special words to denote the different kind of literals (`using f as string suffix; using f as int suffix; using f as fractional suffix;`)

I like the variant with `fractional suffix`.

> 2. Suffixes usually have short names and thus easily clash with other identifiers. Because of that (and also to avoid confusion), it might be useful to disallow referencing (and thus calling) such a function in a regular context, i.e. disallow `m(1)`, but only allow `1 m`. In this example, just writing `m` would result in an "identifier not found" error. The added complication here is that if the effect of `using f as int suffix;` is that `f` is not available as stand-alone identifier, how does the reference to `f` work inside the `using` statement? Maybe we can make an exception to that rule for `using` statements....

So you would like to support functions and literal suffixes with the same name? e.g. `m(1)` is allowed, if there was a function e.g. `m(uint256)` defined. Where suffixes are only allowed to be used as suffixes `1 m`? `m(1)` will be always forbidden, if only defined as user-defined suffix (no matching function e.g. `m(uint256)`).

However, not sure  what `using` restrictions actually apply here. I don't fully understand the restriction with `f` is not available as stand-alone identifier. I think that this should at least work, when `f` was defined as a file-level function, right?

chriseth:
We always have to think about this as being used across files. One file defined the suffix, another file uses it. The using file should explicitly import it. If we just say `import m from "x.sol";`, then we should be able to reference `m` by just writing `m`. But if we allow that, it blocks any other identifier called `m` - this is what we want to avoid.

cameel:
This is now rebased on top of the operators PR (#13790).

cameel:
@aarlt I see that thanks to your refactor I can finally see the errors from the failing AST import tests here. Before it used to just exit silently after printing some dots.

matheusaaguiar:
I am still trying to understand how I managed to push the commit to the wrong branch. Sorry about that. :sweat_smile:

cameel:
Don't worry, but please restore this to the original state :)

I see you reverted the top commit - this does not remove it. You need to move the branch label locally and then push it so that it does not include your commits.

matheusaaguiar:
Should be restored to the previous state now.

cameel:
Still has one commit from your branch on top.

matheusaaguiar:
Removed the remaining commit.

test/libsolidity/semanticTests/operators/userDefined/literal_suffix_as_operator_definition.sol
```diff
@@ -0,0 +1,24 @@
+type B is bool;
+using {bitnot as ~, bitor as |} for B global;
+
+// NOTE: There are no literals of type B so these cannot be actually used on anything
+function bitnot(B x) pure suffix returns (B) {
+    return B.wrap(!B.unwrap(x));
+}
+
+function bitor(B x, B y) pure suffix returns (B) {
```
matheusaaguiar: Now that we have `suffix` specifier, we can check the parameters of the function and be more restrictive. 
I implemented that in #14015 and now this function signature generates a `TypeError` because the parameters must be of type `integer` and `unsigned integer`.

cameel: Good point. The unary one will still work, but I wanted to actually disallow any types that are not implicitly convertible from literals so it will soon stop working anyway.

We should convert this to a syntax test and update expectations.

matheusaaguiar:
For future reference, if needed:
I have been trying to isolate the problematic code in the external tests. 
I chose `ens` because it is the faster to fail. First, I tried to check if our shell script could print the name of files being compiled, but that is not possible directly. I am not sure if hardhat can do that and then instead I just tried erasing the whole `contracts` folder (created in a temp location, like `/tmp/ext-test-ENS-.../`, by our script) and starting with a minimal set of files that could compile without errors. Next, I started adding folders until I got to two interdependent ones which generate a compiler exception: `wrapper` and `ethregistrar`. These have around 30 files, to which I was going to apply the same methodology in the hopes of finding which is the problematic one and then finally be able to isolate the offending block of code. These files reference other projects, so I am not sure if the problem is really here.
Maybe there's a better, easier way to do this, but atm that's what I can do straightforwardly without having to detour and dive deeper into the tools like `hardhat`.

cameel:
@matheusaaguiar Surprisingly, removing the codegen refactor did not make the assert failure in external tests go away (unless it's just a different assert failing now). So we still need to get to the bottom of it after all.

cameel:
Also, a faster way to isolate it might be to try passing individual top-level files to `solc`. As long as you use the `--include-path` option to point at the dependencies, files should generally compile and you should be able to isolate the one that is failing. Then find the spot by cutting out imports.

matheusaaguiar:
I managed to identify what file and what code causes an assertion to fail, in the `ens` external test.
The file is `contracts/wrapper/NameWrapper.sol` and the problem seems to be with function `_canTransfer` at line `607` which overrides a `virtual` function from another contract.
The file is inside the `tmp` folder generated by our test script. Running the following command from the `contracts` folder, reproduces the problem:
`solc --base-path . --include-path ../node_modules/ ./wrapper/NameWrapper.sol --optimize --via-ir`

It makes this assertion fail: https://github.com/ethereum/solidity/blob/c6cc2e4d8198b7b1c0d8a35824eb34c4aefb56bf/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L1073

cameel:
@matheusaaguiar Thanks, that was enough to track it down!

Turns out that the problem is caused by a virtual function that gets overridden with another function that's not exactly identical (`view` rather than mutable). I've got a repro:

```solidity
abstract contract A {
    function f() internal {
        v();
    }

    function v() internal virtual;
}

contract C is A {
    function test() public {
        f();
    }

    function v() internal view override {}
}
```

My assumption was that the types must be identical and this proves it wrong. I'll need to adjust it.

But also good that now we have a test for it. Apparently external tests do that a lot and we did not have even one test case covering it.

cameel:
@ekpyron This is now ready for review.

Current status:
- I extracted small bits into separate PRs. Only the second one is a dependency here.
    - #14103
    - #14104
- There are a few smal leftovers from the checklist above that I decided to do in separate PRs:
    - The PR allows overloading. I still think we should not disallow it, but I prepared a follow-up PR that would do it: #14105.
    - @matheusaaguiar is working on the SMTChecker part (just a warning and tests, not full support).
    - Documentation.
    - Refactor to add a new `CallKind` instead of `isSuffixCall` annotation.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -697,11 +697,48 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+			{
+				for (unsigned i = 0; i < arguments.size(); ++i)
+					acceptAndConvert(*arguments[i], *parameterTypes[i]);
+			}
+			else
+			{
+				solAssert(arguments.size() == 1);
+				solAssert(arguments[0]);
+				auto const* literal = dynamic_cast<Literal const*>(arguments[0].get());
+				Type const& literalType = *literal->annotation().type;
```
nikola-matic: This should go below the `asserts`.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -697,11 +697,48 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+						// NOTE: For string literals we do not need to define the variable. The variable
+						// value will be embedded inside the conversion function.
+						m_context << literalType.literalValue(literal);
+					utils().convertType(literalType, *parameterTypes.at(0));
+				}
+				else
+				{
+					solAssert(parameterTypes.size() == 2);
+
+					auto const* rationalNumberType = dynamic_cast<RationalNumberType const*>(&literalType);
```
nikola-matic: nit: why not universal/forwarding references here, i.e. `auto&&`?

cameel:
I have just posted a summary of the current behavior of literal suffixes: https://github.com/ethereum/solidity/issues/13718#issuecomment-1503203213. I recommend reading it before reviewing.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -997,16 +997,64 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
 		solAssert(!functionType->takesArbitraryParameters());
 
 		vector<string> args;
-		if (functionType->hasBoundFirstArgument())
-			args += IRVariable(_functionCall.expression()).part("self").stackSlots();
+		FunctionDefinition const* functionDef = nullptr;
 
-		for (size_t i = 0; i < arguments.size(); ++i)
-			args += convert(*arguments[i], *parameterTypes[i]).stackSlots();
+		if (!_functionCall.isSuffixCall())
```
ekpyron: Can't the case distinction here be fully reduced to a ``if (_functionCall.isSuffixCall() && parameterTypes.size() == 2)`` and splitting the arguments in that case, while leaving the rest be handled like before?

cameel: It made sense like this while I still had the literal handling part in `visit(Literal)`, separate from handling of ordinary internal calls. You're right that now it can be simplified further

test/libsolidity/semanticTests/literalSuffixes/suffix_making_view_external_call.sol
```diff
@@ -0,0 +1,41 @@
+function suffix(int32 x) pure suffix returns (int32) {
```
ekpyron: This is a really weird test case that has zero relation to suffices...

test/libsolidity/semanticTests/literalSuffixes/suffix_in_new.sol
```diff
@@ -0,0 +1,9 @@
+function twice(uint x) pure suffix returns (uint) { return x * 2; }
+
+contract C {
+    function run() public pure returns (uint[] memory) {
+        return new uint[](5 twice);
```
ekpyron: Why are you testing this separately :-)? Do you want to write every possible program that you can write with any kind of expression and replace that expression with a suffix call for all of those cases :-)?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: Done.

cameel:
I updated #14105 to use the method suggested by @chriseth and it does work, though IMO error messages are very suboptimal. I think we'd still be better off handling that case specially and issuing a better message (which requires the `m_currentFunctionCall` hack).

It works so I added the changes from that PR here. Overloading is now disallowed.

test/libsolidity/semanticTests/literalSuffixes/suffix_making_view_external_call.sol
cameel: It's intentionally trying to do weird things, because technically they're allowed. I wanted to try it to see it it will actually work and it does not hurt to have some weirder test cases.

test/libsolidity/semanticTests/literalSuffixes/suffix_in_new.sol
cameel: Well, good point. I do have it covered by the bigger test case already so I'll remove it.

But the reason why it's here in the first place is that when I took over the PR it had numerous bugs. One of them was that suffixed literals were treated as constants and allowed e.g. in array sizes. When I plugged that hole I also added comprehensive syntax tests, trying to stick suffixes in various contexts and checking what will happen. This is one of them. This is not a function call and not a built-in. It's a construct that's probably processed specially at some point, and I don't know all the details of that so IMO it was prudent to cover it too instead of just assuming that there can be absolutely nothing wrong here.

ekpyron:
> I updated #14105 to use the method suggested by @chriseth and it does work, though IMO error messages are very suboptimal. I think we'd still be better off handling that case specially and issuing a better message (which requires the `m_currentFunctionCall` hack).
> 
> It works so I added the changes from that PR here. Overloading is now disallowed.

Never pay for minorly improved error messages by majorly increased code complexity, it's not worth it - the error messages here will be perfectly fine.

test/libsolidity/semanticTests/literalSuffixes/suffix_making_view_external_call.sol
ekpyron: The point is that this test does not test anything specific to suffices at all. It's testing something completely unrelated and then just straying in a suffix on top. I'm not sure what's the point of tests like that.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,104 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+					2699_error,
+					_function.parameterList().parameters()[0]->location(),
+					"Parameters of fixed-point types are not allowed in literal suffix functions. "
+					"To support fractional literals the suffix function must accept two integer arguments "
+					"(mantissa and exponent) that such literals can be decomposed into."
+				);
+
+			if (
+				!TypeProvider::boolean()->isImplicitlyConvertibleTo(*parameterType) &&
+				// ASSUMPTION: There are no address payable literals.
```
ekpyron, line 458: True

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,104 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+				!TypeProvider::address()->isImplicitlyConvertibleTo(*parameterType) &&
+				// ASSUMPTION: Literal 1 is implicitly convertible to any integer type.
```
ekpyron, line 460: True

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,104 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+				!TypeProvider::rationalNumber(1)->isImplicitlyConvertibleTo(*parameterType) &&
+				// ASSUMPTION: bytes1 is implicitly convertible to any fixed-bytes type.
```
ekpyron, line 462: True

libsolidity/codegen/ExpressionCompiler.cpp
nikola-matic: Too late I guess; looks like I forgot to click on submit.

test/libsolidity/semanticTests/literalSuffixes/suffix_making_view_external_call.sol
cameel: The point is an external call to a `view` function. That call happens directly inside the suffix definition so it's relevant to the feature. You can assume that it should work, but it's still a valid black-box test. The fact that suffixes are no different from functions in that respect is just an assumption and something that should theoretically be tested. Of course, I'm not doing extensive black-box testing here because it would make the test surface unreasonable, but I don't think that including some tests of this kind is a bad idea.

But well, I'll remove it if it bothers you. IMO it's nice to have but it's not like it's essential here or anything. The codegen changes are straightforward enough that I think there's little doubt that suffixes are indeed properly implemented as functions.

libsolidity/codegen/ExpressionCompiler.cpp
cameel: Yeah, I noticed that an fixed that already. But thanks for pointing it out anyway.

libsolidity/codegen/ExpressionCompiler.cpp
cameel: Well, I never really thought about this, and our convention is to explicitly show that it's `const` and a pointer when using `auto`. But I agree that `auto&&` seems very reasonable in a case like this. After all the whole type is visible in `dynamic_cast`, and the point of having `auto` here was not to repeat it.

I replaced all such cases in the PR with `auto&&` now.

cameel:
I dealt with all the comments and also added some minor tweaks (mostly tests, but also things like mentioning location in error message for reference types).

Also note that the docs are ready for review in #14121.

EDIT: SMTChecker support is done in #14118 and also can be merged into this PR when approved.

Also is one of the refactors from the parent issue is in #14124, but note that it's independent of suffixes and the suffix PR can be merged without it.

In general, you can safely focus only on [suffix PRs with the `roadmap` label](https://github.com/ethereum/solidity/pulls?q=is%3Apr+is%3Aopen+label%3Aroadmap+suffix).

test/libsolidity/semanticTests/literalSuffixes/suffix_making_view_external_call.sol
ekpyron: So since ``suffix`` functions can happen in any kind of function that is part of an inheritance tree, we now need a test for any inheritance tree in which a suffix function can occur :-)?
This is **not** related to the feature. If you apply a definition of "relevant to the feature" that is broad enough to include this, then you need to test all possible strings of programs for any change we do :-).

This test in particular didn't bother me that much, but I'm generally bothered by having several thousand lines of mostly unrelated tests ;-). But yeah, it's fine for now - but please try to add *actually relevant* tests only in the future :-).

docs/grammar/SolidityParser.g4
```diff
@@ -182,7 +182,8 @@ locals[
  freeFunctionDefinition:
  	Function (identifier | Fallback | Receive)
  	LParen (arguments=parameterList)? RParen
- 	stateMutability?
+ 	stateMutability? Suffix?
+ 	| Suffix? stateMutability?
```
ekpyron: Does this really do what it's supposed to?
I guess this is meant to read "either ``stateMutability Suffix`` or ``Suffix stateMutability`` while both are optional...
But that's not what I think this says... I think it says:
- An optional stateMutability. Followed by
- Either an optional suffix or an optional suffix. Followed by
- Another optional state mutability.

So this should for example now (incorrectly) accept ``function f() pure suffix pure``...

ekpyron: The proper way to do this is probably in the same manner as this is handled in ``contractFunctionDefinition``.

ekpyron: Maybe the nicest way overall to handle this would have been *not* to split contract functions from free functions, but rather just make "being free" a flag like ``elementaryTypeName`` has one...

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
```diff
@@ -0,0 +1,21 @@
+function uSuffix(uint8) pure suffix returns (uint) {}
+function uSuffix(uint16) pure suffix returns (int) {}
+
+function iSuffix(int8) pure suffix returns (uint) {}
```
aarlt: hmm.. maybe I misunderstood somehow the documentation.. but with respect to ["There are no negative number literals in Solidity." ](https://github.com/ethereum/solidity/pull/14121/files#diff-85812ff9484018acfc04b0dc727b772bded418bb644ba7e1286b59e20d481481R146). Wouldn't it make sense to only support unsigned integers?

docs/grammar/SolidityParser.g4
ekpyron: Not that this is utterly important to fix, but the way it is right now is weird :-).

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
```diff
@@ -0,0 +1,21 @@
...
+
+function iuSuffix(uint8) pure suffix returns (int) {}
+function iuSuffix(int8) pure suffix returns (uint) {}
+
+contract C {
+    int a = 1024 uSuffix;
+    int b = 1024 iSuffix;
+
+    int c = -1024 uSuffix;
+    int d = -1024 iSuffix;
```
aarlt: here we will always receive a positive 1024 inside the `iSuffix` function, right? Maybe there is never a use-case that we will be able to receive an actual negative int?

[``-123 suffix`` is equivalent to ``-(123 suffix)``, so ``suffix`` does not receive ``-123`` as input](https://github.com/ethereum/solidity/pull/14121/files#diff-85812ff9484018acfc04b0dc727b772bded418bb644ba7e1286b59e20d481481R148)

aarlt: ok, maybe if the suffix function is called directly e.g. `iSuffix(-1024)` but normally it will not really happen, right?

aarlt: if we disallow signed int parameters. a user may check the corresponding section in the docs.. also we could generate a nice error, explaining this.

docs/grammar/SolidityParser.g4
cameel: > So this should for example now (incorrectly) accept `function f() pure suffix pure`...

Ah, it's missing parentheses.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,104 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+	{
+		if (_function.stateMutability() != StateMutability::Pure)
+			m_errorReporter.typeError(
+				1716_error,
+				_function.location(),
+				"Only pure functions can be used as literal suffixes"
+			);
+
+		optional<string> parameterCountMessage;
+		if (_function.parameterList().parameters().size() == 0)
```
ekpyron: This would go much smoother if you had ``switch``ed over the number of parameters :-). (And ditched the `optional`. And just used a more generic error message... that the function has to take one or two arguments is by far good enough - there's really no need to distinguish between "none" and "three or more" - nobody's helped by that :-).

docs/grammar/SolidityParser.g4
cameel: No, wait. I just noticed the extra `?`. It was meant to be:

```
stateMutability Suffix?
| Suffix stateMutability?
```

Looks like my suggestion was not applied correctly in https://github.com/ethereum/solidity/pull/14054#discussion_r1148023318 and our tests did not catch it.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,104 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+
+		solAssert(_function.returnParameterList());
+		if (_function.returnParameterList()->parameters().size() != 1)
+		{
+			m_errorReporter.typeError(
+				7848_error,
+				_function.returnParameterList()->location(),
+				"Literal suffix functions must return exactly one value."
+			);
+		}
```
ekpyron:
```suggestion
			m_errorReporter.typeError(
				7848_error,
				_function.returnParameterList()->location(),
				"Literal suffix functions must return exactly one value."
			);
```
If I'm allowed to be nitpicky as well this time ;-).

docs/grammar/SolidityParser.g4
cameel: > Maybe the nicest way overall to handle this would have been *not* to split contract functions from free functions, but rather just make "being free" a flag like ``elementaryTypeName`` has one...

The reason to split it out was that contract functions allow arbitrarily named modifiers and those can be repeated while `suffix` on free functions cannot be repeated.

I guess it's possible to resolve it with a flag, but given how much the modifier lists on those differ, I thought it was cleaner to separate the definitions. And it actually made the grammar more precise. Stuff like `virtual` or `override` is not allowed on free function after all.

libsolidity/analysis/TypeChecker.cpp
cameel: Nitpick as much as you want. This kind of stuff is trivial to fix while I'm applying other suggestions anyway.

libsolidity/ast/Types.cpp
```diff
@@ -314,7 +314,7 @@ char const* Type::categoryName(Type::Category _category)
 	case Category::UserDefinedValueType: return "user-defined value type";
 	case Category::Tuple: return "tuple";
 	case Category::Mapping: return "mapping";
-	case Category::TypeType: return "type of a type";
+	case Category::TypeType: return "type";
```
ekpyron: This is incorrect. Why was this changed?

ekpyron (changes requested):

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
cameel: Yes, we'll always receive 1024, but this is not something we should be assuming in the PR. We should also test it.

docs/grammar/SolidityParser.g4
ekpyron:
```
stateMutability Suffix?
| Suffix stateMutability?
```

means you *always* have a state mutability - and then either an optional suffix or a non-optional suffix - and then another optional state mutability... also doesn't exactly ring right to me ;-).

libsolidity/ast/Types.cpp
aarlt: is `TypeType` not the type of a type anymore? Isn't there a difference between a type and the type of a type?

cameel: Oh, is it? AFAIK `TypeType` is stuff like a struct type, contract type, maybe even built-in types, like `uint`, right? So a type in the language. `Category::Contract` or `Category::Struct` are used for instances of those types.

I got that wrong in #14087 so I'm fixing it but this did not seem important enough on its own to warrant a separate PR.

docs/grammar/SolidityParser.g4
ekpyron: Also: you're aware that these grammars can be used to generate parsers? And case distinctions will be parsed as separate paths in those? A distinction between ``stateMutability Suffix?`` ``Suffix stateMutability?`` is rather weird... as in: the state mutability is parsed as a different field depending on whether there's a suffix in front of it or not...

libsolidity/ast/Types.cpp
cameel: See https://github.com/ethereum/solidity/pull/12656#discussion_r1169057725

I think that, e.g., if you have something like

```solidity
struct S {
    uint x;
}

function f(S s) {
    ....
}
```

And you use `S` in an expression, its AST node is `Category::TypeType`. So it refers to a type (a struct type).

`s` in an expression would be `Category::Struct`. It refers to a value of a struct type.

docs/grammar/SolidityParser.g4
ekpyron: > And it actually made the grammar more precise. Stuff like virtual or override is not allowed on free function after all.

Which is *not* a parser error... hence, technically, this makes the grammar *less* precise ;-).
I.e. the grammar of the actual language *does* allow `virtual` or `override` on free functions.
It's merely an analysis error... do you want to reimplement e.g. the type checker in antlr to cover all analysis errors :-)?
The clear line here should be: whatever passes our parser should pass the antlr grammar and vice versa.

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
aarlt: why not forbid signed integers for suffix functions in general?

aarlt: is there any case where someone could use the suffix functions really as a suffix and receive a negative number there?

docs/grammar/SolidityParser.g4
cameel: > means you always have a state mutability - and then either an optional suffix or a non-optional suffix - and then another optional state mutability... also doesn't exactly ring right to me ;-).

ok, you got me here. Without seeing the values of `stateMutability` It completely slipped my mind that *mutability not set* is one of the allowed cases here.

> Also: you're aware that these grammars can be used to generate parsers?

Yes.

> And case distinctions will be parsed as separate paths in those? A distinction between `stateMutability Suffix?` `Suffix stateMutability?` is rather weird... as in: the state mutability is parsed as a different field depending on whether there's a suffix in front of it or not...

I thought they'd get unified. But now that I think of it, you could use `stateMutability` multiple times so ANTLR must have a way to distinguish those.

ok, I guess this was a bad suggestion. I need to use flags after all.

libsolidity/ast/Types.cpp
ekpyron: No, it's not - how did you get that idea? All of those have their respective category. ``TypeType``, is the type of a type, as in ``type(SomeType)`` - which is considered as the type of ``SomeType``, i.e. the type of the type or TypeType.

I already wasn't happy about that other PR, since it has rather off nomenclature, but this is definitely, plainly incorrect.

ekpyron: No, it doesn't, it's the type of the struct type... ``S`` is not a struct, it's the type of a struct...

ekpyron: These are *type* categories, the categories of *types* of expressions, not categories of expressions themselves - you're thinking on the wrong layer here :-).
So this is definitely still the type of a type and **not** a type. Why would it be called ``TypeType`` otherwise :-)?

ekpyron: https://github.com/ethereum/solidity/pull/14087 already conceptualized this weirdly and inconsistently with anywhere else in the compiler.

aarlt: not sure, but I just write it here, because I was thinking about that after reading the comments.. isn't the type of a type a meta-type? maybe meta-type is a better name for `TypeType`? or is there any difference between a TypeType and a meta-type?

ekpyron: The problem is rather that #14087 has types of expressions print their type categories for the purpose of printing categories of expressions of that type, which makes this horribly confusing and which we have to change, otherwise anybody reading this will get the wrong idea.

docs/grammar/SolidityParser.g4
cameel: Grammar should be fixed now.

> Which is not a parser error... hence, technically, this makes the grammar less precise ;-).

You're right. I did not consider that.

BTW: it's not the only thing that is mismatched that way. Just off the top of my head, unary + is not allowed in the grammar. I suspect there's much more of that. We'd need to have tests against it to have any hope of keeping it consistent.

ekpyron: There's little point in going out of our way to keep this a precise match. It just has to be "close enough" and be intrinsically a grammar that makes sense. There's no point in making the grammar accept always invalid constructs just for the sake of technicalities (and also the other way round). That's why I said *technically* :-).

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
cameel: See my comment on this in the docs PR. Overall this was a design decision with some reasons behind it and if you want it changed, it's probably better to bring it up on the operators channel because here the discussion is getting scattered in comments on multiple PRs and hard to follow.

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_integer_vs_integer.sol
cameel: See https://github.com/ethereum/solidity/pull/12656#discussion_r1169010129.

> is there any case where someone could use the suffix functions really as a suffix and receive a negative number there?

Yes, it's perfectly valid to return a signed value that can be negated. Or a value of an UDVT type that supports `-` operator.

libsolidity/analysis/TypeChecker.cpp
cameel: Done. Using `switch` is actually a nice idea. Thanks for the suggestion.

libsolidity/ast/Types.cpp
ekpyron: > And you use S in an expression, its AST node is Category::TypeType. So it refers to a type (a struct type).

To try to clarify this: this is incorrect - the AST node is an expression :-). The expression *is* itself a type, but that's not the category of the type of the expression - the type of this expression is the type of a type (since the expression is a type). The translation of categories of types to categories of expression was abusive in #14087 already.

ekpyron: But yeah, this is primarily a distraction on this PR - I'd suggest to just remove the change here and then we can argue about this separately. But I'd say we should just revert #14087 and solve this that way, everything else is too much time spent on this.

docs/grammar/SolidityParser.g4
```diff
@@ -172,19 +172,27 @@ locals[
...
+locals[
+	boolean mutabilitySet = false,
+	boolean suffixSet = false,
+]
+:
+	Function (identifier | Fallback | Receive)
+	LParen (arguments=parameterList)? RParen
+	(
+		{!$mutabilitySet}? stateMutability {$mutabilitySet = true;}
+		| {!$suffixSet}? visibility {$suffixSet = true;}
```
ekpyron: visibility?

Also I'm not entirely convinced that in this version it's really that justified to split free functions out from contract function definitions... I'd prefer to have one function definition node that can be free or not and just restrict modifiers and suffix accordingly - otherwise this has a lot of duplication really and does not actually match our AST structure (which again we don't have to be *draconian* in following, but if possible, it's preferable)

libsolidity/ast/Types.cpp
cameel: > No, it's not - how did you get that idea? 

Oh, wow, you're right. I'm wondering this myself because now it does seem obviously wrong in retrospect :) I got this horribly mixed up.

I think it actually comes from the fact that originally the error message was `Type is not callable` when you tried to call something non-callable, like a variable:

https://github.com/ethereum/solidity/blob/7dd6d404815651b2341ecae220709a88aaed4038/test/libsolidity/syntaxTests/scoping/state_variable_function_conflict_former_crash.sol#L9
https://github.com/ethereum/solidity/blob/3f04f00bdc705016ba660ec176bb3239bc32b072/test/libsolidity/syntaxTests/functionCalls/mapping_not_callable.sol#L3

which confused me a lot at first and somewhat warped my original (correct) assumptions about how this works. This was reinforced by the fact that when you try to call a type, you [enter the `TypeType` branch in `visit(FunctionCall)`](https://github.com/ethereum/solidity/blob/a019f400d080aeab3fd02fd4e7b699e965a893f6/libsolidity/analysis/TypeChecker.cpp#L2815-L2847). I now see that the `switch` is dealing with the type of that thing and not the thing itself, which I must have missed originally.

> https://github.com/ethereum/solidity/pull/14087 already conceptualized this weirdly and inconsistently with anywhere else in the compiler.

I finally see why you did not like that change. The descriptions should all be types, while I described them mostly as instances of those types. That's kinda another factor why `type of a type` was bothering me here. That one does not describe an instance, so while it is the wrong way to describe the category, it would fit the error message better, which was the viewpoint I was judging it from.

In any case, I think that the way to fix it without messing with the category stuff any further would be to change the message to `Expression is not callable.`.

I removed the change from this PR.

cameel: > These are type categories, the categories of types of expressions, not categories of expressions themselves - you're thinking on the wrong layer here :-).

You're right. I got that wrong. I was thinking about this:

```solidity
1;          // RationalNumberType, Category::RationalNumber
x;          // IntegerType,        Category::Integer
uint;       // TypeType,           Category::TypeType
type(uint); // MagicType,          Category::Magic

s;          // StructType,         Category::Struct
S;          // TypeType,           Category::TypeType

c;          // ContractType,       Category::Contract
C;          // TypeType,           Category::TypeType
type(C);    // MagicType,          Category::Magic
```

but these indeed describe the type of these things and not the things themselves.

See also https://github.com/ethereum/solidity/pull/12656#discussion_r1170020290.

docs/grammar/SolidityParser.g4
cameel: Damn :) But yeah tests caught this as expected. I just did not get around to look at them yet.

> Also I'm not entirely convinced that in this version it's really that justified to split it out from contract function definitions...

You had a lot of good points about it earlier (i.e. solc parser does not distinguish them and I ended up having to switch back to the flags anyway) so I agree. It ended up being more duplication than I thought it would be originally.

I can merge them back. Do you prefer it here or in a separate PR?

cameel: Ah, wait. The original reason for the split was not a cosmetic refactor, but the issue with modifiers being repeatable but `suffix` not. I can merge it but I guess then we'll have two levels of flags? Won't that be more complicated?

docs/grammar/SolidityParser.g4
cameel: That's also why I also did not jump immediately at changing it back :)

libsolidity/ast/Types.cpp
cameel: > But I'd say we should just revert https://github.com/ethereum/solidity/pull/14087 and solve this that way, everything else is too much time spent on this.

I will do that (separately). I now see what you meant earlier. I'll drop the `categoryName()` function.

docs/grammar/SolidityParser.g4
cameel: > I'd prefer to have one function definition node that can be free or not and just restrict modifiers and suffix accordingly

I tried to do that just now, but I suspect it may and up being pretty complex, unless ANTLR has some special feature for cases like this.

Basically, the issue is that you have to know whether you're dealing with a contract function or a free function to say if something like `suffix suffix` is allowed on the modifier list. Separating them into two definitions achieves that pretty nicely. If they're combined, I need to have a way to tell which one I'm dealing with:
- If rules could be parameterized, I would make that a parameter, but AFAIK ANTLR does not support that.
- I could have a variable set before entering the rule and cleared after exiting. But a stateful rule does not seem like a good thing (if it's even possible in the first place, which I suspect it's not).
- I could make it conditional on the presence of a visibility modifier, because that's the only thing that is always present on contract functions and never present on free functions. But since modifiers can be in any order, it seems like achieving this may require an overcomplicated rule.

Do you have an idea on how to do it cleanly?

ekpyron: Look at ``elementaryTypeName`` - it should be possible to handle "being a free function" similarly to "allowAddressPayable" for that one. So that's pretty much exactly what you need here - and basically means splitting the rule into a free function and a contract function version, just without duplicating it :-). I mentioned that in some other comment, but not sure which one, must have gotten lost there :-).

ekpyron: I.e. you can just make modifiers, resp. suffices (that is whatever differs) conditional on that additional flag of being free and share the rest - and on use of functions just specify whether they're free or not (like e.g. in ``elementaryTypeName`` occurring in ``typeName`` vs. occurring in ``expression``)

cameel: So that's option one. Looks like it's possible after all. Yeah, that's the best option.

I think I saw you mentioning a solution for this before, but it wasn't where I thought it was when I tried to find it later. When I looked for it I saw answers that ANTLR does not allow parametric rules.

ekpyron: Ah, yeah, nice - you agree that this is much nicer and less duplicated than the split now, right :-)? And also less of a relevant change for anyone actually using the grammar (I'd still be curious if anybody actually properly does or whether it's just people merely experimenting a bit with it here and there). And also closer to how our parser and our AST work. Looks good now in any case.

libsolidity/ast/Types.cpp
ekpyron: Yeah, the cause for the confusion may be that for brevity ``Type::Category::*`` does not repeat ``Type`` at the end explicitly in all cases, since given the scope it's implicit - i.e. ``Type::Category::Integer`` is the category of *types* that are integer *types* - and only ``TypeType`` spells that out again since ``Category::Type`` would more likely give the wrong impression :-). So it's not exactly perfect, but now that this is cleared up, we don't have to do anything about it other than https://github.com/ethereum/solidity/pull/12656#discussion_r1170072290

We'll need to rethink all this a bit in our upcoming new type system anyways.

docs/grammar/SolidityParser.g4
cameel: Yes, I agree with the knowledge I have now. I just initially thought it was not possible and earlier I missed some nuances which led me to think that a split matched our parser better.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -426,6 +429,110 @@ bool TypeChecker::visit(FunctionDefinition const& _function)
...
+					"implicitly convertible to its parameter type."
+				);
+
+			break;
+		}
+
+		case 2:
+		{
+			auto&& mantissaType = dynamic_cast<IntegerType const*>(_function.parameterList().parameters()[0]->type());
+			auto&& exponentType = dynamic_cast<IntegerType const*>(_function.parameterList().parameters()[1]->type());
```
ekpyron:
```suggestion
			auto const* mantissaType = dynamic_cast<IntegerType const*>(_function.parameterList().parameters()[0]->type());
			auto const* exponentType = dynamic_cast<IntegerType const*>(_function.parameterList().parameters()[1]->type());
```
Optional nitpick - but yeah, while I'm a fan of ``auto`` in general and of ``auto&&`` in particular, I actually find it easier on the eye here to make it clear it's a ``const*`` :-).

docs/grammar/SolidityParser.g4
```diff
@@ -85,7 +85,7 @@ inheritanceSpecifier: name=identifierPath arguments=callArgumentList?;
  */
 contractBodyElement:
 	constructorDefinition
-	| contractFunctionDefinition
+	| functionDefinition[false]
```
aarlt, line 88: oh interesting.. looks like that I did not look at the grammar for a while now.. I didn't knew that this is possible.. what is the `[false]` and `[true]` doing? why is it needed?

docs/grammar/SolidityParser.g4
```diff
@@ -156,10 +156,11 @@ overrideSpecifier: Override (LParen overrides+=identifierPath (Comma overrides+=
  * Depending on the context in which the function is defined, further restrictions may apply,
  * e.g. functions in interfaces have to be unimplemented, i.e. may not contain a body block.
  */
-contractFunctionDefinition
+functionDefinition[boolean free]
 locals[
 	boolean visibilitySet = false,
 	boolean mutabilitySet = false,
+	boolean suffixSet = false,
```
aarlt, line 163: I'm just wondering now, why we don't use the `free` parameter.. so this seem to be creating the same `locals` for free function definitions and not-free functions

libsolidity/analysis/TypeChecker.cpp
cameel: Haha, for me this is actually one of the few cases where omitting the `const*` is an improvement because we do have the type in the expression :)

Anyway, I changed it because it was suggested by @nikola-matic in another PR. I can change it back but I wonder what's his opinion here.

docs/grammar/SolidityParser.g4
```diff
@@ -169,23 +170,14 @@ locals[
 	(
 		{!$visibilitySet}? visibility {$visibilitySet = true;}
 		| {!$mutabilitySet}? stateMutability {$mutabilitySet = true;}
-		| modifierInvocation
+		| {!$free}? modifierInvocation
```
aarlt, line 173: ah ok.. I see.. so the `free` parameter is used here.. ah and based on this parameter you basically define the needed tokens?

libsolidity/ast/AST.h
```diff
@@ -1019,6 +1024,7 @@ class FunctionDefinition: public CallableDeclaration, public StructurallyDocumen
 	StateMutability m_stateMutability;
 	bool m_free;
 	Token const m_kind;
+	bool m_usableAsSuffix;
```
ekpyron: I know that this is fine and that there's also ``m_free`` above and everything... still I'm also getting itchy when I see POD-members that aren't directly member-initialized :-).
I'd *always* do a
```suggestion
	bool m_usableAsSuffix = false;
```
for *all* cases like this - if this is actually assigned to in the constructor this has no effect whatsoever (i.e. also zero penalty) - but in case this actually *isn't* initialized in a (potentially future) constructor, this will change undefined behaviour into deterministic behaviour...

But yeah - just wanted to remark this generally, both changing it here and keeping it is ultimately fine with me (here and in all other cases like this)

docs/grammar/SolidityParser.g4
cameel: I didn't know either. It basically works like a parameter. See how `functionDefinition` is defined. It has a boolean variable that I can use to choose a branch depending on whether I'm dealing with a free function or not.

docs/grammar/SolidityParser.g4
```diff
@@ -169,23 +170,14 @@ locals[
...
+		| {$free}? {!$suffixSet}? Suffix {$suffixSet = true;}
```
aarlt, line 174: so here it means: if its a free function, where basically the suffix was not yet parsed, but a `Suffix` was parsable after that function, then we just store that we parsed a suffix, right?

docs/grammar/SolidityParser.g4
aarlt: I didn't checked the semantic yet.. but shouldn't it also work with only `{!$free}` - because we always have in this context the parameter defined - e.g. its always `true` or `false`?

docs/grammar/SolidityParser.g4
cameel: Yeah.

docs/grammar/SolidityParser.g4
cameel: Yes.

docs/grammar/SolidityParser.g4
cameel: Because having all of these locals defined always does not hurt even if not all of them are used in call cases. And, honestly, I'm not even sure if they can be declared conditionally. But even if they can, I think it's better as is.

libsolidity/ast/Types.cpp
```diff
@@ -1264,6 +1264,42 @@ FixedPointType const* RationalNumberType::fixedPointType() const
...
+		return {nullptr, TypeProvider::rationalNumber(0)};
+
+	while (unsignedMantissa.denominator() != 1)
+	{
+		unsignedMantissa *= 10;
+		++negatedExponent;
+
+		// NOTE: Technically RationalNumberType::isValidLiteral() should not allow an exponent so
+		// large we cannot store it in 4096 bits. For some reason it does not validate numbers not
+		// in scientific notation though. See https://github.com/ethereum/solidity/issues/14100
```
ekpyron: I don't see how any of this is relevant here.
I.e. the 4096 bits bounds are not relevant for the checks here at all.
So I'd suggest to just remove the comment entirely.

libsolidity/analysis/TypeChecker.cpp
ekpyron: Our general policy on ``auto`` was to make references and pointers explicit.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -697,11 +697,34 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 			// Callee removes them and pushes return values
 
 			evmasm::AssemblyItem returnLabel = m_context.pushNewTag();
-			for (unsigned i = 0; i < arguments.size(); ++i)
-				acceptAndConvert(*arguments[i], *function.parameterTypes()[i]);
+
+			if (!_functionCall.isSuffixCall() || parameterTypes.size() == 1)
```
ekpyron: The negated condition ready much easier and you can just flip the blocks below...

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -997,16 +997,49 @@ void IRGeneratorForStatements::endVisit(FunctionCall const& _functionCall)
...
+			auto&& [mantissa, exponent] = literalRationalType->fractionalDecomposition();
+			solAssert(mantissa && exponent);
+
+			IRVariable mantissaVar(m_context.newYulVariable(), *mantissa);
+			define(mantissaVar) << toCompactHexWithPrefix(mantissa->literalValue(literal)) << "\n";
+			args += convert(mantissaVar, *parameterTypes[0]).stackSlots();
+
+			IRVariable exponentVar(m_context.newYulVariable(), *exponent);
+			define(exponentVar) << toCompactHexWithPrefix(exponent->literalValue(literal)) << "\n";
+			args += convert(exponentVar, *parameterTypes[1]).stackSlots();
```
ekpyron: Couldn't this just simply be
```
args.emplace_back(toCompactHexWithPrefix(mantissa->literalValue(literal)));
args.emplace_back(toCompactHexWithPrefix(exponent->literalValue(literal)));
```
?

libsolidity/ast/AST.h
```diff
@@ -2182,6 +2195,7 @@ class FunctionCall: public Expression
 	std::vector<ASTPointer<Expression>> m_arguments;
 	std::vector<ASTPointer<ASTString>> m_names;
 	std::vector<SourceLocation> m_nameLocations;
+	bool m_isSuffixCall;
```
aarlt: similar as above.. as @ekpyron wrote, it cannot hurt to always initialize PODs

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
```diff
@@ -0,0 +1,31 @@
...
+        1.0e78 u256;
+
+        115792089237316195423570985008687907853269984665640564039457584007913129639936 u8;   // 2**256
+        115792089237316195423570985008687907853269984665640564039457584007913129639936 u16;  // 2**256
+        115792089237316195423570985008687907853269984665640564039457584007913129639936 u256; // 2**256
+    }
+
+    function max() public pure {
+        1e-256 u8;
+        128e-256 mi8eu8;
```
aarlt, line 20: just wondering why you don't test for `1e-1234 u16;` too

libsolidity/analysis/TypeChecker.cpp
ekpyron: Even if you say ``const*`` is redundant here, then so is ``&&``, so then you can just make it a plain ``auto``. `auto&&` in cases in which you obviously don't deal with any kind of reference is rather weird.

ekpyron: So just change it here and anywhere else to ``auto const*``, it's significantly more readable.

cameel: Change reverted.

docs/grammar/SolidityParser.g4
cameel: > but shouldn't it also work with only `{!$free}`

Not sure what you mean. It does work with only `{!$free}`. There's no other variable used here.

libsolidity/ast/AST.h
cameel: Added default initialization for all members.

libsolidity/codegen/ExpressionCompiler.cpp
cameel: Negated.

libsolidity/ast/Types.cpp
cameel: It answers the question that naturally comes up here: is this validation really needed or should it be an assert? I.e. can the exponent actually be so large that it does not fit in `uint`? When I thought about it, I remembered both cases where I could make very it large and cases where compiler complained that it's too large. It seemed like a contradiction that I could not reconcile. Turned out that our validations are inconsistent and only some forms of literals have limits while other, seemingly equivalent, literals do not. It still baffles me that we're fine having something like this in the language.

In any case, I updated the comment to explain it in a different way, hopefully clearer.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
cameel: Good suggestion. Looks like it can. We validate that integers are in the range matching the type so we should be fine without any runtime conversion there.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
cameel: I do, but `1234` is not out of range of a `uint16` exponent so that case is in `exponent_in_range.sol`.

libsolidity/ast/AST.h
cameel: Added default initialization for all members.

libsolidity/ast/Types.cpp
ekpyron: We don't have concrete sharp limits on the language - we prevent expressions that will have small input sizes resulting in very large numbers which can be used to resource-exhaust the compiler quickly and enforce arbitrary limits in some cases for which this is applicable. The limits in those cases are chosen to be larger-than-practical. There is literally no use-case for rational literals that are more than slightly larger than 256 bits (i.e. 512 bits max), since those values cannot be processed in the language anyways.
Working on this would mean spending effort on technicalities that have zero real-world impact, so it's the very definition of wasting time.

libsolidity/ast/Types.cpp
```diff
@@ -1264,6 +1264,42 @@ FixedPointType const* RationalNumberType::fixedPointType() const
...
+		return {nullptr, TypeProvider::rationalNumber(0)};
+
+	while (unsignedMantissa.denominator() != 1)
+	{
+		unsignedMantissa *= 10;
+		++negatedExponent;
+
+		// NOTE: The exponent calculated here can be arbitrarily large because we do not limit precision
+		// of rational literals. While the language does not allow very large exponents in scientific
+		// notation, the number of significant digits in the literal itself is unlimited.
```
ekpyron: Please just remove this comment. There is zero question about why we impose limits here - those limits have no connection whatsoever on the completely unrelated limits you're referring to. This comment is still more confusing than helpful - without it, the code is perfectly clear and fine.

docs/grammar/SolidityParser.g4
ekpyron: The question is whether it could be ``{!$free} modifierInvocation`` instead of ``{!$free}? modifierInvocation``, interpreting the ``?`` as ``$free`` may or may not have a value yet, if I understand it correctly. If I recall correctly (I haven't looked into the details of antlr syntax myself for a while), that's not how this ``?`` is read and it's rather *always* part of conditional rules, i.e. it's always ``{...}? ...`` if the rule is dependent on some boolean condition.

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_decimal_vs_decimal.sol
```diff
@@ -0,0 +1,21 @@
...
+    int b = 1.024 iSuffix;
+
+    int c = -1.024 uSuffix;
+    int d = -1.024 iSuffix;
+
+    int e = 2.55 iuSuffix;
+    int f = -2.55 iuSuffix;
+}
+// ----
+// TypeError 2144: (392-399): No matching declaration found after variable lookup.
```
ekpyron: It's impossible to judge which statement is triggering this error. Should all cases here be invalid, but you're only actually testing one?

ekpyron: There's similar issue with quite a number of other tests when I'm skimming through them.

libsolidity/ast/Types.cpp
cameel: The fact that we're even having this discussion and you need to explain this to me means that at the very least this is not properly documented/commented in the source and is a time waster for anyone who encounters this weird behavior.

But anyway, I don't want to continue this discussion so I removed the comment.

libsolidity/ast/Types.cpp
cameel: I strongly disagree but I don't want to continue this discussion so I removed it.

test/libsolidity/syntaxTests/literalSuffixes/parameters/overloading_decimal_vs_decimal.sol
cameel: It's because they were all valid back when overloading was allowed.

ekpyron: Yes, I know, but do we really want to keep them like that?

cameel: I'll see if I can improve it. The fact that a fatal error stops processing and we don't have a mode where we can let it continue is kinda annoying when you want to test multiple things will less boilerplate, but well...

cameel: I removed lines that are unreachable after a fatal error from some test cases and split some others. I hope it's better now.

Changelog.md
```diff
@@ -23,6 +24,8 @@ Bugfixes:
 
 
 AST Changes:
+ * AST: Add ``isSuffixCall`` field to ``FunctionCall`` AST node, which is ``true`` for calls resulting from applying a literal suffix. Additionally, the ``arguments`` annotation is omitted from such notes.
```
ekpyron: The latter is not a visible AST change, is it? I.e.
```suggestion
 * AST: Add ``isSuffixCall`` field to ``FunctionCall`` AST node, which is ``true`` for calls resulting from applying a literal suffix.
```

cameel: Let me check.

You reminded me that I actually wanted to check if by default it's `null` or an empty list (the former would make it not show up by default). If it's not in the AST at all then that's moot too I guess.

cameel: No, `arguments` is included in the AST for `FunctionCall` nodes: https://github.com/ethereum/solidity/blob/4a8d6618f5b398077f694835905f93d0a3890289/libsolidity/ast/ASTJsonExporter.cpp#L884

Unless you mean that despite this we don't consider it an AST change?

Also, weirdly, I just noticed that my ASTJSON test with a suffix call does have that field in its `FunctionCall` node. Need to check that. This should not be happening unless I'm misunderstanding something here.

ekpyron: That's not the argument types in the expression annotation.
We couldn't skip the actual link to the argument nodes - that would make up a broken AST.

ekpyron: I.e. note that this is not ``make_pair("arguments", toJson(_node.annotation().arguments)),``

ekpyron: So the AST and the fact that this still shows up is perfectly fine - just the changelog entry is wrong :-).

cameel: > I.e. note that this is not `make_pair("arguments", toJson(_node.annotation().arguments)),`

Yeah, just figured that out after stepping though the code in the debugger. It was in my blind spot because I didn't expect there to be two different things called "arguments". Turns out is that we have both an annotation called `arguments` and `FunctionCall::arguments()` (which is not a getter for the annotation). 

Especially the fact the annotation is not exported is a total surprise to me, because I assumed that we pretty much always export annotations (at some point I even used to think it happens automatically).

There are so many slightly different things that all could be referred to as "arguments" in our code that we'd really be better off having more precise names for them. E.g. `ExpressionAnnotation::argumentNamesAndTypes` and `FunctionCall::argumentExpressions()`.

In any case, you're right about the changelog. I removed that bit. Thanks for noticing it.

ekpyron: We definitely don't generally export annotations - rather the opposite, the base AST is what we *always* export. Annotations we only export if they provide valuable additional information for tooling - while the expression arguments annotation (and a lot of the annotations in general) are just an internal convenience store of information that is already otherwise present in the AST in all cases.

libsolidity/codegen/ExpressionCompiler.cpp
```diff
@@ -697,11 +697,34 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 			// Callee removes them and pushes return values
 
 			evmasm::AssemblyItem returnLabel = m_context.pushNewTag();
-			for (unsigned i = 0; i < arguments.size(); ++i)
-				acceptAndConvert(*arguments[i], *function.parameterTypes()[i]);
+
+			if (_functionCall.isSuffixCall() && parameterTypes.size() != 1)
```
ekpyron, line 701: I'm actually not convinced that we should even implement this for legacy code generation.
The legacy optimizer is not powerful enough to inline and compile-time evaluate even the simplest canonical fixed-point literal cases, so it pretty much never de facto does what people will expect (while via-IR this works well enough).

cameel: Is that a strong enough reason not to implement it for legacy though? This is also a convenience feature for UDVTs so it has some merits beyond efficiency. It's not worse than a normal function call in any case. And in the docs I do not hide the fact that this is just a function call syntax at all so I'm not sure the expectation will really be that strong.

I think it may be better to let people decide for themselves. Or we could issue a warning that would kinda discourage them (and explain the situation) but not hard-disallow this.

Also, having differences between the codegens just increases testing surface and is an unnecessary complication. It's simpler for us when things work the same way.

cameel: By the way, this may be a good moment to take a look at the docs PR (#14121) so that I have a chance to do apply any potential fixes this week, while I'm still available.

ekpyron: It's pretty much solely an alternative of having to declare constants with precomputed representations of the UDVT - i.e. ``12.34 f`` is mainly convenience syntax for:
````
Fixed18 constant f12_34 = Fixed18.wrap(12_34 * 10**(18-2)));
````
If the latter is cheaper than the former, people will keep using that, even though it's uglier.

ekpyron: I.e. this is *not* an alternative to function calls, but to that really, so it not being more expensive than function calls doesn't really help that much...

cameel: I still think it's a pretty weak reason for not keeping parity between codegens. In my mind it's a line that I would not cross lightly. But anyway, it's your call. If you want, I can disable suffixes on legacy.

ekpyron: We actually said five features ago that we feature-freeze legacy codegen, I don't know why we still keep implementing stuff for it at all anyways :-). Let alone the fact that there *is no* feature parity between them anyways...

ekpyron: But even if not for that, I'm not sure how "cannot be used for the purpose it is supposed to address" is a weak reason :-D.

libsolidity/ast/AST.h
ekpyron: Except this one :-)?

docs/grammar/SolidityParser.g4
ekpyron: [resolving this based on the assumption that that answered the question]

libsolidity/ast/AST.h
cameel: Ah, my bad. I did not look closely enough to see that it's about a different class. Fixed now.

libsolidity/codegen/ExpressionCompiler.cpp
cameel: > We actually said five features ago that we feature-freeze legacy codegen,

Wait, what? If we're in such a freeze now then I don't think it was communicated clearly enough.

I remember some casual discussions that maybe we should stop implementing features for legacy at some point, and I know that we're generally closing issues that do not exist in the new codegen, but I don't think I've heard it explicitly stated that we should no longer be implementing anything for legacy as a general rule. I don't think I've seen us actually doing that in any recent PR and it was not even suggested in the operators PR (which was the most recent major feature for me).

> But even if not for that, I'm not sure how "cannot be used for the purpose it is supposed to address" is a weak reason :-D.

I think it's weak because IMO the response is disproportionate to the problem. I.e. it's not like it being there actively harms anyone - optimizors will ignore it, other people will decide for themselves if they care for gas or for syntax more. And not all uses of the syntax are inefficient anyway. If anything, this encourages people to migrate to IR. Increasing the feature disparity seems like it would be more annoying to users, on the other hand.

But let's talk about it on the call today. This thread is getting long and this topic seems like something that's quicker to discuss in person.

cameel:
We decided not to merge this feature so I'm closing the PR. See https://github.com/ethereum/solidity/issues/13718#issuecomment-1551488965 for details.

PR #12659, opened 2022-02-13
Add Base64 test cases with and without inline assembly

### LINKED ISSUE #12442: Add an optimizer test case based on OpenZeppelin's Base64 library
OpenZeppelin is currently working on a base-64 library (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2884) and it turns out that our optimizer is not yet good enough to get to the level of hand-optimized assembly in that case. We should add this code as a test case so that we can use it to evaluate our optimizer improvements.

In the PR and its comments you'll find versions of the library with and without inline assembly. Add them both as test cases. I think I'd put them in [`solidity/test/libsolidity/semanticTests/externalContracts/`](https://github.com/ethereum/solidity/tree/develop/test/libsolidity/semanticTests/externalContracts), which would allow us to see gas usage in both cases. Add some expectations with known good base-64 input/output to make sure the code works correctly (the one I originally posted in comments there was actually buggy).

It might also be helpful to have these as command-line tests with `--ir-optimized` ouptut to be able to see changes in the generated Yul.

paarthmadan:
### What's in this PR
Closes #12442

This PR adds test cases for a `Base64` implementation inspired by [OpenZepplin's implementation.](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2884/files)

It adds two test cases, `base64_inline_asm` and `base64_no_inline_asm` to assist in optimizer improvement.

I added assertions to replicate the test vectors specified in the [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#section-10) to ensure both implementations are to specification.

#### Organization
This is my first time contributing, so I don't have strong feelings towards the way I've organized the code. Perhaps it makes more sense to fold both implementations into the one base file as opposed to introduce a different one for each implementation. Happy to defer to someone with expertise.

### DIFF
--- a/test/libsolidity/semanticTests/externalContracts/_base64/base64_inline_asm.sol
+++ b/test/libsolidity/semanticTests/externalContracts/_base64/base64_inline_asm.sol
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Provides a set of functions to operate with Base64 strings.
+ */
+library InlineAsmBase64 {
+    /**
+     * @dev Base64 Encoding/Decoding Table
+     */
+    string internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+    /**
+     * @dev Converts a `bytes` to its Bytes64 `string` representation.
+     */
+    function encode(bytes memory data) internal pure returns (string memory) {
+        /**
+         * Inspired by OpenZepplin Base64 implementation
+         * https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2884/commits/157c32b65a15cb0b58257543643cafa1cebf883a
+         */
+        if (data.length == 0) return "";
+
+        // Loads the table into memory
+        string memory table = _TABLE;
+
+        // Encoding takes 3 bytes chunks of binary data from `bytes` data parameter
+        // and split into 4 numbers of 6 bits.
+        // The final Base64 length should be `bytes` data length multiplied by 4/3 rounded up
+        // - `data.length + 2`  -> Round up
+        // - `/ 3`              -> Number of 3-bytes chunks
+        // - `4 *`              -> 4 characters for each chunk
+        uint256 encodedLen = 4 * ((data.length + 2) / 3);
+
+        // Add some extra buffer at the end required for the writing
+        string memory result = new string(encodedLen);
+
+        assembly {
+            // Store the actual result length in memory
+            mstore(result, encodedLen)
+
+            // Prepare the lookup table
+            let tablePtr := add(table, 1)
+
+            // Prepare input pointer
+            let dataPtr := data
+            let endPtr := add(dataPtr, mload(data))
+
+            // Prepare result pointer, jump over length
+            let resultPtr := add(result, 32)
+
+            // Run over the input, 3 bytes at a time
+            for {
+
+            } lt(dataPtr, endPtr) {
+
+            } {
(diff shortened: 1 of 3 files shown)


test/libsolidity/semanticTests/externalContracts/base64.sol
```diff
@@ -0,0 +1,39 @@
+==== ExternalSource: _base64/base64_inline_asm.sol ====
+==== ExternalSource: _base64/base64_no_inline_asm.sol ====
```
leonardoalt, line 2: Why are these ExternalSource needed here?

test/libsolidity/semanticTests/externalContracts/base64.sol
```diff
@@ -0,0 +1,39 @@
...
+// ----
+// constructor()
+// encode_inline_asm(bytes): 0x20, 0 -> 0x20, 0
+// encode_inline_asm(bytes): 0x20, 1, "f" -> 0x20, 4, "Zg=="
+// encode_inline_asm(bytes): 0x20, 2, "fo" -> 0x20, 4, "Zm8="
+// encode_inline_asm(bytes): 0x20, 3, "foo" -> 0x20, 4, "Zm9v"
+// encode_inline_asm(bytes): 0x20, 4, "foob" -> 0x20, 8, "Zm9vYg=="
+// encode_inline_asm(bytes): 0x20, 5, "fooba" -> 0x20, 8, "Zm9vYmE="
+// encode_inline_asm(bytes): 0x20, 6, "foobar" -> 0x20, 8, "Zm9vYmFy"
+//
```
leonardoalt: I'm not sure whether this blank line causes weirdness in the tests.

leonardoalt:
@paarthmadan thanks for the PR!
Some of the CI tests are failing with "Compiling contract failed". Can you please check the errors by clicking on `Details` on the failing CI tests at the end of this page?

test/libsolidity/semanticTests/externalContracts/_base64/base64_inline_asm.sol
```diff
@@ -0,0 +1,96 @@
...
+                let input := mload(dataPtr)
+
+                // To write each character, shift the 3 bytes (24 bits) chunk 4
+                // times in blocks of 6 bits for each character (18, 12, 6, 0)
+                // and apply logical AND with 0x3F to extract the 6-bit group.
+                // Add the 6-bit group with the table ptr to index into the
+                // table and acquire the character to write.  Finally, write
+                // the character to the result pointer.
+
+                mstore8(resultPtr, mload(add(tablePtr, and(shr(18, input), 0x3F))))
```
ekpyron, line 69: Since there is shifts in here, you will have to specify
``EVMVersion: >=constantinople`` as test option (next to ``compileViaYul``) in the actual tests (shift opcodes were not available in earlier evm versions).

ekpyron:
Parts of the CI failures are missing gas expectations - you can run ``isoltest --optimize`` to have the required gas expecations generated automatically (note that they will only be added if the gas usage is beyond ``100000`` gas, so if you want gas expectations for the runtime calls, you may need to use larger strings).

chriseth:
If we want to get proper runtime numbers, I think we need to encode more data. This does not have to be given explicitly, it can also  be generated e.g. in a loop in the solidity code.
What do you think, @cameel ?

paarthmadan:
@chriseth I've added two functions which perform encode N number of times (arbitrarily chosen to be 100). This provides proper runtime numbers for each of the approaches, and accentuates the advantage of using the inline assembly optimization in the current state of the optimizer (which was the original intent of this change).

I've made some minor adjustments to ensure CI is green.

chriseth (approved):

PR #12662, opened 2022-02-14
Disable running smt tests in ubsan runs.

ekpyron:
We should figure out a more selective way to disable errors on allocation issues in Z3/spacer - both for this and for the asan test runs.
But since this is what we do for the asan runs, maybe it's good enough to stop the constant failure notifications for the time being.


bshastry (approved):
LGTM. The CI failure looks unrelated to the change in this PR.

ekpyron:
Created https://github.com/ethereum/solidity/issues/12664 for looking into this in more detail.

PR #12663, opened 2022-02-14
Assembly annotation as memory-safe using assembly flags.

ekpyron:
Depends on https://github.com/ethereum/solidity/pull/12620


libsolidity/parsing/Parser.cpp
```diff
@@ -1313,11 +1313,32 @@ ASTPointer<InlineAssembly> Parser::parseInlineAssembly(ASTPointer<ASTString> con
...
+		std::string const& dialectString = m_scanner->currentLiteral();
+		auto it = dialectString.begin();
+		auto end = dialectString.end();
+		auto dialectStart = it;
+		while (it != end && !isspace(*it))
+			++it;
+
 		// This can be used in the future to set the dialect.
+		if (string(dialectStart, it) != "evmasm")
+			fatalParserError(4531_error, "Only \"evmasm\" supported.");
```
ekpyron: I'd actually like to just parse the entire string as string and syntax-/type-check it later... but we can't really do that, since we need the dialect to confirm the dialect to actually parse the assembly below...

Conversely, I could move all the flags checks back here to the parser...

docs/assembly.rst
```diff
@@ -358,3 +355,16 @@ in memory is automatically considered memory-safe and does not need to be annota
...
+of solidity, you can use a special comment to annotate an assembly block as memory-safe:
+
+.. code-block:: solidity
+
+    /// @solidity memory-safe-assembly
+    assembly {
+        ...
+    }
+
+Note that this will only be possible until the next breaking release, so if you are not concerned with
```
chriseth: I would propose to keep it at least for 0.9.x

libsolidity/parsing/Parser.cpp
```diff
@@ -1313,11 +1314,32 @@ ASTPointer<InlineAssembly> Parser::parseInlineAssembly(ASTPointer<ASTString> con
 
 	expectToken(Token::Assembly);
 	yul::Dialect const& dialect = yul::EVMDialect::strictAssemblyForEVM(m_evmVersion);
+	ASTPointer<vector<ASTPointer<ASTString>>> dialectFlags;
 	if (m_scanner->currentToken() == Token::StringLiteral)
 	{
-		if (m_scanner->currentLiteral() != "evmasm")
-			fatalParserError(4531_error, "Only \"evmasm\" supported.");
+		std::string const& dialectString = m_scanner->currentLiteral();
+		auto it = dialectString.begin();
```
chriseth: Why don't you just use `boost::split`?

libsolidity/parsing/Parser.cpp
```diff
@@ -1313,11 +1314,32 @@ ASTPointer<InlineAssembly> Parser::parseInlineAssembly(ASTPointer<ASTString> con
...
+		auto end = dialectString.end();
+		auto dialectStart = it;
+		while (it != end && !isWhiteSpace(*it))
+			++it;
+
 		// This can be used in the future to set the dialect.
+		if (string(dialectStart, it) != "evmasm")
```
chriseth: Do we really require 'evmasm' now? Can't it just be the default?

ekpyron: You mean ``assembly "memory-safe" {`` would be fine?
I'm not sure... the idea so far was to require the first world to be the dialect and then allow flags later than that... having the dialect itself just be a flag at arbitrary position... would probably also work, though.

libsolidity/parsing/Parser.cpp
ekpyron: Is that locale-whitespace-safe and sane and everything :-)? I was actually wondering if we had our own custom helper for this, but didn't find one...

libsolidity/parsing/Parser.cpp
ekpyron: Do you like this syntax in general? I'm not sure ``assembly "evmasm memory-safe" {`` really looks good - and there's little reason for not making it ``assembly "evmasm" "memory-safe" {}``, is there? Then we could also distinguish more explicitly between dialect and flags, e.g. as ``assembly "evmasm" ("memory-safe") {}``... and then allow omitting the dialect defaulting to ``"evmasm"``... Or something like that... generally what I mean to say is: I'm not sure about the syntax yet :-).

docs/assembly.rst
ekpyron: Reformulated it to a more neutral "in a future breaking release".

libsolidity/parsing/Parser.cpp
chriseth: If is locale-aware only if you use locale-aware separators.

libsolidity/parsing/Parser.cpp
chriseth: Yeah, I think this requires further discussion.

libsolidity/parsing/Parser.cpp
```diff
@@ -1321,13 +1321,28 @@ ASTPointer<InlineAssembly> Parser::parseInlineAssembly(ASTPointer<ASTString> con
...
+	{
+		flags = make_shared<vector<ASTPointer<ASTString>>>();
+		do
+		{
+			advance();
+			expectToken(Token::StringLiteral, false);
+			flags->emplace_back(make_shared<ASTString>(m_scanner->currentLiteral()));
+			advance();
+		}
+		while(m_scanner->currentToken() == Token::Comma);
```
stackenbotten, line 1335: Coding style error

libsolidity/analysis/DocStringTagParser.cpp
```diff
@@ -202,7 +202,16 @@ bool DocStringTagParser::visit(InlineAssembly const& _assembly)
 				if (valuesSeen.insert(value).second)
 				{
 					if (value == "memory-safe-assembly")
+					{
+						if (_assembly.annotation().markedMemorySafe)
+							m_errorReporter.warning(
```
chriseth, line 207: Do we really want the warning now? We wanted library authors to add the comment as early as possible. If it generates a warning, this kind of defeats the purpose.
What about adding a warning starting from 0.9?

chriseth: Ah wait, don't we also have "info" now?

libsolidity/ast/AST.h
```diff
@@ -1463,19 +1463,26 @@ class InlineAssembly: public Statement
...
 		ASTPointer<ASTString> const& _docString,
 		yul::Dialect const& _dialect,
+		ASTPointer<std::vector<ASTPointer<ASTString>>> _flags,
 		std::shared_ptr<yul::Block> _operations
 	):
-		Statement(_id, _location, _docString), m_dialect(_dialect), m_operations(std::move(_operations)) {}
+	Statement(_id, _location, _docString),
+	m_dialect(_dialect),
+	m_flags(move(_flags)),
+	m_operations(std::move(_operations))
```
chriseth:
```suggestion
		Statement(_id, _location, _docString),
		m_dialect(_dialect),
		m_flags(move(_flags)),
		m_operations(std::move(_operations))
```

libsolidity/ast/AST.h
```diff
@@ -1463,19 +1463,26 @@ class InlineAssembly: public Statement
...
+	{}
 	void accept(ASTVisitor& _visitor) override;
 	void accept(ASTConstVisitor& _visitor) const override;
 
 	yul::Dialect const& dialect() const { return m_dialect; }
 	yul::Block const& operations() const { return *m_operations; }
+	auto const& flags() const { return m_flags; }
```
chriseth: I don't want to look into the private details of a class to find out its interface.

libsolidity/analysis/DocStringTagParser.cpp
ekpyron: The warning is only for
```
/// @solidity memory-safe-assembly
assembly ("memory-safe") {}
```
I.e. doing it both.
That never makes sense - you'd only want to use the comment, if you want backwards compatibility, or only the flag, if not.
Maybe I'll rephrase the text.

libsolidity/ast/AST.h
ekpyron: Haha, it took me a full minute to understand what you mean - I don't even see this stuff :-). But fair enough, you're right :-).

test/libsolidity/syntaxTests/inlineAssembly/natspec_multi.sol
```diff
@@ -0,0 +1,19 @@
...
+    /// @after
+    assembly {}
+    /// @solidity memory-safe-assembly a a a
+    ///           memory-safe-assembly
+    assembly {}
+}
+// ----
+// Warning 6269: (153-164): Unexpected natspec tag in inline assembly: after
+// Warning 6269: (153-164): Unexpected natspec tag in inline assembly: before
+// Warning 8787: (253-264): Unexpected value for @solidity tag in inline assembly: a
```
ekpyron: Wait a second, this seems wrong... Actually in the other one...

libsolidity/analysis/DocStringTagParser.cpp
ekpyron: Rephrased.

ekpyron: Then again, I could also just remove this warning altogether.

Also: *should* I add an info message about the comment being deprecated in the future? I'd tend to not saying anything now and making it a warning in 0.9.

chriseth (approved):

Issue #12460, opened 2021-12-28
Unexpected SLOADs during storing tightly packed struct


k06a:
## Description

Unexpected SLOADs during storing tightly packed struct. The following code should produce 0 SLOADs and 2 SSTOREs from my point of view. But somehow it executes 1 SLOAD before SSTOREs.

## Environment

- Compiler version: 0.8.11
- Target EVM version (as per compiler settings): default (runs=1000000)
- Framework/IDE (e.g. Truffle or Remix): Remix, Hardhat

## Steps to Reproduce

```solidity
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.11;

contract BitPacking {
    struct Data {
        uint48 a;
        uint40 b;
        uint40 c;
        uint128 d;
        uint128 e;
        uint128 f;
    }

    Data public data;

    function func(uint48 _1, uint40 _2, uint40 _3, uint128 _4) external {
        data = Data(_1, _2, _3, _4, _4, _4);
    }
}
```

chriseth:
Solidity is still not very good at optimizing across memory stores (`Data(...)` creates the struct in memory). For now, the more efficient (and maybe also safer) way to do this is to use individual assignments - `data.a = _1; data.b = _2; ...` - since this does not create any copies in memory.

It is weird, though, that slot `0` is even read when doing individual assignments.

k06a:
@chriseth I tried this code too, even before using structs. I have the same SLOADs and SSTOREs counts but no MLOADs and MSTOREs :)

hrkrshnn:
Looked at the IR. The issue is that the function calls for writing to storage, corresponding to `d`, `e` and `f` are not getting inlined. I'd assume the same issue for the legacy codegen. I think improving the inlining heuristics should fix this.

k06a:
@hrkrshnn what do you think about adding tests for such cases?

chriseth:
I don't think we have an issue with inlining in the old codegen, because most routines are always inlined at code gen phase already. I would still like the sload of zero to be investigated.

chriseth:
@k06a we already have tests for that and we know that the gas consumption is far from ideal.

I just re-checked the version where we store in struct members directly, i.e.
```
contract BitPacking {
    struct Data {
        uint48 a;
        uint80 b;
        uint128 c;
    }

    Data data;

    function func(uint48 _1, uint80 _2, uint128 _3) external {
        data.a = _1;
        data.b = _2;
        data.c = _3;
    }
[...]

PR #12672, opened 2022-02-15
Redundant store eliminator limited

chriseth:


libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+	if (!ranges::all_of(funCall->arguments, [](Expression const& _expr) -> bool {
+		return get_if<Identifier>(&_expr) || get_if<Literal>(&_expr);
+	}))
+		return;
+
+	if (
+		*instruction == Instruction::SSTORE ||
+		(!m_ignoreMemory && (
+			*instruction == Instruction::EXTCODECOPY ||
```
chriseth: This should be pulled into into SemanticInformation.h or similar.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,463 @@
...
+		vector<Operation> result;
+		// Unknown read is worse than unknown write.
+		if (sideEffects.memory != SideEffects::Effect::None)
+			result.emplace_back(Operation{Location::Memory, Effect::Read, {}, {}});
+		if (sideEffects.storage != SideEffects::Effect::None)
+			result.emplace_back(Operation{Location::Storage, Effect::Read, {}, {}});
+		return result;
+	}
+
+	switch (*instruction)
```
chriseth: This should be pulled into into SemanticInformation.h or similar.

Of course we cannot work with yul strings here, but we could say:
 - this reads from and writes to memory
 - the read start can be found in argument 2
 - the read length is always 32
 - the write length can be found in argument 7

and so on.

k06a:
Hope this new PR will also help with this issue: https://github.com/ethereum/solidity/issues/12460#issuecomment-1013040347

chriseth:
> Hope this new PR will also help with this issue: [#12460 (comment)](https://github.com/ethereum/solidity/issues/12460#issuecomment-1013040347)

In combination with #12206 it does address the issue, but it might also need some more careful tuning of the inlining parameters to make this always work.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+	UnusedStoreBase::visit(_statement);
+
+	auto const* exprStatement = get_if<ExpressionStatement>(&_statement);
+	if (!exprStatement)
+		return;
+
+	FunctionCall const* funCall = get_if<FunctionCall>(&exprStatement->expression);
+	if (!funCall)
+		return;
```
ekpyron:
```suggestion
	yulAssert(funCall);
```
?

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+	using evmasm::SemanticInformation;
+	bool isStorageWrite = (*instruction == Instruction::SSTORE);
+	bool isMemoryWrite =
+		*instruction == Instruction::EXTCODECOPY ||
+		*instruction == Instruction::CODECOPY ||
+		*instruction == Instruction::CALLDATACOPY ||
+		*instruction == Instruction::RETURNDATACOPY ||
+		*instruction == Instruction::MSTORE ||
+		*instruction == Instruction::MSTORE8;
+	bool isWriteToBeHandled =
```
ekpyron:
```suggestion
	bool isCandidateForRemoval =
```
?
That's what this really is, right?

ekpyron: It also took me a second to see why this is done on the statement level, but we can of course only remove statements anyways... I wonder if in the future, once we'd summarize stores for yul functions, it'll be worthwhile to consider splitting a function into its write part and the computation of its return value part, though...
I.e. try to transform
```
function f(a) -> r { sstore(0, a) r := add(a,1) }
sstore(1, f(42))
sstore(0, 0)
```
```
function f(a) -> r { sstore(0, a) r := add(a,1) }
function f_pure_part(a) -> r { r := add(a,1) }
sstore(1, f_pure_part(42))
sstore(0, 0)
```
and thus ultimately do this on the expression level instead...
But not sure if that likely happened a lot and right now probably just a distraction to consider...

libyul/optimiser/UnusedStoreEliminator.h
```diff
@@ -0,0 +1,117 @@
...
+
+	using UnusedStoreBase::operator();
+	void operator()(FunctionCall const& _functionCall) override;
+	void operator()(FunctionDefinition const&) override;
+	void operator()(Leave const&) override;
+
+	using UnusedStoreBase::visit;
+	void visit(Statement const& _statement) override;
+
+	enum class Location { Storage, Memory };
```
ekpyron:
```suggestion
	using Location = SemanticInformation::Location;
```
and get rid of ``translateLocation``? Or do we expect ``SemanticInformation::Location`` to be extended to calldata or returndata?

ekpyron: Similarly, maybe even for ``Effect``... ``applyOperation`` can just ignore ``None`` (or assert ``!= None``)... but also fine to not change either.

libyul/optimiser/UnusedStoreEliminator.h
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Optimizer component that removes sstore statements if they
+ * are overwritten in all code paths or never read from.
+ *
+ * The m_store member of UnusedStoreBase is only used with the empty yul string
+ * as key in the first dimension.
+ *
+ * Best run in SSA form.
+ *
+ * Prerequisite: Disambiguator, ForLoopInitRewriter.
```
ekpyron, line 50:
```suggestion
 * Prerequisite: Disambiguator, ForLoopInitRewriter, FunctionHoister.
```
?
Otherwise I'd need to wonder if we need to save stuff like ``m_forLoopInfo`` and/or ``m_forLoopNestingDepth`` in the function definition visit...

ekpyron: Oh, sorry, no, this is actually done in ``UnusedStoreBase::operator()(FunctionDefinition const& _functionDefinition)``.... but isn't that missing the ``m_forLoopNestingDepth`` part there? Not that that's critical, but still...

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+		vector<Operation> operations = operationsFromFunctionCall(*funCall);
+		yulAssert(operations.size() == 1, "");
+		m_storeOperations[&_statement] = move(operations.front());
+	}
+}
+
+void UnusedStoreEliminator::finalizeFunctionDefinition(FunctionDefinition const&)
+{
+	changeUndecidedTo(State::Used);
+	scheduleUnusedForDeletion();
```
ekpyron, line 170: Is this needed here?

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+		if (sideEffects.storage != SideEffects::Effect::None)
+			result.emplace_back(Operation{Location::Storage, Effect::Read, {}, {}});
+		return result;
+	}
+
+	using evmasm::SemanticInformation;
+
+	return util::applyMap(SemanticInformation::readWriteOperations(*instruction),
+		[&](SemanticInformation::Operation const& _op) -> Operation
+	{
```
ekpyron: indentation looks off

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include <libevmasm/SemanticInformation.h>
+
+#include <range/v3/algorithm/all_of.hpp>
+
+using namespace std;
+using namespace solidity;
+using namespace solidity::yul;
+
+/// Variable names for special constants that can never appear in actual Yul code.
+static string const zero{"@ 0"};
```
ekpyron, line 46:
```suggestion
static string const zero{"0"};
```
should also be just fine... but doesn't matter... eventually it would probably be worth a thought to support implicit literal values in the knowledge base directly... but fine for now.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+) const
+{
+	KnowledgeBase knowledge(m_dialect, m_ssaValues);
+
+	if (_op1.location != _op2.location)
+		return true;
+	if (_op1.location == Location::Storage)
+	{
+		if (_op1.start && _op2.start)
+			return knowledge.knownToBeDifferent(*_op1.start, *_op2.start);
```
ekpyron: Since we did set the lengths to constant 1 for these, maybe we should assert that they're there and that they are one here.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,404 @@
...
+	UnusedStoreEliminator::Operation const& _covered,
+	UnusedStoreEliminator::Operation const& _covering
+) const
+{
+	if (_covered.location != _covering.location)
+		return false;
+	if (
+		(_covered.start && _covered.start == _covering.start) &&
+		// length is unused for storage
+		(_covered.location == Location::Storage || (_covered.length && _covered.length == _covering.length))
```
ekpyron:
```suggestion
		(_covered.length && _covered.length == _covering.length)
```
Since they're set to 1 in the other PR now for storage, this can be simplified, can't it?

ekpyron (dismissed):
I haven't rechecked the control flow parts in ``UnusedStoreBase`` - I hope they're fine :-).

Apart from that I only had minor comments, but this looks correct to me. At least checking the lengths for the storage side effects would be nice, though, given that we set them to 1 in https://github.com/ethereum/solidity/pull/12693.

libyul/optimiser/UnusedStoreEliminator.cpp
chriseth: How would it work without that?

libyul/optimiser/UnusedStoreEliminator.h
chriseth: Ah that is true, but non-critical, it's just a resource limiter.

libyul/optimiser/UnusedStoreEliminator.h
chriseth: I wanted to avoid the dependency...

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: Won't the ones that are unused at this point just remain unused and ``scheduleUnusedForDeletion()`` in
``UnusedStoreEliminator::run`` will just take care of them later in bulk?

ekpyron: Ah, they'll be removed due to the scoped saving thing in the base class maybe?

ekpyron: Yeah, they will... although we probably wouldn't really have to do that, would we? But even then, better to keep the map small. Yeah, better to just keep it, so nevermind!

libyul/optimiser/UnusedStoreEliminator.h
ekpyron: On the header? Ok, fair enough... two types and two translation functions aren't that low of a price for avoiding that, but fine with me :-).

libyul/optimiser/UnusedStoreEliminator.h
ekpyron: Yeah, and in practice it'll be hoisted anyways, so doesn't matter.

ekpyron: https://github.com/ethereum/solidity/pull/12728 if we still want to fix it.

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: This reverted again now - but well, it also doesn't really matter too much.

test/cmdlineTests/debug_info_in_yul_and_evm_asm_print_all/output
```diff
@@ -201,16 +201,14 @@ object "C_6" {
...
-                    if eq(0x26121ff0, shr(224, calldataload(_2)))
+                    let _1 := 0
+                    if eq(0x26121ff0, shr(224, calldataload(_1)))
                     {
-                        if callvalue() { revert(_2, _2) }
-                        if slt(add(calldatasize(), not(3)), _2) { revert(_2, _2) }
-                        return(_1, _2)
+                        if callvalue() { revert(_1, _1) }
+                        if slt(add(calldatasize(), not(3)), _1) { revert(_1, _1) }
+                        return(memoryguard(0x80), _1)
```
ekpyron, line 211: We could consider to do ``return(x,0) -> return(0,0)`` and ``revert(x,0) -> revert(0,0)`` in the evmasm optimizer, if we don't do that already.
Or on the Yul level ``return(x,0) -> pop(x) return(0,0)`` (we can't just remove the memoryguard, but ``pop(memoryguard(...))`` should for sure be thrown away by libevmasm)...

ekpyron (dismissed):
Might be nice to add more specific tests for the overflow cases - and I haven't yet checked for sufficient test coverage in general.
But I'm approving anyways, since I think everything should be valid.

chriseth:
Maybe it would be good if you could check the test coverage? :)

ekpyron:
> Maybe it would be good if you could check the test coverage? :)

Yeah, my plan was to look into that some more tomorrow in any case.

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_end.yul
```diff
@@ -0,0 +1,20 @@
+{
+    function f() {
+        let x := calldataload(2)
+        sstore(x, 2)
+        // This cannot be removed because we do not know what happens after the function.
+        sstore(x, 3)
```
ekpyron: Doesn't this make more sense with ``mstore``s? The second ``sstore`` couldn't be removed even if it wasn't a function...

test/libyul/yulOptimizerTests/unusedStoreEliminator/unaligned_access.yul
```diff
@@ -0,0 +1,15 @@
+{
+    mstore(0, 0x1234)
+    mstore(4, 0x456)
+    revert(0, 0x24)
```
ekpyron:
```suggestion
    revert(0, 5)
```
? The ``0x24`` case is also covered by ``overlapping.yul``

ekpyron:
Because I was curious: That's an external test benchmark diff between develop and this (although this PR may not be rebased precisely, so the comparison may be slightly off): https://gist.github.com/ekpyron/2af0bea5b23282caffa2d2da43b88d75
Which seems to indicate that without more aggressive inlining and/or dropping the restriction to constants only, this doesn't work extremely well in practice - but regardless it's a good first step.

test/cmdlineTests/debug_info_in_yul_and_evm_asm_print_all/output
chriseth: Sounds good!

chriseth: Although we should rather do that in the yul optimizer shouldn't we?

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_end.yul
chriseth: will change

test/libyul/yulOptimizerTests/unusedStoreEliminator/memoryguard.yul
```diff
@@ -0,0 +1,24 @@
+{
+    mstore(0x40, memoryguard(100))
+    let free_mem_ptr := mload(0x40)
+    // redundant
+    mstore(free_mem_ptr, 100)
+    // redundant
+    mstore8(add(free_mem_ptr, 31), 200)
```
chriseth, line 7: For the future: Even if `free_mem_ptr` is completely unknown and this addition might wrap, we can still remove it, right?
Because if it wraps, the mstore below will revert.

test/libyul/yulOptimizerTests/unusedStoreEliminator/create.yul
```diff
@@ -0,0 +1,15 @@
+{
+    sstore(5, 10)
+    pop(create(0, 0, 0))
```
ekpyron, line 4: Maybe we should add one like this with the creation - or maybe to mix it up a ``call`` - inside a function? (Unless I missed the test case for that, I thought I already saw earlier, but don't see one now).

Also maybe the opposite:

```
{
  function f() -> r {
     if calldataload(0) { f() } // prevent inlining
     r := mload(0x20)
  }
  sstore(5, 10) // should be removed
  mstore(0, 42) // could be removed, but will probably stay?
  pop(f())
  sstore(5, 10)
}
```

test/libyul/yulOptimizerTests/unusedStoreEliminator/covering_calldatacopy.yul
```diff
@@ -0,0 +1,59 @@
+{
+    let start := calldataload(0x10)
```
ekpyron, line 2: This probably only makes sense with ``start`` being a constant in the latest version? Or is this test just for future reference?

test/libyul/yulOptimizerTests/unusedStoreEliminator/memoryguard.yul
chriseth: -> https://github.com/ethereum/solidity/issues/12735

ekpyron: You mean if mstore A is covered (in arbitrary precision reasoning) by mstore B, overflows in mstore A don't matter, since even if mstore A doesn't revert and writes to a wrapper offset (in which case it's not actually covered by B), then mstore B has to revert?
Sounds about right, yeah...

ekpyron: So in that sense overflows are only an issue for "never read", but not for "covered"?

test/cmdlineTests/debug_info_in_yul_and_evm_asm_print_all/output
ekpyron: Yeah, I first thought we couldn't, but ``return(x,0) -> pop(x) return(0,0)`` we *can* do in the yul optimizer.
``pop(memoryguard(...))`` would still stay - we could resolve the memoryguard and try to remove it *after* stack-to-memory, but then we have to be careful not to do anything that really affects stack layout anymore... so since ``pop <constant>`` will just be removed in evmasm, I'd say just doing ``return(x,0) -> pop(x) return(0,0)`` in Yul is good enough.... (and just rely on the evmasm optimizer to remove the pop - at least for now)

test/formal/redundant_store_unrelated.py
```diff
@@ -0,0 +1,63 @@
...
+solver.add(diff <= 2**(n_bits-1))
+
+# We check that the two conditions are conflicting:
+# - overlap
+# - start1 is small
+
+# Overlap:
+# x is a potential point where the memory operations
+# overlap.
+# Note that we do not use wrapping arithmetics
```
ekpyron:
```suggestion
# Note that we do not use wrapping arithmetic
```
codespell complains otherwise apparently

test/formal/redundant_store_unrelated.py
```diff
@@ -0,0 +1,63 @@
+from z3 import Solver, Int, IntVal, unsat
+import sys
```
ekpyron:
```suggestion
import sys
from z3 import Solver, Int, IntVal, unsat
```
python linter apparently complains otherwise

bshastry:
The fuzzer complains that the `mstore` in the following test is optimized out and hence the `sstore` only stores zero at the said location.

```
{
  mstore(2, 0x10000000000000000000000000000000000000000)
  pop(
    call(
      0x100000000000000000000000000000000000000000,
      0x1000000000000000000000000000000000000000000,
      0x10000000000000000000000000000000000000000000,
      0x100000000000000000000000000000000000000000000,
      0x1000000000000000000000000000000000000000000000,
      0,
      0x10000000000000000000000000000000000000000000000
    )
  )
  sstore(0x100000000000000000000000000000000000000000000000, mload(2))
}
```

ekpyron:
Oh, wow... that's right :-)!

The PR assumes that the ``call`` overrides the full output area, but we cannot assume that! Basically we have to assume that it may also just write nothing at all.

Kudos to the fuzzer, this would have been a pretty harsh bug!

bshastry:
~Another~ failing test case(s) reported by the fuzzer


##### staticcall

```
{
  mstore(0x9f, 0xa0)
  return(staticcall(0xa1, 0xff, 0x100, 0x101, 0x10, 0x100), 0x1000)
}
```

The `mstore` is incorrectly optimized because staticcall may not actually overwrite `0x9f`.

##### calldatacopy

```
{
  function foo_m_0() -> x_1, x_2
  {
    sstore(0, mload(2))
  }
  mstore(0, 42)
  calldatacopy(0, 0x3fff, 1)
  let x_7, x_8 := foo_m_0()
}
```

`calldatacopy` is optimized out.

chriseth:
Why is it incorrect to optimize out calldatacopy?
After inlining we get
```
{
  mstore(0, 42)
  calldatacopy(0, 0x3fff, 1)
  sstore(0, mload(2))
}
```
the calldatacopy copies a single byte to location `0`, but the `mload` only reads starting from `2`, so I don't think the calldatacopy is needed.

bshastry:
> Why is it incorrect to optimize out calldatacopy? After inlining we get
> 
> ```
> {
>   mstore(0, 42)
>   calldatacopy(0, 0x3fff, 1)
>   sstore(0, mload(2))
> }
> ```
> 
> the calldatacopy copies a single byte to location `0`, but the `mload` only reads starting from `2`, so I don't think the calldatacopy is needed.

Yup, this is a false positive due to the way the Yul interpreter silently ignores certain writes to memory (which consequently affects loads from memory)

chriseth:
I'll rebase.

chriseth:
Rebased.

chriseth:
Squashed.

libevmasm/SemanticInformation.cpp
```diff
@@ -121,7 +121,9 @@ vector<SemanticInformation::Operation> SemanticInformation::readWriteOperations(
 			Location::Memory,
 			Effect::Write,
 			paramCount - 2,
-			paramCount - 1,
+			// Length is in paramCount - 1, but it is only a max length,
+			// there is no guarantee that the full area is written to.
+			{},
```
ekpyron, line 126: Is this really the only case? What about ``CODECOPY`` and ``EXTCODECOPY``? Do they write zeroes beyond the respective code size or stop?

ekpyron: (The yellowpaper reads like they stop, but I haven't verified)

chriseth: they write zeros.

chriseth: It's the same with calldatacopy.

ekpyron: On the other hand evmone seems to fill in zeroes apparently...

ekpyron: ~~geth code may actually *not* fill in zeroes, though... although I'm not yet sure :-)...~~
geth looks like it also pads with zeroes, so that's probably it.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,378 @@
...
+		// Due to wrapping arithmetic, the below code can have "false positives".
+		// In the situation that the subtraction for computing "diff" wraps,
+		// we can return "unrelated" although the writes are actually related.
+		// In this case, 2.start and 1.start will both be >= largeNumber
+		// (see test/formal/redundant_store_unrelated.py).
+		// This means, both operations revert.
+		// Because of these false positives, this optimizer routine removes write operations
+		// where all subsequent read operations are unrelated or will revert due to high memory
+		// access. Since this second class of read operations will never return a value,
+		// it is irrelevant what value they return, so the write can be safely removed.
```
ekpyron: I wonder whether we should document this somewhere more prominently...
Also: did this have a significant impact beyond the literal-constant-only version? If not, I tend to play it safe and revert to that one... but if it has some effect, it's probably fine to go for it...

chriseth: It was able to optimize this out: https://github.com/ethereum/solidity/pull/12672/files#diff-454ca1e7741b290c37ead677fb52b645bd9c18325aa1aa5d023df0b926b2e1b7R16

I'm not sure what the impact in practice was. Of course, "known unrelated" is always difficult because we need the code to also terminate in the same function in order to say something about memory.

Still, if we want to improve this with the linear solver, we need exactly the same reasoning here.

chriseth: I just checked, that is the only change for isoltest.

Changelog.md
```diff
@@ -5,7 +5,8 @@ Language Features:
 
 
 Compiler Features:
-* JSON-AST: Added selector field for errors and events.
+ * JSON-AST: Added selector field for errors and events.
```
bshastry: Changelog entry is unrelated to this PR.

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: I would have hoped that the linear solver could actually tell us stuff like if ``_op1.start <= _op2.start`` and then the problem just vanishes...

ekpyron: But it might not be able to do that accurately either if either of them could ever have overflowed...

ekpyron: It still seems a bit weird to me to have the optimizer operate under conflicting assumptions - we say both "large memory reads don't return values", but we also allow them to be load-resolved, so we also say "we can assume any memory read to return a value"...
I don't see any issue due to this in practice (yet?), but it will never be formally correct like this at least :-)...

ekpyron: And I actually wouldn't be surprised there was a case, in which this actually happens:
- a store to high memory location A
- a store to high memory location B
- a read from high memory location A

in which B is spuriously unrelated to A... then the write to B would be removed and we may resolve the read from A, after which we can potentially further remove the write to A...
Even if that doesn't happen the way things are set up right now for any concrete values of A and B, we'd have the constant danger of this happening, once we extend the optimizer anywhere...

chriseth: Ok, the interaction with the load resolver is actually a good point.

libevmasm/SemanticInformation.cpp
ekpyron: For the record: I was just too tired to read the yellowpaper properly, so all fine.

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: To do this stuff properly, maybe there is no way around proving that the free memory pointer and arithmetic on it never overflows... but that's probably always going to be tricky (especially across functions calls and such)...

ekpyron: On the other hand: any code we generate in which the free memory pointer does overflow is ill-defined anyways (since we may then remove all kinds of memory writes and reads preventing actual out-of-gas, but have a write after the memory pointer overflow mess with the scratch space or the free memory pointer itself)...

chriseth:
The termination conditions are still wrong...

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,383 @@
...
+		if (coveredStart && coveringStart && coveredLength && coveringLength)
+		{
+			if (
+				*coveringStart <= *coveredStart &&
+				*coveringStart + *coveringLength >= *coveringStart && // no overflow
+				*coveredStart + *coveredLength >= *coveredStart && // no overflow
+				*coveredStart + *coveredLength <= *coveringStart + *coveringLength
+			)
+				return true;
+		}
```
ekpyron:
```suggestion
			if (
				*coveringStart <= *coveredStart &&
				*coveringStart + *coveringLength >= *coveringStart && // no overflow
				*coveredStart + *coveredLength >= *coveredStart && // no overflow
				*coveredStart + *coveredLength <= *coveringStart + *coveringLength
			)
				return true;
```
or just continued ``&&``

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,383 @@
...
+		// i.start <= e.start && e.start + e.length <= i.start + i.length
+		if (!_covered.start || !_covering.start || !_covered.length || !_covering.length)
+			return false;
+		optional<u256> coveredLength = knowledge.valueIfKnownConstant(*_covered.length);
+		optional<u256> coveringLength = knowledge.valueIfKnownConstant(*_covering.length);
+		if (knowledge.knownToBeEqual(*_covered.start, *_covering.start))
+		{
+			if (coveredLength && coveringLength && *coveredLength <= *coveringLength)
+				return true;
+		}
```
ekpyron:
```suggestion
			if (coveredLength && coveringLength && *coveredLength <= *coveringLength)
				return true;
```
or continued ``&&``.

ekpyron (commented):
LGTM apart from remaining minor coding style comments.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+		if (_op1.start && _op1.length && _op2.start && _op2.length)
+		{
+			optional<u256> length1 = knowledge.valueIfKnownConstant(*_op1.length);
+			optional<u256> length2 = knowledge.valueIfKnownConstant(*_op2.length);
+			if (
+				(length1 && *length1 <= 32) &&
+				(length2 && *length2 <= 32) &&
+				knowledge.knownToBeDifferentByAtLeast32(*_op1.start, *_op2.start)
```
ekpyron, line 299: Hm... this could be
```knownToBeDifferentByAtLeast(*_op.start, *_op2.start, max(*length1, *length2))```
thinking of two ``mstore8``s...

And there's nothing really that restricts this case to only ``<= 32``, is there?

Comment (ekpyron) [reply libyul/optimiser/UnusedStoreEliminator.cpp:299]: Probably ``<= 2^255`` is the only boundary, if any?

**Review by ekpyron — APPROVED** [review]

File: libyul/optimiser/UnusedStoreEliminator.cpp (hunk 25 above)
Comment (chriseth) [reply libyul/optimiser/UnusedStoreEliminator.cpp:299]: That is true, but I'm not sure it is worth it delaying the PR :)

**Comment by bshastry** [comment]
The fuzzer reported the following trace divergence (mstore optimized out and hence the keccak256 call returns a different value)

```
{
  mstore(add("z",iszero(0x9f)), 0xa0)
  sstore(0xa1, 0xff)
  sstore(0x100, 0x101)
  sstore(keccak256("z",0x10), 0x100)
}
```

**Comment by ekpyron** [comment]
@bshastry Actually, the ``keccak256`` seems to be evaluated even before this PR... so this resolves to
```
{
            mstore("z", 0xa0)
            sstore(0xa1, 0xff)
            let _1 := 0x100
            sstore(_1, 0x101)
            sstore(110620294328144418057589324861608220015688365608948720310623173341503153578932, _1)
}
```
after which the ``mstore`` is really redundant and *can* be removed...

and it's actually correct - we know the value stored in memory at ``"z"`` is ``0x00...00a0``, so the first ``0x10`` bytes are all zero and the big constant should be the keccak256 hash of 16 zeroes...

So it may be a false positive, but I'm not sure where it's coming from... maybe the interpreter (that's what you use as reference, right?) calculates the keccak hash incorrectly?

**Comment by ekpyron** [comment]
Or actually, wait... that's wrong...

**Comment by ekpyron** [comment]
But it's the keccak256 evaluation that's wrong, isn't it? ``mstore("z", 0xa0)`` does write the byte ``0xa0`` to the byte at memory offset ``"z"``, so the keccak256 should **not** evaluate to the empty keccak hash... should it :-)? I need to confirm - if that's right this may be a severe bug that is already released...

**Comment by ekpyron** [comment]
Ah, the yul interpreter doesn't interpret reads and writes from high memory addresses, does it?
``EVMInstructionInterpreter::accessMemory`` looks like it only allows access up to ``numeric_limits<size_t>::max() - 0xffff``...

And because of that ``EVMInstructionInterpreter::eval`` returns the bogus ``return u256("0x1234cafe1234cafe1234cafe") + arg[0];`` in L185 - and won't get the keccak value right... I think that's it...

**Comment by ekpyron** [comment]
Not sure if there's good reason to restrict the interpreter to only low memory accesses... maybe it should be a setting...

File: libevmasm/SemanticInformation.cpp (hunk 20 above)
Comment (bshastry) [reply libevmasm/SemanticInformation.cpp:126]: What if the address referenced by `extcodecopy` is not valid i.e., there is no code there? Does it still fill in as many zeroes as the `size` parameter dictates?

```
{
  mstore8(2, 100)
  extcodecopy(invalid_address, 0, 0, 32)
  sstore(0, mload(0))
}
```

stores zeroes at storage slot zero post optimization.

Changelog.md
chriseth: This was an old version, wasn't it?

libevmasm/SemanticInformation.cpp
chriseth: There is no such thing as an invalid or non-existing address in the EVM. It should still fill with zeros.

chriseth: Are you saying this is turned into `sstore(0, 0)`? The optimizer should not be able to know that the code is empty at that address.

Changelog.md
bshastry: Yes, sorry. Please ignore this comment.

libevmasm/SemanticInformation.cpp
bshastry: No, what I mean is that the optimizer removed the `mstore` which made me think if removal is correct even when `extcodecopy` does not reference a valid address. Good to know that EVM clients should still fill with zeroes in which case the optimizer is correct.

Issue #12614, opened 2022-02-02
[Docs] Order of contract parts / Solidity Style Guide maintenance


fulldecent:
## Page

https://docs.soliditylang.org/en/latest/style-guide.html?highlight=style%20guide

## Abstract

It is specified:

> Inside each contract, library or interface, use the following order:

But this advice is not followed in examples.

<img width="537" alt="Screen Shot 2022-02-01 at 19 14 07" src="https://user-images.githubusercontent.com/382183/152072835-4a95955e-7a29-40f6-a60b-143660c3f69b.png">
<img width="446" alt="Screen Shot 2022-02-01 at 19 14 25" src="https://user-images.githubusercontent.com/382183/152072837-f652e4f9-647d-4582-a139-a9da150491db.png">

THE QUESTION:

> Should constructor be specified to be above modifiers?
> Or should constructor be moved down with the functions in that example?

## Pull request

(I can provide after answering THE QUESTION)

ekpyron:
Right, thanks, I had not paid attention to constructors when merging https://github.com/ethereum/solidity/pull/12569...
Looking at the examples, I also realize that we don't specify a place for ``error`` declarations - and whatever we'd specify, the ``safe-remote.rst`` example would violate it :-)... and it already violates the event-modifier ordering... and it has the constructor *after* the modifiers... hm...

Also the part about "contract elements" should probably be renamed to "file-level elements" these days and would need to account for the new file-level elements like free functions and file-level error declarations...

Since the "Order of Functions" section considers the constructor a function, it would be "easier" to just move the constructor down in the examples... however, visually I personally find it more appealing to have it before :-)... EDIT: then again you can apply modifiers to constructors...

I'll ping our channel for more opinions!

ekpyron:
We briefly discussed this in our meeting today and we would actually ideally like to hand over the maintenance of the style guide to the community, since we don't want to impose any personal preferences and would rather like the guide to reflect common practice in the community.
Any suggestions on how best to organize that?

fulldecent:
Thank you for reviewing. Yes, that sounds good since it does not have any normative impact on Solidity proper.

Here is a recommend plan. And I volunteer to take all the responsibility and the work for these.

# Work plan

- [ ] Create new style repo https://github.com/ethereum/solidity-style-guide just for the style guide

  - [ ] Name "Solidity Community Style Guide"
  - [ ] Add me as maintainer

[...]

PR #12679, opened 2022-02-16
tweak: expand definition of suggested collision naming convention

devtooligan:
The [current style guide](https://github.com/ethereum/solidity/blob/v0.8.11/docs/style-guide.rst#avoiding-naming-collisions) suggests the use of trailing _ when the desired name of a function or variable collides with a built-in or otherwise reserved word.

However, this convention is widely used to not only apply to reserved words but also to include cases when the name would shadow an existing function or state variable name.

This pr attempts to expand the suggestion to include collisions with existing function or state variable names.

### DIFF
--- a/docs/style-guide.rst
+++ b/docs/style-guide.rst
@@ -1258,7 +1258,8 @@ Avoiding Naming Collisions
 
 * ``single_trailing_underscore_``
 
-This convention is suggested when the desired name collides with that of an existing state variable, function, built-in or otherwise reserved name.
+This convention is suggested when the desired name collides with that of
+an existing state variable, function, built-in or otherwise reserved name.
 
 .. _style_guide_natspec:
 
--- a/docs/style-guide.rst
+++ b/docs/style-guide.rst
@@ -1258,8 +1258,7 @@ Avoiding Naming Collisions
 
 * ``single_trailing_underscore_``
 
-This convention is suggested when the desired name collides with that of a
-built-in or otherwise reserved name.
+This convention is suggested when the desired name collides with that of an existing state variable, function, built-in or otherwise reserved name.
 
 .. _style_guide_natspec:
 


fulldecent:
This style policy is great.

But please update this PR to be a "DRAFT" on GitHub. I'm hoping we can address new opinions like this AFTER other cleanup work.

leonardoalt:
Since it's a small change, I'd suggest we merge it.

Marenz:
Yeah, I think this is minor enough to just go ahead

fulldecent:
Why save all the easy decisions for the existing team?

By presenting this easy one and some other hards ones together when we are ready for broader input, the process will be more engaging.

We need engagement because our goal here is not just to publish a document but to clean up how people write Solidity.

leonardoalt:
> Why save all the easy decisions for the existing team?

Not sure I get this question.

Larger modifications at once are always harder and take a long time to go through/converge.
Small changes on the other hand can be quickly analyzed and accepted/rejected. What's the problem with accepting this now and if the larger change wants to change it we check it there too?

chriseth:
Please make sure that you keep the lines in the documentation source short. This way diffs are easier to read and conflicts are less likely.

fulldecent:
There's no problem accepting this. It makes sense in this repo.

---

But I'm proposing we can have a few PR's like this ready to go (not one big PR) and wait for merge.

Then we (I) can present the list of them to the community or ask specific people for feedback all at once as a checklist.

The end result is we have a nice way to ask new people for their opinions and it is really easy for them to give their feedback.

I think people might appreciate being asked for that feedback—all at once, with some simple questions like this in there. And if we do it that way we might generate some goodwill and more people might read and use this document.

cameel:
I think I see what you're trying to do. Having some specific points to draw in people sounds like a good idea overall, just not sure if it's really necessary to have these sit in unmerged PRs. Small ones are small enough to either merge them quickly or move them to an issue. And doing big ones before they're accepted might be a wasted effort. I think discussing these in issues first might be better.

But anyway, I think it's important that you have full freedom in how you want to organize the process so my personal vote would be to just let you do it the way you want it and see how it goes :)

BTW, I just created a `style guide` label that can be applied to issues and PRs related to this. Actually, do you have enough permissions in the repo to use labels?

devtooligan:
Hey @cameel @fulldecent @leonardoalt I am looking forward to submitting many prs.  Would it be possible to add me as a `contributor`?  @fulldecent tried to add me but it didn't go through (I think that's why I couldn't tag you as a reviewer in the pr's, @fulldecent!)

fulldecent:
@devtooligan Please let's discuss this on the project board I made at https://github.com/users/fulldecent/projects/2

I hope you and I can discuss some things there before we bring them here to the Solidity repo.

---

I do not have access to use labels here. Thank you for creating the label.

cameel:
I think @chriseth has not added you to CODEOWNERS yet as discussed in https://github.com/ethereum/solidity/issues/12614#issuecomment-1034091671. I'll bring that up tomorrow.

Marenz (approved):

PR #12684, opened 2022-02-16
Never allocate more memory than the type size for static returns in external calls

### LINKED ISSUE #12647: external contract can waste caller's gas by return value.
When calling a contract, the compiler always assigns a buffer to hold entire returned data, even if not in use.
even when calling a "void" function, the generated yul code always does
```
    finalize_allocation(_2, returndatasize())
```
regardless if there is a code to use the data or not.

This command, by itself, doesn't cause any significant gas usage. however, any future memory operation will start at higher offset, and the caller contract pays for the entire storage in use, up to the highest offset.

Consider an inner call that does
```
    return (0,1000000)
```
Not only the inner call takes a lot of gas (as expected, to copy this large buffer), but also the calling code now wastes more gas on its own memory allocations.

## Suggestion:
1. optimize out this allocation when calling a "void" function
2. when calling a function with static return size (e.g. `returns (uint)`), use the return buffer size parameter of the call itself, instead of `returndatasize()`. This way, the caller expects (and uses) an exact amount of memory (it is also slightly cheaper than `returndatacopy(memptr, returndatasize())`, but this cost saving is marginal)
3. handling a function with dynamic return (e.g. `returns ( string memory)`) is more complex. This would probably require a language change, to signal the max expected size

## try/catch blocks
The same rules apply to:
```
    try methodCall() { }
    catch {
    }
```

nishant-sachdeva:
closes #12647

### DIFF
--- a/libsolidity/codegen/ReturnInfo.cpp
+++ b/libsolidity/codegen/ReturnInfo.cpp
@@ -54,4 +54,6 @@ ReturnInfo::ReturnInfo(EVMVersion const& _evmVersion, FunctionType const& _funct
 				estimatedReturnSize += retType->decodingType()->calldataEncodedSize();
 		}
 	}
+	if (dynamicReturnSize)
+		solAssert(estimatedReturnSize == 0);
 }
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
@@ -2505,6 +2505,8 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 			appendCode() << "mstore(add(" << m_utils.allocateUnboundedFunction() << "() , " << to_string(returnInfo.estimatedReturnSize) << "), 0)\n";
 	}
 
+	// NOTE: When the expected size of returndata is static, we pass that in to the call opcode and it gets copied automatically.
+    // When it's dynamic, we get zero from estimatedReturnSize() instead and then we need an explicit returndatacopy().
 	Whiskers templ(R"(
 		<?checkExtcodesize>
 			if iszero(extcodesize(<address>)) { <revertNoCode>() }
@@ -2514,22 +2516,29 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		mstore(<pos>, <shl28>(<funSel>))
 		let <end> := <encodeArgs>(add(<pos>, 4) <argumentString>)
 
-		let <success> := <call>(<gas>, <address>, <?hasValue> <value>, </hasValue> <pos>, sub(<end>, <pos>), <pos>, <reservedReturnSize>)
+		let <success> := <call>(<gas>, <address>, <?hasValue> <value>, </hasValue> <pos>, sub(<end>, <pos>), <pos>, <staticReturndataSize>)
 		<?noTryCall>
 			if iszero(<success>) { <forwardingRevert>() }
 		</noTryCall>
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
-			<?dynamicReturnSize>
-				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
-			</dynamicReturnSize>
+			<?isReturndataSizeDynamic>
+				let <returnDataSizeVar> := returndatasize()
+				returndatacopy(<pos>, 0, <returnDataSizeVar>)
+			<!isReturndataSizeDynamic>
+				let <returnDataSizeVar> := <staticReturndataSize>
+				<?supportsReturnData>
+					if gt(<returnDataSizeVar>, returndatasize()) {
+						<returnDataSizeVar> := returndatasize()
+					}
+				</supportsReturnData>
+			</isReturndataSizeDynamic>
 
 			// update freeMemoryPointer according to dynamic return size
-			<finalizeAllocation>(<pos>, <returnSize>)
+			<finalizeAllocation>(<pos>, <returnDataSizeVar>)
 
 			// decode return parameters from external try-call into retVars
-			<?+retVars> <retVars> := </+retVars> <abiDecode>(<pos>, add(<pos>, <returnSize>))
+			<?+retVars> <retVars> := </+retVars> <abiDecode>(<pos>, add(<pos>, <returnDataSizeVar>))
 		}
 	)");
 	templ("revertNoCode", m_utils.revertReasonIfDebugFunction("Target contract does not contain code"));
@@ -2558,21 +2567,18 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 	templ("funSel", IRVariable(_functionCall.expression()).part("functionSelector").name());
 	templ("address", IRVariable(_functionCall.expression()).part("address").name());
 
-	// Always use the actual return length, and not our calculated expected length, if returndatacopy is supported.
-	// This ensures it can catch badly formatted input from external calls.
-	if (m_context.evmVersion().supportsReturndata())
-		templ("returnSize", "returndatasize()");
-	else
-		templ("returnSize", to_string(returnInfo.estimatedReturnSize));
-
-	templ("reservedReturnSize", returnInfo.dynamicReturnSize ? "0" : to_string(returnInfo.estimatedReturnSize));
+	if (returnInfo.dynamicReturnSize)
+		solAssert(m_context.evmVersion().supportsReturndata());
+	templ("returnDataSizeVar", m_context.newYulVariable());
+	templ("staticReturndataSize", to_string(returnInfo.estimatedReturnSize));
+	templ("supportsReturnData", m_context.evmVersion().supportsReturndata());
 
 	string const retVars = IRVariable(_functionCall).commaSeparatedList();
 	templ("retVars", retVars);
 	solAssert(retVars.empty() == returnInfo.returnTypes.empty());
 
 	templ("abiDecode", m_context.abiFunctions().tupleDecoder(returnInfo.returnTypes, true));
-	templ("dynamicReturnSize", returnInfo.dynamicReturnSize);
+	templ("isReturndataSizeDynamic", returnInfo.dynamicReturnSize);
 
 	templ("noTryCall", !_functionCall.annotation().tryCall);
 
--- a/test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
+++ b/test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
@@ -1,0 +1,32 @@
+interface LongReturn {
+    function f() external pure returns (uint[20] memory);
+}
+contract ShortReturn {
+    function f() external pure returns (bytes32) {}
+}
+
+contract Test {
+    function test() public returns (uint) {
+        ShortReturn shortReturn = new ShortReturn();
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        LongReturn(address(shortReturn)).f();
+
+        uint freeMemoryAfter;
+
+        assembly {
+            freeMemoryAfter := mload(0x40)
+        }
+
+        return freeMemoryAfter - freeMemoryBefore;
+    }
+}
+// ====
+// EVMVersion: <=homestead
+// compileViaYul: true
+// ----
+// test() -> 0x0500
+// gas legacy: 131966
--- a/test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected_evm_version_after_homestead.sol
+++ b/test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected_evm_version_after_homestead.sol
@@ -1,0 +1,34 @@
+interface LongReturn {
+    function f() external pure returns (uint[20] memory);
+}
+contract ShortReturn {
+    function f() external pure returns (bytes32) {}
+}
+
+contract Test {
+    function test() public returns (uint) {
+        ShortReturn shortReturn = new ShortReturn();
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        // This reverts. The external call succeeds but ABI decoding fails due to the returned
+        // `bytes32` being much shorter than the expected `uint[20]`.
+        LongReturn(address(shortReturn)).f();
+
+        uint freeMemoryAfter;
+
+        assembly {
+            freeMemoryAfter := mload(0x40)
+        }
+
+        return freeMemoryAfter - freeMemoryBefore;
+    }
+}
+// ====
+// EVMVersion: >homestead
+// compileViaYul: true
+// ----
+// test() -> FAILURE
+// gas legacy: 131966
--- a/test/libsolidity/semanticTests/various/value_insane.sol
+++ b/test/libsolidity/semanticTests/various/value_insane.sol
@@ -21,7 +21,7 @@ contract test {
 // compileViaYul: also
 // ----
 // constructor(), 20 wei ->
-// gas irOptimized: 187835
+// gas irOptimized: 194261
 // gas legacy: 266728
 // gas legacyOptimized: 184762
 // sendAmount(uint256): 5 -> 8
--- a/test/cmdlineTests/standard_debug_info_in_evm_asm_via_ir_location/output.json
+++ b/test/cmdlineTests/standard_debug_info_in_evm_asm_via_ir_location/output.json
@@ -265,34 +265,39 @@ sub_0: assembly {
       return
         /* \"C\":403:411  this.f() */
     tag_36:
-        /* \"C\":79:428  contract C... */
       swap1
       swap2
       pop
-        /* \"C\":403:411  this.f() */
+      0x20
+      swap1
       returndatasize
+      dup3
+      gt
+      tag_41
+      jumpi
+    tag_42:
         /* \"C\":79:428  contract C... */
       0x1f
+      dup3
       add
       not(0x1f)
       and
-      dup3
+      dup4
       add
       swap1
       0xffffffffffffffff
       dup3
       gt
-      dup4
+      dup5
       dup4
       lt
       or
-      tag_41
+      tag_43
       jumpi
       pop
-      tag_40
       swap3
         /* \"C\":403:411  this.f() */
-      tag_43
+      tag_45
         /* \"C\":392:422  stateVar + this.f() + immutVar */
       tag_39
         /* \"C\":79:428  contract C... */
@@ -301,34 +306,41 @@ sub_0: assembly {
       tag_38
         /* \"C\":79:428  contract C... */
       swap4
+      tag_40
+      swap7
       0x40
       mstore
         /* \"C\":403:411  this.f() */
-      returndatasize
       dup2
       add
       swap1
       tag_6
       jump\t// in
-    tag_43:
+    tag_45:
       swap3
       pop
       swap3
       jump(tag_37)
         /* \"C\":79:428  contract C... */
-    tag_41:
+    tag_43:
       shl(0xe0, 0x4e487b71)
       dup2
       mstore
       mstore(0x04, 0x41)
       0x24
-      swap4
+      swap5
       pop
-      swap2
+      swap3
+      pop
       pop
       pop
       revert
         /* \"C\":403:411  this.f() */
+    tag_41:
+      returndatasize
+      swap2
+      pop
+      jump(tag_42)
     tag_34:
         /* \"C\":79:428  contract C... */
       swap3
@@ -345,10 +357,10 @@ sub_0: assembly {
       swap1
       revert
     tag_32:
-      tag_44
+      tag_46
       tag_4
       jump\t// in
-    tag_44:
+    tag_46:
       jump(tag_33)
         /* \"C\":117:119  41 */
     tag_4:
@@ -373,9 +385,9 @@ sub_0: assembly {
       dup5
       sgt
       and
-      tag_45
+      tag_47
       jumpi
-    tag_46:
+    tag_48:
       shl(0xff, 0x01)
       dup3
       swap1
@@ -383,25 +395,25 @@ sub_0: assembly {
       dup4
       slt
       and
-      tag_47
+      tag_49
       jumpi
       add
       swap1
       jump\t// out
-    tag_47:
-      tag_49
+    tag_49:
+      tag_51
       tag_4
       jump\t// in
-    tag_49:
+    tag_51:
       add
       swap1
       jump\t// out
-    tag_45:
-      tag_50
+    tag_47:
+      tag_52
       tag_4
       jump\t// in
-    tag_50:
-      jump(tag_46)
+    tag_52:
+      jump(tag_48)
         /* \"C\":79:428  contract C... */
     tag_6:
       swap1
@@ -410,12 +422,12 @@ sub_0: assembly {
       swap2
       sub
       slt
-      tag_51
+      tag_53
       jumpi
       mload
       swap1
       jump\t// out
-    tag_51:
+    tag_53:
       pop
       pop
       0x00
@@ -732,34 +744,39 @@ sub_0: assembly {
       return
         /* \"C\":403:411  this.f() */
     tag_36:
-        /* \"D\":91:166  contract D is C(3)... */
       swap1
       swap2
       pop
-        /* \"C\":403:411  this.f() */
+      0x20
+      swap1
       returndatasize
+      dup3
+      gt
+      tag_41
+      jumpi
+    tag_42:
         /* \"D\":91:166  contract D is C(3)... */
       0x1f
+      dup3
       add
       not(0x1f)
       and
-      dup3
+      dup4
       add
       swap1
       0xffffffffffffffff
       dup3
       gt
-      dup4
+      dup5
       dup4
       lt
       or
-      tag_41
+      tag_43
       jumpi
       pop
-      tag_40
       swap3
         /* \"C\":403:411  this.f() */
-      tag_43
+      tag_45
         /* \"C\":392:422  stateVar + this.f() + immutVar */
       tag_39
         /* \"D\":91:166  contract D is C(3)... */
@@ -768,34 +785,41 @@ sub_0: assembly {
       tag_38
         /* \"D\":91:166  contract D is C(3)... */
       swap4
+      tag_40
+      swap7
       0x40
       mstore
         /* \"C\":403:411  this.f() */
-      returndatasize
       dup2
       add
       swap1
       tag_6
       jump\t// in
-    tag_43:
+    tag_45:
       swap3
       pop
       swap3
       jump(tag_37)
         /* \"D\":91:166  contract D is C(3)... */
-    tag_41:
+    tag_43:
       shl(0xe0, 0x4e487b71)
       dup2
       mstore
       mstore(0x04, 0x41)
       0x24
-      swap4
+      swap5
       pop
-      swap2
+      swap3
+      pop
       pop
       pop
       revert
         /* \"C\":403:411  this.f() */
+    tag_41:
+      returndatasize
+      swap2
+      pop
+      jump(tag_42)
     tag_34:
         /* \"D\":91:166  contract D is C(3)... */
       swap3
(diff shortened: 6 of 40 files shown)


libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2522,7 +2522,7 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		if <success> {
 			<?dynamicReturnSize>
 				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
+				returndatacopy(<pos>, 0, <returnSize>)
```
chriseth: This part is only executed in the dynamic case, so you do not need to change anything here.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2560,7 +2560,7 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 
 	// Always use the actual return length, and not our calculated expected length, if returndatacopy is supported.
 	// This ensures it can catch badly formatted input from external calls.
-	if (m_context.evmVersion().supportsReturndata())
+	if (m_context.evmVersion().supportsReturndata() || returnInfo.dynamicReturnSize)
```
chriseth: I think it should be `&&`

chriseth: Oh actually no: If supportsReturndata and it is not dynamic, then we have to use something like `max(<returnSize>, returnsize())`

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2515,21 +2515,28 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		let <end> := <encodeArgs>(add(<pos>, 4) <argumentString>)
 
 		let <success> := <call>(<gas>, <address>, <?hasValue> <value>, </hasValue> <pos>, sub(<end>, <pos>), <pos>, <reservedReturnSize>)
+		let <returnDataSize> := <estimatedReturnSize>
+		<?supportsReturnData>
+			if lt(<returnDataSize>, returndatasize()) {
```
chriseth: I think this should be
```suggestion
			if gt(<returnDataSize>, returndatasize()) {
```
and it should be executed only if we do not have dynamic returndata.

Also I think this is better moved a bit further down. You may even use
```
<?dynamicReturnSize>
// what we alread yhave
<!dynamicReturnSize>
// your new code
</dynamicReturnSize>
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
nishant-sachdeva: Edit : Based on the discussion on the group, I've changed it to  `min(<returnSize>, returndatasize()) `

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2520,16 +2520,25 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		</noTryCall>
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
+			let <returnDataSize> := returndatasize()
```
chriseth: `returndatasize()` is an opcode that is not always available, it can only be used in never evm versions. You need to check `supportsreturnData`

chriseth: It's probably better to start with `<estimatedReturnSize>` instead.

chriseth:
I think this can still be simplified. I'll take another look on Monday!

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2558,13 +2566,9 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
-	// This ensures it can catch badly formatted input from external calls.
-	if (m_context.evmVersion().supportsReturndata())
-		templ("returnSize", "returndatasize()");
-	else
-		templ("returnSize", to_string(returnInfo.estimatedReturnSize));
-
+	templ("returnDataSize", m_context.newYulVariable());
+	templ("estimatedReturnSize", to_string(returnInfo.estimatedReturnSize));
+	templ("supportsReturnData", m_context.evmVersion().supportsReturndata());
 	templ("reservedReturnSize", returnInfo.dynamicReturnSize ? "0" : to_string(returnInfo.estimatedReturnSize));
```
chriseth: Actually I think `reservedReturnSize` and `estimatedReturnSize` are the same. Can you add an assertion to the end of the constructor of `ReturnInfo` so that whenever dynamicReturnSize is true, estimatedReturnSize is zero and then only use `estimatedReturnSize` here?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2520,16 +2520,24 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 		</noTryCall>
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
+			let <returnDataSize> := <estimatedReturnSize>
```
chriseth: Maybe using
```suggestion
			let <returnDataSize>
```
here and then putting the assignment into the `!dynamicReturnSize` part is easier to understand?

libsolidity/codegen/ReturnInfo.cpp
```diff
@@ -54,4 +54,5 @@ ReturnInfo::ReturnInfo(EVMVersion const& _evmVersion, FunctionType const& _funct
 				estimatedReturnSize += retType->decodingType()->calldataEncodedSize();
 		}
 	}
+	solAssert(dynamicReturnSize ? estimatedReturnSize == 0 : true);
```
chriseth:
```suggestion
	if (dynamicReturnSize)
		solAssert(estimatedReturnSize == 0);
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,22 +2514,31 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
+			let <returnDataSize>
 			<?dynamicReturnSize>
 				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
+				<returnDataSize> := returndatasize()
+				returndatacopy(<pos>, 0, <returnDataSize>)
+			<!dynamicReturnSize>
+				<returnDataSize> := <estimatedReturnSize>
+				<?supportsReturnData>
+					if gt(<returnDataSize>, returndatasize()) {
```
chriseth: I somehow find `lt(returndatasize(), <returnDataSize>)` more intuitive here, but we can also leave it like that.

test/libsolidity/semanticTests/various/value_insane.sol
```diff
@@ -21,7 +21,7 @@ contract test {
 // compileViaYul: also
 // ----
 // constructor(), 20 wei ->
-// gas irOptimized: 187835
+// gas irOptimized: 195125
```
chriseth: This is very interesting - why is there such a big code size change?

chriseth: It seems the reason is that it does not inline a particular function anymore (because the function to inline into gets a little larger)

chriseth:
Would be nice if you could take a look here @cameel  - this adds a slight gas increase for the regular case (due to the `returnSize := min(returnSize, returndatasize())` but protects against some special case of "gas abuse" by called contracts. Can you confirm this is actually worth it? :)

test/libsolidity/semanticTests/functionCall/return_size_expectation_mismatch.sol
```diff
@@ -0,0 +1,31 @@
...
+
+contract Test {
+    function test() public returns (uint) {
+        LongReturn lr = new LongReturn();
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        lr.f();
```
chriseth: You have to convert it to a `ShortReturn` before you call it so that the compiler thinks it is returning just a single `bytes32`

cameel:
@chriseth Sure, I'll check the benchmarks and let you know.

cameel:
Here's the benchmark diff:

### `ir-optimize-evm+yul`
|           bleeps | **`+0.23% ❌`** |           `0%` |          `+0%` |
|        elementfi | **`+0.61% ❌`** |                |                |
|              ens |                |                |                |
|           gnosis |  **`+0.3% ❌`** |                |                |
|          gnosis2 | **`+0.15% ❌`** |                |                |
|    pool-together | **`+0.34% ❌`** |  **`+0.4% ❌`** |          `-0%` |
|         prb-math |           `0%` |                |                |
|          trident |                |                |                |
| yield_liquidator |  **`+1.3% ❌`** | **`+1.21% ❌`** | **`+0.01% ❌`** |
|         zeppelin | **`+0.28% ❌`** |                |                |        |               |

### `legacy-optimize-evm+yul`
|          project | bytecode_size | deployment_gas |     method_gas |
|:----------------:|--------------:|---------------:|---------------:|
|           bleeps |          `0%` |           `0%` |           `0%` |
|           colony |          `0%` |                |                |
|        elementfi |          `0%` |           `0%` |           `0%` |
|              ens |          `0%` |           `0%` |           `0%` |
|            euler |          `0%` |          `+0%` |          `+0%` |
|           gnosis |          `0%` |                |                |
|          gnosis2 |          `0%` |                |                |
|  perpetual-pools |          `0%` |          `+0%` | **`+0.01% ❌`** |
|    pool-together |          `0%` |          `-0%` |          `+0%` |
|         prb-math |          `0%` |           `0%` |           `0%` |
|          trident |          `0%` |          `+0%` |          `+0%` |
|          uniswap |          `0%` |          `-0%` |          `+0%` |
| yield_liquidator |          `0%` |           `0%` |           `0%` |
|         zeppelin |          `0%` |           `0%` | **`+0.05% ❌`** |

`!V` = version mismatch

[`summarized-benchmarks-3a36ed31-external_contract_wasting_gas_by_return_value.json`](https://github.com/ethereum/solidity/files/8127145/summarized-benchmarks-3a36ed31-external_contract_wasting_gas_by_return_value.json.txt)
[`summarized-benchmarks-e7d93f8-develop.json`](https://github.com/ethereum/solidity/files/8127146/summarized-benchmarks-e7d93f8-develop.json.txt)

cameel:
I wonder where the differences in legacy come from. It's possible that the test suites are just not deterministic.

For example some input values might be randomly generated on the fly by the test suite. Getting a smaller value (with fewer zeros) for an external call can decrease the call cost since zero bytes in calldata are cheaper.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,22 +2514,31 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
+			let <returnDataSize>
 			<?dynamicReturnSize>
 				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
+				<returnDataSize> := returndatasize()
+				returndatacopy(<pos>, 0, <returnDataSize>)
+			<!dynamicReturnSize>
+				<returnDataSize> := <estimatedReturnSize>
```
cameel: You don't have to separately declare the `returnDataSize` variable as if it was an `if` branch. I think it's clearer (and shorter) this way:
```suggestion
				let <returnDataSize> := returndatasize()
				returndatacopy(<pos>, 0, <returnDataSize>)
			<!dynamicReturnSize>
				let <returnDataSize> := <estimatedReturnSize>
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2558,14 +2567,9 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
 
-	// Always use the actual return length, and not our calculated expected length, if returndatacopy is supported.
-	// This ensures it can catch badly formatted input from external calls.
-	if (m_context.evmVersion().supportsReturndata())
-		templ("returnSize", "returndatasize()");
-	else
-		templ("returnSize", to_string(returnInfo.estimatedReturnSize));
-
-	templ("reservedReturnSize", returnInfo.dynamicReturnSize ? "0" : to_string(returnInfo.estimatedReturnSize));
+	templ("returnDataSize", m_context.newYulVariable());
```
cameel: These placeholder names are really confusing to me here. `returnDataSize`, `dynamicReturnSize` and `estimatedReturnSize` all sound like they're numbers but only the last one is. The first one is a variable name and the second one is a boolean. When reading this code I have to constantly look at the definitions to remind myself what each one represents. Can you rename them?
- `returnDataSize` -> `returndataSizeVar`
- `dynamicReturnSize` -> `isReturndataSizeDynamic`

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,22 +2514,31 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
 			if iszero(<success>) { <forwardingRevert>() }
 		</noTryCall>
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
+			let <returnDataSize>
 			<?dynamicReturnSize>
 				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
+				<returnDataSize> := returndatasize()
+				returndatacopy(<pos>, 0, <returnDataSize>)
```
cameel: It took me a bit to process this code because there are two implicit assumptions here:
1) ~`supportsReturndata`~ `!supportsReturndata` and `dynamicReturnSize` can't be both true.
2) Normally we don't need `returndatacopy()` because the CALL opcode will copy the return data automatically. However, when `dynamicReturnSize` is true, we don't know the size upfront. To handle this, `ReturnInfo` is supposed to set `estimatedReturnSize` to zero and we then need to copy the data out manually (based on `returndatasize()`).

Can you add comments explaining these assumptions?

Also, please add an assert below (in C++ code) that ensures that ~`supportsReturndata`~ `!supportsReturndata` and `dynamicReturnSize` really can't be both true.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2558,14 +2567,9 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
-	// Always use the actual return length, and not our calculated expected length, if returndatacopy is supported.
-	// This ensures it can catch badly formatted input from external calls.
-	if (m_context.evmVersion().supportsReturndata())
-		templ("returnSize", "returndatasize()");
-	else
-		templ("returnSize", to_string(returnInfo.estimatedReturnSize));
-
-	templ("reservedReturnSize", returnInfo.dynamicReturnSize ? "0" : to_string(returnInfo.estimatedReturnSize));
+	templ("returnDataSize", m_context.newYulVariable());
+	templ("estimatedReturnSize", to_string(returnInfo.estimatedReturnSize));
```
cameel: I think that something like `staticReturndataSize` would be a better name on this. The current name does not really convey the nuance that this is expected to be zero in the dynamic case.

test/libsolidity/semanticTests/functionCall/return_size_expectation_mismatch.sol
```diff
@@ -0,0 +1,31 @@
+interface ShortReturn {
+    function f() external pure returns (bytes32);
+}
+contract LongReturn {
+    function f() external pure returns (uint[20] memory) {}
+}
+
+contract Test {
+    function test() public returns (uint) {
+        LongReturn lr = new LongReturn();
```
cameel:
```suggestion
        LongReturn longReturn = new LongReturn();
```

test/libsolidity/semanticTests/functionCall/return_size_expectation_mismatch.sol
```diff
@@ -0,0 +1,31 @@
...
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        ShortReturn(address(lr)).f();
```
cameel: Do we have coverage for the opposite case? I.e. returndata size shorter than expected? If not, please add a test.

test/libsolidity/semanticTests/various/value_insane.sol
cameel: There are quite a few cases with high increases like this. Much more than I'd expect from a change that basically adds a single `if` per external call.

Do you think adjusting inlining heuristics will be enough to deal with it?

cameel:
> Can you confirm this is actually worth it? :)

Well, in the few cases where we can get runtime gas usage for optimized IR in external projects, the difference is minimal or zero so in that regard it seems worth it.

On the other hand, bytecode size increases as much as 1.3% for some projects so not great if they're already struggling to stay under bytecode size limit.

test/libsolidity/semanticTests/functionCall/return_size_expectation_mismatch.sol
nishant-sachdeva: Added a test case in the latest push.

chriseth:
I think we should still add this.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,22 +2514,34 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
+				let <returnDataSizeVar> := returndatasize()
+				returndatacopy(<pos>, 0, <returnDataSizeVar>)
+			<!isReturndataSizeDynamic>
+				let <returnDataSizeVar> := <staticReturndataSize>
+				<?supportsReturnData>
+					if gt(<returnDataSizeVar>, returndatasize()) {
+						<returnDataSizeVar> := returndatasize()
+					}
+				</supportsReturnData>
+			</isReturndataSizeDynamic>
```
cameel: Hmm... putting it all there in the generated code is actually a bit verbose. Let's spread it a bit and shorten it like this:

```c++
			<?isReturndataSizeDynamic>
				// explicit returndata copy needed: we don't know the size upfront so we passed zero to the <call>
				returndatacopy(...)
			<!isReturndataSizeDynamic>
				...
				</supportsReturnData>
				// no explicit returndata copy needed: we know the size upfront and we passed it to the <call>
			</isReturndataSizeDynamic>
```

chriseth: I don't think we should put comments there...
It won't look as convoluted anyway

cameel: True, in Yul it'll be simpler. But what I really wanted was a comment at the C++ level in this particular spot. I think that what's happening here and why is not that obvious at first. Maybe we could instead do something like putting the text between `<?comment>` and `</comment>` and defining `comment` variable to be false so that it does not appear in the output?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,25 +2514,40 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
-			<?dynamicReturnSize>
-				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
-			</dynamicReturnSize>
+			<?isReturndataSizeDynamic>
+				<?comment>
+					// explicit returndata copy needed: we don't know the size upfront so we passed zero to the <call>
+				</comment>
```
cameel: I'd do it like this to make it more concise:
```suggestion
				<?comment> explicit returndata copy needed: we don't know the size upfront so we passed zero to the <call> </comment>
```

cameel: By the way, @chriseth suggested putting the comment at the top (i.e. above the `Whiskers templ(R"(` part). I think the comment is most helpful right here but the top would be a reasonable compromise so I'd be fine with that too.

If you change that, use the previous longer version of the comment that addressed both cases of course (and adjust it to make sense in the new spot).

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,25 +2514,40 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
+			<finalizeAllocation>(<pos>, <returnDataSizeVar>)
 
 			// decode return parameters from external try-call into retVars
-			<?+retVars> <retVars> := </+retVars> <abiDecode>(<pos>, add(<pos>, <returnSize>))
+			<?+retVars> <retVars> := </+retVars> <abiDecode>(<pos>, add(<pos>, <returnDataSizeVar>))
 		}
 	)");
 	templ("revertNoCode", m_utils.revertReasonIfDebugFunction("Target contract does not contain code"));
+	templ("comment", false);
+	// we don't want comments to be added to the Yul Code.
```
cameel: How about this?
```suggestion
	// We can put comments inside <?comment> </comment> so that they will not appear in rendered Yul.
	templ("comment", false);
```
But anyway, it's only relevant if we decide to keep the `<comment>` bit and I guess we won't...

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2514,25 +2514,40 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
 		</noTryCall>
 		<?+retVars> let <retVars> </+retVars>
 		if <success> {
-			<?dynamicReturnSize>
-				// copy dynamic return data out
-				returndatacopy(<pos>, 0, returndatasize())
-			</dynamicReturnSize>
+			<?isReturndataSizeDynamic>
+				<?comment>
+					// explicit returndata copy needed: we don't know the size upfront so we passed zero to the <call>
```
chriseth: Maybe that could just be a C++ comment above or below the template explaining how we deal with the return data?

test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
```diff
@@ -0,0 +1,31 @@
+interface LongReturn {
+    function f() external pure returns (bytes16);
+}
+contract ShortReturn {
+    function f() external pure returns (uint[32] memory) {}
+}
```
cameel: `bytes16` is shorter than `uint[32]` :)

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2558,21 +2573,20 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
...
-	if (m_context.evmVersion().supportsReturndata())
-		templ("returnSize", "returndatasize()");
-	else
-		templ("returnSize", to_string(returnInfo.estimatedReturnSize));
-
-	templ("reservedReturnSize", returnInfo.dynamicReturnSize ? "0" : to_string(returnInfo.estimatedReturnSize));
+	solAssert(
+		!(!m_context.evmVersion().supportsReturndata() && returnInfo.dynamicReturnSize),
+		"supportsReturnData and dynamicReturnSize cannot be both true"
+	);
```
chriseth: Is the following semantically equivalent?
```suggestion
	if (returnInfo.dynamicReturnSize)
		solAssert(m_context.evmVersion().supportsReturndata());
```

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2505,6 +2505,10 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 			appendCode() << "mstore(add(" << m_utils.allocateUnboundedFunction() << "() , " << to_string(returnInfo.estimatedReturnSize) << "), 0)\n";
 	}
 
+	// if Return Data Size is dynamic, explicit returndata copy needed:
```
chriseth:
```suggestion
	// If the expected size of the return data is static, we pass it to the call opcode and the data will just be there.
	// If returndatasize() is available, we set the size to the min of the expected size and returndatasize().
	// If it is dynamic, we don't know the size upfront so we pass a length of zero to the call opcode.
	// We obtain the data via the returndatacopy opcode.
```

cameel: How about this?
```suggestion
    // NOTE: When the expected size of returndata is static, we pass that in to the call opcode an it gets copied automatically.
    // When it's dynamic, we get zero from estimatedReturnSize() instead and then we need an explicit returndatacopy().
```

test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
```diff
@@ -0,0 +1,31 @@
...
+
+contract Test {
+    function test() public returns (uint) {
+     LongReturn longReturn = new LongReturn();
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        ShortReturn(address(longReturn)).f();
```
cameel:
```suggestion
        LongReturn(address(shortReturn)).f();
```

test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
```diff
@@ -0,0 +1,32 @@
+interface LongReturn {
+    function f() external pure returns (uint[20] memory);
+}
+contract ShortReturn {
+    function f() external pure returns (bytes32) {}
+}
+
+contract Test {
+    function test() public returns (uint) {
+     ShortReturn shortReturn = new ShortReturn();
```
chriseth: Indentation

chriseth (dismissed):

test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected.sol
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        return freeMemoryAfter - freeMemoryBefore;
+    }
+}
+// ====
+// EVMVersion: <=homestead
+// compileViaYul: true
+// ----
+// test() -> 0x0500
```
cameel, line 31: I was going to say that there must be something wrong here, because it's allocating 0x500 bytes even though we have the condition to reduce it. Just noticed that it's only for `homestead` though. We should really have coverage for newer EVM on this too. Can you add a variant of this test for `EVMVersion: >homestead` too?

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -2505,6 +2505,8 @@ void IRGeneratorForStatements::appendExternalFunctionCall(
 			appendCode() << "mstore(add(" << m_utils.allocateUnboundedFunction() << "() , " << to_string(returnInfo.estimatedReturnSize) << "), 0)\n";
 	}
 
+	// NOTE: When the expected size of returndata is static, we pass that in to the call opcode an it gets copied automatically.
```
```suggestion
	// NOTE: When the expected size of returndata is static, we pass that in to the call opcode and it gets copied automatically.
```
Looks like there was a typo in my suggestion. Sorry for that!

test/libsolidity/semanticTests/functionCall/return_size_shorter_than_expected_evm_version_after_homestead.sol
```diff
@@ -0,0 +1,32 @@
...
+
+contract Test {
+    function test() public returns (uint) {
+        ShortReturn shortReturn = new ShortReturn();
+        uint freeMemoryBefore;
+        assembly {
+            freeMemoryBefore := mload(0x40)
+        }
+
+        LongReturn(address(shortReturn)).f();
```
cameel, line 18: Let's add a comment explaining that `test() -> FAILURE` is the expected result:
```suggestion
        // This reverts. The external call succeeds but ABI decoding fails due to the returned
        // `bytes32` being much shorter than the expected `uint[20]`.
        LongReturn(address(shortReturn)).f();
```

cameel (approved):

ekpyron:
I'd have added a changelog entry for this one...

cameel:
I assumed it does not need one because it's IR. But you're right, I think we decided recently to start adding them. @nishant-sachdeva can you add it in a new PR?

frangio:
What version was this released in? It still isn't in the changelog.

cameel:
It was released in 0.8.13. Looks like we did not add it to the changelog in the end.

@chriseth Should we actually add it now? We don't usually modify old changelogs but maybe it would make sense to do it here?

cameel:
#13307 adds the changelog entry.

PR #12689, opened 2022-02-17
Fix typos on doc

callezenwaka:


docs/contracts/using-for.rst
```diff
@@ -120,7 +120,7 @@ It is also possible to extend elementary types in that way:
     }
 
 Note that all external library calls are actual EVM function calls. This means that
-if you pass memory or value types, a copy will be performed, even of the
+if you pass memory or value types, a copy will be performed, even on the
```
cameel: I think this is not a typo but both `on` and `of` sounds slightly off to me here. How about `in case of`?
```suggestion
if you pass memory or value types, a copy will be performed, even in case of the
```

franzihei: The commas also look slightly confusing to me. The sentence is a bit hard to read to improving it might make sense in any case!

chriseth (approved):

PR #12692, opened 2022-02-17
Add codeowners file.

chriseth:


Marenz (dismissed):

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+docs/style-guide.rst @fulldecent
```
cameel: Github docs say we should be able to see an icon indicating that fulldecent is the code owner. It's supposed to look like this:
codeowners example

I don't see that when I visit https://github.com/ethereum/solidity/blob/addcodeowners/docs/style-guide.rst. I wonder if it just does not show up for PRs (the docs say that it works per branch though) or if it's because it's a RST file with this weird preview that does not let you view source.

cameel: This apparently follows `.gitignore` rules:

> A CODEOWNERS file uses a pattern that follows most of the same rules used in [gitignore](https://git-scm.com/docs/gitignore#_pattern_format) files

So I think you need an initial slash unless you want to match any `docs/style-guide.rst` at any nesting level in the repo.
```suggestion
/docs/style-guide.rst @fulldecent
```

cameel: I wonder if it might be because of this thing, mentioned in an article I found ([Owning Your CODEOWNERS File](https://medium.com/expedia-group-tech/owning-your-codeowners-file-332e288c1d12)):
> It is not enough to write out a username ( or GitHub Team ) in a `CODEOWNERS` file. The *Collaborators & teams* section of the Repository’s Settings MUST contain everyone mentioned in the `CODEOWNERS `file. GitHub’s docs appear persnickety about this as well,

cameel: And another point from that article:

> While GitHub suggests *write* permissions, in practice I’ve found that *maintain* permissions are what you really need. The whole idea of using a `CODEOWNERS` file is to leverage GitHub’s *protected* branches feature where a Code Owner must review a PR before it can be merged to the protected branch. Maintain permissions seem to be required to hit that merge button on a protected branch, this comes in handy when a Code Owner complains “Why can’t I merge this PR that I just approved?”.

It apprears that being in `CODEOWNERS` will not do much beyond automatically adding @fulldecent as a reviewer. It won't require his review though (unless you configure that separately in settings) or won't let him merge code style PRs on his own. Is that enough of should we figure out an alternative that will give him more control over these docs?

cameel: Looks like this `CODEOWNERS` thing works differently than we imagined it. See for example [Is there any way to add people that dont have write permissions fo the repository as CODEOWNERS?](https://github.community/t/is-there-any-way-to-add-people-that-dont-have-write-permissions-fo-the-repository-as-codeowners/2015), which has an official github response. Looks like the file is more of a mechanism to put even more restrictions on some files rather than alter or lift some restrictions for specific people.

chriseth: So we leave it be?

fulldecent: Thank you. Good as-is now. I'll get pinged which is important.

cameel: > So we leave it be?

I'd say that this would be a good reason for having a separate repo after all.

But since @fulldecent says he's fine with it, I guess we can continue as is. I'm going to approve this and we can revisit when that lack of greater permissions starts getting in the way.

cameel (approved):

PR #12693, opened 2022-02-17
Extend side effects

chriseth:
Allows to retrieve more fine-grained effects on memory and storage for instructions. Needed for #12672

### DIFF
--- a/libevmasm/SemanticInformation.cpp
+++ b/libevmasm/SemanticInformation.cpp
@@ -29,6 +29,125 @@ using namespace std;
 using namespace solidity;
 using namespace solidity::evmasm;
 
+vector<SemanticInformation::Operation> SemanticInformation::readWriteOperations(Instruction _instruction)
+{
+	switch (_instruction)
+	{
+	case Instruction::SSTORE:
+	case Instruction::SLOAD:
+	{
+		assertThrow(memory(_instruction) == Effect::None, OptimizerException, "");
+		assertThrow(storage(_instruction) != Effect::None, OptimizerException, "");
+		Operation op;
+		op.effect = storage(_instruction);
+		op.location = Location::Storage;
+		op.startParameter = 0;
+		// We know that exactly one slot is affected.
+		op.lengthConstant = 1;
+		return {op};
+	}
+	case Instruction::MSTORE:
+	case Instruction::MSTORE8:
+	case Instruction::MLOAD:
+	{
+		assertThrow(memory(_instruction) != Effect::None, OptimizerException, "");
+		assertThrow(storage(_instruction) == Effect::None, OptimizerException, "");
+		Operation op;
+		op.effect = memory(_instruction);
+		op.location = Location::Memory;
+		op.startParameter = 0;
+		if (_instruction == Instruction::MSTORE || _instruction == Instruction::MLOAD)
+			op.lengthConstant = 32;
+		else if (_instruction == Instruction::MSTORE8)
+			op.lengthConstant = 1;
+
+		return {op};
+	}
+	case Instruction::REVERT:
+	case Instruction::RETURN:
+	case Instruction::KECCAK256:
+	case Instruction::LOG0:
+	case Instruction::LOG1:
+	case Instruction::LOG2:
+	case Instruction::LOG3:
+	case Instruction::LOG4:
+	{
+		assertThrow(storage(_instruction) == Effect::None, OptimizerException, "");
+		assertThrow(memory(_instruction) == Effect::Read, OptimizerException, "");
+		Operation op;
+		op.effect = memory(_instruction);
+		op.location = Location::Memory;
+		op.startParameter = 0;
+		op.lengthParameter = 1;
+		return {op};
+	}
+	case Instruction::EXTCODECOPY:
+	{
+		assertThrow(memory(_instruction) == Effect::Write, OptimizerException, "");
+		assertThrow(storage(_instruction) == Effect::None, OptimizerException, "");
+		Operation op;
+		op.effect = memory(_instruction);
+		op.location = Location::Memory;
+		op.startParameter = 1;
+		op.lengthParameter = 3;
+		return {op};
+	}
+	case Instruction::CODECOPY:
+	case Instruction::CALLDATACOPY:
+	case Instruction::RETURNDATACOPY:
+	{
+		assertThrow(memory(_instruction) == Effect::Write, OptimizerException, "");
+		assertThrow(storage(_instruction) == Effect::None, OptimizerException, "");
+		Operation op;
+		op.effect = memory(_instruction);
+		op.location = Location::Memory;
+		op.startParameter = 0;
+		op.lengthParameter = 2;
+		return {op};
+	}
+	case Instruction::STATICCALL:
+	case Instruction::CALL:
+	case Instruction::CALLCODE:
+	case Instruction::DELEGATECALL:
+	{
+		size_t paramCount = static_cast<size_t>(instructionInfo(_instruction).args);
+		vector<Operation> operations{
+			Operation{Location::Memory, Effect::Read, paramCount - 4, paramCount - 3, {}},
+			Operation{Location::Storage, Effect::Read, {}, {}, {}}
+		};
+		if (_instruction != Instruction::STATICCALL)
+			operations.emplace_back(Operation{Location::Storage, Effect::Write, {}, {}, {}});
+		operations.emplace_back(Operation{
+			Location::Memory,
+			Effect::Write,
+			paramCount - 2,
+			paramCount - 1,
+			{}
+		});
+		return operations;
+	}
+	case Instruction::CREATE:
+	case Instruction::CREATE2:
+		return vector<Operation>{
+			Operation{
+				Location::Memory,
+				Effect::Read,
+				1,
+				2,
+				{}
+			},
+			Operation{Location::Storage, Effect::Read, {}, {}, {}},
+			Operation{Location::Storage, Effect::Write, {}, {}, {}}
+		};
+	case Instruction::MSIZE:
+		// This is just to satisfy the assert below.
+		return vector<Operation>{};
+	default:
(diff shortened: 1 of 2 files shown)


libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,128 @@ using namespace std;
...
+	case Instruction::LOG1:
+	case Instruction::LOG2:
+	case Instruction::LOG3:
+	case Instruction::LOG4:
+	{
+		Operation op;
+		if (memory(_instruction) == Effect::Write || storage(_instruction) == Effect::Write)
+			op.effect = Effect::Write;
+		else
+			op.effect = Effect::Read;
```
Marenz: Maybe an assert here that `memory(_instruction) != Effect::None` and the same for storage?

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,128 @@ using namespace std;
 using namespace solidity;
 using namespace solidity::evmasm;
 
+vector<SemanticInformation::Operation> SemanticInformation::readWriteOperations(Instruction _instruction)
```
hrkrshnn, line 32: It's a lot easier to review this if this by making the switch more verbose, i.e., minimize the grouping (from `sstore` to  `log4`) and just return the vector corresponding to the instruction.  Because this code is critical in optimizations, I think the extra verbosity is okay.

libevmasm/SemanticInformation.h
```diff
@@ -45,6 +45,32 @@ struct SemanticInformation
...
+	struct Operation
+	{
+		Location location;
+		Effect effect;
+		/// Start of affected area as an index into the parameters.
+		/// Unknown if not provided.
+		std::optional<size_t> startParameter;
+		/// Length of the affected area as an index into the parameters (if this is an opcode).
+		/// Unknown if neither this nor lengthConstant is provided.
+		std::optional<size_t> lengthParameter;
```
hrkrshnn, line 62: I think this parameter only makes sense for memory. We shouldn't set this for storage.

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,128 @@ using namespace std;
...
+			op.lengthParameter = 3;
+		else if (
+			_instruction == Instruction::CALLDATACOPY ||
+			_instruction == Instruction::CODECOPY ||
+			_instruction == Instruction::RETURNDATACOPY
+		)
+			op.lengthParameter = 2;
+		else if (_instruction == Instruction::SSTORE || _instruction == Instruction::SLOAD)
+			// Storage operations, length is unused / non-sensical
+			op.lengthConstant = 1;
```
hrkrshnn: I don't think we should do this. Using length for storage feels error-prone.

libevmasm/SemanticInformation.h
```diff
@@ -45,6 +45,32 @@ struct SemanticInformation
...
+		/// Length as a constant.
+		/// Unknown if neither this nor lengthArgument is provided.
+		std::optional<size_t> lengthConstant;
+	};
+
+	/// @returns the sequence of read write operations performed by the instruction.
+	/// Order matters.
+	/// For external calls, there is just one unknown read and one unknown write operation,
+	/// event though there might be multiple.
+	static std::vector<Operation> readWriteOperations(Instruction _instruction);
```
hrkrshnn, line 72: How about something like `readWriteOperations` that takes a `FunctionCall` (always builtin) rather than just an instruction. I think the returned object should be more expressive. Maybe a variant `Memory(Expression& offset, Expression& length)` and `Storage(Expression& slot)`.

libevmasm/SemanticInformation.h
chriseth: Other team members complained that it is not set for storage :)

libevmasm/SemanticInformation.cpp
chriseth: But it is actually correct. "Unknown" would be wrong IMO

libevmasm/SemanticInformation.h
chriseth: This is inside libevmasm. I was considering putting it into the scope of yul, but in the end this was easier. We can extend it on the yul side to full function side effects if we want, but then it would be a different structure.

libevmasm/SemanticInformation.cpp
chriseth: Are you saying it would be easier to review if we just had one return per case?

libevmasm/SemanticInformation.cpp
chriseth: Actually in the case of `delegatecall`, storage length is set to unknown.

libevmasm/SemanticInformation.cpp
hrkrshnn: Yeah, one return per case or fewer grouping.

chriseth:
I re-grouped it a bit, it's much better now, thanks for the suggestion, @hrkrshnn

libevmasm/SemanticInformation.h
```diff
@@ -45,6 +45,32 @@ struct SemanticInformation
...
+		Effect effect;
+		/// Start of affected area as an index into the parameters.
+		/// Unknown if not provided.
+		std::optional<size_t> startParameter;
+		/// Length of the affected area as an index into the parameters (if this is an opcode).
+		/// Unknown if neither this nor lengthConstant is provided.
+		std::optional<size_t> lengthParameter;
+		/// Length as a constant.
+		/// Unknown if neither this nor lengthArgument is provided.
+		std::optional<size_t> lengthConstant;
```
ekpyron, line 65: Since this and ``lengthParameter`` are mutually exclusive, I wonder if we should make this a variant, e.g.
``std::variant<Parameter, Constant, Unknown> length;`` - but I'm not sure if that'd make things more or less annoying on the usage end (especially since we need additional types encapsulating a simple ``size_t`` for that).

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,135 @@ using namespace std;
...
+	case Instruction::CALLCODE:
+	case Instruction::DELEGATECALL:
+	{
+		vector<Operation> externalEffects =
+			vector<Operation>{Operation{Location::Storage, Effect::Read, {}, {}, {}}} +
+			(
+				_instruction == Instruction::STATICCALL ?
+				vector<Operation>{} :
+				vector<Operation>{Operation{Location::Storage, Effect::Write, {}, {}, {}}}
+			);
```
ekpyron:
```suggestion
		vector<Operation> externalEffects{Operation{Location::Storage, Effect::Read, {}, {}, {}}};
		if (_instruction != Instruction::STATICCALL)
				externalEffects.emplace_back(Operation{Location::Storage, Effect::Write, {}, {}, {}});
			);
```

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,135 @@ using namespace std;
...
+		size_t paramCount = static_cast<size_t>(instructionInfo(_instruction).args);
+		return vector<Operation>{
+			Operation{
+				Location::Memory,
+				Effect::Read,
+				paramCount - 4,
+				paramCount - 3,
+				{}
+			}
+		} + move(externalEffects) + vector<Operation>{
```
ekpyron: Not sure if making it ``effects`` initialized with the memory and storage reads and then conditionally emplacing the storage write and then (unconditionally) the memory write wouldn't be shorter and simpler...

ekpyron (dismissed):
Only some minor suggestions - I'd be fine with merging.

libevmasm/SemanticInformation.cpp
chriseth: Yes, probably.

libevmasm/SemanticInformation.h
chriseth: Yeah, an optional of a variant would be best, espeacilly with a data enum, but we don't have that in C++...

ekpyron: Well, defining ``struct Paramter { size_t value; };`` and ``struct Constant { size_t value; };`` and using them in a ``std::variant`` kind of does that...
Plus either the ``variant`` inside an ``optional``, or with an additional ``struct Unknown {};`` in the variant, or with ``std::monostate`` in the variant (i.e. "empty")...
But yeah, not sure that's really better in the end...

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,132 @@ using namespace std;
...
+	}
+	case Instruction::STATICCALL:
+	case Instruction::CALL:
+	case Instruction::CALLCODE:
+	case Instruction::DELEGATECALL:
+	{
+		vector<Operation> externalEffects{Operation{Location::Storage, Effect::Read, {}, {}, {}}};
+		if (_instruction != Instruction::STATICCALL)
+				externalEffects.emplace_back(Operation{Location::Storage, Effect::Write, {}, {}, {}});
+			);
```
ekpyron:
```suggestion
```
sorry - missed removing that in my suggestion.

ekpyron (approved):

libevmasm/SemanticInformation.cpp
```diff
@@ -29,6 +29,125 @@ using namespace std;
...
+			Operation{Location::Memory, Effect::Read, paramCount - 4, paramCount - 3, {}},
+			Operation{Location::Storage, Effect::Read, {}, {}, {}}
+		};
+		if (_instruction != Instruction::STATICCALL)
+			operations.emplace_back(Operation{Location::Storage, Effect::Write, {}, {}, {}});
+		operations.emplace_back(Operation{
+			Location::Memory,
+			Effect::Write,
+			paramCount - 2,
+			paramCount - 1,
```
ekpyron, line 124: This is what the fuzzer found in https://github.com/ethereum/solidity/pull/12672#issuecomment-1059328877...
This is tricky, since we cannot assume that it actually writes to the full area, but it *may* write to it...

The safest option is to set this to "unknown".

PR #12696, opened 2022-02-17
Fix unreliable ASTJson tests

### LINKED ISSUE #12621: AST JSON tests do not clear AST for parse-only test cases.

wechman:
fixes #12621

ASTJson test variants are generated based on JSON file availability. Then they are validated against "failAfter" marker in a solidity source file. If the "failAfter" marker is not here all generated test variants are accepted. Otherwise, compiler state in test variant  is compared against "failAfter" marker. If it is later in compilation steps order, the test is interrupted with exception.


Aivolt1 (dismissed):

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -141,13 +141,13 @@ TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefi
...
 		{
-			// Ignore non-fatal analysis errors, we only want to export.
-			if (c.state() > CompilerStack::State::Parsed)
-				continue;
-
-			SourceReferenceFormatter formatter(_stream, c, _formatted, false);
-			formatter.printErrorInformation(c.errors());
-			return TestResult::FatalError;
+			// We just want to export so raise fatal analysis errors only
+			if (c.state() < CompilerStack::State::ParsedAndImported)
```
Marenz: You only kind of invert the if-branches here, the logic basically stays the same (or did I miss something?).

After having a quick look, the problem seems to be that, if we call `parseAndAnalyze(AnalysisPerformed)` (one of the m_variants we have), the `c.state()` after it fails is `ParsedAndImported` which will not be caught by the `if` check here (and neither by the code before those changes).

So I believe the right change is to either change the `Parsed` in the old code to `ParsedAndImported` or to change the check here to `<=`

Marenz (changes requested):

test/libsolidity/ASTJSONTest.cpp
wechman: Actually, I made this change to get rid of "continue" instruction. It was responsible for skipping the rest of the test case (including "runTest" function call). Other changes are minor. Nonetheless, I can revert to "<= Parsed" if you feel that would be easier to read.

Marenz: Oooh I see. Good point. Yeah we should keep that change. I think the other parts of my comment are still valid though :)

Marenz: Thinking a bit more about this, one of the goals with the `continue;` was that we can have tests that do the `parseOnly` part and would otherwise fail in the analysis part. So we basically _want_ to be able to parse partially broken code, so we can export it so that other tools can work with it.

cameel: Yeah, I agree with @Marenz that it looks like it should be `<=`, not `<`. I.e. I see that for `not_existing_import.sol` we get both `not_existing_import.json` and `not_existing_import_parseOnly.json` but we should not be able to produce the former.

cameel: > So we basically want to be able to parse partially broken code, so we can export it so that other tools can work with it.

But can we actually generate AST from code that does not even parse? It makes sense for code that parses but does not pass analysis but not sure it's true for code with parsing errors.

wechman: > Yeah, I agree with @Marenz that it looks like it should be `<=`, not `<`. I.e. I see that for `not_existing_import.sol` we get both `not_existing_import.json` and `not_existing_import_parseOnly.json` but we should not be able to produce the former.

What would be a drawback of keeping "< ParsedAndImported" comparison? I assume it is expected to get AST generated for contracts with non-existing import files. Then I am not sure why to raise an error in tests with successful parsing and missing import files.

`not_existing_import.json` and `not_existing_import_parseOnly.json` are the same because the test ends up at compile step. I do not feel it is really bad. We at least make sure that AST is not updated in case of missing imports.

cameel: The problem is that it works only for this specific case where there are errors in import resolution but not for errors at the analysis stage. I think that we should support both.

All the analysis errors you fixed look like genuine bugs in tests and not something meant to use that feature, so we're apparently missing coverage for this. I think you should add a new test meant specifically to test AST generation from code with analysis bugs.

Marenz: > But can we actually generate AST from code that does not even parse? It makes sense for code that parses but does not pass analysis but not sure it's true for code with parsing errors.

No no, as you said, for parse errors we always want to fail, but we do want to be able to export if parsing was successful even if there is an error in the analysis.

cameel: ok. right.

I think it would be good to have some kind of flag saying if the file is supposed to have analysis errors. We did have quite a few bugs here and I guess the cases with intentional errors will be in the minority. We we should fail on analysis errors unless the test specifically tells us not to.

We could do it through file naming (easier but uglier and a bit too "magical" for my taste) or via comments in the file like we do in other isoltest test suites (nicer but might require adding some machinery here for parsing that).

wechman: @cameel, will not the test fail in AST comparison in case of analysis errors?

cameel: It shouldn't. I'm assuming that generating AST in case of analysis errors does not crash of course - if it does, it's a separate matter and probably a bug to fix. But as long as  it does not, you should just get identical output and the comparison should pass.

wechman: > It shouldn't. I'm assuming that generating AST in case of analysis errors does not crash of course - if it does, it's a separate matter and probably a bug to fix. But as long as it does not, you should just get identical output and the comparison should pass.

I did an investigation on that and found the analysis step extends AST by annotations. In result AST differs between ParsedAndImported and AnalysisPerformed states. Can't we rely on that?
Could you give me any solidity code that pass the parsing step and fails on the analysis so I can confirm we cover that case also?
I basically agree it would be nice to have clear information in test case if errors are expected. But it could complicate tests so I want to be sure this is necessary.

wechman: @cameel @Marenz as you suggested, I extended ASTJSON test with possibility to define if an error is expected or not. Please let me know if this is what you expected. By the way, the change did not complicate the test as much as I expected :)

cameel: > Can't we rely on that?

You mean rely on the fact that further stages only add information so there's nothing we'd only find in `parseOnly`? Even with that assumption the `parseOnly` test has an important function: it was added when we implemented the `stopAfter: parsing` option and is basically a regression test that ensures that exporting at that stage does not crash due to missing information (https://github.com/ethereum/solidity/issues/8739#issuecomment-649667330).

> Could you give me any solidity code that pass the parsing step and fails on the analysis so I can confirm we cover that case also?

Any of the ones you fixed in this PR would do that but if you want something specifically for testing, here's a monstrosity that goes just fine through the parser but makes no sense at all to the analyzer (and humans :)):
```solidity
function g() public;

interface I {
    struct S { S s; }

    function f(E storage e) {
        error E;
        emit E();

        ++c;
        uint calldata c = 123.4;
    }
}
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -49,6 +50,15 @@ namespace
 
 string const sourceDelimiter("==== Source: ");
 
+const std::map<string, CompilerStack::State> compilerStateMap = {
```
cameel:
```suggestion
const map<string, CompilerStack::State> compilerStateMap = {
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -105,6 +116,15 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 			);
 			source = string();
 		}
+		else if (boost::algorithm::starts_with(line, failMarker))
+		{
+			string state = line.substr(failMarker.size(), line.size());
```
cameel: The second argument is the length:
```suggestion
			string state = line.substr(failMarker.size(), line.size() - failMarker.size());
```
I mean, your code works because if you specify a value that exceeds string length it will be truncated. But it still looks a bit misleading. If that was your intention, I'd just leave the second arg at the default:

```suggestion
			string state = line.substr(failMarker.size());
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -105,6 +116,15 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
...
+			state.erase(std::remove_if(state.begin(), state.end(),
+									   [](unsigned char x){return std::isspace(x);}),
+						state.end());
```
cameel: Some wonky indentation here. Also a bit hard to see which argument goes into which function and which opening parentheses match which closing ones.

This is the style we typically use:
```suggestion
			state.erase(
				remove_if(
					state.begin(),
					state.end(),
					[](unsigned char x){ return isspace(x); }
				),
				state.end()
			);
```

cameel: Also, you can just use `boost::algorithm::string::trim()` here. I think your version removes all spaces (not just trailing ones) but I don't think it's a good idea. I think that trimming leading and trailing whitespace is enough.

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -141,13 +161,12 @@ TestCase::TestResult ASTJSONTest::run(ostream& _stream, string const& _linePrefi
...
 		if (!c.parseAndAnalyze(variant.stopAfter))
 		{
-			// Ignore non-fatal analysis errors, we only want to export.
-			if (c.state() > CompilerStack::State::Parsed)
-				continue;
-
-			SourceReferenceFormatter formatter(_stream, c, _formatted, false);
-			formatter.printErrorInformation(c.errors());
-			return TestResult::FatalError;
+			if (!m_expectedFailState || m_expectedFailState != c.state())
```
cameel:
```suggestion
			if (!m_expectedFailState.has_value() || m_expectedFailState.value() != c.state())
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -49,6 +50,15 @@ namespace
...
 string const sourceDelimiter("==== Source: ");
 
+const std::map<string, CompilerStack::State> compilerStateMap = {
+	{"Empty", CompilerStack::State::Empty},
+	{"SourcesSet", CompilerStack::State::SourcesSet},
+	{"Parsed", CompilerStack::State::Parsed},
+	{"ParsedAndImported", CompilerStack::State::ParsedAndImported},
+	{"AnalysisPerformed", CompilerStack::State::AnalysisPerformed},
+	{"CompilationSuccessful", CompilerStack::State::CompilationSuccessful}
+};
```
cameel: This will easily go out of sync with `CompilerStack::State` if we add a new state there. I think it would be best to put it in the file that contains the enum declaration.

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -105,6 +116,15 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
...
 			source = string();
 		}
+		else if (boost::algorithm::starts_with(line, failMarker))
+		{
+			string state = line.substr(failMarker.size(), line.size());
+			state.erase(std::remove_if(state.begin(), state.end(),
+									   [](unsigned char x){return std::isspace(x);}),
+						state.end());
+			solAssert(compilerStateMap.count(state), "Unsupported compiler state in sol file");
+			m_expectedFailState = compilerStateMap.at(state);
```
cameel: Please report an error if the state is already set (i.e. the there was more than one directive in the file).

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -105,6 +116,15 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 			);
 			source = string();
 		}
+		else if (boost::algorithm::starts_with(line, failMarker))
+		{
+			string state = line.substr(failMarker.size(), line.size());
+			state.erase(std::remove_if(state.begin(), state.end(),
+									   [](unsigned char x){return std::isspace(x);}),
+						state.end());
+			solAssert(compilerStateMap.count(state), "Unsupported compiler state in sol file");
```
cameel: This validates input so it should not be an assertion. Please use `BOOST_THROW_EXCEPTION()` as in other parts of the file.

BTW, we should really define a proper exception type here and use `solThrow()` macro for exceptions but that's out of the scope of the PR so just FYI.

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -92,6 +102,7 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 	string source;
 	string line;
 	string const delimiter("// ----");
+	string const failMarker("// Fail on:");
```
cameel: How about we call it `failAfter` to make it easier to notice the connection with `stopAfter`?
```suggestion
	string const failMarker("// failAfter:");
```

cameel (commented):
I have some remarks but overall I like the direction it's going in.

test/libsolidity/ASTJSONTest.cpp
wechman: It makes sense.

test/libsolidity/ASTJSONTest.cpp
wechman: Even so I can see that risk, I do not want to mix code of application and tests.

cameel: So how about a static assert that the enum has exactly 6 items? It will fail once someone adds a new value (which is the main case I'm worried about) and the compiler will point exactly at the place that needs to be updated.

wechman: I replaced the map with functions so the compiler will raise errors if case of code update need.

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -49,6 +50,30 @@ namespace
 
 string const sourceDelimiter("==== Source: ");
 
+string compilerStateToString(CompilerStack::State state)
+{
+	switch(state)
```
chriseth:
```suggestion
	switch (state)
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -49,6 +50,30 @@ namespace
 
 string const sourceDelimiter("==== Source: ");
 
+string compilerStateToString(CompilerStack::State state)
```
chriseth:
```suggestion
string compilerStateToString(CompilerStack::State _state)
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -78,11 +103,26 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
...
-	m_variants = {
-		TestVariant(baseName, CompilerStack::State::Parsed),
-		TestVariant(baseName, CompilerStack::State::AnalysisPerformed),
+	const std::vector<CompilerStack::State> variantCompileStates =
+	{
+		CompilerStack::State::Parsed,
+		CompilerStack::State::AnalysisPerformed
 	};
 
+	for (const auto state : variantCompileStates )
```
chriseth:
```suggestion
	for (const auto state: variantCompileStates)
```

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -78,11 +103,26 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
 	string_view baseName = _filename;
 	baseName.remove_suffix(4);
 
-	m_variants = {
-		TestVariant(baseName, CompilerStack::State::Parsed),
-		TestVariant(baseName, CompilerStack::State::AnalysisPerformed),
+	const std::vector<CompilerStack::State> variantCompileStates =
+	{
```
chriseth:
```suggestion
	const std::vector<CompilerStack::State> variantCompileStates = {
```

test/libsolidity/ASTJSONTest.cpp
wechman: tests added

wechman:
I squashed commits and rebased them on the top of the develop branch. 
@Marenz @cameel @chriseth do you see anything that needs to be addressed before we merge?

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -78,11 +103,25 @@ ASTJSONTest::ASTJSONTest(string const& _filename)
...
+			variant.expectation = readFileAsString(variant.astFilename());
+			boost::replace_all(variant.expectation, "\r\n", "\n");
+			m_variants.push_back(variant);
+		}
+	}
+
+	if (m_variants.empty())
+		BOOST_THROW_EXCEPTION(runtime_error("Missing file with expected result for: \"" + _filename + "\"."));
+
 	ifstream file(_filename);
```
ekpyron, line 121: Is there any reason why all the stuff starting from here cannot be done by ``TestCaseReader`` instead? Not that it's strongly related to the issue at hand, but still...

ekpyron:
What will happen, if I add ``failAfter: Parsed`` to a test, but it still has the old expectation file?
Or conversely, what will happen if ``failAfter: Parsed`` is present, but no ``_parseOnly`` file is present?
It might in general help if the PR description explained the intended new logic.

To me it would seem that, ideally, the solidity source file fully determines what variants are expected with its settings and a run of ``isoltest`` will verify that the appropriate expectation files are there and *only* those - and error out specifically if the set of expectation files doesn't match... but from a quick glance, it looks like the PR runs variants based on which files are present?

But I may have missed some discussion on this.

wechman:
@ekpyron Yes, the idea was to run variants based on which files are present. I am not sure if we can depend on *.sol file only. In fail_after_parsing test we expect a compilation fail after parsing. But even so, in the second test variant, a result AST is enriched by "exportedSymbols". Do you think the second variant makes not sense?

ekpyron:
> @ekpyron Yes, the idea was to run variants based on which files are present. I am not sure if we can depend on *.sol file only. In fail_after_parsing test we expect a compilation fail after parsing. But even so, in the second test variant, a result AST is enriched by "exportedSymbols". Do you think the second variant makes not sense?

I don't think there is any way to make the compiler output an AST on the ``fail_after_parsing`` test without specifying ``--stop-after parsing`` (which will lead to the ``_parseOnly`` AST). So the second variant does not really make sense - it just tests some partial intermediate result that will never actually occur anywhere.

wechman:
Thank you for the quick response! It is all clear now. I will update code to validate the test variants based on the "failAfter" marker.

wechman:
@ekpyron I updated PR with JSONs/variants validation. Please let me know if this is what you expected to see.

test/libsolidity/ASTJSON/fail_after_parsing.sol
```diff
@@ -0,0 +1,16 @@
...
+
+interface I {
+    struct S { S s; }
+
+    function f(E storage e) {
+        error E;
+	emit E();
+
+        ++c;
+	uint calldata c = 123.4;
```
ekpyron: Inconsistent indentation here and above

wechman: problem fixed

test/libsolidity/ASTJSONTest.cpp
```diff
@@ -69,20 +94,30 @@ void replaceTagWithVersion(string& _input)
...
 
-	m_variants = {
-		TestVariant(baseName, CompilerStack::State::Parsed),
-		TestVariant(baseName, CompilerStack::State::AnalysisPerformed),
+	const std::vector<CompilerStack::State> variantCompileStates = {
+		CompilerStack::State::Parsed,
+		CompilerStack::State::AnalysisPerformed
 	};
 
+	for (const auto state: variantCompileStates)
```
ekpyron, line 107:
```suggestion
	for (CompilerStack::State state: {
		CompilerStack::State::Parsed,
		CompilerStack::State::AnalysisPerformed
	})
```
or otherwise: the coding style is ``<type> const`` not ``const <type>`` ;-).

ekpyron (dismissed):
I'm still wondering about https://github.com/ethereum/solidity/pull/12696#discussion_r820516375 - but we don't need to change this in the same PR, even if we want to change it, I guess.

Also I'd still consider *always* expecting and checking a ``parseOnly`` file, but arguably it's also fine to merely check it if it's there...

So since regardless of either of that, this PR is definitely an improvement, I'm approving.

chriseth:
Please squash

wechman:
> Please squash

done

ekpyron (approved):

Issue #10329, opened 2020-11-18
wasm binary for 0.5.17 is missing in solc-bin


cameel:
We only have asm.js build for 0.5.17 so it's not present in [solc-bin/wasm](https://github.com/ethereum/solc-bin/tree/gh-pages/wasm). Looks like wasm rebuilds did not include it because they were done before that version was released.

cameel:
I originally thought that we just need a symlink and wanted to add it now but it turns out we need to actually rebuild it. I'm reporting it to get back to it later.

cameel:
When this is built, remember to submit a PR in remix to mark 0.5.17 as a wasm build (see https://github.com/ethereum/remix-project/pull/1133).

blitz-1306:
Hello there. Is there any blocker for adding missing build to [solc-bin WASM distributions](https://binaries.soliditylang.org/wasm/list.json)? There is [a published version in the NPM](https://www.npmjs.com/package/solc/v/0.5.17) that actually bundles `soljson.js` file in the package. Could somebody check if it fits to the distribution and if so, add it?

blitz-1306:
@chriseth sorry for pinging and applying necromancy to an old issue discussion. Still, would it be possible to do anything with the missing 0.5.17 WASM release in historical builds archive?

cameel:
> Hello there. Is there any blocker for adding missing build to [solc-bin WASM distributions](https://binaries.soliditylang.org/wasm/list.json)?

Mostly a matter of time and lack of urgency. Last time I did mass rebuilds it took some fiddling to get stuff below 0.6.0 to build. 

There is an asm.js binary available in [bin/soljson-v0.5.17+commit.d19bba13.js](https://github.com/ethereum/solc-bin/blob/gh-pages/bin/soljson-v0.5.17%2Bcommit.d19bba13.js). Wasm binaries are faster but asm.js are still fully usable. If you're downloading the compiler in a script, you should really have a fallback to `bin/` if you don't find a specific version in `wasm/`.

> There is [a published version in the NPM](https://www.npmjs.com/package/solc/v/0.5.17) that actually bundles `soljson.js` file in the package. Could somebody check if it fits to the distribution and if so, add it?

It's the asm.js binary I mentioned above.

> Still, would it be possible to do anything with the missing 0.5.17 WASM release in historcal builds archive?

I talked with @chriseth and we decided to try to rebuild it now. So it'll become available soon unless it turns out it needs some wild workarounds to build.

[...]

PR #12697, opened 2022-02-18
Revive wasm rebuild scripts

cameel:
Related to #10329.

This is a bunch of tweaks to make wasm rebuild scripts work again. This is only the minimal change I need to build v0.5.17. It uses the original emscripten image used in the rebuilds ([`trzeci/emscripten:sdk-tag-1.39.3-64bit`](https://hub.docker.com/layers/trzeci/emscripten/sdk-tag-1.39.3-64bit/images/sha256-e21b86e327cae16f3428916e09368fc29e78b4cdc22de340670479712477b6f9)), which is a bit outdated by now.

I tried to switch to the buildpack image we're using now but this turned out to be non-trivial. I managed to solve some of the issues but the blocker is that each compiler release has options for emscripten hard-coded in [`EthCompilerSettings.cmake`](https://github.com/ethereum/solidity/blob/develop/cmake/EthCompilerSettings.cmake). We kept updating them as we were switching to newer emscripten versions but older releases still have the old options. So getting it to work requires patching the source, possibly in a different way for different releases. An additional problem is that some of the changed options might also be associated with some changes in compiler sources and those would also need to be pulled in from later releases. This then comes with the risk of breaking things in a way that will not be apparent until someone tries to use it in some specific environment, like a browser.

Overall this is definitely solvable but more work than I want to spend on it right now. So in the end I went with building it with older emscripten. The biggest downside is that the binary won't be compressed as in #12265 and will not be affected by any emscripten config tweaks we might have introduced after 0.5.17. I pushed what I managed to do to the [`modernize-wasm-rebuild`](https://github.com/ethereum/solidity/tree/modernize-wasm-rebuild) branch and left it at that.

One thing I still want to salvage is to make wasm rebuild use the newer bytecode compare setup I have prepared for solc-bin PR check. This will be a separate PR though.


scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh
```diff
@@ -213,7 +213,9 @@ ln -sf /emsdk_portable/emscripten/bin/* /usr/local/bin
 rm -rf /src
 ln -sf /root/project /src
 
+echo "Install dependencies and upgrade system packages."
 apt-get -qq update >/dev/null 2>&1
+apt-get -qq upgrade >/dev/null 2>&1
```
chriseth, line 219: Do we really need the upgrade? This might make it unstable.

chriseth (approved):
Can be merged like this.

scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh
cameel: Yes. This is an old image and has outdated SSL certs. `wget` could not download boost without it.

Maybe I could update only the certs packages but I'd have to figure out which ones and how - just updating `ca-certificates` was not enough when I tried it back then and I already spent some time trying various things until I just up added this upgrade command.

PR #12701, opened 2022-02-20
Fix running path-related tests from any Windows drive

a3d4:
Currently a couple of tests expect the temp directory and the current working directory to be located on the same drive.

As a result, in a popular Windows setup when the system is installed on `C:` (so the default temp directory is on `C:`) and projects are on `D:`, these tests fail.

This PR adjusts the test expectations.

(I'm not fully satisfied though. In an ideal case we would need to run tests from different drives. But I am not even sure if our Windows CI has `D:`, and if that corner case worth the troubles.)

### DIFF
--- a/libsolidity/interface/FileReader.cpp
+++ b/libsolidity/interface/FileReader.cpp
@@ -276,14 +276,8 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
 	// If the path is on the same drive as the working dir, for portability we prefer not to
 	// include the root name. Do this only for non-UNC paths - my experiments show that on Windows
 	// when the working dir is an UNC path, / does not not actually refer to the root of the UNC path.
-	boost::filesystem::path normalizedRootPath = normalizedPath.root_path();
-	if (!isUNCPath(normalizedPath))
-	{
-		boost::filesystem::path workingDirRootPath = canonicalWorkDir.root_path();
-		// Ignore drive letter case on Windows (C:\ <=> c:\).
-		if (boost::filesystem::equivalent(normalizedRootPath, workingDirRootPath))
-			normalizedRootPath = "/";
-	}
+
+	boost::filesystem::path normalizedRootPath = normalizeCLIRootPathForVFS(normalizedPath, canonicalWorkDir);
 
 	// lexically_normal() will not squash paths like "/../../" into "/". We have to do it manually.
 	boost::filesystem::path dotDotPrefix = absoluteDotDotPrefix(normalizedPath);
@@ -306,6 +300,27 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
 		return "/";
 
 	return normalizedPathNoDotDot;
+}
+
+boost::filesystem::path FileReader::normalizeCLIRootPathForVFS(
+	boost::filesystem::path const& _path,
+	boost::filesystem::path const& _workDir
+)
+{
+	solAssert(_workDir.is_absolute(), "");
+
+	boost::filesystem::path absolutePath = boost::filesystem::absolute(_path, _workDir);
+	boost::filesystem::path rootPath = absolutePath.root_path();
+	boost::filesystem::path baseRootPath = _workDir.root_path();
+
+	if (isUNCPath(absolutePath))
+		return rootPath;
+
+	// Ignore drive letter case on Windows (C:\ <=> c:\).
+	if (boost::filesystem::equivalent(rootPath, baseRootPath))
+		return "/";
+
+	return rootPath;
 }
 
 bool FileReader::isPathPrefix(boost::filesystem::path const& _prefix, boost::filesystem::path const& _path)
--- a/test/libsolidity/interface/FileReader.cpp
+++ b/test/libsolidity/interface/FileReader.cpp
@@ -313,8 +313,9 @@ BOOST_AUTO_TEST_CASE(normalizeCLIPathForVFS_should_not_resolve_symlinks_unless_r
 	if (!createSymlinkIfSupportedByFilesystem(tempDir.path() / "abc", tempDir.path() / "sym", true))
 		return;
 
-	boost::filesystem::path expectedPrefixWithSymlinks = "/" / tempDir.path().relative_path();
-	boost::filesystem::path expectedPrefixWithoutSymlinks = "/" / boost::filesystem::weakly_canonical(tempDir).relative_path();
+	boost::filesystem::path expectedRootPath = FileReader::normalizeCLIRootPathForVFS(tempDir);
+	boost::filesystem::path expectedPrefixWithSymlinks = expectedRootPath / tempDir.path().relative_path();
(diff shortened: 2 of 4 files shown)


a3d4:
@cameel Could you have a look please?

Marenz (dismissed):

Marenz:
LGTM, but better get a second reviewer

libsolidity/interface/FileReader.cpp
```diff
@@ -276,14 +276,8 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
...
-	boost::filesystem::path normalizedRootPath = normalizedPath.root_path();
-	if (!isUNCPath(normalizedPath))
-	{
-		boost::filesystem::path workingDirRootPath = canonicalWorkDir.root_path();
-		// Ignore drive letter case on Windows (C:\ <=> c:\).
-		if (boost::filesystem::equivalent(normalizedRootPath, workingDirRootPath))
-			normalizedRootPath = "/";
-	}
+
+	boost::filesystem::path normalizedRootPath = normalizeCLIRootPathForVHS(normalizedPath, canonicalWorkDir);
```
cameel:
```suggestion
	boost::filesystem::path normalizedRootPath = normalizeCLIRootPathForVFS(normalizedPath, canonicalWorkDir);
```

libsolidity/interface/FileReader.cpp
```diff
@@ -308,6 +302,27 @@ boost::filesystem::path FileReader::normalizeCLIPathForVFS(
 	return normalizedPathNoDotDot;
 }
 
+boost::filesystem::path FileReader::normalizeCLIRootPathForVHS(
+	boost::filesystem::path const& _path,
+	boost::filesystem::path const& _base
```
cameel: I'd explicitly call it "working directory" for clarity. It will still work correctly if it's not work dir (because only its root path matters) but `absolutePath` will not really be the absolute version of `_path`.
```suggestion
	boost::filesystem::path const& _workDir
```
Or, alternatively, maybe it would be better to simply accept the work dir root path here?
```suggestion
	boost::filesystem::path const& _workDirRootPath
```

test/libsolidity/interface/FileReader.cpp
```diff
@@ -313,8 +313,9 @@ BOOST_AUTO_TEST_CASE(normalizeCLIPathForVFS_should_not_resolve_symlinks_unless_r
 	if (!createSymlinkIfSupportedByFilesystem(tempDir.path() / "abc", tempDir.path() / "sym", true))
 		return;
 
-	boost::filesystem::path expectedPrefixWithSymlinks = "/" / tempDir.path().relative_path();
-	boost::filesystem::path expectedPrefixWithoutSymlinks = "/" / boost::filesystem::weakly_canonical(tempDir).relative_path();
+	boost::filesystem::path expectedRootPath = FileReader::normalizeCLIRootPathForVHS(tempDir);
+	boost::filesystem::path expectedPrefixWithSymlinks = expectedRootPath / tempDir.path().relative_path();
+	boost::filesystem::path expectedPrefixWithoutSymlinks = expectedRootPath / boost::filesystem::weakly_canonical(tempDir).relative_path();
```
cameel: One thing that's problematic here that after your change we're using the code under test to generate part of the expectations for that same code :)

But I guess, since this particular test is not checking root normalization specifically, it should be fine and is probably less ugly than any alternative.

cameel (changes requested):
Just some naming corrections needed. Other than that it looks ok. Thanks!

cameel:
> (I'm not fully satisfied though. In an ideal case we would need to run tests from different drives. But I am not even sure if our Windows CI has `D:`, and if that corner case worth the troubles.)

Yeah, thoroughly testing stuff that relies on real system-specific locations (Windows drives and UNC in particular) is really problematic. I don't have a good solution and, to be honest, even what we have is a bit heavy for how little actual code it's testing. I.e. most of the corner cases really come from Boost and the FS and not our code and what we're really testing are our assumptions about these things.

I think CI does have more than one drive actually and it's also possible to create new drives with `subst` command. Even so, assuming that we have enough permissions to write there or that we can safely create temp files there without overwriting anything is a bit risky. This stuff is too invasive so I'd prefer not to go that route. I think that adjustments from this PR are good enough.

libsolidity/interface/FileReader.cpp
a3d4: Thanks! Done.

libsolidity/interface/FileReader.cpp
a3d4: OK. The original parameter name was as for [`boost::filesystem::absolute`](https://www.boost.org/doc/libs/1_68_0/libs/filesystem/doc/reference.html#absolute)

> Or, alternatively, maybe it would be better to simply accept the work dir root path here?

I thought about it too. But that would require cutting the root in several places on  the caller side.

Maybe we can say that `_workDir` defines the context on normalization, and taking into account just the root is an implementation detail.

But I would not mind `_workDirRootPath = boost::filesystem::current_path().root_path()` either :).

test/libsolidity/interface/FileReader.cpp
a3d4: Yes, we partially forge expectations, and I do not like it either. But I hope it is still a reasonable pragmatic fix. The entire topic (starting with [a dangerous `current_path()`](https://www.boost.org/doc/libs/1_68_0/libs/filesystem/doc/reference.html#current_path), is somewhat sensitive and fragile.

> I.e. most of the corner cases really come from Boost 

BTW, why do we use `boost::filesystem` and not `std::filesystem`?

cameel: Yeah, I think it's acceptable. Solves the problem with temp dir location on Windows so seems worth it to me.

And as much as `current_path()` is dangerous, we have to live with it. Not sure if there's even an alternative.

As for `std::filesystem`, I think it's just it wasn't available when the project started and now we're rolling with it. I'd be fine with either but unfortunately they have slight differences in behavior that would affect normalization (see for example https://github.com/boostorg/filesystem/issues/88#issuecomment-435102621). At least `FileReader` is now well covered so maybe it would not be as risky as it used to be but I can't vouch for anything using `boost::filesystem` outside of that code.

libsolidity/interface/FileReader.cpp
cameel: > The original parameter name was as for [boost::filesystem::absolute](https://www.boost.org/doc/libs/1_68_0/libs/filesystem/doc/reference.html#absolute)

Ah, ok. For `absolute()` it makes sense. For our normalization we assume work dir though.

> I thought about it too. But that would require cutting the root in several places on the caller side.

It's just a call to `root_path()` so I don't think it's that bad. Having the function do it is more convenient of course but then it's getting more info than it really needs.

In any case, I'm fine with either.

cameel (approved):

chriseth:
If this is just a refactoring for production code, then please merge. Otherwise, change the title of the pull request and add a changelog entry.

cameel:
Yes, just a refactor. Actual logic changes only in tests.

Issue #12515, opened 2022-01-11
Overhaul and modernize CMake setup.


ekpyron:
Our cmake setup is an organically grown mess that today violates pretty much every best practice for writing clean modern cmake code.

One example is the entire mess around ``EthCompilerSettings``. What we should really do is to define an ``INTERFACE`` library to collect flags and pass them to the appropriate targets. Stuff like ``fdiagnostics-color`` has no business being defined there at all, that belongs in the user's configuration. Things like ``CMAKE_CXX_FLAGS_DEBUG`` should remain untouched. Etc. pp...

From all the issues and questionable things I see in this file alone, I'd actually tend to say we should just rewrite the cmake logic entirely - that can be done quickly and cleanly.

For the record: this was triggered by ``$<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>`` in ``EthCompilerSettings`` breaking for @christianparpart, after which we realized that our cmake setup is broken for ninja builds entirely right now. But the entire mechanism there is one single hack that should be replaced :-). 
https://github.com/ethereum/solidity/pull/11422 tried to work on this part once, but to be honest the result was not much better than what we have and also contrary to my understanding of modern cmake best practices.

christianparpart:
Thanks @ekpyron for filing this ticket! We could gather the requirements of what we expect and then see how we can best achieve that. (I'm not saying we should move away from CMake btw - whatever fits best) :)

ekpyron:
We should most definitely stay with cmake - it's the only sane choice. But we should use it properly.

cameel:
> Stuff like `fdiagnostics-color` has no business being defined there at all, that belongs in the user's configuration.

What do you mean by user configuration? Flags passed to CMake? If so then I think we should keep that one. The thing is that by default output is not colored and hard to read. I think it should be enabled by default and that's AFAIK the way to do it.

ekpyron:
The flag is a question of taste and depends on your situation. You may run stuff in an environment, in which colored output is harmful. If you think colored output is more readable in general, then the flag should be in your ``C(XX)FLAGS`` at all times and cmake will automatically pick it up. Stuff like this has nothing to do with the actual build process, so we shouldn't mess with it.

But it's only one of a whole bunch of issues and I don't really mind in this particular case, just saying from my perspective this is also definitely bad style.

ekpyron:
[...]

PR #10359, opened 2020-11-19
Enable the -Wextra-semi warning

axic:
Picked from #10264
Part of #6905.

Feel free to close if disagreeing with it.

PR #12702, opened 2022-02-21
cmake/EthCompilerSettings.cmake: generator expression fix for ninja.

aarlt:
If Ninja is used a CMake generator, it is not possible to build Solidity due to an error `bad $-escape (literal $ must be written as $$)`.

```
➜  build git:(develop) CMAKE_GENERATOR=Ninja cmake ..
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
...
-- Performing Test $<$<COMPILE_LANGUAGECXX>-Wextra-semi>
CMake Error:
  Running

   '/opt/homebrew/bin/ninja' '-C' '/tmp/new/solidity/build/CMakeFiles/CMakeTmp' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:49: bad $-escape (literal $ must be written as $$)





CMake Error at /opt/homebrew/Cellar/cmake/3.22.2/share/cmake/Modules/Internal/CheckSourceCompiles.cmake:92 (try_compile):
  Failed to generate test project build system.
  /opt/homebrew/Cellar/cmake/3.22.2/share/cmake/Modules/Internal/CheckCompilerFlag.cmake:71 (cmake_check_source_compiles)
  /opt/homebrew/Cellar/cmake/3.22.2/share/cmake/Modules/CheckCXXCompilerFlag.cmake:40 (cmake_check_compiler_flag)
  cmake/EthCheckCXXCompilerFlag.cmake:16 (check_cxx_compiler_flag)
  cmake/EthCompilerSettings.cmake:57 (eth_add_cxx_compiler_flag_if_supported)
  CMakeLists.txt:52 (include)


-- Configuring incomplete, errors occurred!
```

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -54,9 +54,10 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
 	add_compile_options(-Wsign-conversion)
 	add_compile_options(-Wconversion)
 
-	eth_add_cxx_compiler_flag_if_supported(
-		$<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
-	)
+	check_cxx_compiler_flag(-Wextra-semi WEXTRA_SEMI)
+	if(WEXTRA_SEMI)
+		add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)
+	endif()
 	eth_add_cxx_compiler_flag_if_supported(-Wfinal-dtor-non-final-class)
 	eth_add_cxx_compiler_flag_if_supported(-Wnewline-eof)
 	eth_add_cxx_compiler_flag_if_supported(-Wsuggest-destructor-override)


cmake/EthCompilerSettings.cmake
```diff
@@ -55,7 +55,7 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
 	add_compile_options(-Wconversion)
 
 	eth_add_cxx_compiler_flag_if_supported(
-		$<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
+        $$<$$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
```
chfast: This think is CMake Generator Expression and should be resolved during CMake generation phase and never reach the build tool (ninja). The "if supported" check probably does not work in this case because it will be checking the expression verbose (generation phase happens later). Simpler fix can be replacing eth_add_cxx_compiler_flag_if_supported with add_compile_option() and generation expression additionally checking if the compiler is Clang.

aarlt: That's true! ninja should never see any generator expression. That means, that this fix is not really solving the issue. Somehow CMake seem not to evaluate the generator expression correctly, if ninja is used.

aarlt:
For now I just set this PR as draft until we solved the underlying problem correctly: somehow cmake seem not to evaluate the generator expression correctly, if ninja is used - ninja should never see these `$`.

cmake/EthCompilerSettings.cmake
chfast: This should fix it if you want to keep the generic compiler check.
```cmake
    check_cxx_compiler_flag(-Wextra-semi WEXTRA_SEMI)
    if(WEXTRA_SEMI)
        add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)
    endif()

Comment (aarlt) [reply cmake/EthCompilerSettings.cmake]: @chfast thx! your fix seem to work!

**Comment by aarlt** [comment]
I changed this PR with the fix suggested by @chfast.

File: cmake/EthCompilerSettings.cmake (hunk 1 above)
Comment (chfast) [reply cmake/EthCompilerSettings.cmake]: For the record, the eth_add_cxx_compiler_flag_if_supported is doing pretty much the same just in this case it would erroneously pass `check_cxx_compiler_flag($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)`.

**Review by chfast — APPROVED** [review]

**Comment by chfast** [comment]
BTW, I recommend using Ninja on CI. It is much more pedantic than Makefiles.

**Comment by aarlt** [comment]
> BTW, I recommend using Ninja on CI. It is much more pedantic than Makefiles.

I was actually thinking the same! I will bring that up in a next meeting. Thx @chfast!

**Comment by cameel** [comment]
Using ninja sounds fine to me. The change should be pretty transparent, I don't think we rely on Makefiles in any major way (other than in this expression by accident). We should keep some jobs on makefiles though just to make sure it still works.

File: cmake/EthCompilerSettings.cmake (hunk 1 above)
Comment (cameel) [reply cmake/EthCompilerSettings.cmake]: I wonder if we really have to check for C++ here? If not, we could just do
```cmake
eth_add_cxx_compiler_flag_if_supported(-Wextra-semi)
```
I suppose there must be some problem with doing it that way or @axic would have used it already in #10359.

Alternatively, I think this should work:
```cmake
if ($<COMPILE_LANGUAGE:CXX>)
    eth_add_cxx_compiler_flag_if_supported(-Wextra-semi)
endif()
```

In any case, the solution suggested by @chfast is good too, but if possible I think it would be better to have all options go through `eth_add_cxx_compiler_flag_if_supported()` since it does some extra processing.

cmake/EthCompilerSettings.cmake
```diff
@@ -54,9 +54,10 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
 	add_compile_options(-Wsign-conversion)
 	add_compile_options(-Wconversion)
 
-	eth_add_cxx_compiler_flag_if_supported(
-		$<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
-	)
+	check_cxx_compiler_flag(-Wextra-semi WEXTRA_SEMI)
+	if(WEXTRA_SEMI)
+		add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)
+	endif()
```
cameel, line 60: Can you check if this would work?
```suggestion
        if ($<COMPILE_LANGUAGE:CXX>)
            eth_add_cxx_compiler_flag_if_supported(-Wextra-semi)
        endif()
```

chfast: This will not work. The `if` is evaluated at "configure" time, the generator expressions are evaluated at "generation" time. How this case works in practice is: at generation time the compile flags are applied to each source file and then it is known what language the file is. During configuration you at most has "enabled" languages (C and C++ in solidity case probably).

cmake/EthCompilerSettings.cmake
chfast: GCC does not like compiling C files with this flag. https://godbolt.org/z/obKK9KT5x.

cmake/EthCompilerSettings.cmake
cameel: Oh, thanks, now I see where the problem is.

In that case how about adding an optional argument to `eth_add_cxx_compiler_flag_if_supported()` to specify the language and moving the generator expression inside?

cmake/EthCompilerSettings.cmake
cameel: I mean, I expected CMake to not report this flag as supported by the compiler in case of C. But I now see why I was wrong - I just realized it's `check_**cxx**_compiler_flag()`, i.e. we only really check if the C++ compiler supports it and ignore the C compiler while the flag is being set for both.

I see that CMake has `check_c_compiler_flag()` too. In that case wouldn't a more robust solution be to change `eth_add_cxx_compiler_flag_if_supported()` to handle C and C++ separately, each with appropriate function? Actually, its name is a bit misleading because it really adds a flag for both C and C++. So maybe we should also rename it to `eth_add_compiler_flag_if_supported()`?

ekpyron:
IMHO the "proper" solution for this would be to not use ``add_compile_option`` ever at all, but to use explicit ``INTERFACE`` targets, equip them with the compiler options using ``target_compile_options`` and explicitly link them to our other targets (except evmc, which we can handle separately, resp. with a shared set of C-compatible options in another interface target - I think evmc has the only C files we compile).
The idea of https://github.com/ethereum/solidity/issues/12515 was, among others, to do just that...

leonardoalt:
What's the conclusion here?

leonardoalt:
@aarlt @cameel @ekpyron

ekpyron (approved):

PR #12703, opened 2022-02-22
scripts/common.sh: 'diff --color' not supported by macOS.

aarlt:
`diff --color` is not supported in macOS.

```
➜  ~ diff --color=auto
diff: unrecognized option `--color=auto'
diff: Try `diff --help' for more information.
```

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -180,7 +180,7 @@ function diff_values
     shift
     shift
 
-    diff --color=auto --unified=0 <(echo "$value1") <(echo "$value2") "$@"
+    diff --unified=0 <(echo "$value1") <(echo "$value2") "$@"
 }
 
 function safe_kill


scripts/common.sh
```diff
@@ -180,7 +180,12 @@ function diff_values
     shift
     shift
 
-    diff --color=auto --unified=0 <(echo "$value1") <(echo "$value2") "$@"
+    local color_option="--color=auto"
+    if [[ $OSTYPE == 'darwin'* ]]; then
+        color_option=
+    fi
+    # shellcheck disable=SC2086
+    diff ${color_option} --unified=0 <(echo "$value1") <(echo "$value2") "$@"
```
cameel: Damn you macOS! that's why we can't have nice things :)

Let's just remove the `--color` option. It's not essential here, not worth doing an error-prone OS detection IMO. I used it because it makes the diff nicer but it's not like we'll see that diff often. It's currently used only for comparing one- and two-step compilation via IR in one command-line test.

aarlt: Ah ok!

cameel (approved):

PR #12704, opened 2022-02-22
Add --import-asm-json input mode.

aarlt:
Replaces #11807.
Replaced by #12799,#12809 &  #12834.


aarlt:
Right now the EVM Assembly JSON import/export fails because of the following contracts.

`solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol`
`libsolidity/syntaxTests/immutable/inheritance_virtual_functions_direct_call.sol`
`solidity/test/libsolidity/syntaxTests/parsing/constructor_super.sol`

After exporting the Assembly JSON of these contracts, the import fails for all of them with the same exception:
```
Uncaught exception:
/solidity/libevmasm/Assembly.cpp(992): Throw in function const solidity::evmasm::LinkerObject &solidity::evmasm::Assembly::assemble() const
Dynamic exception type: boost::wrapexcept<solidity::evmasm::AssemblyException>
std::exception::what: Duplicate tag position.
[solidity::util::tag_comment*] = Duplicate tag position.
```

aarlt:
For `solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol` and `solidity/test/libsolidity/syntaxTests/parsing/constructor_super.sol`  I noted that the exported asm contains a strange `PUSH [tag]`.

`solc solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol --asm-json`

```json
{
...
    {
      "begin": 172,
      "end": 183,
      "name": "PUSH [tag]",
      "source": 0,
      "value": "18446744073709551618"  // = 0x10000000000000002
    }
...
}
```

If I look at the code via `--opcodes` I don't see a push with that strange value.

`solc solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol --opcode`

```
======= solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol:B =======
Opcodes:
PUSH1 0xA0 PUSH1 0x40 MSTORE PUSH1 0x3 PUSH1 0x80 SWAP1 DUP2 MSTORE POP CALLVALUE DUP1 ISZERO PUSH1 0x17 JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP PUSH1 0x80 MLOAD PUSH1 0x3F PUSH1 0x2D PUSH1 0x0 CODECOPY PUSH1 0x0 POP POP PUSH1 0x3F PUSH1 0x0 RETURN INVALID PUSH1 0x80 PUSH1 0x40 MSTORE PUSH1 0x0 DUP1 REVERT INVALID LOG2 PUSH5 0x6970667358 0x22 SLT KECCAK256 0x5D 0x2F 0xC4 EXTCODESIZE 0xE4 0xEF 0xFB SWAP13 0xD3 DUP13 JUMPDEST 0xDD PUSH13 0x4C21DE3CFCB927EF96FEFFDB64 DUP2 0xEA PUSH1 0xA9 0x5E SWAP3 PUSH5 0x736F6C6343 STOP ADDMOD 0xC STOP CALLER

======= solidity/test/libsolidity/syntaxTests/immutable/inheritance_virtual_functions_super.sol:C =======
Opcodes:
PUSH1 0xA0 PUSH1 0x40 MSTORE PUSH1 0x3 PUSH1 0x80 SWAP1 DUP2 MSTORE POP CALLVALUE DUP1 ISZERO PUSH1 0x17 JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP PUSH1 0x27 PUSH1 0x2C PUSH1 0x20 SHL PUSH1 0x9 OR PUSH1 0x20 SHR JUMP JUMPDEST POP PUSH1 0x36 JUMP JUMPDEST PUSH1 0x0 PUSH1 0x80 MLOAD SWAP1 POP SWAP1 JUMP JUMPDEST PUSH1 0x80 MLOAD PUSH1 0x67 PUSH2 0x4E PUSH1 0x0 CODECOPY PUSH1 0x0 PUSH1 0xD ADD MSTORE PUSH1 0x67 PUSH1 0x0 RETURN INVALID PUSH1 0x80 PUSH1 0x40 MSTORE PUSH1 0x0 DUP1 REVERT JUMPDEST PUSH1 0x0 PUSH32 0x0 SWAP1 POP SWAP1 JUMP INVALID LOG2 PUSH5 0x6970667358 0x22 SLT KECCAK256 CALL CALLVALUE 0xEE 0xC8 SELFBALANCE 0xBC 0xB7 0xBB DUP10 PUSH10 0xE660D9BC6E852F5FAADE 0x2B 0xD7 KECCAK256 0x4E 0xA6 0xC0 0xA5 DUP14 0xBB PUSH24 0x946D64736F6C634300080C00330000000000000000000000
```

I guess that I miss something here.

@chriseth @ekpyron @cameel any ideas?

chriseth:
I assume the question above has been resolved, right?

solc/CommandLineParser.cpp
```diff
@@ -451,6 +455,13 @@ void CommandLineParser::parseOutputSelection()
 			CompilerOutputs::componentName(&CompilerOutputs::ewasm),
 			CompilerOutputs::componentName(&CompilerOutputs::ewasmIR),
 		};
+	  static set<string> const evmAssemblyJsonImportModeOutputs = {
```
stackenbotten: Coding style error

solc/CommandLineParser.cpp
```diff
@@ -451,6 +455,13 @@ void CommandLineParser::parseOutputSelection()
...
+			CompilerOutputs::componentName(&CompilerOutputs::asm_),
+			CompilerOutputs::componentName(&CompilerOutputs::binary),
+			CompilerOutputs::componentName(&CompilerOutputs::binaryRuntime),
+			CompilerOutputs::componentName(&CompilerOutputs::opcodes),
+			CompilerOutputs::componentName(&CompilerOutputs::asmJson),
+	  };
```
stackenbotten: Coding style error

scripts/ImportExportTest.sh
```diff
@@ -145,6 +145,40 @@ function testImportExportEquivalence {
                                 done
                             done
 
+                            if ! $SOLC --combined-json asm --pretty-json --import-asm-json expected.asm > obtained_direct_import_export.json 2> obtained_direct_import_export.error
```
aarlt: `--combined-json asm` should not be used here, instead just use `--asm-json`.

solc/CommandLineParser.cpp
```diff
@@ -451,6 +455,13 @@ void CommandLineParser::parseOutputSelection()
 			CompilerOutputs::componentName(&CompilerOutputs::ewasm),
 			CompilerOutputs::componentName(&CompilerOutputs::ewasmIR),
 		};
+		static set<string> const evmAssemblyJsonImportModeOutputs = {
```
aarlt, line 458: @chriseth @ekpyron does the restriction of the output modes make sense for you too?

solc/CommandLineParser.cpp
```diff
@@ -941,9 +959,27 @@ void CommandLineParser::processArgs()
 	for (auto& option: conflictingWithStopAfter)
 		checkMutuallyExclusive({g_strStopAfter, option});
 
+	array<string, 11> const conflictingWithAsmJsonImport{
```
aarlt, line 962: @chriseth @ekpyron and what about this? does it make sense to not allow those?

aarlt:
Regarding the mismatch in the optimised byte-code:

- Export assembly json
  `solc/solc <constract> --asm-json > /tmp/asm.json`

- Equivalence needed for - * _I was assuming this_ *
  `solc/solc <contract> —optimize --opcodes` **(A)**
  `solc/solc --import-asm-json /tmp/asm.json --opcodes --optimize` **(B)**

Basically for these compilations the used optimiser(-steps) are different. That's why we see a difference in the byte-code. **(A)** uses class `CompilerStack`, using`AssemblyStack::optimize()`, and that uses `OptimiserSuite`, where **(B)** uses `Assembly`, `Assembly::optimise(..)`, `Assembly::optimiseInternal(..)`. The final optimisation steps differ in both cases. 

I think it totally make sense to allow `--import-asm-json` with `--optimize`, but not to see it like that we expect that the resulting byte-code of **(A)** and **(B)** need to be the same.

@chriseth @ekpyron What do you think?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
-					try
+	if (!m_evmAssemblyJson.empty())
+	{
+		solAssert(m_importedSources, "");
+		solAssert(m_evmAssemblyJson.size() == 1, "");
+
+		string const evmSourceName = m_evmAssemblyJson.begin()->first;
+		Json::Value const evmJson = m_evmAssemblyJson.begin()->second;
+
+		evmasm::Assembly::OptimiserSettings optimiserSettings;
```
aarlt, line 685: I think I saw somewhere a optimiser-settings transform function.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
+		optimiserSettings.runConstantOptimiser = m_optimiserSettings.runConstantOptimiser;
+		optimiserSettings.runDeduplicate = m_optimiserSettings.runDeduplicate;
+		optimiserSettings.runInliner = m_optimiserSettings.runInliner;
+		optimiserSettings.runJumpdestRemover = m_optimiserSettings.runJumpdestRemover;
+		optimiserSettings.runPeephole = m_optimiserSettings.runPeephole;
+
+		m_contracts[evmSourceName].evmAssembly = make_shared<evmasm::Assembly>(true, evmSourceName);
+		m_contracts[evmSourceName].evmAssembly->loadFromAssemblyJSON(m_evmAssemblyJson[evmSourceName]);
+		if (m_optimiserSettings.enabled)
+			m_contracts[evmSourceName].evmAssembly->optimise(optimiserSettings);
```
aarlt, line 698: @chriseth @ekpyron I do the optimisation here like this. Does it make sense?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
+		m_contracts[evmSourceName].object = m_contracts[evmSourceName].evmAssembly->assemble();
+
+		m_contracts[evmSourceName].evmRuntimeAssembly = make_shared<evmasm::Assembly>(false, evmSourceName);
+		m_contracts[evmSourceName].evmRuntimeAssembly->setSources(m_contracts[evmSourceName].evmAssembly->sources());
+		m_contracts[evmSourceName].evmRuntimeAssembly->loadFromAssemblyJSON(m_evmAssemblyJson[evmSourceName][".data"]["0"], false);
+		if (m_optimiserSettings.enabled)
+			m_contracts[evmSourceName].evmRuntimeAssembly->optimise(optimiserSettings);
```
aarlt, line 705: @chriseth @ekpyron Similar optimisation as above. But for the runtime part. Does it make sense?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
+
+		m_contracts[evmSourceName].evmAssembly = make_shared<evmasm::Assembly>(true, evmSourceName);
+		m_contracts[evmSourceName].evmAssembly->loadFromAssemblyJSON(m_evmAssemblyJson[evmSourceName]);
+		if (m_optimiserSettings.enabled)
+			m_contracts[evmSourceName].evmAssembly->optimise(optimiserSettings);
+		m_contracts[evmSourceName].object = m_contracts[evmSourceName].evmAssembly->assemble();
+
+		m_contracts[evmSourceName].evmRuntimeAssembly = make_shared<evmasm::Assembly>(false, evmSourceName);
+		m_contracts[evmSourceName].evmRuntimeAssembly->setSources(m_contracts[evmSourceName].evmAssembly->sources());
+		m_contracts[evmSourceName].evmRuntimeAssembly->loadFromAssemblyJSON(m_evmAssemblyJson[evmSourceName][".data"]["0"], false);
```
aarlt, line 703: @chriseth @ekpyron Does it make sense to load the runtime assembly hardcoded from `".data"[0]` here? Not sure whether we should do this dynamically - e.g. that the correct run time will be taken automatically. I guess for this I would need to analyse the "deploy" code. Not sure whether we really want this. What are your opinions?

solc/CommandLineParser.cpp
chriseth: @cameel can you answer this?

solc/CommandLineParser.cpp
chriseth: Again @cameel

solc/CommandLineParser.cpp
cameel: Yeah, looks sensible. I think they all should be possible to generate from imported assembly.

My only doubt would be about `binaryRuntime`. If it comes from compiling a contract, we do have the construction/runtime code split. Do we always have it with arbitrary assembly though?

solc/CommandLineParser.cpp
cameel: Not completely sure about `storageLayout`. So we lose that info at assembly level?
And `--gas` - so the estimator must be working at the Solidity level I assume and not by looking at opcodes?

By the way, putting it here seems redundant to me. You already do this check in `parseOutputSelection()` so this will never fail. `conflictingWithStopAfter` was here because it's just a regular option, not an input mode. And `conflictingWithStopAfter` should eventually be integrated into `validOptionInputModeCombinations` above anyway - this was being refactored in #12166 but it's stalled for now.

libevmasm/Assembly.cpp
```diff
@@ -222,13 +222,16 @@ string Assembly::assemblyString(
 	return tmp.str();
 }
 
-Json::Value Assembly::createJsonValue(string _name, int _source, int _begin, int _end, string _value, string _jumpType)
+Json::Value Assembly::createJsonValue(
+	string _name, int _sourceIndex, size_t _modifierDepth, int _begin, int _end, string const& _value, string const& _jumpType)
```
chriseth, line 226: plaese break the line after each `,`

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+
+	auto updateUsedTags = [&](u256 const& data) {
+		auto tag = static_cast<unsigned>(data);
+		if (this->m_usedTags <= tag)
+			this->m_usedTags = tag + 1;
+	};
+
+	auto updateImmutables = [&](string const& _immutableName) -> h256 {
+		h256 hash(util::keccak256(value));
+		this->m_immutables[hash] = _immutableName;
```
chriseth, line 268:
```suggestion
		m_immutables[hash] = _immutableName;
```

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+	auto updateLibraries = [&](string const& _libraryName) -> h256 {
+		h256 hash(util::keccak256(_libraryName));
+		this->m_libraries[hash] = _libraryName;
+		return hash;
+	};
+
+	SourceLocation location;
+	location.start = begin;
+	location.end = end;
+	if (srcIndex > -1 && srcIndex < (int) sources().size())
```
chriseth, line 281:
```suggestion
	if (srcIndex > -1 && srcIndex < static_cast<int>(sources().size()))
```

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+
+	AssemblyItem result(0);
+
+	if (c_instructions.find(name) != c_instructions.end())
+	{
+		AssemblyItem item{c_instructions.at(name), location};
+		item.m_modifierDepth = modifierDepth;
+		if (!value.empty())
+			item.setJumpType(value);
+		result = item;
```
chriseth, line 292:
```suggestion
		return item;
```

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+				item.setJumpType(jumpType);
+			result = item;
+		}
+		else if (name == "PUSH [ErrorTag]")
+			result = {AssemblyItemType::PushTag, data, location};
+		else if (name == "PUSH [tag]")
+		{
+			if (!value.empty())
+				data = u256(value);
+			updateUsedTags(data);
```
chriseth, line 312: Wouldn't it be better to compute that once at the end from the full list of items?

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+		location.sourceName = sources()[static_cast<size_t>(srcIndex)];
+
+	AssemblyItem result(0);
+
+	if (c_instructions.find(name) != c_instructions.end())
+	{
+		AssemblyItem item{c_instructions.at(name), location};
+		item.m_modifierDepth = modifierDepth;
+		if (!value.empty())
+			item.setJumpType(value);
```
chriseth, line 291: Is the jump type stored inside `value`?

chriseth: Oh really - it is. Could we maybe do the string to jump type conversion here (to keep the parsing together) instead of overloading the function?

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+		{
+			if (!value.empty())
+				data = u256(value);
+			updateUsedTags(data);
+			result = {AssemblyItemType::PushTag, data, location};
+		}
+		else if (name == "PUSH [$]")
+		{
+			if (!value.empty())
+				data = u256("0x" + value);
```
chriseth, line 318: Please check that this works properly when the value starts with a zero (i.e. leading zeros may be removed somewhere in the process).

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+			this->m_usedTags = tag + 1;
+	};
+
+	auto updateImmutables = [&](string const& _immutableName) -> h256 {
+		h256 hash(util::keccak256(value));
+		this->m_immutables[hash] = _immutableName;
+		return hash;
+	};
+
+	auto updateLibraries = [&](string const& _libraryName) -> h256 {
```
chriseth, line 272: This is used exactly once, I think it's better readable if you inline it.

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+	{
+		AssemblyItem item{c_instructions.at(name), location};
+		item.m_modifierDepth = modifierDepth;
+		if (!value.empty())
+			item.setJumpType(value);
+		result = item;
+	}
+	else
+	{
+		u256 data;
```
chriseth, line 296: I think it would be better to make this a local variable every time it is needed. You can add / use helpers like `parseHexNumber` instead of relying on it being zero-initialized.

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+		else
+			assertThrow(false, InvalidOpcode, "");
+	}
+	result.m_modifierDepth = modifierDepth;
+	return result;
+}
+
+vector<Json::Value> Assembly::assemblyItemAsJSON(AssemblyItem const& _item, int _sourceIndex) const
+{
+	vector<Json::Value> result;
```
chriseth, line 374: Shouldn't this return a single Json::Value?

libevmasm/Assembly.cpp
```diff
@@ -243,12 +246,276 @@ string Assembly::toStringInHex(u256 _value)
...
+
+	switch (_item.type())
+	{
+	case Operation:
+		result.emplace_back(createJsonValue(
+			instructionInfo(_item.instruction()).name,
+			_sourceIndex,
+			_item.m_modifierDepth,
+			_item.location().start,
+			_item.location().end,
```
chriseth, line 384: Can this be simplified so that we don't have to pass all three values all the time?

libevmasm/Assembly.cpp
```diff
@@ -352,16 +542,75 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+		this->m_items.emplace_back(loadItemFromJSON(it));
+
+	for (auto current = this->m_items.begin(); current != this->m_items.end(); ++current)
+	{
+		// During the assembly json export a `JUMPDEST` is always generated after a `tag`.
+		// So we just ignore exactly these `JUMPDEST`'s.
+		auto const next = std::next(current);
+		if (
+			current->type() == AssemblyItemType::Tag &&
+			next->type() == AssemblyItemType::Operation &&
```
chriseth, line 568: next might be invalid.

libevmasm/Assembly.cpp
```diff
@@ -352,16 +542,75 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	}
+
+	return true;
+}
+
+bool Assembly::loadFromAssemblyJSON(Json::Value const& _json, bool _loadSources /* = true */)
+{
+	if (!_json[".code"].isArray())
+		return false;
+	bool result{true};
```
chriseth, line 581:
```suggestion
	bool success{true};
```

libevmasm/Assembly.h
```diff
@@ -158,6 +162,16 @@ class Assembly
 
 	bool isCreation() const { return m_creation; }
 
+	void setSources(std::vector<std::shared_ptr<std::string const>> _sources) {
```
chriseth, line 165:
```suggestion
	void setSources(std::vector<std::shared_ptr<std::string const>> _sources)
	{
```

chriseth: or put everything on a single line

solc/CommandLineParser.cpp
aarlt: @cameel yeah good point. An initial version of this PR did not support any `binaryRuntime` stuff - because I also saw it like you seem to see it. If I recall correctly, @chriseth wanted to have support for `binaryRuntime` and also support for the `srcmap` & `srcmap-runtime`. I understood it like that it would be great to support as many output modes that are usually also working. To achieve this, of course, we need to have some presumption about the imported assembly json. E.g. structure related to creational code vs. runtime code. Right now not every (well-formed & correct) assembly json will lead be create correct output (e.g. missing "runtime section").

@chriseth @ekpyron  This question was more about the high-level view on this PR: if someone uses the asm json import, what should be the supported output-modes. Are my current assumptions to limit and/or support specific output-modes correct?

The main use-case for this I understood like this: someone wrote in contract in solidity, this contract gets exported to asm json, the user may do some tweaks & optimisations on the given code, but e.g. there is no need to change the initial structure e.g. `creational code` vs `runtime code`. Of course the user could add more data sections - but the initial structure should not get modified (`creational` vs `runtime`). As of now, I also did not really check the behaviour of multiple data sections etc - I guess it should just work, but I'm not entirely sure about this.

The main question is: are my assumptions correct? Do we want to support these output modes for assembly json import? Also for https://github.com/ethereum/solidity/pull/12704#discussion_r826422552 - where my assumptions correct to limit and/or support these?

aarlt:
After talking to @chriseth, we decided to split-up this PR.

solc/CommandLineParser.cpp
cameel: > Also for [#12704 (comment)](https://github.com/ethereum/solidity/pull/12704#discussion_r826422552) - where my assumptions correct to limit and/or support these?

This second check is redundant. It's just the opposite of the supported input modes, isn't it?

PR #12705, opened 2022-02-22
CDCL engine

chriseth:

PR #12706, opened 2022-02-22
Fixes bug with post-analysis

msooseth:
This passes all tests. Adds a few assertions. Handles 1/2/N-long learnt clauses in a way that doesn't break the "all clauses are at least 2 long" invariant. It likely leaks memory of 1- and 2-long clauses, to be fixed.


chriseth:
Added the code to #12705

PR #12707, opened 2022-02-22
[ci] Add simple build job using ninja.

aarlt:
It seem to be quite annoying to build with ninja on circleci. I tried to let it run with `base_ubuntu2004` but there the OOM Killer was always killing the compiler. Even `base_ubuntu2004_xlarge` was not able to let the build finish.

However, maybe its enough to just do 
```
cmake .. -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE:-Release}" $CMAKE_OPTIONS -G "Ninja"
```
Maybe this is enough to check the written cmake files more pedantically.

However, the interesting thing is that the problem of https://github.com/ethereum/solidity/pull/12702 is not happening in CI. CI uses ninja `1.10.0`, I used `1.10.2`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,10 @@ defaults:
       name: Build
       command: scripts/ci/build.sh
 
+  - run_build_ninja: &run_build_ninja
+      name: Build (Ninja)
+      command: scripts/ci/build_ninja.sh
+
   - run_build_ossfuzz: &run_build_ossfuzz
       name: Build_ossfuzz
       command: scripts/ci/build_ossfuzz.sh
@@ -213,6 +217,11 @@ defaults:
             name: Executing solc LSP test suite
             command: ./test/lsp.py ./build/solc/solc
         - gitter_notify_failure_unless_pr
+
+  - steps_build_ninja: &steps_build_ninja
+      steps:
+        - checkout
+        - run: *run_build_ninja
 
   - steps_build: &steps_build
       steps:
@@ -759,6 +768,12 @@ jobs:
     # Enough other jobs depend on it that it's worth it though.
     <<: *base_ubuntu2004_xlarge
     <<: *steps_build
+
+  b_ubu_ninja: &b_ubu_ninja
+    # this runs 2x faster on xlarge but takes 4x more resources (compared to medium).
+    # Enough other jobs depend on it that it's worth it though.
+    <<: *base_ubuntu2004
+    <<: *steps_build_ninja
 
   # x64 ASAN build, for testing for memory related bugs
   b_ubu_asan: &b_ubu_asan
@@ -1448,6 +1463,7 @@ workflows:
 
       # Ubuntu build and tests
       - b_ubu: *workflow_trigger_on_tags
+      - b_ubu_ninja: *workflow_trigger_on_tags
       - t_ubu_cli: *workflow_ubuntu2004
       - t_ubu_soltest_all: *workflow_ubuntu2004
       - t_ubu_soltest_enforce_yul: *workflow_ubuntu2004
--- a/scripts/ci/build_ninja.sh
+++ b/scripts/ci/build_ninja.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+set -ex
+
+ROOTDIR="$(dirname "$0")/../.."
+cd "${ROOTDIR}"
+
+# shellcheck disable=SC2166
+if [ "$CIRCLE_BRANCH" = release -o -n "$CIRCLE_TAG" -o -n "$FORCE_RELEASE" ]
+then
+    echo -n "" >prerelease.txt
+else
+    # Use last commit date rather than build date to avoid ending up with builds for
+    # different platforms having different version strings (and therefore producing different bytecode)
+    # if the CI is triggered just before midnight.
+    TZ=UTC git show --quiet --date="format-local:%Y.%-m.%-d" --format="ci.%cd" >prerelease.txt
+fi
+
+if [ -n "$CIRCLE_SHA1" ]
+then
+    echo -n "$CIRCLE_SHA1" >commit_hash.txt
+fi
+
+mkdir -p build
+cd build
+
+# shellcheck disable=SC2086
+cmake --version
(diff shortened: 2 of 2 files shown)


aarlt:
@chfast do you know whether just running the cmake command would be enough to check the cmake files more pedantically? We could also just run ninja, but it seem to be quite expensive (and/or slow) - so maybe its not needed to actually start a ninja build.

aarlt:
@chfast I'm also very surprised that `1.10.0` vs. `1.10.2` seem to behave that different. Maybe you know why? I did not check ninjas changelog yet.

aarlt:
Maybe the other option of just adding a simple build job that uses ninja, is to change everything from `make` to `ninja`.

chfast:
> @chfast do you know whether just running the cmake command would be enough to check the cmake files more pedantically? We could also just run ninja, but it seem to be quite expensive (and/or slow) - so maybe its not needed to actually start a ninja build.

No. The main gain is that in Ninja all dependencies must be explicit while in Makefiles some implicit dependencies may just work.

> @chfast I'm also very surprised that `1.10.0` vs. `1.10.2` seem to behave that different. Maybe you know why? I did not check ninjas changelog yet.

> Maybe the other option of just adding a simple build job that uses ninja, is to change everything from `make` to `ninja`.

I think it may sense to just convert single build or add single new one. However, in my experience Ninja builds on CI work better. Based on symptoms reported, the problem seems to be the number of parallel builds: Ninja automatically uses parallel builds, but in virtualized environments it may detect number of CPUs incorrectly (similarly cmake --build --parallel).

Add `ninja --help` to CI to see how much CPU cores it thinks it has.

aarlt:
@chfast ok, looks like that you where right. The number of CPUs seem to be detected incorrectly `[default=38 on this system]`. The thing is that for me in the context of https://github.com/ethereum/solidity/pull/12702 it is already failing during `cmake .. -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE:-Release}" $CMAKE_OPTIONS -G "Ninja"`. Let's see how it behaves now.

chfast:
I believe ninja sets the default number of jobs to `nCPU + 2`. So you may try -j4 on 2 vCPU machine. But looks slow anyway.

cameel:
> Ninja automatically uses parallel builds, but in virtualized environments it may detect number of CPUs incorrectly (similarly cmake --build --parallel).

Yeah, can confirm that. I had this problem with pylint in #12443. I didn't investigate to see what number it was getting but with auto-detection it was just hanging the job and sitting there for 10 min until CircleCI killed it for inactivity.

scripts/ci/build_ninja.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo -n "$CIRCLE_SHA1" >commit_hash.txt
+fi
+
+mkdir -p build
+cd build
+
+# shellcheck disable=SC2086
+cmake --version
+cmake .. -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE:-Release}" $CMAKE_OPTIONS -G "Ninja"
+ninja -j 4
```
cameel, line 29: Does ninja read `MAKEFLAGS`? Or maybe it has its own variable for that?

We set this flag in `config.yml` and I think it's a better place - there we can set it appropriately because we know the number of cores corresponding to the `resource_class` or executor we choose.

scripts/ci/build_ninja.sh
```diff
@@ -0,0 +1,29 @@
...
+then
+    echo -n "$CIRCLE_SHA1" >commit_hash.txt
+fi
+
+mkdir -p build
+cd build
+
+# shellcheck disable=SC2086
+cmake --version
+cmake .. -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE:-Release}" $CMAKE_OPTIONS -G "Ninja"
```
cameel, line 28: I'd make the generator type an argument for `build.sh`

.circleci/config.yml
```diff
@@ -760,6 +769,12 @@ jobs:
     <<: *base_ubuntu2004_xlarge
     <<: *steps_build
 
+  b_ubu_ninja: &b_ubu_ninja
+    # this runs 2x faster on xlarge but takes 4x more resources (compared to medium).
+    # Enough other jobs depend on it that it's worth it though.
+    <<: *base_ubuntu2004
+    <<: *steps_build_ninja
```
cameel, line 776: I'd just switch one (or even more) of the existing builds to ninja.

scripts/ci/build_ninja.sh
chfast: No. CMake will generate ninja files (`ninja.build`) instead of Makefile (default generator).

cameel: Yeah, but I was hoping ninja might support this (or more likely) its own variable.

Looks like you're right though. The maintainers of ninja are weirdly opposed to any hint of having an env variable, even suggesting odd workarounds like replacing the binary with a custom script in system dirs (https://github.com/ninja-build/ninja/issues/797, https://github.com/ninja-build/ninja/issues/1370).

Anyway, in that case I think this should be passed via a parameter (or our custom variable) from CI config.

aarlt:
Surprisingly the error described in https://github.com/ethereum/solidity/pull/12702 was not detected on our CI. In our CI we have `cmake`[v3.16.3](https://github.com/Kitware/CMake/releases/tag/v3.16.3) installed. I verified that this version of `cmake` is not triggering the described error. The error is at least triggered with `cmake`[v3.22.2](https://github.com/Kitware/CMake/releases/tag/v3.22.2). We should maybe consider to also upgrade `cmake` in our CI.

cameel:
I think that the CMake version might be constrained by what is available in the Ubuntu images we use. Though not 100% sure - we do in some cases install it from source (e.g. for emscripten or on Windows) so possible we're doing that on Ubuntu too.

In any case, I think that it would be fine to use newer cmake at least in the jobs where we want to use ninja.

ekpyron:
We should switch this to an archlinux build run, which will have the latest upstream version of cmake.

ekpyron:
Closing this, tracking it's recreation in https://github.com/ethereum/solidity/issues/13385

Issue #12677, opened 2022-02-16
Warn about not explicitly requested conversions from rational number types to their mobile type.


donutcrypto:
## Description

The ternary operator doesn't appear to be producing the correct return type.

## Environment

- Compiler version: 8.11
- Framework/IDE (e.g. Truffle or Remix): hardhat
- EVM execution environment / backend / blockchain client: binance smart chain, hardhat testnet

## Steps to Reproduce

Deploying the following contract will overflow on uint8 addition. The expected value of `result` is 385.

``` solidity
contract TestTernary
{
    mapping(uint8 => bool) flags;

    uint256 public result;

    constructor()
    {
        flags[uint8(1)] = true;
        flags[uint8(2)] = true;
        flags[uint8(3)] = true;

        result = 
            (flags[0] ? 100 : 0) +
            (flags[1] ? 110 : 0) +
            (flags[2] ? 125 : 0) +
            (flags[3] ? 150 : 0);
    }
}
```

bshastry:
Thanks for reporting the issue @donutcrypto 

Notes:

Fuzzer wouldn't have helped because both legacy and IR fail on construction. Semantic test below:

```
contract TestTernary
{
    mapping(uint8 => bool) flags;

    uint256 public result;

    constructor()
    {
        flags[uint8(1)] = true;
        flags[uint8(2)] = true;
        flags[uint8(3)] = true;

        result = 
            (flags[0] ? 100 : 0) +
            (flags[1] ? 110 : 0) +
            (flags[2] ? 125 : 0) +
            (flags[3] ? 150 : 0);
    }
}
// ====
// compileViaYul: also
// ----
// constructor() -> FAILURE
// gas irOptimized: 139402
// gas legacy: 140313
// gas legacyOptimized: 135217
```

chriseth:
The reason is that unlimited precision only works as long as you are dealing with arithmetic expressions on literal constants. There are other similar situations where this can happen. I think we could warn when conversions from rational constants to their mobile type happen instead of a conversion to an explicit type.

donutcrypto:
Thanks for the reponse. I just wanted to make sure I understand, but the values of `0`, `100`, `110`, etc. will be compiled to a uint8?

For example:

```
function f(uint8 _x);
function f(uint256 _x);

function g()
{
    f(1); // calls the uint8 version?
}
```

hrkrshnn:
[...]

Issue #12635, opened 2022-02-05
Tuple elements index access


k06a:
## Abstract

Tuples elements can be accessed with indexes. Let's have it!

## Motivation

Other programming languages have this feature, it's quite useful. I believe code could be more expressive with this feature.

## Specification

Tuple could be called anonymous struct, with both name and field names anonymous. We could use indexes to have access to individual fields:

```solidity
function foo() public view returns(uint256 value, bytes memory data) { ... }

foo().0
foo().1
```

## Backwards Compatibility

Not sure there is any way to make this backward compatible.

chriseth:
The main problem I see is that tuples are not "proper" types in solidity.

cameel:
But it's just the fact that we have no tuple-type variables, right? I don't think this particular feature would force us in any way to introduce them.

Though I'm not sure I like the proposed syntax. The index looks like a field name, which it is not. Why not array brackets instead?

chriseth:
The point is that we do not want to have arbitrary access to tuple elements - the element always has to be a compile-time constant, otherwise code generation gets very complicated.

ekpyron:
Yeah, conceptually, as @k06a says, tuples should work like anonymous structs with only compile-time fixed members for accessing tuple elements.
Dynamically accessing tuple elements is not only complicated in code generation, but more importantly, it's impossible to type-check, so that's a no-go, but static access is fine.

But yeah, I'd actually even be down for promoting tuples to proper, even namable types in general - I've always wanted that :-). But even without that, compile-time fixed index access to them would definitely be nice - just not sure about the syntax.

k06a:
@cameel array brackets makes me think elements are of the same type, which is not true.

k06a:
Also I remember someone asked for ability to use tuples as arguments in function calls:
```solidity
token.transfer(getReceiverAndAmount());
```

leonardoalt:
In general I think tuples should be proper types.
I used to dislike this syntax too, but nowadays I actually like it. It's succinct, clear, and already used by other languages.

cameel:
> array brackets makes me think elements are of the same type, which is not true.

I think it depends on what each of us is used to. To me the `.0` syntax feels a bit like I'm abusing some internal mechanism that exposes element indexes as properties on an object. It just looks weird. On the other hand, having used Python a lot, using brackets to index tuples feels completely natural and intuitive.

[...]

PR #12708, opened 2022-02-22
Document behaviour of ternary operator of literals.

nishant-sachdeva:
Helps #12677 

Clarifies the docs and adds tests.

### DIFF
--- a/docs/types/operators.rst
+++ b/docs/types/operators.rst
@@ -26,6 +26,23 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator is used in expressions of the form ``<expression> ? <trueExpression> : <falseExpression>``.
+It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``<expression>``.
+If ``<expression>`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
+
+The result of the ternary operator does not have a rational number type, even if all of its operands are rational number literals.
+The result type is determined from the types of the two operands in the same way as above, converting to their mobile type first if required.
+
+As a consequence, ``255 + (true ? 1 : 0)`` will revert due to arithmetic overflow.
+The reason is that ``(true ? 1 : 0)`` is of ``uint8`` type, which forces the addition to be performed in ``uint8`` as well,
+and 256 exceeds the range allowed for this type.
+
+Another consequence is that an expression like ``1.5 + 1.5`` is valid but ``1.5 + (true ? 1.5 : 2.5)`` is not.
+This is because the former is a rational expression evaluated in unlimited precision and only its final value matters.
+The latter involves a conversion of a fractional rational number to an integer, which is currently disallowed.
+
 .. index:: assignment, lvalue, ! compound operators
 
 Compound and Increment/Decrement Operators
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -463,13 +463,22 @@ There is no additional semantic meaning added to a number literal containing und
 the underscores are ignored.
 
 Number literal expressions retain arbitrary precision until they are converted to a non-literal type (i.e. by
-using them together with a non-literal expression or by explicit conversion).
+using them together with anything else than a number literal expression (like boolean literals) or by explicit conversion).
 This means that computations do not overflow and divisions do not truncate
 in number literal expressions.
 
 For example, ``(2**800 + 1) - 2**800`` results in the constant ``1`` (of type ``uint8``)
 although intermediate results would not even fit the machine word size. Furthermore, ``.5 * 8`` results
 in the integer ``4`` (although non-integers were used in between).
+
+.. warning::
+    While most operators produce a literal expression when applied to literals, there are certain operators that do not follow this pattern:
+
+    - Ternary operator (``... ? ... : ...``),
+    - Array subscript (``<array>[<index>]``).
+
+    You might expect expressions like ``255 + (true ? 1 : 0)`` or ``255 + [1, 2, 3][0]`` to be equivalent to using the literal 256
+    directly, but in fact they are computed within the type ``uint8`` and can overflow.
 
 Any operator that can be applied to integers can also be applied to number literal expressions as
 long as the operands are integers. If any of the two is fractional, bit operations are disallowed
--- a/test/libsolidity/semanticTests/literals/ternary_operator_with_literal_types_overflow.sol
+++ b/test/libsolidity/semanticTests/literals/ternary_operator_with_literal_types_overflow.sol
@@ -1,0 +1,19 @@
+contract TestTernary
+{
+    function h() pure public returns (uint16 b)
+    {
+        b = (true ? 63 : 255) + (false ? 63 : 255);
+    }
+
+    function g() pure public returns (uint16 a)
+    {
+        bool t = true;
+        bool f = false;
+        a = (t ? 63 : 255) + (f ? 63 : 255);
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// g() -> FAILURE, hex"4e487b71", 0x11
+// h() -> FAILURE, hex"4e487b71", 0x11
--- a/test/libsolidity/syntaxTests/literals/ternary_operator_return_type_with_literal_arguments.sol
+++ b/test/libsolidity/syntaxTests/literals/ternary_operator_return_type_with_literal_arguments.sol
@@ -1,0 +1,33 @@
+contract TestTernary
+{
+    function g() pure public
+    {
+        bool t = true;
+        bool f = false;
+        uint8 v255 = 255;
+        uint8 v63 = 63;
+        uint8 a;
+
+        // Currently none of these should produce errors or warnings.
+        // The result of the operator is always a limited-precision integer, even if all arguments are literals.
+
+        a = (t ? 63 : 255) + (f ? 63 : 255);
+        a = (t ? 0x3f : 0xff) + (f ? 0x3f : 0xff);
+        a = (t ? uint8(63) : 255) + (f ? 63 : uint8(255));
+        a = (t ? v63 : 255) + (f ? 63 : v255);
+
+        a = (true ? 63 : 255) + (false ? 63 : 255);
+        a = (true ? 0x3f : 0xff) + (false ? 0x3f : 0xff);
+        a = (true ? uint8(63) : 255) + (false ? 63 : uint8(255));
+        a = (true ? v63 : 255) + (false ? 63 : v255);
+
+        a = (t ? 63 : 255) - (f ? 63 : 255);
+        a = (t ? 63 : 255) * (f ? 63 : 255);
+        a = (t ? 63 : 255) / (f ? 63 : 255);
+
+        a = (t ? (true ? 63 : 255) : (false ? 63 : 255)) + (f ? (t ? 63 : 255) : (f ? 63 : 255));
+        a = uint8(t ? 63 : 255) + uint8(f ? 63 : 255);
+
+    }
+}
+// ----


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1321,6 +1321,7 @@ bool TypeChecker::visit(VariableDeclarationStatement const& _statement)
 				valueComponentType->category() == Type::Category::RationalNumber &&
 				dynamic_cast<RationalNumberType const&>(*valueComponentType).isFractional() &&
 				valueComponentType->mobileType()
+				// Add Check here (?)
```
chriseth: This is only used to generate a nice error message. What is actually happening here is that in line 1313, we check for an implicit conversion to the variable's type. This type is explicit and not somehow generated from the combination of other types, so this case is perfectly fine.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1364,6 +1364,8 @@ void TypeChecker::endVisit(ExpressionStatement const& _statement)
 {
 	if (type(_statement.expression())->category() == Type::Category::RationalNumber)
 		if (!dynamic_cast<RationalNumberType const&>(*type(_statement.expression())).mobileType())
+		// Add Check here (?)
+		// An inadvertent conversion could be happening in RationalType::mobileType()
 			m_errorReporter.typeError(3757_error, _statement.expression().location(), "Invalid rational number.");
```
cameel: This checks literal expressions that are the only thing in a statement. I.e.

```soldidity
1 + 2 * 3;
```

Such statements have no effect so I think there's no point warning about the conversion here.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1411,6 +1413,10 @@ bool TypeChecker::visit(Conditional const& _conditional)
 	else if (trueType && falseType)
 	{
 		commonType = Type::commonType(trueType, falseType);
+		// Add Check here (?)
+		// Seems to be dealing with ternary operators.
+		// Our test case arose out of ternary operators. 
+		// An inadvertent conversion could be happening
```
cameel: Yeah, I think this is the one we want.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1591,6 +1597,8 @@ bool TypeChecker::visit(TupleExpression const& _tuple)
 			// Note: code generation will visit each of the expression even if they are not assigned from.
 			if (types[i]->category() == Type::Category::RationalNumber && components.size() > 1)
 				if (!dynamic_cast<RationalNumberType const&>(*types[i]).mobileType())
+				// Add Check here (?)
+				// An inadvertent conversion could be happening in RationalType::mobileType()
 					m_errorReporter.fatalTypeError(3390_error, components[i]->location(), "Invalid rational number.");
```
cameel: I think this one is for stuff like
```solidity
(uint32 a, uint64 b) = (1, 2);
```
or
```solidity
return (1, 2);
```

Currently you cannot use tuple components in expressions so I think there's no point warning about it. You can only assign/pass the whole tuple or use a destructuring assignment.

This will change though if we decide to implement #12635.

cameel:
@nishant-sachdeva Looks like this check catches some cases where only one argument is a literal or where there already is a conversion:
- Perpetual Pools: `if (negativeY) return absoluteX > absoluteY ? -1 : int8 (1);`
- Euler: `return eTokenAddress == address(0) ? 0 : IERC20(token).balanceOf(eulerAddress);`
- Bleeps: `return eta == 1 ? 0 : eta;`
- Colony: `return unsatisfactory ? 0 : task.payouts[_role][_token];`

test/libsolidity/syntaxTests/literals/ternary_operator_return_type.sol
```diff
@@ -0,0 +1,19 @@
...
+    function g() public returns (uint256 result)
+    {
+        flags[uint8(1)] = true;
+        flags[uint8(2)] = true;
+        flags[uint8(3)] = true;
+
+        result = (flags[0] ? uint256(100) : 0) +
+                    (flags[1] ? 110 : 0) +
+                    (flags[2] ? 125 : 0) +
+                    (flags[3] ? 150 : 0);
```
cameel:
```suggestion
        result =
            (flags[0] ? uint256(100) : 0) +
            (flags[1] ? 110 : 0) +
            (flags[2] ? 125 : 0) +
            (flags[3] ? 150 : 0);
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1411,7 +1411,6 @@ bool TypeChecker::visit(Conditional const& _conditional)
 	else if (trueType && falseType)
 	{
 		commonType = Type::commonType(trueType, falseType);
-
```
cameel: Looks like this PR is no longer modifying anything substantial in this file. Can you remove this change from the commit?

test/libsolidity/syntaxTests/literals/ternary_operator_return_type.sol
```diff
@@ -0,0 +1,19 @@
+contract TestTernary
+{
+    mapping(uint8 => bool) flags;
+
+    function g() public returns (uint256 result)
+    {
+        flags[uint8(1)] = true;
+        flags[uint8(2)] = true;
+        flags[uint8(3)] = true;
```
cameel: Storage is not relevant to the test so I'd make it simpler:
```suggestion
        bool a = true;
        bool b = true;
        bool c = true;
```

test/libsolidity/syntaxTests/literals/ternary_operator_return_type.sol
cameel: Let's include more cases None of them should produce a warning now but if we introduce it later, we'll want to see which cases it affects.

```solidity
        bool t = true;
        bool f = false;
        uint8 v255 = 255;
        uint8 v63 = 63;
        uint8 a;

        // Currently none of these should produce errors or warnings.
        // The result of the operator is always a limited-precision integer, even if all arguments are literals.

        a = (t ? 63 : 255) + (f ? 63 : 255);
        a = (t ? 0x3f : 0xff) + (f ? 0x3f : 0xff);
        a = (t ? uint8(63) : 255) + (f ? 63 : uint8(255));
        a = (t ? v63 : 255) + (f ? 63 : v255);

        a = (true ? 63 : 255) + (false ? 63 : 255);
        a = (true ? 0x3f : 0xff) + (false ? 0x3f : 0xff);
        a = (true ? uint8(63) : 255) + (false ? 63 : uint8(255));
        a = (true ? v63 : 255) + (false ? 63 : v255);

        a = (t ? 63 : 255) - (f ? 63 : 255);
        a = (t ? 63 : 255) * (f ? 63 : 255);
        a = (t ? 63 : 255) / (f ? 63 : 255);

        a = (t ? (true ? 63 : 255) : (false ? 63 : 255)) + (f ? (t ? 63 : 255) : (f ? 63 : 255));
        a = uint8(t ? 63 : 255) + uint8(f ? 63 : 255);
```

Also, let's rename the test to `ternary_operator_return_type_with_literal_arguments.sol`

cameel: I think we should also have a semantic test to verify the current unintuitive behavior. I.e. that these actually overflow:
```solidity
        uint16 a = (t ? 63 : 255) + (f ? 63 : 255);
```
```solidity
        uint16 a = (true ? 63 : 255) + (false ? 63 : 255);
```

docs/types/operators.rst
```diff
@@ -15,6 +15,14 @@ and the type of the operator's result:
    operand, use the type of the right operand,
 3. otherwise, the operation is not allowed.
 
+.. Warning::
```
cameel: The [section about literals](https://docs.soliditylang.org/en/latest/types.html#rational-and-integer-literals) needs clarification too.

First, it says `by using them together with a non-literal expression` and to me this does not imply that `true` and `false` count too. They are not "non-literal expressions".

Second, I think this section needs a warning too, pointing out the behavior of the ternary operator. I'd mention the case of array literals too (i.e. something like `[255][0]` being of `uint8` type and not a literal expression).

cameel: Aren't these case-sensitive?
```suggestion
.. warning::
```

docs/types/operators.rst
```diff
@@ -15,6 +15,14 @@ and the type of the operator's result:
...
+    The "type of the operand" is calculated through a function mobileType() that assigns, in case we're dealing with Rational Literals,
+    the largest possible type that is smaller than the size of the value. An unintended effect of this mechanism is that rational literals
+    ( that are not floats ) get converted to the largest possible integer type that just covers the size of the number.
+
+    For eg, a Rational literal 100 will get converted to uint8.
+    Changes like these can lead to overflow errors and other unintended consequences down the line during various operations.
```
cameel: The C++ `mobileType()` function is an implementation detail so I don't think it should be mentioned in the docs.

Also, we do not have floats :) We do have fixed-point types but something like `1.23` really is just a rational literal and in our type system has the same type as integer literals. Such literals have unlimited precision so it's neither fixed- nor float-point. It's an exact representation.

I'd put that warning lower, either just below `If both are literal numbers, the operation is computed with arbitrary precision.` or at the end of the section. That paragraph already explains what a mobile type is.

And here's how I would reword it:

```rst
.. warning::
    The result of the ternary operator is never a literal, even if all of its operands are literals.
    The result type is the bigger of the mobile types of the integer operands.

    For example ``255 + (true ? 1 : 0)`` will revert due to arithmetic overflow.
    The reason is that ``(true ? 1 : 0)`` is of ``uint8`` type, which forces the addition to be performed in ``uint8`` as well,
    and 256 exceeds the range allowed for this type.

    Another consequence is that an expression like ``1.5 + 1.5`` is valid but ``1.5 + (true ? 1.5 : 2.5)`` is not.
    This is because the former is a literal expression evaluated in unlimited precision and only its final value matters.
    The latter involves a conversion of a fractional rational literal to an integer, which is currently disallowed.
```

cameel (changes requested):

cameel:
Please adjust the title of the PR and modify the description so that it no longer has `Closes #12677` - the PR no longer solves the issue, just clarifies the docs and adds tests.

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+.. warning::
```
chriseth: Some suggestions:
 - please check how we call the operator in other parts of the documentation and at least also add `? ... : ...` as an alternative here, so that everyone knows what we are talking about
 - "literal" is more strict than what we want here. I think we use the term rational constant or something, please check.

At this point in the documentation, I think a "note" is better then a warning - we don't use a warning for `**` either.

docs/types/value-types.rst
```diff
@@ -463,14 +463,22 @@ There is no additional semantic meaning added to a number literal containing und
 the underscores are ignored.
 
 Number literal expressions retain arbitrary precision until they are converted to a non-literal type (i.e. by
-using them together with a non-literal expression or by explicit conversion).
+using them together with a non-literal expression or with booleans like true/false or by explicit conversion).
```
chriseth:
```suggestion
using them together with anything else than a number literal expression (like boolean literals) or by explicit conversion).
```

docs/types/value-types.rst
```diff
@@ -463,14 +463,22 @@ There is no additional semantic meaning added to a number literal containing und
...
 This means that computations do not overflow and divisions do not truncate
 in number literal expressions.
 
 For example, ``(2**800 + 1) - 2**800`` results in the constant ``1`` (of type ``uint8``)
 although intermediate results would not even fit the machine word size. Furthermore, ``.5 * 8`` results
 in the integer ``4`` (although non-integers were used in between).
 
+.. warning::
+    There are cases, such as those involving the ternary operator, where implicit conversions to types of a smaller size can happen.
```
chriseth: type smaller than what?

test/libsolidity/syntaxTests/literals/ternary_operator_return_type_with_literal_arguments.sol
```diff
@@ -0,0 +1,45 @@
...
+        bool y = true;
+        bool z = true;
+        bool w = true;
+
+        result =
+            (x ? uint256(100) : 0) +
+            (y ? 110 : 0) +
+            (z ? 125 : 0) +
+            (w ? 150 : 0);
+
```
cameel: I though you'd just replace this part with the stuff below. I think it's redundant now.

test/libsolidity/semanticTests/literals/ternary_operator_with_literal_types_overflow.sol
```diff
@@ -0,0 +1,21 @@
+contract TestTernary
+{
+    function h() pure public returns (uint16 b)
+    {
+        b = (true ? 63 : 255) + (false ? 63 : 255);
+        return b;
```
cameel: You do not need to use `return` when you have a named return value. It's a bit redundant.

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+.. note::
+    The result of the ternary operator ``... ? ... : ...`` is never a rational, even if all of its operands are rationals.
```
cameel: To be honest, the terminology around literals/constants is a bit unclear to me too. I'd say at least "rational constant" and I saw @chriseth suggest that. Do we actually refer to them as just "rationals"?
```suggestion
    The result of the ternary operator (``... ? ... : ...``) is never a rational constant, even if all of its operands are rational constants.
```

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
...
+    The result type is the bigger of the mobile types of the integer operands.
```
cameel: I originally suggested "integer", but maybe "numeric" would be more fitting here? It applies to fractional numbers too.
```suggestion
    The result type is the bigger of the mobile types of the numeric operands.
```

docs/types/operators.rst
cameel: Actually, after looking at the other file, maybe we should go with "number literal"/"number literal expression" for consistency?

docs/types/value-types.rst
```diff
@@ -463,14 +463,23 @@ There is no additional semantic meaning added to a number literal containing und
...
 in the integer ``4`` (although non-integers were used in between).
 
+.. warning::
+    There are cases, such as those involving the ternary operator, where implicit conversions to sub-types of a smaller size can happen.
+    (For example, a rational number 100 will get converted to uint8 because uint8 is the smallest sized type that is sufficient to accommodate it)
+    This can lead to unintended overflows during calculations further down the pipeline.
+
+    For example ``255 + (true ? 1 : 0)`` will revert due to arithmetic overflow.
+    The reason is that ``(true ? 1 : 0)`` is of ``uint8`` type, which forces the addition to be performed in ``uint8`` as well,
+    and 256 exceeds the range allowed for this type.
```
cameel: Here's my take:
```suggestion
    While most operators produce a literal expression when applied to literals, there are certain operators that do not follow this pattern:
    - Ternary operator (``... ? ... : ...``),
    - Array subscript (``<array>[<index>]``).

    You might expect expressions like ``255 + (true ? 1 : 0)`` or ``255 + [1, 2, 3][0]`` to be equivalent to using the literal 256 directly, but in fact they are computed within the type ``uint8`` and can overflow.
```

docs/types/operators.rst
nishant-sachdeva: I'm pushing changes with "number literal" .
We can replace it further if we decide so

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+.. note::
+    The result of the ternary operator (``... ? ... : ...``) is never a number literal, even if all of its operands are number literals.
+    The result type is the bigger of the mobile types of the numeric operands.
```
chriseth:
```suggestion
    The result type is determined from the types of the two operands in the same way as above, converting to their mobile type first if required.
```

chriseth: The main reason to rephrase this is that "size" is difficult because of signed types.

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
...
+
+    For example ``255 + (true ? 1 : 0)`` will revert due to arithmetic overflow.
```
chriseth:
```suggestion
    As a consequence, ``255 + (true ? 1 : 0)`` will revert due to arithmetic overflow.
```

docs/types/operators.rst
```diff
@@ -26,6 +26,18 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+.. note::
```
chriseth: I'm sorry - I just now opened this page in a wider context, and I think this should just be a section
```
Ternary Operator
--------------
```
instead of a note.

docs/types/operators.rst
```diff
@@ -26,6 +26,19 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+    The result of the ternary operator (``... ? ... : ...``) is never a number literal, even if all of its operands are number literals.
+    The result type is determined from the types of the two operands in the same way as above, converting to their mobile type first if required.
```
cameel: This will probably not render properly if you indent the text like this. The header is enough to create a section, you don't have to indent the content in RST or Markdown. The indented text may be interpreted as a code block.

docs/types/value-types.rst
```diff
@@ -463,14 +463,21 @@ There is no additional semantic meaning added to a number literal containing und
...
 in number literal expressions.
 
 For example, ``(2**800 + 1) - 2**800`` results in the constant ``1`` (of type ``uint8``)
 although intermediate results would not even fit the machine word size. Furthermore, ``.5 * 8`` results
 in the integer ``4`` (although non-integers were used in between).
 
+.. warning::
+    While most operators produce a literal expression when applied to literals, there are certain operators that do not follow this pattern:
+    - Ternary operator (``... ? ... : ...``),
+    - Array subscript (``<array>[<index>]``).
```
cameel: Looks like this does not render properly in RST. I think we need an empty newline before the list.
```suggestion
    While most operators produce a literal expression when applied to literals, there are certain operators that do not follow this pattern:

    - Ternary operator (``... ? ... : ...``),
    - Array subscript (``<array>[<index>]``).
```

cameel (commented):
Looks like we're almost there but still needs minor syntax corrections to render properly in the docs :)

docs/types/operators.rst
```diff
@@ -26,6 +26,19 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The result of the ternary operator (``... ? ... : ...``) is never a number literal, even if all of its operands are number literals.
```
chriseth: Now that this is its own section, maybe we should start with a more generic introduction unrelated to number literals.

docs/types/operators.rst
```diff
@@ -26,6 +26,24 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator evaluates a given expression to a true or false value and then
+executes one of the two given statements depending upon the result of the evaluation.
```
ekpyron: Technically, what you call ``statementA`` and ``statementB`` aren't statements, but also expressions...

docs/types/operators.rst
```diff
@@ -26,6 +26,23 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator is used in expressions of the form ``<expression> ? <trueExpression> : <falseExpression>``.
+It executes one of the two given expressions depending upon the result of the evaluation of the main ``expression``.
+If ``expression`` evaluates to True, then ``<trueExpression>`` will be executed, else, ``<falseExpression>`` is executed.
```
ekpyron:
```suggestion
It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``expression``.
If ``expression`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
```

docs/types/operators.rst
```diff
@@ -26,6 +26,23 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator is used in expressions of the form ``<expression> ? <trueExpression> : <falseExpression>``.
+It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``expression``.
+If ``expression`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
```
ekpyron:
```suggestion
It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``<expression>``.
If ``<expression>`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
```
Otherwise it's inconsistend with the use of ``<trueExpression>`` and ``<falseExpression>`` isn't it?

docs/types/operators.rst
```diff
@@ -26,6 +26,23 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator is used in expressions of the form ``<expression> ? <trueExpression> : <falseExpression>``.
+It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``<expression>``.
+If ``<expression>`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
+
+The result of the ternary operator (``... ? ... : ...``) is never a number literal, even if all of its operands are number literals.
```
ekpyron:
```suggestion
The result of the ternary operator is never a number literal, even if all of its operands are number literals.
```
We don't need to remention the form of ternary expressions now that we started off with it.

But I'm also not sure if we should say that the result "is never a number literal"... That's a bit of a weird thing to say: a complex expression obviously isn't a literal... technically, it would be accurate to say "never has rational number type", but I'm not sure that really makes things clear, so I'm not sure what's best to do here...

docs/types/operators.rst
```diff
@@ -26,6 +26,23 @@ except for comparison operators where the result is always ``bool``.
 The operators ``**`` (exponentiation), ``<<``  and ``>>`` use the type of the
 left operand for the operation and the result.
 
+Ternary Operator
+----------------
+The ternary operator is used in expressions of the form ``<expression> ? <trueExpression> : <falseExpression>``.
+It evaluates one of the latter two given expressions depending upon the result of the evaluation of the main ``<expression>``.
+If ``<expression>`` evaluates to ``true``, then ``<trueExpression>`` will be evaluated, otherwise ``<falseExpression>`` is evaluated.
+
+The result of the ternary operator cannot have a rational number type, even if all of its operands are rational number literals.
```
ekpyron:
```suggestion
The result of the ternary operator does not have a rational number type, even if all of its operands are rational number literals.
```
:-)?

Aivolt1 (approved):

cameel (dismissed):
I think it's good enough in the current form. All the problems that were pointed out seem to have been fixed and I don't see any new ones so let's get it merged :)

Some tests are failing but they all look unrelated to the PR.

chriseth (approved):

PR #770, opened 2016-08-01
Make the Solidity repository standalone

bobsummerwill:
Replaces https://github.com/ethereum/solidity/pull/739.

This commit is the culmination of several months of work to decouple Solidity from the webthree-umbrella so that it can be developed in parallel with cpp-ethereum (the Ethereum C++ runtime) and so that even for the Solidity unit-tests there is no hard-dependency onto the C++ runtime.

The Tests-over-IPC refactoring was a major step in the same process which was already committed.

This commit contains the following changes:
- A subset of the CMake functionality in webthree-helpers was extracted and tailored for Solidity into ./cmake.   Further cleanup is certainly possible.
- A subset of the libdevcore functionality in libweb3core was extracted and tailored for Solidity into ./libdevcore.   Further cleanup is certainly possible
- The gas price constants in EVMSchedule were orphaned into libevmasm.
- Some other refactorings and cleanups were made to sever unnecessary EVM dependencies in the Solidity unit-tests.
- TravisCI and Appveyor support was added, covering builds and running of the unit-tests (Linux and macOS only for now)
- A bug-fix was made to get the Tests-over-IPC running on macOS.
- There are still reliability issues in the unit-tests, which need immediate attention.    The Travis build has been flipped to run the unit-tests 5 times, to try to flush these out.
- The Emscripten automation which was previously in webthree-umbrella was merged into the TravisCI automation here.
- The development ZIP deployment step has been commented out, but we will want to read that ONLY for release branch.

Further iteration on these changes will definitely be needed, but I feel these have got to sufficient maturity than holding them back further isn't winning us anything.    It is go time :-)

PR #12714, opened 2022-02-24
Remove outdated release scripts

cameel:
I think these scripts have not been touched since they were added back in #770, apart from small syntax tweaks and cleanup. They no longer match our release process - we're releasing static binaries on all platforms as single files - and I doubt anyone is using them.


cameel:
Reposting answers from the chat so that this info is discoverable later:

@nishant-sachdeva 
> what are the scripts that have replaced the ones we're deleting in this PR ?

@cameel 
> Nothing, really. I think these scripts were needed just to package the executable and all the required shared libs into a single zip.
> For most platforms that happened long ago, before I even joined the project. Only Windows was a bit late to the party, which is why [in solc-bin we have zips for windows up to 0.7.1](https://github.com/ethereum/solc-bin/tree/gh-pages/windows-amd64).
> At some point we have switched to building completely static binaries and that's already a single file per platform.
> So just like the [checklist says](https://github.com/ethereum/solidity/blob/develop/ReleaseChecklist.md#download-binaries), you grab a binary from CircleCI artifact page and that's it. No need for additional packaging.

nishant-sachdeva (approved):

chriseth (approved):

scripts/release.sh
```diff
@@ -1,56 +0,0 @@
...
-mkdir -p "$ZIP_TEMP_DIR"
-
-# Copy all the solidity executables into a temporary directory prior to ZIP creation
-
-cp "$REPO_ROOT/build/solc/solc" "$ZIP_TEMP_DIR"
-
-# For macOS, we run a fix-up script which alters all of the symbolic links within
-# the executables and dynamic libraries such that the ZIP becomes self-contained, by
-# revectoring all the dylib references to be relative to the directory containing the
-# application, so that the ZIPs are self-contained, with the only external references
```
ghost, line 47: bnm

PR #11282, opened 2021-04-21
Update release checklist.

chriseth:

PR #10000, opened 2020-10-08
Update release checklist.

chriseth:

Issue #9720, opened 2020-09-01
Replace ``commit_hash.txt`` with git archive placeholders (and try also replacing ``prerelease.txt``)


ekpyron:
The github source archives are invalid in that they are lacking proper ``commit_hash.txt`` and ``prerelease.txt`` files and we instead have to manually pack another source archive, which is confusing and error-prone.

The github source archive is created with ``git archive`` and https://git-scm.com/docs/gitattributes#_export_subst would allow us to create a ``commit_hash.txt`` file with one of the placeholders like https://git-scm.com/docs/git-log#Documentation/git-log.txt-emHem

Then our build scripts could check if the ``commit_hash.txt`` contains that placeholder and assume it's a git checkout and determine the hash using ``git`` and otherwise assume it's a release archive and that ``commit_hash.txt`` contains the proper commit hash.

``prerelease.txt`` could be replaced for example by inspecting the top of the changelog - or maybe https://git-scm.com/docs/git-log#Documentation/git-log.txt-emdem can be used to check for tags in the decorators.

cameel:
I think we should also document the mechanism in the docs as a part of this task. It will become less necessary to do this manually but the mechanism is still not obvious and we should have a place with an explanation to point people to.

cameel:
The script that currently generates these files (and the source tarball) is [`create_source_tarball.sh`](https://github.com/ethereum/solidity/blob/develop/scripts/create_source_tarball.sh).

cameel:
Apparently there's an additional use case not mentioned in the issue description:

@ekpyron 
> It would be nice if checking out a release tag via git would build a release without user interaction.

Do we want that as a part of this too? I assumed it's only about github's source packages and that we're just going to keep the current system outside of that usage. If we do want `prerelease.txt` and `commit_hash.txt` to be automatically populated we probably should use git hooks for that.

cameel:
Here's what I think we should do for this issue:
1) Update our scripts and cmake config to treat the file with a placeholder the same way as if the file did not exist at all. Do not try to get the right value automatically when the file is being read and do not fill out the template automatically outside of `git archive`.
    - This way anyone just checking out the repo and building without touching these files will still get the same behavior as now (i.e. always a pre-release build).
2) Provide a script that gets the right values and fills out the templates.
    - Users can easily run it to get a release build after checking out a tag.
[...]

PR #12715, opened 2022-02-24
Put source tarball creation instructions back on the release checklist

cameel:
This bit was removed in #10000, apparently because Travis was uploading the source package automatically. Since then we have retired Travis and in #11282 we have removed a reference to it from the checklist but without adding an alternative.

We're still publishing the source packages (thought hopefully not for long - #9720) so I'm assuming @chriseth went back to building and publishing them manually. If that is the case, these steps should be put back on the list.

### DIFF
--- a/ReleaseChecklist.md
+++ b/ReleaseChecklist.md
@@ -26,7 +26,11 @@
  - [ ] Click the `PUBLISH RELEASE` button on the release page, creating the tag.
  - [ ] Wait for the CI runs on the tag itself.
 
-### Download Binaries
+### Upload Release Artifacts
+ - [ ] Switch to the tag that archives have to be created for.
+ - [ ] Create the ``prerelease.txt`` file: (``echo -n > prerelease.txt``).
+ - [ ] Run ``scripts/create_source_tarball.sh`` while being on the tag to create the source tarball. This will create the tarball in a directory called ``upload``.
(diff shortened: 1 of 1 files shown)


ReleaseChecklist.md
```diff
@@ -26,7 +26,9 @@
  - [ ] Click the `PUBLISH RELEASE` button on the release page, creating the tag.
  - [ ] Wait for the CI runs on the tag itself.
 
-### Download Binaries
+### Upload Release Artifacts
+ - [ ] Run ``scripts/create_source_tarball.sh`` while being on the tag to create the source tarball. Make sure to create ``prerelease.txt`` before: (``echo -n > prerelease.txt``). This will create the tarball in a directory called ``upload``.
```
nishant-sachdeva:
```suggestion
 - [ ] Switch to the tag that archives have to be created for.
 - [ ] Create the ``prerelease.txt`` file : (``echo -n > prerelease.txt``). 
 - [ ] Run ``scripts/create_source_tarball.sh`` while being on the tag to create the source tarball. This will create the tarball in a directory called ``upload``.
```

stackenbotten:
There was an error when running `chk_coding_style` for commit `d9e576ffe320d40cc2e230b96b081d2ea4a455d9`:
```
ReleaseChecklist.md:31: - [ ] Create the ``prerelease.txt`` file : (``echo -n > prerelease.txt``). 

```

chriseth (approved):

PR #975, opened 2016-08-30
solc crashes without 'export LC_ALL=C'

cjgdev:
This pull request fixes #674. The issue is that it is possible for the POSIX system environment to be invalid when the program starts, causing a runtime exception to be thrown. The solution is to call setlocale to test that setting the user defined default succeeds, only if the call fails then adjust the environment setting to a reasonable default (else the only viable course of action for the program would be to exit).

Issue #674, opened 2016-06-20
solc crashes without 'export LC_ALL=C'


AnthonyAkentiev:
#### System information

Geth version: 1.5.0-unstable
OS & Version: Ubuntu 14.04.4
#### Expected behaviour

Solc must run normally.
#### Actual behaviour

It crashes and generates core file.
#### Steps to reproduce the behaviour

``` bash
ubuntu@xxx:~/$ locale -a
locale: Cannot set LC_CTYPE to default locale: No such file or directory
C
C.UTF-8
en_US.utf8
POSIX

ubuntu@xxx:~/$ solc SomeContract.sol
terminate called after throwing an instance of 'std::runtime_error'
  what():  locale::facet::_S_create_c_locale name not valid
Aborted (core dumped)
```

Then i added 'export LC_ALL=C' and everything went normally.

bobsummerwill:
Hey, @AnthonyAkentiev,

Thanks for the report.   This seems to be a common issue across all of our C++ applications, and is the reason we have the following steps listed in the [building from source](http://www.ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/building-from-source/linux-ubuntu.html):

```
sudo apt-get -y install language-pack-en-base
sudo dpkg-reconfigure locales
```

How did you install your `solc`?   Build from source or `apt-get`?   Perhaps those don't happen for an `apt-get install`?    Or perhaps you followed instructions from Solidity readthedocs which were missing those steps?

The real solution here is https://github.com/ethereum/webthree-umbrella/issues/169 - to force "C" locale at runtime.

This issue popped up for me on a Jolla Sailfish phone too, and I've seen it on Raspberry Pi too.    We need to address it at runtime, not rely on the environment to be "good".

AnthonyAkentiev:
I've installed it using `apt-get`

bobsummerwill:
Thanks, @AnthonyAkentiev.

@chriseth FYI ... Looks like the solidity build instructions are missing those locale steps, which would be needed in both "build from source" and "install from PPA", I think.

PR #12716, opened 2022-02-25
Don't set locale used by C++ stdlib to the value from `LC_ALL` env variable even if it's valid

### LINKED ISSUE #12350: Solc fails compiling code with inline assembly when locale is set to Turkish

wechman:
fixes #12350

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,7 @@ Compiler Features:
 
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
-
+ * Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
 
 ### 0.8.12 (2022-02-16)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
  * Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
 
+
 ### 0.8.12 (2022-02-16)
 
 Language Features:
--- a/solc/main.cpp
+++ b/solc/main.cpp
@@ -27,40 +27,16 @@
 
 #include <boost/exception/all.hpp>
 
-#include <clocale>
 #include <iostream>
 
 using namespace std;
 using namespace solidity;
 
-/*
-The equivalent of setlocale(LC_ALL, "C") is called before any user code is run.
-If the user has an invalid environment setting then it is possible for the call
-to set locale to fail, so there are only two possible actions, the first is to
-throw a runtime exception and cause the program to quit (default behaviour),
-or the second is to modify the environment to something sensible (least
-surprising behaviour).
-
-The follow code produces the least surprising behaviour. It will use the user
-specified default locale if it is valid, and if not then it will modify the
-environment the process is running in to use a sensible default. This also means
-that users do not need to install language packs for their OS.
-*/
-static void setDefaultOrCLocale()
-{
-#if __unix__
-	if (!std::setlocale(LC_ALL, ""))
-	{
-		setenv("LC_ALL", "C", 1);
-	}
-#endif
-}
 
 int main(int argc, char** argv)
 {
 	try
 	{
-		setDefaultOrCLocale();
 		solidity::frontend::CommandLineInterface cli(cin, cout, cerr);
 		return cli.run(argc, argv) ? 0 : 1;
 	}
--- a/test/localeTest.sh
+++ b/test/localeTest.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+
+#------------------------------------------------------------------------------
+# Script that tests that the compiler works correctly regardless of the locale
+# setting. As a prerequisite, the following locales must be enabled system-wide:
+# C, tr_TR.utf8, ja_JP.eucjp.
+#
+# Usage:
+#    <script name>.sh <path to solc binary>
+#
+# ------------------------------------------------------------------------------
+# This file is part of solidity.
+#
+# solidity is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# solidity is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with solidity.  If not, see <http://www.gnu.org/licenses/>
+#
+# (c) 2022 solidity contributors.
+#------------------------------------------------------------------------------
+
+set -eo pipefail
+
+REPO_ROOT=$(cd "$(dirname "$0")/.." && pwd)
+# shellcheck source=scripts/common.sh
+source "${REPO_ROOT}/scripts/common.sh"
+
+solc_binary="$1"
+(( $# == 1 )) || fail "Expected exactly 1 argument."
+
+# This test won't work without some specific locales installed
+locale -a | grep -e "^tr_TR\.utf8$" || fail "Locale 'tr_TR.utf8' not available."
+locale -a | grep -e "^ja_JP\.eucjp$" || fail "Locale 'ja_JP.eucjp' not available."
+locale -a | grep -e "^C$" || fail "Locale 'C' not available."
+locale -a | grep -e "^__invalid_locale__$" && fail "'__invalid_locale__' is not supposed to be a valid locale name."
+
+i="i"
+
+test_code=$(cat <<'EOF'
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity *;
+    library L {}
+EOF
+)
+
+# Whatever locale is set by default.
+printTask "Testing the default locale..."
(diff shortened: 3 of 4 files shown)


cameel:
From the description in #674 it looks to me like the reporter had the locale variables in his system set to a locale that was not enabled (i.e. not present or commented-out in `/etc/locale.gen`). The thing I'm not 100% sure about is why it affected solc at all. `setlocale()` docs say that the stdlib does `setlocale(LC_ALL, "C")` at startup, which means that it should completely ignore the locale env vars - so why would invalid locale name in these vars do anything?

The only way I can explain this is that something else down the line tries to use these env variables and fails. Could be Boost or maybe we're spawning a Bash process at some point to run a shell command? If that's the case, it would explain why `setenv()` added by #975 helped - this way we reset the locale variable inherited by the solc process to "C" if its value is invalid. This means that whatever tries to use that variable, now never has to deal with an invalid value.

I hate to be guessing here and given that this locale problem was lurking for several years, I think we should verify that the solution is correct. We need a test that would detect **both** #12350 and #674. This might not be the easiest task if you're not familiar with our CI and scripts so I just tried to create one myself. Here's my [`locale-test`](https://github.com/ethereum/solidity/tree/locale-test) branch. It adds a test script and a CI job that runs it. I'm not submitting it as a PR because it of course fails on `develop` - it detects the problem with the Turkish locale. Please cherry-pick commits from it into your branch and then:
1) Make sure that after your fix it passes.
2) Try removing the `setenv()` line to verify that the test is effective against #674 - if it does not fail after that, the test needs to be improved. If you can't make it fail, it might indicate that we don't really need the `setenv()`.

solc/main.cpp
```diff
@@ -36,31 +36,21 @@ using namespace solidity;
...
+static void setCLocale()
 {
 #if __unix__
-	if (!std::setlocale(LC_ALL, ""))
-	{
-		setenv("LC_ALL", "C", 1);
-	}
+	setenv("LC_ALL", "C", 1);
 #endif
 }
```
cameel: The name and docs should be adjusted to more precisely describe what is happening here and why. I think that this function is no longer named correctly. It does not really set the locale used by the std library any more. It just adjusts the locale set in the env variable if it's invalid. This variable is supposed to be ignored by stdlib though.

cameel:
Just for reference, here's [a failing run of my new `t_ubu_locale` job](https://app.circleci.com/pipelines/github/ethereum/solidity/22448/workflows/709f21c2-45cd-4c27-bca6-daab095e62d9/jobs/985548), which shows that it at least detects the problem from #12350 on `develop.`

cameel:
Oh, one more thing. This fixes a user-visible bug so you should add a new entry in the [changelog](https://github.com/ethereum/solidity/blob/develop/Changelog.md) under "Bugfixes".

wechman:
@cameel  Thank you for your tests! They cover all scenarios I carried out in my local environment. Unfortunately, I was not able to reproduce #674 and neither your tests are. "setenv" removal does not change anything in this matter (seems we can remove it).

Regardless of the above, your guessing sounds reasonable to me. I think, the problem might be related to the dependency we no longer use (e.g. particular version of boost or other library). This could explain why we cannot reproduce it.

chriseth:
I would actually be happier if we could find a way to somehow set an "illegal" locale so that we get an error whenever we perform a locale-dependent operation (and of course remove all locale-dependent operations).

Changelog.md
```diff
@@ -7,7 +7,7 @@ Compiler Features:
 * JSON-AST: Added selector field for errors and events.
 
 Bugfixes:
-
+ * The compiler uses "C" instead of environment locale.
```
cameel: This describes how we fixed it but I think that in the changelog we should focus on the effects of the bug visible to the users (i.e. it caused an ICE under certain conditions):
```suggestion
 * Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
```

cameel (commented):
I think it's good now. I'd just slightly adjust the changelog entry.

cameel:
> I would actually be happier if we could find a way to somehow set an "illegal" locale so that we get an error whenever we perform a locale-dependent operation (and of course remove all locale-dependent operations).

Yeah. would be nice if we could detect locale-dependent code. I think that there might be a lot of such code though (pretty much everything related to capitalization for example) so I'd treat it as a separate issue.

chriseth:
Also, does it really fix the issue? It only fixes it in one binary...
I think we really need to find all those locale-dependent functions. I don't think there are too many, by the way.

cameel:
@chriseth We only use `setlocale()` in this one binary. I saw in history that we used to have it in LLL binary too and also in some other binaries from related repos from before the move to `ethereum/solidity` but that's it. We don't set locale in `yul-phaser` and `solidity-upgrade` for example.

> I think we really need to find all those locale-dependent functions. I don't think there are too many, by the way.

I just checked and you're right in a way, we have only a handful of calls to `tolower()`, `toupper()` and `isprint()` so that part is simple, but I think there are quite a few more things that depend on locale. Doesn't even `std::string` depend on that for some operations? And what about Boost? For me it's more about uncertainty so I'd prefer to see this PR wrapped up, and a new one opened with a clean state where we can deal with it as a separate problem.

chriseth:
I would prefer both be implemented :)
string is not locale-aware, I think it's really mostly the upper/lower functions and maybe case-insensitive regular expressions, but I doubt we us ethem.

cameel:
Ah, I was thinking about `std::locale` but I see that's independent of `std::setlocale()` (and not global).

By the way, the docs for `setlocale()` have a nice list of functions affected by the global state it manages: https://github.com/ethereum/solidity/issues/12720#issuecomment-1054485424. That's over 40 different functions. ~20 if you don't count the ones meant for wide chars.

cameel (dismissed):
> I would prefer both be implemented :)

If possible, sure. But for now it's not even certain that it's possible to set an invalid locale (https://github.com/ethereum/solidity/issues/12720#issuecomment-1054570799). The part that is done here is self-contained and we want it even if we won't find a way to set an invalid locale so I see no reason not to merge it already.

cameel:
@wechman Can you switch it to non-draft and squash the fixup commit?

test/localeTest.sh
```diff
@@ -0,0 +1,79 @@
...
+i="i"
+
+test_code=$(cat <<'EOF'
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity *;
+    library L {}
+EOF
+)
+
+# Whatever locale is set by default.
```
chriseth, line 54: Does this test actually fail before the fix?
Can you make it generate some code and check that it is identical for all locales?

cameel: > Does this test actually fail before the fix?

Yes, see https://github.com/ethereum/solidity/pull/12716#issuecomment-1051261952 and https://github.com/ethereum/solidity/pull/12716#issuecomment-1054096720 above.

> Can you make it generate some code and check that it is identical for all locales?

I assumed that with bad locale it would not compile so this would be enough, but true, we cold write output from all `solc` calls to files (instead of sending it to `/dev/null`) and `diff` it at the end.

test/localeTest.sh
```diff
@@ -0,0 +1,83 @@
...
+printTask "Testing the default locale..."
+echo "$test_code" | "$solc_binary" - --bin > default_locale_result.out
+
+# Plain C locale
+printTask "Testing the C locale..."
+export LC_ALL=C
+[[ ${i^^} == "I" ]] || assertFail
+echo "$test_code" | "$solc_binary" - --bin > c_locale_result.out
+diff_values "$(<default_locale_result.out)" "$(<c_locale_result.out)"
```
cameel: Please avoid creating temp files in the working dir. These scripts should be usable also outside of CI and it's not a good idea to assume that the work dir does not have files with these names or even that you can write to it. You can create a dir when you can write safely with:
```bash
tmp_dir=$(mktemp -d -t "locale-test-XXXXXX")
```
And remember to delete it at the end of the script.

Alternatively, you could capture the output to a variable - I think this is usually preferable:
```suggestion
c_locale_output=$("$test_code" | "$solc_binary" - --bin)
diff_values "$(echo "$default_locale_output")" "$(echo "$c_locale_output")"
```

cameel (commented):
The script should be adjusted not to create temp files in the working dir. Other than that the PR is fine.

wechman:
> The script should be adjusted not to create temp files in the working dir. Other than that the PR is fine.

Ok, script updated. I already squashed all commits and rebased on the newest develop.

Changelog.md
```diff
@@ -11,7 +11,7 @@ Compiler Features:
 
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
-
+ * Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
```
cameel, line 14: Just a minor nitpick, but I'd keep the extra line here to keep spacing consistent.

cameel (dismissed):

Changelog.md
wechman: Problem fixed.

chriseth (approved):

PR #12717, opened 2022-02-25
added commit-hash.txt file under export-subst flag

### LINKED ISSUE #9720: Replace ``commit_hash.txt`` with git archive placeholders (and try also replacing ``prerelease.txt``)

nishant-sachdeva:
closes #9720


chriseth:
Approach looks good. We need to change all the code that reads those files and query the information from git if the contents contain placeholders.

chriseth:
Maybe we can even execute a git command to do the replacement (of course not modify the files)?

cameel:
Not sure if job failures are even related to the PR. There were some CircleCI problems yesterday to I reran it to make sure. It looks like CircleCI had some problems executing the bash code in the config. If it's reproducible, could also be that they changed something in their API/config format but that seems unlikely to me.

cameel:
> Approach looks good. We need to change all the code that reads those files and query the information from git if the contents contain placeholders.

@nishant-sachdeva Does this replace the placeholders only when you explicitly run `git archive` or every time the code is checked out (would be nice to actually have a comment explaining that in the code)? If it's the former, you'll need to look through everything that reads these two files and manually tell either git to generate them before hand or just treat it as if the files did not exist.

ekpyron:
It'll only replace the placeholders on ``git archive``, so yes: we will need to change the logic when reading those files, in particular ``cmake/scripts/buildinfo.cmake``.

If the files do not contain placeholders, we can assume that it's a release tarball and use the commit hash from the file.

The tricky part will be the case, in which the files do contain placeholders: we will probably need to check if there is a release tag at the current commit to be able to tell, if we should build as prerelease or as release. EDIT: ah right - or we can inspect the top of the changelog instead as suggested in https://github.com/ethereum/solidity/issues/9720.

ekpyron:
Using ``$Format:%d$`` is probably too error-prone.
I think we may need to replace the ``prerelease.txt`` mechanism entirely and instead have something like a ``date.txt`` file plus checking the changelog for ``unreleased`` (unless someone comes up with something better).

cameel:
Is there are way for git to create the files during `git archive`? There would be fewer changes needed if they did not exist by default. Or they could exist but under a different name, like `prerelease.txt.template` or something.

ekpyron:
> Is there are way for git to create the files during `git archive`? There would be fewer changes needed if they did not exist by default. Or they could exist but under a different name, like `prerelease.txt.template` or something.

Would that really help that much? In any case, I think the only thing you can do is to *exclude* files from ``git archive``, which is not much help for our case.

The only complicated part I really see for this is to determine whether we're at a release tag on a git checkout.

ekpyron:
Or put differently: for the commit hash this is straightforward, but the prerelease/release logic we will basically have to rewrite for this no matter what.

cameel:
> Would that really help that much?

I think it would. As I see it, the issue is all about source archives created by github. If these files were created only on `git archive`, we'd have proper source archives and keep everything working as it used to in other cases. We'd not have to worry that the files may exist but still contain the template. So we'd be done here already.

cmake/scripts/buildinfo.cmake
```diff
@@ -32,6 +32,19 @@ endif()
 if (EXISTS ${ETH_SOURCE_DIR}/commit_hash.txt)
 	file(READ ${ETH_SOURCE_DIR}/commit_hash.txt SOL_COMMIT_HASH)
 	string(STRIP ${SOL_COMMIT_HASH} SOL_COMMIT_HASH)
+	message(STATUS "SOL_COMMIT_HASH: ${SOL_COMMIT_HASH}")
+	if (SOL_COMMIT_HASH MATCHES [$][F][o][r][m][a][t][:])
+		message(STATUS "SOL_COMMIT_HASH: ${SOL_COMMIT_HASH}")
+		# we are not on a release branch, so we will have to generate the commit hash
```
ekpyron:
```suggestion
		# we are a git checkout, so we will have to generate the commit hash via the git command
```

cmake/scripts/buildinfo.cmake
```diff
@@ -32,6 +32,19 @@ endif()
...
+		execute_process(
+			COMMAND git --git-dir=${ETH_SOURCE_DIR}/.git --work-tree=${ETH_SOURCE_DIR} rev-parse --short=8 HEAD
+			OUTPUT_VARIABLE SOL_COMMIT_HASH OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+		)
+		execute_process(
+			COMMAND git --git-dir=${ETH_SOURCE_DIR}/.git --work-tree=${ETH_SOURCE_DIR} diff HEAD --shortstat
+			OUTPUT_VARIABLE SOL_LOCAL_CHANGES OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+		)
+	endif()
 else()
```
ekpyron: This should be a fatal error now - the file will always be there.

cmake/scripts/buildinfo.cmake
```diff
@@ -32,15 +32,19 @@ endif()
 if (EXISTS ${ETH_SOURCE_DIR}/commit_hash.txt)
```
ekpyron, line 40: We should really make it an error if it doesn't exist instead of just doing this conditionally.
Does the ``file(READ`` thing cause an error if the file doesn't exist? If so we can just remove the ``ìf`` entirely.

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,19 @@ if (NOT ETH_BUILD_PLATFORM)
...
-# Logic here: If prerelease.txt exists but is empty, it is a non-pre release.
-# If it does not exist, create our own prerelease string
-if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
-else()
+# Logic here: If prerelease.txt is empty, or if it has the templated placeholder, it is a non-pre release.
+file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
+string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
+If (SOL_VERSION_PRERELEASE MATCHES [$][F][o][r][m][a][t][:])
```
cameel: Why this weird form? Can't CMake handle regexes like this?
```suggestion
If (SOL_VERSION_PRERELEASE MATCHES \$Format:)
```
Or, at worst, this?
```suggestion
If (SOL_VERSION_PRERELEASE MATCHES [$]Format:)
```

cameel: Also, It would probably be best to require the match to be at the beginning of the line to avoid weird false-positives:
```suggestion
If (SOL_VERSION_PRERELEASE MATCHES ^[$][F][o][r][m][a][t][:])
```

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,19 @@ if (NOT ETH_BUILD_PLATFORM)
 	set(ETH_BUILD_PLATFORM "unknown")
 endif()
 
-# Logic here: If prerelease.txt exists but is empty, it is a non-pre release.
-# If it does not exist, create our own prerelease string
-if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
-else()
+# Logic here: If prerelease.txt is empty, or if it has the templated placeholder, it is a non-pre release.
```
cameel: This comment is wrong. if it has the placeholder, it's a pre-prelease.

cameel (changes requested):

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,19 @@ if (NOT ETH_BUILD_PLATFORM)
...
-# Logic here: If prerelease.txt exists but is empty, it is a non-pre release.
-# If it does not exist, create our own prerelease string
-if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
-else()
+# Logic here: If prerelease.txt has the templated placeholder, it is a pre release.
+file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
+string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
+If (SOL_VERSION_PRERELEASE MATCHES ^[$]Format:)
```
ekpyron: That's not the same as had happened before, is it?
If we are in a source archive, then ``prerelease.txt`` will not contain the placeholder, but it *will* contain whatever ``$Format:%d$`` resolves to - which won't be "empty". So releases will become prereleases... won't they?

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
 	set(ETH_BUILD_PLATFORM "unknown")
 endif()
 
-# Logic here: If prerelease.txt exists but is empty, it is a non-pre release.
-# If it does not exist, create our own prerelease string
+
 if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
+	message(FATAL_ERROR "The usage of the file prerelease.txt is depreciated. Please use the file prerelease.lock instead.")
```
chriseth:
```suggestion
	message(FATAL_ERROR "The usage of the file prerelease.txt is deprecated. Please use the file prerelease.lock instead.")
```

chriseth: also maybe it should be `cmake/prerelease.lock`

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
+endif()
+
+if (SOLC_FORCE_RELEASE)
```
chriseth: When and where is this flag used?

chriseth:
The file `commit_hash.txt` is still mentioned in the cmake file further down. Please do full search for "commit_hash" of the whole repository.

scripts/build.sh
```diff
@@ -11,7 +11,7 @@ else
 fi
 
 if [[ "$(git tag --points-at HEAD 2>/dev/null)" == v* ]]; then
-	touch "${ROOTDIR}/prerelease.txt"
+	touch "${ROOTDIR}/cmake/prerelease.lock"
```
cameel: To build a release you now need to remove `prerelease.lock`.

.gitattributes
```diff
@@ -1 +1,3 @@
 *.sol linguist-language=Solidity
+commit_hash.txt export-subst
```
cameel: Now this file should no longer exist in the repo.

ReleaseChecklist.md
```diff
@@ -28,9 +28,8 @@
 
 ### Upload Release Artifacts
  - [ ] Switch to the tag that archives have to be created for.
- - [ ] Create the ``prerelease.txt`` file: (``echo -n > prerelease.txt``).
  - [ ] Run ``scripts/create_source_tarball.sh`` while being on the tag to create the source tarball. This will create the tarball in a directory called ``upload``.
- - [ ] Take the tarball from the upload directory (its name should be ``solidity_x.x.x.tar.gz``, otherwise ``prerelease.txt`` was missing in the step before) and upload the source tarball to the release page.
+ - [ ] Take the tarball from the upload directory (its name should be ``solidity_x.x.x.tar.gz``) and upload the source tarball to the release page.
```
cameel: This part should be removed. We'll no longer be uploading manually created tarballs.

cmake/prerelease.lock
```diff
@@ -0,0 +1 @@
+$Format:%d$
```
cameel: This is no longer supposed to be a template.

> We create a ``prerelease.txt`` file in the repository, but not in the repository root. I suggest ``cmake/prerelease.lock``. The contents can be something like ``If this file is present, a prerelease will be built by default. You can override this with cmake options ...<mention cmake options below>...``

Here's my suggestion for the content:
```suggestion
When this file is present, an executable that identifies as a pre-release is built by default.
If you want a release build instead, add `SOL_FORCE_RELEASE=On` to your `cmake` command to override the default.
If you want to customize the pre-release part in the version string, you can do it via the `SOL_PRERELEASE_STRING` CMake option.
```
**EDIT**: Replaced `SOLC_` prefix with `SOL_` (see comments below).

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
-# If it does not exist, create our own prerelease string
+
 if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
+	message(FATAL_ERROR "The usage of the file prerelease.txt is deprecated. Please use the file cmake/prerelease.lock instead.")
+endif()
+
+if (SOLC_FORCE_RELEASE)
+	set(SOL_VERSION_PRERELEASE, "")
```
cameel: When `SOLC_FORCE_RELEASE` is true, user must not use this option so it should already be empty. If it's not, report an error.

> If this option is true, ``SOLC_PRERELEASE_STRING`` needs to be unset/empty.

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
 	set(ETH_BUILD_PLATFORM "unknown")
 endif()
 
-# Logic here: If prerelease.txt exists but is empty, it is a non-pre release.
-# If it does not exist, create our own prerelease string
+
 if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
+	message(FATAL_ERROR "The usage of the file prerelease.txt is deprecated. Please use the file cmake/prerelease.lock instead.")
```
cameel: Deprecated sounds as a bit as if it would still be functional for some time. Let's just say it's not supported to make it clear.

And `cmake/prerelease.lock` works in a different way now. We only really kept it for the purpose of creating source archives. I think we should direct users towards the new options instead.

```suggestion
	message(FATAL_ERROR "The usage of the file prerelease.txt is no longer supported. Please use SOL_FORCE_RELEASE option if you want to build a release executable or SOL_PRERELEASE_STRING if you want to adjust the pre-release part of the version string.")
```
**EDIT**: Replaced `SOLC_` prefix with `SOL_` (see comments below).

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
+endif()
+
+if (SOLC_FORCE_RELEASE)
+	set(SOL_VERSION_PRERELEASE, "")
 else()
```
cameel: You can use `elseif` to make the `if` less nested:
```suggestion
elseif(SOL_PRERELEASE_STRING)
```
**EDIT**: Replaced `SOLC_` prefix with `SOL_` (see comments below).

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
-	string(TIMESTAMP SOL_VERSION_PRERELEASE "develop.%Y.%m.%d" UTC)
-	string(REPLACE .0 . SOL_VERSION_PRERELEASE "${SOL_VERSION_PRERELEASE}")
+	if (SOLC_PRERELEASE_STRING)
+		set(SOL_VERSION_PRERELEASE "${SOLC_PRERELEASE_STRING}")
+	else()
+		if (EXISTS ${ETH_SOURCE_DIR}/cmake/prerelease.lock)
```
cameel: Same here:
```suggestion
	elseif(EXISTS ${ETH_SOURCE_DIR}/cmake/prerelease.lock)
```

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
+		else ()
+			set(SOL_VERSION_PRERELEASE, "")
+		endif()
+	endif()
 endif()
 
 if (EXISTS ${ETH_SOURCE_DIR}/commit_hash.txt)
-	file(READ ${ETH_SOURCE_DIR}/commit_hash.txt SOL_COMMIT_HASH)
-	string(STRIP ${SOL_COMMIT_HASH} SOL_COMMIT_HASH)
+	message(FATAL_ERROR "The usage of the file commit_hash.txt is deprecated. Please use the cmake option SOLC_COMMIT_HASH instead.")
```
cameel: For consistency with my other suggestion:
```suggestion
	message(FATAL_ERROR "The usage of the file commit_hash.txt is no longer supported. Please use SOL_COMMIT_HASH option instead.")
```
**EDIT**: Replaced `SOLC_` prefix with `SOL_` (see comments below).

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,32 @@ if (NOT ETH_BUILD_PLATFORM)
...
+endif()
+
+if (SOLC_COMMIT_HASH)
+	set (SOL_COMMIT_HASH "${SOLC_COMMIT_HASH}")
```
cameel: I just noticed that we now have some variables in `SOL_` and `SOLC_` variants.

Let's use `SOL_` for everything to avoid that. I updated the spec in the issue: https://github.com/ethereum/solidity/issues/9720#issuecomment-1059322965

cmake/scripts/buildinfo.cmake
```diff
@@ -50,7 +63,7 @@ endif()
 
 if (NOT SOL_COMMIT_HASH)
 	message(FATAL_ERROR "Unable to determine commit hash. Either compile from within git repository or "
-		"supply a file called commit_hash.txt")
+		"supply a variable called SOLC_COMMIT_HASH")
```
cameel:
```suggestion
		"supply an option called SOL_COMMIT_HASH")
```

docs/conf.py
```diff
@@ -73,11 +73,12 @@ def setup(sphinx):
...
 # The full version, including alpha/beta/rc tags.
-if not os.path.isfile('../prerelease.txt') or os.path.getsize('../prerelease.txt') == 0:
-    release = version
-else:
+if not os.path.isfile('../cmake/prerelease.lock'):
     # This is a prerelease version
     release = version + '-develop'
+else:
+    # This is a Release version
+    release = version
```
cameel: This should be the other way around. When the file exists it's a pre-release. Otherwise it's a release.

docs/conf.py
```diff
@@ -73,11 +73,12 @@ def setup(sphinx):
 with open('../CMakeLists.txt', 'r', encoding='utf8') as f:
     version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1)
 # The full version, including alpha/beta/rc tags.
```
cameel: This comment is now wrong. The file no longer contains the full version.

scripts/create_source_tarball.sh
```diff
@@ -25,9 +25,9 @@ REPO_ROOT="$(dirname "$0")"/..
     git checkout-index -a --prefix="$SOLDIR"
     # Store the commit hash
     echo "$commithash" > "$SOLDIR/commit_hash.txt"
```
cameel: This needs to be removed. We no longer support this file.

scripts/create_source_tarball.sh
```diff
@@ -25,9 +25,9 @@ REPO_ROOT="$(dirname "$0")"/..
...
-    if [ -e prerelease.txt ] && [ ! -s prerelease.txt ]
+    if [ ! -e cmake/prerelease.lock ]
     then
-        cp prerelease.txt "$SOLDIR/"
+        cp cmake/prerelease.lock "$SOLDIR/"
```
cameel: This file will now be stored in the repository so the `git checkout-index` command should already copy it.

cameel: But anyway, we no longer need this script. You can replace most of its content with the `git archive` command.

cameel (changes requested):

cmake/prerelease.lock
```diff
@@ -0,0 +1,3 @@
+-> When this file is present, an executable that identifies as a pre-release is built by default.
```
chriseth: Do the `->` have a special meaning here?

cmake/prerelease.lock
```diff
@@ -0,0 +1,3 @@
...
+-> If you want a release build instead, add `SOL_FORCE_RELEASE=On` to your `cmake` command to override the default.
+-> If you want to customize the pre-release part in the version string, you can do it via the `SOL_PRERELEASE_STRING` CMake option.
```
chriseth: Do we make use / need this feature?

cmake/scripts/buildinfo.cmake
```diff
@@ -50,7 +61,7 @@ endif()
 
 if (NOT SOL_COMMIT_HASH)
 	message(FATAL_ERROR "Unable to determine commit hash. Either compile from within git repository or "
-		"supply a file called commit_hash.txt")
+		"supply am option called SOL_COMMIT_HASH")
```
chriseth:
```suggestion
		"supply an option called SOL_COMMIT_HASH")
```

scripts/build.sh
```diff
@@ -11,7 +11,7 @@ else
 fi
 
 if [[ "$(git tag --points-at HEAD 2>/dev/null)" == v* ]]; then
-	touch "${ROOTDIR}/prerelease.txt"
+	rm "${ROOTDIR}/cmake/prerelease.lock"
```
chriseth: Shouldn't we rather use a cmake option instead of modifying the files?

scripts/create_source_tarball.sh
chriseth: Should we just delete the file?

chriseth: Or at least call `git archive` instead.

cmake/prerelease.lock
cameel: It should not. At least that was not in the spec ;) I'd remove it.

cmake/prerelease.lock
cameel: We use that for nightlies: https://github.com/ethereum/solc-bin/blob/gh-pages/.github/workflows/nightly-emscripten.yml#L53

And for CI builds we also put `ci` there: https://github.com/ethereum/solidity/blob/43f29c00da02e19ff10d43f7eb6955d627c57728/scripts/ci/build.sh#L15

I don't think we strictly need it. We could just as well have all non-release builds identify as `develop`.

This reminds, me, @nishant-sachdeva can you prepare a PR that updates [the nightly build action in solc-bin](https://github.com/ethereum/solc-bin/blob/gh-pages/.github/workflows/nightly-emscripten.yml) to use the new system? We should merge it immediately after we merge this one.

scripts/build.sh
cameel: True. There's a `cmake` command below and we should just add the force release option there.

scripts/create_source_tarball.sh
cameel: I think it would be fine to delete it completely. I don't think we were using it for anything other that release checklist (@nishant-sachdeva please do a search for `create_source_tarball.sh` to make sure) and I doubt people are building their own source packages.

cameel:
As for build errors, please look carefully through the log. The error is there, just not at the very bottom (probably because of the parallel compilation):

```
CMake Error at cmake/scripts/buildinfo.cmake:24 (message):
  The usage of the file prerelease.txt is no longer supported.  Please use
  SOL_FORCE_RELEASE option if you want to build a release executable or
  SOL_PRERELEASE_STRING if you want to adjust the pre-release part of the
  version string.
```

So it looks like we still have `prerelease.txt` there. Please check what is creating it and why. I think it's probably the `build.sh` script that was not correctly updated and does not use the new cmake options yet.

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,30 @@ if (NOT ETH_BUILD_PLATFORM)
...
+
+if (SOL_FORCE_RELEASE)
+	if (SOL_VERSION_PRERELEASE)
+		message(FATAL_ERROR "SOL_VERSION_PRERELEASE is set but SOL_FORCE_RELEASE is set. This is not allowed.")
+	endif()
+elseif(SOL_PRERELEASE_STRING)
+	set(SOL_VERSION_PRERELEASE "${SOL_PRERELEASE_STRING}")
+elseif(EXISTS ${ETH_SOURCE_DIR}/cmake/prerelease.lock)
+		string(TIMESTAMP SOL_VERSION_PRERELEASE "develop.%Y.%m.%d" UTC)
+		string(REPLACE .0 . SOL_VERSION_PRERELEASE "${SOL_VERSION_PRERELEASE}")
```
cameel:
```suggestion
	string(TIMESTAMP SOL_VERSION_PRERELEASE "develop.%Y.%m.%d" UTC)
	string(REPLACE .0 . SOL_VERSION_PRERELEASE "${SOL_VERSION_PRERELEASE}")
```

cmake/scripts/buildinfo.cmake
```diff
@@ -19,19 +19,30 @@ if (NOT ETH_BUILD_PLATFORM)
...
 
 if (EXISTS ${ETH_SOURCE_DIR}/commit_hash.txt)
-	file(READ ${ETH_SOURCE_DIR}/commit_hash.txt SOL_COMMIT_HASH)
-	string(STRIP ${SOL_COMMIT_HASH} SOL_COMMIT_HASH)
+	message(FATAL_ERROR "The usage of the file commit_hash.txt is no longer supported. Please use SOL_COMMIT_HASH option instead.")
+endif()
+
+if (SOL_COMMIT_HASH)
+	set (SOL_COMMIT_HASH "${SOL_COMMIT_HASH}")
 else()
```
cameel: This is now a no-op :)

scripts/create_source_tarball.sh
nishant-sachdeva: I've run the command
`git grep "create_source_tarball.sh" ` on the git repo and I haven't found any references to this file.

So, deleting it looks good to me

cmake/prerelease.lock
nishant-sachdeva: Yea, it doesn't. I felt presenting it as bullet points would make it more readable.
But yea, I'll remove it in the next push

cameel: In that case I'd just use `-` since that's the markdown syntax for bullet points.

But for me this is more like a single paragraph so I don't think these are necessary.

.circleci/config.yml
```diff
@@ -89,8 +89,8 @@ defaults:
   - setup_prerelease_commit_hash: &setup_prerelease_commit_hash
       name: Store commit hash and prerelease
       command: |
-        if [ "$CIRCLE_BRANCH" = release -o -n "$CIRCLE_TAG" ]; then echo -n > prerelease.txt; else date -u +"nightly.%Y.%-m.%-d" > prerelease.txt; fi
-        echo -n "$CIRCLE_SHA1" > commit_hash.txt
+        if [ "$CIRCLE_BRANCH" = release -o -n "$CIRCLE_TAG" ]; then rm cmake/prerelease.lock; else date -u +"nightly.%Y.%-m.%-d" > cmake/prerelease.lock; fi
```
chriseth: I thought we are not using the contents of the file anymore. Shouldn't we set a cmake flag instead?

chriseth: Would it be better to use "git archive" instead for the release build? If we do it differently, then building from an archive might lead to a different result than the compilation of the release build on circleci.

docs/conf.py
```diff
@@ -72,12 +72,12 @@ def setup(sphinx):
...
 with open('../CMakeLists.txt', 'r', encoding='utf8') as f:
     version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1)
-# The full version, including alpha/beta/rc tags.
-if not os.path.isfile('../prerelease.txt') or os.path.getsize('../prerelease.txt') == 0:
-    release = version
-else:
+if os.path.isfile('../cmake/prerelease.lock'):
     # This is a prerelease version
     release = version + '-develop'
+else:
```
chriseth, line 78: Please double-check that the file is actually deleted for a release build.

scripts/docker_deploy_manual.sh
```diff
@@ -19,13 +19,13 @@ cd "$DIR"
 git clone --depth 2 https://github.com/ethereum/solidity.git -b "$branch"
```
chriseth: If this just runs a full checkout, maybe we should just use `git archive` here as well?

scripts/wasm-rebuild/docker-scripts/rebuild_current.sh
```diff
@@ -7,8 +7,6 @@
 
 echo "========== STAGE 1: PREPARE ========== ($(date))"
 COMMIT_DATE="$(git show -s --format=%cI HEAD)"
-git rev-parse --short=8 HEAD >commit_hash.txt
```
chriseth, line 10: This script is meant to build older versions, I think we should keep this here (or make it depend on the version).

.circleci/config.yml
cameel: > I thought we are not using the contents of the file anymore. Shouldn't we set a cmake flag instead?

Yes, this should be using the flag. The flag is now the preferred mechanism.

> Would it be better to use "git archive" instead for the release build?

Sounds good to me.

But we might be better off doing it as a separate PR to let @nishant-sachdeva finally wrap this one up.
Switching to a flag is straightforward while switching to building from source might involve some extra issues with ensuring that dependencies are downloaded, etc. so it might add some additional review cycles here.

.circleci/build_win.ps1
```diff
@@ -3,7 +3,8 @@ $ErrorActionPreference = "Stop"
 cd "$PSScriptRoot\.."
 
 if ("$Env:FORCE_RELEASE" -Or "$Env:CIRCLE_TAG") {
-	New-Item prerelease.txt -type file
+	# New-Item prerelease.txt -type file
+	# commented this line to stop generation of prerelease.txt file. The line should be removed and the script refactored to accomodate this.
```
cameel: You need to replace it with passing a flag on the `..\deps\cmake\bin\cmake` line below. If think we should just put the whole `if` over the `cmake` ccommand.

.circleci/build_win.ps1
```diff
@@ -13,6 +14,8 @@ else {
 	$last_commit_timestamp = git log -1 --date=unix --format=%cd HEAD
 	$last_commit_date = (Get-Date -Date "1970-01-01 00:00:00Z").toUniversalTime().addSeconds($last_commit_timestamp).ToString("yyyy.M.d")
 	-join("ci.", $last_commit_date) | out-file -encoding ascii prerelease.txt
+	# removed the out-file command here, to stop the output to the prereleaste.txt. 
+	# But not fully sure what to do with the rest of the command here.
```
cameel: You should put this `if` over the cmake command and use `SOL_PRERELEASE_STRING` instead of the file.
It should look something like this:

```powershell
if ("$Env:FORCE_RELEASE" -Or "$Env:CIRCLE_TAG") {
	..\deps\cmake\bin\cmake --build . -j 10 --target install --config Release -DSOL_FORCE_RELEASE=On
}
else {
	$last_commit_timestamp = git log -1 --date=unix --format=%cd HEAD
	$last_commit_date = (Get-Date -Date "1970-01-01 00:00:00Z").toUniversalTime().addSeconds($last_commit_timestamp).ToString("yyyy.M.d")
	$prerelease_string = -join("ci.", $last_commit_date) | out-file -encoding ascii prerelease.txt
	..\deps\cmake\bin\cmake --build . -j 10 --target install --config Release -DSOL_PRERELEASE_STRING="$prerelease_string"
}
```

I'm not really sure if the `$prerelease_string = ` line above is correct - please check Powershell syntax for assigning result of such a command to a variable. Also `-DSOL_PRERELEASE_STRING="$prerelease_string"` might not be the correct way to quote on Windows.

You can SSH into a Windows container on CircleCI and try it out in PowerShell before you add it to the PR.

snap/snapcraft.yaml
```diff
@@ -32,7 +32,8 @@ parts:
     override-build: |
       if git describe --exact-match --tags 2> /dev/null
       then
-        echo -n > ../src/prerelease.txt
+        # echo -n > ../src/prerelease.txt
+        # commented this line to stop generation of prerelease.txt. Need to check how to refactor the process to remove this line completely.
```
cameel: Snapcraft config is not actively maintained. See https://github.com/ethereum/solidity/pull/12507#issuecomment-1011971783 for some context/discussion.

In this case the script was creating an empty `prerelease.txt` to get a release build. 
I think that the safest thing to do here is to replace this with a line that removes `prerelease.lock` - since I'm not sure if the `cmake install` command below would be the right place to add the new flag or not. At least with `prerelease.lock` we know that it will work.

scripts/wasm-rebuild/docker-scripts/rebuild_current.sh
cameel: Yeah. For older compilers we still need `prerelease.txt` here. We'll update it later if we ever need to use it to build newer ones. I think I'd just add a comment pointing out that this won't work after 0.8.13:

```bash
# FIXME: The prerelease.txt mechanism is supported only by versions up to 0.8.13.
```

scripts/release_ppa.sh
```diff
@@ -117,11 +117,10 @@ version=$("$(dirname "$0")/get_version.sh")
 commithash=$(git rev-parse --short=8 HEAD)
 commitdate=$(git show --format=%ci HEAD | head -n 1 | cut - -b1-10 | sed -e 's/-0?/./' | sed -e 's/-0?/./')
 
-echo "$commithash" > commit_hash.txt
```
cameel, line 120: I wonder why we were using `commit_hash.txt` here. @chriseth we should still get the right commit hash without it, right?

scripts/docker_deploy_manual.sh
cameel: Yeah, that would be cleaner.

scripts/docker_deploy_manual.sh
```diff
@@ -19,13 +19,13 @@ cd "$DIR"
...
+# echo -n "$commithash" > commit_hash.txt
 version=$("$(dirname "$0")/get_version.sh")
 if [ "$branch" = "release" ] || [ "$branch" = v"$version" ]
 then
-    echo -n > prerelease.txt
+    rm cmake/prerelease.lock
 else
-    date -u +"nightly.%Y.%-m.%-d" > prerelease.txt
+    date -u +"nightly.%Y.%-m.%-d" > cmake/prerelease.lock
 fi
```
cameel: Putting date in `prerelease.txt` won't work. You should replace this with cmake flags.

scripts/ci/build_emscripten.sh
```diff
@@ -49,17 +49,17 @@ cd $WORKSPACE
...
 	# Use last commit date rather than build date to avoid ending up with builds for
 	# different platforms having different version strings (and therefore producing different bytecode)
 	# if the CI is triggered just before midnight.
-	TZ=UTC git show --quiet --date="format-local:%Y.%-m.%-d" --format="ci.%cd" >prerelease.txt
-fi
-if [ -n "$CIRCLE_SHA1" ]
-then
-	echo -n "$CIRCLE_SHA1" >commit_hash.txt
+	TZ=UTC git show --quiet --date="format-local:%Y.%-m.%-d" --format="ci.%cd" >cmake/prerelease.lock
 fi
```
cameel: `prerelease.lock` no longer works like the old `prerelease.txt`. Writing a date there will not make it appear in the version string.

Anyway, please replace this with passing the right flags to the `emcmake cmake` command below.

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,31 @@ if (NOT ETH_BUILD_PLATFORM)
...
-# If it does not exist, create our own prerelease string
+
 if (EXISTS ${ETH_SOURCE_DIR}/prerelease.txt)
-	file(READ ${ETH_SOURCE_DIR}/prerelease.txt SOL_VERSION_PRERELEASE)
-	string(STRIP "${SOL_VERSION_PRERELEASE}" SOL_VERSION_PRERELEASE)
-else()
+	message(FATAL_ERROR "The usage of the file prerelease.txt is no longer supported. Please use SOL_FORCE_RELEASE option if you want to build a release executable or SOL_PRERELEASE_STRING if you want to adjust the pre-release part of the version string.")
+endif()
+
+if (SOL_FORCE_RELEASE)
```
aarlt: Shouldn't `SOL_FORCE_RELEASE` and `SOL_VERSION_PRERELEASE` be somewhere defined as an CMake [`option`](https://cmake.org/cmake/help/latest/command/option.html)?

docs/conf.py
```diff
@@ -72,12 +72,12 @@ def setup(sphinx):
 # The short X.Y version.
 with open('../CMakeLists.txt', 'r', encoding='utf8') as f:
     version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1)
-# The full version, including alpha/beta/rc tags.
-if not os.path.isfile('../prerelease.txt') or os.path.getsize('../prerelease.txt') == 0:
-    release = version
-else:
+if os.path.isfile('../cmake/prerelease.lock'):
     # This is a prerelease version
     release = version + '-develop'
```
aarlt, line 77: Is `-develop` always correct here? Wouldn't it make sense to use `SOL_VERSION_PRERELEASE`? Not sure whether this really makes sense here, but maybe CMake could write `SOL_VERSION_PRERELEASE` to a file somewhere, and this information is then just used here.

aarlt: Ah ok, this line was not touched. So maybe we don't want to change that.

cmake/scripts/buildinfo.cmake
cameel: Yeah, I think they should. Good point.

docs/conf.py
cameel: Yeah, I think it's fine as is. We don't need really that level of granularity for docs. I mean, would be nice to have the exact version there but I also think it's not worth complicating the build system.

cmake/scripts/buildinfo.cmake
cameel: @nishant-sachdeva 
> Could you add some detail to https://github.com/ethereum/solidity/pull/12717#discussion_r834722129
> I'm not very well versed with cmake scripting and am not very sure what you mean by "adding the options"

See how we use it here: https://github.com/ethereum/solidity/blob/c4909e99c1015233d6937d229058cde52615f246/CMakeLists.txt#L35-L37

And also see [docs for `option`](https://cmake.org/cmake/help/latest/command/option.html).

scripts/ci/build_emscripten.sh
cameel: > For https://github.com/ethereum/solidity/pull/12717#discussion_r833473218, are any cmake flags even required ?
> If it's a release, I can just delete the prerelease.lock
> And it's if not a release, I can just leave the prerelease.lock file where it is and the buildinfo.cmake script should take care of generating the earstwhile prerelease string

You should be looking at the other way around. Is `prerelease.lock` even required? If you can achieve it with cmake options, that's a cleaner solution. The only reason we keep the `prerelease.lock` mechanism is that we need it for the archive generated by github.

scripts/docker_deploy_manual.sh
cameel: > Could you elaborate https://github.com/ethereum/solidity/pull/12717#discussion_r833470661 further ?
> Exactly what change are we looking for ?

@chriseth meant here that instead of building directly inside a checked-out repo maybe we should use `git archive` to build a source archive and build from that instead.

cmake/scripts/buildinfo.cmake
aarlt: @cameel exactly. @nishant-sachdeva basically just define `SOL_FORCE_RELEASE` and `SOL_VERSION_PRERELEASE` as an option in `solidity/CMakeLists.txt`.

scripts/docker_deploy_manual.sh
nishant-sachdeva: github does not allow to use git archive remotely on it's repositories. So, I suppose this option is not available to us

cameel: You don't have to use it remotely. You can still clone it. The idea is just to unpack the archive to a fresh dir and build in it rather than in a repo checkout.

Anyway, as I said in another thread here, I think it would be fine to just add an issue for that and do it separately to keep the scope of this PR limited.

.circleci/build_win.ps1
```diff
@@ -2,22 +2,25 @@ $ErrorActionPreference = "Stop"
...
 cd "$PSScriptRoot\.."
 
+mkdir build
+cd build
+$boost_dir=(Resolve-Path $PSScriptRoot\..\deps\boost\lib\cmake\Boost-*)
+
 if ("$Env:FORCE_RELEASE" -Or "$Env:CIRCLE_TAG") {
-	New-Item prerelease.txt -type file
 	Write-Host "Building release version."
+	del cmake\prerelease.lock
```
cameel: Please use `SOL_FORCE_RELEASE` option here.

.circleci/build_win.ps1
```diff
@@ -2,22 +2,25 @@ $ErrorActionPreference = "Stop"
...
 }
 else {
 	# Use last commit date rather than build date to avoid ending up with builds for
 	# different platforms having different version strings (and therefore producing different bytecode)
 	# if the CI is triggered just before midnight.
 	$last_commit_timestamp = git log -1 --date=unix --format=%cd HEAD
 	$last_commit_date = (Get-Date -Date "1970-01-01 00:00:00Z").toUniversalTime().addSeconds($last_commit_timestamp).ToString("yyyy.M.d")
-	-join("ci.", $last_commit_date) | out-file -encoding ascii prerelease.txt
+	-join("ci.", $last_commit_date)
```
cameel: This line seems superfluous now.

.gitattributes
```diff
@@ -1 +1,4 @@
 *.sol linguist-language=Solidity
+prerelease.lock export-ignore
```
cameel: Isn't this file now inside `cmake/` subdir?
```suggestion
/cmake/prerelease.lock export-ignore
```

cameel: Also, the other files might need the `/` at the beginning if paths here work like in `.gitignore`.

.gitattributes
```diff
@@ -1 +1,4 @@
...
+commit_hash.txt export-subst
+current_tag.txt export-subst
```
cameel: Paths here might need the `/` at the beginning if they work like in `.gitignore`.

CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ endif()
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
 option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
+option(SOL_FORCE_RELEASE "Force the creation of a Release Version" OFF)
```
```suggestion
option(SOL_FORCE_RELEASE "Build a release binary" OFF)
```

Maybe we should really call this `SOL_RELEASE_VERSION`? Force sounds a bit as if you were doing something you should not be doing but this is now **the** way to build a release binary.

CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ endif()
...
+option(SOL_VERSION_PRERELEASE "Getting the latest version of a prerelease" ON)
```
```suggestion
option(SOL_VERSION_PRERELEASE "Extra pre-release information to be included in the version string. By default it's the name of the main branch and the date of the last commit in it. Must be empty when doing a release build." ON)
```

cameel: It's a string value. If you set it to `ON`, then that text will be included in the version string.
```suggestion
option(SOL_VERSION_PRERELEASE "Getting the latest version of a prerelease" "")
```

CMakeLists.txt
nishant-sachdeva: Sure, will change it in the next iteration

scripts/release_ppa.sh
ekpyron: It does not pack ``.git`` below, so it won't know the hash anymore. I wonder if we should use ``git archive`` here to create an archive from the git checkout and then append the dependencies into that archive...

cmake/current_tag.txt
```diff
@@ -0,0 +1 @@
+$Format:%(describe)$
```
ekpyron: What's this used for?

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,40 @@ if (NOT ETH_BUILD_PLATFORM)
...
+	if (SOL_COMMIT_HASH STREQUAL "$Format:%H$")
+		execute_process(
+			COMMAND git --git-dir=${ETH_SOURCE_DIR}/.git --work-tree=${ETH_SOURCE_DIR} rev-parse --short=8 HEAD
+			OUTPUT_VARIABLE SOL_COMMIT_HASH OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+		)
+		execute_process(
+			COMMAND git --git-dir=${ETH_SOURCE_DIR}/.git --work-tree=${ETH_SOURCE_DIR} diff HEAD --shortstat
+			OUTPUT_VARIABLE SOL_LOCAL_CHANGES OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+		)
+	endif()
```
ekpyron, line 52: We could also query ``git tag --points-at HEAD`` and check if it's a ``v*`` tag...
Ideally, we would *default* ``SOL_FORCE_RELEASE`` to ``true`` in that case, but that may be a hassle, so we could also just *always* force a release build then...

scripts/build.sh
```diff
@@ -10,14 +10,15 @@ else
...
 
-if [[ "$(git tag --points-at HEAD 2>/dev/null)" == v* ]]; then
-	touch "${ROOTDIR}/prerelease.txt"
-fi
-
 mkdir -p "${BUILDDIR}"
 cd "${BUILDDIR}"
 
-cmake .. -DCMAKE_BUILD_TYPE="$BUILD_TYPE" "${@:2}"
+if [[ "$(git tag --points-at HEAD 2>/dev/null)" == v* ]]; then
```
ekpyron, line 16: Is this robust in the case there are several tags at ``HEAD``?
That usually won't happen, but if it ever does, this might break...
It should really check, whether *any* tag at the ``HEAD`` is a ``v*`` tag, shouldn't it?

.circleci/build_win.ps1
```diff
@@ -2,22 +2,24 @@ $ErrorActionPreference = "Stop"
...
 cd "$PSScriptRoot\.."
 
+mkdir build
+cd build
+$boost_dir=(Resolve-Path $PSScriptRoot\..\deps\boost\lib\cmake\Boost-*)
+
 if ("$Env:FORCE_RELEASE" -Or "$Env:CIRCLE_TAG") {
-	New-Item prerelease.txt -type file
 	Write-Host "Building release version."
+	..\deps\cmake\bin\cmake --build . -j 10 --target install --config Release -DSOL_RELEASE_VERSION="On"
```
ekpyron, line 11: The cmake options here and below don't seem to match the new cmake options defined below, do they?
Also it doesn't seem nice to invoke ``cmake`` here and then again below...

CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ endif()
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
 option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
+option(SOL_FORCE_RELEASE "Build a release binary" OFF)
+option(SOL_RELEASE_VERSION "Extra pre-release information to be included in the version string. By default it's the name of the main branch and the date of the last commit in it. Must be empty when doing a release build." "")
```
ekpyron: This should be
```suggestion
set(SOL_RELEASE_VERSION "" CACHE STRING "Extra pre-release information to be included in the version string. By default it's the name of the main branch and the date of the last commit in it. Must be empty when doing a release build.")
```
shouldn't it?

cmake/scripts/buildinfo.cmake
ekpyron: Why was this marked as resolved?

cmake/scripts/buildinfo.cmake
```diff
@@ -19,20 +19,40 @@ if (NOT ETH_BUILD_PLATFORM)
...
+endif()
+
+if (NOT SOL_COMMIT_HASH)
```
ekpyron: Do we need this case at all? The commit_hash.txt file should always exist, so ``SOL_COMMIT_HASH`` should always be set at this point, shouldn't it?

CMakeLists.txt
ekpyron: Also ``SOL_VERSION_PRERELEASE`` needs to be an option here as well, shouldn't it?

cmake/scripts/buildinfo.cmake
nishant-sachdeva: Removing this case in the latest push.

ekpyron:
I think it's easier to redo this properly at some point rather than keeping this PR alive, so I'm closing for now.

PR #12719, opened 2022-02-28
SAT solver binary, proof logging, min-2-long cls, vector tristate

msooseth:
Hi,

This adds:
- Proof logging using FRAT, see https://github.com/digama0/frat
- All clauses are at least 2-long
- tristate value of variables is now an enum, with a vector, rather than a map, for fast lookup -- O(1) lookup instead of O(log(n))
- Binary that can parse a DIMACS input and generate the solution + proof

Example:
```
$ cat a.cnf 
p cnf 2 4
1 2 0
-1 -2 0
1 -2 0
-1 2 0
$ ./satsolver a.cnf proof.frat 
====
(x0, x1)
(~x0, ~x1)
(x0, ~x1)
(~x0, x1)
====
Propagating.
[...]
 - Propagate resulted in conflict because x1 is also false.
Unsatisfiable
s UNSATISFIABLE
$ ./frat-rs fratchk proof.frat 
4 orig + 6 added - 4 deleted - 6 finalized = 0
6 missing proofs (100.0%)
type 0: 6
matesoos@tiresias:tools$ echo $?
0
```

Hence, `frat-rs` verified our proof. Notice that the lines starting with `o` in `proof.frat` MUST match the input CNF, which they do (except for 1st element, which is simply an arbitrary clause ID):
```
$ grep "^o" proof.frat | cut -d " " -f 3-
1 2 0
-1 -2 0
1 -2 0
-1 2 0
```

matching the input CNF 1-to-1. Hence, `frat-rs` verified that the proof steps were correct.

### DIFF
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -385,4 +385,3 @@ void CDCL::write_final_proof_clauses()
 	}
 	*proof << "f " << unsat_clause_ID << " 0\n";
 }
-
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -45,6 +45,10 @@ CDCL::CDCL(
 
 optional<CDCL::Model> CDCL::solve()
 {
+	if (!ok) {
+		return nullopt;
+	}
+
 	cout << "====" << endl;
 	for (unique_ptr<Clause> const& c: m_clauses)
 		cout << toString(*c) << endl;
@@ -72,8 +76,13 @@ optional<CDCL::Model> CDCL::solve()
 			for (size_t i = 1; i < learntClause.size(); i++)
 				solAssert(isAssignedFalse(learntClause.at(i)));
 
-			addClause(move(learntClause));
-			enqueue(m_clauses.back()->front(), &(*m_clauses.back()));
+			if (learntClause.size() == 1) {
+				assert(currentDecisionLevel() == 0);
+				enqueue(learntClause[0], nullptr);
+			} else {
+				addClause(move(learntClause));
+				enqueue(m_clauses.back()->front(), &(*m_clauses.back()));
+			}
 		}
 		else
 		{
@@ -96,6 +105,14 @@ optional<CDCL::Model> CDCL::solve()
 
 void CDCL::setupWatches(Clause& _clause)
 {
+	assert(_clause.size() >= 2);
+
+	// NOTE: the 2nd literal is always unassigned.
+	// The 1st literal can be assigned ONLY in case we are attaching a learnt clause.
+	assert(!isAssigned(_clause[1]));
+	assert(!isAssigned(_clause[0]) || isAssignedTrue(_clause[0]));
+
+
 	for (size_t i = 0; i < min<size_t>(2, _clause.size()); i++)
 		m_watches[_clause.at(i)].push_back(&_clause);
 }
@@ -115,13 +132,14 @@ optional<Clause> CDCL::propagate()
 		for (; it != end; ++it)
 		{
 			Clause& clause = **it;
+			assert(clause.size() >= 2);
 			cout << " watch clause: " << toString(clause) << endl;
 
 			solAssert(!clause.empty());
 			if (clause.front() != falseLiteral)
 				swap(clause[0], clause[1]);
 			solAssert(clause.front() == falseLiteral);
-			if (clause.size() >= 2 && isAssignedTrue(clause[1]))
+			if (isAssignedTrue(clause[1]))
 			{
 				// Clause is already satisfied, keezp the watch.
 				cout << " -> already satisfied by " << toString(clause[1]) << endl;
@@ -143,12 +161,9 @@ optional<Clause> CDCL::propagate()
 
 			// We did not find a new watch, i.e. all literals starting from index 2
 			// are false, thus clause[1] has to be true (if it exists)
-			if (clause.size() == 1 || isAssignedFalse(clause[1]))
+			if (isAssignedFalse(clause[1]))
 			{
-				if (clause.size() >= 2)
-					cout << " - Propagate resulted in conflict because " << toString(clause[1]) << " is also false." << endl;
-				else
-					cout << " - Propagate resulted in conflict since clause is single-literal." << endl;
+				cout << " - Propagate resulted in conflict because " << toString(clause[1]) << " is also false." << endl;
 				// Copy over the remaining watches and replace.
 				while (it != end) watchReplacement.emplace_back(move(*it++));
 				m_watches[falseLiteral] = move(watchReplacement);
@@ -232,6 +247,36 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
 
 void CDCL::addClause(Clause _clause)
 {
+	if (!ok) {
+		return;
+	}
+
+	vector<Literal> lits;
+	for (const auto& l: _clause.lits) {
+		if (isAssignedTrue(l)) {
+			// Clause is satisfied, nothing to do.
+			return;
+		}
+		if (isAssignedFalse(l)) {
+			// Remove literal from clause.
+			continue;
+		}
+		lits.push_back(l);
+	}
+
+	// Empty clause, set UNSAT and return.
+	if (lits.size() == 0) {
+		*proof << clauseID++ << "0\n";
+		ok = false;
+		return;
+	}
+
+	// Unit clause, enqueue fact.
+	if (lits.size() == 1) {
+		enqueue(lits[0], nullptr);
+		return;
+	}
+	_clause = lits;
 	m_clauses.push_back(make_unique<Clause>(move(_clause)));
 	setupWatches(*m_clauses.back());
 }
@@ -241,7 +286,8 @@ void CDCL::enqueue(Literal const& _literal, Clause const* _reason)
 	cout << "Enqueueing " << toString(_literal) << " @" << currentDecisionLevel() << endl;
 	if (_reason)
 		cout << "  because of " << toString(*_reason) << endl;
-	// TODO assert that assignmnets was unknown
+
+	assert(!isAssigned(_literal));
 	m_assignments[_literal.variable] = _literal.positive;
 	m_levelForVariable[_literal.variable] = currentDecisionLevel();
 	if (_reason)
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -319,7 +319,7 @@ void CDCL::enqueue(Literal const& _literal, Clause const* _reason)
 		cout << "  because of " << toString(*_reason) << endl;
 
 	assert(value(_literal) == TriState::t_unset());
-	m_assignments[_literal.variable] = _literal.positive;
+	m_assignments[_literal.variable] = TriState(_literal.positive);
 	m_levelForVariable[_literal.variable] = currentDecisionLevel();
 	if (_reason)
 		m_reason[_literal] = _reason;
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -75,7 +75,7 @@ optional<CDCL::Model> CDCL::solve()
 				{
 					unsat_clause_ID = clauseID++;
 					*proof << "a " << unsat_clause_ID << " 0\n";
-					write_final_proof_clauses();
+					writeFinalProofClauses();
 				}
 				return nullopt;
 			}
@@ -290,7 +290,7 @@ void CDCL::addClause(const vector<Literal>& _lits)
 	{
 		unsat_clause_ID = clauseID++;
 		*proof << "a " << unsat_clause_ID << " 0\n";
-		write_final_proof_clauses();
+		writeFinalProofClauses();
 		ok = false;
 		return;
 	}
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -376,7 +376,7 @@ string CDCL::toProofString(Clause const& _clause) const
 	return joinHumanReadable(literals, " ");
 }
 
-void CDCL::write_final_proof_clauses()
+void CDCL::writeFinalProofClauses()
 {
 	assert(proof);
 	assert(unsat_clause_ID != 0);
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -39,7 +39,7 @@ CDCL::CDCL(
 	m_variables(move(_variables)),
 	proof(_proof)
 {
-	m_assignments.resize(m_variables.size(), tristate_unset);
+	m_assignments.resize(m_variables.size(), TriState::t_unset());
 	for (const auto& clause: _clauses)
 		addClause(clause);
 
@@ -85,9 +85,9 @@ optional<CDCL::Model> CDCL::solve()
 				*proof << "a " << ID << " " << toProofString(learntClause) << " 0\n";
 
 			solAssert(!learntClause.empty());
-			solAssert(!isAssigned(learntClause.front()));
+			solAssert(value(learntClause.front()) == TriState::t_unset());
 			for (size_t i = 1; i < learntClause.size(); i++)
-				solAssert(isAssignedFalse(learntClause.at(i)));
+				solAssert(value(learntClause[i]) == TriState::t_false());
 
 			if (learntClause.size() == 1)
 			{
@@ -151,7 +151,7 @@ optional<Clause> CDCL::propagate()
 			if (clause.front() != falseLiteral)
 				swap(clause[0], clause[1]);
 			solAssert(clause.front() == falseLiteral);
-			if (isAssignedTrue(clause[1]))
+			if (value(clause[1]) == TriState::t_true())
 			{
 				// Clause is already satisfied, keezp the watch.
 				cout << " -> already satisfied by " << toString(clause[1]) << endl;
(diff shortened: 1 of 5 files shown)


libsolutil/CDCL.cpp
```diff
@@ -230,9 +254,56 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
...
 
-void CDCL::addClause(Clause _clause)
+void CDCL::addClause(const vector<Literal>& _lits)
 {
-	m_clauses.push_back(make_unique<Clause>(move(_clause)));
+	if (!ok) {
+		return;
+	}
+
+	Clause _clause {_lits, clauseID++};
```
chriseth:
```suggestion
	Clause clause{_lits, clauseID++};
```
(underscore only for parameters) and no space in front of expression-level parentheses or similar

libsolutil/CDCL.cpp
```diff
@@ -230,9 +254,56 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
 	return {move(learntClause), backtrackLevel};
 }
 
-void CDCL::addClause(Clause _clause)
+void CDCL::addClause(const vector<Literal>& _lits)
 {
-	m_clauses.push_back(make_unique<Clause>(move(_clause)));
+	if (!ok) {
+		return;
+	}
```
chriseth, line 264:
```suggestion
	if (!ok)
		return;
```
(no braces for a single statement)

libsolutil/CDCL.cpp
```diff
@@ -230,9 +254,56 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
...
+
+	Clause _clause {_lits, clauseID++};
+	if (proof)
+		*proof <<  "o " << _clause.ID << " " << toProofString(_clause) << " 0\n";
+
+	Clause _clause_updated;
```
chriseth:
```suggestion
	Clause clauseUpdated;
```
camelCase

libsolutil/CDCL.h
```diff
@@ -48,16 +49,106 @@ struct Literal
...
+	void push_back(const Literal& l)
+	{
+		lits.push_back(l);
+	}
+
+	std::vector<Literal> lits;
+	uint64_t ID;
+};
+
+enum class TriState {trisate_true, tristate_false, tristate_unset};
```
chriseth: Would be nice to have a class with implicit conversion from bool. Maybe `std::optional<bool>` would be an option, but it could have a rather high overhead.

libsolutil/CDCL.cpp
msooseth: Thanks! This has now been fixed :)

libsolutil/CDCL.cpp
msooseth: Thanks, this has now been fixed, very helpful! :)

libsolutil/CDCL.h
msooseth: Hmmm I think I could make this into a class and fix that. Let me have a go at it. But it can could be miseading to have implicit conversion, though I see what you mean. Let me have a go at it!

libsolutil/CDCL.cpp
msooseth: Thanks, yep, fixed, good catch! I'm learning :)

libsolutil/CDCL.h
msooseth: This was a great idea actually! Let me know what you think!

libsolutil/CDCL.h
```diff
@@ -48,16 +49,137 @@ struct Literal
...
+		{
+			assert(val == 2);
+			return "unset";
+		}
+	}
+
+
+private:
+	// Default value is UNSET
+	char val = 2;
```
chriseth:
```suggestion
	uint8_t val = 2;
```
(char can be a very weird thing depending on the architecture)

libsolutil/CDCL.h
```diff
@@ -48,16 +49,137 @@ struct Literal
...
+
+class TriState {
+public:
+	TriState(const bool b)
+	{
+		if (b)
+			val = 1;
+		else
+			val = 0;
+	}
```
chriseth:
```suggestion
	TriState(const bool b): val(b ? 1 : 0) {}

File: libsolutil/CDCL.h [hunk 7]
```diff
@@ -48,16 +49,137 @@ struct Literal
...
+
+
+private:
+	// Default value is UNSET
+	char val = 2;
+};
+
+const TriState trisate_true = TriState(true);
+const TriState tristate_false = TriState(false);
```
chriseth: Not needed

libsolutil/CDCL.h
```diff
@@ -48,16 +49,137 @@ struct Literal
...
+const TriState tristate_unset = TriState();
```
chriseth: Can you make this a static member of the class, so we can use it as `Tristate::unset` or `Tristate::unset()`?

libsolutil/CDCL.h
```diff
@@ -48,16 +49,137 @@ struct Literal
...
+	bool operator==(const TriState& other) const
+	{
+		return val == other.val;
+	}
+
+	bool operator!=(const TriState& other) const
+	{
+		return val != other.val;
+	}
+
```
chriseth, line 69: Could also add `operator~` (or `!`)

libsolutil/CDCL.h
```diff
@@ -113,7 +237,43 @@ class CDCL
...
 
+inline bool CDCL::isAssigned(Literal const& _literal) const
+{
+	return m_assignments[_literal.variable] != tristate_unset;
+}
+
+inline bool CDCL::isAssignedTrue(Literal const& _literal) const
+{
+	return isAssigned(_literal) &&
+		(m_assignments[_literal.variable].toBool() ^ !_literal.positive) == true;
```
chriseth: Isn't this just
```suggestion
	return m_assignments[_literal.variable] == _literal.positive;
```
now? If yes, do we actually still need the helper?

libsolutil/CDCL.h
msooseth: Yes, indeed. Actually, the worst part about `char*` is that it's always assumed that [it can alias](https://stackoverflow.com/questions/23848188/strict-aliasing-rule-and-char-pointers) to another pointer type so the compiler cannot use some optimizations. I wanted to use `char` just not to confuse, but `uint8_t` is what I would have preferred. Fixing now!

libsolutil/CDCL.h
msooseth: Thanks, cleaner this way! Fixed :)

libsolutil/CDCL.h
msooseth: I was a bit hesitant, because the `~` on `tristate_unset` is unclear. I'd prefer if people had to use `!x.toBool()` which makes it clearer to the user that the underlying value can be more than just a boolean. What do you think? I mean, it is more verbose with `!x.toBool()`, and too much verbosity can also lead to bugs.

libsolutil/CDCL.h
msooseth: The way this is done in most SAT solvers, is to have an inline function `TriState CDCL:value(const Literal& x) const` that returns a `TriState`. So then things get nice in the code, because you can call it like:

```
if (value(cl[0]) == tristate_true) {
  // blah
};
```

Also, `value` is then defined for variables, too, so you don't have to care if you put in a variable or a literal. Obviously, for variables, it just returns `m_assignments[var]` :smiley:

libsolutil/CDCL.h
chriseth: Yeah, I'm also unsure on the semantics of negation, although negation of unset to be unset should be fine.

libsolutil/CDCL.h
chriseth: Indeed, sounds good!

libsolutil/CDCL.h
msooseth: Yep, I made these into a member functions:

```
constexpr static TriState t_true();
constexpr static TriState t_false();
constexpr static TriState t_unset();
```

The constexpr means there is no overhead at all :)

msooseth: Thanks for the feedback, I think it's much cleaner this way!

libsolutil/CDCL.h
msooseth: Added this as part of the `constexpr static` system for `TriState`. We only use `m_assignments` when we need to assign a variable a value, i.e. only inside `enqueue()` and `cancelUntil()`. This makes the code easier to review, in my opinion, as value() cannot change the assignment of a variable.

libsolutil/CDCL.h
msooseth: That's true. I am using it in the code as a helper, that what we are checking is a tristate value. For example:

```
if (value(clause[1]) == TriState::t_true())
{
  // Clause is already satisfied, keep the watch.
```

Would otherwise be:

```
if (value(clause[1]) == true)
{
  // Clause is already satisfied, keep the watch.
```

Which is weird, because one would think that it's possible to then do:

```
if (value(clause[1]))
{
  // Clause is already satisfied, keezp the watch.
```

but for that, we'd need a a `bool` conversion to `TriState`:

```
class TriState{
[...]
	explicit operator bool() const
	{
		if (val == 0) return false;
		else if (val == 1) return true;
		else assert(false);
	}
}
```

But we need to have that `assert()` there, or we could face some serious issues. However, that assert will fire when it's `TriState::t_unset` which it will be, quite often. The reason why we need the `assert` is that if we don't have the assert, then we can only return `true` or `false`, and if we return `false` for both `TriState::t_false()` and `TriState::t_unset()`, then double negation will not be the same as no negation, which can be very unexpected for the programmer.

So overall, as far as I can see, it may be best to compare to the now `static constexpr` member function `TriState::t_true()`? What do you think?

tools/satsolver-main.cpp
```diff
@@ -0,0 +1,180 @@
...
+    parts.push_back(input);
+    return parts;
+}
+
+optional<vector<Literal>> parseLine(std::string& line)
+{
+	vector<Literal> cl;
+	bool end_of_clause = false;
+	auto parts = cutStringBySpace(line);
+	for (const auto& part: parts)
```
chriseth:
```suggestion
	for (const auto& part: line | ranges::views::split(' '))
```

tools/satsolver-main.cpp
```diff
@@ -0,0 +1,180 @@
...
+	{
+		assert(!end_of_clause);
+		const long lit = std::stol(part);
+		const long var = std::abs(lit);
+		if (var == 0)
+		{
+			//end of clause
+			end_of_clause = true;
+			continue;
```
chriseth, line 68: Shouldn't this be a `break`?

tools/satsolver-main.cpp
```diff
@@ -0,0 +1,180 @@
...
+    vector<string> parts;
+
+    size_t pos = 0;
+    while ((pos = input.find(" ")) != string::npos)
+	{
+        parts.push_back(input.substr(0, pos));
+        input.erase(0, pos + 1);
+    }
+    parts.push_back(input);
+    return parts;
```
chriseth:
```suggestion
	return input | ranges::views::split(' ') | ranges::to<vector<string>>();
```
(or just use it inline)

libsolutil/CDCL.h
```diff
@@ -48,16 +49,80 @@ struct Literal
...
+	constexpr static TriState t_true();
+	constexpr static TriState t_false();
+	constexpr static TriState t_unset();
+
+private:
+	// Default value is UNSET
+	uint8_t val = 2;
+};
+
+constexpr TriState TriState::t_true() {
```
chriseth, line 104: Why do you need these? Just to be more explicit?

libsolutil/CDCL.h
chriseth: I agree, the conversion to bool would be dangerous and maybe `value(clause[1]) == true` could look weird, but I would say that in this case, we should also not have an implicit conversion from bool, but only an explicit one.

tools/satsolver-main.cpp
msooseth: This is cool! Note, this requires `set_property(TARGET satsolver PROPERTY CXX_STANDARD 20)` in the `CMakeLIsts.txt` which may be a bit weird (to compile with a higher target than other binaries).

More importantly, it fails compilation at 

```
/home/matesoos/development/solidity/tools/satsolver-main.cpp:65:55: error:
no matching function for call to ‘std::__cxx11::basic_string<char>::basic_string(const std::ranges::split_view<std::ranges::ref_view<std::__cxx11::basic_string<char> >, std::ranges::single_view<char> >::_OuterIter<true>::value_type&)’
```

i.e.

```
const long lit = std::stol(part);
```

doesn't work. But I can't seem to fix this in a simple way :( Any pro tips to fix? I do like this `ranges` stuff, looks super-powerful, but somehow it doesn't spit out things that look like strings or chars. I tried `const long lit = std::stol(string(part));` but that doesn't work either. Do you have a pro tip to fix this?

tools/satsolver-main.cpp
msooseth: I did this so we catch if someone made a mistake, like this in the CNF input:
```
1 2 3 0 4
```

I set `end_of_clause` once I reach `0` and it `continue`-s. Then the 1st line after the `for` is `assert(!end_of_clause)`. This catches the issue:

```
$ cat a.cnf 
p cnf 2 4
1 2 0 3
-1 -2 0
1 -2 0
-1 2 0
$ ./satsolver a.cnf proof.frat 
satsolver: /solidity/tools/satsolver-main.cpp:64: std::optional<std::vector<solidity::util::Literal> > parseLine(std::string&): Assertion `!end_of_clause' failed.
Aborted (core dumped)
```

However, catching it with an `assert` is probably wrong. I now emit a written error and `exit(-1)`, which is better. It's not a code bug (i.e. not a place for an `assert`), but an user bug:

```
$ ./satsolver a.cnf proof.frat 
ERROR: trailing elements after finishing `0` at the end of a clause in CNF file
```

In my experience, one can chase their tail for a long time on issues like this, where their CNF generation script is incorrect. Simple checks like this tend to help the user (often, me :laughing: )

Let me know what you think!

tools/satsolver-main.cpp
msooseth: Hmmm this doesn't seem to compile:
```
/home/matesoos/development/solidity/tools/satsolver-main.cpp:46:66: error: ‘to’ is not a member of ‘std::ranges’
```

And indeed, [it doesn't seem to be](https://en.cppreference.com/w/cpp/ranges)? Maybe it's a typo? I can't seem to be able to find the right function, though. Which function did you mean?

libsolutil/CDCL.h
msooseth: We can add the negation operator, but it would make thing strange. Here is an example:

```
TriState a;
TriState b;
if (a == !b) cout << "Equal" << endl;
else cout << "Not equal" << end;
```

We get `Equal` output for this. However, if the comparison was: `a == b` (instead of `a == !b`), we would **still** get `Equal`... which I think is very strange and may make people start doubting their sanity :laughing:

Before introducing negation, I would rather wait for a good use-case where negation would help clean up the code, given the potential confusion it could cause. What do you think?

tools/satsolver-main.cpp
ekpyron: We use [range-v3](https://ericniebler.github.io/range-v3/) as polyfill until we can transition to C++20 - so you should just be able to link ``satsolver`` against the ``range-v3`` cmake target and include ``<range/v3/view/split.hpp>``. That'll also give you ``ranges::to`` in ``<range/v3/range/conversion.hpp>`` below (that's only staged for C++23, if at all).

libsolutil/CDCL.h
msooseth: Yeah, I think explicit conversion from `true`/`false` would be nicer, too! I just added it, it's a total of 3-line change (1f5445289c835cec534a17b85c2be59a20dea632), and I think it's well worth it. Let me know what you think!

tools/satsolver-main.cpp
ekpyron: But unfortunately, ``part`` will be a range of characters, not a string, so indeed ``std::stol`` won't work...
```
std::from_chars(&*begin(part), &*begin(part) + ranges::distance(part), lit);
```
would work, but I admit that it's not really nice...

msooseth: Oh wow, that's great, fixes it perfectly, thanks! :bowing_man: Fixed in: 3822881f7bcbe38d6178216c30f848b7ecd8b25a

tools/satsolver-main.cpp
msooseth: Ah, now I see! Thanks to both you and @ekpyron , it's now fixed and MUCH cleaner :) Thanks again! Fix in: 3822881f7bcbe38d6178216c30f848b7ecd8b25a

tools/satsolver-main.cpp
ekpyron: ``std::stol(ranges::to<string>(part))`` may be simpler.

tools/satsolver-main.cpp
chriseth: Ah ok I see! The more relaxed version ( https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index-seo.php/SATLINK____DIMACS ) does not require the line to end with a zero, though. Not sure if this is important to us.

tools/satsolver-main.cpp
msooseth: OK, simplified in: cbbf3d572c4d4942381524810c706a664556cf1e Thanks! :)

tools/satsolver-main.cpp
msooseth: In my view, that's not DIMACS, but may be something that a some (set of) authors would like to change it to. The DIMACS specification is from way earlier, this is one from 2002: http://www.satcompetition.org/2004/format-solvers2004.html I have not yet seen a SAT solver that can eat an input without the 0 at the end. The SAT solvers submitted to the SAT competitions can't eat that format AFAIK. So I would try to stick to the DIMACS format from 2002.

So I hope checking for the 0 is OK :) I didn't want to change the DIMACS more than we have to. We will likely have to change it a bit to allow adding the LP part, so I thought it's best to change it as little as possible.

msooseth:
Closing so I can create a nicer one

PR #12723, opened 2022-03-01
[DOCS] Updated translation links / rewording of language influences

franzihei:
- Slightly rephrased the passage about language influences
- Deleted links to very outdated and unmaintained tanslations (I checked all of them and they were both multiple years outdated and not even completely translated, hence of little use for somebody who starts learning Solidity now)
- Added links to actively work-in-progress or complete translations

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,10 +5,8 @@ Solidity is an object-oriented, high-level language for implementing smart
 contracts. Smart contracts are programs which govern the behaviour of accounts
 within the Ethereum state.
 
-Solidity is a `curly-bracket language <https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages>`_.
-It is influenced by C++, Python and JavaScript, and is designed to target the Ethereum Virtual Machine (EVM).
-You can find more details about which languages Solidity has been inspired by in
-the :doc:`language influences <language-influences>` section.
+Solidity is a `curly-bracket language <https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages>`_ designed to target the Ethereum Virtual Machine (EVM).
+It is influenced by C++, Python and JavaScript. You can find more details about which languages Solidity has been inspired by in the :doc:`language influences <language-influences>` section.
 
 Solidity is statically typed, supports inheritance, libraries and complex
 user-defined types among other features.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -5,8 +5,8 @@ Solidity is an object-oriented, high-level language for implementing smart
 contracts. Smart contracts are programs which govern the behaviour of accounts
 within the Ethereum state.
 
-Solidity is a `curly-bracket language <https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages>`_.
-It is influenced by C++, Python and JavaScript, and is designed to target the Ethereum Virtual Machine (EVM).
+Solidity is a `curly-bracket language <https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages>`_ designed to target the Ethereum Virtual Machine (EVM).
+It is most profoundly influenced by C++ but also borrows certain concepts from other languages.
```
ekpyron: What's the trigger for this :-)?
I'm actually not sure emphasizing C++ as primary influence is really all that accurate... beyond being curly-braced and typed there's not all that much in common...

franzihei: Two things: 
1. I wanted to rephrase the sentence so that it makes more sense (with the EVM part in the middle it was a bit mixed up)
2. There was some discussion recently again around it being inspired by JavaScript or not and I just wanted to take the focus away from that, but maybe we could even make it less about other languages here completely and just write "curly bracket language" and then refer to the linked doc for more information. I'm totally fine rephrasing this to whatever makes more sense.

chriseth: I always think that C++ had a heavy influence but Daniel is actually right, it is more the family of curly-bracket languages. Nevertheless, I think it is good to give a list of languages.

franzihei: Okay so should I go back to the original sentence "It is influenced by C++, Python and JavaScript"?

Then the whole section could be as follows:
```Solidity is a `curly-bracket language <https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages>`_ designed to target the Ethereum Virtual Machine (EVM).
It is influenced by C++, Python and JavaScript. You can find more details about which languages Solidity has been inspired by in
the :doc:`language influences <language-influences>` section.```

chriseth: sounds good!

franzihei:
I tried to rebase this but I'm not sure if I did this correctly, please help 😅

chriseth (approved):

PR #12724, opened 2022-03-01
Implement & use markers in existing test framework

### LINKED ISSUE #12601: LSP test framework improvements
After some discussions with @Marenz and now @chriseth, I'm writing down what we agreed on:

In order to avoid messing with file offsets, the idea is to reference cursor locations and source location ranges via Solidity comments.
Using one or more `^` below a line to mark the line above at these columns as a source location.

If it's a single `^`, then that location can be used as "position" (e.g. for client's text cursor), and a sequence of one or more `^` can be used to express a source location range.

Right next to the sequence of  `^`('s) will be the ID, starting with an `@` and followed by an identifier. 

Multiline source locations will have a single `^( @Identifier` marking the first line and a `^) @Identifier` marking the last line.

The new test framework should be similar to `isoltest`, having .sol files for each small test case. Multiple files are split by `// ==== ileA.sol` lines, if needed.

The format for the request(s) and expectation(s) can be similar to the following example:

```solidity
contract C {
    function f(uint256 var1) public pure returns { return 2 * var1; }
                   //  ^^^^ @var1-def                         ^ @use-of-var1
}
// ----
// send: textDocument/definition
// params: { @use-of-var1 }
// reply.result.0: { @var1-def }
```

On failure, an editor could be popped up (like isoltest) to interactively fix the mismatch, or maybe provide an auto-fix iff(!) the actual location is visualized.

Marenz:
refs #12601

This is a draft and only the first step:

This implements a rough version of parsing file tags/markers and using them in the pre-existing test framework.
Main goal was functionality, not readability or performance.

Next steps are to make it more "file based" so that the expectations are also in the files..

### DIFF
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -7,8 +7,10 @@ import os
 import subprocess
 import sys
 import traceback
+import re
 
 from typing import Any, List, Optional, Tuple, Union
+from enum import Enum, auto
 
 import colorama # Enables the use of SGR & CUP terminal VT sequences on Windows.
 from deepdiff import DeepDiff
@@ -17,6 +19,7 @@ from deepdiff import DeepDiff
 class BadHeader(Exception):
     def __init__(self, msg: str):
         super().__init__("Bad header: " + msg)
+
 
 class JsonRpcProcess:
     exe_path: str
@@ -116,12 +119,13 @@ SGR_STATUS_FAIL = '\033[1;31m'
 
 class ExpectationFailed(Exception):
     def __init__(self, actual, expected):
-        self.actual = actual
-        self.expected = expected
-        diff = DeepDiff(actual, expected)
+        self.actual = json.dumps(actual, sort_keys=True)
+        self.expected = json.dumps(expected, sort_keys=True)
+        diff = json.dumps(DeepDiff(actual, expected), indent=4)
         super().__init__(
-            f"Expectation failed.\n\tExpected {expected}\n\tbut got {actual}.\n\t{diff}"
+            f"\n\tExpected {self.expected}\n\tbut got {self.actual}.\n\t{diff}"
         )
+
 
 def create_cli_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Solidity LSP Test suite")
@@ -168,12 +172,25 @@ class Counter:
     passed: int = 0
     failed: int = 0
 
+
+class Marker(Enum):
+    SimpleRange = auto()
+    MultilineRange = auto()
+
+
+# Returns the given marker with the end extended by 'amount'
+def extendEnd(marker, amount=1):
+    marker["end"]["character"] += amount
+    return marker
+
+
 class SolidityLSPTestSuite: # {{{
     test_counter = Counter()
     assertion_counter = Counter()
     print_assertions: bool = False
     trace_io: bool = False
     test_pattern: str
+    marker_regexes: {}
 
     def __init__(self):
         colorama.init()
@@ -184,6 +201,10 @@ class SolidityLSPTestSuite: # {{{
         self.print_assertions = args.print_assertions
         self.trace_io = args.trace_io
         self.test_pattern = args.test_pattern
+        self.marker_regexes = {
+            Marker.SimpleRange: re.compile(R"(?P<range>[\^]+) (?P<tag>@\w+)"),
+            Marker.MultilineRange: re.compile(R"\^(?P<delimiter>[()]) (?P<tag>@\w+)$")
+        }
 
         print(f"{SGR_NOTICE}test pattern: {self.test_pattern}{SGR_RESET}")
 
@@ -206,9 +227,8 @@ class SolidityLSPTestSuite: # {{{
                 with JsonRpcProcess(self.solc_path, ["--lsp"], trace_io=self.trace_io) as solc:
                     test_fn(solc)
                     self.test_counter.passed += 1
-            except ExpectationFailed as e:
+            except ExpectationFailed:
                 self.test_counter.failed += 1
-                print(e)
                 print(traceback.format_exc())
             except Exception as e: # pragma pylint: disable=broad-except
                 self.test_counter.failed += 1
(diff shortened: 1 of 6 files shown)


ekpyron:
Who'll port this to ``isoltest`` as well ;-P?

Marenz:
> Who'll port this to isoltest as well ;-P?

Well, I was given freedom on how exactly to do this. The plan is certainly to have an interface _like_ isoltest. 
Do you think it's worth the time & resources to port it to isoltest?

@christianparpart  maybe you can elaborate a bit why you chose python to do this :)

ekpyron:
> 
> Well, I was given freedom on how exactly to do this. The plan is certainly to have an interface _like_ isoltest. Do you think it's worth the time & resources to port it to isoltest?
> 
> @christianparpart maybe you can elaborate a bit why you chose python to do this :)

Sorry, my comment was probably a bit misleading - I'm not saying this stuff should be ported into isoltest (as a matter of fact I'm not sure if we hadn't been better off to write isoltest as a set of scripts) - I meant that this way of reporting locations is nice and it'd be worth a thought to change the isoltest syntax tests to report source locations just like this (IIRC we may even have an issue about that).
But yeah, for this PR, you can basically disregard my comment.

Marenz:
Should be ready to review & merge (only the first steps! more PRs will come)

Marenz:
Mind you, I am no python expert, so suggestions welcome

wechman:
@Marenz I have two remarks that are not in code lines you changed, but makes the test results with markers harder to read:
- `print(e)` in `except ExpectationFailed as e:` is redundant due to data it provides is printed also in the very next line of the code (`print(traceback.format_exc())`)
- Because python dictionaries are unordered, in case of a test failure, we end up with a confusing massage like this: 
  `Expected {'start': {'line': 7, 'character': 7}, 'end': {'line': 7, 'character': 20}}`
  `but got {'end': {'character': 21, 'line': 7}, 'start': {'character': 8, 'line': 7}}.`
 
  The simplest fix I can imagine is to `json.dump(dict, sort_keys=True)` before `super().__init__` in `class ExpectationFailed(Exception)`.

Other than that I see only minor issues like code formatting that does not adhere to PEP8 rules:
- variable and function names should be lowercase
- two spaces are expected before inline comment
- two blank lines are expected between class definitions

wechman:
@Marenz I have two remarks that are not in code lines you changed, but makes the test results with markers harder to read:
- "print(e)" in `except ExpectationFailed as e:` is redundant due to it prints data which is also printed in the very next line of the code (`print(traceback.format_exc())`)
- Because python dictionaries are unordered, in case of a test failure we end up with confusing massages like this: 
  `Expected {'start': {'line': 7, 'character': 7}, 'end': {'line': 7, 'character': 20}}`
  `but got {'end': {'character': 21, 'line': 7}, 'start': {'character': 8, 'line': 7}}.`
 
  The simplest fix I can imagine is to `json.dump(dict, sort_keys=True)` before `super().__init__` in `class ExpectationFailed(Exception)`.

Other than that I see only minor issues like code formatting that does not adhere to PEP8 rules:
- variable and function names should be lowercase
- two spaces are expected before inline comment
- two blank lines are expected between class definitions

Marenz:
I fixed some of the PEP8 things you mentioned, though probably not all. I want to focus more on my additions in this PR.
Also enhanced the exception reporting as per your suggestion

christianparpart:
sorry for the late response. I didn't find this PR at first. will review that ASAP @Marenz (pls request review next time :) )

test/lsp.py
```diff
@@ -375,10 +399,12 @@ def test_publish_diagnostics_warnings(self, solc: JsonRpcProcess) -> None:
...
 
+        markers = self.getFileMarkers(TEST_NAME)
+
         self.expect_equal(len(diagnostics), 3, "3 diagnostic messages")
-        self.expect_diagnostic(diagnostics[0], code=6321, lineNo=13, startEndColumns=(44, 48))
-        self.expect_diagnostic(diagnostics[1], code=2072, lineNo= 7, startEndColumns=( 8, 19))
-        self.expect_diagnostic(diagnostics[2], code=2072, lineNo=15, startEndColumns=( 8, 20))
+        self.expect_diagnostic(diagnostics[0], code=6321, marker=markers["@unusedReturnVariable"])
+        self.expect_diagnostic(diagnostics[1], code=2072, marker=markers["@unusedVariable"])
+        self.expect_diagnostic(diagnostics[2], code=2072, marker=markers["@unusedContractVariable"])
```
christianparpart, line 455: @Marenz these expect_diagnostic calls should not be needed anymore. Just like isoltest for all the other tests, if one wants to add a test, only such test files should be needed to be touched and not this lsp.py file. :)

christianparpart, line 455: @Marenz these expect_diagnostic calls should not be needed anymore. Just like isoltest for all the other tests, if one wants to add a test, only such test files should be needed to be touched and not this lsp.py file. :)

christianparpart (changes requested):
It's good to not be required  to manually specify line/column offsets for the source locations anymore. But I think it's best to avoid editing the lsp.py for any new test being added. Just like isoltest based tests.

test/lsp.py
Marenz: I mean, yeah, that is the idea. This PR is only the first step as I keep repeating.

christianparpart:
> @christianparpart maybe you can elaborate a bit why you chose python to do this :)

The first implementation was in C++ but I rewrote it in Python because we initially thought it's helpful to use an **existing** LSP client library. So after having switched to Python utilizing an existing LSP client library, we realized that I had to maintain a fork of that library to actually be useful. So We decided to drop that dependency and simply do the JSON-RPC I/O explicitly, which was a good choice.

I'd have favored C++ probably also for this refactor, now, since we know that the transport layer is so thin that it's not really complicated at all. And the the goal of this refactor is also to keep the tests in the .sol files only, so when the isoltest-like tool is being refactored/rewritten into something else/C++, then no tests would be needed to be touched. On the other hand, the motivation is that Python (nor C++) should be needed to be touched just to add a new LSP test.

Marenz:
Rebased.

test/lsp.py
```diff
@@ -480,11 +509,48 @@ def test_textDocument_didOpen_with_relative_import(self, solc: JsonRpcProcess) -
         report = published_diagnostics[1]
         self.expect_equal(report['uri'], self.get_test_file_uri('lib'), "Correct file URI")
         self.expect_equal(len(report['diagnostics']), 1, "one diagnostic")
-        self.expect_diagnostic(report['diagnostics'][0], code=2072, lineNo=31, startEndColumns=(8, 19))
+        marker = self.getFileMarkers("lib")["@diagnostics"]
+        self.expect_diagnostic(report['diagnostics'][0], code=2072, marker=marker)
+
+    def getFileMarkers(self, test_name: str, verbose=False):
```
chriseth: `get_file_markers`? Do we have coding style checks for python?

Marenz: The style is so all over the place I wasn't sure what to follow. I am fine with either, just pick one :D

christianparpart (changes requested):
as discussed in the call. some little more documentation on top of a few functions to clarify their purpose and input/output would be nice.

Other than that, this is stage1-PR, the rest ought to be coming in follow-up PRs.

test/lsp.py
christianparpart: ^|^ that comment was from last week. Thanks Github. :)

christianparpart (approved):
LGTM with above notes.

PR #12726, opened 2022-03-01
Update buildpacks to include `jq`.

aarlt:
The asm json import/export use `jq` to compare the output (see https://github.com/ethereum/solidity/pull/12704).

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -22,7 +22,7 @@
 # (c) 2016-2021 solidity contributors.
 #------------------------------------------------------------------------------
 FROM gcr.io/oss-fuzz-base/base-clang:latest as base
-LABEL version="15"
+LABEL version="16"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
@@ -32,7 +32,7 @@ RUN apt-get update; \
 		software-properties-common \
 		ninja-build git wget \
 		libbz2-dev zlib1g-dev git curl uuid-dev \
-		pkg-config openjdk-8-jdk liblzma-dev unzip mlton m4; \
+		pkg-config openjdk-8-jdk liblzma-dev unzip mlton m4 jq; \
     apt-get install -qy python3-pip;
 
 # Install cmake 3.21.2 (minimum requirement is cmake 3.10)
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
 #------------------------------------------------------------------------------
 FROM buildpack-deps:focal AS base
-LABEL version="10"
+LABEL version="11"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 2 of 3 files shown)


scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
```diff
@@ -32,14 +32,14 @@ RUN apt-get update; \
...
-# Install cmake 3.21.2 (minimum requirement is cmake 3.10)
-RUN wget https://github.com/Kitware/CMake/releases/download/v3.21.2/cmake-3.21.2-Linux-x86_64.sh; \
-    test "$(sha256sum cmake-3.21.2-Linux-x86_64.sh)" = "3310362c6fe4d4b2dc00823835f3d4a7171bbd73deb7d059738494761f1c908c  cmake-3.21.2-Linux-x86_64.sh"; \
-    chmod +x cmake-3.21.2-Linux-x86_64.sh; \
-    ./cmake-3.21.2-Linux-x86_64.sh --skip-license --prefix="/usr"
+# Install cmake 3.22.2 (minimum requirement is cmake 3.10)
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2-linux-x86_64.sh; \
+    test "$(sha256sum cmake-3.22.2-linux-x86_64.sh)" = "faaa0465b72fe467fa193954c9bb67141048a2c7acc3236975ac20fdf241582e  cmake-3.22.2-linux-x86_64.sh"; \
+    chmod +x cmake-3.22.2-linux-x86_64.sh; \
+    ./cmake-3.22.2-linux-x86_64.sh --skip-license --prefix="/usr"
```
aarlt: For `ubuntu1604.clang.ossfuzz` there is already a cmake installed in `/usr/local/bin/cmake` v3.21.1, where `PATH` is set to `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/out` - so the "old" version has higher priorities. Maybe we could just change the `prefix` to install the new version in `/usr/local`.

aarlt: I will now change cmake installation `prefix` to `/usr/local`.

scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
```diff
@@ -37,12 +37,18 @@ RUN set -ex; \
...
 		; \
 	apt-get install -qy python3-pip python3-sphinx; \
 	pip3 install codecov; \
 	rm -rf /var/lib/apt/lists/*
 
+# Install cmake 3.22.2 (minimum requirement is cmake 3.10)
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2-linux-x86_64.sh; \
+    test "$(sha256sum cmake-3.22.2-linux-x86_64.sh)" = "faaa0465b72fe467fa193954c9bb67141048a2c7acc3236975ac20fdf241582e  cmake-3.22.2-linux-x86_64.sh"; \
+    chmod +x cmake-3.22.2-linux-x86_64.sh; \
+    ./cmake-3.22.2-linux-x86_64.sh --skip-license --prefix="/usr/local"
```
ekpyron: This doesn't seem like a good idea - we shouldn't be testing against a customly installed version of ``cmake``.
If we want to test builds against newer versions of ``cmake`` than what is packaged in the base images (resp. available for ubuntu LTS), we can use archlinux build runs instead.

aarlt: Ah ok! Basically we don't want to change the build tools contained in the default ubuntu distribution. Yeah that makes sense. I will checkout the archlinux build.

ekpyron (approved):

PR #12728, opened 2022-03-02
Save and restore for loop nesting level in UnusedStoreBase.

ekpyron:
Came up in https://github.com/ethereum/solidity/pull/12672#discussion_r817566550
It doesn't really matter in practice, but it's weird not to do it.


ekpyron:
Closing in favor of https://github.com/ethereum/solidity/pull/12730

PR #12729, opened 2022-03-02
Split out disassemble to remove numeric from instruction includes.

chriseth:
Needed for https://github.com/ethereum/solidity/pull/12672


ekpyron (dismissed):
Haha, ok - +135/-67 diff here puts the translation functions and duplicate types on the other end into another perspective :-).
Still nicer this way!

ekpyron:
test/libsolidity/SolidityOptimizer.cpp seems to be missing the additional include

ekpyron (approved):

PR #12730, opened 2022-03-02
Also save and restore the nesting depth.

chriseth:
Mentioned in https://github.com/ethereum/solidity/pull/12672

This is just a performance parameter, so it was not a semantic bug.


ekpyron (approved):
Now we have two of those :-D.

chriseth:
Sorry, didn't see that! Interesting how we both chose `0` instead of `{}` :)

PR #12734, opened 2022-03-02
Adds TriState, min-2-long clause invariant, and SAT solver binary

msooseth:
This is a squashed, minimized change from the previous PR so it's easier to merge and causes less issues in reviewing.

### DIFF
--- a/libsolutil/CDCL.cpp
+++ b/libsolutil/CDCL.cpp
@@ -1,0 +1,322 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	solidity is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with solidity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+// SPDX-License-Identifier: GPL-3.0
+
+#include <libsolutil/CDCL.h>
+
+#include <liblangutil/Exceptions.h>
+
+// TODO remove before merge
+#include <iostream>
+#include <libsolutil/StringUtils.h>
+
+using namespace std;
+using namespace solidity;
+using namespace solidity::util;
+
+
+CDCL::CDCL(
+	vector<string> _variables,
+	vector<Clause> const& _clauses,
+	std::function<std::optional<Clause>(std::map<size_t, bool> const&)> _theorySolver
+):
+	m_theorySolver(_theorySolver),
+	m_variables(move(_variables))
+{
+	for (Clause const& clause: _clauses)
+		addClause(clause);
+
+	// TODO some sanity checks like no empty clauses, no duplicate literals?
+}
+
+optional<CDCL::Model> CDCL::solve()
+{
+	cout << "====" << endl;
+	for (unique_ptr<Clause> const& c: m_clauses)
+		cout << toString(*c) << endl;
+	cout << "====" << endl;
+	while (true)
+	{
+		optional<Clause> conflictClause = propagate();
+		if (!conflictClause && m_theorySolver)
+		{
+			conflictClause = m_theorySolver(m_assignments);
+			if (conflictClause)
+				cout << "Theory gave us conflict: " << toString(*conflictClause) << endl;
+		}
+		if (conflictClause)
+		{
+			if (currentDecisionLevel() == 0)
+			{
+				cout << "Unsatisfiable" << endl;
+				return nullopt;
+			}
+			auto&& [learntClause, backtrackLevel] = analyze(move(*conflictClause));
+			cancelUntil(backtrackLevel);
+			solAssert(!learntClause.empty());
+			solAssert(!isAssigned(learntClause.front()));
+			for (size_t i = 1; i < learntClause.size(); i++)
+				solAssert(isAssignedFalse(learntClause.at(i)));
+
+			addClause(move(learntClause));
+			enqueue(m_clauses.back()->front(), &(*m_clauses.back()));
+		}
+		else
+		{
+			if (auto variable = nextDecisionVariable())
+			{
+				m_decisionPoints.emplace_back(m_assignmentTrail.size());
+				cout << "Deciding on " << m_variables.at(*variable) << " @" << currentDecisionLevel() << endl;
+				enqueue(Literal{false, *variable}, nullptr);
+			}
+			else
+			{
+				cout << "satisfiable." << endl;
+				for (auto&& [var, value]: m_assignments)
+					cout << " " << m_variables.at(var) << ": " << (value ? "true" : "false") << endl;
+				return m_assignments;
+			}
+		}
+	}
+}
+
+void CDCL::setupWatches(Clause& _clause)
+{
+	for (size_t i = 0; i < min<size_t>(2, _clause.size()); i++)
+		m_watches[_clause.at(i)].push_back(&_clause);
+}
+
+optional<Clause> CDCL::propagate()
+{
+	cout << "Propagating." << endl;
+	for (; m_assignmentQueuePointer < m_assignmentTrail.size(); m_assignmentQueuePointer++)
+	{
+		Literal toPropagate = m_assignmentTrail.at(m_assignmentQueuePointer);
+		Literal falseLiteral = ~toPropagate;
+		cout << "Propagating " << toString(toPropagate) << endl;
+		// Go through all watched clauses where this assignment makes the literal false.
+		vector<Clause*> watchReplacement;
+		auto it = m_watches[falseLiteral].begin();
+		auto end = m_watches[falseLiteral].end();
+		for (; it != end; ++it)
+		{
+			Clause& clause = **it;
+			cout << " watch clause: " << toString(clause) << endl;
+
+			solAssert(!clause.empty());
+			if (clause.front() != falseLiteral)
+				swap(clause[0], clause[1]);
+			solAssert(clause.front() == falseLiteral);
+			if (clause.size() >= 2 && isAssignedTrue(clause[1]))
+			{
+				// Clause is already satisfied, keezp the watch.
+				cout << " -> already satisfied by " << toString(clause[1]) << endl;
+				watchReplacement.emplace_back(&clause);
+				continue;
+			}
+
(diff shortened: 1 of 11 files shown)


libsolutil/CDCL.h
```diff
@@ -113,7 +174,22 @@ class CDCL
 	std::vector<size_t> m_decisionPoints;
 	/// Index into assignmentTrail: All assignments starting there have not yet been propagated.
 	size_t m_assignmentQueuePointer = 0;
+
+	// Current state of the solver. If FALSE, we are in an UNSAT state.
+	bool ok = true;
```
chriseth:
```suggestion
	bool m_ok = true;
```
(although I also find "ok" not descriptive enough)

libsolutil/CDCL.h
```diff
@@ -68,25 +132,22 @@ class CDCL
 	std::pair<Clause, size_t> analyze(Clause _conflictClause);
 	size_t currentDecisionLevel() const { return m_decisionPoints.size(); }
 
-	void addClause(Clause _clause);
+	void addClause(const Clause& _lits);
```
chriseth:
```suggestion
	void addClause(Clause const& _lits);
```
the type always comes first

libsolutil/CDCL.h
```diff
@@ -48,16 +49,79 @@ struct Literal
...
 class CDCL
 {
 public:
-	using Model = std::map<size_t, bool>;
+	using Model = std::vector<TriState>;
 	CDCL(
 		std::vector<std::string> _variables,
 		std::vector<Clause> const& _clauses,
-		std::function<std::optional<Clause>(std::map<size_t, bool> const&)> _theoryPropagator = {}
+		std::function<std::optional<Clause>(std::vector<TriState> const&)> _theoryPropagator = {}
```
chriseth: Oh is this still called Propagator here? Should be Solver

libsolutil/CDCL.h
```diff
@@ -48,16 +49,79 @@ struct Literal
 		return std::make_tuple(positive, variable) < std::make_tuple(_other.positive, _other.variable);
 	}
 };
+
 using Clause = std::vector<Literal>;
 
+class TriState {
+public:
+	explicit constexpr TriState(const bool b) : val(b ? 1 : 0) {}
```
chriseth:
```suggestion
	explicit constexpr TriState(bool const b): val(b ? 1 : 0) {}
```

libsolutil/CDCL.h
```diff
@@ -48,16 +49,79 @@ struct Literal
...
+		}
+	}
+
+	constexpr static TriState t_true();
+	constexpr static TriState t_false();
+	constexpr static TriState t_unset();
+
+private:
+	// Default value is UNSET
+	uint8_t val = 2;
```
chriseth:
```suggestion
	uint8_t m_value = 2;
```

libsolutil/CDCL.h
```diff
@@ -48,16 +49,79 @@ struct Literal
...
+		else if (val == 0)
+			return "false";
+		else
+		{
+			assert(val == 2);
+			return "unset";
+		}
+	}
+
+	constexpr static TriState t_true();
```
chriseth: We usually don't have `t_` prefixes, but simple `true` / `false` would not work. Let me see...

libsolutil/CDCL.cpp
```diff
@@ -121,7 +136,7 @@ optional<Clause> CDCL::propagate()
 			if (clause.front() != falseLiteral)
 				swap(clause[0], clause[1]);
 			solAssert(clause.front() == falseLiteral);
-			if (clause.size() >= 2 && isAssignedTrue(clause[1]))
+			if (value(clause[1]) == TriState::t_true())
```
chriseth:
```suggestion
			if (value(clause[1]) == TriState{true})
```
(shorter and does not need the constexpr function)

libsolutil/CDCL.cpp
```diff
@@ -230,9 +245,40 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
...
+
+	Clause clause{_lits};
+	Clause clause_updated;
+	for (const auto& l: clause)
+	{
+		// Clause is satisfied, nothing to do.
+		if (value(l) == TriState::t_true())
+			return;
+
+		// Remove literal from clause.
```
chriseth, line 259: Why can I do that? Is this only for learnt clauses? If yes, we should rename the function

libsolutil/CDCL.cpp
```diff
@@ -230,9 +245,40 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
 	return {move(learntClause), backtrackLevel};
 }
 
-void CDCL::addClause(Clause _clause)
+void CDCL::addClause(const Clause& _lits)
```
chriseth:
```suggestion
void CDCL::addClause(Clause _clause)
```

libsolutil/CDCL.cpp
```diff
@@ -230,9 +245,40 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
...
 {
-	m_clauses.push_back(make_unique<Clause>(move(_clause)));
+	if (!ok) return;
+
+	Clause clause{_lits};
```
chriseth: This is not needed if the parameter is passed by value

libsolutil/CDCL.cpp
```diff
@@ -230,9 +245,40 @@ std::pair<Clause, size_t> CDCL::analyze(Clause _conflictClause)
...
+	Clause clause_updated;
```
chriseth:
```suggestion
	Clause updatedClause;
```

libsolutil/CDCL.cpp
```diff
@@ -45,6 +46,8 @@ CDCL::CDCL(
 
 optional<CDCL::Model> CDCL::solve()
 {
+	if (!ok) return nullopt;
+
```
chriseth, line 50: Doesn't this need some preprocessing here to enforce the 1-/2-long clause invariant?

libsolutil/CDCL.h
msooseth: Good pointer, actually/ Fixing to:

```
enum class State {unknown, unsat};
State m_state = State::unknown;
```

Which I think captures it much better?

libsolutil/CDCL.h
msooseth: Oops, good find, sorry!

libsolutil/CDCL.h
msooseth: :+1:

libsolutil/CDCL.h
msooseth: Ah, now I understand the rule! OK, got it! Quite a good rule actually, I should use it in my own code.

libsolutil/CDCL.h
msooseth: Fixed!

libsolutil/CDCL.cpp
msooseth: We can do this because if there is a literal that's already satisfied according to the current assignment, then the clause is already satisfied, so it has no effect.

This can occur if we have two clauses that come in, in sequence:

```
1 0
1 2 0
```

In this case, the 2nd clause can be removed, because variable `1` is already set to `TriState{true}` so the clause has no effect, it's already satisfied by literal `1`. Conversely, in the case of:

```
1 0
-1 2 3 0
```

The 2nd clause will be simplified to `2 3 0`, since `1` is already `TriState{true}`, and so the literal `-1` evaluates to `TriState{false}` and has no effect.

libsolutil/CDCL.cpp
msooseth: Fixed, thanks!

libsolutil/CDCL.cpp
msooseth: Thanks, I see, now I understand the style better :)

libsolutil/CDCL.cpp
msooseth: Yes, I believe it's now fixed, thanks for pointing that out!

libsolutil/CDCL.h
msooseth: Yep, I was unsure what to do here. The `t_` can be misleading in retrospect, since it's not a type, and sometimes `t_` is used to designate types.

libsolutil/CDCL.cpp
msooseth: Yep, I fixed it. Actually, I can remove the constexpr in retrospect. This is cleaner.

msooseth: Thanks for sticking to this. You were right, it's cleaner this way :)

libsolutil/CDCL.cpp
chriseth: But again, is this only for learnt clauses?

libsolutil/CDCL.h
chriseth: Sounds good!

libsolutil/CDCL.cpp
msoos: Ah sorry. With the new system, `addClause` is not called for learnt clauses at all. We deal with 1- and 2- long learnt clauses post-analysis, and the larger ones we attach. So the invariant holds for all clauses, but this code is not called for learnt clauses.

libsolutil/CDCL.cpp
msoos: I enforce it at `addClause` and after analysis. They are the only places, AFAIK, that can add clauses to the DB. Maybe I missed a place? :S

libsolutil/CDCL.h
msooseth: Thanks :)

libsolutil/CDCL.h
msooseth: Actually, now that I removed the constexpr for `t_true` and `t_false`, we don't need these at all. The unset can simply be `unset`, as that doesn't clash. This `TriState{true}` and `TriState{false}` really did make things clearer in the end.

msooseth: OK, resolving now, because this `t_` business is not needed at all now :)

libsolutil/CDCL.cpp
msooseth: I think this can now be resolved, since this is not called with learnt clauses, so it should be fine.

libsolutil/CDCL.cpp
msooseth: As we discussed out-of-band, thankfully this needs no check here, due to both addClause enforcing the invariant, and the learnt clause adding enforcing the invariant, too. I am resolving. Please re-open in case there are some concerns.

cameel:
> Those are being rewritten in Rust anyway and are not part of the roadmap, so my guess would be close

Closing.

PR #12736, opened 2022-03-03
Re-enable running tests via IR in external tests in cases where they don't pass due to Hardhat heuristics

cameel:
~Depends on #12195.~ Merged.

So far only compilation was enabled in IR presets for OpenZeppelin, PRBMath, ElementFi and (after #12197) Gnosis. This is due to https://github.com/nomiclabs/hardhat/issues/2115, which is still unresolved. Unfortunately this does not give us any benchmarks via IR for these projects. In this PR I'm re-enabling tests and instead disabling individual test cases that do not pass.

~The PR also disables one particularly flaky test file in ElementFi. It has one case that fails often and has been reported upstream (https://github.com/element-fi/elf-contracts/issues/240) but also many tests in the file seem to depend on an external service, which has resulted in a mass failure due to network timeout at least once when the service was down.~ **EDIT**: This part moved to #12764.

### DIFF
--- a/test/externalTests/elementfi.sh
+++ b/test/externalTests/elementfi.sh
@@ -42,13 +42,12 @@ function elementfi_test
     local config_file="hardhat.config.ts"
     local config_var=config
 
-    local compile_only_presets=(
-        ir-optimize-evm+yul       # Compiles but tests fail. See https://github.com/nomiclabs/hardhat/issues/2115
-    )
+    local compile_only_presets=()
     local settings_presets=(
         "${compile_only_presets[@]}"
         #ir-no-optimize           # Compilation fails with "YulException: Variable var_amount_9311 is 10 slot(s) too deep inside the stack."
         #ir-optimize-evm-only     # Compilation fails with "YulException: Variable var_amount_9311 is 10 slot(s) too deep inside the stack."
+        ir-optimize-evm+yul
         legacy-no-optimize
         legacy-optimize-evm-only
         legacy-optimize-evm+yul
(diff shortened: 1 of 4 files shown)


cameel:
There are 15 failures for ElementFi via IR unrelated to https://github.com/NomicFoundation/hardhat/issues/2115 (see [job 994880](https://app.circleci.com/pipelines/github/ethereum/solidity/22668/workflows/5b0f2e5b-cc89-406f-bf91-c2a954c9219c/jobs/994880)):
1) 2 cases of address not matching expectations.
    - It's a compiler issue: #12738
2) 5 cases of a revert with wrong reason string.
    - It's a Hardhat issue: https://github.com/NomicFoundation/hardhat/issues/2453
3) 8 cases of unexpected reverts.
    - Could be related to (2). Still investigating this one.

cameel:
Turns out that all 15 failures were caused by #12738. The reverts were caused by external calls to wrongly calculated address and that's why they did not have a revert message. The issue I found in Hardhat is still real but it's a different thing that I just happened to find by accident.

cameel:
The failure in OZ test seems to be just a case of https://github.com/NomicFoundation/hardhat/issues/2453 triggered under more specific conditions (see https://github.com/NomicFoundation/hardhat/issues/2453#issuecomment-1059467546).

@ekpyron's #12744 fixes the failures on ElementFi so I won't be adding workarounds for them in this PR. It does not fix the OZ one though so I'm disabling this test case in OZ. With that, all tests should start passing here once #12744 is merged.

ekpyron:
Looks like this has a merge conflict now - but apart from that there's nothing blocking this anymore, is there?

cameel:
Yeah, a small conflict with #12764. They both add some new lines at the same spot.

Conflict resolved.

cameel:
> but apart from that there's nothing blocking this anymore, is there?

It's ready unless it turns out that some tests are still failing.

ekpyron (approved):
The failing euler test is unrelated and hopefully fixed by https://github.com/ethereum/solidity/pull/12768 (which I just merged).

ekpyron:
I'll just go ahead and un-draft and merge this.

cameel:
Ah, right. Forgot to undraft it. Thanks!

And yeah, euler test failure was unrelated, should be fixed by #12768.

PR #12740, opened 2022-03-04
Replace all locale-dependent operations with locale-agnostic counterparts

### LINKED ISSUE #12720: Use invalid locale instead of "C" to detect and fix any attempted locale-dependent operations in the codebase
Requested in https://github.com/ethereum/solidity/pull/12716#issuecomment-1054141878.
Related to #12350.

> I would actually be happier if we could find a way to somehow set an "illegal" locale so that we get an error whenever we perform a locale-dependent operation (and of course remove all locale-dependent operations).

wechman:
fixes #12720

### DIFF
--- a/libsolutil/StringUtils.h
+++ b/libsolutil/StringUtils.h
@@ -27,7 +27,9 @@
 #include <libsolutil/CommonData.h>
 #include <libsolutil/Numeric.h>
 
+#include <algorithm>
 #include <limits>
+#include <locale>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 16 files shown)


chriseth (commented):
I think it would be nice to have functions `toUpper(char)` and `toUpper(string)` (and lower) in our string utils header or something - same for isDigit.

libyul/backends/evm/EVMDialect.cpp
```diff
@@ -142,8 +142,8 @@ map<YulString, BuiltinFunctionForEVM> createBuiltins(langutil::EVMVersion _evmVe
 	map<YulString, BuiltinFunctionForEVM> builtins;
 	for (auto const& instr: evmasm::c_instructions)
 	{
-		string name = instr.first;
-		transform(name.begin(), name.end(), name.begin(), [](unsigned char _c) { return tolower(_c); });
+		string name = instr.first;
```
stackenbotten: Coding style error

wechman:
[@chriseth](https://github.com/chriseth) I created string util functions as you suggested. However, [CommonSyntaxTest.cpp](https://github.com/ethereum/solidity/blob/develop/test/CommonSyntaxTest.cpp) still needs small cleanup around `isDigit` function. It is related to `using namespace` directive abuse I have found. The problem is described in #12743.

libsolutil/StringUtils.h
```diff
@@ -197,4 +199,45 @@ inline std::optional<unsigned> toUnsignedInt(std::string const& _value)
...
+/// @param _c value to be converted
+/// @return the converted value
+inline char toUpper(char _c)
+{
+	return toupper(_c, std::locale::classic());
+}
+
+/// Converts parameter _s to its lowercase equivalent. It uses the classic "C" locale semantics.
+/// @param _s value to be converted
+inline void toLower(std::string& _s)
```
chriseth: Can you change this so that it returns a copy of the string? This way it might be rather error prone.

wechman: Sure! Fixup commit already added to the PR.

libsolutil/StringUtils.h
```diff
@@ -197,4 +199,47 @@ inline std::optional<unsigned> toUnsignedInt(std::string const& _value)
...
+}
+
+/// Converts parameter _s to its lowercase equivalent. It uses the classic "C" locale semantics.
+/// @param _s value to be converted
+/// @return the converted value
+inline std::string toLower(std::string _s)
+{
+	std::transform(_s.begin(), _s.end(), _s.begin(), [](char _c) {
+		return toLower(_c);
+	});
```
ekpyron, line 225: Doesn't this work?
```suggestion
	std::transform(_s.begin(), _s.end(), _s.begin(), toLower);
```

ekpyron: Ah, it doesn't because it's overloaded... yeah, then the current way is probably nicest. So sorry, nevermind :-).

chriseth:
Can you squash this please?

chriseth (approved):

PR #12741, opened 2022-03-04
[Doc] Fix leftover use of divModWithSlacks

david-k:
The option `divModWithSlacks` was previously changed to `divModNoSlacks`.
However, this was not reflected in the documentation.


leonardoalt (approved):
Thanks!

PR #12744, opened 2022-03-04
Store whether a subassembly is creation code and optimize accordingly.

### LINKED ISSUE #11118: Understanding creation context when optimizing Yul subobjects
Currently, we assume that all Yul subobjects are in runtime context: https://github.com/ethereum/solidity/blob/develop/libyul/AssemblyStack.cpp#L177 (the second parameter `false`.)

However, this is not true in general.

Take for example, 

```
  contract C {
	  constructor () {
		  assembly {
			  sstore(0, shl(180, 1))
		  }
		  new D();
	  }
  }
  contract D {
	  constructor() {
		  assembly {
			  sstore(1, shl(180, 1))
		  }
	  }
	  fallback() external {
		assembly {
			sstore(2, shl(180, 1))
		}
	  }
  }
```

From the context of `C`, both `D`'s constructor and `fallback` will have `shl(180, 1)` expanded. In this case, this would lead to more expensive construction code.

-----

One solution is that the context of each sub-object should be stored in the C++ data structure. But this runs into another problem for the following code:

```
  contract C {
	  D d;
	  constructor () {
		  d = new D();
		  d.f();
	  }
	  fallback() external {
		  d.f();
	  }
  }
  contract D {
	  constructor() {
		  assembly {
			  sstore(1, shl(180, 1))
		  }
	  }
	  function f() public {
		  assembly {
			  sstore(2, shl(180, 1))
		  }
	  }
  }
```

In this example, optimizing the sub-object `D.f` with creation context would mean that the calling C's constructor is cheaper, but making the call to fallback more expensive at runtime.  Somehow we have to understand which among the two options would be cheaper overall and optimize according to that. We might require a gas estimator for Yul to properly do this.

### LINKED ISSUE #12738: `creationCode` via IR does not match the compiler's `--bin` output when the contract has a constructor with `bytes` arguments
## Description
In contracts compiled with the `--experimental-via-ir` option, the bytecode returned from `type(C).creationCode` does not match the one the compiler returns from `--bin` when there's a constructor accepting a `bytes` argument.

```solidity
contract C {
    constructor(bytes memory x) {}
}
```
It does not happen if the constructor has a `unit` parameter or no parameters at all.

I discovered this when I re-enabled tests in the `ir-optimize-evm+yul` preset of ElementFi external test (#12736). Note that in that PR I had to disable some tests due to known issues with Hardhat's heuristics not working with the new codegen. The remaning tests pass with the legacy codegen but 15 of them fail via IR. The bug is the cause of some of these failures.

It's the easiest to reproduce with a framework like Hardhat because this way it's possible to compare compiler output directly. We'll need #11424 to be able to compare the bytecode in isoltest (and #12737 could also be an issue).

## Environment

- Compiler version: 0.8.12 and `develop`
- Framework/IDE: Hardhat 2.9.0

The issue is reproducible as far back as 0.8.2. On 0.8.1 Hardhat crashes. On 0.8.0, 0.7.6 and 0.7.5 the compiler crashes. On 0.7.4 and earlier `viaIR` option was not available.

## Steps to Reproduce

Run this in the shell:
```bash
mkdir test/ contracts/

cat <<EOF > hardhat.config.js
require("@nomiclabs/hardhat-waffle");

module.exports = {
    solidity: {
        compilers: [{
            version: "0.8.12",
            settings: {
                viaIR: true,
                optimizer: { enabled: true },
                metadata: { "bytecodeHash": "none" },
            },
        }]
    }
};
EOF

cat <<EOF > test/CreationCodeTest.js
const { expect } = require("chai");
const { ethers } = require("hardhat");

const targetJSON = require("../artifacts/contracts/CreationCodeTest.sol/Target.json");

it("Bytecode creationCode should match the bytecode in the artifact", async () => {
    const Factory = await ethers.getContractFactory("Factory");
    const factory = await Factory.deploy();

    expect(await factory.targetCreationCode()).to.equal(targetJSON.bytecode);
});
EOF

cat <<EOF > contracts/CreationCodeTest.sol
contract Target {
    constructor(bytes memory x) {}
}

contract Factory {
    bytes public targetCreationCode = type(Target).creationCode;
}
EOF

npm install hardhat @nomiclabs/hardhat-waffle chai
npx hardhat test
```

```
Compiled 1 Solidity file successfully


  1) Bytecode creationCode should match the bytecode in the artifact

  0 passing (635ms)
  1 failing

  1) Bytecode creationCode should match the bytecode in the artifact:

      AssertionError: expected '0x6080604052346100e95761015d803803806…' to equal '0x6080604052346100e85761015b803803806…'
      + expected - actual

      -0x6080604052346100e95761015d8038038061001981610105565b92833981019060209081818403126100e157805167ffffffffffffffff918282116100d7570183601f820112156100c15780519182116100ca575b610066601f8301601f19168401610105565b938285528383830101116100c15781906000905b848383106100a95750501161009b575b505050604051602490816101398239f35b60009201015238808061008a565b8193508281939201015182828801015201829161007a565b50505050600080fd5b6100d26100ee565b610054565b5050505050600080fd5b505050600080fd5b600080fd5b50634e487b7160e01b600052604160045260246000fd5b6040519190601f01601f1916820167ffffffffffffffff81118382101761012b57604052565b6101336100ee565b60405256fe6080604052600080fdfea26c6578706572696d656e74616cf564736f6c634300080c0018
      +0x6080604052346100e85761015b8038038061001981610104565b92833981019060209081818403126100e05780516001600160401b03918282116100d6570183601f820112156100c05780519182116100c9575b610065601f8301601f19168401610104565b938285528383830101116100c05781906000905b848383106100a85750501161009a575b505050604051602490816101378239f35b600092010152388080610089565b81935082819392010151828288010152018291610079565b50505050600080fd5b6100d16100ed565b610053565b5050505050600080fd5b505050600080fd5b600080fd5b50634e487b7160e01b600052604160045260246000fd5b6040519190601f01601f191682016001600160401b0381118382101761012957604052565b6101316100ed565b60405256fe6080604052600080fdfea26c6578706572696d656e74616cf564736f6c634300080c0018

      at Context.<anonymous> (test/CreationCodeTest.js:10:51)
```

The expected value above comes from the `bytecode` field of an artifact produced by Hardhat. This value comes directly from compiler's Standard JSON output, which can be verified by compiling the contract without Hardhat:
```
solc contracts/CreationCodeTest.sol --experimental-via-ir --optimize --metadata-hash none --bin
```
```
======= contracts/CreationCodeTest.sol:Target =======
Binary:
6080604052346100e85761015b8038038061001981610104565b92833981019060209081818403126100e05780516001600160401b03918282116100d6570183601f820112156100c05780519182116100c9575b610065601f8301601f19168401610104565b938285528383830101116100c05781906000905b848383106100a85750501161009a575b505050604051602490816101378239f35b600092010152388080610089565b81935082819392010151828288010152018291610079565b50505050600080fd5b6100d16100ed565b610053565b5050505050600080fd5b505050600080fd5b600080fd5b50634e487b7160e01b600052604160045260246000fd5b6040519190601f01601f191682016001600160401b0381118382101761012957604052565b6101316100ed565b60405256fe6080604052600080fdfea26c6578706572696d656e74616cf564736f6c634300080c0018
```

ekpyron:
Attempt to fix https://github.com/ethereum/solidity/issues/12738
Should also fix https://github.com/ethereum/solidity/issues/11118

Not ready yet:
- [ ] needs Changelog entry
- [ ] needs unit tests
- [ ] needs a decision on whether we want a better way to determine whether a yul object is creation code than checking for the ``_deployed`` suffix.
- [ ] needs to check if the change in legacy code gen in ``libsolidity/codegen/CompilerUtils.cpp`` is needed - and should either get a test for it or have the change reverted (if it affects legacy codegen, we also have to consider if it's an "important bug")

### DIFF
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -203,7 +203,7 @@ void Assembly::assemblyStream(
 		for (size_t i = 0; i < m_subs.size(); ++i)
 		{
 			_out << endl << _prefix << "sub_" << i << ": assembly {\n";
-			m_subs[i]->assemblyStream(_out, _debugInfoSelection, _prefix + "    ", _sourceCodes);
+			m_subs[i].first->assemblyStream(_out, _debugInfoSelection, _prefix + "    ", _sourceCodes);
 			_out << _prefix << "}" << endl;
 		}
 	}
@@ -352,7 +352,7 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) 
 		{
 			std::stringstream hexStr;
 			hexStr << hex << i;
-			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices);
+			data[hexStr.str()] = m_subs[i].first->assemblyJSON(_sourceIndices);
 		}
 	}
 
@@ -435,9 +435,8 @@ map<u256, u256> const& Assembly::optimiseInternal(
 	for (size_t subId = 0; subId < m_subs.size(); ++subId)
 	{
 		OptimiserSettings settings = _settings;
-		// Disable creation mode for sub-assemblies.
-		settings.isCreation = false;
-		map<u256, u256> const& subTagReplacements = m_subs[subId]->optimiseInternal(
+		settings.isCreation = m_subs[subId].second;
+		map<u256, u256> const& subTagReplacements = m_subs[subId].first->optimiseInternal(
 			settings,
 			JumpdestRemover::referencedTags(m_items, subId)
 		);
@@ -582,7 +581,7 @@ LinkerObject const& Assembly::assemble() const
 	map<u256, pair<string, vector<size_t>>> immutableReferencesBySub;
 	for (auto const& sub: m_subs)
 	{
-		auto const& linkerObject = sub->assemble();
+		auto const& linkerObject = sub.first->assemble();
 		if (!linkerObject.immutableReferences.empty())
 		{
 			assertThrow(
@@ -592,7 +591,7 @@ LinkerObject const& Assembly::assemble() const
 			);
 			immutableReferencesBySub = linkerObject.immutableReferences;
 		}
-		for (size_t tagPos: sub->m_tagPositionsInBytecode)
+		for (size_t tagPos: sub.first->m_tagPositionsInBytecode)
 			if (tagPos != numeric_limits<size_t>::max() && tagPos > subTagSize)
 				subTagSize = tagPos;
 	}
@@ -626,7 +625,7 @@ LinkerObject const& Assembly::assemble() const
 
 	unsigned bytesRequiredIncludingData = bytesRequiredForCode + 1 + static_cast<unsigned>(m_auxiliaryData.size());
 	for (auto const& sub: m_subs)
-		bytesRequiredIncludingData += static_cast<unsigned>(sub->assemble().bytecode.size());
+		bytesRequiredIncludingData += static_cast<unsigned>(sub.first->assemble().bytecode.size());
 
 	unsigned bytesPerDataRef = numberEncodingSize(bytesRequiredIncludingData);
 	uint8_t dataRefPush = static_cast<uint8_t>(pushInstruction(bytesPerDataRef));
@@ -780,7 +779,7 @@ LinkerObject const& Assembly::assemble() const
 		vector<size_t> const& tagPositions =
 			subId == numeric_limits<size_t>::max() ?
 			m_tagPositionsInBytecode :
-			m_subs[subId]->m_tagPositionsInBytecode;
+			m_subs[subId].first->m_tagPositionsInBytecode;
 		assertThrow(tagId < tagPositions.size(), AssemblyException, "Reference to non-existing tag.");
 		size_t pos = tagPositions[tagId];
 		assertThrow(pos != numeric_limits<size_t>::max(), AssemblyException, "Reference to tag without position.");
@@ -870,7 +869,7 @@ Assembly const* Assembly::subAssemblyById(size_t _subId) const
 	Assembly const* currentAssembly = this;
 	for (size_t currentSubId: subIds)
 	{
-		currentAssembly = currentAssembly->m_subs.at(currentSubId).get();
+		currentAssembly = currentAssembly->m_subs.at(currentSubId).first.get();
 		assertThrow(currentAssembly, AssemblyException, "");
 	}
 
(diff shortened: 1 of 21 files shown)


libevmasm/Assembly.cpp
```diff
@@ -435,9 +435,8 @@ map<u256, u256> const& Assembly::optimiseInternal(
 	for (size_t subId = 0; subId < m_subs.size(); ++subId)
 	{
 		OptimiserSettings settings = _settings;
-		// Disable creation mode for sub-assemblies.
-		settings.isCreation = false;
-		map<u256, u256> const& subTagReplacements = m_subs[subId]->optimiseInternal(
+		settings.isCreation = m_subs[subId].second;
```
ekpyron, line 438: If I change this back to hard-coded ``false;``, then I still get ``ffffffffffffffff42`` in the https://github.com/ethereum/solidity/issues/12738#issuecomment-1059158952 setting - it only vanishes with both this and the yul optimizer changes...

But I haven't found a case in which something like this would happen in legacy codegen... aren't created contracts also sub-assemblies for that and should get the wrong value here?

ekpyron: In other words: from the looks of this, this should also affect legacy codegen, but I can't reproduce it for that.

libsolidity/codegen/CompilerUtils.cpp
```diff
@@ -1521,7 +1521,7 @@ void CompilerUtils::copyContractCodeToMemory(ContractDefinition const& contract,
 				_context.compiledContract(contract) :
 				_context.compiledContractRuntime(contract);
 			// pushes size
-			auto subroutine = _context.addSubroutine(assembly);
+			auto subroutine = _context.addSubroutine(assembly, _creation);
```
ekpyron, line 1524: Or put differently: I'm not sure if we need this change - I'd expect weird thing to happen without it, but couldn't come up with a case.

libevmasm/Assembly.h
```diff
@@ -89,7 +89,7 @@ class Assembly
 
 	/// Adds a subroutine to the code (in the data section) and pushes its size (via a tag)
```
ekpyron, line 90: "subroutine" is a weird name for this actually

libevmasm/Assembly.h
```diff
@@ -204,7 +204,7 @@ class Assembly
 	std::map<util::h256, bytes> m_data;
 	/// Data that is appended to the very end of the contract.
 	bytes m_auxiliaryData;
-	std::vector<std::shared_ptr<Assembly>> m_subs;
+	std::vector<std::pair<std::shared_ptr<Assembly>, bool>> m_subs;
```
chriseth, line 207: Did you consider making the bool part of the `Assembly` class?

ekpyron: Yeah, but only after I had this initial version - if we generally agree that this is the way to go, I can look into moving it into ``Assembly`` and we can go for that, if it works smoothly.

ekpyron:
[rebased to ``develop`` again - will now look into moving the flag into ``Assembly``]

ekpyron:
Closing this in favour of https://github.com/ethereum/solidity/pull/12748 (will reopen if we end up preferring the solution here instead).

Issue #12743, opened 2022-03-04
Remove  "using namespace util" directive from ASTAnnotations.h


wechman:
Because header inclusion can be very deeply nested, "using namespace" directive located in header files can unexpectedly change the meaning of the code. It can also lead to the name collision. For instance, it is not possible to define isDigit function in CommonSyntaxTest.cpp.

TODO:
1) Remove [using namespace util;](https://github.com/ethereum/solidity/blob/develop/libsolidity/ast/ASTAnnotations.h#L50)
2) Update [CommonSyntaxTest.cpp](https://github.com/ethereum/solidity/blob/develop/test/CommonSyntaxTest.cpp) so it has only one local definition of  "isDigit" function (can be addressed after #12740 merge).

YashAtwal:
I'd like to fix this issue

tfire:
Resolved

cameel:
So is the `isDigit()` thing resolved? Can we close this?

tylerfxn:
Yes, resolved @cameel

PR #12745, opened 2022-03-05
Removed "using namespace util" directive from ASTAnnotations.h #12743

YashAtwal:
removed "using namespace util;"

### DIFF
--- a/libsolidity/ast/ASTAnnotations.h
+++ b/libsolidity/ast/ASTAnnotations.h
@@ -47,7 +47,6 @@ namespace solidity::frontend
 
 class Type;
 class ArrayType;
-using namespace util;
 
 struct CallGraph;
 


chriseth:
Thank you for your pull request! Did you try to compile it? It seems that `SetOnce` now needs to be qualified.

YashAtwal:
How would SetOnce be qualified?

chriseth:
I would assume using `util::SetOnce`

nishant-sachdeva (commented):
I tried removing the "using namespace util" line from the file "ASTAnnotations.h" . Looks like this change spawns errors down the line in many files and we'll have to individually pick and fix all of those changes.

Are we sure if this removal is worth it?

cameel:
Looks like we got a similar PR just merged: #12746.

This one was first but the other one addresses the problem fully, passes tests and is already merged so I'm going to close this one.

PR #12746, opened 2022-03-07
Remove use of `using namespace` in header file (`ASTAnnotations.h`)

tylerfxn:
#12743

> 2. Update [CommonSyntaxTest.cpp](https://github.com/ethereum/solidity/blob/develop/test/CommonSyntaxTest.cpp) so it has only one local definition of "isDigit" function.

I'm not sure that this is actually a need. `isdigit` is a native function and not being defined elsewhere.


wechman:
Thank you for the heads-up! Indeed `isdigit` is not there yet. This part can be addressed after #12740 merge. I already updated #12743 description.

wechman:
Hi @tfire! You are going in the right direction with your changes, but there are still things to be addressed:
1) Tests are failing so require update.
2) There is no need to have [using namespace directive](https://github.com/ethereum/solidity/pull/12746/files#diff-c94fd953390e50182709ef3eea4b17c1b46a078b345e1b8d6befeb29d3ab0b8bR30) when [fully qualified names](https://github.com/ethereum/solidity/pull/12746/files#diff-c94fd953390e50182709ef3eea4b17c1b46a078b345e1b8d6befeb29d3ab0b8bR133) are in use.

tylerfxn:
got it building locally @wechman

wechman:
#12740 is already merged so if you like you can also address [CommonSyntaxTest.cpp](https://github.com/ethereum/solidity/blob/develop/test/CommonSyntaxTest.cpp) related part.

tylerfxn:
@wechman will be done later, merge from latest introduced a bunch of calls that aren't qualified

tylerfxn:
@wechman should be done! 🙂

ekpyron (approved):

PR #12748, opened 2022-03-07
Store whether an evmasm Assembly is creation code.

### LINKED ISSUE #11118: Understanding creation context when optimizing Yul subobjects

### LINKED ISSUE #12738: `creationCode` via IR does not match the compiler's `--bin` output when the contract has a constructor with `bytes` arguments

ekpyron:
Alternative to https://github.com/ethereum/solidity/pull/12744
Fixes https://github.com/ethereum/solidity/issues/12738
Fixes https://github.com/ethereum/solidity/issues/11118

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ Compiler Features:
 * JSON-AST: Added selector field for errors and events.
 
 Bugfixes:
+ * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
 
 
 ### 0.8.12 (2022-02-16)
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -1126,6 +1126,11 @@ Above, ``Block`` refers to ``Block`` in the Yul code grammar explained in the pr
 
 .. note::
 
+    An object with a name that ends in ``_deployed`` is treated as deployed code by the Yul optimizer.
+    The only consequence of this is a different gas cost heuristic in the optimizer.
+
+.. note::
+
     Data objects or sub-objects whose names contain a ``.`` can be defined
     but it is not possible to access them through ``datasize``,
     ``dataoffset`` or ``datacopy`` because ``.`` is used as a separator
@@ -1172,17 +1177,17 @@ An example Yul Object is shown below:
 
             // now return the runtime object (the currently
             // executing code is the constructor code)
-            size := datasize("runtime")
+            size := datasize("Contract1_deployed")
             offset := allocate(size)
             // This will turn into a memory->memory copy for Ewasm and
             // a codecopy for EVM
-            datacopy(offset, dataoffset("runtime"), size)
+            datacopy(offset, dataoffset("Contract1_deployed"), size)
             return(offset, size)
         }
 
         data "Table2" hex"4123"
 
-        object "runtime" {
+        object "Contract1_deployed" {
             code {
                 function allocate(size) -> ptr {
                     ptr := mload(0x40)
@@ -1204,7 +1209,7 @@ An example Yul Object is shown below:
                 // code here ...
             }
 
-            object "runtime" {
+            object "Contract2_deployed" {
                 code {
                     // code here ...
                 }
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -411,13 +411,15 @@ map<u256, u256> const& Assembly::optimiseInternal(
 	if (m_tagReplacements)
 		return *m_tagReplacements;
 
+	assertThrow(_settings.isCreation == m_creation, OptimizerException, "Mismatching creation settings.");
+
 	// Run optimisation for sub-assemblies.
 	for (size_t subId = 0; subId < m_subs.size(); ++subId)
 	{
 		OptimiserSettings settings = _settings;
-		// Disable creation mode for sub-assemblies.
-		settings.isCreation = false;
-		map<u256, u256> const& subTagReplacements = m_subs[subId]->optimiseInternal(
+		Assembly& sub = *m_subs[subId];
+		settings.isCreation = sub.isCreation();
+		map<u256, u256> const& subTagReplacements = sub.optimiseInternal(
 			settings,
 			JumpdestRemover::referencedTags(m_items, subId)
 		);
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -411,14 +411,11 @@ map<u256, u256> const& Assembly::optimiseInternal(
 	if (m_tagReplacements)
 		return *m_tagReplacements;
 
-	assertThrow(_settings.isCreation == m_creation, OptimizerException, "Mismatching creation settings.");
-
 	// Run optimisation for sub-assemblies.
 	for (size_t subId = 0; subId < m_subs.size(); ++subId)
 	{
 		OptimiserSettings settings = _settings;
 		Assembly& sub = *m_subs[subId];
-		settings.isCreation = sub.isCreation();
 		map<u256, u256> const& subTagReplacements = sub.optimiseInternal(
 			settings,
 			JumpdestRemover::referencedTags(m_items, subId)
@@ -438,7 +435,7 @@ map<u256, u256> const& Assembly::optimiseInternal(
 				m_items,
 				_tagsReferencedFromOutside,
 				_settings.expectedExecutionsPerDeployment,
-				_settings.isCreation,
+				isCreation(),
 				_settings.evmVersion
 			}.optimise();
 
@@ -539,8 +536,8 @@ map<u256, u256> const& Assembly::optimiseInternal(
 
 	if (_settings.runConstantOptimiser)
 		ConstantOptimisationMethod::optimiseConstants(
-			_settings.isCreation,
-			_settings.isCreation ? 1 : _settings.expectedExecutionsPerDeployment,
+			isCreation(),
+			isCreation() ? 1 : _settings.expectedExecutionsPerDeployment,
 			_settings.evmVersion,
 			*this
 		);
--- a/libevmasm/Assembly.h
+++ b/libevmasm/Assembly.h
@@ -117,7 +117,6 @@ public:
 
 	struct OptimiserSettings
 	{
-		bool isCreation = false;
 		bool runInliner = false;
 		bool runJumpdestRemover = false;
 		bool runPeephole = false;
--- a/libevmasm/Assembly.h
+++ b/libevmasm/Assembly.h
@@ -48,7 +48,7 @@ using AssemblyPointer = std::shared_ptr<Assembly>;
 class Assembly
 {
 public:
-	explicit Assembly(std::string _name = std::string()):m_name(std::move(_name)) { }
+	Assembly(bool _creation, std::string _name): m_creation(_creation), m_name(std::move(_name)) { }
 
 	AssemblyItem newTag() { assertThrow(m_usedTags < 0xffffffff, AssemblyException, ""); return AssemblyItem(Tag, m_usedTags++); }
 	AssemblyItem newPushTag() { assertThrow(m_usedTags < 0xffffffff, AssemblyException, ""); return AssemblyItem(PushTag, m_usedTags++); }
@@ -157,6 +157,8 @@ public:
 	std::vector<size_t> decodeSubPath(size_t _subObjectId) const;
 	size_t encodeSubPath(std::vector<size_t> const& _subPath);
 
+	bool isCreation() const { return m_creation; }
+
 protected:
 	/// Does the same operations as @a optimise, but should only be applied to a sub and
 	/// returns the replaced tags. Also takes an argument containing the tags of this assembly
@@ -214,6 +216,8 @@ protected:
 	mutable std::vector<size_t> m_tagPositionsInBytecode;
 
 	int m_deposit = 0;
+	/// True, if the assembly contains contract creation code.
+	bool const m_creation = false;
 	/// Internal name of the assembly object, only used with the Yul backend
 	/// currently
 	std::string m_name;
--- a/libyul/backends/evm/EVMObjectCompiler.cpp
+++ b/libyul/backends/evm/EVMObjectCompiler.cpp
@@ -30,6 +30,8 @@
 #include <libyul/Object.h>
 #include <libyul/Exceptions.h>
 
+#include <boost/algorithm/string.hpp>
+
 using namespace solidity::yul;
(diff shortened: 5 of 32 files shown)


libevmasm/Assembly.cpp
```diff
@@ -431,13 +431,16 @@ map<u256, u256> const& Assembly::optimiseInternal(
 	if (m_tagReplacements)
 		return *m_tagReplacements;
 
+	assertThrow(_settings.isCreation == m_creation, OptimizerException, "Mismatching creation settings.");
```
ekpyron: We could also remove the flag in the settings entirely...

test/cmdlineTests/viair_subobject_optimization/input.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >0.0.0;
+
+contract C {
+  constructor(uint x) {
+    // The constant should be optimized away and occur neither in the bytecode of C
+    // nor embedded in the bytecode of D.
```
ekpyron: Not sure if we have - or should come up with - a better way to test this...

It might even make sense to compile each and every semantics test contract directly and compare the bytecode with compiling it together with another contract that returns its ``type(...).creationCode``....

test/cmdlineTests/viair_subobject_optimization/output
```diff
@@ -0,0 +1,8 @@
...
+
+======= viair_subobject_optimization/input.sol:D =======
+Binary:
+6080806040523461001657610203908161001d8239f35b50600080fdfe6080806040526004361015610015575b50600080fd5b600090813560e01c6326121ff01461002d575061000f565b3461009657816003193601126100965760ce603f8101601f1916820167ffffffffffffffff81118382101761007f5761007b93506040528082526100f260208301396040519182918261009a565b0390f35b5050634e487b7160e01b8252506041600452602490fd5b5080fd5b919091602080825283519081818401526000945b8286106100db5750508060409394116100ce575b601f01601f1916010190565b60008382840101526100c2565b85810182015184870160400152948101946100ae56fe6080806040523460525760ce80380390601f19601f8301168301918383106001600160401b03841117605857808492602094604052833981010312605257604590516072565b604051604d908160818239f35b50600080fd5b50505050634e487b7160e01b600052604160045260246000fd5b60be600160481b031460525756fe6080604052600080fdfea26c6578706572696d656e74616cf564736f6c63782b302e382e31332d646576656c6f702e323032322e332e372b636f6d6d69742e37633931646430352e6d6f640041a26c6578706572696d656e74616cf564736f6c63782b302e382e31332d646576656c6f702e323032322e332e372b636f6d6d69742e37633931646430352e6d6f640041
```
ekpyron: Note that on develop this contains ``ffffffffffffffff42``, while the one for ``C`` doesn't.

test/cmdlineTests/viair_subobject_optimization/output
```diff
@@ -0,0 +1,378 @@
...
+          pop
+          pop
+          pop
+          pop
+          mstore(0x00, shl(0xe0, 0x4e487b71))
+          mstore(0x04, 0x41)
+          revert(0x00, 0x24)
+            /* "viair_subobject_optimization/input.sol":76:278  constructor(uint x) {... */
+        tag_1:
+          sub(shl(0x48, 0x01), 0xbe)
```
ekpyron, line 357: Note that this is ``0xffffffffffffffff42`` on develop, while it's still constant-optimized in the version above.

test/cmdlineTests/yul_optimize_runs/output
```diff
@@ -5,12 +5,12 @@ Pretty printed source:
...
         {
-            let _1 := datasize("Runtime")
-            datacopy(0, dataoffset("Runtime"), _1)
+            let _1 := datasize("Runtime_deployed")
+            datacopy(0, dataoffset("Runtime_deployed"), _1)
             return(0, _1)
         }
     }
-    object "Runtime" {
+    object "Runtime_deployed" {
```
ekpyron, line 13: This case now depends on the name of the subobject... which is not really nice...

Changelog.md
```diff
@@ -8,6 +8,7 @@ Compiler Features:
 * JSON-AST: Added selector field for errors and events.
 
 Bugfixes:
+ * Yul Optimizer: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
```
chriseth: Shouldn't this also say Yul IR codegen?

docs/yul.rst
```diff
@@ -1124,6 +1124,10 @@ regular strings in native encoding. For code,
 
 Above, ``Block`` refers to ``Block`` in the Yul code grammar explained in the previous chapter.
 
+.. note::
+
+    An object with a name that ends in ``_deployed`` is treated as runtime code by the Yul optimizer.
```
chriseth: We should say that the only consequence of this is the tuning of the gas cost heuristic of the optimizer.

libevmasm/Assembly.cpp
chriseth: Yeah, that actually sounds better.

libevmasm/Assembly.h
```diff
@@ -157,6 +157,8 @@ class Assembly
 	std::vector<size_t> decodeSubPath(size_t _subObjectId) const;
 	size_t encodeSubPath(std::vector<size_t> const& _subPath);
 
+	bool isCreation() const { return m_creation; }
```
chriseth, line 159: What about changing this to "executedOnce"?

ekpyron: The effect in the optimizer is not only adjusting the runs parameter, but also stuff like
``m_params.isCreation ? GasCosts::txDataNonZeroGas(m_params.evmVersion) : GasCosts::createDataGas``
in the constant optimizer calculations...

chriseth: Yeah but as I said, for tx data gas, it should be relevant whether it is a top-level object.

ekpyron: Yeah, but right now the flag does both, so it's really a "creation mode" flag so far...

If we want to additionally consider whether something is a top-level object, then we run into the same problem again: the compiler would output different bytecode for ``C`` compiled in isolation compared to ``C`` used from another contract ``D``... what we really need is per-contract optimizer settings for this, don't we?

libevmasm/Assembly.cpp
ekpyron: Done in another commit now.

docs/yul.rst
```diff
@@ -1124,6 +1124,11 @@ regular strings in native encoding. For code,
...
```
chriseth:
```suggestion
    An object with a name that ends in ``_deployed`` is treated as deployed code by the Yul optimizer.
```
(creation code is also runtime code)

libevmasm/Assembly.h
```diff
@@ -48,7 +48,7 @@ using AssemblyPointer = std::shared_ptr<Assembly>;
 class Assembly
 {
 public:
-	explicit Assembly(std::string _name = std::string()):m_name(std::move(_name)) { }
+	explicit Assembly(bool _creation, std::string _name): m_creation(_creation), m_name(std::move(_name)) { }
```
chriseth:
```suggestion
	Assembly(bool _creation, std::string _name): m_creation(_creation), m_name(std::move(_name)) { }
```

chriseth (dismissed):

libyul/backends/evm/EVMObjectCompiler.cpp
```diff
@@ -48,7 +50,8 @@ void EVMObjectCompiler::run(Object& _object, bool _optimize)
 	for (auto const& subNode: _object.subObjects)
 		if (auto* subObject = dynamic_cast<Object*>(subNode.get()))
 		{
-			auto subAssemblyAndID = m_assembly.createSubAssembly(subObject->name.str());
+			bool isCreation = !boost::ends_with(subObject->name.str(), "_deployed");
```
chriseth, line 53: This could actually use a (new) function of IRNames instead of a fixed string, but only if we do that everywhere.

chriseth (approved):
Using IRNames is probably not viable.

PR #12749, opened 2022-03-07
[CMake] Refactor libsolc linker flag for EXPORTED_FUNCTIONS and make it more readable.

christianparpart:
Extraction from #12418


ekpyron:
Can you also add the functions in ``EXTRA_EXPORTED_RUNTIME_METHODS`` in ``cmake/EthCompilerSettings.cmake``?
``EXTRA_EXPORTED_RUNTIME_METHODS`` has been deprecated for a while - I've been meaning to move them to ``EXPORTED_FUNCTIONS`` for a while, but never got around to.

EDIT: Actually wait... the warning is
```
em++: warning: EXTRA_EXPORTED_RUNTIME_METHODS is deprecated, please use EXPORTED_RUNTIME_METHODS instead [-Wdeprecated]
```

so those should actually end up in ``EXPORTED_RUNTIME_METHODS`` not in ``EXPORTED_FUNCTIONS``....
So we can do it separately, i.e. nevermind for this PR.

libsolc/CMakeLists.txt
```diff
@@ -1,8 +1,39 @@
+
+# CreateExportedFunctions(OUTPUT_VARIABLE Symbol1 Symbol2 ... SymbolN)
+macro(CreateExportedFunctions)
```
ekpyron: Can't you just add the output variable as ``macro(CreateExportedFunctions OUTPUT_VARIABLE)`` instead of all the special case handling below?

libsolc/CMakeLists.txt
```diff
@@ -1,8 +1,39 @@
...
+	if("${ARGN}" STREQUAL "")
+		message(FATAL_ERROR "CreateExportedFunctions: No arguments passed.")
+	endif()
+	set(result "")
+	foreach(_name IN ITEMS ${ARGN})
+		if("${result}" STREQUAL "") # skip first element and start JS array
+			set(result "[")
+		else()
+			if(NOT("${result}" STREQUAL "["))
+				set(result "${result}, ")
```
ekpyron: The PR doesn't seem to work (see the ``t_ems`` runs)...
Might be the space here.

libsolc/CMakeLists.txt
cameel: I'd put the macro in some utils file in `cmake/`. Other than that, having this does seem like an improvement as long as it works correctly.

christianparpart: it made it a little bit shorter. Thanks.

ekpyron:
Since we require cmake 3.13 anyways, you can also simply use https://cmake.org/cmake/help/latest/command/list.html#join on a list, can't you?

christianparpart:
> Since we require cmake 3.13 anyways, you can also simply use https://cmake.org/cmake/help/latest/command/list.html#join on a list, can't you?

@ekpyron It's not just joining but alsp prefixing the symbols with `_`. This is done to make it easier on the caller side, to allow word-based auto-completion and avoid accidental mistypings of the symbol names.

cmake/EthUtils.cmake
```diff
@@ -48,3 +48,21 @@ function(detect_stray_source_files FILELIST DIRECTORY)
...
+	foreach(_name IN ITEMS ${ARGN})
+		if(NOT("${result}" STREQUAL ""))
+			set(result "${result},")
+		endif()
+		set(result "${result}\"_${_name}\"") # Append and prepend leading underscore.
+	endforeach()
+	set(result "[${result}]")
+	message(STATUS "CreateExportedFunctionsForEMSDK: ${result}")
+	set(${ARGV0} "${result}" PARENT_SCOPE)
+endfunction()
```
ekpyron:
```suggestion
function(CreateExportedFunctionsForEMSDK OUTPUT_VARIABLE)
  list(TRANSFORM ARGN PREPEND "\"_")
  list(TRANSFORM ARGN APPEND "\"")
  list(JOIN ARGN "," ARGN)
  set(${OUTPUT_VARIABLE} "[${ARGN}]" PARENT_SCOPE)
endfunction()
```
That's my point.

ekpyron (approved):
Looks good to me if tests pass.

PR #12750, opened 2022-03-07
typeCheckAbiEncodeCallFunction should type check the arguments on functionPointerType->asExternallyCallableFunction instead of the plain function type

### LINKED ISSUE #12718: `abi.encodeCall` does not properly convert the function type to an externally-called function
## Abstract

The following gives an error:

```solidity
pragma solidity 0.8.12;

contract Sender {
    function main(string memory helloworld) external {
        // TypeError: Cannot implicitly convert component at position 0 from "string" to "string".
        abi.encodeCall(IReceiver.methodName, (helloworld));
    }
}

interface IReceiver {
    function methodName(string calldata) external;
}
```

## Motivation

When encoding data for a call, it shouldn't really matter if it's calldata or memory. If the interface has calldata, but if the value that we want to pass is in memory, it triggers the typecheck error abi.encodeCall. 

## Specification

abi.encodeCall should not typecheck data location.

## Backwards Compatibility

This should not have any breakings.

nishant-sachdeva:
Closes #12718


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2190,7 +2190,9 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	for (size_t i = 0; i < numParameters; i++)
 	{
 		Type const& argType = *type(*callArguments[i]);
-		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]);
+		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->asExternallyCallableFunction(
+								functionPointerType->kind() == FunctionType::Kind::DelegateCall ? true:false
```
ekpyron: Why not
```suggestion
								functionPointerType->kind() == FunctionType::Kind::DelegateCall
```
?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2190,7 +2190,11 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	for (size_t i = 0; i < numParameters; i++)
 	{
 		Type const& argType = *type(*callArguments[i]);
-		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]);
+		BoolResult result = argType.isImplicitlyConvertibleTo(
+								*functionPointerType->asExternallyCallableFunction(
+									functionPointerType->kind() == FunctionType::Kind::DelegateCall
```
ekpyron: Looking at the code above: this is always false, isn't it :-)? ``functionPointerType->kind() == FunctionType::Kind::DelegateCall`` is an error case above that returns early, right?

nishant-sachdeva: Yes. You're right. Will change it in the next push.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2190,7 +2190,10 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	for (size_t i = 0; i < numParameters; i++)
 	{
 		Type const& argType = *type(*callArguments[i]);
-		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]);
+		// functionPointerType is never equal to DelegateCall ( it would have thrown error above )
+		BoolResult result = argType.isImplicitlyConvertibleTo(
+								*functionPointerType->asExternallyCallableFunction(false)->parameterTypes()[i]
+							);
```
chriseth: Indentation is off here

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2196,7 +2196,10 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	for (size_t i = 0; i < numParameters; i++)
 	{
 		Type const& argType = *type(*callArguments[i]);
-		BoolResult result = argType.isImplicitlyConvertibleTo(*functionPointerType->parameterTypes()[i]);
+		// functionPointerType is never equal to DelegateCall ( it would have thrown error above )
```
ekpyron: I'd say you can either just drop the comment or if it should be more explicit make it a proper
```suggestion
		solAssert(functionPointerType->kind() != FunctionType::Kind::Declaration);
```

chriseth: See line 2124. I would actually do the conversion to external callable function much closer to where the pointer is defined.

ekpyron: Yeah, I was actually wondering whether to suggest that as well!

test/libsolidity/semanticTests/abiEncoderV1/bool_out_of_bounds.sol
```diff
@@ -3,9 +3,10 @@ contract C {
 	function f(bool b) public pure returns (bool) { return b; }
 }
 // ====
+// compileViaYul: false
```
chriseth: Please revert these, there is no `abi.encodeCall` in these tests, so they should not change.

chriseth:
@ekpyron I'm seeing that ExpressionCompiler.cpp just calls `abiEncode` without a target type. Is this actually safe?

ekpyron:
> @ekpyron I'm seeing that ExpressionCompiler.cpp just calls `abiEncode` without a target type. Is this actually safe?

We should at least change it, but let me think if I can come up with an example that breaks.

ekpyron:
``IRGeneratorForStatements.cpp`` also doesn't use the parameter types of the function type, but the encoding types of the argument tuple...

test/libsolidity/semanticTests/abiEncoderV1/bool_out_of_bounds.sol
nishant-sachdeva: Reverted the changes in the latest push.

Changelog.md
```diff
@@ -10,6 +10,7 @@ Compiler Features:
 
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
+* TypeChecker: ABIEncodeCallFunction properly converts function type to externally called function.
```
```suggestion
* TypeChecker: Use external function parameter types for `abi.encodeCall`.
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2154,7 +2154,10 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 		return;
 	}
 
-	solAssert(!functionPointerType->takesArbitraryParameters(), "Function must have fixed parameters.");
+	solAssert(functionPointerType->kind() != FunctionType::Kind::DelegateCall);
+	auto const externallyCallableFunction = functionPointerType->asExternallyCallableFunction(false);
```
cameel: I'd call it `externalFunctionType` since it's still a function type.

Also, we generally avoid `auto` unless the type is obvious from context - as it was at the point of declaration of `functionPointerType` due to `dynamic_cast`. Here it's not. And when we use auto, we at least include the `*` or `&` to make it clear that it's a pointer/reference.

```suggestion
	FunctionType const* externalFunctionType = functionPointerType->asExternallyCallableFunction(false);
```

Changelog.md
```diff
@@ -10,6 +10,7 @@ Compiler Features:
 
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
+* TypeChecker: Use external function parameter types for `abi.encodeCall`.
```
Marenz: I think a better description would be "Use location independent type checking for function parameter types used with `abi.encodeCall`"

libsolidity/analysis/TypeChecker.cpp
Marenz: Just a though, might it be better to overwrite `functionPointerType` with the result of this call instead? It is not used afterwards but could lead to accidentally be used in the future if one is not careful.

Marenz (changes requested):

test/libsolidity/syntaxTests/abiEncoder/passing_memory_to_interface_that_expects_calldata_should_not_throw.sol
```diff
@@ -0,0 +1,10 @@
+contract Sender {
```
cameel: I think this should be more thoroughly tested. The reporter provided one case but when implementing a fix, please always try to come up with other similar cases and add some of them. For example here:
- I'd test all the 6 combinations (`calldata`/`calldata`, `memory`/`calldata`, `calldata`/`memory`, `memory`/`memory`, `storage`/`calldata`, `storage`/`memory`). `memory`/`memory` is probably already covered by existing tests but please check.
- Are there any other types that could cause a mismatch like this?
    - So far I managed to come up with three (see https://github.com/ethereum/solidity/issues/12718#issuecomment-1064322093). Please add these as test cases here and check if your fix already covers them. Also, please try to see if you can come up with more.
    - For a good measure I'd still add some more types in this case: `bytes`, static and dynamic array, struct.
- A semantic test for the `calldata`/`memory` case would not hurt - so that we're sure that codegen can actually handle it correctly despite the slight difference in types. A function pointer case might also warrant a semantic test since I remember that we did have some asserts failing at some point because they did not expect slightly different pointer types. I'd verify it works at least.

cameel (changes requested):

Changelog.md
cameel: It's actually not only about location. You get the same thing with `address` vs `address payable` for example.

But I agree that this changelog entry could be a bit less terse. It's technically correct but as a user I think I'd have trouble figuring out what it really means.

libsolidity/analysis/TypeChecker.cpp
cameel: Not sure about others but personally I prefer choosing names that are distinct enough and keeping the original variable over overwriting the same variable with something slightly different. I actually wish variables were `const` by default, which would somewhat discourage overwriting them. You might still want to access the previous value later or you could move a chunk of code not realizing that at the new location the variable has a new meaning.

cameel:
I just tested the `b_ubu_static` artifact from this PR with the cases I listed in https://github.com/ethereum/solidity/issues/12718#issuecomment-1064322093 and looks like the PR does fix them. I.e. it fixes the first one and the other two are actually expected to error out due to types that are not implicitly convertible.

cameel:
Ah, wait. The case with function pointers is actually valid. This is the same exact function type and it's rejected simply because the function type has a `calldata` argument. For `memory` it does not happen. The PR does not fix that problem.

libsolidity/analysis/TypeChecker.cpp
Marenz: If you need to access something later, you can always change it back, in that case it is at least a very conscious decision to do so.
An alternative could be to never declare `functionPointerType` in the function scope but a tiny sub-scope where the only thing it does is assigning to an `externallyCallableFunction` (or however you want to call it) variable in the function scope.
As before, if you DO need to access it at a later time, you can always move it out of the scope.

cameel: > An alternative could be to never declare `functionPointerType` in the function scope

Yeah, I agree, this would be much better.

ekpyron:
> Ah, wait. The case with function pointers is actually valid. This is the same exact function type and it's rejected simply because the function type has a `calldata` argument. For `memory` it does not happen. The PR does not fix that problem.

That's a separate issue, though, isn't it? That also doesn't work for directly calling a function...
<details>
<pre>
interface I {
      function f(function (string calldata) external) external;
  }
  
  contract C {
      function g(string calldata) external {}
  
      function main() external view {
          I x;
          x.f(this.g);
          abi.encodeCall(I.f, (this.g));
      }
  }
</pre>
fails with both
<pre>
Error: Invalid type for argument in function call. Invalid implicit conversion from function (string memory) external to function (string calldata) external requested.
  --> test.sol:10:6:
   |
10 |    x.f(this.g);
   |        ^^^^^^

Error: Cannot implicitly convert component at position 0 from "function" to "function".
  --> test.sol:11:29:
   |
11 |         abi.encodeCall(I.f, (this.g));
   |                             ^^^^^^^^
</pre>
after this PR

</details>

So that should be fixed in ``asExternallyCallableFunction`` (or whatever it's coming from) instead.

cameel:
> That's a separate issue, though, isn't it?

True. Looks like it's not a problem with `abi.encodeCall()` but a more general one. I'll report a separate issue so that we can focus only on `encodeCall()` here.

cameel:
Reported in #12778.

chriseth:
there is a coding style problem

cameel:
Also, still needs more test cases in my opinion.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,61 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+				"Expected first argument to be a function pointer, not \"" +
+				type(*arguments.front())->toString() +
+				"\"."
+			);
+			return;
+		}
+
+		externalFunctionType = functionPointerType->asExternallyCallableFunction(
+													functionPointerType->kind() == FunctionType::Kind::DelegateCall
+												);
```
```suggestion
		externalFunctionType = functionPointerType->asExternallyCallableFunction(
			functionPointerType->kind() == FunctionType::Kind::DelegateCall
		);
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,61 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
-		m_errorReporter.typeError(
-			5511_error,
-			arguments.front()->location(),
-			"Expected first argument to be a function pointer, not \"" +
-			type(*arguments.front())->canonicalName() +
-			"\"."
-		);
-		return;
+		auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front()));
+		if (!functionPointerType)
```
cameel: Another way to avoid having this extra variable visible in the whole function, without introducing an artificial block, would be to do this:
```suggestion
		if (auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front())))
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,61 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+				arguments.front()->location(),
+				"Expected first argument to be a function pointer, not \"" +
+				type(*arguments.front())->toString() +
+				"\"."
+			);
+			return;
+		}
+
+		externalFunctionType = functionPointerType->asExternallyCallableFunction(
+													functionPointerType->kind() == FunctionType::Kind::DelegateCall
```
cameel: Can this actually be a delegatecall in practice? I think you had an assert before and that seemed fine to me.

test/libsolidity/syntaxTests/abiEncoder/abi_encode_conversions.sol
```diff
@@ -0,0 +1,30 @@
...
+contract testContract {
+    function g(string calldata) external {}
+    function h(string memory) external {}
+    string s;
+
+    function main() external view {
+        abi.encodeCall(testInterface.A, (address(0)));
+        abi.encodeCall(testInterface.B, (s));
+        abi.encodeCall(testInterface.C, (s));
+        abi.encodeCall(testInterface.D, (s));
```
cameel, line 27: This covers the `storage`/`calldata` combination. Can you also add others for full coverage, like I mentioned in https://github.com/ethereum/solidity/pull/12750/files#r823996395?
- Encoding a `calldata` value as an input for a `calldata` argument
- Encoding a `calldata` value as an input for a `memory` argument
- Encoding a `memory` value as an input for a `calldata` argument
- Encoding a `storage` value as an input for a `memory` argument

test/libsolidity/syntaxTests/abiEncoder/abi_encode_conversions.sol
```diff
@@ -0,0 +1,30 @@
...
+    function D(string calldata) external;
+}
+
+contract testContract {
+    function g(string calldata) external {}
+    function h(string memory) external {}
+    string s;
+
+    function main() external view {
+        abi.encodeCall(testInterface.A, (address(0)));
```
cameel: Can you also add a test the other way around? I.e. encoding `address payable` as an input for an `address` argument?

test/libsolidity/syntaxTests/abiEncoder/passing_memory_to_interface_that_expects_calldata_should_not_throw.sol
cameel: Could you add some of these?

libsolidity/analysis/TypeChecker.cpp
nishant-sachdeva: I felt this made it much more readable.
Plus there is a check in the code below, that carries an assert that it can't be DelegateCall. 
So, having an assert here as well seemed repetitive.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,58 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 		return;
 	}
 
-	auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front()));
-
-	if (!functionPointerType)
+	FunctionType const* externalFunctionType = nullptr;
+	if (auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front())))
+		externalFunctionType = functionPointerType->asExternallyCallableFunction(
+			functionPointerType->kind() == FunctionType::Kind::DelegateCall
```
chriseth: This should notbe true - see below.

Changelog.md
```diff
@@ -37,6 +37,7 @@ Bugfixes:
  * General: Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
  * Type Checker: Fix incorrect type checker errors when importing overloaded functions.
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
+* TypeChecker: Use external function parameter types for `abi.encodeCall`.
```
chriseth: This has to be moved to the new version.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,59 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
 
-	auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front()));
-
-	if (!functionPointerType)
+	FunctionType const* externalFunctionType = nullptr;
+	if (auto const functionPointerType = dynamic_cast<FunctionTypePointer>(type(*arguments.front())))
+	{
+		// functionPointerType->kind() != FunctionType::Kind::DelegateCall
+		externalFunctionType = functionPointerType->asExternallyCallableFunction(false);
+	}
```
```suggestion
		// this cannot be a library function, that is checked below
		externalFunctionType = functionPointerType->asExternallyCallableFunction(false);
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2111,57 +2111,59 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
 			"Expected first argument to be a function pointer, not \"" +
-			type(*arguments.front())->canonicalName() +
+			type(*arguments.front())->toString() +
 			"\"."
 		);
 		return;
 	}
 
 	if (
-		functionPointerType->kind() != FunctionType::Kind::External &&
```
chriseth, line 2129: Please do this check on the original type. Otherwise I think we would introduce a bug.

chriseth: Ah ok, it currently is the same thing. Then please add an assertion above that the kind is the same for both type pointers.

test/libsolidity/syntaxTests/abiEncoder/abi_encode_convert_address_to_address_payable.sol
```diff
@@ -1,11 +0,0 @@
-interface I {
```
chriseth, line 1: Why did you delete these tests?

Changelog.md
chriseth: What about `Convert parameters of function type to how they would be called for abi.encodeCall.`?

test/libsolidity/syntaxTests/abiEncoder/abi_encode_convert_address_to_address_payable.sol
nishant-sachdeva: Seemed redundant to have them spread across different files. Now we have them in two files based on whether the code is working correctly or not.

Edit : It's not a perfectly working/not-working split. The idea of the split is to keep the test cases concerning function pointers (something that the this PR does not fix and which will be fixed with #12778) in a separate file and all the other test cases concerning this issue in one file.

chriseth: Ah, you just moved them!
I don't see that they are properly split into working / not working, though.

nishant-sachdeva: Ah yes, it's not a perfect split. 
The issue #12778 that should be solving the error statements coming in these functions only mentioned functions. So, I didn't move the address pointer test to the other file

chriseth: Please do it properly. When we address #12778 we can still change the tests.

Changelog.md
```diff
@@ -8,6 +8,7 @@ Compiler Features:
 
 
 Bugfixes:
+* TypeChecker: Convert parameters of function type to how they would be called for abi.encodeCall.
```
```suggestion
* TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
```

Changelog.md
```diff
@@ -38,7 +39,6 @@ Bugfixes:
  * Type Checker: Fix incorrect type checker errors when importing overloaded functions.
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
 
-
```
chriseth: Please undo this.

chriseth (approved):

PR #12754, opened 2022-03-07
Some more peephole optimizer rules.

chriseth:


ekpyron (commented):
Needs maybe a changelog entry, and needs legacy non-optimized gas expecation updates.

Apart from that: looks good and nice that it seems to occur quite often!

chriseth:
On the example from twitter:
```
Gas legacy pre opt:
construction:
   93 + 40200 = 40293
external:
   count():	2261
   foo():	24431

Gas IR pre opt:
construction:
   99 + 41400 = 41499
external:
   count():	2258
   foo():	24399


Post optimization:
legacy:
construction:
   93 + 40000 = 40093
external:
   count():	2261
   foo():	24428

via IR:
construction:
   99 + 40600 = 40699
external:
   count():	2258
   foo():	24398

```

ekpyron:
That's ``--gas`` estimates? Why are they that far off :-)?
If I actually call ``foo`` in a semantics test with ``--enforce-gas-cost-min-value 1`` I get
```
contract Memo {
        uint public count;
        function foo() external { ++count; }
}
// ====
// compileViaYul: also
// ----
// foo() ->
// gas irOptimized: 43362
// gas legacyOptimized: 43392
```
(with this PR, without it 1 gas more in irOptimized and 3 gas more in legacyOptimized)

chriseth:
`--gas` is just the execution gas, without the transaction gas.

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -46,6 +46,14 @@ struct ApplyRule
 {
 };
```
ekpyron:
```suggestion
{
	template <size_t... Indices>
	static bool applyRule(AssemblyItems::const_iterator _in, back_insert_iterator<AssemblyItems> _out, index_sequence<Indices...>)
	{
		return Method::applySimple(_in[Indices]..., _out);
	}

	static bool applyRule(AssemblyItems::const_iterator _in, back_insert_iterator<AssemblyItems> _out)
	{
		return applyRule(_in, _out, make_index_sequence<Arguments>{});
	}
};
```
and removing all the explicit specializations should just work, shouldn't it?

chriseth: maybe it does in the meantime, let me check.

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -256,6 +234,65 @@ struct IsZeroIsZeroJumpI: SimplePeepholeOptimizerMethod<IsZeroIsZeroJumpI, 4>
...
+	)
+	{
+		if (
+			_eq == Instruction::EQ &&
+			_iszero == Instruction::ISZERO &&
+			_pushTag.type() == PushTag &&
+			_jumpi == Instruction::JUMPI
+		)
+		{
+			*_out = AssemblyItem(Instruction::XOR, _eq.location());
```
chriseth: SUB would have the same effect and could be better for formal analysis

libevmasm/PeepholeOptimiser.cpp
ekpyron: We could probably simplify it even further, s.t. there is no need for spelling out the size anywhere including in ``SimplePeepholeOptimizerMethod<..., WindowSize>``... but not sure that's worth it...

ekpyron:
```c++

template<typename FunctionType>
struct GetArgumentCount;
template<typename R, typename... Args>
struct GetArgumentCount<R(Args...)>
{
	static constexpr auto value = sizeof...(Args);
};

template <class Method>
struct SimplePeepholeOptimizerMethod
{
	template <size_t... Indices>
	static bool applyRule(AssemblyItems::const_iterator _in, back_insert_iterator<AssemblyItems> _out, index_sequence<Indices...>)
	{
		return Method::applySimple(_in[Indices]..., _out);
	}
	static bool apply(OptimiserState& _state)
	{
		static constexpr size_t WindowSize = GetArgumentCount<decltype(Method::applySimple)>::value - 1;
		if (
			_state.i + WindowSize <= _state.items.size() &&
			applyRule(_state.items.begin() + static_cast<ptrdiff_t>(_state.i), _state.out, make_index_sequence<WindowSize>{})
		)
		{
			_state.i += WindowSize;
			return true;
		}
		else
			return false;
	}
};

```

ekpyron: With that there shouldn't be any need for ``ApplyRule`` at all anymore and the window size template argument can be dropped from all methods... although this is just rather silly distraction...

chriseth:
I'll actually change it to SUB.

chriseth:
Haha, changing it to SUB activates some more optimizer opportunities in the CSE I think :)

But it also makes `semanticTests/functionCall/external_call_to_nonexisting_debugstrings.sol` 636 gas more expensive...

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -233,6 +233,65 @@ struct IsZeroIsZeroJumpI: SimplePeepholeOptimizerMethod<IsZeroIsZeroJumpI>
...
+		std::back_insert_iterator<AssemblyItems> _out
+	)
+	{
+		if (
+			_pushTag1.type() == PushTag &&
+			_jumpi == Instruction::JUMPI &&
+			_pushTag2.type() == PushTag &&
+			_jump == Instruction::JUMP &&
+			_tag1.type() == Tag &&
+			_pushTag1.data() == _tag1.data()
```
ekpyron, line 281: This is robust against the subassembly encoding in the tags, since obviously the tag is in the same subassembly, right? Might be worth a comment.

Changelog.md
```diff
@@ -6,6 +6,7 @@ Language Features:
 
 Compiler Features:
 * JSON-AST: Added selector field for errors and events.
+ * Peephole Optimizer: Optimize comparisons and double jumps in front of conditional jumps.
```
```suggestion
 * Peephole Optimizer: Optimize comparisons in front of conditional jumps and conditional jumps across a single unconditional jump.
```
?

ekpyron (approved):

PR #12756, opened 2022-03-08
Refactored solc --link to use LinkerObject

### LINKED ISSUE #10308: [CLI] Rewrite solc --link to use LinkerObject
Extracted from https://github.com/ethereum/solidity/pull/10297#discussion_r524260360:

@axic
> Why do we have this messy code here and why isn't this part of `LinkerObject`? Also `solc-js` has this messy code duplicated.

> I would prefer a clear way of adding a `LinkerObject::parse(bytes)` feature and unit testing, and using `LinkerObject::parse(bin).link(addresses)` in the cli.

> You can take inspiration from https://github.com/ethereum/solc-js/blob/master/linker.js (see `findLinkReferences`).

bmoore01:
I have refactored solc/CommandLineInterface.cpp Fixes #10308


solc/CommandLineInterface.cpp
```diff
@@ -916,60 +913,69 @@ void CommandLineInterface::serveLSP()
...
-			else
-				serr() << "Reference \"" << foundPlaceholder << "\" in file \"" << src.first << "\" still unresolved." << endl;
-			it += placeholderSize;
 		}
+
+		evmasm::LinkerObject linkerObject
+			= {fromHex(bytecode),
+			   linkReferences,
+			   map<u256, std::pair<std::string, std::vector<size_t>>>(),
+			   std::map<std::string, evmasm::LinkerObject::FunctionDebugData>()};
```
cameel: That's how we usually wrap it (more readable IMO). Also, you had a mix of tabs and spaces here. We only use tabs for indentation in C++ code.
```suggestion
		evmasm::LinkerObject linkerObject = {
			fromHex(bytecode),
			linkReferences,
			map<u256, std::pair<std::string, std::vector<size_t>>>(),
			std::map<std::string, evmasm::LinkerObject::FunctionDebugData>()
		};
```

solc/CommandLineInterface.cpp
```diff
@@ -916,60 +913,69 @@ void CommandLineInterface::serveLSP()
...
 
+vector<string> extractLines(string const& _source)
+{
+	auto stream = stringstream{_source};
+	vector<string> result;
+	string line;
+	while(getline(stream, line, '\n'))
+	{
+		result.push_back(line);
+	}
```
```suggestion
	while (getline(stream, line, '\n'))
		result.push_back(line);
```

solc/CommandLineInterface.cpp
```diff
@@ -916,60 +913,69 @@ void CommandLineInterface::serveLSP()
...
 	int const placeholderSize = 40; // 20 bytes or 40 hex characters
 	for (auto const& library: m_options.linker.libraries)
 	{
 		string const& name = library.first;
-		// Library placeholders are 40 hex digits (20 bytes) that start and end with '__'.
-		// This leaves 36 characters for the library identifier. The identifier used to
-		// be just the cropped or '_'-padded library name, but this changed to
-		// the cropped hex representation of the hash of the library name.
-		// We support both ways of linking here.
-		librariesReplacements["__" + evmasm::LinkerObject::libraryPlaceholder(name) + "__"] = library.second;
```
cameel: Please preserve comments if they still make sense after your changes.

solc/CommandLineInterface.cpp
```diff
@@ -916,60 +913,69 @@ void CommandLineInterface::serveLSP()
...
+	auto stream = stringstream{_source};
+	vector<string> result;
+	string line;
+	while(getline(stream, line, '\n'))
+	{
+		result.push_back(line);
+	}
+
+	return result;
+}
```
cameel: You could replace this with `boost::split()`.

cameel (changes requested):
Thanks for the PR! I'll review it in detail later but here's some early feedback. For now just a few superficial things that stood out.

Also, test are failing. Please take a look at the failed CI runs. I think you must have removed some imports that are actually used.

cameel:
Two more things:
- Please add `Fixes #10308` to your description so that github will automatically close the issue when we merge this.
- I'd recommend not submitting PRs from the `develop` branch of your fork. Keeping it up to date with `develop` in the upstream repo is always a bit awkward. In my experience it's much more convenient to use a feature branch.

bmoore01:
@cameel Thanks for the review, I have made another commit addressing your initial issues so hopefully this fixes any issues!

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
 		// be just the cropped or '_'-padded library name, but this changed to
```
chriseth: I don't think we need this comment any more, at least not at this point in the code.

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
+		string bytecode;
+		vector<string> sourceLines;
+		boost::split(sourceLines, src.second, boost::is_any_of("\n"));
+
+		if (sourceLines.size() < 3)
+			// no bytecode found
+			continue;
+
+		size_t bytecodeIndex;
+		// Bytecode can be on line 0 or 3, depending on if file header is included
```
chriseth: Do we really want to account for the output formatting of the non-file-based commandline interface?

bmoore01: Ideally yes, because this is required to allow the tests to pass although, I'm away that it's not the most elegant solution. I also considered using a regex to find the index instead of hard coding it but, this seemed simpler.

cameel: I'd do one of these here:
1) Assume a rigid file structure and validate it more strictly. For example when you skip files shorter than 3 lines, verify that these lines do not contain the bytecode (i.e. a string consisting only of hex chars without whitespace and possibly placeholders between them)
2) Detect the bytecode and ignore everything else. This will be more robust against things like someone just taking the compiler output (with multiple bytecode pieces in it), putting it in a file and running the linker on it.

The current version is neither of these.

Given how the linker has been working so far (basically accepting anything, regardless of it's bytecode or not), I think (2) would be better.

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
-
-			string foundPlaceholder(it, it + placeholderSize);
-			if (librariesReplacements.count(foundPlaceholder))
+			string const& name = library.first;
+			// Bytecode we need is on the third line there's a blank line at the start
+			bytecode = sourceLines.at(bytecodeIndex);
+			size_t libraryStartOffset = bytecode.find("__$");
+
+			for (size_t i = 0; i < placeholderSize; i++)
+				bytecode[libraryStartOffset + i] = '0';
```
cameel: This code used to throw on invalid placeholders. Now it just looks for `__$` and assumes that anything it founds must be a valid placeholder. But it could be something else (`$__` missing) or it could be truncated and you'll go past the end of the `bytecode` array.

cameel: This also reminds me that we have no tests covering this code at all. This is because we can't use our usual [command-line tests](https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests) for this - linker is a bit unusual compared to other input modes in that it modifies files in place and never prints them to stdout.

Recently [test/solc/CommandLineInterface.cpp](https://github.com/ethereum/solidity/blob/develop/test/solc/CommandLineInterface.cpp) was added though so we could test it there. It would be something similar to [`cli_input` test](https://github.com/ethereum/solidity/blob/develop/test/solc/CommandLineInterface.cpp#L174-L219) but with different options and would need to create input files and then read output from created ones (we have helpers for most of this, see other tests in the file).

Or we could refactor `link()` into a static method (taking `libraries` and `sourceCodes` as arguments) and then test that method. This would make the test much simpler.

In any case, we need at least three tests: input with placeholders, input with no placeholders, input with broken placeholders.

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
-				copy(hexStr.begin(), hexStr.end(), it);
+				// Divide by 2 because it's going to be converted to bytes
+				linkReferences[libraryStartOffset / 2] = name;
 			}
-			else
-				serr() << "Reference \"" << foundPlaceholder << "\" in file \"" << src.first << "\" still unresolved." << endl;
-			it += placeholderSize;
 		}
+
+		evmasm::LinkerObject linkerObject= {
```
```suggestion
		evmasm::LinkerObject linkerObject = {
```

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
+			fromHex(bytecode),
+			linkReferences,
+			map<u256, std::pair<std::string, std::vector<size_t>>>(),
+			std::map<std::string, evmasm::LinkerObject::FunctionDebugData>()
```
cameel: I think you should be able to use empty initializer lists here and the compiler should infer the types:
```suggestion
			{},
			{}
```

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
+		};
+
+		linkerObject.link(librariesReplacements);
+		sourceLines.at(bytecodeIndex) = toHex(linkerObject.bytecode);
+
+		std::string resolvedBytecode = boost::algorithm::join(sourceLines, "\n");
```
```suggestion
		string resolvedBytecode = boost::algorithm::join(sourceLines, "\n");
```

solc/CommandLineInterface.cpp
```diff
@@ -931,45 +932,61 @@ void CommandLineInterface::link()
...
+
+			if (libraryStartOffset != bytecode.length())
 			{
-				string hexStr(toHex(librariesReplacements.at(foundPlaceholder).asBytes()));
-				copy(hexStr.begin(), hexStr.end(), it);
+				// Divide by 2 because it's going to be converted to bytes
+				linkReferences[libraryStartOffset / 2] = name;
 			}
-			else
-				serr() << "Reference \"" << foundPlaceholder << "\" in file \"" << src.first << "\" still unresolved." << endl;
```
cameel: Why is this being removed?

solc/CommandLineInterface.cpp
cameel: I thought it was still relevant but looking closer at the PR now, it removes support for old-style placeholders and without it it's not necessary.

Are we fine with that though? Modern compiler versions do not produce them so I'd be fine with that but it's technically a breaking change.

cameel (changes requested):

bmoore01:
I added the recommended changes and I've added in some unit tests to cover this functionality too.

solc/CommandLineInterface.cpp
```diff
@@ -653,7 +654,8 @@ void CommandLineInterface::processInput()
 		assemble(m_options.assembly.inputLanguage, m_options.assembly.targetMachine);
 		break;
 	case InputMode::Linker:
-		link();
+		solAssert(m_options.input.mode == InputMode::Linker, "");
```
cameel: We have recently made the description optional in asserts:
```suggestion
		solAssert(m_options.input.mode == InputMode::Linker);
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
 		solThrow(CommandLineExecutionError, "LSP terminated abnormally.");
 }
 
-void CommandLineInterface::link()
+static bool isBytecodeChar(char c)
 {
-	solAssert(m_options.input.mode == InputMode::Linker, "");
+	return isxdigit(c) || c == '_' || c == '$';
```
cameel: Just FYI, we have recently added wrappers for functions like `isdigit()` to ensure we always use the C locale (#12740). I see from the docs that `isxdigit()` is not affected by locale though so I guess using it directly is fine.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+				hasError = true;
+			}
 		}
+
+		evmasm::LinkerObject linkerObject = {
+			fromHex(bytecode),
+			linkReferences,
+			{},
+			{}
+		};
```
```suggestion
		evmasm::LinkerObject linkerObject = {fromHex(bytecode), linkReferences, {}, {}};
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
-			string foundPlaceholder(it, it + placeholderSize);
-			if (librariesReplacements.count(foundPlaceholder))
+			if (!startMarkerFound && endMarkerFound)
 			{
-				string hexStr(toHex(librariesReplacements.at(foundPlaceholder).asBytes()));
-				copy(hexStr.begin(), hexStr.end(), it);
+				solThrow(
+					CommandLineExecutionError,
+					"Error in binary object file " + src.first + " unbounded placeholder without start point __$ at " + to_string(libraryEndOffset) + "\n"
+							);
```
```suggestion
				solThrow(
					CommandLineExecutionError,
					"Error in binary object file " + src.first + " unbounded placeholder without start point __$ at " + to_string(libraryEndOffset) + "\n"
				);
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+			{
+				if ((libraryEndOffset - libraryStartOffset) != placeholderSize - 3)
+				{
+					solThrow(
+						CommandLineExecutionError,
+						"Error in binary object file " + src.first + " truncated placeholder found at " + to_string(libraryStartOffset) +
+							" placeholder addresses should be " + to_string(placeholderSize) + " characters long (including address markers __$ and $__)\n"
+					);
+				}
+			}
```
cameel: You don't need braces if there's just one instruction inside `if` or `for`. The code is more concise without them:
```suggestion
			if (startMarkerFound && endMarkerFound)
				if ((libraryEndOffset - libraryStartOffset) != placeholderSize - 3)
					solThrow(
						CommandLineExecutionError,
						"Error in binary object file " + src.first + " truncated placeholder found at " + to_string(libraryStartOffset) +
							" placeholder addresses should be " + to_string(placeholderSize) + " characters long (including address markers __$ and $__)\n"
					);
```
There are also a few other places like this in the PR.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+			for (size_t i = 0; i < placeholderSize; i++)
+			{
+				foundPlaceholder += bytecode[libraryStartOffset + i];
+				bytecode[libraryStartOffset + i] = '0';
+			}
+
+			if (libraryStartOffset != bytecode.length())
+			{
+				// Divide by 2 because it's going to be converted to bytes
+				linkReferences[libraryStartOffset / 2] = name;
```
cameel: `linkReferences` is not used before this place. I'd put it's declaration above this `if`. Similarly for a few other declarations I think it's generally more readable to have them close to the place where they're used.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
-			if (it == end) break;
-			if (
-				end - it < placeholderSize ||
-				*(it + 1) != '_' ||
-				*(it + placeholderSize - 2) != '_' ||
-				*(it + placeholderSize - 1) != '_'
-			)
+			// If all characters are bytecode characters
+			if (sourceLines[i].size() > 0
+				&& all_of(sourceLines[i].begin(), sourceLines[i].end(), isBytecodeChar))
```
```suggestion
			if (
				sourceLines[i].size() > 0 &&
				all_of(sourceLines[i].begin(), sourceLines[i].end(), isBytecodeChar)
			)
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
 	{
-		auto end = src.second.end();
-		for (auto it = src.second.begin(); it != end;)
+		// Map from how the libraries will be named inside the bytecode to their addresses.
+		map<size_t, string> linkReferences;
+		string bytecode;
+		vector<string> sourceLines;
+		boost::split(sourceLines, src.second, boost::is_any_of("\n"));
+
+		size_t bytecodeIndex;
```
cameel: Please always initialize variables of primitive types with some sane/neutral value. Even if the initial value is not meant to be used in the end, it's better to avoid having random junk in them. In case there's a bug and the value gets used after all, the behavior will be more deterministic.
```suggestion
		size_t bytecodeIndex = sourceLines.size();
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+		bool bytecodeFound = false;
+		for (size_t i = 0; i < sourceLines.size(); ++i)
```
cameel: Why not use the index variable as the loop counter?
```suggestion
		bool bytecodeFound = false;
		size_t lineIndex = 0;
		for (lineIndex; i < sourceLines.size(); ++lineIndex)
```

Alternatively, you could use range loop:
```suggestion
		bool bytecodeFound = false;
		size_t lineIndex = 0;
		for (string const& line: sourceLines)
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+				&& all_of(sourceLines[i].begin(), sourceLines[i].end(), isBytecodeChar))
+			{
+				bytecodeFound = true;
+				bytecodeIndex = i;
+				break;
+			}
+		}
+
+		if (!bytecodeFound)
+			continue;
```
cameel: You don't really need this variable:
```suggestion
		if (bytecodeIndex >= sourceLines.size())
			continue;
```

solc/CommandLineInterface.cpp
cameel: I see that the only reason you need the index is to the get the line from the array. Why not use `string_view` instead and just store a reference to the line itself?

Also you could replace the whole loop with a single call to `std::find_if()`. Then access the string via the returned iterator and assign it to a `string_view`.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
 
-			string foundPlaceholder(it, it + placeholderSize);
-			if (librariesReplacements.count(foundPlaceholder))
+			if (!startMarkerFound && endMarkerFound)
 			{
-				string hexStr(toHex(librariesReplacements.at(foundPlaceholder).asBytes()));
-				copy(hexStr.begin(), hexStr.end(), it);
+				solThrow(
+					CommandLineExecutionError,
+					"Error in binary object file " + src.first + " unbounded placeholder without start point __$ at " + to_string(libraryEndOffset) + "\n"
```
```suggestion
					"Error in binary object file " + src.first + " unbounded placeholder without start marker __$ at " + to_string(libraryEndOffset) + "\n"
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+			}
+
+			if (startMarkerFound && endMarkerFound)
+			{
+				if ((libraryEndOffset - libraryStartOffset) != placeholderSize - 3)
+				{
+					solThrow(
+						CommandLineExecutionError,
+						"Error in binary object file " + src.first + " truncated placeholder found at " + to_string(libraryStartOffset) +
+							" placeholder addresses should be " + to_string(placeholderSize) + " characters long (including address markers __$ and $__)\n"
```
```suggestion
						"Error in binary object file \"" + src.first + "\": placeholder of invalid length found in column " + to_string(libraryStartOffset) + ".\n" + 
							"Placeholders must be exactly " + to_string(placeholderSize) + " characters long, including address markers __$ and $__.\n"
```

You could also include line number in the message.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+				}
+			}
+
+			string foundPlaceholder;
+
+			for (size_t i = 0; i < placeholderSize; i++)
+			{
+				foundPlaceholder += bytecode[libraryStartOffset + i];
+				bytecode[libraryStartOffset + i] = '0';
+			}
```
cameel: You could do this with `substr()` and `replace()`:
```suggestion
			string foundPlaceholder = bytecode.substr(libraryStartOffset, placeholderSize);
			bytecode.replace(libraryStartOffset, placeholderSize, "0000000000000000000000000000000000000000");
```
Not tested but I hope you get the idea.

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+		if (!bytecodeFound)
+			continue;
+
+		for (auto const& library: libraries)
+		{
+			string const& name = library.first;
+			// Bytecode we need is on the third line there's a blank line at the start
+			bytecode = sourceLines.at(bytecodeIndex);
+			size_t libraryStartOffset = bytecode.find("__$");
+			size_t libraryEndOffset = bytecode.find("$__");
```
```suggestion
			size_t startMakerOffset = bytecode.find("__$");
			size_t endMarkerOffset = bytecode.find("$__");
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
 }
 
-void CommandLineInterface::link()
+static bool isBytecodeChar(char c)
 {
-	solAssert(m_options.input.mode == InputMode::Linker, "");
+	return isxdigit(c) || c == '_' || c == '$';
+}
 
+void CommandLineInterface::link(frontend::FileReader& fileReader, map<string, h160> const& libraries, bool& hasError, ostream& serr)
```
```suggestion
void CommandLineInterface::link(frontend::FileReader& fileReader, map<string, h160> const& libraries, bool& hasOutput, ostream& serr)
```

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+			{
+				// Divide by 2 because it's going to be converted to bytes
+				linkReferences[libraryStartOffset / 2] = name;
 			}
 			else
-				serr() << "Reference \"" << foundPlaceholder << "\" in file \"" << src.first << "\" still unresolved." << endl;
-			it += placeholderSize;
+			{
+				serr << "Reference \"" << foundPlaceholder << "\" in file \"" << src.first << "\" still unresolved."<< endl;
+				hasError = true;
```
cameel: I'd prefer to avoid manually setting `m_hasOutput`. Can you instead add a new overload of `serr()` take accepts the stream and the flag variable as parameters?

solc/CommandLineInterface.cpp
```diff
@@ -916,67 +918,134 @@ void CommandLineInterface::serveLSP()
...
+			fromHex(bytecode),
+			linkReferences,
+			{},
+			{}
+		};
+
+		linkerObject.link(librariesReplacements);
+		sourceLines.at(bytecodeIndex) = toHex(linkerObject.bytecode);
+
+		string resolvedBytecode = boost::algorithm::join(sourceLines, "\n");
```
cameel, line 1041: It's not just bytecode. It's the whole file reassembled.
```suggestion
		string resolvedSrc = boost::algorithm::join(sourceLines, "\n");
```

cameel, line 1041: I'm not sure if it's a good idea to rewrite the newlines. At least it changes the current behavior because the previous implementation would preserve them.

solc/CommandLineInterface.cpp
cameel: Maybe we should allow whitespace as well. At least at the start/end of the line. The previous implementation would handle it just fine.

test/solc/CommandLineInterface.cpp
```diff
@@ -278,6 +278,113 @@ BOOST_AUTO_TEST_CASE(cli_not_a_file)
...
+	string hashString = "1234567890123456789012345678901234567890";
+	map<string, util::FixedHash<20>> libraries = {
+		{"test/libsolidity/semanticTests/libraries/stub.sol:L", FixedHash<20>("0x" + hashString)}
+	};
+
+	bool hasError = false;
+
+	CommandLineInterface::link(fileReader, libraries, hasError, cerr);
+	BOOST_TEST(!hasError);
+	BOOST_TEST(fileReader.sourceUnits().find("bytecodeInput.bin")->second.find(hashString) != string::npos);
```
cameel: Why not just compare the result directly? Less chance of false positives by mistake and it's not that long anyway:

```suggestion
	map<string, string> expectedSources = {{
		"bytecodeInput.bin",
		"565b6040516100579190610158565b60405180910390f35b600073" + hashString + "63b3de648b8360405"
	}};
	BOOST_TEST(fileReader.sourceUnits() == expectedSources);
```

test/solc/CommandLineInterface.cpp
```diff
@@ -278,6 +278,113 @@ BOOST_AUTO_TEST_CASE(cli_not_a_file)
...
+		[&](auto const& _exception) { BOOST_TEST(_exception.what() == expectedMessage); return true; }
+	);
+}
+
+BOOST_AUTO_TEST_CASE(linker_unmatched_markers_no_start)
+{
+	frontend::FileReader fileReader({},{},{});
+	fileReader.setSourceUnits({
+								  {"bytecodeInput.bin", "565b6040516100579190610158565b60405180910390f35b600073$__b3157e226f2c4dddae135e6cab4ed4e74763b3de648b8360405"}
+							  });
```
cameel:
```suggestion
	fileReader.setSourceUnits({
		{"bytecodeInput.bin", "565b6040516100579190610158565b60405180910390f35b600073$__b3157e226f2c4dddae135e6cab4ed4e74763b3de648b8360405"}
	});
```

test/solc/CommandLineInterface.cpp
```diff
@@ -278,6 +278,113 @@ BOOST_AUTO_TEST_CASE(cli_not_a_file)
...
+
+	string hashString = "1234567890123456789012345678901234567890";
+	map<string, util::FixedHash<20>> libraries = {
+		{"test/libsolidity/semanticTests/libraries/stub.sol:L", FixedHash<20>("0x" + hashString)}
+	};
+
+	bool hasError = false;
+
+	string expectedMessage ="Error in binary object file bytecodeInput.bin unbounded placeholder without start point __$ at 54\n";
```
cameel:
```suggestion
	string expectedMessage = "Error in binary object file bytecodeInput.bin unbounded placeholder without start point __$ at 54\n";
```

test/solc/CommandLineInterface.cpp
```diff
@@ -278,6 +278,113 @@ BOOST_AUTO_TEST_CASE(cli_not_a_file)
...
+	CommandLineInterface::link(fileReader, libraries, hasError, cerr);
+	BOOST_TEST(!hasError);
+	BOOST_TEST(fileReader.sourceUnits().find("bytecodeInput.bin")->second.find(hashString) != string::npos);
+}
+
+BOOST_AUTO_TEST_CASE(linker_unmatched_markers_no_end)
+{
+	frontend::FileReader fileReader({},{},{});
+	fileReader.setSourceUnits({
+		{"bytecodeInput.bin", "565b6040516100579190610158565b60405180910390f35b600073__$b3157e226f2c4dddae135e6cab4ed4e74763b3de648b8360405"}
```
cameel, line 408: Can you add a few more test cases? I think that the code might not work correctly with some of them:

- Old-style placeholder:
    ```
    565b604051610057919__lib.sol:L_____________________________48b8360405
    ```
- Two starts, single end:
    ```
    565b6040516100579190__$b36__$b3157e226f2c4dddae135e6cab4ed4e747$__63b3de648b8360405
    ```
- Placeholder in a placeholder:
    ```
    565b6040516100579190__$b36__$b3157e226f2c4dddae135e6cab4ed4e747$__63b3de$__648b8360405
    ```
- Double-sided marker:
    ```
    565b604051610057919__$__157e226f2c4dddae135e6cab4ed4e7__$__48b8360405
    ```
    I think this was valid in the old implementation and would be treated as an old-style placeholder.
- Marker without `$`:
    ```
    565b604051610057919___b3157e226f2c4dddae135e6cab4ed4e747___48b8360405
    ```
    Also technically a valid old-style placeholder.

solc/CommandLineInterface.cpp
cameel: I think that the detection will unfortunately need to be a bit more complex to handle old-style placeholders since they could contain arbitrary chars. So it would be something like this on the outside of `__$` and `$__` and anything in between.

Maybe let's try this:
- The line has to contain `__`.
- All chars before the first and after the last `__` must be valid hex chars.
- There can be anything between the first `__` and last `__`.

cameel: Unless we can make our lives easier and actually drop support for these old-style placeholders.
@chriseth Would that be acceptable without a breaking release?

cameel (changes requested):
Here's more feedback. Sorry for a ton of comments but most of these are just small style corrections, improvement suggestions and error message tweaks. Here's a summary of the more important ones:
- More test cases needed.
- This removes support for old-style placeholders.
- This changes how newlines are handled (they're now always rewritten to `\n`).
- Maybe we should allow leading/trailing whitespace on bytecode lines as the old implementation did?

cameel:
By the way, please don't get discouraged by the comments, I think this is overall going in the right direction :) Especially I like that we now will have this covered with some tests!

bmoore01:
@cameel  No probblem, i understand, i'll get these changes done, always worth doing right!

test/solc/CommandLineInterface.cpp
bmoore01: Should we be supporting the placeholder in a placeholder or double-sided marker cases?

cameel: The double-sized one could be a valid old-style placeholder so yes, unless we drop support for old-style placeholders. The `__` on the inside basically counts as part of the library identifier in that case.
But I'm mentioning that case mostly because it's a corner case that looks like it could easily confuse regex based detection.

As for nested ones, if the outer `__` pair has the correct distance between markers, then again, it could be a valid old-style placeholder. But if it's the inner one that has the right length then the outer one should be treated as unmatched markers and result in an error.

leonardoalt:
Seems to be waiting for review @cameel

bmoore01:
@leonardoalt yeah, I'm happy with it so just waiting for review now.

bmoore01:
@cameel Are you happy for me to merge this one now? It's been open for a while just wanted to see if you wanted any more changes.

cameel:
@bmoo0 Hello, sorry for not going back to this one sooner but there's just been too much happening. We still have a huge heap of PRs to get through and on top of that right now most of the team (including me) is still at Devconnect.

So for now, just a few short comments. I discussed this with @chriseth earlier and the problem is really that this refactor complicates the mechanism more than we expected it would. There were two major conclusions from that discussion:
- Currently there are very few restrictions on how the `.bin` object would look like. Now with this PR we have to detect the bytecode so I had to push for move validations and this unfortunately makes things complicated. We also have to support old format for placeholders.

    One idea to make things simpler is basically to make it a breaking change scheduled for 0.9.0. This way we could drop all the legacy stuff and inconvenient restrictions. To do that you would have to change the base branch of the PR to a branch called `breaking`.
- A very valuable part of the PR as it is currently are the tests. If you extracted them into a separate PR we'd gladly accept them - or at least the part that would still work on top of `develop` branch.

What do you think? If you agree with these changes I could try to go over the PR again and give some hints on what specifically should be carried over and what would not be necessary when we make the change breaking.

bmoore01:
Sounds good to me. I can actually break it out into two PRs if that's better. One containing the tests branched from the develop branch and the other adding the breaking change which can stem from the breaking branch, if that makes things simpler.

cameel:
Great! Tests should go into `develop` since they do not have to wait for a breaking release.

Marenz:
It seems the PR needs a rebase to resolve the current conflicts.
I see the plan was to split out the tests and make this one breaking.. any updates on that? Can we help or should we take over parts possible?

solc/CommandLineInterface.cpp
```diff
@@ -918,12 +917,64 @@ void CommandLineInterface::serveLSP()
 		solThrow(CommandLineExecutionError, "LSP terminated abnormally.");
 }
 
-static bool isBytecodeChar(char c)
```
axic: I think keeping this helper may still be nice.

solc/CommandLineInterface.cpp
```diff
@@ -918,12 +917,64 @@ void CommandLineInterface::serveLSP()
...
+static bool isBytecodeString(string s)
 {
-	return isxdigit(c) || c == '_' || c == '$';
```
axic: Unrelated to this PR, but:
> isxdigit(std::locale) | checks if a character is classified as a hexadecimal digit by a locale(function template)

Perhaps we'd be better off with just a manual check and could make it `constexpr` that case.

leonardoalt:
Since this PR seems stale is going to be split up anyway, I'm closing for now. The code is still available in the branch.

bmoore01:
Hey sorry I was away for awhile count' work on this. I think the tests can still be merged if you think they're useful I just accidentally pulled too many changes into my branch and need to look up how to only merge my new tests.

PR #12757, opened 2022-03-09
Take termination into account in data flow analyzer

### LINKED ISSUE #12755: Take termination into account in data flow analyzer
```
{
    let _1 := memoryguard(0x80)
    mstore(64, _1)
    let _2 := 0
    switch shr(224, calldataload(_2))
    case 0x06661abd {
        mstore(_1, sload(_2))
        return(_1, 32)
    }
    case 0xc2985578 {
        return(mload(64), _2)
    }
}
// ----
// step: fullSuite
//
// {
//     {
//         let _1 := memoryguard(0x80)
//         mstore(64, _1)
//         switch shr(224, calldataload(0))
//         case 0x06661abd {
//             mstore(_1, sload(0))
//             return(_1, 32)
//         }
//         case 0xc2985578 { return(mload(64), 0) }
//     }
// }
```

The mload at the end should be replaced by `_1`, but somehow it is not. The `mstore` in the first swich case seems to destroy it.

Similar example:
```
let x := 1
if calldataload(0) {
  x := 8
  revert(0, 0)
}
mstore(0, x)
```
The DataFlowAnalyzer will not be aware of the value of x in the last statement. It should take into account if a branching control-flow terminates or not before it clears knowledge of local variables and storage / memory.

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12755

### DIFF
--- a/libyul/optimiser/DataFlowAnalyzer.cpp
+++ b/libyul/optimiser/DataFlowAnalyzer.cpp
@@ -136,16 +136,18 @@ void DataFlowAnalyzer::operator()(Switch& _switch)
 		unordered_map<YulString, YulString> storage = m_storage;
 		unordered_map<YulString, YulString> memory = m_memory;
 		(*this)(_case.body);
+		//TODO throw away knowledge if it is terminating
 		joinKnowledge(storage, memory);
 
 		set<YulString> variables = assignedVariableNames(_case.body);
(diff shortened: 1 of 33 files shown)


chriseth:
Shouldn't we do the same thing for the unused store base?

PR #12758, opened 2022-03-09
Provide generic access to type definition for user-defined types.

chriseth:
Needed for #12430 and "using for".


chriseth:
For the LSP it might make sense to also include contracts, isn't it, @christianparpart ?
If we do this, it needs a special case for "using for", but I guess it's fine.

chriseth:
It seems LSP does not need this, so we can merge it unmodified.

ekpyron (approved):

Issue #12668, opened 2022-02-14
Run semantics tests using the proper viaIR pipeline and make optimizer requirements explicit in the tests.


ekpyron:
Basically we should do something along the lines of what https://github.com/ethereum/solidity/pull/11949 tried.

ekpyron:
Note that since the old PR, we're enabling minimal Yul optimizations by default, so this will change the setup. If we're lucky, none of the tests will require more optimization than the default minimal ones and we can get rid of the logic to re-run with the optimizer entirely (i.e. just run with default settings without ``--optimize`` and with optimizer when running with ``--optimize``).
Conversely, the fact that we *do* have minimal optimizations by default now and the semantics tests use different settings makes addressing this more pressing.

There may be a complication with preserving the "homestead" test runs, since they are less resilient against stack issues. It's ugly, but if needed, we can preserve the "run twice" logic for those (no need to make that explicit in the tests, if it *only* applies to homestead), until we finally drop homestead support.

PR #12760, opened 2022-03-09
Mark viaIR code generation as non-experimental.

ekpyron:
~~Depends on https://github.com/ethereum/solidity/pull/12748~~
Did I miss anything?
Somebody should probably really do https://github.com/ethereum/solidity/issues/12668 before we merge this...


Changelog.md
```diff
@@ -5,6 +5,8 @@ Language Features:
 
 
 Compiler Features:
+ * Commandline Interface: Allow the use of ``--via-ir`` in place of ``--experimental-via-ir``.
+ * Compilation via Yul IR is no longer experimental.
```
```suggestion
 * Compilation via Yul IR is no longer marked as experimental.
```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1677,7 +1677,7 @@ bytes CompilerStack::createCBORMetadata(Contract const& _contract, bool _forIR)
 	else
 		solAssert(m_metadataHash == MetadataHash::None, "Invalid metadata hash");
 
-	if (experimentalMode || _forIR)
+	if (experimentalMode)
```
chriseth, line 1680: I think this requires updating the gas cost expectations :)

solc/CommandLineParser.cpp
```diff
@@ -578,7 +579,11 @@ General Information)").c_str(),
 		)
 		(
 			g_strExperimentalViaIR.c_str(),
-			"Turn on experimental compilation mode via the IR (EXPERIMENTAL)."
+			"Deprecated synonym of --via-ir."
+		)
+		(
+			g_strViaIR.c_str(),
+			"Turn on experimental compilation mode via the IR."
```
chriseth:
```suggestion
			"Turn on compilation mode via the IR."
```

chriseth:
Do we have a test that shows that the old flag still works?

solc/CommandLineParser.cpp
ekpyron: It was actually in there twice, yeah :-).

ekpyron:
> Do we have a test that shows that the old flag still works?

No, but I can add one back. [EDIT: done - changed the command line test script to compare against both flags; EDIT2: removed that again now and added a small test for the command line parser instead]

libsolidity/interface/CompilerStack.cpp
ekpyron: I think we're compensating for it in the test setup already.

test/cmdlineTests.sh
```diff
@@ -354,13 +354,20 @@ function test_via_ir_equivalence()
...
+    )
+    bin_output_experimental_via_ir=$(
         echo "$solidity_code" |
         msg_on_error --no-stderr "$SOLC" - --experimental-via-ir --bin "${optimizer_flags[@]}" |
         sed '/^======= <stdin>/d' |
         sed '/^Binary:$/d'
     )
 
     diff_values "$bin_output_two_stage" "$bin_output_via_ir" --ignore-space-change --ignore-blank-lines
+    diff_values "$bin_output_two_stage" "$bin_output_experimental_via_ir" --ignore-space-change --ignore-blank-lines
```
cameel: Not sure if it's worth keeping this. We're compiling some external contracts here so it's slow (this part of the cmdline tests adds 30-60 secs which is a significant part of the whole runtime). Maybe some lighter test would suffice?

test/solc/CommandLineParser.cpp
```diff
@@ -175,7 +175,7 @@ BOOST_AUTO_TEST_CASE(cli_mode_options)
 		expectedOptions.output.dir = "/tmp/out";
 		expectedOptions.output.overwriteFiles = true;
 		expectedOptions.output.evmVersion = EVMVersion::spuriousDragon();
-		expectedOptions.output.experimentalViaIR = true;
+		expectedOptions.output.viaIR = true;
```
cameel, line 179: There's `--experimental-via-ir` higher up, in the `vector<string> commandLine` initializer. I'd actually add `--via-ir` next to it, instead of replacing it there - both flags are still supported after all and this test is supposed to use all possible flags.

cameel, line 179: Also, I'd add a separate small test in this file that checks that both `--experiemental-via-ir` and `--ir` on their own set `expectedOptions.output.viaIR` to true. You can use this one as a base - just remove most of the flags and most of the expectations (they should stay at defaults without the flags).

cameel:
> Did I miss anything?
- `layout-in-storage.rst`, `optimizer.rst` and `ir-breaking-changes.rst` refer to the IR mode as experimental and sometimes mention the `--experimental-via-ir` flag. [@ekpyron: done I think]
- Some comments still call it experimental in `CompilerStack.h` [@ekpyron: done I think]
- We still have this warning in `StandardCompiler.cpp` and `CommandLineInterface.cpp`:
    ```
    Yul is still experimental. Please use the output with care.
    ```
    - There are also a few different places in tests that cut out this message from test output. We should remove that too.

ekpyron:
>     * We still have this warning in `StandardCompiler.cpp` and `CommandLineInterface.cpp`:
>       ```
>       Yul is still experimental. Please use the output with care.
>       ```

That's compilation of pure Yul, not solidity code generation via-IR, though, so its not exactly the same thing. Marking solidity compilation via IR as non-experimental covers only a subset of all possible Yul programs... that being said it'd be a bit weird to keep referring to yul as experimental... I'll remove this in a separate commit.

cameel:
Actually, did we still consider Yul experimental until now or is it just leftover?

ekpyron:
Is Yul really stable :-)? Especially the object format and stuff like the "optimizer hint builtins" I can still see changing... that doesn't affect the stability of compiling *via Yul*, but it does affect stuff written in the language Yul itself...

cameel:
Interesting. I always considered Yul itself less experimental than the codegen :) We even had normal changelog for it and not for the codegen.

chriseth:
I would say it's just a leftover. We have been using yul in production in the abi encoder for years now.

chriseth (approved):

PR #12762, opened 2022-03-09
Optimize ``return(x,0) -> pop(x) return(0,0)`` (and also for ``revert``).

ekpyron:
Nothing fancy, but a few gas here and there...
Kind of depends on https://github.com/ethereum/solidity/issues/12913, but wouldn't be fatal to merge without.


libyul/optimiser/StructuralSimplifier.cpp
```diff
@@ -87,6 +88,44 @@ void StructuralSimplifier::simplify(std::vector<yul::Statement>& _statements)
...
 				return {std::move(_forLoop.pre.statements)};
 			return {};
+		},
+		[&](ExpressionStatement& _stmt) -> OptionalStatements {
+			FunctionCall& functionCall = get<FunctionCall>(_stmt.expression);
+			if (dynamic_cast<EVMDialect const*>(&m_dialect))
+				if (
+					functionCall.functionName.name == "revert"_yulstring ||
+					functionCall.functionName.name == "return"_yulstring
+				)
```
ekpyron: Maybe we can do it for all memory reading opcodes based on the start/length fields in the side-effects...

chriseth: Yep, I think that's better.

chriseth:
Is there an optimizer step that might be more fitting? Structural sounds more like it operates on the structure of the code...

ekpyron:
> Is there an optimizer step that might be more fitting? Structural sounds more like it operates on the structure of the code...

I didn't see anything obvious... the expression simplifier only acts on expressions and can't easily add the ``pop``... and the structural simplifier already has the literal rematerializer as soft-requirement, s.t. we likely get the ``0`` as literal...
We could extend the expression simplifier to expression statements instead I guess...

test/libsolidity/semanticTests/smoke/constructor.sol
```diff
@@ -14,7 +14,7 @@ contract C {
 // compileViaYul: also
 // ----
 // constructor(), 2 wei: 3 ->
-// gas irOptimized: 111699
+// gas irOptimized: 112569
```
ekpyron: This one I don't really understand - may be worth a closer look :-).

chriseth:
> I didn't see anything obvious... the expression simplifier only acts on expressions and can't easily add the `pop`... and the structural simplifier already has the literal rematerializer as soft-requirement, s.t. we likely get the `0` as literal... We could extend the expression simplifier to expression statements instead I guess...

Or start a new one?

ekpyron:
Actually the ``ExpressionSimplifier`` is suitable after all - we already restrict simplification to variables to avoid side-effect issues anyways, so we can also just do that for this (and then don't need to insert a ``pop`` at all).

test/libsolidity/semanticTests/smoke/constructor.sol
```diff
@@ -14,7 +14,7 @@ contract C {
 // compileViaYul: also
 // ----
 // constructor(), 2 wei: 3 ->
-// gas irOptimized: 107627
+// gas irOptimized: 108707
```
ekpyron: This warrants a closer look. There's also occasionally other regressions by <10 gas, but this is the single significant ones.

test/cmdlineTests/debug_info_in_yul_and_evm_asm_print_all/output
```diff
@@ -194,7 +194,7 @@ object "C_6" {
                     {
                         if callvalue() { revert(_1, _1) }
                         if slt(add(calldatasize(), not(3)), _1) { revert(_1, _1) }
-                        return(memoryguard(0x80), _1)
+                        return(_1, _1)
```
ekpyron, line 197: These were the main target and it seems to work nicely for them.

ekpyron: Although: how does the ``memoryguard`` vanish completely here? Didn't we make it non-removable?
Maybe we didn't, arguing exactly that it will always still be used somewhere in the end...
Hm... that's a problem - it vanishing entirely may prevent stack-to-memory...

ekpyron: Yeah, IIRC, the idea was that if we make ``memoryguard`` movable, not non-removable, we may accumulate a lot of weird ``pop(memoryguard(...))``s over time... and at the time, the value actually being written to memory guaranteed that at least one would remain - but that's already no longer true since after the unused store eliminator...

One way to fix this would probably be a special kind of side-effect "can not be *entirely* removed"... or just generally a special case in the ``UnusedPruner``, i.e. have it keep track of how many ``memoryguard``s there are and only consider one removable, if it is not the last one, while otherwise keeping a ``pop(memoryguard(...))`` around...

But none of that is really nice :-)...

ekpyron: Hm... one could argue:
If code contains a ``memoryguard`` before optimization, then if all ``memoryguard``s to vanish during optimization, that has to mean that all dynamic memory uses could be eliminated... and we can just add the original ``memoryguard`` back... but that's also a bit weird...

axic:
Weird that this change mostly increases gas -- or is that only for the isolated test cases and when it is used in the complete optimisation suite it helps?

ekpyron:
> Weird that this change mostly increases gas -- or is that only for the isolated test cases and when it is used in the complete optimisation suite it helps?

This is the gas diffs on our semantics test suite
<details><summary>Click for a table of gas differences</summary>

| File name                                                   |   IR-optimized (%) |   Legacy-Optimized (%) |   Legacy (%) |
|-------------------------------------------------------------|--------------------|------------------------|--------------|
| events/event_indexed_string.sol                             |       -0.000899607 |                      0 |            0 |
| events/event_dynamic_nested_array_storage_v2.sol            |       -0.00162047  |                      0 |            0 |
| events/event_dynamic_array_storage_v2.sol                   |       -0.00264278  |                      0 |            0 |
| events/event_emit_from_other_contract.sol                   |       -0.64311     |                      0 |            0 |
| events/event_dynamic_array_storage.sol                      |       -0.00264278  |                      0 |            0 |
| functionCall/gas_and_value_basic.sol                        |       -0.164804    |                      0 |            0 |
| functionCall/gas_and_value_brace_syntax.sol                 |       -0.164804    |                      0 |            0 |
| externalContracts/base64.sol                                |        0.000561615 |                      0 |            0 |
| externalContracts/FixedFeeRegistrar.sol                     |        1.04122     |                      0 |            0 |
| abiEncoderV2/abi_encode_calldata_slice.sol                  |       -0.00407541  |                      0 |            0 |
| userDefinedValueType/calldata.sol                           |       -0.0139169   |                      0 |            0 |
| abiEncoderV1/abi_encode_calldata_slice.sol                  |       -0.00407541  |                      0 |            0 |
| viaYul/array_storage_index_access.sol                       |       -7.3568e-05  |                      0 |            0 |
| viaYul/array_storage_push_empty.sol                         |       -0.000546533 |                      0 |            0 |
| viaYul/copy_struct_invalid_ir_bug.sol                       |        0.0176993   |                      0 |            0 |
| viaYul/array_storage_index_zeroed_test.sol                  |       -0.000181742 |                      0 |            0 |
| array/dynamic_array_cleanup.sol                             |       -0.000577349 |                      0 |            0 |
| array/dynamic_arrays_in_storage.sol                         |       -0.00179453  |                      0 |            0 |
| array/fixed_array_cleanup.sol                               |       -0.00751977  |                      0 |            0 |
| array/push/nested_bytes_push.sol                            |       -0.00167436  |                      0 |            0 |
| array/push/push_no_args_2d.sol                              |       -0.00524586  |                      0 |            0 |
| array/push/push_no_args_bytes.sol                           |       -0.0016341   |                      0 |            0 |
| array/copying/array_copy_storage_storage_dyn_dyn.sol        |       -0.00538725  |                      0 |            0 |
| array/pop/byte_array_pop_long_storage_empty_garbage_ref.sol |       -0.00210321  |                      0 |            0 |
| structs/struct_delete_storage_with_array.sol                |       -0.00246589  |                      0 |            0 |
| structs/structs.sol                                         |        0.00223159  |                      0 |            0 |
| various/contract_binary_dependencies.sol                    |       -0.641184    |                      0 |            0 |
| various/swap_in_storage_overwrite.sol                       |       -0.00273488  |                      0 |            0 |
</details>

not sure that qualifies as "mostly increases gas" :-). But I'm just looking at the ``FixedFeeRegistrar.sol`` case... what's happening there is that due to the change, two tags in the final assembly end up slightly different, while they were the same before (allowing deduplication) causing a slightly larger code size... (We should also really distinguish code size from runtime gas cost in our tests...)

So, the gas differences are little, but still usually favorable, I'd say, apart from corner cases like that.

Unfortunately, there's some issue with some external test apparently, so so far I'm not getting gas diffs on the external projects, that'd be more interesting.

ekpyron:
### `ir-optimize-evm+yul`
|           bleeps | **`-0.63% ✅`** |           `0%` |          `-0%` |
|            brink | **`-0.02% ✅`** |                |                |
|        chainlink |  **`-0.2% ✅`** | **`-0.23% ✅`** | **`-0.03% ✅`** |
|        elementfi |          `+0%` |                |                |
|              ens |           `0%` | **`-0.19% ✅`** |          `+0%` |
|            euler |  **`-0.1% ✅`** |  **`-0.1% ✅`** | **`+0.06% ❌`** |
|           gnosis | **`-0.05% ✅`** |          `-0%` |          `+0%` |
|              gp2 | **`+0.09% ❌`** | **`+0.04% ❌`** | **`+0.01% ❌`** |
|  perpetual-pools | **`-0.06% ✅`** | **`-0.06% ✅`** | **`+0.01% ❌`** |
|    pool-together | **`-0.21% ✅`** | **`-0.23% ✅`** |          `-0%` |
|          trident | **`-0.11% ✅`** | **`-0.16% ✅`** | **`-0.02% ✅`** |
|          uniswap | **`-0.12% ✅`** | **`+0.06% ❌`** | **`-0.04% ✅`** |
| yield_liquidator | **`-0.24% ✅`** | **`-0.24% ✅`** | **`-0.01% ✅`** |

Yeah... I'd call that "mixed, slightly favorable" or so :-).

test/externalTests/chainlink.sh
```diff
@@ -92,6 +92,10 @@ function chainlink_test
     sed -i "s|\(it\)\(('cannot remove a consumer from a nonexistent subscription'\)|\1.skip\2|g" test/v0.8/dev/VRFCoordinatorV2Mock.test.ts
     sed -i "s|\(it\)\(('cannot remove a consumer after it is already removed'\)|\1.skip\2|g" test/v0.8/dev/VRFCoordinatorV2Mock.test.ts
     sed -i "s|\(it\)\(('fails to fulfill without being a valid consumer'\)|\1.skip\2|g" test/v0.8/dev/VRFCoordinatorV2Mock.test.ts
+    # TODO: check why these two are needed due to this PR.
+    sed -i "s|\(it\)\(('cannot fund a nonexistent subscription'\)|\1.skip\2|g" test/v0.8/dev/VRFCoordinatorV2Mock.test.ts
+    sed -i "s|\(it\)\(('can cancel a subscription'\)|\1.skip\2|g" test/v0.8/dev/VRFCoordinatorV2Mock.test.ts
+
```
ekpyron, line 98: Most likely just another heuristics failure, but has to be investigated before merging this.

Marenz:
I rebased this PR

Marenz:
I also re-ran the tests and got the same problems, so the rebase didn't change anything in that regard

test/externalTests/chainlink.sh
cameel: @matheusaaguiar I think this was the failed run: [1093685](https://app.circleci.com/pipelines/github/ethereum/solidity/24838/workflows/51df4e72-9b91-47f0-9284-def047e9e9f6/jobs/1093685)
```
1352 passing (9m)
  65 pending
  2 failing

  1) VRFCoordinatorV2Mock
       #fundSubscription
         cannot fund a nonexistent subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  2) VRFCoordinatorV2Mock
       #cancelSubscription
         can cancel a subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
```

It looks similar to https://github.com/NomicFoundation/hardhat/issues/2453 which we just disable but the message is not the same. Rather than no revert reason or generic `Error`, here we have `TransactionExecutionError`.

There is a chance that it might be just newer `ethers.js` reporting the same problem in a different way. My suggestion would be to remove other patches for https://github.com/NomicFoundation/hardhat/issues/2453 from `chainlink.sh` and see if they have the same error message as this new problem. If so, I think we can safely assume they're the same thing. If they're different, we need to investigate deeper.

matheusaaguiar: @cameel this looks like the same issue we had [here](https://github.com/ethereum/solidity/issues/13264) also.

matheusaaguiar: > There is a chance that it might be just newer `ethers.js` reporting the same problem in a different way. My suggestion would be to remove other patches for [NomicFoundation/hardhat#2453](https://github.com/NomicFoundation/hardhat/issues/2453) from `chainlink.sh` and see if they have the same error message as this new problem. If so, I think we can safely assume they're the same thing. If they're different, we need to investigate deeper.

The error messages are still different. I will try to investigate.

matheusaaguiar: All `VRFCoordinatorV2Mock` tests fail with:
```
TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
```
Meanwhile, the other tests fail with:
```
Transaction reverted: function was called with incorrect parameters
```
In `develop` branch, the same tests also fail with the same messages, except for the two aforementioned tests in this PR, which are ok.
So, these new two failing tests produce the same error message as other `VRFCoordinatorV2Mock` failing tests.
I am not sure if it is just a heuristics problem or if it is something else we should look into further.

<details><summary> Failed tests error output </summary>

```
  1354 passing (8m)
  48 pending
  17 failing

  1) Flags
       #getFlag
         if the access control is turned on
           reverts:
     AssertionError: Expected transaction to be reverted with No access, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  2) Flags
       #getFlags
         respects the access controls of #getFlag:
     AssertionError: Expected transaction to be reverted with No access, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  3) FluxAggregator
       #setRequesterPermissions
         when called by the owner
           when permission is removed by the owner
             does not allow the specified address to start new rounds:
     AssertionError: Expected transaction to be reverted with not authorized requester, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  4) FluxAggregator
       #setRequesterPermissions
         when called by a stranger
           reverts:
     AssertionError: Expected transaction to be reverted with not authorized requester, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  5) AuthorizedForwarder
       #setAuthorizedSenders
         when called by the owner
           setting 0 authorized senders
             reverts with a minimum senders message:
     AssertionError: Expected transaction to be reverted with Must have at least 1 authorized sender, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  6) AuthorizedForwarder
       #setAuthorizedSenders
         when called by a non-owner
           cannot add an authorized node:
     AssertionError: Expected transaction to be reverted with Cannot set authorized senders, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  7) ArbitrumSequencerUptimeFeed
       Protect reads on AggregatorV2V3Interface functions
         should disallow reads on AggregatorV2V3Interface functions when consuming contract is not whitelisted:
     AssertionError: Expected transaction to be reverted with No access, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  8) CrossDomainOwnable
       #transferL1Ownership
         should not be callable by non-owners:
     AssertionError: Expected transaction to be reverted with Only callable by L1 owner, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  9) CrossDomainOwnable
       #acceptL1Ownership
         should not be callable by non pending-owners:
     AssertionError: Expected transaction to be reverted with Only callable by proposed L1 owner, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  10) VRFCoordinatorV2Mock
       #addConsumer
         cannot add a consumer to a nonexistent subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  11) VRFCoordinatorV2Mock
       #removeConsumer
         cannot remove a consumer from a nonexistent subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  12) VRFCoordinatorV2Mock
       #removeConsumer
         cannot remove a consumer after it is already removed:
     AssertionError: Expected transaction to be reverted with InvalidConsumer, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  13) VRFCoordinatorV2Mock
       #fundSubscription
         cannot fund a nonexistent subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  14) VRFCoordinatorV2Mock
       #cancelSubscription
         can cancel a subscription:
     AssertionError: Expected transaction to be reverted with InvalidSubscription, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  15) VRFCoordinatorV2Mock
       #fulfillRandomWords
         fails to fulfill without being a valid consumer:
     AssertionError: Expected transaction to be reverted with InvalidConsumer, but other exception was thrown: TransactionExecutionError: VM Exception while processing transaction: revert with unrecognized return data or custom error
  

  16) Flags
       #getFlag
         if the access control is turned on
           reverts:
     AssertionError: Expected transaction to be reverted with No access, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
  

  17) Flags
       #getFlags
         respects the access controls of #getFlag:
     AssertionError: Expected transaction to be reverted with No access, but other exception was thrown: Error: Transaction reverted: function was called with incorrect parameters
```

</details>

ekpyron:
I'd say it's fine to merge given https://github.com/ethereum/solidity/pull/12762#discussion_r900162365

leonardoalt:
Did anyone ever approve this? We could rebase & merge if someone did

ekpyron:
I'll rebase it in any case. But yeah, it's basically stuck on missing review.

ekpyron:
Rebased... and well, it *has* been marked ``priority review`` for over a month...

matheusaaguiar (approved):
I don't see anything that could be improved. 
Regarding the failing tests in https://github.com/ethereum/solidity/pull/12762#discussion_r900162365 , I confirmed again that they have the same error msg produced by previous disabled tests due to https://github.com/NomicFoundation/hardhat/issues/2453, so it should be fine.

PR #12763, opened 2022-03-09
doc: improve explanation of interfaces

k4nedito:
Something that tricked me when I was learning solidity back in the days. I thought that the interface can only be constructed from external functions. When in reality, contract implementation can be public, but it MUST be external in the interface. Therefore, here is the small change in docs that might help someone.


chriseth (commented):
Nice, thanks!

PR #12765, opened 2022-03-09
Increase Hardhat timeout in Euler ext test

cameel:
Euler has prepared a branch that increases the timeout to 100 sec (https://github.com/euler-xyz/euler-contracts/issues/97#issuecomment-1061000033) so we should start seeing fewer failures due to this.

~Unfortunately I don't see any way of properly evaluating it other than just merging it and seeing if we keep running into timeouts. They happen regularly but even if I just rerun tests a few times here, it won't prove that the issue is fixed.~

### DIFF
--- a/test/externalTests/common.sh
+++ b/test/externalTests/common.sh
@@ -233,12 +233,34 @@ function force_truffle_compiler_settings
 function name_hardhat_default_export
 {
     local config_file="$1"
-    local config_var_name="$2"
 
     local import="import {HardhatUserConfig} from 'hardhat/types';"
     local config="const config: HardhatUserConfig = {"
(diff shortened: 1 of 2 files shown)


cameel:
I decided to just patch the timeout on our side instead of switching to the branch. Looks like the cause of these timeouts might be that they're too short for unoptimized code but Euler always runs optimized, so I'm not certain that the branch will end up being merged. Also, the new helper function will be handy for bypassing timeouts in any other ext tests if they pop up.

ekpyron (approved):

ekpyron:
Looks like this didn't work:
https://app.circleci.com/pipelines/github/ethereum/solidity/22877/workflows/224204c3-78fc-457c-ae2a-b0a980550af6/jobs/1004400 still seems to run with a 20s timeout.

PR #12766, opened 2022-03-10
Remove unneeded header files.

chriseth:


ekpyron (approved):
Yep, I noticed that, too - should probably have gone for changing it right away :-).

PR #12769, opened 2022-03-10
Take termination into account in data flow analyzer

### LINKED ISSUE #12755: Take termination into account in data flow analyzer

chriseth:
Fixes https://github.com/ethereum/solidity/issues/12755


test/libyul/yulOptimizerTests/fullSuite/revert_in_switch.yul
```diff
@@ -0,0 +1,15 @@
...
+    }
+    case 1 {
+        if 0 { revert(0, 0) }
+    }
+    sstore(0, x)
+}
+// ----
+// step: fullSuite
+//
+// { { sstore(0, 0) } }
```
chriseth: That's what is currently generated. It is wrong.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -158,15 +159,14 @@ void DataFlowAnalyzer::operator()(Switch& _switch)
...
-			// postState is not always an ancestor of m_storage
 			joinKnowledge(postState);
-			clearValues(assignedVariableNames(_case.body));
+			assignedVariables += assignedVariableNames(_case.body);
 
 			postState = move(m_state);
 		}
 	}
 	m_state = move(postState);
+	clearValues(assignedVariables);
```
chriseth: joinKnowledge only acts on memory and storage, so the assignment `postState = move(m_state)` will just copy over the variable values as they were in the last case.

I'm not yet sure if we also need to clear after joinKnowledge or if it is OK here. We might have to clear above as well.

test/libyul/yulOptimizerTests/unusedAssignEliminator/if_revert.yul
```diff
@@ -0,0 +1,27 @@
+{
```
chriseth: I think this should go to a different PR.

chriseth:
This probably needs some more tests.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -430,3 +421,32 @@ std::optional<YulString> DataFlowAnalyzer::isSimpleLoad(
 				return key->name;
 	return {};
 }
+
+bool DataFlowAnalyzer::hasFlowOutControlFlow(Block const& _block) const
+{
+	return
+		_block.statements.empty() ||
+		TerminationFinder{m_dialect, &m_controlFlowSideEffects}.
+			controlFlowKind(_block.statements.back()) == TerminationFinder::ControlFlow::FlowOut;
```
ekpyron, line 462: I find this rather creepy... I'd much rather have ``TerminationFinder`` run on the full AST up front and map block pointers to out flow or something like that...
Every time I see this, I expect bugs due to ``{ break leave }`` or ``{ continue revert(0,0) }`` or ``{ if someCond { continue } revert(0,0) }`` or something...
Although right now I'm not finding a case that will *actually* break...

Do we have tests for these things? Maybe we should have.

But maybe I'm paranoid here... we do remove unreachable stuff like that anyways and the for loop handling in here should still remain sound...

test/libyul/yulOptimizerTests/rematerialiser/non_movable_function.yul
```diff
@@ -12,7 +12,7 @@
 //     let a := 1
 //     let b := f(1)
 //     let c := 1
-//     mstore(add(1, b), 1)
+//     mstore(add(1, f(1)), 1)
```
ekpyron: The name of the test case suggests that the idea was for ``f`` to be non-movable and for this *not* to happen...
maybe add a ``call`` to ``f`` or something like that?

ekpyron (commented):
Some minor comments.

Apart from those I would have said this looks good, but it has properly failing semantics tests, so something is still off, but I actually don't see it...

Also some test cases have noticeably increased code size, I'd also be curious why that happens...

Apart from that: more tests, test updates and changelog entry.

ekpyron:
I spent some time looking at the ``call_options_overload`` test that properly fails in CI...
I think it's a homestead run (I can only reproduce it for that locally)...

But still I neither saw what's wrong in Yul nor in the generated asm even...

Until I finally noticed: it's an actual ``out-of-gas``... changing it to ``3000`` gas makes it pass... :-).

ekpyron:
Hahaha - I think now I see what really happened :-).

The homestead test failure is due to https://github.com/ethereum/solidity/issues/12668 !

Unoptimized it's a stack-too-deep case. I haven't verified, but I think before this PR, it needed the full optimizer to recover from that, hence gas cost for the function call was low in practice.
After this PR, I think the minimal optimizations are enough to fix the stack-too-deep error - but the function is not sufficiently optimized for the call to go through with 1000 gas. So only ``--evm-version homestead --optimize`` passes, but ``--evm-version homestead`` fails.

So yeah, we can just bump the gas value there :-).

libyul/optimiser/DataFlowAnalyzer.cpp
chriseth: But this function is totally fine even if we didn't remove dead code already, isn't it?

1) This function returning "true" can always be wrong because of some complicated invariants / conditions
2) If the last statement in the block is a non-flowout statement, but there is additional control-flow, then we must handle that correctly already anyway (continue / break)

Note that `leave` is always ignored in this component.

test/libyul/yulOptimizerTests/commonSubexpressionEliminator/for_reassign_continue.sol
```diff
@@ -0,0 +1,22 @@
...
+        x := 0
+        if calldataload(0) { x := 2 continue }
+    }
+}
+// ----
+// step: commonSubexpressionEliminator
+//
+// {
+//     let x := 0
+//     for { } sload(0) { let t := 0 }
```
chriseth, line 18: The optimizer currently replaces this by `let t := x`, which is wrong - `x` was reassigned in line 18.

To fix this, we need to store the state at each continue and break statement and take this properly into account like we do in the unused store base.

test/libyul/yulOptimizerTests/loadResolver/for_break_sstore.yul
```diff
@@ -0,0 +1,51 @@
+{
+    sstore(0, 123213)
+    let x := 0
+    for {} sload(0) { mstore(0x50, sload(0)) } {
```
chriseth, line 4: The sload(0) in the post could still be replaced, but maybe this is a bit too much.

libyul/optimiser/DataFlowAnalyzer.h
```diff
@@ -184,9 +180,27 @@ class DataFlowAnalyzer: public ASTModifier
 		std::unordered_map<YulString, YulString> storage;
 		std::unordered_map<YulString, YulString> memory;
 	};
+
+	/// Joins knowledge about storage and memory with an older point in the control-flow.
+	/// This only works if the current state is a direct successor of the older point,
```
chriseth, line 185: This is not true anymore.

libyul/optimiser/Rematerialiser.cpp
```diff
@@ -87,6 +87,11 @@ void Rematerialiser::visit(Expression& _e)
 	DataFlowAnalyzer::visit(_e);
 }
 
+void LiteralRematerialiser::run(OptimiserStepContext& _context, Block& _ast)
```
chriseth, line 90: We might not need this.

bshastry:
Minor update from the fuzzer: There are no divergences between unoptimized and optimized programs found due to this PR after several days of fuzzing.

leonardoalt:
Needs some update I guess, based on comments.

libyul/optimiser/DataFlowAnalyzer.h
```diff
@@ -168,14 +154,24 @@ class DataFlowAnalyzer: public ASTModifier
...
 	/// Side-effects of user-defined functions. Worst-case side-effects are assumed
 	/// if this is not provided or the function is not found.
 	std::map<YulString, SideEffects> m_functionSideEffects;
+	std::map<YulString, ControlFlowSideEffects> m_controlFlowSideEffects;
 
 private:
 	struct State
 	{
+		/// Latest assignment counter per variable.
+		std::map<YulString, size_t> latestAssignment;
```
ekpyron, line 174: I'd somehow feel better with a ``std::map<YulString, Statement const*> latestAssignment;`` or something like that - which would be enough, since we only need to check for equality and which wouldn't raise the question of whether ``<`` on the values has special significance or not...
But getting a nice suitable pointer in each case would be a hassle, so I guess a unique counter value is just fine...

ekpyron (commented):
I read through it once now and I don't see anything obviously wrong with the logic, but I'll need another look to convince myself that it's always safe and we might need some more tests.

ekpyron:
Hm... should have proceeded with this sooner... now the conflicts list looks like this will make for an uncomfortable rebase...

leonardoalt:
> Hm... should have proceeded with this sooner... now the conflicts list looks like this will make for an uncomfortable rebase...

Damn yea true. We could try and see how much effort that'd be. Most conflicts are in tests which can just be ignored and rerun.

ekpyron:
> > Hm... should have proceeded with this sooner... now the conflicts list looks like this will make for an uncomfortable rebase...
> 
> Damn yea true. We could try and see how much effort that'd be. Most conflicts are in tests which can just be ignored and rerun.

Nope, I meant the actual conflicts in basically everything that was changed here - just had a peek, it's definitely non-trivial to rebase, although probably feasible.

ekpyron:
After a quick chat with @chriseth we decided to close this for now, since the data flow analyzer is up for a refactor anyways and it may be easier to recreate this afterwards.

PR #12770, opened 2022-03-10
Adding a selector field for errors and events

### LINKED ISSUE #12581: Extend the language to have an EVENT.selector and ERROR.selector
## Abstract

Add `ev.selector` and `err.selector` fields that would work similarly to `fun.selector`, extracting event signature (topic 0) from its definition

```
interface MyInterface {
    event someEvent();
    error someError();
    function someFunction() external;
}

contract Test {
    function nameIsIrrelevant() public {
        bytes4  x = MyInterface.someFunction.selector; // This works
        bytes32 y = MyInterface.someEvent.selector; // This doesn't, that is what I'd like to add
        bytes32 z = MyInterface.someError.selector; // This doesn't, that is what I'd like to add
    }
}
```

## Motivation

Some cross chain mechanism, such as polygon's [FxBaseRootTunnel](https://github.com/fx-portal/contracts/blob/main/contracts/tunnel/FxBaseRootTunnel.sol), perform event inclusion check that require the event signature. This is likely to become a more frequent need with inclusion proofs being an important part of rollups.

## Specification

Add a bytes32 field named `.selector`, `.topic0` or `.signature` to events (and a similar one to errors), that the compiler would replace with the event's signature (keccak256 of the event description)

## Backwards Compatibility

I can't imagine this would cause any backward compatibility issue

nishant-sachdeva:
closes #12581

### DIFF
--- a/test/libsolidity/syntaxTests/events/selector_field_for_events_and_errors.sol
+++ b/test/libsolidity/syntaxTests/events/selector_field_for_events_and_errors.sol
@@ -1,0 +1,15 @@
+interface MyInterface {
+    event someEvent();
+    error someError();
+    function someFunction() external;
+}
+
+contract Test {
+    function nameIsIrrelevant() pure public {
+        bytes4  x = MyInterface.someFunction.selector; // This works
+        bytes32 y = MyInterface.someEvent.selector; // This doesn't, that is what I'd like to add
+        bytes32 z = MyInterface.someError.selector; // This doesn't, that is what I'd like to add
+    }
+}
+// ----
+// TypeError 9582: (262-283): Member "someEvent" not found or not visible after argument-dependent lookup in type(contract MyInterface).


test/libsolidity/syntaxTests/events/selector_field_for_events_and_errors.sol
```diff
@@ -0,0 +1,15 @@
...
+    event someEvent();
+    error someError();
+    function someFunction() external;
+}
+
+contract Test {
+    function nameIsIrrelevant() pure public {
+        bytes4  x = MyInterface.someFunction.selector; // This works
+        bytes32 y = MyInterface.someEvent.selector; // This doesn't, that is what I'd like to add
+        bytes32 z = MyInterface.someError.selector; // This doesn't, that is what I'd like to add
```
chriseth, line 11: error selector is only four bytes

nishant-sachdeva: So, we already have a selector for errors ?
We only need to implement it for events then?

chriseth: I don't think we have it, but it will be bytes4

PR #12771, opened 2022-03-10
Clarifying a doc entry on Address.balance

### LINKED ISSUE #12772: Documentation for Address.balance should include additional information
## Page

https://github.com/ethereum/solidity/blob/develop/docs/units-and-global-variables.rst

## Abstract

The value returned by address.balance is slightly different for msg.sender. In that case it is reduced by the transaction's gas limit multiplied by the tx.gasprice.

Code that attempts to compare balances and make decisions should be aware of this.

I have proposed a tweak to the documentation for address.balance. Please review and edit as appropriate.

## Pull request

https://github.com/ethereum/solidity/pull/12771

dsmitchell:
Fixes #12772.

The value returned by `address.balance` is slightly different for `msg.sender`. In that case it is reduced by the transaction's gas limit multiplied by the `tx.gasprice`.

Code that attempts to compare balances and make decisions should be aware of this.

I have proposed a tweak to the documentation for `address.balance`. Please review and edit as appropriate.

Thank you.

### DIFF
--- a/docs/units-and-global-variables.rst
+++ b/docs/units-and-global-variables.rst
@@ -246,7 +246,7 @@ Members of Address Types
 ------------------------
 
 ``<address>.balance`` (``uint256``)
-    balance of the :ref:`address` in Wei
+    balance of the :ref:`address` in Wei. In the case of msg.sender, the returned value is reduced by the expected gas cost of the transaction (tx gas limit * ``tx.gasprice``)
 
 ``<address>.code`` (``bytes memory``)
     code at the :ref:`address` (can be empty)


chriseth:
In my opinion, this just explains a very special case and is not really useful at this point. `address.balance` is always the current balance at that address. There is a multitude of factors that can influence this number.

dsmitchell:
Maybe then (to keep it simple) just mention that there are factors that can influence the value, to make people aware that it is not necessarily the same as the balance at the moment the transaction is submitted. I'm sure I'm not the only one who will ever be affected by the lack of this knowledge

chriseth:
Sure! Then maybe change it to "current balance..."?

dsmitchell:
Problem with just "current balance" is that I can create a tx with a gas limit much higher than needed, bringing the supposed "current balance" to 0, which would not be true when the transaction completes

dsmitchell:
While I like that the balance returns the "worst case" balance at the end of the tx (to be safe), knowing this could've helped me with both contract code and the client code submitting the tx

dsmitchell:
Anyway, `msg.sender.balance` is neither the balance at the moment the transaction was submitted, and likely not the balance upon completion of the transaction...unless the gas limit is specified perfectly.

chriseth:
It is the balance at the moment the statement is executed. And isn't that the case for any other statement in solidity?

dsmitchell:
I would say something to the effect that it is the balance minus the amount reserved in order to execute the transaction.

I will think more on how to succinctly capture this info with simple text, but will also gladly accept suggestions from those in the community that are far more familiar with the vocabulary and the factors involved in its calculation

chriseth:
Note that whenever you sent ether during the call, the value will change again. If you want, you can find a better wording, but I will be closing this PR in the meantime.

PR #12773, opened 2022-03-10
Fix encodeCall bug.

ekpyron:
Came up in https://github.com/ethereum/solidity/pull/12750#issuecomment-1062636745
Needs at least a changelog entry and potentially a bug list entry.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 ### 0.8.13 (unreleased)
 
+Important Bugfixes:
+ * Code Generator: Correctly encode literals used in ``abi.encodeCall`` in place of fixed bytes arguments.
+
+
 Language Features:
  * General: Allow annotating inline assembly as memory-safe to allow optimizations and stack limit evasion that rely on respecting Solidity's memory model.
  * General: ``using M for Type;`` is allowed at file level and ``M`` can now also be a brace-enclosed list of free functions or library functions.
--- a/docs/bugs.json
+++ b/docs/bugs.json
@@ -1,5 +1,16 @@
 [
     {
+        "uid": "SOL-2022-1",
+        "name": "AbiEncodeCallLiteralAsFixedBytesBug",
+        "summary": "Literals used for a fixed length bytes parameter in ``abi.encodeCall`` were encoded incorrectly.",
+        "description": "For the encoding, the compiler only considered the types of the expressions in the second argument of ``abi.encodeCall`` itself, but not the parameter types of the function given as first argument. In almost all cases the abi encoding of the type of the expression matches the abi encoding of the parameter type of the given function. This is because the type checker ensures the expression is implicitly convertible to the respective parameter type. However this is not true for number literals used for fixed bytes types shorter than 32 bytes, nor for string literals used for any fixed bytes type. Number literals were encoded as numbers instead of being shifted to become left-aligned. String literals were encoded as dynamically sized memory strings instead of being converted to a left-aligned bytes value.",
+        "link": "https://blog.soliditylang.org/2022/03/16/encodecall-bug/",
+        "introduced": "0.8.11",
+        "fixed": "0.8.13",
+        "severity": "very low"
+
+    },
+    {
         "uid": "SOL-2021-4",
         "name": "UserDefinedValueTypesBug",
         "summary": "User defined value types with underlying type shorter than 32 bytes used incorrect storage layout and wasted storage",
@@ -8,7 +19,6 @@
         "introduced": "0.8.8",
         "fixed": "0.8.9",
         "severity": "very low"
-
     },
     {
         "uid": "SOL-2021-3",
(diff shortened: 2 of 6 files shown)


stackenbotten:
There was an error when running `chk_coding_style` for commit `46d2d8581d0654e12d15784c5b05cb2d942529b3`:
```
Changelog.md:15: * Code Generator: Correctly encode literals used in ``abi.encodeCall`` in place of fixed bytes arguments. 

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `61b2b61c2b1c462063a2d1bbe91dcbf63fa31cd1`:
```
Changelog.md:15: * Code Generator: Correctly encode literals used in ``abi.encodeCall`` in place of fixed bytes arguments. 

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `697fa1dc98b4938c281939d25fca9db959407f5f`:
```
Changelog.md:15: * Code Generator: Correctly encode literals used in ``abi.encodeCall`` in place of fixed bytes arguments. 

```

test/libsolidity/semanticTests/abiencodedecode/abi_encode_call_uint_bytes.sol
```diff
@@ -0,0 +1,20 @@
+contract C {
+	function removeSignature(bytes calldata x) external pure returns (bytes memory) {
+		return x[4:];
+	}
```
ekpyron, line 4: I can get away with this, since gas is irrelevant, but we should really think about how we can do memory slices :-).

Changelog.md
```diff
@@ -11,6 +11,8 @@ Compiler Features:
 Bugfixes:
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
 
+Important Bugfixes:
```
chriseth: Should be first entry.

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-1",
+        "name": "AbiEncodeCallLiteralAsFixedBytesBug",
+        "summary": "Literals used in place of a fixed bytes argument in ``abi.encodeCall`` were encoded incorrectly",
```
chriseth:
```suggestion
        "summary": "Literal numbers used for a fixed length bytes parameter in ``abi.encodeCall`` were encoded incorrectly.",
```
(I assume "123" works correctly)

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
...
+        "description": "The compiler only considered the types of the expressions in the second argument of ``abi.encodeCall`` itself, but not the parameter types of the function given as first argument. While in almost all cases the abi encoding of the type of the expression matches the abi encoding of the parameter type of the given function, since the type checker ensures the expression is implicitly convertible to the respective parameter type, this is not true for literals that are implicitly convertible to fixed bytes types shorter than 32 bytes, which have to be shifted to become left-aligned. This shifting was not performed.",
```
chriseth:
```suggestion
        "description": "For the encoding, the compiler only considered the types of the expressions in the second argument of ``abi.encodeCall`` itself, but not the parameter types of the function given as first argument. In almost all cases the abi encoding of the type of the expression matches the abi encoding of the parameter type of the given function. This is because the type checker ensures the expression is implicitly convertible to the respective parameter type. However, this is not true for literal numbers that are implicitly convertible to fixed bytes types shorter than 32 bytes. They were encoded as numbers instead of being shifted to become left-aligned.",
```

docs/bugs.json
ekpyron: Kept it at "literals"... decimal literals and hex literals of incorrect length won't convert, i.e. fail in the type checker... string literals, however, are even more messed up and encoded as ``string memory``... which can also happen for any string shorter than 32 bytes and any larger fixed bytes type (i.e. ``"ab"`` will pass type checking as ``bytes32`` and encode a tail pointer to a memory string...)

docs/bugs.json
ekpyron: Rephrased somewhat following most of the suggestion, but also added the string literal case.

ekpyron: Should we mention that ``very low`` is due to any single test on an occurrence of the bug will catch it?

chriseth: We never justified our severity ratings.

ekpyron:
Anything left here? Looks good to me.

chriseth:
Could use some squashing.

chriseth (approved):

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12775, opened 2022-03-10
Refactor data flow analyzer state access.

chriseth:


test/libyul/KnowledgeBaseTest.cpp
```diff
@@ -58,7 +58,7 @@ class KnowledgeBaseTest
 		for (auto const& [name, expression]: m_ssaValues.values())
 			m_values[name].value = expression;
 
-		return KnowledgeBase(m_dialect, m_values);
+		return KnowledgeBase(m_dialect, [this] -> decltype(auto) { return m_values; });
```
ekpyron:
```suggestion
		return KnowledgeBase(m_dialect, [this]() -> decltype(auto) { return m_values; });
```
also here should finally do it, I'd hope.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -47,7 +48,7 @@ DataFlowAnalyzer::DataFlowAnalyzer(
 ):
 	m_dialect(_dialect),
 	m_functionSideEffects(std::move(_functionSideEffects)),
-	m_knowledgeBase(_dialect, m_value)
+	m_knowledgeBase(_dialect, [this] () -> decltype(auto) { return allValues(); })
```
ekpyron:
```suggestion
	m_knowledgeBase(_dialect, [this]() -> decltype(auto) { return allValues(); })
```
?

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -218,6 +219,18 @@ void DataFlowAnalyzer::operator()(Block& _block)
 	assertThrow(numScopes == m_variableScopes.size(), OptimizerException, "");
 }
 
+optional<YulString> DataFlowAnalyzer::storageValue(YulString _key) const
+{
+	auto it = m_state.storage.find(_key);
+	return it == m_state.storage.end() ? nullopt : optional<YulString>{it->second};
```
ekpyron:
```suggestion
	if (auto const* value = util::valueOrNullptr(m_state.storage, _key))
		return *value;
	else
		return nullopt;
```
Or we could add something like ``util::valueOrNullopt`` or ``util::valueAsOptional`` or so... but well, also fine as is...

libyul/optimiser/KnowledgeBase.cpp
```diff
@@ -80,8 +80,8 @@ bool KnowledgeBase::knownToBeZero(YulString _a)
 
 optional<u256> KnowledgeBase::valueIfKnownConstant(YulString _a)
 {
-	if (m_variableValues.count(_a))
-		if (Literal const* literal = get_if<Literal>(m_variableValues.at(_a).value))
+	if (m_variableValues().count(_a))
+		if (Literal const* literal = get_if<Literal>(m_variableValues().at(_a).value))
 			return valueOfLiteral(*literal);
```
ekpyron:
```suggestion
	if (Literal const* literal = get_if<Literal>(util::valueOrNullptr(m_variableValues(), _a)))
		return valueOfLiteral(*literal);
```

ekpyron (dismissed):
Looks good - only minor comments.

libyul/optimiser/KnowledgeBase.cpp
chriseth: That doesn't work because we still need to access the `.value` member of the map value.

chriseth:
I think this is working now, but I don't like the public / protected division in the class. I'll clean it up a bit better.

libyul/optimiser/KnowledgeBase.h
```diff
@@ -37,15 +38,16 @@ struct AssignedValue;
...
- *
- * The reference to the map of values provided at construction is assumed to be updating.
  */
 class KnowledgeBase
 {
 public:
-	KnowledgeBase(Dialect const& _dialect, std::map<YulString, AssignedValue> const& _variableValues):
+	KnowledgeBase(
+		Dialect const& _dialect,
+		std::function<std::map<YulString, AssignedValue> const&()> _variableValues
```
ekpyron: I wonder if it wouldn't actually be nicer to make this a
``std::function<AssignedValue const*(YulString)>``

But this would require further changes down to ``SimplificationRules`` and ``Pattern``, so maybe not worth it at least right away... even though at a quick glance I haven't seen anything actually traversing this thing, so it might actually be possible - but maybe I just missed it.

libyul/optimiser/DataFlowAnalyzer.cpp
```diff
@@ -153,11 +154,11 @@ void DataFlowAnalyzer::operator()(FunctionDefinition& _fun)
...
 	// but this could be difficult if it is subclassed.
-	ScopedSaveAndRestore valueResetter(m_value, {});
+	ScopedSaveAndRestore valueResetter(m_state.value, {});
 	ScopedSaveAndRestore loopDepthResetter(m_loopDepth, 0u);
-	ScopedSaveAndRestore referencesResetter(m_references, {});
-	ScopedSaveAndRestore storageResetter(m_storage, {});
-	ScopedSaveAndRestore memoryResetter(m_memory, {});
+	ScopedSaveAndRestore referencesResetter(m_state.references, {});
+	ScopedSaveAndRestore storageResetter(m_state.storage, {});
+	ScopedSaveAndRestore memoryResetter(m_state.memory, {});
```
ekpyron: So this still can't just be a ``stateResetter`` now?

ekpyron (dismissed):
https://github.com/ethereum/solidity/pull/12775/files#r826191625 may be worth a thought, if it can be changed.
Apart from that, everything should remain equivalent after this change, so I'm approving.

libyul/optimiser/DataFlowAnalyzer.cpp
chriseth: Good point

libyul/optimiser/KnowledgeBase.h
chriseth: CommonSubexpressionEliminator.cpp:109 does the traversal

chriseth: Ah wait, but that's a subclass

ekpyron (approved):

Sticknmove (commented):
Thanks

Sticknmove (commented):
Thanks

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12781, opened 2022-03-11
Fix import directive visits in type checker and view pure checker.

### LINKED ISSUE #12780: Cannot import shadowed functions in Solidity 0.8.12
## Description

Importing overloaded methods in Solidity 0.8.12 gives an error instead of loading the methods, even if the methods are not called.

Other info that might be relevant:
- Checking the Solidity change list I found something that *might* be relevant
> Inheritance: Consider functions in all ancestors during override analysis.

- @chriseth once said that Solidity performs conflict resolution when the method is called. However, this doesn't seem to be the case anymore.
https://github.com/ethereum/solidity/issues/12296#issuecomment-971705133

## Environment

- Compiler version: 0.8.12
- Target EVM version (as per compiler settings): compiler default (but happens with all versions)
- Framework/IDE (e.g. Truffle or Remix): Remix, but happens everywhere
- EVM execution environment / backend / blockchain client: any
- Operating system: any

## Steps to Reproduce

Consider these 2 files created in Remix IDE

- `Math.sol` - defines 2 methods, same name, different arguments

```solidity
function sub(uint256 x, uint256 y) pure {}
function sub(uint256 x, int256 y) pure {}
```

- `ImportMath.sol` - just tries to import the `sub` method 

```solidity
import {sub} from "./Math.sol";
```

Output error on compilation:

```
TypeError: No matching declaration found after variable lookup.
 --> ImportMath.sol:1:9:
  |
1 | import {sub} from "./Math.sol";
  |         ^^^
```

That means I don't know how to import those methods.

ekpyron:
Fixes https://github.com/ethereum/solidity/issues/12780

This was an unfortunate omission in https://github.com/ethereum/solidity/pull/12494

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@ Compiler Features:
 
 Bugfixes:
  * General: Fix internal error for locales with unusual capitalization rules. Locale set in the environment is now completely ignored.
+ * Type Checker: Fix incorrect type checker errors when importing overloaded functions.
  * Yul IR Code Generation: Optimize embedded creation code with correct settings. This fixes potential mismatches between the constructor code of a contract compiled in isolation and the bytecode in ``type(C).creationCode``, resp. the bytecode used for ``new C(...)``.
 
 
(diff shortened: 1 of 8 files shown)


ekpyron:
Probably just also needs this overwritten in the smt checker.

ekpyron (commented):
Approving the changes by @chriseth.

chriseth (approved):

PR #12786, opened 2022-03-14
Moves comment above the line for which it was intended.

RominHalltari:
That comment above the `Voter storage delegate_ = voters[to];` line is confusing, because that's not what that line of code does. This moves that comment above the line for which it was intended.


ekpyron:
Nope, the comment is right where it's supposed to...

PR #12796, opened 2022-03-15
Refactor stack compressor

chriseth:


libyul/optimiser/StackCompressor.cpp
```diff
@@ -156,62 +168,81 @@ set<YulString> chooseVarsToEliminate(
...
+	map<YulString, map<size_t, vector<tuple<YulString, set<YulString>>>>> candidates = selector.candidates();
+
+	set<YulString> varsToEliminate;
+	for (auto const& [functionName, numVariables]: _numVariables)
+	{
+		yulAssert(numVariables > 0);
+		varsToEliminate += chooseVarsToEliminate(candidates[functionName], static_cast<size_t>(numVariables));
+	}
+
+	// TODO this also did not set the last parameter (only selected variables) to true before the refactor.
```
chriseth: @ekpyron what do you think about this?

libyul/optimiser/StackCompressor.cpp
```diff
@@ -50,31 +50,41 @@ namespace
...
 	{
-		map<size_t, vector<tuple<YulString, set<YulString>>>> cand;
-		for (auto const& candidate: m_candidates)
+		map<YulString, map<size_t, vector<tuple<YulString, set<YulString>>>>> cand;
+		for (auto const& [functionName, candidate]: m_candidates)
 		{
 			if (size_t const* cost = util::valueOrNullptr(m_expressionCodeCost, candidate))
 			{
 				size_t numRef = m_numReferences[candidate];
 				set<YulString> const* ref = references(candidate);
```
chriseth, line 72: Oh wow! This uses the references to the candidate after the whole thing has run.
It should be zero at that point.

chriseth: Ok, the only influence this has is that we stop looking for candidates in the non-optimized codegen as soon as we would want to eliminate a variable that references one that has already been chosen for elimination.

So in the worst case, we eliminate more variables than we actually should.

libyul/optimiser/StackCompressor.cpp
ekpyron: Yeah, I only specifically added the "only selected variable" mode for the new code transform.
Changing it for the old code transform path wreaked havoc IIRC - not sure if it brought back more stack-too-deep errors, since the old code transform is too optimistic in its deficit guess, or whether it was gas changes... if run via the old code transform also the ``"fDnTOc g"`` sequence is run afterwards... changing any of that had various effects, so I thought it easiest to just keep the old path as it was before...

ekpyron (approved):

PR #12797, opened 2022-03-15
Improve Error Reporting of SemVer Parser

vinayman:


vinayman:
Hi @chriseth 
Apologies for taking so long to coming around to this. I've made some of the edits you mentioned but I see some tests failing , e.g.,:

`/solidity/test/boostTest.cpp:120: error: in "syntaxTests/pragma/broken_version_5": Exception during extracted test: /solidity/libsolidity/parsing/Parser.cpp(166): Throw in function void solidity::frontend::Parser::parsePragmaVersion(const solidity::langutil::SourceLocation &, const vector<solidity::langutil::Token> &, const vector<std::string> &)
Dynamic exception type: boost::wrapexcept<solidity::langutil::SemVerError>
std::exception::what: Invalid versionPart: _
[solidity::util::tag_comment*] = Invalid versionPart: _`

`/solidity/test/boostTest.cpp:120: error: in "syntaxTests/pragma/broken_version_4": Exception during extracted test: /solidity/libsolidity/parsing/Parser.cpp(166): Throw in function void solidity::frontend::Parser::parsePragmaVersion(const solidity::langutil::SourceLocation &, const vector<solidity::langutil::Token> &, const vector<std::string> &)
Dynamic exception type: boost::wrapexcept<solidity::langutil::SemVerError>
std::exception::what: Invalid versionPart: (
[solidity::util::tag_comment*] = Invalid versionPart: (`

Whilst I debug them I was hoping you might be able to take a look and see whether this is more inline with what you were thinking following your comments.

I had to decline the old PR as trying to sync with the branch I accidentally pulled a load of unwanted commits.

Thanks

liblangutil/SemVerHandler.h
```diff
@@ -25,17 +25,27 @@
...
 #include <optional>
 #include <utility>
 #include <vector>
+#include <variant>
 
 namespace solidity::langutil
 {
 
-class SemVerError: public util::Exception
+struct SemVerError: public Error
```
chriseth: I think it would be easier if this was not part of the Solidity error inheritance hierarchy. You don't make use of most of the features (source location etc) and you would not need to assign a new error id for each of the tiny ways the parser can fail.

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,7 +156,13 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
 		vector<Token> tokens(_pragma.tokens().begin() + 1, _pragma.tokens().end());
 		vector<string> literals(_pragma.literals().begin() + 1, _pragma.literals().end());
 		SemVerMatchExpressionParser parser(tokens, literals);
-		auto matchExpression = parser.parse();
+		auto matchExpressionOrError = parser.parse();
+		if (std::holds_alternative<SemVerError>(matchExpressionOrError))
+		{
+			auto matchExpressionError = std::get<SemVerError>(matchExpressionOrError);
+			BOOST_THROW_EXCEPTION(matchExpressionError);
```
chriseth: This shouldn't directly throw exception but rather use `m_errorReporter.syntaxError` instead.

chriseth: Oh and we certainly need a source location here!

libsolidity/parsing/Parser.cpp
```diff
@@ -159,7 +159,13 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
...
 {
 	SemVerMatchExpressionParser parser(_tokens, _literals);
-	auto matchExpression = parser.parse();
+	auto matchExpressionOrError = parser.parse();
+	if (std::holds_alternative<SemVerError>(matchExpressionOrError))
+	{
+		auto matchExpressionError = std::get<SemVerError>(matchExpressionOrError);
+		BOOST_THROW_EXCEPTION(matchExpressionError);
+	}
+	auto matchExpression = std::get<std::optional<SemVerMatchExpression>>(matchExpressionOrError);
```
chriseth: Is there still a need for the `optional`? Isn't it either an error or a valid match expression?

chriseth:
This looks very good already!

libsolidity/parsing/Parser.cpp
vinayman: I think you're right because of the way the match expression is handled after. I just wasn't sure whether `m_tokens.empty()` which would return a `nullopt` is a valid and handled return type (Line 163 of SemVerHandler.cpp)

I will alter it to just an error or valid match expression.

libsolidity/analysis/SyntaxChecker.cpp
vinayman: The one thing i wasn't sure about here - would it technically be a `fatalError` in which case we should have a `fatalSyntaxError` which we would throw? Or the `syntaxError` be enough where we would add to `m_errorList` and only throw if there are excessive errors?

I wonder this because in the line below we have a `solAssert` which would throw an exception if the `matchExpression` has no value

chriseth: Actually I think the parser ("parsePragmaVersion") already generates a fatal error if there is anything wrong. This means at this point here, there should not be any error and you could use `solAssert` to check that.

liblangutil/SemVerHandler.cpp
```diff
@@ -171,14 +171,19 @@ optional<SemVerMatchExpression> SemVerMatchExpressionParser::parse()
 			if (m_pos >= m_tokens.size())
 				break;
 			if (currentToken() != Token::Or)
-				BOOST_THROW_EXCEPTION(SemVerError());
+			{
+				string tokenString = TokenTraits::toString(currentToken());
+				BOOST_THROW_EXCEPTION(
+					SemVerError() << util::errinfo_comment("Invalid token or more than one version not delimited by ||; token: " + (tokenString.empty() ? "" : tokenString))
```
chriseth:
```suggestion
					SemVerError() << util::errinfo_comment("Invalid token or more than one version not delimited by || token: " + (tokenString.empty() ? "" : tokenString))
```

liblangutil/SemVerHandler.h
```diff
@@ -25,17 +25,19 @@
...
-#include <optional>
 #include <utility>
 #include <vector>
+#include <variant>
 
 namespace solidity::langutil
 {
 
-class SemVerError: public util::Exception
+struct SemVerError : public util::Exception
```
chriseth:
```suggestion
struct SemVerError: public util::Exception
```

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,11 +156,20 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
...
+		{
+			auto matchExpressionError = std::get<SemVerError>(matchExpressionOrError);
+			m_errorReporter.syntaxError(
+				9981_error,
+				_pragma.location(),
+				matchExpressionError.what()
+			);
+		}
+		auto matchExpression = std::get<SemVerMatchExpression>(matchExpressionOrError);
 		// An unparsable version pragma is an unrecoverable fatal error in the parser.
```
chriseth: This comment essentially says that the above can never happen.

libsolidity/parsing/Parser.cpp
```diff
@@ -159,17 +159,20 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
 void Parser::parsePragmaVersion(SourceLocation const& _location, vector<Token> const& _tokens, vector<string> const& _literals)
 {
 	SemVerMatchExpressionParser parser(_tokens, _literals);
-	auto matchExpression = parser.parse();
-	if (!matchExpression.has_value())
+	auto matchExpressionOrError = parser.parse();
+	if (std::holds_alternative<SemVerError>(matchExpressionOrError))
+	{
+		auto matchExpressionError = std::get<SemVerError>(matchExpressionOrError);
```
```suggestion
	if (SemVerError const* matchExpressionError = std::get_if<SemVerError>(&matchExpressionOrError))
	{
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -41,7 +41,7 @@ BOOST_AUTO_TEST_SUITE(SemVerMatcher)
 namespace
 {
 
-SemVerMatchExpression parseExpression(string const& _input)
+SemVerMatchExpressionParser createSemVerMatchExpressionParser(string const& _input)
```
chriseth: Why did you split the function?

leonardoalt:
Needs review again

vinayman:
Hey @chriseth - just a friendly chase on this PR - the only outstanding thing I believe is now the failing test `broken_version_7.sol` mentioned previously and a subsequent re-review. Any suggestions on how to progress so we can get this merged?

cameel:
@nishant-sachdeva Maybe you could help @vinayman with the test here?

liblangutil/SemVerHandler.h
```diff
@@ -25,17 +25,19 @@
...
-#include <optional>
 #include <utility>
 #include <vector>
+#include <variant>
 
 namespace solidity::langutil
 {
 
-class SemVerError: public util::Exception
+struct SemVerError: public util::Exception
```
Marenz, line 37: Why this change?

liblangutil/SemVerHandler.h
```diff
@@ -87,8 +89,8 @@ class SemVerMatchExpressionParser
 public:
 	SemVerMatchExpressionParser(std::vector<Token> _tokens, std::vector<std::string> _literals);
 
-	/// Returns an expression if it was parseable, or nothing otherwise.
-	std::optional<SemVerMatchExpression> parse();
+	/// Returns an expression if it was parseable, or error otherwise.
```
Marenz: Nitpicking, not really important: `an error`

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,11 +156,11 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
 		vector<Token> tokens(_pragma.tokens().begin() + 1, _pragma.tokens().end());
 		vector<string> literals(_pragma.literals().begin() + 1, _pragma.literals().end());
 		SemVerMatchExpressionParser parser(tokens, literals);
-		auto matchExpression = parser.parse();
+		std::variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
+		SemVerMatchExpression const& matchExpression = std::get<SemVerMatchExpression>(matchExpressionOrError);
```
Marenz: how is the error case handled here when `std::get()` throws `bad_variant_access` here? 
Are errors not possible at this place? If so, please add comment. And in that case you could also do

```
SemVerMatchExpression const& matchExpression = std::get<SemVerMatchExpression>(parser.parse());
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -41,7 +41,8 @@ BOOST_AUTO_TEST_SUITE(SemVerMatcher)
 namespace
 {
 
-SemVerMatchExpression parseExpression(string const& _input)
+
```
Marenz: we don't need that extra new line

test/libsolidity/syntaxTests/pragma/broken_version_7.sol
```diff
@@ -1,3 +1,3 @@
 pragma solidity >0.5.0<;
 // ----
-// ParserError 1684: (0-24): Found version pragma, but failed to parse it. Please ensure there is a trailing semicolon.
+// ParserError 1684: (0-24): Invalid token in version Pragma: �
```
Marenz: Where does this `ÿ` come from? Maybe an uninitialized variable?

Marenz (dismissed):

leonardoalt:
is @nishant-sachdeva taking over this PR?

matheusaaguiar:
I am taking over instead of @nishant-sachdeva. Unless, OP decides to return, but it seems a long time has already passed with no contact...

liblangutil/SemVerHandler.h
matheusaaguiar: I guess OP reasoned that since there's no virtual functions and no members, it should be a struct, as recommended by the coding style document.

test/libsolidity/syntaxTests/pragma/broken_version_7.sol
matheusaaguiar: After it gets the `<` at the end, there's no more tokens to read from, but still it tries to print the `char(-1)` value generated by function `current_char`.

matheusaaguiar: I think it is not worth to implement a solution to cover such a corner case. What do you think?

cameel: Is it that hard to fix though? We should simply issue a different error when we reach the end - that we expected something more but the pragma suddenly ended.

matheusaaguiar: Yes, I guess I wasn't clear. I meant that I don't think it is worth to implement a solution to still print the offending char/token... but definitely, should produce an error message.

libsolidity/analysis/SyntaxChecker.cpp
matheusaaguiar: It seems like an unparsable pragma is a fatal error (mentioned in this resolved [comment](https://github.com/ethereum/solidity/pull/12797#discussion_r829041295)).

liblangutil/SemVerHandler.h
matheusaaguiar: fixed.

test/libsolidity/SemVerMatcher.cpp
matheusaaguiar: fixed.

matheusaaguiar:
Summary of changes since takeover:
 - small case pragma in error msgs
 - minor changes in string concatenation in error msgs
 - assert no `SemVerError` before getting value out of variant at line 160 of `SyntaxChecker.cpp`
 - add message when pragma version suddenly ends (broken test)
 - removed error code 9981 (added earlier i guess but then not used currently)
 - rebased

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,11 +156,12 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
 		vector<Token> tokens(_pragma.tokens().begin() + 1, _pragma.tokens().end());
 		vector<string> literals(_pragma.literals().begin() + 1, _pragma.literals().end());
 		SemVerMatchExpressionParser parser(tokens, literals);
-		auto matchExpression = parser.parse();
+		std::variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
```
nikola-matic:
```suggestion
		variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
```

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,11 +156,12 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
...
 		// An unparsable version pragma is an unrecoverable fatal error in the parser.
-		solAssert(matchExpression.has_value(), "");
+		solAssert(!std::holds_alternative<SemVerError>(matchExpressionOrError));
+		SemVerMatchExpression const& matchExpression = std::get<SemVerMatchExpression>(matchExpressionOrError);
```
nikola-matic:
```suggestion
		SemVerMatchExpression const& matchExpression = get<SemVerMatchExpression>(matchExpressionOrError);
```

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -156,11 +156,12 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
 		vector<Token> tokens(_pragma.tokens().begin() + 1, _pragma.tokens().end());
 		vector<string> literals(_pragma.literals().begin() + 1, _pragma.literals().end());
 		SemVerMatchExpressionParser parser(tokens, literals);
-		auto matchExpression = parser.parse();
+		std::variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
 		// An unparsable version pragma is an unrecoverable fatal error in the parser.
-		solAssert(matchExpression.has_value(), "");
+		solAssert(!std::holds_alternative<SemVerError>(matchExpressionOrError));
```
nikola-matic:
```suggestion
		solAssert(!holds_alternative<SemVerError>(matchExpressionOrError));
```

libsolidity/parsing/Parser.cpp
```diff
@@ -160,17 +160,19 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
 void Parser::parsePragmaVersion(SourceLocation const& _location, vector<Token> const& _tokens, vector<string> const& _literals)
 {
 	SemVerMatchExpressionParser parser(_tokens, _literals);
-	auto matchExpression = parser.parse();
-	if (!matchExpression.has_value())
+	std::variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
```
nikola-matic:
```suggestion
	variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
```

libsolidity/parsing/Parser.cpp
```diff
@@ -160,17 +160,19 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
...
+	if (SemVerError const* matchExpressionError = std::get_if<SemVerError>(&matchExpressionOrError))
```
nikola-matic:
```suggestion
	if (SemVerError const* matchExpressionError = get_if<SemVerError>(&matchExpressionOrError))
```

libsolidity/parsing/Parser.cpp
```diff
@@ -160,17 +160,19 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
...
+	{
 		m_errorReporter.fatalParserError(
 			1684_error,
 			_location,
-			"Found version pragma, but failed to parse it. "
-			"Please ensure there is a trailing semicolon."
+			matchExpressionError->what()
 		);
+	}
+	SemVerMatchExpression const& matchExpression = std::get<SemVerMatchExpression>(matchExpressionOrError);
```
nikola-matic:
```suggestion
	SemVerMatchExpression const& matchExpression = get<SemVerMatchExpression>(matchExpressionOrError);
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -41,7 +41,7 @@ BOOST_AUTO_TEST_SUITE(SemVerMatcher)
 namespace
 {
 
-SemVerMatchExpression parseExpression(string const& _input)
+std::variant<SemVerError, SemVerMatchExpression> parseExpression(string const& _input)
```
nikola-matic:
```suggestion
variant<SemVerError, SemVerMatchExpression> parseExpression(string const& _input)
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -58,17 +58,35 @@ SemVerMatchExpression parseExpression(string const& _input)
...
 
-	auto expression = SemVerMatchExpressionParser(tokens, literals).parse();
-	BOOST_REQUIRE(expression.has_value());
-	BOOST_CHECK_MESSAGE(
-		expression->isValid(),
-		"Expression \"" + _input + "\" did not parse properly."
-	);
-	return *expression;
+	auto expressionOrError = SemVerMatchExpressionParser(tokens, literals).parse();
+	if (SemVerMatchExpression const* expression = std::get_if<SemVerMatchExpression>(&expressionOrError))
```
nikola-matic:
```suggestion
	if (SemVerMatchExpression const* expression = get_if<SemVerMatchExpression>(&expressionOrError))
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -159,11 +177,15 @@ BOOST_AUTO_TEST_CASE(positive_range)
 	for (auto const& t: tests)
 	{
 		SemVerVersion version(t.second);
-		SemVerMatchExpression expression = parseExpression(t.first);
-		BOOST_CHECK_MESSAGE(
-			expression.matches(version),
-			"Version \"" + t.second + "\" did not satisfy expression \"" + t.first + "\""
-		);
+		auto expressionOrError = parseExpression(t.first);
+		if (std::holds_alternative<SemVerMatchExpression>(expressionOrError))
```
nikola-matic:
```suggestion
		if (holds_alternative<SemVerMatchExpression>(expressionOrError))
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -159,11 +177,15 @@ BOOST_AUTO_TEST_CASE(positive_range)
...
+		{
+			auto expression = std::get<SemVerMatchExpression>(expressionOrError);
```
nikola-matic:
```suggestion
			auto expression = get<SemVerMatchExpression>(expressionOrError);
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -235,12 +257,16 @@ BOOST_AUTO_TEST_CASE(negative_range)
...
 	{
 		SemVerVersion version(t.second);
-		SemVerMatchExpression expression = parseExpression(t.first);
-		BOOST_CHECK_MESSAGE(
-			!expression.matches(version),
-			"Version \"" + t.second + "\" did satisfy expression \"" + t.first + "\" " +
-			"(although it should not)"
-		);
+		auto expressionOrError = parseExpression(t.first);
+		if (std::holds_alternative<SemVerMatchExpression>(expressionOrError))
```
nikola-matic:
```suggestion
		if (holds_alternative<SemVerMatchExpression>(expressionOrError))
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -235,12 +257,16 @@ BOOST_AUTO_TEST_CASE(negative_range)
...
+		{
+			auto expression = std::get<SemVerMatchExpression>(expressionOrError);
```
nikola-matic:
```suggestion
			auto expression = get<SemVerMatchExpression>(expressionOrError);
```

nikola-matic (dismissed):
I've left what feels like a million comments, but they're all trivial and related to using the `std::` qualifier; rather, our code style is such that in `.cpp` files, we use `using namespace std;`, and therefore prefer calling functions from the `std` namespace without the `std::` qualifier, with a noted exception for `std::move`.

Please take a look, make the changes, rebase, and in the meantime we'll make sure to do another set of reviews so that we can merge this in :)

matheusaaguiar:
Rebased on develop after applying review suggestions.

liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +271,16 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
 		{
 			c = currentChar();
 			if (v * 10 < v || v * 10 + static_cast<unsigned>(c - '0') < v * 10)
-				BOOST_THROW_EXCEPTION(SemVerError());
+				BOOST_THROW_EXCEPTION(SemVerError() << util::errinfo_comment("Invalid token in version pragma: "s + (c ? c : '\0')));
```
nikola-matic: Not sure if the `NULL` check for `c` is necessary here, and if yes, what example `pragma ...` statement would cause this to hit the null check branch, and if there is one, it would be nice to have a test for it.

matheusaaguiar: Yes. I agree that it is not needed. The value returned from `currentChar` and assigned to `c` is checked in the loop condition (line `270`) and should never be `0`.

matheusaaguiar: Then, I think the second part of the condition of the `if`, `v * 10 + static_cast<unsigned>(c - '0') < v * 10` should always be false.

matheusaaguiar: Also thinking that the whole `if` condition of line `273` is unnecessary... 
 - `v` is initialized in line `268` with a value in the range `[1, 9]`, guaranteed by the condition of the `if` in line `266`. So `v * 10 < v` should be false the first time the loop runs.
 - the loop condition always checks that `currentChar` is part of the same token and that it is in the range `[0,9]`, which when added to `v` should never make its value decrease. Then, `v * 10 + static_cast<unsigned>(c - '0') < v * 10` should always be false in every loop iteration. 
 - from the previous,  `v * 10 < v` shall be false in loop iterations after the first.

Is that correct?

nikola-matic: If in doubt, `solAssert` :)

matheusaaguiar: Anyway, I think the possible removal of the `if` block should be done in a separate PR. For now, I will just remove the ternary operator checking `NULL` for `c`.

nikola-matic:
@matheusaaguiar another rebase pleae :D

liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +271,16 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
...
 			v = v * 10 + static_cast<unsigned>(c - '0');
 			nextChar();
 		}
 		return v;
 	}
+	else if (c == char(-1))
+		BOOST_THROW_EXCEPTION(SemVerError() << util::errinfo_comment("Expected version pragma number but reached end of text"));
 	else
-		BOOST_THROW_EXCEPTION(SemVerError());
+		BOOST_THROW_EXCEPTION(SemVerError() << util::errinfo_comment("Invalid token in version pragma: "s + (c ? c : '\0')));
```
nikola-matic:
```suggestion
		BOOST_THROW_EXCEPTION(SemVerError() << util::errinfo_comment("Invalid token in version pragma: "s + c));
```

liblangutil/SemVerHandler.h
```diff
@@ -87,8 +89,8 @@ class SemVerMatchExpressionParser
 public:
 	SemVerMatchExpressionParser(std::vector<Token> _tokens, std::vector<std::string> _literals);
 
-	/// Returns an expression if it was parseable, or nothing otherwise.
-	std::optional<SemVerMatchExpression> parse();
+	/// Returns an expression if it was parseable, or an error otherwise.
```
nikola-matic:
```suggestion
	/// Returns an expression if it was parsable, or an error otherwise.
```

liblangutil/SemVerHandler.cpp
```diff
@@ -52,7 +52,7 @@ SemVerVersion::SemVerVersion(string const& _versionString)
 		if (level < 2)
 		{
 			if (i == end || *i != '.')
-				BOOST_THROW_EXCEPTION(SemVerError());
+				BOOST_THROW_EXCEPTION(SemVerError() << util::errinfo_comment("Invalid versionString: "s + _versionString));
```
nikola-matic: I'd add `using namespace solidity::util` up top, and have this be `errinfo_comment` instead of `util::errinfo_comment`.

matheusaaguiar:
Rebased!

nikola-matic:
Hmm, looks good now, but I'm wondering whether this will need a changelog entry due to the changed error messages? @ekpyron

matheusaaguiar:
I have defaulted to a 'yes' and added an entry, as recommended in the other PR.

matheusaaguiar:
Rebased.

liblangutil/SemVerHandler.cpp
```diff
@@ -171,14 +172,20 @@ optional<SemVerMatchExpression> SemVerMatchExpressionParser::parse()
 			if (m_pos >= m_tokens.size())
 				break;
 			if (currentToken() != Token::Or)
-				BOOST_THROW_EXCEPTION(SemVerError());
+			{
+				string tokenString = TokenTraits::toString(currentToken());
+				BOOST_THROW_EXCEPTION(
+					SemVerError() << errinfo_comment("Invalid token or "
+					"more than one version not delimited by || token: "s + tokenString)
```
cameel:
```suggestion
					SemVerError() << errinfo_comment(
						"Invalid token or more than one version not delimited by || token: "s + tokenString
					)
```

liblangutil/SemVerHandler.h
```diff
@@ -87,8 +89,8 @@ class SemVerMatchExpressionParser
 public:
 	SemVerMatchExpressionParser(std::vector<Token> _tokens, std::vector<std::string> _literals);
 
-	/// Returns an expression if it was parseable, or nothing otherwise.
-	std::optional<SemVerMatchExpression> parse();
+	/// Returns an expression if it was parsable, or an error otherwise.
+	std::variant<SemVerError, SemVerMatchExpression> parse();
```
cameel: Catching a `SemVerError` instance and then passing it around seems bizarre to me. It's an exception - why don't you throw it?

And for cases where we do want to return a message in case of failure, we typically use [`Result`](https://github.com/ethereum/solidity/blob/v0.8.17/libsolutil/Result.h#L39-L84).

test/libsolidity/syntaxTests/pragma/broken_version_7.sol
```diff
@@ -1,3 +1,3 @@
 pragma solidity >0.5.0<;
 // ----
-// ParserError 1684: (0-24): Found version pragma, but failed to parse it. Please ensure there is a trailing semicolon.
+// ParserError 1684: (0-24): Expected version pragma number but reached end of text
```
cameel:
```suggestion
// ParserError 1684: (0-24): Expected version number but reached end of pragma.
```

test/libsolidity/syntaxTests/pragma/empty_version.sol
```diff
@@ -1,3 +1,3 @@
 pragma solidity;
 // ----
-// ParserError 1684: (0-16): Found version pragma, but failed to parse it. Please ensure there is a trailing semicolon.
+// ParserError 1684: (0-16): Empty pragma version
```
cameel:
```suggestion
// ParserError 1684: (0-16): Empty version pragma.
```

test/libsolidity/syntaxTests/pragma/unterminated_version.sol
```diff
@@ -1,4 +1,4 @@
 pragma solidity 0.4.3
 pragma abicoder v2;
 // ----
-// ParserError 1684: (0-41): Found version pragma, but failed to parse it. Please ensure there is a trailing semicolon.
+// ParserError 1684: (0-41): Invalid token in version pragma: p
```
cameel: Well, the `p` here is technically not a token. Or at least calling it that is confusing because normally `pragma` as a whole would be a token.

The hint about the semicolon is also still useful in this case.

How about this?
```suggestion
// ParserError 1684: (0-41): Expected the start of a version number. Found character 'p'. Version number is invalid or the pragma is not terminated with a semicolon.
```

libsolidity/parsing/Parser.cpp
```diff
@@ -160,17 +160,19 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
...
-	if (!matchExpression.has_value())
+	variant<SemVerError, SemVerMatchExpression> matchExpressionOrError = parser.parse();
+	if (SemVerError const* matchExpressionError = get_if<SemVerError>(&matchExpressionOrError))
+	{
 		m_errorReporter.fatalParserError(
 			1684_error,
 			_location,
-			"Found version pragma, but failed to parse it. "
-			"Please ensure there is a trailing semicolon."
+			matchExpressionError->what()
```
cameel: Since all these errors have the same error code, I think they should have a common prefix. For example:
```suggestion
			"Invalid version pragma. " + matchExpressionError->what()
```

liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +272,16 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
 		{
 			c = currentChar();
 			if (v * 10 < v || v * 10 + static_cast<unsigned>(c - '0') < v * 10)
-				BOOST_THROW_EXCEPTION(SemVerError());
+				BOOST_THROW_EXCEPTION(SemVerError() << errinfo_comment("Invalid token in version pragma: "s + c));
```
cameel:
```suggestion
				BOOST_THROW_EXCEPTION(SemVerError() << errinfo_comment("Integer too large to be used in a version number."));
```
You might also include the current value of `v`, though that's not the complete integer so it might be misleading...

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -58,17 +58,35 @@ SemVerMatchExpression parseExpression(string const& _input)
...
+			"Expression \"" + _input + "\" did not parse properly."
+		);
+		return expressionOrError;
+	}
+	return expressionOrError;
 }
 
 }
 
+BOOST_AUTO_TEST_CASE(check_exceptions)
```
cameel:
```suggestion
BOOST_AUTO_TEST_CASE(exception_on_invalid_version_in_semverversion_constructor)
```

liblangutil/SemVerHandler.cpp
cameel: Also, there seems to be no test case covering this.

liblangutil/SemVerHandler.cpp
cameel: This message should be more precise. As is, it may be confusing because normally we do not require `||` between versions - if you do not use it it's just interpreted as `&&` instead and it's perfectly fine. The one situation where that's not the case is a range of versions (e.g. `0.8.3 - 0.8.17`). Looking at `parseMatchExpression()`, I see we can have multiple versions next to each other but not multiple ranges. Or a range next to a version.

Not sure how to phrase it though. This would be precise, though it still does not explain why:
```c++
"Invalid token: "s + tokenString + ". Note that conjunction is not allowed at this position."
```

Maybe we can come up with something better.

To be honest, this is a weird limitation. I don't see why conjunction would have to be disallowed here.

cameel: Please also add a test case covering this. Currently we don't seem to have any.

cameel (changes requested):

liblangutil/SemVerHandler.h
cameel: This does not seem resolved yet.

liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +272,16 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
 		{
 			c = currentChar();
 			if (v * 10 < v || v * 10 + static_cast<unsigned>(c - '0') < v * 10)
```
cameel, line 274: By the way, since `v` is `unsigned`, this will result in different limits on different platforms. We should use a type of a fixed size here, e.g. `uint32_t`. Maybe we should even consider this a bug.

I would not change it in this PR though. I guess it's actually a breaking change, even if trivial. Probably better to just report it as an issue for now.

liblangutil/SemVerHandler.cpp
cameel: > Then, I think the second part of the condition of the `if`, `v * 10 + static_cast<unsigned>(c - '0') < v * 10` should always be false.

Not really. The `while` only ensures that `c` is a character between `'0'` and `'9'`. It does not ensure that adding the character to the number does not cause overflow in `v`.

I think that you can hit this case with this example:
```solidity
pragma solidity 4294967296;
```

`4294967296` is `2**32`. When you're at the final `6`, `v` is `429496729`. Multiplying it by `10` gives you `4294967290`, which still fits in `unsigned` and does not overflow so the first part of the `if` is false. But then adding `6` to that brings it over the edge so the second part of the `if` is true.

liblangutil/SemVerHandler.cpp
cameel: Reported as #13706.

liblangutil/SemVerHandler.cpp
matheusaaguiar: Right, at the time I didn't think about extreme values and overflow.

Changelog.md
```diff
@@ -4,13 +4,13 @@ Language Features:
...
- * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
- * Language Server: Add basic document hover support.
- * Optimizer: Added optimization rule ``and(shl(X, Y), shl(X, Z)) => shl(X, and(Y, Z))``.
+* Commandline Interface: Return exit code ``2`` on uncaught exceptions.
+* Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the bytecode.
+* Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
+* Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
+* Language Server: Add basic document hover support.
+* Optimizer: Added optimization rule ``and(shl(X, Y), shl(X, Z)) => shl(X, and(Y, Z))``.
+* Parser: Improved error messages about the version pragma.
```
cameel: There should be a 1-char indent here. #13692 fixed it and I think you kept the old version here when rebasing.

liblangutil/SemVerHandler.cpp
```diff
@@ -265,14 +276,21 @@ unsigned SemVerMatchExpressionParser::parseVersionPart()
...
 	}
+	else if (c == char(-1))
+		BOOST_THROW_EXCEPTION(SemVerError() << errinfo_comment("Expected version number but reached end of pragma."));
 	else
-		BOOST_THROW_EXCEPTION(SemVerError());
+		BOOST_THROW_EXCEPTION(
+			SemVerError() << errinfo_comment(
+				fmt::format("Expected the start of a version number but instead found character '{}'. "
+				"Version number is invalid or the pragma is not terminated with a semicolon.", c)
+			)
```
cameel:
```suggestion
			SemVerError() << errinfo_comment(fmt::format(
				"Expected the start of a version number but instead found character '{}'. "
				"Version number is invalid or the pragma is not terminated with a semicolon.",
				c
			))
```

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -153,22 +153,28 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
...
+				m_errorReporter.syntaxError(
+					3997_error,
+					_pragma.location(),
+					"Source file requires different compiler version (current compiler is " +
+					string(VersionString) + ") - note that nightly builds are considered to be "
+					"strictly less than the released version"
+				);
+			m_versionPragmaFound = true;
+		}
+		catch(SemVerError const&)
```
cameel:
```suggestion
		catch (SemVerError const&)
```

libsolidity/parsing/Parser.cpp
```diff
@@ -160,27 +160,31 @@ ASTPointer<SourceUnit> Parser::parse(CharStream& _charStream)
...
+			if (!m_parserErrorRecovery)
+				m_errorReporter.fatalParserError(
+					5333_error,
+					_location,
+					"Source file requires different compiler version (current compiler is " +
+					string(VersionString) + ") - note that nightly builds are considered to be "
+					"strictly less than the released version"
+				);
+	}
+	catch(SemVerError const& matchError)
```
cameel:
```suggestion
	catch (SemVerError const& matchError)
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -159,11 +175,19 @@ BOOST_AUTO_TEST_CASE(positive_range)
...
+			SemVerMatchExpression matchExpression = parseExpression(t.first);
+			BOOST_CHECK_MESSAGE(
+				matchExpression.matches(version),
+				"Version \"" + t.second + "\" did not satisfy expression \"" + t.first + "\""
+			);
+		}
+		catch(SemVerError const& matchError)
+		{
+
+		}
```
cameel: You should assert here that it does not throw. You can also move both of these handlers into `parseExpression()`.

liblangutil/SemVerHandler.cpp
cameel: So, after discussing this with @matheusaaguiar and @nikola-matic, let's make the message here something like "You can only combine version ranges using the || operator."

test/libsolidity/SemVerMatcher.cpp
matheusaaguiar: Moved to `parseExpression()`.

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -159,11 +175,19 @@ BOOST_AUTO_TEST_CASE(positive_range)
...
-		);
+		try
+		{
+			SemVerMatchExpression matchExpression = parseExpression(t.first);
+			BOOST_CHECK_MESSAGE(
+				matchExpression.matches(version),
+				"Version \"" + t.second + "\" did not satisfy expression \"" + t.first + "\""
+			);
+		}
+		catch(SemVerError const& matchError)
```
nikola-matic: Surprised the compiler is not complaining against the unused `matchError`.

```suggestion
		catch(SemVerError const&)
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -235,12 +259,19 @@ BOOST_AUTO_TEST_CASE(negative_range)
...
+		try
+		{
+			auto matchExpression = parseExpression(t.first);
+			BOOST_CHECK_MESSAGE(
+				!matchExpression.matches(version),
+				"Version \"" + t.second + "\" did satisfy expression \"" + t.first + "\" " +
+				"(although it should not)"
+			);
+		}
+		catch(SemVerError const& matchError)
```
nikola-matic:
```suggestion
		catch(SemVerError const&)
```

test/libsolidity/SemVerMatcher.cpp
matheusaaguiar: This was my bad. I committed this without a proper check.

nikola-matic: While you're at it, can you add a comment like `// Ignored` in the `catch` block as well?

Changelog.md
```diff
@@ -12,6 +12,7 @@ Compiler Features:
  * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
  * Language Server: Add basic document hover support.
  * Optimizer: Added optimization rule ``and(shl(X, Y), shl(X, Z)) => shl(X, and(Y, Z))``.
+ * Parser: Improved error messages about the version pragma.
```
```suggestion
 * Parser: More detailed error messages about invalid version pragmas.
```

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -153,22 +153,28 @@ bool SyntaxChecker::visit(PragmaDirective const& _pragma)
...
+					"Source file requires different compiler version (current compiler is " +
+					string(VersionString) + ") - note that nightly builds are considered to be "
+					"strictly less than the released version"
+				);
+			m_versionPragmaFound = true;
+		}
+		catch (SemVerError const&)
+		{
+			// An unparsable version pragma is an unrecoverable fatal error in the parser.
+			solAssert(false, "");
```
cameel:
```suggestion
			solAssert(false);
```

test/libsolidity/SemVerMatcher.cpp
```diff
@@ -58,15 +59,39 @@ SemVerMatchExpression parseExpression(string const& _input)
...
+			matchExpression.isValid(),
+			"Expression \"" + _input + "\" did not parse properly."
+		);
+
+		return matchExpression;
+	}
+	catch (SemVerError const&)
+	{
+		// Ignored, since a test case should have a parsable version
+		soltestAssert(false, "");
```
cameel:
```suggestion
		soltestAssert(false);
```

test/libsolidity/syntaxTests/pragma/missing_disjunction_token.sol
```diff
@@ -0,0 +1,3 @@
+pragma solidity 0.4.1 - 0.6.2 0.8.17;
```
cameel: Can you add also these?
- `pragma solidity 0.8.17 0.4.1 - 0.6.2;` (invalid)
- `pragma solidity 0.8.17 || 0.4.1 - 0.6.2;` (valid)
- `pragma solidity 0.4.0 - 0.4.1 0.4.1 - 0.6.2;` (invalid)
- `pragma solidity 0.4.0 - 0.4.1 || 0.4.1 - 0.6.2;` (valid)

test/libsolidity/syntaxTests/pragma/empty_version.sol
```diff
@@ -1,3 +1,3 @@
 pragma solidity;
```
cameel, line 1: Can you also add a case for this: `pragma solidity ;` (i.e. a space after `solidity`)? Should trigger the same message.

cameel (commented):
This is close to being done but I'd still like a few more tests and some small tweaks. See my comments below.

matheusaaguiar:
@cameel I have added the tests and made the changes you suggested. 
Also rebased.

Changelog.md
```diff
@@ -15,7 +15,7 @@ Compiler Features:
  * Optimizer: Added optimization rule ``and(shl(X, Y), shl(X, Z)) => shl(X, and(Y, Z))``.
  * SMTChecker: Support Eldarica as a Horn solver for the CHC engine when using the CLI option ``--model-checker-solvers eld``. The binary `eld` must be available in the system.
  * SMTChecker: Make ``z3`` the default solver for the BMC and CHC engines instead of all solvers.
-
+ * Parser: More detailed error messages about invalid version pragmas.
 
 Bugfixes:
```
cameel: There should be 2 empty lines between sections.

test/libsolidity/syntaxTests/pragma/invalid_range_conjunction_1.sol
```diff
@@ -0,0 +1,3 @@
+pragma solidity 0.4.1 - 0.6.2 0.8.17;
```
cameel: Please give these tests more detailed names. Avoid just adding numbers - we have a lot of that and it makes it harder to see which cases we have covered. It's ok to leave existing ones as is but new ones should not use that convention.

I'd suggest naming it like this:
- `invalid_range_conjunction_1.sol` - > `invalid_range_conjunction_range_first.sol`
- `invalid_range_conjunction_2.sol` - > `invalid_range_conjunction_range_second.sol`
- `invalid_range_conjunction_3.sol` - > `invalid_range_conjunction_two_ranges.sol`
- `valid_range_disjunction_1.sol` -> `valid_range_disjunction_range_second.sol`
- `valid_range_disjunction_2.sol` -> `valid_range_disjunction_two_ranges.sol`

cameel (dismissed):
Approving but needs some cosmetic tweaks.

cameel:
Also review fixes should be squashed.

test/libsolidity/syntaxTests/pragma/invalid_range_conjunction_1.sol
matheusaaguiar: Right. I will remember that for future PRs.

cameel (approved):

PR #12798, opened 2022-03-15
LSP.py: Implement expectations directly in the test files

Marenz:
This is a rough draft that is already usable. Not all goto tests where converted yet.

I plan to make the code a lot nicer and convert all goto tests and others that are feasable.

### DIFF
--- a/test/libsolidity/lsp/goto_definition.sol
+++ b/test/libsolidity/lsp/goto_definition.sol
@@ -2,13 +2,16 @@
 pragma solidity >=0.8.0;
 
 import "./lib.sol";
+//      ^ @importDirective
 
 interface I
 {
     function f(uint x) external returns (uint);
+    //       ^ @functionF
 }
 
 contract IA is I
+    //   ^^ @IASymbol
 {
     function f(uint x) public pure override returns (uint) { return x + 1; }
 }
@@ -21,6 +24,7 @@ contract IB is I
 library IntLib
 {
     function add(int self, int b) public pure returns (int) { return self + b; }
+    //       ^^^ @IntLibAdd
 }
 
 contract C
@@ -29,40 +33,175 @@ contract C
     function virtual_inheritance() public payable
     {
         obj = new IA();
+        //        ^ @usingIASymbol
         obj.f(1); // goto-definition should jump to definition of interface.
+    //      ^ @virtualFunctionLookup
     }
 
     using IntLib for *;
     function using_for(int i) pure public
     {
         i.add(5);
+ //       ^ @usingIntAdd
         14.add(4);
     }
 
     function useLib(uint n) public payable returns (uint)
     {
         return Lib.add(n, 1);
+        //     ^ @LibSymbol
+        //         ^ @LibAddSymbol
     }
 
     function enums(Color c) public pure returns (Color d)
+    //             ^ @ColorSymbolInParameter
     {
         Color e = Color.Red;
+        //    ^ @eVariableDeclaration
+        //              ^ @RedEnumMemberAccess
         if (c == e)
+        //       ^ @eVariableAccess
             d = Color.Green;
         else
             d = c;
     }
 
     type Price is uint128;
+    //   ^^^^^ @PriceDeclaration
     function udlTest() public pure returns (uint128)
     {
         Price p = Price.wrap(128);
+    //  ^ @PriceSymbol
+    //            ^ @PriceInWrap
         return Price.unwrap(p);
     }
 
     function structCtorTest(uint8 v) public pure returns (uint8 result)
     {
         RGBColor memory c = RGBColor(v, 2 * v, 3 * v);
+        //                       ^ @RGBColorCursor
         result = c.red;
+        int a;
+//      ^^^^^ @unusedLocalVar
     }
 }
+// ----
+// goto_definition: @unusedLocalVar 2072
+// lib: @diagnostics 2072
+// -> textDocument/definition {
+//     "position": @importDirective
+// }
+// <- [
+//     {
+//         "range": {
+//             "end": {
+//                 "character": 0,
+//                 "line": 0
+//             },
+//             "start": {
+//                 "character": 0,
+//                 "line": 0
+//             }
+//         },
+//         "uri": "lib.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @usingIASymbol
+// }
+// <- [
+//     {
+//         "range": @IASymbol,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @virtualFunctionLookup
+// }
+// <- [
+//     {
+//         "range": @functionF,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @usingIntAdd
+// }
+// <- [
+//     {
+//         "range": @IntLibAdd,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @LibSymbol
+// }
+// <- [
+//     {
+//         "range": @LibLibrary,
+//         "uri": "lib.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @LibAddSymbol
+// }
+// <- [
+//     {
+//         "range": @addSymbol,
+//         "uri": "lib.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @ColorSymbolInParameter
+// }
+// <- [
+//     {
+//         "range": @ColorEnum,
+//         "uri": "lib.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @RedEnumMemberAccess
+// }
+// <- [
+//     {
+//         "range": @EnumMemberRed,
+//         "uri": "lib.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @eVariableAccess
+// }
+// <- [
+//     {
+//         "range": @eVariableDeclaration,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @PriceSymbol
+// }
+// <- [
+//     {
+//         "range": @PriceDeclaration,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @PriceInWrap
+// }
+// <- [
+//     {
+//         "range": @PriceDeclaration,
+//         "uri": "goto_definition.sol"
+//     }
+// ]
+// -> textDocument/definition {
+//     "position": @RGBColorCursor
+// }
+// <- [
+//     {
+//         "range": @RGBColorStruct,
+//         "uri": "lib.sol"
+//     }
+// ]
--- a/test/libsolidity/lsp/lib.sol
+++ b/test/libsolidity/lsp/lib.sol
@@ -5,6 +5,7 @@ pragma solidity >=0.8.0;
 error E(uint, uint);
 
 enum Weather {
+//   ^^^^^^^ @whetherEnum
     Sunny,
     Cloudy,
     Rainy
@@ -12,8 +13,10 @@ enum Weather {
 
 /// Some custom Color enum type holding 3 colors.
 enum Color {
+//   ^^^^^ @ColorEnum
     /// Red color.
     Red,
+//  ^^^ @EnumMemberRed
     /// Green color.
     Green,
     /// Blue color.
@@ -21,9 +24,11 @@ enum Color {
 }
 
 library Lib
+//   @  ^^^ @LibLibrary
 {
     function add(uint a, uint b) public pure returns (uint result)
 // ^( @addFunction
+//           ^^^ @addSymbol
     {
         result = a + b;
     }
@@ -37,8 +42,11 @@ library Lib
 }
 
 struct RGBColor
+//     ^^^^^^^^ @RGBColorStruct
 {
     uint8 red;
     uint8 green;
     uint8 blue;
 }
+// ----
+// lib: @diagnostics 2072
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -8,12 +8,92 @@ import subprocess
 import sys
 import traceback
 import re
-
+import tty
+import functools
+from collections import namedtuple
+from copy import deepcopy
 from typing import Any, List, Optional, Tuple, Union
+from itertools import islice
+
 from enum import Enum, auto
 
 import colorama # Enables the use of SGR & CUP terminal VT sequences on Windows.
 from deepdiff import DeepDiff
+
+"""
+Named tuple that holds various regexes used to parse the test specification.
+"""
+TestRegexesTuple = namedtuple("TestRegexesTuple", [
+    "sendRequest", # regex to find requests to be sent & tested
+    "findQuotedTag", # regex to find tags wrapped in quotes
+    "findTag", # regex to find tags
+    "fileDiagnostics", # regex to find diagnostic expectations for a file
+    "diagnostic" # regex to find a single diagnostic within the file expectations
+])
+"""
+Instance of the named tuple holding the regexes
+"""
+TEST_REGEXES = TestRegexesTuple(
+    re.compile(R'^// -> (?P<method>[\w\/]+) {'),
+    re.compile(R'(?P<tag>"@\w+")'),
+    re.compile(R'(?P<tag>@\w+)'),
+    re.compile(R'// (?P<testname>\w+):[ ]?(?P<diagnostics>[\w @]*)'),
+    re.compile(R'(?P<tag>@\w+) (?P<code>\d\d\d\d)')
+)
+
+"""
+Named tuple holding regexes to find tags in the solidity code
+"""
+TagRegexesTuple = namedtuple("TagRegexestuple", ["simpleRange", "multilineRange"])
+TAG_REGEXES = TagRegexesTuple(
+    re.compile(R"(?P<range>[\^]+) (?P<tag>@\w+)"),
+    re.compile(R"\^(?P<delimiter>[()]{1,2}) (?P<tag>@\w+)$")
+)
+
+
+def count_index(lines, start=0):
+    """
+    Takes an iterable of lines and adds the current byte index so it's available
+    when iterating or looping.
+    """
+    n = start
+    for elem in lines:
+        yield n, elem
+        n += 1 + len(elem)
+
+def tags_only(lines, start=0):
+    """
+    Filter the lines for tag comments and report line number that tags refer to.
+    """
+    n = start
+    numCommentLines = 0
+
+    def hasTag(line):
+        if line.find("// ") != -1:
+            for _, regex in TAG_REGEXES._asdict().items():
+                if regex.search(line[len("// "):]) is not None:
+                    return True
+        return False
+
+    for line in lines:
+        if hasTag(line):
+            numCommentLines += 1
+            yield n - numCommentLines, line
+        else:
+            numCommentLines = 0
+
+        n += 1
+
+
+def prepend_comments(sequence):
+    """
+    Prepends a comment indicator to each element
+    """
+    result = ""
+    for line in sequence.splitlines(True):
+        result = result + "// " + line
+    return result
+
 
 # {{{ JsonRpcProcess
 class BadHeader(Exception):
@@ -59,7 +139,7 @@ class JsonRpcProcess:
         while True:
             # read header
             line = self.process.stdout.readline()
-            if line == '':
+            if len(line) == 0:
                 # server quit
                 return None
             line = line.decode("utf-8")
@@ -118,12 +198,30 @@ SGR_STATUS_OKAY = '\033[1;32m'
 SGR_STATUS_FAIL = '\033[1;31m'
 
 class ExpectationFailed(Exception):
-    def __init__(self, actual, expected):
-        self.actual = json.dumps(actual, sort_keys=True)
-        self.expected = json.dumps(expected, sort_keys=True)
-        diff = json.dumps(DeepDiff(actual, expected), indent=4)
+    class Part(Enum):
+        Diagnostics = auto()
+        Methods = auto()
+
+    def __init__(self, reason: str, part):
+        self.part = part
+        super().__init__(reason)
+
+class JSONExpectationFailed(ExpectationFailed):
+    def __init__(self, actual, expected, part):
+        self.actual = actual
+        self.expected = expected
+
+        expected_pretty = ""
+
+        if expected is not None:
+            expected_pretty = json.dumps(expected, sort_keys=True)
+
+        diff = DeepDiff(actual, expected)
+
         super().__init__(
-            f"\n\tExpected {self.expected}\n\tbut got {self.actual}.\n\t{diff}"
+            f"\n\tExpected {expected_pretty}" + \
+            f"\n\tbut got {json.dumps(actual, sort_keys=True)}.\n\t{diff}",
+            part
         )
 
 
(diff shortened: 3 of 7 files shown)


test/libsolidity/lsp/goto_definition.sol
```diff
@@ -29,40 +33,211 @@ contract C
...
+        //                       ^ @RGBColorCursor
         result = c.red;
+        int a;
+//      ^^^^^ @unusedLocalVar
     }
 }
+// ----
+// goto_definition: @unusedLocalVar 2072
+// lib: @diagnostics 2072
+// -> textDocument/definition {
```
christianparpart, line 91: Nice idea to use arrows to dictate direction :+1:

test/libsolidity/lsp/goto_definition.sol
```diff
@@ -29,40 +33,211 @@ contract C
...
+//             "uri": "lib.sol"
+//         }
+//     ]
+// }
+// -> textDocument/definition {
+//     "position": @usingIASymbol
+// }
+// <- {
+//     "jsonrpc": "2.0",
+//     "result": [
```
christianparpart: Forward message is slightly abstracted. Good.
Backward message is however bare, which is probably a little bit confusing. I'd suggest to only require passing the response value, such as I'm going to suggest below.

test/libsolidity/lsp/goto_definition.sol
```diff
@@ -29,40 +33,211 @@ contract C
...
+//         {
+//             "range": @functionF,
+//             "uri": "goto_definition.sol"
+//         }
+//     ]
+// }
```
christianparpart:
```suggestion
// <- [
//     { "range": @functionF, "uri": "goto_definition.sol"}
// ]
```

test/libsolidity/lsp/goto_definition.sol
```diff
@@ -29,40 +33,211 @@ contract C
...
+// <- {
+//     "jsonrpc": "2.0",
+//     "result": [
+//         {
+//             "range": @RGBColorStruct,
+//             "uri": "lib.sol"
+//         }
+//     ]
+// }
```
christianparpart: 243 lines, I think we agreed on splitting the file up into distinct test cases, such that this file doesn't explode too much in LoC's. What do you think?

test/lsp.py
```diff
@@ -8,13 +8,76 @@
...
+)
+
+
+def countIndex(sequence, start=0):
+    n = start
+    for elem in sequence:
+        yield n, elem
+        n += 1 + len(elem)
+
+# Filter the lines for tag comments and report line number that tags refer to
```
christianparpart: document via `"""` two lines below instead. :)

test/lsp.py
```diff
@@ -39,8 +106,37 @@ def __enter__(self):
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE
         )
+        self.async_read_loop()
         return self
 
+    # Reads a message without blocking
+    def read_message(self, peek=False, block=True):
```
christianparpart:
```suggestion
    def read_message(self, peek: bool =False, block: bool =True):
```

test/lsp.py
```diff
@@ -173,15 +273,146 @@ class Counter:
...
 # Returns the given marker with the end extended by 'amount'
 def extendEnd(marker, amount=1):
-    marker["end"]["character"] += amount
-    return marker
+    newMarker = deepcopy(marker)
+    newMarker["end"]["character"] += amount
+    return newMarker
+
+class ReplacingTag:
+    def __init__(self, tagName, tagRange):
```
christianparpart: type annotations?

christianparpart (changes requested):

test/lsp.py
```diff
@@ -8,13 +8,76 @@
...
+        if hasTag(line):
+            numCommentLines += 1
+            yield n - numCommentLines, line
+        else:
+            numCommentLines = 0
+
+        n += 1
+
+
+def preceedComments(sequence):
```
aarlt: Just scrolling through the code. This is really cool PR! What is `preceed`? Is that a typo?

Marenz: No, that's deliberate: https://www.thefreedictionary.com/preceed "to put in front" "to put before" basically :)

test/libsolidity/lsp/goto_definition.sol
Marenz: I mean, that's up to the test file creator, either works with the code :)

test/libsolidity/lsp/goto_definition.sol
Marenz: I can probably easily remove the jsonrpc, but for this stage/PR I wouldn't go much further with minimalizing the responses. They are also automatically created, so it's not like you ahve to type them

test/lsp.py
aarlt: Hehe :) Ah! Looks like that there are different (correct) ways to write that. I thought it is written `precede`. Maybe its just British vs. American English.

Marenz: God damn, you're right. I was so sure.

test/lsp.py
Marenz: dead code ;)

test/lsp.py
```diff
@@ -118,6 +214,10 @@ def send_notification(self, name: str, params: Optional[dict] = None) -> None:
 SGR_STATUS_FAIL = '\033[1;31m'
```
aarlt, line 198: Interesting! I only knew those as ANSI escape sequences, so I thought `ANSI` as prefix would be better. But after checking what `SGR` really means, it seem to be more accurate :)

aarlt:
I was wondering how this script is actually called. So I was using grep and found this one 
`"$REPO_ROOT/scripts/test_solidity_lsp.py" "${SOLIDITY_BUILD_DIR}/solc/solc"` in `scripts/tests.sh`. But this script seem not exist.

test/lsp.py
christianparpart: @aarlt while these are ANSI VT sequences, the specific term here is actually SGR for "Select Graphics Rendition", as these VT sequences are to change the graphics rendition of the painting ANSI text cursor. :-)

christianparpart:
> I was wondering how this script is actually called. So I was using grep and found this one `"$REPO_ROOT/scripts/test_solidity_lsp.py" "${SOLIDITY_BUILD_DIR}/solc/solc"` in `scripts/tests.sh`. But this script seem not exist.

`./test/lsp.py ./build/solc/solc` - given that your local solc is at `./build/solc/solc` as it is the binary to be spawned for each test case and assuming you are in project root.

Marenz:
That still leaves the question what 
`"$REPO_ROOT/scripts/test_solidity_lsp.py" "${SOLIDITY_BUILD_DIR}/solc/solc"` in `scripts/tests.sh.` actually is and does :D

christianparpart:
> That still leaves the question what `"$REPO_ROOT/scripts/test_solidity_lsp.py" "${SOLIDITY_BUILD_DIR}/solc/solc"` in `scripts/tests.sh.` actually is and does :D

the script got renamed to `lsp.py` but where this script got referenced, apparently not. :)

test/libsolidity/lsp/lib.sol
```diff
@@ -5,25 +5,30 @@ pragma solidity >=0.8.0;
 error E(uint, uint);
 
 enum Weather {
+//   ^^^^^^^ @whetherEnum
```
christianparpart, line 8: style breakage. sometimes lowerCamel sometimes UpperCammels :)

test/lsp.py
```diff
@@ -8,13 +8,77 @@
...
+)
+
+TagRegexesTuple = namedtuple("TagRegexestuple", ["simpleRange", "multilineRange"])
+TAG_REGEXES = TagRegexesTuple(
+    re.compile(R"(?P<range>[\^]+) (?P<tag>@\w+)"),
+    re.compile(R"\^(?P<delimiter>[()]{1,2}) (?P<tag>@\w+)$")
+)
+
+
+def countIndex(sequence, start=0):
```
christianparpart:
```suggestion
def countIndex(sequence: YourTypeHere, start=0):
```
I'd also prefer a little documentation :)

test/lsp.py
```diff
@@ -8,13 +8,77 @@
...
+    n = start
+    for elem in sequence:
+        yield n, elem
+        n += 1 + len(elem)
+
+def tagsOnly(lines, start=0):
+    """
+    Filter the lines for tag comments and report line number that tags refer to
```
christianparpart:
```suggestion
    Filter the lines for tag comments and report line number that tags refer to.
```

test/lsp.py
```diff
@@ -26,11 +90,15 @@ class JsonRpcProcess:
     exe_args: List[str]
     process: subprocess.Popen
     trace_io: bool
+    messages: queue.Queue
+    last_message: None
```
christianparpart: type?

test/lsp.py
```diff
@@ -39,8 +107,45 @@ def __enter__(self):
...
+        for _ in range(num):
+            try:
+                self.read_message(False, block)
+            except queue.Empty:
+                return
+
+
+    # Reads a message without blocking
+    def read_message(self, peek: bool = False, block: bool = True):
+#        print("read_message(" + str(peek) + ", " + str(block))
```
christianparpart:
```suggestion
    def read_message(self, peek: bool = False, block: bool = True):
        """
        Reads a message without blocking
        """

File: test/lsp.py [hunk 15]
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+
+class TestParser:
+    RequestAndResponse = namedtuple('RequestAndResponse',
+        "method, request, response, responseBegin, responseEnd",
+        defaults=(None, None, None, None)
+    )
+    Diagnostics = namedtuple('Diagnostics', 'tests start end has_header')
+    Diagnostic = namedtuple('Diagnostic', 'marker code')
+
+    TEST_START = "// ----"
```
christianparpart, line 313:
```suggestion
    TEST_START = "\n// ----\n"
```
or similar. should be required to stand at the beginning of a line and the **only** content of that line.

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+
+    def __init__(self, content: str):
+        self.content = content
+        self.lines = None
+        self.currentLine = None
+
+    def parse(self):
+        testDefStartIdx = self.content.rfind(self.TEST_START)
```
christianparpart: with the above change, no need to rfind (just raises questions) :)

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+    Diagnostic = namedtuple('Diagnostic', 'marker code')
+
+    TEST_START = "// ----"
+
+    def __init__(self, content: str):
+        self.content = content
+        self.lines = None
+        self.currentLine = None
+
+    def parse(self):
```
christianparpart, line 320:
```suggestion
    def parse(self):
    """
    document me :)
    """
```

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+            if self.currentLine is None:
+                raise TestParserException(ret, "Response footer not found")
+
+        return ret
+
+    def nextLine(self):
+        self.currentLine = next(self.lines, None)
+
+    def line(self):
+        return self.currentLine[1]
```
christianparpart:
```suggestion
    def currentLine[(self):
        return self.currentLineTuple[1]
```

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+
+    # Returns current byte position
+    def position(self):
+        if self.currentLine is None:
+            return len(self.content)
+        return self.currentLine[0]
+
+class FileTestRunner:
```
christianparpart, line 462:
```suggestion
class FileTestRunner:
    """
    fill me :)
    """
```

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+        self.test_name = test_name
+        self.suite = suite
+        self.solc = solc
+        self.open_tests = []
+        self.content = self.suite.get_test_file_contents(self.test_name)
+        self.markers = self.suite.get_file_tags(self.test_name)
+        self.parsed_testcases = None
+        self.expected_diagnostics = None
+
+    def test_diagnostics(self):
```
christianparpart, line 485: can we split this function into smaller ones? for easier review/maintenance?

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+                    )
+
+            return True
+        except:
+            self.close_open_files()
+            raise
+
+    def close_open_files(self):
+        # Do a non-blocking test read to make sure we didn't receive more
+        # diagnostics than expected
```
christianparpart:
```suggestion
    def close_all_open_files(self):
        """
        Do a non-blocking test read to make sure we didn't receive more
        diagnostics than expected
        """
```

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+            raise ExpectationFailed("Received more diagnostics than expected!")
+        except queue.Empty:
+            pass
+
+        for test in self.open_tests:
+            self.solc.send_message(
+                'textDocument/didClose',
+                { 'textDocument': { 'uri': self.suite.get_test_file_uri(test) }}
+            )
+            print(f"closing: {test}")
```
christianparpart:
```suggestion
```
or a debug-print facility? :)

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+
+
+        self.solc.clear_received_messages(len(self.open_tests), True)
+        self.open_tests.clear()
+
+    def test_methods(self):
+        try:
+            # Now handle each request/response pair in the test definition
+            for testcase in self.parsed_testcases:
```
christianparpart, line 561: this is a huge for-loop-body. can it be factored out?

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+                'textDocument/didClose',
+                { 'textDocument': { 'uri': self.suite.get_test_file_uri(test) }}
+            )
+            print(f"closing: {test}")
+
+
+        self.solc.clear_received_messages(len(self.open_tests), True)
+        self.open_tests.clear()
+
+    def test_methods(self):
```
christianparpart:
```suggestion
    def test_methods(self):
        """"
        document me :)
        """"
```

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+                    actualResponsePretty = self.suite.replace_ranges_with_tags(actualResponseJson)
+
+                    if expectedResponseJson is None:
+                        print("Failed to parse expected response, received: \n" + actualResponsePretty)
+                    else:
+                        print("Expected:\n" + \
+                            self.suite.replace_ranges_with_tags(expectedResponseJson) + \
+                            "\nbut got:\n" + actualResponsePretty
+                        )
+                    print("(u)pdate/(r)etry/(i)gnore?")
```
christianparpart: interactive part into factored out method?

test/lsp.py
```diff
@@ -180,15 +288,332 @@ class Counter:
...
+            return True
+        except TestParserException as e:
+            print(e)
+            print(e.result)
+            raise
+        finally:
+            self.close_open_files()
+
+
+    def parse_json_with_tags(self, content, markersFallback):
```
christianparpart, line 641: probably document what we've just talked about? :)

test/lsp.py
```diff
@@ -322,17 +742,72 @@ def wait_for_diagnostics(self, solc: JsonRpcProcess, count: int) -> List[dict]:
         Return `count` number of published diagnostic reports sorted by file URI.
         """
         reports = []
+
+#        import pdb; pdb.set_trace()
```
christianparpart:
```suggestion
```

test/lsp.py
```diff
@@ -322,17 +742,72 @@ def wait_for_diagnostics(self, solc: JsonRpcProcess, count: int) -> List[dict]:
...
+
+    def update_diagnostics_in_file(
+        self,
+        solc: JsonRpcProcess,
+        test,
+        content,
+        current_diagnostics:
+        TestParser.Diagnostics
+    ):
+        content = content[:current_diagnostics.start] + \
```
christianparpart, line 867: maybe use real if statements (not expression) here to make the construction more readable?

test/lsp.py
```diff
@@ -446,44 +921,79 @@ def expect_goto_definition_location(
...
 
-    # {{{ actual tests
-    def test_publish_diagnostics_warnings(self, solc: JsonRpcProcess) -> None:
-        self.setup_lsp(solc)
-        TEST_NAME = 'publish_diagnostics_1'
-        published_diagnostics = self.open_file_and_wait_for_diagnostics(solc, TEST_NAME)
 
-        self.expect_equal(len(published_diagnostics), 1, "One published_diagnostics message")
-        report = published_diagnostics[0]
+    def find_tag_with_range(self, test, target_range):
```
christianparpart, line 1008: docstring

test/lsp.py
```diff
@@ -446,44 +921,79 @@ def expect_goto_definition_location(
...
 
-        markers = self.get_file_tags(TEST_NAME)
+        return None
 
-        self.expect_equal(len(diagnostics), 3, "3 diagnostic messages")
-        self.expect_diagnostic(diagnostics[0], code=6321, marker=markers["@unusedReturnVariable"])
-        self.expect_diagnostic(diagnostics[1], code=2072, marker=markers["@unusedVariable"])
-        self.expect_diagnostic(diagnostics[2], code=2072, marker=markers["@unusedContractVariable"])
+    def replace_ranges_with_tags(self, content):
+        # Replace matching ranges with "@<tagname>"
```
christianparpart:
```suggestion
        """ Replace matching ranges with "@<tagname>" """
```

test/lsp.py
```diff
@@ -446,44 +921,79 @@ def expect_goto_definition_location(
...
+                        self.get_file_tags.cache_clear()
+                        continue
+                    elif userResponse == "r":
+                        print("retrying...")
+                        # pragma pylint: disable=no-member
+                        self.get_file_tags.cache_clear()
+                        return False
+                    elif userResponse == "i":
+                        print("ignoring...")
+                        return True
```
christianparpart: nested user communication. I think that would better fit into a sub-function.

test/lsp.py
```diff
@@ -632,9 +1141,43 @@ def verify_didOpen_with_import_diagnostics(
...
-        marker = self.get_file_tags('lib')["@diagnostics"]
+        markers = self.get_file_tags('lib')
+        marker = markers["@diagnostics"]
         self.expect_diagnostic(report['diagnostics'][0], code=2072, marker=marker)
 
+    def test_generic(self, solc: JsonRpcProcess) -> None:
+        self.setup_lsp(solc)
+
+
+        TESTS = ['goto_definition', 'goto_definition_imports', 'publish_diagnostics_1', 'publish_diagnostics_2', 'lib']
```
christianparpart: auto-construct me

christianparpart (changes requested):

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -171,6 +171,10 @@ void LanguageServer::compileAndUpdateDiagnostics()
 	{
 		Json::Value extra;
 		extra["openFileCount"] = Json::UInt64(diagnosticsBySourceUnit.size());
+		for (auto&& [name, val] : diagnosticsBySourceUnit)
+			extra["diag"][name] = "diag";
+		for (auto& name : m_openFiles)
```
stackenbotten: Coding style error

test/libsolidity/lsp/goto_definition.sol
```diff
@@ -29,40 +33,199 @@ contract C
...
+//             "range": @PriceDeclaration,
+//             "uri": "goto_definition.sol"
+//         }
+//     ]
+// }
+// -> textDocument/definition {
+//     "position": @PriceInWrap
+// }
+// <- {
+//     "result": [
```
christianparpart: Only keep child element of "result" :)

test/lsp.py
```diff
@@ -355,9 +880,9 @@ def open_file_and_wait_for_diagnostics(
                 }
             }
         )
-        return self.wait_for_diagnostics(solc_process, max_diagnostic_reports)
+        return self.wait_for_diagnostics(solc_process)
 
-    def expect_equal(self, actual, expected, description="Equality") -> None:
+    def expect_equal(self, actual, expected, description="Equality", part=ExpectationFailed.Part.Diagnostics) -> None:
```
christianparpart: coding style

test/lsp.py
```diff
@@ -446,44 +983,94 @@ def expect_goto_definition_location(
...
+                    self.update_diagnostics_in_file(solc, test, content, current_diagnostics)
+                    break
+                # pragma pylint: disable=broad-except
+                except Exception as e:
+                    print(e)
+                    if ret := self.user_interaction_failed_autoupdate(test):
+                        return ret
+        elif user_response == 's':
+            return True
+        else:
```
christianparpart: what if user types `'t'`?

christianparpart: loop until valid input?

test/lsp.py
```diff
@@ -446,44 +983,94 @@ def expect_goto_definition_location(
...
+            # pragma pylint: disable=no-member
+            self.get_file_tags.cache_clear()
+        elif user_response == "s":
+            print("skipping...")
 
-        self.expect_equal(len(diagnostics), 3, "3 diagnostic messages")
-        self.expect_diagnostic(diagnostics[0], code=9574, marker=markers["@conversionError"])
-        self.expect_diagnostic(diagnostics[1], code=6777, marker=markers["@argumentsRequired"])
-        self.expect_diagnostic(diagnostics[2], code=6160, marker=markers["@wrongArgumentsCount"])
+        return True
```
christianparpart, line 1097: what if user types something else than r/e/s?

test/lsp.py
```diff
@@ -526,7 +1113,7 @@ def test_textDocument_didOpen_with_relative_import(self, solc: JsonRpcProcess) -
         marker = self.get_file_tags("lib")["@diagnostics"]
         self.expect_diagnostic(report['diagnostics'][0], code=2072, marker=marker)
 
-
+    @functools.lru_cache # pragma pylint: disable=lru-cache-decorating-method
```
christianparpart, line 1144: Nice!

test/lsp.py
```diff
@@ -632,9 +1217,42 @@ def verify_didOpen_with_import_diagnostics(
...
 
-        marker = self.get_file_tags('lib')["@diagnostics"]
+        markers = self.get_file_tags('lib')
+        marker = markers["@diagnostics"]
         self.expect_diagnostic(report['diagnostics'][0], code=2072, marker=marker)
 
+    def test_generic(self, solc: JsonRpcProcess) -> None:
+        self.setup_lsp(solc)
+
+        STATIC_TESTS = ['didChange_template', 'didOpen_with_import', 'publish_diagnostics_3']
```
christianparpart, line 1255: still hard-coded?

christianparpart: ah! ok.

test/lsp.py
```diff
@@ -632,9 +1217,42 @@ def verify_didOpen_with_import_diagnostics(
...
+
+        tests = filter(
+            lambda x: x not in STATIC_TESTS,
+            map(lambda x: x[:-len(".sol")], os.listdir(self.project_root_dir))
```
christianparpart, line 1259: what about sub directories?

christianparpart (changes requested):

test/lsp.py
christianparpart: if that's so complex, amybe next time just use what is well used already (prepend vs append) :)

test/lsp.py
```diff
@@ -180,15 +282,411 @@ class Counter:
...
+            for diagnostics in published_diagnostics:
+                testname = diagnostics["uri"].replace(self.suite.project_root_uri + "/", "")[:-len(".sol")]
+
+                expected_diagnostics = tests[testname]
+                self.suite.expect_equal(
+                    len(diagnostics["diagnostics"]),
+                    len(expected_diagnostics),
+                    description="Unexpected amount of diagnostics"
+                )
+                lMarkers = self.suite.get_file_tags(testname)
```
christianparpart: no abbreviations :)

test/lsp.py
```diff
@@ -180,15 +282,411 @@ class Counter:
...
+                for diagnostic in diagnostics["diagnostics"]:
+                    expected_diagnostic = next((x for x in
+                        expected_diagnostics if diagnostic['range'] ==
+                        lMarkers[x.marker]), None)
```
christianparpart:
```suggestion
                    expected_diagnostic = next((diagnostic for diagnostic in
                        expected_diagnostics if diagnostic['range'] ==
                        lMarkers[diagnostic.marker]), None)
```

test/lsp.py
```diff
@@ -180,15 +282,411 @@ class Counter:
...
+            print(e)
+            raise
+        finally:
+            self.close_all_open_files()
+
+    def user_interaction_failed_method_test(self, testcase, actual, expected):
+        actual_pretty = self.suite.replace_ranges_with_tags(actual)
+
+        if expected is None:
+            print("Failed to parse expected response, received: \n" + actual)
```
christianparpart:
```suggestion
            print("Failed to parse expected response, received:\n" + actual)
```

test/lsp.py
```diff
@@ -180,15 +282,411 @@ class Counter:
...
+                with open(self.suite.get_test_file_path(self.test_name), mode="w", encoding="utf-8", newline='') as f:
+                    f.write(self.content)
+                return self.TestResult.Reparse
+            if user_response == "r":
+                return self.TestResult.Reparse
+
+            print("Invalid response.")
+
+
+    def run_testcase(self, testcase):
```
christianparpart:
```suggestion
    def run_testcase(self, testcase: Tuple[str, str]):
```
or similar - is it possible to specify the type? Otherwise it should be documented at least.

christianparpart (changes requested):

test/lsp.py
christianparpart: any updates?

christianparpart (commented):
The sub directories-question is still open. Is this working?

test/lsp.py
Marenz: We said we do those in another PR?

christianparpart (approved):

PR #12799, opened 2022-03-15
libevmasm: refactor asm-json export & add support for source list.

aarlt:
Needed for #11787.
Replaces part of #12704.

This PR refactors the asm-json export and adds support for source list (e.g. needed for srcmaps during asm-json import).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,9 +5,11 @@ Language Features:
 
 Compiler Features:
  * Peephole Optimizer: Remove operations without side effects before simple terminations.
+ * Assembly-Json: Export: Include source list in `sourceList` field.
 
 
 Bugfixes:
+* Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
 
 
 
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -222,122 +222,58 @@ string Assembly::assemblyString(
 	return tmp.str();
 }
 
-Json::Value Assembly::createJsonValue(string _name, int _source, int _begin, int _end, string _value, string _jumpType)
-{
-	Json::Value value{Json::objectValue};
-	value["name"] = _name;
-	value["source"] = _source;
-	value["begin"] = _begin;
-	value["end"] = _end;
-	if (!_value.empty())
-		value["value"] = _value;
-	if (!_jumpType.empty())
-		value["jumpType"] = _jumpType;
-	return value;
-}
-
-string Assembly::toStringInHex(u256 _value)
-{
-	std::stringstream hexStr;
-	hexStr << std::uppercase << hex << _value;
-	return hexStr.str();
-}
-
-Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) const
+Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices, bool _includeSourceList) const
 {
 	Json::Value root;
 	root[".code"] = Json::arrayValue;
-
-	Json::Value& collection = root[".code"];
-	for (AssemblyItem const& i: m_items)
+	Json::Value& code = root[".code"];
+	for (AssemblyItem const& item: m_items)
 	{
 		int sourceIndex = -1;
-		if (i.location().sourceName)
+		if (item.location().sourceName)
 		{
-			auto iter = _sourceIndices.find(*i.location().sourceName);
+			auto iter = _sourceIndices.find(*item.location().sourceName);
 			if (iter != _sourceIndices.end())
 				sourceIndex = static_cast<int>(iter->second);
 		}
 
-		switch (i.type())
+		auto [name, data] = item.nameAndData();
+		Json::Value jsonItem;
+		jsonItem["name"] = name;
+		jsonItem["begin"] = item.location().start;
+		jsonItem["end"] = item.location().end;
+		if (item.m_modifierDepth != 0)
+			jsonItem["modifierDepth"] = static_cast<int>(item.m_modifierDepth);
+		std::string jumpType = item.getJumpTypeAsString();
+		if (!jumpType.empty())
+			jsonItem["jumpType"] = jumpType;
+		if (name == "PUSHLIB")
+			data = m_libraries.at(h256(data));
+		else if (name == "PUSHIMMUTABLE" || name == "ASSIGNIMMUTABLE")
+			data = m_immutables.at(h256(data));
+		if (!data.empty())
+			jsonItem["value"] = data;
+		jsonItem["source"] = sourceIndex;
+		code.append(move(jsonItem));
+
+		if (item.type() == AssemblyItemType::Tag)
 		{
-		case Operation:
-			collection.append(
-				createJsonValue(
-					instructionInfo(i.instruction()).name,
-					sourceIndex,
-					i.location().start,
-					i.location().end,
-					i.getJumpTypeAsString())
-				);
-			break;
-		case Push:
-			collection.append(
-				createJsonValue("PUSH", sourceIndex, i.location().start, i.location().end, toStringInHex(i.data()), i.getJumpTypeAsString()));
-			break;
-		case PushTag:
-			if (i.data() == 0)
-				collection.append(
-					createJsonValue("PUSH [ErrorTag]", sourceIndex, i.location().start, i.location().end, ""));
-			else
-				collection.append(
-					createJsonValue("PUSH [tag]", sourceIndex, i.location().start, i.location().end, toString(i.data())));
-			break;
-		case PushSub:
-			collection.append(
-				createJsonValue("PUSH [$]", sourceIndex, i.location().start, i.location().end, toString(h256(i.data()))));
-			break;
-		case PushSubSize:
-			collection.append(
-				createJsonValue("PUSH #[$]", sourceIndex, i.location().start, i.location().end, toString(h256(i.data()))));
-			break;
-		case PushProgramSize:
-			collection.append(
-				createJsonValue("PUSHSIZE", sourceIndex, i.location().start, i.location().end));
-			break;
-		case PushLibraryAddress:
-			collection.append(
-				createJsonValue("PUSHLIB", sourceIndex, i.location().start, i.location().end, m_libraries.at(h256(i.data())))
-			);
-			break;
-		case PushDeployTimeAddress:
-			collection.append(
-				createJsonValue("PUSHDEPLOYADDRESS", sourceIndex, i.location().start, i.location().end)
-			);
-			break;
-		case PushImmutable:
-			collection.append(createJsonValue(
-				"PUSHIMMUTABLE",
-				sourceIndex,
-				i.location().start,
-				i.location().end,
-				m_immutables.at(h256(i.data()))
-			));
-			break;
-		case AssignImmutable:
-			collection.append(createJsonValue(
-				"ASSIGNIMMUTABLE",
-				sourceIndex,
-				i.location().start,
-				i.location().end,
-				m_immutables.at(h256(i.data()))
-			));
-			break;
-		case Tag:
-			collection.append(
-				createJsonValue("tag", sourceIndex, i.location().start, i.location().end, toString(i.data())));
-			collection.append(
-				createJsonValue("JUMPDEST", sourceIndex, i.location().start, i.location().end));
-			break;
-		case PushData:
-			collection.append(createJsonValue("PUSH data", sourceIndex, i.location().start, i.location().end, toStringInHex(i.data())));
-			break;
-		case VerbatimBytecode:
-			collection.append(createJsonValue("VERBATIM", sourceIndex, i.location().start, i.location().end, util::toHex(i.verbatimData())));
-			break;
-		default:
-			assertThrow(false, InvalidOpcode, "");
+			Json::Value jumpdest;
+			jumpdest["name"] = "JUMPDEST";
+			jumpdest["begin"] = item.location().start;
+			jumpdest["end"] = item.location().end;
+			jumpdest["source"] = sourceIndex;
+			if (item.m_modifierDepth != 0)
+				jumpdest["modifierDepth"] = static_cast<int>(item.m_modifierDepth);
+			code.append(move(jumpdest));
 		}
+	}
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& jsonSourceList = root["sourceList"];
+		for (auto const& [name, index]: _sourceIndices)
+			jsonSourceList[index] = name;
 	}
 
 	if (!m_data.empty() || !m_subs.empty())
@@ -346,17 +282,17 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) 
 		Json::Value& data = root[".data"];
 		for (auto const& i: m_data)
 			if (u256(i.first) >= m_subs.size())
-				data[toStringInHex((u256)i.first)] = util::toHex(i.second);
+				data[util::toHex(toBigEndian((u256)i.first), util::HexPrefix::DontAdd, util::HexCase::Upper)] = util::toHex(i.second);
 
 		for (size_t i = 0; i < m_subs.size(); ++i)
 		{
 			std::stringstream hexStr;
 			hexStr << hex << i;
-			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices);
+			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices, /*_includeSourceList = */false);
 		}
 	}
 
-	if (m_auxiliaryData.size() > 0)
+	if (!m_auxiliaryData.empty())
 		root[".auxdata"] = util::toHex(m_auxiliaryData);
 
 	return root;
--- a/libevmasm/Assembly.h
+++ b/libevmasm/Assembly.h
@@ -39,6 +39,7 @@
 #include <sstream>
 #include <memory>
 #include <map>
+#include <utility>
 
 namespace solidity::evmasm
 {
@@ -147,7 +148,8 @@ public:
 
 	/// Create a JSON representation of the assembly.
 	Json::Value assemblyJSON(
-		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>()
+		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>(),
+		bool _includeSourceList = true
 	) const;
 
 	/// Mark this assembly as invalid. Calling ``assemble`` on it will throw.
@@ -167,16 +169,6 @@ protected:
 	unsigned codeSize(unsigned subTagSize) const;
 
 private:
-	static Json::Value createJsonValue(
-		std::string _name,
-		int _source,
-		int _begin,
-		int _end,
-		std::string _value = std::string(),
-		std::string _jumpType = std::string()
-	);
-	static std::string toStringInHex(u256 _value);
-
 	bool m_invalid = false;
 
 	Assembly const* subAssemblyById(size_t _subId) const;
@@ -222,6 +214,7 @@ protected:
 	std::string m_name;
 
 	langutil::SourceLocation m_currentSourceLocation;
+
 public:
 	size_t m_currentModifierDepth = 0;
 };
--- a/libevmasm/AssemblyItem.cpp
+++ b/libevmasm/AssemblyItem.cpp
@@ -21,6 +21,7 @@
 #include <libevmasm/Assembly.h>
 
 #include <libsolutil/CommonData.h>
+#include <libsolutil/CommonIO.h>
 #include <libsolutil/Numeric.h>
 #include <libsolutil/StringUtils.h>
 #include <libsolutil/FixedHash.h>
@@ -35,6 +36,18 @@ using namespace solidity::evmasm;
 using namespace solidity::langutil;
 
 static_assert(sizeof(size_t) <= 8, "size_t must be at most 64-bits wide");
+
+namespace
+{
+
+string toStringInHex(u256 _value)
+{
+	std::stringstream hexStr;
+	hexStr << std::uppercase << hex << _value;
+	return hexStr.str();
+}
+
+}
 
 AssemblyItem AssemblyItem::toSubAssemblyTag(size_t _subId) const
 {
@@ -54,6 +67,44 @@ pair<size_t, size_t> AssemblyItem::splitForeignPushTag() const
 	size_t subId = static_cast<size_t>((combined >> 64) - 1);
 	size_t tag = static_cast<size_t>(combined & 0xffffffffffffffffULL);
 	return make_pair(subId, tag);
+}
+
+pair<string, string> AssemblyItem::nameAndData() const
+{
+	switch (type())
+	{
+	case Operation:
+		return {instructionInfo(instruction()).name, m_data != nullptr ? toStringInHex(*m_data) : ""};
+	case Push:
+		return {"PUSH", toStringInHex(data())};
+	case PushTag:
+		if (data() == 0)
+			return {"PUSH [ErrorTag]", ""};
+		else
+			return {"PUSH [tag]", util::toString(data())};
+	case PushSub:
+		return {"PUSH [$]", toString(util::h256(data()))};
+	case PushSubSize:
+		return {"PUSH #[$]", toString(util::h256(data()))};
+	case PushProgramSize:
+		return {"PUSHSIZE", ""};
+	case PushLibraryAddress:
+		return {"PUSHLIB", toString(util::h256(data()))};
+	case PushDeployTimeAddress:
+		return {"PUSHDEPLOYADDRESS", ""};
+	case PushImmutable:
+		return {"PUSHIMMUTABLE", toString(util::h256(data()))};
+	case AssignImmutable:
+		return {"ASSIGNIMMUTABLE", toString(util::h256(data()))};
+	case Tag:
+		return {"tag", util::toString(data())};
+	case PushData:
+		return {"PUSH data", toStringInHex(data())};
+	case VerbatimBytecode:
+		return {"VERBATIM", util::toHex(verbatimData())};
+	default:
+		assertThrow(false, InvalidOpcode, "");
+	}
 }
 
 void AssemblyItem::setPushTagSubIdAndTag(size_t _subId, size_t _tag)
--- a/libevmasm/AssemblyItem.h
+++ b/libevmasm/AssemblyItem.h
@@ -106,6 +106,13 @@ public:
 	u256 const& data() const { assertThrow(m_type != Operation, util::Exception, ""); return *m_data; }
 	void setData(u256 const& _data) { assertThrow(m_type != Operation, util::Exception, ""); m_data = std::make_shared<u256>(_data); }
 
+	/// This function is used in `Assembly::assemblyJSON`.
+	/// It returns the name & data of the current assembly item.
+	/// @returns a pair, where the first element is the json-assembly
+	/// item name, where second element is the string representation
+	/// of it's data.
+	std::pair<std::string, std::string> nameAndData() const;
+
 	bytes const& verbatimData() const { assertThrow(m_type == VerbatimBytecode, util::Exception, ""); return std::get<2>(*m_verbatimBytecode); }
 
 	/// @returns the instruction of this item (only valid if type() == Operation)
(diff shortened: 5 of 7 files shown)


libevmasm/Assembly.cpp
```diff
@@ -242,13 +245,153 @@ string Assembly::toStringInHex(u256 _value)
 	hexStr << std::uppercase << hex << _value;
 	return hexStr.str();
 }
+vector<Json::Value> Assembly::assemblyItemAsJSON(AssemblyItem const& _item, int _sourceIndex) const
```
aarlt: Sadly this need to be a vector, because `Tag` is emitting two json objects, `tag` and `JUMPDEST`.

libevmasm/Assembly.cpp
```diff
@@ -242,13 +245,153 @@ string Assembly::toStringInHex(u256 _value)
...
+
+	switch (_item.type())
+	{
+	case Operation:
+		result.emplace_back(createJsonValue(
+			instructionInfo(_item.instruction()).name,
+			_sourceIndex,
+			_item.m_modifierDepth,
+			_item.location().start,
+			_item.location().end,
```
chriseth: Can you find a solution where we don't have to pass there four arguments all the time to each of the function calls?

aarlt: True! That makes sense!

test/cmdlineTests/asm_json/output
```diff
@@ -450,9 +450,9 @@ EVM assembly:
         {
           "begin": 77,
           "end": 158,
+          "jumpType": "[in]",
```
aarlt, line 453: hmm.. I guess that should not be like that..

libevmasm/Assembly.cpp
```diff
@@ -242,13 +247,123 @@ string Assembly::toStringInHex(u256 _value)
...
+vector<Json::Value> Assembly::assemblyItemAsJSON(AssemblyItem const& _item, int _sourceIndex) const
+{
+	vector<Json::Value> result;
+
+	switch (_item.type())
+	{
+	case Operation:
+		result.emplace_back(createJsonValue(
+			instructionInfo(_item.instruction()).name,
+			_sourceIndex,
```
chriseth: I'm still not happy that the `_sourceIndex` is pulled all the way through.
Would it work to create a (or two) json value that only contains the values that depend on the type of the assembly item and then the rest is filled in Assembly::assemblyJSON ?

libevmasm/Assembly.h
```diff
@@ -158,6 +160,16 @@ class Assembly
 
 	bool isCreation() const { return m_creation; }
 
+	void setSources(std::vector<std::shared_ptr<std::string const>> _sources) {
```
chriseth: Coding style.

Could you split this up further and only include the change to `Assembly::assemblyJSON` - maybe not even the `_includeSourceList` parameter?

libevmasm/Assembly.cpp
aarlt: Sure! It's much better like that!

libevmasm/Assembly.h
aarlt: good idea!

chriseth:
I think this can be even further simplified by creating a function (can be a method of AssemblyItem) that returns `pair<string, string>` which is the name and the value. The function would then just be
```
switch (_item.type())
{
case Operation: return {instructionInfo(_item.instruction()).name, ""};
case Push: return {"PUSH", toStringInHex(_item.data())};
...
}
```
The rest of the values is then added in the loop inside Assembly.cpp and inside this loop, we also add the additional jumpdest for tags.

What do you think?

libevmasm/Assembly.h
```diff
@@ -146,9 +147,7 @@ class Assembly
 	) const;
 
 	/// Create a JSON representation of the assembly.
-	Json::Value assemblyJSON(
-		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>()
-	) const;
+	Json::Value assemblyJSON(std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>(), bool _includeSourceList = true) const;
```
cameel: This line is getting long. I'd keep it the way it was split originally:
```suggestion
	Json::Value assemblyJSON(
		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>(),
		bool _includeSourceList = true
	) const;
```

libevmasm/Assembly.h
```diff
@@ -39,6 +39,7 @@
 #include <sstream>
 #include <memory>
 #include <map>
+#include <utility>
```
cameel, line 42: Interesting. Never knew that C++ had such a catch-all header for basic stuff. I wonder if we should be using it more often instead of including things individually.

libevmasm/Assembly.cpp
```diff
@@ -222,17 +222,21 @@ string Assembly::assemblyString(
...
-	value["begin"] = _begin;
-	value["end"] = _end;
+	value["begin"] = _item.location().start;
+	value["end"] = _item.location().end;
+	if (_item.m_modifierDepth != 0)
+		value["modifierDepth"] = static_cast<int>(_item.m_modifierDepth);
 	if (!_value.empty())
 		value["value"] = _value;
-	if (!_jumpType.empty())
-		value["jumpType"] = _jumpType;
```
cameel: I see that after your changes, `"jumpType"` field is no longer being filled. On the other hand you're now filling `"modifierDepth"`. This seems like an API change rather than just a refactor. Is that intentional?

libevmasm/Assembly.cpp
```diff
@@ -222,17 +222,21 @@ string Assembly::assemblyString(
...
+	if (!_item.getJumpTypeAsString().empty())
+	{
+		assertThrow(_name == "JUMP" && value.isMember("value") == false, AssemblyException, "");
```
cameel: Is this an assertion or input validation?

If it's validation, it should have a message. If it's an assertion, I think you should use `solAssert()` instead to make it clear.

libevmasm/Assembly.cpp
```diff
@@ -242,13 +246,101 @@ string Assembly::toStringInHex(u256 _value)
...
+{
+	vector<Json::Value> result;
 
-Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) const
+	switch (_item.type())
+	{
+	case Operation:
+		result.emplace_back(createJsonValue(
+			instructionInfo(_item.instruction()).name,
+			_item));
```
cameel: Wrapping does not match our style. Same for other `case`s in this `switch`.
```suggestion
		result.emplace_back(createJsonValue(
			instructionInfo(_item.instruction()).name,
			_item
		));
```

libevmasm/Assembly.cpp
```diff
@@ -242,13 +246,101 @@ string Assembly::toStringInHex(u256 _value)
...
+			_item,
+			m_immutables.at(h256(_item.data()))));
+		break;
+	case Tag:
+		result.emplace_back(createJsonValue(
+			"tag",
+			_item,
+			toString(_item.data())));
+		result.emplace_back(createJsonValue(
+			"JUMPDEST",  _item));
```
cameel:
```suggestion
		result.emplace_back(createJsonValue(
			"JUMPDEST",
			_item
		));
```

libevmasm/Assembly.cpp
cameel:
```suggestion
		assertThrow(_name == "JUMP" && !value.isMember("value"), AssemblyException, "");
```

libevmasm/Assembly.cpp
```diff
@@ -259,86 +351,22 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
-			break;
-		case PushData:
-			collection.append(createJsonValue("PUSH data", sourceIndex, i.location().start, i.location().end, toStringInHex(i.data())));
-			break;
-		case VerbatimBytecode:
-			collection.append(createJsonValue("VERBATIM", sourceIndex, i.location().start, i.location().end, util::toHex(i.verbatimData())));
-			break;
-		default:
-			assertThrow(false, InvalidOpcode, "");
+			item["source"] = sourceIndex;
```
cameel: I'd assert that `source` is not yet in `item` to make sure we're not overwriting anything:
```suggestion
			solAssert(!item.isMember("source"));
			item["source"] = sourceIndex;
```

libevmasm/Assembly.cpp
```diff
@@ -242,13 +246,101 @@ string Assembly::toStringInHex(u256 _value)
 	hexStr << std::uppercase << hex << _value;
 	return hexStr.str();
 }
+vector<Json::Value> Assembly::assemblyItemAsJSON(AssemblyItem const& _item) const
```
cameel: Function names here are confusing to me. I had to read `createJsonValue()` and `assemblyItemAsJSON()` to really understand what they do. From the names and arguments alone, they both look like they're just translating an `AssemblyItem` to JSON.

It would be really helpful to have a term for the thing that `createJsonValue()` returns. Can we call it an opcode? Assembly instruction? Line of assembly? It's a smaller unit than assembly item but it also includes tags which aren't opcodes/instructions. We should use that term in the name.

libevmasm/Assembly.cpp
```diff
@@ -222,17 +222,21 @@ string Assembly::assemblyString(
...
+	if (_item.m_modifierDepth != 0)
+		value["modifierDepth"] = static_cast<int>(_item.m_modifierDepth);
 	if (!_value.empty())
 		value["value"] = _value;
-	if (!_jumpType.empty())
-		value["jumpType"] = _jumpType;
+	if (!_item.getJumpTypeAsString().empty())
+	{
+		assertThrow(_name == "JUMP" && value.isMember("value") == false, AssemblyException, "");
+		value["value"] = _item.getJumpTypeAsString();
```
cameel: Isn't this supposed to be `jumpType` rather than `value`? You have already filled `value` above and now you're overwriting it.
```suggestion
		value["jumpType"] = _item.getJumpTypeAsString();
```

libevmasm/Assembly.cpp
```diff
@@ -259,86 +351,22 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& sourceList = root["sourceList"];
+		vector<string> sources(_sourceIndices.size());
+		for (auto const& item: _sourceIndices)
+			sources[item.second] = item.first;
+		for (auto const& item: sources)
+			sourceList.append(item);
+	}
```
cameel: This also looks like an actual JSON change rather than a refactor.

It would be easier to review if the changes that affect the JSON output were separate from ones that just make the code nicer.

libevmasm/Assembly.cpp
```diff
@@ -259,86 +351,22 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+			code.append(item);
 		}
 	}
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& sourceList = root["sourceList"];
+		vector<string> sources(_sourceIndices.size());
+		for (auto const& item: _sourceIndices)
+			sources[item.second] = item.first;
```
cameel:
```suggestion
		for (auto const& [name, index]: _sourceIndices)
			sources[index] = name;
```

libevmasm/Assembly.cpp
```diff
@@ -259,86 +351,22 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
-		default:
-			assertThrow(false, InvalidOpcode, "");
+			item["source"] = sourceIndex;
+			code.append(item);
 		}
 	}
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& sourceList = root["sourceList"];
```
cameel: I think the code would be more readable if you used `root["sourceList"]` directly. An extra variable means an extra thing to keep track of while reading so I think it's not worth it unless the original way to access it is overly long or if it's a function call that would have to be recomputed on every access.

If it's to avoid the lookup, I'd at least add `json` in front of it to make it clear it's part of the JSON:
```suggestion
		Json::Value& jsonSourceList = root["sourceList"];
```

libevmasm/Assembly.cpp
```diff
@@ -259,86 +351,22 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+		vector<string> sources(_sourceIndices.size());
```
cameel:
```suggestion
		vector<string> sourceNames(_sourceIndices.size());
```

libevmasm/Assembly.cpp
```diff
@@ -352,11 +380,11 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
 		{
 			std::stringstream hexStr;
 			hexStr << hex << i;
-			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices);
+			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices, false);
```
cameel:
```suggestion
			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices, /*_includeSourceList = */ false);
```
or, alternatively:
```suggestion
			bool const includeSourceList = false;
			data[hexStr.str()] = m_subs[i]->assemblyJSON(_sourceIndices, includeSourceList);
```

test/cmdlineTests/asm_json/output
```diff
@@ -1582,5 +1582,10 @@ EVM assembly:
         }
       ]
     }
-  }
+  },
```
cameel, line 1585: Why is the removal of `"jumpType"` and addition of `"modifierDepth"` not reflected in this test? I'd expect it to fail after these changes.

aarlt: I'm now also wondering. Yesterday I changed exactly this part. The thing is that in our current tests we have it like this:
```json
        {
          "begin": 77,
          "end": 158,
          "name": "JUMP",
          "source": 0,
          "value": "[in]"
        }
```
So I saw yesterday, that the `jumpType` is stored in `value` - so I just changed that - without really checking the original code. The strange thing is that the original code reads like that there should be an `jumpType` node. Something seem to be wrong. I will look into this. Thx for that hint!

aarlt: For `modifierDepth` it is just needed to be able to import the asm json correctly. Without this information it is not possible to generate the original byte code.

cameel: So that JSON the function creates is not the final form but later goes through some extra processing?

aarlt: Not sure what you mean. I think `modifierDepth` was just forgotten when `asm-json` was first implemented.

cameel: Oh. From your comment I thought that `modifierDepth` is used internally but then not written out on purpose - so there must be some postprocessing step that removes it (and maybe does other stuff).

libevmasm/Assembly.cpp
aarlt: Yes, I removed `jumpType` yesterday, because I saw that it is just stored in `value` (for `Operation`). Regarding the old code, `jumpType` is only set for `PUSH`. Not sure why.

libevmasm/Assembly.cpp
aarlt: Yeah this PR is not just a refactor, because it also adds this source list support. Maybe I should have splitt this PR in a refactor PR & the add source list PR.

libevmasm/Assembly.cpp
aarlt: @chriseth why is `jumpType` needed for `PUSH`?

aarlt: Or is it just a typo in the code?

chriseth: I don't see a problem with the old code. Yes, jumptype was output for push, but also for operation.

chriseth: Oh you are right, it was stored in the wrong location. Why did nobody ever complain?

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,68 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
-		case PushData:
-			collection.append(createJsonValue("PUSH data", sourceIndex, i.location().start, i.location().end, toStringInHex(i.data())));
-			break;
-		case VerbatimBytecode:
-			collection.append(createJsonValue("VERBATIM", sourceIndex, i.location().start, i.location().end, util::toHex(i.verbatimData())));
-			break;
-		default:
-			assertThrow(false, InvalidOpcode, "");
+			if (name == "JUMP")
+				item["value"] = jumpType;
```
chriseth: Please add a comment that this is replicating a bug.

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,68 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+			if (name == "PUSH")
```
chriseth:
```suggestion
			else
```

libevmasm/AssemblyItem.cpp
```diff
@@ -56,6 +57,51 @@ pair<size_t, size_t> AssemblyItem::splitForeignPushTag() const
 	return make_pair(subId, tag);
 }
 
+string AssemblyItem::toStringInHex(u256 _value)
```
chriseth: Don't we already have a utility function for that? In any case, please turn this into a function inside an anonnymous namespace at least.

libevmasm/Assembly.cpp
aarlt: I guess nobody really used it yet. Decision from call: store jump type's always in `jumpType` node.

test/cmdlineTests/asm_json/output
aarlt: I just needed to add `modifierDepth`. It was just missing in the original `asm-json` implementation. There is no post processing or any other extra processing involved. This information was just missing, and now this information will be exported correctly.

libevmasm/Assembly.cpp
aarlt: I replaced this now with 
```
		if (!jumpType.empty())
			item["jumpType"] = jumpType;
```
so whenever a jump type was defined, it will be stored in an `jumpType` node.

The original behaviour for `JUMP` (and maybe also other operations) was to store it in `value`. We will now store this in `jumpType`.

libevmasm/Assembly.cpp
aarlt: We will store this now always in `jumpType`.

libevmasm/AssemblyItem.cpp
aarlt: Yes, but I was not able to put it in exactly the needed hex format with the other utility functions. Moved to anonymous namespace.

libevmasm/Assembly.cpp
aarlt: Yeah you where totally right. I removed `createJsonValue()` now. I think after I did https://github.com/ethereum/solidity/pull/12799#issuecomment-1070640194 the code is now more readable.

test/cmdlineTests/asm_json/output
aarlt: Just to also write it here in this thread: decision from call: always store jump type in a `jumpType` node. @cameel thx for making us aware of that!

chriseth:
Since this also fixes a bug now, it needs a changelog entry.

libevmasm/Assembly.cpp
```diff
@@ -222,33 +222,11 @@ string Assembly::assemblyString(
...
-
-Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices) const
+Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices, bool _includeSourceList) const
 {
 	Json::Value root;
 	root[".code"] = Json::arrayValue;
-
-	Json::Value& collection = root[".code"];
+	Json::Value& code = root[".code"];
 	for (AssemblyItem const& i: m_items)
```
chriseth: Maybe call this `item` and the other `itemJson`?

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+		if (name == "PUSHLIB")
+			data = m_libraries.at(h256(data));
+		else if (name == "PUSHIMMUTABLE" || name == "ASSIGNIMMUTABLE")
+			data = m_immutables.at(h256(data));
+		if (!data.empty())
+			item["value"] = data;
+		item["source"] = sourceIndex;
+		code.append(item);
+
+		if (name == "tag")
```
chriseth: I would feel better if this as a condition on the type instead of the returned string. It is much clearer that there can be no collisions.

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
-		case VerbatimBytecode:
-			collection.append(createJsonValue("VERBATIM", sourceIndex, i.location().start, i.location().end, util::toHex(i.verbatimData())));
-			break;
-		default:
-			assertThrow(false, InvalidOpcode, "");
+			Json::Value jumpdest;
+			jumpdest["name"] = "JUMPDEST";
+			jumpdest["begin"] = i.location().start;
+			jumpdest["end"] = i.location().end;
+			jumpdest["source"] = sourceIndex;
```
chriseth, line 265: no modifier depth?

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+			code.append(jumpdest);
```
chriseth:
```suggestion
			code.append(move(jumpdest));
```

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
 		}
 	}
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& jsonSourceList = root["sourceList"];
```
chriseth, line 274: This looks a bit complicated. What about move-assigning it at the end?

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+		vector<string> sourceNames(_sourceIndices.size());
```
chriseth: Are we sure they are contiguous?

libevmasm/Assembly.cpp
```diff
@@ -259,104 +237,63 @@ Json::Value Assembly::assemblyJSON(map<string, unsigned> const& _sourceIndices)
...
+		for (auto const& [name, index]: _sourceIndices)
+			sourceNames[index] = name;
+		for (auto const& item: sourceNames)
+			jsonSourceList.append(item);
```
chriseth: Can't you make this a json array and then avoid the vector?

libevmasm/Assembly.h
```diff
@@ -147,7 +148,8 @@ class Assembly
 
 	/// Create a JSON representation of the assembly.
 	Json::Value assemblyJSON(
-		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>()
+		std::map<std::string, unsigned> const& _sourceIndices = std::map<std::string, unsigned>(),
+		bool _includeSourceList = true
```
chriseth, line 152: Shouldn't this default to false to retain the old behaviour? Do we need a default value?

libevmasm/AssemblyItem.h
```diff
@@ -105,6 +105,7 @@ class AssemblyItem
 	AssemblyItemType type() const { return m_type; }
 	u256 const& data() const { assertThrow(m_type != Operation, util::Exception, ""); return *m_data; }
 	void setData(u256 const& _data) { assertThrow(m_type != Operation, util::Exception, ""); m_data = std::make_shared<u256>(_data); }
+	std::pair<std::string, std::string> nameAndData() const;
```
chriseth, line 114: Please document. Mainly mention that this is used for json export.

chriseth:
Can you add a modifier to the commandline test so we can test the modifier depth?

libevmasm/Assembly.cpp
aarlt: Good catch! :)

libevmasm/Assembly.h
aarlt: Right now I set `includeSourceList` to `false`, only if a sub-assembly is exported. In `CompilerStack::assemblyJSON(..)` we want to have the source list included. If called with default value, the top-level assembly json will always include the source list, but it's sub-assemblies not.

libevmasm/Assembly.cpp
aarlt: I think [yes](https://github.com/ethereum/solidity/blob/develop/libsolidity/interface/CompilerStack.cpp#L941).

libevmasm/Assembly.cpp
aarlt: The actual order will be defined with the first for loop. The json array seem not to allow the insertion of items at specific positions. I just use the vector to store the final list of source names, ordered according to its index.

Changelog.md
```diff
@@ -8,6 +8,7 @@ Compiler Features:
 
 
 Bugfixes:
+ * Assembly: Fix assembly json export to store jump types of operations in `jumpType` field (instead of `value`).
```
```suggestion
 * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
```

libevmasm/Assembly.cpp
```diff
@@ -222,141 +222,80 @@ string Assembly::assemblyString(
...
```
chriseth:
```suggestion
		root["sourceList"] = Json::arrayValue;
		for (auto const& [name, index]: _sourceIndices)
			root["sourceList"][index] = name;
```
I'm pretty sure this does it.

libevmasm/Assembly.h
chriseth: Oh, this should also be mentioned in the changelog.

aarlt: Thats true!

Changelog.md
aarlt: Oh, great! Somehow I did not check the operators. Thx!

libevmasm/Assembly.cpp
aarlt: Oh, great! Somehow I did not check the operators. Thx!

Changelog.md
```diff
@@ -8,6 +8,8 @@ Compiler Features:
 
 
 Bugfixes:
+* Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
+* Assembly-Json: Fix assembly json to include source list in `sourceList` field.
```
chriseth: Isn't this a feature?

libevmasm/Assembly.cpp
```diff
@@ -222,141 +222,78 @@ string Assembly::assemblyString(
...
+			if (item.m_modifierDepth != 0)
+				jumpdest["modifierDepth"] = static_cast<int>(item.m_modifierDepth);
+			code.append(move(jumpdest));
 		}
 	}
+	if (_includeSourceList)
+	{
+		root["sourceList"] = Json::arrayValue;
+		Json::Value& jsonSourceList = root["sourceList"];
+		vector<string> sourceNames(_sourceIndices.size());
```
chriseth:
```suggestion
```
(why didn't the compiler complain?)

test/cmdlineTests/asm_json/output
chriseth: Is this comment still relevant?

chriseth:
Only cosmetic changes left.

libevmasm/Assembly.cpp
aarlt: Interesting.. yeah that's strange..

Changelog.md
aarlt: yes, maybe :)

chriseth (approved):

Sticknmove:
Thanks

PR #12801, opened 2022-03-16
CSE optimization

### LINKED ISSUE #12406: Large array declaration hangs compiler when optimization enabled
## Description

For private deployment of Ethereum I would like to deploy "larger" contracts than on mainnet.
Here's example of such smart contract: https://gist.github.com/citizen-stig/a25be3d125969c64f0f2b94b28a0d160

solc is able to compile this smart contract normally, but hangs when optimization is enabled.

## Environment

- Compiler version: 0.8.10+commit.fc410830.Linux.g++
- Target EVM version (as per compiler settings): Default
- Framework/IDE (e.g. Truffle or Remix): No
- EVM execution environment / backend / blockchain client: Besu
- Operating system:  x86_64 GNU/Linux 5.15.6-arch2 Archlinux

## Steps to Reproduce

Download example source code from [this gist](https://gist.github.com/citizen-stig/a25be3d125969c64f0f2b94b28a0d160)

Run this command:

```
solc --bin -o build_check --optimize --optimize-runs=200 --no-optimize-yul ./MediumVerifier.sol
```

### Expected result

Code is compiled, same way as without optimization.

### Actual result

Compiler hangs, consuming 100% of single cpu thread.

### Notes

Outcome is the same if `--optimize-runs=1`

wechman:
partially address: #12406

Proposed changes:
1) Fix the problem of missing reference in range-for with non trivial type in CSECodeGenerator::generateClassElement.
2) During optimization, the [CommonSubexpressionEliminator](https://github.com/ethereum/solidity/blob/develop/libevmasm/CommonSubexpressionEliminator.h#L62) spends much time accessing elements of [m_classPositions](https://github.com/ethereum/solidity/blob/develop/libevmasm/CommonSubexpressionEliminator.h#L161) and [m_neededBy](https://github.com/ethereum/solidity/blob/develop/libevmasm/CommonSubexpressionEliminator.h#L157). By converting these two to hash maps we gain constant access time in average scenario.
3) Because [m_neededBy](https://github.com/ethereum/solidity/blob/develop/libevmasm/CommonSubexpressionEliminator.h#L157) is a multimap it is usually more efficient to check element presence with `find` function call instead `count`.

With all above changes [MediumVerifier.sol](https://gist.github.com/citizen-stig/a25be3d125969c64f0f2b94b28a0d160) compilation time in my local environment is reduced from 2h10min to 24min.

### DIFF
--- a/libevmasm/CommonSubexpressionEliminator.cpp
+++ b/libevmasm/CommonSubexpressionEliminator.cpp
@@ -219,7 +219,7 @@ void CSECodeGenerator::addDependencies(Id _c)
 {
 	if (m_classPositions.count(_c))
 		return; // it is already on the stack
-	if (m_neededBy.count(_c))
+	if (m_neededBy.find(_c) != m_neededBy.end())
 		return; // we already computed the dependencies for _c
 	ExpressionClasses::Expression expr = m_expressionClasses.representative(_c);
 	assertThrow(expr.item, OptimizerException, "");
@@ -300,8 +300,8 @@ void CSECodeGenerator::addDependencies(Id _c)
 
 void CSECodeGenerator::generateClassElement(Id _c, bool _allowSequenced)
 {
-	for (auto it: m_classPositions)
-		for (auto p: it.second)
+	for (auto const& it: m_classPositions)
+		for (int p: it.second)
 			if (p > m_stackHeight)
 			{
 				assertThrow(false, OptimizerException, "");
--- a/libevmasm/CommonSubexpressionEliminator.h
+++ b/libevmasm/CommonSubexpressionEliminator.h
@@ -24,11 +24,12 @@
 
 #pragma once
 
-#include <vector>
 #include <map>
+#include <ostream>
 #include <set>
 #include <tuple>
-#include <ostream>
+#include <unordered_map>
+#include <vector>
 #include <libsolutil/CommonIO.h>
 #include <libsolutil/Exceptions.h>
 #include <libevmasm/ExpressionClasses.h>
@@ -154,11 +155,11 @@ private:
 	/// Current height of the stack relative to the start.
 	int m_stackHeight = 0;
 	/// If (b, a) is in m_requests then b is needed to compute a.
-	std::multimap<Id, Id> m_neededBy;
+	std::unordered_multimap<Id, Id> m_neededBy;
 	/// Current content of the stack.
 	std::map<int, Id> m_stack;
 	/// Current positions of equivalence classes, equal to the empty set if already deleted.
-	std::map<Id, std::set<int>> m_classPositions;
+	std::unordered_map<Id, std::set<int>> m_classPositions;
 
 	/// The actual equivalence class items and how to compute them.
 	ExpressionClasses& m_expressionClasses;


libevmasm/CommonSubexpressionEliminator.h
```diff
@@ -154,11 +155,11 @@ class CSECodeGenerator
 	/// Current height of the stack relative to the start.
 	int m_stackHeight = 0;
 	/// If (b, a) is in m_requests then b is needed to compute a.
-	std::multimap<Id, Id> m_neededBy;
+	std::unordered_multimap<Id, Id> m_neededBy;
```
chriseth, line 158: Can this make compilation in any way depend on the specific hash function?

libevmasm/CommonSubexpressionEliminator.cpp
```diff
@@ -219,7 +219,7 @@ void CSECodeGenerator::addDependencies(Id _c)
 {
 	if (m_classPositions.count(_c))
 		return; // it is already on the stack
-	if (m_neededBy.count(_c))
+	if (m_neededBy.find(_c) != m_neededBy.end())
```
chriseth, line 222: Can you use the `contains` helper?

ekpyron (dismissed):
From the looks of it, none of the traversals of those ``unordered_*`` maps depend on the order of traversal, so from what I can see this should be safe, resp. remain equivalent!

chriseth:
Can you squash to a single commit please?

libevmasm/CommonSubexpressionEliminator.cpp
```diff
@@ -300,7 +300,7 @@ void CSECodeGenerator::addDependencies(Id _c)
 
 void CSECodeGenerator::generateClassElement(Id _c, bool _allowSequenced)
 {
-	for (auto it: m_classPositions)
+	for (auto const& it: m_classPositions)
 		for (auto p: it.second)
```
chriseth: can you make this an explicit `int`?

wechman: done

wechman:
> Can you squash to a single commit please?

done

libevmasm/CommonSubexpressionEliminator.cpp
wechman: I am not sure of perfomance impact of such change.

ekpyron: Good catch... we should actually specialize ``contains`` in ``libsolutil/CommonData.h`` for types that have ``.find`` similar to how it's done for ``findOffset``... but no need to do it in this PR, so I'd say just keeping it as is, is fine.

ekpyron (approved):

libevmasm/CommonSubexpressionEliminator.h
wechman: Taking into consideration, code does not rely on the order of the elements in this container, hashing function should have no impact on the output. But, I can imagine that it can have an impact on the performance in case of a poor hashing algorithm.

libevmasm/CommonSubexpressionEliminator.cpp
chriseth: Oh, I was hoping that std::find does it, but of course it can't.

PR #12809, opened 2022-03-16
scripts/ImportExportTest.sh: replace scripts/ASTImportTest.sh.

aarlt:
Replaces part of #12704.
Moved to #12834.

- replaces `scripts/ASTImportTest.sh` with `scripts/ImportExportTest.sh`
- `scripts/ImportExportTest.sh` is already prepared to test asm-json import/export, but right now these tests are just not triggered.


scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                echo "Compiler stderr:"
+                cat ./stderr.txt
+                echo
+                echo "Compiler stdout:"
+                cat ./obtained.json
+                return 1
+            fi
+            set +e
+            DIFF="$(diff expected.json obtained.json)"
+            set +e
```
wechman, line 70: Shouldn't it be `set -e`?

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+set -e
+
+# Bash script to test the ast-import option of the compiler by
+# first exporting a .sol file to JSON, then loading it into the compiler
+# and exporting it again. The second JSON should be identical to the first
+READLINK=readlink
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    READLINK=greadlink
+fi
+IMPORT_TEST_TYPE=${1}
```
wechman, line 12: Is `${1}` somehow better than `"$1"`?

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                                        if [ "$DIFFVIEW" == "" ]
+                                        then
+                                            echo -e "ERROR: JSONS differ for $1: \n $DIFF \n"
+                                            echo "Expected:"
+                                            cat  expected.asm
+                                            echo "Obtained:"
+                                            cat obtained_direct_import_export.asm
+                                        else
+                                            # Use user supplied diff view binary
+                                            $DIFFVIEW expected.asm obtained_direct_import_export.asm
```
wechman, line 174: Looks like duplication of lines 124-143. Could you take out this code to the helper function?

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+    exit 1
+fi
+
+# function tests whether exporting and importing again leaves the JSON ast unchanged
+# Results are recorded by adding to FAILED or UNCOMPILABLE.
+# Also, in case of a mismatch a diff and the respective ASTs are printed
+# Expected parameters:
+# $1 name of the file to be exported and imported
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
```
wechman, line 40: It would be great to have this function divided into smaller pieces since it is quite big by now.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+FAILED=0
+UNCOMPILABLE=0
+TESTED=0
+
+if [[ "$(find . -maxdepth 0 -type d -empty)" == "" ]]; then
+    echo "Test directory not empty. Skipping!"
+    exit 1
+fi
+
+# function tests whether exporting and importing again leaves the JSON ast unchanged
```
wechman, line 34: I think the comment may require an update.

wechman (commented):
It would be nice to have at least one test with asm-json before a merge.

scripts/ImportExportTest.sh
aarlt: Oh that's true! That seem to be a typo! Thx!

scripts/ImportExportTest.sh
aarlt: Nope, not really.. but you're right, this should be `"${1}"`.

scripts/ImportExportTest.sh
aarlt: True! I will add some helpers.

scripts/ImportExportTest.sh
aarlt: I agree!

scripts/ASTImportTest.sh
```diff
@@ -1,155 +0,0 @@
-#!/usr/bin/env bash
```
cameel, line 1: Could you do the file rename in a separate commit? I think this would help git detect that it's still basically the same script and show a better diff.

scripts/ImportExportTest.sh
cameel, line 70:
```suggestion
            DIFF="$(diff expected.json obtained.json || true)"
```

scripts/ImportExportTest.sh
cameel: They're equivalent but `${}` ensures that the text after the variable name won't be interpreted as a part of that name. Personally, I use the `"$x"` form only there's nothing else inside the quotes. If there's extra text, other variables, etc., I always use `${x}`.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if $SOLC --bin "$nth_input_file" "${all_input_files[@]}" > /dev/null 2>&1
+    then
+        ! [[ -e stderr.txt ]] || { echo "stderr.txt already exists. Refusing to overwrite."; exit 1; }
+
+        if [ "${IMPORT_TEST_TYPE}" == "ast" ]
```
cameel, line 48: Let's use the modern Bash variant.

```suggestion
        if [[ $IMPORT_TEST_TYPE == "ast" ]]
```
Same in the `elif` below.

test/cmdlineTests.sh
```diff
@@ -645,11 +645,11 @@ SOLTMPDIR=$(mktemp -d)
 )
 rm -r "$SOLTMPDIR"
 
-printTask "Testing AST import..."
+printTask "Testing AST import/export..."
 SOLTMPDIR=$(mktemp -d)
 (
     cd "$SOLTMPDIR"
-    if ! "$REPO_ROOT/scripts/ASTImportTest.sh"
+    if ! "$REPO_ROOT/scripts/ImportExportTest.sh" ast
```
cameel, line 652:
```suggestion
    if ! "$REPO_ROOT/scripts/importExportTest.sh" ast
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+    else
+        UNCOMPILABLE=$((UNCOMPILABLE + 1))
+    fi
+}
+
+WORKINGDIR=$PWD
+NSOURCES=0
+
+# check whether SOLC works.
+if ! $SOLC --version > /dev/null 2>&1
```
cameel, line 216: I wonder why shellcheck does not complain about this not being quoted:
```suggestion
if ! "$SOLC" --version > /dev/null 2>&1
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+    echo "$SOLC not found. aborting."
+    exit 1
+fi
+
+# check whether jq can be found.
+if ! jq --version > /dev/null 2>&1
+then
+    echo "jq needed. please install. aborting."
+    exit 1
+fi
```
cameel, line 227: You could make it a helper function and add it to [`scripts/common.sh`](https://github.com/ethereum/solidity/blob/develop/scripts/common.sh).

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+if ! $SOLC --version > /dev/null 2>&1
+then
+    echo "$SOLC not found. aborting."
+    exit 1
+fi
+
+# check whether jq can be found.
+if ! jq --version > /dev/null 2>&1
+then
+    echo "jq needed. please install. aborting."
```
cameel, line 225: This should be written to stderr.

But it would be best to just use the [`fail()`](https://github.com/ethereum/solidity/blob/v0.8.13/scripts/common.sh#L81-L87) helper.

scripts/ImportExportTest.sh
cameel, line 12: I'd put commandline argument processing as the first thing in the script. Just after `set -e`.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
+#!/usr/bin/env bash
+
+set -e
```
cameel, line 3: Would this script work with these?
```suggestion
set -euo pipefail
```

I wanted to do that last time I was going through these scripts but I did not have time to test each one individually and make sure it does not break anything. Now that you're rewriting it, it looks like a good opportunity to try this.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+then
+    echo "jq needed. please install. aborting."
+    exit 1
+fi
+
+# for solfile in $(find $DEV_DIR -name *.sol)
+# boost_filesystem_bug specifically tests a local fix for a boost::filesystem
+# bug. Since the test involves a malformed path, there is no point in running
+# AST tests on it. See https://github.com/boostorg/filesystem/issues/176
+if [ "${IMPORT_TEST_TYPE}" == "ast" ]
```
cameel, line 233:
```suggestion
if [[ $IMPORT_TEST_TYPE == "ast" ]]
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+then
+    IMPORT_TEST_FILES=$(find "${SYNTAXTESTS_DIR}" "${ASTJSONTESTS_DIR}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
+elif [ "${IMPORT_TEST_TYPE}" == "evm-assembly" ]
+then
+    IMPORT_TEST_FILES=$(find "${SYNTAXTESTS_DIR}" "${SEMANTICTESTS_DIR}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
+else
+    echo "unknown import test type. aborting. please specify $0 [ast|evm-assembly]."
+    exit 1
```
cameel, line 241:
```suggestion
    fail "Unknown import test type. Aborting. Please specify ${0} [ast|evm-assembly]."
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+fi
+
+# for solfile in $(find $DEV_DIR -name *.sol)
+# boost_filesystem_bug specifically tests a local fix for a boost::filesystem
+# bug. Since the test involves a malformed path, there is no point in running
+# AST tests on it. See https://github.com/boostorg/filesystem/issues/176
+if [ "${IMPORT_TEST_TYPE}" == "ast" ]
+then
+    IMPORT_TEST_FILES=$(find "${SYNTAXTESTS_DIR}" "${ASTJSONTESTS_DIR}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
```
cameel, line 235: To avoid repeating the whole command, you could just set the list of dirs here:
```suggestion
    TEST_DIRS=("$SYNTAXTESTS_DIR" "$ASTJSONTESTS_DIR")
```
Then have this for both cases:
```bash
    IMPORT_TEST_FILES=$(find "${TEST_DIRS[@]}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
```

aarlt:
Moved to #12834. Closing.

cameel (commented):
Oh, I was in the middle of reviewing this...
Looks like I'll need to move my comments to the new PR.

aarlt:
> Oh, I was in the middle of reviewing this...
> Looks like I'll need to move my comments to the new PR.

Oh!! Sorry for that! I think there is no need to move the comments.. I will try to integrate your feedback from here on the other PR.. I hope it will not create too much confusion

cameel:
Well, you did say on the meeting that you'll be moving this to a bigger PR where tests actually test something so it's my fault. But no worries. Just copying them fortunately was not that bad. They're all in #12834 now (except 1 or 2 that were basically just responses to other comments).

PR #12810, opened 2022-03-16
Fixed ambiguity in the create2 address computation doc

hroussille:
specify that arguments are encoded using abi.encode in both the text and the example code

### DIFF
--- a/docs/control-structures.rst
+++ b/docs/control-structures.rst
@@ -283,7 +283,7 @@ which only need to be created if there is a dispute.
                 salt,
                 keccak256(abi.encodePacked(
                     type(D).creationCode,
-                    arg
+                    abi.encode(arg)
                 ))
             )))));
 


docs/control-structures.rst
```diff
@@ -250,7 +250,7 @@ use a different mechanism to come up with the address of the new contract:
 
 It will compute the address from the address of the creating contract,
 the given salt value, the (creation) bytecode of the created contract and the constructor
-arguments.
+arguments encoded using ``abi.encode``.
```
chriseth: While his is true, we don't specify how the other elements are used specifically to compute the address. Furthermore, as it is now, it is not clear if `abi.encode` only applies to the arguments or to the whole thing.

Maybe it's better to limit this change to only adding `abi.encode` in the code below?

ekpyron:
Replacing ``abi.encodePacked`` altogether with ``bytes.concat`` would also work, right?
I.e.
```
            keccak256(bytes.concat(
                type(D).creationCode,
                abi.encode(arg)
            ))
```
would be nicer.

docs/control-structures.rst
hroussille: That's true... The only important change is the code example anyway, after re-reading the text, it is a more high level explanation, giving details on every encoding here would probably be too heavy to read and redundant with the code example.

Do you want another PR focusing only on the code change ?

chriseth: No need for a new PR, you can amend the commit ("git commit --amend") and then force-push ("git push --force").

chriseth (approved):

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12804, opened 2022-03-16
Benchmark differ

cameel:
~~**Depends on #12868.**~~

This is the initial version of my script for comparing benchmarks.

I still have a few small features to add (displaying absolute values in the table, filtering presets, etc.) but I'll be doing these in separate PRs. Adding it to CI so that diffs are generated automatically as artifacts will also be a separate PR.

PR #12818, opened 2022-03-18
Benchmark downloader (script + CI job)

cameel:
~Depends on #12804.~ Merged.

The script makes several CirecleCI and Github API requests to get the JSON with benchmark results for a specific PR or branch. I think it will be very handy for manual use, though my primary goal here is to use it in `c_ext_benchmarks` job to get benchmarks from the base PR and provide the diff. The PR extends that job to generate such diffs and attach them as artifacts.

After the experiences with #12181 and https://github.com/ethereum/solc-bin/pull/35 I decided to use Python to make this more maintainable. This way I'll be able to cover it with tests, and, compared to my earlier Bash scripts, it was easy to make a module with helpers that we'll be able to reuse in other situations that require API access. While it isn't that short as a whole, it's more concise compared to the previous Bash scripts - the main logic specific to this task is concentrated in one function and the rest are reusable helpers and command-line argument definitions (which is more like docs than real code).

I'm on purpose not trying to do too much error handling to keep things simple. I.e. it either works or you get an exception. Only the cases that would not just crash on unexpected response (like finding no matching results) are explicitly checked.

~This is still work in progress. I still need to~ Everything is done:
- [x] Add some unit tests.
- [x] Handle the case where the job has not finished yet.
- [x] Test some corner cases to make sure things fail properly on errors.
- [x] Getting artifacts for the specific revision of base branch that the requested PR/branch is based on. This would be helpful in cases where base branch has moved on since the PR/branch was created and has extra commits that affect benchmark results.

### Usage of the script
To get benchmarks from your branch and from develop, you just run this:
```bash
scripts/externalTests/download_benchmarks.py --branch develop
scripts/externalTests/download_benchmarks.py
```
This stores `summarized-benchmarks-<branch>-<commit hash>.json` and `all-benchmarks-<branch>-<commit hash>.json` in the current directory. Now you can run the script from #12804 on the files.

If you want to compare someone else's PR you probably have the PR number handy - you can pass that to the script using the `--pr` option instead of the branch.

You can also use `--base-of-pr` option to get artifacts from the branch your PR is based on (instead of `develop`). By default the script finds a CI run matching not only the base branch name but also the exact commit. This guarantees that even if the base branch has moved on  and has extra commits, your comparison will not be influenced by these commits.

Sometimes you might still want to get the benchmark from the latest state of the base branch, e.g. if there was no successful run on the commit you're based on and you know that any extra commits are not going to distort results all that much - in that case you can use `--any-commit` flag to bypass the commit check.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -690,9 +690,16 @@ jobs:
           name: Install pip
           command: apt -q update && apt install -y python3-pip
       - run:
-          name: Install pylint
-          command: python3 -m pip install pylint z3-solver pygments-lexer-solidity parsec tabulate deepdiff colorama
-          # also z3-solver, parsec and tabulate to make sure pylint knows about this module, pygments-lexer-solidity for docs
+          name: Install pylint and dependencies of the scripts that will be linted
+          command: python3 -m pip install
+            pylint
+            z3-solver
+            pygments-lexer-solidity
+            parsec
+            tabulate
+            deepdiff
+            colorama
+            requests
       - run:
           name: Linting Python Scripts
           command: ./scripts/pylint_all.py
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1235,15 +1235,64 @@ jobs:
       - attach_workspace:
           at: .
       - run:
+          name: Install dependencies of helper scripts
+          command: |
+            sudo apt update
+            sudo apt install python3-pip --assume-yes --no-install-recommends
+            python3 -m pip install requests --user
+      - run:
           name: Combine benchmark reports
           command: cat reports/externalTests/benchmark-*.json | scripts/externalTests/merge_benchmarks.sh > reports/externalTests/all-benchmarks.json
       - run:
           name: Summarize reports
           command: cat reports/externalTests/all-benchmarks.json | scripts/externalTests/summarize_benchmarks.sh > reports/externalTests/summarized-benchmarks.json
+      - run:
+          name: Download reports from base branch
+          command: |
+            if [[ $CIRCLE_PULL_REQUEST != "" ]]; then
+              mkdir reports/externalTests/base-branch/
+              cd reports/externalTests/base-branch/
+
+              pr_id=$(echo "$CIRCLE_PULL_REQUEST" | sed 's|\(.*\)\/||')
+              scripts_dir=../../../scripts
+
+              "${scripts_dir}/externalTests/download_benchmarks.py" --base-of-pr "$pr_id"
+            fi
+      - run:
+          name: Diff benchmarks
+          command: |
+            if [[ $CIRCLE_PULL_REQUEST != "" ]]; then
+              cd reports/externalTests/
+              mkdir diff/
+              scripts_dir=../../scripts
+
+              "${scripts_dir}/externalTests/benchmark_diff.py" table \
+                --output-format markdown \
+                --style humanized \
+                base-branch/summarized-benchmarks-*.json \
+                summarized-benchmarks.json > diff/benchmark-diff-summarized-table-markdown-humanized.md
+              "${scripts_dir}/externalTests/benchmark_diff.py" table \
+                --output-format markdown \
+                --style absolute \
+                base-branch/summarized-benchmarks-*.json \
+                summarized-benchmarks.json > diff/benchmark-diff-summarized-table-markdown-absolute.md
+              "${scripts_dir}/externalTests/benchmark_diff.py" inplace \
+                --style absolute \
+                base-branch/summarized-benchmarks-*.json \
+                summarized-benchmarks.json > diff/benchmark-diff-summarized-inplace-absolute.md
+              "${scripts_dir}/externalTests/benchmark_diff.py" inplace \
+                --style absolute \
+                base-branch/all-benchmarks-*.json \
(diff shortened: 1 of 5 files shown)


cameel:
This is now done and ready for review.

.circleci/config.yml
```diff
@@ -1223,16 +1230,65 @@ jobs:
...
+                base-branch/summarized-benchmarks-*.json \
+                summarized-benchmarks.json > diff/benchmark-diff-summarized-table-markdown-absolute.md
+              "${scripts_dir}/externalTests/benchmark_diff.py" inplace \
+                --style absolute \
+                base-branch/summarized-benchmarks-*.json \
+                summarized-benchmarks.json > diff/benchmark-diff-summarized-inplace-absolute.md
+              "${scripts_dir}/externalTests/benchmark_diff.py" inplace \
+                --style absolute \
+                base-branch/all-benchmarks-*.json \
+                all-benchmarks.json > diff/benchmark-diff-all-table-inplace-absolute.md
```
cameel, line 1286: We might not need all of these. `benchmark-diff-summarized-table-markdown-humanized.md` will probably be the most useful one in practice. It will soon be possible to include extra columns with absolute values in it as well.

On the other hand the overhead of having all of them is negligible and also this acts as an end to end test for the script so I included them here anyway.

cameel:
Note, about the `c_ext_benchmarks` failure:
```
Looking for pipelines that ran on branch benchmark-differ, commit fd0ec2e30a8d9c60bfbb15a984cf90bb0f571908.
[ERROR] Job c_ext_benchmarks has failed or is still running. Current status: blocked.
```

This is a limitation of the script with regard to PRs that are based one on the other. The job tries to access artifacts from the same job on the base PR. If both were pushed at the same time, it's likely that that base job has not finished yet. In that case if just needs to be reran after the base job finishes.

test/scripts/test_externalTests_benchmark_downloader.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python3
```
ekpyron, line 1: I always find it admirable how commited you are to unit-testing everything ;-). But I'm not entirely sure, in this particular case, that this is worth 250 lines in the repo :-). You also don't run this anywhere, right?
I'm fine with keeping it - it may also just serve to document the helper scripts... but I also wouldn't say we really need this :-).

ekpyron (dismissed):
This is a lot of infrastructure for just downloading some files :-). But probably not much one can do to avoid that...

I haven't reviewed the python scripts at all, but - as usual for these PRs - I'd say it's fine to merge based on them working properly in CI - and it being nice to have them, but them being non-critical.

If anyone is up for going through the python scripts once, though, it wouldn't hurt.

test/scripts/test_externalTests_benchmark_downloader.py
cameel: > You also don't run this anywhere, right?

True, you cannot run this on the command line (it's not even executable), the shebang should not be there :) I'll remove it.

> But I'm not entirely sure, in this particular case, that this is worth 250 lines in the repo :-).

Honestly, I'd say that the amount of tests here is pretty low for a scripted language. In other projects I could rarely get away with fewer lines in tests than in the script itself :) Interpreted code is just so brittle if you don't execute everything at least once. One of the things I hate when dealing with our Bash scripts is that I often cannot be sure that my changes do not break something. This is often a problem with ones that are not even plugged into CI (like PPA build or wasm rebuilds). It was actually a big obstacle back when I was doing a mass refactor to enable shellcheck. With Python we can avoid that and I think that each script should have at least some minimal coverage.

And this **is** minimal. These aren't even proper unit tests covering each function in isolation. I'm only calling the top-level function with the each important combination of options. And only ~100 lines here are really tests. The rest are JSON fixtures of responses (which I could move to standalone files in `fixtures/` but I thought it was better to have them here).

It also did actually help me to find some subtle bugs I had. E.g. initially I was sorting it wrong and not actually choosing the latest run, which is not that easy to tell unless you actually tediously inspect all the responses. I could do more thorough manual testing but I really find it more convenient to have any tests I did in the form of a test suite I can rerun. Maybe I'm just spoiled by tests covering our C++ code. I want the same thing for scripts :)

cameel:
> This is a lot of infrastructure for just downloading some files :-). But probably not much one can do to avoid that...

Unfortunately, dealing with APIs reliably is just verbose. This is not the first time I need to access their APIs and I always end up with a really ugly Bash script. We really needed some wrappers for that that we can later reuse as needed.

CircleCI actually has its own CLI utility called `circleci` and I was initially hoping I'd get away just using that instead of writing it in Python. But it does not really have the same use case. It does not have commands for fetching info from the API, only for controlling the pipelines from CLI.

Also, to be fair, it adds more value than just "downloading some files". It automates finding the exact runs that you need which is more tedious than it seems at first. E.g. if your branch is based on `develop` and something got merged in the meantime, you have to dig into history find the old commit ID, then locate the right run via this horrible list of pipelines in CircleCI's UI with infinite scroll. It's painful to use, especially If what you're looking for is not immediately on the first page. With the script you just give it a PR number and it either gives you the right benchmark or fails. Or you go to the artifacts in your PR and find the diff attached there. You don't have to think about it, which is how it should work.

chriseth (approved):

PR #12826, opened 2022-03-21
use special prerelease version precedence check only if major, minor and patch identifiers match

### LINKED ISSUE #12812: `pragma solidity` with partial versions in match expression broken on prerelease builds
## Description

```
pragma solidity >= 0.8;

contract MyContract { }
```

gives

```
Error: Source file requires different compiler version (current compiler is 0.8.13-develop.2022.3.14+commit.57e012da.Linux.g++) - note that nightly builds are considered to be strictly less than the released version
 --> mvce._version.sol:1:1:
  |
1 | pragma solidity >= 0.8;
  | ^^^^^^^^^^^^^^^^^^^^^^^
```

but there shouldn't have been an error as `0.8.13-develop....` is strictly greater than `0.8.12` which in turn is greater than `0.8`.

The following test cases work as expected:

```
pragma solidity >= 0.8.13; // error but that's correct since the compiler is the prerelease version
pragma solidity >= 0.8.12; // no error; correct
pragma solidity >= 0.7; // no error; correct
```

## Environment

- Compiler version: 0.8.13-develop.2022.3.14+commit.57e012da.Linux.g++
- Target EVM version (as per compiler settings):
- Framework/IDE (e.g. Truffle or Remix): Local
- EVM execution environment / backend / blockchain client: N/A
- Operating system: Ubuntu 20.04

## Steps to Reproduce

```
pragma solidity >= 0.8;

contract MyContract { }
```

compile with a prerelease build of the compiler whose version is `0.8.x`.

YashasSamaga:
Item 11.3 in semvar specs says prerelease check is performed if and only if the major, minor and patch versions match. This commit adds a check to perform prerelease check only when the major, minor and patch version match.

fixes #12812 

**Pending:**
- add tests

### DIFF
--- a/liblangutil/SemVerHandler.cpp
+++ b/liblangutil/SemVerHandler.cpp
@@ -116,7 +116,7 @@ bool SemVerMatchExpression::MatchComponent::matches(SemVerVersion const& _versio
 				cmp = static_cast<int>(_version.numbers[i] - version.numbers[i]);
 			}
 
-		if (cmp == 0 && !_version.prerelease.empty() && didCompare)
+		if (cmp == 0 && levelsPresent == 3 && !_version.prerelease.empty() && didCompare)
 			cmp = -1;
 
 		switch (prefix)


liblangutil/SemVerHandler.cpp
```diff
@@ -116,7 +116,7 @@ bool SemVerMatchExpression::MatchComponent::matches(SemVerVersion const& _versio
 				cmp = static_cast<int>(_version.numbers[i] - version.numbers[i]);
 			}
 
-		if (cmp == 0 && !_version.prerelease.empty() && didCompare)
+		if (cmp == 0 && levelsPresent == 3 && !_version.prerelease.empty() && didCompare)
```
cameel, line 119: Oh, I just noticed that `0.8` is apparently equivalent to `0.8.*` and not `0.8.0`. I did not expect that. Seems correct though, since the `semver` utility behaves the same way.

cameel, line 119: Can you add a test case specifically for the `0.8.13-develop...` case you're fixing?

Some other similar cases would be useful too. In general, please try to think about other, potentially problematic cases related to this and add them to the test suite.

leonardoalt:
Needs update

leonardoalt:
Doesn't look like anyone is working on this. Closing for now.

PR #12828, opened 2022-03-21
Use error message for unary operators.

chriseth:


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1663,19 +1663,22 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
...
 		_operation.subExpression().accept(*this);
 	Type const* subExprType = type(_operation.subExpression());
-	Type const* t = type(_operation.subExpression())->unaryOperatorResult(op);
-	if (!t)
+	TypeResult result = type(_operation.subExpression())->unaryOperatorResult(op);
+	if (!result)
 	{
 		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->toString();
+		if (!result.message().empty())
+			description += ": " + result.message();
```
cameel: I think that a dot might look better since we already have one colon in the message.
```suggestion
			description += ". " + result.message();
```

cameel (dismissed):

nishant-sachdeva (commented):
LGTM

libsolidity/analysis/TypeChecker.cpp
leonardoalt: Agree this should be changed.

leonardoalt (dismissed):
Please replace colon by period.

cameel (approved):

PR #11550, opened 2021-06-17
Constant for default --runs

cameel:
Related to https://github.com/ethereum/solidity/pull/11518#discussion_r653526253.

PR #10286, opened 2020-11-13
Optimise libevmasm in yul

axic:
Fixes #9386.
Depends on #11604.

PR #12834, opened 2022-03-22
libevmasm: Add support to import evm assembly json.

### LINKED ISSUE #11787: Compiler mode to reimport assembly json.

aarlt:
Replaces part of #12704.
Replaced by #13673.

**Depends on #13576, ~#13577~, ~#13578~ and #13579**

Closes #11787.


scripts/ASTImportTest.sh
```diff
@@ -1,155 +0,0 @@
-#!/usr/bin/env bash
```
cameel, line 1: Could you do the file rename in a separate commit? I think this would help git detect that it's still basically the same script and show a better diff.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
+#!/usr/bin/env bash
+
+set -e
```
cameel: Would this script work with these?
```suggestion
set -euo pipefail
```

I wanted to do that last time I was going through these scripts but I did not have time to test each one individually and make sure it does not break anything. Now that you're rewriting it, it looks like a good opportunity to try this.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+
+# Bash script to test the ast-import option of the compiler by
+# first exporting a .sol file to JSON, then loading it into the compiler
+# and exporting it again. The second JSON should be identical to the first
+READLINK=readlink
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    READLINK=greadlink
+fi
+IMPORT_TEST_TYPE=${1}
```
cameel: I'd put commandline argument processing as the first thing in the script. Just after `set -e`.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if $SOLC --bin "$nth_input_file" "${all_input_files[@]}" > /dev/null 2>&1
+    then
+        ! [[ -e stderr.txt ]] || { echo "stderr.txt already exists. Refusing to overwrite."; exit 1; }
+
+        if [ "${IMPORT_TEST_TYPE}" == "ast" ]
```
cameel: Let's use the modern Bash variant.

```suggestion
        if [[ $IMPORT_TEST_TYPE == "ast" ]]
```
Same in the `elif` below.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                echo "Compiler stderr:"
+                cat ./stderr.txt
+                echo
+                echo "Compiler stdout:"
+                cat ./obtained.json
+                return 1
+            fi
+            set +e
+            DIFF="$(diff expected.json obtained.json)"
+            set +e
```
cameel:
```suggestion
            DIFF="$(diff expected.json obtained.json || true)"
```
There are also a few more places like this below.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+    else
+        UNCOMPILABLE=$((UNCOMPILABLE + 1))
+    fi
+}
+
+WORKINGDIR=$PWD
+NSOURCES=0
+
+# check whether SOLC works.
+if ! $SOLC --version > /dev/null 2>&1
```
cameel: I wonder why shellcheck does not complain about this not being quoted:
```suggestion
if ! "$SOLC" --version > /dev/null 2>&1
```
Looks like there are many instances of this not being quoted all around the file.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+then
+    echo "$SOLC not found. aborting."
```
cameel: This should be printed to stderr.

But it would be best to just use the [`fail()`](https://github.com/ethereum/solidity/blob/v0.8.13/scripts/common.sh#L81-L87) helper.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+    exit 1
+fi
+
+# check whether jq can be found.
+if ! jq --version > /dev/null 2>&1
+then
+    echo "jq needed. please install. aborting."
+    exit 1
+fi
```
cameel: You could make it a helper function and add it to [`scripts/common.sh`](https://github.com/ethereum/solidity/blob/develop/scripts/common.sh).

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+
+# for solfile in $(find $DEV_DIR -name *.sol)
+# boost_filesystem_bug specifically tests a local fix for a boost::filesystem
+# bug. Since the test involves a malformed path, there is no point in running
+# AST tests on it. See https://github.com/boostorg/filesystem/issues/176
+if [ "${IMPORT_TEST_TYPE}" == "ast" ]
```
cameel:
```suggestion
if [[ $IMPORT_TEST_TYPE == "ast" ]]
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+then
+    IMPORT_TEST_FILES=$(find "${SYNTAXTESTS_DIR}" "${ASTJSONTESTS_DIR}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
```
cameel: To avoid repeating the whole command, you could just set the list of dirs here:
```suggestion
    TEST_DIRS=("$SYNTAXTESTS_DIR" "$ASTJSONTESTS_DIR")
```
Then have this for both cases:
```bash
    IMPORT_TEST_FILES=$(find "${TEST_DIRS[@]}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+elif [ "${IMPORT_TEST_TYPE}" == "evm-assembly" ]
+then
+    IMPORT_TEST_FILES=$(find "${SYNTAXTESTS_DIR}" "${SEMANTICTESTS_DIR}" -name "*.sol" -and -not -name "boost_filesystem_bug.sol")
+else
+    echo "unknown import test type. aborting. please specify $0 [ast|evm-assembly]."
+    exit 1
```
cameel:
```suggestion
    fail "Unknown import test type. Aborting. Please specify ${0} [ast|evm-assembly]."
```

solc/CommandLineInterface.cpp
```diff
@@ -161,7 +161,11 @@ static bool coloredOutput(CommandLineOptions const& _options)
 
 void CommandLineInterface::handleBinary(string const& _contract)
 {
-	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
+	solAssert(
+		m_options.input.mode == InputMode::Compiler ||
+		m_options.input.mode == InputMode::CompilerWithASTImport ||
+		m_options.input.mode == InputMode::CompilerWithEvmAssemblyJsonImport, ""
```
cameel: A message is now optional with `solAssert()` macro.
```suggestion
		m_options.input.mode == InputMode::CompilerWithEvmAssemblyJsonImport
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                return 2
+            fi
+            TESTED=$((TESTED + 1))
+            rm expected.json obtained.json
+            rm -f stderr.txt
+        elif [ "${IMPORT_TEST_TYPE}" == "evm-assembly" ]
+        then
+            local types=( "asm" "bin" "bin-runtime" "opcodes" "srcmap" "srcmap-runtime" )
+            local _TESTED=1
+            if ! $SOLC --combined-json bin,bin-runtime,opcodes,asm,srcmap,srcmap-runtime --pretty-json "$nth_input_file" "${all_input_files[@]}" > expected.json 2> expected.error
```
cameel:
```suggestion
            if ! $SOLC --combined-json "$(IFS=, ; echo "${types[*]}")" --pretty-json "$nth_input_file" "${all_input_files[@]}" > expected.json 2> expected.error
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                fi
+                FAILED=$((FAILED + 1))
+                return 2
+            fi
+            TESTED=$((TESTED + 1))
+            rm expected.json obtained.json
+            rm -f stderr.txt
+        elif [ "${IMPORT_TEST_TYPE}" == "evm-assembly" ]
+        then
+            local types=( "asm" "bin" "bin-runtime" "opcodes" "srcmap" "srcmap-runtime" )
```
cameel:
```suggestion
            local outputs=( "asm" "bin" "bin-runtime" "opcodes" "srcmap" "srcmap-runtime" )
```

scripts/ImportExportTest.sh
cameel: If JSON readability is important here (and it seems to be given that we're using `--pretty-json`) I'd also add `--json-indent 4`:
```suggestion
            if ! $SOLC --combined-json bin,bin-runtime,opcodes,asm,srcmap,srcmap-runtime --pretty-json --json-indent 4 "$nth_input_file" "${all_input_files[@]}" > expected.json 2> expected.error
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+            local _TESTED=1
+            if ! $SOLC --combined-json bin,bin-runtime,opcodes,asm,srcmap,srcmap-runtime --pretty-json "$nth_input_file" "${all_input_files[@]}" > expected.json 2> expected.error
+            then
+                printf "\n"
+                echo "$nth_input_file"
```
cameel: For consistency, either
```suggestion
                echo
                echo "$nth_input_file"
```
or
```suggestion
                printf "\n%s" "$nth_input_file"
```
There are a few more below.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                cat expected.error
+                UNCOMPILABLE=$((UNCOMPILABLE + 1))
+                return 0
+            else
```
cameel: Since the branch above just returns, you don't need the `else`. Since this goes on through several levels below, dropping the `else` would reduce nesting quite a bit.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                for contract in $(jq '.contracts | keys | .[]' expected.json 2> /dev/null)
```
cameel: Why silence stderr here? I think `jq` does not print anything there by default anyway and If something goes wrong here, we won't see the error.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if $SOLC --bin "$nth_input_file" "${all_input_files[@]}" > /dev/null 2>&1
+    then
+        ! [[ -e stderr.txt ]] || { echo "stderr.txt already exists. Refusing to overwrite."; exit 1; }
+
+        if [ "${IMPORT_TEST_TYPE}" == "ast" ]
+        then
```
cameel: I'd put the content of each branch of this `if` in a separate function. They're pretty much independent and we get less nesting and more structure that way.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                echo "$nth_input_file"
+                cat expected.error
+                UNCOMPILABLE=$((UNCOMPILABLE + 1))
+                return 0
+            else
+                for contract in $(jq '.contracts | keys | .[]' expected.json 2> /dev/null)
+                do
+                    for type in "${types[@]}"
+                    do
+                        jq --raw-output ".contracts.${contract}.\"${type}\"" expected.json > "expected.${type}"
```
cameel: The individual outputs are still JSON:
```suggestion
                        jq --raw-output ".contracts.${contract}.\"${type}\"" expected-all.json > "expected-${type}.json"
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                    done
+
+                    assembly=$(cat expected.asm)
+                    if [ "$assembly" != "" ] && [ "$assembly" != "null" ]
+                    then
+                        if ! $SOLC --combined-json bin,bin-runtime,opcodes,asm,srcmap,srcmap-runtime --pretty-json --import-asm-json expected.asm > obtained.json 2> obtained.error
+                        then
```
cameel: As pointed out by @wechman in #12809 already, this is repeating the bits above and could be extracted into a helper function.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                                        if [ "$DIFFVIEW" == "" ]
+                                        then
+                                            echo -e "ERROR: JSONS differ for $1: \n $DIFF \n"
+                                            echo "Expected:"
+                                            cat  "expected.${type}"
+                                            echo "Obtained:"
+                                            cat "obtained.${type}"
+                                        else
+                                            # Use user supplied diff view binary
+                                            $DIFFVIEW expected.json obtained.json
```
cameel:
```suggestion
                                            "$DIFFVIEW" expected.json obtained.json
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                                        fi
```
cameel: We have a [`diff_values()`](https://github.com/ethereum/solidity/blob/v0.8.13/scripts/common.sh#L174-L184) helper. How about adding `diff_files()` next to it and moving this part there so that it's reusable? You could then also add support for `$DIFFVIEW` to `diff_values()`.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                                    set -e
+                                    if [ "$DIFF" != "" ]
+                                    then
+                                        if [ "$DIFFVIEW" == "" ]
+                                        then
+                                            echo -e "ERROR: JSONS differ for $1: \n $DIFF \n"
+                                            echo "Expected:"
+                                            cat  "expected.${type}"
+                                            echo "Obtained:"
+                                            cat "obtained.${type}"
```
cameel: This should be printed to stderr.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                        else
+                            for type in "${types[@]}"
+                            do
+                                for obtained_contract in $(jq '.contracts | keys | .[]' obtained.json  2> /dev/null)
+                                do
+                                    jq --raw-output ".contracts.${obtained_contract}.\"${type}\"" obtained.json > "obtained.${type}"
+                                    set +e
+                                    DIFF="$(diff "expected.${type}" "obtained.${type}")"
+                                    set -e
+                                    if [ "$DIFF" != "" ]
```
cameel: I think it would be better to rely on the return value from `diff` here rather than on the output being empty.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                                    fi
+                                done
+                            fi
+
+                            rm obtained.json
+                            rm -f obtained.error
+                            for type in "${types[@]}"
+                            do
+                                rm "obtained.${type}"
+                            done
```
cameel:
```suggestion
                            rm "$(printf "obtained.%s " "${types[@]}")"
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,302 @@
...
+                        fi
+
+                        for type in "${types[@]}"
+                        do
+                            rm "expected.${type}"
+                        done
+                    fi
+                done
+                rm expected.json
```
cameel: I think there's no point removing these files given that the function runs in a loop that creates a new temporary dir in each cycle. So it will be gone anyway and for debugging you might want to inspect the files and it's easier to comment-out the single line deleting the whole dir than all the lines removing individual files.

solc/CommandLineInterface.cpp
```diff
@@ -187,7 +191,11 @@ void CommandLineInterface::handleBinary(string const& _contract)
 
 void CommandLineInterface::handleOpcode(string const& _contract)
 {
-	solAssert(m_options.input.mode == InputMode::Compiler || m_options.input.mode == InputMode::CompilerWithASTImport, "");
+	solAssert(
+		m_options.input.mode == InputMode::Compiler ||
+		m_options.input.mode == InputMode::CompilerWithASTImport ||
+		m_options.input.mode == InputMode::CompilerWithEvmAssemblyJsonImport, ""
+	);
```
cameel: Maybe you could create a constant representing all three modes? This used to be just a quick assert and now it got pretty verbose in quite a few places.
```suggestion
	solAssert(compilationInputModes.count(m_options.input.mode) > 0);
```

cameel (commented):
I originally wanted to review #12809 so this is just a review of the import test script or now.

scripts/ASTImportTest.sh
aarlt: I renamed the script in a separate commit, but GitHub seem to forget this.. However, maybe be this commit helps https://github.com/ethereum/solidity/pull/12834/commits/0b7afaf056e21e33fbb1ea1fd21dba36dca9f7de

scripts/ImportExportTest.sh
aarlt: Sadly its not just working. But you're right, it would make sense to do this. I will look into this tomorrow.

scripts/ImportExportTest.sh
aarlt: I added two helpers in `scripts/common.sh` - a helper for diffing files `diff_files`, and an helper that checks for executables `check_executable`.

scripts/ImportExportTest.sh
aarlt: True!

scripts/ImportExportTest.sh
aarlt: Ah that's nice!

scripts/ImportExportTest.sh
aarlt: Ah true, I still need to add `$DIFFVIEW` support for `diff_values`.

scripts/ImportExportTest.sh
aarlt: You're right!

scripts/ImportExportTest.sh
aarlt: Ok, I did this in https://github.com/ethereum/solidity/pull/12834/commits/a36a1a2c2dbea4ebfdb7c14c23e46276c1d43297. But somehow it feels like a hack.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+        cat ./stderr.txt >&2
+        printError
+        printError "Compiler stdout:"
+        cat ./obtained.json >&2
+        return 1
+    fi
+    set +e
+    diff_files expected.json obtained.json
+    DIFF=$?
+    set -euo pipefail
```
aarlt: hmm.. maybe its just ok to do `set -e` here.. not sure whether the other settings saved. @cameel do you know?

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+# Results are recorded by adding to FAILED or UNCOMPILABLE.
+# Also, in case of a mismatch a diff and the respective ASTs are printed
+# Expected parameters:
+# $1 name of the file to be exported and imported
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if [ -z ${all_input_files+x} ]
```
aarlt: Somehow those if's feel a bit like a hack. But maybe it's totally ok, because if a variable is not yet defined in any way, we provide with this the default value. @cameel How would you fix such unbound variable issues?

scripts/ImportExportTest.sh
cameel: Yeah, just `+e`/`-e` is probably enough.

cameel: But you can avoid this pattern completely with something like
```bash
if ! diff_files expected.json obtained.json
```

scripts/common.sh
```diff
@@ -26,6 +26,11 @@ set -e
 # changes directory. The paths returned by `caller` are relative to it.
 _initial_work_dir=$(pwd)
 
+if [ -z ${CIRCLECI+x} ]
+then
+    CIRCLECI=0
+fi
+
 if [ "$CIRCLECI" ]
```
cameel: Why not just this?
```suggestion
if [[ ${CIRCLECI:-} != "" ]]
```

I.e. just put the check directly wherever we need this variable. Setting it to `0` globally (especially in `common.sh` which gets sourced by many other scripts) could be interpreted in some places as true. Including the condition you have here (it treats an empty string as false).

Also the more modern `if` syntax and `${var:-default}`, to give it an empty string as a default value when the variable is undefined.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+
+        exit 1
+    fi
+
+    cd "$WORKINGDIR"
+    # Delete temporary files
+    rm -rf "$FILETMP"
+done
+
+echo ""
```
cameel:
```suggestion
echo
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+
+if [ "$FAILED" = 0 ]
```
cameel:
```suggestion
if (( $FAILED == 0 ))
```

scripts/ImportExportTest.sh
cameel:
```suggestion
    if [[ ${all_input_files:-} == "" ]]
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+# Expected parameters:
+# $1 name of the file to be exported and imported
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local nth_input_file="$1"
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if [ -z ${all_input_files+x} ]
+    then
+        all_input_files=( "" )
```
cameel: Isn't this equivalent to a single input file with an empty name? Didn't you mean this?

```suggestion
        all_input_files=()
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+        ! [[ -e stderr.txt ]] || { printError "stderr.txt already exists. Refusing to overwrite."; exit 1; }
+
+        if [[ $IMPORT_TEST_TYPE == "ast" ]]
+        then
+            Ast_ImportExportEquivalence "$nth_input_file" "${all_input_files[@]}"
+        elif [[ $IMPORT_TEST_TYPE == "evm-assembly" ]]
+        then
+            JsonEvmAsm_ImportExportEquivalence "$nth_input_file" "${all_input_files[@]}"
+        else
+            fail "unknown import test type. aborting."
```
cameel:
```suggestion
            fail "Unknown import test type. Aborting."
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,270 @@
...
+    IFS=" " read -r -a all_input_files <<< "$2"
+
+    if [ -z ${all_input_files+x} ]
+    then
+        all_input_files=( "" )
+    fi
+
+    if $SOLC --bin "$nth_input_file" "${all_input_files[@]}" > /dev/null 2>&1
+    then
+        ! [[ -e stderr.txt ]] || { printError "stderr.txt already exists. Refusing to overwrite."; exit 1; }
```
cameel:
```suggestion
        ! [[ -e stderr.txt ]] || fail "stderr.txt already exists. Refusing to overwrite."
```

scripts/ImportExportTest.sh
cameel: Maybe `command_available` would be a better name?

scripts/ImportExportTest.sh
aarlt: Yeah I also tried this one. Sadly it seem not to work.
```
line 162: all_input_files[@]: unbound variable
```
Looks like that there must be at least one element inside.

scripts/common.sh
aarlt: Thats much better!

scripts/ImportExportTest.sh
```diff
@@ -224,29 +231,35 @@ do
...
     then
         NSOURCES=$((NSOURCES - 1))
-        for i in $OUTPUT;
+        # shellcheck disable=SC2034
+        for i in $OUTPUT
         do
-            testImportExportEquivalence "$i" "$OUTPUT"
             NSOURCES=$((NSOURCES + 1))
         done
+        testImportExportEquivalence "$solfile" "${OUTPUT[@]}"
```
aarlt: @cameel not sure why this was done within the for loop before... but shouldn't this be also correct?

scripts/ImportExportTest.sh
cameel: Oh. But does it work? Because the problem I see is that `solc` gets an empty value as a file name and will try to load that from disk. Or does `( "" )` evaluate to an empty array or something?

scripts/ASTImportTest.sh
cameel: Thanks. Still looks like a new file in the combined diff but at least I changes in individual commits look reasonable. That helps!

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,283 @@
...
+#
+# solidity is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with solidity.  If not, see <http://www.gnu.org/licenses/>
+#
+# (c) 2016-2022 solidity contributors.
```
cameel: I don't think we'll ever be updating these dates so I'd put only the initial one here:
```suggestion
# (c) 2016 solidity contributors.
```

scripts/ImportExportTest.sh
```diff
@@ -108,44 +124,41 @@ function JsonEvmAsm_ImportExportEquivalence
...
-                            if [[ ${DIFF} != 0 ]]
-                            then
-                                _TESTED=
-                                FAILED=$((FAILED + 1))
-                                return 0
-                            fi
-                        done
+                        # reformat jsons - yeah, somehow its not generating the same format
+                        # todo: fix that
+                        jq . expected.asm.json > expected.asm.json.pretty
```
cameel: If it's just about indentation try `jq --indent 4`. Or are you seeing other differences?

aarlt: I did not investigate this further yet. I only saw that there was a diff regarding the arrays.
e.g.
```json
{
  "bla": [
    "a",
    "b"
  ]
}
```
and 
```json
{
  "bla": 
  [
    "a",
    "b"
  ]
}
```
Thats why I just reformat the output with `jq` here. I think there is something wrong with `--pretty-json`, but I did not look into this yet.

scripts/ImportExportTest.sh
cameel: True, running it for each file seems redundant. I think that the idea here was that you compile each file only with those files it actually depends on. There's no mechanism for checking that though and the script just passes in all the files :)

For some context, originally this had a bug where `$OUTPUT` was not quoted so it was only actually using the first file from the list (f712662a663e98733f9ad33ed141a25558921db7). It could be that the author noticed that it's not compiling all files but could not figure out why and made it a loop for a good measure. Or maybe he implemented it with just `$i` at first, then noticed that files are also needed but did not realize that it makes running in a loop pointless :)

In any case, your change here seems fine. I'd put it in a separate commit though instead of mixing it with the other stuff. Overall I think that keeping small commits, one per logical change is better for reviewing.

scripts/common.sh
```diff
@@ -185,7 +181,21 @@ function diff_values
...
-    diff --unified=0 <(echo "$value1") <(echo "$value2") "$@"
+    if ! diff --unified=0 <(echo "$value1") <(echo "$value2") "$@"
+    then
+        if [ "${DIFFVIEW:-}" == "" ]
+        then
+            printError "ERROR: values differ: '${value1}' vs. '${value2}'"
+            printError "Expected:"
+            printError "${value1}"
+            printError "Obtained:"
+            printError "${value2}"
```
cameel: Why repeat the values twice?

scripts/common.sh
```diff
@@ -185,7 +181,21 @@ function diff_values
...
+        else
+            # Use user supplied diff view binary
+            "$DIFFVIEW" "${file1}" "${file2}"
```
cameel, line 218: In this case I'd still print the error, just without showing full diff on the console.

scripts/ImportExportTest.sh
cameel: Both are valid ways to format this. In JSON whitespace does not matter so each formatter can do it its own way. solc uses json-cpp, which is apparently does not match `jq`.

Passing it through `jq` to get consistent formatting is fine.

scripts/ImportExportTest.sh
```diff
@@ -108,44 +124,41 @@ function JsonEvmAsm_ImportExportEquivalence
...
+                        jq . obtained_direct_import_export.json > obtained_direct_import_export.json.pretty
+                        set +e
+                        if ! diff_files expected.asm.json.pretty obtained_direct_import_export.json.pretty
```
cameel: Now that you have the diff command in the `if`, you do not need the `set +e`/`set -e` commands here.

scripts/ImportExportTest.sh
```diff
@@ -224,29 +231,35 @@ do
...
+        testImportExportEquivalence "$solfile" "$solfile"
     else
         # All other return codes will be treated as critical errors. The script will exit.
-        printError "\nGot unexpected return code ${SPLITSOURCES_RC} from ${SPLITSOURCES}. Aborting."
-        printError "\n${OUTPUT}\n"
+        echo >&2
+        printError "Got unexpected return code ${SPLITSOURCES_RC} from ${SPLITSOURCES}. Aborting."
+        echo >&2
+        printError "${OUTPUT}"
+        echo >&2
```
cameel: Why not keep the `\n` here?

leonardoalt:
PR seems to be waiting for code changes.

scripts/ImportExportTest.sh
aarlt: renamed.

solc/CommandLineInterface.cpp
aarlt: Will do the changes regarding the c++ code, after the import/export script seem to be ready.

scripts/ImportExportTest.sh
aarlt: Ah ok! Interesting! 

> Overall I think that keeping small commits, one per logical change is better for reviewing.

Yeah that's true!

scripts/ImportExportTest.sh
aarlt: I remove the year completely

scripts/ImportExportTest.sh
aarlt: Ok! I leave it with reformatting using `jq`.

scripts/common.sh
aarlt: True!

scripts/ImportExportTest.sh
aarlt: Oh! I thought that those escape sequences are not getting interpreted, but that seem to work with `printError`. I will revert this change.

aarlt: Ok, it looks like that under some circumstances escape sequences are not getting interpreted. I changed now the `print*` functions  in `common.sh` to use `echo -e` instead of just `echo`.

cameel:
@aarlt Does this PR close #11787 or will there be more coming? If so, please add it in the description. Also, other related PRs should mention that issue too in their descriptions so that it's easier to locate them later.

scripts/ImportExportTest.sh
aarlt: Script uses now newly added `diff_files` (defined in `common.sh`).

scripts/ImportExportTest.sh
aarlt: Sorry, for the late answer I needed to check again why I actually did this. The reason is simple, we have some files that only consists of free functions. So the resulting combined json files does not have any keys. To not get spammed of those errors eg. `jq: error: null (null) has no keys`, I redirected the errors to `/dev/null/`.

libevmasm/Assembly.h
```diff
@@ -152,6 +152,12 @@ class Assembly
 		bool _includeSourceList = true
 	) const;
 
+	/// Loads the JSON representation of assembly.
+	/// @param _json JSON object containing assembly
+	/// @param _loadSources true, if source list should be included, false otherwise.
+	/// @returns true on success, false otherwise
+	bool loadFromAssemblyJSON(Json::Value const& _json, bool _loadSources = true);
```
chriseth: Shouldn't this be a static factory function instead?

libevmasm/Assembly.h
```diff
@@ -160,6 +166,22 @@ class Assembly
...
+	}
+
+	/// Set the source list from simple vector<string>.
+	void setSources(std::vector<std::string> const& _sources)
+	{
+		for (auto const& item: _sources)
+			m_sources.emplace_back(std::make_shared<std::string>(item));
+	}
+
+	/// @returns List of sources.
```
chriseth: It is the list of source names, right?

libevmasm/Assembly.h
```diff
@@ -168,6 +190,14 @@ class Assembly
 
 	unsigned codeSize(unsigned subTagSize) const;
 
+	/// Add all assembly items from given JSON array.
+	void addAssemblyItemsFromJSON(Json::Value const& _code);
+
+	/// Creates an AssemblyItem from a given JSON representation.
+	/// @param _json JSON representation of an assembly item
+	/// @returns AssemblyItem from a given JSON representation.
```
chriseth: These two are totally redundant given the generic description string.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -414,6 +414,28 @@ void CompilerStack::importASTs(map<string, Json::Value> const& _sources)
...
 
+void CompilerStack::importEvmAssemblyJson(std::map<std::string, Json::Value> const& _sources)
+{
+	solAssert(_sources.size() == 1, "");
+	solAssert(m_sources.empty(), "");
+	solAssert(m_sourceOrder.empty(), "");
+	if (m_stackState != Empty)
+		solThrow(CompilerError, "Must call importEvmAssemblyJson only before the SourcesSet state.");
+
+	Json::Value jsonValue = _sources.begin()->second;
```
chriseth, line 428: This performs a deep copy.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -414,6 +414,28 @@ void CompilerStack::importASTs(map<string, Json::Value> const& _sources)
...
+	if (jsonValue.isMember("sourceList"))
+		for (auto const& item: jsonValue["sourceList"])
+		{
+			Source source;
+			source.charStream = std::make_shared<CharStream>(item.asString(), "");
+			m_sources.emplace(std::make_pair(item.asString(), source));
+			m_sourceOrder.push_back(&m_sources[item.asString()]);
+		}
+	m_evmAssemblyJson[_sources.begin()->first] = jsonValue;
```
chriseth: This performs a deep copy again.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
+		solAssert(m_importedSources, "");
+		solAssert(m_evmAssemblyJson.size() == 1, "");
+
+		string const evmSourceName = m_evmAssemblyJson.begin()->first;
+		Json::Value const evmJson = m_evmAssemblyJson.begin()->second;
+
+		evmasm::Assembly::OptimiserSettings optimiserSettings;
+		optimiserSettings.evmVersion = m_evmVersion;
+		optimiserSettings.expectedExecutionsPerDeployment = m_optimiserSettings.expectedExecutionsPerDeployment;
+		optimiserSettings.runCSE = m_optimiserSettings.runCSE;
```
chriseth: Don't we have a translation function here?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
-			if (auto contract = dynamic_cast<ContractDefinition const*>(node.get()))
-				if (isRequestedContract(*contract))
-				{
-					try
+	if (!m_evmAssemblyJson.empty())
+	{
+		solAssert(m_importedSources, "");
+		solAssert(m_evmAssemblyJson.size() == 1, "");
+
+		string const evmSourceName = m_evmAssemblyJson.begin()->first;
```
chriseth: Is this just the source name or also the name of the contract?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
+		if (m_optimiserSettings.enabled)
+			m_contracts[evmSourceName].evmRuntimeAssembly->optimise(optimiserSettings);
+		m_contracts[evmSourceName].runtimeObject = m_contracts[evmSourceName].evmRuntimeAssembly->assemble();
+	}
+	else
+	{
+		for (Source const* source: m_sourceOrder)
+			for (ASTPointer<ASTNode> const& node: source->ast->nodes())
+				if (auto contract = dynamic_cast<ContractDefinition const*>(node.get()))
+					if (isRequestedContract(*contract))
```
chriseth, line 717: This logic is completely missing above.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -649,55 +674,85 @@ bool CompilerStack::compile(State _stopAfter)
...
 	map<ContractDefinition const*, shared_ptr<Compiler const>> otherCompilers;
 
-	for (Source const* source: m_sourceOrder)
-		for (ASTPointer<ASTNode> const& node: source->ast->nodes())
-			if (auto contract = dynamic_cast<ContractDefinition const*>(node.get()))
-				if (isRequestedContract(*contract))
-				{
-					try
+	if (!m_evmAssemblyJson.empty())
+	{
```
chriseth, line 684: Can this if-body be somehow be combined with (some part of) `CompilerStack::assemble(...)`?

libsolidity/interface/CompilerStack.h
```diff
@@ -498,6 +502,7 @@ class CompilerStack: public langutil::CharStreamProvider
 	std::map<std::string const, Source> m_sources;
 	// if imported, store AST-JSONS for each filename
 	std::map<std::string, Json::Value> m_sourceJsons;
+	std::map<std::string, Json::Value> m_evmAssemblyJson;
```
chriseth: Please document. Also I'm not super happy to have this here...

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+		// During the assembly json export a `JUMPDEST` is always generated after a `tag`.
+		// So we just ignore exactly these `JUMPDEST`'s.
+		auto const next = std::next(current);
+		if (
+			next != m_items.end() &&
+			current->type() == AssemblyItemType::Tag &&
+			next->type() == AssemblyItemType::Operation &&
+			next->instruction() == Instruction::JUMPDEST
+		)
+			m_items.erase(next);
```
chriseth, line 95: Don't you need `current = m_items.erase(next)`? I might also be wrong...

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
 	}
 }
 
+void Assembly::addAssemblyItemsFromJSON(Json::Value const& _code)
+{
+	solAssert(_code.isArray(), "");
```
chriseth, line 80: Should this also assert that `m_items` is empty?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+			next->type() == AssemblyItemType::Operation &&
+			next->instruction() == Instruction::JUMPDEST
+		)
+			m_items.erase(next);
+	}
+}
+
+AssemblyItem Assembly::loadItemFromJSON(Json::Value const& _json)
+{
+	std::string name = _json["name"].isString() ? _json["name"].asString() : "";
```
chriseth: Shouldn't these all fail if the type is wrong? Unless it is optional, of course.

chriseth: Also could use some helpers `asStringOrDefault` / `asIntOrDefault`?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+	int srcIndex = _json["source"].isInt() ? _json["source"].asInt() : -1;
+	size_t modifierDepth = _json["modifierDepth"].isInt() ? static_cast<size_t>(_json["modifierDepth"].asInt()) : 0;
+	std::string value = _json["value"].isString() ? _json["value"].asString() : "";
+	std::string jumpType = _json["jumpType"].isString() ? _json["jumpType"].asString() : "";
+	solAssert(!name.empty(), "");
+
+	auto updateUsedTags = [&](u256 const& data) {
+		auto tag = static_cast<unsigned>(data);
+		if (m_usedTags <= tag)
+			m_usedTags = tag + 1;
```
chriseth: maybe
```suggestion
		m_usedTags = max(m_usedTags, static_cast<unsigned>(data) + 1);
```

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+		return data;
+	};
+
+	auto immutableHash = [&](string const& _immutableName) -> h256 {
+		h256 hash(util::keccak256(value));
```
chriseth: Where does `value` come from?

chriseth: Maybe make it an explicit parameter?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+	};
+
+	auto libraryHash = [&](string const& _libraryName) -> h256 {
+		h256 hash(util::keccak256(value));
+		m_libraries[hash] = _libraryName;
+		return hash;
+	};
+
+	SourceLocation location;
+	location.start = begin;
```
chriseth: Why not assign it right at the beginning / do you need the temporary variables begin and end?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+	location.end = end;
+	if (srcIndex > -1 && srcIndex < static_cast<int>(sources().size()))
```
chriseth, line 138: Should we allow out-of-bounds source indices?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+		location.sourceName = sources()[static_cast<size_t>(srcIndex)];
+
+	AssemblyItem result(0);
+
+	if (c_instructions.find(name) != c_instructions.end())
```
chriseth: `contains`

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+	{
+		AssemblyItem item{c_instructions.at(name), location};
+		item.m_modifierDepth = modifierDepth;
```
chriseth: This is assigned at the end.

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+		if (!jumpType.empty())
```
chriseth, line 146: Can't we do that at the end as well?

libevmasm/Assembly.cpp
```diff
@@ -74,6 +74,117 @@ unsigned Assembly::codeSize(unsigned subTagSize) const
...
+		else if (name == "ASSIGNIMMUTABLE")
+			result = {AssemblyItemType::AssignImmutable, immutableHash(value), location};
+		else if (name == "tag")
+			result = {AssemblyItemType::Tag, updateUsedTags(u256(value)), location};
+		else if (name == "PUSH data")
+			result = {AssemblyItemType::PushData, u256("0x" + value), location};
+		else if (name == "VERBATIM")
+		{
+			AssemblyItem item(fromHex(value), 0, 0);
+			item.setLocation(location);
```
chriseth: You could call `setLocation` at end end and just don't provide it in all the other constructors.

libevmasm/Assembly.h
aarlt: Yeah that might be better.

libsolidity/interface/CompilerStack.cpp
aarlt: changed to const ref

libsolidity/interface/CompilerStack.cpp
aarlt: Thats true! I saw that once too, but then I was not able to find it anymore. However, found it now `CompilerContext::translateOptimiserSettings` - refactored it - so there is now a static function `evmasm::Assembly::OptimiserSettings::translateSettings`.

libsolidity/interface/CompilerStack.cpp
aarlt: This is just the source name. Basically evm assembly don't know anything about contracts anymore.

libsolidity/interface/CompilerStack.cpp
aarlt: I would say this logic is not needed.. we can only import exactly one evm assembly json file.. or what do you mean?

libsolidity/interface/CompilerStack.cpp
aarlt: I guess that would be possible.. at least for now `CompilerStack::compile(...)` is right now not calling `CompilerStack::assemble(...)` - but I guess it would be better if the whole logic could be somehow moved to `CompilerStack::assemble(...)`. Will look into this later.

libsolidity/interface/CompilerStack.h
aarlt: Yeah, I know.. I was also not very sure whether I should just do it like this. Will think about this.

libevmasm/Assembly.cpp
aarlt: It is not needed. The iterator returned by `m_items.erase(next)` would point to the element after next (if it exists). After that, `current` would then be incremented again in the loop and point to the next (next) element. However, I was  not very sure so I just checked the  behaviour with `current = ...`: the import/export test script is not happy with that.

libevmasm/Assembly.cpp
aarlt: True. Maybe it's safer if we assert `m_items` is empty here. I think we don't really need to be able to add assembly items from json multiple times, and if we need that in the future, we can just remove the assert.

libevmasm/Assembly.cpp
aarlt: Yeah those helpers would make sense!

aarlt: True, only the 'name' is asserted so far. Need add an `assert` for the other members.

libevmasm/Assembly.cpp
aarlt: Oh wow! Those where typos! `immutableHash` and `libraryHash` should not use `value` here.

E.g.

```c++
	auto immutableHash = [&](string const& _immutableName) -> h256 {
		h256 hash(util::keccak256(_immutableName));
		m_immutables[hash] = _immutableName;
		return hash;
	};
```

libevmasm/Assembly.cpp
aarlt: True, this is not really needed especially after introducing those helpers `asStringOrDefault` / `asIntOrDefault`.

libsolidity/interface/CompilerStack.cpp
cameel: I originally did it like that in #10286, but back then @axic objected that it would introduce unnecessary cross dependencies (https://github.com/ethereum/solidity/pull/10286#issuecomment-827969786):

> If we are not doing it, I'd prefer the original version where we just copied the settings within libyul, without creating this weird set off cross dependencies between libevmasm <-> libyul <-> libsolidity.

libsolidity/interface/OptimiserSettings.h
```diff
@@ -149,6 +149,8 @@ struct OptimiserSettings
 	/// This specifies an estimate on how often each opcode in this assembly will be executed,
 	/// i.e. use a small value to optimise for size and a large value to optimise for runtime gas usage.
 	size_t expectedExecutionsPerDeployment = 200;
+	/// Flag reflecting whether optimizer is enabled.
+	bool enabled = false;
```
cameel, line 159: Do we really need this flag in `OptimiserSettings`? As far as I can tell it's not even respected everywhere - you're only handling it for EVM asm in `CompilerStack`. Other places ignore it so optimization might be performed even if it's `false`.

Another problem is that `optimizer.enabled = false` in Standard JSON (or the lack of `--optimize`) actually does not disable the optimizer completely. It corresponds to `OptimiserSettings::minimal()` and some optimizations are still performed. So there's no clear "disabled" state. The flag would be a bit misleading in that context.

If you still think it's a good idea to add it then I think it would be much better to have this and any changes related to it in a separate PR. I suspect that such a change will require modifications in many more places and is not all that much related to the asm import.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+READLINK=readlink
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    READLINK=greadlink
+fi
+REPO_ROOT=$(${READLINK} -f "$(dirname "$0")"/..)
+SOLIDITY_BUILD_DIR=${SOLIDITY_BUILD_DIR:-${REPO_ROOT}/build}
+SOLC=${SOLIDITY_BUILD_DIR}/solc/solc
+SPLITSOURCES=${REPO_ROOT}/scripts/splitSources.py
+
+# shellcheck disable=SC1090,SC1091
```
cameel: If you do this, shellcheck will actually find and analyze the sourced file:
```suggestion
# shellcheck source=scripts/common.sh
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+        printError "\n${OUTPUT[*]}\n"
+        echo >&2
+        testImportExportEquivalence "$solfile" "$solfile"
+    else
+        # All other return codes will be treated as critical errors. The script will exit.
+        echo >&2
+        printError "\nGot unexpected return code ${SPLITSOURCES_RC} from ${SPLITSOURCES}. Aborting."
+        echo >&2
+        printError "\n${OUTPUT[*]}\n"
+        echo >&2
```
cameel: These echos look equivalent to just adding more `\n` inside the print:
```suggestion
        printError "\n\nGot unexpected return code ${SPLITSOURCES_RC} from ${SPLITSOURCES}. Aborting."
        printError "\n\n${OUTPUT[*]}\n\n"
```

Same above.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+WORKINGDIR=$PWD
+NSOURCES=0
+
+command_available "${SOLC}" --version
+command_available jq --version
+
+# for solfile in $(find $DEV_DIR -name *.sol)
+# boost_filesystem_bug specifically tests a local fix for a boost::filesystem
+# bug. Since the test involves a malformed path, there is no point in running
+# AST tests on it. See https://github.com/boostorg/filesystem/issues/176
```
cameel: This comment refers to the ignored `boost_filesystem_bug.sol` so it should really be above the `IMPORT_TEST_FILES=` line below.

Also, the `for solfile` line is just some earlier, commented-out code. You can remove it.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+    OUTPUT=$("$SPLITSOURCES" "$solfile")
+    SPLITSOURCES_RC=$?
+    set -e
+    if [ ${SPLITSOURCES_RC} == 0 ]
+    then
+        OIFS=${IFS}
+        IFS=' ' read -ra OUTPUT_ARRAY <<< "${OUTPUT}"
+        IFS=${OIFS}
+        NSOURCES=$((NSOURCES - 1 + ${#OUTPUT_ARRAY[@]}))
+        testImportExportEquivalence "$solfile" "${OUTPUT[@]}"
```
cameel:
```suggestion
        testImportExportEquivalence "$solfile" "${OUTPUT_ARRAY[@]}"
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+    elif [ ${SPLITSOURCES_RC} == 1 ]
+    then
+        testImportExportEquivalence "$solfile" "$solfile"
+    elif [ ${SPLITSOURCES_RC} == 2 ]
+    then
+        # The script will exit with return code 2, if an UnicodeDecodeError occurred.
+        # This is the case if e.g. some tests are using invalid utf-8 sequences. We will ignore
+        # these errors, but print the actual output of the script.
+        echo >&2
+        printError "\n${OUTPUT[*]}\n"
```
cameel: `$OUTPUT` is not an array. Or did you mean `${OUTPUT_ARRAY[*]}` here? I'd use `${OUTPUT}` anyway.
```suggestion
        printError "\n${OUTPUT}\n"
```
Same below.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+        else
+            fail "Unknown import test type. Aborting."
+        fi
+    else
+        UNCOMPILABLE=$((UNCOMPILABLE + 1))
+    fi
+}
+
+WORKINGDIR=$PWD
+NSOURCES=0
```
cameel: `NSOURCES=0` seems unused. You just overwrite it below.

scripts/ImportExportTest.sh
cameel: `"${OUTPUT[@]}"` will expand into as many arguments as there are array items but `testImportExportEquivalence` currently only uses the first one (i.e. `$2`) and ignores the rest.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+
+    set +e
+    OUTPUT=$("$SPLITSOURCES" "$solfile")
+    SPLITSOURCES_RC=$?
+    set -e
+    if [ ${SPLITSOURCES_RC} == 0 ]
+    then
+        OIFS=${IFS}
+        IFS=' ' read -ra OUTPUT_ARRAY <<< "${OUTPUT}"
+        IFS=${OIFS}
```
cameel: You don't need to preserve the original value of `$IFS` here. The assignment takes effect only for the duration of the `read` command and then the original value is automatically restored.
```suggestion
        IFS=' ' read -ra OUTPUT_ARRAY <<< "${OUTPUT}"
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+    echo -n "."
+    # create a temporary sub-directory
+    FILETMP=$(mktemp -d)
+    cd "$FILETMP"
+
+    set +e
+    OUTPUT=$("$SPLITSOURCES" "$solfile")
+    SPLITSOURCES_RC=$?
+    set -e
+    if [ ${SPLITSOURCES_RC} == 0 ]
```
cameel:
```suggestion
    if [[ ${SPLITSOURCES_RC} == 0 ]]
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+# Also, in case of a mismatch a diff and the respective ASTs are printed
+# Expected parameters:
+# $1 name of the file to be exported and imported
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local sol_file="$1"
+    local input_files="$2"
+    if $SOLC --bin "${input_files}" > /dev/null 2>&1
+    then
+        ! [[ -e stderr.txt ]] || { fail "stderr.txt already exists. Refusing to overwrite."; }
```
cameel: You only need the `{}` here if you have multiple commands.
```suggestion
        ! [[ -e stderr.txt ]] || fail "stderr.txt already exists. Refusing to overwrite."
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+# function tests whether exporting and importing again leaves the JSON ast unchanged
+# Results are recorded by adding to FAILED or UNCOMPILABLE.
+# Also, in case of a mismatch a diff and the respective ASTs are printed
+# Expected parameters:
+# $1 name of the file to be exported and imported
+# $2 any files needed to do so that might be in parent directories
+function testImportExportEquivalence {
+    local sol_file="$1"
+    local input_files="$2"
+    if $SOLC --bin "${input_files}" > /dev/null 2>&1
```
cameel:
```suggestion
    if "$SOLC" --bin "${input_files}" > /dev/null 2>&1
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+
+        if [[ $IMPORT_TEST_TYPE == "ast" ]]
+        then
+            Ast_ImportExportEquivalence "${sol_file}" "${input_files}"
+        elif [[ $IMPORT_TEST_TYPE == "evm-assembly" ]]
+        then
+            JsonEvmAsm_ImportExportEquivalence "${sol_file}" "${input_files}"
+        else
+            fail "Unknown import test type. Aborting."
+        fi
```
cameel:
```suggestion
        case "$IMPORT_TEST_TYPE" in
            ast) Ast_ImportExportEquivalence "${sol_file}" "${input_files}" ;;
            evm-assembly) JsonEvmAsm_ImportExportEquivalence "${sol_file}" "${input_files}" ;;
            *) fail "Unknown import test type. Aborting." ;;
        esac
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+# AST tests on it. See https://github.com/boostorg/filesystem/issues/176
+if [[ $IMPORT_TEST_TYPE == "ast" ]]
+then
+    TEST_DIRS=("${SYNTAXTESTS_DIR}" "${ASTJSONTESTS_DIR}")
+elif [[ $IMPORT_TEST_TYPE == "evm-assembly" ]]
+then
+    TEST_DIRS=("${SYNTAXTESTS_DIR}" "${SEMANTICTESTS_DIR}")
+else
+    fail "Unknown import test type. Aborting. Please specify ${0} [ast|evm-assembly]."
+fi
```
cameel:
```suggestion
case "$IMPORT_TEST_TYPE" in
    ast) TEST_DIRS=("${SYNTAXTESTS_DIR}" "${ASTJSONTESTS_DIR}") ;;
    evm-assembly) TEST_DIRS=("${SYNTAXTESTS_DIR}" "${SEMANTICTESTS_DIR}") ;;
    *) fail "Unknown import test type. Aborting. Please specify ${0} [ast|evm-assembly]." ;;
esac
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+
+    if [ -n "${_TESTED}" ]
+    then
+        TESTED=$((TESTED + 1))
+    fi
+}
+
+# function tests whether exporting and importing again leaves the JSON ast unchanged
+# Results are recorded by adding to FAILED or UNCOMPILABLE.
+# Also, in case of a mismatch a diff and the respective ASTs are printed
```
cameel: This docstring is now out of date.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+
+FAILED=0
+UNCOMPILABLE=0
+TESTED=0
+
+if [[ "$(find . -maxdepth 0 -type d -empty)" == "" ]]; then
+    fail "Test directory not empty. Skipping!"
+fi
+
+function Ast_ImportExportEquivalence
```
cameel: I'd stick to either camel case or snake case here.
```suggestion
function astImportExportEquivalence
```
```suggestion
function ast_import_export_equivalence
```

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+{
+    local sol_file="$1"
+    local input_files="$2"
```
cameel, line 59: I see that now `sol_file` is only used in error messages. It looks a bit redundant. Maybe we could get rid of it and only have `input_files` here? For example we could print the original input file name before running this function so that it's not necessary to pass it in.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+            if ! diff_files expected.asm.json.pretty obtained_direct_import_export.json.pretty
+            then
+                _TESTED=
+                FAILED=$((FAILED + 1))
+                return 0
+            fi
+        fi
+    done
+
+    if [ -n "${_TESTED}" ]
```
cameel:
```suggestion
    if [[ $_TESTED == "" ]]
```

scripts/ImportExportTest.sh
cameel: Makes sense. Could you put that in a comment though? I think we should leave stderr enabled and only disable it when it's too noisy. And it helps if it's explained why it's disabled in a particular case.

scripts/ImportExportTest.sh
```diff
@@ -0,0 +1,276 @@
...
+    for contract in $(jq '.contracts | keys | .[]' expected.json 2> /dev/null)
+    do
+        for output in "${outputs[@]}"
+        do
+            jq --raw-output ".contracts.${contract}.\"${output}\"" expected.json > "expected.${output}.json"
+        done
+
+        assembly=$(cat expected.asm.json)
+        if [ "$assembly" != "" ] && [ "$assembly" != "null" ]
+        then
```
cameel:
```suggestion
        assembly=$(cat expected.asm.json)
        [[ $assembly != "" && $assembly != "null" ]] || return
```

solc/CommandLineInterface.cpp
aarlt: I would prefer to be more explicit. But I agree that it became too verbose. Your suggestion is not really reflecting what modes are exactly allowed. Maybe I will create some helper functions, so that e.g. something like this is possible `solAssert(or(m_options.input.mode, {InputMode::Compiler, InputMode::CompilerWithASTImport, InputMode::CompilerWithEvmAssemblyJsonImport})` - where `{InputMode::Compiler, InputMode::CompilerWithASTImport, InputMode::CompilerWithEvmAssemblyJsonImport}` could be defined somewhere and just get reused. Let me think a bit more to find maybe a even better solution. But you're totally right, it's getting too verbose.

aarlt: Ok, I introduced `any_of` and `none_of` that are now used in these macros. See commit https://github.com/ethereum/solidity/pull/12834/commits/d6f7aeac41bf4db2b56cfcffdee160a7477fa60a. I added those function in `solc/CommandLineParser.h`. Maybe we should move those to a more generic place. What would you say, where is a good place for these functions?

cameel: > Your suggestion is not really reflecting what modes are exactly allowed.

`compilationInputModes` was meant to be a constant `map` or `set` (I was hoping it was obvious from `count()`) that has all the `Compiler*` modes set. With that it would work, wouldn't it?

And sure, `any_off`/`none_of` are fine but you would not need them if you used a `set` :)

> What would you say, where is a good place for these functions?

We already have [`std::ranges::all_of`, `std::ranges::any_of`, `std::ranges::none_of`](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of) in ranges v3. Why not use those?

libsolidity/interface/CompilerStack.cpp
aarlt: Changed that back to a deep copy, not sure how to avoid that here. However, I changed line `434` to use `std::move`.

libsolidity/interface/CompilerStack.cpp
aarlt: Changed that to use `std::move(jsonValue)`.

libsolidity/interface/CompilerStack.cpp
aarlt: Hmm.. I can see that the dependency between `libevmasm <-> libsolidity` was already introduced by https://github.com/ethereum/solidity/pull/11550. I saw that you where asking for feedback from @axic in that PR. However, it got merged like this. Thats why I would argue that it is ok to have these dependencies - because they are already there.

Looking at the current dependencies again: `libsolidity/codegen/CompilerContext.cpp` is now using the translate function defined in "libevmasm/Assembly.cpp". So we have a dependency between `libsolidity->libevmasm`. This function uses a type `frontend::OptimiserSettings` from `libsolidity` - but we also use `evmasm::AssemblyItems` here. That means we have now `libsolidity <-> libevmasm`. `libsolidity/codegen/CompilerContext.cpp` also uses other `libevmasm` types like `evmasm::AssemblyItem`. Looking at `libyul/AssemblyStack.cpp` again, that was now renamed to `libyul/YulStack.cpp`. It is also having the dependency to `libevmasm`, because this dependency is already defined by the copied "translateOptimiserSettings" function - but this dependency is already there, because of the usage of e.g. `evmasm::LinkerObject` or `evmasm::Assembly`. So we have already a strong interdependence between these module. 

So the these cross dependencies where are already in place.

I think it would be good to replace the code replication in `libyul/YulStack.cpp`, so that `YulStack::assembleEVMWithDeployed` will now also use `Assembly::OptimiserSettings::translateSettings`.

aarlt: After talking to @cameel, we decided to just remove the translation function from this PR. I will create another PR that will only introduce that translation function.

aarlt: See https://github.com/ethereum/solidity/pull/13405

libsolidity/interface/CompilerStack.h
aarlt: Ok, I just removed that. Current version will use only `m_sourceJsons`.

libsolidity/interface/CompilerStack.cpp
aarlt: Looking at this again. I think it is ok how this is done right now. I don't see anything that could be moved into the `CompilerStack::assemble(...)` method.

libevmasm/Assembly.cpp
aarlt: I introduced in `libsolutil/JSON.h` the helper functions `ofType`, `ofTypeIfExists` and `getOrDefault`. These functions are now used here.

libevmasm/Assembly.cpp
```diff
@@ -98,14 +98,23 @@ void Assembly::addAssemblyItemsFromJSON(Json::Value const& _code)
...
+	solAssert(ofType<int>(_json, "end"));
+	solAssert(ofType<int>(_json, "source"));
+	solAssert(ofTypeIfExists<std::string>(_json, "value"));
+	solAssert(ofTypeIfExists<int>(_json, "modifierDepth"));
+	solAssert(ofTypeIfExists<std::string>(_json, "jumpType"));
+
+	std::string name = getOrDefault<std::string>(_json, "name", "");
+	solAssert(!name.empty());
+
+	int begin = getOrDefault<int>(_json, "begin", -1);
```
aarlt: Theoretically we only need a `get` function for the mandatory elements `name`, `begin`, `end` and `source` here.

aarlt: I added now a function `get`. I think its more consistent to have this function together with `getOrDefault`.

solc/CommandLineInterface.cpp
aarlt: True! That should definitely work!

aarlt: I just looked now superficially into this. But `std::ranges::all_of` does not work on single elements, it need to be a "collection". The usage of `ranges` would look like `solAssert(ranges::any_of(std::set<InputMode>{m_options.input.mode}, IsValidInputMode));` where `IsValidInputMode` is a predicate instead of a vector or set. Not sure whether this is nice. I could use `std::ranges::any_of` to implement `any_of`. 

However, I think I finally really understood your `compilationInputModes.count(m_options.input.mode)` suggestion now. I totally misread it. I agree, that's a good solution. So I guess I will go with your suggestion and remove the `any_of` stuff totally.

aarlt: Ok, I removed the `any_of` stuff - I used your suggestion now.

libevmasm/Assembly.cpp
aarlt: I would say no, I just added an additional assert for `location.sourceName`.

libevmasm/Assembly.cpp
aarlt: Using `c_instructions.count(name))` now

libsolidity/interface/OptimiserSettings.h
aarlt: The main problem here was that I basically was not able to find out whether the optimiser need to get run on the assembly or not. So I could either always optimize, or never. However, I simplified now the mechanism to set it in `CompilerStack::setOptimiserSettings(bool _optimize, size_t _runs)`, where I set the `enabled` flag to `_optimize`.

aarlt: Maybe also interesting in this context: https://github.com/ethereum/solidity/pull/12704#issuecomment-1067363681

scripts/ImportExportTest.sh
aarlt: I think I tried that once, but there was somehow a problem.. let me try again

aarlt: ah nice! it seem to work

scripts/ImportExportTest.sh
aarlt: True!

scripts/ImportExportTest.sh
aarlt: I thought that `"${OUTPUT[@]}"` will get expanded into one string because of `"`. So that the content of `"${OUTPUT[@]}"` is the second parameter.. need to check this

aarlt: I just tested it with
```bash
#!/bin/bash

function test()
{
  echo $1
  echo $2
}

array_name=(value1 value2 value3)

test "hello" "world"
test "hello" "${array_name[*]}"
```
Output:
```
hello
world
hello
value1 value2 value3
```
Should be ok then, right?

scripts/ImportExportTest.sh
aarlt: True!

scripts/ImportExportTest.sh
aarlt: Yeah you're right, but I would say we just leave it for now like this. We could refactor this later.

scripts/ImportExportTest.sh
aarlt: That's really good to prevent unneeded indentation. I put `[ $assembly != "" && $assembly != "null" ]] || continue` here - because we are in the loop that we still want to process.

solc/CommandLineInterface.cpp
```diff
@@ -497,8 +504,8 @@ void CommandLineInterface::readInputFiles()
 		string fileContent = readFileAsString(infile);
 		if (m_options.input.mode == InputMode::StandardJson)
 		{
-			solAssert(!m_standardJsonInput.has_value(), "");
-			m_standardJsonInput = std::move(fileContent);
+			solAssert(!m_standardJsonInput.has_value());
+			m_standardJsonInput = move(fileContent);
```
stackenbotten: Coding style error

scripts/ImportExportTest.sh
aarlt: I just went through the comments again. It looks like that I tested the change here - but I forgot to also actually change `testImportExportEquivalence "$solfile" "${OUTPUT[@]}"` to `testImportExportEquivalence "$solfile" "${OUTPUT[*]}"`

PR #12835, opened 2022-03-22
Fix MultiSigWallet.sol order of functions

dtedesco1:
Aligned order of functions with [style guide](https://docs.soliditylang.org/en/v0.8.12/style-guide.html#order-of-functions): 
- Constructor should appear before receive function
- Public functions should appear before internal functions

### DIFF
--- a/test/compilationTests/MultiSigWallet/MultiSigWallet.sol
+++ b/test/compilationTests/MultiSigWallet/MultiSigWallet.sol
@@ -88,18 +88,6 @@ contract MultiSigWallet {
         _;
     }
 
-    /// @dev Receive function allows to deposit ether.
-    receive()
-        external
-        payable
-    {
-        if (msg.value > 0)
-            emit Deposit(msg.sender, msg.value);
-    }
-
-    /*
-     * Public functions
-     */
     /// @dev Contract constructor sets initial owners and required number of confirmations.
     /// @param _owners List of initial owners.
     /// @param _required Number of required confirmations.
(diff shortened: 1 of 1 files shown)


dtedesco1:
@fulldecent would you mind reviewing this fix?

fulldecent:
@dtedesco1 This modifies a file in the test directory.

All tests are exempt from style guide updates.

Also please rebase, hopefully that will fix those pesky errors (I don't think you introduced them).

If you can fix errors then this is approved from me and then on the fence to core devs if they want this change or reject on account of not wanting to have any change on testing code.

chriseth:
I think this can be closed because it only touches test code, isn't it?

fulldecent:
I think we are good to update test code for typos and style fixed.

But if there is something in the test code that is specifically testing a typo or some thing that is against the style guide (such as some functions being defined in some order that against it) then that should be a comment and we will not change anything per those comments.

dtedesco1:
Discovered some checks were failing due to additional issues that I'm fixing:
- executeTransaction function shadowed builtin tx. Replaced with a different name.
- isConfirmed returns bool but did not specify false result conditions. Added this condition.

fulldecent:
Always be squashing

cameel:
Please rebase on `develop`. It already has a fix for the failing CI jobs (#12868).

dtedesco1:
Thanks @cameel.

Check passed and ready for approving review.

cameel (approved):

Issue #12830, opened 2022-03-22
Should `using` be applicable for interfaces and libraries?


hrkrshnn:
```solidity
// Bad interface
interface IERC20 {
    function safeTransferFrom(address from, address to, uint value) external;
}
library SafeTransfer {
    function safeTransferFrom(IERC20 token, address from, address to, uint value) internal {
        // do nothing
    }
}
// This type checks, but would produce an error when used
// Error: Member "safeTransferFrom" not unique after argument-dependent lookup in contract IERC20.
using SafeTransfer for IERC20;

contract C {
    // using SafeTransfer for IERC20;
    IERC20 constant token = IERC20(address(0));
    function stake(uint value) external {
        // uncomment to see the error
        // token.safeTransferFrom(msg.sender, address(this), value);
    }
}
```

chriseth:
We retained the same mechanism that was in place already for "using" inside a contract.

hrkrshnn:
The main issue now is that you can ship a file with a global using statement  and only realize at callsite that the statement cannot be used. I think we need to introduce more checks at the using statement definition.

Unfortunately, this would be a breaking change 😢

hrkrshnn:
The decision was to break `using X for I` where `I` is interface or contract type. Will write a blog post regarding this.

chriseth:
`using SafeTransfer for IERC20 global` turned out to be an unintended feature, i.e. a bug.

Whether or not we should disallow `using X for I` for interfaces and contracts should probably be discussed with more user feedback, but I personally tend to disallowing it.

Essentially: If there is a different way to define member functions for types, the other way should be used.

PR #12836, opened 2022-03-23
Fix using global with libraries.

chriseth:
Related to #12830.

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -3638,11 +3638,36 @@ void TypeChecker::endVisit(Literal const& _literal)
 
 void TypeChecker::endVisit(UsingForDirective const& _usingFor)
 {
+	if (_usingFor.global())
+	{
+		if (m_currentContract || !_usingFor.typeName())
+		{
+			solAssert(m_errorReporter.hasErrors());
+			return;
+		}
+		solAssert(_usingFor.typeName()->annotation().type);
+		if (Declaration const* typeDefinition = _usingFor.typeName()->annotation().type->typeDefinition())
+		{
+			if (typeDefinition->scope() != m_currentSourceUnit)
+				m_errorReporter.typeError(
+					4117_error,
+					_usingFor.location(),
+					"Can only use \"global\" with types defined in the same source unit at file level."
+				);
+		}
+		else
+			m_errorReporter.typeError(
+				8841_error,
+				_usingFor.location(),
+				"Can only use \"global\" with user-defined types."
+			);
+	}
+
 	if (!_usingFor.usesBraces())
 	{
 		solAssert(_usingFor.functionsOrLibrary().size() == 1);
 		ContractDefinition const* library = dynamic_cast<ContractDefinition const*>(
-				_usingFor.functionsOrLibrary().front()->annotation().referencedDeclaration
+			_usingFor.functionsOrLibrary().front()->annotation().referencedDeclaration
 		);
 		solAssert(library && library->isLibrary());
 		// No type checking for libraries
@@ -3661,28 +3686,6 @@ void TypeChecker::endVisit(UsingForDirective const& _usingFor)
 		_usingFor.typeName()->annotation().type
 	);
 	solAssert(normalizedType);
-
-	if (_usingFor.global())
-	{
-		if (m_currentContract)
-			solAssert(m_errorReporter.hasErrors());
-		if (Declaration const* typeDefinition = _usingFor.typeName()->annotation().type->typeDefinition())
-		{
-			if (typeDefinition->scope() != m_currentSourceUnit)
-				m_errorReporter.typeError(
-					4117_error,
-					_usingFor.location(),
-					"Can only use \"global\" with types defined in the same source unit at file level."
-				);
-		}
-		else
-			m_errorReporter.typeError(
-				8841_error,
-				_usingFor.location(),
-				"Can only use \"global\" with user-defined types."
-			);
-	}
-
 
 	for (ASTPointer<IdentifierPath> const& path: _usingFor.functionsOrLibrary())
 	{
--- a/test/libsolidity/syntaxTests/using/global_library_for_defined_elsewhere.sol
+++ b/test/libsolidity/syntaxTests/using/global_library_for_defined_elsewhere.sol
@@ -1,0 +1,6 @@
+using L for L.S global;
+library L {
+    struct S { uint x; }
+}
+// ----
+// TypeError 4117: (0-23): Can only use "global" with types defined in the same source unit at file level.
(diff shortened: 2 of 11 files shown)


test/libsolidity/syntaxTests/using/global_library_for_defined_elsewhere.sol
```diff
@@ -0,0 +1,6 @@
+using L for L.S global;
+library L {
+    struct S { uint x; }
+}
```
cameel, line 4: Can you also add a positive case for this? I.e. is this supposed to work?
```solidity
using L for S global;
struct S { uint x; }
library L {}
```

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
```diff
@@ -0,0 +1,6 @@
+interface I {}
```
cameel: Not sure I understand the naming convention for these tests. Shouldn't this one be `global_library_for_interface.sol`?

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
```diff
@@ -0,0 +1,6 @@
...
+using L for I global;
+library L {
+}
```
cameel: Again, can you add a positive test? This would be allowed without `global`, right?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3632,6 +3632,27 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+	{
+		if (m_currentContract)
+			solAssert(m_errorReporter.hasErrors());
+		if (Declaration const* typeDefinition = _usingFor.typeName()->annotation().type->typeDefinition())
+		{
+			if (typeDefinition->scope() != m_currentSourceUnit)
+				m_errorReporter.typeError(
+					4117_error,
+					_usingFor.location(),
+					"Can only use \"global\" with types defined in the same source unit at file level."
```
cameel, line 3655: Is something like this supposed to work?
`b.sol`
```solidity
import {S} from "./a.sol";

using L for S global;
library L {}
```
`a.sol`
```solidity
struct S { uint x; }
```

It currently compiles without errors on 0.8.13. If it is then this error message is not entirely correct.

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
cameel: I think we could use a bit more coverage with this. What types is this supposed to work for? I see that for example it worked for `uint` on 0.8.13. I think we should have at least this as a test case:
```solidity
using L for uint global;
library L {}
```

But we should really cover more types. For example do we have any tests at all covering `using for global` with enums? Or function pointers?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3632,6 +3632,27 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+				);
+		}
+		else
+			m_errorReporter.typeError(
+				8841_error,
+				_usingFor.location(),
+				"Can only use \"global\" with user-defined types."
+			);
+	}
+
```
cameel: Something must be still uninitialized at this point because [soltest is crashing due to a memory violation](https://app.circleci.com/pipelines/github/ethereum/solidity/23256/workflows/076c71e3-cd11-4dce-9d0a-63012cb529bf/jobs/1021108):
```
memory access violation at address: 0x00000000: no mapping at fault address
```
And it's this test that crashes it:
```solidity
using {f} for * global;
function f(uint) pure{}
```

There are also crashes in the fuzzer during CLI tests (e.g. [1021105](https://app.circleci.com/pipelines/github/ethereum/solidity/23256/workflows/076c71e3-cd11-4dce-9d0a-63012cb529bf/jobs/1021105)) but they could just be caused by the same memory violation.

ekpyron:
I feel quite bad for approving https://github.com/ethereum/solidity/pull/12288 without seeing any of this and without asking for tests for these...

libsolidity/analysis/TypeChecker.cpp
chriseth: It should not work - here is a test: https://github.com/ethereum/solidity/blob/develop/test/libsolidity/syntaxTests/using/global_for_type_from_other_file.sol

test/libsolidity/syntaxTests/using/global_library_for_defined_elsewhere.sol
chriseth: We already have a test: https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/using/using_global_library.sol

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
chriseth: Renamed.

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
chriseth: Added a test.

test/libsolidity/syntaxTests/using/using_global_library_for_interface.sol
chriseth: Added some more tests. Enum should work, function pointers not.

cameel:
Updated the description to link it to #12830. I guess we can't say it solves it because the issue is also about deciding whether to disallow it for interfaces completely?

libsolidity/analysis/TypeChecker.cpp
cameel: My example does compile on 0.8.13 though while `global_for_type_from_other_file.sol` does not. Apparently libraries are treated differently. Can you add my example and make sure it's reported as error?

cameel: Or just add a library to that existing test.

test/libsolidity/syntaxTests/using/global_library_for_defined_elsewhere.sol
cameel: ok. I mean in that example it's an UDVT not a struct but I guess they should be handled by the same code, right?

test/libsolidity/syntaxTests/using/global_with_asterisk.sol
```diff
@@ -0,0 +1,5 @@
+using {f} for * global;
```
cameel, line 1: Maybe it would be a good idea to also have `using L for * global;` covered?

cameel (dismissed):
I have two remaining suggestions about tests that should be added. The case with a struct in a separate file that was not detected as error on 0.8.13 and using a library with `*`.

I'm already approving though since these are just nice to have, not strictly required. I checked the first one and it does get fixed by the PR.

libsolidity/analysis/TypeChecker.cpp
chriseth: Added as "global_library_foreign.sol"

chriseth:
Merging since I added the tests exactly as @cameel requested.

PR #12837, opened 2022-03-23
Rephrase comment about arrays of arrays in code example.

aathan:
Clarify comment using language similar to that in the Array section of the documentation. Previously it said simply "Because of that..." but what the word "that" was about, was not evident.

### DIFF
--- a/docs/types/reference-types.rst
+++ b/docs/types/reference-types.rst
@@ -379,8 +379,10 @@ Array Members
         uint[2**20] aLotOfIntegers;
         // Note that the following is not a pair of dynamic arrays but a
         // dynamic array of pairs (i.e. of fixed size arrays of length two).
-        // Because of that, T[] is always a dynamic array of T, even if T
-        // itself is an array.
+        // In Solidity, T[k] and T[] are always arrays with elements of type T,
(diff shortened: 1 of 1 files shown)


docs/types/reference-types.rst
```diff
@@ -379,8 +379,10 @@ Array Members
         uint[2**20] aLotOfIntegers;
         // Note that the following is not a pair of dynamic arrays but a
         // dynamic array of pairs (i.e. of fixed size arrays of length two).
-        // Because of that, T[] is always a dynamic array of T, even if T
-        // itself is an array.
+        // In Solidity, T[n] is always an array containing n elements of type T
```
chriseth: I agree that "because of that" is wrong here, but maybe it's confusing to talk about both dynamic and statically-sized arrays here. What about `In Solidity, T[k] and T[] are always arrays with elements of type T, even if T is itself an array. Because of that, bool[2][] is a dynamic array...`?

And maybe we could also delete the first two lines?

Marenz:
No response from author so far, I am marking that as take-over for now. Feel free to update and I'll remove the label.

leonardoalt (approved):

PR #12839, opened 2022-03-23
lsp.py: Adds --fail-fast (-f) to terminate early on first failure.

christianparpart:
Breakout from PR #12468 

Looks like more will be broken out into its own tiny PRs.

### DIFF
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -129,6 +129,13 @@ class ExpectationFailed(Exception):
 
 def create_cli_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Solidity LSP Test suite")
+    parser.set_defaults(fail_fast=False)
+    parser.add_argument(
+        "-f, --fail-fast",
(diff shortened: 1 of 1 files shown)


Marenz (approved):

test/lsp.py
```diff
@@ -129,6 +129,13 @@ def __init__(self, actual, expected):
 
 def create_cli_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Solidity LSP Test suite")
+    parser.set_defaults(fail_fast=False)
+    parser.add_argument(
+        "-f, --fail-fast",
+        dest="fail_fast",
+        action="store_true",
+        help="Terminates the running tests on first failure."
+    )
```
cameel, line 138: Why not use `default=False` argument of `add_arguments()`?

cameel (commented):
Looks like I'm too late to review this but looks good anyway.

Still, I feel like we're reimplementing `unittest` here :)

PR #12843, opened 2022-03-24
value-types.rst: Clarify the purpose of `.wrap()`

aathan:
Just some clarification.

### DIFF
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -708,8 +708,10 @@ type with 18 decimals and a minimal library to do arithmetic operations on the t
 
 Notice how ``UFixed256x18.wrap`` and ``FixedMath.toUFixed256x18`` have the same signature but
 perform two very different operations: The ``UFixed256x18.wrap`` function returns a ``UFixed256x18``
-that has the same data representation as the input, whereas ``toUFixed256x18`` returns a
-``UFixed256x18`` that has the same numerical value.
+that has the same data representation and literal integer value as the input, whereas
+``toUFixed256x18`` returns a ``UFixed256x18`` that has a different literal integer value: The scaled
+one value of ``a`` that represents the same number of use with the arithmetic functions defined in
+the library. ``wrap`` is used within the library to store these scaled values without further scaling.
 
 .. index:: ! function type, ! type; function
 


docs/types/value-types.rst
```diff
@@ -708,8 +708,10 @@ type with 18 decimals and a minimal library to do arithmetic operations on the t
 
 Notice how ``UFixed256x18.wrap`` and ``FixedMath.toUFixed256x18`` have the same signature but
 perform two very different operations: The ``UFixed256x18.wrap`` function returns a ``UFixed256x18``
-that has the same data representation as the input, whereas ``toUFixed256x18`` returns a
-``UFixed256x18`` that has the same numerical value.
+that has the same data representation and literal integer value as the input, whereas
```
chriseth, line 711: I would actually like to avoid the term "literal integer value" - it is not clear if you are talking about the internal representation or about the integer value (i.e. truncated) of a fixed point number.

aathan: How about "of which the internal storage, which is a ``uint256`` has the same data representation and integer value as the input (and therefore who's semantic value is different from what it would be if interpreted as a ``UFixed256x18``, whereas"

... but hey, it's not like the original text was terrible. As I re-read some of these docs I'm just trying to pay attention to areas where a new reader may have to pause for a moment to really understand what is being said, and offer some additional details that reduce the degrees of freedom in the interpretation.

Marenz: > How about "of which the internal storage, which is a uint256 has the same data representation and integer value as the input (and therefore who's semantic value is different from what it would be if interpreted as a UFixed256x18, whereas"

I feel both descriptions here are overly complicated. I am also not sure if that whole paragraph is needed at all. I may be biased, but to me it tries to explain something that was never confusing.
But its mere existence implies that there is more going on than is obvious in the source code, thus adding confusion. 
The user then has to read it several times to make sure they didn't miss anything besides the obvious...

At least that was my experience after reading up the whole page again.

But if we decide to keep it here is my suggestion:

`that simply wraps the input data without modifying it whereas "toUFixed256x18" tries to interpret and modify the value according to the logic of the class`

Marenz (changes requested):

docs/types/value-types.rst
cameel: I think that some confusion is introduced by saying at the beginning of the section that `.wrap()` and `.unwrap()` **convert to/from** the custom type. I think it would be clearer to say that they **set/get the internal representation** of the value. Then we could later unambiguously call `toUFixed256x18()` a conversion function and explain things like this:

```rst
``FixedMath`` provides ``toUFixed256x18`` function, which can be used to convert an ``uint`` value into ``UFixed256x18``.
Note that this is different from wrapping that ``uint`` value.
``UFixed256x18.wrap()`` also turns an integer into an instance of ``UFixed256x18`` but that instance does not
represent the same value and is not a 1:1 conversion.
```

Or maybe adding an extra overload of ``toUFixed256x18()`` with a different type would show that better.  `wrap()` works only with the underlying type while there could be conversions from many other types. For example we could add something like this (not tested)

```solidity
function toUFixed256x18(UFixed128x9 a) internal pure returns (UFixed256x18) {
    return UFixed256x18.wrap(a.unwrap() * (multiplier / UFixed128x9.multiplier));
}
```

But overall, I agree with @Marenz that maybe we're just overexplaining it here. I also don't feel like this paragraph is needed all that much.

ekpyron:
Closing this, since - while there may be call for changing this section in general - the consensus seems to be against the particular changes in the PR, so we can reopen a fresh one, if we want to proceed with changing at all.

PR #12844, opened 2022-03-24
Update operators.rst

aathan:
Based on my understanding I think this description is more accurate than "arbitrary precision" in that the language is not picking some potentially insufficient "arbitrary" precision, but rather, does not effectively limit the precision of the literal expression.

### DIFF
--- a/docs/types/operators.rst
+++ b/docs/types/operators.rst
@@ -18,7 +18,9 @@ and the type of the operator's result:
 In case one of the operands is a :ref:`literal number <rational_literals>` it is first converted to its
 "mobile type", which is the smallest type that can hold the value
 (unsigned types of the same bit-width are considered "smaller" than the signed types).
-If both are literal numbers, the operation is computed with arbitrary precision.
+If both are literal numbers, the operation is computed with effectively unlimited precision in
+that the expression is evaluated to whatever precision is necessary so that none is lost
+when the result is used with a non-literal type.
 
 The operator's result type is the same as the type the operation is performed in,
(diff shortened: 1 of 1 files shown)


docs/types/operators.rst
```diff
@@ -18,7 +18,9 @@ and the type of the operator's result:
 In case one of the operands is a :ref:`literal number <rational_literals>` it is first converted to its
 "mobile type", which is the smallest type that can hold the value
 (unsigned types of the same bit-width are considered "smaller" than the signed types).
-If both are literal numbers, the operation is computed with arbitrary precision.
+If both are literal numbers, the operation is computed with effectively unlimited precision in
+that the expression is evaluated to whatever precision is necessary so that no precision is lost
```
chriseth: Yes, this sounds better! Is there a way to avoid the third repetition of the word "precision"?

What I don't understand is why you say "when the result is used with a non-literal type". Is that really necessary (and correct)?

aathan: "so that no precision is lost" --> "so that none is lost" fixes that.

I believe yes, it is necessary, because my reading of the docs (which I have not yet verified by reading a Solidity compiler implementation) is that these literal expressions are maintained in a computable but not computed form until the conversion to a "physical" type happens. I.e., that the "effectively unlimited precision" is a result not of actually storing the "unlimited precision" value, but rather, its fractional rational representation until that value can be computer *up to the limit of the destination type*.

leonardoalt (approved):

PR #12845, opened 2022-03-24
Rephrase documentation about literals and rational types.

aathan:
(1) The word "arbitrary" is not strictly correct as I understand it, although I would concede that in this context, many people would understand it to mean "any." I believe, however, that strictly speaking, Solidity treats these expressions as if they had effectively unlimited precision, or equivalently, at minimum the precision of the target conversion.

A highly technical reader might be left wondering if Solidity literally is selecting an arbitrary precision vs the necessary precision such that the precision seems unlimited.

(Dictionary definition of "arbitrary": Determined by chance, whim, or impulse, and not by necessity, reason, or principle)

(2) The note about number literal types seems to be unnecessarily typing implementation details of the compiler relative to number literal types, but I think the important point the documentation is trying to convey simply has to do with the equivalence of literal expressions to the type of the result -- I am therefore proposing an alternative wording.

(3) I believe my proposed wording is a bit less confusing for the final note. I am not sure how to use :ref: correctly, so please fix that if incorrect. I also noticed a whhile typo in my pull request, which should be fixed if accepted.

### DIFF
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -111,7 +111,7 @@ the operands, division on integers always results in an integer.
 In Solidity, division rounds towards zero. This means that ``int256(-5) / int256(2) == int256(-2)``.
 
 Note that in contrast, division on :ref:`literals<rational_literals>` results in fractional values
-of arbitrary precision.
+of arbitrary precision. Here, the word "arbitrary" means "any" or "unlimited."
 
 .. note::
   Division by zero causes a :ref:`Panic error<assert-and-require>`. This check can **not** be disabled through ``unchecked { ... }``.
@@ -465,7 +465,9 @@ the underscores are ignored.
 Number literal expressions retain arbitrary precision until they are converted to a non-literal type (i.e. by
 using them together with anything other than a number literal expression (like boolean literals) or by explicit conversion).
 This means that computations do not overflow and divisions do not truncate
-in number literal expressions.
+in number literal expressions. One way to achieve this is to represent fractional rational numbers as
+quotients of integers, converting to limited precision only when necessary (such as when assigning to a
+value type variable).
 
 For example, ``(2**800 + 1) - 2**800`` results in the constant ``1`` (of type ``uint8``)
 although intermediate results would not even fit the machine word size. Furthermore, ``.5 * 8`` results
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
     Division on integer literals used to truncate in Solidity prior to version 0.4.0, but it now converts into a rational number, i.e. ``5 / 2`` is not equal to ``2``, but to ``2.5``.
 
 .. note::
-    Solidity has a number literal type for each rational number.
-    Integer literals and rational number literals belong to number literal types.
+    In solidity, rational numbers and integers expressed as literals are of number literal type.
     Moreover, all number literal expressions (i.e. the expressions that
-    contain only number literals and operators) belong to number literal
-    types.  So the number literal expressions ``1 + 2`` and ``2 + 1`` both
-    belong to the same number literal type for the rational number three.
+    contain only number literals and operators) are of number literal
+    type.  The number literal expressions ``1 + 2`` and ``2 + 1`` are both
+    of number literal type, just the same as ``3``. Technically speaking, the internal representation
+    for fractional rational numbers like ``1.5`` is different than that of ``3``, specifically to support
+    arbitrary precision. However, all number literals are treated as effectively the same type.
 
 
 .. note::
     Number literal expressions are converted into a non-literal type as soon as they are used with non-literal
-    expressions. Disregarding types, the value of the expression assigned to ``b``
-    below evaluates to an integer. Because ``a`` is of type ``uint128``, the
-    expression ``2.5 + a`` has to have a proper type, though. Since there is no common type
-    for the type of ``2.5`` and ``uint128``, the Solidity compiler does not accept
-    this code.
+    expressions. If not for type conversion considerations, when ``a=1`` the value of the expression ``2.5 + a + 0.5``
+    would be the integer ``4``. However, ``a`` is of type ``uint128``. Therefore, an implicit conversion to a non-literal
+    type must occur when evaluating the first part of the expression ``2.5 + a``. Since there is no such conversion
+    between the types of ``2.5`` (a fractional rational literal number type) and ``uint128`` (an unsigned integer
+    value type), the Solidity compiler does not accept this code.  See :ref:Operators<operator>
 
 .. code-block:: solidity
 


docs/types/value-types.rst
```diff
@@ -462,10 +462,11 @@ Underscores are only allowed between two digits and only one consecutive undersc
 There is no additional semantic meaning added to a number literal containing underscores,
 the underscores are ignored.
 
-Number literal expressions retain arbitrary precision until they are converted to a non-literal type (i.e. by
+Number literal expressions retain unlimited precision until they are converted to a non-literal type (i.e. by
 using them together with anything other than a number literal expression (like boolean literals) or by explicit conversion).
 This means that computations do not overflow and divisions do not truncate
-in number literal expressions.
+in number literal expressions. This is achieved by representing fractional rational numbers as quotients of integers, and
```
chriseth: Is it important to document how we do it internally? I'm asking because we need to find this sentence and modify it whenever we do a change.

docs/types/value-types.rst
```diff
@@ -494,21 +495,21 @@ regardless of the type of the right (exponent) operand.
...
 .. note::
-    Solidity has a number literal type for each rational number.
-    Integer literals and rational number literals belong to number literal types.
+    In solidity, rational numbers and integers expressed as literals are of number literal type.
     Moreover, all number literal expressions (i.e. the expressions that
-    contain only number literals and operators) belong to number literal
-    types.  So the number literal expressions ``1 + 2`` and ``2 + 1`` both
-    belong to the same number literal type for the rational number three.
+    contain only number literals and operators) are of number literal
+    type.  The number literal expressions ``1 + 2`` and ``2 + 1`` are both
```
chriseth, line 502: I think it's important her eto state that it is exactly the same type.

docs/types/value-types.rst
```diff
@@ -494,21 +495,21 @@ regardless of the type of the right (exponent) operand.
...
+    of number literal type, just the same as ``3``. There are number literal subtypes, such
```
chriseth: Why is it a "subtype" and what do you mean by the term?

docs/types/value-types.rst
```diff
@@ -494,21 +495,21 @@ regardless of the type of the right (exponent) operand.
...
 
 .. note::
     Number literal expressions are converted into a non-literal type as soon as they are used with non-literal
-    expressions. Disregarding types, the value of the expression assigned to ``b``
-    below evaluates to an integer. Because ``a`` is of type ``uint128``, the
-    expression ``2.5 + a`` has to have a proper type, though. Since there is no common type
-    for the type of ``2.5`` and ``uint128``, the Solidity compiler does not accept
-    this code.
+    expressions. Disregarding types, the value of the expression ``2.5 + a + 0.5`` where ``a=1`` would be the
+    integer ``4``. However, ``a`` is of type ``uint128``. Therefore, an implicit conversion to a non-literal
```
chriseth: What I meant with "Disregarding types" was that the value that is assigned to `b` is always an integer, regardless of the type of `a` or of any other sub-expression in the assigned value. What we can maybe say is that the expression `2.5 + 1 + 0.5` evaluates to `4` (if you think that this is not clear to the reader).

docs/types/value-types.rst
aathan: Fair enough, however that is the main point of this whole topic. I.e., that fractional rational numbers are represented as fractional rational numbers (which specifically means as quotients of integers. In other words, if that implementation changed, this whole section and every mention of "unlimited precision" (or "arbitrary precision") in the docs would probably have to be revisited.

docs/types/value-types.rst
aathan: Ok. As a new reader the words "belong to the same number literal type" in the context of the rest of the original wording did not actually convey "that it is exactly the same type". Instead, the picture that was formed in my head, reading the original doc, is that "there is a set of number literal types which are fungible/compatible with each other, and that 1+2 and 2+1 could be of one of those types, and 3 of another of those types, but all of them are in the same family.

So, if it is true that they are literally "the one and the same type" the wording can be even more improved from my suggestion. But, in my experiments, I found that some implementations (foundry) seems to treat a fractional rational literal as one type, and a non-fractional rational literal as another type, but that those types are "fully compatible" with each other.

That's why I chose this wording.

ekpyron:
> (1) The word "arbitrary" is not strictly correct as I understand it, although I would concede that in this context, many people would understand it to mean "any." I believe, however, that strictly speaking, Solidity treats these expressions as if they had effectively unlimited precision, or equivalently, at minimum the precision of the target conversion.
> 
> A highly technical reader might be left wondering if Solidity literally is selecting an arbitrary precision vs the necessary precision such that the precision seems unlimited.
> 
> (Dictionary definition of "arbitrary": Determined by chance, whim, or impulse, and not by necessity, reason, or principle)
> 

Note that [Arbitrary Precision](https://en.wikipedia.org/wiki/Arbitrary-precision_arithmetic) is an idiomatic, technical term in this context that means precisely what we are talking about here.

aathan:
Fair enough :)

docs/types/value-types.rst
```diff
@@ -111,7 +111,7 @@ the operands, division on integers always results in an integer.
 In Solidity, division rounds towards zero. This means that ``int256(-5) / int256(2) == int256(-2)``.
 
 Note that in contrast, division on :ref:`literals<rational_literals>` results in fractional values
-of arbitrary precision.
+of arbitrary precision. Here, the word "arbitrary" means "any" or "unlimited."
```
cameel, line 114: As pointed out by @ekpyron this is really idiomatic and explaining the word here feels just weird to me. How about we just link to Wikpiedia for those who are not familiar with the term? I think this would be less obtrusive for others, who are:

```suggestion
of `arbitrary precision <https://en.wikipedia.org/wiki/Arbitrary-precision_arithmetic>`_.
```
```
```
v

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
     Division on integer literals used to truncate in Solidity prior to version 0.4.0, but it now converts into a rational number, i.e. ``5 / 2`` is not equal to ``2``, but to ``2.5``.
 
 .. note::
-    Solidity has a number literal type for each rational number.
-    Integer literals and rational number literals belong to number literal types.
+    In solidity, rational numbers and integers expressed as literals are of number literal type.
```
cameel, line 499: Your version as a different meaning. The original sentence says that e.g. `1` and `2` belong to different types (called `int_const 1` and `int_const 2`; you can't use these types for variables but you can sometimes see them in error messages), while `1` and `1.0` are of the same type (both are `int_const 1`).

The important takeaway is that there are no separate types for integers and fractional numbers used in literals. In the AST `1.0` is indistinguishable from `1`. That distinction only appears when you convert the literal expression to a concrete integer or fixed-point type (note also that fixed-point types are currently not fully implemented).

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
...
     Moreover, all number literal expressions (i.e. the expressions that
-    contain only number literals and operators) belong to number literal
-    types.  So the number literal expressions ``1 + 2`` and ``2 + 1`` both
-    belong to the same number literal type for the rational number three.
+    contain only number literals and operators) are of number literal
+    type.  The number literal expressions ``1 + 2`` and ``2 + 1`` are both
+    of number literal type, just the same as ``3``. Technically speaking, the internal representation
+    for fractional rational numbers like ``1.5`` is different than that of ``3``, specifically to support
+    arbitrary precision. However, all number literals are treated as effectively the same type.
```
cameel, line 505: It's not different. They're both internally represented using the same multi-precision type. If you took the same number as a fractional vs integer (e,g `3.0` vs `3`) there would be absolutely no difference internally. You can consider them not just convertible but exactly equivalent for all intents and purposes is Solidity.

This is really getting into compiler's implementation details that are not relevant to understanding semantics of literals. I'd remove this part.

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
...
-    expressions. Disregarding types, the value of the expression assigned to ``b``
-    below evaluates to an integer. Because ``a`` is of type ``uint128``, the
-    expression ``2.5 + a`` has to have a proper type, though. Since there is no common type
-    for the type of ``2.5`` and ``uint128``, the Solidity compiler does not accept
-    this code.
+    expressions. If not for type conversion considerations, when ``a=1`` the value of the expression ``2.5 + a + 0.5``
+    would be the integer ``4``. However, ``a`` is of type ``uint128``. Therefore, an implicit conversion to a non-literal
+    type must occur when evaluating the first part of the expression ``2.5 + a``. Since there is no such conversion
+    between the types of ``2.5`` (a fractional rational literal number type) and ``uint128`` (an unsigned integer
+    value type), the Solidity compiler does not accept this code.  See :ref:Operators<operator>
```
cameel, line 514:
```suggestion
    value type), the Solidity compiler does not accept this code.  See :ref:`Operators<operator>`.
```

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
...
 
 .. note::
     Number literal expressions are converted into a non-literal type as soon as they are used with non-literal
-    expressions. Disregarding types, the value of the expression assigned to ``b``
-    below evaluates to an integer. Because ``a`` is of type ``uint128``, the
-    expression ``2.5 + a`` has to have a proper type, though. Since there is no common type
-    for the type of ``2.5`` and ``uint128``, the Solidity compiler does not accept
-    this code.
+    expressions. If not for type conversion considerations, when ``a=1`` the value of the expression ``2.5 + a + 0.5``
+    would be the integer ``4``. However, ``a`` is of type ``uint128``. Therefore, an implicit conversion to a non-literal
```
cameel, line 511: Technically, it could be a literal type if a common literal type existed.
```suggestion
    would be the integer ``4``. However, ``a`` is of type ``uint128``. Therefore, an implicit conversion to a common
```

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
...
+    type must occur when evaluating the first part of the expression ``2.5 + a``. Since there is no such conversion
```
cameel, line 512:
```suggestion
    type must occur when evaluating the first part of the expression (``2.5 + a``). Since there is no such conversion
```

docs/types/value-types.rst
```diff
@@ -494,21 +496,22 @@ regardless of the type of the right (exponent) operand.
...
+    between the types of ``2.5`` (a fractional rational literal number type) and ``uint128`` (an unsigned integer
```
cameel, line 513: There are no fractional rational number types distinct from integer rational number types. Integer and fractional literals belong to the same rational type if their values are equal.
```suggestion
    between the types of ``2.5`` (a rational literal number type) and ``uint128`` (an unsigned integer
```

docs/types/value-types.rst
cameel, line 514: It sounds as if you're talking about the type of `uint128` (i.e. type of a type), while you should be talking about the type of `a`.
```suggestion
    between the type of ``2.5`` (a fractional rational literal number type) and ``uint128``
    the Solidity compiler does not accept this code.  See :ref:Operators<operator>
```

cameel (changes requested):

docs/types/value-types.rst
cameel: I also think this is unnecessary. It's a compiler implementation detail not necessary to understand the semantics of literals in Solidity.

docs/types/value-types.rst
cameel: I think you misunderstood what this paragraph is talking about see: https://github.com/ethereum/solidity/pull/12845#discussion_r845359000

Which, on the other hand, does indicate that we're not clear enough and it should be reworded. Just not the way it's currently in the PR :)

cameel:
Squashed and rebased on `develop`.

leonardoalt:
This PR is waiting for answers for a long time, and has a lot of comments left. I'm closing it for now, please reopen if needed.

PR #12846, opened 2022-03-24
Update operators.rst

aathan:
I am pretty sure the original text referencing a conversion to integer is wrong, but I am only mostly sure my proposed edits are strictly correct.

### DIFF
--- a/docs/types/operators.rst
+++ b/docs/types/operators.rst
@@ -20,6 +20,9 @@ In case one of the operands is a :ref:`literal number <rational_literals>` it is
 (unsigned types of the same bit-width are considered "smaller" than the signed types).
 If both are literal numbers, the operation is computed with arbitrary precision.
 
+Certain imipicit conversions between number types are disallowed, such as implicit
+conversions between fractional rational literals and fixed point value types.
+
 The operator's result type is the same as the type the operation is performed in,
 except for comparison operators where the result is always ``bool``.
 
@@ -41,7 +44,8 @@ and 256 exceeds the range allowed for this type.
 
 Another consequence is that an expression like ``1.5 + 1.5`` is valid but ``1.5 + (true ? 1.5 : 2.5)`` is not.
 This is because the former is a rational expression evaluated in unlimited precision and only its final value matters.
-The latter involves a conversion of a fractional rational number to an integer, which is currently disallowed.
+The latter involves a conversion of a fractional rational number to an unsigned fixed point type (``ufixed8x1``, the
+minimal type that fits ``1.5``). This is a disallowed implicit conversion as mentioned above.
 
 .. index:: assignment, lvalue, ! compound operators
 


docs/types/operators.rst
```diff
@@ -20,6 +20,9 @@ In case one of the operands is a :ref:`literal number <rational_literals>` it is
 (unsigned types of the same bit-width are considered "smaller" than the signed types).
 If both are literal numbers, the operation is computed with arbitrary precision.
 
+Certain imipicit conversions between number types are disallowed, such as implicit
```
chriseth, line 23: Does this provide any value if you don't go into detail?

docs/types/operators.rst
```diff
@@ -41,7 +44,8 @@ and 256 exceeds the range allowed for this type.
 
 Another consequence is that an expression like ``1.5 + 1.5`` is valid but ``1.5 + (true ? 1.5 : 2.5)`` is not.
 This is because the former is a rational expression evaluated in unlimited precision and only its final value matters.
-The latter involves a conversion of a fractional rational number to an integer, which is currently disallowed.
+The latter involves a conversion of a fractional rational number to an unsigned fixed point type (``ufixed8x1``, the
```
chriseth, line 47: The main problem here is that the implementation of fixed point types in Solidity is not finished. We decided to take them out of the language again, so maybe it is better to not talk about this in detail and just remove fixed point types from the language altogether.

chriseth:
Thank you for all the good suggestions! I think your PRs have value, but maybe you could slim them down a bit, so we can merge those parts that do not require any discussion first?

aathan:
I'm a noob at making doc suggestion PRs. Happy to conform to whatever process is best for your workflow. Just let me know. Can you clarify for example what you mean by "slim them down"?  Do you mean you'd like me to submit each individual paragraph being changed as an individual PR? 

Or do you just literally mean "make your changes shorter?"  <--- that is exactly counter to adding words that yes, lengthen the docs, but hopefully at the same time reduce ambiguity. It has been my experience that both some amount of repetition and some amount of stating things 2-3 different ways, helps humans absorb new information.

chriseth:
I meant that some of your changes are fine as they are and some others are being discussed. Since discussion on wording is usually lengthy on pull requests, you could extract the parts that do not need any discussion and we can merge them right away.

aathan:
Understood. By the way, is there a forum where we could chat about proposed changes *before* I even write one? For example, I'm currently reading some of the docs on the `address` type members where I could add something to the docs, but I'm not sure it's the right place etc and would like to ask first...

chriseth:
Sure! Just come to our chat room as explained in https://docs.soliditylang.org/en/latest/contributing.html#workflow-for-pull-requests

leonardoalt:
The reviews seem to disagree that the changes here are wanted. Closing for now.

PR #12847, opened 2022-03-24
Yul interpreter: Fix overflows in memory pointer read/writes.

bshastry:
After the redundant store eliminator optimization was merged (https://github.com/ethereum/solidity/pull/12672), the fuzzer started reporting false positives due to overflows in memory pointer arithmetic in the Yul interpreter.

Previously, these did not matter since the overflow affected unoptimized and optimized programs equivalently but after the merge of RSE, this started to matter because an overflowing memory read/write may only occur in the unoptimized version in situations where the overflowing memory read/write is redundant and hence removed.

Example

```
mstore(not(0), 1) // wraps around and writes a one to memory byte addressed by 31
sstore(0, mload(0)) // reads from the overflown write
```

In the example above, the RSE would remove the first `mstore` but the unoptimized program contains it. Since the unoptimized program actually writes to the memory word starting at address zero, the interpretation of the `sstore` would contain the said value. In the post-RSE-optimized program, the `mstore` would be missing and hence the `sstore` would not contain the said value.


bshastry:
BTW, this PR does not fix the following false positive

```
contract C {
    function f() external
    {
        assembly {
            let x := not(0)
            mstore(x, 7)
            sstore(0, mload(x))
        }
    }
}
```

Here, the unoptimized IR generator would generate code that would revert (due to high memory write). The optimized IR generator would simplify the assembly block to `sstore(0, 7)`. Technically, this is not a false positive because this is how the real code generators also behave, so I'm going to ignore them for now.

bshastry:
I feel the elementfi external test failure is unrelated although it looks like the failing tests make use of the EVM interpreter which may not be the interpreter this PR touches in any case. Any thoughts @cameel since you may be more familiar with that external test?

cameel:
Looks completely unrelated:
```
     ProviderError: Unable to complete request at this time.
      at HttpProvider.request (node_modules/hardhat/src/internal/core/providers/http.ts:49:19)
      at JsonRpcClient._send (node_modules/hardhat/src/internal/hardhat-network/jsonrpc/client.ts:319:39)
      at JsonRpcClient._perform (node_modules/hardhat/src/internal/hardhat-network/jsonrpc/client.ts:249:34)
      at async ForkStateManager.getContractStorage (node_modules/hardhat/src/internal/hardhat-network/provider/fork/ForkStateManager.ts:206:25)
```
This is a network problem - ElementFi tests use alchemy.io to get mainnet state. Apparently there was some problem with the request (maybe alchemy.io is down?).

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -270,15 +270,17 @@ u256 EVMInstructionInterpreter::eval(
...
-		accessMemory(arg[0], 0x20);
-		writeMemoryWord(arg[0], arg[1]);
+		if (accessMemory(arg[0], 0x20))
+			writeMemoryWord(arg[0], arg[1]);
 		return 0;
 	case Instruction::MSTORE8:
-		accessMemory(arg[0], 1);
-		m_state.memory[arg[0]] = uint8_t(arg[1] & 0xff);
+		if (accessMemory(arg[0], 1))
+			m_state.memory[arg[0]] = uint8_t(arg[1] & 0xff);
```
cameel, line 287: There are still quite a few unguarded `accessMemory()` calls in this file. I guess we're assuming that we're nowhere close to the edge there? Should we replace them with something like `soltestAssert(accessMemory())` to make sure we're not missing any other cases?

cameel:
Do the memory writes wrap on the actual EVM? In the Yellow Paper I see this:
> MSTORE
> Save word to memory.
> μ0m[μs[0] . . . (μs[0] + 31)] ≡ μs[1]
> μ0i ≡ max(μi , d(μs[0] + 32) ÷ 32e)
> The addition in the calculation of μ0i is not subject to the 2^256 modulo.

It specifically says that for `μ0i`, the `+` does not wrap. So this means that for `μ0m` it does, right?

> In the example above, the RSE would remove the first `mstore` but the unoptimized program contains it.

Isn't that a bug in the optimizer? If memory writes wrap around then the optimizer removing them changes semantics.

> Here, the unoptimized IR generator would generate code that would revert (due to high memory write).

Why would it revert? Does codegen insert a check against that? If so, does that check get optimized out?

> The optimized IR generator would simplify the assembly block to `sstore(0, 7)`.

It would actually store `7 * 256`, right? The first byte of the range is still addressable so the `7` would go into the second byte of the slot if I understand it correctly.

bshastry:
> Why would it revert? Does codegen insert a check against that? If so, does that check get optimized out?

It would revert because the IR code generator assumes an upper bound on the value of the free memory pointer, accesses beyond this value revert i.e., the code generator generates something like

```
if free_mem_ptr > 0xfffffffffff
  revert
```

No, the check does not get optimized out.

> It would actually store 7 * 256, right? The first byte of the range is still addressable so the 7 would go into the second byte of the slot if I understand it correctly.

Right, please ignore the specifics. I just wanted to point out that the sstore stores a non zero value.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
bshastry: > I guess we're assuming that we're nowhere close to the edge there?

Each of those unguarded `accessMemory()` calls is in conjunction with a `logTrace()` call that writes to the interpreter trace. For this reason, we can be sure that trace of such calls is not silently ignored.

A `soltestAssert(accessMemory())` would not be useful since it's failure (not accessing bounded memory) does not indicate a bug, rather could lead to lots of false positives.

bshastry:
> Isn't that a bug in the optimizer? If memory writes wrap around then the optimizer removing them changes semantics.

I think there may be a misunderstanding here. The optimiser does not remove wrap around memory store, it only removes redundant stores.

Please note that the following wrap around memory write is correctly accounted for (and optimised) by the Yul optimiser

```
{
  mstore(add(not(0), 1), 7)
  sstore(0, mload(0))
}
```

is optimised to

```
{
    sstore(0, 7)
}
```

test/libyul/yulInterpreterTests/access_large_memory_offsets.yul
```diff
@@ -7,8 +7,6 @@
 // ----
 // Trace:
 // Memory dump:
-//      0: ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff07
-//   FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE0: 00000000000000000000000000000000000000000000000000000000000000ff
+//      0: 0000000000000000000000000000000000000000000000000000000000000007
 // Storage dump:
 //   0000000000000000000000000000000000000000000000000000000000000000: 0000000000000000000000000000000000000000000000000000000000000007
-//   0000000000000000000000000000000000000000000000000000000000000001: ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```
ekpyron, line 14: If this change removes this write (resp. makes it a ``sstore(1,0)``), won't this cause *more* false positives?
For the optimizer it's just fine to resolve the ``mload(sub(0,1))`` to ``sub(0,1)`` - which would result in ``sub(0,1)`` be written to storage slot ``1``...

bshastry: You are right, this may cause more false positives because loads from previously written to (large) memory addresses won't be interpreted.

I don't have a good solution to the problem this PR seeks to address.

ekpyron:
*A* solution to the underlying problem here may be to just drop the bounds on memory accesses in the interpreter... which bad implications might that have? It'd need to be a lot of accesses of ``0xffff`` bytes to actually exhaust memory, resp. we could still limit the overall size of accessed memory instead.

ekpyron:
What do we do with this PR? What about the solution I proposed here https://github.com/ethereum/solidity/pull/12847#issuecomment-1298409296?

ekpyron:
Closing for now, as discussed with @bshastry

PR #12850, opened 2022-03-24
Properly check data location in inheritance.

### LINKED ISSUE #10900: Mismatching data locations are allowed in overrides
It is possible to override functions that expect `memory` arguments with implementations that expect them in `calldata`. Offending code compiles with no warnings, but causes runtime errors.

I've observed this behavior in both 0.7 and 0.8 versions, but haven't tested how far back it goes.

Sample contract:
```solidity

pragma solidity ^0.8.0;

pragma experimental ABIEncoderV2;

abstract contract Base {
    struct Foo {
        uint256 foo;
    }
    
    function bar(Foo memory f) public pure returns (uint256) {
        return baz(f);
    }
    
    function baz(Foo memory f) internal virtual pure returns (uint256);
}

contract FaultyDerived is Base {
    function baz(Foo calldata f) internal override pure returns (uint256) {
        return f.foo;       
    }
}

contract CorrectDerived is Base {
    function baz(Foo memory f) internal override pure returns (uint256) {
        return f.foo;       
    }
}
```

I haven't tested too much the runtime issues, but I have seen some inconsistent behavior. When trying the example above on Remix, CorrectDerived works correctly as expected, but FaultyDerived returns 0 always. This is also true if `foo` is an address instead of a uint. 

However, on my local Hardhat environment in the project where I found this, I get reverts when attempting to access fields in the offending struct. I imagine this is related to either the fact that my structs are more complex, or that my project has already performed a few memory allocations at the moment the issue is triggered.

chriseth:
~Depends on #13027~. Merged.
Fixes https://github.com/ethereum/solidity/issues/10900

Note that the bug is also relevant for the return types.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 Important Bugfixes:
  * ABI Encoder: When ABI-encoding values from calldata that contain nested arrays, correctly validate the nested array length against ``calldatasize()`` in all cases.
+ * Override Checker: Allow changing data location for parameters only when overriding external functions.
 
 
 Compiler Features:
--- a/docs/bugs.json
+++ b/docs/bugs.json
@@ -1,5 +1,16 @@
 [
     {
+        "uid": "SOL-2022-3",
+        "name": "DataLocationChangeInInternalOverride",
+        "summary": "It was possible to change the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa while overriding internal and public functions. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
+        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public and internal functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function interprets it as a calldata pointer or vice-versa.",
+        "link": "https://blog.soliditylang.org/2022/05/17/data-location-inheritance-bug/",
+        "introduced": "0.6.9",
+        "fixed": "0.8.14",
+        "severity": "very low"
+
+    },
+    {
         "uid": "SOL-2022-2",
         "name": "NestedCallataArrayAbiReencodingSizeValidation",
         "summary": "ABI-reencoding of nested dynamic calldata arrays did not always perform proper size checks against the size of calldata and could read beyond ``calldatasize()``.",
--- a/docs/bugs_by_version.json
+++ b/docs/bugs_by_version.json
@@ -1345,6 +1345,7 @@
     },
     "0.6.10": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1356,6 +1357,7 @@
     },
     "0.6.11": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1367,6 +1369,7 @@
     },
     "0.6.12": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1477,6 +1480,7 @@
     },
     "0.6.9": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1489,6 +1493,7 @@
     },
     "0.7.0": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1500,6 +1505,7 @@
     },
     "0.7.1": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1512,6 +1518,7 @@
     },
     "0.7.2": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1523,6 +1530,7 @@
     },
     "0.7.3": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1533,6 +1541,7 @@
     },
     "0.7.4": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1542,6 +1551,7 @@
     },
     "0.7.5": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1551,6 +1561,7 @@
     },
     "0.7.6": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1560,6 +1571,7 @@
     },
     "0.8.0": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1569,6 +1581,7 @@
     },
     "0.8.1": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1578,12 +1591,14 @@
     },
     "0.8.10": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation"
         ],
         "released": "2021-11-09"
     },
     "0.8.11": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "AbiEncodeCallLiteralAsFixedBytesBug"
         ],
@@ -1591,6 +1606,7 @@
     },
     "0.8.12": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "AbiEncodeCallLiteralAsFixedBytesBug"
         ],
@@ -1598,12 +1614,14 @@
     },
     "0.8.13": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation"
         ],
         "released": "2022-03-16"
     },
     "0.8.2": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory",
@@ -1613,6 +1631,7 @@
     },
     "0.8.3": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables",
             "ABIDecodeTwoDimensionalArrayMemory"
@@ -1621,6 +1640,7 @@
     },
     "0.8.4": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables"
         ],
@@ -1628,6 +1648,7 @@
     },
     "0.8.5": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables"
         ],
@@ -1635,6 +1656,7 @@
     },
     "0.8.6": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables"
         ],
@@ -1642,6 +1664,7 @@
     },
     "0.8.7": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "SignedImmutables"
         ],
@@ -1649,6 +1672,7 @@
     },
     "0.8.8": {
         "bugs": [
+            "DataLocationChangeInInternalOverride",
             "NestedCallataArrayAbiReencodingSizeValidation",
             "UserDefinedValueTypesBug",
             "SignedImmutables"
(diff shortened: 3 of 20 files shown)


chriseth:
Patch for openzeppelin (only one contract affected): https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3293

chriseth:
Blog post: https://github.com/ethereum/solidity-blog/pull/173

cameel:
Patch for Gnosis Safe: https://github.com/gnosis/safe-contracts/pull/394 (also only one contract affected)

And Bleeps is only affected through the `Governor` contract from OpenZeppelin.

cameel:
I just pushed a commit with a workaround for this issue in Gnosis, Zeppelin and Bleeps. It's a simple replacement of the affected lines. We can just drop it from the PR if they fix the problem quickly or merge it if they don't.

leonardoalt:
Tests are failing. Is it just rebase or are there problems in the PR?

cameel:
`chk_errorcodes` probably just needs a small tweak. I haven't reviewed the PR yet though so I don't know how complete it is.

As for external tests, they are failing because this change is technically not backwards-compatible (we're treating it as a bugfix). @chriseth took care of submitting fixes to OZ and ENS (https://github.com/ensdomains/ens-contracts/pull/65), I did Gnosis. Bleeps breakage turned out to be coming from the OZ dependency.

The only reason why Gnosis, OZ and Bleeps passed here was that I added a quick and dirty workaround in case this does not get fixed upstream quickly. I did not do ENS because I did not investigate that one.

Anyway, looks like OZ and ENS merged our PRs. The only remaining one is Gnosis (zero reaction so far). So I guess I'll remove the workarounds for everything except Gnosis and ext tests should pass. Other than that, this still needs the fix for `chk_errorcode` and a review.

cameel:
I just reverted workarounds for OZ and Bleeps. Looks like my PR will be merged in Gnosis soon and then we'll be able to remove that one too.

cameel:
Looks like for ENS we'll actually have to switch to the `master` branch. We're currently at a tag but they don't have too many tags and the fix is only in `master`.

And for Bleeps we'll have to wait for an OZ release (or hard-code it to use a pre-release). [A new OZ release is currently in an open review period, which will end around April 22nd](https://forum.openzeppelin.com/t/release-candidate-for-contracts-4-6-open-review-period/27075).

libsolidity/analysis/OverrideChecker.cpp
```diff
@@ -322,6 +322,15 @@ FunctionType const* OverrideProxy::functionType() const
 	}, m_item);
 }
 
+FunctionType const* OverrideProxy::specificFunctionType() const
+{
+	return std::visit(GenericVisitor{
+		[&](FunctionDefinition const* _item) { return TypeProvider::function(*_item); },
+		[&](VariableDeclaration const*) -> FunctionType const* { solAssert(false, "Requested specifig function type of variable."); return nullptr; },
```
aarlt: "specifig" -> "specific"

cameel:
Current status regarding external tests:
- **ENS**: there's a fix upstream but we're using an earlier, tagged version. Because of this there was an extra commit here to switch ENS from that tag to `master` branch. It won't pass otherwise. This commit has to be restored (I still have it locally).
- **Bleeps**: Turns out it's actually [hard-coded for OZ 4.3.2](https://github.com/wighawag/bleeps/blob/bb90cd033f91c56e3148f0609ebc128a55c9381f/contracts/package.json#L30) so it does not pick up the latest 4.6.0 despite the fact that we're removing `package.lock`. So we need to force the newer one in `bleeps.sh`:
    ```bash
    npm install @openzeppelin/contracts@>=4.6.0
    ```
    - The problem is that on 4.6.0 there's an unrelated error:
        ```
        DeclarationError: Identifier not found or not unique.
          --> src/bleepsdao/BleepsDAOGovernor.sol:31:9:
           |
        31 |         ERC20Votes _token,
           |         ^^^^^^^^^^
        ```
        Maybe `ERC20Votes` was moved to a different location in one of the newer versions of OZ. Or it used to be imported by some transitive dependency and now no longer is. Unfortunately I think we'll have to patch that ourselves.
- **Brink**: this is a new test and we have not reported the problem there. Probably best to patch it on our side now.
- **Gnosis**: [My upstream PR](https://github.com/safe-global/safe-contracts/pull/394) has not been merged yet so patching on our side this is still needed.

cameel:
The ext tests should all be passing now. I restored the ENS and Gnosis workarounds and added a new one for Brink (also submitted a PR upstream: https://github.com/brinktrade/brink-core/pull/52).

For Bleeps it turns out that updating to OZ 4.6.0 and patching the single compilation error that appears after that makes a ton of other compilation errors pop up. So I gave up, kept it at OZ 4.3.2 and just copied the patch command we had here for OZ formerly.

cameel:
ENS is now failing. My commit updates it to the latest version and in that version apparently they have some low-level functions for doing reverts ([`LowLevelCallUtils.propagateRevert()`](https://github.com/ensdomains/ens-contracts/blob/43b483db7ae3b13b985b3600779aa4e65ca22964/contracts/utils/LowLevelCallUtils.sol#L45-L53)) or just accessing the returndata ([`LowLevelCallUtils.readReturnData()`](https://github.com/ensdomains/ens-contracts/blob/43b483db7ae3b13b985b3600779aa4e65ca22964/contracts/utils/LowLevelCallUtils.sol#L33-L43)).

[`UniversalResolver.callWithOffchainLookupPropagation()`](https://github.com/ensdomains/ens-contracts/blob/43b483db7ae3b13b985b3600779aa4e65ca22964/contracts/utils/UniversalResolver.sol#L67-L73) uses them like this:
```solidity
        bool result = LowLevelCallUtils.functionStaticCall(target, data);
        uint256 size = LowLevelCallUtils.returnDataSize();

        if(result) {
            return LowLevelCallUtils.readReturnData(0, size);
        }
```

So apparently #12860 does break some things (though in their case fix is trivial, just make these functions `view`).

cameel:
Just to be clear, this is not caused directly by this PR, merging #12860 broke it. It only shows up here due to the ENS update (which we do to get an upstream fix for the override issue and avoid having to do even more patching in external tests).

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-3",
+        "name": "DataLocationChange",
+        "summary": "It was possible to override a function and changing the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa. This caused invalid code to be generated when calling such a function internally.",
+        "description": "TODO",
+        "link": "TODO",
```
ekpyron: Those still need to be filled in I guess - apart from that the fix generally looks good.

chriseth:
Since we plan to revert https://github.com/ethereum/solidity/issues/12860 that should fix the ens test breakage.

libsolidity/analysis/OverrideChecker.cpp
```diff
@@ -589,21 +598,54 @@ void OverrideChecker::checkOverride(OverrideProxy const& _overriding, OverridePr
...
+
+		// The override proxy considers calldata and memory the same data location.
+		// Here we do a more specific check:
+		// Data locations of parameters and return variables have to match
+		// unless we have a public function overriding an external one.
+		if (
+			_overriding.isFunction() &&
+			!returnTypesDifferAlready &&
+			_super.visibility() != Visibility::External &&
+			_overriding.functionKind() != Token::Fallback
```
ekpyron, line 630: Is this really necessary? Doesn't fallback imply external visibility (both of ``_overriding`` and of ``_super``)?
Not that it hurts in any case.

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-3",
+        "name": "DataLocationChange",
+        "summary": "It was possible to override a function and change the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
```
ekpyron:
```suggestion
        "summary": "It was possible to override internal and public function while changing the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
```
?

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
...
+        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function expects calldata pointer or vice-versa.",
```
ekpyron:
```suggestion
        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function expects a calldata pointer or vice-versa.",
```

ekpyron (commented):
Only minor comments. Near-merge-ready.

libsolidity/analysis/OverrideChecker.cpp
chriseth: Yes, maybe, but we can also just leave it in...

ekpyron (dismissed):

Changelog.md
```diff
@@ -1,5 +1,9 @@
 ### 0.8.14 (unreleased)
 
+Important Bugfixes:
+ * Type Checker: Only allow changing data location between ``memory`` and ``calldata`` when overriding external functions.
```
cameel: I think it would be clearer to state what is getting disallowed:
```suggestion
 * Type Checker: Disallow changing data location between ``memory`` and ``calldata`` when overriding internal functions.
```

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-3",
+        "name": "DataLocationChange",
+        "summary": "It was possible to override internal and public function while changing the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
```
```suggestion
        "summary": "It was possible to override internal and public functions while changing the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
```

cameel: It sounds a bit weird in that order. I'd reverse that sentence:
```suggestion
        "summary": "It was possible to change the data location of the parameters or return variables from ``calldata`` to ``memory`` and vice-versa while overriding internal and public functions. This caused invalid code to be generated when calling such a function internally through virtual function calls.",
```

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
...
+        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function expects a calldata pointer or vice-versa.",
```
cameel: I'd state directly that it's being interpreted as calldata pointer and used to address calldata.
```suggestion
        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function interprets it as a calldata pointer or vice-versa.",
```

```suggestion
        "description": "When calling external functions, it is irrelevant if the data location of the parameters is ``calldata`` or ``memory``, the encoding of the data does not change. Because of that, changing the data location when overriding external functions is allowed. The compiler incorrectly also allowed a change in the data location for overriding public and internal functions. Since public functions can be called internally as well as externally, this causes invalid code to be generated when such an incorrectly overridden function is called internally through the base contract. The caller provides a memory pointer, but the called function expects a calldata pointer or vice-versa.",
```

docs/bugs_by_version.json
```diff
@@ -1333,6 +1333,7 @@
     },
     "0.6.10": {
         "bugs": [
+            "DataLocationChange",
```
cameel: This sounds very generic. Maybe it would be better to make it more specific?
```suggestion
            "DataLocationChangeInInternalOverride",
```

cameel (commented):
Some wording suggestions.

Reviewing the actual code now.

Changelog.md
chriseth: if we write 'internal', it could mean that it does not refer to public functions.

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-3",
+        "name": "DataLocationChange",
```
chriseth:
```suggestion
        "name": "DataLocationChangeInInternalOverride",
```

libsolidity/analysis/OverrideChecker.h
```diff
@@ -84,7 +84,10 @@ class OverrideProxy
 	/// @returns receive / fallback / function (only the latter for modifiers and variables);
 	langutil::Token functionKind() const;
 
-	FunctionType const* functionType() const;
+	/// @returns the externally callable function type
+	FunctionType const* externalFunctionType() const;
+	/// @returns the (unmodified) function type
+	FunctionType const* specificFunctionType() const;
```
cameel: Maybe "original" or "native" would be a better word here?
```suggestion
	/// @returns the function type exactly as in its original definition
	FunctionType const* originalFunctionType() const;
```

test/libsolidity/syntaxTests/inheritance/dataLocation/return_type_data_location_illegal.sol
```diff
@@ -0,0 +1,8 @@
+abstract contract A {
```
cameel: I'd make the test name more specific: `return_type_data_location_illegal.sol` -> `return_type_data_location_change_illegal.sol`.

test/libsolidity/syntaxTests/inheritance/dataLocation/memory_calldata_public.sol
```diff
@@ -0,0 +1,16 @@
...
+    function f(uint256[1] memory a) public virtual returns (uint256);
+    function test() external returns (uint) {
+        uint[1] memory t;
+        t[0] = 7;
+        return f(t);
+    }
+}
+
+contract B is A {
+    function f(uint256[1] calldata a) public override returns (uint256) {
```
cameel: I think we should also have the other direction (`calldata` -> `memory`) covered.

libsolidity/analysis/OverrideChecker.cpp
```diff
@@ -589,21 +598,54 @@ void OverrideChecker::checkOverride(OverrideProxy const& _overriding, OverridePr
...
+		)
+		{
+			if (!_overriding.specificFunctionType()->hasEqualParameterTypes(*_super.specificFunctionType()))
+				overrideError(
+					_overriding,
+					_super,
+					7723_error,
+					"Data locations of parameters have to be the same when overriding, but they differ.",
```
cameel: The current message might mislead users into thinking that changing location is never allowed. I'd make it more specific:
```suggestion
					"Data locations of parameters have to be the same when overriding functions that are not external, but they differ.",
```
Same in the other message below.

test/libsolidity/syntaxTests/inheritance/dataLocation/memory_calldata_internal.sol
```diff
@@ -0,0 +1,16 @@
...
+    function f(uint256[1] memory a) internal virtual returns (uint256);
+    function test() external returns (uint) {
+        uint[1] memory t;
+        t[0] = 7;
+        return f(t);
+    }
+}
+
+contract B is A {
+    function f(uint256[1] calldata a) internal override returns (uint256) {
```
cameel: Can you add an equivalent test for changing location in an overridden modifier? We have it covered but not specifically in case of changing just the data location.

test/libsolidity/syntaxTests/inheritance/dataLocation/memory_calldata_internal.sol
```diff
@@ -0,0 +1,16 @@
+abstract contract A {
```
cameel: How about this test naming?
- `memory_calldata_internal.sol` -> `parameter_data_location_change_illegal_internal.sol`
- `memory_calldata_public.sol` -> `parameter_data_location_change_illegal_public.sol`

They really should have more specific names. It's already a bit hard to check if we have a particular situation covered and the name is all you have without looking inside the file.

Changelog.md
cameel: Good point. Maybe non-external then?

cameel (dismissed):
I have a few more suggestions but overall the PR seems correct. I haven't found any serious problems so I'm approving already.

docs/bugs.json
cameel: What about this one? It's `internal and public` in the summary but `public` in the description.

cameel (dismissed):
Looks fine overall. There's just one last small inconsistency in https://github.com/ethereum/solidity/pull/12850#discussion_r871516277

cameel (dismissed):

cameel:
You can rebase this on #13027 if you want to be sure all external tests pass.

chriseth:
Rebased, so please don't merge before #13027

ekpyron (approved):

cameel (approved):

PR #12851, opened 2022-03-24
Tests for the equivalence of one and two-stage compilation via IR ports multi-file contracts

### LINKED ISSUE #11877: Tests verifying that compilation with `--experimental-via-yul` produces same EVM assembly as two-stage compilation via Yul

wechman:
fixes #11877

This PR adds support for multi-file contracts and multi-contract files in tests of the equivalence of one- and two-stage compilation via IR. Following contract has been added to the test scenarios:
[PRBMathSD59x18.sol](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/externalContracts/_prbmath/PRBMathSD59x18.sol) - library 
[PRBMathUD60x18.sol](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/externalContracts/_prbmath/PRBMathUD60x18.sol)  - library 
[snark.sol](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/externalContracts/snark.sol) - multi-contract file
[TokenCreation.sol](https://github.com/ethereum/solidity/blob/develop/test/externalTests/solc-js/DAO/TokenCreation.sol) - multi-file contracts and multi-contract files

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
         fail "The second argument must be --optimize if present."
     fi
 
-    local solidity_code="$1"
+    local solidity_file="$1"
     local optimize_flag="$2"
+
+    filename=$(basename "$1" .sol)
 
     local optimizer_flags=()
     [[ $optimize_flag == "" ]] || optimizer_flags+=("$optimize_flag")
+    [[ $optimize_flag == "" ]] || filename+="_optimize"
 
-    local ir_output
-    ir_output=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --ir-optimized --debug-info location "${optimizer_flags[@]}" |
-        sed '/^Optimized IR:$/d'
-    )
+    msg_on_error --no-stderr "$SOLC" --ir-optimized --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
+    sed '/^Optimized IR:$/d' |
+    awk -v RS= "{print > (\"${filename}_\"NR \".yul\")}"
+
+    for yul_file in $(find . -name "$filename*.yul" | sort -V); do
+        msg_on_error --no-stderr "$SOLC" --strict-assembly --asm "${optimizer_flags[@]}" "$yul_file" |
+        sed '/^======= <stdin>/d' |
+        sed '/^Text representation:$/d' >> "${yul_file/.yul/.asm}"
+    done
 
     local asm_output_two_stage asm_output_via_ir
-    asm_output_two_stage=$(
-        echo "$ir_output" |
-        msg_on_error --no-stderr "$SOLC" - --strict-assembly --asm "${optimizer_flags[@]}" |
-        sed '/^======= <stdin>/d' |
-        sed '/^Text representation:$/d'
-    )
+    for asm_file in $(find . -name "$filename*.asm" | sort -V); do
+        asm_output_two_stage+=$(sed '/^asm_output_two_stage:$/d' "$asm_file" | sed '/^=======/d')
+    done
+
     asm_output_via_ir=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --via-ir --asm --debug-info location "${optimizer_flags[@]}" |
+        msg_on_error --no-stderr "$SOLC" --via-ir --asm --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
         sed '/^======= <stdin>/d' |
-        sed '/^EVM assembly:$/d'
+        sed '/^EVM assembly:$/d' |
+        sed '/^asm_output_via_ir/d' |
+        sed '/^=======/d'
     )
 
     diff_values "$asm_output_two_stage" "$asm_output_via_ir" --ignore-space-change --ignore-blank-lines
 
     local bin_output_two_stage bin_output_via_ir
-    bin_output_two_stage=$(
-        echo "$ir_output" |
-        msg_on_error --no-stderr "$SOLC" - --strict-assembly --bin "${optimizer_flags[@]}" |
-        sed '/^======= <stdin>/d' |
-        sed '/^Binary representation:$/d'
-    )
+
+    for yul_file in $(find . -name "$filename*.yul" | sort -V); do
+        bin_output_two_stage+=$(
+	    msg_on_error --no-stderr "$SOLC" --strict-assembly --bin "${optimizer_flags[@]}" "$yul_file" |
+            sed '/^======= <stdin>/d' |
+            sed '/^Binary representation:$/d' |
+            sed '/^=======/d'
+        )
+    done
+
     bin_output_via_ir=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --via-ir --bin "${optimizer_flags[@]}" |
+        msg_on_error --no-stderr "$SOLC" --via-ir --bin "${optimizer_flags[@]}" "$solidity_file" |
         sed '/^======= <stdin>/d' |
-        sed '/^Binary:$/d'
+        sed '/^Binary:$/d' |
+        sed '/^=======/d'
     )
 
     diff_values "$bin_output_two_stage" "$bin_output_via_ir" --ignore-space-change --ignore-blank-lines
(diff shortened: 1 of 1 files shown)


test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
 
-    local ir_output
-    ir_output=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --ir-optimized --debug-info location "${optimizer_flags[@]}" |
-        sed '/^Optimized IR:$/d'
-    )
+    msg_on_error --no-stderr "$SOLC" --ir-optimized --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
+    sed '/^Optimized IR:$/d' |
+    awk -v RS= "{print > (\"${filename}_\"NR \".yul\")}"
```
cameel, line 331:
```suggestion
    msg_on_error --no-stderr "$SOLC" --ir-optimized --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
        sed '/^Optimized IR:$/d' |
        awk -v RS= "{print > (\"${filename}_\"NR \".yul\")}"
```

cameel, line 331: Could you create a small helper function out of this and put it in `common.sh`? Putting a name on it would make it much easier to deciper what it's supposed to do:
```bash
split_on_empty_lines_into_numbered_files "$path_prefix" "$path_suffix"
```

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
         fail "The second argument must be --optimize if present."
     fi
 
-    local solidity_code="$1"
+    local solidity_file="$1"
     local optimize_flag="$2"
 
+    filename=$(basename "$1" .sol)
```
cameel, line 323:
```suggestion
    output_file_prefix=$(basename "$1" .sol)
```

test/cmdlineTests.sh
ekpyron, line 331: What does this do exactly :-)?

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
+    local asm_output_two_stage asm_output_via_ir
+    for asm_file in $(find . -name "$filename*.asm" | sort -V); do
+        asm_output_two_stage+=$(sed '/^asm_output_two_stage:$/d' "$asm_file" | sed '/^=======/d')
+    done
+
     asm_output_via_ir=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --via-ir --asm --debug-info location "${optimizer_flags[@]}" |
+        msg_on_error --no-stderr "$SOLC" --via-ir --asm --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
         sed '/^======= <stdin>/d' |
```
cameel, line 346: You no longer supply code on stdin so this is superfluous.

Same in the solc command above.

test/cmdlineTests.sh
ekpyron: Ah, I think this splits at empty lines and writes into numbered separate yul files... which in turn together with the ``sed`` will result in one yul object per file... which is what we need...

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
-    local asm_output_two_stage asm_output_via_ir
-    asm_output_two_stage=$(
-        echo "$ir_output" |
-        msg_on_error --no-stderr "$SOLC" - --strict-assembly --asm "${optimizer_flags[@]}" |
+    for yul_file in $(find . -name "$filename*.yul" | sort -V); do
+        msg_on_error --no-stderr "$SOLC" --strict-assembly --asm "${optimizer_flags[@]}" "$yul_file" |
         sed '/^======= <stdin>/d' |
-        sed '/^Text representation:$/d'
-    )
+        sed '/^Text representation:$/d' >> "${yul_file/.yul/.asm}"
```
cameel, line 336: Why append? is the `.asm` file supposed to already exist?
```suggestion
        sed '/^Text representation:$/d' > "${yul_file/.yul/.asm}"
```

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
-    )
+    msg_on_error --no-stderr "$SOLC" --ir-optimized --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
+    sed '/^Optimized IR:$/d' |
+    awk -v RS= "{print > (\"${filename}_\"NR \".yul\")}"
 
-    local asm_output_two_stage asm_output_via_ir
-    asm_output_two_stage=$(
-        echo "$ir_output" |
-        msg_on_error --no-stderr "$SOLC" - --strict-assembly --asm "${optimizer_flags[@]}" |
+    for yul_file in $(find . -name "$filename*.yul" | sort -V); do
```
cameel, line 333:
```suggestion
    for yul_file in $(find . -name "${filename}*.yul" | sort -V); do
```

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
+    done
+
     asm_output_via_ir=$(
-        echo "$solidity_code" |
-        msg_on_error --no-stderr "$SOLC" - --via-ir --asm --debug-info location "${optimizer_flags[@]}" |
+        msg_on_error --no-stderr "$SOLC" --via-ir --asm --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
         sed '/^======= <stdin>/d' |
-        sed '/^EVM assembly:$/d'
+        sed '/^EVM assembly:$/d' |
+        sed '/^asm_output_via_ir/d' |
```
cameel, line 348: Not sure what this line is meant to do. `asm_output_via_ir` is a variable but here you're expecting to find its name in compiler output. It would make more sense if you just forgot `$` but then it wouldn't work either because it's in single quotes, because it's not yet defined at this point and because this sed command requires a single-line string.

Same with `asm_output_two_stage` in the loop.

ekpyron (approved):
After staring at this for a while and thinking through its ``awk``wardness, I'm reasonably convinced that this works fine.

But I'd actually suggest to rewrite this script more cleanly in python or something like that eventually - it has become too complicated for a bash script... and if we nicely rewrite it we could also allow selecting single tests or sub-test-suites and everything... I'll create an issue for that...

test/cmdlineTests.sh
```diff
@@ -317,47 +317,56 @@ function test_via_ir_equivalence()
...
+    for yul_file in $(find . -name "$filename*.yul" | sort -V); do
+        msg_on_error --no-stderr "$SOLC" --strict-assembly --asm "${optimizer_flags[@]}" "$yul_file" |
         sed '/^======= <stdin>/d' |
-        sed '/^Text representation:$/d'
-    )
+        sed '/^Text representation:$/d' >> "${yul_file/.yul/.asm}"
+    done
+
+    local asm_output_two_stage asm_output_via_ir
+    for asm_file in $(find . -name "$filename*.asm" | sort -V); do
```
cameel, line 340: This pattern could pick up files from from another run of this function because the temporary dir is reused between multiple runs. Please move the temp dir creation and removal inside the function.

cameel:
Oh, this got merged while I was reviewing it. I think it's correct in terms of functionality but could really use some cleanup. For example the apparently useless `asm_output_via_ir` part or the fact that the dir is not cleared between function runs.

cameel:
Can you submit corrections as a new PR?

wechman:
> Can you submit corrections as a new PR?

@cameel sure, I will do that!

test/cmdlineTests.sh
wechman: What you described is exactly what the commands do :)

PR #12853, opened 2022-03-24
Clarify scope of unchecked blocks.

aathan:


docs/control-structures.rst
```diff
@@ -532,7 +532,7 @@ a failing assertion.
 The ``unchecked`` block can be used everywhere inside a block, but not as a replacement
 for a block. It also cannot be nested.
 
-The setting only affects the statements that are syntactically inside the block.
+The setting only affects the statements that are lexically inside the block.
```
leonardoalt, line 535: Why is that?

aathan: I am not sure what you are asking. I can't make any assertions about the original statement. I assumed it was correct since it was in the documentation. However, I think the word "lexically" may be more precisely what is meant here? Based on other examples I saw, for example, it was not clear whether the "unchecked" flag affected code in functions called from an unchecked block. I think no. If no, then those calls may be considered "syntactically" inside the block, but they are certainly not lexically in the block...

ekpyron: To me it would seem to be rather clear that the body of a called function is neither syntactically nor lexically inside the unchecked block itself :-).
But if you think "lexically" is clearer, I'm fine with changing it. In the end it should become clear either way, since the very next sentence clarifies it anyways.

ekpyron (approved):

PR #12854, opened 2022-03-24
control-structures.rst: Clarify that low-level calls to precompiles always succeed

aathan:
Correcting the use of the word "account" to "address" ... also incorporating the issue of precompiled contact addresses mentioned elsewhere in the docs.

### DIFF
--- a/docs/control-structures.rst
+++ b/docs/control-structures.rst
@@ -584,8 +584,9 @@ of an exception instead of "bubbling up".
 .. warning::
     The low-level functions ``call``, ``delegatecall`` and
     ``staticcall`` return ``true`` as their first return value
-    if the account called is non-existent, as part of the design
-    of the EVM. Account existence must be checked prior to calling if needed.
+    if the address called is non-existent or is a precompiled contract address,
+    as part of the design of the EVM. Address existence must be checked
+    prior to calling if needed.
 
 Exceptions can contain error data that is passed back to the caller
 in the form of :ref:`error instances <errors>`.


Marenz (approved):
LGTM, maybe one other person can look over it

docs/control-structures.rst
```diff
@@ -584,8 +584,9 @@ of an exception instead of "bubbling up".
 .. warning::
     The low-level functions ``call``, ``delegatecall`` and
     ``staticcall`` return ``true`` as their first return value
-    if the account called is non-existent, as part of the design
-    of the EVM. Account existence must be checked prior to calling if needed.
+    if the address called is non-existent or is a precompiled contract address,
```
cameel, line 587: I wonder if we can really assert that about precompiles. I checked the Yellow Paper and it does not state this explicitly. It only says this:
> An account is empty when it has no code, zero nonce
> and zero balance:
> (14) EMPTY(σ, a) ≡ σ[a]c = KEC (()) ∧σ[a]n = 0∧σ[a]b = 0
> Even callable precompiled contracts can have an empty
> account state. This is because their account states do not
> usually contain the code describing its behavior.

And it's true that for empty accounts `true` is returned but would that kick in for a precompile or do they actually explicitly return a value?

If it's the latter, I think none of the current precompiles can ever revert (other than due to running out of gas) but then can we assume this about all future precompiles too?

docs/control-structures.rst
```diff
@@ -584,8 +584,9 @@ of an exception instead of "bubbling up".
...
+    as part of the design of the EVM. Address existence must be checked
+    prior to calling if needed.
```
cameel, line 589: I think the version with "account" was clearer. What does it even meant for address not to exist? They're just numbers, they all exist.

The Yellow Paper calls such a case an "empty account" and that's what I'd use too:
```suggestion
    if the account called is empty or is a precompiled contract address,
    as part of the design of the EVM. This must be checked
    prior to calling if needed.
```

cameel (commented):
Honestly, I'm unsure about this specific detail of the EVM. Would be best if @axic could look at it.

docs/control-structures.rst
axic: I do not think it is correct, some precompiles do revert on invalid/short input, albeit it is the minority.

cameel: Thanks. In that case I guess we should close the PR? The incorrect part about precompiles always succeeding seems to be its main point.

leonardoalt:
Closing this PR for now since the added info isn't fully correct.

PR #12858, opened 2022-03-25
function-modifiers.rst: Clarification about modifier arguments

aathan:
The original pronoun usage had me wondering, for example, if there wasn't some kind of syntax magic like the ``_;`` that may have promoted access to locals up to the lexical scope of the modifier. This would have been strange at the before the ``_;`` but plausible after it.

I verified with foundry tests the statements I make

### DIFF
--- a/docs/contracts/function-modifiers.rst
+++ b/docs/contracts/function-modifiers.rst
@@ -109,7 +109,17 @@ Multiple modifiers are applied to a function by specifying them in a
 whitespace-separated list and are evaluated in the order presented.
 
 Modifiers cannot implicitly access or change the arguments and return values of functions they modify.
-Their values can only be passed to them explicitly at the point of invocation.
+Those identifiers can, however, be used to explicitly pass them into the modifier as its arguments.
+When one of those values happens to be a complex value reference type, the modified members of the referenced
+value will of course be visible to all accessors of the referenced item, including the modified function and
+the scope it ultimately returns to.
+
+The reference itself and items that are non-reference value types follow the usual rules: The
+visibility of changes to them is limited to the lexical scope of the modifier function. You
+can think of the called function as shadowing the symbol of the same name. Note that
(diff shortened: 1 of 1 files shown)


docs/contracts/function-modifiers.rst
```diff
@@ -109,7 +109,17 @@ Multiple modifiers are applied to a function by specifying them in a
...
+When one of those values happens to be a complex value reference type, the modified members of the referenced
+value will of course be visible to all accessors of the referenced item, including the modified function and
+the scope it ultimately returns to.
+
+The reference itself and items that are non-reference value types follow the usual rules: The
+visibility of changes to them is limited to the lexical scope of the modifier function. You
+can think of the called function as shadowing the symbol of the same name. Note that
+as a consequence, that means changes to the value made within the
+``_;`` (i.e. the modified function) are not visible within the modifier function, even
+after the ``_;``.
```
ekpyron, line 122: I don't think this part is helpful, but rather confusing here. Basically, regular reference and value semantics applies.

aathan: Not sure what to say. Ok. The thing I'm getting at, which there may be a better way of stating, is to clarify for the reader whether this weird special thing ``_;`` operates anything like a preprocessor macro in C, and give explicit answers to some of the questions that likely pop up in readers' minds (like they did in mine).

... but it's not really worth the time/discussion per @chriseth's guidance regarding the team's time. I'll close the PR.

ekpyron: I was only referring to the cited part of the text, which specifically talks about reference types - the behaviour of which are not particularly different here than anywhere else, which is documented elsewhere.

For the very first sentence changed in this PR, I'd agree that it could use some improvement - and the last paragraph changed could also indeed be a bit clearer about local variables in the respective function body. Although in both cases, I'd not say it urgently needs improvement.

Issue #9581, opened 2020-08-06
Improve reviewability of extracted tests (for SMTChecker tests and source locations in general).


ekpyron:
Especially for the SMTChecker tests reviews would be much simpler, if one could check out the branch locally, build isoltest and have it run all tests touched in the diff from develop. In this mode ``isoltest`` should single-step through the test expectations and highlight each one (even if succeesding) in the source with full error message (including secondary location - although we might add those to the expectations anyways).

If we'd want to be fancy, this mode in isoltest could even have the option to either agree with the test or open an editor, in which a message could be typed that would automatically be added as a comment to the PR (that *should* work using the github command line tools) - but even without that, it'd be a big help.

(ping @leonardoalt for comments)

chriseth:
So the main issue here is that when looking at the test file itself, it is not easily visible what the error message refers to, right?
Or is it just about the secondary locations?

Is there a way to fix that in the test format itself?

Not specific to SMT checker, some ideas:
```
contract C {
  function f() {} //Error: No visibility specified. Did you intend to add "public"?
}
```
- 
```
contract C {
  function f() {} //E[1]
}
// [1]: Error: No visibility specified. Did you intend to add "public"?
```
-
```
contract C {
  /*E[1]*/function f() {}/*[1]E*/
}
// [1]: Error: No visibility specified. Did you intend to add "public"?
```

ekpyron:
It's a bit of both... With the SMT checker tests, especially after the counterexamples, I'm not sure they'd fit into an inline expectation all that well... and even if so, then having three of those might inflate the test case that much, s.t. it's hard to read it at once... (and SMTChecker tests often justifiably have several assertions in them...)

On the other hand, if isoltest highlighted the assertion for me and displayed me a full counterexample call chain or model for each assertion, one by one, then reviewing would be trivial...

But yeah, if we found a way to improve this without having to locally compile a PR branch and run something, it'd of course be better...

Your suggestion with the markers could work, would need to see it in action...

leonardoalt:
> But yeah, if we found a way to improve this without having to locally compile a PR branch and run something, it'd of course be better...

Also that, I think having to compile a branch locally in order to review a PR sounds like a lot of overhead

ekpyron:
[...]

Issue #8153, opened 2020-01-15
Strengthen the definition of ``pure`` potentially disallowing e.g. ``codecopy`` - or "the future of ``pure``".


ekpyron:
Came up in https://github.com/ethereum/solidity/issues/3835, in particular https://github.com/ethereum/solidity/issues/3835#issuecomment-574854014 states the issue.

The main question is: since ``view`` already allows ``staticcall``, is there any use for the coarse definition of ``pure``, that allows for expressions that are *not* compile-time constant?

Personally, I'd say we should officially strengthen pure, so that everything pure can be used in place of compile-time-constant expressions, e.g. the most basic definition of pure is "everything that can be used as array size of a statically sized array".

We *could* also move towards the opposite direction and e.g. *allow* reading the address of ``this`` in ``pure`` functions and weaken ``pure`` to mainly mean "does not read from storage". But then we need yet another stricter concept for compile-time-constant expressions. So I'm not a fan of that course of action :-).

axic:
This is what pure was supposed to be, but it was softened up somehow :wink:

chriseth:
I agree, pure functions should not be allowed to read from data that can be modified by a constructor.

ekpyron:
I agree, but then ``SemanticInformation::invalidInPureFunctions`` should be extended by ``CODECOPY`` - and there may be others, we should check. I'll move this to "accepted in 0.7.0" right away then (unless we call it "bug" and want to do it earlier).

MicahZoltu:
(sorry for the double-post, missed the comment that discussion about pure should be moved here)

@ekpyron Great breakdown (in the linked issue) of the two potential definitions of pure.  I have always treated pure as along the lines of "_if you call this method on this contract multiple times with the same inputs, you'll get the same result every time no matter what other conditions change between calls_".  This interpretation generally aligns with what I want as a dapp developer, and it also makes sense that `codecopy` would be allowed.  To me, `pure` does not mean "_this function would yield the same result no matter what contract it was implemented in_".

I can kind of see the merits to the other definitions, but IMO "it yields the same result if called multiple times" feels both the most useful and the most natural.

Can someone express the argument as to why a _user_ would want pure to mean "provides the same result no matter what bytecode the function is deployed with" is?  I want to be careful that we are implementing pure in the most pure (pun intended) way because it offers the best developer experience, not just because it is the strictest definition we can come up with.

chriseth:
[...]

Issue #4620, opened 2018-07-30
Potential improvement of syntax test annotations


chriseth:
This is inspired by the test syntax of flint by @franklinsch : https://github.com/flintlang/flint/blob/master/Tests/SemanticTests/invalid_types.flint

Perhaps we should move expected errors in syntax tests that span only a single line as a comment into that line:

```
contract C {
  function f() {} // 2-15: TypeError: ...
}
```

where the numbers indicate the covered columns. This way, modifications in other lines will not create conflicts and it is easy to see where the error applies (i.e. it might be possible to mix failing and succeeding cases).

If an error spans multiple lines, it would still appear at the bottom.

One downside is that this would not check the order of the errors.

ekpyron:
Thinking about this some more, I'm starting to like it. Maybe we should use ``///`` instead to distinguish from "normal" comments (or something else like ``//!``). We can also consider such a comment in an otherwise empty line to refer to the line above or below - however, it may be problematic that it's not quite clear which one it would be.

Maybe even:
```
contract C {
  function f(uint a) {}
         /// ^----^  Warning: Unused function parameter
}
```
Not sure about that, though.

This would be compatible with multiple expectation sections (for EVM version and pragma settings) - errors expected with this syntax would just be added to all sections.

I'll keep it in mind for the next test infrastructure overhaul.

chriseth:
I think `//!` is a good idea (`///` interferes with natspec), but putting it in a line below is probably not too useful, especially with non-fixed-width fonts or tabs.

PR #12861, opened 2022-03-25
Disallow RETURNDATACOPY and RETURNDATASIZE in inline assembly in pure functions

### LINKED ISSUE #12860: RETURNDATACOPY opcode is not disallowed in pure functions
The main feature of a pure function is that it cannot read blockchain state, which includes not only reading storage but also things like accessing external contract code, checking balance, checking remaining gas, etc. There are also cases where this is not so clear cut, e.g. CODECOPY (#12260, #8153) but we tend towards disallowing that as well. Another such case is RETURNDATACOPY. It was recently pointed out that it should not be accessible in pure functions.

The fact that it's allowed should probably be considered a bug. Otherwise it's a breaking change.

## Environment

- Compiler version: 0.8.13

## Steps to reproduce
This code compiles just fine:
```solidity
function foo() pure {
    assembly {
        returndatacopy(0, 32, 64)
    }
}
```

While the same would not work with e.g. EXTCODECOPY:
```solidity
function foo() pure {
    assembly {
        extcodecopy(0, 0, 32, 64)
    }
}
```
```
Error: Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
 --> test.sol:3:13:
  |
3 |             extcodecopy(0, 0, 32, 64)
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Extra information
Here's the full list of instructions currently disallowed in `pure` functions
https://github.com/ethereum/solidity/blob/c4909e99c1015233d6937d229058cde52615f246/libevmasm/SemanticInformation.cpp#L465-L520

cameel:
Fixes #12860.

The PR also adds comprehensive test cases for opcodes allowed/disallowed at each visibility level.


libevmasm/SemanticInformation.cpp
```diff
@@ -479,6 +479,8 @@ bool SemanticInformation::invalidInPureFunctions(Instruction _instruction)
 	case Instruction::EXTCODESIZE:
 	case Instruction::EXTCODECOPY:
 	case Instruction::EXTCODEHASH:
+	case Instruction::RETURNDATASIZE:
+	case Instruction::RETURNDATACOPY:
```
hrkrshnn, line 483: I'm not sure what should or should not be pure. AFAIK, there is no proper definition for what is a pure function in EVM. So this is entirely arbitrary.

The only restriction I can think about is related to compile time constant evaluation. Which we don't properly have yet. If that is indeed the case, there should be more instructions here.

Changelog.md
```diff
@@ -11,6 +11,7 @@ Compiler Features:
 Bugfixes:
 * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
 * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
+* View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view.
```
hrkrshnn: Should be about marking them as not pure!

libevmasm/SemanticInformation.cpp
cameel: Well, we don't even have full agreement if `pure` should refer to stuff that can be evaluated at compilation time or to stuff that does not touch state (#8153). So I think this out of scope of this PR.

This particular change to RETURNDATACOPY was explicitly requested by @chriseth:

> 0age just mentioned it the ethsecurity chat that returndatacopt is not marked as invalid in pure functions - can someone please create an issue about it and potentially even fix?

What I'm not clear about is whether this should really be considered a bug. I'd consider it a breaking change but from that comment it seems to me like @chriseth sees it as a bugfix.

Changelog.md
cameel: Well, maybe. I'll reword to clarify but I based this on the entry that we have for `chainid`, which was a similar situation:

> * View Pure Checker: Mark ``chainid`` as view.

cameel: How about now?

hrkrshnn: Looks good!

libevmasm/SemanticInformation.cpp
aarlt: Hmm.. why is this actually needed?

aarlt: Ahh.. I guess because `pure` is considered as being evaluatable at compile time, right?

cameel: Not exactly. Currently `pure` is closer to "not reading or writing", though there are discussions about changing/clarifying that. Reading returndata is a bit borderline IMO. I could see it being allowed.

aarlt: Yeah, seem to depend on how `pure` is exactly defined. I also understood it (now) as you said - not reading or writing from state (after reading the docs again). Isn't it correct to see it like that `RETURNDATACOPY` & `RETURNDATASIZE` is not reading or writing any state? 

Why should `RETURNDATACOPY` & `RETURNDATASIZE` not be `pure`? What is the big advantage to change that? Or isn't it like that `RETURNDATACOPY` & `RETURNDATASIZE`  can basically only be used, if a call was done before? `STATICCALL` is already considered as reading the state - it is already not allowed in pure functions. 

If `RETURNDATACOPY` & `RETURNDATASIZE` is used without a call (e.g. in pure functions) I see it like that `RETURNDATACOPY` will just not do anything. Where `RETURNDATASIZE` will always return `0`.

For all other cases, calls are needed, that are not allowed in `pure` functions.

hrkrshnn:
Notes from the call: allow this. Might need to think about memory operations in the future.

Also, `returndatasize` may be used by gas optimizers instead of `push1 0`. In that sense, this can break some contracts.

libevmasm/SemanticInformation.cpp
aarlt: After having a short discussion with @chriseth  and @hrkrshnn, I finally understand the underlying issue. The problem is basically, that it is possible to do a call before calling a pure function that maybe using `RETURNDATACOPY` and `RETURNDATASIZE`. In this case the values depend on the previous calls.

Thats why it makes totally sense to disallow `RETURNDATACOPY` and `RETURNDATASIZE` in pure functions.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
```diff
@@ -0,0 +1,90 @@
...
+            pop(mulmod(0, 1, 2))
+            pop(signextend(0, 1))
+            pop(keccak256(0, 1))
+            pop(0)
+            pop(mload(0))
+            mstore(0, 1)
+            mstore8(0, 1)
+            //pop(sload(0))
+            //sstore(0, 1)
+            //pop(gas())
```
aarlt, line 37: I did not check the actual current behaviour. But wouldn't it be better to not comment those out, so that we can see the error messages later?

aarlt, line 37: I did not check the actual current behaviour. But wouldn't it be better to not comment those out, so that we can see the error messages later?

cameel: Yeah, that's why I also added `inline_assembly_instructions_disallowed_pure.sol`. It contains only those commented-out opcodes.

I left them commented-out here because otherwise it's not that easy to see which ones are missing from the list.

Also, I want this case to pass without error because then it goes all the way through codegen. The failing one (obviously) stops at the analysis stage.

aarlt: not uncomment, but let the errors there.. e.g.
```
// ====
// EVMVersion: >=london
// ----
// Warning 5740: (94-1721): Unreachable code.
...
// TypeError 8961: (1759-1774): Function cannot be declared as pure because this expression (potentially) modifies the state.
// TypeError 8961: (1809-1819): Function cannot be declared as pure because this expression (potentially) modifies the state.
...
// TypeError 8961: (1919-1941): Function cannot be declared as pure because this expression (potentially) modifies the state.
```

aarlt: not uncomment, but let the errors there.. e.g.
```
// ====
// EVMVersion: >=london
// ----
// Warning 5740: (94-1721): Unreachable code.
...
// TypeError 8961: (1759-1774): Function cannot be declared as pure because this expression (potentially) modifies the state.
// TypeError 8961: (1809-1819): Function cannot be declared as pure because this expression (potentially) modifies the state.
...
// TypeError 8961: (1919-1941): Function cannot be declared as pure because this expression (potentially) modifies the state.
```

cameel: Duplicate comment.

cameel:
> Notes from the call: allow this. Might need to think about memory operations in the future.

Allow what? Allow RETURNDATACOPY in pure functions or allow this PR to be merged?

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
aarlt: Right now I would prefer having those errors.. but I also find the error messages a bit poor.. I think it would be better to know exactly what expression actually caused it

aarlt: Maybe two versions would make sense.. one where the stuff is commented out, and one that will produce an error?

cameel: I don't understand. If you don't uncomment them, how would you get the error messages?

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_disallowed_pure.sol
```diff
@@ -0,0 +1,84 @@
...
+            //pop(pc())
+        }
+    }
+}
+// ====
+// EVMVersion: >=london
+// ----
+// Warning 5740: (742-1153): Unreachable code.
+// TypeError 2527: (79-87): Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
+// TypeError 8961: (101-113): Function cannot be declared as pure because this expression (potentially) modifies the state.
```
aarlt, line 51: Ah sorry! Somehow I missed this.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
aarlt: Ah sorry, somehow I did not see the other tests.. so you are already testing both versions, allowed and disallowed.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_disallowed_pure.sol
cameel: Yeah. For each visibility level there are two files: one with allowed and one with disallowed opcodes :)

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
aarlt: > I don't understand. If you don't uncomment them, how would you get the error messages?

Everything is good! I did not saw that you already did what I meant with the `*disallowed*` tests. That makes totally sense now. Somehow I just missed those. However, I meant the error messages that are shown e.g. in the `*disallowed*` tests.. E.g. `TypeError 2527: (153-162): Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".` I was missing somehow the exact reference in the source code.. I was expecting that the thing is somehow written directly in the message. However, I also noticed now that the reference is shown, if `solc` is used:
```
Error: Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
  --> /Users/alex/git/solidity/test/libsolidity/syntaxTests/test2.sol:68:17:
   |
68 |             pop(chainid())
   |                 ^^^^^^^^^
```

aarlt: If we have e.g. `TypeError 2527: (153-162)`... wouldn't it be nice to extract the corresponding string (eg. string between characters `153-162`).. so we could also read the tests more easily.. e.g. `TypeError 2527: (153-162):  'chainid()' Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view"`.

cameel: Yeah, there were similar ideas in #4620 and #9581. Maybe we should restart the discussion there. @nishant-sachdeva also recently commented on the fact that we don't even have line numbers in tests, which is another similar idea.

The thing is, we don't really have an agreement on what to do so it requires some design discussion.

I think that adding line numbers would probably be uncontroversial, easy to implement and would help a bit so maybe we should create a `good first issue` for that? Not sure if we have enough manpower now to push through a bigger change.

aarlt:
> > Notes from the call: allow this. Might need to think about memory operations in the future.
> 
> Allow what? Allow RETURNDATACOPY in pure functions or allow this PR to be merged?

I think @hrkrshnn meant allow the disallow ;)

hrkrshnn:
I meant to say that allow this change ;)

Just do add one more comment. These opcodes `returndatasize` and `returndatacopy` are indeed used in high level calls. For example:

```solidity
contract C {
    function f() pure external returns (uint) {}
}

contract Test {
    function test() pure external returns (uint ret) {
        ret = C(address(123)).f();
    }
}
```

aarlt:
> I meant to say that allow this change ;)
> 
> Just do add one more comment. These opcodes `returndatasize` and `returndatacopy` are indeed used in high level calls. For example:
> 
> ```solidity
> contract C {
>     function f() pure external returns (uint) {}
> }
> 
> contract Test {
>     function test() pure external returns (uint ret) {
>         ret = C(address(123)).f();
>     }
> }
> ```

@cameel Maybe we should rename this PR (and commit messages) to communicate that this is only disallowing this in inline assembly? 
@hrkrshnn I guess we don't want to allow this for high-level calls, right? Shouldn't be all calls disallowed in pure?

hrkrshnn:
@aarlt About high level calls: I'd argue, yes. In particular, we should make pure to be as close to compile time evaluation as possible. Changing the behaviour for high level calls would definitely be a breaking change. Technically, this one as well, but we're trying to argue that we haven't mentioned the assembly part in the documentation.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
aarlt: Sure, having line numbers would make the navigation in files much more easy. But it will not help during reviews, because the user still need to look at numbers. I just created #12890 that adds the location string to the message. I think that increases the readability of such tests.

cameel:
@aarlt
> Maybe we should rename this PR (and commit messages) to communicate that this is only disallowing this in inline assembly?

Good point. Done.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
cameel: That's a monster PR :)

I think line numbers still help a bit. You can see them in diffs. With them it's much easier than what I do now when reviewing, i.e. counting errors manually.

ekpyron (approved):
I'm actually still a bit wary about calling this a non-breaking bugfix - especially since occurrences *after* an external call to an external pure function are arguably just fine :-)...
But it's probably fine and the code change and tests look good, so I'm approving!

We could consider mentioning this in the release notes though, or something like that.

cameel:
Personally I think it would be better to treat it as breaking because it's more of an omission than a real bug. But on the other hand we were called out on this which means that people already expect it to work that way...

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_disallowed.sol
```diff
@@ -0,0 +1,28 @@
...
+            memoryguard(0)
+            verbatim_1i_1o(hex"600202", 0)
+
+            pop(msize())
+            pop(pc())
+        }
+    }
+}
+// ----
+// SyntaxError 6553: (47-362): The msize instruction cannot be used when the Yul optimizer is activated because it can change its semantics. Either disable the Yul optimizer or do not use the instruction.
```
ekpyron, line 19: How does this actually work :-)? We only run these tests with enabled yul optimizer? I wouldn't have guessed :-).

cameel: Apparently. I was surprised too since I usually run `isoltest` without the `--optimize` flag.

ekpyron:
> Personally I think it would be better to treat it as breaking because it's more of an omission than a real bug. But on the other hand we were called out on this which means that people already expect it to work that way...

I'd consider it a proper bug - if these are called before any external call in the pure function itself, then it's definitely contrary to the intended meaning of ``pure`` (under *every* of the different views on what ``pure`` means ;-)) - you can construct pure functions that read state like that after all... (respectively pure functions with different behaviour based on some state that affected the previous external call).
We could of course take drastic measures and do control flow analysis and only flag this in paths, in which there is no pure external call up front - but that's significant effort and in 0.9 we may disallow external calls in pure functions anyways (undecided though)... and it also wouldn't be perfect and would also have false positives... so whatever we do, it's not perfectly nice and maybe disallowing as non-breaking bug-fix is indeed the least harmful way to deal with it.

cameel:
Is returndata really a part of state though? I see it more like some random piece of memory, which is why it seems like a grey area and is just easier to lump together with state than treat as a thing of its own.

ekpyron:
> Is returndata really a part of state though? I see it more like some random piece of memory, which is why it seems like a grey area and is just easier to lump together with state than treat as a thing of its own.

Yeah, I was a bit torn on this when writing that comment myself :-)... I mean - in a lot of cases the returndata of the previous call will literally be lying around somewhere in memory and thereby be accessible via ``mload``... so is ``mload`` also impure :-)?

cameel:
We need to maintain a map of polluted memory slots :P

Anyway, I think the change is good, just not 100% convinced that the case is clear enough that it should be a bugfix.

chriseth (approved):

ekpyron:
Another argument for considering this breaking after all is
https://github.com/ensdomains/ens-contracts/blob/master/contracts/utils/LowLevelCallUtils.sol

I guess those helper functions were specifically marked ``pure`` because before this change the compiler actively *suggested* to *make* them pure... after which making this an error in a non-breaking release is probably a bit mean...

chriseth:
Yeah, let's make it breaking.

PR #12868, opened 2022-03-28
Workarounds for ext test breakage (ethers.js downgrade, disabled polygon test in zeppelin, switch to compiler-only in elementfi)

cameel:
This is a bunch of workarounds for recent breakage in external tests. The PR was originally addressing just the ethers.js one but breakage keeps piling up. I decided to just add new fixes here so that it can be merged without special provileges. Otherwise the ethers.js fix on its own would not pass tests and would require a force merge. 
1) Workaround for breakage caused by https://github.com/ethers-io/ethers.js/discussions/2849.
2) Removal of the workaround added in #12733 since that seems to have been fixed upstream.
3) OpenZeppelin has a new test related to polygon that hits https://github.com/NomicFoundation/hardhat/issues/2115 and had to be disabled
4) ElementFi now completely stopped working without alchemy.io key. All I could do was to switch all presets to compile-only mode. We won't be able to run the test suite but at least we can still run compilation via IR and we'll get bytecode size in benchmarks.
    - I originally thought that only some of their tests ran with alchemy. Apparently that's pretty much all of them. They had alchemy.io key hard-coded in the config file and the current breakage is caused by it finally being removed. This explains why this ext test was so flaky.

### DIFF
--- a/test/externalTests/ens.sh
+++ b/test/externalTests/ens.sh
@@ -65,6 +65,10 @@ function ens_test
     force_hardhat_compiler_settings "$config_file" "$(first_word "$SELECTED_PRESETS")"
     yarn install
 
+    # With ethers.js 5.6.2 many tests for revert messages fail.
+    # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
+    yarn add ethers@5.6.1
+
     replace_version_pragmas
     neutralize_packaged_contracts
 
--- a/test/externalTests/euler.sh
+++ b/test/externalTests/euler.sh
@@ -68,6 +68,10 @@ function euler_test
     force_hardhat_unlimited_contract_size "$config_file"
     npm install
 
+    # With ethers.js 5.6.2 many tests for revert messages fail.
+    # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
+    npm install ethers@5.6.1
+
     replace_version_pragmas
     neutralize_packaged_contracts
 
--- a/test/externalTests/perpetual-pools.sh
+++ b/test/externalTests/perpetual-pools.sh
@@ -66,11 +66,6 @@ function perpetual_pools_test
     force_hardhat_unlimited_contract_size "$config_file" "$config_var"
     yarn install
 
-    # The project depends on @openzeppelin/hardhat-upgrades, which is currently not prepared
-    # for the parallel compilation introduced in Hardhat 2.9.0.
-    # TODO: Remove when https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/528 is fixed.
-    yarn add hardhat@2.8.4
-
     # With ethers.js 5.6.2 many tests for revert messages fail.
     # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
     yarn add ethers@5.6.1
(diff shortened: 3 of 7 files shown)


cameel:
Closing since it got accidentally merged along with #12874.

cameel:
Looks like I was wrong about it being merged. I.e. it got merged but not into develop. Reopening.

test/externalTests/euler.sh
```diff
@@ -68,6 +68,10 @@ function euler_test
     force_hardhat_unlimited_contract_size "$config_file"
     npm install
 
+    # With ethers.js 5.6.2 many tests for revert messages fail.
+    # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
+    npm install ethers@5.6.1
```
aarlt, line 73: I'm wondering why we use sometimes `npm` and sometimes `yarn`. Is there any special reason for this?

test/externalTests/ens.sh
```diff
@@ -65,6 +65,10 @@ function ens_test
     force_hardhat_compiler_settings "$config_file" "$(first_word "$SELECTED_PRESETS")"
     yarn install
 
+    # With ethers.js 5.6.2 many tests for revert messages fail.
+    # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
+    yarn add ethers@5.6.1
```
aarlt, line 70: I don't know `yarn` or `npm` very well, but what is the reason for doing `yarn add ethers@5.6.1` after `yarn install`. The same pattern I can also see later with `npm`. What is the semantic here? That `yarn install`/`npm install`  is installing a newer `ether.js`, that's why we overwrite that with an older version?

test/externalTests/perpetual-pools.sh
```diff
@@ -66,10 +66,9 @@ function perpetual_pools_test
     force_hardhat_unlimited_contract_size "$config_file" "$config_var"
     yarn install
 
-    # The project depends on @openzeppelin/hardhat-upgrades, which is currently not prepared
-    # for the parallel compilation introduced in Hardhat 2.9.0.
-    # TODO: Remove when https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/528 is fixed.
-    yarn add hardhat@2.8.4
+    # With ethers.js 5.6.2 many tests for revert messages fail.
+    # TODO: Remove when https://github.com/ethers-io/ethers.js/discussions/2849 is resolved.
+    yarn add ethers@5.6.1
```
aarlt, line 71: An here it is totally ok to just switch from `hardhat` to `ethers.js`?

test/externalTests/euler.sh
cameel: I'm trying to stick to the one the project itself uses in its README and `package.json`.

Either one would probably work.

test/externalTests/ens.sh
cameel: Yes. It's to get specifically version 5.6.1. The `install` command installs whathever `package.json` and `package-lock.json`/`yarn.lock` tell it to. For example for ENS:
```json
"ethers": "^5.1.4",
```
so it installs the latest ethers.js in the 5.x.y series (currently 5.6.2).

One way to override it would be to patch `package.json`. Another is just to install the version we want afterwards. I used to do the former but installing it afterwards is cleaner and less likely to break.

test/externalTests/perpetual-pools.sh
cameel: These are two separate workarounds that just happen to be next to each other. The `hardhat` add command was a workaround for #12733 and is no longer needed. See commit descriptions (and PR description).

cameel:
I just pushed commits adding fixes for the new breakage in zeppelin and elementfi. See updated PR description for details.

wechman (approved):
LGTM

PR #12871, opened 2022-03-28
inheritance.rst: Clarification about parameters of internal types in external functions and constructors

aathan:
The prior text said "You can use internal parameters in a constructor" ... and I had to re-read the paragraph a few times to decide exactly what was meant by "use" there (e.g., did it mean in the body? in the declaration?) etc. This proposed change disambiguates that, and also gives a relatable example that has already been discussed in the documentation.

### DIFF
--- a/docs/contracts/inheritance.rst
+++ b/docs/contracts/inheritance.rst
@@ -430,9 +430,16 @@ equivalent to ``constructor() {}``. For example:
         constructor() {}
     }
 
-You can use internal parameters in a constructor (for example storage pointers). In this case,
-the contract has to be marked :ref:`abstract <abstract-contract>`, because these parameters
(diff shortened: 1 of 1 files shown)


ekpyron:
Here, I don't really agree with the suggested changes:
- The section is about constructors, so discussing external functions is out of place. It may be unclear, what "internal parameters" means, but in that case we should rather link to another section that clarifies that terminology, if possible.
- constructors are not really "implicitly external" - they can be called externally or internally during inheritance. Or put differently: visibility does not really apply to constructors -that's why it's not specified for them. So what really happens - just as documented - is that internal argument types prevent external construction, forcing the contract to be abstract.

So I'll close this PR for now. We can consider different clarifications here (in a new PR), though.

PR #12873, opened 2022-03-28
abstract-contracts.rst: Clarify when a contract may or must be marked abstract

aathan:
Adds the other case where a contract must be abstract. I've noted the Solidity docs do not use certain words in a "strict" way. For example, the word "define" vs "declare" vs "implemented."  I would propose in future versions of the documentation, the word "define" should never be used if the word "declare" can be used instead. I think for most people coming from other compiled languages, the word "definition" is closer to the word "implementation" ... i.e., a definition of a thing includes all of what's necessary to instantiate the thing.

### DIFF
--- a/docs/contracts/abstract-contracts.rst
+++ b/docs/contracts/abstract-contracts.rst
@@ -6,12 +6,15 @@
 Abstract Contracts
 ******************
 
-Contracts need to be marked as abstract when at least one of their functions is not implemented.
-Contracts may be marked as abstract even though all functions are implemented.
+Contracts must be marked as abstract when at least one of their functions is not implemented or when
+they do not provide arguments for all of their base contract constructors.
+Even if this is not the case, a contract may still be marked abstract, such as when you do not intend
+for the contract to be created directly. Abstract contracts are similar to :ref:`interfaces` but an
+interface is more limited in what it can declare.
 
-This can be done by using the ``abstract`` keyword as shown in the following example. Note that this contract needs to be
-defined as abstract, because the function ``utterance()`` was defined, but no implementation was
-provided (no implementation body ``{ }`` was given).
+An abstract contract is declared using the ``abstract`` keyword as shown in the following example.
+Note that this contract needs to be defined as abstract, because the function ``utterance()`` is declared,
+but no implementation was provided (no implementation body ``{ }`` was given).
 
 .. code-block:: solidity
 


docs/contracts/abstract-contracts.rst
```diff
@@ -6,12 +6,15 @@
...
-Contracts need to be marked as abstract when at least one of their functions is not implemented.
-Contracts may be marked as abstract even though all functions are implemented.
-
-This can be done by using the ``abstract`` keyword as shown in the following example. Note that this contract needs to be
-defined as abstract, because the function ``utterance()`` was defined, but no implementation was
-provided (no implementation body ``{ }`` was given).
+Contracts must be marked as abstract when at least one of their functions is not implemented or when
+their constructors (or declarations) do not provide arguments for all of their base contract constructors.
+A contract may be marked abstract even if it must not be marked abstract, such as when you do not intend
+for the contract to be created directly. Abstract contracts are similar to :ref:`Interfaces` but an
```
cameel: I think the `:ref:` ID must match the case of the anchor, which is defined as
```rst
.. _interfaces:
```
so this should be:
```suggestion
for the contract to be created directly. Abstract contracts are similar to :ref:`interfaces` but an
```

docs/contracts/abstract-contracts.rst
```diff
@@ -6,12 +6,15 @@
...
 
-Contracts need to be marked as abstract when at least one of their functions is not implemented.
-Contracts may be marked as abstract even though all functions are implemented.
-
-This can be done by using the ``abstract`` keyword as shown in the following example. Note that this contract needs to be
-defined as abstract, because the function ``utterance()`` was defined, but no implementation was
-provided (no implementation body ``{ }`` was given).
+Contracts must be marked as abstract when at least one of their functions is not implemented or when
+their constructors (or declarations) do not provide arguments for all of their base contract constructors.
+A contract may be marked abstract even if it must not be marked abstract, such as when you do not intend
```
cameel: Isn't this contradictory? I.e. "must not be marked abstract" = "may never be marked abstract" in this context? How about this?
```suggestion
Even if this is not the case, a contract may still be marked abstract, such as when you do not intend
```

docs/contracts/abstract-contracts.rst
```diff
@@ -6,12 +6,15 @@
...
+for the contract to be created directly. Abstract contracts are similar to :ref:`Interfaces` but an
+interface is more limited in what it can declare.
```
cameel: I think it's a good practice to break the line at the dot whenever possible. Makes for less rewrapping when the text is later updated and in the rendered version line breaks are removed anyway.
```suggestion
for the contract to be created directly.
Abstract contracts are similar to :ref:`Interfaces` but an interface is more limited in what it can declare.
```

docs/contracts/abstract-contracts.rst
```diff
@@ -6,12 +6,15 @@
...
 ******************
 
-Contracts need to be marked as abstract when at least one of their functions is not implemented.
-Contracts may be marked as abstract even though all functions are implemented.
-
-This can be done by using the ``abstract`` keyword as shown in the following example. Note that this contract needs to be
-defined as abstract, because the function ``utterance()`` was defined, but no implementation was
-provided (no implementation body ``{ }`` was given).
+Contracts must be marked as abstract when at least one of their functions is not implemented or when
+their constructors (or declarations) do not provide arguments for all of their base contract constructors.
```
cameel: By "or declarations" I assume you mean the inheritance list? I don't think this is the best way to word it - the first thing that comes to mind for me were the state variable declarations. I'd omit that detail. Or you could link to https://docs.soliditylang.org/en/latest/contracts.html#arguments-for-base-constructors if you think this requires that detail.
```suggestion
Contracts must be marked as abstract when at least one of their functions is not implemented or when
they do not provide arguments for all of their base contract constructors.
```

cameel (commented):
Thanks for the PR! Here's some feedback from me.

Also, one general remark. I see you're making lots of tiny PRs (which is completely fine) but they all have only the file name in their titles and commit descriptions and the branch name is a cryptic `patch-N`. This makes it hard to keep track of them, especially when there are multiple edits for the same page. In the future it will be hard to see what each one was about just by looking at the PR or commit list or to find them in history. Could you please differentiate them by giving them more concrete titles, descriptions and branch names?

cameel:
By the way, don't worry about the failing `test_ext` jobs. #12868 fixes that. They'll pass when that PR is merged into `develop` and you rebase on top of it.

cameel:
> Adds the other case where a contract must be abstract. I've noted the Solidity docs do not use certain words in a "strict" way. For example, the word "define" vs "declare" vs "implemented." I would propose in future versions of the documentation, the word "define" should never be used if the word "declare" can be used instead. I think for most people coming from other compiled languages, the word "definition" is closer to the word "implementation" ... i.e., a definition of a thing includes all of what's necessary to instantiate the thing.

True. I noticed the same and I think we should try to be stricter about it.

docs/contracts/abstract-contracts.rst
aathan: ok thanks. I just haven't found the time to read up on the markup language being used in order to figure out exactly how :ref: works :)

docs/contracts/abstract-contracts.rst
aathan: I'm used to RFC's etc where the words MAY and MUST are often capitalized and have very specific meanings. Something "may" be a certain way, even when it "must" not be that way. Your suggestion is totally fine, but perhaps it's enough to just changedthe word "marked" to "declared" in my original ... and/or in yours.

From a workflow perspective, I'm not sure if it's up to me to make suggested changes to my PRs and then resubmit them?? Or will you just make the change and accept it??

docs/contracts/abstract-contracts.rst
aathan: In general the docs don't do a great job of using strictly defined words for things like "declaration" vs "definition" and what I meant here is the declaration of the contract itself (as in ``contract foo is bar ...`` vs ``contract foo is bar(1)...``

aathan:
chriseth had asked me to break up my PRs into smaller chunks. FYI, I haven't pulled down the whole repo yet (thinking I was just going to make a few spot adjustments) and have been using the github editing tools to create the patches and PRs directly on github. I'll figure out how to rename the patches here or, start editing in a local repo.

docs/contracts/abstract-contracts.rst
cameel: That's all the more reason not to use the somewhat ambiguous "declaration" here. If you really want to have it here I'd suggest "inheritance list". But I still think that just not going into that detail here is the best option. We have a whole separate section on the topic.

docs/contracts/abstract-contracts.rst
cameel: When in doubt, you take a look at the rendered version, which should let you spot problems with links or formatting.

You can find the rendered version attached as an artifact to the `b_docs` CI job (look at the list of CI jobs at the bottom of the PR). For example the artifacts of the most recent run of this job are here: [1025753](https://app.circleci.com/pipelines/github/ethereum/solidity/23355/workflows/f05b1119-afb8-4ba1-90f9-f8ae8c41c3c4/jobs/1025753/artifacts). `docs-html/contracts.html` is the document where you can find the content your currently editing.

docs/contracts/abstract-contracts.rst
cameel: > I'm used to RFC's etc where the words MAY and MUST are often capitalized and have very specific meanings. 

To be honest, the problem of these docs that they're a mix between high-level user guide and a low-level language spec. At least as it is now, I don't really see the strict use as in RFCs fitting there. I think it's good to remove ambiguities, but it should also read well in plain English.

By the way, there's an ongoing initiative to create a more rigorous language spec. See #12739.

> Something "may" be a certain way, even when it "must" not be that way. Your suggestion is totally fine, but perhaps it's enough to just changedthe word "marked" to "declared" in my original ... and/or in yours.

My problem with your version was really the conflict between "may" and "must" here and I'm not sure how the "marked"/"declared" change would help with that. I.e. it was written as if you meant "not(must(be marked abstract))" but I read it as "must(not(be marked abstract))" and I think this is the usual interpretation of "must not" in English.

> From a workflow perspective, I'm not sure if it's up to me to make suggested changes to my PRs and then resubmit them?? Or will you just make the change and accept it??

Not sure I fully understand the question, are you asking about the suggested edits I added in my comments? It's up to you to incorporate them or not. You can just take them directly, rewrite them your way or just ignore them if you think they make no sense. We can discuss them until all the threads here are resolved. Then we can approve and merge the PR. The new changes can be in new commits if that's more convenient to you but the commits will need to be squashed before the PR is merged anyway.

By the way, it would be best to mark these comment threads as resolved only once both the change and the discussion is  fully resolved. I.e. if it's a simple change with no further discussion needed, it's fine to resolve them. If not, as is the case in this one, it's also fine to keep them open so that reviewer can spot them.

cameel:
> chriseth had asked me to break up my PRs into smaller chunks. FYI, I haven't pulled down the whole repo yet (thinking I was just going to make a few spot adjustments) and have been using the github editing tools to create the patches and PRs directly on github. I'll figure out how to rename the patches here or, start editing in a local repo.

Sure, breaking it down is good. It means we will be able to discuss and merge them piecemeal instead of being stuck for a long time in a huge thread with a mix of everything.

I think it would be best to be editing locally, especially since these PRs often need a rebase (e.g. you'll need to rebase on `develop` to get rid of the CI failures because the fix has been merged there) but if you prefer doing it on github, that's fine too. We can help you squash/rebase/rename the commit before merging the PR. It would be nice to at least rename the PR though since you can do this via github's UI, just click "edit" near the title at the top of the page here.

docs/contracts/abstract-contracts.rst
cameel: I'm marking this unresolved since it looks like it has not been changed in the code yet and you did not comment anything here.

To be clear, it's fine to disagree and not change this if you think this is not a good change, but please at least leave a comment saying why :)

docs/contracts/abstract-contracts.rst
aathan: Oh!  I see the confusion that my text may create now.

aathan:
Oh ok, I thought you meant you wanted me to change the name of the branch itself (patch-18 e.g.) as github creates it when I click "Suggest Changes" after doing an edit. I posted a new issue to the github community discussion about that--I feel that button should give you the ability to instantly change the patch name. Unless I missed it it's otherwise many clicks to get to the spot where you can do so!

aathan:
I can't seem to figure out a way to make a change to my PR that is in addition to or different from clicking "Commit Suggestion" on one of your suggestions. Thanks in advance for the usage tips :)

cameel (dismissed):
Approved.

I squashed the commits and changed the commit description. Other than that it was good as is.

cameel:
> Oh ok, I thought you meant you wanted me to change the name of the branch itself (patch-18 e.g.) as github creates it when I click "Suggest Changes" after doing an edit.

I mean, I did, but I didn't know you were creating the PRs via github UI. In that case there's probably not much you can do. I'd say, just keep PR titles distinct and it's good enough for us.

aathan:
Just FYI, you can actually change the patch name. You just have to navigate to it awkwardly (back through your repos, click branches -- which only shows up if your window is wide enough -- in the repo view, then rename the branch/patch there, and create the pull request from there).

cameel:
Rebased on `develop`, which now has a fix for the failing CI jobs.

cameel (approved):

PR #12874, opened 2022-03-29
Fix broken `${}` syntax in `externalTests.sh`

cameel:
**Depends on #12868.**

I just noticed that the script is using `{$x}` instead of `${x}` that it was supposed to use. Weird that no one complained :)


wechman:
LGTM

wechman (approved):

cameel:
@wechman This PR depends on #12868 so it should have waited until that was merged. Merging it merged them both as a single branch...

wechman:
> @wechman This PR depends on #12868 so it should have waited until that was merged. Merging it merged them both as a single branch...

@cameel I am not sure why you introduced this dependency. Taking into consideration what changes are delivered in this PR I did not expect any dependencies. Anyway, next time I will be more careful.

cameel:
@wechman Standard practice when there's breakage in `develop`. Without #12868 tests won't pass and you'd not be able to merge the PR anyway. It would have to be rebased on it sooner or later. I could wait until it's merged but I don't have to because github actually handles this scenario pretty nicely - when you create a PR on top of another, github will automatically change your base branch once that other PR is merged. So it makes sense to start new branches on top of it until we have the fix in `develop`.

It would be nice if github blocked merging PRs with dependencies in addition to that but unfortunately there's no such option. All we can do is to notify the reviewer to be careful. I used to mark PRs as draft and make a note in the description but this was confusing reviewers so now we have the `has depedencies` label instead. Which, I, admittedly, forgot to apply here so sorry about that - I was busy with something else and only quickly switched the base branch when I realized that tests won't pass otherwise.

Anyway, I just noticed that your merge merged it into #12868's branch and not into `develop` as I assumed. So this still needs a proper merge. I can restore the branch but I can't reopen a PR once it's merged so I'm going to create a new one.

PR #12875, opened 2022-03-29
assembly.rst: Clarify location of zero memory following allocation pointer

aathan:
The original text may be confusing to anyone that doesn't already know the length of the memory pointer.

### DIFF
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -263,8 +263,9 @@ Here is an assembly snippet you can use for allocating memory that follows the p
     }
 
 The first 64 bytes of memory can be used as "scratch space" for short-term
-allocation. The 32 bytes after the free memory pointer (i.e., starting at ``0x60``)
-are meant to be zero permanently and is used as the initial value for
+allocation. The free memory pointer is itself 32 bytes. The following 32 bytes
+after the free memory pointer (i.e., starting at ``0x60``)
+are meant to be zero permanently, and they are used as the initial value for
 empty dynamic memory arrays.
 This means that the allocatable memory starts at ``0x80``, which is the initial value
 of the free memory pointer.
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -247,8 +247,11 @@ first.
 Memory Management
 =================
 
+This section is an abbreviated version of the information at
+:ref:`Reserved Memory Areas` and :ref:`Layout in Memory`
+
 Solidity manages memory in the following way. There is a "free memory pointer"
-at position ``0x40`` in memory. If you want to allocate memory, use the memory
+at position ``0x40`` (64) in memory. If you want to allocate memory, use the memory
 starting from where this pointer points at and update it.
 There is no guarantee that the memory has not been used before and thus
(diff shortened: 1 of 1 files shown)


docs/assembly.rst
```diff
@@ -263,8 +263,9 @@ Here is an assembly snippet you can use for allocating memory that follows the p
     }
 
 The first 64 bytes of memory can be used as "scratch space" for short-term
-allocation. The 32 bytes after the free memory pointer (i.e., starting at ``0x60``)
-are meant to be zero permanently and is used as the initial value for
+allocation. The free memory pointer is itself 32 bytes. The following 32 bytes
+after the free memory pointer (i.e., starting at ``0x60``)
+are meant to be zero permanently, and they are used as the initial value for
 empty dynamic memory arrays.
```
hrkrshnn: > he 32 bytes after the free memory pointer (i.e., starting at ``0x60``) are meant to be zero permanently and is used as the initial value forempty dynamic memory arrays.

I don't think this statement is true anymore.

docs/assembly.rst
```diff
@@ -263,8 +263,9 @@ Here is an assembly snippet you can use for allocating memory that follows the p
     }
 
 The first 64 bytes of memory can be used as "scratch space" for short-term
-allocation. The 32 bytes after the free memory pointer (i.e., starting at ``0x60``)
-are meant to be zero permanently and is used as the initial value for
+allocation. The free memory pointer is itself 32 bytes. The following 32 bytes
```
hrkrshnn: > The free memory pointer is itself 32 bytes

While that is true, it's somewhat implicit as Yul variables are the same size as EVM words (32 bytes), there is nothing special here.

docs/assembly.rst
aathan: This is mentioned also in the Layout in Memory section. If it's no longer true, should this description be left in, or perhaps changed to say something like "this memory should be left zeroed out for legacy reasons."?

docs/assembly.rst
aathan: The point of stating the free memory pointer size here was to allow new users to understand the pointer math that yields 0x80.

Actually this whole PR will be closed once you answer the above question(s). This section was changed a lot by another PR to address the Layout in Memory changes.

docs/assembly.rst
ekpyron: > > he 32 bytes after the free memory pointer (i.e., starting at `0x60`) are meant to be zero permanently and is used as the initial value forempty dynamic memory arrays.
> 
> I don't think this statement is true anymore.

As a matter of fact, the statement still is true :-) (in legacy codegen and also via-IR):
https://github.com/ethereum/solidity/blob/31b548577994397f7f849e8dd005d9d7aaa48c76/libsolidity/codegen/YulUtilFunctions.cpp#L4121

docs/assembly.rst
cameel: So, since the other question was answered, should this PR still be reviewed or will you be closing it?

leonardoalt:
Same with this one, unfortunately (https://github.com/ethereum/solidity/pull/12884).

PR #12876, opened 2022-03-29
[Docs] clarify the reason for the memory-unsafe assembly example being unsafe

aathan:
I'm not 100% sure my proposed clarification is in fact correct. Maybe the non-memory-safe nature of the first code snippet is due to some other reason. However, the immediately preceeding documentation says that accessing memory at location 0 *is* memory safe, and the example non-memory-safe snippet accesses memory at location 0!

Therefore, why is it unsafe? This is the reason I came up with. If it's wrong, lets state what the actual reason is.

### DIFF
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -313,7 +313,8 @@ Furthermore, if the assembly block assigns to Solidity variables in memory, you 
 the Solidity variables only access these memory ranges.
 
 Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block
-reverts or terminates. As an example, the following assembly snippet is not memory safe:
+reverts or terminates. As an example, the following assembly snippet is not memory safe, because the value of
+``returndatasize()`` may exceed the 64 byte scratch space:
 
 .. code-block:: solidity
 
@@ -322,7 +323,8 @@ reverts or terminates. As an example, the following assembly snippet is not memo
       revert(0, returndatasize())
     }
 
-But the following is:
+On the other hand, the following code *is* memory safe, because memory beyond the location pointed to by the
+free memory pointer can safely be used as temporary scratch space:
 
 .. code-block:: solidity
 


docs/assembly.rst
```diff
@@ -313,7 +313,9 @@ Furthermore, if the assembly block assigns to Solidity variables in memory, you
 the Solidity variables only access these memory ranges.
 
 Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block
-reverts or terminates. As an example, the following assembly snippet is not memory safe:
+reverts or terminates. As an example, the following assembly snippet is not memory safe in general, because the value of
+returndatasize() may exceed the 64 byte scratch space (however, if the returndatasize() is guaranteed to be less than 64,
+then this snippet would be memory safe):
```
ekpyron: Yeah, that's the correct reason for this to be non-memory-safe...
However, I'm a bit wary about the proviso in the parentheses... while this is true and perfectly fine, short of actually doing ``if iszero(lt(returndatasize(), 64)) { revert(0, 0) }``, there is rarely much of a "guarantee" that ``returndatasize()`` is less than 64...

So maybe, keeping this simpler is actually better? E.g. something like this:

```suggestion
reverts or terminates. As an example, the following assembly snippet is not memory safe, because the value of
``returndatasize()`` may exceed the 64 byte scratch space:
```

ekpyron:
I now commited my suggested simplification. Would you say this is still enough of a clarification?

aathan:
I made a small addition. Looks good.

docs/assembly.rst
```diff
@@ -322,7 +323,8 @@ reverts or terminates. As an example, the following assembly snippet is not memo
       revert(0, returndatasize())
     }
 
-But the following is:
+On the other hand, the following code *is* memory safe, because memory beyond the location pointed to by the
+free memory pointed is considered scratch space:
```
ekpyron:
```suggestion
On the other hand, the following code *is* memory safe, because memory beyond the location pointed to by the
free memory pointer can safely be used as temporary scratch space:
```
Rather maybe?

ekpyron (dismissed):

cameel (approved):

cameel:
Commits squashed and PR rebased on `develop`.

Issue #12864, opened 2022-03-27
Compiler is not evaluating all if/elseif/else branches to check for “Immutable variables cannot be initialized inside an if statement"


StErMi:
## Abstract

Currently, solidity compiler is correctly raising an error if an `immutable` variable is initialized only inside one branch of an if/else statement.

The following contract will raise the following compiler error: “Cannot write to immutable here: Immutable variables cannot be initialized inside an if statement.”

```solidity
// SPDX-License-Identifier: Unlicense
pragma solidity 0.8.13;

contract Immutable {
    uint256 public immutable iamImmutable;

    constructor(bool shouldInitWithZero) {
        if (shouldInitWithZero) {
            iamImmutable = 0;
        }
}
```

The same error is thrown also if the `immutable` variable is initialized in the `else` branch of the `if/else` statement.

```solidity
// SPDX-License-Identifier: Unlicense
pragma solidity 0.8.13;

contract Immutable {
    uint256 public immutable iamImmutable;

    constructor(bool shouldInitWithZero) {
        if (shouldInitWithZero) {
            iamImmutable = 0;
        } else {
            iamImmutable = 1;
        }
    }
}
```

But it's **not thrown** if we use the ternary operator to initialize the variable

```solidity
// SPDX-License-Identifier: Unlicense
pragma solidity 0.8.13;

contract Immutable {
    uint256 public immutable iamImmutable;

    constructor(bool shouldInitWithZero) {
        iamImmutable = shouldInitWithZero ? 0 : 1;
    }
}
```

## Motivation

The compiler error should be thrown **only** if the `immutable` state variable is not initialized **in all** the if/elseif/else branch.

## Backwards Compatibility

I don't think that there should be any backward compatibility problem.

ekpyron:
Thank you very much for the issue, but we won't do this for two reasons:
- No analysis of whether an immutable is written in all cases is prefect, so rather than eliminating invalid cases, we'd just keeping to accumulate special casing (think ``constructor(bool hasPropertyA) { bool doesNotHavePropertyA = !hasPropertyA; if (hasPropertyA) immutable = 1; if (doesNotHavePropertyA) immutable = 0;``).
- It hides the actual *cost* of the operation. The write to the immutable operates on memory, not on stack, so we shouldn't encourage doing it more than once anyways.
- We will be replacing immutables by a code data location in the future, which will not have the "write-once" restriction in the constructor at all, which will implicitly address this issue.

[...]

PR #12878, opened 2022-03-30
Make ImmutableValidator more precise while analyzing branches

StrongerXi:
## Motivation

Addresses #12864

## Mechanism

1. Delay error generation. Originally we issue error as soon as we see any write to immutable state variable inside a branch, that's can be too conservative, e.g., when both branches write to the variable.
2. Collect all immutable state variable written in a branch (and their position, obtained via the write expression). After analyzing both branch, we can inspect the set difference, and determine which ones are not written (thus initialized) in both paths. Then we issue error for those variables, _and_ collect the all written to continue this process while avoiding duplicated error generation.

## Backward Compatibility

This shouldn't break backward compatibility because we are generating less errors, i.e., correct programs remain correct. The error message is changed, but we can discuss and tweak that.

## Test Case

```solidity
contract Immutable {
    uint256 public immutable iamImmutable;

    constructor(bool shouldInitWithZero) {
        if (shouldInitWithZero) {
            if (shouldInitWithZero) {
                iamImmutable = 0;
            } else {
                iamImmutable = 1;
            }
        } else {
            if (true) {
                iamImmutable = 1;
            }
        }
    }
}
```
This generates a single error at the last write to `iamImmutable`.
If we remove the last conditional `if (true)`, no error is issued.

## Alternative

I actually think the Java way of handling such errors is worth to consider, i.e., it's no less precise than what this PR aims to achieve, and it issues a single error message at the variable declaration site, saying `the final field ... may not have been initialized`. Which is similar to error 2658 in Solidity.

This approach makes the algorithm a bit simpler -- we just need to keep track and filter out the variables-not-initialized-in-both-branches when we issue error 2658, but we lose the location of the problematic write.

### DIFF
--- a/libsolidity/analysis/ImmutableValidator.cpp
+++ b/libsolidity/analysis/ImmutableValidator.cpp
@@ -21,6 +21,7 @@
 #include <libsolutil/CommonData.h>
 
 #include <range/v3/view/reverse.hpp>
+#include <vector>
 
 using namespace solidity::frontend;
 using namespace solidity::langutil;
@@ -104,17 +105,39 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
 
 bool ImmutableValidator::visit(IfStatement const& _ifStatement)
 {
+	// store old context variables
 	bool prevInBranch = m_inBranch;
+	VariableDeclarationSet prevWrites = std::move(m_immutableStateVariablesWrittenInBranch);
+	m_immutableStateVariablesWrittenInBranch.clear();
 
 	_ifStatement.condition().accept(*this);
 
 	m_inBranch = true;
+	m_immutableStateVariablesWrittenInBranch.clear();
 	_ifStatement.trueStatement().accept(*this);
 
-	if (auto falseStatement = _ifStatement.falseStatement())
+	// find out immutable state variables that are written in then/else branches
+	VariableDeclarationSet thenWrites = std::move(m_immutableStateVariablesWrittenInBranch);
+	m_immutableStateVariablesWrittenInBranch.clear();
+	VariableDeclarationSet elseWrites;
+	if (Statement const* falseStatement = _ifStatement.falseStatement())
+	{
+		m_immutableStateVariablesWrittenInBranch.clear();
 		falseStatement->accept(*this);
+		elseWrites = std::move(m_immutableStateVariablesWrittenInBranch);
+		m_immutableStateVariablesWrittenInBranch.clear();
+	}
 
+	// restore and update context
 	m_inBranch = prevInBranch;
+	m_immutableStateVariablesWrittenInBranch = prevWrites;
+	for (VariableDeclaration const* var: thenWrites)
+		if (elseWrites.count(var))
+		{
+			m_immutableStateVariablesWrittenInBranch.insert(var);
+			m_initializedStateVariables.emplace(var);
+		}
+	m_stateVariablesInitializedInOnlyOneBranch += (elseWrites - thenWrites) + (thenWrites - elseWrites);
 
 	return false;
 }
@@ -211,13 +234,7 @@ void ImmutableValidator::analyseVariableReference(VariableDeclaration const& _va
 				_expression.location(),
 				"Cannot write to immutable here: Immutable variables cannot be initialized inside a loop."
 			);
-		else if (m_inBranch)
-			m_errorReporter.typeError(
-				4599_error,
-				_expression.location(),
-				"Cannot write to immutable here: Immutable variables cannot be initialized inside an if statement."
-			);
-		else if (m_initializedStateVariables.count(&_variableReference))
+		else if (writtenInCurrentControlFlow(_variableReference))
 		{
 			if (!read)
 				m_errorReporter.typeError(
@@ -238,12 +255,15 @@ void ImmutableValidator::analyseVariableReference(VariableDeclaration const& _va
 				_expression.location(),
 				"Immutable variables must be initialized using an assignment."
 			);
-		m_initializedStateVariables.emplace(&_variableReference);
+		if (m_inBranch)
+			m_immutableStateVariablesWrittenInBranch.emplace(&_variableReference);
+		else
+			m_initializedStateVariables.emplace(&_variableReference);
 	}
 	if (
 		read &&
 		m_inCreationContext &&
-		!m_initializedStateVariables.count(&_variableReference)
+		!writtenInCurrentControlFlow(_variableReference)
 	)
 		m_errorReporter.typeError(
 			7733_error,
@@ -258,14 +278,21 @@ void ImmutableValidator::checkAllVariablesInitialized(solidity::langutil::Source
 	{
 		for (VariableDeclaration const* varDecl: contract->stateVariables())
 			if (varDecl->immutable())
-				if (!util::contains(m_initializedStateVariables, varDecl))
+			{
+				if (util::contains(m_stateVariablesInitializedInOnlyOneBranch, varDecl))
+					m_errorReporter.typeError(
+						4599_error,
+						varDecl->nameLocation(),
+						"Immutable is not initialized in every branch of control flow."
+					);
+				else if (!util::contains(m_initializedStateVariables, varDecl))
 					m_errorReporter.typeError(
 						2658_error,
 						_location,
 						solidity::langutil::SecondarySourceLocation().append("Not initialized: ", varDecl->location()),
 						"Construction control flow ends without initializing all immutable state variables."
 					);
-
+			}
 		// Don't check further than the current c'tors contract
 		if (contract == m_currentConstructorContract)
 			break;
@@ -279,4 +306,13 @@ void ImmutableValidator::visitCallableIfNew(Declaration const& _declaration)
 
 	if (m_visitedCallables.emplace(_callable).second)
 		_declaration.accept(*this);
+}
+
+
+bool ImmutableValidator::writtenInCurrentControlFlow(VariableDeclaration const& _variableReference) const
+{
+	if (m_inBranch)
+		return m_immutableStateVariablesWrittenInBranch.count(&_variableReference);
+	else
+		return m_initializedStateVariables.count(&_variableReference);
 }
--- a/libsolidity/analysis/ImmutableValidator.h
+++ b/libsolidity/analysis/ImmutableValidator.h
@@ -32,13 +32,13 @@ namespace solidity::frontend
  * cannot be read before being initialized
  * cannot be read when initializing state variables inline
  * must be initialized outside loops (only one initialization)
- * must be initialized outside ifs (must be initialized unconditionally)
- * must be initialized exactly once (no multiple statements)
+ * must be initialized exactly once in each control flow (no multiple statements in the same control flow)
  * must be initialized exactly once (no early return to skip initialization)
 */
 class ImmutableValidator: private ASTConstVisitor
 {
 	using CallableDeclarationSet = std::set<CallableDeclaration const*, ASTNode::CompareByID>;
+	using VariableDeclarationSet = std::set<VariableDeclaration const*, ASTNode::CompareByID>;
 
 public:
 	ImmutableValidator(langutil::ErrorReporter& _errorReporter, ContractDefinition const& _contractDefinition):
@@ -66,17 +66,21 @@ private:
 
 	void visitCallableIfNew(Declaration const& _declaration);
 
+	bool writtenInCurrentControlFlow(VariableDeclaration const& _variableReference) const;
+
 	ContractDefinition const& m_mostDerivedContract;
 
 	CallableDeclarationSet m_visitedCallables;
 
-	std::set<VariableDeclaration const*, ASTNode::CompareByID> m_initializedStateVariables;
+	VariableDeclarationSet m_initializedStateVariables;
+	VariableDeclarationSet m_stateVariablesInitializedInOnlyOneBranch;
 	langutil::ErrorReporter& m_errorReporter;
 
 	FunctionDefinition const* m_currentConstructor = nullptr;
 	ContractDefinition const* m_currentConstructorContract = nullptr;
 	bool m_inLoop = false;
 	bool m_inBranch = false;
+	VariableDeclarationSet m_immutableStateVariablesWrittenInBranch;
 	bool m_inCreationContext = true;
 };
 
--- a/test/libsolidity/syntaxTests/immutable/conditionally_initialized.sol
+++ b/test/libsolidity/syntaxTests/immutable/conditionally_initialized.sol
@@ -1,9 +1,0 @@
-contract C {
-    uint immutable x;
-    constructor() {
-        if (false)
-            x = 1;
-    }
-}
-// ----
-// TypeError 4599: (86-87): Cannot write to immutable here: Immutable variables cannot be initialized inside an if statement.
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+* TypeChecker: Support initializing immutable variables inside if-else branches.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 
 
--- a/test/libsolidity/analysis/FunctionCallGraph.cpp
+++ b/test/libsolidity/analysis/FunctionCallGraph.cpp
@@ -2009,16 +2009,19 @@ BOOST_AUTO_TEST_CASE(conversions_and_struct_array_constructors)
 BOOST_AUTO_TEST_CASE(immutable_initialization)
 {
 	unique_ptr<CompilerStack> compilerStack = parseAndAnalyzeContracts(R"(
-		function free() pure returns (uint) { return 42; }
+		function free1() pure returns (uint) { return 42; }
+		function free2() pure returns (uint) { return 42; }
 
 		contract Base {
-			function ext() external pure returns (uint) { free(); }
-			function inr() internal pure returns (uint) { free(); }
+			function ext1() external pure returns (uint) { free1(); }
+			function ext2() external pure returns (uint) { free2(); }
+			function inr1() internal pure returns (uint) { free1(); }
+			function inr2() internal pure returns (uint) { free2(); }
 		}
 
 		contract C is Base {
-			uint immutable extImmutable = this.ext();
-			uint immutable inrImmutable = inr();
+			uint immutable extImmutable = this.ext1();
+			uint immutable inrImmutable = inr2();
 		}
 
 		contract D is Base {
@@ -2026,8 +2029,16 @@ BOOST_AUTO_TEST_CASE(immutable_initialization)
 			uint immutable inrImmutable;
 
 			constructor () {
-				extImmutable = this.ext();
-				inrImmutable = inr();
+				if (true) {
+					extImmutable = this.ext1();
+				} else {
+					extImmutable = this.ext2();
+				}
+				if (true) {
+					inrImmutable = inr2();
+				} else {
+					inrImmutable = inr1();
+				}
 			}
 		}
 	)"s);
@@ -2036,28 +2047,36 @@ BOOST_AUTO_TEST_CASE(immutable_initialization)
 	map<string, EdgeNames> expectedCreationEdges = {
 		{"Base", {}},
 		{"C", {
-			{"Entry", "function Base.inr()"},
-			{"function Base.inr()", "function free()"},
+			{"Entry", "function Base.inr2()"},
+			{"function Base.inr2()", "function free2()"},
 		}},
 		{"D", {
 			{"Entry", "constructor of D"},
-			{"constructor of D", "function Base.inr()"},
-			{"function Base.inr()", "function free()"},
+			{"constructor of D", "function Base.inr1()"},
+			{"constructor of D", "function Base.inr2()"},
+			{"function Base.inr1()", "function free1()"},
+			{"function Base.inr2()", "function free2()"},
 		}},
 	};
 
 	map<string, EdgeNames> expectedDeployedEdges = {
 		{"Base", {
-			{"Entry", "function Base.ext()"},
-			{"function Base.ext()", "function free()"},
+			{"Entry", "function Base.ext1()"},
+			{"Entry", "function Base.ext2()"},
+			{"function Base.ext1()", "function free1()"},
+			{"function Base.ext2()", "function free2()"},
 		}},
 		{"C", {
-			{"Entry", "function Base.ext()"},
-			{"function Base.ext()", "function free()"},
+			{"Entry", "function Base.ext1()"},
+			{"function Base.ext1()", "function free1()"},
+			{"Entry", "function Base.ext2()"},
+			{"function Base.ext2()", "function free2()"},
 		}},
 		{"D", {
-			{"Entry", "function Base.ext()"},
-			{"function Base.ext()", "function free()"},
+			{"Entry", "function Base.ext1()"},
+			{"Entry", "function Base.ext2()"},
+			{"function Base.ext1()", "function free1()"},
+			{"function Base.ext2()", "function free2()"},
 		}},
 	};
 
--- a/test/libsolidity/syntaxTests/immutable/conditionally_initialized_multiple_writes_fail.sol
+++ b/test/libsolidity/syntaxTests/immutable/conditionally_initialized_multiple_writes_fail.sol
@@ -1,0 +1,14 @@
+contract C {
+    uint immutable multipleWrites;
+
+    constructor() {
+        if (true) {
+            multipleWrites = 1;
+        } else {
+            multipleWrites = 1;
+        }
+        multipleWrites = 2; // error
+    }
+}
+// ----
+// TypeError 1574: (188-202): Immutable state variable already initialized.
--- a/test/libsolidity/syntaxTests/immutable/conditionally_initialized_multiple_writes_pass.sol
+++ b/test/libsolidity/syntaxTests/immutable/conditionally_initialized_multiple_writes_pass.sol
@@ -1,0 +1,12 @@
+contract C {
+    uint immutable multipleWrites;
+
+    constructor() {
+        if (true) {
+            multipleWrites = 1;
+        } else {
+            multipleWrites = 2;
+        }
+    }
+}
+// ----
--- a/test/libsolidity/syntaxTests/immutable/conditionally_initialized_nested_branch_write_fail.sol
+++ b/test/libsolidity/syntaxTests/immutable/conditionally_initialized_nested_branch_write_fail.sol
@@ -1,0 +1,14 @@
+contract C {
+    uint immutable nestedBranchWrite; // error
+
+    constructor() {
+        if (true) {
+            nestedBranchWrite = 1;
+        } else {
+            if (true)
+                nestedBranchWrite = 2;
+        }
+    }
+}
+// ----
+// TypeError 4599: (32-49): Immutable is not initialized in every branch of control flow.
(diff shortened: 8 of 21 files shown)


StrongerXi:
The more I think about it, the more I believe the alternative approach is preferable. Especially true for cases like this:
```solidity
contract Immutable {
    uint256 public immutable iamImmutable;

    constructor(bool shouldInitWithZero) {
        if (shouldInitWithZero) {
            if (shouldInitWithZero) {
                iamImmutable = 0;
            }
        } else {
            if (true) {
                iamImmutable = 1;
            }
        }
    }
}
```

chriseth:
Thank you for the pull request! I fear that the complete solution is a bit more complicated. Can you just ignore all the cases after the `else if (m_inBranch)`?

StrongerXi:
> I fear that the complete solution is a bit more complicated

Would you mind elaborating on what edge case I might be missing? Based on my limited understanding of Solidity, it seems sound if we take the alternative approach, i.e., mark any variable written only in 1 of the if/else branches to be "not properly initialized".

> Can you just ignore all the cases after the else if (m_inBranch)?

I might be misunderstanding, do you mean deleting all these chunks?
```c++
		else if (m_initializedStateVariables.count(&_variableReference))
		{
			if (!read)
				m_errorReporter.typeError(
					1574_error,
					_expression.location(),
					"Immutable state variable already initialized."
				);
			else
				m_errorReporter.typeError(
					2718_error,
					_expression.location(),
					"Immutable variables cannot be modified after initialization."
				);
		}
		else if (read)
			m_errorReporter.typeError(
				3969_error,
				_expression.location(),
				"Immutable variables must be initialized using an assignment."
			);
		m_initializedStateVariables.emplace(&_variableReference)
```

StrongerXi:
Rebase.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,49 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
 
 bool ImmutableValidator::visit(IfStatement const& _ifStatement)
 {
-	bool prevInBranch = m_inBranch;
+	auto prevInBranch = m_inBranch;
+	auto prevWrites = std::move(m_immutableStateVariablesWrittenInBranch);
```
Marenz, line 109: It seems dangerous to me to use `std::move()` here. You also access the moved variable a a few lines down and then move it again, yet to my understanding variables that have been moved should no longer be used...

Marenz: It seems dangerous to me to use std::move() here. You also access the moved variable a a few lines down and then move it again, yet to my understanding variables that have been moved should no longer be used...

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,49 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
 
 bool ImmutableValidator::visit(IfStatement const& _ifStatement)
 {
-	bool prevInBranch = m_inBranch;
+	auto prevInBranch = m_inBranch;
```
Marenz: Why the change away from `bool`?

Marenz (changes requested):

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Good point. Reverted it to keep changes minimal

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: I thought about that at first, then [this post](https://stackoverflow.com/questions/7027523/what-can-i-do-with-a-moved-from-object) convinced me that it's safe to use `clear` on it, which would put it back into a valid state.

The motivation behind using a `std::move` here is to avoid unnecessary copying. It'll be a no-issue if we use an immutable map here.

Regardless, I'm still waiting to hear feedback on the alternative approach, since it'll lead to simpler code and better error message IMO.

Marenz:
I am not sure what exactly your suggestion for the alternative approach is.

> This approach makes the algorithm a bit simpler -- we just need to keep track and filter out the variables-not-initialized-in-both-branches when we issue error 2658, but we lose the location of the problematic write.

Problematic write means a conditional write?

Could you explain the alternative approach a bit?

libsolidity/analysis/ImmutableValidator.cpp
Marenz: True, with the info in your link, the usage of `std::move` here seems okay.

chriseth:
The main issue I have with this PR is that I'm not convinced that it properly detects multiple writes to a variable in the same branch. Reporting the error about "not initialized" on the variable seems fine to me.

chriseth:
The main issue I have with this PR is that I'm not convinced that it properly detects multiple writes to a variable in the same branch. Reporting the error about "not initialized" on the variable seems fine to me.

StrongerXi:
@chriseth @Marenz I updated using the alternative approach. The extra check in `ImmutableValidator::checkAllVariablesInitialized` should make the high-level logic pretty clear.

Regarding proper detection of "multiple writes to a variable in the same branch", I believe error 2718 already takes care of it. I also added a test case to illustrate that.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -212,11 +233,7 @@ void ImmutableValidator::analyseVariableReference(VariableDeclaration const& _va
 				"Cannot write to immutable here: Immutable variables cannot be initialized inside a loop."
 			);
 		else if (m_inBranch)
-			m_errorReporter.typeError(
-				4599_error,
-				_expression.location(),
-				"Cannot write to immutable here: Immutable variables cannot be initialized inside an if statement."
-			);
+			m_immutableStateVariablesWrittenInBranch.emplace(&_variableReference);
```
chriseth: I think if this is not always an error case, we cannot continue with "else if".

Can you create a test case that detects this bug?

StrongerXi: Thanks for catching this; I think I have a better understanding of your earlier concerns now.

This PR essentially (1) relaxes the multiple write check to allow single write in each branch, and (2) checks for at least a single write in each branch. I made necessary changes for (1) and augmented test case to demonstrate that it falls back to other existing checks in relaxed cases.

StrongerXi:
Rebase.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -105,16 +106,40 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+	auto prevWrites = std::move(m_immutableStateVariablesWrittenInBranch);
 
 	_ifStatement.condition().accept(*this);
 
 	m_inBranch = true;
+	m_immutableStateVariablesWrittenInBranch.clear();
 	_ifStatement.trueStatement().accept(*this);
 
+	auto then_writes = std::move(m_immutableStateVariablesWrittenInBranch);
+	decltype(m_immutableStateVariablesWrittenInBranch) else_writes;
```
Marenz: We're using this set template at least three times. I would suggest to make an alias for it and use it  here and in the class declaration..

Marenz: if we have an alias we can also use it for `prevWrites` instead of `auto`

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -105,16 +106,40 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+		m_immutableStateVariablesWrittenInBranch.clear();
 		falseStatement->accept(*this);
+		else_writes = std::move(m_immutableStateVariablesWrittenInBranch);
+	}
 
+	// restore and update context
 	m_inBranch = prevInBranch;
+	m_immutableStateVariablesWrittenInBranch = std::move(prevWrites);
+	std::vector<ASTNode::SourceLocation const*> bad_write_locs;
+	for (auto var: then_writes) {
```
Marenz: Please keep using the style with `{` on a new line

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -105,16 +106,40 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+		if (else_writes.count(var))
+		{
+			m_immutableStateVariablesWrittenInBranch.emplace(var);
+			m_initializedStateVariables.emplace(var);
+		} else {
```
Marenz: same here, 
```suggestion
		}
		else
		{
```

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -280,3 +310,11 @@ void ImmutableValidator::visitCallableIfNew(Declaration const& _declaration)
 	if (m_visitedCallables.emplace(_callable).second)
 		_declaration.accept(*this);
 }
+
+
+bool ImmutableValidator::mustHaveBeenWrittenInCurrentControlFlow(VariableDeclaration const& _variableReference) const {
```
Marenz:
```suggestion
bool ImmutableValidator::writtenInCurrentControlFlow(VariableDeclaration const& _variableReference) const {
```

This name seems easier to read to me and conveys the same information

test/libsolidity/syntaxTests/immutable/conditionally_initialized.sol
```diff
@@ -1,9 +1,83 @@
 contract C {
-    uint immutable x;
+    uint immutable singleBranchWrite; // error
```
Marenz: The test case is comprehensive, but it would be nice to have one test case for each variable-test here, instead of one large one.
Easier to read, faster to see what fails, etc..

Marenz (changes requested):
I didn't find any code logic errors, mostly style.
Oh, a changelog entry is also missing.
Other than that, looks good to me.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Good call, done.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done.

test/libsolidity/syntaxTests/immutable/conditionally_initialized.sol
StrongerXi: Agree, done.

StrongerXi:
> I didn't find any code logic errors, mostly style. Oh, a changelog entry is also missing. Other than that, looks good to me.

Thanks!

1. All comments are addressed
2. Changelog entry added
3. `ImmutableValidator.h` documentation updated.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -105,16 +106,43 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
 	bool prevInBranch = m_inBranch;
+	VariableDeclarationSet prevWrites = std::move(m_immutableStateVariablesWrittenInBranch);
 
 	_ifStatement.condition().accept(*this);
 
 	m_inBranch = true;
+	m_immutableStateVariablesWrittenInBranch.clear();
 	_ifStatement.trueStatement().accept(*this);
 
+	auto then_writes = std::move(m_immutableStateVariablesWrittenInBranch);
```
Marenz: I didn't catch it earlier, but we use `camelCase` here, so `thenWrites` would be the correct style, same for `elseWrites`

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -280,3 +313,11 @@ void ImmutableValidator::visitCallableIfNew(Declaration const& _declaration)
 	if (m_visitedCallables.emplace(_callable).second)
 		_declaration.accept(*this);
 }
+
+
+bool ImmutableValidator::writtenInCurrentControlFlow(VariableDeclaration const& _variableReference) const {
```
Marenz: Style `{`

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -105,16 +106,43 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+		{
+			m_immutableStateVariablesWrittenInBranch.emplace(var);
+			m_initializedStateVariables.emplace(var);
+		}
+		else
+		{
+			m_stateVariablesNotInitializedInBothBranches.emplace(var);
+		}
+	}
+	for (auto var: else_writes) {
```
Marenz: `{`

Marenz (changes requested):

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done.

StrongerXi:
1. Addressed all comments.
2. Added comments to help clarify code logic.
3. Rebase.

Marenz (dismissed):

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,47 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
 
 bool ImmutableValidator::visit(IfStatement const& _ifStatement)
 {
+	// store old context variables
 	bool prevInBranch = m_inBranch;
+	VariableDeclarationSet prevWrites = std::move(m_immutableStateVariablesWrittenInBranch);
```
chriseth, line 110: At this point, `m_immutableStateVariablesWrittenInBranch` is in an undefined state. We should clear it an assert it is cleared after the `accept`. Why is it safe to move it here anyway? Can't there be a write operation inside the if condition?

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,47 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+	{
+		m_immutableStateVariablesWrittenInBranch.clear();
 		falseStatement->accept(*this);
+		elseWrites = std::move(m_immutableStateVariablesWrittenInBranch);
+	}
 
+	// restore and update context
 	m_inBranch = prevInBranch;
+	m_immutableStateVariablesWrittenInBranch = std::move(prevWrites);
+	std::vector<ASTNode::SourceLocation const*> bad_write_locs;
```
chriseth:
```suggestion
	vector<ASTNode::SourceLocation const*> badWriteLocations;
```
(style guide also says "no abbreviations", or at least keep them to a minimum)

chriseth: Oh it's actually not even used?

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,47 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+		else
+		{
+			m_stateVariablesNotInitializedInBothBranches.emplace(var);
+		}
+	}
+	for (auto var: elseWrites)
+	{
+		if (!thenWrites.count(var))
+			m_stateVariablesNotInitializedInBothBranches.emplace(var);
+	}
```
chriseth:
```suggestion
	for (auto var: thenWrites)
		if (elseWrites.count(var))
			m_initializedStateVariables.emplace(var);
	m_stateVariablesNotInitializedInBothBranches += (elseWrites - thenWrites) + (thenWrites - elseWrites);
```

libsolidity/analysis/ImmutableValidator.h
```diff
@@ -66,17 +66,21 @@ class ImmutableValidator: private ASTConstVisitor
...
 
+	bool writtenInCurrentControlFlow(VariableDeclaration const& _variableReference) const;
+
 	ContractDefinition const& m_mostDerivedContract;
 
 	CallableDeclarationSet m_visitedCallables;
 
-	std::set<VariableDeclaration const*, ASTNode::CompareByID> m_initializedStateVariables;
+	VariableDeclarationSet m_initializedStateVariables;
+	VariableDeclarationSet m_stateVariablesNotInitializedInBothBranches;
```
chriseth: Wouldn't `m_stateVariablesInitializedInOnlyOneBranch` be a better name?

chriseth: (Variables not initialized in any of the branches are not part of the set)

chriseth:
Does this need updated documentation?

libsolidity/analysis/ImmutableValidator.cpp
Marenz: That's not strictly true. It is in an `unspecified` state. 

from https://stackoverflow.com/questions/7027523/what-can-i-do-with-a-moved-from-object:

> Objects of types defined in the C++ standard library may be moved from (12.8). Move operations may be explicitly specified or implicitly generated. Unless otherwise specified, such moved-from objects shall be placed in a valid but unspecified state.

The main question now is if `emplace` is an operation that requires preconditions or not. I am still researching that question

Marenz: Or another phrasing from cppreference.com:

> in a "valid but unspecified state", meaning the object's class invariants hold (so functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from):

I interpret that, that it is completely okay to use `emplace` or `insert` after it was moved. I am not aware of any `valid` state that would prevent insertions in a `std::set` (e.g. you can't do anything to an instance to prevent `emplace` from working except memory exhaustion)

Marenz: That said, having an explicit clear after the move would def. make it failsafe and will avoid the next code reader wondering the same thing.

leonardoalt (changes requested):
I'm revoking the approval so that this doesn't get accidentally merged since there are still reviewer questions/comments.

Marenz:
@StrongerXi would you like to address the open comments or should we take the PR over from here?

StrongerXi:
> @StrongerXi would you like to address the open comments or should we take the PR over from here?

My apologies. Got very busy during the last few weeks. Will push a commit by end of today.

Marenz:
> My apologies. Got very busy during the last few weeks. Will push a commit by end of today.

All good, just pinging to see whats up :)

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Added explicit `clear` for applicable cases.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: My bad -- left from old commit. Removed.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done -- also added `m_initializedStateVariables.emplace(var);` in the if branch.

StrongerXi:
> Does this need updated documentation?

I already updated the documentation in the header.

libsolidity/analysis/ImmutableValidator.h
StrongerXi: Good call, done.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -258,14 +286,22 @@ void ImmutableValidator::checkAllVariablesInitialized(solidity::langutil::Source
...
 			if (varDecl->immutable())
-				if (!util::contains(m_initializedStateVariables, varDecl))
+			{
+				if (util::contains(m_stateVariablesNotInitializedInBothBranches, varDecl))
+					m_errorReporter.typeError(
+						4599_error,
+						varDecl->nameLocation(),
+						"Immutable is not initialized in every branch of control flow."
+					);
+				else if (util::contains(m_stateVariablesNotInitializedInBothBranches, varDecl) ||
```
StrongerXi: Removed `util::contains(m_stateVariablesNotInitializedInBothBranches, varDecl)` -- it's redundant

Marenz:
> I already updated the documentation in the header.

In this case "documentation" means the files under `docs/`, our user-side documentation

StrongerXi:
> > I already updated the documentation in the header.
> 
> In this case "documentation" means the files under `docs/`, our user-side documentation

I see. Just did some `grep`s in `docs/` and the only relevant section seems to be [this](https://github.com/ethereum/solidity/blob/develop/docs/contracts/constant-state-variables.rst#immutable). It doesn't mention anything about initialization constraints tho.

What do you think we should do here?

StrongerXi:
Rebase.

Marenz (dismissed):

test/libsolidity/syntaxTests/immutable/conditionally_initialized_multiple_writes.sol
```diff
@@ -0,0 +1,14 @@
+contract C {
+    uint immutable multipleWrites;
+
+    constructor() {
+        if (true) {
+            multipleWrites = 1;
+        } else {
+            multipleWrites = 1;
+        }
+        multipleWrites = 2; // error
```
nikola-matic: Is there a "happy case" for this, i.e. a test like this one here, except without the third write, e.g.

```
contract C {
    uint immutable multipleWrites;

    constructor() {
        if (true) {
            multipleWrites = 1;
        } else {
            multipleWrites = 2;
        }
    }
}
```

StrongerXi: Good call. Updated all test cases to add such cases.

StrongerXi:
Address comment, rebase, & update changelog.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,39 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+	m_immutableStateVariablesWrittenInBranch.clear();
 
 	_ifStatement.condition().accept(*this);
 
 	m_inBranch = true;
+	m_immutableStateVariablesWrittenInBranch.clear();
 	_ifStatement.trueStatement().accept(*this);
 
+	// find out immutable state variables that are written in then/else branches
+	auto thenWrites = std::move(m_immutableStateVariablesWrittenInBranch);
```
nikola-matic: @chriseth is not a big fan of generous `auto` usage, and whilst I somewhat disagree with him, it's part of our code style, so please unless the type is present in the statement, e.g. `auto const ptr = std::make_shared<Type>(value)`, replace all instances of `auto` with the actual type.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,39 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+	{
+		m_immutableStateVariablesWrittenInBranch.clear();
 		falseStatement->accept(*this);
+		elseWrites = std::move(m_immutableStateVariablesWrittenInBranch);
+		m_immutableStateVariablesWrittenInBranch.clear();
+	}
 
+	// restore and update context
 	m_inBranch = prevInBranch;
+	m_immutableStateVariablesWrittenInBranch = std::move(prevWrites); // last use of `prevWrites`
```
nikola-matic: So I know @Marenz had a concern with the usage of `std::move` up top, but in this case I would prefer to simply do an  assignment
```c++
m_immutableStateVariablesWrittenInBranch = prevWrites;
```
as we might add code later on that will access `prevWrites` after the move, causing us to spend time figuring out why we're getting undefined behaviour.

libsolidity/analysis/ImmutableValidator.cpp
```diff
@@ -104,17 +105,39 @@ bool ImmutableValidator::visit(MemberAccess const& _memberAccess)
...
+	for (auto var: thenWrites)
+		if (elseWrites.count(var))
+		{
+			m_immutableStateVariablesWrittenInBranch.emplace(var);
```
nikola-matic: Not sure what the type of `var` is (hence the `auto` comment), but unless it's using perfect forwarding, this is gonna be equivalent to `insert(...)`. This is a nit though, so feel free to ignore this comment even if I'm right.

test/libsolidity/syntaxTests/immutable/conditionally_initialized_nested_branch_write.sol
```diff
@@ -0,0 +1,29 @@
...
+        if (true) {
+            nestedBranchWrite = 1;
+        } else {
+            if (true)
+                nestedBranchWrite = 2;
+        }
+    }
+}
+
+contract C2 {
```
nikola-matic: Please move the happy case to a separate file.

nikola-matic (changes requested):
I believe you also need to add a boost test to `test/libsolidity/analysis/FunctionCallGraph.cpp`, but I may be mistaken. See `BOOST_AUTO_TEST_CASE(immutable_initialization)`.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: This is a meant to be an efficient "restoration" of the `m_immutableStateVariablesWrittenInBranch` context member variable. So `prevWrites` will not (and should not) be used for any other purposes.

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Done

libsolidity/analysis/ImmutableValidator.cpp
StrongerXi: Agree, changed `auto` to explicit type and `emplace` to `insert`.

test/libsolidity/syntaxTests/immutable/conditionally_initialized_nested_branch_write.sol
StrongerXi: Done.

StrongerXi:
> I believe you also need to add a boost test to `test/libsolidity/analysis/FunctionCallGraph.cpp`, but I may be mistaken. See `BOOST_AUTO_TEST_CASE(immutable_initialization)`.

Thanks, done.

StrongerXi:
Offline chat -- a more general approach is preferred instead of covering control flow one-by-one (e.g., another case would be #13433).

nikola-matic (commented):
Hey @StrongerXi, so we've discussed this on our team language design call, and the opinions are somewhat split regarding whether to merge this in, or close it in lieu of a more general solution. Most are fine with merging it in, but since @ekpyron would prefer a general approach, and he's off until October, we'll have to wait for his approval, so if we forget, ping us from October 1st to resume discussion.

On the other hand, we're happy with the contribution, so feel free to pick up another issue in the meantime if you feel like it :)

ekpyron:
I'm sorry about not doing this sooner, but especially since we will replace the immutable logic by a code data location that will not require the write-once restriction in the constructor at all, the added complexity here is not worth the temporary benefits it may or may not bring.

So thank you very much for the contribution and for your efforts, and again I apologize for not clarifying this sooner, but I'll be closing this PR now.

PR #12879, opened 2022-03-30
Clarify leading relative path segments in import paths

aathan:
Note: The "Preview" tab on github did *NOT* show the .. code:: block around line ~290 even before my changes. Not sure what's up with that. The rst seems correct, and it displays properly  at ``https://docs.soliditylang.org/en/latest/path-resolution.html#import-callback``

Also, I inferred the rule for ``.`` in the leading part of the import path based on the description that the import path itself is first normalized. My edit may therefore be incorrect if ``..`` segments in the leading part of the import path behave the same way as they do in the instant rule (e.g. if ``././../foo.bar`` in the import path means ``./foo.bar``). Doubtful.

Also note, the "normalization" of the import path is not exactly the normalization referred to in the CLI path normalization, since it remains relative. Probably ok not to mention explicitly.

### DIFF
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -252,12 +252,14 @@ The compiler computes a source unit name from the import path in the following w
 
 1. First a prefix is computed
 
-    - Prefix is initialized with the source unit name of the importing source unit.
-    - The last path segment with preceding slashes is removed from the prefix.
-    - Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
+    a) Prefix is initialized with the source unit name of the importing source unit.
+    b) The last path segment with preceding slashes is removed from the prefix.
+    c) Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
       characters is considered.
-      For every ``..`` segment found in this part the last path segment with preceding slashes is
-      removed from the prefix.
+      For every ``.`` segment found in this part of the import path, the segment is removed from the
+      import path.
+      For every ``..`` segment found in this part of the import path the last path segment with
+      preceding slashes is removed from the prefix, and the corresponding element in the import path
 
 2. Then the prefix is prepended to the normalized import path.
    If the prefix is non-empty, a single slash is inserted between it and the import path.
@@ -291,6 +293,14 @@ Here are some examples of what you can expect if they are not:
     import "../util/../array/util.sol"; // source unit name: lib/src/array/util.sol
     import "../.././../util.sol";       // source unit name: util.sol
     import "../../.././../util.sol";    // source unit name: util.sol
+
+.. note::
+
+   This rule may lead to counterintuitive behavior for prefixes that end in relative path segments.
+   For example, when used to compute the import path ``../bar.sol``, the prefix ``a/b/..``
+   (e.g., from a source file ``a/b/../foo.sol``) becomes ``a/b`` and the import path becomes ``bar.sol``.
+   The result is ``a/b/bar.sol``. You may have expected the result to be ``bar.sol`` because
+   ``a/b/../../bar.sol`` evaluates to ``bar.sol`` when normalized as a full path without this rule.
 
 .. note::
 


docs/path-resolution.rst
```diff
@@ -252,12 +252,14 @@ The compiler computes a source unit name from the import path in the following w
...
-    - The last path segment with preceding slashes is removed from the prefix.
-    - Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
+    a) Prefix is initialized with the source unit name of the importing source unit.
+    b) The last path segment with preceding slashes is removed from the prefix.
+    c) Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
       characters is considered.
-      For every ``..`` segment found in this part the last path segment with preceding slashes is
-      removed from the prefix.
+      For every ``.`` segment found in this part of the import path, the segment is removed from the
+      import path.
```
leonardoalt, line 260:
```suggestion
      Every ``.`` segment found in this part of the import path is removed.
```

docs/path-resolution.rst
```diff
@@ -252,12 +252,14 @@ The compiler computes a source unit name from the import path in the following w
...
+      For every ``..`` segment found in this part of the import path the last path segment with
+      preceding slashes is removed from the prefix, and the corresponding element in the import path
```
leonardoalt, line 262:
```suggestion
      preceding slashes is removed from the prefix.
```

leonardoalt:
@cameel wanna take a quick look on the added `note`?

docs/path-resolution.rst
```diff
@@ -252,12 +252,14 @@ The compiler computes a source unit name from the import path in the following w
 
 1. First a prefix is computed
 
-    - Prefix is initialized with the source unit name of the importing source unit.
-    - The last path segment with preceding slashes is removed from the prefix.
-    - Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
+    a) Prefix is initialized with the source unit name of the importing source unit.
+    b) The last path segment with preceding slashes is removed from the prefix.
+    c) Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
```
cameel, line 257: I don't think this is a valid list in RST. You can use `#.` for ordered lists though:
```suggestion
    #. Prefix is initialized with the source unit name of the importing source unit.
    #. The last path segment with preceding slashes is removed from the prefix.
    #. Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
```

docs/path-resolution.rst
cameel: ~I thought it was obvious. If we want to be pedantic here, we should also say that the slashes following the segment are removed too.~

docs/path-resolution.rst
cameel: The sentence seems cut off but if we really want to add the sentence about `.`, it does make sense to say that about `..` too:

```suggestion
      preceding slashes is removed from the prefix, and the corresponding element in the import path with any slashes following it is removed as well.
```

docs/path-resolution.rst
```diff
@@ -292,6 +294,14 @@ Here are some examples of what you can expect if they are not:
     import "../.././../util.sol";       // source unit name: util.sol
     import "../../.././../util.sol";    // source unit name: util.sol
 
+.. note::
+
+   This rule may lead to counterintuitive behavior for prefixes that end in relative path segments.
+   For example, when used to compute the import path ``../bar.sol``, the prefix ``a/b/..``
+   (e.g., from a source file ``a/b/../foo.sol``) becomes ``a/b`` and the import path becomes ``bar.sol``.
+   The result is ``a/b/bar.sol``. You may have expected the result to be ``bar.sol`` because
+   ``a/b/../../bar.sol`` evaluates to ``bar.sol`` when normalized as a full path without this rule.
```
cameel, line 303: The examples above already show this case and more.

The initial draft of this page went into detail about this and other unintuitive cases but we intentionally tuned that down not to distract from the main point. I don't think it's a good idea to go back to that version.

cameel (changes requested):
Sorry, but I don't think that the proposed changes really add much to the page.

In any case, they would need some corrections first.

docs/path-resolution.rst
cameel: Wait, no, this was actually omitted because it's implied by the import path being normalized.

cameel:
> Also note, the "normalization" of the import path is not exactly the normalization referred to in the CLI path normalization, since it remains relative. Probably ok not to mention explicitly.

It's not the CLI normalization. The normalization is explicitly defined later in the same section:
>> The normalization rules are the same as for UNIX paths, namely:
>> - All the internal ``.`` segments are removed.
>> - Every internal ``..`` segment backtracks one level up in the hierarchy.
>> - Multiple slashes are squashed into a single one.

> Also, I inferred the rule for ``.`` in the leading part of the import path based on the description that the import path itself is first normalized. My edit may therefore be incorrect if ``..`` segments in the leading part of the import path behave the same way as they do in the instant rule (e.g. if ``././../foo.bar`` in the import path means ``./foo.bar``). Doubtful.

Yes, the rule for `.` is implied by the fact that the import path is normalized. It's just that normalization can still leave leading `..` segments in the path. E.g. `.././../a//b/..//c/./d.sol` would be normalized to `../../a/c/d.sol` according to the rules stated in that section.

I think this whole confusion comes from the fact that the description does not explicitly say to remove these leading `..` segments from the import path before joining it with the prefix. I guess we can consider that a documentation bug.

Still, I don't think this PR is the best way to reword it. I have submitted my own, which fixes that and simplifies the whole description: #13572.

PR #12882, opened 2022-03-31
Reword the section about cleaning up variables `variable_cleanup.rst`

aathan:
The original section title is misleading, because it sounds like it would be a discussion of memory allocation cleanup -- and that's not really a thing in Solidity or the EVM. The section actually deals with whether padding bits are zeroed. There's also some really low level discussion of JUMPI with quite a head-scratcher of an example!

### DIFF
--- a/docs/internals/variable_cleanup.rst
+++ b/docs/internals/variable_cleanup.rst
@@ -1,33 +1,34 @@
 .. index: variable cleanup
 
-*********************
-Zeroing of Padding
-*********************
+*************************
+Cleaning of Padding Bits
+*************************
 
 When a value is shorter than 256 bit, in some cases the remaining (padding) bits
-must be zeroed. The Solidity compiler is designed to zero any padding bits before operations
-that might be adversely affected by garbage in the padding bits.
+must be cleaned up, by sign extending the shorter value. The Solidity
+compiler is designed to clean up padding bits before operations
+that might be adversely affected by garbage in those padding bits.
 For example, before writing a value to  memory, the padding bits need
-to be cleared because the memory contents could be used for computing
+to be cleaned because the memory contents could be used for computing
 hashes or sent as the data of a message call. Similarly, before
 storing a value in storage, padding bits need to be cleaned
-because otherwise their value can be observed, and may leak information.
+because otherwise their value can be observed.
 
-Note that inline assembly does not automatically zero padding bits
+Note that inline assembly instructions do not automatically clean padding bits
 of values used therein.
-If you use inline assembly to access and modify Solidity variables
+Furthermore, if you use inline assembly to access and modify Solidity variables
 shorter than 256 bits, the compiler does not do any automated
-cleanup of the padding bits. It is up to you to store zeroed
+cleanup of the padding bits either. It is up to you to store cleaned
 values back into the source memory or storage location.
 
 Moreover, the assembly emitted by the compiler may not zero the
-padding bits of temporary values that are not stored anywhere,
+padding bits of values,
 particularly when the following instruction(s) would behave the
 same irrespective of the content of the padding bits. Low level
 temporary values that are inputs to the `JUMPI` opcode are one
 example.
 
-In addition to the design principle above, the Solidity compiler
+In addition, the Solidity compiler
 cleans input data when it is loaded onto the stack.
 
 Different types have different rules for cleaning up invalid values:
--- a/docs/internals/variable_cleanup.rst
+++ b/docs/internals/variable_cleanup.rst
@@ -12,7 +12,9 @@ For example, before writing a value to  memory, the padding bits need
 to be cleaned because the memory contents could be used for computing
 hashes or sent as the data of a message call. Similarly, before
 storing a value in storage, padding bits need to be cleaned
-because otherwise their value can be observed.
+because otherwise their value, which may not be deterministic across
+EVM or Solidity versions, may affect operations such as cryptographic
+functions or other operations which observe those values.
 
 Note that inline assembly instructions do not automatically clean padding bits
 of values used therein.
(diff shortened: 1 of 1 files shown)


docs/internals/variable_cleanup.rst
```diff
@@ -1,29 +1,31 @@
...
-must be cleaned.
-The Solidity compiler is designed to clean such remaining bits before any operations
-that might be adversely affected by the potential garbage in the remaining bits.
-For example, before writing a value to  memory, the remaining bits need
-to be cleared because the memory contents can be used for computing
-hashes or sent as the data of a message call.  Similarly, before
-storing a value in the storage, the remaining bits need to be cleaned
-because otherwise the garbled value can be observed.
+When a value is shorter than 256 bit, in some cases the remaining (padding) bits
+must be zeroed. The Solidity compiler is designed to zero any padding bits before operations
```
hrkrshnn: Cleaned is the correct word. For signed integers, zeroing is incorrect, the right operation is `signextend`, which would make the higher order bits to be 1 for some cases.

docs/internals/variable_cleanup.rst
```diff
@@ -1,29 +1,31 @@
...
+hashes or sent as the data of a message call. Similarly, before
+storing a value in storage, padding bits need to be cleaned
+because otherwise their value can be observed, and may leak information.
 
-Note that access via inline assembly is not considered such an operation:
-If you use inline assembly to access Solidity variables
-shorter than 256 bits, the compiler does not guarantee that
-the value is properly cleaned up.
+Note that inline assembly does not automatically zero padding bits
+of values used therein.
```
hrkrshnn:
```suggestion
```
Isn't this redundant? The following lines mention this.

docs/internals/variable_cleanup.rst
```diff
@@ -1,29 +1,31 @@
 .. index: variable cleanup
 
 *********************
-Cleaning Up Variables
+Zeroing of Padding
```
hrkrshnn: See my other comment, zeroing is not generally correct.

docs/internals/variable_cleanup.rst
aathan: Well, the word "cleaned" was ambiguous to me. I was thinking, did it mean "zero" or did it mean sign extend?

The reason I went on this excursion is that the original text contains this statement: "the remaining bits need to be cleared because the memory contents can be used for computing hashes"

Note the word "cleared", which maybe it was meant to read "cleaned".

I can edit this documentation to clarify what is meant by "cleaned" if it means sign-extend even in the context of computing hashes. Please advise.

docs/internals/variable_cleanup.rst
aathan: This sentence is about what the assembly itself does. The following lines are about what the compiler does.

I'll clean up the language when I clean up the prior issue.

docs/internals/variable_cleanup.rst
ekpyron: It should indeed say ``cleaned`` instead of ``cleared``.
The table in the end of this section should explain, what "cleaning" means, though  (not that things cannot be improved to make this clearer).

cameel:
Rebased on `develop`.

docs/internals/variable_cleanup.rst
```diff
@@ -1,29 +1,31 @@
...
-storing a value in the storage, the remaining bits need to be cleaned
-because otherwise the garbled value can be observed.
+When a value is shorter than 256 bit, in some cases the remaining (padding) bits
+must be zeroed. The Solidity compiler is designed to zero any padding bits before operations
+that might be adversely affected by garbage in the padding bits.
+For example, before writing a value to  memory, the padding bits need
+to be cleared because the memory contents could be used for computing
+hashes or sent as the data of a message call. Similarly, before
+storing a value in storage, padding bits need to be cleaned
+because otherwise their value can be observed, and may leak information.
```
cameel: "Leak" sounds as if it was about security. I think it's really more about potentially affecting the results of operations. There's no memory protection so contract could get at these values anyway if it wanted to.

chriseth:
Please avoid changing the location of newlines. It makes this much harder to review.

docs/internals/variable_cleanup.rst
aathan: Maybe so, but note that "their value can be observed" is from the original text. Not really sure what concern that would be referring to other than a concern over being observed (i.e., an information leakage concern). That being said, I'll just avoid using the word "leak."

aathan:
Hey, I don't want to burden you with doc review so I've simply changed the wording of this PR back to "cleaned" (from cleared/zeroed) given @[hrkrshnn](https://github.com/hrkrshnn)'s comments. I've also avoided attempting to reword the table that follows which talks about "Invalid Values"

I will note however that it seems like that content is dated and misplaced? It seems like the "Invalid Values Mean" column is referring to how the compiler deals not so much with cleaning, but with stored values or operations that overflow the underlying type???  Thus, where the table says "in the future exceptions will be thrown" is now outdated right?

docs/internals/variable_cleanup.rst
cameel: Yes, the concern is about it being observed. But it's not a concern because the information is secret (as calling it a "leak" implies). It's because relying on it can make the behavior of your contract non-deterministic. While it's not really random, it may depend on low-level details of contract execution and compiler implementation details.

cameel: For example the values before being cleaned could be different between the legacy and IR codegen. Or with and without the optimizer.

docs/internals/variable_cleanup.rst
```diff
@@ -1,31 +1,36 @@
 .. index: variable cleanup
 
-*********************
-Cleaning Up Variables
-*********************
+*************************
+Cleaning of Padding Bits
```
chriseth, line 4: In my view, "padding" is only something that is used in byte arrays, i.e. data structures of dynamic length.

leonardoalt:
The reviews seem to indicate that the changes are not really needed/consistent. Reviewers please comment otherwise.

Issue #12832, opened 2022-03-22
Error when array length is an expression involving only literals and type conversions: `Invalid array length, expected integer literal or constant expression`


bshastry:
```
contract C {
  function f() external {
    uint[uint256((((((77203342890060379470995536562582626741748008201257474317563024391158357352195 ** uint152(1020873840434952056135838088180821062407121094)) >> 56910752605302543330500802966993239962334264168338881516084502707322707229319) ** 82705070667290639185515540116081796374718317434275556705609245196574547402355) + 66683721687286726833199636002753785367603597234526287466620826223496546038896) & 87485231878771483395216945770994664543542803176944483799433554461795287649940))] memory x;
  }
}
```

throws the said error although the length literal is an integer literal and a constant expression.

chriseth:
I think the issue is that the constant is too large - we should improve the error message.

bshastry:
> I think the issue is that the constant is too large - we should improve the error message.

```
contract C {
  function ()[uint8(5)] x;
}
```

also errors similarly. But here there is something else wrong.

bshastry:
On debugging, it looks like the constant evaluator does not evaluate explicitly typed constant literals like `uint8(1)`.

bshastry:
Minor update: This error is also encountered when statically defined array size is a compile-time constant but is not in a form that will be evaluated as such because it is not of a `RationalNumberType` e.g., `bool[uint8(1)] x;`

StrongerXi:
> Minor update: This error is also encountered when statically defined array size is a compile-time constant but is not in a form that will be evaluated as such because it is not of a `RationalNumberType` e.g., `bool[uint8(1)] x;`

I believe this should be dealt with in `ConstantEvaluator.cpp`. Indeed it's not visiting such AST variant at the moment. We can raise another issue if needed, and I can send a PR for that.

cameel:
The constant is definitely too large. We don't really support exponentiation in arbitrary precision with such large numbers. If you remove explicit conversions, you get this error message:
```
Error: Operator ** not compatible with types int_const 7720...(69 digits omitted)...2195 and int_const 1020...(38 digits omitted)...1094
```
And you don't even need numbers that large to get this error. You'd get it even with something like `10 ** 5000`.

> Minor update: This error is also encountered when statically defined array size is a compile-time constant but is not in a form that will be evaluated as such because it is not of a `RationalNumberType` e.g., `bool[uint8(1)] x;`

[...]

PR #12885, opened 2022-04-01
Improve error message for static array with size too large

StrongerXi:
# Motivation
Addresses #12832

# Mechanism
Simply disambiguate the (1) array with size too large, and (2) array with non-constant integer size cases and generate a new error for (1).

# Backward Compatibility
I fixed all the existing test cases, and hopefully this PR should only make the error messages better, rather than breaking existing applications.

### DIFF
--- a/libsolidity/analysis/DeclarationTypeChecker.cpp
+++ b/libsolidity/analysis/DeclarationTypeChecker.cpp
@@ -297,7 +297,7 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
 		else if (optional<ConstantEvaluator::TypedRational> value = ConstantEvaluator::evaluate(m_errorReporter, *length))
 			lengthValue = value->value;
 
-		if (!lengthValue || lengthValue > TypeProvider::uint256()->max())
+		if (!lengthValue)
 			m_errorReporter.typeError(
 				5462_error,
 				length->location(),
@@ -309,6 +309,12 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
 			m_errorReporter.typeError(3208_error, length->location(), "Array with fractional length specified.");
 		else if (*lengthValue < 0)
 			m_errorReporter.typeError(3658_error, length->location(), "Array with negative length specified.");
+		else if (lengthValue > TypeProvider::uint256()->max())
+			m_errorReporter.typeError(
+				1847_error,
+				length->location(),
+				"Array length too large, maximum is 2**256 - 1."
+			);
 
 		_typeName.annotation().type = TypeProvider::array(
 			DataLocation::Storage,
--- a/test/libsolidity/syntaxTests/array/length/too_large.sol
+++ b/test/libsolidity/syntaxTests/array/length/too_large.sol
@@ -1,5 +1,8 @@
 contract C {
     uint[8**90] ids;
+    uint[2**256-1] okay;
+    uint[2**256] tooLarge;
 }
 // ----
-// TypeError 5462: (22-27): Invalid array length, expected integer literal or constant expression.
+// TypeError 1847: (22-27): Array length too large, maximum is 2**256 - 1.
+// TypeError 1847: (68-74): Array length too large, maximum is 2**256 - 1.
--- a/test/libsolidity/syntaxTests/array/length/bytes32_too_large.sol
+++ b/test/libsolidity/syntaxTests/array/length/bytes32_too_large.sol
@@ -2,4 +2,4 @@ contract C {
     bytes32[8**90] ids;
 }
 // ----
-// TypeError 5462: (25-30): Invalid array length, expected integer literal or constant expression.
+// TypeError 1847: (25-30): Array length too large, maximum is 2**256 - 1.
--- a/test/libsolidity/syntaxTests/array/length/bytes32_too_large_multidim.sol
+++ b/test/libsolidity/syntaxTests/array/length/bytes32_too_large_multidim.sol
@@ -2,4 +2,4 @@ contract C {
     bytes32[8**90][500] ids;
 }
 // ----
-// TypeError 5462: (25-30): Invalid array length, expected integer literal or constant expression.
+// TypeError 1847: (25-30): Array length too large, maximum is 2**256 - 1.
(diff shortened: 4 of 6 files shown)


libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -297,12 +297,18 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
 		else if (optional<ConstantEvaluator::TypedRational> value = ConstantEvaluator::evaluate(m_errorReporter, *length))
 			lengthValue = value->value;
 
-		if (!lengthValue || lengthValue > TypeProvider::uint256()->max())
```
StrongerXi, line 300: Apparently the comparison in this old check is ~missing an asterisk~, when I just copy-pasted this, an edge case failed (`arr[2**256-1]`) when it shouldn't. I added this as a regression test too

---
Edit:
The issue seems to be more than just this, comparing a `rational` against a `u256` seems to give some weird result (somehow `2**28` as a rational is considered bigger than `uint256()->max()`).

This is fixed once I use `lengthValue->numerator()`. It's safe because there is already a check for `lengthValue->denominator() == 1`.

bshastry: > somehow 2**28 as a rational is considered bigger than uint256()->max()

Thanks for reporting this. Could you please open this as a separate issue, perhaps it has implications elsewhere in the code?

bshastry: > This is fixed once I use lengthValue->numerator(). It's safe because there is already a check for lengthValue->denominator() == 1

Where is this check exactly?

StrongerXi: > > This is fixed once I use lengthValue->numerator(). It's safe because there is already a check for lengthValue->denominator() == 1
> 
> Where is this check exactly?

It's a few "else if" above where I added the new check.

cameel: > somehow `2**28` as a rational is considered bigger than `uint256()->max()`

Sounds like it might be doing a signed comparison - with `uint256()->max()` getting converted to `int256`, giving -1.

StrongerXi: > > somehow 2**28 as a rational is considered bigger than uint256()->max()
> 
> Thanks for reporting this. Could you please open this as a separate issue, perhaps it has implications elsewhere in the code?

Sorry it's a false alarm. It's an error from `ArrayType::validForLocation` which makes sure memory data isn't larger than `2^32`. It doesn't affect the type checker here. Let me know if you want me to use the old way for comparison here.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -309,6 +309,12 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
 			m_errorReporter.typeError(3208_error, length->location(), "Array with fractional length specified.");
 		else if (*lengthValue < 0)
 			m_errorReporter.typeError(3658_error, length->location(), "Array with negative length specified.");
+		else if (lengthValue->numerator() > TypeProvider::uint256()->max())
+			m_errorReporter.typeError(
+				1847_error,
+				length->location(),
+				"Array length too large, maximum is 2^256-1."
+			);
```
cameel: Please add a sanity check here:
```suggestion
		else
		{
			solAssert(lengthValue->denominator() == 1);
			if (lengthValue->numerator() > TypeProvider::uint256()->max())
				m_errorReporter.typeError(
					1847_error,
					length->location(),
					"Array length too large, maximum is 2^256-1."
				);
		}
```
and also include a short comment explaining why the comparison is being done this way (this is not something obvious).

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -309,6 +309,12 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
 			m_errorReporter.typeError(3208_error, length->location(), "Array with fractional length specified.");
 		else if (*lengthValue < 0)
 			m_errorReporter.typeError(3658_error, length->location(), "Array with negative length specified.");
+		else if (lengthValue->numerator() > TypeProvider::uint256()->max())
+			m_errorReporter.typeError(
+				1847_error,
+				length->location(),
+				"Array length too large, maximum is 2^256-1."
```
cameel:
```suggestion
				"Array length too large, maximum is 2**256 - 1."
```

test/libsolidity/syntaxTests/array/length/too_large.sol
```diff
@@ -1,5 +1,8 @@
 contract C {
     uint[8**90] ids;
+    uint[2**256-1] okay;
+    uint[2**256] tooLarge;
```
cameel, line 4: Can you add also `uint[uint(2**256-1)]`? Possibly as a separate test file with appropriate name and some more type conversion cases along with it.

I'm assuming this will still trigger an error due to the `!lengthValue` check.

cameel (changes requested):
Thanks! Looks like a good change. This should make the message less confusing, at least in some cases.

libsolidity/analysis/DeclarationTypeChecker.cpp
cameel: Yes, I'd revert to the old one if it works because it's more straightforward (does not require a sanity check or any special explanations).

libsolidity/analysis/DeclarationTypeChecker.cpp
StrongerXi: Reverting to the comparison so we don't need this assertion or explanation.

libsolidity/analysis/DeclarationTypeChecker.cpp
StrongerXi: Done

test/libsolidity/syntaxTests/array/length/too_large.sol
StrongerXi: Done

leonardoalt:
Could you please rebase over develop? This should take care of the failing tests.

StrongerXi:
> Could you please rebase over develop? This should take care of the failing tests.

Done.

cameel (dismissed):
Thanks! Looks ready to merge.

The only thing I wanted to request was to adjust the name of the new test to be more specific but to avoid adding another review pass I just modified and pushed that myself.

cameel (approved):

PR #12890, opened 2022-04-02
[test] Add location string to syntax tests.

aarlt:
After reviewing #12861 I thought that adding the actual location string would improve the readability of the syntax tests (see https://github.com/ethereum/solidity/pull/12861/files#r840725679).

TODO
- [ ] update `chk_errorcodes` script
- [ ] update `scripts/isolate_tests.py`
- [ ] update all other tests (e.g. also those with `SMTIgnoreOS: macos`)


leonardoalt:
Marked for takeover but not really a prio. Closing the PR, code is still there if someone wants to reuse it.

aarlt:
@leonardoalt just saw that you closed this. I think this would be still very useful, especially for reviewing syntax tests. I know its not a priority, but it could accelerate the review process.

PR #12872, opened 2022-03-28
inheritance.rst: Illustrate the case of abstract derived contract not providing base constructor arguments

aathan:
The prior text left questions open about what this might look like, and how it might be used.

PR #12891, opened 2022-04-02
Clarify try-catch success block (Update control-structures.rst)

aathan:


docs/control-structures.rst
```diff
@@ -840,7 +840,8 @@ that also involves internal function calls), only a revert happening inside the
 call itself. The ``returns`` part (which is optional) that follows declares return variables
 matching the types returned by the external call. In case there was no error,
 these variables are assigned and the contract's execution continues inside the
-first success block. If the end of the success block is reached, execution continues after the ``catch`` blocks.
+first. This block is called the "success block". If the end of the success block
```
ekpyron:
```suggestion
first block of the ``try`` statement. This block is called the "success block". If the end of the success block
```
maybe? Otherwise I'd be wondering the first *what*, when reading this...

cameel:
```suggestion
first. This block is called the *success block*. If the end of the success block
```

cameel:
Something seems to have gone horribly wrong here :) Github says that your branch was renamed to `patch-16`. Which is weird because that was the branch from #12872. Did you maybe rename the wrong branch? It would explain why #12872 cannot be reopened.

PR #12898, opened 2022-04-05
Create split_on_empty_lines_into_numbered_files function in scripts/common.sh

wechman:
This PR is needed to address problems raised in accidentally merged #12851

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -246,3 +246,12 @@ function first_word
 
     echo "$words" | cut -d " " -f 1
 }
+
+# Function reads from stdin. Therefore it has to be used with pipes.
+function split_on_empty_lines_into_numbered_files
+{
+    path_prefix="${1}"
+    path_suffix="${2}"
+
+    awk -v RS= "{print > (\"${path_prefix}_\"NR \"${path_suffix}\")}"
+}
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -310,6 +310,9 @@ function test_solc_assembly_output
 
 function test_via_ir_equivalence()
 {
+    SOLTMPDIR=$(mktemp -d)
+    pushd "$SOLTMPDIR"
+
     (( $# <= 2 )) || fail "This function accepts at most two arguments."
 
     if [[ $2 != --optimize ]] && [[ $2 != "" ]]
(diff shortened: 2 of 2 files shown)


scripts/common.sh
```diff
@@ -246,3 +246,14 @@ function first_word
...
+
+function split_on_empty_lines_into_numbered_files
+{
+    local input=""
+    [[ -p /dev/stdin ]] && input="$(cat -)"
+
+    path_prefix="${1}"
+    path_suffix="${2}"
+
+    echo "${input}" | awk -v RS= "{print > (\"${path_prefix}_\"NR \"${path_suffix}\")}"
```
ekpyron:
```suggestion
    path_prefix="${1}"
    path_suffix="${2}"
    awk -v RS= "{print > (\"${path_prefix}_\"NR \"${path_suffix}\")}"
```
Won't this just work?

wechman: @ekpyron I can see one problem with a suggested solution. It gets stuck in case of an empty stdin. To avoid that we can keep current code or have a failing assertion for an empty stdin. I have no strong feelings on which solution is better.

ekpyron: You mean just plainly using ``split_on_empty_lines_into_numbered_files X Y`` gets stuck? I wouldn't say that's a huge problem. And even ``echo -n "" | split_on_empty_lines_into_numbered_files X Y`` should still work, shouldn't it?

wechman: Yes, plainly used `split_on_empty_lines_into_numbered_files X Y` gets stuck and bash is hanging without any user notification. And, I can see the lack of notification to be a little problem. That's because, it can lead to CI fails with only timeout information, nothing more. However, if you still prefer a simplified version, I can do an update :)
`echo -n "" | split_on_empty_lines_into_numbered_files X Y` works just fine.

ekpyron: I'd prefer the simpler version :-).
I don't think it's an issue if a new PR using this function incorrectly fails CI - it's not like it'll cause random CI failures. You can just add a comment to the function that states that you have to pipe something to it.

wechman: Code updated :)

scripts/common.sh
```diff
@@ -246,3 +246,12 @@ function first_word
 
     echo "$words" | cut -d " " -f 1
 }
+
+# Function reads from stdin. Therefore it can be used with pipes.
```
ekpyron:
```suggestion
# Function reads from stdin. Therefore it has to be used with pipes.
```

ekpyron (dismissed):
Minor comment only; otherwise looks good to me, but @cameel should confirm ;-).

scripts/common.sh
wechman: commits squashed

test/cmdlineTests.sh
```diff
@@ -320,56 +323,52 @@ function test_via_ir_equivalence()
...
     local optimizer_flags=()
     [[ $optimize_flag == "" ]] || optimizer_flags+=("$optimize_flag")
-    [[ $optimize_flag == "" ]] || filename+="_optimize"
+    [[ $optimize_flag == "" ]] || output_file_prefix+="_optimize"
 
     msg_on_error --no-stderr "$SOLC" --ir-optimized --debug-info location "${optimizer_flags[@]}" "$solidity_file" |
-    sed '/^Optimized IR:$/d' |
-    awk -v RS= "{print > (\"${filename}_\"NR \".yul\")}"
+        sed '/^Optimized IR:$/d' |
+        split_on_empty_lines_into_numbered_files $output_file_prefix ".yul"
```
cameel, line 334: Can you add this indent for other `msg_on_error` calls where a multiline command is used as well? This is not the only one.

I generally indent them like this because otherwise they look like separate commands. The only exception is when this stuff is already in a parenthesis (as it was before your previous PR) because then it's not as ambiguous.

test/cmdlineTests.sh
```diff
@@ -310,6 +310,9 @@ function test_solc_assembly_output
 
 function test_via_ir_equivalence()
 {
+    SOLTMPDIR=$(mktemp -d)
+    cd "$SOLTMPDIR"
```
cameel:
```suggestion
    pushd "$SOLTMPDIR"
```
And then `popd` at the end of the function. So that calling the function still preserves your working dir.

cameel (commented):
Two small remarks, otherwise it's perfect.

test/cmdlineTests.sh
wechman: done

test/cmdlineTests.sh
wechman: Good point! Code updated accordingly.

ekpyron (approved):
Reapproving after review fixes.

PR #12905, opened 2022-04-06
Update SolidityParser.g4 add throw statement

stolyarrrov:
At the moment, my team and I are writing a utility for automatically auditing smart contracts. While writing the parser, I checked SolidityParser.g4 and noticed a potential inaccuracy.

### DIFF
--- a/docs/grammar/SolidityParser.g4
+++ b/docs/grammar/SolidityParser.g4
@@ -429,6 +429,7 @@ statement:
 	| breakStatement
 	| tryStatement
 	| returnStatement
+	| throwStatement
 	| emitStatement
 	| revertStatement
 	| assemblyStatement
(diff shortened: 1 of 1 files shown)


ekpyron:
Thanks for the PR and sounds interesting! However, note that ``throw`` statements had long been deprecated and were finally completely disallowed back in 0.5.0.
The antlr grammar in this repo is only meant to parse exactly the syntax of the most recent version of solidity (and - as far as possible - fail for syntax that is invalid in the most recent version of solidity).
So given that, it very much intentionally doesn't contain throw statements anymore.

stolyarrrov:
> Thanks for the PR and sounds interesting! However, note that `throw` statements had long been deprecated and were finally completely disallowed back in 0.5.0. The antlr grammar in this repo is only meant to parse exactly the syntax of the most recent version of solidity (and - as far as possible - fail for syntax that is invalid in the most recent version of solidity). So given that, it very much intentionally doesn't contain throw statements anymore.

Thanks for the clarification, I'll keep that in mind

PR #12907, opened 2022-04-07
Calldatasize-check during ABI encoding from calldata values with nested dynamic arrays.

ekpyron:
Mainly needs reviews on the wording in the bug list, a decision on severity and a blog post.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 ### 0.8.14 (unreleased)
 
+Important Bugfixes:
+ * ABI Encoding: When ABI-encoding values from calldata that contain nested arrays, correctly validate the nested array length against ``calldatasize()`` in all cases.
+
+
 Language Features:
 
 
--- a/docs/bugs.json
+++ b/docs/bugs.json
@@ -1,5 +1,15 @@
 [
     {
+        "uid": "SOL-2022-2",
+        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "summary": "ABI-reencoding of nested dynamic calldata arrays did not always perform proper size checks against the size of calldata and could read beyond ``calldatasize()``.",
+        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may not be performed, when ABI encoding such nested types again directly from calldata. For instance, this can happen, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continued reading values from beyond ``calldatasize()`` (i.e. zero values).",
+        "link": "https://blog.soliditylang.org/2022/05/17/calldata-reencode-size-check-bug/",
+        "introduced": "0.5.8",
+        "fixed": "0.8.14",
+        "severity": "very low"
+    },
+    {
         "uid": "SOL-2022-1",
         "name": "AbiEncodeCallLiteralAsFixedBytesBug",
         "summary": "Literals used for a fixed length bytes parameter in ``abi.encodeCall`` were encoded incorrectly.",
--- a/test/libsolidity/semanticTests/abiEncoderV2/calldata_array_short.sol
+++ b/test/libsolidity/semanticTests/abiEncoderV2/calldata_array_short.sol
@@ -1,0 +1,12 @@
+pragma abicoder v2;
+
+contract C {
+	function f(uint[] calldata) public {}
+}
+// ====
+// revertStrings: debug
+// compileViaYul: also
+// ----
+// f(uint256[]): 0x20, 0 ->
+// f(uint256[]): 0x20, 1 -> FAILURE, hex"08c379a0", 0x20, 0x2b, "ABI decoding: invalid calldata a", "rray stride"
+// f(uint256[]): 0x20, 2 -> FAILURE, hex"08c379a0", 0x20, 0x2b, "ABI decoding: invalid calldata a", "rray stride"
--- a/test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
+++ b/test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
@@ -14,10 +14,10 @@ contract C {
 // revertStrings: debug
 // ----
 // h(uint256[][]): 0x20, 1, 0x20, 0 ->
-// h(uint256[][]): 0x20, 1, 0x20, 1 ->
-// h(uint256[][]): 0x20, 1, 0x20, 2 ->
+// h(uint256[][]): 0x20, 1, 0x20, 1 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
+// h(uint256[][]): 0x20, 1, 0x20, 2 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
 // h(uint256[][]): 0x20, 1, 0x20, 3 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
 // f((uint256[])): 0x20, 0x20, 0 ->
-// f((uint256[])): 0x20, 0x20, 1 ->
-// f((uint256[])): 0x20, 0x20, 2 ->
+// f((uint256[])): 0x20, 0x20, 1 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
+// f((uint256[])): 0x20, 0x20, 2 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
 // f((uint256[])): 0x20, 0x20, 3 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
--- a/test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
+++ b/test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
@@ -1,0 +1,23 @@
+pragma abicoder v2;
+
+contract C {
+	function h(uint[][] calldata a) public {
+		abi.encode(a);
+	}
+	struct S { uint[] x; }
+	function f(S calldata a) public {
+		abi.encode(a);
+	}
+}
+// ====
+// compileViaYul: also
+// revertStrings: debug
+// ----
+// h(uint256[][]): 0x20, 1, 0x20, 0 ->
+// h(uint256[][]): 0x20, 1, 0x20, 1 ->
+// h(uint256[][]): 0x20, 1, 0x20, 2 ->
+// h(uint256[][]): 0x20, 1, 0x20, 3 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
+// f((uint256[])): 0x20, 0x20, 0 ->
+// f((uint256[])): 0x20, 0x20, 1 ->
+// f((uint256[])): 0x20, 0x20, 2 ->
+// f((uint256[])): 0x20, 0x20, 3 -> FAILURE, hex"08c379a0", 0x20, 0x1e, "Invalid calldata access stride"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -990,7 +990,7 @@ jobs:
       - gitter_notify_failure_unless_pr
 
   t_ubu_soltest_all: &t_ubu_soltest_all
-    <<: *base_ubuntu2004
+    <<: *base_ubuntu2004_large
     parallelism: 50
     <<: *steps_soltest_all
 
--- a/docs/bugs_by_version.json
+++ b/docs/bugs_by_version.json
@@ -1036,6 +1036,7 @@
     },
     "0.5.10": {
         "bugs": [
+            "NestedCallataArrayAbiReencodingSizeValidation",
             "ABIDecodeTwoDimensionalArrayMemory",
             "KeccakCaching",
             "EmptyByteArrayCopy",
@@ -1051,6 +1052,7 @@
     },
     "0.5.11": {
         "bugs": [
+            "NestedCallataArrayAbiReencodingSizeValidation",
             "ABIDecodeTwoDimensionalArrayMemory",
             "KeccakCaching",
             "EmptyByteArrayCopy",
@@ -1065,6 +1067,7 @@
     },
     "0.5.12": {
         "bugs": [
+            "NestedCallataArrayAbiReencodingSizeValidation",
             "ABIDecodeTwoDimensionalArrayMemory",
             "KeccakCaching",
             "EmptyByteArrayCopy",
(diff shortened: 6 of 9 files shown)


test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
```diff
@@ -0,0 +1,14 @@
+pragma abicoder v2;
+
+contract C {
+        function h(uint[][] calldata a) public {
+                abi.encode(a);
+        }
+}
```
cameel:
```suggestion
contract C {
    function h(uint[][] calldata a) public {
        abi.encode(a);
    }
}
```

Changelog.md
```diff
@@ -1,5 +1,9 @@
 ### 0.8.14 (unreleased)
 
+Important Bugfixes:
+ * ABI Encoding: When ABI-encoding values in calldata that contain nested arrays, correctly validate the nested array length against ``calldatasize()`` in all cases.
```
cameel: Does it really affect only nested arrays? From the code it looks like it would affect 1D arrays as well. Is `calldataAccessFunction()` only called for nested arrays or something?

test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
```diff
@@ -0,0 +1,14 @@
+pragma abicoder v2;
+
+contract C {
+        function h(uint[][] calldata a) public {
```
cameel: Can you add a case with a 1D array as well? Looking through the test dir I don't see a case covering that.

cameel (commented):
The fix looks correct.

Changelog.md
ekpyron: ``calldataAccessFunction`` is only used with the base type of an array or the member of a struct - and the bug is only in the part where that type is dynamically sized.
So
```
contract C {
  struct S { uint[] x; }
  function f(S calldata a) public { abi.encode(a); }
}
```
also has it - but I'd also have called that a "nested array". I can test that explicitly as well, though.
I can also add tests for one-dimensional arrays, but they don't have the bug.

test/libsolidity/semanticTests/abiEncoderV2/calldata_nested_array_reencode.sol
ekpyron: One-dimensional arrays are immediately validated during the initial abi decoding, so that'll revert even for an empty function body - I'd think (hope :-)) there's some tests for that somewhere, but I'll add another just to be sure.

test/libsolidity/semanticTests/abiEncoderV2/calldata_array_short.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
+	function f(uint[] calldata) public {}
+}
+// ====
+// revertStrings: debug
+// compileViaYul: also
+// ----
+// f(uint256[]): 0x20, 0 ->
+// f(uint256[]): 0x20, 1 -> FAILURE, hex"08c379a0", 0x20, 0x2b, "ABI decoding: invalid calldata a", "rray stride"
+// f(uint256[]): 0x20, 2 -> FAILURE, hex"08c379a0", 0x20, 0x2b, "ABI decoding: invalid calldata a", "rray stride"
```
ekpyron, line 12: Not sure if we should change all those revert reasons to something better while we're at it :-)... but it may unnecessarily confuse tooling that expects them like this...

Changelog.md
cameel: ok, would be good to have that as a test case.

And yeah, I'd still add cases for 1D arrays just for better coverage. It also makes it easy to verify during review that this case is really bug-free.

docs/bugs.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "uid": "SOL-2022-2",
+        "name": "AbiEncodeFromCalldataNestedArraySizeValidationBug",
```
cameel: My naming suggestion from the chat:
```suggestion
        "name": "AbiEncodeNestedCalldataArraySizeValidation",
```

ekpyron: I went for the later suggestion of ``NestedCallataArrayAbiReencodingSizeValidation`` for now. Open to changing it further.

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
...
+        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "summary": "ABI-encoding values in calldata that contain nested dynamic arrays sometimes read beyond ``calldatasize()`` instead of reverting.",
+        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may fail, when ABI encoding such nested types again directly from calldata. For instance, this can, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continue reading values from beyond ``calldatasize()`` (i.e. zero values).",
+        "link": "TODO",
+        "introduced": "0.5.8",
+        "fixed": "0.8.14",
+        "severity": "very low"
```
ekpyron, line 10: Any opinions on the severity? Is ``very low`` an underestimation?
In most cases, I'd expect little harm to come from reading zeroes from calldata, but it may be worth thinking about whether weird cases can happen...
Also worth a thought whether the inconsistency in the revert behaviour depending on directly accessing or re-encoding calldata has any serious implications for proxy patterns or the like.

ekpyron: On the other hand one can also argue for the direct effect of the bug, that this (in certain cases) can undermine the safety guard against transactions with corrupted calldata... based on that one could also go as high as ``medium``...

ekpyron:
Created https://github.com/ethereum/solidity/issues/12922 for increasing test coverage for these cases in order not to stall the bug fix.

docs/bugs.json
ekpyron: @chriseth in channel: "I think it's very low"

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "uid": "SOL-2022-2",
+        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "summary": "ABI-encoding values in calldata that contain nested dynamic arrays sometimes read beyond ``calldatasize()`` instead of reverting.",
+        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may fail, when ABI encoding such nested types again directly from calldata. For instance, this can, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continue reading values from beyond ``calldatasize()`` (i.e. zero values).",
```
```suggestion
        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may not fail, when ABI encoding such nested types again directly from calldata. For instance, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continued reading values from beyond ``calldatasize()`` (i.e. zero values).",
```

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "uid": "SOL-2022-2",
+        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "summary": "ABI-encoding values in calldata that contain nested dynamic arrays sometimes read beyond ``calldatasize()`` instead of reverting.",
```
```suggestion
        "summary": "ABI-encoding values in calldata that contain nested dynamic arrays sometimes reads beyond ``calldatasize()`` instead of reverting.",
```

cameel: Or how about this?
```suggestion
        "summary": "ABI-reencoding of nested dynamic calldata arrays may not detect invalid size and read beyond array bounds.",
```
It may happen at any point inside, crossing into another nested array. It's not just about reading past calldata, right?

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
...
+        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may fail, when ABI encoding such nested types again directly from calldata. For instance, this can, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continue reading values from beyond ``calldatasize()`` (i.e. zero values).",
+        "link": "TODO",
```
cameel: Should this be kept open until we have a link or will we update it in a separate PR?

cameel (commented):
Some tweaks to bug wording.

Other than that it seems fine, though I think it would not hurt to add some more manually created test cases.

docs/bugs.json
ekpyron: ``read`` was meant to be past tense :-). (the summaries generally seem to be formulated in past tense)
``"ABI-reencoding of nested dynamic calldata arrays did not always perform proper size checks against the size of calldata and read beyond ``calldatasize``."``

It's actually not the array bounds that are violated - that'll fortunately never happen, it'd be much more serious. It's just that the array may not fit into the explicitly passed calldata entirely.

docs/bugs.json
ekpyron: Changed ``this can`` to ``this can happen`` and applied the ``continued`` check.

With regards to ``this check may *not* fail``: in my mind, we have a check that is supposed to revert in these cases, and this check failing means, it does not revert, i.e. ``this check may fail`` seems right to me... but it's indeed a bit ambiguous, so maybe I'll try to rephrase generally...

cameel: Makes sense but the way it's written it's really a bit ambiguous and could be interpreted either way. How about clarifying it like this?

> While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may **be performed incorrectly**, when ABI encoding such nested types again directly from calldata.

docs/bugs.json
cameel: > It's actually not the array bounds that are violated - that'll fortunately never happen, it'd be much more serious. It's just that the array may not fit into the explicitly passed calldata entirely.

Right. I incorrectly remembered that we're checking if length fits in the available space but we're actually only checking if it does not overflow past the end of calldata. So the description does match the code.

But in that case - do we even validate that array items do not overlap other items or go past the space allocated for the whole array? I.e. what if the space allocated for an item as indicated by head offsets is too short compared to item lengths?

cameel: > `read` was meant to be past tense :-).

Fair enough. Looks like I just read it wrong :)

ekpyron: > But in that case - do we even validate that array items do not overlap other items or go past the space allocated for the whole array? I.e. what if the space allocated for an item as indicated by head offsets is too short compared to item lengths?

- do we even validate that array items do not overlap other items?

Not sure what you mean exactly, but generally encodings with overlapping arrays are valid, i.e.
```
contract C { function f(uint[] calldata x, uint[] calldata y) external { ... } }
// ----
// f(uint256[],uint256[]): 0x40, 0x40, 3, 1, 2, 3
```
is just fine and passes the very same array as ``x`` and ``y``.

- go past the space allocated for the whole array?

Also not sure what you mean, but there is no such thing as "space allocated for the array" - i.e. the following is also perfectly fine:
```
contract C { function f(uint[] calldata x, uint[] calldata y) external { ... } }
// ----
// f(uint256[],uint256[]): 0x80, 0x40, 3, 1, 2, 3, 4
```
which will pass ``[3, 4]`` as ``x`` and ``[1, 2, 3]`` as ``y``.
(Yes, ABI-encoding is weird ;-))

cameel: >  generally encodings with overlapping arrays are valid

Oh, that's interesting to know. I expected these checks to be much stricter because otherwise you can have a situations similarly "bad" to what we have with this bug, just without going past the calldata. I.e. if the client computes the length incorrectly and it's too long, the decoder will just grab the data belonging to another item. Just like here you can have wrong length and grab extra zeros from the empty space after calldata and here we consider it a bug - even though the result is actually more predictable and potentially even safer. I could actually see someone trying to be clever and using it intentionally to avoid paying for a bunch of trailing zeros :)

> Also not sure what you mean, but there is no such thing as "space allocated for the array"

Not sure how to call it properly but I was thinking about the space between the offset of your item and the offset of the next one. But you're right, the offsets do not even have to increase monotonically and very weird cases are possible.

It's just a bit odd that we care about not going beyond the end of calldata if we do not care about restricting these weird cases. Re-encoding them runs into the same potential problem with proxies you mentioned on the chat where data encoded again will be different from what you get as input.

chriseth (dismissed):

chriseth:
(still need squashing)

docs/bugs.json
ekpyron: > It's just a bit odd that we care about not going beyond the end of calldata if we do not care about restricting these weird cases. Re-encoding them runs into the same potential problem with proxies you mentioned on the chat where data encoded again will be different from what you get as input.

Well, hence "very low" severity :-).

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
...
+        "summary": "ABI-reencoding of nested dynamic calldata arrays did not always perform proper size checks against the size of calldata and could read beyond ``calldatasize()``.",
+        "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may fail, when ABI encoding such nested types again directly from calldata. For instance, this can happen, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continued reading values from beyond ``calldatasize()`` (i.e. zero values).",
+        "link": "https://blog.soliditylang.org/2022/05/17/calldata-reencode-size-check-bug/",
```
ekpyron, line 7: We'll need to double-check that this will be the final link before merging.

Changelog.md
chriseth: maybe "from calldata"?

chriseth (dismissed):

chriseth (dismissed):

ekpyron:
@chriseth I had to restrict the latest test case to abiencoder-v2-only with the pragma - no other changes.

ekpyron:
I added another commit increasing the CircleCI resource size, since it was hitting OOM here as well now.

chriseth (approved):

cameel (approved):
~The PR itself looks fine but the [blog post](https://github.com/ethereum/solidity-blog/pull/173) says severity is low and that the bug was introduced in 0.6.9. So one of them needs adjustment.~

EDIT: My bad, that's the blog post about the other bug :)

PR #12908, opened 2022-04-07
Moved `docs.sh` script to `docs/`

### LINKED ISSUE #12906: Move `docs.sh` script to `docs/`
Our translation repositories in [solidity-docs ](https://github.com/solidity-docs) only preserve the `docs/` directory from the main repository. The problem is that the script we use to build the documentation is currently under [`scripts/docs.sh`](https://github.com/ethereum/solidity/blob/develop/scripts/docs.sh) so it cannot be easily used by translators to render the docs locally. It would be better to have this script in `docs/`.

- [ ] Move the script.
- [ ] Adjust CI and paths for the new location.

tharun634:
- resolves #12906


cameel (approved):
Thanks! Looks good.

I'm actually surprised that you did not need any changes inside the script. I see that it's because we always run it from the repo root dir in CI so relative paths still work. I added a commit to make the script independent of the working dir.

I also updated your commit description - please do not include the issue number there. It adds a lot of noise in issues when commits are rebased.

I'm approving but not merging yet - I'm the one who reported the issue so it would be good for someone else from the team to at least take a look and say if there might be any reasons not to move the script.

PR #12909, opened 2022-04-07
Add `.DS_Store` and `Thumbs.db` to `.gitignore`

bytecurl:
Added OS specific local files .DS_Store (macOS) and Thumbs.db (Windows) that often sneak their way in when not explicitly included in the .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,7 @@ CMakeLists.txt.user
 
 # place to put local temporary files
 tmp
+
+# OS specific local files
+.DS_Store
+Thumbs.db


cameel (dismissed):
Thanks!

cameel:
Ah, wait. There are some weird merge commits in your branch. Can you rebase on `develop` and squash them?

A more concrete commit description would also be nice - i.e. in what specifically got changed about `.gitignore`. Only saying that it updates `.gitignore` is a bit useless because that's obvious from the file list.

cameel (approved):
Rebased and updated commit description.

PR #12751, opened 2022-03-07
updates to gas section

andreb0x:
small modifications for clarity

PR #12910, opened 2022-04-07
Some updates to gas and delegatecall.

chriseth:
Follow-up to https://github.com/ethereum/solidity/pull/12751

### DIFF
--- a/docs/introduction-to-smart-contracts.rst
+++ b/docs/introduction-to-smart-contracts.rst
@@ -396,26 +396,33 @@ returns that code when executed.
 Gas
 ===
 
-Upon creation, each transaction is charged with a certain amount of **gas**.
-Imposing a cost on transactions serves to secure the network:
-it deincentivizes bad actors from spamming, and pays validators (i.e. miners and stakers)
-for the work that is required to validate new blocks. While the EVM executes the
+Upon creation, each transaction is charged with a certain amount of **gas**
+that has to be paid for by the originator of the transaction (``tx.origin``).
+While the EVM executes the
 transaction, the gas is gradually depleted according to specific rules.
-
-The **gas price** is a value set by the originator of the transaction, i.e. `tx.origin`, who
-has to pay ``gas_price * gas`` up front from the sending account.
-If some gas is left after execution, it is refunded to the transaction originator.
-
 If the gas is used up at any point (i.e. it would be negative),
-an out-of-gas exception is triggered, which reverts all modifications
+an out-of-gas exception is triggered, which ends execution and reverts all modifications
 made to the state in the current call frame.
+
+This mechanism incentivizes economical use of EVM execution time
+and also compensates EVM executors (i.e. miners / stakers) for their work.
+Since each block has a maximum amount of gas, it also limits the amount
+of work needed to validate a block.
+
+The **gas price** is a value set by the originator of the transaction, who
+has to pay ``gas_price * gas`` up front to the EVM executor.
+If some gas is left after execution, it is refunded to the transaction originator.
+In case of an exception that reverts changes, already used up gas is not refunded.
+
+Since EVM executors can choose to include a transaction or not,
+transaction senders cannot abuse the system by setting a low gas price.
 
 .. index:: ! storage, ! memory, ! stack
 
 Storage, Memory and the Stack
 =============================
 
-The Ethereum Virtual Machine has three areas where it can store data-
+The Ethereum Virtual Machine has three areas where it can store data:
 storage, memory and the stack.
 
 Each account has a data area called **storage**, which is persistent between function calls
@@ -504,10 +511,8 @@ Delegatecall / Callcode and Libraries
 
 There exists a special variant of a message call, named **delegatecall**
 which is identical to a message call apart from the fact that
-the code at the target address is executed in the context of the calling
-contract. In practice this means ``msg.sender`` and ``msg.value`` retain
-the values that were passed to the delegator, while executing code that
-lives in the delegated contract.
+the code at the target address is executed in the context (i.e. at the address) of the calling
+contract and ``msg.sender`` and ``msg.value`` do not change their values.
 
 This means that a contract can dynamically load code from a different
 address at runtime. Storage, current address and balance still


ekpyron (dismissed):

docs/introduction-to-smart-contracts.rst
```diff
@@ -504,10 +511,8 @@ Delegatecall / Callcode and Libraries
 
 There exists a special variant of a message call, named **delegatecall**
 which is identical to a message call apart from the fact that
-the code at the target address is executed in the context of the calling
-contract. In practice this means ``msg.sender`` and ``msg.value`` retain
-the values that were passed to the delegator, while executing code that
-lives in the delegated contract.
+the code at the target address is executed in the context (i.e. at the address) of the calling
+contract and ``msg.sender`` and ``msg.value`` do not change thair values.
```
cameel:
```suggestion
contract and ``msg.sender`` and ``msg.value`` do not change their values.
```

cameel: Since this is an intro, might be worth adding a few words to clarify the consequences:
```suggestion
contract and ``msg.sender`` and ``msg.value`` retain the values they have in that context,
effectively letting the callee impersonate the calling contract and freely modify its state.
```

docs/introduction-to-smart-contracts.rst
```diff
@@ -396,26 +396,33 @@ returns that code when executed.
...
-Imposing a cost on transactions serves to secure the network:
-it deincentivizes bad actors from spamming, and pays validators (i.e. miners and stakers)
-for the work that is required to validate new blocks. While the EVM executes the
+Upon creation, each transaction is charged with a certain amount of **gas**
+that has to be paid for by the originator of the transaction (``tx.origin``).
+While the EVM executes the
 transaction, the gas is gradually depleted according to specific rules.
+If the gas is used up at any point (i.e. it would be negative),
+an out-of-gas exception is triggered, which ends execution and reverts all modifications
+made to the state in the current call frame.
```
cameel: I wanted to comment that it's not just the current call frame but the whole transaction since you cannot silence out-of-gas. Then I realized the caller could forward less gas and handle the revert. Maybe it would be a good idea to mention that the caller can decide how much gas he forwards?

cameel (dismissed):
Looks good. Found just one typo.

I also have two suggestions but they're not all that important.

docs/introduction-to-smart-contracts.rst
cameel: Looks like `chk_coding_style` failed because of trailing whitespace here:
```suggestion
made to the state in the current call frame.
```

docs/introduction-to-smart-contracts.rst
ekpyron: I must be going blind...

cameel: @ekpyron What do you think about this one? Once we resolve this, I could fix the two typos, merge and we'd be one PR lighter :)

ekpyron: Hm... "context" is a bit underdefined for that :-). Earlier we say by "context" we mean "at the address of the calling
contract"... but `msg.sender` and `msg.value` don't have values "at that address"... they retain the values they had at the call site...

And in general the sentence is getting too long... I'd rather keep the sentence as in the PR now - and we can think about adding a new sentence about "effectively allowing to impersonate"... on the other hand the next paragraph also kind of says that...

docs/introduction-to-smart-contracts.rst
```diff
@@ -504,10 +511,8 @@ Delegatecall / Callcode and Libraries
...
 which is identical to a message call apart from the fact that
-the code at the target address is executed in the context of the calling
-contract. In practice this means ``msg.sender`` and ``msg.value`` retain
-the values that were passed to the delegator, while executing code that
-lives in the delegated contract.
+the code at the target address is executed in the context (i.e. at the address) of the calling
+contract and ``msg.sender`` and ``msg.value`` do not change thair values.
 
 This means that a contract can dynamically load code from a different
 address at runtime. Storage, current address and balance still
```
ekpyron, line 518:
```suggestion
This means that a contract can dynamically load code from a different
address at runtime and execute that code as if it was one of its internal functions. Storage, current address and balance still
```

Maybe that's a good way to clarify what you wanted to have in https://github.com/ethereum/solidity/pull/12910/files#r845484979 ? Executing the foreign code as if it was an internal function is basically the opposite way of putting the called code impersonating the calling contract, but to me it would seem easier to understand what it means...

docs/introduction-to-smart-contracts.rst
cameel: > on the other hand the next paragraph also kind of says that...

A, right. Didn't look too far ahead :) OK, let's drop this suggestion in that case.

docs/introduction-to-smart-contracts.rst
cameel: Sounds good to me. I'll add that to the PR then.

ekpyron: On the other hand "as if it was one of its internal functions" might read like the called contract would somehow magically get access to the symbols in the calling contracts scope... but maybe I'm just slowly going mad over thinking about docs ;-).

cameel: Maybe. I thought it still gives you a better picture of what will happen.

But yeah, could be misleading. Let's just merge it as is with only typos fixes.

docs/introduction-to-smart-contracts.rst
cameel: Fixed.

ekpyron (approved):
Looks good to me (with or without https://github.com/ethereum/solidity/pull/12910#discussion_r845491118 ;-)).

cameel:
Just pushed typo fixes and left the rest of the text as is.

docs/introduction-to-smart-contracts.rst
ekpyron: Yeah, I'm not sure anymore, maybe I'm starting to overthink what stuff might imply or not ;-). I'm fine with adding this and with not adding this.

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12923, opened 2022-04-11
[Circle CI] Parameterize b_bytecode_ubu and b_bytecode_osx build job

super1ha1:


.circleci/config.yml
```diff
@@ -81,6 +81,29 @@ commands:
           event: success
           condition: on_success
 
+  run_test_byte_code:
+    description: "Run test byte code on a environment."
```
cameel: Let's call this command `prepare_bytecode_report`.
```suggestion
  prepare_bytecode_report:
    description: "Generate bytecode report and upload it as an artifact."
```

.circleci/config.yml
```diff
@@ -81,6 +81,29 @@ commands:
...
+    parameters:
+      os_type:
+        type: enum
+        enum: ["ubuntu", "osx"]
```
cameel: This does not really control which OS this runs on, only file naming, so I think it would be better to just call this `name` or `label`.
```suggestion
      label:
        type: string
```

.circleci/config.yml
```diff
@@ -81,6 +81,29 @@ commands:
...
+    steps:
+      - run: mkdir test-cases/
+      - run: cd test-cases && ../scripts/isolate_tests.py ../test/
+      - run: cd test-cases && ../scripts/bytecodecompare/prepare_report.py ../build/solc/solc --interface standard-json --report-file ../bytecode-report-${<< parameters.os_type >>}-json.txt
```
cameel: Anything with bash variables in it should be quoted just in case.
```suggestion
      - run: cd test-cases && ../scripts/bytecodecompare/prepare_report.py ../build/solc/solc --interface standard-json --report-file "../bytecode-report-${<< parameters.os_type >>}-json.txt"
```

.circleci/config.yml
```diff
@@ -1297,41 +1320,17 @@ jobs:
...
-          root: .
-          paths:
-            - bytecode-report-osx-json.txt
-            - bytecode-report-osx-cli.txt
-      - gitter_notify_failure_unless_pr
+      - run_test_byte_code:
+          os_type: "osx"
 
   b_bytecode_win:
     <<: *base_win_cmd
```
cameel, line 1426: Now that I think of it, we could have the command handle windows and emscripten as well, using `when` blocks. This way we'd be able to share at least the part that persists and workspace and artifacts. The Python commands might actually work too given that on Windows the `/` separators are supported.

Could you try that?

cameel:
The `b_ubu` job seems to be failing due to resource constraints. CircleCI runs it on your account with less memory than it needs and the process gets killed.

This seems to be because CircleCI runs it on your account rather than our org account. This seems to often happen with external PRs. Maybe you can try to log into CircleCI with your github account and unfollow your fork? See also [Why aren't pull requests triggering jobs on my organization?](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-). If that does not help, in the end we can always retrigger CI manually, but it would be more convenient if just worked for you.

.circleci/config.yml
super1ha1: Hi @cameel yes, I think we can further share some part of the b_bytecode_win and b_bytecode_ems

yet currently since these job are quite different versus each other,

so if we split to smaller steps, would that affect the code readability?

super1ha1:
> 

Thanks @cameel for your valuable comment, I have updated the pull request and unfollow my fork, how can we trigger the build again?

cameel:
oh well... I triggered the build by making a branch in on the same commit in the main repo. All checks should complete now.

.circleci/config.yml
cameel: It would but duplication is not great either. Especially when the duplicated bits are no longer next to each other.

I mean, I'll still accept the PR even without this (especially given that you're having problems triggering the CI so it might be hard to get right) but I think I think that it would be an improvement.

cameel:
@super1ha1 Are you still interested in working on this?

leonardoalt:
@cameel do you think this can be merged soon? Otherwise I'd close it

cameel (dismissed):
@leonardoalt Well, it should be really generalized further to handle Windows and Emscripten in the same command.

But that's also not something we want to put much effort into right now and as is it already makes things better so we could merge in the current form.

I rebased and fixed one problem I saw (there should be no `${}` around parameters). If it works, we can squash and merge.

cameel:
@nishant-sachdeva Can you take it over? It only requires squashing commits and then we can merge.

cameel:
`Fix spelling mistakes and CI spellcheck job`? :) Something went wrong here...

cameel:
BTW, when squashing please make sure to preserve the author of the original commit for later attribution in the release notes. I see that you got @Marenz for some reason even though I think he had nothing to do with this PR :)

cameel (dismissed):

cameel (approved):

Issue #12925, opened 2022-04-11
CircleCI dependency caching broken on macOS


cameel:
Restoring cache in macOS jobs takes ~17 minutes. See for example [recent `b_osx` run on `develop`](https://app.circleci.com/pipelines/github/ethereum/solidity/23649/workflows/01a8fe85-debd-4892-874c-2745e02c10cc/jobs/1038415). There are lots of errors about symlinks being skipped, which might have something to do with it:
```
...
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/env.rb" - open /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/env.rb: permission denied
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/environment_preserver.rb" - open /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/environment_preserver.rb: permission denied
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler
...
```

For now we have disabled the cache because just installing the dependencies is quicker than that (4 min). Still, the cache used to save us some time so we need to investigate and fix the problem. Might be a good idea to report it to CircleCI.

cameel:
For the record, @r0qs tried re-enabling the cache and it still does not work the way we used to do. I think we need a different approach, and that would be, broadly speaking, caching the things we download, not the things that get installed. I think that the download is the slow part and the installation is pretty quick. That should also be more robust against breakage like this, because we would not have to make any assumptions about the contents of the installation dirs (especially that we can just replace it wholesale).

Also, there are two distinct parts to this - caching Homebrew packages and caching the other libs we install in `osx_install_dependencies.sh` (i.e. evmone and z3). Each could be done independently and the libs are much less complicated to handle than Homebrew.

See [the comment thread in the PR](https://github.com/ethereum/solidity/pull/14485#issuecomment-1689819247) for detailed discussion of this approach and the problems we encountered so far.

PR #12924, opened 2022-04-11
Invalidate dependency cache on macOS in CI

cameel:
Restoring cache in macOS jobs takes ~17 minutes. See for example [recent `b_osx` run on `develop`](https://app.circleci.com/pipelines/github/ethereum/solidity/23649/workflows/01a8fe85-debd-4892-874c-2745e02c10cc/jobs/1038415). There are lots of errors about symlinks being skipped, which might have something to do with it:
```
...
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/env.rb" - open /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/env.rb: permission denied
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/environment_preserver.rb" - open /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler/environment_preserver.rb: permission denied
Skipping writing "usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.8/lib/ruby/2.6.0/bundler
...
```

I think that the Homebrew in CircleCI's image must have been updated and the errors might be because we're restoring our cached, older version. If that's the case, cache invalidation should help.

PR #12926, opened 2022-04-11
Disable dependency caching on macOS in CI

cameel:
Replaces #12924.

Looks like invalidating the cache does not fix the problem so I added an issue to figure it out later (#12925) and here I'm disabling the cache for now as a temporary workaround.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -246,23 +246,11 @@ defaults:
 
   - steps_install_dependencies_osx: &steps_install_dependencies_osx
       steps:
-        - restore_cache:
-            keys:
-              - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
-        # DO NOT EDIT between here and save_cache, but rather edit ./circleci/osx_install_dependencies.sh
-        # WARNING! If you do edit anything here instead, remember to invalidate the cache manually.
+        # FIXME: We used to cache dependencies on macOS but now it takes longer than just installing
+        # them each time. See https://github.com/ethereum/solidity/issues/12925.
         - run:
             name: Install build dependencies
             command: ./.circleci/osx_install_dependencies.sh
-        - save_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -900,23 +900,25 @@ jobs:
...
-            - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
-      # DO NOT EDIT between here and save_cache, but rather edit ./circleci/osx_install_dependencies.sh
-      # WARNING! If you do edit anything here instead, remember to invalidate the cache manually.
+      # FIXME: Dependency caching disabled on macOS because it takes longer than just installing them each time.
+      # See https://github.com/ethereum/solidity/issues/12925
+      #- restore_cache:
+      #    keys:
+      #      - dependencies-osx-{{ arch }}-{{ checksum ".circleci/osx_install_dependencies.sh" }}
+      ## DO NOT EDIT between here and save_cache, but rather edit ./circleci/osx_install_dependencies.sh
+      ## WARNING! If you do edit anything here instead, remember to invalidate the cache manually.
```
cameel: Is it better to leave it commented out like this so that it's easy to restore or should I just delete it?

ekpyron: It's nothing really special - and we still have it in the git history... I'd just remove it, but also fine commenting it.

cameel:
It seemed that disabling caching revealed some breakage:
```
./.circleci/soltest.sh: line 39: realpath: command not found
```

but after looking into it, it's just that the `t_osx` jobs assumed that `b_osx` will always create cache that can be restored. Now that it's gone, they don't use cache but don't install these dependencies either.

To fix that I made all osx jobs consistently use `steps_install_dependencies_osx`. This also makes disabling easier - it needs to be done just in one place.

.circleci/config.yml
cameel: ok. Removed commented-out code and left only the FIXME.

ekpyron (approved):
Looks good if tests pass.

PR #12928, opened 2022-04-12
Propagate purity information for member access to foreign pure variables

### LINKED ISSUE #12927: Should support referring to foreign-constants as compile-time constants
`A.sol`:
```
import "./B.sol";

contract A {
    uint256 private constant LIMIT = B.MAX * B.SCALE;
}
```

`B.sol`:
```
library B {
    uint256 internal constant MAX = 1;
    uint256 internal constant SCALE = 100;
}
```

Results in:
```
TypeError: Initial value for constant variable has to be compile-time constant.
 --> contracts/A.sol:4:38:
  |
4 |     uint256 private constant LIMIT = B.MAX * B.SCALE;
  |                                      ^^^^^^^^^^^^^^^
```

The problem is not the multiplication, but rather that `B.MAX` or `B.SCALE` is not considered compile-time constant.

StrongerXi:
Fixes #12927.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@ Language Features:
 
 
 Compiler Features:
-
+ * TypeChecker: Support using library constants in initializers of other constants.
 
 Bugfixes:
 
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -3214,6 +3214,13 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
 
 			annotation.isPure = isPure;
 		}
+	if (
+		auto const* varDecl = dynamic_cast<VariableDeclaration const*>(annotation.referencedDeclaration);
+		!annotation.isPure.set() &&
+		varDecl &&
+		varDecl->isConstant()
+	)
+		annotation.isPure = true;
 
 	if (auto magicType = dynamic_cast<MagicType const*>(exprType))
 	{
--- a/test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_constants.sol
+++ b/test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_constants.sol
@@ -1,0 +1,25 @@
+uint256 constant MAX = 1;
+
+library L1 {
+    uint256 internal constant INT = 100;
+}
+
+contract C1 {
+    uint256 internal constant CONST = 20 + L2.INT; // forward reference
+    uint256 internal constant LIMIT = MAX * L1.INT;  // same file & external library constant
+    uint256 internal constant NESTED = LIMIT + CONST; // nested & same contract constant
+}
+
+contract C2 is C1 {
+    uint256 internal constant INHERITED = NESTED + CONST * LIMIT; // inherited constants
+}
+
+contract C3 is C2 {
+    uint256 internal constant NESTED_INHERITED = INHERITED + NESTED + CONST * LIMIT; // nest-inherited constants
+}
+
+library L2 {
+    uint256 internal constant INT = 100;
+}
+
+// ----
--- a/test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
+++ b/test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
@@ -1,0 +1,13 @@
+==== Source: A.sol ====
+import "B.sol";
+
+uint256 constant A = B + 1;
+
+==== Source: B.sol ====
+import "A.sol";
+
+uint256 constant B = A + 1;
+
+// ----
+// TypeError 6161: (B.sol:17-43): The value of the constant B has a cyclic dependency via A.
+// TypeError 6161: (A.sol:17-43): The value of the constant A has a cyclic dependency via B.
--- a/test/libsolidity/syntaxTests/constants/constant_variables_as_static_array_length.sol
+++ b/test/libsolidity/syntaxTests/constants/constant_variables_as_static_array_length.sol
@@ -1,0 +1,29 @@
+uint256 constant MAX = 1;
+
+library L1 {
+    uint256 internal constant INT = 100;
+}
+
+contract C1 {
+    uint256 internal constant CONST1 = L1.INT;
+
+    uint256[L1.INT] internal arr1; // error, backward reference
+    uint256[L2.INT] internal arr2; // error, forward reference
+}
+
+contract C2 is C1 {
+    uint256 internal constant CONST2 = CONST1;
+
+    uint256[CONST1] internal arr3; // error, inherited constants
+    uint256[CONST2] internal arr4; // error, same contract constant
+}
+
+library L2 {
+    uint256 internal constant INT = 100;
+}
+
+// ----
+// TypeError 5462: (158-164): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (222-228): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (356-362): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (421-427): Invalid array length, expected integer literal or constant expression.
--- a/test/libsolidity/syntaxTests/constants/constant_with_dependencies_as_array_sizes.sol
+++ b/test/libsolidity/syntaxTests/constants/constant_with_dependencies_as_array_sizes.sol
@@ -1,0 +1,35 @@
+uint256 constant MAX = 1;
+
+library L1 {
+    uint256 internal constant INT = 100;
+}
+
+contract C1 {
+    uint256 internal constant CONST = 20 + L2.INT; // forward reference
+    uint256 internal constant LIMIT = MAX * L1.INT;  // same file & external library constant
+    uint256 internal constant NESTED = LIMIT + CONST; // nested & same contract constant
+
+    uint256[L1.INT] internal arr1; // error, backward reference
+    uint256[L2.INT] internal arr2; // error, forward reference
+}
+
+contract C2 is C1 {
+    uint256 internal constant INHERITED = NESTED + CONST * LIMIT; // inherited constants
+}
+
+contract C3 is C2 {
+    uint256 internal constant NESTED_INHERITED = INHERITED + NESTED + CONST * LIMIT; // nest-inherited constants
+
+    uint256[CONST] internal arr3;            // error, nest-inherited constants
+    uint256[NESTED_INHERITED] internal arr4; // error, same contract constant
+}
+
+library L2 {
+    uint256 internal constant INT = 100;
+}
+
+// ----
+// TypeError 5462: (366-372): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (430-436): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (742-747): Invalid array length, expected integer literal or constant expression.
+// TypeError 5462: (822-838): Invalid array length, expected integer literal or constant expression.
--- a/test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_file_level_constants.sol
+++ b/test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_file_level_constants.sol
@@ -1,0 +1,11 @@
+==== Source: A.sol ====
+import "B.sol" as B;
+
+uint constant X = 1;
+uint constant Y = B.Y;
+
+==== Source: B.sol ====
+import "A.sol" as A;
+
+uint constant X = A.X;
+uint constant Y = 2;
--- a/test/libsolidity/syntaxTests/constants/constants_with_dependencies_on_library_constants_multi_file.sol
+++ b/test/libsolidity/syntaxTests/constants/constants_with_dependencies_on_library_constants_multi_file.sol
@@ -1,0 +1,17 @@
+==== Source: A.sol ====
+import "B.sol";
+
+library L {
+    uint constant X = 1;
+    uint constant Y = K.Y;
+}
+
+==== Source: B.sol ====
+import "A.sol";
+
+library K {
+    uint constant X = L.X;
+    uint constant Y = 2;
+}
+
+// ====
--- a/test/libsolidity/syntaxTests/constants/cyclic_dependency_file_and_library_constants.sol
+++ b/test/libsolidity/syntaxTests/constants/cyclic_dependency_file_and_library_constants.sol
@@ -1,0 +1,15 @@
+==== Source: A.sol ====
+import "B.sol";
+
+uint256 constant A = B.VAL + 1;
+
+==== Source: B.sol ====
+import "A.sol";
+
+library B {
+    uint256 constant VAL = A + 1;
+}
+
+// ----
+// TypeError 6161: (B.sol:33-61): The value of the constant VAL has a cyclic dependency via A.
+// TypeError 6161: (A.sol:17-47): The value of the constant A has a cyclic dependency via VAL.
--- a/test/libsolidity/syntaxTests/constants/cyclic_dependency_library_constants.sol
+++ b/test/libsolidity/syntaxTests/constants/cyclic_dependency_library_constants.sol
@@ -1,0 +1,11 @@
+library A {
+    uint256 constant VAL = B.VAL + 1;
+}
+
+library B {
+    uint256 constant VAL = A.VAL + 1;
+}
+
+// ----
+// TypeError 6161: (16-48): The value of the constant VAL has a cyclic dependency via VAL.
+// TypeError 6161: (69-101): The value of the constant VAL has a cyclic dependency via VAL.


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3158,6 +3158,12 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
 
 			annotation.isPure = isPure;
 		}
+	if (
+		VariableDeclaration const* varDecl = dynamic_cast<decltype(varDecl)>(annotation.referencedDeclaration);
+		!annotation.isPure.set() &&
```
chriseth, line 3219: This looks like it depends on whether it is a forward-reference or not - I'm not sure we really want that.

chriseth: How does it work for file-level constants?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3158,6 +3158,12 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
+		varDecl && varDecl->value() && varDecl->value()->annotation().isPure.set()
```
```suggestion
		varDecl &&
		varDecl->value() &&
		varDecl->value()->annotation().isPure.set()
```

libsolidity/analysis/TypeChecker.cpp
cameel: > This looks like it depends on whether it is a forward-reference or not - I'm not sure we really want that.

So we basically need a separate pass that walks the AST, keeps track of dependencies and can go back to them? Or maybe just does it recursively?

> How does it work for file-level constants?

Yeah, I think we need test cases covering that. Also a mix between file-level and contract constants.

test/libsolidity/semanticTests/constants/define_constants_with_external_constant_member_var.sol
```diff
@@ -0,0 +1,25 @@
+==== Source: external.sol ====
```
cameel: This should be in `syntaxTests/`. The returned values are not relevant to us and in a semantic test you don't see the returned error messages.

test/libsolidity/semanticTests/constants/define_constants_with_external_constant_member_var.sol
```diff
@@ -0,0 +1,25 @@
...
+
+library External {
+    uint256 internal constant MAX = 1;
+}
+
+==== Source: usage.sol ====
```
cameel: This could be a single-file test. I don't think that the fact that these are in separate files affects this in any way.

cameel (changes requested):

test/libsolidity/semanticTests/constants/define_constants_with_external_constant_member_var.sol
StrongerXi: Done

test/libsolidity/semanticTests/constants/define_constants_with_external_constant_member_var.sol
StrongerXi: Done

libsolidity/analysis/TypeChecker.cpp
StrongerXi: These are good concerns, but fortunately I think we can bypass them. TL;DR: we should probably use `varDecl->isConstant()` here.

At the type checker level, when we see a member reference, we should simply respect its declaration type. In this case, use `varDecl->isConstant()`. Afaik this is what most of the type checkers do, and what they should do (else you need type inference).

This way, we don't need to worry about annotation of the actual `varDecl->value()`, i.e., if it's not pure, type checker will complain when visiting that `varDecl`.

If we have a cycle, [ConstStateVarCircularReferenceChecker](https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/PostTypeChecker.cpp#L148-L149) will take care of it.

I included 4 tests which should cover all the file/contract level constant interactions.

libsolidity/analysis/TypeChecker.cpp
StrongerXi: Done

StrongerXi:
Btw, where does the "constant variable evaluation & substitution" step happen? I wonder if this would affect it, since we are allowing member access in rhs of constant variables now.

StrongerXi:
Looking deeper, we were handling this correctly for file-level constant because of [this line](https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/TypeChecker.cpp#L3506). Member access is similar to identifier, so this should give more confidence to this PR.

cameel:
> Btw, where does the "constant variable evaluation & substitution" step happen?

See [`ConstantEvaluator.h`](https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/ConstantEvaluator.h) and the places where it's used.

StrongerXi:
> > Btw, where does the "constant variable evaluation & substitution" step happen?
> 
> See [`ConstantEvaluator.h`](https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/ConstantEvaluator.h) and the places where it's used.

Hmm I did check all the usages but none of them is doing what I was looking for.

Then I dumped the IR and things make sense now. At least at the IR level, it seems that each constant (or all?) variable is turned into a function, and the initialization expression is linearized into IR statements in the function body. I assume constant propagation will clean things up further?

Regardless, this PR shouldn't affect things there.

StrongerXi:
Rebase.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3158,6 +3158,14 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
 			annotation.isPure = isPure;
 		}
+	if (
+		VariableDeclaration const* varDecl = dynamic_cast<decltype(varDecl)>(annotation.referencedDeclaration);
+		!annotation.isPure.set() &&
+		varDecl &&
+		varDecl->isConstant()
+	) {
+		annotation.isPure = true;
+	}
```
```suggestion
	)
		annotation.isPure = true;
```

test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
```diff
@@ -0,0 +1,13 @@
+==== Source: A.sol ====
+import "B.sol";
+
+uint256 constant A = B + 1;
+
+==== Source: B.sol ====
+import "A.sol";
+
+uint256 constant B = A + 1;
```
cameel, line 9: These cyclic cases are fine but we also need some that will actuallyl compile and where the visiting order could have an impact. For example something like this:

```solidity
==== Source: A.sol ====
import "B.sol";

library L {
    uint constant X = 1;
    uint constant Y = K.Y;
}
==== Source: B.sol ====
import "A.sol";

library K {
    uint constant X = L.X;
    uint constant Y = 2;
}
```

or
```solidity
==== Source: A.sol ====
import "B.sol" as B;

uint constant X = 1;
uint constant Y = B.Y;
==== Source: B.sol ====
import "A.sol" as A;

uint constant X = A.X;
uint constant Y = 2;
```

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
```diff
@@ -0,0 +1,11 @@
+uint256 constant MAX = 1;
```
cameel: I'd call this test `constant_with_dependencies_on_constants.sol` and I think we should test more cases here:
- Dependency on another constant from the same contract.
- Dependency on inherited constants.
- Dependency on file-level constants.

test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
cameel, line 9: The fact that these are in separate files does not seem relavant to the test. It would still fail if it was a single file, wouldn't it?

libsolidity/analysis/TypeChecker.cpp
cameel: Great! This looks like a much better approach.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3158,6 +3158,14 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
 
 			annotation.isPure = isPure;
 		}
+	if (
+		VariableDeclaration const* varDecl = dynamic_cast<decltype(varDecl)>(annotation.referencedDeclaration);
```
```suggestion
		auto const* variableDeclaration = dynamic_cast<VariableDeclaration const*>(annotation.referencedDeclaration);
```

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
```diff
@@ -0,0 +1,11 @@
...
+
+library Internal {
+    uint256 internal constant SCALE = 100;
+}
+
+contract Usage {
+    uint256 private constant LIMIT = MAX * Internal.SCALE;
+}
```
cameel: Can you add a case with a deeper tree of dependencies that goes through several levels? We should especially have one that ensures that `LIMIT` can now be used as a value in another constant definition.

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
```diff
@@ -0,0 +1,11 @@
+uint256 constant MAX = 1;
+
+library Internal {
+    uint256 internal constant SCALE = 100;
+}
+
+contract Usage {
```
cameel: Let's switch to our typical naming that lets you see at a glance what is a contract, what is a library, etc.

```suggestion
library L {
    uint256 internal constant SCALE = 100;
}

contract C {
```

cameel (commented):
This looks great. Tests need to be a bit stronger (see my suggestions) but other than that it's a nice enhancement that users will probably appreciate!

Also, please add a changelog entry.

libsolidity/analysis/TypeChecker.cpp
StrongerXi: Done.

test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
StrongerXi: Right, but I thought of these as sanity checks to prove that this PR won't mistakenly allow such illegal file-level cyclic dependencies.

StrongerXi: Added as test case.

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
StrongerXi: Done

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
StrongerXi: Done

test/libsolidity/syntaxTests/constants/define_constants_with_external_constant_member_var.sol
StrongerXi: Done

libsolidity/analysis/TypeChecker.cpp
StrongerXi: Done. I'm keeping variable name though since `varDecl` seems to be the name used for `VariableDeclaration` elsewhere.

test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
cameel: Can you add the one with libraries too? It's actually the more important one of the two since it's a case the PR should fix while the one with file-level constants already works.

cameel: Fair enough.

libsolidity/analysis/TypeChecker.cpp
cameel: I think some of the older code is overusing these shortened names. Personally I'd use full words but fair enough, we do have these shortened names in some places.

Changelog.md
```diff
@@ -5,6 +5,7 @@ Language Features:
 
 Compiler Features:
 * LSP: Add rudimentary support for semantic highlighting.
+* TypeChecker: Support using foreign constants as compile-time constants
```
cameel: I think this wording might be a little clearer to users:
```suggestion
* TypeChecker: Support using library constants in initializers of other constants.
```

The change seems more generic but looks like in practice it only affected library constants.

chriseth:
Please also add some tests where the constants are used as array lengths of statically-sized arrays, both as forward and as backward references.

test/libsolidity/syntaxTests/constants/cyclic_dependency_file_constants.sol
StrongerXi: Ah I missed it. Done.

Changelog.md
StrongerXi: Agree, done.

StrongerXi:
> Please also add some tests where the constants are used as array lengths of statically-sized arrays, both as forward and as backward references.

I think this is more involved than expected. The logic chain is as follows:

- To use a constant variable as array lengths, `DeclarationTypeChecker::endVisit(ArrayTypeName)` requires the variable to either (1) already be a rational number, e.g., `X = 42`, or (2) can be evaluated to a rational number by `ConstantEvaluator`.
- To use `X = L.INT` as array length, we must go with (2).
- `ConstantEvaluator` currently doesn't evaluate `MemberAccess`. But to support it, we must check for circular dependencies in constant declarations first, which is currently done in `PostTypeChecker`.

I think this reveals a larger issue than this PR is concerned -- We probably want to augment `ConstantEvaluator` to evaluate more things like member access, but that eventually entails moving constant circular dependence check to earlier places in the compilation stack.

Currently the compilation stack is a bit awkward for constant evaluation, i.e.,

```
...
DeclarationTypeChecker (use of constant variable as array size is checked)
...
TypeChecker (purity information is propagated, e.g., this PR)
...
PostTypeChecker (constant circular dependence is checked)
...
```

I think the logics in parenthesis should be reversed...?

StrongerXi:
Rebase.

chriseth:
Sorry, I should have clarified: Please add tests so that we see that it does not catastrophically fail. I did not mean to implement this feature for the constant evaluator.

And you are right, the constant evaluator is not at the right place, we have a larger issue planned to extend it.

StrongerXi:
> Sorry, I should have clarified: Please add tests so that we see that it does not catastrophically fail. I did not mean to implement this feature for the constant evaluator.
> 
> And you are right, the constant evaluator is not at the right place, we have a larger issue planned to extend it.

Got it. Done.

test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_constants.sol
```diff
@@ -0,0 +1,35 @@
...
+    uint256 internal constant INT = 100;
+}
+
+contract C1 {
+    uint256 internal constant CONST = 20 + L2.INT; // forward reference
+    uint256 internal constant LIMIT = MAX * L1.INT;  // same file & external library constant
+    uint256 internal constant NESTED = LIMIT + CONST; // nested & same contract constant
+
+    uint256[L1.INT] internal arr1; // error, backward reference
+    uint256[L2.INT] internal arr2; // error, forward reference
```
cameel: Can you put the part testing constants in array sizes in a separate test? The dependencies that we were testing before need a positive test that causes no errors. Some errors are fatal so if there's an error, you cannot be 100% sure that there would be no other errors if you removed it.

You are really testing two different things here. They should be in separate files named in such a way that we can easily discern them.

cameel (dismissed):
One last thing.

It's minor enough that I'm approving this already but I really think these tests need to be separate.

test/libsolidity/syntaxTests/constants/constant_with_dependencies_on_constants.sol
StrongerXi: Agree, done.

cameel: I meant something slightly different, i.e.:
1) A passing test that only tests dependencies (i.e. basically the test you had originally).
2) A failing tests that uses constants with dependencies as array sizes.

nishant-sachdeva: Pushed changes to resolve this comment

Changelog.md
```diff
@@ -12,6 +12,7 @@ Language Features:
 Compiler Features:
  * LSP: Add rudimentary support for semantic highlighting.
  * Yul Optimizer: Improve inlining heuristics for via IR code generation and pure Yul compilation.
+ * TypeChecker: Support using library constants in initializers of other constants
```
cameel: @nishant-sachdeva Now that we released 0.8.15 the PR needs to be rebased and this moved to 0.8.16 changelog.

cameel (dismissed):
It's good. I'd merge this already but we have just released 0.8.15 so the changelog entry needs to be moved.

ekpyron (approved):
Approving, partly based on @cameel's earlier approval.

PR #12929, opened 2022-04-12
[CI] Automate preparation of release binaries for publishing (via workspace)

### LINKED ISSUE #9933: Automate publishing of release binaries

cameel:
Replaces #12181.
Closes #9933.

This is a simplified version of #12181 (as discussed in https://github.com/ethereum/solidity/pull/12181#pullrequestreview-882043322). Turns out it's possible to do it via workspace after all.

It's not complete automation of the publishing process as stated in #9933 but it at least takes care of the most tedious and error prone part of the process so I think it's good enough.

The PR adds a CI job that gathers release binaries from other jobs and attaches them as artifacts, under two naming conventions - for github release page and for solc-bin.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -418,6 +418,13 @@ defaults:
         tags:
           only: /.*/
 
+  - workflow_trigger_on_releases: &workflow_trigger_on_releases
+      filters:
+        tags:
+          only: /^v.*/
+        branches:
+          ignore: /.*/
+
   - workflow_ubuntu2004: &workflow_ubuntu2004
       <<: *workflow_trigger_on_tags
       requires:
@@ -1416,6 +1423,46 @@ jobs:
           path: all-bytecode-reports.zip
       - gitter_notify_failure_unless_pr
 
+  c_release_binaries:
+    <<: *base_ubuntu2004
+    steps:
+      - checkout
+      - attach_workspace:
+          at: workspace
+      - run:
+          name: Gather and rename binaries from dependent jobs
+          command: |
+            mkdir github/
+            cp workspace/solc/solc             github/solc-static-linux
+            cp workspace/build/solc/solc       github/solc-macos
+            cp workspace/solc/Release/solc.exe github/solc-windows.exe
+            cp workspace/soljson.js            github/soljson.js
+
+            cd github/
+            tar --create --file ../github-binaries.tar *
+      - store_artifacts:
+          path: github-binaries.tar
+      - run:
+          name: Rename binaries to solc-bin naming convention
+          command: |
+            full_version=$(
+              github/solc-static-linux --version |
+              sed -En 's/^Version: ([0-9.]+.*\+commit\.[0-9a-f]+(\.mod)?).*$/\1/p'
+            )
+
+            mkdir -p solc-bin/{linux-amd64,macosx-amd64,windows-amd64,bin}
+
+            mv github/solc-static-linux "solc-bin/linux-amd64/solc-linux-amd64-${full_version}"
+            mv github/solc-macos        "solc-bin/macosx-amd64/solc-macosx-amd64-${full_version}"
+            mv github/solc-windows.exe  "solc-bin/windows-amd64/solc-windows-amd64-${full_version}.exe"
+            mv github/soljson.js        "solc-bin/bin/soljson-${full_version}.js"
+
+            cd solc-bin/
+            tar --create --file ../solc-bin-binaries.tar *
+      - store_artifacts:
+          path: solc-bin-binaries.tar
+      - gitter_notify_failure_unless_pr
+
 workflows:
   version: 2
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1401,6 +1419,40 @@ jobs:
...
+            full_version=$(
+              github/solc-static-linux --version |
+              sed -En 's/^Version: ([0-9.]+.*\+commit\.[0-9a-f]+(\.mod)?).*$/\1/p'
+            )
+
+            mkdir -p "solc-bin/{linux-amd64,macosx-amd64,windows-amd64,bin}"
+            mv github/solc-static-linux "solc-bin/linux-amd64/solc-linux-amd64-${full_version}"
+            mv github/macosx-amd64      "solc-bin/macosx-amd64/solc-macosx-amd64-${full_version}"
+            mv github/solc-windows.exe  "solc-bin/windows-amd64/solc-windows-amd64-${full_version}.exe"
+            mv github/soljson.js        "solc-bin/bin/soljson-${full_version}.js"
```
cameel: This is something I could also put in a script (see `rename-circleci-binaries-for-solc-bin.sh` in #12181), which would allow running it locally instead of renaming manually (this workflow will become obsolete after this PR though). It also fits better with the idea of not keeping bigger pieces bash code outside of CI config.

The way it's done here is much simpler though.

cameel:
Looks like it works. Removing the debug commit.

For reference, here's a successful run: [1040499](https://app.circleci.com/pipelines/github/ethereum/solidity/23697/workflows/dc5b77e3-ee4b-4c78-8f76-4469ff514767/jobs/1040499/artifacts)

.circleci/config.yml
```diff
@@ -58,6 +58,10 @@ commands:
 
             [[ "<< parameters.event >>" == "failure" ]] && message=" ❌ [${workflow_name}] Job ${job} failed on **${CIRCLE_BRANCH}**. Please see [build ${CIRCLE_BUILD_NUM}](${CIRCLE_BUILD_URL}) for details."
             [[ "<< parameters.event >>" == "success" ]] && message=" ✅ [${workflow_name}] Job ${job} succeeded on **${CIRCLE_BRANCH}**. Please see [build ${CIRCLE_BUILD_NUM}](${CIRCLE_BUILD_URL}) for details."
+            [[ "<< parameters.event >>" == "release" ]] && message=" 📦 Release binaries for version **${CIRCLE_TAG}** are ready and attached as artifacts to [build ${CIRCLE_BUILD_NUM}](${CIRCLE_BUILD_URL}). **Please make sure the whole workflow succeeded before using them.**"
+
+            # The release notification only makes sense on tagged commits
+            [[ "<< parameters.event >>" == "release" ]] && { [[ $CIRCLE_TAG != "" ]] || { echo "Not a tagged commit - notification skipped."; exit 0; } }
```
Marenz, line 64: I am a bit confused, the message in line 64 basically locally echos that it will skip the notification, while line 61 actually does post the notification.. wouldn't line 61 need that extra condition? a local echo that we won't post a notification on its own doesn't seem very useful to me. But I am probably missing some logic here..?

cameel: Line 61 just sets the value of the `message` variable. The code that actually posts it is the `curl` command below. If line 64 exits, the message won't actually be used.

I could move line 61 lower or put 61 and 64 under a single `if` but I thought it was nicer to have these messages side by side for easy comparison.

Marenz: I think the positioning is fine, but maybe add in the comment for 64 something like "bail out early for untagged commits"?

cameel: Done.

Marenz (dismissed):

cameel:
I switched to creating a tarball so that the executable flag is preserved. Here are the artifacts from a successful run: [1058937/artifacts](https://app.circleci.com/pipelines/github/ethereum/solidity/24104/workflows/cf0ba70a-6930-40d5-9e2e-c83224fe241b/jobs/1058937/artifacts). Now I'm removing the debug commit that enables this job outside of releases.

ekpyron (approved):

PR #12731, opened 2022-03-02
Remove inlining constraint for large functions when targetting the new code transform.

ekpyron:
**Depends on #12937.**

PR #12937, opened 2022-04-13
Disable external test cases in which Hardhat's heuristics break after relaxing inlining constraints

cameel:
~Depends on #12936.~ Merged.
**Should not be merged separately from #12731!**

The PR disables a bunch of tests where Hardhat can no longer detect correct revert reasons, most likely due to https://github.com/NomicFoundation/hardhat/issues/2453. I think that heuristics depend on specific opcode sequences and inlining those sequences probably throws them off.

This is really a part of #12731 because these tests work fine on `develop` but I need it in a separate PR to have a fair benchmark comparison. Disabling these tests lowers gas usage so we can't compare directly with `develop`.

I'm marking it as draft for now because it should not be merged until we decide to merge #12731 as well.


ekpyron:
Rebasing this to be able to rebase https://github.com/ethereum/solidity/pull/12731

cameel:
PR updated to account for tests newly broken due to more relaxed heuristics. Now everything in #12731 should pass.

chriseth (approved):

PR #12938, opened 2022-04-14
Script for keeping main branches of external test repos in sync with upstream

cameel:
Some time ago when the upstream of the [perpetual-pools-contracts](https://github.com/solidity-external-tests/perpetual-pools-contracts) repo became unavailable I created forks of all our external tests, even those where we use upstream directly. This way we'll be able to switch to the fork when this happens again. Keeping these forks up to date is a hassle though. Today I created a new fork for the recently merged #12562 so I though I'd also create a script to automated updating it.

It would probably be a good idea to have this run automatically in the nightly CI run. In this PR I'm only including the script thought since that would require setting up tokens in CI to be able to push to `solidity-external-docs` org.

### DIFF
--- a/scripts/externalTests/update_external_repos.sh
+++ b/scripts/externalTests/update_external_repos.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+#------------------------------------------------------------------------------
+# Clones external test repositories from solidity-external-tests organization
+# and for each of them pulls latest upstream changes from the main branch and
+# pushes them to our fork.
+#
+# The script assumes that the current user has write access to
+# solidity-external-tests and that git is configured to be able to push there
+# without specifying the password (e.g. with the key already unlocked and loaded
+# into ssh-agent). Otherwise git will keep asking for password for each repository.
+#
+# Usage:
+#
+#    ./update_external_repos.sh [<target_dir>]
+#
+# <target_dir>: directory where the clones of the repositories are stored.
+#               If omitted, a temporary directory will be created.
+#
+# ------------------------------------------------------------------------------
+# This file is part of solidity.
+#
+# solidity is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# solidity is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with solidity.  If not, see <http://www.gnu.org/licenses/>
+#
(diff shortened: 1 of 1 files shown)


scripts/externalTests/update_external_repos.sh
```diff
@@ -0,0 +1,120 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with solidity.  If not, see <http://www.gnu.org/licenses/>
+#
+# (c) 2022 solidity contributors.
+#------------------------------------------------------------------------------
+
+set -euo pipefail
+
+target_dir="${1:-$(mktemp -d -t update_external_repos_XXXXXX)}"
```
wechman, line 41: It makes a target directory when it is not defined as input parameter. What about opposite situation? Don't we want to make a directory as well?

scripts/externalTests/update_external_repos.sh
```diff
@@ -0,0 +1,120 @@
...
+sync_branch brink-core                master
+sync_branch dappsys-monolithic        master
+sync_branch elf-contracts             main
+sync_branch ens                       master
+sync_branch ens-contracts             master
+sync_branch euler-contracts           master
+sync_branch gp-v2-contracts           main
+sync_branch mock-contract             master
+sync_branch safe-contracts            main
+sync_branch util-contracts            main
```
wechman, line 110: I get an error when syncing util-contracts:
`hint: 'main' matched more than one remote tracking branch.
hint: We found 2 remotes with a reference that matched.`

scripts/externalTests/update_external_repos.sh
wechman, line 41: Don't we want to remove a temporary directory at the end of the script?

cameel: Good point. I can make it create it. The assumption is that you already have these repos stored on disk if you're giving a path but yeah, the first time you run it you might want it to create the dir for you.

cameel: Not necessarily. You might want to inspect contents or rerun the script, now giving it that tmp path. It's a relatively heavy operation so I prefer to err on the side of not removing it automatically. My assumption is that the temp dir does not really hurt being there for a short time and it will be cleared anyway on next reboot (or whenever the user cleans his tmp if it's not on tmpfs).

scripts/externalTests/update_external_repos.sh
cameel: Hmm... let me check.

cameel: Fixed. It was because `git checkout <branch>` fails to create a tracking branch when it's present on more than one remote. I now switched to explicitly specifying branch name and branch to track. I now also force branch creation when it already exists.

scripts/externalTests/update_external_repos.sh
cameel: Done.

matheusaaguiar (approved):
I did some testing using personal repositories and it seems to be working fine.

wechman (approved):

PR #12941, opened 2022-04-17
yulopti: Add support for Yul Objects

### LINKED ISSUE #12856: Support Yul objects in yulopti
## Abstract
[yulopti](https://github.com/ethereum/solidity/blob/develop/test/tools/yulopti.cpp) is a simple tool for executing [optimizer step](https://docs.soliditylang.org/en/latest/internals/optimizer.html#optimizer-steps) sequences step by step.

Currently it only supports raw Yul input (example: [`medium.yul`](https://github.com/ethereum/solidity/blob/develop/test/libyul/yulOptimizerTests/fullSuite/medium.yul)). It would be convenient if it supported Yul objects as well (example: [`subObjectAccess.yul`](https://github.com/ethereum/solidity/blob/adc58c67b432ada0d7a4fc68f3de610ba78358c0/test/libyul/objectCompiler/subObjectAccess.yul)).

## Motivation
The compiler never produces raw Yul as output. When invoked with `--ir` or `--ir-optimized`, the Yul code is always wrapped in an object. The object part needs to be stripped for it to be usable with yulopti.

## Specification
- Use `ObjectParser` for parsing Yul objects (see how it was done in yul-phaser: #8513).
- yulopti should optimize all objects present in the file, including nested ones.
- (optional) add a command-line flag for selecting a specific object via a dotted path (e.g. `--object x.y.z`)

## Backwards Compatibility
This is an internal tool. We don't guarantee backwards-compatibility.

dflupu:
This fixes [#12856](https://github.com/ethereum/solidity/issues/12856).

~I've made a small change to ObjectParser as part of this to set the `.subId` fields as they didn't seem to be set anywhere.~

This is my first contribution to this project so be aware of that when reviewing the changes.

### DIFF
--- a/libyul/Object.cpp
+++ b/libyul/Object.cpp
@@ -31,12 +31,24 @@
 
 #include <boost/algorithm/string.hpp>
 
+#include <range/v3/view/drop.hpp>
 #include <range/v3/view/transform.hpp>
 
 using namespace solidity;
 using namespace solidity::langutil;
 using namespace solidity::util;
 using namespace solidity::yul;
+
+
+Object const* Object::at(YulString _name) const
+{
+	yulAssert(
+		this->subIndexByName.count(_name),
+		"Assembly object <" + _name.str() + "> not found."
+	);
+	size_t subIndex = this->subIndexByName.at(_name);
+	return dynamic_cast<Object const*>(this->subObjects[subIndex].get());
+}
 
 std::string Data::toString(Dialect const*, DebugInfoSelection const&, CharStreamProvider const*) const
 {
@@ -129,11 +141,8 @@ std::set<YulString> Object::qualifiedDataNames() const
 	return qualifiedNames;
 }
 
-std::vector<size_t> Object::pathToSubObject(YulString _qualifiedName) const
+void Object::visitPath(YulString _qualifiedName, std::function<bool(Object const*)> const& _visitor) const
 {
-	yulAssert(_qualifiedName != name, "");
-	yulAssert(subIndexByName.count(name) == 0, "");
-
 	if (boost::algorithm::starts_with(_qualifiedName.str(), name.str() + "."))
 		_qualifiedName = YulString{_qualifiedName.str().substr(name.str().length() + 1)};
 	yulAssert(!_qualifiedName.empty(), "");
@@ -141,21 +150,51 @@ std::vector<size_t> Object::pathToSubObject(YulString _qualifiedName) const
 	std::vector<std::string> subObjectPathComponents;
 	boost::algorithm::split(subObjectPathComponents, _qualifiedName.str(), boost::is_any_of("."));
 
-	std::vector<size_t> path;
 	Object const* object = this;
 	for (std::string const& currentSubObjectName: subObjectPathComponents)
 	{
 		yulAssert(!currentSubObjectName.empty(), "");
-		auto subIndexIt = object->subIndexByName.find(YulString{currentSubObjectName});
-		yulAssert(
-			subIndexIt != object->subIndexByName.end(),
-			"Assembly object <" + _qualifiedName.str() + "> not found or does not contain code."
-		);
-		object = dynamic_cast<Object const*>(object->subObjects[subIndexIt->second].get());
-		yulAssert(object, "Assembly object <" + _qualifiedName.str() + "> not found or does not contain code.");
-		yulAssert(object->subId != std::numeric_limits<size_t>::max(), "");
-		path.push_back({object->subId});
-	}
 
+		YulString subObjectName = YulString{currentSubObjectName};
+		object = object->at(subObjectName);
+		if (object && _visitor(object))
+			break;
+	}
+}
+
+std::vector<size_t> Object::pathToSubObject(YulString _qualifiedName) const
+{
+	std::vector<size_t> path;
+	yulAssert(_qualifiedName != name, "");
+	yulAssert(subIndexByName.count(name) == 0, "");
+
+	this->visitPath(_qualifiedName, [&](Object const* _object) -> bool {
+		yulAssert(_object->subId != std::numeric_limits<size_t>::max(), "");
+		path.push_back(_object->subId);
+		return false;
+	});
 	return path;
+}
+
+Object const* Object::subObjectAt(YulString _qualifiedName)
+{
+	yulAssert(!_qualifiedName.empty(), "");
+
+	// If there is no `.` in the given `_qualifiedName`, the target
+	// object name is considered to be equal to the `_qualifiedName`, otherwise,
+	// the target object name is the last element in the path given by `_qualifiedName`.
+	YulString targetObjectName = _qualifiedName;
+	size_t targetObjectPos = _qualifiedName.str().find_last_of(".");
+	if (targetObjectPos != std::string::npos)
+		targetObjectName = YulString(_qualifiedName.str().substr(targetObjectPos + 1));
+
+	Object const* foundObject = nullptr;
+	this->visitPath(_qualifiedName, [&](Object const* _subObject) -> bool {
+		if (targetObjectName != _subObject->name)
+			return false;
+		foundObject = _subObject;
+		return true;
+	});
+
+	return foundObject;
 }
--- a/libyul/Object.h
+++ b/libyul/Object.h
@@ -31,6 +31,7 @@
 
 #include <memory>
 #include <set>
+#include <string>
 #include <limits>
 #include <json/json.h>
 
@@ -115,6 +116,21 @@ public:
 	/// pathToSubObject("A1.E2") == {1}
 	/// The path must not lead to a @a Data object (will throw in that case).
 	std::vector<size_t> pathToSubObject(YulString _qualifiedName) const;
+
+	/// Searches for a subobject at @param _qualifiedName within the current object.
+	/// @returns a pointer to the subobject or a nullptr if it was not found.
+	Object const* subObjectAt(YulString _qualifiedName);
+
+	/// Visits all subobjects in the path given by the @a _qualifiedName
+	/// of the current object applying the function @a _visitor.
+	/// If @a _visitor returns `true` the visiting stops, otherwise,
+	/// it continues until the last subobject in the path is reached.
+	void visitPath(
+		YulString _qualifiedName,
+		std::function<bool(Object const*)> const& _visitor
+	) const;
+
+	Object const* at(YulString _name) const;
 
 	/// sub id for object if it is subobject of another object, max value if it is not subobject
 	size_t subId = std::numeric_limits<size_t>::max();
--- a/test/tools/yulopti.cpp
+++ b/test/tools/yulopti.cpp
@@ -29,6 +29,7 @@
 #include <libyul/AST.h>
 #include <libyul/AsmParser.h>
 #include <libyul/AsmPrinter.h>
+#include <libyul/ObjectParser.h>
 #include <libyul/Object.h>
 #include <liblangutil/SourceReferenceFormatter.h>
 
@@ -90,33 +91,29 @@ public:
 	{
 		ErrorList errors;
 		ErrorReporter errorReporter(errors);
-		CharStream _charStream(_input, "");
+		CharStream charStream(_input, "");
+
 		try
 		{
-			m_ast = yul::Parser(errorReporter, m_dialect).parse(_charStream);
-			if (!m_ast || !errorReporter.errors().empty())
+			ObjectParser parser(errorReporter, m_dialect);
+
+			auto scanner = make_shared<Scanner>(charStream);
+			m_inputIsCodeBlock = (scanner->currentToken() == Token::LBrace);
+			m_object = parser.parse(scanner, false);
+
+			if (!m_object || !errorReporter.errors().empty())
 			{
 				cerr << "Error parsing source." << endl;
-				printErrors(_charStream, errors);
-				throw std::runtime_error("Could not parse source.");
+				printErrors(charStream, errors);
+				solThrow(Exception, "Could not parse source.");
 			}
-			m_analysisInfo = make_unique<yul::AsmAnalysisInfo>();
-			AsmAnalyzer analyzer(
-				*m_analysisInfo,
-				errorReporter,
-				m_dialect
-			);
-			if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
-			{
-				cerr << "Error analyzing source." << endl;
-				printErrors(_charStream, errors);
-				throw std::runtime_error("Could not analyze source.");
-			}
+
+			runCodeAnalyzer(errorReporter);
 		}
 		catch(...)
 		{
 			cerr << "Fatal error during parsing: " << endl;
-			printErrors(_charStream, errors);
+			printErrors(charStream, errors);
 			throw;
 		}
 	}
@@ -129,7 +126,7 @@ public:
 		yulAssert(_columns > 0);
 		auto const& optimiserSteps = OptimiserSuite::stepAbbreviationToNameMap();
 		auto hasShorterString = [](auto const& a, auto const& b) { return a.second.size() < b.second.size(); };
-		size_t longestDescriptionLength = std::max(
+		size_t longestDescriptionLength = max(
 			max_element(optimiserSteps.begin(), optimiserSteps.end(), hasShorterString)->second.size(),
 			max_element(_extraOptions.begin(), _extraOptions.end(), hasShorterString)->second.size()
 		);
@@ -170,28 +167,156 @@ public:
 		}
 	}
 
-	void disambiguate()
+	void applyFunctionToObjectAndSubobjects(Object& _object, function<void(Object&)> _function)
 	{
-		*m_ast = std::get<yul::Block>(Disambiguator(m_dialect, *m_analysisInfo)(*m_ast));
-		m_analysisInfo.reset();
-		m_nameDispenser.reset(*m_ast);
+		for (auto const& subObjectNode: _object.subObjects)
+		{
+			auto subObject = dynamic_pointer_cast<Object>(subObjectNode);
+
+			if (subObject != nullptr)
+				applyFunctionToObjectAndSubobjects(*subObject, _function);
+		}
+
+		_function(_object);
 	}
 
-	void runSteps(string _source, string _steps)
+	void runCodeAnalyzer(ErrorReporter& _errorReporter)
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				_object.analysisInfo = make_shared<yul::AsmAnalysisInfo>();
+
+				AsmAnalyzer analyzer(
+					*_object.analysisInfo,
+					_errorReporter,
+					m_dialect,
+					{},
+					_object.qualifiedDataNames()
+				);
+
+				bool success = analyzer.analyze(*_object.code);
+				yulAssert(success && !_errorReporter.hasErrors(), "Invalid assembly/yul code.");
+			}
+		);
+	}
+
+	void runCodeDisambiguator()
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				_object.code = make_shared<yul::Block>(
+					get<yul::Block>(Disambiguator(m_dialect, *_object.analysisInfo)(*_object.code))
+				);
+
+				_object.analysisInfo.reset();
+			}
+		);
+	}
+
+	void runSequence(string_view _steps)
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				OptimiserStepContext context = createOptimiserStepContext(_object);
+				OptimiserSuite{context}.runSequence(_steps, *_object.code);
+			}
+		);
+	}
+
+	void runVarNameCleaner()
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				OptimiserStepContext context = createOptimiserStepContext(_object);
+				VarNameCleaner::run(context, *_object.code);
+			}
+		);
+	}
+
+	void runStackCompressor()
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object) { StackCompressor::run(m_dialect, _object, true, 16); }
+		);
+	}
+
+	void printObject(string const& _objectPath)
+	{
+		if (!m_inputIsCodeBlock)
+		{
+			if (_objectPath.empty() || m_object->name.str() == _objectPath)
+				cout << m_object->toString(&m_dialect) << endl;
+			else
+			{
+				Object const* subObject = m_object->subObjectAt(YulString(_objectPath));
+				if (subObject == nullptr)
+					solThrow(Exception, "Assembly object not found.");
+
+				cout << subObject->toString(&m_dialect) << endl;
+			}
+		}
+		else
+		{
+			yulAssert(
+				m_object->subObjects.empty(),
+				"Unexpected subObjects found."
+			);
+
+			if (!_objectPath.empty())
+				solThrow(Exception, "Object path argument cannot be used. Input is a code block.");
+
+			cout << AsmPrinter{m_dialect}(*m_object->code) << endl;
+		}
+	}
+
+	void resetNameDispenser()
+	{
+		m_nameDispenser = make_shared<NameDispenser>(
+			m_dialect,
+			m_reservedIdentifiers
+		);
+	}
+
+	void parseAndRunSteps(string const& _source, string const& _steps)
 	{
 		parse(_source);
-		disambiguate();
-		OptimiserSuite{m_context}.runSequence(_steps, *m_ast);
-		cout << AsmPrinter{m_dialect}(*m_ast) << endl;
+		runCodeDisambiguator();
+		runSequence(_steps);
 	}
 
-	void runInteractive(string _source, bool _disambiguated = false)
+	OptimiserStepContext createOptimiserStepContext(Object& _object)
 	{
+		bool isCreation = m_object.get() == &_object || !boost::ends_with(_object.name.str(), "_deployed");
+
+		return OptimiserStepContext{
+			m_dialect,
+			*m_nameDispenser,
+			m_reservedIdentifiers,
+			isCreation ? nullopt : make_optional(OptimiserSettings::standard().expectedExecutionsPerDeployment)
+		};
+	}
+
+	void runInteractive(string _source, string const& _objectPath, bool _disambiguated = false)
+	{
+		ErrorList errors;
+		ErrorReporter errorReporter(errors);
 		bool disambiguated = _disambiguated;
+
+		parse(_source);
+
 		while (true)
 		{
-			parse(_source);
-			disambiguated = disambiguated || (disambiguate(), true);
+			disambiguated = disambiguated || (runCodeDisambiguator(), true);
+
 			map<char, string> const& extraOptions = {
 				// QUIT starts with a non-letter character on purpose to get it to show up on top of the list
 				{'#', ">>> QUIT <<<"},
@@ -213,24 +338,18 @@ public:
 					case '#':
 						return;
 					case ',':
-						VarNameCleaner::run(m_context, *m_ast);
+						runVarNameCleaner();
 						// VarNameCleaner destroys the unique names guarantee of the disambiguator.
 						disambiguated = false;
 						break;
 					case ';':
-					{
-						Object obj;
-						obj.code = m_ast;
-						StackCompressor::run(m_dialect, obj, true, 16);
+						runStackCompressor();
 						break;
-					}
 					default:
-						OptimiserSuite{m_context}.runSequence(
-							std::string_view(&option, 1),
-							*m_ast
-						);
+						runSequence(string_view(&option, 1));
 				}
-				_source = AsmPrinter{m_dialect}(*m_ast);
+
+				resetNameDispenser();
 			}
 			catch (...)
 			{
@@ -238,22 +357,20 @@ public:
 				cerr << boost::current_exception_diagnostic_information() << endl;
 			}
 			cout << "----------------------" << endl;
-			cout << _source << endl;
+			printObject(_objectPath);
 		}
 	}
 
 private:
-	shared_ptr<yul::Block> m_ast;
+	shared_ptr<yul::Object> m_object;
+	bool m_inputIsCodeBlock = false;
+
 	Dialect const& m_dialect{EVMDialect::strictAssemblyForEVMObjects(EVMVersion{})};
-	unique_ptr<AsmAnalysisInfo> m_analysisInfo;
 	set<YulString> const m_reservedIdentifiers = {};
-	NameDispenser m_nameDispenser{m_dialect, m_reservedIdentifiers};
-	OptimiserStepContext m_context{
+	shared_ptr<NameDispenser> m_nameDispenser = make_shared<NameDispenser>(
 		m_dialect,
-		m_nameDispenser,
-		m_reservedIdentifiers,
-		solidity::frontend::OptimiserSettings::standard().expectedExecutionsPerDeployment
-	};
+		m_reservedIdentifiers
+	);
 };
 
 int main(int argc, char** argv)
@@ -264,7 +381,7 @@ int main(int argc, char** argv)
 		po::options_description options(
 			R"(yulopti, yul optimizer exploration tool.
 	Usage: yulopti [Options] <file>
-	Reads <file> as yul code and applies optimizer steps to it,
+	Reads <file> containing either a yul object or a yul code block and applies optimizer steps to it,
 	interactively read from stdin.
 	In non-interactive mode a list of steps has to be provided.
 	If <file> is -, yul code is read from stdin and run non-interactively.
@@ -282,6 +399,14 @@ int main(int argc, char** argv)
 				"steps",
 				po::value<string>(),
 				"steps to execute non-interactively"
+			)
+			(
+				"object",
+				po::value<string>()->value_name("path"),
+				"Dotted path to a specific Yul object in the source to print after running the steps. "
+				"If not specified, the top-level object will be selected. "
+				"The object will be printed recursively. "
+				"Only valid if the source actually contains objects (rather than a block of Yul code)."
 			)
 			(
 				"non-interactive,n",
@@ -307,6 +432,8 @@ int main(int argc, char** argv)
 		}
 
 		string input;
+		string objectPath;
+
 		if (arguments.count("input-file"))
 		{
 			string filename = arguments["input-file"].as<string>();
@@ -330,20 +457,28 @@ int main(int argc, char** argv)
 			return 1;
 		}
 
+		if (arguments.count("object"))
+			objectPath = arguments["object"].as<string>();
+
 		YulOpti yulOpti;
 		bool disambiguated = false;
+
 		if (!nonInteractive)
-			cout << input << endl;
+		{
+			yulOpti.parse(input);
+			yulOpti.printObject(objectPath);
+		}
+
 		if (arguments.count("steps"))
 		{
 			string sequence = arguments["steps"].as<string>();
 			if (!nonInteractive)
 				cout << "----------------------" << endl;
-			yulOpti.runSteps(input, sequence);
+			yulOpti.parseAndRunSteps(input, sequence);
 			disambiguated = true;
 		}
 		if (!nonInteractive)
-			yulOpti.runInteractive(input, disambiguated);
+			yulOpti.runInteractive(input, objectPath, disambiguated);
 
 		return 0;
 	}


libyul/ObjectParser.cpp
```diff
@@ -100,7 +100,10 @@ shared_ptr<Object> ObjectParser::parseObject(Object* _containingObject)
 			fatalParserError(8143_error, "Expected keyword \"data\" or \"object\" or \"}\".");
 	}
 	if (_containingObject)
+	{
+		ret->subId = _containingObject->subObjects.size();
```
chriseth: Currently, this is only set in the EVM object compiler. I think it is a reference to something from Assembly.cpp, so I don't think we should set it to an arbitrary value here.

test/tools/yulopti.cpp
```diff
@@ -265,10 +369,11 @@ int main(int argc, char** argv)
 		po::options_description options(
 			R"(yulopti, yul optimizer exploration tool.
 	Usage: yulopti [Options] <file>
-	Reads <file> as yul code and applies optimizer steps to it,
+	Reads <file> containing a yul object and applies optimizer steps to it,
 	interactively read from stdin.
 	In non-interactive mode a list of steps has to be provided.
 	If <file> is -, yul code is read from stdin and run non-interactively.
+	If <object> is provided then only the object matching the given path will be read.
```
chriseth: Can you explain what `<object>` is?

test/tools/yulopti.cpp
```diff
@@ -87,32 +88,44 @@ class YulOpti
 		}.printErrorInformation(_errors);
 	}
 
-	void parse(string const& _input)
+	shared_ptr<Object> getSubObject(shared_ptr<Object> const& _rootObject, string const& _path)
+	{
+		if (_path.empty())
+			return _rootObject;
+
+		auto pathToSubObject = _rootObject->pathToSubObject(YulString(_path));
```
chriseth: `pathToSubObject` was written to be used during code generation. Because of that, it directly returns code-generation internal IDs. What we would need here is an actual Object. I think it would be better to use a separate path resolution function - maybe there are ways to share code.

test/tools/yulopti.cpp
```diff
@@ -171,28 +184,118 @@ class YulOpti
 		}
 	}
 
+	void objectApply(function<void(Object&)> _fn)
+	{
+		vector<shared_ptr<Object>> stack;
```
chriseth: I think it's totally fine to use the C++ machine stack.

test/tools/yulopti.cpp
```diff
@@ -171,28 +184,118 @@ class YulOpti
...
+					stack.push_back(subObject);
+			}
+
+			_fn(*object);
+		}
+	}
+
+	void analyze(ErrorReporter& errorReporter)
+	{
+		objectApply([&](Object& object) -> void {
```
chriseth: Do you need this explicit `-> void`?

chriseth:
This looks good in general, but I think we are getting some complications due to the feature that we can select a single sub-object to run on. Is this actually needed? Also I didn't take a look at the output yet, but could you check that if the object is only a single object with only code, that it prints the code without the object?

libyul/ObjectParser.cpp
dflupu: Got it, I've removed this line and the use of `pathToSubObject`. The value I had set was required for `pathToSubObject` to work.

test/tools/yulopti.cpp
dflupu: I've changed this to `An <object> flag may be provided, specifying a dotted path to an object in the input.`. Hope that's better.

test/tools/yulopti.cpp
dflupu: Removed the use of `pathToSubObject` and simply did the search in `YulOpti::getSubObject`.

test/tools/yulopti.cpp
dflupu: I assumed this means making the function recursive. If so, done.

test/tools/yulopti.cpp
dflupu: Fixed

dflupu:
> This looks good in general, but I think we are getting some complications due to the feature that we can select a single sub-object to run on. Is this actually needed? Also I didn't take a look at the output yet, but could you check that if the object is only a single object with only code, that it prints the code without the object?

Thanks for the review! I've addressed your comments and made sure that the output for a code block is still a code block. I can't comment on the utility of the `<object>` flag myself. The author of the issue might be able to. At this point the implementation for that feature is a pretty neat chunk of code that doesn't call anything weird.

leonardoalt:
This is still waiting for reviews. Any team member that sees this message is welcome to review it.

test/tools/yulopti.cpp
```diff
@@ -87,32 +88,68 @@ class YulOpti
...
+
+		yulAssert(
+			boost::algorithm::starts_with(_qualifiedPath, _object->name.str() + "."),
+			"Assembly object not found."
+		);
+
+		const auto subObjectPath = _qualifiedPath.substr(_object->name.str().length() + 1);
+		const auto subObjectName = subObjectPath.substr(0, subObjectPath.find_first_of('.'));
+
+		auto subObjectIt = find_if(
```
matheusaaguiar: Maybe we could use `ranges::find_if` instead of `std::find_if`.

test/tools/yulopti.cpp
```diff
@@ -87,32 +88,68 @@ class YulOpti
...
+			_object->subObjects.begin(),
+			_object->subObjects.end(),
+			[subObjectName](auto const& subObject) { return subObject->name.str() == subObjectName; }
```
matheusaaguiar:
```suggestion
			[subObjectName](auto const& _subObject) { return _subObject->name.str() == subObjectName; }
```
Our code style is to use underscore for parameter variable names.

test/tools/yulopti.cpp
```diff
@@ -87,32 +88,68 @@ class YulOpti
...
+		);
+
+		return getSubObject(subObject, subObjectPath);
+	}
+
+	void parse(string const& _input, string const& _objectPath)
 	{
 		ErrorList errors;
 		ErrorReporter errorReporter(errors);
 		CharStream _charStream(_input, "");
```
matheusaaguiar: I am aware that this was not introduced by you, but we could seize the opportunity to conform the var name to our code style, i.e, remove the underscore prefix.
@ekpyron, I saw this was introduced in commit 95c973d... maybe it was overlooked at the time, right?

test/tools/yulopti.cpp
```diff
@@ -171,28 +208,118 @@ class YulOpti
 		}
 	}
 
+	void objectApply(shared_ptr<Object> _object, function<void(Object&)> _fn)
```
matheusaaguiar: Could we come up with a more descriptive name for this function? Apply what to object?

test/tools/yulopti.cpp
```diff
@@ -171,28 +208,118 @@ class YulOpti
...
+			auto subObject = dynamic_pointer_cast<Object>(subObjectNode);
+
+			if (subObject != nullptr)
+				objectApply(subObject, _fn);
+		}
+
+		_fn(*_object);
+	}
+
+	void analyze(ErrorReporter& errorReporter)
```
matheusaaguiar: I think that the name of this function can be more explicit too. It is not clear what it is analyzing. Is is it the errors? Or the object source code?
Also, coding style.
```suggestion
	void analyze(ErrorReporter& _errorReporter)
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +208,118 @@ class YulOpti
...
+	{
+		objectApply(m_object, [&](Object& object) {
```
matheusaaguiar: Maybe it is just me, but I think that the definition of the lambda expression here makes it a little bit hard to read, since it is fairly long. If you think it is a good idea, we could assign it to a local var before and then pass it to `objectApply` so it is more readable.
Other option, since there are breaking lines, could be to break at opening parenthesis of `objectApply`:
```suggestion
		objectApply(
			m_object,
			[&](Object& _object)
			{
```
And do we need `[&]` ? Nothing is being captured from outer scope ...
Also, coding style (underscore for param var name).

test/tools/yulopti.cpp
```diff
@@ -171,28 +208,118 @@ class YulOpti
...
+			yulAssert(success && !errorReporter.hasErrors(), "Invalid assembly/yul code.");
+		});
+	}
+
 	void disambiguate()
 	{
-		*m_ast = std::get<yul::Block>(Disambiguator(m_dialect, *m_analysisInfo)(*m_ast));
-		m_analysisInfo.reset();
-		m_nameDispenser.reset(*m_ast);
+		objectApply(m_object, [&](Object& object) {
```
matheusaaguiar: Same as before regarding breaking lines, capture and coding style. 
Please make sure to also check the next set of functions that have this lambda as a parameter of `objectApply`.

test/tools/yulopti.cpp
```diff
@@ -214,47 +341,50 @@ class YulOpti
 					case '#':
 						return;
 					case ',':
-						VarNameCleaner::run(m_context, *m_ast);
+						runVarNameCleaner();
 						// VarNameCleaner destroys the unique names guarantee of the disambiguator.
 						disambiguated = false;
 						break;
 					case ';':
 					{
```
matheusaaguiar: I guess braces can be removed here.

test/tools/yulopti.cpp
```diff
@@ -214,47 +341,50 @@ class YulOpti
...
-						StackCompressor::run(m_dialect, obj, true, 16);
+						runStackCompressor();
 						break;
 					}
 					default:
-						OptimiserSuite{m_context}.runSequence(
-							std::string_view(&option, 1),
-							*m_ast
-						);
+						runSequence(std::string_view(&option, 1));
```
matheusaaguiar:
```suggestion
						runSequence(string_view(&option, 1));
```
We are already using namespace `std`.

matheusaaguiar (commented):
Some suggestions and coding style fixes.

nikola-matic:
@dflupu heya, do you plan on continuing work on this? We're actively working on reducing the number of open PRs, and I'd like to know where we stand with this, so we can decide whether to have someone from the team to take it over, or to close it.

dflupu:
> @dflupu heya, do you plan on continuing work on this? We're actively working on reducing the number of open PRs, and I'd like to know where we stand with this, so we can decide whether to have someone from the team to take it over, or to close it.

I'll address the comments by the end of the week.

test/tools/yulopti.cpp
dflupu: Done

test/tools/yulopti.cpp
dflupu: Done

test/tools/yulopti.cpp
dflupu: Done

test/tools/yulopti.cpp
dflupu: Renamed to `objectApplyFn`. Not sure if that's an improvement or not. Couldn't come up with anything more appropriate since it's a pretty generic function.

test/tools/yulopti.cpp
dflupu: Renamed to `runCodeAnalyzer`

test/tools/yulopti.cpp
dflupu: Went with the line split since the local var would've required some kind of name that'd be a duplicate of something.

test/tools/yulopti.cpp
dflupu: Fixed

test/tools/yulopti.cpp
dflupu: Fixed

test/tools/yulopti.cpp
dflupu: Fixed

dflupu:
I think I've covered everything. Feel free to review again.

test/tools/yulopti.cpp
matheusaaguiar: I think `objectApplyFunction` would be better.

dflupu: Renamed to `objectApplyFunction`

nikola-matic:
Heya @dflupu, thanks for picking this up again :)
A couple of pointers for future reference - when someone leaves a comment, and the fix is trivial and doesn't require a discussion (e.g. rename of a variable name), please don't leave a comment in the thread (i.e. done, fixed, etc.) as it this will send a ton of notifications and is generally annoying; in such cases, feel free to just click on resolve conversation.
Second - I've not taken a detailed look, but you've got a couple of strange failures, which will most likely be resolved if you rebase against the fresh upstream develop (e.g. the `b_archlinux` is failing due to a Z3 dependency version which was bumped some time ago).

dflupu:
@nikola-matic Okay! I've rebased it.

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,72 @@ class YulOpti
...
+			boost::algorithm::starts_with(_qualifiedPath, _object->name.str() + "."),
+			"Assembly object not found."
+		);
+
+		const auto subObjectPath = _qualifiedPath.substr(_object->name.str().length() + 1);
+		const auto subObjectName = subObjectPath.substr(0, subObjectPath.find_first_of('.'));
+
+		auto subObjectIt = ranges::find_if(
+			_object->subObjects,
+			[subObjectName](auto const& _subObject) { return _subObject->name.str() == subObjectName; }
```
matheusaaguiar:
```suggestion
			[&subObjectName](auto const& _subObject) { return _subObject->name.str() == subObjectName; }
```

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,72 @@ class YulOpti
...
+		if (_qualifiedPath.empty() || _qualifiedPath == _object->name.str())
+			return _object;
+
+		yulAssert(
+			boost::algorithm::starts_with(_qualifiedPath, _object->name.str() + "."),
+			"Assembly object not found."
+		);
+
+		const auto subObjectPath = _qualifiedPath.substr(_object->name.str().length() + 1);
+		const auto subObjectName = subObjectPath.substr(0, subObjectPath.find_first_of('.'));
```
matheusaaguiar:
```suggestion
		const string subObjectPath = _qualifiedPath.substr(_object->name.str().length() + 1);
		const string subObjectName = subObjectPath.substr(0, subObjectPath.find_first_of('.'));
```
I don't see any benefit of using `auto` here.

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,72 @@ class YulOpti
...
+
+			if (!m_inputWasCodeBlock && scanner->currentToken() == Token::LBrace)
+				m_inputWasCodeBlock = true;
+
+			auto content = parser.parse(scanner, false);
+
+			if (content != nullptr)
+				m_object = getSubObject(content, _objectPath);
+
+			if (!m_object || !errorReporter.errors().empty())
```
matheusaaguiar:
```suggestion
			shared_ptr<Object> object = parser.parse(scanner, false);
			
			if (!object || !errorReporter.errors().empty())
```
I think it is better to check this way. If there is a parsing error, either with a `nullptr` being returned or errors present, it should be detected before continuing. 
Since `getSubObject` fails the `YulAssert` at line `117` when `nullptr` is found, it should be fine to have just the following assignment after line `151`:
``` 
m_object = getSubObject(object, _objectPath);
``` 
We could even put an assertion to guarantee `m_object` has value, but I guess it would be redundant with the one already present inside the function.

test/tools/yulopti.cpp
```diff
@@ -171,28 +208,138 @@ class YulOpti
 		}
 	}
 
-	void disambiguate()
+	void objectApplyFunction(shared_ptr<Object> _object, function<void(Object&)> _fn)
```
matheusaaguiar:
```suggestion
	void applyFunctionToObject(shared_ptr<Object> _object, function<void(Object&)> _function)
```
I think we should definitely expand the `fn` abbreviation, and maybe the method name looks better this way?
I guess we could also consider being even more specific and using `applyFunctionToObjectAndSubObjects`... but maybe then we cross the line of readability/simplicity.

test/tools/yulopti.cpp
```diff
@@ -214,47 +361,48 @@ class YulOpti
...
+				runCodeAnalyzer(errorReporter);
 			}
 			catch (...)
 			{
 				cerr << endl << "Exception during optimiser step:" << endl;
 				cerr << boost::current_exception_diagnostic_information() << endl;
 			}
 			cout << "----------------------" << endl;
-			cout << _source << endl;
+			printObject();
```
matheusaaguiar: @cameel I think this is more a design question about the behavior of the tool. Now that it can read and work with nested objects, should their code also be printed?

cameel: Yes. I haven't reviewed the changes here but I'm assuming that now the tool can work both with a Yul block (as it used to) and with a full object. In that case, with an object, the optimizer is also processing subobjects and I'd expect to see them in the output. Stripping them may give you an object that does not even compile (e.g. when you have `datacopy()` that refers to a subobject). If the user does not want to see subobjects, he can pass in a block or remove them from the input.

We may also want to add an option to ignore nested objects but I wouldn't add it until it turns out that this is something that's actually useful in practice.

test/tools/yulopti.cpp
nikola-matic: Also, `const` qualifier has to come after the type as per our coding style, so in this case, `string const` instead of `const string`.

test/tools/yulopti.cpp
matheusaaguiar: @dflupu, please consider this.

dflupu: This was how I had intended it to work, but it often didn't because I didn't properly initialize `m_inputWasCodeBlock` to false. Please try running the tool again and let me know if it is printing nested objects as expected now.

matheusaaguiar: Ok, I confirmed that it now prints subobjects as well.

matheusaaguiar (dismissed):
I think it is good, just need to squash commits and rebase.

nikola-matic:
> I think it is good, just need to squash commits and rebase.

I've squashed the two fixup commits and rebase against develop. I'll give another review as well, and @matheusaaguiar, in the meantime, you can slap another approve if you're still OK with this, since the force push dismissed it.

Also, as far as squashing everything into one commit - I guess we can just use the `squash and merge` option.

test/tools/yulopti.cpp
```diff
@@ -265,10 +411,11 @@ int main(int argc, char** argv)
 		po::options_description options(
 			R"(yulopti, yul optimizer exploration tool.
 	Usage: yulopti [Options] <file>
-	Reads <file> as yul code and applies optimizer steps to it,
+	Reads <file> containing a yul object and applies optimizer steps to it,
```
nikola-matic: Why this change? What happens if no object is present in the file?

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,70 @@ class YulOpti
 		}.printErrorInformation(_errors);
 	}
 
-	void parse(string const& _input)
+	shared_ptr<Object> getSubObject(shared_ptr<Object> const& _object, string const& _qualifiedPath)
```
nikola-matic: From what I understand, the purpose of this function is to be given a top level object, and then based on the `_qualifiedPath` to find the 'sub object' recursively; however, is it possible for the top level input object to be the the one that we've specified via the `_qualifiedPath` parameter? If yes, then this function should be called `getObject`. 

In any case, it would be good to document this function to mention that it searches for the sub object recursively based on the `_qualifiedPath` parameter.

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,70 @@ class YulOpti
...
+	{
+		if (_qualifiedPath.empty() || _qualifiedPath == _object->name.str())
+			return _object;
+
+		yulAssert(
```
nikola-matic: I don't think the usage of `yulAssert` or any assert is appropriate here. `_qualifiedPath` is user provided and the user may make a typo, which is a perfectly valid scenario. Asserts should not be used for error handling. You could instead throw an exception here (although I'd argue that the user providing and incorrect object path is not exactly exceptional either), or even better, return a `nullptr` indicating that the object was not found, and then handle appropriately elsewhere. Same thing goes for the assert on line 110. The one on line 117 seems OK though.

test/tools/yulopti.cpp
cameel: Yeah, it should generally be possible to get the top-level object too.

test/tools/yulopti.cpp
matheusaaguiar: The PR enables that objects (and nested objects) can be read by the tool. But you can have just a raw yul block too.
If there is no object nor yul block present in the file, then it terminates with an exception due to parsing error.
@dflupu , I think that it should say that a raw yul block is also accepted.

test/tools/yulopti.cpp
cameel: We could add this helper to `Object`. We already have `Object::pathToSubObject()`, which overlaps some of the logic here. It could be reused if you replace the assertions with proper errors.

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,70 @@ class YulOpti
...
+			subObjectIt != _object->subObjects.end(),
+			"Assembly object not found."
+		);
+
+		auto subObject = dynamic_pointer_cast<Object>(*subObjectIt);
+
+		yulAssert(
+			subObject != nullptr,
+			"Assembly object may not contain code."
```
cameel: Currently it sounds as if it's not allowed to contain code.
```suggestion
			"Assembly object does not contain code."
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +206,138 @@ class YulOpti
 		}
 	}
 
-	void disambiguate()
+	void applyFunctionToObject(shared_ptr<Object> _object, function<void(Object&)> _function)
```
cameel:
```suggestion
	void applyFunctionToObjectAndSubobjects(shared_ptr<Object> _object, function<void(Object&)> _function)
```

cameel:
```suggestion
	void applyFunctionToObject(Object& _object, function<void(Object&)> _function)
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +206,138 @@ class YulOpti
...
+			OptimiserStepContext{
+				m_dialect,
+				*m_nameDispenser,
+				m_reservedIdentifiers,
+				solidity::frontend::OptimiserSettings::standard().expectedExecutionsPerDeployment
+			}
+		);
+	}
+
+	void runSteps(string _source, string _objectPath, string _steps)
```
cameel:
```suggestion
	void runSteps(string const& _source, string const& _objectPath, string const& _steps)
```

test/tools/yulopti.cpp
```diff
@@ -331,20 +485,25 @@ int main(int argc, char** argv)
...
 
+		if (arguments.count("object"))
+			objectPath = arguments["object"].as<string>();
+
 		YulOpti yulOpti;
 		bool disambiguated = false;
+
 		if (!nonInteractive)
-			cout << input << endl;
+			yulOpti.parseAndPrint(input, objectPath);
```
cameel: This helper is used only once and I think it's actually clearer to have it directly here since it's so simple:
```suggestion
			{
				yulOpti.parse(input, objectPath);
				yulOpti.printObject();
			}
```

test/tools/yulopti.cpp
```diff
@@ -331,20 +485,25 @@ int main(int argc, char** argv)
 			return 1;
 		}
 
+		if (arguments.count("object"))
+			objectPath = arguments["object"].as<string>();
```
cameel, line 461: We should report an error when this is non-empty and the input is a block, not an object.

test/tools/yulopti.cpp
```diff
@@ -284,6 +431,11 @@ int main(int argc, char** argv)
 				po::value<string>(),
 				"steps to execute non-interactively"
 			)
+			(
+				"object",
+				po::value<string>(),
```
cameel, line 436:
```suggestion
				po::value<string>()->value_name("path"),
```

test/tools/yulopti.cpp
```diff
@@ -284,6 +431,11 @@ int main(int argc, char** argv)
...
+				"path to a yul object in the input"
```
cameel:
```suggestion
				po::value<string>()->value_name("path"),
				"Dotted path to a specific Yul object in the source. "
				"If not specified, the top-level object is used. "
				"Only valid if the source actually contains objects (rather than a block of Yul code)."
```

test/tools/yulopti.cpp
```diff
@@ -265,10 +411,11 @@ int main(int argc, char** argv)
 		po::options_description options(
 			R"(yulopti, yul optimizer exploration tool.
 	Usage: yulopti [Options] <file>
-	Reads <file> as yul code and applies optimizer steps to it,
+	Reads <file> containing a yul object and applies optimizer steps to it,
 	interactively read from stdin.
 	In non-interactive mode a list of steps has to be provided.
 	If <file> is -, yul code is read from stdin and run non-interactively.
+	An <object> flag may be provided, specifying a dotted path to an object in the input.
```
cameel:
```suggestion
	An --object flag may be provided, specifying a dotted path to an object in the input.
```
But I think this is not necessary here. The option description below is already enough.

test/tools/yulopti.cpp
```diff
@@ -171,28 +206,138 @@ class YulOpti
...
+		parse(_source, _objectPath);
+		printObject();
+	}
+
+	void printObject()
+	{
+		if (!m_inputWasCodeBlock)
+			cout << m_object->toString(&m_dialect) << endl;
+		else
+			cout << AsmPrinter{m_dialect}(*m_object->code)  << endl;
```
cameel:
```suggestion
			cout << AsmPrinter{m_dialect}(*m_object->code) << endl;
```

test/tools/yulopti.cpp
```diff
@@ -214,47 +359,48 @@ class YulOpti
...
 			cout << "----------------------" << endl;
-			cout << _source << endl;
+			printObject();
 		}
 	}
 
 private:
-	shared_ptr<yul::Block> m_ast;
+	shared_ptr<yul::Object> m_object;
+	bool m_inputWasCodeBlock = false;
```
cameel: Isn't it still a code block? A block should not be converted to an object by the tool.
```suggestion
	bool m_inputIsCodeBlock = false;
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +206,138 @@ class YulOpti
...
+	void parseAndPrint(string _source, string _objectPath)
+	{
+		parse(_source, _objectPath);
+		printObject();
+	}
+
+	void printObject()
+	{
+		if (!m_inputWasCodeBlock)
+			cout << m_object->toString(&m_dialect) << endl;
```
cameel: Please assert that the object has no subobjects or data objects in this case.

EDIT: I actually meant the other case, when it's a block. I see I selected the wrong branch.

test/tools/yulopti.cpp
cameel: Now that you added `runSequence()`, both sound like they would do the same thing. Let's make the difference between them clear:
```suggestion
	void parseAndRunSteps(string _source, string _objectPath, string _steps)
```

test/tools/yulopti.cpp
```diff
@@ -87,37 +89,70 @@ class YulOpti
...
-			if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
-			{
-				cerr << "Error analyzing source." << endl;
-				printErrors(_charStream, errors);
-				throw std::runtime_error("Could not analyze source.");
+				printErrors(charStream, errors);
+				throw runtime_error("Could not parse source.");
 			}
+
+			m_object = getSubObject(object, _objectPath);
```
cameel: Don't we have to keep the whole object? If you do this, I'd expect `parse()` to fail if called multiple times (unless the nested object happens to have the same path inside it - but even then it would be a different subobject than requested).

cameel (changes requested):

test/tools/yulopti.cpp
cameel: Ah, I see we don't. We only run `parse()` once. So I guess this is fine as long as a nested object cannot refer to any objects on upper levels and I think that currently it can't.

test/tools/yulopti.cpp
```diff
@@ -171,28 +206,138 @@ class YulOpti
...
+	void runSequence(string_view _steps)
+	{
+		applyFunctionToObject(
+			m_object,
+			[&](Object& _object)
+			{
+				OptimiserSuite{*m_context}.runSequence(_steps, *_object.code);
+			}
+		);
+	}
```
cameel: I recently realized that there's a big problem here. This code ignores the distinction between *creation* and *deployed* objects (see [Specification of Yul Object](https://docs.soliditylang.org/en/latest/yul.html#specification-of-yul-object)). Normally the optimizer considers the top-level object and any objects with names not ending with `_deployed` to be creation objects. For these `expectedExecutionsPerDeployment` (i.e. `--optimize-runs`) is set to 1 and it affects the result of the optimization. This logic sits in `YulStack::optimize()` but the code here ignores it and just walks the object treating every sub-object the same way.

This wasn't a problem as long as we were working with only one object in isolation. With this PR it becomes misleading because when you pass it the whole object you'll get slightly different output than you would get from the compiler.

test/tools/yulopti.cpp
dflupu: Added to `Object` as a static function.

Not sure what kind of errors I'm supposed to throw. `Object::pathToSubObject()` uses only asserts. For now I've made it return null if the path isn't found as suggested in another comment.

test/tools/yulopti.cpp
dflupu: I've added this check in `parse()`

test/tools/yulopti.cpp
dflupu: Renamed.

The tool now uses `ObjectParser::parse` which has a special case for code blocks. That makes it possible to handle objects and code blocks the same for the most part. Are there issues with that approach?

test/tools/yulopti.cpp
dflupu: I see. Would it make sense to call `YulStack::optimize()` here instead or should I implement the same logic?

I'm assuming that the top-level object will be the top-level object from the input and never the object given by `--object`?

cameel: I'd rather avoid reimplementing all of it because we'll easily forget to update yulopti if the code diverges. On the other hand we already make assumptions about what the optimizer does so I don't think we can fully avoid that.

The part I'd like to avoid duplicating the most is the logic for determining if an object is a creation object. I think it would be acceptable to keep the current solution of walking the object on your own if you extract a common method with logic for determining it and make both yulopti and `YulStack` use it. You could put it in `Object`.

Also, I'd recommend doing this (and the refactor I suggested in https://github.com/ethereum/solidity/pull/12941#discussion_r1018349654) in a separate PR. Changes outside of yulopti (which is just a non-user-facing testing tool) have more potential to break things so would be better to have them separately and then to get more people to look specifically at them.

cameel: > I'm assuming that the top-level object will be the top-level object from the input and never the object given by `--object`?

We could do it either way - with the top being fixed or with it being the subobject one specified by `--object`- but I think the former would be more intuitive. I.e. we always optimize the whole object and `--object` works simply as a filter to show you just the one object you care about - as opposed to cutting out the subobject and optimizing it as if it was the whole thing (which you can always achieve just by modifying the file).

It would be a good idea to say that in the help string of the option.

test/tools/yulopti.cpp
```diff
@@ -171,28 +182,138 @@ class YulOpti
...
+				StackCompressor::run(m_dialect, _object, true, 16);
+			}
+		);
+	}
+
+	void printObject()
+	{
+		if (!m_inputIsCodeBlock)
+			cout << m_object->toString(&m_dialect) << endl;
+		else {
```
matheusaaguiar:
```suggestion
		else
		{
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +182,138 @@ class YulOpti
 		}
 	}
 
-	void disambiguate()
+	void applyFunctionToObjectAndSubobjects(Object& _object, function<void(Object&)> _function)
+	{
+		for (auto const& subObjectNode: _object.subObjects) {
```
matheusaaguiar:
```suggestion
		for (auto const& subObjectNode: _object.subObjects)
		{
```

test/tools/yulopti.cpp
```diff
@@ -87,37 +88,47 @@ class YulOpti
...
+				cerr << "Error parsing source." << endl;
+				printErrors(charStream, errors);
+				throw runtime_error("Could not parse source.");
+			}
+
+			m_object = Object::objectAt(object, _objectPath);
+
+			if (m_object == nullptr) {
+				throw runtime_error("Assembly object not found.");
 			}
```
matheusaaguiar:
```suggestion
			if (m_object == nullptr)
				throw runtime_error("Assembly object not found.");
```
Also, maybe we should throw something a different type such as `util::Exception` or `yul::YulException`  instead of `runtime_error`.

test/tools/yulopti.cpp
```diff
@@ -171,28 +182,138 @@ class YulOpti
...
-		parse(_source);
-		disambiguate();
-		OptimiserSuite{m_context}.runSequence(_steps, *m_ast);
-		cout << AsmPrinter{m_dialect}(*m_ast) << endl;
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				OptimiserSuite{*m_context}.runSequence(_steps, *_object.code);
+			}
```
matheusaaguiar:
```suggestion
			[&](Object& _object) { OptimiserSuite{*m_context}.runSequence(_steps, *_object.code); }
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +182,138 @@ class YulOpti
...
 
-	void runInteractive(string _source, bool _disambiguated = false)
+	void runVarNameCleaner()
 	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				VarNameCleaner::run(*m_context, *_object.code);
+			}
```
matheusaaguiar:
```suggestion
			[&](Object& _object) { VarNameCleaner::run(*m_context, *_object.code); }
```

test/tools/yulopti.cpp
```diff
@@ -171,28 +182,138 @@ class YulOpti
...
+	}
+
+	void runStackCompressor()
+	{
+		applyFunctionToObjectAndSubobjects(
+			*m_object,
+			[&](Object& _object)
+			{
+				StackCompressor::run(m_dialect, _object, true, 16);
+			}
```
matheusaaguiar:
```suggestion
			[&](Object& _object) { StackCompressor::run(m_dialect, _object, true, 16); }
```

matheusaaguiar (commented):
Mostly nitpicks about code style.

test/tools/yulopti.cpp
dflupu: I've updated the PR, adding proper handling of creation objects. For now I've duplicated the check code. I'm not sure how the check can be properly refactored into `Object` since knowing whether an object is a creation object or not requires knowing if the object is at the root of the tree. Some possible method declarations would be:
```
bool Object::isCreation(string _qualifiedPath);
bool Object::isCreation(int depth);
bool Object::isCreation(bool isRootObject);
```

Maybe just `bool Object::isDeployed()` would do?

I had done the refactor suggested in [#12941 (comment)](https://github.com/ethereum/solidity/pull/12941#discussion_r1018349654) on this PR branch previously. I can of course move it into a separate PR.

libyul/Object.h
```diff
@@ -125,6 +126,13 @@ struct Object: public ObjectNode
 
 	/// @returns the name of the special metadata data object.
 	static std::string metadataName() { return ".metadata"; }
+
+	/// Recursively searches for an Object at @param _qualifiedPath within @param _object.
+	/// @returns a shared_ptr to the Object or a nullptr if it was not found.
+	static std::shared_ptr<Object> objectAt(
+		std::shared_ptr<Object> const& _object,
+		std::string const& _qualifiedPath
```
cameel: I'd call it `qualifiedName` just `pathToSubObject()`:
```suggestion
		std::string const& _qualifiedName
```

But you have a point that it's a path. It's just that in this file by path we seem to mean specifically a vector with indexes so I'd stick to that for consistent terminology.

libyul/Object.h
```diff
@@ -125,6 +126,13 @@ struct Object: public ObjectNode
 
 	/// @returns the name of the special metadata data object.
 	static std::string metadataName() { return ".metadata"; }
+
+	/// Recursively searches for an Object at @param _qualifiedPath within @param _object.
+	/// @returns a shared_ptr to the Object or a nullptr if it was not found.
+	static std::shared_ptr<Object> objectAt(
```
cameel: Any reason not to make it a member function and call it `subObjectAt()`? You need an `Object` instance to use it anyway.

libyul/Object.cpp
```diff
@@ -144,3 +145,28 @@ vector<size_t> Object::pathToSubObject(YulString _qualifiedName) const
...
+	);
+
+	if (subObjectIt == _object->subObjects.end())
+		return nullptr;
+
+	auto subObject = dynamic_pointer_cast<Object>(*subObjectIt);
+	yulAssert(subObject, "Assembly object <" + subObject->name.str() + "> does not contain code.");
+
+	return objectAt(subObject, subObjectPath);
+}
```
cameel: I think it would be better to use `pathToSubObject()` in the implementation here. They both traverse the object tree in the same way and just return different things. Having this not construct a vector of indexes is a bit more efficient but in the grand scheme of things the difference is negligible and being sure that they won't get out of sync is more than worth it.

For example we already have a PR that extends `pathToSubObject()` to support data objects (#13707) and if your function was calling it, we'd get that functionality for free here, without having to do anything.

libyul/Object.h
cameel: Then I'd also move it next to `pathToSubObject()`.

test/tools/yulopti.cpp
```diff
@@ -91,33 +92,32 @@ class YulOpti
...
 		try
 		{
-			m_ast = yul::Parser(errorReporter, m_dialect).parse(_charStream);
-			if (!m_ast || !errorReporter.errors().empty())
+			ObjectParser parser(errorReporter, m_dialect);
+
+			auto scanner = make_shared<Scanner>(charStream);
+
+			if (!m_inputIsCodeBlock && scanner->currentToken() == Token::LBrace)
+				m_inputIsCodeBlock = true;
```
cameel: Any reason this is not updated when `m_inputIsCodeBlock` is already true?
```suggestion
			m_inputIsCodeBlock = (scanner->currentToken() == Token::LBrace);
```

I don't think it makes much difference in the current implementation (it won't randomly flip from one to the other anyway) and just assigning it every time makes the logic clearer.

test/tools/yulopti.cpp
```diff
@@ -171,28 +171,152 @@ class YulOpti
...
+				AsmAnalyzer analyzer(
+					*_object.analysisInfo,
+					_errorReporter,
+					m_dialect,
+					{},
+					_object.qualifiedDataNames()
+				);
+
+				bool success = analyzer.analyze(*_object.code);
+				yulAssert(success && !_errorReporter.hasErrors(), "Invalid assembly/yul code.");
```
cameel, line 200: Before this PR an analysis error resulted in the errors being printed with `printErrors()`. I think we should keep that (unless I'm missing something and this behavior is somehow preserved?).

It would also throw `std::runtime_error`, which is semantically different from an assertion, even though currently both are being handled the same way in `catch (...)`. We should really refactor this some day to throw a more specific exception class on validation/runtime errors and handle them differently from failed assertions.

test/tools/yulopti.cpp
```diff
@@ -265,7 +382,7 @@ int main(int argc, char** argv)
 		po::options_description options(
 			R"(yulopti, yul optimizer exploration tool.
 	Usage: yulopti [Options] <file>
-	Reads <file> as yul code and applies optimizer steps to it,
+	Reads <file> containing a either a yul object or a yul code block and applies optimizer steps to it,
```
cameel:
```suggestion
	Reads <file> containing either a yul object or a yul code block and applies optimizer steps to it,
```

test/tools/yulopti.cpp
```diff
@@ -214,47 +338,40 @@ class YulOpti
...
-						OptimiserSuite{m_context}.runSequence(
-							std::string_view(&option, 1),
-							*m_ast
-						);
+						runSequence(string_view(&option, 1));
 				}
-				_source = AsmPrinter{m_dialect}(*m_ast);
+
+				resetNameDispenser();
+				runCodeAnalyzer(errorReporter);
```
cameel: Rerunning analysis after every step is probably superfluous. Or at least, `yulopti` was not doing that before. Any reason to introduce it now?

test/tools/yulopti.cpp
cameel: It's fine as long as you don't print an object when the input was a block.

test/tools/yulopti.cpp
cameel: It uses asserts because the caller is expected to check for these things (they're basically preconditions). We have proper validations against them at other layers that are reported as compilation errors and once you get to actually using the `Object`, they should never happen.

I think it would be best to take the same approach in yulopti but if there is some specific case where it's problematic, we could raise an exception. Either one of those already available in `Exceptions.h` or a new one defined specifically for this.

test/tools/yulopti.cpp
cameel: > ```solidity
> bool Object::isCreation(bool isRootObject);
> ```

I'd go with this one. The one taking the path might actually be nicer but in some places where we need this (in particular `YulStack`), we don't have the path at hand so we'd have spend extra effort to construct it. We could also add both, one using the other under the hood, but that's optional.

> Maybe just `bool Object::isDeployed()` would do?

If you mean a method only checking for `_deployed` in the name, I think that would be a bit misleading because for the top-level object it could return `true` even if it's a creation object. I see `isDeployed` as the opposite of `isCreation`.

> I had done the refactor suggested in https://github.com/ethereum/solidity/pull/12941#discussion_r1018349654 on this PR branch previously. I can of course move it into a separate PR.

I'd really recommend moving it to `Object` and doing that in a separate PR. We have several places that have this condition hard-coded and I think it would be best if they all used a single method. To be honest, I'm not even confident that they deal with the top-level object the same way so it would be a good opportunity to check that and maybe add some tests to cover it.

[Specification of Yul Object](https://docs.soliditylang.org/en/latest/yul.html#specification-of-yul-object) does not even mention the top-level object so we could also update it in that PR.

NunoFilipeSantos:
Hi @dflupu ! 👋 are you planning to come back soon to this PR, or should we close it for now?

libyul/Object.cpp
r0qs: I proposed a solution for this, please see https://github.com/ethereum/solidity/pull/12941/commits/70c816c54bc2a4bfdfa34ae5f42dd7d6ef84a897 :)

test/tools/yulopti.cpp
r0qs: I removed it since I didn't see a reason for it there either.

libyul/Object.cpp
```diff
@@ -144,33 +145,62 @@ set<YulString> Object::qualifiedDataNames() const
...
-		_qualifiedName = YulString{_qualifiedName.str().substr(name.str().length() + 1)};
 	yulAssert(!_qualifiedName.empty(), "");
 
 	vector<string> subObjectPathComponents;
 	boost::algorithm::split(subObjectPathComponents, _qualifiedName.str(), boost::is_any_of("."));
 
-	vector<size_t> path;
-	Object const* object = this;
-	for (string const& currentSubObjectName: subObjectPathComponents)
+	for (size_t i = 0; i < subObjectPathComponents.size(); i++)
```
matheusaaguiar, line 155: why not iterators?

matheusaaguiar: Why not range-based for?

```suggestion
	for (auto const& currentSubObjectName: subObjectPathComponents)
```

libyul/Object.h
matheusaaguiar: Marked as resolved, but suggested changes were not applied.

libyul/Object.cpp
r0qs: No specific reason. Fixed :)

libyul/Object.h
```diff
@@ -116,6 +117,14 @@ struct Object: public ObjectNode
 	/// The path must not lead to a @a Data object (will throw in that case).
 	std::vector<size_t> pathToSubObject(YulString _qualifiedName) const;
 
+	/// Traverses the @a _object tree applying the function @a _visitor at every subobject
+	/// in the path given by the @a _qualifiedName.
+	static void traverseObjectTree(
+		Object const* _object,
+		YulString _qualifiedName,
+		std::function<bool(Object const*)> _visitor
+	);
```
cameel: - If something requires a single instance of the class as an argument, in most cases it should just be a member function.
- I'd call it `visitPath()`. The current name is a bit ambiguous - to me at first it sounded as if it would visit the whole tree.
- You don't say what happens when the `_visitor` returns `false`. Or when the object at the path does not exist. Or is a `Data` object. Or in which direction the path is traversed.

libyul/Object.cpp
```diff
@@ -144,33 +145,61 @@ set<YulString> Object::qualifiedDataNames() const
...
 	return path;
 }
+
+std::shared_ptr<Object> Object::objectAt(std::shared_ptr<Object> const& _object, string const& _qualifiedName)
+{
+	std::shared_ptr<Object> object = nullptr;
+	traverseObjectTree(_object.get(), YulString(_qualifiedName), [&](Object const* _obj) -> bool {
+		if (_qualifiedName.empty() || _qualifiedName == _obj->name.str())
+		{
+			object = make_shared<Object>(*_obj);
```
cameel: `make_shared()` will create a copy of the object. We have all the objects neatly organized in a tree and we don't want extra copies floating around outside, with references to the ones inside. You should avoid casually copying them. That would only be needed if you were going to modify the tree structure.

If you're going to return a `shared_ptr` to an `Object`, it should be pointing at one of the objects in the tree. But in this case, you don't really need `shared_ptr` at all. Just return a pointer or a reference. Use smart pointers in contexts where you need to express ownership. E.g. when declaring a class member that holds the reference to it long-term. Informational functions like this need not be concerned with ownership.

cameel: Also, you have `_object`, `object` and `_obj` here. That does not help understand what each of them is for. Names should be more specific.

libyul/Object.cpp
```diff
@@ -144,33 +145,61 @@ set<YulString> Object::qualifiedDataNames() const
...
+		return false;
+	});
 	return path;
 }
+
+std::shared_ptr<Object> Object::objectAt(std::shared_ptr<Object> const& _object, string const& _qualifiedName)
+{
+	std::shared_ptr<Object> object = nullptr;
+	traverseObjectTree(_object.get(), YulString(_qualifiedName), [&](Object const* _obj) -> bool {
+		if (_qualifiedName.empty() || _qualifiedName == _obj->name.str())
```
cameel: Note that `name` is unqualified. I.e. for qualified path `A.B.C` the name is `C`. I don't think this comparison will work as intended.

Also the check for empty qualified path looks like it should be handled before the traversal. Do we even have to treat it as a special case? It looks like it would work regardless.

libyul/Object.cpp
```diff
@@ -144,33 +145,61 @@ set<YulString> Object::qualifiedDataNames() const
...
 		yulAssert(!currentSubObjectName.empty(), "");
-		auto subIndexIt = object->subIndexByName.find(YulString{currentSubObjectName});
-		yulAssert(
-			subIndexIt != object->subIndexByName.end(),
-			"Assembly object <" + _qualifiedName.str() + "> not found or does not contain code."
+
+		auto subObjectIt = ranges::find_if(
+			_object->subObjects,
+			[&currentSubObjectName](auto const& _subObject) { return _subObject->name.str() == currentSubObjectName; }
 		);
```
cameel: This replaces search in a hash map with linear search through a vector. The old version was both more efficient (for large objects) and more concise.

Since this assumes that the key must be present anyway, we could even make the old version more concise by using `operator[]` (and asserting beforehand with `map::count()`) instead of `map::find()`. Then we would not have to deal with iterators here at all.

The `if` below should also be an assert.

libyul/Object.cpp
r0qs: Right! Thanks, I changed it to a pointer and renamed the other variables :)

libyul/Object.cpp
r0qs: Indeed!! Fixed ;)

libyul/Object.h
r0qs: Now it is resolved :)

libyul/Object.cpp
r0qs: Done. I added the `operator::at` instead of `operator[]` since, to keep the same semantics of the later, we would need to add a new element if it does not exists.

r0qs:
@cameel I noticed that you were reviewing this work in another branch: https://github.com/ethereum/solidity/commits/yulopti-object-support and decided to apply the suggestions there that still make sense in the current state of the PR! Thanks a lot ;)

~The only one that I didn't apply was the change suggested in this commit: https://github.com/ethereum/solidity/commit/f34171f8063d1939fd73b525ac5d76267f28b2ed because I use some `std::string` methods like `find_last_of` and `substr` to determine the name of the target subobject, so if we change the function to receive a `YulString` instead of `std::string` we would need to make conversions between them inside `subObjectAt` anyway. But it would be consistent with the other methods...so yeah, maybe we should change it indeed.~. I decided to apply it as well ;)

libyul/Object.cpp
```diff
@@ -38,6 +38,13 @@ using namespace solidity::langutil;
 using namespace solidity::util;
 using namespace solidity::yul;
 
+
+Object const* Object::at(YulString _name) const
+{
+	size_t subIndex = this->subIndexByName.at(_name);
```
nikola-matic, line 49: This can throw.

r0qs: Isn't this the expected behavior of `at` operator?

nikola-matic: Yeah. Is it handled properly if it throws?

r0qs: Yes, or at least, I thought so. But looking at the code again, all exceptions in yulopti are handled in the main try-catch loop as expected, although @cameel mentioned that we could improve on that eventually as well https://github.com/ethereum/solidity/pull/12941#discussion_r1103136596. However, as this code is now also shared by `Object::pathToSubObject` through the common `Object::visitPath` this may indeed throw in [EVMDialect](https://github.com/ethereum/solidity/blob/1b5c6f66930509d45c7b97ccd803e649c54a8702/libyul/backends/evm/EVMDialect.cpp#L222) when creating builtins. So yeah, good point. I will check again how the `pathToSubObject` used to work before this change to ensure the same behavior.

r0qs: It appears that the function would return an empty path if there is no [subObjectPathComponents](https://github.com/ethereum/solidity/blob/develop/libyul/Object.cpp#L142) and an assertion failure may occur if the  [subObject is not found](https://github.com/ethereum/solidity/blob/develop/libyul/Object.cpp#L155), which still the case with the current change. Also, there is an assertion for `object->subIndexByName.count(subObjectName)` right before the invocation of [object->at(subObjectname)](https://github.com/ethereum/solidity/pull/12941/files#diff-85786c576d9ffa5957420e25761d465d531c5e9648f3a1d3f5cb91c40e201c5eR153). Therefore, in this scenario, it should not throw an error in the `Object::at` function, instead, it should trigger the assertion failure in `visitPath`.

r0qs: Also, I just found a bug now. If you don't provide an object with the `--object` command-line flag, it should use the top-level object as default but it doesn't.

r0qs: Fixed. I added an assert in any case. You can test with:
```
# raw Yul as input:
./build/test/tools/yulopti test/libyul/yulOptimizerTests/fullSuite/medium.yul
# Yul objects as input:
./build/test/tools/yulopti test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object "A.B" test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object "A.B.C.D" test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object "B.E" test/libyul/objectCompiler/subObjectAccess.yul
```

libyul/Object.cpp
```diff
@@ -129,33 +141,64 @@ std::set<YulString> Object::qualifiedDataNames() const
...
 
 	std::vector<std::string> subObjectPathComponents;
 	boost::algorithm::split(subObjectPathComponents, _qualifiedName.str(), boost::is_any_of("."));
 
-	std::vector<size_t> path;
 	Object const* object = this;
-	for (std::string const& currentSubObjectName: subObjectPathComponents)
+
+	// Iterate over sub objects ignoring the top-level object in the path.
+	for (std::string const& currentSubObjectName: subObjectPathComponents | ranges::views::drop(1))
```
r0qs: I'm not completely sure if it is fine to drop the first path element in the context of `Object::pathToSubObject`.

r0qs: Actually it is not, because of the use of `starts_with` and `substr` there. I will fix it.

r0qs: It should be fixed now:)

libyul/Object.cpp
```diff
@@ -31,13 +31,25 @@
...
 
+
+Object const* Object::at(YulString _name) const
+{
+	yulAssert(
+		this->subIndexByName.count(_name),
+		"Assembly object <" + _name.str() + "> not found."
+	);
+	size_t subIndex = this->subIndexByName.at(_name);
+	return dynamic_cast<Object const*>(this->subObjects[subIndex].get());
```
nikola-matic, line 50:
```suggestion
	size_t subIndex = subIndexByName[_name];
	return dynamic_cast<Object const*>(subObjects[subIndex].get());
```

No need for `at` now that you've asserted that the object is indeed present and produces a valid index.

nikola-matic: Actually - keep the `this->`, but none of the members in `Object` are prepended with `m_` to indicate that they're indeed members, which is not great; but this isn't the place to fix it.

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
...
+
+	Object const* foundObject = nullptr;
+	this->visitPath(_qualifiedName, [&](Object const* _subObject) -> bool {
+		if (targetObjectName != _subObject->name)
+			return false;
+		foundObject = _subObject;
+		return true;
+	});
+
+	return foundObject;
```
nikola-matic, line 199: I find this somewhat counter intuitive, perhaps due to the naming, or due to lack of asserts; in any case, we're making sure that `_qualifiedName` is not empty, so regardless of whether we're using it as is, or as a last qualified name delimited by `.`, we're always going to have a qualified name.

Then we search for the sub-object using that name, and store into `foundObject`, but the same `foundObject` can end up being a `nullptr`. The question now is - do we always expect `_qualifiedName` to produce a valid subjobject - if yes, this should have an assert.

libyul/Object.cpp
r0qs: > No need for `at` now that you've asserted that the object is indeed present and produces a valid index.

Indeed. However, this would make the method non-const. I would prefer to keep it const, but I can make the change if necessary.

> Actually - keep the `this->`, but none of the members in `Object` are prepended with `m_` to indicate that they're indeed members, which is not great; but this isn't the place to fix it.

Yeah, I thought about fix it here, but it is better to open a new PR for that.

libyul/Object.cpp
r0qs: Note that `_qualifiedName` is not modified, and this function expects it to be in the format: `(A.*)+`, where `A` is a Yul object name. The search is performed using `_qualifiedName`, and `targetObjectName` is the name of the sub-object we are looking for. It will be `nullptr` only if the object is not found, i.e., it is not a sub-object of the top-level object.

But I don't think we always expect that the `_qualifiedName` will always contain valid sub-objects, since this is an user input. Some asserts are already in place to validate parts of it (see `Object::visitPath`), but without test coverage, I'm unsure if we are covering everything. For instance, these inputs are invalid and fail as expected:

```
./build/test/tools/yulopti --object "A." test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object ".A" test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object "." test/libyul/objectCompiler/subObjectAccess.yul
./build/test/tools/yulopti --object "..." test/libyul/objectCompiler/subObjectAccess.yul
# A is not a sub-object of B
./build/test/tools/yulopti --object "B.A" test/libyul/objectCompiler/subObjectAccess.yul
```

r0qs: Should I validate it with a regex? Although it could be any valid Yul Object identifier, so not really sure if we need more validation than we already have.

libyul/Object.cpp
nikola-matic: > Indeed. However, this would make the method non-const. I would prefer to keep it const, but I can make the change if necessary.

Ah, you're right, didn't even think about that.

libyul/Object.cpp
nikola-matic: > Should I validate it with a regex? Although it could be any valid Yul Object identifier, so not really sure if we need more validation than we already have.

Nope, no need.

nikola-matic (approved):

nikola-matic:
@cameel there's still a CR from your side on this PR (from Nov 9th).

cameel:
I'll give it a final look tomorrow.

libyul/Object.cpp
cameel: I'd actually remove `this->`. We never use it explicitly and it was always pointed out in my PRs.

cameel: Also, I don't think `at()` is a good name if this method ignores `Data` objects. It's easy to take it for a general accessor like `map::at()` and assume that the index is empty when `nullptr` is returned.

The fact that this happens is not even documented in its docstring.

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
...
-			subIndexIt != object->subIndexByName.end(),
-			"Assembly object <" + _qualifiedName.str() + "> not found or does not contain code."
-		);
-		object = dynamic_cast<Object const*>(object->subObjects[subIndexIt->second].get());
-		yulAssert(object, "Assembly object <" + _qualifiedName.str() + "> not found or does not contain code.");
-		yulAssert(object->subId != std::numeric_limits<size_t>::max(), "");
-		path.push_back({object->subId});
+
+		YulString subObjectName = YulString{currentSubObjectName};
+		object = object->at(subObjectName);
```
cameel, line 159: Just a general remark about naming: please avoid renaming things in a way that does not make it clear what the distinction between them really is. Here both `subObjectName` and `currentSubObjectName` are pretty much the same thing and the difference is not that one is not "current". It's that one is a `YulString`. So something like `currentSubObjectNameYul` would be a much clearer name for the former. But in this situation I find that `YulString{currentSubObjectName}` is not much longer so I'd just use it directly:

```suggestion
		object = object->at(YulString{currentSubObjectName});
```

libyul/Object.h
```diff
@@ -116,6 +117,21 @@ struct Object: public ObjectNode
...
 	std::vector<size_t> pathToSubObject(YulString _qualifiedName) const;
 
+	/// Searches for a subobject at @param _qualifiedName within the current object.
+	/// @returns a pointer to the subobject or a nullptr if it was not found.
+	Object const* subObjectAt(YulString _qualifiedName);
+
+	/// Visits all subobjects in the path given by the @a _qualifiedName
+	/// of the current object applying the function @a _visitor.
+	/// If @a _visitor returns `true` the visiting stops, otherwise,
+	/// it continues until the last subobject in the path is reached.
```
cameel, line 127: I'd make it return `false` to stop. That's how it works in AST visitors so doing the opposite here will only make mistakes more likely.

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
...
+		if (object && _visitor(object))
+			break;
```
cameel, line 161: You continue the loop when `object` is null and the next cycle may try to access it. This will segfault.

`test.yul`
```yul
object "A" {
    code {}
    data "B" ""
}
```
```
yulopti --input-file test.yul --object A.B.C
```
```
Segmentation fault (core dumped)
```

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
 	return qualifiedNames;
 }
 
-std::vector<size_t> Object::pathToSubObject(YulString _qualifiedName) const
+void Object::visitPath(YulString _qualifiedName, std::function<bool(Object const*)> const& _visitor) const
 {
-	yulAssert(_qualifiedName != name, "");
-	yulAssert(subIndexByName.count(name) == 0, "");
```
cameel, line 135: These asserts are still relevant here and should not be moved.

The second one checks an assumption for the `if` below that strips `name` from `_qualifiedName`. The assumption is that an object never contains a sub-object of the same name. If that was not true, such paths would be ambiguous.

The first one ensures that we'll visit at least one subobject, just like `yulAssert(!_qualifiedName.empty(), "");` below.

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
...
+	// the target object name is the last element in the path given by `_qualifiedName`.
+	YulString targetObjectName = _qualifiedName;
+	size_t targetObjectPos = _qualifiedName.str().find_last_of(".");
+	if (targetObjectPos != std::string::npos)
+		targetObjectName = YulString(_qualifiedName.str().substr(targetObjectPos + 1));
+
+	Object const* foundObject = nullptr;
+	this->visitPath(_qualifiedName, [&](Object const* _subObject) -> bool {
+		if (targetObjectName != _subObject->name)
+			return false;
```
cameel, line 194: I don't think this is correct. What if `targetObjectName` appears earlier on the path? E.g. for `A.B.A.B` the function would give you the object corresponding to `A.B`.

You should just let `visitPath()` iterate the whole `_qualifiedName`, just unconditionally updating `foundObject`. When it stops, `foundObject` will be the last object on the path.

libyul/Object.h
```diff
@@ -116,6 +117,21 @@ struct Object: public ObjectNode
 	/// The path must not lead to a @a Data object (will throw in that case).
 	std::vector<size_t> pathToSubObject(YulString _qualifiedName) const;
 
+	/// Searches for a subobject at @param _qualifiedName within the current object.
+	/// @returns a pointer to the subobject or a nullptr if it was not found.
```
cameel, line 121: Please clarify what happens when the path leads to a `Data` object.

Also, it's not entirely true that it returns null if not found. The function has a check against that but it will trigger an assert in `Object::at()` before it even gets to that check.

libyul/Object.h
cameel: Also, the docstring does not specify what happens when the path leads to an object that does not exist or is a `Data` object.

libyul/Object.cpp
cameel: Having asserts instead of validations here is fine for now - in the compiler we have those validations already elsewhere and yulopti is an internal tool. I mean, I'd include a validation if I were writing it myself, but not having one is not a big deal as long as it has an assert with a good message.

Just, if we're using asserts, I'd try to make it somewhat consistent with how other functions behave. Now `subObjectAt()`, `visitPath()`, and `pathToSubObject()` each has a different idea of what to do on objects that do not exist and on `Data` objects. Since `pathToSubObject()` already existed and we use it elsewhere in the compiler, IMO it would be sensible if all of them followed its behavior (i.e. just assert against all of these cases).

test/tools/yulopti.cpp
```diff
@@ -90,33 +91,29 @@ class YulOpti
...
-				errorReporter,
-				m_dialect
-			);
-			if (!analyzer.analyze(*m_ast) || !errorReporter.errors().empty())
-			{
-				cerr << "Error analyzing source." << endl;
-				printErrors(_charStream, errors);
-				throw std::runtime_error("Could not analyze source.");
+				printErrors(charStream, errors);
+				solThrow(Exception, "Could not parse source.");
```
cameel, line 108: The exception goes from here to the `catch` block, which will print the errors again. We should print them only once.

Also, `runCodeAnalyzer()` should use `solThrow()` too instead of asserting that it expects no errors (which is not true).

cameel (changes requested):
There are still significant problems in the changes to `Object`. I found a case where it will segfault and also the new functions are pretty inconsistent in how they behave in corner cases, which may easily lead to future bugs and is a regression compared to the current situation.

I haven't reviewed the yulopti part very thoroughly.

libyul/Object.cpp
```diff
@@ -129,33 +141,60 @@ std::set<YulString> Object::qualifiedDataNames() const
...
+{
+	std::vector<size_t> path;
+	yulAssert(_qualifiedName != name, "");
+	yulAssert(subIndexByName.count(name) == 0, "");
+
+	this->visitPath(_qualifiedName, [&](Object const* _object) -> bool {
+		yulAssert(_object->subId != std::numeric_limits<size_t>::max(), "");
+		path.push_back(_object->subId);
+		return false;
+	});
```
cameel, line 175: By the way `visitPath()` would be more idiomatic if it used iterators instead of callbacks. It would let us use ranges:

```
visitPath(_qualifiedName) | transform(objectToSubId) | to<vector>();
```

This would make most of these new helpers obsolete. I'd not do it in this PR though. We want to be finally done with it :)

test/tools/yulopti.cpp
dharmendrarawal: Renamed to runCodeAnalyzer

PR #12945, opened 2022-04-19
Encode path conditions

chriseth:
The goal of this PR is to find out how good an SMT-based optimizer step can be:
It uses z3 as a backend, so we can assume that it can solve queries at least as fast as a hand-written LP-solver can do it.

The current state is that this PR did not reduce the gas cost by much. This means we either need to refine the encoding of yul into SMT queries, we need to find better conditions or that the code is already written in a way where such a component cannot improve it by much.


cameel:
> Those are being rewritten in Rust anyway and are not part of the roadmap, so my guess would be close

Closing.

PR #12948, opened 2022-04-20
Get rid of MSVC message "The contents of <span> are available only with C++20 or later."

a3d4:
MSVC `<span>` header contains
```
#if !_HAS_CXX20
#pragma message("The contents of <span> are available only with C++20 or later.")
. . . .
```
and shows the message if the header gets compiled with earlier version of C++.

`range-v3` includes `<span>` if it is available (e.g. in `range/v3/range/concepts.h`):
```
#ifdef __has_include
#if __has_include(<span>)
#include <span>
#endif
```

That leads to 204 messages during MSVC compilation.

`range-v3` [has a fix](https://github.com/ericniebler/range-v3/commit/2a3f9cb999), but it is not in the release. This PR applies corresponding patch.

I'm not very happy introducing a new `patches` directory, and not fully convinced it worth troubles. But a repeating compilation message is also annoying.

### DIFF
--- a/scripts/check_style.sh
+++ b/scripts/check_style.sh
@@ -6,6 +6,7 @@ ERROR_LOG="$(mktemp -t check_style_XXXXXX.log)"
 
 EXCLUDE_FILES=(
     "libsolutil/picosha2.h"
+    "patches/avoid-including-empty-span-header-when-using-msvc.patch"
     "test/cmdlineTests/strict_asm_only_cr/input.yul"
     "test/cmdlineTests/strict_asm_only_cr/err"
     "test/libsolutil/UTF8.cpp"
--- a/cmake/range-v3.cmake
+++ b/cmake/range-v3.cmake
@@ -6,16 +6,26 @@ else()
     set(RANGE_V3_CMAKE_COMMAND ${CMAKE_COMMAND})
 endif()
 
+# MSVC reports "The contents of <span> are available only with C++20 or later." 200+ times.
+# A fix (https://github.com/ericniebler/range-v3/commit/2a3f9cb999) is available,
+# but not included in the latest release.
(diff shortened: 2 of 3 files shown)


cameel:
> `range-v3` [has a fix](https://github.com/ericniebler/range-v3/commit/2a3f9cb999), but it is not in the release. This PR applies corresponding patch.

Looks like there was a release just before the fix went into `master` (August 2020) and then releases stopped. It's about time for a new one and there's even an issue asking for that: https://github.com/ericniebler/range-v3/issues/1694.

The patch is not that big though so I actually would not mind it that much as long as we'd be able to remove it soon.

scripts/check_style.sh
```diff
@@ -6,6 +6,7 @@ ERROR_LOG="$(mktemp -t check_style_XXXXXX.log)"
 
 EXCLUDE_FILES=(
     "libsolutil/picosha2.h"
+    "patches/avoid-including-empty-span-header-when-using-msvc.patch"
```
a3d4, line 9: Because the .patch file contains trailing spaces.

a3d4:
@cameel @christianparpart I rebased. Note however that there is recent activity in the linked task. We should keep monitoring it.

a3d4:
@cameel @christianparpart 
New version of `range-v3` is released. Perhaps we should update and if the message `"The contents of <span> are available only with C++20 or later."` does not appear any longer, close this PR.

cameel:
Yeah, we definitely should update it! Created an issue: #13242.

cameel:
Ranges-v3 updated (#13242) so I'm closing this. Let me know if it turns out that this did not really fix the problem but hopefully that was all we needed.

PR #12902, opened 2022-04-05
Fix MSVC Debug crash

a3d4:
After 44093f2ed63782e02ee3611e701c92b2bdf9bd0c `soltest` reports corrupted stack in in MSVC **Debug**.

The commit looks OK and it is likely a MSVC issue. The following code

```
template<typename A, typename B> struct S : A, B{};
template<typename A, typename B> S(A, B)->S<A, B>;

int main()
{
    S s{ [](){}, [](){} };
}
```
causes an error "_Run-Time Check Failure #2 - Stack around the variable 's' was corrupted._" for **C++17** **Debug**. Observed in **VS2019** and **VS2022**. It does not occur for **C++20**, and it does not occur in **Release**.

PR #12950, opened 2022-04-20
Refix MSVC Debug crash

a3d4:
This PR addresses the same issue as #12902, but in another way. It makes sure sure that empty base optimization is enabled for `struct GenericVisitor`.

That workaround was [suggested by MS](https://developercommunity.visualstudio.com/t/10005513#T-N10018465).

### DIFF
--- a/libsolutil/Visitor.h
+++ b/libsolutil/Visitor.h
@@ -56,6 +56,17 @@ struct VisitorFallback<R> { template<typename T> R operator()(T&&) const { retur
 template<>
 struct VisitorFallback<> { template<typename T> void operator()(T&&) const {} };
 
-template <typename... Visitors> struct GenericVisitor: Visitors... { using Visitors::operator()...; };
+// MSVC. Empty base class optimization does not happen in some scenarios.
+// Enforcing it with __declspec(empty_bases) avoids MSVC Debug test crash
+// (Run-Time Check Failure #2 - Stack around the variable '....' was corrupted).
+// See https://docs.microsoft.com/en-us/cpp/cpp/empty-bases,
+//     https://developercommunity.visualstudio.com/t/10005513.
+#if defined(_MSC_VER)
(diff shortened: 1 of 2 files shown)


a3d4:
@ekpyron @christianparpart FYI :).

libsolutil/Visitor.h
```diff
@@ -56,6 +56,16 @@ struct VisitorFallback<R> { template<typename T> R operator()(T&&) const { retur
 template<>
 struct VisitorFallback<> { template<typename T> void operator()(T&&) const {} };
 
-template <typename... Visitors> struct GenericVisitor: Visitors... { using Visitors::operator()...; };
+// MSVC. Empty base class optimization does not happen in some scenarios
+// (https://docs.microsoft.com/en-us/cpp/cpp/empty-bases).
+// Enforcing it with __declspec(empty_bases) avoids MSVC C++17 Debug test crash
+// (Run-Time Check Failure #2 - Stack around the variable '....' was corrupted).
```
christianparpart, line 61: Sounds like you investigated a bit into this. Did you try to recreate a minimized source version that can still get it to corrupt the stack and yet fix it via `__declspec(empty_bases)`? If so, it might be worth reporting the bug to MSVC and link that bug report here. What do you think?

a3d4: I could reproduce the issue with the following code:

```
template<typename A, typename B> struct S : A, B{};
template<typename A, typename B> S(A, B)->S<A, B>;

int main()
{
    S s{ [](){}, [](){} };
}
```

It is reported, see https://developercommunity.visualstudio.com/t/10005513 (and in particular https://developercommunity.visualstudio.com/t/10005513#T-N10018465).

christianparpart (approved):
LGTM :)

libsolutil/Visitor.h
christianparpart: Thanks :)

chriseth (approved):

PR #12952, opened 2022-04-22
Add `mypy` checks to CI

### LINKED ISSUE #12857: Add a mypy job to CI and fix type errors in our Python code
## Abstract
We need a CI job that runs [mypy](https://github.com/python/mypy) on every PR and reports type errors in Python code.

## Motivation
Currently type annotations in our Python scripts are not checked at all. This is because the interpreter does not enforce them and an external tool is needed. mypy is currently the most popular one (though there are some others and we could use one of them if there are good arguments for doing so, see [ethanhs/python-typecheckers](https://github.com/ethanhs/python-typecheckers)). 

We already have CI jobs for [running Python unit tests](https://github.com/ethereum/solidity/blob/develop/.circleci/config.yml#L724-L741) and [pylint](https://github.com/ethereum/solidity/blob/develop/.circleci/config.yml#L661-L675) on Python code so there's no reason not to have mypy there as well.

## Specification
- Add a mypy config next to our [`pylintrc`](https://github.com/ethereum/solidity/blob/develop/scripts/pylintrc).
- Add a wrapper script in `scripts/` for easily running mypy with our config locally on all Python scripts in the repo..
- Run mypy on all existing Python scripts and fix any errors it finds. If there are any false-positives, you can disable them. For ease of review, the fixes should be a separate PR.
- Add a CI job that runs our wrapper script.

## Backwards Compatibility
No backwards-compatibility concerns here.

gurukamath:
closes #12857 

Added jobs to perform type annotation checks of python scripts in CI. Updated the current scripts to to conform to the checks by using the generic `Any` type.

All the fixes are in a separate commit for ease of review.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -680,6 +680,21 @@ jobs:
           command: ./scripts/pylint_all.py
       - gitter_notify_failure_unless_pr
 
+  chk_mypy:
+    <<: *base_buildpack_focal_small
+    steps:
+      - checkout
+      - run:
+          name: Install pip
+          command: apt -q update && apt install -y python3-pip
+      - run:
+          name: Install mypy
+          command: python3 -m pip install mypy
+      - run:
+          name: Python Scripts Type Annotations
+          command: ./scripts/mypy_all.py
+      - gitter_notify_failure_unless_pr
+
   chk_antlr_grammar:
     <<: *base_buildpack_focal_small
     steps:
@@ -1419,6 +1434,7 @@ workflows:
       - chk_buglist: *workflow_trigger_on_tags
       - chk_proofs: *workflow_trigger_on_tags
       - chk_pylint: *workflow_trigger_on_tags
+      - chk_mypy: *workflow_trigger_on_tags
       - chk_errorcodes: *workflow_trigger_on_tags
       - chk_antlr_grammar: *workflow_trigger_on_tags
       - chk_docs_pragma_min_version: *workflow_trigger_on_tags
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,6 +17,7 @@ import sys
 import os
 import re
 
+from typing import Any
 from pygments_lexer_solidity import SolidityLexer, YulLexer
 
 # If extensions (or modules to document with autodoc) are in another directory,
@@ -27,7 +28,7 @@ ROOT_PATH = os.path.dirname(os.path.realpath(__file__))
 
 sys.path.insert(0, os.path.join(ROOT_PATH, 'ext'))
 
-def setup(sphinx):
+def setup(sphinx: Any) -> None:
     sphinx.add_lexer('Solidity', SolidityLexer)
     sphinx.add_lexer('Yul', YulLexer)
 
@@ -71,7 +72,7 @@ project_copyright = '2016-2021, Ethereum'
 #
 # The short X.Y version.
 with open('../CMakeLists.txt', 'r', encoding='utf8') as f:
-    version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1)
+    version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1) # type: ignore
 # The full version, including alpha/beta/rc tags.
 if not os.path.isfile('../prerelease.txt') or os.path.getsize('../prerelease.txt') == 0:
     release = version
@@ -237,7 +238,7 @@ latex_elements = {
 
 # Additional stuff for the LaTeX preamble.
 #'preamble': '',
-}
+} # type: Any
 
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
@@ -272,7 +273,7 @@ latex_documents = [
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
 man_pages = [
-]
+] # type: Any
 
 # If true, show URL addresses after external links.
 #man_show_urls = False
@@ -284,7 +285,7 @@ man_pages = [
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-]
+] # type: Any
 
 # Documents to append as an appendix to all manuals.
 #texinfo_appendices = []
--- a/scripts/mypy_all.py
+++ b/scripts/mypy_all.py
@@ -1,0 +1,103 @@
+#! /usr/bin/env python3
+
+"""
+Runs mypy on all Python files in project directories known to contain Python scripts.
+"""
+
+from argparse import ArgumentParser
+from os import path, walk
+from textwrap import dedent
+import subprocess
+import sys
+
+from typing import Any
+
+PROJECT_ROOT = path.dirname(path.dirname(path.realpath(__file__)))
+MYPY_CONFIG = f"{PROJECT_ROOT}/scripts/mypy.ini"
+
+SGR_INFO = "\033[1;32m"
+SGR_CLEAR = "\033[0m"
+
+def mypy_all_filenames(dev_mode: Any, rootdirs: Any) -> bool:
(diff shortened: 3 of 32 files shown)


.circleci/config.yml
```diff
@@ -680,6 +680,21 @@ jobs:
           command: ./scripts/pylint_all.py
       - gitter_notify_failure_unless_pr
 
+  chk_mypy:
```
christianparpart, line 683:
```suggestion
  chk_python_types:
```
Given that this name will be shown up in the UI, it should be probably slightly more meaningful to those not knowing about "mypy" but trying to avoid too long names (sadly), I think that the above proposal is a good sweet spot. What do you think?

.circleci/config.yml
```diff
@@ -1419,6 +1434,7 @@ workflows:
       - chk_buglist: *workflow_trigger_on_tags
       - chk_proofs: *workflow_trigger_on_tags
       - chk_pylint: *workflow_trigger_on_tags
+      - chk_mypy: *workflow_trigger_on_tags
```
christianparpart, line 1437: if line above gets changed, this one needs to be adapted.

test/lsp.py
```diff
@@ -256,7 +256,7 @@ def main(self) -> int:
 
         return min(max(self.test_counter.failed, self.assertion_counter.failed), 127)
 
-    def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root=True):
+    def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root: Any=True) -> None:
```
christianparpart, line 259: why `Any` and not `bool` here? Or did you simply walk through all complains and added `Any` to it? Which is fine, I'd just like to know about it as we're touching this file quite regularly here. :)

test/lsp.py
```diff
@@ -278,20 +278,20 @@ def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root=True):
...
-        }
+        } # type: Any
         if not expose_project_root:
             params['rootUri'] = None
         lsp.call_method('initialize', params)
         lsp.send_notification('initialized')
 
     # {{{ helpers
-    def get_test_file_path(self, test_case_name):
+    def get_test_file_path(self, test_case_name: Any) -> Any:
```
christianparpart, line 288:
```suggestion
    def get_test_file_path(self, test_case_name: str) -> str:
```
But I could contribute that after this PR on top as well, too. hmmm (same for the other type annotation changes in this file at least)

christianparpart (changes requested):
I agree that adding type annotations to existing non-owning python code might become very hard, so I'd probably be fine in just setting them to `Any` - as very well done already in this PR  - but leave the specialization to future touches on the specific files. In my case, I'd touch lsp.py anyways, so I'm fine in having these change done later and yet benefit from mypi PR.

In terms of change-request, probably only (or at least) have the check name changed, because it's a displayable unit and mypy isn't really self-explaining to the reader.

docs/conf.py
```diff
@@ -71,7 +72,7 @@ def setup(sphinx):
 #
 # The short X.Y version.
 with open('../CMakeLists.txt', 'r', encoding='utf8') as f:
-    version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1)
+    version = re.search('PROJECT_VERSION "([^"]+)"', f.read()).group(1) # type: ignore
```
wechman, line 75: If we suppress errors like this, most probably they won't be fixed anytime soon. So maybe it would be better to use a proper type right away: `version: Optional[re.Match] = re.search(....)`. Or, we can create a follow-up ticket to address them, which I guess is not that bad either.

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+PROJECT_ROOT = path.dirname(path.dirname(path.realpath(__file__)))
+MYPY_CONFIG = f"{PROJECT_ROOT}/scripts/mypy.ini"
+
+SGR_INFO = "\033[1;32m"
+SGR_CLEAR = "\033[0m"
+
+def mypy_all_filenames(dev_mode: Any, rootdirs: Any) -> bool:
+    """ Performs type annotation checks on all python files within given root directory (recursively).  """
+
+    BARE_COMMAND = [
```
wechman, line 24: Variable in function should be lowercase.

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+
+from argparse import ArgumentParser
+from os import path, walk
+from textwrap import dedent
+import subprocess
+import sys
+
+from typing import Any
+
+PROJECT_ROOT = path.dirname(path.dirname(path.realpath(__file__)))
```
wechman, line 15: Would be nice to use `pathlib` to deal with filesystem. Also, we can move it to the main function.

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+        "--install-types",
+        "--non-interactive"
+    ]
+
+    filenames = []
+    for rootdir in rootdirs:
+        for rootpath, _, filenames_w in walk(rootdir):
+            for filename in filenames_w:
+                if filename.endswith('.py'):
+                    filenames.append(path.join(rootpath, filename))
```
wechman, line 38: With `pathlib` it can be one line: `filenames += [str(file.resolve()) for file in Path(rootdir).glob("**/*.py") ]`

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+        "mypy",
+        f"--config-file={MYPY_CONFIG}",
+        "--namespace-packages",
+        "--explicit-package-bases",
+        "--install-types",
+        "--non-interactive"
+    ]
+
+    filenames = []
+    for rootdir in rootdirs:
```
wechman, line 34: Usually we try to avoid abbreviations. Wouldn't `for directory in root_directories:` look better?

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+        for rootpath, _, filenames_w in walk(rootdir):
+            for filename in filenames_w:
+                if filename.endswith('.py'):
+                    filenames.append(path.join(rootpath, filename))
+
+    if not dev_mode:
+        command_line = BARE_COMMAND + filenames
+        return subprocess.run(command_line, check=False).returncode == 0
+
+    for i, filename in enumerate(filenames):
```
wechman, line 44: Please use `enumerate(filenames, start=1)`

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+
+from typing import Any
+
+PROJECT_ROOT = path.dirname(path.dirname(path.realpath(__file__)))
+MYPY_CONFIG = f"{PROJECT_ROOT}/scripts/mypy.ini"
+
+SGR_INFO = "\033[1;32m"
+SGR_CLEAR = "\033[0m"
+
+def mypy_all_filenames(dev_mode: Any, rootdirs: Any) -> bool:
```
wechman, line 21: Since it is a new file, it would be nice to deliver it with a proper typing.

scripts/mypy_all.py
```diff
@@ -0,0 +1,103 @@
...
+    print()
+    return True
+
+
+def parse_command_line() -> Any:
+    script_description = dedent("""
+        Runs type annotation checks on all Python files in project directories known to contain Python scripts.
+
+        This script is meant to be run from the CI but can also be easily used in the local dev
+        environment.
```
wechman, line 66:
```suggestion
        This script is meant to be run from the CI but can also be easily used in the local development
        environment.
```

leonardoalt:
I think someone needs to take this PR over, if we want it.

gurukamath:
I will take care of the suggested changes

leonardoalt:
This PR has been stale with many reviewer requested changes for a while. I'm closing for now, the code is still available in the original branch. Please re-open if desired

PR #12954, opened 2022-04-23
Link to address literals from address value types

aathan:
Questions have come up in the chat about EIP-55 mixed case literals. Also the address value type section had two separate notes that both discuss payables, and which could therefore be merged. Linked to the literals section, which is easily missed otherwise.

### DIFF
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -210,6 +210,11 @@ an exception to this rule.
     declare its type as ``address payable`` to make this requirement visible. Also,
     try to make this distinction or conversion as early as possible.
 
+    The distinction between ``address`` and ``address payable`` was introduced with version 0.5.0.
+    Also starting from that version, contracts are not implicitly convertible to the ``address`` type, but can still be explicitly converted to
+    ``address`` or to ``address payable``, if they have a receive or payable fallback function.
+
+
 Operators:
 
 * ``<=``, ``<``, ``==``, ``!=``, ``>=`` and ``>``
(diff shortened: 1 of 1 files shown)


docs/types/value-types.rst
```diff
@@ -210,6 +210,11 @@ an exception to this rule.
     declare its type as ``address payable`` to make this requirement visible. Also,
     try to make this distinction or conversion as early as possible.
 
+    The distinction between ``address`` and ``address payable`` was introduced with version 0.5.0.
+    Also starting from that version, contracts do not derive from the address type, but can still be explicitly converted to
```
cameel: I see that this was here before your PR but it seems weird. Was this actually inheritance? I always thought that it was an explicit conversion plus the presence of contract members (which has a similar effect but it's technically not the same thing):

```suggestion
    Also starting from that version, contracts are not implicitly convertible to the ``address`` type, but can still be explicitly converted to
```

docs/types/value-types.rst
```diff
@@ -223,9 +228,7 @@ Operators:
     or you can use ``address(uint160(uint256(b)))``, which results in ``0x777788889999AaAAbBbbCcccddDdeeeEfFFfCcCc``.
 
 .. note::
-    The distinction between ``address`` and ``address payable`` was introduced with version 0.5.0.
-    Also starting from that version, contracts do not derive from the address type, but can still be explicitly converted to
-    ``address`` or to ``address payable``, if they have a receive or payable fallback function.
+    See :ref:`Address Literals`, which may be written as mixed-case hexadecimals conforming to `EIP-55 <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md>`_.
```
cameel: This seems like something that could just be integrated into the text instead of keeping it in a note:

```suggestion
Mixed-case hexadecimal numbers conforming to `EIP-55 <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md>`_ are automatically treated as literals of the ``address`` type.
See :ref:`Address Literals`.
```

cameel (dismissed):
Could be good enough even in the current form but I'd still apply some small tweaks.

docs/types/value-types.rst
cameel: Actually, there's an error here. I think that referencing other headings only works only works within the same document. You probably need to use an explicit anchor.

leonardoalt:
I've updated this PR but building the docs isn't working locally for me atm. @cameel can you check whether the anchor I added works properly?

leonardoalt:
Just managed to check it, looks fine.

leonardoalt (approved):

PR #12958, opened 2022-04-25
Yul grammar generator: Bound memory accesses.

### LINKED ISSUE #6091: Returning large segments of memory slows the yul interpreter down
## Description

```
{return(441964094,88877776)}
```

bshastry:
Fixes #6091.

PR to bound memory accesses such that stack-to-memory mover may be supported in an upcoming PR.


test/tools/ossfuzz/protoToYul.h
```diff
@@ -344,6 +344,8 @@ class ProtoConverter
 	static unsigned constexpr s_modOutputParams = 5;
 	/// Hard-coded identifier for a Yul object's data block
 	static auto constexpr s_dataIdentifier = "datablock";
+	/// Upper bound on memory writes
+	static unsigned constexpr s_maxMemory = 1024;
```
hrkrshnn: Why 1024? It's somewhat low. I'd go for `2**32`. See EIP-1985 as a reference.

leonardoalt: @bshastry wanna change that? We could get this merged soon

test/tools/ossfuzz/protoToYul.cpp
```diff
@@ -619,7 +619,14 @@ void ProtoConverter::visit(UnaryOp const& _x)
 		break;
 	}
 	m_output << "(";
-	visit(_x.operand());
+	if (op == UnaryOp::MLOAD)
+	{
+		m_output << "mod(";
```
leonardoalt, line 641: why `mod`?

bshastry: Because I use `mod` to place an upper bound like `mod(rand, upper)` is the bounded value

test/tools/ossfuzz/protoToYul.h
bshastry: The bound suggested by @hrkrshnn would not work with the current state of the Yul interpreter because it does not properly interpret memory accesses to high locations such as 2**32

ekpyron: It could easily do so, though, couldn't it? Resp. the bound in `EVMInstructionInterpreter::accessMemory` is 
`numeric_limits<size_t>::max() - 0xffff`... on most platforms `numeric_limits<size_t>::max()` should be 2**64... or do you mean a different limitation in there?

bshastry: I was wrong. Thanks for correcting @ekpyron . I pushed an update with a few more changes.

test/tools/ossfuzz/protoToYul.cpp
```diff
@@ -1074,7 +1094,15 @@ void ProtoConverter::visit(StoreFunc const& _x)
 		m_output << "mstore8(";
 		break;
 	}
-	visit(_x.loc());
+	// Write to memory within bounds, storage is unbounded
```
leonardoalt, line 1125: Why is storage unbounded?

bshastry: Storage is unbounded because it is unaffected by the stack-to-memory mover optimization step that this PR seeks to start testing. Memory needs to be bounded because the stack-to-memory mover needs a guarantee from the programmer that the program access memory within pre-defined bounds.

leonardoalt (dismissed):
Needs a rebase

axic:
I think this should be merged after #13516 is sorted.

test/libyul/yulInterpreterTests/external_call_to_self.yul
```diff
@@ -14,7 +14,6 @@
 // ----
 // Trace:
 //   CALL(153, 0x11111111, 0, 64, 32, 256, 32)
-//   RETURN(0, 0)
```
ekpyron: Why remove all these traces? They technically don't access memory, but the actual return (resp. log/create) still happens...

bshastry: > Why remove all these traces?

Mostly because the optimizer converts `return(12345, 0)` to `return(0, 0)` and consequently diverge traces.

ekpyron: Hm... we could print ``RETURN(0, 0)`` for all empty returns in the trace, maybe :-)?
The danger in just removing the traces is that we might miss issues where some return is actually missing, isn't it? Returning with an empty return vs not returning at all is a huge difference...

bshastry: Good point, I'll update the PR accordingly

bshastry: Updated.

ekpyron (approved):

cameel:
I restarted the failing `t_osx_cli`. May need a rebase if that does not pass.

PR #12966, opened 2022-04-28
Fix ICE caused by an immutable struct with mapping

### LINKED ISSUE #12953: ICE in `StructType::nativeMembers()` when the type of an immutable is a struct containing a mapping
## Description
Get an error `InternalCompilerError: Solidity assertion failed` which is not helpful. A descriptive error is shown in 0.8.11 (see below)

## Environment

- Compiler version: 0.8.12

## Steps to Reproduce
Run the code below in 0.8.12 or 0.8.13, you get the compiler output:
`InternalCompilerError: Solidity assertion failed`

```solidity
contract MyContract {
    
  struct CircularQueue {
    mapping(uint256 => address) queue;
    uint256 first;
    uint256 end;
    uint256 capacity;
  }

  CircularQueue immutable queue;

  constructor(
  ) {
    queue = CircularQueue({capacity: 2});
  }
}
```

While this code is not correct, it should still give a valid compiler warning. In 0.8.11 you correctly get:
```TypeError: Immutable variables cannot have a non-value type.
--> contracts/MyContract.sol:11:3:
|
11 | CircularQueue immutable queue;
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: Types in storage containing (nested) mappings cannot be assigned to.
--> contracts/MyContract.sol:15:5:
|
15 | queue = CircularQueue({capacity: 2});
| ^^^^^

TypeError: Struct containing a (nested) mapping cannot be constructed.
--> contracts/MyContract.sol:15:13:
|
15 | queue = CircularQueue({capacity: 2});
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

a3d4:
Fixes #12953.

The ICE was introduced as a side effect of 7c0a121e456b966826afe575929bad225aa0cb15, see 
https://github.com/ethereum/solidity/commit/7c0a121e456b966826afe575929bad225aa0cb15#diff-0e0e69b13aacc21916c6074c34a2ee68a0118b5914767495870c13b383061a94R135.

It makes `DocStringAnalizer` dependent on `TypeProvider`, which requires `TypeChecker`.

Moving `DocStringAnalyser` after `TypeCherker` fixes the ICE. Is it somewhat OK or too invasive?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Compiler Features:
 Bugfixes:
  * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
  * Yul Optimizer: Do not remove ``returndatacopy`` in cases in which it might perform out-of-bounds reads that unconditionally revert as out-of-gas. Previously, any ``returndatacopy`` that wrote to memory that was never read from was removed without accounting for the out-of-bounds condition.
+ * DocString Parser: Fix ICE caused by an immutable struct with mapping.
 
 
 ### 0.8.14 (2022-05-17)
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -479,12 +479,6 @@ bool CompilerStack::analyze()
 			if (auto sourceAst = source->ast)
 				noErrors = contractLevelChecker.check(*sourceAst);
 
-		// Requires ContractLevelChecker
-		DocStringAnalyser docStringAnalyser(m_errorReporter);
-		for (Source const* source: m_sourceOrder)
-			if (source->ast && !docStringAnalyser.analyseDocStrings(*source->ast))
-				noErrors = false;
-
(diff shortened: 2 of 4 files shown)


a3d4:
@Marenz The ICE seems to be related to your commit. Could you have a look at the PR please?

Marenz:
I would prefer a way that we only conditionally query the type, but I couldn't find any good conditions.. In that light, I think your approach is fine. You probably need to rebase for the CI though.

Marenz:
LGTM but needs rebasing again and CI fixes

ekpyron:
> LGTM but needs rebasing again and CI fixes

Rebasing alone should do it - that should also fix CI (the failures are just CI running out-of-memory and https://github.com/ethereum/solidity/pull/12907 upped the CI resources for those runs)

wechman:
@ekpyron, @Marenz I have rebased changes. You can continue a review.

Marenz (approved):

Issue #12897, opened 2022-04-04
Rewrite command line test script cleanly in python.


ekpyron:
Came up in https://github.com/ethereum/solidity/pull/12851.
``test/cmdlineTests.sh`` has grown more and more complicated over time - I'd say it's time to consider making it a python script and rewriting it nicely.
This would for example allow to easily filter by its sub-test-suites and individual tests, etc.

On the other hand it shares logic with other script, so whether this is a good idea may depend on whether we want to transition them as well.

cameel:
I think it's a good idea and here's the plan for doing it I came up with with @Marenz last year:
1) Add support for a new kind of CLI test - one that has a Python or Bash script inside.
2) Split the [Bash tests that are currently tacked on at the end of `cmdlineTests.sh`](https://github.com/ethereum/solidity/blob/v0.8.13/test/cmdlineTests.sh#L291-L674) into such smaller tests.
    - When naming dirs for these tests use a prefix to makes them sort at the end to much the current behavior (they currently run after all the other tests and I think we want to keep it that way since they can be long). `~` is a good candidate for such a prefix.
3) Rewrite the smaller scripts in Python.
4) Rewrite `cmdlineTests.sh` in Python.

I actually have an old branch where I did (1) and started (2). I'm going to assign myself here for now since I'd like to finish that when I have time. Then (3) and (4) could be done piece by piece.

christianparpart:
@cameel you self-assigned this, but we already have soo many PRs (you've 10 open ones). Let's try to close them first? :-)

cameel:
I assigned this because I already started working on this last year. There just was no issue for this back then but I have a branch with (1) and (2) almost done. I had to switch to more important tasks but if we're doing this then I want to update that branch and submit it as a PR when we're done with PR squashing.

christianparpart:
ah, perfect. thx. ;)

PR #12969, opened 2022-04-28
[SMTChecker] Add Eldarica as a Horn solver

leonardoalt:
Currently Eldarica can only be used via solc-js's callback, which is not great. This PR adds support to call Eldarica via its binary if installed in the system.


cameel:
Should this still be a draft? Many tests seem to be failing.

libsmtutil/CHCSmtLib2Interface.cpp
```diff
@@ -203,12 +209,46 @@ string CHCSmtLib2Interface::querySolver(string const& _input)
...
+
+	if (m_enabledSolvers.eld)
+	{
+		auto tempDir = boost::filesystem::temp_directory_path();
+		auto queryFileName = tempDir / "query.smt2";
+
+		auto queryFile = boost::filesystem::ofstream(queryFileName);
+		queryFile << _input;
+
+		auto eldBin = boost::process::search_path("eld");
```
axic: I think generally not a good idea to have such filesystem/execution dependent code in this library, as it gets included in the wasm/JS build which could become complex.

It should be supported in the CLI, but that needs some new interface.

leonardoalt: What exactly becomes complex? I think it should also be supported in JSON mode

axic: The "JSON mode" abstracts all this via callbacks, it is not using any `boost::filesystem` or `boost::process` libraries.

axic: See the build failures:

```
[100%] Linking CXX executable soljson.js
em++: warning: EXTRA_EXPORTED_RUNTIME_METHODS is deprecated, please use EXPORTED_RUNTIME_METHODS instead [-Wdeprecated]
error: undefined symbol: $PIPEFS (referenced by __syscall_pipe__deps: ['$PIPEFS','$SYSCALLS'], referenced by top-level compiled C/C++ code)
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
warning: PIPEFS may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: __syscall_wait4 (referenced by top-level compiled C/C++ code)
warning: Link with `-s LLD_REPORT_UNDEFINED` to get more information on undefined symbols
warning: ___syscall_wait4 may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: execve (referenced by top-level compiled C/C++ code)
warning: _execve may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: fork (referenced by top-level compiled C/C++ code)
warning: _fork may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
Error: Aborting compilation due to previous errors
em++: error: '/emsdk/node/14.15.5_64bit/bin/node /emsdk/upstream/emscripten/src/compiler.js /tmp/tmp6dnsntii.txt' failed (returned 1)
```

leonardoalt:
@axic how about now? Please check first commit

test/cmdlineTests/model_checker_solvers_eld/err
```diff
@@ -0,0 +1 @@
+Warning: CHC analysis was not possible since Eldarica was the only Horn solver enabled, but it was not found in the system.
```
ekpyron: So this test will fail on all systems that happen to have Eldarica installed :-)?

leonardoalt: Yes indeed. I think we do the same for cvc4 maybe?

ekpyron: Do we? It's suboptimal in any case, but yeah, I'd ultimately not stall this because of that...

leonardoalt: Yea, most people won't have eld anyway. And this is only a problem for the tests to be 100% precise.

libsmtutil/CHCSmtLib2Interface.cpp
```diff
@@ -203,12 +206,14 @@ string CHCSmtLib2Interface::querySolver(string const& _input)
 	util::h256 inputHash = util::keccak256(_input);
 	if (m_queryResponses.count(inputHash))
 		return m_queryResponses.at(inputHash);
-	if (m_smtCallback)
+
+	if ((m_enabledSolvers.smtlib2 || m_enabledSolvers.eld) && m_smtCallback)
```
leonardoalt: maybe unnecessary in the current version to test which solvers were enabled here

libsmtutil/CHCSmtLib2Interface.cpp
```diff
@@ -16,6 +16,7 @@
 */
 // SPDX-License-Identifier: GPL-3.0
 
+#include <boost/filesystem/operations.hpp>
```
leonardoalt: not needed

libsmtutil/CHCSmtLib2Interface.h
```diff
@@ -35,6 +35,7 @@ class CHCSmtLib2Interface: public CHCSolverInterface
 	explicit CHCSmtLib2Interface(
 		std::map<util::h256, std::string> const& _queryResponses = {},
 		frontend::ReadCallback::Callback _smtCallback = {},
+		SMTSolverChoice _enabledSolvers = SMTSolverChoice::All(),
```
leonardoalt, line 38: maybe not needed

libsolidity/interface/SMTSolverCLI.h
```diff
@@ -0,0 +1,50 @@
...
+public:
+	SMTSolverCLI(std::string _solverCmd);
+
+	/// Receives a @p _sourceUnitName that refers to a source unit in compiler's virtual filesystem
+	/// and attempts to interpret it as a path and read the corresponding file from disk.
+	/// The read will only succeed if the canonical path of the file is within one of the @a allowedDirectories().
+	/// @param _kind must be equal to "source". Other values are not supported.
+	/// @return Content of the loaded file or an error message. If the operation succeeds, a copy of
+	/// the content is retained in @a sourceUnits() under the key of @a _sourceUnitName. If the key
+	/// already exists, previous content is discarded.
```
leonardoalt: Comments are outdated.

libsmtutil/SolverInterface.h
```diff
@@ -42,21 +42,24 @@ namespace solidity::smtutil
 struct SMTSolverChoice
 {
 	bool cvc4 = false;
+	bool eld = false;
 	bool smtlib2 = false;
 	bool z3 = false;
```
cameel, line 47: Not sure about the other tests yet but I think that [`t_archlinux_soltest
`](https://app.circleci.com/pipelines/github/ethereum/solidity/25900/workflows/85ffec54-b5d6-4a5f-a11b-904281e61ae4/jobs/1143820) is failing because the new option is not included in the expectations:

https://github.com/ethereum/solidity/blob/a229d7219bcc7100364ed9c636fc3bbabf4fb4bb/test/solc/CommandLineParser.cpp#L210-L219

But I wonder why the compiler is not complaining that the struct initializer is missing one item.

cameel, line 47: Not sure about the other tests yet but I think that [`t_archlinux_soltest
`](https://app.circleci.com/pipelines/github/ethereum/solidity/25900/workflows/85ffec54-b5d6-4a5f-a11b-904281e61ae4/jobs/1143820) is failing because the new option is not included in the expectations:

https://github.com/ethereum/solidity/blob/a229d7219bcc7100364ed9c636fc3bbabf4fb4bb/test/solc/CommandLineParser.cpp#L210-L219

But I wonder why the compiler is not complaining that the struct initializer is missing one item.

cameel: By the way, you should include the new option on the argument list in these tests: https://github.com/ethereum/solidity/blob/a229d7219bcc7100364ed9c636fc3bbabf4fb4bb/test/solc/CommandLineParser.cpp#L145-L152

leonardoalt: Right yea, those are still missing.

libsolidity/interface/FileAndSMTCallbacks.h
```diff
@@ -0,0 +1,54 @@
...
+			auto reader = m_fileReader.reader();
+			auto solver = m_solver.solver();
+			if (_kind == ReadCallback::kindString(ReadCallback::Kind::ReadFile))
+				return reader(_kind, _data);
+			else if (_kind == ReadCallback::kindString(ReadCallback::Kind::SMTQuery))
+				return solver(_kind, _data);
+			solAssert(false);
+			return {};
+		};
+	}
```
cameel: As for the memory violation in [`t_ubu_release_soltest_all`](https://app.circleci.com/pipelines/github/ethereum/solidity/25900/workflows/85ffec54-b5d6-4a5f-a11b-904281e61ae4/jobs/1143809/steps), it happens during tests for `--allow-paths` but it's reproducible by just compiling a file containing an import. For example something as simple as this does the trick:
```solidity
import 'x.sol';
```
```bash
solc test.sol
```

The memory violation indeed happens in `FileReader`, when it's processing include paths:
https://github.com/ethereum/solidity/blob/a229d7219bcc7100364ed9c636fc3bbabf4fb4bb/libsolidity/interface/FileReader.cpp#L114
But it's `fileAndSMTCallbacks()` that invokes it and the problem seems to be there. Looks as if the list of include paths was corrupted or something.

cameel: I don't see the problem yet but passing `this` to the callback looks suspicious. Do you make sure that it outlives the file reader?

scripts/error_codes.py
```diff
@@ -230,7 +230,7 @@ def examine_id_coverage(top_dir, source_id_to_file_names, new_ids_only=False):
         "1584", "1823",
         "1988", "2066", "2833", "3356",
         "3893", "3996", "4010", "4802",
-        "5272", "5622", "7128", "7400",
+        "5272", "5622", "7128", "7267", "7400",
```
cameel: Isn't this covered by a CLI test now? Not really sure we want such a test that depends on you having/not having specific software but just commenting for now to go back to it once I review that test.

test/libsolidity/smtCheckerTests/eldarica/smoke.sol
```diff
@@ -0,0 +1,8 @@
+contract C {
+	function f(uint x) public pure {
+		assert(x > 0);
+	}
+}
+// ====
+// SMTEngine: chc
+// SMTSolvers: eld
```
cameel: This expects no errors because the assumption is that you don't have Eldarica installed? I think the file should have a comment saying that. Might not be obvious.

test/cmdlineTests/standard_model_checker_solvers_eld/input.json
```diff
@@ -0,0 +1,18 @@
+{
+	"language": "Solidity",
+	"sources":
+	{
+		"A":
+		{
+			"content": "// SPDX-License-Identifier: GPL-3.0\npragma solidity >=0.0;\n\ncontract C { function f(uint x) public pure { assert(x > 0); } }"
```
cameel: Can you put the snippet in a separate file and load it via `urls`?

docs/smtchecker.rst
```diff
@@ -614,10 +614,11 @@ which is primarily an SMT solver and makes `Spacer
...
 The user can choose which solvers should be used, if available, via the CLI
-option ``--model-checker-solvers {all,cvc4,smtlib2,z3}`` or the JSON option
+option ``--model-checker-solvers {all,cvc4,eld,smtlib2,z3}`` or the JSON option
 ``settings.modelChecker.solvers=[smtlib2,z3]``, where:
 
 - ``cvc4`` is only available if the ``solc`` binary is compiled with it. Only BMC uses ``cvc4``.
+- ``eld`` is used via its binary which must be installed in the system. Only CHC uses ``eld``.
 - ``smtlib2`` outputs SMT/Horn queries in the `smtlib2 <http://smtlib.cs.uiowa.edu/>`_ format.
   These can be used together with the compiler's `callback mechanism <https://github.com/ethereum/solc-js>`_ so that
   any solver binary from the system can be employed to synchronously return the results of the queries to the compiler.
```
cameel, line 624: I guess the sentence below ("This is currently the only way to use Eldarica, for example, since it does not have a C++ API.", can't comment on it) is not longer correct in this PR? It should be updated.

test/cmdlineTests/model_checker_solvers_eld/err
cameel: Not sure if this test is a good idea. Not being able to pass all SMT tests if you don't have a specific solver would be unfortunate but understandable. Not being able to pass them when you **do** have a specific solver on the other hand would IMO be much more weird and annoying.

I see that detection is based on the presence of `eld` command in path. We could make it more robust by hiding the executable from bash. I haven't tried it with boost but from what I've read, creating a function like `function eld { false; }` would do that. You could put it in `cmdlineTests.sh` with a note that it completely hides `eld` from all CLI tests and is a hack. After #12897 we'd be able to customize each test more so we'd make it less of a hack and move it into that specific test.

test/cmdlineTests/standard_model_checker_solvers_eld/output.json
```diff
@@ -0,0 +1,3 @@
+{"errors":[{"component":"general","errorCode":"7267","formattedMessage":"Warning: CHC analysis was not possible since Eldarica was the only Horn solver enabled, but it was not found in the system.
+
+","message":"CHC analysis was not possible since Eldarica was the only Horn solver enabled, but it was not found in the system.","severity":"warning","type":"Warning"}],"sources":{"A":{"id":0}}}
```
cameel: Would be best to add an `args` file with `--pretty-print --json-indent 4` to have the output formatted nicely.

solc/CommandLineInterface.h
```diff
@@ -27,6 +27,7 @@
 #include <libsolidity/interface/CompilerStack.h>
 #include <libsolidity/interface/DebugSettings.h>
 #include <libsolidity/interface/FileReader.h>
+#include <libsolidity/interface/SMTSolverCLI.h>
```
cameel: Is this needed in the header? There are no other changes here so looks like it could be in the .cpp file instead.

libsolidity/interface/SMTSolverCLI.cpp
```diff
@@ -0,0 +1,94 @@
...
+		return ReadCallback::Result{false, "Exception in SMTQuery callback: " + boost::diagnostic_information(_exception)};
+	}
+	catch (std::exception const& _exception)
+	{
+		return ReadCallback::Result{false, "Exception in SMTQuery callback: " + boost::diagnostic_information(_exception)};
+	}
+	catch (...)
+	{
+		return ReadCallback::Result{false, "Unknown exception in SMTQuery callback: " + boost::current_exception_diagnostic_information()};
+	}
```
cameel: You can drop the first two handlers here. The `current_exception_diagnostic_information()` used by `catch (...)` will get us all the same info `diagnostic_information()` would.

libsolidity/interface/SMTSolverCLI.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+		auto tempDir = boost::filesystem::temp_directory_path();
+		auto queryFileName = tempDir / "query.smt2";
+
+		auto queryFile = boost::filesystem::ofstream(queryFileName);
+		queryFile << _query;
+
+		auto eldBin = boost::process::search_path("eld");
+		string res;
+		if (!eldBin.empty())
```
cameel: Is it really a good idea to do nothing but still report success when the `eld` binary cannot be found? As a user I'd expect some kind of error because the compiler failed to do what I requested.

libsolidity/interface/SMTSolverCLI.cpp
```diff
@@ -0,0 +1,94 @@
...
+		{
+			boost::process::ipstream is;
```
cameel:
```suggestion
			boost::process::ipstream pipe;
```

libsolidity/interface/SMTSolverCLI.cpp
```diff
@@ -0,0 +1,94 @@
...
+			solAssert(false, "SMTQuery callback used as callback kind " + _kind);
+
+		auto tempDir = boost::filesystem::temp_directory_path();
+		auto queryFileName = tempDir / "query.smt2";
+
+		auto queryFile = boost::filesystem::ofstream(queryFileName);
+		queryFile << _query;
+
+		auto eldBin = boost::process::search_path("eld");
+		string res;
```
cameel:
```suggestion
		string callback_output;
```

libsolidity/interface/SMTSolverCLI.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+			vector<string> data;
+			string line;
+			while (eld.running() && std::getline(is, line))
+				if (!line.empty())
+					data.push_back(line);
+
+			eld.wait();
+
+			res = boost::join(data, "\n");
```
cameel: Dumping a string to a file and spawning a child with output capture seem like things that might be handy from time to time. Maybe you could make them into helpers in `CommonIO.h`?

libsolidity/interface/FileAndSMTCallbacks.h
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: GPL-3.0
+#pragma once
+
+#include <libsolidity/interface/FileReader.h>
+#include <libsolidity/interface/SMTSolverCLI.h>
+
+namespace solidity::frontend
+{
+
+class FileAndSMTCallbacks
```
cameel: It would be best to consistently refer to the `FileReader`'s callback as "import callback".
```suggestion
class ImportAndSMTCallbackAdapter
```
or how about we just call it an "universal callback"?
```suggestion
class UniversalCallback
```

libsolidity/interface/FileAndSMTCallbacks.h
```diff
@@ -0,0 +1,54 @@
...
+	{
+		return [this](std::string const& _kind, std::string const& _data) -> ReadCallback::Result {
+			auto reader = m_fileReader.reader();
+			auto solver = m_solver.solver();
+			if (_kind == ReadCallback::kindString(ReadCallback::Kind::ReadFile))
+				return reader(_kind, _data);
+			else if (_kind == ReadCallback::kindString(ReadCallback::Kind::SMTQuery))
+				return solver(_kind, _data);
+			solAssert(false);
+			return {};
```
cameel: If the compiler complains here, we have recently added a helper for such cases:
```suggestion
			unreachable();
```
It's in `libsolutil/Assertions.h`.

But also recently, someone noticed that not skipping the message in `solAssert()` makes the compiler properly detect this as unreachable so you could also do that instead. It's only the new macros without the message that have this problem.

libsolidity/interface/FileAndSMTCallbacks.h
cameel: Yeah. The problem is in this bit: 
```c++
	if (modelChecker.solvers.eld)
	{
		auto smtCLI = SMTSolverCLI("eld");
		m_compiler = make_unique<CompilerStack>(FileAndSMTCallbacks(m_fileReader, smtCLI).fileAndSMTCallbacks());
	}
```
So `this` is fine, but `m_solver` goes out of scope immediately after creation. You need to store it in `CommandLineInterface`. Maybe it would be cleaner (and harder to make this mistake) if the new callback owned the other callbacks instead of storing references?

libsolidity/interface/FileAndSMTCallbacks.h
```diff
@@ -0,0 +1,54 @@
...
+
+class FileAndSMTCallbacks
+{
+public:
+	FileAndSMTCallbacks(FileReader& _fileReader, SMTSolverCLI& _solver) :
+		m_fileReader(_fileReader),
+		m_solver(_solver)
+	{}
+
+	frontend::ReadCallback::Callback fileAndSMTCallbacks()
```
cameel: How about we call it just `callback()`?
```suggestion
	frontend::ReadCallback::Callback callback()
```

Actually, I'd be tempted to also rename `reader()` and `solver()` to `callback()` as well.

libsolidity/interface/SMTSolverCLI.h
```diff
@@ -0,0 +1,50 @@
...
+#pragma once
+
+#include <libsolidity/interface/ReadFile.h>
+
+#include <boost/filesystem.hpp>
+
+namespace solidity::frontend
+{
+
+class SMTSolverCLI
```
cameel: How about this? I was a bit confused about the CLI in the name until I realized you really meant mostly the 'command' part rather than the 'interface' part :)

```suggestion
class SMTSolverCommand
```

Also, currently it's hard-coded for Eldarica so I'd rename it to `EldaricaCommand`. Or you could make it actually more generic and accept `modelChecker.solvers` as input and decide based on that (i.e. just do nothing for other solvers but at least keep that decision here). And then I'd hard-code the solver command here (rather than in the CLI).

libsolidity/formal/CHC.cpp
```diff
@@ -99,6 +101,20 @@ void CHC::analyze(SourceUnit const& _source)
...
+	)
+	{
+		m_errorReporter.warning(
+			7267_error,
+			SourceLocation(),
+			"CHC analysis was not possible since Eldarica was the only"
+			" Horn solver enabled, but it was not found in the system."
+		);
+		return;
+	}
```
cameel: Shouldn't we generally warn the user if a solver is selected but not available even when some other solver is? I'd understand not wanting to issue an error and fail but this is just a warning.

libsmtutil/CHCSmtLib2Interface.cpp
cameel: If the callback is not allowed for other solvers, shouldn't we report an error instead of ignoring it?

cameel (changes requested):

leonardoalt:
Thanks for the review @cameel . I haven't had time to look at the first one yet, but will try to look at the whole thing this week.

cameel:
Sure, no hurry, I'm at a hackathon now so I won't be able to go back to it before Monday anyway :) At least I finally managed to digest it all.

Overall I didn't find many serious problems - mostly the bug that CI was crashing on, rest are small things so we should be able to get this merged quckly once I'm back.

test/cmdlineTests/model_checker_solvers_eld/err
leonardoalt: We don't even support the newest cvc (cvc5) and haven't used it for ages. I think it's a pragmatic choice to just say we don't use cvc4 in the tests.

libsolidity/interface/FileAndSMTCallbacks.h
leonardoalt: Yea that would be better for the smt one, but not sure we should move `m_fileReader` into it. Maybe make a copy.

leonardoalt:
I'm fixing it now.

libsmtutil/CHCSmtLib2Interface.cpp
leonardoalt: it's not that it's not allowed, it's just not used

libsolidity/interface/SMTSolverCLI.cpp
leonardoalt: Good point. Changed it to an error.

libsolidity/interface/SMTSolverCLI.cpp
leonardoalt: Removed the variable completely.

libsolidity/formal/CHC.cpp
leonardoalt: I think this is how it's done already. It's done separately in each engine because they use different solvers.

libsolidity/interface/SMTSolverCLI.cpp
leonardoalt: It requires some tweaking because here we have the nice `ReadCallback::Result` type. Is it ok to just do it in another PR?

libsolidity/interface/FileAndSMTCallbacks.h
leonardoalt: Changed to `UniversalCallback`.

libsolidity/interface/FileAndSMTCallbacks.h
leonardoalt: Ah ok, added string.

libsolidity/interface/FileAndSMTCallbacks.h
leonardoalt: Change the new one. I'd change the other ones separately, if at all.

libsolidity/interface/SMTSolverCLI.h
leonardoalt: Changed the name.
This class is not hardcoded to Eldarica though, that's a constructor argument.

leonardoalt:
@cameel @ekpyron can we decide in a sync way what to do with the tests here? I don't wanna try different things that might be rejected, just wanna make a decision before.

libsolidity/formal/ModelChecker.cpp
```diff
@@ -135,14 +141,55 @@ vector<string> ModelChecker::unhandledQueries()
...
 #ifdef HAVE_CVC4
 	available.cvc4 = true;
 #endif
 	return available;
 }
+
+SMTSolverChoice ModelChecker::checkRequestedSolvers(SMTSolverChoice _enabled, ErrorReporter& _errorReporter)
+{
+	auto availableSolvers = ModelChecker::availableSolvers();
+
```
stackenbotten: Coding style error

stackenbotten: Coding style error

test/cmdlineTests/model_checker_solvers_eld/err
cameel: I mean, I'd see it as pragmatic if there was no other choice but I think that the solution with shadowing `eld` inside `cmdlineTests.sh` is pretty simple and better IMO. Why not do that?

leonardoalt: Well, we both agree it's a hack, and not even sure that works. boost::process will look for the `eld` binary in the system, so we'd also need to add a wrapper in the system somehow? Also, if we do add a binary that we expect to be called as eldarica, it will influence the expected output, in which case we're basically deciding to always/never have the solver in the system.

libsolidity/formal/ModelChecker.cpp
stackenbotten: Coding style error

libsolidity/formal/ModelChecker.cpp
```diff
@@ -135,14 +141,55 @@ vector<string> ModelChecker::unhandledQueries()
...
 #ifdef HAVE_CVC4
 	available.cvc4 = true;
 #endif
 	return available;
 }
+
+SMTSolverChoice ModelChecker::checkRequestedSolvers(SMTSolverChoice _enabled, ErrorReporter& _errorReporter)
+{
+	SMTSolverChoice availableSolvers{ModelChecker::availableSolvers()};
+
```
stackenbotten: Coding style error

libsmtutil/CHCSmtLib2Interface.cpp
leonardoalt: Actually, since this is CHC, we could add an assertion that the first arg of the conjunction should be true. What do you think @cameel ?

libsmtutil/SolverInterface.h
```diff
@@ -42,35 +42,31 @@ namespace solidity::smtutil
...
+	static constexpr SMTSolverChoice CVC4() noexcept { return {true, false, false, false}; }
+	static constexpr SMTSolverChoice ELD() noexcept { return {false, true, false, false}; }
+	static constexpr SMTSolverChoice SMTLIB2() noexcept { return {false, false, true, false}; }
+	static constexpr SMTSolverChoice Z3() noexcept { return {false, false, false, true}; }
+	static constexpr SMTSolverChoice None() noexcept { return {false, false, false, false}; }
 
 	static std::optional<SMTSolverChoice> fromString(std::string const& _solvers)
 	{
 		SMTSolverChoice solvers;
-		if (_solvers == "all")
```
leonardoalt, line 57: Removed this `all` option in favor of a default CLI option.

libsolidity/formal/ModelCheckerSettings.h
```diff
@@ -153,7 +153,7 @@ struct ModelCheckerSettings
 	ModelCheckerEngine engine = ModelCheckerEngine::None();
 	ModelCheckerInvariants invariants = ModelCheckerInvariants::Default();
 	bool showUnproved = false;
-	smtutil::SMTSolverChoice solvers = smtutil::SMTSolverChoice::All();
+	smtutil::SMTSolverChoice solvers = smtutil::SMTSolverChoice::Z3();
```
leonardoalt, line 156: Changed the default to `z3` because it's the only solver that both BMC and CHC can use plus it's the one we ship best.

solc/CommandLineParser.cpp
```diff
@@ -828,7 +828,7 @@ General Information)").c_str(),
 		)
 		(
 			g_strModelCheckerSolvers.c_str(),
-			po::value<string>()->value_name("all,cvc4,z3,smtlib2")->default_value("all"),
+			po::value<string>()->value_name("cvc4,eld,z3,smtlib2")->default_value("z3"),
```
leonardoalt, line 831: If we keep it at `all`, we always get warnings saying some solver was enabled but is not available. Better to change the default to `z3`.

libsmtutil/SolverInterface.h
cameel: Do we still consider the SMTChecker an experimental feature? Because if not, this is technically breaking.

libsolidity/formal/ModelChecker.cpp
```diff
@@ -135,14 +141,55 @@ vector<string> ModelChecker::unhandledQueries()
 	return m_bmc.unhandledQueries() + m_chc.unhandledQueries();
 }
 
-solidity::smtutil::SMTSolverChoice ModelChecker::availableSolvers()
+SMTSolverChoice ModelChecker::availableSolvers()
 {
 	smtutil::SMTSolverChoice available = smtutil::SMTSolverChoice::SMTLIB2();
-#ifdef HAVE_Z3
+	available.eld = !boost::process::search_path("eld").empty();
+#if defined(HAVE_Z3) || defined(HAVE_Z3_DLOPEN)
```
cameel: As far as I can tell `HAVE_Z3_DLOPEN` is never defined without `HAVE_Z3` so `HAVE_Z3` should be enough.

If you're doing it just in case then maybe we should just have an assert that `!defined(HAVE_Z3) || defined(HAVE_Z3_DLOPEN)` combination is not possible?

libsolidity/formal/ModelCheckerSettings.h
cameel: This should probably be in the changelog.

Actually, the changelog entry for the whole feature is missing.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -577,6 +577,11 @@ bool CompilerStack::analyze()
...
 		if (noErrors)
 		{
+			/// m_modelCheckerSettings is spread to engines and solver interfaces,
+			/// so we need to check whether the enabled ones are available before building the classes.
+			if (m_modelCheckerSettings.engine.any())
+				m_modelCheckerSettings.solvers = ModelChecker::checkRequestedSolvers(m_modelCheckerSettings.solvers, m_errorReporter);
+
 			ModelChecker modelChecker(m_errorReporter, *this, m_smtlib2Responses, m_modelCheckerSettings, m_readFile);
 			auto allSources = util::applyMap(m_sourceOrder, [](Source const* _source) { return _source->ast; });
 			modelChecker.enableAllEnginesIfPragmaPresent(allSources);
```
cameel: Shouldn't this new `if` be placed after `enableAllEnginesIfPragmaPresent()`? I think that otherwise you won't get the new warnings when using the pragma.

libsmtutil/CHCSmtLib2Interface.cpp
cameel: Ah, I didn't notice that it's guaranteed to be true. Sure, assert sounds right in that case.

libsolidity/formal/BMC.cpp
```diff
@@ -61,15 +62,14 @@ BMC::BMC(
 
 void BMC::analyze(SourceUnit const& _source, map<ASTNode const*, set<VerificationTargetType>, smt::EncodingContext::IdCompare> _solvedTargets)
 {
-	if (m_interface->solvers() == 0)
+	/// At this point every enabled solver is available.
```
cameel:
```suggestion
	// At this point every enabled solver is available.
```

Actually, I just noticed that you're using these docstring comments in many places in implementation files. Why?

libsolidity/formal/CHC.cpp
```diff
@@ -994,10 +986,11 @@ void CHC::resetSourceAnalysis()
 	ArraySlicePredicate::reset();
 	m_blockCounter = 0;
 
-	bool usesZ3 = false;
+	/// At this point every enabled solver is available.
+	/// If more than one Horn solver is selected we go with z3.
```
cameel: Why? This sounds like something that should be documented somewhere to make it clear to the user. E.g. in the CLI option description.

Or maybe we should at least issue a warning in this case?

libsolidity/formal/CHC.cpp
```diff
@@ -994,10 +986,11 @@ void CHC::resetSourceAnalysis()
...
+	/// We still need the ifdef because of Z3CHCInterface.
 #ifdef HAVE_Z3
```
cameel: You could move the `#ifdef` into the `if` though. And maybe inside add an assert that `HAVE_Z3` is true. Same with a similar `#ifdef` below.

solc/CommandLineParser.cpp
cameel: Sounds reasonable.

libsolidity/interface/UniversalCallback.h
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: GPL-3.0
+#pragma once
+
+#include <libsolidity/interface/FileReader.h>
+#include <libsolidity/interface/SMTSolverCommand.h>
+
+namespace solidity::frontend
+{
+
+class UniversalCallback
```
cameel, line 29: The class and the file should have at least a basic docstring. Same with `SMTSolverCommand`.

libsolidity/interface/SMTSolverCommand.cpp
```diff
@@ -0,0 +1,85 @@
...
+		if (_kind != ReadCallback::kindString(ReadCallback::Kind::SMTQuery))
+			solAssert(false, "SMTQuery callback used as callback kind " + _kind);
+
+		auto tempDir = boost::filesystem::temp_directory_path();
+		auto queryFileName = tempDir / "query.smt2";
+
+		auto queryFile = boost::filesystem::ofstream(queryFileName);
+		queryFile << _query;
+
+		auto eldBin = boost::process::search_path("eld");
```
cameel: Shouldn't this use `m_solverCmd`?

solc/CommandLineInterface.cpp
```diff
@@ -692,7 +694,12 @@ void CommandLineInterface::compile()
 {
 	solAssert(CompilerInputModes.count(m_options.input.mode) == 1);
 
-	m_compiler = make_unique<CompilerStack>(m_fileReader.reader());
+	auto modelChecker = m_options.modelChecker.settings;
```
```suggestion
	auto modelCheckerSettings = m_options.modelChecker.settings;
```

solc/CommandLineInterface.cpp
```diff
@@ -692,7 +694,12 @@ void CommandLineInterface::compile()
...
+	auto universalCallback{UniversalCallback(m_fileReader, SMTSolverCommand("eld"))};
+	if (modelChecker.solvers.eld)
+		m_compiler = make_unique<CompilerStack>(universalCallback.callback());
```
cameel: You can put the callback inside the `if` branch and avoid making a copy of the reader when eld is not used.

libsolidity/interface/UniversalCallback.h
```diff
@@ -0,0 +1,51 @@
...
+	{}
+
+	frontend::ReadCallback::Callback callback()
+	{
+		return [this](std::string const& _kind, std::string const& _data) -> ReadCallback::Result {
+			auto solver = m_solver.solver();
+			if (_kind == ReadCallback::kindString(ReadCallback::Kind::ReadFile))
+				return m_fileReader.readFile(_kind, _data);
+			else if (_kind == ReadCallback::kindString(ReadCallback::Kind::SMTQuery))
+				return solver(_kind, _data);
```
cameel: Solver is used only inside the `else`. I'd move it inside.

solc/CommandLineInterface.cpp
```diff
@@ -692,7 +694,12 @@ void CommandLineInterface::compile()
 {
 	solAssert(CompilerInputModes.count(m_options.input.mode) == 1);
 
-	m_compiler = make_unique<CompilerStack>(m_fileReader.reader());
+	auto modelChecker = m_options.modelChecker.settings;
+	auto universalCallback{UniversalCallback(m_fileReader, SMTSolverCommand("eld"))};
```
cameel: This leaves us with two copies of the file reader (`m_fileReader` and the one inside the universal callback) which can easily go out of sync. This is IMO just asking for trouble. The command-line interface keeps referring to `m_fileReader`, while the callback pulls files from its own copy.

libsolidity/interface/SMTSolverCLI.h
cameel: It actually is, and the argument seems unused, though I suspect it's just a bug.

But ignoring that, is it usable with any other solver? It gives the binary arguments like `-ssol` and `-scex`. I'm assuming that's specific to Eldarica?

libsolidity/interface/SMTSolverCLI.cpp
cameel: Sure. It's not that important overall, just a suggestion.

libsmtutil/SolverInterface.h
leonardoalt: I guess we do since the pragma is still supported.

libsolidity/formal/ModelCheckerSettings.h
leonardoalt: Yea, will add.

libsolidity/formal/ModelChecker.cpp
leonardoalt: True, will remove the second.

libsolidity/interface/CompilerStack.cpp
leonardoalt: It needs to go before the constructor of `ModelChecker`, because it propagates the settings into the engines and solvers. But good point, I'll check.

test/cmdlineTests/model_checker_solvers_eld/err
cameel: ok. good points. Anyway, I still think we might do better than assuming it's not installed, one way or another, but since we decided drop that test the point is moot - so marking this as resolved.

libsolidity/formal/BMC.cpp
leonardoalt: Yea, no idea lol

libsolidity/formal/CHC.cpp
leonardoalt: Warning and docs sounds good. The reason why is the same as why we're setting z3 as the default, just better support from our side, and we don't have a CHC portfolio like we do with BMC which does support multiple solvers at the same time.

libsolidity/interface/SMTSolverCommand.cpp
leonardoalt: Yes!

solc/CommandLineInterface.cpp
leonardoalt: Somehow we can't. That's what we had before and what causes crashes

solc/CommandLineInterface.cpp
leonardoalt: I know, but if we put it inside the branch it causes exceptions.

cameel: Ah, that was the problem with it going out of scope.

Why not make it a member variable then and have it keep a reference to the reader?

solc/CommandLineInterface.cpp
cameel: Ah right. So like in my other comment - I'd suggest making it a member variable of the `CommandLineInterface`. That would solve lifetime issues and IMO in a cleaner way than creating copies.

libsmtutil/SolverInterface.h
cameel: Good. Switching to Z3 as the default makes sense so having go consider it breaking would be annoying.

libsolidity/interface/CompilerStack.cpp
leonardoalt: Fixed.

solc/CommandLineInterface.cpp
leonardoalt: @cameel if we put the callback inside the if, `universalCallback` goes out of scope right after, causing the `this` in the callback to be trash.

solc/CommandLineInterface.cpp
leonardoalt: I think it should be fine. `m_fileReader` is only used as read at line 769: `m_compiler->setSources(m_fileReader.sourceUnits());`, and after that the compiler is already called. I don't see how the objects can go out of sync.

leonardoalt: Ultimately, I can also change this callback to go directly into the `ModelCheckerSettings` only as `SMTSolverCommand` instead of the compiler-wide `ReadCallback`, since it doesn't really use the C callback but is only a wrapper.

libsolidity/interface/SMTSolverCLI.h
leonardoalt: Yea I removed those in this version actually, because we don't use them anyway in the current version. It should be usabled with any solver.

leonardoalt:
Ok it's now fixed with the SMT command callback and the Universal callback as class members.
@cameel lmk if fine then I'll squash.

solc/CommandLineInterface.cpp
```diff
@@ -692,7 +692,15 @@ void CommandLineInterface::compile()
...
 	solAssert(CompilerInputModes.count(m_options.input.mode) == 1);
 
-	m_compiler = make_unique<CompilerStack>(m_fileReader.reader());
+	auto modelCheckerSettings = m_options.modelChecker.settings;
+	if (modelCheckerSettings.solvers.eld)
+	{
+		m_solverCommand = make_unique<SMTSolverCommand>("eld");
+		m_universalCallback = make_unique<UniversalCallback>(m_fileReader, *m_solverCommand);
+		m_compiler = make_unique<CompilerStack>(m_universalCallback->callback());
+	}
```
cameel: I just realized that this will only run on the CLI. What about Standard JSON?

Also, since the new solver is in model checker settings, we'll actually accept `eld` in Standard JSON but just do nothing.

libsolidity/interface/SMTSolverCommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+ReadCallback::Result SMTSolverCommand::solve(string const& _kind, string const& _query)
+{
+	try
+	{
+		if (_kind != ReadCallback::kindString(ReadCallback::Kind::SMTQuery))
+			solAssert(false, "SMTQuery callback used as callback kind " + _kind);
+
+		auto tempDir = boost::filesystem::temp_directory_path();
+		auto queryFileName = tempDir / "query.smt2";
```
cameel: When I saw this earlier I assumed you were creating a new tmp dir but after running it, I see that you just leave `query.smt2` directly in the tmp dir. That will be problematic when someone tries to run multiple instances of `solc` in parallel. And it's easy to avoid. You should create the file/dir with a unique name, and preferably delete it once you no longer need it.

I actually have a utility for handling that better, and I'd suggest using it, though you'd have to move it from tests to solutil:
https://github.com/ethereum/solidity/blob/9db2da0385c5abec0d1c3eab468648ec85fb5caa/test/TemporaryDirectory.h#L41-L56

cameel (commented):
ok, so the stuff I found before is all fixed but I found two new things that need addressing :(

solc/CommandLineInterface.cpp
leonardoalt: Standard JSON is tricky because we can't be sure whether it's running from the command line or the browser, for example.

libsolidity/interface/SMTSolverCommand.cpp
leonardoalt: Yea good point. Fixed it.

solc/CommandLineInterface.cpp
leonardoalt: Running via Standard JSON with `eld` we get:
```
"Warning: CHC analysis was not possible. No Horn solver was available. None of the installed solvers was enabled.\n\n"
```

libsolidity/interface/UniversalCallback.h
```diff
@@ -0,0 +1,53 @@
...
+{
+public:
+	UniversalCallback(FileReader&  _fileReader, SMTSolverCommand& _solver) :
+		m_fileReader{_fileReader},
+		m_solver{_solver}
+	{}
+
+	frontend::ReadCallback::Callback callback()
+	{
+		return [this](std::string const& _kind, std::string const& _data) -> ReadCallback::Result {
```
nikola-matic, line 39: Is there a reason `_kind` could not be `ReadCallbackKind::Kind` instead of a string?

leonardoalt: let me check

leonardoalt: good point, a few places are doing `readFile(ReadCallback::kindString(ReadCallback::Kind::ReadFile), ...)` for the `FileReader` callback which looks silly, so will change to `Kind` directly.

nikola-matic: Sure thing - however if this requires changing it in too many places, feel free to open a separate issue for this in case you don't want to pollute this PR with a refactor.

leonardoalt: I think the reason was to keep the callback generic actually, as in you can pass in any kind you want, cus ultimately this callback may call something outside the compiler (the callback can be given via the C interface of the compiler, which `solcjs` uses also for file stuff). So I wouldn't change it in the `ReadCallback` itself, but here it might make sense to change it.

nikola-matic: Generic is fine, but passing the callback type as a string is just weird to me.

leonardoalt: So, this requires deeper changes in `FileReader` and `ReadFile` as well, so I wouldn't touch all that in this PR if it's fine.

nikola-matic: Yup, sounds good, don't worry about.

solc/CommandLineInterface.cpp
cameel: ok, we do not exactly do nothing because part of it is handled in `ModelChecker`. So the warning will still appear.

I checked with solc-js and I also get the message about `eld` not being available in the system, but I think that may be confusing to the user. It does not make it clear that it's not allowed, user is likely to think he's just doing it wrong and waste time tinkering with it. After all, with `--standard-json` on the CLI it technically could work so I think we should really have a validation against `eld` being selected as one of the solvers in Standard JSON.

And anyway, using filesystem stuff like `boost::process::search_path("eld")` anywhere outside of the CLI (e.g. in `ModelChecker.cpp`) is questionable in general. Can we even assume that emscripten will always handle that in a reasonable way? To be honest, I'm surprised that it does not raise some kind of exception or crash.

test/TemporaryDirectoryTest.cpp
```diff
@@ -16,9 +16,9 @@
 */
 // SPDX-License-Identifier: GPL-3.0
```
cameel, line 18: I'd move this test file to `test/libsolutil/`.

test/cmdlineTests/standard_model_checker_divModSlacks_default_all/output.json
```diff
@@ -1,76 +1,4 @@
 {
-    "auxiliaryInputRequested":
-    {
-        "smtlib2queries":
-        {
-            "0x72b32f47722af7f889ab5740b229c1799f27d0a693995a581c00f65438255a15": "(set-option :produce-models true)
```
cameel, line 6: These auxiliary inputs disappeared because smtlib is now not enabled by default I guess?

leonardoalt: Exactly.

solc/CommandLineInterface.cpp
leonardoalt: Ah yea true... that's actually the reason why we moved it entirely into the CLI before I added that eld bit. Sigh this is so annoying.

leonardoalt: @cameel we can't easily do such validation and modification of the JSON input in the CLI because there it's still just a string. Here's an idea I talked about with Chris today:

- we don't do any of the CLI shenanigans, no internal flag in settings saying whether the `eld` command is available or not
- if the user selected `eld` as a solver, we assume it's available and create the proper `UniversalCallback`
- when we make the call to the cmd, we `ifdef` check whether we're running in emscripten mode and don't even try to make the call if yes
- otherwise we make the call, and if it turns out the binary is not available it fails properly, which is fine

cameel: Sounds good!

I'd just add that if the `ifdef` is false, the message should be different (i.e. say that eld is not usable under emscripten rather than say that the binary was not found).

libsolidity/formal/ModelChecker.cpp
```diff
@@ -146,3 +155,43 @@ solidity::smtutil::SMTSolverChoice ModelChecker::availableSolvers()
...
+		);
+	}
+
+	if (_enabled.eld && !availableSolvers.eld)
+	{
+		_enabled.eld = false;
+		_errorReporter.warning(
+			4458_error,
+			SourceLocation(),
+			"Solver Eldarica was selected for SMTChecker but it was not found in the system."
```
cameel, line 180: Like I said in https://github.com/ethereum/solidity/pull/12969#discussion_r1029406607, can we give a different message here when the `ifdef` is false?

libsolidity/formal/ModelChecker.cpp
```diff
@@ -135,9 +141,12 @@ vector<string> ModelChecker::unhandledQueries()
 	return m_bmc.unhandledQueries() + m_chc.unhandledQueries();
 }
 
-solidity::smtutil::SMTSolverChoice ModelChecker::availableSolvers()
+SMTSolverChoice ModelChecker::availableSolvers()
 {
 	smtutil::SMTSolverChoice available = smtutil::SMTSolverChoice::SMTLIB2();
+#if defined(__linux) || defined(__APPLE__)
```
cameel, line 147: Don't we also need one for windows here? What about the `EMSCRIPTEN` flag that Chris mentioned - does it exist?

cameel: Looks like we at least have it as a CMake option:
https://github.com/ethereum/solidity/blob/3109ce2dbcb07606df0b177b8e0378f6d6b80835/CMakeLists.txt#L139-L141

libsolidity/formal/ModelChecker.cpp
```diff
@@ -20,6 +20,13 @@
 #ifdef HAVE_Z3
 #include <libsmtutil/Z3Interface.h>
 #endif
+#ifdef HAVE_Z3_DLOPEN
+#include <z3_version.h>
+#endif
+
+#if defined(__linux) || defined(__APPLE__)
```
cameel, line 27: Same here. `EMSCRIPTEN`?

solc/CommandLineInterface.cpp
```diff
@@ -692,7 +692,10 @@ void CommandLineInterface::compile()
 {
 	solAssert(CompilerInputModes.count(m_options.input.mode) == 1);
 
-	m_compiler = make_unique<CompilerStack>(m_fileReader.reader());
+	if (m_options.modelChecker.settings.solvers.eld)
+		m_compiler = make_unique<CompilerStack>(m_universalCallback.callback());
+	else
+		m_compiler = make_unique<CompilerStack>(m_fileReader.reader());
```
cameel: If we're always using the universal callback for StandardJSON I guess we could just as well do it for CLI too to simplify things? I.e. SMT checker won't use the callback anyway if eld is not enabled, right?

libsolidity/formal/ModelChecker.cpp
leonardoalt: we don't really support using the smtchecker on windows anyway without compiling the compiler yourself, and we don't test on windows, so left it out for now. That's also why I didn't try the EMSCRIPTEN one

libsolidity/formal/ModelChecker.cpp
leonardoalt: Same explanation.

solc/CommandLineInterface.cpp
leonardoalt: Yep, wanted to do that after you checked the current version

libsolidity/formal/ModelChecker.cpp
leonardoalt: added

libsolidity/formal/ModelChecker.cpp
cameel: ok. I guess it's fine then as long as the error message says that eld is not supported on this platform (rather than emscripten specifically).

cameel (dismissed):
LGTM

cameel (approved):

PR #12971, opened 2022-04-29
fix(parser): error for unexpected token

shekhirin:
With adding custom error definitions, error message for unexpected token on the source unit level wasn't updated.


chriseth (approved):
Cool, thanks a lot!

PR #12974, opened 2022-04-29
update smt tests z3 4.8.16

leonardoalt:
I have NO IDEA why the new version is outputting so many new invariants


blishko (commented):
@leonardoalt , there is something weird about all this invariants.
From what I have seen in the diff, it looks like it adds `|| true` to the invariants it discovered Which is weird because if you add `or true` to a formula you basically get just `true`. Which is an invariant but not very useful.
Are you sure this should be happening?

Also there are contract invariants now that look like `true || true || true`, which again, does not seem very useful...

leonardoalt:
> @leonardoalt , there is something weird about all this invariants. From what I have seen in the diff, it looks like it adds `|| true` to the invariants it discovered Which is weird because if you add `or true` to a formula you basically get just `true`. Which is an invariant but not very useful. Are you sure this should be happening?
> 
> Also there are contract invariants now that look like `true || true || true`, which again, does not seem very useful...

Yea also noticed that they look weird. I haven't yet checked if the weirdness is coming from the solver or a bug on our side. However the `|| true` stuff is not coming only from this version, many invariants in our tests already had that stuff before.

hrkrshnn (approved):

Sticknmove (commented):
Wow thanks

Sticknmove:
Thanks

Sticknmove:
Thanks

PR #12976, opened 2022-05-01
Add simplification rules for `mod(mul(X, Y), A)` & `mod(add(X, Y), A)`

### LINKED ISSUE #10688: Convert certain mod(mul(X, Y), A) into mulmod(X, Y, A)
If `A` is a constant that divides `2**256` (so basically of the form `2**N`), then the two expressions should be the same. `mod(mul(X, Y), A)` costs 10 and `mulmod(X, Y, A)` costs `8` (excluding the shorter deploy code.)

dflupu:
This fixes https://github.com/ethereum/solidity/issues/10688

I'm not sure about the placement of these rules, it just seemed like they should be somewhere above the following rule: `Replace MOD X, <power-of-two> with AND X, <power-of-two> - 1`

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,8 @@ Language Features:
 Compiler Features:
  * TypeChecker: Support using library constants in initializers of other constants.
  * Yul IR Code Generation: Improved copy routines for arrays with packed storage layout.
+ * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
+ * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
 
 
 Bugfixes:
--- a/libevmasm/RuleList.h
+++ b/libevmasm/RuleList.h
@@ -275,6 +275,7 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart4_5(
 
 template <class Pattern>
 std::vector<SimplificationRule<Pattern>> simplificationRuleListPart5(
+	bool _forYulOptimizer,
 	Pattern A,
 	Pattern B,
 	Pattern,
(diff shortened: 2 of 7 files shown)


hrkrshnn:
Let me rebase this to see if the tests pass.

wechman:
I have rebased changes to make clear what still needs to be fixed.

ekpyron:
The assertion hit by the tests is ``libevmasm/CommonSubexpressionEliminator.cpp:L386`` - I'd actually say that it's not worth the effort to actually implement that part for legacy code generation - so the easiest solution to this would be to only add these simplification rules for via-IR code generation (which would require passing that down as a flag to the rule list generation).

dflupu:
> The assertion hit by the tests is `libevmasm/CommonSubexpressionEliminator.cpp:L386` - I'd actually say that it's not worth the effort to actually implement that part for legacy code generation - so the easiest solution to this would be to only add these simplification rules for via-IR code generation (which would require passing that down as a flag to the rule list generation).

Thanks for clarifying! The output wasn't really telling me much.

I could make that change. Do you have a name in mind for the flag? And do you recommend passing that flag down to each `simplificationRuleListPart` or just adding another "Part" that is wrapped in an if statement?

I notice that there's a few rules, `evmRuleList` to be specific, that are added if the `_evmVersion` is set (it is only set from libyul).  I figure that there is the option of making use of that. The only issue is that those rules are added at the end.

ekpyron:
> Thanks for clarifying! The output wasn't really telling me much.
> 
> I could make that change. Do you have a name in mind for the flag? And do you recommend passing that flag down to each `simplificationRuleListPart` or just adding another "Part" that is wrapped in an if statement?
> 
> I notice that there's a few rules, `evmRuleList` to be specific, that are added if the `_evmVersion` is set (it is only set from libyul). I figure that there is the option of making use of that. The only issue is that those rules are added at the end.

Ah, right! I didn't realize myself that we don't set the evm version for legacy code generation. In that case, it makes sense to just reuse that. You can add a flag only to the rule list ``Part`` in which you add the new rules only and set it based on ``_evmVersion.has_value()``. As for a name, maybe something like ``_forYulOptimizer``.

I'd also add a comment before conditionally adding the new rules stating something like:
```
// The libevmasm optimizer does not support rules resulting in opcodes with more than two arguments.
```

dflupu:
@ekpyron Looks like that solved it! Tests are passing now 🙌

ekpyron:
We'll still need a changelog entry for this - apart from that I think we can merge.

dflupu:
I've added changelog entries as well. Feel free to edit if necessary.

ekpyron (dismissed):

wechman (approved):

PR #12977, opened 2022-05-02
change error message for Unicode character in non-unicode string literal

### LINKED ISSUE #12968: Error message that results from using Unicode in string literal is confusing
## Description

Suppose I try to compile the following contract:

```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Repro {
  function run() public pure returns (string memory) {
    return "¡¡¡";
  }
}
```

I get the following error message:

```
ParserError: Invalid character in string.
 --> project:/contracts/Repro.sol:6:12:
  |
6 |     return "¡¡¡";
  |            ^^
```

Now, I know Solidity well enough to realize that, oh right, this is because Unicode strings have to be prefixed with the word `unicode`.  However, someone newer to Solidity, who hasn't read all the documentation, likely wouldn't realize this, and would likely conclude that Unicode strings simply aren't allowed at all.

The error message should be rewriten to point out the existence of `unicode` strings. Say something like, "Unicode character in non-unicode string literal; literal must be marked as unicode" or something like that.

## Environment

- Compiler version: 0.8.13
- Framework/IDE (e.g. Truffle or Remix): Truffle

## Steps to Reproduce

See above.

super1ha1:
Fixes #12968.

change error message for Unicode character in non-unicode string literal

### DIFF
--- a/liblangutil/Scanner.cpp
+++ b/liblangutil/Scanner.cpp
@@ -76,7 +76,7 @@ string to_string(ScannerError _errorCode)
 		case ScannerError::IllegalHexDigit: return "Hexadecimal digit missing or invalid.";
 		case ScannerError::IllegalCommentTerminator: return "Expected multi-line comment-terminator.";
 		case ScannerError::IllegalEscapeSequence: return "Invalid escape sequence.";
-		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
+		case ScannerError::IllegalCharacterInString: return "Invalid character in string. Unicode character in non-unicode string literal; literal must be marked as unicode";
 		case ScannerError::IllegalStringEndQuote: return "Expected string end-quote.";
 		case ScannerError::IllegalNumberSeparator: return "Invalid use of number separator '_'.";
 		case ScannerError::IllegalExponent: return "Invalid exponent.";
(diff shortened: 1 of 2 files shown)


liblangutil/Scanner.cpp
```diff
@@ -76,7 +76,7 @@ string to_string(ScannerError _errorCode)
 		case ScannerError::IllegalHexDigit: return "Hexadecimal digit missing or invalid.";
 		case ScannerError::IllegalCommentTerminator: return "Expected multi-line comment-terminator.";
 		case ScannerError::IllegalEscapeSequence: return "Invalid escape sequence.";
-		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
+		case ScannerError::IllegalCharacterInString: return "Invalid character in string. Unicode character in non-unicode string literal; literal must be marked as unicode";
```
chriseth, line 79: This is used for any non-printable ascii character, so we cannot know that the user intended this to be unicode. Maybe something like `"If you want to use unicode characters, use a unicode\"...\" string literal."`

test/libsolidity/syntaxTests/string/string_unicode_without_prefix.sol
```diff
@@ -4,4 +4,4 @@ contract test {
     }
```
cameel, line 4: Just as @chriseth said, this message could also be printed for example for [control characters](https://en.wikipedia.org/wiki/Control_character). Could you add a test that includes some?

Issue #12650, opened 2022-02-08
Constant-complexity dispatcher idea


k06a:
Today I spent some time on prototyping and was managed to achieve O(1)-complexity dispatcher. Check out few my messages here: https://github.com/vyperlang/vyper/issues/2386#issuecomment-1033117651

And find code prototype here: https://gist.github.com/k06a/737fd1f389bb6f2e66f0ef76ea73729b

## Abstract

I believe dispatcher implementation initially was O(N), but now it may be tree-like O(log(n)). I assume having an O(1) dispatcher when some contract has 5+ methods is a cool achievement.

## Motivation

Contracts with tens of methods could have really slow dispatcher for most of the methods. Computing keccak256 and 2 jumps is a relatively cheap way to dispatch.

## Specification

We just need to brute-force `magic` values at compile time to achieve all selectors being mapped into 0..N-1 values. The complexity of such brute force should be `N^N/N!`. If the number of selectors is really huge we could, for example, find the first function for splitting into a few groups by 8, and then use different magic for each group. Having 1 magic number you could split 32 selectors into 4 groups by 8 and then find magic for each group. Finding magic for a group of 8 would take `8^8/8! = 416` iterations in average. Seems possible to do this in compile-time.

Pseudocode:
```solidity
const index = keccak256(abi.encodePacked(selector + (magic << 32))) % selectors.length;
JUMP(READ(table_start + index));
```

## Backwards Compatibility

The resulting dispatcher code should be fully compatible for external calls with any other versions of the dispatcher.

k06a:
I will write detailed specification, hope you like the idea and have desire to implement it. I assume dispatching can cost 100-150 gas for up to 512 methods and bytecode size could be smaller than now.

chriseth:
I actually spent some time months ago to do something similar.
I would say: Why not?
One problem is that we still need to do the final comparison in the end, but we already do that even in the binary search case.
Another problem might be if we want to get rid of dynamic jumps at some point. Oh and this won't be compatible with code generation via Yul unless...
we implement this for the yul to evm code generation for `switch` statements in general :)

chriseth:
Ah, maybe I misread your algorithm: Why do you need `keccak`? Shouldn't the following also work:

```
let n := mod(byte(0, mul(X, calldataload(0))), N)
codecopy(0, add(start, mul(n, 4)))
let d := shr(240, mload(0))
jump(d)
```

[...]

PR #12978, opened 2022-05-02
Use jump table for some switch statements

chriseth:


philbawa:
@philbawa hi, how do I transfer BNB

chriseth:
This is looking really good already! We have to be careful to find a good balance between this being generic enough and not too complicated, but I would suggest the following items (that could maybe go into future pull requests):
 - implement this in optimized evm transform
 - use gas measurements and the "runs" parameter to decide whether or not to do the optimization
 - allow cases that are not exactly "enum-like" - for example it might not be such a big deal if some of the values are missing - we just use the same tag as the default case
 - think about how we could use this in the function dispatch - see https://github.com/ethereum/solidity/issues/12650

jaa2:
Item 3 is now completed. Any range of cases spanning no more than 16 values is now supported by subtracting the minimum case if necessary (underflows desired), and empty cases are allowed to be stored in the pushed value, which jump to the default case.

chriseth:
I'm not sure if that was already your reason for the relative offsets, but if we use relative offsets and can ensure them to be at most 256, we can actually fit 32 cases in a single push constant instead of just 16.

Also, @ekpyron suggested that as an alternative, we could also use a fully fixed-layout jump table instead, which would allow us to have no restriction on the size. The downside is that it needs another jump for each case and 4 additional bytes per case
```
let x :=<switch value>
if ge(x, <max value>) { /* something else */ }
jump(add(<start>, mul(x, <byte size of case>)))
// for each case:
jumpdest
pushtag <actual tag>
jump
```

jaa2:
The primary reason I included relative offsets was to eliminate a conditional jump for gas savings, but you're right, we could fit up to 32 cases in a single push, especially when the cases are sparse. The reason I haven't done it that way yet is for architecture reasons: the jump table code itself is currently built in EVMCodeTransform, and I wanted to make the fewest assumptions about the addresses where the cases begin. For instance, if we were to handle 32 cases and a case's address is >= 256 from the default case, the compilation would fail when deciding the jump addresses because the relative offset would not fit in a single byte.

But the compilation could currently fail anyway with backward jumps because the offsets are relative. Those aren't possible with the unoptimized version since all non-default cases come after the default case, but maybe with the optimizer that will become an issue.

Is there a way I can shift the assembly code generation of the jump table to be done after the case addresses are known? That way we could then decide if we could fit 32 cases and also whether we need to handle backward jumps. The code size would vary, so I'm not sure how I would handle that properly, maybe by taking up the maximum possible size and using some instruction for no-ops if the actual size is less.

> fully fixed-layout jump table

Ha! That was my first idea. It is still limited by the range of the cases (we can't fit cases spanning a range of 20,000) and uses more gas (59 rather than 35), but if you have a larger range of cases it could be useful.

libevmasm/Assembly.cpp
```diff
@@ -574,6 +576,23 @@ LinkerObject const& Assembly::assemble() const
 			ret.bytecode.resize(ret.bytecode.size() + bytesPerTag);
 			break;
 		}
+		case JumpTablePushTag:
+		{
+			unsigned pushSize = (unsigned) i.jumpTableTags().size() * bytesPerJumpTableTag;
+			if (pushSize == 0)
```
ekpyron: Can this happen? Should we maybe rather assert that there's at least one tag already when constructing JumpTablePushTag items?

libevmasm/JumpdestRemover.cpp
```diff
@@ -59,11 +59,27 @@ set<size_t> JumpdestRemover::referencedTags(AssemblyItems const& _items, size_t
...
+		case PushTag:
 		{
 			auto subAndTag = item.splitForeignPushTag();
 			if (subAndTag.first == _subId)
 				ret.insert(subAndTag.second);
+			break;
+		}
+		case JumpTablePushTag:
+		{
+			// TODO: Do we need to look at the sub here?
```
ekpyron: Jump table tags should always refer to the current sub-assembly, shouldn't they? We may want to assert this here, but don't need more than that.

libevmasm/Assembly.cpp
jaa2: I tried to write this in a way that was as permissive as possible, so technically a jump table with no cases would work here. But since there is not much use for a zero-case jump table, I have added asserts when constructing a JumpTablePushTag and when doing the final assembly.

libevmasm/JumpdestRemover.cpp
jaa2: That's what I was thinking, but I wasn't exactly sure of the sub-assembly architecture. Thank you for confirming that. I don't store the sub in the JumpTablePushTag anyway, so I just removed the comment.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -341,6 +373,47 @@ void ControlFlowGraphBuilder::operator()(Switch const& _switch)
...
+	map<u256, CFG::BasicBlock*> cases;
+	for (Case const& _case: _switch.cases)
+	{
+		if (_case.value)
+		{
+			u256 caseVal = valueOfLiteral(*_case.value);
+			cases[caseVal] = &m_graph.makeBlock(debugDataOf(_case.body));
+	std::cout << "Visiting case value " << caseVal << std::endl;
+			m_currentBlock = cases[caseVal];
+			(*this)(_case.body);
```
ekpyron: After visiting each case (also the default case above), you'll need to tie things up by jumping to ``afterSwitch``.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -341,6 +373,47 @@ void ControlFlowGraphBuilder::operator()(Switch const& _switch)
...
+		}
+	}
+
+	m_currentBlock = switchBlock;
+	makeSwitch(debugDataOf(_switch), switchExpr, defaultCaseBlock,
+		cases, afterSwitch);
```
ekpyron: In case you have a default case, the switch itself never jumps to ``afterSwitch``, so I think ``afterSwitch`` should only be conditionally in a switch just in case that there is no default case.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -493,6 +527,15 @@ list<pair<CFG::BasicBlock const*, CFG::BasicBlock const*>> StackLayoutGenerator:
...
 			[&](CFG::BasicBlock::FunctionReturn const&) {},
 			[&](CFG::BasicBlock::Terminated const&) {},
+			[&](CFG::BasicBlock::Switch const& _switch)
+			{
+				// TODO: Confirm that there are no backwards jumps here
+				if (_switch.defaultCase != nullptr)
+					_addChild(_switch.defaultCase);
+				for (auto const& [caseValue, caseBlock]: _switch.cases)
+					_addChild(caseBlock);
+				_addChild(_switch.target);
```
ekpyron: Just to note it here again: it's only possible to directly jump to the target here, if there is no default case, so if there is a default case, ``_switch`` should not contain a ``target``.
Alternatively, you could reuse the same field in ``Switch`` for both and call it ``defaultCaseOrAfterSwitch`` (or just ``defaultCase`` with a comment that it will point to the block after the switch, in case there was no explicit default case).

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -810,6 +921,46 @@ void StackLayoutGenerator::fillInJunk(CFG::BasicBlock const& _block)
 			},
 			[&](CFG::BasicBlock::FunctionReturn const&) {},
 			[&](CFG::BasicBlock::Terminated const&) {},
+			[&](CFG::BasicBlock::Switch const& _switch)
+			{
+				/*std::vector<CFG::BasicBlock*> exits;
```
ekpyron, line 898: This part will be less of a concern after https://github.com/ethereum/solidity/pull/12132 (see https://github.com/ethereum/solidity/pull/12132#discussion_r894413634). But it's not crucial to reproduce this for switches in a first implementation at all, so you're fine leaving this commented, resp. removing this and ignoring the ``fillInJunk`` part apart from the child visits below.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -540,6 +583,47 @@ void StackLayoutGenerator::stitchConditionalJumps(CFG::BasicBlock const& _block)
 			},
 			[&](CFG::BasicBlock::FunctionReturn const&)	{},
 			[&](CFG::BasicBlock::Terminated const&) { },
+			[&](CFG::BasicBlock::Switch const& _switch)
+			{
+				/*Stack exitLayout = info.exitLayout;
```
ekpyron: You will need to keep this commented in, this part is crucial.

The commented code actually looks pretty much correct, apart from the fact that (as commented elsewhere), you don't jump to the target if there is a default case here.

ekpyron:
For debugging the optimized evm code generation path, it may be good to know that you can visualize the control flow graphs and stack layouts by running our tests:

You need to have something installed that can consume graphviz dot files on stdin - I usually use ``xdot`` for that.

Then you can run (assuming a linux system, otherwise however you can set environment variables):
```
export ISOLTEST_DISPLAY_GRAPHS_ON_SUCCESS_COMMAND="xdot -"
export ISOLTEST_DISPLAY_GRAPHS_ON_FAILURE_COMMAND="xdot -"
build/test/tools/isoltest -t yulControlFlowGraph*
```
to get visualization of all control flow graph tests and

```
export ISOLTEST_DISPLAY_GRAPHS_SUCCESS="xdot -"
export ISOLTEST_DISPLAY_GRAPHS_FAILURE="xdot -"
build/test/tools/isoltest -t yulStackLayout*
```
to get visualization of all stack layout tests.
(It should actually be the same environment variables in both cases, we should probably align them :-))

The problem there, of course, is that those test runs won't understand the new ``Switch`` exit, so you will only be able to use this on ``develop``, until you extend ``test/libyul/ControlFlowGraphTest.cpp`` and ``test/libyul/StackLayoutGeneratorTest.cpp`` to handle the new switch exit... but it may still be helpful to look at the existing graphs and stack layouts. And adjusting those tests may help in properly building the graphs and stack layouts by allowing to visualize them.

libyul/backends/evm/StackLayoutGenerator.cpp
jaa2: The reason I stored the two separately to begin with is that in the most optimized jump table, the default case block must come before all the other case blocks, since offsets from the default case block are stored in the jump table. But I now realize that it's also possible to arrange the jump table the other way around, i.e. to require the default case block to come after all the other case blocks and instead of storing a positive offset, we store negative offsets in the jump table. This way, we can neatly keep the default case block and `target` together in the way you suggested.

ekpyron: I actually hadn't considered that, but yeah :-)!

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -497,7 +497,133 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
...
+					m_assembly.appendInstruction(evmasm::Instruction::POP);
+					m_stack.pop_back();
+					createStackLayout(debugDataOf(_switch), m_stackLayout.blockInfos.at(_switch.defaultCase).entryLayout);
+					std::cout << "Checking default case..." << std::endl;
+					assertLayoutCompatibility(m_stack, m_stackLayout.blockInfos.at(_switch.defaultCase).entryLayout);
+					(*this)(*_switch.defaultCase);
+				}
+			}
+			else
+				m_assembly.appendJumpTo(m_blockLabels[_switch.target]);
```
ekpyron: I'd expect something like
```suggestion
{
					ScopeGuard stackRestore([storedStack = m_stack, this]() {
						m_stack = move(storedStack);
						m_assembly.setStackHeight(static_cast<int>(m_stack.size()));
					});
					m_assembly.appendInstruction(evmasm::Instruction::POP);
					m_stack.pop_back();
					createStackLayout(debugDataOf(_switch), m_stackLayout.blockInfos.at(_switch.target).entryLayout);
					assertLayoutCompatibility(m_stack, m_stackLayout.blockInfos.at(_switch.target).entryLayout);
					(*this)(*_switch.target);
}
```
here

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -341,6 +376,49 @@ void ControlFlowGraphBuilder::operator()(Switch const& _switch)
 {
 	yulAssert(m_currentBlock, "");
 	shared_ptr<DebugData const> preSwitchDebugData = debugDataOf(_switch);
+	auto switchBlock = m_currentBlock;
+	std::cout << "CFGB: Visiting switch expression" << std::endl;
+	StackSlot switchExpr = std::visit(*this, *_switch.expression);
```
ekpyron:
```suggestion
	std::cout << "CFGB: Visiting switch expression" << std::endl;
	StackSlot switchExpr = std::visit(*this, *_switch.expression);
	auto switchBlock = m_currentBlock;
```

The switch block only really starts *after* the expression is fully evaluated - like the ``if`` condition is part of the block *before* the ``if``, the switch expression is part of the block *before* the ``switch``.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -497,7 +497,133 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
 			CFG::BuiltinCall const* builtinCall = get_if<CFG::BuiltinCall>(&_block.operations.back().operation);
 			yulAssert(builtinCall, "");
 			yulAssert(builtinCall->builtin.get().controlFlowSideEffects.terminatesOrReverts(), "");
-		}
+		},
+		[&](CFG::BasicBlock::Switch const& _switch)
+		{
+			std::cout << "In OptimizedEVMCodeTransform: creating switch statement" << std::endl;
```
ekpyron: You will need
```
		   createStackLayout(debugDataOf(_switch), blockInfo.exitLayout);
```
here, which will be the optimal layout from which all the switch cases can be generated cheaply below.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -497,7 +497,133 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
...
+					ScopeGuard stackRestore([storedStack = m_stack, this]() {
+						m_stack = move(storedStack);
+						m_assembly.setStackHeight(static_cast<int>(m_stack.size()));
+					});
+					m_assembly.appendInstruction(evmasm::Instruction::POP);
+					m_stack.pop_back();
+					std::cout << "Adding case block " << caseValue << std::endl;
+					createStackLayout(debugDataOf(_switch), m_stackLayout.blockInfos.at(caseBlock).entryLayout);
+					assertLayoutCompatibility(m_stack, m_stackLayout.blockInfos.at(caseBlock).entryLayout);
+					(*this)(*caseBlock);
```
ekpyron: Actually the real problem, I think is here (and actually similarly for the default and target cases above).

This should actually just be nothing else than

```suggestion
					(*this)(*caseBlock);
```

The main problem here is that only the ``(*this)(*caseBlock);`` will actually emit the label for the block ``caseBlock`` that you jumped to.
So the jump you generate earlier will just jump right past the ``POP`` you generated above and thus have a malformed stack.

The validation goes through since you're also changing ``m_stack``, in a sense pretending like you did not jump with the layout you had when you actually jumped above, but one less slot. That's something to watch out for: the ``OptimizedCodeTransform`` really must keep ``m_stack`` valid, even across jumps, otherwise all the validation it does is worthless :-).

However, this also means that in the ``StackLayoutGenerator`` you will have to keep the switch expression in the entry layout of the case blocks, I'll add another comment about that there.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -540,6 +594,49 @@ void StackLayoutGenerator::stitchConditionalJumps(CFG::BasicBlock const& _block)
 			},
 			[&](CFG::BasicBlock::FunctionReturn const&)	{},
 			[&](CFG::BasicBlock::Terminated const&) { },
+			[&](CFG::BasicBlock::Switch const& _switch)
+			{
+				Stack exitLayout = info.exitLayout;
+				// The last block must have produced the switch expression at the stack top.
+				yulAssert(!exitLayout.empty(), "");
+				yulAssert(exitLayout.back() == _switch.switchExpr, "");
+				// The switch expression is consumed by the switch.
```
ekpyron: Given my other comment, I think it's easier to generate layouts in which each switch case still *has* a copy of the switch expression in its entry layout - since that's how you'll actually end up jumping to them.
Not sure if this is the only place that needs adjustments due to that - but once you adjust the ``StackLayoutGeneratorTest``s you should be able to confirm if you get layouts that have the temporary switch expression slot as entires of the switch cases.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -497,7 +497,133 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
...
+		[&](CFG::BasicBlock::Switch const& _switch)
+		{
+			std::cout << "In OptimizedEVMCodeTransform: creating switch statement" << std::endl;
+
+			// Create labels for the targets, if not already present.
+			if (_switch.defaultCase != nullptr && !m_blockLabels.count(_switch.defaultCase))
+				m_blockLabels[_switch.defaultCase] = m_assembly.newLabelId();
+			for (auto const& [caseValue, caseBlock]: _switch.cases)
+				if (!m_blockLabels.count(caseBlock))
+					m_blockLabels[caseBlock] = m_assembly.newLabelId();
```
ekpyron, line 513: Thinking about it, there is one different option, other than adding the switch expression back to the stack layouts of the switch cases:

Instead of creating the actual block labels here, you could just store "case dispatch" labels in a vector and jump to those, and then below, emit those labels, pop the switch expression and only then visit the case blocks. That way you could keep the case stack layouts free of the switch expression and we don't need to readjust things for the jump table implementation.

But either way is fine - we'll in any case first need a traditional implementation with switch exits that works reliably before we can move to jump tables.

ekpyron: But yeah, that may be nicer - conceptually the case blocks doesn't really *want* the switch expression in their entry layout, so the fact that a traditional switch dispatch needs them is more of an implementation detail of the switch dispatch.

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
jaa2: Added in [e8a3f92](https://github.com/ethereum/solidity/pull/12978/commits/e8a3f92578904dddbabacdb86f4c88ebe6a8440f)

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
jaa2: Added in 3d947dd8

jaa2:
The switch block type is now in ControlFlowGraphTest and StackLayoutGeneratorTest. The new graphs seem correct to me, but I haven't yet committed updates to the test cases.

libyul/backends/evm/ControlFlowGraphBuilder.cpp
```diff
@@ -628,3 +687,32 @@ void ControlFlowGraphBuilder::jump(
...
+		m_currentBlock = &target;
+	}
+	else
+	{
+		defaultCase->entries.emplace_back(m_currentBlock);
+		m_currentBlock = nullptr;//&target;
+	}
+	for (auto const& [caseVal, caseBlock]: cases)
+		caseBlock->entries.emplace_back(m_currentBlock);
+}
```
ekpyron:
```suggestion
	if (defaultCase == nullptr)
		target.entries.emplace_back(m_currentBlock);
	else
		defaultCase->entries.emplace_back(m_currentBlock);
	for (auto const& [caseVal, caseBlock]: cases)
		caseBlock->entries.emplace_back(m_currentBlock);
	m_currentBlock = nullptr;
}
```

This is the last thing I saw: all cases need entries from the *switch block*, but you've overwritten ``m_currentBlock`` (not sure why) above, so it's no longer the switch block here.

libyul/backends/evm/StackLayoutGenerator.cpp
```diff
@@ -540,6 +572,41 @@ void StackLayoutGenerator::stitchConditionalJumps(CFG::BasicBlock const& _block)
...
+				for (auto const& [caseValue, caseBlock]: _switch.cases)
+				{
+					auto& caseInfo = m_layout.blockInfos.at(caseBlock);
+					caseInfo.entryLayout = fixJumpTargetEntry(caseInfo.entryLayout);
+					_addChild(caseBlock);
+				}
+
+				_addChild(_switch.defaultCase);
+				for (auto const& [caseValue, caseBlock]: _switch.cases)
+					_addChild(caseBlock);
```
ekpyron: This is redundant

libyul/backends/evm/OptimizedEVMCodeTransform.cpp
```diff
@@ -497,7 +497,60 @@ void OptimizedEVMCodeTransform::operator()(CFG::BasicBlock const& _block)
...
+			{
+				// Restore the stack afterwards for the non-zero case below.
+				ScopeGuard stackRestore([storedStack = m_stack, this]() {
+					m_stack = move(storedStack);
+					m_assembly.setStackHeight(static_cast<int>(m_stack.size()));
+				});
+				createStackLayout(debugDataOf(_switch), m_stackLayout.blockInfos.at(_switch.defaultCase).entryLayout);
+				std::cout << "Checking default case..." << std::endl;
+				assertLayoutCompatibility(m_stack, m_stackLayout.blockInfos.at(_switch.defaultCase).entryLayout);
+				(*this)(*_switch.defaultCase);
```
ekpyron:
```suggestion
				(*this)(*_switch.defaultCase);
```
Similarly to the cases below.

ekpyron:
I've, out of curiosity, just built the latest PR version with additionally my last three comments addressed - and that version in fact passes all of our semantics tests, so that's great!

However, it has some adverse effects on the gas costs in quite a few cases, so it would be nice to look into some cases in more detail, compare the generated assembly and see if we can figure out where the increased cost is coming from and if we can do something about it. My guess would be that either the order of visitation of the blocks in the stack layout generator changed, resulting in less optimal layouts, or that it has something to do with the precise layout of code in the switch visit of the optimized code transform. But it's hard to tell without an in-depth look. It should be possible to get back similar gas results with only a few minor tweaks, though, I hope.

libyul/backends/evm/StackLayoutGenerator.cpp
jaa2: Fixed in [f4c11dd](https://github.com/ethereum/solidity/pull/12978/commits/f4c11dd29942f1deaa45a32fb04be2a92caa5599)

ekpyron:
The external tests actually all passed - here's the relevant part of the benchmarks from using ``scripts/externalTests/benchmark_diff.py`` between the ``summarized-benchmarks-develop-b80f4baa.json`` and ``summarized-benchmarks.json`` artifacts of the ``c_ext_benchmarks`` CI run:

### `ir-optimize-evm+yul`
|           bleeps | **`+0.47% ❌`** |           `0%` | **`+0.03% ❌`** |
|            brink | **`+0.88% ❌`** |                |                |
|        chainlink | **`+0.66% ❌`** | **`+0.64% ❌`** | **`+0.02% ❌`** |
|        elementfi | **`+1.01% ❌`** |                |                |
|              ens |           `0%` | **`+0.71% ❌`** | **`+0.01% ❌`** |
|            euler |           `!V` |           `!V` |           `!V` |
|           gnosis | **`+0.62% ❌`** | **`+0.61% ❌`** | **`+0.02% ❌`** |
|              gp2 | **`+0.87% ❌`** | **`+0.78% ❌`** | **`+0.03% ❌`** |
|  perpetual-pools | **`+0.95% ❌`** | **`+0.93% ❌`** | **`+0.13% ❌`** |
|    pool-together | **`+0.77% ❌`** | **`+0.75% ❌`** |          `+0%` |
|         prb-math | **`+1.32% ❌`** | **`+1.32% ❌`** |           `0%` |
|          trident |  **`+0.7% ❌`** | **`+0.73% ❌`** | **`+0.53% ❌`** |
|          uniswap | **`+1.34% ❌`** | **`+1.24% ❌`** | **`+1.18% ❌`** |
| yield_liquidator | **`+0.68% ❌`** | **`+0.71% ❌`** |          `-0%` |

So there is some work to be done - we'll probably need to look at some assembly diffs to figure out why the new version with the ``Switch`` exit so far generates slightly worse code, but this strongly suggests that the current version already generates *correct* code again!

jaa2:
Current issues:

- The switch CFG type as I implemented it has increased bytecode size and same or slightly worse execution cost for regular switches. I think this is because the Yul insertion technique with ghost variables takes advantage of the optimizer in ways that hard-coded assembly cannot. For example, it does not rely on hard-coded dup3 instructions and can have a slightly more flexible stack layout.
- If we do use the switch CFG type, the stack layouts for the case blocks will be different between regular switches and jump table switches, since after the jump, regular switches will have the switch expression left on the stack while a jump table will not. One way to fix this (as discussed with @ekpyron) is to keep the case block entry stack layouts without the switch expression and simply add a pop instruction before each case block when creating the EVM instructions for the regular switch type. The problem is this will produce suboptimal stack layouts when the case blocks could have been produced without the pop instruction.

I am tempted to put the decision to use a jump table into the CFG building step. That way, the Yul insertion with ghost variables method can still be used to make all non-jump tables and produce the best code for them. The CFG switch type would only be used for jump tables then. Of course, this is much less clean than having all switch types use the same CFG block and producing the code in the code transform step.

ekpyron:
For reference the gas differences of the latest version:
<details>

(The perpetual pools regression is probably merely due to non-determinism)

### `ir-no-optimize`
|:----------------:|--------------:|---------------:|-----------:|
|           bleeps |               |                |            |
|            brink |               |                |            |
|        chainlink |               |                |            |
|           colony |          `0%` |                |            |
|        elementfi |               |                |            |
|              ens |               |                |            |
|            euler |               |                |            |
|           gnosis |               |                |            |
|              gp2 |               |                |            |
|  perpetual-pools |               |                |            |
|    pool-together |               |                |            |
|         prb-math |               |                |            |
|          trident |               |                |            |
|          uniswap |               |                |            |
| yield_liquidator |               |                |            |
|         zeppelin |               |                |            |

### `ir-optimize-evm+yul`
|           bleeps |           `0%` |           `0%` |           `0%` |
|            brink |           `0%` |                |                |
|        chainlink |           `0%` |          `+0%` |          `-0%` |
|        elementfi |           `0%` |                |                |
|              ens |           `0%` |          `+0%` |           `0%` |
|            euler | **`-0.01% ✅`** | **`-0.01% ✅`** | **`-0.09% ✅`** |
|           gnosis |           `0%` |          `-0%` |          `-0%` |
|              gp2 |           `0%` |          `+0%` |          `+0%` |
|  perpetual-pools |           `0%` |          `+0%` | **`+0.02% ❌`** |
|    pool-together |           `0%` |          `-0%` |          `-0%` |
|          trident |           `0%` |          `-0%` |          `-0%` |
|          uniswap |           `0%` |          `+0%` |          `+0%` |
| yield_liquidator |           `0%` |          `-0%` |           `0%` |
|         zeppelin |           `0%` |          `-0%` |          `+0%` |

### `ir-optimize-evm-only`
|:----------------:|--------------:|---------------:|-----------:|
|           bleeps |               |                |            |
|            brink |               |                |            |
|        chainlink |               |                |            |
|           colony |          `0%` |                |            |
|        elementfi |               |                |            |
|              ens |               |                |            |
|            euler |               |                |            |
|           gnosis |               |                |            |
|              gp2 |               |                |            |
|  perpetual-pools |               |                |            |
|    pool-together |               |                |            |
|         prb-math |               |                |            |
|          trident |               |                |            |
|          uniswap |               |                |            |
| yield_liquidator |               |                |            |
|         zeppelin |               |                |            |

### `legacy-no-optimize`
|            brink |           `0%` |                |                |
|        chainlink |           `0%` |                |                |
|        elementfi |           `0%` |                |                |
|            euler | **`-0.02% ✅`** | **`-0.02% ✅`** | **`-0.19% ✅`** |
|           gnosis |           `0%` |          `-0%` |          `+0%` |
|              gp2 |           `0%` |                |                |
|  perpetual-pools |           `0%` |          `-0%` | **`-0.01% ✅`** |
|    pool-together |           `0%` |          `-0%` |          `+0%` |
|          trident |           `0%` |           `0%` |          `+0%` |
|          uniswap |           `0%` |          `+0%` |          `-0%` |
| yield_liquidator |           `0%` |          `+0%` |           `0%` |
|         zeppelin |           `0%` |          `+0%` |          `-0%` |

### `legacy-optimize-evm+yul`
|           bleeps |           `0%` |           `0%` |           `0%` |
|            brink |           `0%` |                |                |
|        chainlink |           `0%` |          `+0%` |          `+0%` |
|        elementfi |           `0%` |                |                |
|              ens |           `0%` |           `0%` |           `0%` |
|            euler | **`-0.03% ✅`** | **`-0.03% ✅`** | **`-0.19% ✅`** |
|           gnosis |           `0%` |          `+0%` |          `-0%` |
|              gp2 |           `0%` |          `+0%` |          `+0%` |
|  perpetual-pools |           `0%` |          `+0%` | **`-0.02% ✅`** |
|    pool-together |           `0%` |          `+0%` |          `+0%` |
|          trident |           `0%` |          `-0%` |          `-0%` |
|          uniswap |           `0%` |          `+0%` |          `+0%` |
| yield_liquidator |           `0%` |          `-0%` |          `-0%` |
|         zeppelin |           `0%` |          `-0%` |          `+0%` |

|            brink |           `0%` |                |                |
|        chainlink |           `0%` |          `+0%` |          `-0%` |
|        elementfi |           `0%` |                |                |
|              ens |           `0%` |          `+0%` |           `0%` |
|            euler | **`-0.02% ✅`** | **`-0.02% ✅`** | **`-0.13% ✅`** |
|           gnosis |           `0%` |          `-0%` |          `-0%` |
|              gp2 |           `0%` |           `0%` |          `+0%` |
|  perpetual-pools |           `0%` |           `0%` |          `+0%` |
|    pool-together |           `0%` |           `0%` |          `+0%` |
|          trident |           `0%` |           `0%` |          `-0%` |
|          uniswap |           `0%` |           `0%` |          `+0%` |
| yield_liquidator |           `0%` |          `+0%` |           `0%` |
|         zeppelin |           `0%` |          `-0%` |          `+0%` |


`!V` = version mismatch

</details>

So it already kicks in without us e.g. changing the external dispatch to something that allows jump tables.

I hope I'll have time for a closer look at the implementation soon.

PR #12983, opened 2022-05-04
Recommend `via-ir` whenever a stack too deep error is encountered.

### LINKED ISSUE #12912: Suggest trying to compile via IR in the stack-too-deep error message in legacy compilation.
Since it seems like via IR compilation works in a lot of such cases these days, this might help against stack-too-deep frustration and help in promoting via-IR codegen.

hrkrshnn:
Closes https://github.com/ethereum/solidity/issues/12912

### DIFF
--- a/libsolidity/codegen/CompilerContext.cpp
+++ b/libsolidity/codegen/CompilerContext.cpp
@@ -43,6 +43,7 @@
 
 #include <libsolutil/Whiskers.h>
 #include <libsolutil/FunctionSelector.h>
+#include <libsolutil/StackTooDeepString.h>
 
 #include <liblangutil/ErrorReporter.h>
 #include <liblangutil/Scanner.h>
@@ -422,7 +423,7 @@ void CompilerContext::appendInlineAssembly(
 			BOOST_THROW_EXCEPTION(
 				StackTooDeepError() <<
 				errinfo_sourceLocation(nativeLocationOf(_identifier)) <<
-				util::errinfo_comment("Stack too deep (" + to_string(stackDiff) + "), try removing local variables.")
+				util::errinfo_comment(util::stackTooDeepString)
 			);
 		if (_context == yul::IdentifierContext::RValue)
 			_assembly.appendInstruction(dupInstruction(static_cast<unsigned>(stackDiff)));
--- a/libsolutil/StackTooDeepString.h
+++ b/libsolutil/StackTooDeepString.h
@@ -1,0 +1,28 @@
+/*
+	This file is part of solidity.
+
+	solidity is free software: you can redistribute it and/or modify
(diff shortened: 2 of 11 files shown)


libsolutil/StackTooDeepString.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <string>
+
+namespace solidity::util
+{
+	static std::string stackTooDeepString =
+		"Stack too deep. "
+		"Try compiling with `--via-ir` (cli) or the equivalent `viaIR: true` (standard JSON). "
+		"Otherwise, try removing local variables.";
```
leonardoalt: This removes the information of how much deeper than allowed it actually is.
I'm fine with that, but consensus?

leonardoalt:
I guess it needs a rebase for the smt tests?

Also, do we not have tests for this?!

libsolidity/codegen/CompilerContext.cpp
```diff
@@ -422,7 +423,7 @@ void CompilerContext::appendInlineAssembly(
 			BOOST_THROW_EXCEPTION(
 				StackTooDeepError() <<
 				errinfo_sourceLocation(nativeLocationOf(_identifier)) <<
-				util::errinfo_comment("Stack too deep (" + to_string(stackDiff) + "), try removing local variables.")
```
chriseth, line 425: If this is always using the same text, does it make sense to add the comment in the constructor of the error?

libsolutil/StackTooDeepString.h
chriseth: I also think it's not such a useful information.
But I think we should suggest to also activate the optimizer, shouldn't we?

hrkrshnn: Just added info about optimizer.

libsolidity/codegen/CompilerContext.cpp
hrkrshnn: What do you mean?

ekpyron (approved):
Something like https://github.com/ethereum/solidity/pull/12983/files#r865705579 would be nice, but at a quick glance, I'm not sure if it'd still properly play together e.g. with ``assertThrow`` and such...
So I'd say let's just merge this as is - might be nice to have in the upcoming release.

ekpyron:
> Also, do we not have tests for this?!

Technically, it's an ICE and we don't test for ICEs, since we usually fix them instead :-D...
Although we might want to think about dragging a proper error reporter to codegen for this... then we could also start emitting info messages about stack limit evasion and such...

ekpyron:
Damn, I thought I merged this before the release... merging it now anyways - we can still do more cosmetics afterwards, but we should definitely have this suggestion in general...

Issue #12603, opened 2022-01-31
Support remappings.txt in LSP


christianparpart:
Seems like vscode-solidity supports the remappings.txt format of dapptools now, too. It would make sense to adopt it then for our LSP server.

See: https://github.com/gakonst/dapptools-template/blob/master/remappings.txt

cameel:
My proposal would be to make it a general feature (not just for LSP). We could add a CLI option for specifying remappings in a file, e.g. `--remappings=<path>` or (`--remappings-file=<path>` as proposed by @christianparpart on gitter). The file would have the same simple structure as `remappings.txt` (i.e. just a plain list of remappings, one per line). Since `remappings.txt` is getting some traction, we could make make it the default value for the LSP to reduce the amount of configuration needed (but IMO there should still be a setting to override it). For plain CLI I'd always require it to be specified explicitly and without it only accept remappings on the CLI as we do now.

This would not even be the first place where we support something like this. The `--libraries` option also accepts a file.

chriseth:
We should be careful here. We have to avoid that the LSP behaves differently than the compiler itself. In addition to that, remappings should be less and less common now that we have include paths.

For now, I think it would be better to handle this mechanism inside the plugins instead of the LSP, at least until we have some more generic project configuration file that can be read by both the lsp and the compiler itself.

hrkrshnn:
Foundry also uses `remappings.txt`. https://github.com/gakonst/foundry. A common template (https://github.com/ZeframLou/foundry-template) uses it, so I think `remappings.txt` will be prevalent.

stevennevins:
It's worth mentioning that while the remappings.txt is still there in ZeframLou's repo, that I think foundry is moving away from it in favor of the foundry.toml file also in the repo: https://github.com/ZeframLou/foundry-template/blob/main/foundry.toml

gakonst:
> My proposal would be to make it a general feature (not just for LSP). We could add a CLI option for specifying remappings in a file, e.g. `--remappings=<path>` or (`--remappings-file=<path>` as proposed by @christianparpart on gitter). The file would have the same simple structure as `remappings.txt` (i.e. just a plain list of remappings, one per line). Since `remappings.txt` is getting some traction, we could make make it the default value for the LSP to reduce the amount of configuration needed (but IMO there should still be a setting to override it). For plain CLI I'd always require it to be specified explicitly and without it only accept remappings on the CLI as we do now.
> 
[...]

PR #12994, opened 2022-05-09
LSP to resolve custom include paths (e.g. Hardhat)

### LINKED ISSUE #12833: LSP: bare imports are not resolved
## Description

I am note sure whether this is actually a bug or rather a new feature.

When having a bare import like: 

```
import "hardhat/console.sol";
```
I get the following error message from the LSP:
```
ParserError: Source "hardhat/console.sol" not found: File not found. Searched the following locations: "/home/<someProjectDir>".
```

The file to import is part of a dependency within `node_modules`. The 'searched' location from the error msg is the project root.

The problem persists in truffle and hardhat projects as far as I can see. I'd also figure that nvim native LSP or Coc are not the issue.

Relative imports seem to be fine.

## Environment

- Compiler / LSP version: 0.8.13
- Target EVM version (as per compiler settings): 0.8.4
- Framework/IDE (e.g. Truffle or Remix): NVIM v0.7.0-dev+1292-ge9b53f3fb + native LSP, Truffle / hardhat
- EVM execution environment / backend / blockchain client: -

## Steps to Reproduce

Create a new sample hardhat typescript project and open `Greeter.sol`

christianparpart:
This PR is WIP, but it will fix #12833 later on.

Big question is, how do we want to tell `solc --lsp` where to look for include paths. I remember there was some project file format that does include that information, but what was that? @cameel, do you remember?

### Checklist:

- [x] factor our file path resolving in FileRepository: `tryResolvePath(.)`
- [x] fix `sourceUnitNameToUri()` by making use of new path resolver.
- [x] custom include paths via JSON settings object
- [x] always add `{projectRoot}/node_modules` to include paths (vscode-solidity is doing that too and users are used to it)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -30,7 +30,8 @@ Compiler Features:
  * LSP: Add rudimentary support for semantic highlighting.
  * Type Checker: Warn about assignments involving multiple pushes to storage ``bytes`` that may invalidate references.
  * Yul Optimizer: Improve inlining heuristics for via IR code generation and pure Yul compilation.
-
+ * Language Server: Always add ``{project_root}/node_modules`` to include search paths.
+ * Language Server: Adds support for configuring ``include-paths`` JSON settings object that can be passed during LSP configuration stage.
 
 Bugfixes:
  * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
--- a/libsolidity/lsp/FileRepository.cpp
+++ b/libsolidity/lsp/FileRepository.cpp
@@ -22,10 +22,13 @@
 #include <libsolutil/StringUtils.h>
 #include <libsolutil/CommonIO.h>
 
+#include <range/v3/algorithm/none_of.hpp>
 #include <range/v3/range/conversion.hpp>
 #include <range/v3/view/transform.hpp>
 
 #include <regex>
+
+#include <boost/algorithm/string/predicate.hpp>
 
 using namespace std;
 using namespace solidity;
@@ -34,25 +37,49 @@ using namespace solidity::frontend;
 
 using solidity::util::readFileAsString;
 using solidity::util::joinHumanReadable;
+using solidity::util::Result;
 
-FileRepository::FileRepository(boost::filesystem::path _basePath): m_basePath(std::move(_basePath))
+FileRepository::FileRepository(boost::filesystem::path _basePath, std::vector<boost::filesystem::path> _includePaths):
+	m_basePath(std::move(_basePath)),
+	m_includePaths(std::move(_includePaths))
 {
+}
+
+void FileRepository::setIncludePaths(std::vector<boost::filesystem::path> _paths)
+{
+	m_includePaths = std::move(_paths);
 }
 
 string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
 {
 	regex const windowsDriveLetterPath("^[a-zA-Z]:/");
 
+	auto const ensurePathIsUnixLike = [&](string inputPath) -> string {
+		if (!regex_search(inputPath, windowsDriveLetterPath))
+			return inputPath;
+		else
+			return "/" + move(inputPath);
+	};
+
 	if (m_sourceUnitNamesToUri.count(_sourceUnitName))
+	{
+		solAssert(boost::starts_with(m_sourceUnitNamesToUri.at(_sourceUnitName), "file://"), "");
 		return m_sourceUnitNamesToUri.at(_sourceUnitName);
+	}
 	else if (_sourceUnitName.find("file://") == 0)
 		return _sourceUnitName;
 	else if (regex_search(_sourceUnitName, windowsDriveLetterPath))
 		return "file:///" + _sourceUnitName;
-	else if (_sourceUnitName.find("/") == 0)
-		return "file://" + _sourceUnitName;
-	else
+	else if (
+		auto const resolvedPath = tryResolvePath(_sourceUnitName);
+		resolvedPath.message().empty()
+	)
+		return "file://" + ensurePathIsUnixLike(resolvedPath.get().generic_string());
+	else if (m_basePath.generic_string() != "/")
 		return "file://" + m_basePath.generic_string() + "/" + _sourceUnitName;
+	else
+		// Avoid double-/ in case base-path itself is simply a UNIX root filesystem root.
+		return "file:///" + _sourceUnitName;
 }
 
 string FileRepository::uriToSourceUnitName(string const& _path) const
@@ -69,6 +96,52 @@ void FileRepository::setSourceByUri(string const& _uri, string _source)
 	m_sourceCodes[sourceUnitName] = std::move(_source);
 }
 
+Result<boost::filesystem::path> FileRepository::tryResolvePath(std::string const& _strippedSourceUnitName) const
+{
+	if (
+		boost::filesystem::path(_strippedSourceUnitName).has_root_path() &&
+		boost::filesystem::exists(_strippedSourceUnitName)
+	)
+		return boost::filesystem::path(_strippedSourceUnitName);
+
+	vector<boost::filesystem::path> candidates;
+	vector<reference_wrapper<boost::filesystem::path const>> prefixes = {m_basePath};
+	prefixes += (m_includePaths | ranges::to<vector<reference_wrapper<boost::filesystem::path const>>>);
+	auto const defaultInclude = m_basePath / "node_modules";
+	if (m_includePaths.empty())
+		prefixes.emplace_back(defaultInclude);
+
+	auto const pathToQuotedString = [](boost::filesystem::path const& _path) { return "\"" + _path.string() + "\""; };
+
+	for (auto const& prefix: prefixes)
+	{
+		boost::filesystem::path canonicalPath = boost::filesystem::path(prefix) / boost::filesystem::path(_strippedSourceUnitName);
+
+		if (boost::filesystem::exists(canonicalPath))
+			candidates.push_back(move(canonicalPath));
+	}
+
+	if (candidates.empty())
+		return Result<boost::filesystem::path>::err(
+			"File not found. Searched the following locations: " +
+			joinHumanReadable(prefixes | ranges::views::transform(pathToQuotedString), ", ") +
+			"."
+		);
+
+	if (candidates.size() >= 2)
+		return Result<boost::filesystem::path>::err(
+			"Ambiguous import. "
+			"Multiple matching files found inside base path and/or include paths: " +
+			joinHumanReadable(candidates | ranges::views::transform(pathToQuotedString), ", ") +
+			"."
+		);
+
+	if (!boost::filesystem::is_regular_file(candidates[0]))
+		return Result<boost::filesystem::path>::err("Not a valid file.");
+
+	return candidates[0];
+}
+
 frontend::ReadCallback::Result FileRepository::readFile(string const& _kind, string const& _sourceUnitName)
 {
 	solAssert(
@@ -83,53 +156,11 @@ frontend::ReadCallback::Result FileRepository::readFile(string const& _kind, str
 			return ReadCallback::Result{true, m_sourceCodes.at(_sourceUnitName)};
 
 		string const strippedSourceUnitName = stripFileUriSchemePrefix(_sourceUnitName);
+		Result<boost::filesystem::path> const resolvedPath = tryResolvePath(strippedSourceUnitName);
+		if (!resolvedPath.message().empty())
+			return ReadCallback::Result{false, resolvedPath.message()};
 
-		if (
-			boost::filesystem::path(strippedSourceUnitName).has_root_path() &&
-			boost::filesystem::exists(strippedSourceUnitName)
-		)
-		{
-			auto contents = readFileAsString(strippedSourceUnitName);
-			solAssert(m_sourceCodes.count(_sourceUnitName) == 0, "");
-			m_sourceCodes[_sourceUnitName] = contents;
-			return ReadCallback::Result{true, move(contents)};
-		}
-
-		vector<boost::filesystem::path> candidates;
-		vector<reference_wrapper<boost::filesystem::path>> prefixes = {m_basePath};
-		prefixes += (m_includePaths | ranges::to<vector<reference_wrapper<boost::filesystem::path>>>);
-
-		auto const pathToQuotedString = [](boost::filesystem::path const& _path) { return "\"" + _path.string() + "\""; };
-
-		for (auto const& prefix: prefixes)
-		{
-			boost::filesystem::path canonicalPath = boost::filesystem::path(prefix) / boost::filesystem::path(strippedSourceUnitName);
-
-			if (boost::filesystem::exists(canonicalPath))
-				candidates.push_back(move(canonicalPath));
-		}
-
-		if (candidates.empty())
-			return ReadCallback::Result{
-				false,
-				"File not found. Searched the following locations: " +
-				joinHumanReadable(prefixes | ranges::views::transform(pathToQuotedString), ", ") +
-				"."
-			};
-
-		if (candidates.size() >= 2)
-			return ReadCallback::Result{
-				false,
-				"Ambiguous import. "
-				"Multiple matching files found inside base path and/or include paths: " +
-				joinHumanReadable(candidates | ranges::views::transform(pathToQuotedString), ", ") +
-				"."
-			};
-
-		if (!boost::filesystem::is_regular_file(candidates[0]))
-			return ReadCallback::Result{false, "Not a valid file."};
-
-		auto contents = readFileAsString(candidates[0]);
+		auto contents = readFileAsString(resolvedPath.get());
 		solAssert(m_sourceCodes.count(_sourceUnitName) == 0, "");
 		m_sourceCodes[_sourceUnitName] = contents;
 		return ReadCallback::Result{true, move(contents)};
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -130,7 +130,7 @@ LanguageServer::LanguageServer(Transport& _transport):
 		{"textDocument/semanticTokens/full", bind(&LanguageServer::semanticTokensFull, this, _1, _2)},
 		{"workspace/didChangeConfiguration", bind(&LanguageServer::handleWorkspaceDidChangeConfiguration, this, _2)},
 	},
-	m_fileRepository("/" /* basePath */),
+	m_fileRepository("/" /* basePath */, {} /* no search paths */),
 	m_compilerStack{m_fileRepository.reader()}
 {
 }
@@ -148,6 +148,26 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
 	m_settingsObject = _settings;
+	Json::Value jsonIncludePaths = _settings["include-paths"];
+	int typeFailureCount = 0;
+
+	if (jsonIncludePaths && jsonIncludePaths.isArray())
+	{
+		vector<boost::filesystem::path> includePaths;
+		for (Json::Value const& jsonPath: jsonIncludePaths)
+		{
+			if (jsonPath.isString())
+				includePaths.emplace_back(boost::filesystem::path(jsonPath.asString()));
+			else
+				typeFailureCount++;
+		}
+		m_fileRepository.setIncludePaths(move(includePaths));
+	}
+	else
+		++typeFailureCount;
+
+	if (typeFailureCount)
+		m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
 }
 
 void LanguageServer::compile()
@@ -155,7 +175,7 @@ void LanguageServer::compile()
 	// For files that are not open, we have to take changes on disk into account,
 	// so we just remove all non-open files.
 
-	FileRepository oldRepository(m_fileRepository.basePath());
+	FileRepository oldRepository(m_fileRepository.basePath(), m_fileRepository.includePaths());
 	swap(oldRepository, m_fileRepository);
 
 	for (string const& fileName: m_openFiles)
@@ -302,7 +322,7 @@ void LanguageServer::handleInitialize(MessageID _id, Json::Value const& _args)
 	else if (Json::Value rootPath = _args["rootPath"])
 		rootPath = rootPath.asString();
 
-	m_fileRepository = FileRepository(rootPath);
+	m_fileRepository = FileRepository(rootPath, {});
 	if (_args["initializationOptions"].isObject())
 		changeConfiguration(_args["initializationOptions"]);
 
--- a/test/libsolidity/lsp/include-paths/using-custom-includes.sol
+++ b/test/libsolidity/lsp/include-paths/using-custom-includes.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+   import "otherlib/otherlib.sol";
+// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ @NotFound
+
+contract MyContract
+{
+}
+// ----
+// using-custom-includes: @NotFound 6275
--- a/test/libsolidity/lsp/other-include-dir/otherlib/otherlib.sol
+++ b/test/libsolidity/lsp/other-include-dir/otherlib/otherlib.sol
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+library OtherLib
+{
+    function f(uint n) public returns (uint) { return n + 1; }
+}
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -1450,7 +1450,7 @@ class SolidityLSPTestSuite: # {{{
         self.setup_lsp(solc)
 
         sub_dirs = filter(
-            lambda filepath: filepath.is_dir(),
+            lambda filepath: filepath.is_dir() and filepath.name != 'node_modules',
             os.scandir(self.project_root_dir)
         )
 
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -1381,6 +1381,34 @@ class SolidityLSPTestSuite: # {{{
         return markers
 
 
+    def test_custom_includes(self, solc: JsonRpcProcess) -> None:
+        self.setup_lsp(solc, expose_project_root=False)
+        solc.send_notification(
+            'workspace/didChangeConfiguration', {
+                'settings': {
+                    'include-paths': [
+                        f"{self.project_root_dir}/other-include-dir"
+                    ]
+                }
+            }
+        )
+        published_diagnostics = self.open_file_and_wait_for_diagnostics(solc, 'include-paths/using-custom-includes')
+
+        self.expect_equal(len(published_diagnostics), 2, "Diagnostic reports for 2 files")
+
+        # test file
+        report = published_diagnostics[0]
+        self.expect_equal(report['uri'], self.get_test_file_uri('using-custom-includes', 'include-paths'))
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 0, "no diagnostics")
+
+        # imported file
+        report = published_diagnostics[1]
+        self.expect_equal(report['uri'], f"{self.project_root_uri}/other-include-dir/otherlib/otherlib.sol")
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 1, "no diagnostics")
+        self.expect_diagnostic(diagnostics[0], code=2018, lineNo=5, startEndColumns=(4, 62))
+
     def test_didChange_in_A_causing_error_in_B(self, solc: JsonRpcProcess) -> None:
         # Reusing another test but now change some file that generates an error in the other.
         self.test_textDocument_didOpen_with_relative_import(solc)
--- a/libsolidity/lsp/FileRepository.h
+++ b/libsolidity/lsp/FileRepository.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <libsolidity/interface/FileReader.h>
+#include <libsolutil/Result.h>
 
 #include <string>
 #include <map>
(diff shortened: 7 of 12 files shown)


cameel:
> I remember there was some project file format that does include that information, but what was that? @cameel, do you remember?

Do you mean the dappfile ([EIP 82](https://github.com/ethereum/EIPs/issues/82))? Or one of the config files actually used by frameworks?

christianparpart:
> > I remember there was some project file format that does include that information, but what was that? @cameel, do you remember?
> 
> Do you mean the dappfile ([EIP 82](https://github.com/ethereum/EIPs/issues/82))? Or one of the config files actually used by frameworks?

ah right thanks! Dappfile is what I remembered, but this time, I in fact meant a framework that introduced a file that is now also supported by another one. I am talking vaguely because I forgot the names. ;-(

I might look into Dappfile though.

christianparpart:
@cameel if we can't come up with something immediately, what do you think about adding a CLI option, such as `--lsp-search-paths PATH,...`. NB: We once agreed in explicitly not messing with standard-json or related CLI args, as these are in fact two different topics, IIRC.

cameel:
> ah right thanks! Dappfile is what I remembered, but this time, I in fact meant a framework that introduced a file that is now also supported by another one. I am talking vaguely because I forgot the names. ;-(

Oh, you meant `remappings.txt` (https://github.com/ethereum/solidity/issues/12603). But it's not really for include paths. It's for remappings (though in practice they get used to get the same effect).

cameel:
> if we can't come up with something immediately, what do you think about adding a CLI option,

Why not just support `--include-paths` instead of adding a new arg?

christianparpart:
> > if we can't come up with something immediately, what do you think about adding a CLI option,
> 
> Why not just support `--include-paths` instead of adding a new arg?

I can do that. thanks. But I am still hoping for some kind of auto-configuration, where users do not need to do too much.

Suppose you are working in more than one project, and they all do not necessarily share the same include paths.

I think we could auto-detect at least the well known frameworks and adjust the include paths accordingly (on top of manually added ones, if added). What do you think, @cameel ?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,6 +101,21 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
 	m_settingsObject = _settings;
+	if (_settings["include-paths"] && _settings["include-paths"].isArray())
+	{
+		std::vector<boost::filesystem::path> includePaths;
+		Json::Value jsonIncludePaths = _settings["includes"];
+		for (Json::Value jsonPath : jsonIncludePaths)
```
stackenbotten: Coding style error

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
 	m_settingsObject = _settings;
+	Json::Value jsonIncludePaths = _settings["include-paths"];
+	if (jsonIncludePaths && jsonIncludePaths.isArray())
+	{
+		std::vector<boost::filesystem::path> includePaths;
+		for (Json::Value jsonPath : jsonIncludePaths)
```
stackenbotten: Coding style error

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
 	m_settingsObject = _settings;
+	Json::Value jsonIncludePaths = _settings["include-paths"];
+	if (jsonIncludePaths && jsonIncludePaths.isArray())
```
cameel, line 154: Shouldn't we do something if `include-paths` is present but is not an array?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+	{
+		std::vector<boost::filesystem::path> includePaths;
```
```suggestion
		vector<boost::filesystem::path> includePaths;
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+		for (Json::Value jsonPath: jsonIncludePaths)
```
cameel: Shouldn't this be a reference? Or does it actually dispense copies?
```suggestion
		for (Json::Value const& jsonPath: jsonIncludePaths)
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+		{
+			if (!jsonPath.isString())
+			{
+				m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
+				continue;
+			}
+			includePaths.emplace_back(boost::filesystem::path(jsonPath.asString()));
```
cameel: Isn't this equivalent to a simple `if`/`else`?
```suggestion
			if (!jsonPath.isString())
				m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
			else
				includePaths.emplace_back(boost::filesystem::path(jsonPath.asString()));
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -101,14 +101,29 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
 	m_settingsObject = _settings;
+	Json::Value jsonIncludePaths = _settings["include-paths"];
+	if (jsonIncludePaths && jsonIncludePaths.isArray())
+	{
+		std::vector<boost::filesystem::path> includePaths;
+		for (Json::Value jsonPath: jsonIncludePaths)
+		{
+			if (!jsonPath.isString())
+			{
+				m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
```
cameel: If there are 10 non-string values, you'll get 10 identical messages. Wouldn't it be better to either issue just one (about the whole array) or have each one mention the invalid value?

Changelog.md
```diff
@@ -13,6 +13,7 @@ Compiler Features:
  * Commandline Interface: option ``--pretty-json`` works also with the following options: ``--abi``, ``--asm-json``, ``--ast-compact-json``, ``--devdoc``, ``--storage-layout``, ``--userdoc``.
  * SMTChecker: Support ``abi.encodeCall`` taking into account the called selector.
  * Language Server: Allow full filesystem access to language server.
+ * Language Server: Always adds ``{project_root}/node_modules`` to include search paths.
```
Marenz: I suggest to use `Always add` instead of `Always adds` as it is more consistent with the other entries

libsolidity/lsp/LanguageServer.cpp
Marenz: Should this be maybe a proper error with throwing and all?

Marenz (dismissed):
In the commit with the descrp. `[lsp] Fixes sourceUnitNameToUri() path path resolving in relation to ... …include paths being used.`
Maybe one `path` is enough in the commit message? :P

libsolidity/lsp/FileRepository.cpp
```diff
@@ -34,9 +34,25 @@ using namespace solidity::frontend;
...
 {
+	setIncludePaths(move(_includePaths));
+}
+
+void FileRepository::setIncludePaths(std::vector<boost::filesystem::path> _paths)
+{
+	// Prefix relative paths with base-path.
+	for (auto& path: _paths)
+		if (path.is_relative())
+			path = m_basePath / path;
```
cameel: Not sure if this makes sense. Normally the include paths are relative to working dir, not base path.

So this kinda depends on how LSP handles working dir. If it ignores it completely, this might be ok but if it's affected by it in anyway then probably not.

libsolidity/lsp/FileRepository.cpp
```diff
@@ -34,9 +34,25 @@ using namespace solidity::frontend;
...
+using solidity::util::Result;
 
-FileRepository::FileRepository(boost::filesystem::path _basePath): m_basePath(std::move(_basePath))
+FileRepository::FileRepository(boost::filesystem::path _basePath, std::vector<boost::filesystem::path> _includePaths):
+	m_basePath(std::move(_basePath))
 {
+	setIncludePaths(move(_includePaths));
+}
+
+void FileRepository::setIncludePaths(std::vector<boost::filesystem::path> _paths)
```
cameel, line 48: This is not a simple setter. It makes some opinionated decisions. I think we should rename it to make it very clear:.
```suggestion
void FileRepository::setUpIncludePaths(std::vector<boost::filesystem::path> _requestedPaths)
```

libsolidity/lsp/FileRepository.cpp
```diff
@@ -51,6 +67,8 @@ string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
 		return "file:///" + _sourceUnitName;
 	else if (_sourceUnitName.find("/") == 0)
 		return "file://" + _sourceUnitName;
+	else if (auto const resolvedPath = tryResolvePath(_sourceUnitName); resolvedPath.message().empty())
```
cameel: The actual condition in such ifs is really easy to miss. I'd split that in two lines (or just avoid this syntax):
```suggestion
	else if (
		auto const resolvedPath = tryResolvePath(_sourceUnitName);
		resolvedPath.message().empty()
	)
```

cameel: Isn't this supposed to kick in when the resolved path is **not** empty?
```suggestion
	else if (auto const resolvedPath = tryResolvePath(_sourceUnitName); !resolvedPath.message().empty())
```

libsolidity/lsp/FileRepository.cpp
```diff
@@ -51,6 +67,8 @@ string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
...
+		return "file://" + resolvedPath.get().generic_string();
 	else
 		return "file://" + m_basePath.generic_string() + "/" + _sourceUnitName;
```
cameel, line 79: Does `generic_string()` always strip the trailing `/`? What happens when base path is `/`? Won't the resulting path be interpreted as UNC path giving weird results?

libsolidity/lsp/FileRepository.cpp
```diff
@@ -51,6 +67,8 @@ string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
 		return "file:///" + _sourceUnitName;
```
cameel, line 72: Regarding the first part of the `if` above (I can't comment on it directly):
```c++
if (m_sourceUnitNamesToUri.count(_sourceUnitName))
		return m_sourceUnitNamesToUri.at(_sourceUnitName);
```
Is the assumption that this is already an URI, i.e. starts with `protocol://`? If so then maybe we should have an assert here?

libsolidity/lsp/FileRepository.cpp
```diff
@@ -51,6 +67,8 @@ string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
...
 	else if (_sourceUnitName.find("/") == 0)
 		return "file://" + _sourceUnitName;
```
cameel: Doesn't this break resolution for files that are placed directly under one of the include dirs? I.e. if you have LSP set up equivalent to `--base-path /a/ --include-path b/c` then you won't be able to do `import "x.sol"` to import `/a/b/c/x.sol`, it will look for `/a/x.sol` instead.

test/libsolidity/lsp/include-paths/default-include.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+import "my-module/test.sol";
```
cameel: Can you add a test that imports just `test.sol`?

cameel:
> I think we could auto-detect at least the well known frameworks and adjust the include paths accordingly (on top of manually added ones, if added). What do you think, @cameel ?

I'm not really convinced that this should be baked into the compiler. It will lead to weird behavior in corner cases, when you're not on the "happy path" with a well supported framework and old compiler binaries will be less usable with new frameworks. I think we should instead give frameworks a way to provide this info to the compiler. I think that some non-framework-specific config file that the LSP expects to find under the base path might be a better idea.

libsolidity/lsp/LanguageServer.cpp
christianparpart: Not with throwing, at most reporting (trace does report towards the client), as this is an error caused by the client.

libsolidity/lsp/LanguageServer.cpp
christianparpart: At most, report (just like what if not having strings inside). Otherwise, still continue functioning. So: report & ignore (continue).

libsolidity/lsp/LanguageServer.cpp
Marenz: Then maybe reporting it at a higher log level tahn trace?

libsolidity/lsp/LanguageServer.cpp
christianparpart: I wasn't sure about what style to chose. But yes ;)

libsolidity/lsp/FileRepository.cpp
christianparpart: the `m_basePath` does contain the project's root directory, as the information is passed to the server from the client (IDE).

libsolidity/lsp/FileRepository.cpp
```diff
@@ -34,9 +34,25 @@ using namespace solidity::frontend;
...
+
+void FileRepository::setIncludePaths(std::vector<boost::filesystem::path> _paths)
+{
+	// Prefix relative paths with base-path.
+	for (auto& path: _paths)
+		if (path.is_relative())
+			path = m_basePath / path;
+
+	// Always have {projectRoot}/node_modules in the include-paths.
+	_paths.emplace_back(m_basePath / "node_modules");
```
chriseth: I don't think this is a good idea. As discussed in the meeting, I think it would be better to have `{"PROJECT_ROOT/node_modules"}` as the set of default includes paths. If you do it as it is done now there is no way to remove `node_modules` from the include paths.

libsolidity/lsp/FileRepository.cpp
cameel: But does the working dir of the LSP have any influence on its path resolution?

libsolidity/lsp/FileRepository.cpp
christianparpart: done.

libsolidity/lsp/FileRepository.cpp
christianparpart: I just tried that locally in my browser and having too many leading slashes in your file-URI works. FF simply ignores, Chrome does rewrite the URL.

libsolidity/lsp/FileRepository.cpp
christianparpart: Yes, right below, the function named `setSourceByUri`  This is called in a context where the remote end is providing us URIs. I'm not sure an assert() is really necessary (?) as in that case, the client would just receive as URI what it initially has sent as URI (if it could be mapped to a local file). Hmmm.

libsolidity/lsp/FileRepository.cpp
christianparpart: We have much longer lines than that. I don't think that's particularly unreadable. But this is subjective. I'm not sure.

libsolidity/lsp/FileRepository.cpp
christianparpart: @cameel: If at all, I think this if-stmt should be removed all together. Meaning, it should not happen (AFAICT). I might be wrong. But if at all, I'd prefer to remove this in a seperate PR, because these lines are unrelated to the actual PR that should not become too big. What do you think?

test/libsolidity/lsp/include-paths/default-include.sol
christianparpart: We have general import tests already for import path names that are meant to be relatively imported. What do you think should we expect of your import-test.sol-testcase?

libsolidity/lsp/FileRepository.cpp
cameel: Fair enough. If it was there already I think it's fine to address it separately.

test/libsolidity/lsp/include-paths/default-include.sol
cameel: I expect it to trigger the problem I mentioned in https://github.com/ethereum/solidity/pull/12994#discussion_r874118206.

libsolidity/lsp/FileRepository.cpp
cameel: The thing I'm not sure about is: can a non-URI end up in `m_sourceUnitNamesToUri`? My impression from this function is that it's trying hard to always have the result prefixed with `file://` but it wasn't obvious to me if this particular `if` always guarantees that prefix too. My guess is yes and if that's the case, an assert would make it obvious. If not, I think it needs a comment saying how/why a non-URI can end up in `m_sourceUnitNamesToUri`.

libsolidity/lsp/FileRepository.cpp
cameel: For me it's not about the line length but about the condition blending in with the variable declaration. In most cases in our code such `if`s do not have the condition part after `;` and the condition is the variable. If it's not then IMO it's very easy to miss (I caught myself on that many times) and I think the condition should always be on a line of its own to make it more prominent.

libsolidity/lsp/FileRepository.cpp
christianparpart: done

libsolidity/lsp/FileRepository.cpp
christianparpart: @cameel I've added a test (bad-include) and it triggers a diagnostic that the given file "test.sol" could not be found.

cameel:
`t_win_soltest` is failing. And I just noticed that it runs LSP tests, which I didn't expect. I think this was originally a separate step and it made more sense that way. Then you can tell just from the job name what failed - I wouldn't even bother looking inside if I saw it's a failing LSP test in a LSP-related PR. And with this running after soltest, you have to first wait 20+ minutes for the Windows binary, and then 15+ minutes for soltest to finish. You'd get feedback much faster if this ran independently.

Changelog.md
```diff
@@ -8,6 +8,8 @@ Language Features:
 Compiler Features:
  * LSP: Add rudimentary support for semantic highlighting.
  * Yul Optimizer: Improve inlining heuristics for via IR code generation and pure Yul compilation.
+ * Language Server: Always add ``{project_root}/node_modules`` to include search paths.
+ * Language Server: Adds support for configuring extra include paths ``include-paths`` JSON settings object that can be passed during LSP configuration stage.
```
```suggestion
 * Language Server: Adds support for configuring ``include-paths`` JSON settings object that can be passed during LSP configuration stage.
```

libsolidity/lsp/FileRepository.cpp
cameel: What about this one? Looks like it got lost among other comments.

test/lsp.py
```diff
@@ -562,7 +562,11 @@ def test_diagnostics(self):
                 testname_and_subdir = diagnostics["uri"].replace(self.suite.project_root_uri + "/", "")[:-len(".sol")]
                 testname, sub_dir = split_path(testname_and_subdir)
 
-                expected_diagnostics = tests[testname]
+                # Only add expected diagnostics, if available.
+                expected_diagnostics = {}
+                if testname in tests:
+                    expected_diagnostics = tests[testname]
```
```suggestion
                expected_diagnostics = tests.get(testname, {})
```

test/libsolidity/lsp/include-paths/default-include.sol
```diff
@@ -0,0 +1,14 @@
...
+
+contract MyContract
+{
+    function f(uint a, uint b) public pure returns (uint)
+    {
+        return MyModule.add(a, b);
+    }
+}
+// ----
+// file:///home/christianparpart/work/solidity/test/libsolidity/lsp/node_modules/my-module/test:
```
cameel: This path looks like it's specific to your system. It shouldn't be here, should it?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -146,14 +146,30 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+		vector<boost::filesystem::path> includePaths;
+		int typeFailures = 0;
+		for (Json::Value const& jsonPath: jsonIncludePaths)
+		{
+			if (jsonPath.isString())
+				includePaths.emplace_back(boost::filesystem::path(jsonPath.asString()));
+			else
+				typeFailures++;
+		}
+		if (typeFailures)
```
cameel: I think this would be much more readable. From the name it looks like a boolean but it isn't.
```suggestion
		if (typeFailureCount > 0)
```

libsolidity/lsp/FileRepository.cpp
```diff
@@ -69,6 +91,52 @@ void FileRepository::setSourceByUri(string const& _uri, string _source)
...
+	if (
+		boost::filesystem::path(_strippedSourceUnitName).has_root_path() &&
+		boost::filesystem::exists(_strippedSourceUnitName)
+	)
+		return boost::filesystem::path(_strippedSourceUnitName);
+
+	vector<boost::filesystem::path> candidates;
+	vector<reference_wrapper<boost::filesystem::path const>> prefixes = {m_basePath};
+	prefixes += (m_includePaths | ranges::to<vector<reference_wrapper<boost::filesystem::path const>>>);
+	auto const defaultInclude = m_basePath / "node_modules";
```
cameel, line 110: This looks like a better place for adding special paths than `sourceUnitNameToUri()` but this still makes it impossible not to have any include paths at all - and this is a very common case. Maybe a more logical place to add this would be `changeConfiguration()`? That function can discern between "include paths not set" (i.e. `null`) and "include paths set is empty" (i.e. `[]`).

Unless the LSP clients do not care and would just always send `[]`... What do you think?

libsolidity/lsp/FileRepository.cpp
cameel: ok, so you'll be fixing that in a separate PR? I think it would make sense to report it as an issue unless you'll be doing that right away.

test/libsolidity/lsp/include-paths/default-include.sol
christianparpart: I was addressing this to @Marenz already. He can improve `lsp.py` seperately, I'd say. :)

libsolidity/lsp/FileRepository.cpp
christianparpart: I think the current behavior is fine. I've tested it with both, and the default configuration should suffice (works like solidity-vscode extension by default works, too), but gives the ability to customize and also _not_ to include the above mentioned default include of "${projectRoot}/node_modules".

test/libsolidity/lsp/include-paths/default-include.sol
cameel: But is this something enforced by the test suite? If so, won't this fail for anyone but you? If not, why not remove it?

libsolidity/lsp/FileRepository.cpp
cameel: But that's only if you include some other path, right? Because the condition below is `m_includePaths.empty()` and that's going to always add it if you have no include paths.

christianparpart: Yes, it's going to add that default include only if the user hasn't supplied any. Do you explicitly want to support dropping the default include and have zero other include paths set? I think this case is rather artificial. The default include path is for ease-of-use, as that's most likely the most often used include path being used in the wild (AFAICT). If people chose to roll their own, they can do, which would not auto-add the default include path. Hmmm....

test/libsolidity/lsp/include-paths/default-include.sol
christianparpart: @Marenz what's that line exactly there for? Maybe you can jump in?

Marenz: I am not sure how or about what you two are misunderstanding each other, but yes, this path should be relative to the root dir or rather the test dir. I believe chris pushed this only so I can investigate.

libsolidity/lsp/FileRepository.cpp
cameel: Yeah, I think it would be best if one could drop the default path from the set. I mean, it probably won't hurt anyone in a vast majority of cases but it just does not seem quite right. If the client has a way to say explicitly "no extra paths" (and this depends on client's UI), the server should have a way to respect that. In some corner cases you could have a project that does not use `node_modules` for contracts but still have a stray dir there because you were installing some extra tools via npm.

Now that I think of it - is there a mechanism for the server to provide the "default" settings for the client to use as a base value? Because the "right" way to solve this IMO would be to just put `node_modules` path in these defaults so that user does not have to do anything but is also able to modify them.

Overall this is not a deal-breaker and I'm going to accept the PR regardless of this but I'm just thinking that maybe we could solve this better.

libsolidity/lsp/LanguageServer.cpp
cameel: Not sure if I missed it earlier, or if it changed, but in the current code `jsonIncludePaths.isArray() == false` does not result in an `Invalid JSON configuration passed.` report.

test/libsolidity/lsp/include-paths/bad_include.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+   import "test.sol";
+// ^^^^^^^^^^^^^^^^^^ @BadInclude
```
cameel: As discussed on the chat, the include is valid, it's just that the file is missing - so the error (and test name) should say just that.

cameel: Also, we should have a positive test for this case too. I.e. a file placed directly under base path or one of the include paths should be successfully imported.

libsolidity/lsp/FileRepository.cpp
cameel: I see this has been removed so marking as resolved.

libsolidity/lsp/LanguageServer.cpp
cameel: Looking at it now, I'm still not sure if the change we did on the call is 100% correct. I.e. now it will add a trace when settings contains `"include-paths"` key but it's `null`. Wouldn't it be better to treat it the same way as the key not being there?

test/libsolidity/lsp/other-include-dir/otherlib/otherlib.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+library MyRootLib
```
cameel:
```suggestion
library OtherLib
```

libsolidity/lsp/FileRepository.cpp
cameel: Marking as resolved.

We discussed it on the call and looks like there is not way for the client to query the configuration. So my idea is not viable after all.

cameel (commented):
State of the review after our call and subsequent changes:
1. There are two minor comments to resolve that I added now (lib name and trace for `null` settings).
2. Some tests are failing.
3. We agreed that there should be some tests covering the function for setting configuration and non-default include paths.

libsolidity/lsp/LanguageServer.cpp
christianparpart: @cameel isn't that subjective? I don't mind, and I have no too strong opinion about it. Feels minor to me, I prefer to be strict.  And if someone explicitly sets it to null, to be honest, to me that reads like he doesn't want any include paths. So the code IMHO(!) semantically does make sense, doesn't it?

cameel: I'm fine with being strict, just not sure if the LSP clients are actually that strict themselves. And if they aren't, we'll have to adapt to them not they to us. I can see them just not being consistent and deciding to set a field to `null` instead of removing it in some cases. We had been doing that ourselves in JSON exports for a long time.

Anyway, I also think this is minor. Just wanted to let you know that this could potentially lead to a false-positive warning, depending on what clients do.

test/libsolidity/lsp/include-paths/using-custom-includes.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.0;
+
+   import "otherlib/otherlib.sol";
+// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ @SomeTag
```
cameel: Shouldn't this succeed? And what kind of error is `SomeTag`?

test/lsp.py
```diff
@@ -1381,6 +1381,37 @@ def get_test_tags(self, test_name: TestName, sub_dir=None, verbose=False):
...
 
+    def test_custom_includes(self, solc: JsonRpcProcess) -> None:
+        self.setup_lsp(solc, expose_project_root=False)
+        solc.send_notification(
+            'workspace/didChangeConfiguration',
+            {
+                'settings':
+                {
+                    'include-paths':
+                    [
```
```suggestion
                'settings': {
                    'include-paths': [
```

test/libsolidity/lsp/include-paths/using-custom-includes.sol
christianparpart: Not in this test. But in the lsp.py there is an explicit test with the custom include path set, THERE it should and will pass. :)

cameel (dismissed):

Marenz (approved):

cameel (approved):

narghev:
Hi, just had this problem on nvim with newly installed version of `solc` using `brew`. 
```Version: 0.8.15+commit.e14f2714.Darwin.appleclang```

Wanted to check if this is released already or not.

Edit: Checked and the commit with hash `e14f2714` thus the `v0.8.15` was released on Jun 15th which is before this merge. Is this going to be included in `0.8.16`?

cameel:
Yes, this will go into 0.8.16, which we are releasing tomorrow.

PR #12997, opened 2022-05-10
Support new z3 AST node

leonardoalt:
We need to support this for z3 4.8.16.

### DIFF
--- a/libsmtutil/Z3Interface.cpp
+++ b/libsmtutil/Z3Interface.cpp
@@ -298,6 +298,7 @@ Expression Z3Interface::fromZ3Expr(z3::expr const& _expr)
 		arguments.push_back(fromZ3Expr(_expr.arg(i)));
 
 	auto kind = _expr.decl().decl_kind();
+
 	if (_expr.is_ite())
 		return Expression::ite(arguments[0], arguments[1], arguments[2]);
 	else if (_expr.is_not())
@@ -371,7 +372,10 @@ Expression Z3Interface::fromZ3Expr(z3::expr const& _expr)
 		return Expression("dt_accessor_" + _expr.decl().name().str(), arguments, sort);
 	else if (kind == Z3_OP_DT_IS)
 		return Expression("dt_is", {arguments.at(0)}, sort);
-	else if (kind == Z3_OP_UNINTERPRETED)
+	else if (
+		kind == Z3_OP_UNINTERPRETED ||
+		kind == Z3_OP_RECURSIVE
+	)
 		return Expression(_expr.decl().name().str(), arguments, fromZ3Sort(_expr.get_sort()));
 
 	smtAssert(false, "");
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@ Bugfixes:
  * Type Checker: Properly check restrictions of ``using ... global`` in conjunction with libraries.
  * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
+ * SMTChecker: Fix ABI compatibility with z3 >=4.8.15.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
  * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
 
--- a/docs/smtchecker.rst
+++ b/docs/smtchecker.rst
@@ -629,6 +629,11 @@ option ``--model-checker-solvers {all,cvc4,smtlib2,z3}`` or the JSON option
   - if a dynamic ``z3`` library of version 4.8.x is installed in a Linux system (from Solidity 0.7.6);
   - statically in ``soljson.js`` (from Solidity 0.6.9), that is, the Javascript binary of the compiler.
 
+.. note::
+  z3 version 4.8.15 broke ABI compatibility with previous versions and can only
+  be used with solc <=0.8.13. If you are using z3 >=4.8.15 please use solc
+  >=0.8.14.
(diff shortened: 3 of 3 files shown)


leonardoalt:
We should have this before the release.

libsmtutil/Z3Interface.cpp
```diff
@@ -371,7 +372,10 @@ Expression Z3Interface::fromZ3Expr(z3::expr const& _expr)
 		return Expression("dt_accessor_" + _expr.decl().name().str(), arguments, sort);
 	else if (kind == Z3_OP_DT_IS)
 		return Expression("dt_is", {arguments.at(0)}, sort);
-	else if (kind == Z3_OP_UNINTERPRETED)
+	else if (
+		kind == Z3_OP_UNINTERPRETED ||
+		kind == Z3_OP_RECURSIVE
+	)
```
cameel, line 378: Is it possible to add some kind of test to verify that it's handled correctly?

What is the result of just feeding the expression string directly into `Expression`? Will it just somehow go through without issues or can something in our code fail when it's encountered?

leonardoalt: 1) Not really. We don't really control what the solver's proof has, and it doesn't seem to consistently use this AST node.

leonardoalt: 2. We could also do that, worst case it would make the counterexamples/invariants look weird or break it in which case we just don't show counterexamples/invariants. But the worst part is not noticing that we potentially should be handling a certain AST node properly which we aren't.

I have a tendency to keep it as is because this is the approach we used early on to detect what kind of AST nodes we should handle, but fine with either if you think the second option is safer.

ekpyron:
Should this get a changelog entry?

leonardoalt:
> Should this get a changelog entry?

Not sure this one needs a changelog entry though. It's not user facing.

ekpyron:
> > Should this get a changelog entry?
> 
> Not sure this one needs a changelog entry though. It's not user facing.

Isn't it? I haven't read through the trigger for this yet, but won't this break the counterexamples if you use e.g. solc 0.8.13 with dynamically loaded recent z3?

cameel:
solc 0.8.13 actually crashes if you dynamically load Z3 4.8.16. We discussed this earlier on the chat and I think we should treat it as ABI incompatibility. Basically, older solc versions support only Z3 <= 4.8.14. I think Z3 should have released this as 4.9.

leonardoalt:
Yea ok. Will add changelog entry and write something in the docs.

libsmtutil/Z3Interface.cpp
cameel: With (2) I was actually asking what happens if we do `Expression(_expr.decl().name().str(), ...)` and it turns out the string was something invalid. Will it crash at some point or just go unnoticed. I'm asking because I have no idea if this is the right way to handle this and if it will fail with something invalid, it would at least give me more confidence that the fix is not wrong :)

leonardoalt: Ah no, the string should be fine. From what I read in their docs/code both are functions, but the RECURSIVE one has a `recursive` flag.

cameel: > I have a tendency to keep it as is because this is the approach we used early on to detect what kind of AST nodes we should handle, but fine with either if you think the second option is safer.

We could actually have the compiler detect any unhandled values at compile time if you replaced the `if` with a `switch` (could be a separate PR though). It would not catch the issue at runtime though so I guess we still need the assert even then.

leonardoalt: there are way too many of these :grimacing: 
don't wanna add all of them here

cameel: Ah, ok. We have so many listed that I thought it's pretty much all of them. So those we don't have actually never happen with our queries?

cameel: > Ah no, the string should be fine. From what I read in their docs/code both are functions, but the RECURSIVE one has a `recursive` flag.

OK then. Would be best to have it tested but if we can't really trigger it easily then I guess there's not much we can do.

cameel:
So I think it's ok to merge when you add the changelog entry/docs.

ekpyron:
Since newer solc versions will (if I understood things correctly) still keep working with Z3 < 4.8.16, a changelog entry alone should be fine... not much we can do about older solc versions breaking with newer Z3, but yeah, that's z3's fault for doing something like this without bumping the so version...

libsmtutil/Z3Interface.cpp
leonardoalt: > Ah, ok. We have so many listed that I thought it's pretty much all of them. So those we don't have actually never happen with our queries?

Exactly.

leonardoalt: > OK then. Would be best to have it tested but if we can't really trigger it easily then I guess there's not much we can do.

Yea I wouldn't really know how to trigger it on purpose. I can try to do that with more time next week.

leonardoalt:
> Since newer solc versions will (if I understood things correctly) still keep working with Z3 < 4.8.16, a changelog entry alone should be fine... not much we can do about older solc versions breaking with newer Z3, but yeah, that's z3's fault for doing something like this without bumping the so version...

yea but doesn't hurt adding it to the docs, people might be confused/annoyed if it happens

stackenbotten:
There was an error when running `chk_coding_style` for commit `931f489947d82f6a36c5eca5671ce08c04d6396f`:
```
docs/smtchecker.rst:636: 

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `931f489947d82f6a36c5eca5671ce08c04d6396f`:
```
docs/smtchecker.rst:636: 

```

PR #13001, opened 2022-05-10
Don't add an empty data slot to ABI-encoded empty strings from storage

### LINKED ISSUE #12880: Bug in ABI encoding empty storage array
The following function does ABI encoding of strings in storage. However, this has a bug when the string in storage is empty. If the string is short, it will still have an empty data that's part of the ABI. But that is not how empty strings are encoded in ABI.

https://github.com/ethereum/solidity/blob/c4909e99c1015233d6937d229058cde52615f246/libsolidity/codegen/ABIFunctions.cpp#L699-L703

For reference, the ABI encoder v1 version correctly encodes the data.
```solidity
// pragma abicoder v1;
contract C {
    string message;
    event Message(string);
    function emitMessage() external {
        emit Message(message);
    }
}
```
---
Found in https://github.com/gakonst/foundry/issues/1027#issuecomment-1079880060

The ABI encoding for the empty string should be  (48-bytes)
```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000000
```

but ends up being

```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000000
```

matheusaaguiar:
Fixes https://github.com/ethereum/solidity/issues/12880


Changelog.md
```diff
@@ -13,9 +13,10 @@ Compiler Features:
 
 Bugfixes:
  * Type Checker: Properly check restrictions of ``using ... global`` in conjunction with libraries.
-* Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
-* TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
-* View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
+ * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
```
chriseth: Thanks for fixing this, but I think we should do this in its own pull request.

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -695,11 +695,14 @@ string ABIFunctions::abiEncodingFunctionCompactStorageArray(
 					let slotValue := sload(value)
 					let length := <byteArrayLengthFunction>(slotValue)
 					pos := <storeLength>(pos, length)
+					ret := pos      //assume array is empty
```
chriseth: Maybe it's better to move this inside the `case 0`

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -695,11 +695,14 @@ string ABIFunctions::abiEncodingFunctionCompactStorageArray(
...
 					switch and(slotValue, 1)
 					case 0 {
 						// short byte array
 						mstore(pos, and(slotValue, not(0xff)))
```
chriseth, line 701: I think this should also go inside the if

Changelog.md
```diff
@@ -13,9 +13,10 @@ Compiler Features:
...
+ * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
+ * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
+ * ABI Encoding: Fix bug when encoding string in storage that is empty.
```
chriseth: Maybe `remove padding for empty storage strings`?

chriseth:
This looks very good already!

About the tests: I think you fixed it so that both abi encoders (v1 and v2) behave in the same way. In this case, you can move the test file somewhere else and remove the "pragma abicoder", because then the test framework will automatically run both encoders on it.

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -698,8 +698,11 @@ string ABIFunctions::abiEncodingFunctionCompactStorageArray(
 					switch and(slotValue, 1)
 					case 0 {
 						// short byte array
-						mstore(pos, and(slotValue, not(0xff)))
-						ret := add(pos, <lengthPaddedShort>)
+						ret := pos		//assume empty array
```
chriseth: We usually don't use tabs / spaces to align things after the initial tabs - this will always turn out messy.

```suggestion
						ret := pos // assume empty array
```

Changelog.md
chriseth: Also maybe `ABI Encoder` - we usually list a component of the compiler.

Changelog.md
```diff
@@ -22,6 +22,7 @@ Bugfixes:
  * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
  * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
+ * ABI Encoder: Fix bug when encoding string in storage that is empty.
```
cameel: We should be more specific about what the bug was:
```suggestion
 * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
```

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -698,8 +698,11 @@ string ABIFunctions::abiEncodingFunctionCompactStorageArray(
 					switch and(slotValue, 1)
 					case 0 {
 						// short byte array
-						mstore(pos, and(slotValue, not(0xff)))
-						ret := add(pos, <lengthPaddedShort>)
+						ret := pos //assume empty array
+						if gt(length, 0) {
+							mstore(pos, and(slotValue, not(0xff)))
+							ret := add(pos, <lengthPaddedShort>)
+						}
```
cameel: Following @chriseth's suggestion in https://github.com/ethereum/solidity/issues/12880#issuecomment-1122292315, this should look like this:
```suggestion
						mstore(pos, and(slotValue, not(0xff)))
						ret := add(pos, mul(<lengthPaddedShort>, iszero(length)))
```

Your version requires a jump. `JUMPI` opcode is 10 gas while `MUL` is just 5.

But it would be best to measure it to be sure (the final cost difference might be different due to the optimizer). You might also be able to come up with something even cheaper.

cameel: You can compare costs of different versions by using `isoltest` to update gas expectations in sematic tests:
```bash
isoltest --accept-updates --enforce-gas-cost --optimize
```
When you run this, isoltest will update these tests in your working copy and you'll be able to see what the gas difference is. The gas update for all tests takes some time so for quick testing you might want to select just one specific test. You can do that via the `--test` option. E.g. `--test=semanticTests/various/skip_dynamic_types_for_structs`.

Also, when you're done, you'll have to update these expectations and commit them for all soltest runs in CI to pass. To update them all, you need to run the above `isoltest` command twice - once with and once without `--optimize`. For example right now the `t_ubu_soltest_all` job is failing in your PR exactly for this reason:

```xml
<testcase assertions="4" classname="semanticTests.various" name="skip_dynamic_types_for_structs" time="0.358608">
<failure message="failure" type="assertion error"><![CDATA[
ASSERTION FAILURE:
- file   : boostTest.cpp
- line   : 111
- message: Test expectation mismatch.
Running via Yul:
Expected result:
// g() -> 2, 6
// gas irOptimized: 178805
// gas legacy: 180753
// gas legacyOptimized: 179472

Obtained result:
// g() -> 2, 6
// gas irOptimized: 178868 [+63 (+0%)]
// gas legacy: 180753
// gas legacyOptimized: 179472
```

By the way, something seems broken since [CircleCI does not show the failing test output](https://app.circleci.com/pipelines/github/ethereum/solidity/24096/workflows/6e82f6f8-585f-410c-b863-7e5147a86cf2/jobs/1058493/tests). We'll need to fix that but for now I was able to find the failing test by looking at the artifacts directly:

test/libsolidity/semanticTests/strings/empty_storage_string_3.sol
```diff
@@ -0,0 +1,14 @@
+contract C {
+    string message;
+    string message2;
+    event Message(string, string);
+    function f() external {
+        message = "msg";
```
cameel: You can initialize these strings inline
```suggestion
    string message = "msg";
    string message2 = "";
    event Message(string, string);
    function f() external {
```

test/libsolidity/semanticTests/strings/empty_storage_string_3.sol
```diff
@@ -0,0 +1,14 @@
+contract C {
+    string message;
+    string message2;
```
cameel:
```suggestion
    string emptyMessage;
    string nonEmptyMessage;
```

test/libsolidity/semanticTests/strings/empty_storage_string_1.sol
```diff
@@ -0,0 +1,12 @@
+contract C {
+    string message;
+    event Message(string);
+    function f() external {
+        emit Message(message);
+    }
+}
```
cameel: I'd put cases 1-5 all in a single test. Originally I wanted to suggest to give them better, more specific names but they're such minor variations that I think that just having one function going through these cases makes more sense and would be much more compact. Especially given that you're testing what happens when these values are reassigned.

test/libsolidity/semanticTests/strings/empty_storage_string_1.sol
```diff
@@ -0,0 +1,12 @@
+contract C {
```
cameel: Some more test ideas to make sure we're not missing any other cases:
- Encoding an empty string from calldata and from memory.
- Encoding an empty string by returning it from external function (rather than emitting it in an event). Or including it in an error. Or using `abi.encode()` explicitly.
- Encoding empty `bytes`. Or an empty string/bytes literal.

cameel (changes requested):

Changelog.md
chriseth: At least for me, "slot" is a term that only applies to storage.

cameel: Technically yes, but I don't think we have a word for "a slot-sized amount of memory" and I think that "slot" is still understandable in this context.

How would you word it without referring to it as a slot?

test/libsolidity/smtCheckerTests/abi/abi_decode_simple.sol
```diff
@@ -21,14 +21,23 @@ contract C {
...
+// Warning 7812: (191-205): BMC: Assertion violation might happen here.
+// Warning 7812: (209-223): BMC: Assertion violation might happen here.
+// Warning 7812: (242-256): BMC: Assertion violation might happen here.
+// Warning 7812: (260-274): BMC: Assertion violation might happen here.
+// Warning 7812: (359-373): BMC: Assertion violation might happen here.
+// Warning 7812: (392-406): BMC: Assertion violation might happen here.
+// Warning 7812: (425-434): BMC: Assertion violation might happen here.
+// Warning 7812: (505-519): BMC: Assertion violation might happen here.
+// Warning 7812: (538-552): BMC: Assertion violation might happen here.
+// Warning 8084: BMC analysis was not possible. No SMT solver (Z3 or CVC4) was available. None of the installed solvers was enabled.
```
cameel: Where did these changes some from? These look like they're related to SMTChecker but this PR does not touch it at all.

@leonardoalt These aren't legit expectation updates, are they?

leonardoalt:
```
 // Warning 3996: CHC analysis was not possible. No Horn solver was available. None of the installed solvers was enabled. 
  // Warning 8084: BMC analysis was not possible. No SMT solver (Z3 or CVC4) was available. None of the installed solvers was enabled.
```

these are there so looks like @matheusaaguiar just forgot to use `--no-smt` when running the tests

matheusaaguiar: Should I run with --no-smt ?

cameel: Ah, right.

@matheusaaguiar Only if you don't have Z3 and CVC4 installed. You need them to recreate the same results in tests.

leonardoalt: If you don't have latest z3 installed yes

leonardoalt: But also if your change doesn't affect SMT tests it also makes your tests much faster

matheusaaguiar: Ok, running with --no-smt now.

leonardoalt:
You need a rebase on develop also, cus we updated the z3 version and CI build is failing

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -699,7 +699,7 @@ string ABIFunctions::abiEncodingFunctionCompactStorageArray(
 					case 0 {
 						// short byte array
 						mstore(pos, and(slotValue, not(0xff)))
-						ret := add(pos, <lengthPaddedShort>)
+						ret := add(pos, mul( <lengthPaddedShort>, iszero( iszero(length) ) ) )
```
cameel:
```suggestion
						ret := add(pos, mul(<lengthPaddedShort>, iszero(iszero(length))))
```

Changelog.md
```diff
@@ -21,6 +21,8 @@ Bugfixes:
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
  * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
+ * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
```
cameel: This should not be here. It's a part of a reverted PR.

test/libsolidity/semanticTests/strings/empty_storage_string.sol
```diff
@@ -0,0 +1,69 @@
+contract C {
+
+    string emptyString;
+    string emptyString2 = "";
+    string nonEmptyString = "This is a non empty string";
+    string str = "Another string";
```
cameel:
```suggestion
    string uninitializedString;
    string emptyString = "";
    string nonEmptyString1 = "This is a non empty string";
    string nonEmptyString2 = "Another string";
```

test/libsolidity/semanticTests/strings/empty_storage_string.sol
```diff
@@ -0,0 +1,69 @@
...
+    error EmptyError(string);
```
cameel, line 9: The previous version also was emitting an event, wasn't it? I'd add that too for better coverage.

test/libsolidity/semanticTests/strings/empty_storage_string.sol
```diff
@@ -0,0 +1,69 @@
...
+
+    function f() public returns (string memory) {
+      return emptyString;
+    }
+
+    function g() public returns (string memory, string memory) {
+      return (emptyString, emptyString2);
+    }
```
cameel:
```suggestion
    function f() public returns (string memory) {
        return emptyString;
    }

    function g() public returns (string memory, string memory) {
        return (emptyString, emptyString2);
    }
```

test/libsolidity/semanticTests/strings/empty_storage_string_1.sol
cameel: I see you added most of these but there could you add these as well?
- Emitting `""` literal directly (without assigning it anywhere).
- Storage variable of type `bytes`.
- The test that uses abicoder v1 from the original issue that returns correct results (unless we have that covered somewhere already).

matheusaaguiar: Sure! Working on it now.

matheusaaguiar: > I see you added most of these but there could you add these as well?
> 
>     * Emitting `""` literal directly (without assigning it anywhere).
Created a test function which emits an event object with the "" literal.
> 
>     * Storage variable of type `bytes`.
Created a test function which returns an uninitialized bytes variable from storage.
> 
>     * The test that uses abicoder v1 from the original issue that returns correct results (unless we have that covered somewhere already).
I guess this test is already present, it would be the test that emits an event with an empty string. Should I add a test with a line with "pragma abicoder v1;" ?

Changelog.md
```diff
@@ -21,6 +21,7 @@ Bugfixes:
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
  * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
+ * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
```
chriseth: This needs to be moved now.

chriseth:
The previous version also avoided the mstore, but it's probably better to optimize for the case where the length is not zero. The current version is cheaper in that case, is it?

matheusaaguiar:
The previous version used `if( gt( length, 0) )` which would be 10 (`if`) + 3 (`gt`) = 13 gas addition to avoid the mstore for length zero. It would also add an assignment (`ret := pos`) before the `if`.
Current version adds 5 (`mul`) + 6(`iszero` x 2) = 11 gas.
So I think the current version is cheaper for the case when length is **not**  zero. When the length is indeed zero, it will do the mstore anyway, but It should cost only 1 gas more than doing the `if(gt(...)` operation: 3 (`mstore`) + 5 (`mul`) + 6(`iszero` x 2) = 14. Right?

matheusaaguiar:
Also, after reading https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a0-1-memory-expansion I guess that the mstore operation for length zero could cost 0 gas, right?
Edit: Actually, after reading it again, I see that it does not mean that...

ekpyron:
> The previous version used `if( gt( length, 0) )` which would be 10 (`if`) + 3 (`gt`) = 13 gas addition to avoid the mstore for length zero. It would also add an assignment (`ret := pos`) before the `if`. Current version adds 5 (`mul`) + 6(`iszero` x 2) = 11 gas. So I think the current version is cheaper for the case when length is **not** zero. When the length is indeed zero, it will do the mstore anyway, but It should cost only 1 gas more than doing the `if(gt(...)` operation: 3 (`mstore`) + 5 (`mul`) + 6(`iszero` x 2) = 14. Right?

The constants also cost and the assignment is probably also not free - without thinking through it completely, I'm pretty sure the non-branching version is better :-).

ekpyron (approved):
Looks good to me!

chriseth (approved):

PR #12921, opened 2022-04-11
Adding event and error selector fields

nishant-sachdeva:
closes #12581

PR #13005, opened 2022-05-11
Display human readable type name in conversion error message

### LINKED ISSUE #12896: Custom error is described as "function() pure" instead of "error" in error messages in case of explicit conversions
## Description

Say we have a custom error defined like this in an interface:

```solidity
interface MyInterface {
    error MyCustomError();
}
```

The compiler mistakenly considers `MyInterface.MyCustomError` to be of type `function() pure` instead of type `error`. The latter is what should be used because that's what ends up being set in the ABI.

Casting `MyInterface.MyCustomError` to `bytes4` should work, but it doesn't:

> Explicit type conversion not allowed from `function() pure` to `bytes4`.

```text
Error:
   0: Compiler run failed
      TypeError: Explicit type conversion not allowed from "function () pure" to "bytes4".
        --> /path/to/my/test/file.t.sol:33:25:
         |
      33 |         vm.expectRevert(bytes4(MyInterface.MyCustomError));
         |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Environment

- Compiler version: v0.8.13+commit.abaa5c0e
- Target EVM version (as per compiler settings): default for v0.8.13
- Framework/IDE (e.g. Truffle or Remix): forge
- EVM execution environment / backend / blockchain client: HEVM (forge v0.2.0 (a0db055))

- Operating system: macOS v12.3.1

## Steps to Reproduce

```solidity
interface MyInterface {
    error MyCustomError();
}

contract MyContract {
    function myFunction() external {
        bytes4(MyInterface.MyCustomError);
    }
}
```

StrongerXi:
Closes #12896.

As discussed, I'm adding in `string Type::humanReadableName()`, which defaults to `toString()` and is overridden for error type.

### DIFF
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -110,6 +110,14 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string toStringInParentheses(TypePointers const& _types, bool _short)
+{
+	return '(' + util::joinHumanReadable(
+		_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
+		","
+	) + ')';
+}
+
 }
 
 MemberList::Member::Member(Declaration const* _declaration, Type const* _type):
@@ -3090,6 +3098,19 @@ string FunctionType::canonicalName() const
 	return "function";
 }
 
+string FunctionType::humanReadableName() const
+{
+	switch (m_kind)
+	{
+	case Kind::Error:
+		return "error " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _short */ true);
+	case Kind::Event:
+		return "event " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _short */ true);
+	default:
+		return toString(/* _short */ false);
+	}
+}
+
 string FunctionType::toString(bool _short) const
 {
 	string name = "function ";
@@ -3101,20 +3122,15 @@ string FunctionType::toString(bool _short) const
 			name += *contract->annotation().canonicalName + ".";
 		name += functionDefinition->name();
 	}
-	name += '(';
-	for (auto it = m_parameterTypes.begin(); it != m_parameterTypes.end(); ++it)
-		name += (*it)->toString(_short) + (it + 1 == m_parameterTypes.end() ? "" : ",");
-	name += ")";
+	name += toStringInParentheses(m_parameterTypes, _short);
 	if (m_stateMutability != StateMutability::NonPayable)
 		name += " " + stateMutabilityToString(m_stateMutability);
 	if (m_kind == Kind::External)
 		name += " external";
 	if (!m_returnParameterTypes.empty())
 	{
-		name += " returns (";
-		for (auto it = m_returnParameterTypes.begin(); it != m_returnParameterTypes.end(); ++it)
-			name += (*it)->toString(_short) + (it + 1 == m_returnParameterTypes.end() ? "" : ",");
-		name += ")";
+		name += " returns ";
+		name += toStringInParentheses(m_returnParameterTypes, _short);
 	}
 	return name;
 }
--- a/test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
+++ b/test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
@@ -1,0 +1,11 @@
+interface MyInterface {
+    error MyCustomError(uint256, bool);
+}
+
+contract Test {
+    function test() public returns(bytes4) {
+        return bytes4(MyInterface.MyCustomError);
+    }
+}
+// ----
+// TypeError 9640: (143-176): Explicit type conversion not allowed from "error MyCustomError(uint256,bool)" to "bytes4".
--- a/test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
+++ b/test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
@@ -1,0 +1,9 @@
+contract Test {
+    event MyCustomEvent(uint256);
+
+    function test() public returns(bytes4) {
+        return bytes4(MyCustomEvent);
+    }
+}
+// ----
+// TypeError 9640: (111-132): Explicit type conversion not allowed from "event MyCustomEvent(uint256)" to "bytes4".
--- a/test/libsolidity/syntaxTests/conversion/explicit_conversion_from_error.sol
+++ b/test/libsolidity/syntaxTests/conversion/explicit_conversion_from_error.sol
@@ -1,0 +1,28 @@
+interface MyInterface {
+    enum MyEnum { E1, E2 }
+    error CustomError1(
+        uint256,
+        bool,
+        bool[],
+        address payable,
+        MyInterface,
+        MyEnum,
+        function (string memory) external returns (uint)
+    );
+}
+
+contract Test {
+    function testFunction(string memory) external returns (uint) {}
+
+    function test() public {
+        MyInterface instance = MyInterface(msg.sender);
+        bool[] calldata arr;
+        address payable addr;
+        bytes4(MyInterface.CustomEror1);
+        bytes4(MyInterface.CustomError1());
+        bytes4(MyInterface.CustomError1(1, true, arr, addr, instance, MyInterface.MyEnum.E1, this.testFunction));
+        address(MyInterface.CustomError1);
+    }
+}
+// ----
+// TypeError 9582: (495-518): Member "CustomEror1" not found or not visible after argument-dependent lookup in type(contract MyInterface).
--- a/test/libsolidity/syntaxTests/conversion/explicit_conversion_from_event.sol
+++ b/test/libsolidity/syntaxTests/conversion/explicit_conversion_from_event.sol
@@ -1,0 +1,34 @@
+interface MyInterface {
+    enum MyEnum { E1, E2 }
+}
+
+contract Test {
+    function testFunction(string memory) external returns (uint) {}
+
+    event CustomEvent1(
+        uint256,
+        bool,
+        bool[],
+        address payable,
+        MyInterface,
+        MyInterface.MyEnum,
+        function (string memory) external returns (uint)
+    );
+
+
+    function test() public {
+        MyInterface instance = MyInterface(msg.sender);
+        bool[] calldata arr;
+        address payable addr;
+        bytes4(CustomEvent1);
+        bytes4(CustomEvent1());
+        bytes4(CustomEvent1(1, true, arr, addr, instance, MyInterface.MyEnum.E1, this.testFunction));
+        address(CustomEvent1);
+    }
+}
+// ----
+// TypeError 9640: (502-522): Explicit type conversion not allowed from "event CustomEvent1(uint256,bool,bool[],address payable,contract MyInterface,enum MyInterface.MyEnum,function (string) external returns (uint256))" to "bytes4".
+// TypeError 6160: (539-553): Wrong argument count for function call: 0 arguments given but expected 7.
+// TypeError 9640: (532-554): Explicit type conversion not allowed from "tuple()" to "bytes4".
+// TypeError 9640: (564-656): Explicit type conversion not allowed from "tuple()" to "bytes4".
+// TypeError 9640: (666-687): Explicit type conversion not allowed from "event CustomEvent1(uint256,bool,bool[],address payable,contract MyInterface,enum MyInterface.MyEnum,function (string) external returns (uint256))" to "address".
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -1842,7 +1842,7 @@ Type const* TypeChecker::typeCheckTypeConversionAndRetrieveReturnType(
 					_functionCall.location(),
 					ssl,
 					"Explicit type conversion not allowed from non-payable \"address\" to \"" +
-					resultType->toString() +
+					resultType->humanReadableName() +
 					"\", which has a payable fallback function."
 				);
 			}
@@ -1856,9 +1856,9 @@ Type const* TypeChecker::typeCheckTypeConversionAndRetrieveReturnType(
 					5030_error,
 					_functionCall.location(),
 					"Explicit type conversion not allowed from \"" +
-					argType->toString() +
+					argType->humanReadableName() +
 					"\" to \"" +
-					resultType->toString() +
+					resultType->humanReadableName() +
 					"\". To obtain the address of the contract of the function, " +
 					"you can use the .address member of the function."
 				);
@@ -1867,9 +1867,9 @@ Type const* TypeChecker::typeCheckTypeConversionAndRetrieveReturnType(
 					9640_error,
 					_functionCall.location(),
 					"Explicit type conversion not allowed from \"" +
-					argType->toString() +
+					argType->humanReadableName() +
 					"\" to \"" +
-					resultType->toString() +
+					resultType->humanReadableName() +
 					"\".",
 					result.message()
 				);
--- a/libsolidity/ast/Types.h
+++ b/libsolidity/ast/Types.h
@@ -339,6 +339,7 @@ public:
 	std::string toString() const { return toString(false); }
 	/// @returns the canonical name of this type for use in library function signatures.
 	virtual std::string canonicalName() const { return toString(true); }
+	virtual std::string humanReadableName() const { return toString(); }
 	/// @returns the signature of this type in external functions, i.e. `uint256` for integers
 	/// or `(uint256,bytes8)[2]` for an array of structs. If @a _structsByName,
 	/// structs are given by canonical name like `ContractName.StructName[2]`.
@@ -1378,6 +1379,7 @@ public:
 	TypeResult unaryOperatorResult(Token _operator) const override;
 	TypeResult binaryOperatorResult(Token, Type const*) const override;
 	std::string canonicalName() const override;
+	std::string humanReadableName() const override;
 	std::string toString(bool _short) const override;
 	unsigned calldataEncodedSize(bool _padded) const override;
 	bool canBeStored() const override { return m_kind == Kind::Internal || m_kind == Kind::External; }


StrongerXi:
I intentionally limited the usage of `humanReadableName` to the relevant conversion, because using it for all applicable type errors probably deserves another PR...?

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3090,14 @@ string FunctionType::canonicalName() const
 	return "function";
 }
 
+string FunctionType::humanReadableName() const
+{
+	switch (m_kind) {
```
nishant-sachdeva:
```suggestion
	switch (m_kind) 
	{
```
Changing to adhere to the style guide.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3090,14 @@ string FunctionType::canonicalName() const
...
+		case Kind::Error: return "error";
```
nishant-sachdeva:
```suggestion
		case Kind::Error: return "error " + m_declaration->name();
```
We could add the name of the error as requested by @cameel  in https://github.com/ethereum/solidity/issues/12896#issuecomment-1108853261

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3090,14 @@ string FunctionType::canonicalName() const
...
+		default: return Type::toString();
```
nishant-sachdeva:
```suggestion
        case Kind::Event: return "event " + m_declaration->name();
		default: return Type::toString();
```
We could add this extra case for Events as well as mentioned in https://github.com/ethereum/solidity/issues/12896#issuecomment-1095438214

libsolidity/ast/Types.cpp
StrongerXi: Thanks, done.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3090,16 @@ string FunctionType::canonicalName() const
 	return "function";
 }
 
+string FunctionType::humanReadableName() const
+{
+	switch (m_kind)
+	{
+		case Kind::Error: return "error " + m_declaration->name();
```
chriseth: I think this should also include the signature like `toString` does, shouldn't it?

nishant-sachdeva: As mentioned by @cameel  in the Solidity Dev chat, It's about function parameters
For eg. 
`function abc(uint x, bytes calldata b) public view {}`
The signature is `abc(uint,bytes)`, not just `abc`
Since errors are similar to function types, the same goes for them. Their signatures work the same way.

nishant-sachdeva (commented):
@StrongerXi  Are you still working on the PR :) ?
 This PR is blocking another one of our PRs #12921 , so we were looking for this to be wrapped up quickly.

StrongerXi:
> @StrongerXi Are you still working on the PR :) ? This PR is blocking another one of our PRs #12921 , so we were looking for this to be wrapped up quickly.

Definitely! I've been waiting for feedbacks:).

What's left to do? I see some checks are failing, but they don't seem to be caused by this PR?

StrongerXi:
Rebase.

nishant-sachdeva:
> Rebase.
Please check the changes requested by @chriseth . If needed, you can come to the Solidity Chat and we can discuss how to implement the requested changes.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3090,54 @@ string FunctionType::canonicalName() const
...
+			{
+				name += " returns (";
+				for (auto it = m_returnParameterTypes.begin(); it != m_returnParameterTypes.end(); ++it)
+					name += (*it)->toString(true) + (it + 1 == m_returnParameterTypes.end() ? "" : ",");
+				name += ")";
+			}
+			break;
+		case Kind::Event:
+			name = "event " + m_declaration->name();
+			name += '(';
```
StrongerXi: Should we refactor out the logic for stringifying parameter types? Now it's duplicated here and in `FunctionType::toString`?

StrongerXi: Should we refactor out the logic for stringifying parameter types? Now it's duplicated here and in `FunctionType::toString`?

StrongerXi:
1. Refactored out logic for stringifying parameter/return types.
2. Do not display extra type info such as purity for error/event types.

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,15 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
...
 }
 
+std::string typesToString(TypePointers const& _types, bool _short) {
+	std::string str;
+	str += '(';
+	for (auto it = _types.begin(); it != _types.end(); ++it)
+		str += (*it)->toString(_short) + (it + 1 == _types.end() ? "" : ",");
+	str += ")";
+	return str;
+}
```
ekpyron:
```suggestion
string typesToString(TypePointers const& _types, bool _short)
{
	string str;
	str += '(';
	for (auto it = _types.begin(); it != _types.end(); ++it)
		str += (*it)->toString(_short) + (it + 1 == _types.end() ? "" : ",");
	str += ")";
	return str;
}
```

as some minor coding style remarks...
We could probably also simplify further to something like

```suggestion
string typesToString(TypePointers const& _types, bool _short)
{
  return '(' + joinHumanReadable(_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }) + ')';
}
```

ekpyron (commented):
Only a minor style remark, apart from that this looks reasonable!

libsolidity/ast/Types.cpp
ekpyron: (For reference: https://github.com/ethereum/solidity/blob/develop/CODING_STYLE.md describes most of the general coding style we try to adhere to)

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,12 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+std::string typesToString(TypePointers const& _types, bool _short) {
+	return '(' + solidity::util::joinHumanReadable(
```
ekpyron:
```suggestion
string typesToString(TypePointers const& _types, bool _short)
{
	return '(' + util::joinHumanReadable(
```

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,12 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
...
+			_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); } )
```
ekpyron:
```suggestion
			_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); })
```

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,12 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+std::string typesToString(TypePointers const& _types, bool _short) {
+	return '(' + util::joinHumanReadable(
```
ekpyron:
```suggestion
string typesToString(TypePointers const& _types, bool _short)
{
	return '(' + util::joinHumanReadable(
```

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,13 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string typesToString(TypePointers const& _types, bool _short)
+{
+	return '(' + util::joinHumanReadable(
+			_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); })
+		) + ')';
```
ekpyron: (Even though also discussed via chat)
This probably needs to be
```suggestion
	return '(' + util::joinHumanReadable(
			_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
			","
		) + ')';
```
in order for things to stay consistent with the old behaviour of ``toString``.

ekpyron (dismissed):
Looks good if tests pass!

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,15 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string typesToString(TypePointers const& _types, bool _short)
+{
+	return '(' + util::joinHumanReadable(
+				_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
+				","
+			) +
+		')';
```
ekpyron:
```suggestion
			) + ')';
```
Either *always* break or *never* break is what the coding style says :-).
Also looking at this again, the indentation is off... it's multiple levels of indentation while it should only be one...

StrongerXi: Suggestion applied.

ekpyron:
I think the external test failures are most probably unrelated to the PR itself. We had quite a few fixes in those external tests in the recent past - my guess would be that just rebasing to current ``develop`` should fix them.

ekpyron (dismissed):

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,14 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string typesToString(TypePointers const& _types, bool _short)
+{
+	return '(' + util::joinHumanReadable(
+				_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
+				","
+			) + ')';
```
cameel:
```suggestion
	return '(' + util::joinHumanReadable(
		_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
		","
	) + ')';
```

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3098,26 @@ string FunctionType::canonicalName() const
...
+			name += typesToString(m_parameterTypes, true);
+			break;
+		case Kind::Event:
+			name = "event " + m_declaration->name();
+			name += typesToString(m_parameterTypes, true);
+			break;
+		default:
+			name = Type::toString();
+			break;
+	}
```
cameel:
```suggestion
	switch (m_kind)
	{
	case Kind::Error:
		name = "error " + m_declaration->name();
		name += typesToString(m_parameterTypes, true);
		break;
	case Kind::Event:
		name = "event " + m_declaration->name();
		name += typesToString(m_parameterTypes, true);
		break;
	default:
		name = Type::toString();
}
```

libsolidity/ast/Types.cpp
StrongerXi: Done.

libsolidity/ast/Types.cpp
StrongerXi: Done.

StrongerXi:
Rebase.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3098,25 @@ string FunctionType::canonicalName() const
...
+		name += typesToString(m_parameterTypes, true);
+		break;
+	case Kind::Event:
+		name = "event " + m_declaration->name();
+		name += typesToString(m_parameterTypes, true);
+		break;
+	default:
+		name = Type::toString();
+	}
+	return name;
```
cameel: This would be simpler without the intermediate `name` variable.
```suggestion
	switch (m_kind)
	{
	case Kind::Error:
		return "error " + m_declaration->name() + typesToString(m_parameterTypes, true);
	case Kind::Event:
		return "event " + m_declaration->name() + typesToString(m_parameterTypes, true);
	default:
		return Type::toString();
	}
```

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,14 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string typesToString(TypePointers const& _types, bool _short)
```
cameel: I think that the name of this function is too generic for what it does. It specifically creates a parameters for use in a signature and makes assumptions related to that, i.e. that there should be no space after `,` or that you need the parentheses. From the name I would assume that I could use it for example to print the parameter list for a function definition but that's not the case (you need the space). So I'd change it like this:
```suggestion
string formatSignatureParameterList(TypePointers const& _types)
```
Then assume also that `_short` is always `true`.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3098,25 @@ string FunctionType::canonicalName() const
...
 }
 
+string FunctionType::humanReadableName() const
+{
+	string name;
+	switch (m_kind)
+	{
+	case Kind::Error:
+		name = "error " + m_declaration->name();
+		name += typesToString(m_parameterTypes, true);
```
cameel: We already have a function you could use here:
```suggestion
		name = "error " + m_declaration->externalSignature();
```

I think that you should make `externalSignature()` use your new `typesToString()` though.

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
```diff
@@ -0,0 +1,11 @@
+interface MyInterface {
+    error MyCustomError(uint256, bool);
+}
+
+contract Test {
+    function test() public returns(bytes4) {
+        return bytes4(MyInterface.MyCustomError);
+    }
+}
```
cameel, line 9: There are 3 errors in `TypeChecker.cpp` that you changed but this test triggers only one. Can you add more conversions here so that we hit all 3 of them?

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
```diff
@@ -0,0 +1,11 @@
+interface MyInterface {
+    error MyCustomError(uint256, bool);
+}
+
+contract Test {
+    function test() public returns(bytes4) {
+        return bytes4(MyInterface.MyCustomError);
```
cameel, line 7: Would `bytes4(MyInterface.MyCustomError(1, true))` also trigger this error message? How about `bytes4(MyInterface.MyCustomError())`? Can you add these to the test?

Same for the event test below.

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
```diff
@@ -0,0 +1,11 @@
...
+    }
+}
+// ----
+// TypeError 9640: (143-176): Explicit type conversion not allowed from "error MyCustomError(uint256,bool)" to "bytes4".
```
cameel, line 11: Explicit conversions are not the only case where we should use `humanReadableName()`. For example if you do this in the function above:
```solidity
return MyInterface.MyCustomError;
```
you will get
```
Error: Return argument type function (uint256,bool) pure is not implicitly convertible to expected type (type of first return variable) bytes4.
```

The original issue only mentions the explicit conversion, but I think we really should do a more thorough job here. Can you look for other cases like this? If you prefer, I think it would be fine to just report them as a new issue instead of adding that to this PR so that we can finally merge it but I think we need to fix them eventually.

cameel (changes requested):

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
```diff
@@ -0,0 +1,9 @@
+contract Test {
+    event MyCustomEvent(uint256);
```
cameel, line 2: Can you add more types of parameters here to make sure that when we change the PR to use `externalSignature()`, the result is still what we want? I'd add:
- a reference type, e.g. an array
- a function type with some parameters and return values
- an enum
- a payable address
- an interface

Same in the error test case.

libsolidity/ast/Types.cpp
cameel: After discussing with @ekpyron, `externalSignature()` is probably not what we want (it will print types like they are in the ABI) but what I really meant is that we definitely already have code for printing these parameters somewhere.

Maybe you could look inside `toString()` and check how it constructs the parameter list of a function. If there's a function for doing that - use it. If not, create such a function so that it can be reused.

libsolidity/ast/Types.cpp
StrongerXi: Done.

libsolidity/ast/Types.cpp
StrongerXi: Name updated, but I think we still need `bool _short` in order to reuse this in `FunctionType::toString`?

libsolidity/ast/Types.cpp
StrongerXi: > If not, create such a function so that it can be reused.

Done (the current strategy already does this).

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
StrongerXi: Both `MyInterface.MyCustomError(1, true)` and `MyInterface.MyCustomError()` end up being an empty tuple type, i.e., `tuple()`, with the former causing an additional arity error. So the error message is like the following, which is unrelated to this PR imo.

```
Explicit type conversion not allowed from "tuple()" to "bytes4".
```

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
StrongerXi: Removed 1 case, and added test for the other case.

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
StrongerXi: > when we change the PR to use externalSignature()

Since we are not using `externalSignature()` anymore per comments above, I assume this comment becomes outdated too?

StrongerXi:
Rebase.

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
cameel: ok. I'd still include those just for more coverage. One might expect them to work like the ones where you use the type.

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
cameel: Actually, I'd still add this. `externalSignature()` was the initial motivation but it would still be good to have something that claerly shows how these types are handled. For example I'm not really sure if it will print `bytes memory` or `bytes` like in an error/event definition.

libsolidity/ast/Types.cpp
```diff
@@ -3090,6 +3098,19 @@ string FunctionType::canonicalName() const
...
+string FunctionType::humanReadableName() const
+{
+	switch (m_kind)
+	{
+	case Kind::Error:
+		return "error " + m_declaration->name() + signatureParameterList(m_parameterTypes, true);
+	case Kind::Event:
+		return "event " + m_declaration->name() + signatureParameterList(m_parameterTypes, true);
+	default:
+		return Type::toString();
```
cameel: Can you explicitly pass `false` to `toString()` and also add the parameter name?
```suggestion
	case Kind::Error:
		return "error " + m_declaration->name() + signatureParameterList(m_parameterTypes, /* _short */ true);
	case Kind::Event:
		return "event " + m_declaration->name() + signatureParameterList(m_parameterTypes, /* _short */ true);
	default:
		return Type::toString(/* _short */ false);
```

Otherwise it's not that easy to spot. I just noticed this and started wondering why we don't use the same value. So apparently `_short` affects only whether we print the location and nothing else. Might be good to add a comment about that because that's not obvious.

libsolidity/ast/Types.cpp
cameel: Right. You do need `_short`. I missed the fact that in the `toString()` we do use it.

libsolidity/ast/Types.cpp
cameel: Haha, sorry, while writing this I somehow missed the fact that the code I was looking at was actually inside one of the specializations of `toString()` :)

What I really meant its that I was sure we have code for printing of events/errors, including their signature implemented somewhere. But looks like I might be wrong - I just looked for it again and could not find anything. Apparently we really just print them as functions. In places that matter (like the output of `solc --hashes` we do use `externalSignature()`).

So ok, looks like this is the right way do it after all.

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
StrongerXi: Done.

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
StrongerXi: Can function be used as event/error parameter? I keep getting `TypeError 3417 Internal or recursive type is not allowed as error parameter type` for declaration like:
```solidity
error MyCustomError(function (string memory) returns (uint));
```

All other cases you mentioned are added.

libsolidity/ast/Types.cpp
StrongerXi: Done.

StrongerXi:
Rebase.

libsolidity/ast/Types.cpp
```diff
@@ -110,6 +110,14 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
+string signatureParameterList(TypePointers const& _types, bool _short)
```
chriseth: If you look at the implementation of this function, there is no "parameter" to be found, yet it is in the name of the function.
Maybe `toStringInParentheses` would be a better name?

StrongerXi: Done.

StrongerXi: Agree, done.

StrongerXi:
Rebase.

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
cameel: It can but it must be an external function. I.e.
```solidity
error MyCustomError(function (string memory) external returns (uint));
```
Function pointers are internal by default.

libsolidity/ast/Types.cpp
cameel: > there is no "parameter" to be found, yet it is in the name of the function.

I suggested that name because that's the intended usage and the function makes some small assumptions related to that (like the lack of space after the comma; I originally wanted it to skip `_short` too but we need it after all). It's not really a generic `toString()` so I think it's misleading to call it that.

But ok, let's leave it at `toStringInParentheses()` for now to avoid going back and forth on this.

test/libsolidity/syntaxTests/conversion/explicit_conversion_from_event.sol
```diff
@@ -0,0 +1,23 @@
+interface MyInterface {
+    enum MyEnum { E1, E2 }
+}
+
+contract Test {
+    error MyCustomEvent(uint256, bool, bool[], address payable, MyInterface, MyInterface.MyEnum);
```
cameel:
```suggestion
    event MyCustomEvent(uint256, bool, bool[], address payable, MyInterface, MyInterface.MyEnum);
```

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
cameel: Can you create an issue for this?

nishant-sachdeva: Has this issue been created ? @cameel

cameel: Doesn't seem so.

test/libsolidity/syntaxTests/conversion/explicit_conversion_event_to_bytes4.sol
cameel: @nishant-sachdeva This thread was still unresolved due to this part:

> - a function type with some parameters and return values

It's still not done. Can you add a function pointer to `explicit_conversion_from_error.sol` and `explicit_conversion_from_event.sol`?

test/libsolidity/syntaxTests/conversion/explicit_conversion_from_event.sol
```diff
@@ -0,0 +1,34 @@
...
+
+    event CustomEvent1(
+        uint256,
+        bool, 
+        bool[],
+        address payable,
+        MyInterface,
+        MyInterface.MyEnum,
+        function (string memory) external returns (uint)
+    );
```
cameel: `chk_coding_style` is failing because of this. Please configure your editor to strip trailing whitespace.
```suggestion
    );
```

test/libsolidity/syntaxTests/conversion/explicit_conversion_from_error.sol
```diff
@@ -0,0 +1,32 @@
...
+}
+
+contract Test {
+    function testFunction(string memory) external returns (uint) {}
+
+    function test() public {
+        MyInterface instance = MyInterface(msg.sender);
+        bool[] calldata arr;
+        address payable addr;
+        bytes4(MyIrnterface.CustomEror1);
```
cameel:
```suggestion
        bytes4(MyInterface.CustomError1);
```

cameel (dismissed):
There are some typos that make CI fail and we still need https://github.com/ethereum/solidity/pull/13005#discussion_r877162339 done but other than that there are no big issues to solve here so I'm approving.

test/libsolidity/syntaxTests/conversion/explicit_conversion_error_to_bytes4.sol
nishant-sachdeva: Created issue at #13106

ekpyron (approved):

cameel (approved):

PR #13006, opened 2022-05-11
Fix asorted compilation issues with GCC 12.

ekpyron:
Apparently gcc 12 started warning about this - and strictly speaking it's actually right in doing so, even though in our particular case, the current ordering shouldn't cause issues.

### DIFF
--- a/libsolutil/Numeric.h
+++ b/libsolutil/Numeric.h
@@ -29,7 +29,15 @@
 #error "Unsupported Boost version. At least 1.65 required."
 #endif
 
+// TODO: do this only conditionally as soon as a boost version with gcc 12 support is released.
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
 #include <boost/multiprecision/cpp_int.hpp>
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
+#pragma GCC diagnostic pop
+#endif
 
 #include <limits>
 
--- a/test/Metadata.cpp
+++ b/test/Metadata.cpp
@@ -94,7 +94,10 @@ public:
 				else if (value == 21)
 					return "true";
 				else
+				{
 					assertThrow(false, CBORException, "Unsupported simple value (not a boolean).");
+					return ""; // unreachable, but prevents compiler warning.
+				}
 			}
 			default:
 				assertThrow(false, CBORException, "Unsupported value type.");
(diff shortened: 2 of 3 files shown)


cameel (approved):

cameel (dismissed):

leonardoalt:
it doesn;t

libsolutil/Numeric.h
```diff
@@ -29,7 +29,15 @@
 #error "Unsupported Boost version. At least 1.65 required."
 #endif
 
+// TODO: do this only conditionally as soon as a boost version with gcc 12 support is released.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
```
ekpyron: This is rather ugly... I'd have expected the build by default to consider boost headers as system libraries and not to have ``-Werror`` apply to them - not entirely sure why that's not the case...

libsolutil/Numeric.h
```diff
@@ -29,7 +29,15 @@
 #error "Unsupported Boost version. At least 1.65 required."
 #endif
 
+// TODO: do this only conditionally as soon as a boost version with gcc 12 support is released.
```
leonardoalt, line 32: maybe open an issue to keep track of this?

test/Metadata.cpp
```diff
@@ -94,7 +94,10 @@ class TinyCBORParser
 				else if (value == 21)
 					return "true";
 				else
+				{
 					assertThrow(false, CBORException, "Unsupported simple value (not a boolean).");
+					return ""; // unreachable, but prevents compiler warning.
```
leonardoalt, line 99: bad compiler

libsolutil/Numeric.h
```diff
@@ -29,7 +29,15 @@
...
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
 #include <boost/multiprecision/cpp_int.hpp>
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
+#pragma GCC diagnostic pop
+#endif
```
leonardoalt, line 40: wow this is ugly

leonardoalt (approved):
the solution isn't really nice, but if you feel like that's the easiest way, fine with me

libsolutil/Numeric.h
ekpyron: https://github.com/ethereum/solidity/issues/13008

PR #13000, opened 2022-05-10
[SMTChecker] Fix check that solver is available

leonardoalt:
Fixes https://github.com/ethereum/solidity/issues/12998

PR #13009, opened 2022-05-11
Support new z3 AST node

leonardoalt:
Somehow https://github.com/ethereum/solidity/pull/12997 got merged so reopening here

Depends on https://github.com/ethereum/solidity/pull/13000

DO NOT MERGE before /\

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@ Bugfixes:
  * Type Checker: Properly check restrictions of ``using ... global`` in conjunction with libraries.
  * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
+ * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
  * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
 
--- a/docs/smtchecker.rst
+++ b/docs/smtchecker.rst
@@ -629,6 +629,11 @@ option ``--model-checker-solvers {all,cvc4,smtlib2,z3}`` or the JSON option
   - if a dynamic ``z3`` library of version 4.8.x is installed in a Linux system (from Solidity 0.7.6);
   - statically in ``soljson.js`` (from Solidity 0.6.9), that is, the Javascript binary of the compiler.
 
+.. note::
+  z3 version 4.8.16 broke ABI compatibility with previous versions and cannot
+  be used with solc <=0.8.13. If you are using z3 >=4.8.16 please use solc
+  >=0.8.14.
+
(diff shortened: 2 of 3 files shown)


docs/smtchecker.rst
```diff
@@ -629,6 +629,11 @@ option ``--model-checker-solvers {all,cvc4,smtlib2,z3}`` or the JSON option
   - if a dynamic ``z3`` library of version 4.8.x is installed in a Linux system (from Solidity 0.7.6);
   - statically in ``soljson.js`` (from Solidity 0.6.9), that is, the Javascript binary of the compiler.
 
+.. note::
+  z3 version 4.8.15 broke ABI compatibility with previous versions and can only
```
ekpyron:
```suggestion
  z3 version 4.8.15 broke ABI compatibility with previous versions and cannot
```
:-)?

leonardoalt: done

leonardoalt:
I changed the z3 version in this PR docs also, the change was in 4.8.16 and not 15.

Changelog.md
```diff
@@ -15,6 +15,7 @@ Bugfixes:
  * Type Checker: Properly check restrictions of ``using ... global`` in conjunction with libraries.
  * Assembly-Json: Fix assembly json export to store jump types of operations in `jumpType` field instead of `value`.
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
+ * SMTChecker: Fix ABI compatibility with z3 >=4.8.15.
```
```suggestion
 * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
```
IIUC :-)?

leonardoalt: sigh

ekpyron (approved):

Andy53:
We currently have this in our CMakeLists:
```
set(LATEST_Z3_VERSION "4.11.0")
set(MINIMUM_Z3_VERSION "4.8.0")
find_package(Z3)
if (${Z3_FOUND})
  if (${STRICT_Z3_VERSION})
    if (NOT ("${Z3_VERSION_STRING}" VERSION_EQUAL ${LATEST_Z3_VERSION}))
      message(
        FATAL_ERROR
        "SMTChecker tests require Z3 ${LATEST_Z3_VERSION} for all tests to pass.\n\
Build with -DSTRICT_Z3_VERSION=OFF if you want to use a different version. \
You can also use -DUSE_Z3=OFF to build without Z3. In both cases use --no-smt when running tests."
      )
    endif()
  else()
    if ("${Z3_VERSION_STRING}" VERSION_LESS ${MINIMUM_Z3_VERSION})
      message(
        FATAL_ERROR
        "Solidity requires Z3 ${MINIMUM_Z3_VERSION} or newer. You can also use -DUSE_Z3=OFF to build without Z3."
      )
    endif()
  endif()
endif()
```

I think this covers this issue however if that is not the case can someone please elaborate on where this is lacking. I have been able to build solidity using version 4.11.0 and 4.9.1. Both built however the only issue I ran into is that when building Z3 from source or using the latest release they cause an error since they are newer than 4.11.0 (4.11.3/4.11.2)

Is it worth us just updating `LATEST_Z3_VERSION` to `4.11.3`?

PR #13013, opened 2022-05-12
Revert "Disallow RETURNDATASIZE and RETURNDATACOPY in inline assembly blocks in pure functions"

matheusaaguiar:
This reverts commit f567eb1fb2a24dbd956fe65512d4b9015ba7f5ed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,8 +21,6 @@ Bugfixes:
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
  * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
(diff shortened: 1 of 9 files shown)


cameel:
To add some context to this PR: we're reverting #12861 because it turned out that there are projects that actually use `RETURNDATACOPY` in pure functions. We're going to merge that feature into `breaking` instead.

cameel (commented):
Looks good. But please squash the corrections into the original commit. Our general practice is for each commit to be a single logical change, without changing things back and forth. This keeps history easier to understand.

ekpyron:
> Looks good. But please squash the corrections into the original commit. Our general practice is for each commit to be a single logical change, without changing things back and forth. This keeps history easier to understand.

(I'd usually use ``git rebase -i develop`` for that, in case you're wondering; that'll open up an editor with some comments which explain the rest)

ekpyron (approved):

cameel (approved):

Sticknmove (commented):
Thanks

Sticknmove:
Thanks

Issue #13017, opened 2022-05-13
Confusing error message when using function calls that don't return anything as expressions.


0xalpharush:
## Description

The behavior of dynamic storage arrays with struct type is inconsistent and the compiler output is not helpful. I would expect that `pop1` and `pop3` be semantically equivalent; that is, `.pop()` would return the last array element and delete it. However, the error message for `pop3` suggests that a type of tuple is return by `.pop()`. 

When I treat it as a tuple as suggested, the error message says there's a different number of components... Ideally one could cast this tuple to the struct type or receive the struct type and manually reference values, but neither works with `.pop()`. 

See also that the number of components is different when accessed via index versus using pop (`pop2` vs `pop4`).

## Environment

- Compiler version: Version: 0.8.13+commit.abaa5c0e.Darwin.appleclang
- Target EVM version (as per compiler settings): NA
- Framework/IDE (e.g. Truffle or Remix): NA
- EVM execution environment / backend / blockchain client:  NA
- Operating system: MacOS

## Steps to Reproduce
Run `solc test.sol`

```solidity
pragma solidity =0.8.13;
contract Test {
  struct MyStruct{
    uint num;
    bytes32 info;
  }
  MyStruct[] myStruct;
  function add(uint a, bytes32 b) external {
    myStruct.push(MyStruct({num: a, info: b}));
  }
  function pop1() external returns(MyStruct[] memory myMemoryStruct) {
   if (myStruct.length != 0) {
     myMemoryStruct = new MyStruct[](1);
     myMemoryStruct[0] = myStruct[myStruct.length - 1];
     delete myStruct[myStruct.length - 1];
   }
  }
  function pop2() external returns(MyStruct[] memory myMemoryStruct) {
   if (myStruct.length != 0) {
     myMemoryStruct = new MyStruct[](1);
     (uint a, bytes32 b) = myStruct[myStruct.length - 1];
   }
  }

  function pop3() external returns(MyStruct[] memory myMemoryStruct){
   if (myStruct.length != 0) {
     myMemoryStruct = new MyStruct[](1);
     myMemoryStruct[0] = myStruct.pop();
   }
  }

  function pop4() external returns(MyStruct[] memory myMemoryStruct) {
   if (myStruct.length != 0) {
      myMemoryStruct = new MyStruct[](1);
     (uint a, bytes32 b) = myStruct.pop();
   }
  }
}

```
```
Error: Different number of components on the left hand side (2) than on the right hand side (1).
[...]

PR #13018, opened 2022-05-13
document that pop does not return last element

0xalpharush:
It's already documented for push and this would clarify my incorrect assumption (that pop returns a value as in other languages) that caused confusion with https://github.com/ethereum/solidity/issues/13017


ekpyron (approved):

ekpyron:
Thank you very much, this should indeed have been documented more clearly!

PR #13019, opened 2022-05-13
Remove enforce-via-yul CI run and instead always force via yul in regular test runs.

ekpyron:
I noticed when https://circleci.com/gh/ethereum/solidity/1060053 ran out of memory.
It's rather wasteful to run the smt tests in this run again - they're already covered and not affected by the option.

We might want to look for other cases like this...
Also we could think about restricting this further... the flag really only affects the semantics tests, doesn't it? But most of it is cheap anyways, so the SMT tests are the most wasteful ones I guess.


.circleci/config.yml
```diff
@@ -1022,7 +1022,7 @@ jobs:
     parallelism: 20
     environment:
       EVM: << pipeline.parameters.evm-version >>
-      SOLTEST_FLAGS: --enforce-via-yul
+      SOLTEST_FLAGS: --enforce-via-yul --no-smt
```
cameel: I see that `soltest_all.sh` uses `-t !smtCheckerTests` to disable the SMT tests. Now I wonder if there's any difference between these...

cameel: > Also we could think about restricting this further... the flag really only affects the semantics tests, doesn't it? But most of it is cheap anyways, so the SMT tests are the most wasteful ones I guess.

How about this? I wonder if this would actually work given the space after `-t`.
```suggestion
      SOLTEST_FLAGS: --enforce-via-yul -t semanticTests
```

cameel (dismissed):
I have some suggestions but overall the PR seems good.

.circleci/config.yml
ekpyron: ``-t semanticsTests`` you mean? But yeah, not sure if it'd work :-). I also haven't checked yet whether the flag is really *only* relevant for the extracted semantics tests.

ekpyron:
I'm actually not quite sure why we have a separate run for this in general :-)... is that to prevent slow-downs of the tests? Not sure that this would still make a huge difference... almost all tests should work via yul as well anyways with only a handful of exceptions for IR-breaking-changes... maybe we should instead just enable this flag by default? Resp. remove the flag altogether and just always check this?

ekpyron:
So maybe doing this merely conditionally is just outdated :-).

.circleci/config.yml
ekpyron: Probably pretty much equivalent...

cameel: > `-t semanticsTests` you mean?

Right, fixed.

Also, I checked now and looks like the `--enforce-via-yul` flag is handled specifically by `SemanticTest.h`. Also the only place where the `compileViaYul` directive ever appears are tests in `semanticTests/`. So it should be safe.

cameel:
> I'm actually not quite sure why we have a separate run for this in general :-)

```c++
	if (m_testCaseWantsLegacyRun)
		result = runTest(_stream, _linePrefix, _formatted, false, false);

	if ((m_testCaseWantsYulRun || m_enforceViaYul) && result == TestResult::Success)
		result = runTest(_stream, _linePrefix, _formatted, true, false);
```

This means that it does add some overhead. We'll run via Yul even in tests that are known to work only via legacy. But yeah, not sure now much overhead that adds. Might be negligible now that almost everything works via Yul. Overhead could have been bigger in the past.

I think it would be a good idea to eliminate this run.

ekpyron:
Yeah, *all* tests except the very few ones that test for legacy behaviour on IR breaking changes should have ``m_testCaseWantsYulRun`` set to true anyways - I guess that's negligible overhead. I'm just looking into eliminating the run.

ekpyron:
We can compare the test run times against develop, but I don't expect significant changes. If we want to merge this, we have to remove the run as required CI run, though.

cameel:
I changed it to not required in settings.

ekpyron:
That's the exhaustive list:
```
test/libsolidity/semanticTests/abiEncoderV1/abi_encode_empty_string.sol:// compileViaYul: false
test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol:// compileViaYul: false
test/libsolidity/semanticTests/calldata/calldata_array_dynamic_bytes.sol:// compileViaYul: false
test/libsolidity/semanticTests/cleanup/cleanup_bytes_types_shortening_OldCodeGen.sol:// compileViaYul: false
test/libsolidity/semanticTests/constructor_inheritance_init_order_3_legacy.sol:// compileViaYul: false
test/libsolidity/semanticTests/modifiers/function_modifier_loop.sol:// compileViaYul: false
test/libsolidity/semanticTests/modifiers/function_modifier_multi_invocation.sol:// compileViaYul: false
test/libsolidity/semanticTests/modifiers/function_modifier_multi_with_return.sol:// compileViaYul: false
test/libsolidity/semanticTests/operators/shifts/shift_bytes_cleanup.sol:// compileViaYul: false
test/libsolidity/semanticTests/revertStrings/empty_v1.sol:// compileViaYul: false
test/libsolidity/semanticTests/revertStrings/function_entry_checks_v1.sol:// compileViaYul: false
test/libsolidity/semanticTests/types/mapping_enum_key_v1.sol:// compileViaYul: false
test/libsolidity/semanticTests/uninitializedFunctionPointer/uninitialized_internal_storage_function_legacy.sol:// compileViaYul: false
test/libsolidity/semanticTests/userDefinedValueType/assembly_access_bytes2_abicoder_v1.sol:// compileViaYul: false
test/libsolidity/semanticTests/userDefinedValueType/cleanup_abicoderv1.sol:// compileViaYul: false
test/libsolidity/semanticTests/userDefinedValueType/conversion_abicoderv1.sol:// compileViaYul: false
test/libsolidity/semanticTests/various/negative_stack_height.sol:// compileViaYul: false
```

So it's 17 out of over 1300 - should be fine.
Did we ever check that all of those 17 cases are all legitimately still there :-)? I guess so...

EDIT: ah no, that's actually wrong... those are the ones that will not even run via yul in the enforce-via-yul run...

cameel:
You mean did we check if they already work via Yul? That's what the job was supposed to check :)

But most of them seem to require abicoder v1 and AFAIK that's only available on legacy.

cameel (dismissed):

ekpyron:
Yeah, I meant whether for each of them there's a reason for them still *not* working via yul. But it was the wrong set of cases...
We actually have this in ``SemanticTest.cpp``:

```
	// Do not enforce via yul and ewasm, if via yul was explicitly denied.
	if (compileViaYul == "false")
	{
		m_enforceViaYul = false;
		m_enforceCompileToEwasm = false;
	}
```

so these cases actually won't even be run in the enforce-via-yul run anyways...
So it's really only those *without* a ``compileViaYul`` setting...

That's those:
```
abiEncoderV1/bool_out_of_bounds.sol
abiEncoderV1/cleanup/cleanup.sol
abiEncoderV1/enums.sol
abiencodedecode/abi_encode_empty_string_v1.sol
arithmetics/checked_add_v1.sol
cleanup/bool_conversion_v1.sol
cleanup/cleanup_address_types_v1.sol
cleanup/cleanup_bytes_types_v1.sol
operators/shifts/shift_right_garbled_signed_v1.sol
operators/shifts/shift_right_garbled_v1.sol
operators/shifts/shift_right_negative_lvalue_signextend_int16_v1.sol
operators/shifts/shift_right_negative_lvalue_signextend_int32_v1.sol
operators/shifts/shift_right_negative_lvalue_signextend_int8_v1.sol
revertStrings/calldata_too_short_v1.sol
revertStrings/enum_v1.sol
revertStrings/invalid_abi_decoding_calldata_v1.sol
revertStrings/invalid_abi_decoding_memory_v1.sol
types/mapping_enum_key_getter_v1.sol
types/mapping_enum_key_library_v1.sol
```

and all of them indeed seem to be abi-encoder-v1 related tests.

So we could actually probably even go ahead and add ``compileViaYul: false`` to them... and then just either require each test to have a ``compileViaYul`` setting or make ``compileViaYul: also`` the default...

ekpyron:
Ok, so new summary:

Tests with an explicit ``compileViaYul: false`` were not even tried by ``enforceViaYul``. It only kicked in for tests *without any* ``compileViaYul`` option.
And looking at those cases, it turns out that all of them were ``// ABIEncoderV1Only: true`` cases.

So I changed the logic to the following:

- Every test case that has ``// ABIEncoderV1Only: true`` is implicitly ``compileViaYul: false`` (I removed it where it was additionally added explicitly and made having ``compileViaYul`` together with ``ABIEncoderV1Only`` an error to ensure consistency).
- Every other test *needs* to explicitly specify ``compileViaYul``.


If this version now passes in CI, we can actually consider making ``compileViaYul: also`` the default and remove it from all test cases... or just merge it.

cameel:
> Every test case that has `// ABIEncoderV1Only: true` is implicitly `compileViaYul: false`

I actually think it's clearer to have it explicitly set to false. At least as long as `compileViaYul` is still a thing. For me at least for a long time it was not obvious at all that abicoder v1 only runs with legacy codegen and that we silently switch to v2 under the hood on IR. 

I'd be fine with changing the default to `also` though and just not requiring that directive by default.

ekpyron:
> > Every test case that has `// ABIEncoderV1Only: true` is implicitly `compileViaYul: false`
> 
> I actually think it's clearer to have it explicitly set to false. At least as long as `compileViaYul` is still a thing. For me at least for a long time it was not obvious at all that abicoder v1 only runs with legacy codegen and that we silently switch to v2 under the hood on IR.
> 
> I'd be fine with changing the default to `also` though and just not requiring that directive by default.

Ok, I'll switch it around and require them to be explicitly set to ``compileViaYul: false``. I don't mind either way - just wanted it to be the same everywhere :-).

ekpyron:
Ok, now ``// ABIEncoderV1Only: true`` *requires* ``// compileViaYul: false``.

I still haven't made ``compileViaYul: also`` the default, though, and currently it still *requires* a ``compileViaYul`` setting - but that's a bit convenient, so we should still change it to just make ``compileViaYul: also`` the default. But that requires some more adjustments in the update logic, I won't do that today.

test/libsolidity/SemanticTest.cpp
```diff
@@ -60,27 +59,32 @@ SemanticTest::SemanticTest(
...
+	m_runWithABIEncoderV1Only = m_reader.boolSetting("ABIEncoderV1Only", false);
+	if (m_runWithABIEncoderV1Only && !solidity::test::CommonOptions::get().useABIEncoderV1)
+		m_shouldRun = false;
+
 	string compileViaYul = m_reader.stringSetting("compileViaYul", "default");
+	if (m_runWithABIEncoderV1Only && compileViaYul != "false")
+		BOOST_THROW_EXCEPTION(runtime_error(
+			"ABIEncoderV1Only tests cannot be run via yul, "
+			"so they need to also specify ``compileViaYul: false``"
+			));
```
cameel:
```suggestion
		BOOST_THROW_EXCEPTION(runtime_error(
			"ABIEncoderV1Only tests cannot be run via yul, "
			"so they need to also specify ``compileViaYul: false``"
		));
```

test/libsolidity/SemanticTest.cpp
```diff
@@ -60,27 +59,32 @@ SemanticTest::SemanticTest(
...
 	if (!util::contains(compileViaYulAllowedValues, compileViaYul))
 		BOOST_THROW_EXCEPTION(runtime_error("Invalid compileViaYul value: " + compileViaYul + "."));
 	m_testCaseWantsYulRun = util::contains(yulRunTriggers, compileViaYul);
 	m_testCaseWantsLegacyRun = util::contains(legacyRunTriggers, compileViaYul);
 
 	// Do not enforce via yul and ewasm, if via yul was explicitly denied.
```
cameel:
```suggestion
	// Do not enforce ewasm, if via yul was explicitly denied.
```

cameel (dismissed):

test/libsolidity/SemanticTest.cpp
```diff
@@ -307,7 +299,7 @@ TestCase::TestResult SemanticTest::run(ostream& _stream, string const& _linePref
 	if (m_testCaseWantsLegacyRun)
 		result = runTest(_stream, _linePrefix, _formatted, false, false);
 
-	if ((m_testCaseWantsYulRun || m_enforceViaYul) && result == TestResult::Success)
+	if (m_testCaseWantsYulRun && result == TestResult::Success)
```
cameel, line 302: > I still haven't made ``compileViaYul: also`` the default, though, and currently it still *requires* a ``compileViaYul`` setting

In that case maybe it would be better to behave here as if `m_enforceViaYul` was always true? Because otherwise this basically disables the enforcement and I think this makes sense only if the default is to run via yul.

cameel: Just to clarify this - this is not needed after all because the PR initially disallowed omitting `compileViaYul` and now it made it default to `also` instead. So we don't really need enforcement - the Yul run can only be be skipped if you explicitly request it.

cameel (approved):
Looks good. I haven't really reviewed all the test changes, only scrolled through about half of them until github refused to load more diffs. The C++ code looks correct though and tests pass so I'm assuming it must all be fine.

ekpyron:
> Looks good. I haven't really reviewed all the test changes, only scrolled through about half of them until github refused to load more diffs. The C++ code looks correct though and tests pass so I'm assuming it must all be fine.

I basically just removed ``// compileViaYul: also`` using sed and then removed ``// ====``, if immediately followed by ``// ----`` - should be safe :-).

ekpyron:
FYI: as an additional sanity test, I just added ``revert(0,0)`` in IR codegen to the beginning of every contract and ran the tests - and they indeed fail as expected, so the tests *are* still run via Yul :-). (Just because from CI one wouldn't be able to tell :-)).

PR #13022, opened 2022-05-14
Add tests for `abi.encodeCall()` ICE on internal function passed in arg of different type

### LINKED ISSUE #12774: ICE in `abi.encodeCall()` when an internal function is passed in for an argument of a different type
## Description
When you use `abi.encodeCall()` with a function that takes at least one argument, and you pass in an internal function pointer, the compiler reports an Internal Compiler Error.

```
/solidity/libsolidity/ast/Types.cpp(3039): Throw in function virtual std::string solidity::frontend::FunctionType::canonicalName() const
```

This happens only specifically with internal function pointers and not with internal types in general. For example having such a pointer wrapped in a struct does not cause an ICE. Passing in an external function pointer is fine too.

### Passing in an internal function pointer
```solidity
contract C {
    function f(uint) external {}

    function main() external view {
        function () h;
        abi.encodeCall(this.f, (h));
    }
}
```
```
Internal compiler error:
/solidity/libsolidity/ast/Types.cpp(3039): Throw in function virtual std::string solidity::frontend::FunctionType::canonicalName() const
std::exception::what: Solidity assertion failed
[solidity::util::tag_comment*] = Solidity assertion failed
```

The example uses a function type variable but passing in an existing function by name also causes this.

### Passing in an internal type
```solidity
struct S {
    function () f;
}

contract C {
    function f(uint) external {}

    function main() external view {
        S memory s;
        abi.encodeCall(this.f, (s));
    }
}
```
```
Error: Cannot implicitly convert component at position 0 from "S" to "uint256".
  --> test.sol:10:32:
   |
10 |         abi.encodeCall(this.f, (s));
   |               
```

## Environment
- Compiler version: 0.8.12

nishant-sachdeva:
issue has already been fixed. Adding test cases for confirmation

closes #12774

### DIFF
--- a/test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_args_internal_function_pointer_for_uint.sol
+++ b/test/libsolidity/syntaxTests/specialFunctions/encodeCall_fail_args_internal_function_pointer_for_uint.sol
@@ -1,0 +1,10 @@
+contract C {
+    function f(uint) external {}
+
+    function main() external view {
+        function () h;
+        abi.encodeCall(this.f, (h));
+    }
+}
+// ----
+// TypeError 5407: (137-140): Cannot implicitly convert component at position 0 from "function ()" to "uint256".
--- a/test/libsolidity/syntaxTests/specialFunctions/encodecall_fail_args_internal_struct_for_uint.sol
+++ b/test/libsolidity/syntaxTests/specialFunctions/encodecall_fail_args_internal_struct_for_uint.sol
@@ -1,0 +1,43 @@
+struct S {
+    function () f;
+}
+
+contract C {
+    enum testEnum { choice1, choice2, choice3 }
+
+    function f1(uint8) external {}
+    function f2(uint32) external {}
+    function f3(uint) external {}
+    function g1(bytes memory) external {}
+    function g2(bytes32) external {}
+    function h(string memory) external {}
+    function i(bool) external {}
+    function j(address) external {}
+    function k(address payable) external {}
+    function l(testEnum) external {}
+
+    function main() external view {
+        S memory s;
+        abi.encodeCall(this.f1, (s));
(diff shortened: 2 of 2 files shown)


test/libsolidity/syntaxTests/abiEncoder/passing_internal_function_pointer_to_abi_encodecall.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
```
cameel: The original tests for `abi.encodeCall()` are in `syntaxTests/specialFunctions/`. I'd move this one there too to keep them in one place.

This one could be `syntaxTests/specialFunctions/encodeCall_fail_args_internal_function_ptr_for_uint.sol` and the other `syntaxTests/specialFunctions/encodeCall_fail_args_internal_struct_for_uint.sol`.

test/libsolidity/syntaxTests/abiEncoder/passing_internal_type_to_abi_encodecall.sol
```diff
@@ -0,0 +1,14 @@
...
+}
+
+contract C {
+    function f(uint) external {}
+
+    function main() external view {
+        S memory s;
+        abi.encodeCall(this.f, (s));
+    }
+}
```
cameel: Could you add another test that goes over more types? We're clearly missing coverage here. It would be good to have one that tries as many different types as possible.

cameel:
#13027 has a fix for the failing external tests.

ekpyron (dismissed):

ekpyron (approved):

Sticknmove (commented):
Thanks

Sticknmove:
Thanks everyone

Sticknmove:
Thanks

PR #13023, opened 2022-05-15
Changes to error message for Unicode character in non-unicode string literal

### LINKED ISSUE #12968: Error message that results from using Unicode in string literal is confusing

SamuelOsewa:
Fixes #12968.

Changed error message for Unicode character in non-unicode string literal


liblangutil/Scanner.cpp
```diff
@@ -76,7 +76,7 @@ string to_string(ScannerError _errorCode)
 		case ScannerError::IllegalHexDigit: return "Hexadecimal digit missing or invalid.";
 		case ScannerError::IllegalCommentTerminator: return "Expected multi-line comment-terminator.";
 		case ScannerError::IllegalEscapeSequence: return "Invalid escape sequence.";
-		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
+		case ScannerError::IllegalCharacterInString: return "Invalid character in string. If said character is a unicode, use a unicode string literal.";
```
cameel:
```suggestion
		case ScannerError::IllegalCharacterInString: return "Invalid character in string. If you want to use unicode characters, use a unicode\"...\" string literal.";
```

cameel (changes requested):
Another contributor has already submitted a PR to fix this ~2 weeks before yours: #12977. So I'd be inclined to merge that other one. Still, we got no response there to a request for changes so far so if your PR is ready for merge first, I guess I'd merge yours and close the other one.

So what's needed here:
1) You need to update the message in test expectations (run `isoltest` locally with `--update` option).
2) Like I said in the other PR, it would be best to have a test that triggers this message with a non-unicode character as well to show that it's not just about unicode characters and that the message still fits there.
3) Squash your commits (unless they actually represent separate logical changes rather than simple corrections)
4) Rebase on current `develop` to have the external tests pass.

liblangutil/Scanner.cpp
nishant-sachdeva: @cameel  Regarding your point 2 : Could you give an example of a non-unicode character ? You've mentioned `control characters` in https://github.com/ethereum/solidity/pull/12977#discussion_r874924451 but from what I understand, control characters are also a part of the unicode set.

I've added a test case for the same in `control_character_returned_in_string.sol` . Maybe you could suggest appropriate edits for the same ?

cameel: I do mean control characters. By "Unicode characters" I really meant non-ASCII Unicode characters.

liblangutil/Scanner.cpp
```diff
@@ -76,7 +76,7 @@ string to_string(ScannerError _errorCode)
 		case ScannerError::IllegalHexDigit: return "Hexadecimal digit missing or invalid.";
 		case ScannerError::IllegalCommentTerminator: return "Expected multi-line comment-terminator.";
 		case ScannerError::IllegalEscapeSequence: return "Invalid escape sequence.";
-		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
+		case ScannerError::IllegalCharacterInString: return "Invalid character in string. If you want to use unicode characters, use a unicode\"...\" string literal.";
```
cameel: OK, let's make it more precise:
```suggestion
		case ScannerError::IllegalCharacterInString: return "Invalid character in string. If you want to use non-ASCII characters, use a unicode\"...\" string literal.";
```

test/libsolidity/syntaxTests/string/control_character_returned_in_string.sol
```diff
@@ -0,0 +1,9 @@
+contract test {
+    function f() public pure returns (string memory) {
+        return "ج";
+        // this is not a control character. It's a unicode character that gives the same error
+        // this needs to be replaced with an appropriate control character.
```
cameel: You can easily get some control chars like this:
```bash
echo -e "\x1\x2\x3\x4\x5\x6\x7\x8\x9" > /tmp/control-chars.txt
```

test/libyul/yulSyntaxTests/invalid/unicode_string_direction_override.sol
```diff
@@ -3,4 +3,4 @@
     let s := unicode"underflow ‬";
 }
 // ----
-// ParserError 1465: (35-47): Illegal token: Invalid character in string.
+// ParserError 1465: (35-47): Illegal token: Invalid character in string. If you want to use unicode characters, use a unicode"..." string literal.
```
cameel: The message is now wrong here. Note that this **is** a `unicode` string. We just disallow some BiDi characters.

So we need to make the message more specific or start issuing a different error in this case.

cameel (dismissed):

liblangutil/Scanner.h
```diff
@@ -160,6 +161,8 @@ class Scanner
 	/// Called by the parser during FunctionDefinition parsing to clear the current comment
 	void clearCurrentCommentLiteral() { m_skippedComments[Current].literal.clear(); }
 
+	ScannerKind scannerKind() { return m_kind; }
```
ekpyron:
```suggestion
	ScannerKind scannerKind() const { return m_kind; }
```

test/liblangutil/Scanner.cpp
```diff
@@ -123,7 +123,14 @@ BOOST_AUTO_TEST_CASE(string_nonprintable)
 		if (v == '\n' || v == '\v' || v == '\f' || v == '\r')
 			BOOST_CHECK_EQUAL(scanner.currentError(), ScannerError::IllegalStringEndQuote);
 		else
-			BOOST_CHECK_EQUAL(scanner.currentError(), ScannerError::IllegalCharacterInString);
+		{
+			BOOST_CHECK_EQUAL(
+				scanner.currentError(),
+				(scanner.scannerKind() == ScannerKind::Solidity) ?
```
ekpyron: Does this actually ever run with ``scanner.scannerKind() == ScannerKind::Yul``?

liblangutil/Scanner.cpp
```diff
@@ -76,7 +76,8 @@ string to_string(ScannerError _errorCode)
 		case ScannerError::IllegalHexDigit: return "Hexadecimal digit missing or invalid.";
 		case ScannerError::IllegalCommentTerminator: return "Expected multi-line comment-terminator.";
 		case ScannerError::IllegalEscapeSequence: return "Invalid escape sequence.";
-		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
+		case ScannerError::IllegalCharacterInStringSolMode: return "Invalid character in string. If you are trying to use Unicode characters, use a unicode\"...\" string literal.";
+		case ScannerError::IllegalCharacterInStringYulMode: return "Invalid character in string.";
```
ekpyron:
```suggestion
		case ScannerError::UnicodeCharacterInNonUnicodeString: return "Invalid character in string. If you are trying to use Unicode characters, use a unicode\"...\" string literal.";
		case ScannerError::IllegalCharacterInString: return "Invalid character in string.";
```
Maybe better to choose names according to the intention of the messages rather than the scanner mode?

test/liblangutil/Scanner.cpp
nishant-sachdeva: It does not. I changed it in the tests and they're still running as expected. 
Will remove it in the latest code push.

ekpyron (dismissed):

ekpyron (approved):

PR #13028, opened 2022-05-16
Disallow RETURNDATASIZE and RETURNDATACOPY in inline assembly blocks in pure functions

matheusaaguiar:
(cherry picked from commit f567eb1fb2a24dbd956fe65512d4b9015ba7f5ed)

This PR recreates #12861 on `breaking`, after it was reverted from `develop`  in #13013.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ Breaking changes:
  * Commandline Interface: Remapping targets are not automatically added to allowed paths.
  * Commandline Interface: Assembler mode no longer enables all outputs by default.
  * General: The identifier ``basefee`` is a reserved identifier in Yul for all EVM versions.
+ * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
 
 
 ### 0.8.15 (unreleased)
--- a/test/externalTests/ens.sh
+++ b/test/externalTests/ens.sh
@@ -75,6 +75,9 @@ function ens_test
     # In some cases Hardhat does not detect revert reasons properly via IR.
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2115 gets fixed.
     sed -i "s|it\(('Does not allow wrapping a name you do not own',\)|it.skip\1|g" test/wrapper/NameWrapper.js
+    # Related to disallow returndatasize and returndatacopy in inline assembly blocks in pure functions https://github.com/ethereum/solidity/pull/13028
+    # TODO: Remove this after release 0.9 and ENS updated to support it.
+    sed -i 's/pure/view/' contracts/utils/LowLevelCallUtils.sol
 
     find . -name "*.sol" -exec sed -i -e 's/^\(\s*\)\(assembly\)/\1\/\/\/ @solidity memory-safe-assembly\n\1\2/' '{}' \;
 
--- a/test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
+++ b/test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
@@ -47,8 +47,8 @@ contract C {
             //codecopy(0, 1, 2)
             //pop(extcodesize(0))
             //extcodecopy(0, 1, 2, 3)
-            pop(returndatasize())
-            returndatacopy(0, 1, 2)
+            //pop(returndatasize())
+            //returndatacopy(0, 1, 2)
             //pop(extcodehash(0))
             //pop(create(0, 1, 2))
             //pop(create2(0, 1, 2, 3))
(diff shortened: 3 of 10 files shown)


Changelog.md
```diff
@@ -33,6 +33,7 @@ Bugfixes:
  * SMTChecker: Fix bug when z3 is selected but not available at runtime.
  * SMTChecker: Fix ABI compatibility with z3 >=4.8.16.
  * TypeChecker: Convert parameters of function type to how they would be called for ``abi.encodeCall``.
+ * View Pure Checker: Mark ``returndatasize`` and ``returndatacopy`` as view to disallow them in inline assembly blocks in pure functions.
```
chriseth: This needs to be moved to 0.9.0 now.

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_pure.sol
```diff
@@ -85,6 +85,6 @@ contract C {
 // ====
 // EVMVersion: >=london
 // ----
-// Warning 5740: (94-1755): Unreachable code.
-// Warning 5740: (1768-1780): Unreachable code.
-// Warning 5740: (1823-1832): Unreachable code.
+// Warning 5740: (94-1759): Unreachable code.
+// Warning 5740: (1772-1784): Unreachable code.
+// Warning 5740: (1827-1836): Unreachable code.
```
cameel: There are some weird CI failures. One is about this file but it complains that it's missing these:
```
  TypeError 2527: (1247-1257): Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
  TypeError 2527: (1271-1288): Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
```
Not sure why since the instructions that would trigger it are now commented-out.

cameel:
Two of the CI failures seem to be coming from other PRs. For example `negation.sol`:
```
Expected result:
  TypeError 4907: (97-99): Unary operator - cannot be applied to type uint256: Unary negation is only allowed for signed integers.
Obtained result:
  TypeError 4907: (97-99): Unary operator - cannot be applied to type uint256. Unary negation is only allowed for signed integers.
```

This really weird. The test was added in #12828 and I remember that that PR had a last-minute change to replace the `:` with `.`. But the version with `:` was just changed in place. It was never merged into `develop` or `breaking`. Now this PR behaves as if it had `:`.

What'e ever weirder, I just checked state of `TypeChecker.cpp` as of this PR and there's clearly `.` there:
https://github.com/ethereum/solidity/blob/932fb3ef21cd0e7042f73a0dc915e80f2b37aa05/libsolidity/analysis/TypeChecker.cpp#L1665

So it's as if the binary used to run tests here was not built from the code at the right commit.

cameel:
@matheusaaguiar I updated the description to link to the old PR. If your PR is directly related to some other PRs please try to always link to them. It helps a lot if, for whatever reason, you later need to go back and figure out what actually happened.

Also when github folds part of the title into the description, just put it back in the title. Or shorten it - whichever you prefer.

cameel:
I took a closer look at history and I think that the extra failures are actual errors on `breaking` that probably have not been fixed after the last merge from `develop`.

The `negation.sol` change for example:
- That change in `TypeChecker.cpp` was originally done on `breaking` in #11843. Back then it had `:` in the message.
- Recently the same change was introduced on `develop` in #12828. I thought it was something completely new but now I realized that it must have just been cherry-picked from `breaking`.
    - I suggested a small tweak in that PR: we changed `:` to `.` as a fixup. That's why I was confused as to why `:` is still there. I thought that PR was the only thing changing that code and that we had `:` only in its initial version.
- Later `develop` was merged into `breaking` and there must have been a conflict due to that tweak. I think it must have been resolved by taking `TypeChecker.cpp` from one PR but `negation.sol` from the other. Since such a merge usually triggers a wave of fail notifications, the fact that `breaking` was broken because of that was probably not noticed.

So, in short, these seem to be actual problems on `breaking` and should be fixed in a separate PR to `breaking`.

ekpyron:
Since we said we want to have @wechman fix the test failures due to merging develop into breaking separately, I now merged https://github.com/ethereum/solidity/pull/13041 - but this PR should now be fine again by rebasing and just dropping the test fixes that are unrelated to this PR from here, since they're now already in ``breaking``.

test/externalTests/ens.sh
```diff
@@ -75,6 +75,9 @@ function ens_test
     # In some cases Hardhat does not detect revert reasons properly via IR.
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2115 gets fixed.
     sed -i "s|it\(('Does not allow wrapping a name you do not own',\)|it.skip\1|g" test/wrapper/NameWrapper.js
+    #Related to disallow returndatasize and returndatacopy in inline assembly blocks in pure functions https://github.com/ethereum/solidity/pull/13028
+    #TODO: Remove this after release 0.9 and ENS updated to support it.
```
ekpyron:
```suggestion
    # Related to disallow returndatasize and returndatacopy in inline assembly blocks in pure functions https://github.com/ethereum/solidity/pull/13028
    # TODO: Remove this after release 0.9 and ENS updated to support it.
```
maybe

ekpyron (approved):
Looks good!

test/externalTests/ens.sh
matheusaaguiar: Done.

cameel (approved):
Looks good.

I'm going to merge this despite the failing ext test since it's unrelated and it might take some time before the fix gets merged into `breaking`.

PR #13040, opened 2022-05-18
Do not remove potentially reverting returndatacopy cases.

### LINKED ISSUE #13039: [Optimizer] Does not respect returndatacopy specification
The optimizer incorrectly optimizes out `returndatacopy` calls in the following contracts. Found by fuzzing and inputs from @hrkrshnn 

```
contract C {
  fallback() external
  {
    assembly
    {
      returndatacopy(0, 0, 1)
    }
  }
}
```

and

```
contract C {
  fallback() external
  {
    assembly
    {
      returndatacopy(0, 1, 0)
    }
  }
}
```

From matrix chat

@hrkrshnn quotes from the spec

> This opcode has similar semantics to CALLDATACOPY, but instead of copying data from the call data, it copies data from the return data buffer. Furthermore, accessing the return data buffer beyond its size results in a failure; i.e. if start + length overflows or results in a value larger than RETURNDATASIZE, the current call stops in an out-of-gas condition. In particular, reading 0 bytes from the end of the buffer will read 0 bytes; reading 0 bytes from one-byte out of the buffer causes an exception.

So technically the optimizer should retain the returndatacopy call so that it would run out of gas.

ekpyron:
I'm wondering where and how we should also generally mark this in the semantics information...
Although libevmasm marks it as having side effects and it becomes non-movable and non-removable already...
We could consider the revert behaviour as "write to other state" instead, but that's a bit weird as well...
So maybe special handling in the ``UnusedStoreEliminator`` is just enough...

Fixes https://github.com/ethereum/solidity/issues/13039

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ Compiler Features:
 
 Bugfixes:
  * ABI Encoder: When encoding an empty string coming from storage do not add a superfluous empty slot for data.
+ * Yul Optimizer: Do not remove ``returndatacopy`` in cases in which it might perform out-of-bounds reads that unconditionally revert as out-of-gas. Previously, any ``returndatacopy`` that wrote to memory that was never read from was removed without accounting for the out-of-bounds condition.
 
 
 ### 0.8.14 (2022-05-17)
--- a/libyul/optimiser/UnusedStoreEliminator.cpp
+++ b/libyul/optimiser/UnusedStoreEliminator.cpp
@@ -157,7 +157,25 @@ void UnusedStoreEliminator::visit(Statement const& _statement)
 	yulAssert(isCandidateForRemoval == (isStorageWrite || (!m_ignoreMemory && isMemoryWrite)));
 	if (isCandidateForRemoval)
 	{
-		m_stores[YulString{}].insert({&_statement, State::Undecided});
+		State initialState = State::Undecided;
+		if (*instruction == Instruction::RETURNDATACOPY)
+		{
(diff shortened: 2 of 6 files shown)


chriseth:
To fix this properly, I think returndatacopy should be marked the same way as `revert` (plus writing to memory) and we should override that in the unused store eliminator.

ekpyron:
> To fix this properly, I think returndatacopy should be marked the same way as `revert` (plus writing to memory) and we should override that in the unused store eliminator.

I need to check again, but I think it actually may be just like that... it's marked as "has side effects" in ``libevmasm/Instruction.cpp`` and neither revert nor returndatacopy are handled in any way specially beyond that in ``backends/evm/EVMDialect.cpp``... and just the ``UnusedStoreEliminator`` itself treats everything specifically on instruction basis, mostly ignoring side effects...
But yeah, let me double check...

ekpyron:
But yeah, I guess the ``UnusedStoreEliminator`` could/should reject anything "non-removable" except for white-listed cases.

ekpyron:
Looking through this again: the only way in which ``revert`` is marked differently is that it's marked as *always* reverting, which we obviously can't do for ``returndatacopy``... and *all* the instructions handled in the ``UnusedStoreEliminator`` are not movable/removable... so not sure there's much more to be done here, short of restructuring our side-effects in general (and not sure there's that much call for that based on this issue)...

chriseth (dismissed):

chriseth (approved):

Issue #11602, opened 2021-07-01
[ecrecover] Legacy codegen cleans dirty storage bits on read but Sol->Yul on write itself


bshastry:
```
contract C {
  bytes s0;
  constructor() {
    ecrecover(
        0,
        1,
        bytes32(0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa),
        bytes32(0x1337133713371337133713371337133713371337133713371337133713371337)
    ).code;
    s0=("thisisa33bytearraythisisa33bytear");
  }
  function f() external returns (bytes32 r) {
      assembly {
          mstore(0, s0.slot)
          r := sload(add(keccak256(0, 32), 1))
      }
  }
}
// ====
// compileViaYul: also
// ----
// f() -> 0x7237133713371337133713371337133713371337133713371337133713371337
```

This test checks the value obtained via inline assembly from the storage slot 32 bytes after the start of storage variable `s0`. 

Ideally, this slot should return the last byte of the 33-byte bytes array literal that is being assigned to `s0` (which would be the hex of the ascii value of `r` i.e., `0x72`). However, what is actually returned includes "dirty bits" that seem to be from the last argument to the `ecrecover` call preceeding the storage variable assignment in the constructor.

The semantic test above passes via legacy code gen but fails via Yul. Via Yul the return value from `f()` is `0x7200...00`.

Please note that the legacy codegen does clean up the dirty bits but lazily such that non inline assembly accesses e.g., `r = s0[32];` returns `0x7200...00` as expected i.e., the cleaned up value from the adjacent slot.

bshastry:
I'm not sure if this is a "bug" as such. Perhaps sufficient to document in IR breaking changes docs? We already document the fact that dirty storage bits may be retained by the legacy codegen using the following example. But there the "dirtying" is visible to the auditor's eyes unlike the present example (via a call to ecrecover)

```
    contract C {
        bytes x;
        function f() public returns (uint _r) {
            bytes memory m = "tmp";
            assembly {
                mstore(m, 8)
                mstore(add(m, 32), "deadbeef15dead")
            }
            x = m;
            assembly {
                _r := sload(x.slot)
            }
        }
    }
```

chriseth:
Whether or not the dirtying is visible or not does not change the issue in my opinion.

bshastry:
[...]

PR #13045, opened 2022-05-19
Bytes array to storage bug

ekpyron:
Poor-man's-fix for the last cases in https://github.com/ethereum/solidity/issues/11602

This costs, but it's significantly easier than trying to rewrite the copy routine in low-level-style to account for this...

Blog post draft: https://github.com/ethereum/solidity-blog/pull/183


test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol
```diff
@@ -16,4 +16,4 @@ contract C {
 // ====
 // compileViaYul: false
 // ----
-// f() -> 0x6465616462656566313564656164000000000000000000000000000000000010
+// f() -> 0x6465616462656566000000000000000000000000000000000000000000000010
```
ekpyron, line 19: This test means we were fine with this in the past... did we realize back than that the dirty values can actually be accessed without inline assembly when using ``.push()`` on the array?
Given that are we *still* fine with this?
We could also just disable empty ``.push()`` on bytes instead... or have it actually write a zero value... or at the very least document this :-)...

bshastry: > did we realize back than that the dirty values can actually be accessed without inline assembly when using .push() on the array?

No, not as far as I can recall.

chriseth: I think this was overlooked. But it seems like nobody cared anyway

wechman (commented):
@ekpyron Could you put some light on the root cause?

ekpyron:
@wechman The underlying situation is that we do not clear memory to zero after use, but rather only clear it on new allocations. That can result in the memory bytes *after* the data area of a ``bytes memory`` to contain "dirty", i.e. non-zero, values.

Now when copying ``bytes`` from memory to storage, we copied full 256-bit words in bulk. Consequently, the dirty bytes from memory could end up being copied to storage.

Unused storage, on the other hand, is generally assumed to be zero, so on ``.push`` on such a storage ``bytes`` only increases the length field, but does not explicitly write zero values. So such a ``.push`` on storage ``bytes`` could result in the new element of the ``bytes`` state variable to be *non-zero*, even though the docs are very clear on that it is supposed to be zero.

wechman (commented):
@ekpyron Thank you for the explanation! I am aware of the rule to keep unused storage in a zero state. I just wonder why we cannot use [code that do the storage cleanup](https://github.com/ethereum/solidity/blob/7a8ac072ba0783e4af1aa6fa22701fcc48a833fd/libsolidity/codegen/ArrayUtils.cpp#L272). Because we cannot, right? Is it because when we copy bytes from memory to storage we copy full 256-bit words so after that the storage reference is  set not on data end but next slot?

ekpyron:
@wechman Yeah, the current ``copyArrayToStorage`` routine copies ``bytes`` just as if it was ``uint256[]`` - that kind of works, since the copy loop increments and checks slots/offsets during the loop and not element indices (at least that's how I remember it right now from looking at it last month). It would probably be possible to adjust the routine to do something similar as ``m_context.utilFunctions().copyByteArrayToStorageFunction``, i.e. first copy the bulk of the ``bytes`` array as 256-bit words and then copy the remaining single bytes separately... but I'm not sure it's worth the effort messing with the low-level routine for that...

libsolidity/codegen/ArrayUtils.cpp
```diff
@@ -49,9 +49,8 @@ void ArrayUtils::copyArrayToStorage(ArrayType const& _targetType, ArrayType cons
 	// stack layout: [source_ref] [source length] target_ref (top)
 	solAssert(_targetType.location() == DataLocation::Storage, "");
 
-	Type const* uint256 = TypeProvider::uint256();
-	Type const* targetBaseType = _targetType.isByteArrayOrString() ? uint256 : _targetType.baseType();
-	Type const* sourceBaseType = _sourceType.isByteArrayOrString() ? uint256 : _sourceType.baseType();
```
ekpyron, line 55: @wechman Here, this original piece of code basically results in copying ``bytes`` as ``uint256[]``... so the logic below is no longer properly aware of the possibility of a tail that ought to be cleared or anything like that... to properly account for this, we'd probably basically have to rewrite the entire function specially for the ``bytes`` case instead of reusing the general array code - so we might as well just call the yul implementation... even though it may be slightly more expensive, but we want people to move to Yul IR codegen anyways, so I wouldn't spend too much effort on making this optimal for legacy codegen.

ekpyron:
Just rebased this. We should check the external test benchmarks - but judging from the ``isoltest`` gas reports, the difference in optimized legacy gas cost is not actually that bad - it's mainly some increases in code size and, of course, unoptimized code suffers a lot...

ekpyron:
### `ir-no-optimize`
|          project |  bytecode_size | deployment_gas | method_gas |
|:----------------:|---------------:|---------------:|-----------:|
|           bleeps |                |                |            |
|            brink |                |                |            |
|           colony | **`+0.13% ❌`** |                |            |
|        elementfi |                |                |            |
|              ens |                |                |            |
|            euler |                |                |            |
|           gnosis |                |                |            |
|  perpetual-pools |                |                |            |
|    pool-together |                |                |            |
|         prb-math |                |                |            |
|          trident |                |                |            |
|          uniswap |                |                |            |
| yield_liquidator |                |                |            |
|         zeppelin |                |                |            |

### `ir-optimize-evm+yul`
|          project |  bytecode_size | deployment_gas | method_gas |
|:----------------:|---------------:|---------------:|-----------:|
|           bleeps |           `0%` |           `0%` |       `0%` |
|            brink |           `0%` |                |            |
|           colony | **`+0.03% ❌`** |                |            |
|        elementfi |           `0%` |                |            |
|              ens |           `0%` |          `-0%` |       `0%` |
|            euler |           `0%` |          `-0%` |      `-0%` |
|           gnosis |           `0%` |          `+0%` |      `+0%` |
|  perpetual-pools |           `0%` |          `-0%` |      `-0%` |
|    pool-together |           `0%` |          `+0%` |      `+0%` |
|         prb-math |           `0%` |           `0%` |       `0%` |
|          trident |           `0%` |          `-0%` |      `-0%` |
|          uniswap |           `0%` |          `+0%` |      `-0%` |
| yield_liquidator |           `0%` |          `+0%` |       `0%` |
|         zeppelin |           `0%` |          `-0%` |      `+0%` |

### `ir-optimize-evm-only`
|:----------------:|--------------:|---------------:|-----------:|
|           bleeps |               |                |            |
|            brink |               |                |            |
|           colony | **`+0.1% ❌`** |                |            |
|        elementfi |               |                |            |
|              ens |               |                |            |
|            euler |               |                |            |
|           gnosis |               |                |            |
|  perpetual-pools |               |                |            |
|    pool-together |               |                |            |
|         prb-math |               |                |            |
|          trident |               |                |            |
|          uniswap |               |                |            |
| yield_liquidator |               |                |            |
|         zeppelin |               |                |            |

### `legacy-no-optimize`
|            brink | **`+3.14% ❌`** |                |                |
|           colony | **`+0.13% ❌`** |                |                |
|        elementfi | **`+1.97% ❌`** |                |                |
|            euler | **`+0.32% ❌`** | **`+0.22% ❌`** |          `+0%` |
|           gnosis |  **`+1.9% ❌`** | **`+2.73% ❌`** | **`+0.03% ❌`** |
|  perpetual-pools | **`+2.09% ❌`** |  **`+0.9% ❌`** | **`+0.06% ❌`** |
|    pool-together | **`+1.54% ❌`** | **`+0.08% ❌`** |          `+0%` |
|          trident | **`+0.46% ❌`** | **`+0.05% ❌`** |          `+0%` |
|          uniswap |           `0%` |          `+0%` |          `+0%` |
| yield_liquidator |  **`+0.9% ❌`** | **`+0.02% ❌`** |           `0%` |
|         zeppelin | **`+2.91% ❌`** | **`+0.78% ❌`** | **`+0.01% ❌`** |

### `legacy-optimize-evm+yul`
|           bleeps | **`+0.37% ❌`** |           `0%` |           `0%` |
|            brink | **`+0.98% ❌`** |                |                |
|           colony | **`+0.03% ❌`** |                |                |
|        elementfi | **`+0.54% ❌`** |                |                |
|              ens |           `0%` | **`+0.41% ❌`** |          `-0%` |
|            euler | **`+0.11% ❌`** | **`+0.07% ❌`** |          `+0%` |
|           gnosis | **`+0.87% ❌`** | **`+1.01% ❌`** | **`-0.02% ✅`** |
|  perpetual-pools | **`+0.88% ❌`** | **`+0.39% ❌`** | **`+0.02% ❌`** |
|    pool-together | **`+0.59% ❌`** | **`+0.03% ❌`** |          `+0%` |
|          trident | **`+0.17% ❌`** | **`+0.02% ❌`** |          `-0%` |
|          uniswap |           `0%` |          `-0%` |          `-0%` |
| yield_liquidator | **`+0.25% ❌`** | **`-0.05% ✅`** |           `0%` |
|         zeppelin | **`+1.16% ❌`** | **`+0.32% ❌`** |          `+0%` |

|            brink | **`+1.79% ❌`** |                |                |
|           colony |  **`+0.1% ❌`** |                |                |
|        elementfi | **`+1.12% ❌`** |                |                |
|              ens |           `0%` | **`+0.67% ❌`** |          `-0%` |
|            euler | **`+0.19% ❌`** | **`+0.12% ❌`** |          `+0%` |
|           gnosis | **`+1.31% ❌`** | **`+1.48% ❌`** |          `+0%` |
|  perpetual-pools |  **`+1.4% ❌`** | **`+0.58% ❌`** | **`-0.01% ✅`** |
|    pool-together | **`+1.04% ❌`** | **`+0.05% ❌`** |          `-0%` |
|          trident | **`+0.29% ❌`** | **`+0.03% ❌`** |          `-0%` |
|          uniswap |           `0%` |          `-0%` |          `+0%` |
| yield_liquidator | **`+0.64% ❌`** | **`-0.02% ✅`** |           `0%` |
|         zeppelin | **`+1.95% ❌`** | **`+0.52% ❌`** | **`-0.05% ✅`** |


`!V` = version mismatch

test/libsolidity/semanticTests/byte_array_to_storage_cleanup.sol
```diff
@@ -0,0 +1,37 @@
+contract C {
+  event ev0(uint[] i0, uint);
+  bytes public s2;
+  function h() external returns (bytes memory) {
```
cameel: Bad indent :)
```suggestion
    event ev0(uint[] i0, uint);
    bytes public s2;
    function h() external returns (bytes memory) {
```

test/libsolidity/semanticTests/byte_array_to_storage_cleanup.sol
```diff
@@ -0,0 +1,37 @@
+contract C {
+  event ev0(uint[] i0, uint);
```
cameel: Why is one named and the other is not?
```suggestion
  event ev0(uint[], uint);
```

test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol
cameel: > or have it actually write a zero value...

Actually, why not this? It seems like a more natural solution to me. For one thing - it would then affect just empty pushes to `bytes` arrays so it would be less of a performance hit - especially given that these seem rare. For another - are we sure that there are no other situations where dirty bytes can be copied? For example how about ABI encoding with result being assigned directly to storage?

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "uid": "SOL-2022-5",
+        "name": "DirtyBytesArrayToStorage",
+        "summary": "Copying ``bytes`` arrays from memory or calldata to storage may result in dirty storage values.",
+        "description": "Copying ``bytes`` arrays from memory or calldata to storage is done in chunks of 32 bytes. Thereby, dirty values in calldata or memory can be written to storage, which may then become observable after a ``.push()`` on the bytes array in storage.",
```
cameel: I think we could make it a little more detailed. Especially I'm not sure if "dirty values" on its own is self-explanatory to everyone. How about this version?

```suggestion
        "description": "Copying ``bytes`` arrays from memory or calldata to storage is done in chunks of 32 bytes even if the length is not a multiple of 32. Thereby, extra bytes past the end of the array may be copied from calldata or memory to storage. These dirty bytes were assumed not to be observable without the use of inline assembly. However, a ``.push()`` without arguments to a ``bytes`` array assumes that the storage is already zeroed and only increments the array length, which may result in dirty bytes becoming a part of the array.",
```

```suggestion
        "description": "Copying ``bytes`` arrays from memory or calldata to storage is done in chunks of 32 bytes. Thereby, dirty values in calldata or memory can be written to storage, which may then become observable after a ``.push()`` on the bytes array in storage. This bug affects only the legacy code generation pipeline. The IR-based pipeline always cleans dirty bytes when copying array elements.",
```

test/libsolidity/semanticTests/byte_array_to_storage_cleanup.sol
ekpyron: I really should start watching that :-).

test/libsolidity/semanticTests/byte_array_to_storage_cleanup.sol
ekpyron: Directly took it from the issue, but I can remove the name.

test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol
ekpyron: I don't think empty ``push``es to bytes are really *that* rare :-). Merely ones that don't write anything eventually anyways... and actually writing zeroes is a huge hit for those cases, unless the write can be optimized away...
And it'd be weird to write zeroes here all of a sudden, but in no other cases...

docs/bugs.json
ekpyron: I wonder if we should start using ``"viaIR": true`` and ``"viaIR": false`` in the conditions field...

ekpyron: I went for something in between for now.

cameel: Yeah, sounds like a good idea.

ekpyron: Actually ``docs/bugs.rst`` doesn't document even the ``yulOptimizer`` one so far...
Thinking about it, I'm also not sure if we should change that to plain ``optimizer: true`` for the other bug as - since you don't *explicitly* need to enable the yul optimizer anymore these days...

test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol
cameel: But I thought we write zeros in all other cases? If we don't then why is it a problem only with `bytes` arrays.

docs/bugs.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "uid": "SOL-2022-5",
+        "name": "DirtyBytesArrayToStorage",
+        "summary": "Copying ``bytes`` arrays from memory or calldata to storage may result in dirty storage values.",
+        "description": "Copying ``bytes`` arrays from memory or calldata to storage is done in chunks of 32 bytes even if the length is not a multiple of 32. Thereby, extra bytes past the end of the array may be copied from calldata or memory to storage. These dirty bytes may then become observable after a ``.push()`` without arguments to the bytes array in storage, i.e. such a push will not result in a zero value at the end of the array as expected. This bug only affects the legacy code generation pipeline, the new code generation pipeline via IR is not affected.",
```
cameel, line 6:
```suggestion
        "description": "Copying ``bytes`` arrays from memory or calldata to storage is done in chunks of 32 bytes even if the length is not a multiple of 32. Thereby, extra bytes past the end of the array may be copied from calldata or memory to storage. These dirty bytes may then become observable after a ``.push()`` without arguments to the ``bytes`` array in storage, i.e. such a push will not result in a zero value at the end of the array as expected. This bug only affects the legacy code generation pipeline, the new code generation pipeline via IR is not affected.",
```

cameel (approved):
For now I still think we might be better off just making `.push()` write a zero (https://github.com/ethereum/solidity/pull/13045#discussion_r896038667) but maybe I just have not yet noticed all the downsides :)

In any case, it's probably a bit late to change that. In case we decide to stay with the current solution, this PR seems ready.

docs/bugs.json
cameel: Does `optimizer: true` mean "happens only with optimizer" or "happens with optimizer too"?

cameel: Also, about the new flag, maybe @axic would like to chime in.

ekpyron: > Does `optimizer: true` mean "happens only with optimizer" or "happens with optimizer too"?

See ``docs/bugs.rst`` (even though it's apparently incomplete): 
```
conditions
    Conditions that have to be met to trigger the bug. The following
    keys can be used:
    ``optimizer``, Boolean value which
    means that the optimizer has to be switched on to enable the bug.
    ``evmVersion``, a string that indicates which EVM version compiler
    settings trigger the bug. The string can contain comparison
    operators. For example, ``">=constantinople"`` means that the bug
    is present when the EVM version is set to ``constantinople`` or
    later.
    If no conditions are given, assume that the bug is present.
```

I.e. it means happens *only* with optimizer.

test/libsolidity/semanticTests/array/copying/dirty_memory_bytes_to_storage_copy.sol
ekpyron: We never write zeroes :-). Storage is just always zeroed. For packed arrays writing an explicit zero is an ``sload`` and an ``store`` for each element (or at least each slot, but only if the optimizer works well on it) - that's among the most costly things you can do :-).

What we do is that we write zeroes on *deallocation*, i.e. on ``delete`` and ``.pop()`` and the like, s.t. we *can* assume storage to always be zero.

The bug is that copies from calldata/memory to storage do *not* respect the general principle that storage never contains dirty values, thereby violating the assumption that all unused storage is zero at all times.

ekpyron:
> For now I still think we might be better off just making `.push()` write a zero ([#13045 (comment)](https://github.com/ethereum/solidity/pull/13045#discussion_r896038667)) but maybe I just have not yet noticed all the downsides :)
> 
> In any case, it's probably a bit late to change that. In case we decide to stay with the current solution, this PR seems ready.

Just commented again on that - we definitely don't want the ``.push()`` to write a zero, so I'm merging as is :-).

PR #13047, opened 2022-05-20
Workaround for euler ext test failing due to changed chainid behavior

matheusaaguiar:
Patches issue https://github.com/euler-xyz/euler-contracts/issues/119.
After issue is resolved, this should be reverted/changed.

### DIFF
--- a/test/externalTests/euler.sh
+++ b/test/externalTests/euler.sh
@@ -66,6 +66,10 @@ function euler_test
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2453 gets fixed.
     sed -i "/expectError: 'JUNK_UPGRADE_TEST_FAILURE'/d" test/moduleUpgrade.js
 
+    # One of the tests is currently broken in the upstream, due to changed behavior of chainid in @ethereumjs/vm.
+    # TODO: Remove this when https://github.com/euler-xyz/euler-contracts/issues/119 gets fixed.
+    sed -i 's/domainSeparatorMatch)\.to\.equal(true/domainSeparatorMatch)\.to\.equal(false/g' test/eulStakes.js
+
     neutralize_package_lock
     neutralize_package_json_hooks
     force_hardhat_compiler_binary "$config_file" "$BINARY_TYPE" "$BINARY_PATH"


test/externalTests/euler.sh
```diff
@@ -66,6 +66,10 @@ function euler_test
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2453 gets fixed.
     sed -i "/expectError: 'JUNK_UPGRADE_TEST_FAILURE'/d" test/moduleUpgrade.js
 
+    # Workaround to make euler test pass
+    # TODO: Remove this when https://github.com/euler-xyz/euler-contracts/issues/119 gets fixed.
+    sed -i 's/domainSeparatorMatch).to.equal(true/domainSeparatorMatch).to.equal(false/g' test/eulStakes.js
```
cameel: You need to escape `.` to match actual dots rather than any character:
```suggestion
    sed -i 's/domainSeparatorMatch)\.to\.equal(true/domainSeparatorMatch)\.to\.equal(false/g' test/eulStakes.js
```
Probably won't make a difference here but still, better to keep it correct.

test/externalTests/euler.sh
```diff
@@ -66,6 +66,10 @@ function euler_test
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2453 gets fixed.
     sed -i "/expectError: 'JUNK_UPGRADE_TEST_FAILURE'/d" test/moduleUpgrade.js
 
+    # Workaround to make euler test pass
```
cameel: I'd include a bit more info here. All of these patches are workarounds so it's good to have a quick way to discern them without having to visit the URL :)
```suggestion
    # One of the tests is currently broken in the upstream, due to changed behavior of chainid in @ethereumjs/vm.
```

cameel (commented):
Only needs some cosmetic tweaks and we can merge it.

Also, I'd avoid including issue numbers in git commits (even those from external repos). It spams the issue with a mention every time you push code after rebasing.

chriseth:
This blocks tests, so I'll make the changes and merge.

chriseth (approved):

Issue #12934, opened 2022-04-13
More concise "Cheatsheet" page


cameel:
## Page
- [Cheatsheet](https://docs.soliditylang.org/en/latest/cheatsheet.html#global-variables)
- [Units and Globally Available Variables](https://docs.soliditylang.org/en/latest/units-and-global-variables.html)

## Abstract
Currently the "Cheatsheet" page mostly contains information already available in "Units and Globally Available Variables". The information is already concise in that original source so the cheatsheet just looks like an inferior copy. Also, having two copies is just an extra maintenance burden that's not worth it unless the cheatsheet actually adds something of value. I think we can make it more valuable by adding summaries of information from other pages and streamlining the way it's presented where possible.

Here's what I think we should do:
- Remove excessive detail from the cheatsheet. If the info is only available there, move it to a different location.
    - [x] Notes/warnings should not be there
    - [ ] (Optional) "Global Variables" section would be better off as a table. This may require tweaking the descriptions or presenting signatures in a more compact way.
        - Marking this as optional since we tried it and it did not look good (#13595). I still think it's possible but will require some editing of the content to make it actually suitable for a table.
    - [x] Members of `address`, `type`, etc. should be in a separate section, not under "Global Variables".
    - [x] The operator precedence should be under [Types > Operators](https://docs.soliditylang.org/en/latest/types.html#operators). Maybe it makes sense to have a copy here but that's also extra maintenance so I'd just move it.
    - [x] I think that the list of keywords should be moved to "Units and Globally Available Variables". It does not fit that page perfectly but is a bit related.
- Add missing info
    - [ ] Members available on functions, errors, events, calldata/storage variables, arrays, etc. (`slot`, `offset`, `length`, `address`, `selector`, ...).
        - These should be added to "Units and Globally Available Variables" as well.
    - [ ] Make sure that the list of "Global Variables" contains every item from "Units and Globally Available Variables"

SamuelOsewa:
I created a pull request with the notes section removed from the cheat sheet. Was that done correctly? @cameel

victorknox:
@cameel I have created a PR for moving the keywords, hope it's okay
I also noticed that separating members address, type etc. would make it very similar to "Units and Globally Variables"
would it be okay to proceed?

cameel:
[...]

PR #13048, opened 2022-05-21
Move operator precedence from cheatsheet to operators

shadyy41:
#12934 
Moved the operator precedence section from the cheatsheet to types > operators.


chriseth:
Sorry, but I fear it does not work this way. You moved it to the integer type, but operator precedence is unrelated to the type.

shadyy41:
> Sorry, but I fear it does not work this way. You moved it to the integer type, but operator precedence is unrelated to the type.

#12934 
Point 4 here says it should be moved from the cheatsheet to Types > Operators, did I move it to the wrong place? Please correct me, I'm new.

cameel:
> Sorry, but I fear it does not work this way. You moved it to the integer type, but operator precedence is unrelated to the type.

It's not nested inside integer type docs. It's under "Operators", which is directly under "Types".

docs/cheatsheet.rst
```diff
@@ -2,71 +2,6 @@
 Cheatsheet
 **********
 
-.. index:: precedence
-
-.. _order:
-
-Order of Precedence of Operators
-================================
```
cameel: Now that I think of it, we could leave a link to it here. Can you leave the section here, with just a link inside?

docs/types/operators.rst
```diff
@@ -108,3 +108,68 @@ value it referred to previously.
             assert(y.length == 0);
         }
     }
+
+.. index:: precedence
```
cameel: While at it, I'd improve the index entry. Let's make it the main entry for this topic and nest it under `operators`.

```suggestion
.. index:: ! operators; precedence
```

docs/cheatsheet.rst
```diff
@@ -4,68 +4,10 @@ Cheatsheet
 
 .. index:: precedence
```
cameel: I'd either remove this or at least use the same index entry as in the other place (without `!` though since it's not the main entry on the topic).
```suggestion
.. index:: operators; precedence
```

cameel:
The failing CI job is unrelated to this PR and has a fix in `develop`. Please rebase.

cameel:
Something weird happened here. I think you did a merge but there are your commits on both sides of it. Github shows tons of unrelated commits in your PR for that reason.

We don't merge `develop` into PRs to update them since that makes history very messy. We prefer a rebase workflow. I push a cleaned up version here in a moment.

cameel (dismissed):

cameel (approved):

PR #13049, opened 2022-05-21
nit: ABI docs use canonical representation

josepot:
Hi!

First and foremost: thanks a lot for these amazing docs! They truly are outstandingly good!

However, there is something really small that's been bothering me a bit and I've decided to open a PR: IMO the "selector-like" signatures should be consistent and always use the canonical type representations.

I mean, IMHO I think that it's totally ok to display solidity code examples that uses the shorthand type definitions, ofc. However, I _think_ that when displaying "selector-like" function definitions, then the docs should stick with the canonical type representations.

I'm aware that this is a super small thing. So, I totally get it if you rather not to consider this change.

### DIFF
--- a/docs/abi-spec.rst
+++ b/docs/abi-spec.rst
@@ -308,7 +308,7 @@ In total:
 Use of Dynamic Types
 ====================
 
-A call to a function with the signature ``f(uint,uint32[],bytes10,bytes)`` with values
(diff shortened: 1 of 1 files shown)


hrkrshnn (approved):
Thanks!

hrkrshnn:
Need to wait for #13047 to make the test pass :(

PR #13063, opened 2022-05-25
External test workarounds for Gnosis and Euler external test failures

cameel:
A fresh batch of external test workarounds:

### Gnosis
Job [1072860](https://app.circleci.com/pipelines/github/ethereum/solidity/24386/workflows/1d379105-8a89-41e6-88d3-442c5339827d/jobs/1072860) fails at `hardhat compile` step:
```
An unexpected error occurred:

Error: Debug Failure. False expression: Non-string value passed to `ts.resolveTypeReferenceDirective`, likely by a wrapping package working with an outdated `resolveTypeReferenceDirectives` signature. This is probably not a problem in TS itself.
```

Some kind of incompatibility with latest [typescript package](https://www.npmjs.com/package/typescript?activeTab=versions). Typescript 4.7.2 was released today. Until now the project was using 4.6.4 (there were only pre-releases in 4.7.1 and 4.7.0 series). My workaround is to cap it at `< 4.7.0`.

### Euler
Job [1072861](https://app.circleci.com/pipelines/github/ethereum/solidity/24386/workflows/1d379105-8a89-41e6-88d3-442c5339827d/jobs/1072861) has 3 failing tests with this error message:
```
337 passing (9m)
  3 failing

  1) batch operations
       sub-account transfers:
     Error: VM Exception while processing transaction: reverted with custom error 'BatchDispatchSimulation(<error parameters>)'
        ...
```

This is somehow related to changes in ethers.js. We earlier capped ethers.js for this external test at 5.6.1 in #13027. Removing this older workaround now fixes the problem.

While at it I also removed this workaround from all other external tests except for Gnosis and Uniswap. In #13051 @matheusaaguiar determined that Gnosis is the case that still fails with 5.6.8 and for me Uniswap also fails here without the workaround (see [1073153](https://app.circleci.com/pipelines/github/ethereum/solidity/24397/workflows/555dd7a6-bd3f-4164-b7ef-49646687da25/jobs/1073153)).

Issue #13057, opened 2022-05-24
Switch from deprecated `circleci/buildpack-deps` to another Ubuntu image for our Dockerfiles


cameel:
CircleCI has deprecated a bunch of docker images, including `circleci/buildpack-deps` (see [Legacy Convenience Image Deprecation](https://discuss.circleci.com/t/legacy-convenience-image-deprecation)). As of 2022 these images are still available but are not receiving any updates:
> Dec 31, 2021 - Legacy Convenience Images are no longer supported. This will be the last publishing day of these images. Existing tags will still exist on Docker Hub but will remain unchanged. There will be no support for bug or security fixes for existing tags. No new tags will be published after this day.

We rely on this image in [Dockerfile.ubuntu2004](https://github.com/ethereum/solidity/blob/develop/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004) and [Dockerfile.ubuntu2004.clang](https://github.com/ethereum/solidity/blob/develop/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004.clang). We also [use it directly as a base image](https://github.com/ethereum/solidity/blob/v0.8.14/.circleci/config.yml#L327-L341) for some CI jobs.

The CI jobs are easy to deal with and I'm already working on it as a part of a general CI update to apply feedback we received from CircleCI. I managed to replace the image there with `cimg/base:current`.

Docker images are a bigger problem because there we want a specific Ubuntu version and `cimg/base` does not have tags for that. It's based on latest LTS instead so by using it we'd no longer be in control of when to update. We should find another image that's suitable as a base, maybe just a plain `ubuntu` image. This will require some adjustments in the dockerfile though.

I'm creating an issue instead of doing it right away because it's best to do this together with the update to Ubuntu 22.04. While we're still on focal, we can keep using the old image.

cameel:
Actually, there's nothing more to do here. Removal of `circleci/buildpack-deps` from CI config in #13068 was all that was needed. The [`buildpack-deps` image that we use in docker images](https://hub.docker.com/_/buildpack-deps) is actually a separate thing and that still appears to be maintained. I guess `circleci/buildpack-deps` must have been just CircleCI's fork of it and only that was dropped?

Issue #13056, opened 2022-05-24
Parallelize bytecode report generation in CI


cameel:
The scripts we use for bytecode report generation in CI are single-threaded and do not take advantage of multiple cores we have available there. They could finish much faster if parallelized. This is suboptimal especially on macOS and Windows where the compiler build takes almost 20 minutes and these jobs cannot start before then.

This is the timing from [one of the recent CI runs on `develop`](https://app.circleci.com/pipelines/github/ethereum/solidity/23305/workflows/ea43cc31-b20f-4a9f-aafa-f639ef7688aa):

| job              | duration | cores |
|------------------|----------|-------|
| `b_bytecode_ubu` |  8m 47s  | 1     |
| `b_bytecode_ems` |  6m 11s  | 1     |
| `b_bytecode_win` | 11m 01s  | 4     |
| `b_bytecode_osx` | 13m 14s  | 4     |

There are two ways to address this:
1) Use multiple threads or processes in [`prepare_report.py`](https://github.com/ethereum/solidity/blob/develop/scripts/bytecodecompare/prepare_report.py) and [`prepare_report.js`](https://github.com/ethereum/solidity/blob/develop/scripts/bytecodecompare/prepare_report.js).
2) Split the set of input files into smaller batches, run separately on each batch and then merge the reports.

I think that option (2) is better given that with (1) we'd have to implement the same mechanism twice, once in JS and once in Python.

Once the scripts are parallelized, first run enough threads/processes in parallel to saturate the available cores. Then try switching to bigger machines with more cores and see if that gives us an even bigger boost relative to the cost. Finally, also try the `parallelism` option.

PR #13068, opened 2022-05-25
CI config tweaks after CircleCI review

cameel:
~Depends on #13063.~ Merged.

This is a bunch of CI tweaks addressing some of the [feeback we received in CircleCI's config review](https://gist.github.com/jenny-miggin/9578d2e52b383b3787e265c80e5f6b03). Here's an overview of what I'm changing:
1) I think that `b_win` job is superfluous given that we have `b_win_release` as well. They pretty much always fail or succeed together and for some reason `b_win` is actually sometimes not showing all compiler errors in its output and `b_win_release` does not have this problem. We're doing just fine with one job for macOS so I think we can do with one for Windows. macOS and Windows jobs are also the ones we're paying the most for so removing this is is going to be a big saving.
    - ~While at it I noticed that `b_osx` is actually running a release config, just like `b_win_release` so I also renamed it to `b_osx_release`.~ **EDIT 2022-10-24**: I misunderstood that. It was called `b_win_release` because it's built with `FORCE_RELEASE=ON`, not because it's a release build. All our macOS and Windows builds are release builds. So instead I'm now leaving it as is and renaming `b_ubu_release` to `b_ubu_force_release` to make this distinction clearer.
    - **2022-10-24**: I'm now removing `b_win_release` and leaving `b_win` in instead. We can't use `b_win_release` for bytecode comparison since it won't match version string of other, non-force-release binaries.
2) CircleCI has deprecated some of the images we're using (see [Legacy Convenience Image Deprecation](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034)) so I'm switching to alternatives:
    - `cimg/node` instead `circleci/node`. The downside here is that the tags for major node.js versions like 12, 14, 16, etc. are no longer available and it doesn't looks like they'll come back anytime soon (https://github.com/CircleCI-Public/cimg-node/issues/130). ~For now I had to specify it down to patch version.~ And we'll have to remember about keeping them up to date now. **EDIT** It's actually possible to specify it only down to minor version which improves the situation a little.
    - `cimg/base` instead of `circleci/buildpack-deps`. In some cases we were using specifically `circleci/buildpack-deps:focal` and this tag is not available for `cimg/base` but this was for jobs like pylint or grammar so I don't think this matters.
        - This actually revealed a small problem with our bytecode report scripts. Looks like the new image has a different default locale, which affects the sort order for Bash globs and for Python's `sorted()`. This affects cases like `input_sol.sol` vs `in_sol.sol` and makes the comparison fail. To fix this I explicitly set locale to `C` whenever we use these scripts.

There are two more things that require more work and are not urgent so I only created issues for them:
- #13057
- #13056

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -735,7 +735,7 @@ jobs:
           name: Install Java
           command: |
             sudo apt -q update
-            sudo apt install -y openjdk-16-jdk
+            sudo apt install -y openjdk-17-jdk
       - run:
           name: Run tests
           command: ./scripts/test_antlr_grammar.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -356,17 +356,9 @@ defaults:
         TERM: xterm
         MAKEFLAGS: -j 10
 
-  - base_buildpack_focal_small: &base_buildpack_focal_small
+  - base_cimg_small: &base_cimg_small
       docker:
-        - image: buildpack-deps:focal
-      resource_class: small
-      environment:
-        TERM: xterm
-        MAKEFLAGS: -j 2
-
-  - base_buildpack_latest_small: &base_buildpack_latest_small
-      docker:
-        - image: buildpack-deps:latest
+        - image: cimg/base:current
       resource_class: small
       environment:
         TERM: xterm
@@ -681,12 +673,14 @@ jobs:
       - gitter_notify_failure_unless_pr
 
   chk_coding_style:
-    <<: *base_buildpack_focal_small
+    <<: *base_cimg_small
     steps:
       - checkout
       - run:
           name: Install shellcheck
-          command: apt -q update && apt install -y shellcheck
+          command: |
+            sudo apt -q update
+            sudo apt install -y shellcheck
       - run:
           name: Check for C++ coding style
           command: ./scripts/check_style.sh
@@ -708,12 +702,14 @@ jobs:
       - gitter_notify_failure_unless_pr
 
   chk_pylint:
-    <<: *base_buildpack_focal_small
+    <<: *base_cimg_small
     steps:
       - checkout
       - run:
           name: Install pip
-          command: apt -q update && apt install -y python3-pip
+          command: |
+            sudo apt -q update
+            sudo apt install -y python3-pip
       - run:
           name: Install pylint and dependencies of the scripts that will be linted
           command: python3 -m pip install
@@ -732,12 +728,14 @@ jobs:
       - gitter_notify_failure_unless_pr
 
   chk_antlr_grammar:
-    <<: *base_buildpack_focal_small
+    <<: *base_cimg_small
     steps:
       - checkout
       - run:
           name: Install Java
-          command: apt -q update && apt install -y openjdk-16-jdk
+          command: |
+            sudo apt -q update
+            sudo apt install -y openjdk-16-jdk
       - run:
           name: Run tests
           command: ./scripts/test_antlr_grammar.sh
@@ -759,14 +757,14 @@ jobs:
       - gitter_notify_failure_unless_pr
 
   chk_proofs:
-    <<: *base_buildpack_latest_small
+    <<: *base_cimg_small
     steps:
       - checkout
       - run:
           name: Z3 python deps
           command: |
-            apt-get -qq update
-            apt-get -qy install python3-pip
+            sudo apt-get -qq update
+            sudo apt-get -qy install python3-pip
             pip3 install --user z3-solver
       - run: *run_proofs
       - gitter_notify_failure_unless_pr
@@ -1405,6 +1403,10 @@ jobs:
 
   b_bytecode_ubu:
     <<: *base_ubuntu2004_small
+    environment:
+      TERM: xterm
+      MAKEFLAGS: -j 2
+      LC_ALL: C
     steps:
       - checkout
       - attach_workspace:
@@ -1414,6 +1416,10 @@ jobs:
 
   b_bytecode_osx:
     <<: *base_osx
+    environment:
+      TERM: xterm
+      MAKEFLAGS: -j 5
+      LC_ALL: C
     steps:
       - checkout
       - attach_workspace:
@@ -1449,6 +1455,7 @@ jobs:
     <<: *base_node_small
     environment:
       SOLC_EMSCRIPTEN: "On"
+      LC_ALL: C
     steps:
       - checkout
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -527,11 +527,6 @@ defaults:
       requires:
         - b_win
 
-  - workflow_win_release: &workflow_win_release
-      <<: *workflow_trigger_on_tags
-      requires:
-        - b_win_release
-
   # --------------------------------------------------------------------------
   # Parameterized Job Templates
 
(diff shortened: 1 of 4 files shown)


cameel:
Looks like I also need to set the locale on Windows somehow...

wechman:
PR changes look fine and reflect what I can find in the description. In regard to the node version, can we maybe have a script that provides warning whenever a new release is available?

nikola-matic:
I rebased and made it worse :muscle:

.circleci/config.yml
```diff
@@ -742,12 +736,14 @@ jobs:
...
-    <<: *base_buildpack_focal_small
+    <<: *base_cimg_small
     steps:
       - checkout
       - run:
           name: Install Java
-          command: apt -q update && apt install -y openjdk-16-jdk
+          command: |
+            sudo apt -q update
+            sudo apt install -y openjdk-16-jdk
```
nikola-matic:
```
E: Unable to locate package openjdk-16-jdk
```
Not sure why this worked before, but apparently `openjdk-16` is [not a part of any Debian release](https://tracker.debian.org/pkg/openjdk-16); I get the same error when trying to install it locally. Can we bump this to `openjdk-17`?

.circleci/soltest.sh
```diff
@@ -92,7 +92,7 @@ do
 done
 
 # wait for individual processes to get their exit status
-for pid in ${PIDs[*]}
+for pid in "${PIDs[*]}"
```
nikola-matic: Without quotes (original)
```
SC2048 (warning): Use "${array[@]}" (with quotes) to prevent whitespace problems.
```
With quotes (current)
```
SC2066 (error): Since you double quoted this, it will not word split, and the loop will only run once.
```
What's the correct approach to this, and is there actually one other than disabling this check? Also, having looked [shellcheck docs](https://www.shellcheck.net/wiki/Ignore), it seems like an error can only be disable on a per-file basis, and not on a per line basis (more fine grained), which I feel is overkill.

nikola-matic (commented):
I see bytecodecompare has different compiler versions, which I guess is the source of failure (one of, haven't checked in detail whether there's anything else).

.circleci/soltest.sh
cameel: This must be either `${PIDs[*]}` (as in the original) or `"${PIDs[@]}"`. The latter will correctly handle array items with spaces in them, though, given that these are process IDs, we don't really expect any spaces in them and the former worked just fine. In both cases every PID will be a separate item for the loop.

`"${PIDs[*]}"` is incorrect because it will join the array into a single string (with spaces in it) and will be seen as a single item by the loop.

The gotcha here is simply that the `"${x[@]}"` syntax is counter-intuitive Bash magic. I.e. it has quotes around it so it looks as if the result was a single quoted thing but in fact it will evaluate to multiple separate quoted items.

For example with `PIDs=(1 23 "456 789")`:
- `${PIDs[*]}` evaluates to `1`, `23`, `456`, `789`
- `"${PIDs[@]}"` evaluates to `1`, `23`, `456 789`
- `"${PIDs[*]}"` evaluates to `1 23 456 789`

.circleci/config.yml
cameel: It did work and yeah, if version 17 works, it's good.

.circleci/soltest.sh
cameel: Fixed.

.circleci/config.yml
cameel: Switched to version 17.

cameel:
This is now ready. All jobs pass except for the hardhat one, which is broken for unrelated reasons. Note that there are a few jobs that look like they're still running but that's because the were renamed/removed in the PR while they are still marked as required on github. Once the PR is approved, I'll adjust repo config to fix this, just before this is merged.

One important point to consider when reviewing this: are we fine with dropping one of the windows builds?

cameel:
> I see bytecodecompare has different compiler versions, which I guess is the source of failure (one of, haven't checked in detail whether there's anything else).

It was failing because I removed one of the windows build jobs and switched the bytecode comparison it to use the other - which was running with `FORCE_RELEASE=ON`. That flag makes it identify as a release, which affects the version string - so version looked different than on other platforms. I fixed that by switching to removing the other windows job.

nikola-matic:
> This is now ready. All jobs pass except for the hardhat one, which is broken for unrelated reasons. Note that there are a few jobs that look like they're still running but that's because the were renamed/removed in the PR while they are still marked as required on github. Once the PR is approved, I'll adjust repo config to fix this, just before this is merged.
> 
> One important point to consider when reviewing this: are we fine with dropping one of the windows builds?

Can you explain in more detail why you dropped `b_win_release` (I saw the description about version difference, but still don't understand)? Also, what does `FORCE_RELEASE` do if the builds are already in release mode?

cameel:
The primary motivation for dropping it was that it was one of the recommendations from [CircleCI's config review](https://gist.github.com/jenny-miggin/9578d2e52b383b3787e265c80e5f6b03):

> The `b_win` and `b_win_release` jobs are largely duplicated, only with an additional environment variable included in the latter. The same occurs for the `t_win` and `t_win_soltest` jobs. Consider optimising this further, by merging these jobs into a more efficient flow, and also perhaps including a [`when`](https://support.circleci.com/hc/en-us/articles/360043638052-Conditional-steps-in-jobs-and-conditional-workflows) clause to run only one workflow depending on whether the `FORCE_RELEASE:` environment variable is set or not, or by only running one of these on the `main` branch.

Windows and macOS jobs are pretty expensive and the difference between `b_win` and `b_win_release` is very small. At some point we considered moving one of them to the nightly run but really, but I'm not sure if such a build gives us all that much if it's not running on every PR.

I don't really remember any case where one of them failed and the other did not. At most, I noticed on multiple occasions that the build log from `b_win_release` is missing some error messages while they do show up in `b_win`. Not sure why it happens and we should just fix it but so far we never had much motivation to spend time on it.

> Also, what does `FORCE_RELEASE` do if the builds are already in release mode?

When we build a release we include an empty `prerelease.txt` file in repo root. Setting `FORCE_RELEASE=ON` does the same thing. Based on that our CMake config sets a different version string (it no longer says it's a development build) and the C++ code has access to that via a preprocessor macro. In this mode the compiler also no longer emits a warning that it's a pre-release version. There are probably some minor things it can affect in the build, but I'd have to look for them to tell. One you might be able to notice is if you break stripping of the pre-release warning and CI starts failing for only one of these builds. Other differences are unlikely to be very noticeable, it's just that it's too late to fix them when we're doing an actual release. We want to catch them earlier.

We still have a Linux build in this mode and it might be enough. For some reason we never had a macOS or emscripten build like this.

.circleci/config.yml
```diff
@@ -403,6 +395,11 @@ defaults:
         name: win/default
         shell: cmd.exe
```
r0qs: You replaced `base_win_cmd` with `base_win_bash` in the `b_bytecode_win` and `base_win_cmd` is not being used anymore. Can we remove it?

r0qs (changes requested):
LGTM. Just one comment.

.circleci/config.yml
cameel: Good point. I thought it was still in use by other jobs but apparently not. Removed.

r0qs (approved):

nikola-matic (approved):

PR #13072, opened 2022-05-26
Added option to allow soltest in CI print errors to stderr in addition to the XML report.

### LINKED ISSUE #13066: soltest in CI should print errors to stderr in addition to the XML report
[`soltest.sh`](https://github.com/ethereum/solidity/blob/develop/.circleci/soltest.sh) passes `--logger=JUNIT` option to soltest, which makes it produce an XML report that CircleCI can then analyze to list failed test cases on the TESTS tab (see [CircleCI docs > Collecting test data](https://circleci.com/docs/2.0/collect-test-data/)). Currently using this option means that errors are no longer printed to stderr, which makes analyzing the output harder. Especially combined with bugs like #12862, which break processing of the XML report (and we actually still haven't fixed that completlely, it happens in some cases).

The task here is to make it print both to stderr and to the XML file, possibly with an extra option (though I think it would be fine as the default behavior too).

soltest is based on [Boost Test](https://www.boost.org/doc/libs/1_79_0/libs/test/doc/html/index.html) which, I think, does not have options for printing to both out of the box. Please check that first though - if it does, solving this could be as simple as specifying an extra option in `soltest.sh`. Otherwise it will have to be implemented manually in C++.

matheusaaguiar:
Resolves #13066

### DIFF
--- a/.circleci/soltest.sh
+++ b/.circleci/soltest.sh
@@ -75,6 +75,7 @@ do
         "--color_output=no"
         "--show_progress=yes"
         "--logger=JUNIT,error,test_results/$(get_logfile_basename "$((CPUs * CIRCLE_NODE_INDEX + run))").xml"
+        "--logger=HRF,error,stdout"
         "${BOOST_TEST_ARGS[@]}"
     )
     SOLTEST_ARGS=("--evm-version=$EVM" "${SOLTEST_FLAGS[@]}")
--- a/.circleci/soltest.ps1
+++ b/.circleci/soltest.ps1
@@ -6,7 +6,7 @@ cd "$PSScriptRoot\.."
 if ( -not $? ) { throw "Cannot execute solc --version." }
 
 mkdir test_results
-.\build\test\Release\soltest.exe --color_output=no --show_progress=yes --logger=JUNIT,error,test_results/result.xml -- --no-smt
+.\build\test\Release\soltest.exe --color_output=no --show_progress=yes --logger=JUNIT,error,test_results/result.xml --logger=HRF,error,stdout -- --no-smt
 if ( -not $? ) { throw "Unoptimized soltest run failed." }
-.\build\test\Release\soltest.exe --color_output=no --show_progress=yes --logger=JUNIT,error,test_results/result_opt.xml -- --optimize --no-smt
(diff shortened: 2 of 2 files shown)


.circleci/soltest.sh
```diff
@@ -75,6 +75,7 @@ do
         "--color_output=no"
         "--show_progress=yes"
         "--logger=JUNIT,error,test_results/$(get_logfile_basename "$((CPUs * CIRCLE_NODE_INDEX + run))").xml"
+        "--logger=HRF,error,stderr"
```
cameel: Can you add this also in `soltest.ps1`? And do a test in CI by introducing some artificial error to make sure it behaves as expected on Windows.

cameel (commented):
Looks great! This solves the problem nicely. Once you add this for Windows too, we can merge it.

.circleci/soltest.sh
cameel: Regarding https://github.com/ethereum/solidity/issues/13066#issuecomment-1138951346, I think these should go to stdout because they're part of the normal output of the command, not problems in `soltest` itself:
```suggestion
        "--logger=HRF,error,stdout"
```

But I'd also be fine with printing them to stderr if you can think of a good use case that would benefit from having them there rather on stdout.

matheusaaguiar: Yes, I will be doing that.

matheusaaguiar: I set stderr for no special reason and can't think of anything to prevent it being changed to stdout... I am going to change it as you suggested.

cameel (commented):
Looks like it works.

The failed test run for later reference: [f7ddb246-899e-43d5-b83a-8c6799cdb41f](https://app.circleci.com/pipelines/github/ethereum/solidity/24429/workflows/f7ddb246-899e-43d5-b83a-8c6799cdb41f) (I see that the artificial error has already been removed and now tests pass).

Please just squash your commits. I.e. remove the artificial error and the revert - we don't need that in history. The first two could stay but it's really something I'd put into a single commit since it's very closely related so I'd recommend squashing them into one.

matheusaaguiar:
Yes, sure. I explicitly let those commits stay for you to check the history :)
I will squash everything now.

cameel (approved):
Ah, ok then, thanks! Approved.

PR #13075, opened 2022-05-27
 Print settings options such evmVersion and optimize in isoltest/soltest failed test logs.

### LINKED ISSUE #13065: More detailed info about failed semantic tests in soltest
Currently, when a semantic test fails, soltest/isoltest does not print any information about options it was executed with. In particular the EVM version used (`--evm-version` option) and the optimier status (`--optimize` option). This information should be included both in the console output for each failed test and in the XML file for CircleCI that it produces when `--logger=JUNIT` option is used.

matheusaaguiar:
Resolves #13065.

### DIFF
--- a/test/Common.cpp
+++ b/test/Common.cpp
@@ -20,14 +20,19 @@
 #include <iostream>
 #include <test/Common.h>
 #include <test/EVMHost.h>
+#include <test/libsolidity/util/SoltestErrors.h>
 
 #include <libsolutil/Assertions.h>
+#include <libsolutil/StringUtils.h>
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
+#include <range/v3/all.hpp>
 
 namespace fs = boost::filesystem;
 namespace po = boost::program_options;
+
+using namespace std;
 
 namespace solidity::test
 {
@@ -207,6 +212,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
 	return true;
 }
 
+string CommonOptions::toString(vector<string> const& _selectedOptions) const
+{
+	if (_selectedOptions.empty())
+		return "";
+
+	auto boolToString = [](bool _value) -> string { return _value ? "true" : "false"; };
+	// Using std::map to avoid if-else/switch-case block
+	map<string, string> optionValueMap = {
+		{"evmVersion", evmVersion().name()},
+		{"optimize", boolToString(optimize)},
+		{"useABIEncoderV1", boolToString(useABIEncoderV1)},
+		{"batch", to_string(selectedBatch + 1) + "/" + to_string(batches)},
+		{"ewasm", boolToString(ewasm)},
+		{"enforceCompileToEwasm", boolToString(enforceCompileToEwasm)},
+		{"enforceGasTest", boolToString(enforceGasTest)},
+		{"enforceGasTestMinValue", enforceGasTestMinValue.str()},
+		{"disableSemanticTests", boolToString(disableSemanticTests)},
+		{"disableSMT", boolToString(disableSMT)},
+		{"showMessages", boolToString(showMessages)},
+		{"showMetadata", boolToString(showMetadata)}
+	};
+
+	soltestAssert(ranges::all_of(_selectedOptions, [&optionValueMap](string const& _option) { return optionValueMap.count(_option) > 0; }));
+
+	vector<string> optionsWithValues = _selectedOptions |
+		ranges::views::transform([&optionValueMap](string const& _option) { return _option + "=" + optionValueMap.at(_option); }) |
+		ranges::to<vector>();
+
+	return solidity::util::joinHumanReadable(optionsWithValues);
+}
+
+void CommonOptions::printSelectedOptions(ostream& _stream, string const& _linePrefix, vector<string> const& _selectedOptions) const
+{
+	_stream << _linePrefix << "Run Settings: " << toString(_selectedOptions) << endl;
+}
 
 langutil::EVMVersion CommonOptions::evmVersion() const
 {
--- a/test/Common.h
+++ b/test/Common.h
@@ -81,6 +81,12 @@ struct CommonOptions
 	// Throws a ConfigException on error
 	virtual void validate() const;
 
+	/// @returns string with a key=value list of the options separated by comma
+	/// Ex.: "evmVersion=london, optimize=true, useABIEncoderV1=false"
+	virtual std::string toString(std::vector<std::string> const& _selectedOptions) const;
+	/// Helper to print the value of settings used
+	virtual void printSelectedOptions(std::ostream& _stream, std::string const& _linePrefix, std::vector<std::string> const& _selectedOptions) const;
+
 	static CommonOptions const& get();
 	static void setSingleton(std::unique_ptr<CommonOptions const>&& _instance);
 
--- a/test/libsolidity/SemanticTest.cpp
+++ b/test/libsolidity/SemanticTest.cpp
@@ -315,6 +315,14 @@ TestCase::TestResult SemanticTest::run(ostream& _stream, string const& _linePref
 				throw;
 		}
 	}
+
+	if (result != TestResult::Success)
+		solidity::test::CommonOptions::get().printSelectedOptions(
+			_stream,
+			_linePrefix,
+			{"evmVersion", "optimize", "useABIEncoderV1", "batch"}
+		);
+
 	return result;
 }
 


test/tools/isoltest.cpp
```diff
@@ -146,6 +152,16 @@ class TestTool
 
 bool TestTool::m_exitRequested = false;
 
+void TestTool::printOptions(ostream& _stream, string const& _linePrefix, bool) const
+{
+	_stream << _linePrefix << "// ====" << endl;
+	_stream << _linePrefix << "// evmVersion: " << m_options.evmVersion().name() << endl;
+	_stream << _linePrefix << "// optimize: " << m_options.optimize << endl;
+	_stream << _linePrefix << "// useABIEncoderV1: " << m_options.useABIEncoderV1 << endl;
```
cameel: This will print `0` or `1`, right? `true`/`false` would be better.

test/tools/isoltest.cpp
```diff
@@ -146,6 +152,16 @@ class TestTool
...
+	_stream << _linePrefix << "// batches: " << m_options.batches << endl;
+	_stream << _linePrefix << "// selectedBatch: " << m_options.selectedBatch << endl;
```
cameel: The isoltest output is already pretty verbose since it prints the whole source. I'd try to make this more compact.

For example I'd make this look something like: `batch: 3/125` instead having each part on a separate line.

BTW, what are these set to when not using the `--batches` option? Is this `1`/`1` or zeros or something else?

cameel: This format makes it look like the expectations you can add in the test files but you cannot use it there. I guess we can do it like this too but I was really thinking about something more compact since these values are just extra information that you do not always need. They should not be some prominent.

Maybe it would be better to have them on a single line? I.e. something like:
```
Run settings: EVM=london, optimize, batch=3/125
```
with boolean values like `optimize` and `abiencoderv1` just being omitted when false.

Or, alternatively:
```
Run settings: evmVersion=london, optimize=true, abiencoderv1=false, batch=3/125
```

test/tools/isoltest.cpp
```diff
@@ -178,6 +194,7 @@ TestTool::Result TestTool::process()
 						AnsiColorized(cout, formatted, {BOLD, CYAN}) << "  Contract:" << endl;
 						m_test->printSource(cout, "    ", formatted);
 						m_test->printSettings(cout, "    ", formatted);
+						printOptions(cout, "    ", formatted);
```
cameel: Can you make sure that `isoltest` won't add this to the rewritten test file when the test fails and it offers to update the expectations for you and you press `u` to accept?

test/tools/isoltest.cpp
```diff
@@ -146,6 +152,16 @@ class TestTool
 
 bool TestTool::m_exitRequested = false;
 
+void TestTool::printOptions(ostream& _stream, string const& _linePrefix, bool) const
```
cameel: Why include `_formatted` if you're not using it? It does not look like you're overriding this function.
```suggestion
void TestTool::printOptions(ostream& _stream, string const& _linePrefix) const
```

test/tools/isoltest.cpp
matheusaaguiar: I guess it would output something like `0/1` if not using --batches. That's what I got when running isoltest locally with `./test/tools/isoltest --enforce-gas-cost --optimize --no-smt`:
```
...
 // ====
    // evmVersion: london
    // optimize: 1
    // useABIEncoderV1: 0
    // batches: 1
    // selectedBatch: 0
```

test/tools/isoltest.cpp
matheusaaguiar: ops, forgot to delete that ...

test/tools/isoltest.cpp
matheusaaguiar: I would prefer the second form because you leave no room for doubt about the omitted/false values...

test/tools/isoltest.cpp
matheusaaguiar: Sorry, I didn't quite understand what you meant here ... Do you mean preventing someone in the future doing this? Or do you mean making sure that it can't be reached from inside method `updateCase` ?

test/tools/isoltest.cpp
cameel: I think we should add 1 to batch number in that case. I.e. so that's `1/1` by default and so that it's normally from `1/n` to `n/n` rather than from `0/n` to `n-1/n`.

cameel: Sure, I'm fine with either.

test/tools/isoltest.cpp
cameel: Yes, I mean `updateTestCase()`.

Have you seen what happens when you run `isoltest` and the expectations do not match? `soltest` just fails, but `isoltest` stops and asks you what to do. One of the options is to rewrite the test case with new expectations. I just wanted you to make sure this part is only printed on the screen but not included in the rewritten file (because it would fail then).

test/tools/isoltest.cpp
```diff
@@ -146,6 +151,15 @@ class TestTool
 
 bool TestTool::m_exitRequested = false;
 
+void TestTool::printOptions(ostream& _stream, string const& _linePrefix) const
+{
+	_stream << _linePrefix << "// Run settings:" <<
+				" EVM=" << m_options.evmVersion().name() <<
+				", optimize=" << (m_options.optimize ? "true" : "false") <<
+				", useABIEncoderV1=" << (m_options.useABIEncoderV1 ? "true" : "false") <<
+				", batch=" << (m_options.selectedBatch+1) << "/"  << m_options.batches << endl;
```
cameel:
```suggestion
	_stream << _linePrefix << "// Run settings:" <<
		" EVM=" << m_options.evmVersion().name() <<
		", optimize=" << (m_options.optimize ? "true" : "false") <<
		", useABIEncoderV1=" << (m_options.useABIEncoderV1 ? "true" : "false") <<
		", batch=" << (m_options.selectedBatch+1) << "/"  << m_options.batches << endl;
```

cameel:
```suggestion
	_stream << _linePrefix << "// Run settings:" <<
				" EVM=" << m_options.evmVersion().name() <<
				", optimize=" << (m_options.optimize ? "true" : "false") <<
				", useABIEncoderV1=" << (m_options.useABIEncoderV1 ? "true" : "false") <<
				", batch=" << (m_options.selectedBatch + 1) << "/" << m_options.batches << endl;
```

test/tools/isoltest.cpp
```diff
@@ -115,6 +115,11 @@ class TestTool
 
 	Result process();
 
+	void printOptions(
+		ostream& _stream,
+		string const& _linePrefix = ""
+	) const;
```
cameel: This is short enough to be on a single line now.
```suggestion
	void printOptions(ostream& _stream, string const& _linePrefix = "") const;
```

cameel (commented):
Just minor style fixes left.

If the thing with expectations I mentioned earlier works fine, it should be good to go.

As the final test though, could you add some errors to expectations so that we can see how the output looks like now in CI? That was the primary motivation for this after all. Would be good to see how it looks when multiple processes are printing output simultaneously. You can just add a commit with an error, let it do a complete run and then remove that commit and just post a link to the failed run.

EDIT: Also, I guess it would be best to rebase the PR on #13072 first so that the output actually gets captured (unless we just merge it before then).

test/tools/isoltest.cpp
matheusaaguiar: Ah, ok!
Yes, I am sure it won't update the files. I have tested locally, including updating a failed test (both with --accept-updates and manually pressing `u` when prompted).

cameel: ok, sounds good then.

cameel:
I just merged #13072 so when you're done you can simply rebase on `develop`.

matheusaaguiar:
Ok.

Failed test run: [5b537588-6425-40b3-92e5-05a5414aff8a](https://app.circleci.com/pipelines/github/ethereum/solidity/24435/workflows/5b537588-6425-40b3-92e5-05a5414aff8a).

Rebasing and removing errors now.

cameel:
> Failed test run: [5b537588-6425-40b3-92e5-05a5414aff8a](https://app.circleci.com/pipelines/github/ethereum/solidity/24435/workflows/5b537588-6425-40b3-92e5-05a5414aff8a).

Looks like something's wrong. I can't see the `Run settings` output in any of the failed runs.

matheusaaguiar:
I am investigating.

matheusaaguiar:
Changed the way the options information is added to test failure messages. Added it to the end of the `outputMessage` returned by the test and colored it yellow.I think it will show in CI now.

matheusaaguiar:
I figured out what happened. I only looked at isoltest and forgot about soltest. So the changes I made were only affecting the former. So here's what I did: I implemented method `toString()` and helper function `printOptions` in `CommonOptions`.  I think this is a good way to be able to access the options values in both testers, isoltest and soltest, or any other test that uses `CommonOptions`.

matheusaaguiar:
Now the options are showing in CI failed tests also: [076cf593-a2f5-4e95-9609-7aa537b52855](https://app.circleci.com/pipelines/github/ethereum/solidity/24466/workflows/076cf593-a2f5-4e95-9609-7aa537b52855)

test/Common.cpp
```diff
@@ -207,6 +207,74 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+{
+	std::stringstream options;
+	if(!m_singleton)
+		return "";
+
+	for(std::string& opt : _filter)
+	{
+		switch (options) {
+			case "evmVersion":
+				options << "evmVersion=" << evmVersion().name() << " ";
```
matheusaaguiar, line 220: I think adding a functionality like this could be helpful in the future if we want to print other options... But I don't know if this is a probable scenario or if it is worth the effort... some inputs about this would be appreciated.

matheusaaguiar:
Rebased on develop.
I am thinking that maybe in the future we could want to print other options. Is it worth to implement a way of specifying what option values are desired in the `printOptions` functions ? Maybe use std::initializer_list or a STL container like std::vector?
The thing that is bugging me is that now it is just hardcoded  to some options...

test/Common.cpp
```diff
@@ -207,6 +207,38 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+		{"enforceCompileToEwasm", "enforceCompileToEwasm=" + boolToString(enforceCompileToEwasm)},
+		{"enforceGasTest", "enforceGasTest=" + boolToString(enforceGasTest)},
+		{"enforceGasTestMinValue", "enforceGasTestMinValue=" + enforceGasTestMinValue.str()},
+		{"disableSemanticTests", "disableSemanticTests=" + boolToString(disableSemanticTests)},
+		{"disableSMT", "disableSMT=" + boolToString(disableSMT)},
+		{"showMessages", "showMessages=" + boolToString(showMessages)},
+		{"showMetadata", "showMetadata=" + boolToString(showMetadata)}
+	};
+
+	for(auto optionName : _filter)
```
cameel: Our style is to always but a space betwene `if` and `(` or `for` and `(`.
```suggestion
	for (auto optionName: _filter)
```

test/Common.cpp
```diff
@@ -207,6 +207,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
 }
 
+std::string CommonOptions::toString(const std::vector<std::string>& _filter) const
+{
+	std::string options;
+	if(!m_singleton || _filter.empty())
+		return "";
+
+	//Using std::map to avoid if-else/switch-case block
+	std::map<std::string, std::string> option_value_map = {
```
cameel:
```suggestion
	std::map<std::string, std::string> optionValueMap = {
```

test/Common.cpp
```diff
@@ -207,6 +207,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+		{"evmVersion", "evmVersion=" + evmVersion().name()},
+		{"optimize", "optimize=" + boolToString(optimize)},
+		{"useABIEncoderV1", "useABIEncoderV1=" + boolToString(useABIEncoderV1)},
+		{"batch", "batch=" + std::to_string(selectedBatch+1) + "/" + std::to_string(batches)},
```
cameel:
```suggestion
		{"batch", "batch=" + std::to_string(selectedBatch + 1) + "/" + std::to_string(batches)},
```

test/Common.cpp
```diff
@@ -207,6 +207,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+	for(const std::string& optionName : _filter)
+	{
+		auto opt = option_value_map.find(optionName);
+		if(opt != option_value_map.end())
+			options.append(opt->second + ", ");
+	}
+
+	//erase extra ', '
+	if(!options.empty())
+		options.erase(options.size()-2);
```
cameel: You should use ranges v3 and our `joinHumanReadable()` here.

test/Common.cpp
```diff
@@ -207,6 +207,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
 	return true;
 }
 
+std::string CommonOptions::toString(std::vector<std::string> const& _filter) const
```
cameel: Let's just add `using namespace std;` and drop the `std::` prefixes. We do this everywhere.

test/Common.cpp
```diff
@@ -207,6 +207,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+{
+	if(!m_singleton || _filter.empty())
+		return "";
+
+	//Using std::map to avoid if-else/switch-case block
```
cameel:
```suggestion
	// Using std::map to avoid if-else/switch-case block
```

test/Common.cpp
```diff
@@ -277,4 +312,14 @@ bool loadVMs(CommonOptions const& _options)
 	return true;
 }
 
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options, std::vector<std::string> const& _filter)
+{
+	_stream << _linePrefix << "Run Settings: " << _options.toString(_filter) << std::endl;
+}
+
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options)
```
cameel: It would make more sense to call this something like `printImportantOptions()`. Without the argument I'd actually expect it to print everyting, not some random subset.
```suggestion
void printImportantOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options)
```

test/Common.h
```diff
@@ -102,6 +107,13 @@ bool isValidSemanticTestPath(boost::filesystem::path const& _testPath);
 
 bool loadVMs(CommonOptions const& _options);
 
+///Helper function to print the value of settings used (CommonOptions struct)
+///@p _filter should be the sequence of option names to be printed.
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options, std::vector<std::string> const& _filter);
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options);
+
+inline std::string const boolToString(bool b) { return b ? "true" : "false"; }
```
cameel: I'd either make this a lambda for local use in `toString()` or put it in [`CommonData.h`](https://github.com/ethereum/solidity/blob/develop/libsolutil/CommonData.h) if you really think it makes sense to make it generic.

cameel:
```suggestion
inline std::string const boolToString(bool value) { return value ? "true" : "false"; }
```

test/libsolidity/semanticTests/strings/empty_storage_string.sol
```diff
@@ -84,9 +84,9 @@ contract C {
 // g() -> 0x40, 0x60, 0, 0
 // h() -> 0x40, 0x60, 0, 0x1a, 38178759162904981154304545770567765692299154484752076569098748838215919075328
 // i() -> 0x40, 0x80, 0x1a, 38178759162904981154304545770567765692299154484752076569098748838215919075328, 0
-// j(string): 0x20, 0, "" -> 0x20, 0
+// j(string): 0x80, 0, "" -> FAILURE
 // k() -> 0x20, 0
-// l(string): 0x20, 0, "" -> 0x20, 0x40, 0x20, 0
+// l(string): 0x40, 0, "" -> 0x20, 0x40, 0x20, 0
```
cameel: This looks unrelated to this PR. Why is it here?

test/Common.h
```diff
@@ -102,6 +107,13 @@ bool isValidSemanticTestPath(boost::filesystem::path const& _testPath);
 
 bool loadVMs(CommonOptions const& _options);
 
+///Helper function to print the value of settings used (CommonOptions struct)
+///@p _filter should be the sequence of option names to be printed.
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options, std::vector<std::string> const& _filter);
```
cameel: Maybe `_selectedOptions` would be more self-explanatory than just `_filter`?
```suggestion
void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options, std::vector<std::string> const& _selectedOptions);
```

test/Common.h
```diff
@@ -102,6 +107,13 @@ bool isValidSemanticTestPath(boost::filesystem::path const& _testPath);
...
+void printOptions(std::ostream& _stream, std::string const& _linePrefix, CommonOptions const& _options);
```
cameel: These take `CommonOptions` as an argument. It seems to me like they should just be its methods rather than standalone functions.

cameel:
> I am thinking that maybe in the future we could want to print other options. 

Maybe, but I'd avoid overgeneralizing it. Adding a new option would not have been a major effort even with the code you originally had here. But since you have already written it, we can just keep the general version now :)

> Is it worth to implement a way of specifying what option values are desired in the printOptions functions ? Maybe use std::initializer_list or a STL container like std::vector?
> The thing that is bugging me is that now it is just hardcoded to some options...

A vector seems fine. If we wanted it to be really clean, I'd use an enum instead of strings for specifying options. But like I said, not much point extending it too much. What you have now is fine.

I'd keep only the version of `printOptions()` that takes the list of options and I'd hard-code the options we want at the point of call. After all the ones we want are specific to our use case (i.e. debugging a semantic test). If you want to avoid had-coding some options in two places maybe just create a constant for the default vector.

test/boostTest.cpp
```diff
@@ -108,6 +107,7 @@ void runTestCase(TestCase::Config const& _config, TestCase::TestCaseCreator cons
 				case TestCase::TestResult::Success:
 					break;
 				case TestCase::TestResult::Failure:
+					solidity::test::printOptions(errorStream, "", solidity::test::CommonOptions::get());
```
cameel: Are you sure that printing here will make these options show up only for semantic tests? That's the only situation in which the test result may be affected by them so it does not make much sense to print them otherwise.

test/libsolidity/semanticTests/strings/empty_storage_string.sol
matheusaaguiar: I guess it has to do with local tests and I must have overlooked when staging and committing. I am going to restore to the correct version.

test/Common.cpp
matheusaaguiar: I have changed it to `printSelectedOptions`.

test/Common.h
matheusaaguiar: I guess a lambda is what I was looking for since the beginning, so I'll go with it.

test/boostTest.cpp
matheusaaguiar: You are right, I lost sight of that important detail and was thinking about all tests. I have already moved it to only print in failed semantic tests.

matheusaaguiar:
I have implemented all your suggestions @cameel.

test/Common.cpp
```diff
@@ -207,6 +211,45 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
 
+string CommonOptions::toString(vector<string> const& _selectedOptions) const
+{
+	if (!m_singleton || _selectedOptions.empty())
+		return "";
+
+	auto boolToString = [](bool value) -> string
+		{
+			return value ? "true" : "false";
+		};
```
cameel:
```suggestion
	auto boolToString = [](bool value) -> string
	{
		return value ? "true" : "false";
	};
```

But it's short enough that I'd just put it on a single line.

test/Common.cpp
```diff
@@ -207,6 +211,45 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+		{"showMessages", "showMessages=" + boolToString(showMessages)},
+		{"showMetadata", "showMetadata=" + boolToString(showMetadata)}
+	};
+
+	using namespace ranges;
+	vector<string> options = _selectedOptions |
+		views::transform([&optionValueMap](auto& optionName) { return optionValueMap.find(optionName); }) |
+		views::filter([&optionValueMap](auto it) { return it != optionValueMap.end(); }) |
+		views::transform([](auto it) { return it->second; }) |
+		to<vector>();
```
cameel: 1) I just noticed that you just ignore options that are not present in `optionValueMap`. I think we should require them to be present instead. Option names are not user input so we do not need to handle them gracefully and in practice, when you make a typo in the name it's better to get an error than for it to be silently skipped.
2) Since the `option=` part is always identical to the key, why not remove it from the map and add it here? Less duplication and the map would be more readable.

```suggestion
	soltestAssert(views::set_difference(_selectedOptions, optionValueMap | views::keys).empty());

	vector<string> optionsWithValues = _selectedOptions |
		views::transform([&optionValueMap](string const& _option) { return _option + "=" + optionValueMap.at(_option); }) |
		to<vector>();
```

test/Common.cpp
```diff
@@ -207,6 +211,45 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+
+	return solidity::util::joinHumanReadable(options);
+}
+
+void CommonOptions::printSelectedOptions(std::ostream& _stream, std::string const& _linePrefix, std::vector<std::string> const& _selectedOptions) const
+{
+	_stream << _linePrefix << "Run Settings: " << toString(_selectedOptions) << std::endl;
+}
```
cameel:
```suggestion
void CommonOptions::printSelectedOptions(ostream& _stream, string const& _linePrefix, vector<string> const& _selectedOptions) const
{
	_stream << _linePrefix << "Run Settings: " << toString(_selectedOptions) << endl;
}
```

test/Common.h
```diff
@@ -81,6 +81,12 @@ struct CommonOptions
 	// Throws a ConfigException on error
 	virtual void validate() const;
 
+	// @returns string with a key=value list of the options separated by comma
+	// Ex.: "evmVersion=london, optimitze=true, useABIEncoderV1=false"
+	virtual std::string toString(std::vector<std::string> const& _selectedOptions) const;
+	///Helper to print the value of settings used
```
cameel:
```suggestion
	/// @returns string with a key=value list of the options separated by comma
	/// Ex.: "evmVersion=london, optimize=true, useABIEncoderV1=false"
	virtual std::string toString(std::vector<std::string> const& _selectedOptions) const;
	/// Helper to print the value of settings used
```

test/Common.cpp
matheusaaguiar: 1. Yes, it really makes sense to do so. I did not think about erroneous values at the time I came up with that code.
2. I should have done it like that since the beginning.
Thanks for the excellent review and suggestions!

Jessejames1180 (approved):

Jessejames1180 (dismissed):

test/Common.cpp
matheusaaguiar: I could not use `set_differences` because `_selectedOptions` should be sorted prior to that, but then it is a `const` reference. So I used `all_of` instead...

test/Common.cpp
```diff
@@ -207,6 +212,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
...
+		{"enforceCompileToEwasm", boolToString(enforceCompileToEwasm)},
+		{"enforceGasTest", boolToString(enforceGasTest)},
+		{"enforceGasTestMinValue", enforceGasTestMinValue.str()},
+		{"disableSemanticTests", boolToString(disableSemanticTests)},
+		{"disableSMT", boolToString(disableSMT)},
+		{"showMessages", boolToString(showMessages)},
+		{"showMetadata", boolToString(showMetadata)}
+	};
+
+	soltestAssert(ranges::all_of(_selectedOptions, [&optionValueMap](string const& _option) { return optionValueMap.find(_option) != optionValueMap.end(); }));
```
matheusaaguiar: I could not use `set_differences` because `_selectedOptions` should be sorted prior to that, since the `optionValueMap` keys are sorted. But then, it is a `const` reference,  so I used `all_of` instead...

matheusaaguiar:
Rebased on develop.

test/Common.cpp
```diff
@@ -207,6 +212,41 @@ bool CommonOptions::parse(int argc, char const* const* argv)
 	return true;
 }
 
+string CommonOptions::toString(vector<string> const& _selectedOptions) const
+{
+	if (!m_singleton || _selectedOptions.empty())
+		return "";
```
cameel, line 218: This function operates on the current instance of options and does not access `m_singleton`. Is this condition actually needed?

The ` _selectedOptions.empty()` part is fine but it seems that function would still work fine without it so I'd remove it. It's such a useless case that it's not really worth optimizing for.

test/Common.cpp
cameel: Ah, you probably need `_selectedOptions | to<set>()` for it to work. I usually find the set versions more concise but either is fine really.

You could just use `count()` to shorten it since you don't use the iterator returned by `find()` for anything:
```suggestion
	soltestAssert(ranges::all_of(_selectedOptions, [&optionValueMap](string const& _option) { return optionValueMap.count(_option) > 0; }));
```

cameel (commented):
Two more tweaks and we're done here.

test/Common.cpp
matheusaaguiar: Yes, I should have changed that once I made it become a member of the struct.

matheusaaguiar:
Rebased and addressed suggested changes.

cameel (dismissed):

cameel (dismissed):

cameel (approved):

PR #13082, opened 2022-05-30
Revert "Move operator precedence from cheatsheet to operators"

chriseth:
Reverts ethereum/solidity#13048

I think we need to clean this up in general, but this specific PR made it worse instead of better:

Operators should not be a sub-section of types:
image

And a full heading that just contains a link also looks very bad IMO:
image


chriseth:
We should probably have a new section about the language grammar / structure and that should contain operator precedence.
This is currently mostly covered by "layout of a source file" or "structure of a contract". In the future, those two sections could be combined, since contract are no longer the only top-level elements. Instead, we could maybe split everything into:

- declarations
- statements and control structures
- expressions

Then operator precedence could be part of expressions.

cameel:
> Operators should not be a sub-section of types:

Maybe, but that PR did not change it :) It just moved the table.

> Then operator precedence could be part of expressions.

That sounds good but until we have that, why isn't the "Operators" section a more logical place to find the precedence table than the cheatsheet?

> And a full heading that just contains a link also looks very bad IMO:

Maybe we could just change the heading to something like "Tables" or "Charts and tables"? This is a neat table so it does make sense to have quick access to it from the cheatsheet and I thought a link would be good enough. There are more such tables scattered around the docs. My idea was to eventually have a section with quick links to them in the cheatsheet.

docs/cheatsheet.rst
```diff
@@ -2,12 +2,70 @@
 Cheatsheet
 **********
 
-.. index:: operator; precedence
+.. index:: precedence
```
cameel, line 5: Even if we're ultimately going to revert it, I'd keep the new index entry (and with `!` if we restore the table here).

PR #13083, opened 2022-05-30
lsp.py: Fix various problems with subdirectory edge cases

Marenz:
* fix properly path parsing with nested subdirectories
* fix regex to be able to handle `file://..`
* fix not checking all expected diagnostics
* Some extra comments and renames

### DIFF
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -69,7 +69,7 @@ TEST_REGEXES = TestRegexesTuple(
     re.compile(R'^// -> (?P<method>[\w\/]+) {'),
     re.compile(R'(?P<tag>"@\w+")'),
     re.compile(R'(?P<tag>@\w+)'),
-    re.compile(R'// (?P<testname>\w+):[ ]?(?P<diagnostics>[\w @]*)'),
+    re.compile(R'// (?P<testname>\S+):([ ](?P<diagnostics>[\w @]*))?'),
     re.compile(R'(?P<tag>@\w+) (?P<code>\d\d\d\d)')
 )
 
@@ -86,7 +86,7 @@ def split_path(path):
     """
     Return the test name and the subdir path of the given path.
     """
-    sub_dir_separator = path.find("/")
+    sub_dir_separator = path.rfind("/")
 
     if sub_dir_separator == -1:
         return (path, None)
@@ -105,7 +105,8 @@ def count_index(lines, start=0):
 
 def tags_only(lines, start=0):
     """
-    Filter the lines for tag comments and report line number that tags refer to.
+    Filter the lines for tag comments and report the line number that the tags
+    _refer_ to (which is not the line they are on!).
     """
     n = start
     numCommentLines = 0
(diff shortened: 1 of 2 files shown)


test/lsp.py
```diff
@@ -586,6 +591,12 @@ def test_diagnostics(self):
                         marker=markers[expected_diagnostic.marker]
                     )
 
+            if len(expected_diagnostics_per_file) > 0:
+                raise ExpectationFailed(
+                    f"Expected but received no diagnostics for {expected_diagnostics_per_file}",
```
christianparpart: "Expected but received no diagnostics" sounds rather strange to me, grammer-wise. Did you miss something in this sentence?

test/lsp.py
```diff
@@ -889,22 +900,31 @@ def wait_for_diagnostics(self, solc: JsonRpcProcess) -> List[dict]:
 
         return sorted(reports, key=lambda x: x['uri'])
 
+    def normalizeUri(self, uri, no_suffix = True):
+        relUri = uri.replace(self.project_root_uri + "/", "")
```
christianparpart:
```suggestion
        relativeURI = uri.replace(self.project_root_uri + "/", "")
```

christianparpart (changes requested):
thanks. also the naming changes come quite in handy. much more self-explanatory :)

test/lsp.py
Marenz: "*grammar" your mean :P

 I don't see a problem with the sentence.. "[We] expected[, ]but received no diagnostics for .. "

test/lsp.py
```diff
@@ -889,22 +900,31 @@ def wait_for_diagnostics(self, solc: JsonRpcProcess) -> List[dict]:
 
         return sorted(reports, key=lambda x: x['uri'])
 
+    def normalizeUri(self, uri, no_suffix = True):
```
christianparpart: Second parameter is **never** used so it seems? Did I overlook it?

Marenz: Ah, there was a function I wanted to use it but then decided otherwise

christianparpart: then probably clean this argument up and not have that if-statement below.

christianparpart (approved):

christianparpart (changes requested):
`lsp.py: error: unrecognized arguments: --batch` needs a fix

test/lsp.py
```diff
@@ -272,27 +273,34 @@ def create_cli_parser() -> argparse.ArgumentParser:
...
         dest="fail_fast",
         action="store_true",
         help="Terminates the running tests on first failure."
     )
+    parser.set_defaults(batch=False)
+    parser.add_argument(
+        "-b", "--batch",
+        dest="batch",
+        action="store_true",
+        help="Prevent interactive queries and just fail instead."
```
christianparpart, line 286: Honestly, is `--batch` then the right word to go with? Why not `--non-interactive` to avoid interactive use (such as in CI runs)?

christianparpart (approved):

PR #13084, opened 2022-05-31
Revert workaround introduced in euler tests because it is no longer needed

matheusaaguiar:
This reverts #13047

### DIFF
--- a/test/externalTests/euler.sh
+++ b/test/externalTests/euler.sh
@@ -67,10 +67,6 @@ function euler_test
     sed -i "/expectError: 'JUNK_UPGRADE_TEST_FAILURE'/d" test/moduleUpgrade.js
     sed -i "/et\.expect(errMsg)\.to\.contain('e\/collateral-violation');/d" test/flashLoanNative.js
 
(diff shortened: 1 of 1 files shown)


cameel:
> This reverts commit [98efb46](https://github.com/ethereum/solidity/commit/98efb460315bfc31452347560acc1c8495fea86b).

Better to link to the PR rather than to the commit. I think it was #13047 that's being reverted here.

cameel (approved):

matheusaaguiar:
> > This reverts commit [98efb46](https://github.com/ethereum/solidity/commit/98efb460315bfc31452347560acc1c8495fea86b).
> 
> Better to link to the PR rather than to the commit. I think it was #13047 that's being reverted here.

Should I change the commit message too?

cameel:
> Should I change the commit message too?

No, the commit message is fine the way it is. It's actually best not to have the PR number there to avoid github spamming mentioned PR with useless notes on each rebase :)

We only want that on github, in PR and issue descriptions because it's **very** convenient to have them all interlinked. So that when you're later investigating something you can just follow the whole chain of events via links :) It's still possible when you post a commit ID but it's more tedious and github does not post a link in the other issue/PR.

PR #13087, opened 2022-06-01
Workaround for the spurious `control reaches end of non-void function` warning in GCC 12.1

cameel:
**Depends on #13088**.

GCC 12.1 warns that `control reaches end of non-void function [-Wreturn-type]` in some places where our assertion macros guarantee that this can never happen.

This started happening for me when GCC 12.1 hit the core Arch Linux repos. The problem does not exist on earlier GCC or on Clang. [I have already reported it in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105794) but for now we need a workaround. I added an explicit `throw` to make GCC shut up.

### DIFF
--- a/libsmtutil/CVC4Interface.cpp
+++ b/libsmtutil/CVC4Interface.cpp
@@ -19,6 +19,7 @@
 #include <libsmtutil/CVC4Interface.h>
 
 #include <libsolutil/CommonIO.h>
+#include <libsolutil/Exceptions.h>
 
 #include <cvc4/util/bitvector.h>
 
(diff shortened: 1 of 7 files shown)


cameel:
As to why `b_archlinux` does not fail there: apparently this warning does not appear when building in release mode. I suspect that GCC's optimizer might be just removing the code paths its analyzer is complaining about. I logged in via SSH and checked my repo and it is reproducible there. But adding `-O2` when compiling that repro makes the warning disappear.

leonardoalt (approved):

PR #13090, opened 2022-06-02
Fixes to build with boost-1.79.0 and Clang on FreeBSD

yurivict:
It was failing with messages like this:
> error: implicit conversion loses integer precision: 'typename std::enable_if<number_category<cpp_int_backend<0, 0, boost::multiprecision::signed_magnitude, boost::multiprecision::unchecked, allocator<unsigned long long>>>::value == number_kind_integer, std::size_t>::type' (aka 'unsigned long') to 'unsigned int'

### DIFF
--- a/libevmasm/GasMeter.cpp
+++ b/libevmasm/GasMeter.cpp
@@ -184,7 +184,7 @@ GasMeter::GasConsumption GasMeter::estimateMax(AssemblyItem const& _item, bool _
 				if (*value)
 				{
 					// Note: msb() counts from 0 and throws on 0 as input.
-					unsigned const significantByteCount  = (boost::multiprecision::msb(*value) + 1 + 7) / 8;
+					unsigned const significantByteCount  = (unsigned)((boost::multiprecision::msb(*value) + 1 + 7) / 8);
 					gas += GasCosts::expByteGas(m_evmVersion) * significantByteCount;
 				}
 			}
--- a/libsolidity/analysis/ConstantEvaluator.cpp
+++ b/libsolidity/analysis/ConstantEvaluator.cpp
@@ -49,7 +49,7 @@ bool fitsPrecisionExp(bigint const& _base, bigint const& _exp)
 
 	size_t const bitsMax = 4096;
 
-	unsigned mostSignificantBaseBit = boost::multiprecision::msb(_base);
+	auto mostSignificantBaseBit = boost::multiprecision::msb(_base);
 	if (mostSignificantBaseBit == 0) // _base == 1
 		return true;
 	if (mostSignificantBaseBit > bitsMax) // _base >= 2 ^ 4096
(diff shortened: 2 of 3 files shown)


libevmasm/GasMeter.cpp
```diff
@@ -184,7 +184,7 @@ GasMeter::GasConsumption GasMeter::estimateMax(AssemblyItem const& _item, bool _
 				if (*value)
 				{
 					// Note: msb() counts from 0 and throws on 0 as input.
-					unsigned const significantByteCount  = (boost::multiprecision::msb(*value) + 1 + 7) / 8;
+					unsigned const significantByteCount  = (unsigned)((boost::multiprecision::msb(*value) + 1 + 7) / 8);
```
cameel, line 187:
```suggestion
					unsigned const significantByteCount  = static_cast<unsigned>((boost::multiprecision::msb(*value) + 1 + 7) / 8);
```

But I think it would be cleaner to just change the type of `significantByteCount` to `unsigned long`:
```suggestion
					unsigned long const significantByteCount = (boost::multiprecision::msb(*value) + 1 + 7) / 8;
```

libsolidity/analysis/ConstantEvaluator.cpp
```diff
@@ -49,7 +49,7 @@ bool fitsPrecisionExp(bigint const& _base, bigint const& _exp)
 
 	size_t const bitsMax = 4096;
 
-	unsigned mostSignificantBaseBit = boost::multiprecision::msb(_base);
+	auto mostSignificantBaseBit = boost::multiprecision::msb(_base);
```
cameel, line 52: We generally prefer explicit types:
```suggestion
	unsigned long mostSignificantBaseBit = boost::multiprecision::msb(_base);
```

libsolutil/Numeric.cpp
```diff
@@ -31,7 +31,7 @@ bool solidity::fitsPrecisionBaseX(bigint const& _mantissa, double _log2OfBase, u
 
 	size_t const bitsMax = 4096;
 
-	unsigned mostSignificantMantissaBit = boost::multiprecision::msb(_mantissa);
+	auto mostSignificantMantissaBit = boost::multiprecision::msb(_mantissa);
```
cameel, line 34:
```suggestion
	unsigned long mostSignificantMantissaBit = boost::multiprecision::msb(_mantissa);
```

cameel (commented):
Thanks for the PR. I wonder if it's really due too Boost 1.77. We have a CI job running with latest Boost, latest GCC, etc. and it's not failing. There must be something else going on here.

Can you give me more details about your environment? Platform, OS, compiler version, etc.

I guess merging this won't hurt if it does not make our other jobs fail but it would be best if we could reproduce this and catch this problem already in PRs.

cameel:
> Can you give me more details about your environment? Platform, OS, compiler version, etc.

Looking at your other issue I see it's FreeBSD + Clang 13. I compiled it locally with Clang 13.0.1 and Boost 1.78.0 and I don't get this warning. So it must be specific to FreeBSD and unfortunately neither CircleCI nor Github Actions have FreeBSD-based runners. So unfortunately we won't be able to detect these easily, we'll have to rely on bug reports.

We can still merge this PR though with the tweaks I suggested above.

yurivict:
My environment:
boost-libs-1.79.0
clang-13.0.0
FreeBSD 13.1

PR #13092, opened 2022-06-02
Disabling more OpenZeppelin tests that expect reverts with specific error

nishant-sachdeva:
Disabling more zeppline tests that expect reverts with specific errors and they're broken via IR due to Hardhat heuristics.

### DIFF
--- a/test/externalTests/zeppelin.sh
+++ b/test/externalTests/zeppelin.sh
@@ -106,6 +106,11 @@ function zeppelin_test
     sed -i "s|it(\('other accounts cannot pause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('other accounts cannot unpause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('prevents initialization'\)|it.skip(\1|g" test/proxy/utils/Initializable.test.js
+    sed -i "s|it(\('divide by 0'\)|it.skip(\1|g" test/utils/math/Math.test.js
+    # CAUTION:: The following two sed commands depend on the order of occurrence of the relevant patterns in the mentioned files.
+    # Could result in an error in the future.
+    sed -zi "s|it(\('deposit'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC20TokenizedVault.test.js
+    sed -zi "s|it(\('withdraw'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC20TokenizedVault.test.js
 
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2115 gets fixed.
     sed -i "s|describe\(('Polygon-Child'\)|describe.skip\1|g" test/crosschain/CrossChainEnabled.test.js


test/externalTests/zeppelin.sh
```diff
@@ -106,6 +106,10 @@ function zeppelin_test
     sed -i "s|it(\('other accounts cannot pause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('other accounts cannot unpause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('prevents initialization'\)|it.skip(\1|g" test/proxy/utils/Initializable.test.js
+    #TODO: Add more cases here to fix new errors
+    sed -i "s|it(\('deposit'\)|it.skip(\1|g/3" test/token/ERC20/extensions/ERC20TokenizedVault.test.js
```
matheusaaguiar: You are almost there.
That `sed` command will substitute (`s`) every occurrence (`g`) that matches the given pattern. But it will do this for **each line**.  In this case, we can tell `sed` to view the whole file as a single line (`-z`) and then match and replace only the third occurrence:
```
sed -zi "s|it(\('deposit'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC20TokenizedVault.test.js
```
Or you can also specify the desired line:
```
sed -i "208s|it(\('deposit'\)|it.skip(\1|" test/token/ERC20/extensions/ERC20TokenizedVault.test.js
```

test/externalTests/zeppelin.sh
```diff
@@ -106,6 +106,10 @@ function zeppelin_test
     sed -i "s|it(\('other accounts cannot pause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('other accounts cannot unpause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('prevents initialization'\)|it.skip(\1|g" test/proxy/utils/Initializable.test.js
+    #TODO: Add more cases here to fix new errors
```
cameel: This TODO is now done, isn't it? :)

matheusaaguiar:
I ran the test locally and everything passes now. I guess it just a matter of addressing cameel's suggestions and maybe adding a commentary for future reference about the workaround.

stackenbotten:
There was an error when running `chk_coding_style` for commit `75f3c5ec8993c0cb855d845d8c56e450ba03657a`:
```
test/externalTests/zeppelin.sh:110:    # CAUTION:: The following two sed commands depend on the order of occurence of the relevant patterns in the mentioned files. 

```

test/externalTests/zeppelin.sh
nishant-sachdeva: @cameel  Pushed changes to address this comment. Will squash commits once you approve.

cameel: I meant that this line can just be removed.

cameel (dismissed):

cameel:
Weird that `c_ext_benchmarks` is failing. It should be fine, only the OZ test can create `benchmark-zeppelin-ir-optimize-evm+yul.json`. Rerunning tests to see if it's reproducible.

test/externalTests/zeppelin.sh
nishant-sachdeva: @matheusaaguiar  mentioned that we could add some explanation in his comment :  https://github.com/ethereum/solidity/pull/13092#issuecomment-1146380471

So, I added this extra comment. I can remove it if required.

cameel:
@nishant-sachdeva I think know why `c_ext_benchmarks` is failing :P You committed an empty `benchmark-zeppelin-ir-optimize-evm+yul.json`. This file is not supposed to be in the repo.

cameel (approved):

cameel:
Now `c_ext_benchmarks` is failing for a different reason: the previous run on `develop` failed so benchmarks are not available. This is unrelated to the PR here. I should have made it succeed in that case and just skip the diff. I'll do that in a separate PR.

Since benchmark job is optional, I'm going to merge this now despite the failure.

PR #13093, opened 2022-06-02
lsp.py: Add missing check for non-interactive

Marenz:


test/lsp.py
```diff
@@ -654,6 +654,9 @@ def user_interaction_failed_method_test(self, testcase, actual, expected):
                 "\nbut got:\n" + actual_pretty
```
cameel, line 654: Unrelated to the PR, but I see that you have lots of `\` line continuations in this file. In Python you don't need it if the line break is inside parentheses or any other construct that makes it clear that the code will be continued. I.e. you'd need them in
```python
message = "Expected:\n" + \
    self.suite.replace_ranges_with_tags(expected, self.sub_dir) + \
    "\nbut got:\n" + actual_pretty
```

but not in
```python
message = (
    "Expected:\n" +
    self.suite.replace_ranges_with_tags(expected, self.sub_dir) +
    "\nbut got:\n" + actual_pretty
)
```

cameel (dismissed):
The failing OZ test is unrelated (@nishant-sachdeva is working on it).
The pylint failure is already fixed on `develop`.

cameel:
> GitGuardian Security Checks

I see that github added something new.

test/lsp.py
Marenz: Understood, good to know :) 
Thanks!

cameel:
I wanted to merge it but I took a look at `chk_pylint` details just in case and I noticed that it has one extra warning compared to what I fixed on `develop`. I think it actually comes from this PR:

```
test/lsp.py:646:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```

test/lsp.py
```diff
@@ -654,6 +654,9 @@ def user_interaction_failed_method_test(self, testcase, actual, expected):
...
             )
 
+        if self.suite.non_interactive:
+            return
```
cameel: pylint seems right. This function normally returns `TestResult`. This would implicitly return `None`. Is `None` even a valid thing to return here?

Marenz: Ops, I probably was to hasty with my fix

Marenz: Finally got around to fix this

cameel (approved):

PR #13094, opened 2022-06-02
Fix missing STL include in tools/solidityUpgrade/SourceUpgrade.cpp

### LINKED ISSUE #13091: #include <fstream> is missing in tools/solidityUpgrade/SourceUpgrade.cpp for std::ofstream
std::ofstream is used there but because there's no ```#include <fstream>``` clang-13 fails on FreeBSD 13.1

yurivict:
Fixes #13091.


cameel (approved):

toshihidemitumar (commented):
image

PR #13096, opened 2022-06-05
Moved Keywords section from 'Cheatsheet' to 'Units and Globally Available Variables'

victorknox:
Part of https://github.com/ethereum/solidity/issues/12934.
Moved the list of keywords to "Units and Globally Available Variables" from "CheatSheet"

### DIFF
--- a/docs/cheatsheet.rst
+++ b/docs/cheatsheet.rst
@@ -112,13 +112,3 @@ Modifiers
 - ``override``: States that this function, modifier or public state variable changes
   the behaviour of a function or modifier in a base contract.
 
-Reserved Keywords
-=================
-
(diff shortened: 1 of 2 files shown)


victorknox:
@cameel sorry about the tag, but could you please review my PR? 
I'm new around so would help me get started

cameel (dismissed):
Looks ok but needs a rebase on current `develop` (the failing CI job is already fixed there).

Also, maybe it would be a nice idea to keep it both here and in the cheatsheet via a shared page (like I did with the operator table in #13138). But I guess what's in the PR now is fine too.

Marenz (approved):

PR #13097, opened 2022-06-06
Skipping benchmark diff when previous run failed

cameel:
This is a fix for the failure of `c_ext_benchmarks` that we observed in #13092.

Currently `download_benchmarks.py` fails when benchmarks on the base branch are not available for whatever reason. They may be unavailable either because `c_ext_benchmarks` on that branch failed (or did not run because of a failure of a dependency) or because the workflow is still in progress. This is what we want for CLI usage, since not being able to provide the benchmarks is technically a failure of the script. When running in CI, however, we only really want the job to fail in a situation where it's a fixable issue that requires attention. For example if we have a syntax error in the script, the job should fail but if the previous benchmark predictably is not (and will not be) available, we should just skip the diff and still provide the current benchmark so that the next run can use it.

This PR updates `download_benchmarks.py` so that it can signal what exactly happened via its exit code and makes `c_ext_benchmarks` job in CI use the code to decide whether to skip diffing or not. The script will also now print more concrete error messages when the job is not successful.

One further improvement we might make would be to retry the script if the workflow on the base branch is still running. For now I just went with the simplest solution: the script fails. This is actually the one situation where failing makes sense because the failure is transient and you can later use "Restart from failed" option in CircleCI to rerun just the benchmark job.

### DIFF
--- a/scripts/externalTests/download_benchmarks.py
+++ b/scripts/externalTests/download_benchmarks.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 from argparse import ArgumentParser, Namespace
+from enum import Enum, unique
 from pathlib import Path
 from typing import Mapping, Optional
 import sys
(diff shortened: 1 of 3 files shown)


nishant-sachdeva (commented):
LGTM

scripts/externalTests/download_benchmarks.py
```diff
@@ -141,32 +154,40 @@ def download_benchmarks(
...
     except requests.exceptions.HTTPError as exception:
         print(f"[ERROR] {exception}", file=sys.stderr)
-        return 1
+        return Status.ERROR.value
     except RuntimeError as exception:
         print(f"[ERROR] {exception}", file=sys.stderr)
-        return 1
+        return Status.ERROR.value
 
 if __name__ == '__main__':
```
wechman, line 193: Could you add an extra empty line before if statement to keep formatting consistent?

wechman (dismissed):
Changes looks fine.

scripts/externalTests/download_benchmarks.py
cameel: Done.

wechman (approved):

PR #13100, opened 2022-06-07
Fix unused store inline assembly

chriseth:


libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -76,7 +78,13 @@ void UnusedStoreEliminator::run(OptimiserStepContext& _context, Block& _ast)
 		ignoreMemory
 	};
 	rse(_ast);
-	rse.changeUndecidedTo(State::Unused, Location::Memory);
+	if (
+		auto evmDialect = dynamic_cast<EVMDialect const*>(&_context.dialect);
+		evmDialect->providesObjectAccess()
```
ekpyron:
```suggestion
		evmDialect && evmDialect->providesObjectAccess()
```

ekpyron:
Fix looks good in general - but we'll have to check under which circumstances this could happen precisely to decide on severity.

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: I guess this is causing the CI segfaults.

chriseth:
It turns out this is only triggered when there is an inline assembly block that does not reference any outside variables.

docs/bugs.json
```diff
@@ -1,4 +1,17 @@
 [
+    {
+        "uid": "SOL-2022-4",
+        "name": "InlineAssemblyMemorySideEffects",
+        "summary": "The Yul optimizer may incorrectly remove memory writes from inline assembly blocks, that do not access solidity variables.",
+        "description": "The Yul optimizer considers all memory writes that are never read from in the outermost Yul block as unused and removes them. This is valid for a full Yul program generated with the new via-IR code generation pipeline. However the legacy code generation path (which is still the default) also runs the Yul optimizer individually on all inline assembly blocks, which do not access local Solidity variables. Consequently, memory writes by such inline assembly blocks are removed as well, if the written memory is never read from in the same assembly block, even if the written memory is accessed later for example by a subsequent inline assembly block.",
```
cameel: I'd reword it a bit to make it clearer why asm blocks in IR are not affected. How about this?

```suggestion
        "description": "The Yul optimizer considers all memory writes in the outermost Yul block that are never read from as unused and removes them. This is valid when that Yul block is the entire Yul program, which is always the case for the Yul code generated by the new via-IR pipeline. Inline assembly blocks are never optimized in isolation when using that pipeline. Instead they are optimized as a part of the whole Yul input. However, the legacy code generation pipeline (which is still the default) runs the Yul optimizer individually on an inline assembly block if the block does not refer to any local variables defined in the surrounding Solidity code. Consequently, memory writes in such inline assembly blocks are removed as well, if the written memory is never read from in the same assembly block, even if the written memory is accessed later, for example by a subsequent inline assembly block.",
```

test/libsolidity/semanticTests/inlineAssembly/optimize_memory_store_multi_block.sol
```diff
@@ -0,0 +1,23 @@
...
+		}
+		return a + b + myGetX();
+	}
+
+	function myGetX() internal view returns (uint256) {
+		assembly {
+			mstore(1, 0x123456789abcdef)
+		}
+		return x;
+	}
```
cameel: I think we had more affected cases on the channel. Maybe it would be a good idea to add them all?

docs/bugs.json
ekpyron: Nice, thanks :-)!

test/libsolidity/semanticTests/inlineAssembly/optimize_memory_store_multi_block.sol
ekpyron: Added another test file with simpler cases. I don't think we need more than that, but if you have some in mind, I'll happily add.

cameel (approved):
So looks like we're done here?

ekpyron:
For the record: @chriseth in channel regarding severity:
> i think medium/low is probably good
more important to release quickly

PR #13102, opened 2022-06-07
Better way to annotate unreachability in C++

christianparpart:
At first my code (after OS upgrade) wasn't compiling anymore so I fixed it an then realized @cameel was faster in fixing (or working around) what GCC 12 is reporting.

I think however, that the above suggested way is probably the better way, generally, and also for future-use. What do you think @cameel ?

PR relates to #13087.

### DIFF
--- a/libsolutil/Assertions.h
+++ b/libsolutil/Assertions.h
@@ -40,6 +40,28 @@ namespace solidity::util
 #define ETH_FUNC __func__
 #endif
 
+#if defined(__GNUC__)
+// GCC 4.8+, Clang, Intel and other compilers compatible with GCC (-std=c++0x or above)
+[[noreturn]] inline __attribute__((always_inline)) void unreachable()
+{
+	__builtin_unreachable();
+}
+
+#elif defined(_MSC_VER) // MSVC
+
+[[noreturn]] __forceinline void unreachable()
+{
+	__assume(false);
+}
+
+#else
+
+[[noreturn]] inline void unreachable()
+{
+	solThrow(Exception, "Unreachable");
+}
+#endif
+
 namespace assertions
 {
 
--- a/libsmtutil/CVC4Interface.cpp
+++ b/libsmtutil/CVC4Interface.cpp
@@ -292,7 +292,7 @@ CVC4::Expr CVC4Interface::toCVC4Expr(Expression const& _expr)
 	smtAssert(false);
 
 	// FIXME: Workaround for spurious GCC 12.1 warning (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105794)
-	throw exception();
+	util::unreachable();
 }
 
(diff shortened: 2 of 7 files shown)


libsolutil/Assertions.h
```diff
@@ -40,6 +40,27 @@ namespace solidity::util
...
+[[noreturn]] __forceinline void unreachable()
+{
+    __assume(false);
+}
+
+#else
+
+[[noreturn]] inline void unreachable()
+{
+}
```
ekpyron: I guess, it'd be safer if this actually did not return, i.e. if it threw a bogus exception... that should then also work for all compilers without case distinction, shouldn't it?

cameel: Did you manage to get this variant with anything? I wonder if the `[[noreturn]]` is enough to make it work.

cameel (commented):
Why not. I just wonder if the specialized versions are really worth it over the simpler default one but other than that I would not be against merging it.

I think that for just the GCC 12 workaround a simple `throw` was good enough since we'll want to remove it as soon as it's fixed anyway but I also remember some situations before where I needed an explicit return even though code was unreachable so I think this will be useful anyway.

libsolutil/Assertions.h
christianparpart: right!

libsolutil/Assertions.h
```diff
@@ -40,6 +40,27 @@ namespace solidity::util
...
+[[noreturn]] inline __attribute__((always_inline)) void unreachable()
+{
+    __builtin_unreachable();
+}
+
+#elif defined(_MSC_VER) // MSVC
+
+[[noreturn]] __forceinline void unreachable()
+{
+    __assume(false);
```
cameel: You used tabs instead of spaces for indentation. That's why `chk_coding_style` is failing.
```suggestion
	__builtin_unreachable();
}

#elif defined(_MSC_VER) // MSVC

[[noreturn]] __forceinline void unreachable()
{
	__assume(false);
```

libsolutil/Assertions.h
christianparpart: I test-wise disabled the obove #if-branches to force this one. The compiler complains on a `[[noreturn]]` function that still returns. So now I've added a `solThrow(Exception, ...)`. But in most (if not all) supported cases, the two above should hit.

libsolutil/Assertions.h
```diff
@@ -40,6 +40,28 @@ namespace solidity::util
...
+
+#else
+
+[[noreturn]] inline void unreachable()
+{
+	solThrow(Exception, "Unreachable");
+}
+
+#endif
 namespace assertions
```
cameel: I'd add an empty line to separate these.

cameel (dismissed):
Still not sure if we really need the specialized versions but having them also does not hurt either so I'd just merge it and be done with it :)

cameel (approved):

Issue #12814, opened 2022-03-17
Overflow checks should not avoid overflow


chriseth:
I'm not totally sure anymore why we implemented overflow checks without using overflow, but I think we should change that and use the cheapest method possible.
I can think of two reasons:
1. it might be easier for a possible future reasoning engine in the compiler
2. it might be easier to detect overflow by external tools

The reasoning engine we currently have in mind should be powerful enough to deal with overflows. Furthermore, we make use of wrapping arithmetic in the optimizer rules already (subtraction is turned into addition for example).

chriseth:
It looks like we can save 7 gas per addition:
```
let z := add(x, y)
if gt(x, not(y)) { revert(0, 0) }
```
wrt
```
let z := add(x, y)
if lt(z, x) { revert(0, 0) }
```

cameel:
Just to clarify: you mean we should have used the `ovf` and `sovf` flags from [EIP-1051: Overflow checking for the EVM](https://eips.ethereum.org/EIPS/eip-1051)?

How do we access them at Yul level?

EDIT: I see that the EIP is stagnant so it's probably not this mechanism.

ekpyron:
I think the main reason for doing the checks as they are done now, was that we expected a weaker reasoning engine to be able to deal with them more easily that way - but the plan back then was not to use any of this in production before the optimizer could remove most of the checks and to revisit how to do the check, when working on the optimizer to remove them...
Given how things turned out since then, it makes sense to hand-optimize them to the cheapest ones without optimization right now...

ekpyron:
Anything left here?

PR #13104, opened 2022-06-07
Do not avoid overflow in arithmetic operations checks

matheusaaguiar:
Changes overflow checks for arithmetic operations as discussed in #12814


test/formal/checked_int_mul_16.py
```diff
@@ -26,16 +26,23 @@
...
 	# Constants
 	maxValue = BVSignedMax(type_bits, n_bits)
 	minValue = BVSignedMin(type_bits, n_bits)
 
 	# Overflow and underflow checks in YulUtilFunction::overflowCheckedIntMulFunction
-	overflow_check_1 = AND(AND(SGT(X, 0), SGT(Y, 0)), GT(X, DIV(maxValue, Y)))
-	underflow_check_1 = AND(AND(SGT(X, 0), SLT(Y, 0)), SLT(Y, SDIV(minValue, X)))
-	underflow_check_2 = AND(AND(SLT(X, 0), SGT(Y, 0)), SLT(X, SDIV(minValue, Y)))
-	overflow_check_2 = AND(AND(SLT(X, 0), SLT(Y, 0)), SLT(X, SDIV(maxValue, Y)))
+	if type_bits == 256:
```
ekpyron: This should probably be:
```suggestion
	if type_bits == n_bits:
```

Here and in the other ones as well.

The underlying problem here is that the SMT solvers don't live up to proving a lot of these up to bit vectors as large as 256 bits - so we usually have to reduce the bit size and argue that it doesn't matter, if you stop at 256 bits or at 16 bits :-).
For the multiplication overflow checks - depending on what we end up doing in the end - we should take care that we have a case where ``n_bits/2 < type_bits < n_bits``, though.
We may need to change ``type_bits *= 2`` to ``type_bits += 4`` below or something like that.

matheusaaguiar: Done.

test/formal/checked_int_mul_16.py
```diff
@@ -26,18 +26,22 @@
...
 	minValue = BVSignedMin(type_bits, n_bits)
+	bitMask =  2**(type_bits-1)
 
 	# Overflow and underflow checks in YulUtilFunction::overflowCheckedIntMulFunction
-	overflow_check_1 = AND(AND(SGT(X, 0), SGT(Y, 0)), GT(X, DIV(maxValue, Y)))
-	underflow_check_1 = AND(AND(SGT(X, 0), SLT(Y, 0)), SLT(Y, SDIV(minValue, X)))
-	underflow_check_2 = AND(AND(SLT(X, 0), SGT(Y, 0)), SLT(X, SDIV(minValue, Y)))
-	overflow_check_2 = AND(AND(SLT(X, 0), SLT(Y, 0)), SLT(X, SDIV(maxValue, Y)))
+	if type_bits == n_bits:
+		overflow_check = AND(ISZERO(ISZERO(Y)), ISZERO(EQ(X, SDIV(product, Y))))
```
matheusaaguiar, line 38: This check seems to be failing. However the corresponding one in `YulUtilFunction::overflowCheckedIntMulFunction` is passing. Not sure if there is something I am missing about the bit operation `EQ` for example.

test/formal/checked_uint_mul_16.py
```diff
@@ -25,13 +25,17 @@
...
 	Y = BVUnsignedUpCast(Y_short, n_bits)
+	product = MUL(X, Y)
 
 	# Constants
 	maxValue = BVUnsignedMax(type_bits, n_bits)
 
 	# Overflow check in YulUtilFunction::overflowCheckedIntMulFunction
-	overflow_check = AND(ISZERO(ISZERO(X)), GT(Y, DIV(maxValue, X)))
+	if type_bits == n_bits:
+		overflow_check = AND(ISZERO(ISZERO(Y)), ISZERO(EQ(X, DIV(product, Y))))
```
matheusaaguiar, line 35: This check has the same problem.

matheusaaguiar:
For the multiplication case of 256bit, I could not come up with something better than checking if `x == product / y` (with y != 0). I think it is safe to assume that in case of an overflow/underflow, the truncated result should not "cycle" back to `x`...

test/formal/checked_uint_mul_16.py
ekpyron: I think there's two problems here: one is just that Z3 is surprisingly bad to handle this with larger bit sizes (probably due to the division).
For ``type_bits == 16`` and ``n_bits == 16``, it just reports "Unable to prove the rule", but does not provide a counterexample, which means, it just times out and cannot tell, if the rule is correct.

If you set the values even more conservatively to ``type_bits = 4`` and ``n_bits = 8`` above, it proves the rule for
``type_bits == 4 && n_bits == 8`` as well as for ``type_bits == 8 && n_bits == 8``.
However, it reports a counterexample for ``type_bits == 6 && n_bits == 8``. That's the tricky case, in which ``n_bits / 2 < type_bits < n_bits``... in our setting, that would for example be the multiplication of two ``uint160``s. For that case, we'll need to come up with a different check, since the current one actually doesn't work... if we don't find anything better, we could just keep the previous checks for that, though.

ekpyron: The signed case may even be more complicated :-).

matheusaaguiar: Interesting. Do you have the values of the counterexample for `type_bits == 6 && n_bits == 8`?

ekpyron: For the record from chat: ``Model: [Y = 40, X = 7]``

 40 * 7 does overflow in 6-bit arithmetic, 40 * 7 = 280. But 280 modulo 256 (since we assume 8-bits in the ambient type) is 24, so product is 24. maxValue, however, in 6 bits is 63... so the check ``product > maxValue``, i.e. ``24 > 63``, is false, even though you did overflow.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -624,16 +624,28 @@ string YulUtilFunctions::overflowCheckedIntAddFunction(IntegerType const& _type)
...
+						// overflow, if x >= 0 and sum > maxValue
+						if and(iszero(slt(x, 0)), sgt(sum, <maxValue>)) { <panic>() }
+						// underflow, if x < 0 and sum < minValue
+						if and(slt(x, 0), slt(sum, <minValue>)) { <panic>() }
+					</256bit>
 				<!signed>
-					// overflow, if x > (maxValue - y)
-					if gt(x, sub(<maxValue>, y)) { <panic>() }
+					<?256bit>
+						// overflow, if x > sum
```
chriseth: This comment is fully redundant.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -624,16 +624,28 @@ string YulUtilFunctions::overflowCheckedIntAddFunction(IntegerType const& _type)
...
+						if gt(x, sum) { <panic>() }
+					<!256bit>
+						// overflow, if sum > maxValue
```
chriseth: Same here.

chriseth:
Let's finish the addition part first and get it merged. Then work on multiplication in a new PR.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -795,15 +816,28 @@ string YulUtilFunctions::overflowCheckedIntSubFunction(IntegerType const& _type)
...
+					<!256bit>
+						// underflow, if y >= 0 and diff < minValue
+						if and(iszero(slt(y, 0)), slt(diff, <minValue>)) { <panic>() }
+						// overflow, if y < 0 and diff > maxValue
+						if and(slt(y, 0), sgt(diff, <maxValue>)) { <panic>() }
+					</256bit>
 				<!signed>
-					if lt(x, y) { <panic>() }
+					<?256bit>
+						//underflow if diff > x
```
chriseth: This comment is fully redundant.

matheusaaguiar:
> Let's finish the addition part first and get it merged. Then work on multiplication in a new PR.

Ok,  #13130  is a separate PR with only add/sub checks.

matheusaaguiar:
PR was separated in two: #13130 (add/sub) and #13177 (mul).

PR #13109, opened 2022-06-08
Fix commandline tests not detecting a snippet with a broken free function definition

sourabhxyz:
Example is incorrect in the sense that "Free functions" cannot have visibility specified.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -517,15 +517,16 @@ SOLTMPDIR=$(mktemp -d)
         opts=()
         # We expect errors if explicitly stated, or if imports
         # are used (in the style guide)
-        if grep -E "This will not compile|import \"" "$f" >/dev/null
+        if grep -E "// This will not compile" "$f" >/dev/null ||
+            sed -e 's|//.*||g' "$f" | grep -E "import \"" >/dev/null
         then
             opts=(--expect-errors)
         fi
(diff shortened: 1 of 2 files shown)


cameel:
Thanks for the correction. It's weird that it actually passed our tests - we do test all these snippets. Turns out that our tests expect the snippet to fail if it has any imports - these won't work in the docs. This snippet triggered that condition because it has an import inside a comment. So we were expecting it to fail and it was indeed failing - but not due to the import :) Now that you fixed it our tests actually started failing because the snippet is no longer producing an error.

I adjusted our regexes to skip commented-out imports and pushed that as an extra commit to your branch. This should make CI pass.

test/cmdlineTests.sh
```diff
@@ -517,15 +517,16 @@ SOLTMPDIR=$(mktemp -d)
...
         then
             opts=(--expect-errors)
         fi
-        if grep "This will report a warning" "$f" >/dev/null
+        if grep "// This will report a warning" "$f" >/dev/null
         then
             opts+=(--expect-warnings)
         fi
-        if grep "This may report a warning" "$f" >/dev/null
+        if grep "// This may report a warning" "$f" >/dev/null
```
cameel, line 529: To make this more robust I think we should also be checking that the output actually contains an import error and not just any error. I won't be doing that in this PR though.

Marenz (approved):

PR #13110, opened 2022-06-08
AST changes for LSP

### LINKED ISSUE #13064: Ensure every string in the AST has a location associated with it
This is a requirement for many LSP features, most recently the `semanticTokens` support for highlighting and the `rename` support.

Currently, the following AST types have strings which have no source location available for them:

* `IdentifierPath`: the individual strings that make up the path are only a vector of `ASTString`s, no location information for the individual parts are available
* * [x] https://github.com/ethereum/solidity/pull/13182
* * [x] https://github.com/ethereum/solidity/pull/13183
* `MemberAccess`: member has no location
*  * [x] https://github.com/ethereum/solidity/pull/13184
* `FunctionCall`: The list of argument names for a named call contains no source code locations
* * [x] https://github.com/ethereum/solidity/pull/13185

Marenz:
fixes #13064


libsolidity/analysis/NameAndTypeResolver.cpp
```diff
@@ -185,8 +185,20 @@ vector<Declaration const*> NameAndTypeResolver::nameFromCurrentScope(ASTString c
...
 Declaration const* NameAndTypeResolver::pathFromCurrentScope(vector<ASTString> const& _path) const
+{
+	auto declarations = pathFromCurrentScopeWithAllDeclarations(_path);
+
+	if (declarations.empty())
+		return nullptr;
+
+	return declarations.back();
+}
+
```
christianparpart:
```suggestion
{
	if (auto declarations = pathFromCurrentScopeWithAllDeclarations(_path); !declarations.empty())
	    return declarations.back();
	else
		return nullptr;
}

```

libsolidity/ast/AST.h
```diff
@@ -587,10 +587,16 @@ class ContractDefinition: public Declaration, public StructurallyDocumented, pub
...
 public:
-	IdentifierPath(int64_t _id, SourceLocation const& _location, std::vector<ASTString> _path):
-		ASTNode(_id, _location), m_path(std::move(_path)) {}
+	IdentifierPath(
+		int64_t _id,
+		SourceLocation const& _location,
+		std::vector<ASTString> _path,
+		std::vector<SourceLocation> _pathLocations
+	):
+		ASTNode(_id, _location), m_path(std::move(_path)), m_pathLocations(std::move(_pathLocations)) {}
```
christianparpart:
```suggestion
		ASTNode(_id, _location), m_path(std::move(_path)), m_pathLocations(std::move(_pathLocations))
		{
		    solAssert(m_pathLocations.size() == m_path.size(), "");
		}
```
I think they must always match, right?

libsolidity/ast/AST.h
```diff
@@ -600,6 +606,8 @@ class IdentifierPath: public ASTNode
 	void accept(ASTConstVisitor& _visitor) const override;
 private:
 	std::vector<ASTString> m_path;
+	// These are currently not exported to JSON!
```
christianparpart:
```suggestion
	// These are currently not exported to JSON to not break backwards compatibility!
```

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -334,7 +334,11 @@ ASTPointer<IdentifierPath> ASTJsonImporter::createIdentifierPath(Json::Value con
 		astAssert(!s.empty(), "Expected non-empty string for IdentifierPath element.");
 		namePath.emplace_back(s);
 	}
-	return createASTNode<IdentifierPath>(_node, namePath);
+	return createASTNode<IdentifierPath>(
+		_node,
+		namePath,
+		vector<SourceLocation>{ /* Data not exported, create dummy array. */ }
```
christianparpart: I just suggested an solAssert() above, I think in that case, we should change that assert then to either match explicitly the counts or have zero source locations.

libsolidity/parsing/Parser.cpp
```diff
@@ -1061,14 +1061,27 @@ ASTPointer<IdentifierPath> Parser::parseIdentifierPath()
 	RecursionGuard recursionGuard(*this);
 	ASTNodeFactory nodeFactory(*this);
 	nodeFactory.markEndPosition();
+
+	SourceLocation identifierLocation{currentLocation()};
```
christianparpart:
```suggestion
	SourceLocation identifierLocation = currentLocation();
```
I personally like the `{}`-initialization, but I think that would break style (here) otherwise.

libsolidity/ast/AST.h
Marenz: Depends on how we want to do the json import..

libsolidity/ast/AST.h
```diff
@@ -587,10 +587,19 @@ class ContractDefinition: public Declaration, public StructurallyDocumented, pub
...
-		ASTNode(_id, _location), m_path(std::move(_path)) {}
+	IdentifierPath(
+		int64_t _id,
+		SourceLocation const& _location,
+		std::vector<ASTString> _path,
+		std::vector<SourceLocation> _pathLocations
+	):
+		ASTNode(_id, _location), m_path(std::move(_path)), m_pathLocations(std::move(_pathLocations))
+	{
+		solAssert(m_pathLocations.size() == m_path.size());
```
christianparpart, line 598:
```suggestion
		solAssert(m_pathLocations.empty() || m_pathLocations.size() == m_path.size());
```
Shouldn't it be that? Because of AST JSON import still supporting old ASTs?

Marenz: I am creating an array with fitting size but default source locations

libsolidity/ast/AST.h
```diff
@@ -600,6 +609,8 @@ class IdentifierPath: public ASTNode
 	void accept(ASTConstVisitor& _visitor) const override;
 private:
 	std::vector<ASTString> m_path;
+	// These are currently not exported to JSON to keep backwards compatibility!!
```
chriseth: Why is this a problem? This just adds something to the ast - so there should be no issue with backwards compatibility, should there?

libsolidity/ast/AST.h
```diff
@@ -587,10 +587,19 @@ class ContractDefinition: public Declaration, public StructurallyDocumented, pub
 class IdentifierPath: public ASTNode
 {
 public:
-	IdentifierPath(int64_t _id, SourceLocation const& _location, std::vector<ASTString> _path):
-		ASTNode(_id, _location), m_path(std::move(_path)) {}
+	IdentifierPath(
+		int64_t _id,
+		SourceLocation const& _location,
+		std::vector<ASTString> _path,
+		std::vector<SourceLocation> _pathLocations
```
chriseth, line 594: I assume that the path locations are the source locations of the individual elements, i.e. in `a.b.c.d`, the location of the `a`, then of the `b`, then of the `c` and so on. In contrast to other ast elements, these are not nested. So the other interpretation could be that it is the location of `a`, then of `a.b` then of `a.b.c` and so on.
Please document which one is the case.

libsolidity/ast/AST.h
Marenz: We introduce a new c'tor parameter that needs to be filled when creating the AST from importing JSON. In that case we only have dummy data in this variable.

libsolidity/ast/ASTAnnotations.h
```diff
@@ -256,6 +256,9 @@ struct IdentifierPathAnnotation: ASTAnnotation
 	Declaration const* referencedDeclaration = nullptr;
```
chriseth, line 256: Is this duplicated in the vector?

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -334,7 +334,11 @@ ASTPointer<IdentifierPath> ASTJsonImporter::createIdentifierPath(Json::Value con
 		astAssert(!s.empty(), "Expected non-empty string for IdentifierPath element.");
 		namePath.emplace_back(s);
 	}
-	return createASTNode<IdentifierPath>(_node, namePath);
+	return createASTNode<IdentifierPath>(
+		_node,
+		namePath,
+		vector<SourceLocation>(namePath.size() /* Data not exported, create dummy array. */ )
```
chriseth: Please export it.

libsolidity/ast/ASTAnnotations.h
Marenz: Yes. I wanted to avoid changes all over the project, but I can only use the vector or even add a small function to access the last element

libsolidity/analysis/NameAndTypeResolver.h
```diff
@@ -93,6 +93,10 @@ class NameAndTypeResolver
 	/// Should only be called during the initial resolving phase.
 	/// @note Returns a null pointer if any component in the path was not unique or not found.
 	Declaration const* pathFromCurrentScope(std::vector<ASTString> const& _path) const;
+	/// Resolves a path starting from the "current" scope, but also searches parent scopes.
+	/// Should only be called during the initial resolving phase.
+	/// @note Returns an empty vector if any component in the path was not unique or not found.
```
chriseth: Please document what it returns in the success case as well.

libsolidity/ast/ASTJsonImporter.cpp
Marenz: In that case it is no longer backwards compatible though, we couldn't import earlier ASTs anymore. Well, I guess we could gracefully check for the existence of the field and use if it's there

Marenz:
I've split this PR into several smaller ones:

* https://github.com/ethereum/solidity/pull/13182
* https://github.com/ethereum/solidity/pull/13183
* https://github.com/ethereum/solidity/pull/13184
* https://github.com/ethereum/solidity/pull/13185

PR #13117, opened 2022-06-09
Remove all-whitespace lines from docs.

ekpyron:
Followup on https://github.com/ethereum/solidity/pull/13115


cameel (approved):
Argh... Looks like I should not have skipped CI for #13115 :|
It looked like there was nothing that could break it but I was wrong.

ekpyron:
> Argh... Looks like I should not have skipped CI for #13115 :| It looked like there was nothing that could break it but I was wrong.

Happens :-).

PR #13119, opened 2022-06-10
Fix ternary operator type deduction

### LINKED ISSUE #13078: Ternary operator does not correctly deduce common type for `-1` and `0` to be `int8`
## Description

Using trinary comparison operators causes type inference to fail.

## Environment

- Compiler version: 0.8.14
- Target EVM version (as per compiler settings): unknown
- Operating system: Linux

## Steps to Reproduce

```solidity
    function ab(int a, int b) public returns (int) {
        return a < b ? -1 : 0;
    }
```

Yields

```
TypeError: True expression's type int8 does not match false expression's type uint8.
    |
16 |         return a < b ? -1
    |                ^ (Relevant source part starts here and spans across multiple lines).
```

So there are really two problems:

1. The trinary operator's true and false operands are not appropriately unified to `int8` (which works for both operands)
2. The trinary operator's true and false operands are not correctly widened from `int8` to the return type `int`.

The solution is a bit ugly:

```solidity
    function ab(int a, int b) public returns (int) {
        return a < b ? int(-1) : int(0);
    }
```

wechman:
fixes #13078 

Type deduction for ternary operator starts with a check if an implicit conversion between one operand type into a mobile type of the other one is possible. That allows to use all the information of the operand types (e.g. literal types) before it is lost in result of conversion to the mobile type.

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -1467,24 +1467,25 @@ bool TypeChecker::visit(Conditional const& _conditional)
 	_conditional.trueExpression().accept(*this);
 	_conditional.falseExpression().accept(*this);
 
-	Type const* trueType = type(_conditional.trueExpression())->mobileType();
-	Type const* falseType = type(_conditional.falseExpression())->mobileType();
-
+	Type const* trueType =  type(_conditional.trueExpression());
+	Type const* falseType = type(_conditional.falseExpression());
+	Type const* trueMobileType = trueType->mobileType();
+	Type const* falseMobileType = falseType->mobileType();
 	Type const* commonType = nullptr;
 
-	if (!trueType)
-		m_errorReporter.typeError(9717_error, _conditional.trueExpression().location(), "Invalid mobile type in true expression.");
-	else
+	if (trueMobileType)
 		commonType = trueType;
-
-	if (!falseType)
-		m_errorReporter.typeError(3703_error, _conditional.falseExpression().location(), "Invalid mobile type in false expression.");
 	else
-		commonType = falseType;
+		m_errorReporter.typeError(9717_error, _conditional.trueExpression().location(), "Invalid mobile type in true expression.");
 
-	if (!trueType && !falseType)
+	if (falseMobileType)
+		commonType = falseType;
+	else
+		m_errorReporter.typeError(3703_error, _conditional.falseExpression().location(), "Invalid mobile type in false expression.");
+
+	if (!trueMobileType && !falseMobileType)
 		BOOST_THROW_EXCEPTION(FatalError());
-	else if (trueType && falseType)
+	else if (trueMobileType && falseMobileType)
 	{
 		commonType = Type::commonType(trueType, falseType);
 
@@ -1494,14 +1495,14 @@ bool TypeChecker::visit(Conditional const& _conditional)
 					1080_error,
 					_conditional.location(),
 					"True expression's type " +
-					trueType->humanReadableName() +
+					trueMobileType->humanReadableName() +
 					" does not match false expression's type " +
-					falseType->humanReadableName() +
+					falseMobileType->humanReadableName() +
 					"."
 					);
 			// even we can't find a common type, we have to set a type here,
 			// otherwise the upper statement will not be able to check the type.
-			commonType = trueType;
+			commonType = trueMobileType;
 		}
 	}
 
--- a/test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
+++ b/test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
@@ -1,0 +1,22 @@
+contract C {
+    function f(int a, int b) public returns (int) {
+        return a < b ? -1 : 1;
+    }
+
+    function g(int a) public returns (int) {
+        int8 b = -1;
+        return a < 0 ? b : 1;
+    }
+
+    function h(int a) public returns (int) {
+        int8 b = -1;
+        return a < 0 ? 1 : b;
+    }
+}
+// ----
+// f(int256,int256): 0, 2 -> -1
+// f(int256,int256): 2, 0 -> 1
+// g(int256): -2 -> -1
+// g(int256): 2 -> 1
+// h(int256): -2 -> 1
+// h(int256): 2 -> -1
--- a/test/libsolidity/syntaxTests/literals/ternary_operator_with_address_arguments.sol
+++ b/test/libsolidity/syntaxTests/literals/ternary_operator_with_address_arguments.sol
@@ -1,0 +1,14 @@
+contract C {
+    function f(bool cond) public {
+         // OK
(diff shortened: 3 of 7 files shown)


test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
```diff
@@ -0,0 +1,22 @@
...
+
+    function g(int a) public returns (int) {
+        int8 b = -1;
+        return a < 0 ? b : 1;
+    }
+
+    function h(int a) public returns (int) {
+        int8 b = -1;
+        return a < 0 ? b : 1;
+    }
```
cameel: These two functions are identical. Wasn't one supposed to differ in some way?

test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
```diff
@@ -0,0 +1,22 @@
+contract C {
+    function f(int a, int b) public returns (int) {
+        return a < b ? -1 : 1;
+    }
```
cameel, line 4: I think this feature needs much more test coverage. Just looking at the code I can't really be sure that this will behave as expected in all the possible combinations. Our coverage of ternary operator with various types seems pretty sparse in general so we need to increase it.

Here are some suggestions of things I'd try to test:
- More sizes. E.g. `int8` vs `uint16`, etc. Also literals that are large enough to require these sizes.
- More types: `bytesNN`, booleans, strings. For example:
    - `x ? 1 : bytes1(1)`
    - `x ? "12345678" : bytes32(1)`
- `address` and `address payable`. The latter is convertible to the former and note that `address` literals exist (e.g. `0x1234567890123456789012345678901234567890` is of address type, not `uint160` or `bytes20`).
- Fixed point types. Do these work now?
    - `x ? 1.0 : 1.1`
    - `x ? 0 : 1.1`
- Are there any type combinations where both implicit conversions are possible so the result depends on whether we consider the true or false part first?
    - What about `x ? bs : bm` where `bs` is a storage bytes array and bm is memory bytes array? Will the result be one or the other type depending on the order?
- Array literals? Arrays vs slices? Slices are convertible to arrays under some conditions.

Also, it would be good to know which of these already worked and which ones are new.

cameel (commented):
The implementation seems fine but it's really hard be sure of its exact consequences without more test coverage.

chriseth:
I don't think we should complicate type conversion rules. It is much more confusing if there are more special cases than simple rules that require you to do a bit more typing.

chriseth:
If we want, we could improve the suggestion how to fix the error.

test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
wechman: I added suggested test cases. Let me know if there is anything else that is missing.
Here are all test cases that were failing before and are successful after this PR changes:

**bytes1 v2 = cond ? bytes1("a") : "a";**
TypeError 1080: (142-166): True expression's type bytes1 does not match false expression's type string memory.

**bytes2 v3 = cond ? bytes2("a") : "a";**
TypeError 1080: (189-213): True expression's type bytes2 does not match false expression's type string memory.

**int8 v1 = cond ? -1 : 1;**
TypeError 1080: (82-95): True expression's type int8 does not match false expression's type uint8.

**int16 v2 = cond ? -32768 : 1;**
TypeError 1080: (117-134): True expression's type int16 does not match false expression's type uint8.

**int16 v5 = cond ? int16(1) : 1;**
TypeError 1080: (241-260): True expression's type int16 does not match false expression's type uint8.

**ufixed8x1 v7 = cond ? 1.0 : 1.1;**
TypeError 1080: (329-345): True expression's type uint8 does not match false expression's type ufixed8x1.**
TypeError 9574: (314-345): Type uint8 is not implicitly convertible to expected type ufixed8x1.

**ufixed8x1 v8 = cond ? 0 : 1.1;**
TypeError 1080: (371-385): True expression's type uint8 does not match false expression's type ufixed8x1.**
TypeError 9574: (356-385): Type uint8 is not implicitly convertible to expected type ufixed8x1.

**fixed8x1 v9 = cond ? -1.0 : -1.1;**
TypeError 1080: (410-428): True expression's type int8 does not match false expression's type fixed8x1.
TypeError 9574: (396-428): Type int8 is not implicitly convertible to expected type fixed8x1.

**fixed8x1 v10 = cond ? -1.1 : 1;**
TypeError 1080: (454-469): True expression's type fixed8x1 does not match false expression's type uint8.

wechman:
> I don't think we should complicate type conversion rules. It is much more confusing if there are more special cases than simple rules that require you to do a bit more typing.

Actually the only thing we need to do is to drop premature conversion to the mobile type. I updated draft with that. Let me know if it looks better now.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1459,24 +1459,25 @@ bool TypeChecker::visit(Conditional const& _conditional)
...
+	else
+		m_errorReporter.typeError(3703_error, _conditional.falseExpression().location(), "Invalid mobile type in false expression.");
 
-	if (!trueType && !falseType)
+	if (!trueMobileType && !falseMobileType)
 		BOOST_THROW_EXCEPTION(FatalError());
-	else if (trueType && falseType)
+	else if (trueMobileType && falseMobileType)
 	{
 		commonType = Type::commonType(trueType, falseType);
```
chriseth, line 1490: So this is the main "change" here, right? So if we have types that are not their own mobile types, the common type will be computed from the non-mobile types instead of the mobile types, right?
What are the common types in these cases?

chriseth: To be honest, I still cannot really grasp the consequences of this change.

wechman: Yes, this is the main change. The common type does not change:
- It is a mobile type of `a` if  `b` is implicitly convertible to mobile type of `a`. 
- It is a mobile type of `b` if `a` is implicitly convertible to mobile type of `b`. 
- Otherwise it is `nullptr`. 

Please notice that `Type::commonType` function checks if conversion between type of one parameter and mobile type of another parameter is possible. It converts to mobile type when it is necessary and also returns mobile type. However, when we convert parameters to their mobile type before we pass them to the function, we loose information that might be crucial. For instance, we loose information about a literal type exact value. This is why I think we should call `Type::commonType` with full types.

leonardoalt:
This hasn't changed in more than a month. Is @wechman working on it or is it a `takeover` PR?

wechman:
@leonardoalt Actually, there is no much to do. The PR is ready to be reviewed.

test/libsolidity/semanticTests/literals/ternary_operator_with_rationals.sol
wechman: The test case is already reworked.

NunoFilipeSantos:
We're ok with the current behavior, so we're closing it.

cameel:
Well, I'm not ok with the current behavior. Still, this is very low priority and needs some careful review to make sure it does not break anything so better close it because I'm not sure if we'll manage to get back to this any time soon.

Issue #12783, opened 2022-03-12
`UnimplementedFeatureError` with no line indication when copying an array of structs to storage


barakman:
## Description

I am getting an error with no indication of which line it refers to:
```
UnimplementedFeatureError: Copying of type struct MyStruct memory[] memory to storage not yet supported.
```

## Environment

- Compiler version: 0.8.12

## Steps to Reproduce

Well, I need to dig through my code in order to find it, but on the compiler side, the problem is probably in ArrayUtils.cpp:
```
solUnimplemented("Copying of type " + _sourceType.toString(false) + " to storage not yet supported.");
```

Thanks :)

wechman:
@barakman  Thank you for your ticket! Could you share the code we can use to reproduce the issue? Hope you have managed to find it. Otherwise, even if we made adjustments you suggest, we wouldn't be sure the problem is fixed.

cameel:
This looks like the same thing as #3446.

And it probably won't be implemented in the current syntax but #2435 will provide a different mechanism for the same thing.

I think we should replace the `UnimplementedFeatureError` with a proper error at the analysis stage. Letting it blow up in the codegen is not a great user experience. And even when we have `copyof` syntax will be different so will still need to block this case.

barakman:
> @barakman Thank you for your ticket! Could you share the code we can use to reproduce the issue? Hope you have managed to find it. Otherwise, even if we made adjustments you suggest, we wouldn't be sure the problem is fixed.

With no line indication on your side, all I can do is paste my entire project here, which you'll probably agree is senseless.
I suggest that you add a line indication on your side (as pointed out in my initial message), then similar issues in the future will also include a short snippet of the client code (solidity) causing the problem.

As you might understand, the main issue in this context is not the compilation error itself, but rather the message issued by the compiler, which lacks the information required for the developer to fix (or workaround) the error.

cameel:
Yeah, the error should be better.

But just to help you locate it in your particular case - you can try commenting out lines that assign something of type `MyStruct` stored in memory to a storage variable. If doing this makes the error go away, it's the same issue.

timweri:
I'm interested in fixing this. Will start digging to see what I can find.

barakman:
So what's the status on this? Still getting it on solc 0.8.17.

cameel:
Sorry, no one is working actively on this due to other tasks with higher priorites.

[...]

PR #13128, opened 2022-06-13
docs: specify memory-safe is available since 0.8.13 for inline assembly

CJ42:
# What does this PR introduce?

A very simple mention in the **inline-assembly** section in the docs about the new `("memory-safe")` feature.
Since it was introduced recently, I thought it would be good to mention it is available since 0.8.13, as devs using older compiler version like 0.8.7 might not guess it from the docs.

Not sure if it's in your conventions tho to mention the versions number directly in the docs content body.

### DIFF
--- a/docs/assembly.rst
+++ b/docs/assembly.rst
@@ -364,8 +364,9 @@ in memory is automatically considered memory-safe and does not need to be annota
     an assembly block as memory-safe, but violate one of the memory assumptions, this **will** lead to incorrect and
     undefined behaviour that cannot easily be discovered by testing.
 
+The annotation was introduced in version 0.8.13 and is not supported by older compilers.
 In case you are developing a library that is meant to be compatible across multiple versions
-of solidity, you can use a special comment to annotate an assembly block as memory-safe:
+of Solidity, you can use a special Natspec comment that has the same effect but is ignored in older versions:
 
 .. code-block:: solidity
 


docs/assembly.rst
```diff
@@ -292,7 +292,7 @@ in an incompatible way. Therefore, moving stack variables to memory and addition
 by default, disabled in the presence of any inline assembly block that contains a memory operation or assigns
 to solidity variables in memory.
 
-However, you can specifically annotate an assembly block to indicate that it in fact respects Solidity's memory
+However since v0.8.13, you can specifically annotate an assembly block to indicate that it in fact respects Solidity's memory
```
cameel:
```suggestion
However since version 0.8.13, you can specifically annotate an assembly block to indicate that it in fact respects Solidity's memory
```

cameel (commented):
We don't normally mention when features were introduced though I guess we could add it if you think it adds some value. But maybe it would be less out of place in the paragraph near the end that talks about the special natspec comment (`In case you are developing a library that is meant to be compatible across multiple versions ...`)? That's where the version is relevant.

cameel:
Hello, @CJ42 are you still working on the PR? What do you think about the changes I requested above?

CJ42:
@cameel went back to the PR.
I have changed the mention to be at the end, where you mention about natspec comments and that _"annotation via comment will be disallowed in a future breaking release"_.

I still think it should be mentioned in a bit more obvious manner that this is a recently new feature (maybe via a link to changelog? a callout?)

If you are new and want to use the memory safe feature in assembly using an older/more stable version of the compiler like 0.8.7 for instance, the error message from the Solidity compiler does not help to know that.

_example:_

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.7;

contract Draft {
    
    function test() public {

        assembly ("memory-safe") {
            // ...
        }
    }
}
```

image

A little hint in the docs would help. How about mentioning it in a blue callout under the sub-heading **Memory Safety[](https://docs.soliditylang.org/en/v0.8.15/assembly.html#memory-safety)**?

cameel:
Thanks for adding more context. This explains why you want it there but, honestly, I'm still not convinced that this particular case requires special treatment. We introduce new features all the time and using them with older versions will often produce similarly unfriendly errors. If it's really an issue, we should address it more comprehensively. For example, I would not be against having the version listed explicitly for each major feature. You could submit a feature request for that and we could discuss it further. For now we're not doing it though so it would be a bit out of place. I think that having the version mentioned where it's relevant (i.e. in the paragraph about the comment) is more in line with the rest of the docs

By the way, I just pushed some small tweaks to the text. Please see if maybe this version would be better? We could also make it a note, but I think it makes more sense as part of the main text.

CJ42:
Ok @cameel  I understand the context of new features better and not mention the version they were introduced everywhere.

I think the change and way it's described in the paragraph now look good to me.

I am new to the Solidity language but fairly new to contributing Open Source to the Solidity repository (and very keen to do so!)

What is the best place to start? Do you have some areas (docs/features) that need particular attention? If yes where can I find more info? 🙂

cameel:
ok then. In that case I'm going to squash commits here and merge it. Thanks for the PR!

> What is the best place to start? Do you have some areas (docs/features) that need particular attention? If yes where can I find more info? slightly_smiling_face

We have some issues labeled with [`good first issue`](https://github.com/ethereum/solidity/issues?q=is%3Aissue+is%3Aopen+label%3A"good+first+issue"). These are always good starting points and you'll find easier and harder ones, both actual C++ code and docs. Note that many already have pending PRs (these should be visible in the issue though).

Though honestly, our problem right now is that we don't yet have enough manpower to review all PRs efficiently so they're piling up, especially in areas that are not a high priority - as you have probably seen by how long this PR has been waiting :) If you'd like something that would get more attention, perhaps #12783 would be a good one to try. This one might require some guidance though.

cameel (approved):

Issue #13086, opened 2022-06-01
Include all emitted events in the contract ABI and userdoc in a non-breaking way by ignoring conflicting Natspec


cameel:
## Abstract
Currently the contract ABI lists only events defined in the contract and inherited from other contracts. It does not list events defined in libraries that the contract emits and (after #12921) also events defined in other contracts.

Just like we do with errors, the ABI should list both defined and emitted events.

## Motivation
A contract can emit events defined in a variety of places:

```solidity
library   L { /** A library event */             event EL(uint8); }
interface I { /** An interface event */          event EI(uint16); }
contract  C { /** A foreign contract event */    event EC(uint24); }
contract  D { /** An inherited contract event */ event ED(uint32); }

contract E is D {
    /// Own event
    event EE(uint40);

    function foo() public {
        emit L.EL(1);
        emit I.EI(2);
        emit C.EC(3);
        emit ED(4);
        emit EE(5);
    }
}
```
Not all of them get listed in the ABI:
```json
[
    {
        "anonymous": false,
        "inputs":
        [{"indexed": false, "internalType": "uint32", "name": "", "type": "uint32"}],
        "name": "ED",
        "type": "event"
    },
    {
        "anonymous": false,
        "inputs":
        [{"indexed": false, "internalType": "uint40", "name": "", "type": "uint40"}],
        "name": "EE",
        "type": "event"
    },
    {
        "inputs": [],
        "name": "foo",
        "outputs": [],
        "stateMutability": "nonpayable",
        "type": "function"
    }
]
```

This poses a problem for tools like Truffle, which use the ABI to determine which events the contract might emit.

The `solc --userdoc` output compiled from the Natspec docstrings also only lists this subset of events:
```json
{
    "events":
    {
        "ED(uint32)": {"notice": "An inherited contract event "},
        "EE(uint40)": {"notice": "Own event"}
    },
    "kind": "user",
    "methods": {},
    "version": 1
}
```

## Specification
The only real problem here is the userdoc. The current structure only allows one set of docs for each signature. We are planning to change the structure in Natspec v2 (#11114) but this will be a breaking change.

[...]

PR #13135, opened 2022-06-14
Revert access to foreign events and clean up event/error selector tests/docs

cameel:
Related to #13086.

This reverts the part of #12921 that enabled full access to events defined in foreign contracts and interfaces.

I also cleaned up some mess I noticed in tests and docs while doing the revert.


Changelog.md
```diff
@@ -6,7 +6,6 @@ Important Bugfixes:
 
 Language Features:
  * Add `E.selector` for a non-anonymous event `E` to access the 32-byte selector topic.
- * Errors and Events allow qualified access from other contracts.
```
cameel, line 9: This was partially wrong. #12921 only allowed access for events. Foreign errors were already accessible before.

test/libsolidity/semanticTests/error/selector.sol
```diff
@@ -1,18 +0,0 @@
...
-}
-contract C {
-    function f() public pure returns (bytes4, bytes4) {
-        assert(L.E.selector == T.E.selector);
-        assert(L.E.selector != S.E.selector);
-        return (L.E.selector, S.E.selector);
-    }
-}
-// ----
-// f() -> 0x92bbf6e800000000000000000000000000000000000000000000000000000000, 0x2ff06700000000000000000000000000000000000000000000000000000000
```
cameel, line 18: This test is completely redundant. Probably should have been deleted when `error_selector.sol` was added.

cameel: Also, it was in `semanticTests/error/` while we also have `semanticTests/errors/`. I'm merging these two dirs here and leaving only `errors/`.

test/libsolidity/syntaxTests/errors/error_selector_syntax.sol
```diff
@@ -1,48 +1,34 @@
...
+        assert(librarySelector == L.E.selector);
+        assert(E.selector == B.E.selector);
 
-    function test3() pure external returns (bytes4) {
-        return (F.selector);
+        if (condition)
+            revert E(E.selector, true, "123");
+        else
+            revert L.E((B.E.selector), true, "123");
     }
```
cameel, line 29: I have rewritten this test for errors and the other, almost identical one for events because they we already have almost identical semantic tests. They weren't really testing all that much about the selector syntax. I made it a bit more diverse.

ekpyron (approved):

PR #13138, opened 2022-06-14
Operator precedence table shared between Cheatsheet and Operators

cameel:
Related to #12934.
Replaces and closes #13082.

This way we can have it in both places (until we restructure Operators).


ekpyron (approved):
Ok, I only now realized that "operators" was already filed controversially under "types"... given that, yeah, I guess this is not the worst solution for now :-).

PR #13139, opened 2022-06-14
Warn about multiple assignments to storage byte pushes.

ekpyron:


test/libsolidity/syntaxTests/tupleAssignments/warn_multiple_storage_storage_copies.sol
```diff
@@ -4,6 +4,11 @@ contract C {
 	function f() public {
 		(x, y) = (y, x);
 	}
+	function g() public {
+		uint z;
+		((x, y), z) = ((y, x), 0);
```
ekpyron, line 9: This was missed earlier due to not counting recursively...

test/libsolidity/syntaxTests/array/bytes_push_assign_multi.sol
```diff
@@ -0,0 +1,21 @@
...
+    bytes z;
+    function f() public {
+        (x.push(), x.push()) = (0,0);
+        (((x.push())), (x.push())) = (0,0);
+        ((x.push(), x.push()), x.push()) = ((0,0), 0);
+        (x.push(), x[0]) = (0, 0);
+        bytes storage y = x;
+        (x.push(), y.push()) = (0,0);
+        // The following is a false positive.
+        (x.push(), z.push()) = (0,0);
```
ekpyron: Not sure if we should be bothered by such false positives or not.
For comparison: the other warning regarding copies to storage has similar false positives - it'll warn about ``(x,y) = (y,x)``, but also about ``(x,y) = (z, w)`` (even if they're all definitely distinct).

libsolidity/ast/ASTUtils.h
```diff
@@ -38,4 +38,8 @@ bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
 /// Returns the innermost AST node that covers the given location or nullptr if not found.
 ASTNode const* locateInnermostASTNode(int _offsetInFile, SourceUnit const& _sourceUnit);
 
+/// @returns @a _expr itself, in case it is not an unary tuple expression. Otherwise it descends recursively
```
cameel:
```suggestion
/// @returns @a _expr itself, in case it is not a unary tuple expression. Otherwise it descends recursively
```

libsolidity/ast/ASTUtils.h
```diff
@@ -38,4 +38,8 @@ bool isConstantVariableRecursive(VariableDeclaration const& _varDecl);
...
+/// into unary tuples and returns the contained expression.
+Expression const* resolveUnaryTuples(Expression const* _expr);
```
cameel: This does nothing for any unary tuples nested somewhere deeper in the expression. Just the outer ones, right?
```suggestion
Expression const* resolveOuterUnaryTuples(Expression const* _expr);
```

test/libsolidity/syntaxTests/tupleAssignments/warn_multiple_storage_storage_copies.sol
```diff
@@ -4,6 +4,11 @@ contract C {
...
+	}
 }
 // ----
 // Warning 7238: (79-94): This assignment performs two copies to storage. Since storage copies do not first copy to a temporary location, one of them might be overwritten before the second is executed and thus may have unexpected effects. It is safer to perform the copies separately or assign to storage pointers first.
+// Warning 7238: (134-159): This assignment performs two copies to storage. Since storage copies do not first copy to a temporary location, one of them might be overwritten before the second is executed and thus may have unexpected effects. It is safer to perform the copies separately or assign to storage pointers first.
```
cameel, line 14: This is was already here before the PR so not necessarily something to fix right now but I just wanted to note that the message could be more precise. It's not really about simultaneous copies by themselves but simultaneous reads and writes of the same storage location, right?

test/libsolidity/syntaxTests/array/bytes_push_assign_multi.sol
```diff
@@ -0,0 +1,21 @@
+contract C {
+    bytes x;
+    bytes z;
+    function f() public {
+        (x.push(), x.push()) = (0, 0);
+        (((x.push())), (x.push())) = (0, 0);
+        ((x.push(), x.push()), x.push()) = ((0, 0), 0);
+        (x.push(), x[0]) = (0, 0);
```
cameel, line 8: Isn't this one a false-positive too? `x[0]` is not a reference so it will be resolved correctly, whether it's before or after the push, right?

test/libsolidity/syntaxTests/tupleAssignments/warn_multiple_storage_storage_copies.sol
```diff
@@ -4,6 +4,11 @@ contract C {
 	function f() public {
 		(x, y) = (y, x);
```
cameel, line 5: Does it warn also when you do this?
```suggestion
		(x, y) = (x, y);
```

test/libsolidity/syntaxTests/array/bytes_assign_multi_fine.sol
```diff
@@ -0,0 +1,7 @@
+contract C {
+	bytes x;
+	function f() public {
+		(x[0], x[1]) = (0, 0);
```
cameel, line 4: Are these fine too?
- `(x[0], x[1]) = (x[0], x[1])`
- `(x[0], x[1]) = (x[1], x[0])`

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -134,6 +180,16 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
...
 		);
+
+	if (storageByteArrayPushes >= 1 && storageByteAccesses >= 2)
+		m_errorReporter.warning(
+			7239_error,
+			_assignment.location(),
+			"This assignment involves multiple accesses to a bytes array in storage while simultaneously enlarging it. "
+			"When a bytes array is enlarged, it may transition from short storage layout to long storage layout, "
+			"which invalidates all references to its elements. It is safer to only enlarge byte arrays in a single "
+			"operation one element at a time."
```
```suggestion
			"operation, one element at a time."
```

ekpyron:
Rebased.

cameel (commented):
I'm not done yet but pushing some of my comments to give you something to work on.

Overall, seems good so far. Just some new test cases and small typos.

test/libsolidity/syntaxTests/tupleAssignments/warn_multiple_storage_storage_copies.sol
ekpyron: Simultaneous reads to the same locations are what the warning is for, but it has false positives, so it will also warn about unrelated copies, i.e. for
```
uint[] x;
uint[] y;
uint[] z;
uint[] w;
function f(bool c) public {
   uint[] storage r = c ? y : z;
   uint[] storage r2 = c ? x : w;
   (x,y) = (r,r2);
}
```
it's impossible to tell at compile time if the case is fine or not, so it also warns for
```
uint[] x;
uint[] y;
uint[] z;
uint[] w;
function f() public {
   uint[] storage r = z;
   uint[] storage r2 = w;
   (x,y) = (r,r2);
}
```
even though a deep analysis could realize that it's fine.

test/libsolidity/syntaxTests/array/bytes_push_assign_multi.sol
ekpyron: ``x[0]`` is a reference, so this is not a false positive.
In this case
``(x.push(), x[0]) = (0,0);`` the same problem exists, just for ``x`` already being one element smaller.

I.e. in ``(x.push(), x[0])``, ``x[0]`` is a reference to the first element of ``x``, which may still be in the same slot as the length. After the ``push`` that element reference may become a reference to the length field, but the reference still refers to the length slot. So same situation really.

test/libsolidity/syntaxTests/tupleAssignments/warn_multiple_storage_storage_copies.sol
ekpyron: Yes, but that did not change here :-).

test/libsolidity/syntaxTests/array/bytes_assign_multi_fine.sol
ekpyron: Yes, in those cases, the rhs actually performs the accesses to the actual values, i.e. already reads from storage, before the left hand side is even evaluated.

cameel (approved):
I'd prefer to spend a bit more time to think through corner cases and potential test cases but with the time I had I did not manage to find any problems so it seems good to go given that we need to move on with the release.

PR #13140, opened 2022-06-14
Changed sed command in `zeppelin.sh` to match renamed test file in openzeppelin project.

matheusaaguiar:
After the merge of [openzeppelin-contracts#3467](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3467) file `ERC20TokenizedVault.test.js` was renamed to `ERC4626.test.js`. There are two `sed` commands in the `zeppelin.sh` script which referenced that file and needed to be changed accordingly.


cameel (approved):

ekpyron:
Looks like it passed - enabling auto-merge.

PR #13141, opened 2022-06-14
Add documentation section about dangling references.

ekpyron:
Not quite sure how much of this we need and how much details and best practices advice this should give, etc...

### DIFF
--- a/docs/types/reference-types.rst
+++ b/docs/types/reference-types.rst
@@ -468,6 +468,119 @@ Array Members
         }
     }
 
+.. index:: ! array;dangling storage references
+
+Dangling References to Storage Array Elements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When working with storage arrays, you need to take care to avoid dangling references.
+A dangling reference is a reference that points to something that no longer exists or has been
+moved without updating the reference. A dangling reference can for example occur, if you store a
+reference to an array element in a local variable and then ``.pop()`` from the containing array:
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.0 <0.9.0;
+
+    contract C {
+        uint[][] s;
+
+        function f() public {
+            // Stores a pointer to the last array element of s.
+            uint[] storage ptr = s[s.length - 1];
+            // Removes the last array element of s.
+            s.pop();
+            // Writes to the array element that is no longer within the array.
+            ptr.push(0x42);
+            // Adding a new element to ``s`` now will not add an empty array, but
+            // will result in an array of length 1 with ``0x42`` as element.
+            s.push();
+            assert(s[s.length - 1][0] == 0x42);
+        }
+    }
+
+The write in ``ptr.push(0x42)`` will **not** revert, despite the fact that ``ptr`` no
+longer refers to a valid element of ``s``. Since the compiler assumes that unused storage
+is always zeroed, a subsequent ``s.push()`` will not explicitly write zeroes to storage,
+so the last element of ``s`` after that ``push()`` will have length ``1`` and contain
+``0x42`` as its first element.
+
+Note that Solidity does not allow to declare references to value types in storage. These kinds
+of explicit dangling references are restricted to nested reference types. However, dangling references
+can also occur temporarily when using complex expressions in tuple assignments:
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.0 <0.9.0;
+
+    contract C {
+        uint[] s;
+        uint[] t;
+        constructor() {
+            // Push some initial values to the storage arrays.
+            s.push(0x07);
+            t.push(0x03);
+        }
+
+        function g() internal returns (uint[] storage) {
+            s.pop();
+            return t;
+        }
+
+        function f() public returns (uint[] memory) {
+            // The following will first evaluate ``s.push()`` to a reference to a new element
+            // at index 1. Afterwards, the call to ``g`` pops this new element, resulting in
+            // the left-most tuple element to become a dangling reference. The assignment still
+            // takes place and will write outside the data area of ``s``.
+            (s.push(), g()[0]) = (0x42, 0x17);
+            // A subsequent push to ``s`` will reveal the value written by the previous
+            // statement, i.e. the last element of ``s`` at the end of this function will have
+            // the value ``0x42``.
+            s.push();
+            return s;
+        }
+    }
+
+It is always safer to only assign to storage once per statement and to avoid
+complex expressions on the left-hand-side of an assignment.
+
+You need to take particular care when dealing with references to elements of
+``bytes`` arrays, since a ``.push()`` on a bytes array may switch :ref:`from short
+to long layout in storage<bytes-and-string>`.
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.0 <0.9.0;
+
+    contract C {
+        bytes x = "012345678901234567890123456789";
+
+        function test() external returns(uint) {
+            (x.push(), x.push()) = (0x01, 0x02);
+            return x.length;
+        }
+    }
+
+Here, when the first ``x.push()`` is evaluated, ``x`` is still stored in short
+layout, thereby ``x.push()`` returns a reference to an element in the first storage slot of
+``x``. However, the second ``x.push()`` switches the bytes array to large layout.
+Now the element that ``x.push()`` referred to is in the data area of the array while
+the reference still points at its original location, which is now a part of the length field
+and the assignment will effectively garble the length of ``x``.
+To be safe, only enlarge bytes arrays by at most one element during a single
+assignment and do not simultaneously index-access the array in the same statement.
+
+While the above describes the behaviour of dangling storage references in the
+current version of the compiler, any code with dangling references should be
+considered to have *undefined behaviour*. In particular, this means that
+any future version of the compiler may change the behaviour of code that
+involves dangling references.
+
+Be sure to avoid dangling references in your code!
+
 .. index:: ! array;slice
 
 .. _array-slices:


docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
...
+
+While the above describes the behaviour of dangling storage references in the
+current version of the compiler, any code with dangling references should be
+considered to have *undefined behaviour*. In particular, this means that
+any future version of the compiler may change the behaviour of code that
+involves dangling references.
+
+Be sure to avoid dangling references in your code!
+
+.. index:: ! array;dangling storage references
```
cameel: I think this is supposed to go somewhere close to the beginning of the section.

docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
 
 .. index:: ! array;slice
 
+Dangling References to Storage Array Elements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
cameel, line 474: Maybe we should mention the "bugs" related to this in the blog post and then link to this section?

docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
...
+        constructor() {
+           // Push some initial values to the storage arrays.
+           s.push(0x07);
+           t.push(0x03);
+        }
+
+        function g() internal returns (uint[] storage) {
+             s.pop();
+             return t;
+        }
```
cameel:
```suggestion
    contract C {
        uint[] s;
        uint[] t;
        constructor() {
            // Push some initial values to the storage arrays.
            s.push(0x07);
            t.push(0x03);
        }

        function g() internal returns (uint[] storage) {
            s.pop();
            return t;
        }
```

docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
...
+            return s;
+        }
+    }
+
+It is always safer to only assign to storage once per statement and to avoid
+complex expressions on the left-hand-side of an assignment.
+
+You need to take particular care when dealing with references to elements of
+``bytes`` arrays, since a ``.push()`` on a bytes array may switch from short
+to long layout in storage.
```
cameel: I think this should link to the docs that explain the short vs long layout.

docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
...
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.0 <0.9.0;
+
+   contract C {
```
cameel:
```suggestion
    contract C {
```

docs/types/reference-types.rst
```diff
@@ -470,6 +470,114 @@ Array Members
 
 .. index:: ! array;slice
 
+Dangling References to Storage Array Elements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When working with storage arrays, you need to take care to avoid dangling references.
+A dangling reference can for example occur, if you store a reference to an array element in
+a local variable and then ``.pop()`` from the containing array:
```
cameel: Might be worth explicitly mentioning that Solidity does not have a type that would represent a reference to a value type but that you can encounter such references as temporary values. The examples show it but I think it's nice to have these details spelled out in the text.

cameel: It would be good to have a more explicit definition of a dangling reference here. I.e. a reference that points at something that no longer exist or has been moved without updating the reference.

cameel (commented):
I only did a brief pass without reading it in detail. Here's what I found so far.

docs/types/reference-types.rst
ekpyron: I'm not sure I follow - for nested arrays, we *do* have storage reference types, and they are explicitly named in the example here.

docs/types/reference-types.rst
ekpyron: We could, but I'd also say we don't have to :-).

cameel: I think it would show that we took it seriously even if it's not really a bug that we can fix.

docs/types/reference-types.rst
cameel: But we do not have storage references to value types. You can have references to stuff in storage, but only to reference types. I think it would be good to point that out because it's not necessarily obvious and we sometimes rely on this. For example you need to know this to see why the impact of changing the string from short to long is limited to this weird `push()`/tuple case.

ekpyron: I added a sentence about this after the first example.

By the way: it's not only ``(byteArray.push(), byteArray.push()) = ...``, but it can also be
``(byteArray[0], byteArray.push()) = ...`` (or the other way round) - should I make that explicit or is it fine as is? The warnings in the other PR should hit both cases.

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+        }
+    }
+
+The write in ``ptr.push(0x42)`` will **not** revert, despite the fact that ``ptr`` no
+longer refers to a valid element of ``s``. Since the compiler assumes that unused storage
+is always zeroed, a subsequent ``s.push()`` will not explicitly write zeroes to storage,
+so the last element of ``s`` after that ``push()`` will have length ``1`` and contain
+``0x42`` as its first element.
+
+Note that Solidity does not allow to declare references to value types in storage, these kinds
```
cameel:
```suggestion
Note that Solidity does not allow to declare references to value types in storage. These kinds
```

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+
+    contract C {
+        uint[][] s;
+
+        function f() public {
+            // Stores a pointer to the last array element of s.
+            uint[] storage ptr = s[s.length - 1];
+            // Removes the last array element of s.
+            s.pop();
+            // Writes to the array element that is no longer at a valid index in s.
```
cameel: How about this?
```suggestion
            // Writes to the array element that is no longer within the array.
```

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+            t.push(0x03);
+        }
+
+        function g() internal returns (uint[] storage) {
+            s.pop();
+            return t;
+        }
+
+        function f() public returns (uint[] memory) {
+            // The following will first evaluate ``s.push()`` to a reference to a new element at index 1.
```
cameel: This comment is a bit too long to fit in code snippet window in docs and adds a scrollbar. I think it would be better to wrap just after `at`.

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+        bytes x = "012345678901234567890123456789";
+
+        function test() external returns(uint) {
+            (x.push(), x.push()) = (0x01, 0x02);
+            return x.length;
+        }
+    }
+
+Here, when the first ``x.push()`` is evaluated, ``x`` is still stored in short
+layout, thereby ``x.push()`` refers to an element in the first storage slot of
```
cameel:
```suggestion
layout, thereby ``x.push()`` returns a reference to an element in the first storage slot of
```

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+``x``. However, the second ``x.push()`` switches the bytes array to large layout.
+Now the reference created by the first ``x.push()`` refers to parts of the length
+field of ``x``, so the assignment will effectively garble the length of ``x``.
```
cameel: How about this?
```suggestion
Now the element that ``x.push()`` referred to is in the data area of the array while
the reference still points at its original location, which is now a part of the length field
and the assignment will effectively garble the length of ``x``.
```

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+            // Stores a pointer to the last array element of s.
+            uint[] storage ptr = s[s.length - 1];
+            // Removes the last array element of s.
+            s.pop();
+            // Writes to the array element that is no longer at a valid index in s.
+            ptr.push(0x42);
+            // Adding a new element to ``s`` now will not add an empty array, but
+            // will result in an array of length 1 with ``0x42`` as element.
+            s.push();
+            assert(s[s.length - 1][0] == 0x42);
```
cameel, line 499: This part feels like it comes out of nowhere since the text before only mentions dangling references and this feels like a separate issue - even if you understand dangling references you might not be aware that `push()` does not zero memory.

I think that we should first give an example showing that a dangling reference can let you write outside of the array and then, separately, show another one with `push()` to give it the proper emphasis.

But since we're pressed for time, I think this is something that could be changed after the release rather than in this PR.

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
         }
     }
 
+.. index:: ! array;dangling storage references
+
+Dangling References to Storage Array Elements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When working with storage arrays, you need to take care to avoid dangling references.
```
cameel, line 476: What about memory arrays? There's really nothing outside of inline assembly that can give you a dangling reference in memory?

docs/types/reference-types.rst
```diff
@@ -468,6 +468,117 @@ Array Members
...
+
+The write in ``ptr.push(0x42)`` will **not** revert, despite the fact that ``ptr`` no
+longer refers to a valid element of ``s``. Since the compiler assumes that unused storage
+is always zeroed, a subsequent ``s.push()`` will not explicitly write zeroes to storage,
+so the last element of ``s`` after that ``push()`` will have length ``1`` and contain
+``0x42`` as its first element.
+
+Note that Solidity does not allow to declare references to value types in storage, these kinds
+of explicit dangling references are restricted to nested reference types. However, dangling references
+can also occur temporarily when using complex expressions in tuple assignments:
```
cameel, line 511: It's not really **only** tuples, right? For example `s.push() = s.push() = 0x42` would also break things this way, right?

cameel (dismissed):
There are some things that might need slight adjustment but overall they're don't seem critical so I guess it would even to fine to merge it as is.

docs/types/reference-types.rst
ekpyron: This is the entire point... if this assertion would not hold in the end as a result of the dangling references, none of this would be particularly noteworthy :-).

docs/types/reference-types.rst
ekpyron: Nope, I don't think so.

docs/types/reference-types.rst
ekpyron: I think this really resolves to
```
bytes1 x = (s.push() = 0x42);
s.push() = x;
```
And thereby is fine (just checked and didn't get a mangled length in this case).

So the only cases I'm aware of do involve tuples... but I'm not entirely sure I'm not missing anything either.

cameel (approved):

docs/types/reference-types.rst
cameel: Is it? It's one of the more serious consequences but I think that dangling references would be problem even if `push()` was zeroing storage.

docs/types/reference-types.rst
cameel: Ah, right. For this to break you need something that will evaluate pushes first and only then do the assignment.

PR #13145, opened 2022-06-15
Fix GCC compilation errors due to Boost 1.79 update

ekpyron:
Replaces and closes #13090.

Attempt to fix https://app.circleci.com/pipelines/github/ethereum/solidity/24748/workflows/e6c3ff3b-a9f2-4832-aa2f-d64b66f03c5b/jobs/1089693

### DIFF
--- a/libsolutil/Numeric.h
+++ b/libsolutil/Numeric.h
@@ -33,6 +33,9 @@
 #if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#pragma GCC diagnostic ignored "-Wstringop-overread"
+#pragma GCC diagnostic ignored "-Waggressive-loop-optimizations"
 #endif
 #include <boost/multiprecision/cpp_int.hpp>
 #if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
--- a/libevmasm/GasMeter.cpp
+++ b/libevmasm/GasMeter.cpp
@@ -184,7 +184,7 @@ GasMeter::GasConsumption GasMeter::estimateMax(AssemblyItem const& _item, bool _
 				if (*value)
 				{
 					// Note: msb() counts from 0 and throws on 0 as input.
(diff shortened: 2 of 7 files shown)


cameel:
Interesting. The changes are pretty much the same as in #13090. So it apparently wasn't just a BSD problem after all. The PR said it's because of Boost 1.77 but I dismissed it because I was already running 1.78 and it wasn't happening for me. Now I see that Boost 1.79 just hit Arch repos and, looking closer at the comment from the contributor, he later said he's actually using 1.79.

ekpyron:
My local builds just passed, so I hope this now was all of them.

ekpyron:
> Interesting. The changes are pretty much the same as in #13090. So it apparently wasn't just a BSD problem after all. The PR said it's because of Boost 1.77 but I dismissed it because I was already running 1.78 and it wasn't happening for me. Now I see that Boost 1.79 just hit Arch repos and, looking closer at the comment from the contributor, he later said he's actually using 1.79.

Yeah... I was actually hoping that the next boost release played nicer with gcc 12, but apparently they still only support gcc 11 and it actually made things worse :-).
Might also be the reason for the arch update taking longer than usual for it.

cameel (approved):
Changes look harmless so I think it's good if it passes tests.

libsolutil/Numeric.h
```diff
@@ -33,6 +33,9 @@
 #if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 12)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#pragma GCC diagnostic ignored "-Wstringop-overread"
+#pragma GCC diagnostic ignored "-Waggressive-loop-optimizations"
 #endif
```
ekpyron, line 39: At some point we should really figure out why we need this at all - if the build was set up properly, it should ignore issues inside system headers in general and we shouldn't have to disable these here in particular... might just be that it's hard for gcc to tell where the error is actually coming from, since all of these happen in template instantiations...

Anyways, I'd just keep it like this for now.

cameel: But if it's the templates, GCC 11 should have had the same problem...

ekpyron: It's GCC 12-only warnings that occur in template instantiations of boost multiprecision templates (and boost also apparently changed some of those template definitions in 1.79).

PR #13150, opened 2022-06-15
LSP analyze all files in project

christianparpart:
Analyzes all files within a opened project. That includes those that are not opened. This is the default option (as it seems to match what other LSPs do) but can be explicitly turned off using custom settings object that can be passed during initialization.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ Language Features:
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
 
 
 Bugfixes:
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -58,14 +58,15 @@ namespace fs = boost::filesystem;
 namespace
 {
 
-bool resolvesToRegularFile(boost::filesystem::path _path)
+bool resolvesToRegularFile(boost::filesystem::path _path, int maxRecursionDepth = 10)
 {
 	fs::file_status fileStatus = fs::status(_path);
 
-	while (fileStatus.type() == fs::file_type::symlink_file)
+	while (fileStatus.type() == fs::file_type::symlink_file && maxRecursionDepth > 0)
 	{
 		_path = boost::filesystem::read_symlink(_path);
 		fileStatus = fs::status(_path);
+		maxRecursionDepth--;
 	}
 
 	return fileStatus.type() == fs::file_type::regular_file;
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -53,8 +53,23 @@ using namespace solidity::lsp;
 using namespace solidity::langutil;
 using namespace solidity::frontend;
 
+namespace fs = boost::filesystem;
+
 namespace
 {
+
+bool resolvesToRegularFile(boost::filesystem::path _path)
+{
+	fs::file_status fileStatus = fs::status(_path);
+
+	while (fileStatus.type() == fs::file_type::symlink_file)
+	{
+		_path = boost::filesystem::read_symlink(_path);
+		fileStatus = fs::status(_path);
+	}
+
+	return fileStatus.type() == fs::file_type::regular_file;
+}
 
 int toDiagnosticSeverity(Error::Type _errorType)
 {
@@ -198,22 +213,18 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 
 vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
 {
-	namespace fs = boost::filesystem;
-
-	std::vector<fs::path> collectedPaths{};
+	vector<fs::path> collectedPaths{};
 
 	// We explicitly decided against including all files from include paths but leave the possibility
 	// open for a future PR to enable such a feature to be optionally enabled (default disabled).
 
 	auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
 	for (fs::directory_entry const& dirEntry: directoryIterator)
-	{
 		if (
-			dirEntry.status().type() == fs::file_type::regular_file &&
-			dirEntry.path().extension() == ".sol"
+			dirEntry.path().extension() == ".sol" &&
+			(dirEntry.status().type() == fs::file_type::regular_file || resolvesToRegularFile(dirEntry.path()))
 		)
-		collectedPaths.push_back(dirEntry.path());
-	}
+			collectedPaths.push_back(dirEntry.path());
 
 	return collectedPaths;
 }
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -207,8 +207,13 @@ vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() co
 
 	auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
 	for (fs::directory_entry const& dirEntry: directoryIterator)
-		if (dirEntry.path().extension() == ".sol")
-			collectedPaths.push_back(dirEntry.path());
+	{
+		if (
+			dirEntry.status().type() == fs::file_type::regular_file &&
+			dirEntry.path().extension() == ".sol"
+		)
+		collectedPaths.push_back(dirEntry.path());
+	}
 
 	return collectedPaths;
 }
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -32,6 +32,7 @@
 #include <liblangutil/SourceReferenceExtractor.h>
 #include <liblangutil/CharStream.h>
 
+#include <libsolutil/CommonIO.h>
 #include <libsolutil/Visitor.h>
 #include <libsolutil/JSON.h>
 
@@ -41,6 +42,8 @@
 
 #include <ostream>
 #include <string>
+
+#include <fmt/format.h>
 
 using namespace std;
 using namespace std::string_literals;
@@ -118,7 +121,7 @@ LanguageServer::LanguageServer(Transport& _transport):
 		{"cancelRequest", [](auto, auto) {/*nothing for now as we are synchronous */}},
 		{"exit", [this](auto, auto) { m_state = (m_state == State::ShutdownRequested ? State::ExitRequested : State::ExitWithoutShutdown); }},
 		{"initialize", bind(&LanguageServer::handleInitialize, this, _1, _2)},
-		{"initialized", [](auto, auto) {}},
+		{"initialized", bind(&LanguageServer::handleInitialized, this, _1, _2)},
 		{"$/setTrace", [this](auto, Json::Value const& args) { setTrace(args["value"]); }},
 		{"shutdown", [this](auto, auto) { m_state = State::ShutdownRequested; }},
 		{"textDocument/definition", GotoDefinition(*this) },
@@ -147,6 +150,26 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
+	// The settings item: "file-load-strategy" (enum) defaults to "project-directory" if not (or not correctly) set.
+	// It can be overridden during client's handshake or at runtime, as usual.
+	//
+	// If this value is set to "project-directory" (default), all .sol files located inside the project directory or reachable through symbolic links will be subject to operations.
+	//
+	// Operations include compiler analysis, but also finding all symbolic references or symbolic renaming.
+	//
+	// If this value is set to "directly-opened-and-on-import", then only currently directly opened files and
+	// those files being imported directly or indirectly will be included in operations.
+	if (_settings["file-load-strategy"])
+	{
+		auto const text = _settings["file-load-strategy"].asString();
+		if (text == "project-directory")
+			m_fileLoadStrategy = FileLoadStrategy::ProjectDirectory;
+		else if (text == "directly-opened-and-on-import")
+			m_fileLoadStrategy = FileLoadStrategy::DirectlyOpenedAndOnImported;
+		else
+			lspAssert(false, ErrorCode::InvalidParams, "Invalid file load strategy: " + text);
+	}
+
 	m_settingsObject = _settings;
 	Json::Value jsonIncludePaths = _settings["include-paths"];
 
@@ -173,6 +196,23 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 	}
 }
 
+vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
+{
+	namespace fs = boost::filesystem;
+
+	std::vector<fs::path> collectedPaths{};
+
+	// We explicitly decided against including all files from include paths but leave the possibility
+	// open for a future PR to enable such a feature to be optionally enabled (default disabled).
+
+	auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
+	for (fs::directory_entry const& dirEntry: directoryIterator)
+		if (dirEntry.path().extension() == ".sol")
+			collectedPaths.push_back(dirEntry.path());
+
+	return collectedPaths;
+}
+
 void LanguageServer::compile()
 {
 	// For files that are not open, we have to take changes on disk into account,
@@ -181,6 +221,18 @@ void LanguageServer::compile()
 	FileRepository oldRepository(m_fileRepository.basePath(), m_fileRepository.includePaths());
 	swap(oldRepository, m_fileRepository);
 
+	// Load all solidity files from project.
+	if (m_fileLoadStrategy == FileLoadStrategy::ProjectDirectory)
+		for (auto const& projectFile: allSolidityFilesFromProject())
+		{
+			lspDebug(fmt::format("adding project file: {}", projectFile.generic_string()));
+			m_fileRepository.setSourceByUri(
+				m_fileRepository.sourceUnitNameToUri(projectFile.generic_string()),
+				util::readFileAsString(projectFile)
+			);
+		}
+
+	// Overwrite all files as opened by the client, including the ones which might potentially have changes.
 	for (string const& fileName: m_openFiles)
 		m_fileRepository.setSourceByUri(
 			fileName,
@@ -269,6 +321,7 @@ bool LanguageServer::run()
 			{
 				string const methodName = (*jsonMessage)["method"].asString();
 				id = (*jsonMessage)["id"];
+				lspDebug(fmt::format("received method call: {}", methodName));
 
 				if (auto handler = util::valueOrDefault(m_handlers, methodName))
 					handler(id, (*jsonMessage)["params"]);
@@ -277,6 +330,10 @@ bool LanguageServer::run()
 			}
 			else
 				m_client.error({}, ErrorCode::ParseError, "\"method\" has to be a string.");
+		}
+		catch (Json::Exception const&)
+		{
+			m_client.error(id, ErrorCode::InvalidParams, "JSON object access error. Most likely due to a badly formatted JSON request message."s);
 		}
 		catch (RequestError const& error)
 		{
@@ -345,6 +402,12 @@ void LanguageServer::handleInitialize(MessageID _id, Json::Value const& _args)
 	replyArgs["capabilities"]["renameProvider"] = true;
 
 	m_client.reply(_id, move(replyArgs));
+}
+
+void LanguageServer::handleInitialized(MessageID, Json::Value const&)
+{
+	if (m_fileLoadStrategy == FileLoadStrategy::ProjectDirectory)
+		compileAndUpdateDiagnostics();
 }
 
 void LanguageServer::semanticTokensFull(MessageID _id, Json::Value const& _args)
--- a/libsolidity/lsp/LanguageServer.h
+++ b/libsolidity/lsp/LanguageServer.h
@@ -37,6 +37,23 @@ class RenameSymbol;
 enum class ErrorCode;
 
 /**
+ * Enum to mandate what files to take into consideration for source code analysis.
+ */
+enum class FileLoadStrategy
+{
+	/// Takes only those files into consideration that are explicitly opened and those
+	/// that have been directly or indirectly imported.
+	DirectlyOpenedAndOnImported = 0,
+
+	/// Takes all Solidity (.sol) files within the project root into account.
+	/// Symbolic links will be followed, even if they lead outside of the project directory
+	/// (`--allowed-paths` is currently ignored by the LSP).
+	///
+	/// This resembles the closest what other LSPs should be doing already.
+	ProjectDirectory = 1,
+};
+
+/**
  * Solidity Language Server, managing one LSP client.
  * This implements a subset of LSP version 3.16 that can be found at:
  * https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/
@@ -68,6 +85,7 @@ private:
 	/// Reports an error and returns false if not.
 	void requireServerInitialized();
 	void handleInitialize(MessageID _id, Json::Value const& _args);
+	void handleInitialized(MessageID _id, Json::Value const& _args);
 	void handleWorkspaceDidChangeConfiguration(Json::Value const& _args);
 	void setTrace(Json::Value const& _args);
 	void handleTextDocumentDidOpen(Json::Value const& _args);
@@ -82,6 +100,9 @@ private:
 
 	/// Compile everything until after analysis phase.
 	void compile();
+
+	std::vector<boost::filesystem::path> allSolidityFilesFromProject() const;
+
 	using MessageHandler = std::function<void(MessageID, Json::Value const&)>;
 
 	Json::Value toRange(langutil::SourceLocation const& _location);
@@ -100,6 +121,7 @@ private:
 	/// Set of source unit names for which we sent diagnostics to the client in the last iteration.
 	std::set<std::string> m_nonemptyDiagnostics;
 	FileRepository m_fileRepository;
+	FileLoadStrategy m_fileLoadStrategy = FileLoadStrategy::ProjectDirectory;
 
 	frontend::CompilerStack m_compilerStack;
 
--- a/libsolutil/CommonIO.h
+++ b/libsolutil/CommonIO.h
@@ -48,46 +48,6 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
 namespace util
 {
 
-namespace detail
-{
-
-template <typename Predicate>
-struct RecursiveFileCollector
-{
-	Predicate predicate;
-	std::vector<boost::filesystem::path> result {};
-
-	RecursiveFileCollector& operator()(boost::filesystem::path const& _directory)
-	{
-		if (!boost::filesystem::is_directory(_directory))
-			return *this;
-		auto iterator = boost::filesystem::directory_iterator(_directory);
-		auto const iteratorEnd = boost::filesystem::directory_iterator();
-
-		while (iterator != iteratorEnd)
-		{
-			if (boost::filesystem::is_directory(iterator->status()))
-				(*this)(iterator->path());
-
-			if (predicate(iterator->path()))
-				result.push_back(iterator->path());
-
-			++iterator;
-		}
-		return *this;
-	}
-};
-
-template <typename Predicate>
-RecursiveFileCollector(Predicate) -> RecursiveFileCollector<Predicate>;
-}
-
-template <typename Predicate>
-std::vector<boost::filesystem::path> findFilesRecursively(boost::filesystem::path const& _rootDirectory, Predicate _predicate)
-{
-	return detail::RecursiveFileCollector{_predicate}(_rootDirectory).result;
-}
-
 /// Retrieves and returns the contents of the given file as a std::string.
 /// If the file doesn't exist, it will throw a FileNotFound exception.
 /// If the file exists but is not a regular file, it will throw NotAFile exception.
--- a/libsolutil/CommonIO.h
+++ b/libsolutil/CommonIO.h
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
 namespace util
 {
 
+namespace detail
+{
+
+template <typename Predicate>
+struct RecursiveFileCollector
+{
+	Predicate predicate;
+	std::vector<boost::filesystem::path> result {};
+
+	RecursiveFileCollector& operator()(boost::filesystem::path const& _directory)
+	{
+		if (!boost::filesystem::is_directory(_directory))
+			return *this;
+		auto iterator = boost::filesystem::directory_iterator(_directory);
+		auto const iteratorEnd = boost::filesystem::directory_iterator();
+
+		while (iterator != iteratorEnd)
+		{
+			if (boost::filesystem::is_directory(iterator->status()))
+				(*this)(iterator->path());
+
+			if (predicate(iterator->path()))
+				result.push_back(iterator->path());
+
+			++iterator;
+		}
+		return *this;
+	}
+};
+
+template <typename Predicate>
+RecursiveFileCollector(Predicate) -> RecursiveFileCollector<Predicate>;
+}
+
+template <typename Predicate>
+std::vector<boost::filesystem::path> findFilesRecursively(boost::filesystem::path const& _rootDirectory, Predicate _predicate)
+{
+	return detail::RecursiveFileCollector{_predicate}(_rootDirectory).result;
+}
+
 /// Retrieves and returns the contents of the given file as a std::string.
 /// If the file doesn't exist, it will throw a FileNotFound exception.
 /// If the file exists but is not a regular file, it will throw NotAFile exception.
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -539,16 +539,9 @@ class TestParser:
         return self.current_line_tuple is None
 
 class FileLoadStrategy(Enum):
-    Undefined = auto()
-    ProjectDirectory = auto()
-    DirectlyOpenedAndOnImport = auto()
-
-    def lsp_name(self):
-        if self == FileLoadStrategy.ProjectDirectory:
-            return 'project-directory'
-        elif self == FileLoadStrategy.DirectlyOpenedAndOnImport:
-            return 'directly-opened-and-on-import'
-        return None
+    Undefined = None
+    ProjectDirectory = 'project-directory'
+    DirectlyOpenedAndOnImport =  'directly-opened-and-on-import'
 
 class FileTestRunner:
     """
@@ -932,7 +925,6 @@ class SolidityLSPTestSuite: # {{{
             # Enable traces to receive the amount of expected diagnostics before
             # actually receiving them.
             'trace': 'messages',
-            # 'initializationOptions': {},
             'capabilities': {
                 'textDocument': {
                     'publishDiagnostics': {'relatedInformation': True}
@@ -949,7 +941,7 @@ class SolidityLSPTestSuite: # {{{
 
         if file_load_strategy != FileLoadStrategy.Undefined:
             params['initializationOptions'] = {}
-            params['initializationOptions']['file-load-strategy'] = file_load_strategy.lsp_name()
+            params['initializationOptions']['file-load-strategy'] = file_load_strategy.value
 
         if custom_include_paths is not None and len(custom_include_paths) != 0:
             if params['initializationOptions'] is None:
@@ -1337,7 +1329,7 @@ class SolidityLSPTestSuite: # {{{
     # }}}
 
     # {{{ actual tests
-    def test_analyze_all_project_files1(self, solc: JsonRpcProcess) -> None:
+    def test_analyze_all_project_files_flat(self, solc: JsonRpcProcess) -> None:
         """
         Tests the option (default) to analyze all .sol project files even when they have not been actively
         opened yet. This is how other LSPs (at least for C++) work too and it makes cross-unit tasks
@@ -1370,18 +1362,18 @@ class SolidityLSPTestSuite: # {{{
         self.expect_equal(report['uri'], self.get_test_file_uri('E', SUBDIR), "Correct file URI")
         self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
 
-    def test_analyze_all_project_files2(self, solc: JsonRpcProcess) -> None:
+    def test_analyze_all_project_files_nested(self, solc: JsonRpcProcess) -> None:
         """
         Same as first test on that matter but with deeper nesting levels.
         """
         SUBDIR = 'include-paths-nested'
-        EXPECTED_FILES = [
+        EXPECTED_FILES = {
             "A/B/C/foo",
             "A/B/foo",
             "A/foo",
             "foo",
-        ]
-        EXPECTED_URIS = [self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES]
+        }
+        EXPECTED_URIS = {self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES}
         self.setup_lsp(
             solc,
             file_load_strategy=FileLoadStrategy.ProjectDirectory,
@@ -1389,23 +1381,22 @@ class SolidityLSPTestSuite: # {{{
         )
         published_diagnostics = self.wait_for_diagnostics(solc)
         self.expect_equal(len(published_diagnostics), len(EXPECTED_FILES), "Test number of files analyzed.")
-        for report in published_diagnostics:
-            self.expect_true(report['uri'] in EXPECTED_URIS, "Correct file URI")
-            self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+        self.expect_equal({report['uri'] for report in published_diagnostics}, EXPECTED_URIS)
+        self.expect_equal([len(report['diagnostics']) for report in published_diagnostics], [0] * len(EXPECTED_URIS))
 
-    def test_analyze_all_project_files3(self, solc: JsonRpcProcess) -> None:
+    def test_analyze_all_project_files_nested_with_include_paths(self, solc: JsonRpcProcess) -> None:
         """
         Same as first test on that matter but with deeper nesting levels.
         """
         SUBDIR = 'include-paths-nested-2'
-        EXPECTED_FILES = [
+        EXPECTED_FILES = {
             "A/B/C/foo",
             "A/B/foo",
             "A/foo",
             "foo",
-        ]
+        }
         IMPLICITLY_LOADED_FILE_COUNT = 1
-        EXPECTED_URIS = [self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES]
+        EXPECTED_URIS = {self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES}
         self.setup_lsp(
             solc,
             file_load_strategy=FileLoadStrategy.ProjectDirectory,
@@ -1426,9 +1417,9 @@ class SolidityLSPTestSuite: # {{{
 
         # Check last report (should be the custom imported lib).
         # This file is analyzed because it was imported via "A/B/C/foo.sol".
-        report = published_diagnostics[len(EXPECTED_URIS)]
-        self.expect_equal(report['uri'], f"{self.project_root_uri}/other-include-dir/otherlib/second.sol", "Correct file URI")
-        self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+        last_report = published_diagnostics[len(EXPECTED_URIS)]
+        self.expect_equal(last_report['uri'], self.get_test_file_uri('second', 'other-include-dir/otherlib'), "Correct file URI")
+        self.expect_equal(len(last_report['diagnostics']), 0, "no diagnostics")
 
 
     def test_publish_diagnostics_errors_multiline(self, solc: JsonRpcProcess) -> None:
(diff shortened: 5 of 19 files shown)


Marenz:
Is this PR still relevant or was it superceeded?

christianparpart:
> Is this PR still relevant or was it superceeded?

@Marenz relevant. and soon ready :)

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
+};
+
+template <typename Predicate>
+RecursiveFileCollector(Predicate) -> RecursiveFileCollector<Predicate>;
+}
+
+template <typename Predicate>
+std::vector<boost::filesystem::path> findFilesRecursively(boost::filesystem::path const& _rootDirectory, Predicate _predicate)
+{
+	return detail::RecursiveFileCollector{_predicate}(_rootDirectory).result;
```
cameel:
```suggestion
	return detail::RecursiveFileCollector{std::move(_predicate)}(_rootDirectory).result;
```

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
 {
 
+namespace detail
+{
+
+template <typename Predicate>
+struct RecursiveFileCollector
+{
+	Predicate predicate;
+	std::vector<boost::filesystem::path> result {};
```
cameel: It looks like an empty code block if you write it this way :)
```suggestion
	std::vector<boost::filesystem::path> result{};
```

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
+
+	RecursiveFileCollector& operator()(boost::filesystem::path const& _directory)
+	{
+		if (!boost::filesystem::is_directory(_directory))
+			return *this;
+		auto iterator = boost::filesystem::directory_iterator(_directory);
```
cameel: Why not use `recursive_directory_iterator` instead of doing the recursion by hand?

Alternatively, we also have a reusable [`BreadthFirstSearch`](https://github.com/ethereum/solidity/blob/v0.8.15/libsolutil/Algorithms.h#L80-L126) implementation that could be used.

libsolutil/CommonIO.h
cameel:
```suggestion
	std::vector<boost::filesystem::path> collectedPaths{};
```

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
+		auto const iteratorEnd = boost::filesystem::directory_iterator();
+
+		while (iterator != iteratorEnd)
+		{
+			if (boost::filesystem::is_directory(iterator->status()))
+				(*this)(iterator->path());
+
+			if (predicate(iterator->path()))
+				result.push_back(iterator->path());
```
cameel: For me it would be more intuitive to include a directory before its content. Is there any reason why it's the other way around?

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
+	{
+		if (!boost::filesystem::is_directory(_directory))
+			return *this;
+		auto iterator = boost::filesystem::directory_iterator(_directory);
+		auto const iteratorEnd = boost::filesystem::directory_iterator();
+
+		while (iterator != iteratorEnd)
+		{
+			if (boost::filesystem::is_directory(iterator->status()))
+				(*this)(iterator->path());
```
cameel: You already have a check against it being a dir at the beginning of the function. This one seems superfluous and, given how heavy disk operations typically are, removing it should save more than we lose on an extra function call.
```suggestion
			(*this)(iterator->path());
```

Changelog.md
```diff
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+ * LSP: Analyzes all files in a project by default (can be customized by setting ``'analyze-all-files-in-project'`` to ``false`` in LSP settings object).
```
cameel: We seem to have a mix of `LSP` and `Language Server` in the changelog. We should settle on one :)

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -147,6 +150,16 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
+	try
+	{
+		if (_settings["analyze-all-files-in-project"])
+			m_analyzeAllFilesFromProject = _settings["analyze-all-files-in-project"].asBool();
+	}
+	catch (...)
+	{
+		throw RequestError{ErrorCode::InvalidParams};
```
cameel: What specific exception are you expecting here? I think it would be best to handle only that exception and let others bubble up to `run()` (which should be one central place to deal with unexpected errors).

Also, before this happens, `run()` will log the invalid parameters, right? If not, you should log them here.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -147,6 +150,16 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
+	try
+	{
+		if (_settings["analyze-all-files-in-project"])
+			m_analyzeAllFilesFromProject = _settings["analyze-all-files-in-project"].asBool();
```
cameel: We should document this somewhere. Preferably in docs but for now at least a comment explaining its purpose would be the bare minimum.

Especially the default value should be clearly documented.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +186,18 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
 
+vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
+{
+	// Don't iterate through all .sol files recursively when project root is system root.
+	if (m_fileRepository.basePath() == "/")
+		return {};
+
+	return util::findFilesRecursively(
+		m_fileRepository.basePath(),
+		[&](boost::filesystem::path const& p) { return p.extension() == ".sol"; }
```
```suggestion
		[](boost::filesystem::path const& p) { return p.extension() == ".sol"; }
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +186,18 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 	}
 }
 
+vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
+{
+	// Don't iterate through all .sol files recursively when project root is system root.
+	if (m_fileRepository.basePath() == "/")
+		return {};
```
cameel: This does not look like a robust safeguard.
First - will it still be `/` on Windows or `\\` or something like `C:`?
Second what is the actual thing you're trying to guard against? If it's that it the filesystem hierarchy can be deep? There are many other dirs like that, so maybe we should limit depth instead. Or if it's about access to system files, then the the iterator will throw when trying to access a dir that user does not have permissions for, won't it?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +186,18 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+
+	return util::findFilesRecursively(
+		m_fileRepository.basePath(),
```
cameel: What about files from include dirs?

And files accessible through symlinks? I think that `is_directory()` does not return `true` for symlinked dirs so `findFilesRecursively()` will skip them.

libsolutil/CommonIO.h
```diff
@@ -48,6 +48,46 @@ inline std::ostream& operator<<(std::ostream& os, bytes const& _bytes)
...
+		return *this;
+	}
+};
+
+template <typename Predicate>
+RecursiveFileCollector(Predicate) -> RecursiveFileCollector<Predicate>;
+}
+
+template <typename Predicate>
+std::vector<boost::filesystem::path> findFilesRecursively(boost::filesystem::path const& _rootDirectory, Predicate _predicate)
```
cameel: This needs a docstring. It should at least say how errors are reported. And how the dirs are traversed - is any order guaranteed, are inaccessible dirs and broken links skipped, does it follow symlinks, are symlinked and hardlinked files included, etc.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -181,6 +206,18 @@ void LanguageServer::compile()
 	FileRepository oldRepository(m_fileRepository.basePath(), m_fileRepository.includePaths());
 	swap(oldRepository, m_fileRepository);
 
+	// Load all solidity files from project.
+	if (m_analyzeAllFilesFromProject)
+		for (auto const& projectFile: allSolidityFilesFromProject())
```
cameel, line 243: How is performance of rescanning the dir on every compilation? Is compilation and file read itself always heavy enough for it not to matter or can it be a bottleneck?

cameel: Actually, just recompiling all files in the project every time is probably very heavy. I haven't tried running it myself but judging by what I've seen with external tests, I'd expect recompiling real-life codebases like OpenZeppelin or Uniswap via IR this way to make the LSP unusable when this feature is enabled.

Seems like not recompiling unchanged files would be pretty crucial for this to work efficiently. Maybe it would make more sense to have the feature disabled by default until that's implemented?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -181,6 +206,18 @@ void LanguageServer::compile()
...
+		{
+			lspDebug(fmt::format("adding project file: {}", projectFile.generic_string()));
+			m_fileRepository.setSourceByUri(
+				m_fileRepository.sourceUnitNameToUri(projectFile.generic_string()),
+				util::readFileAsString(projectFile)
+			);
+		}
+
+	// Overwrite all files as opened by the client and might potentially have changes.
```
cameel: Seems like something is missing in this sentence:
```suggestion
	// Overwrite all files as opened by the client, including the ones which might potentially have changes.
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -347,6 +385,13 @@ void LanguageServer::handleInitialize(MessageID _id, Json::Value const& _args)
 	m_client.reply(_id, move(replyArgs));
 }
 
+void LanguageServer::handleInitialized(MessageID, Json::Value const&)
+{
+	lspDebug(fmt::format("handle initialized notification: {}", m_analyzeAllFilesFromProject ? "analyze-all-files-in-project" : "analyze-files-on-demand"));
```
cameel: Actually, maybe this setting would be better of as an enum? It would be easier to add more strategies later and it would be more readable. For example:
```
enum FileLoadStrategy
    PreloadProjectDir
    OnImport
```

test/lsp.py
```diff
@@ -898,18 +899,29 @@ def main(self) -> int:
 
         return min(max(self.test_counter.failed, self.assertion_counter.failed), 127)
 
-    def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root=True):
+    def setup_lsp(
+        self,
+        lsp: JsonRpcProcess,
+        expose_project_root=True,
+        analyze_all_files_in_project=False,
+        project_root_suffix = None
```
```suggestion
        project_root_suffix=None
```

cameel: Or with annotations:

```suggestion
        lsp: JsonRpcProcess,
        expose_project_root: bool = True,
        analyze_all_files_in_project: bool = False,
        project_root_suffix: Optional[str] = None
```

And yeah, I know it's weird to have different spacing based on whether you have an annotation or not but [that's what PEP8 actually recommends](https://peps.python.org/pep-0008/#other-recommendations). Personally, I'd be for just always having them there :)

test/lsp.py
```diff
@@ -898,18 +899,29 @@ def main(self) -> int:
...
             'processId': None,
-            'rootUri': self.project_root_uri,
+            'rootUri': rootUri,
             # Enable traces to receive the amount of expected diagnostics before
             # actually receiving them.
             'trace': 'messages',
-            'initializationOptions': {},
+            'initializationOptions': {
+                'analyze-all-files-in-project': analyze_all_files_in_project,
+            },
```
cameel: We should also have the case when this setting is not present covered.

test/lsp.py
```diff
@@ -1296,6 +1308,40 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
 
     # {{{ actual tests
 
+    def test_analyze_all_project_files(self, solc: JsonRpcProcess) -> None:
+        """
+        Tests the option (default) to analyze all .sol project files even when they have not been actively
+        opened yet. This is how other LSP's (at least for C++) work too and it makes cross-unit tasks
```
```suggestion
        opened yet. This is how other LSPs (at least for C++) work too and it makes cross-unit tasks
```

test/lsp.py
```diff
@@ -1296,6 +1308,40 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+        self.expect_equal(report['uri'], self.get_test_file_uri('C', SUBDIR), "Correct file URI")
+        self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+        # D.sol
+        report = published_diagnostics[1]
+        self.expect_equal(report['uri'], self.get_test_file_uri('D', SUBDIR), "Correct file URI")
+        self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+        # E.sol
+        report = published_diagnostics[2]
```
cameel, line 1361: You should also check that no other files were found.

test/lsp.py
```diff
@@ -1296,6 +1308,40 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
 
     # {{{ actual tests
 
+    def test_analyze_all_project_files(self, solc: JsonRpcProcess) -> None:
```
cameel: Can you add a more complex test, that would also have some include dirs, and some files that should **not** be detected?

Preferably also symlinks but I guess it would not be a good idea to store these in the repo so you'd have to do that in a temp dir... Does the current testing framework support that?

cameel: We should also have a test for the case of `analyze_all_files_in_project=False`. Overall, I think this is a bit undertested given how many corner cases there could be.

test/lsp.py
```diff
@@ -898,18 +899,29 @@ def main(self) -> int:
...
+        analyze_all_files_in_project=False,
+        project_root_suffix = None
+    ):
         """
         Prepares the solc LSP server by calling `initialize`,
         and `initialized` methods.
         """
+        rootUri = self.project_root_uri
+        if project_root_suffix is not None:
+            rootUri = rootUri + '/' + project_root_suffix
```
cameel: Is this meant to be a hard-coded `/` on all platforms or should it be a platform-specific separator? I'm guessing the latter - pathlib has `/` operator for that:
```suggestion
            rootUri = rootUri / project_root_suffix
```

test/lsp.py
cameel: Also, I see that you're passing in a string for `project_root_suffix` while `self.project_root_uri` is a subclass of `pathlib.Path`. Wouldn't it be better to stick to the latter everywhere?

test/lsp.py
```diff
@@ -898,18 +899,29 @@ def main(self) -> int:
...
+        lsp: JsonRpcProcess,
+        expose_project_root=True,
+        analyze_all_files_in_project=False,
+        project_root_suffix = None
+    ):
         """
         Prepares the solc LSP server by calling `initialize`,
         and `initialized` methods.
         """
+        rootUri = self.project_root_uri
```
```suggestion
        root_uri = self.project_root_uri
```

test/lsp.py
cameel: I find the naming here confusing because `rootUri` and `self.project_root_uri` sound like the same thing and the "suffix" actually ends up being the middle part in full paths. Maybe this would be better?
- `rootUri` -> `project_root_uri_with_subdir`
- `project_root_suffix` -> `project_root_subdir`

cameel (changes requested):

libsolutil/CommonIO.h
christianparpart: @cameel I'm fine in adding directories first, of course. Why I wasn't doing it, is purely because in my use case that doesn't matter at all. I'm simply adding files to a list, and their order it irrelevant.

test/lsp.py
christianparpart: @cameel it's a string containing a URI, which is always using forward slash for path separation. So simply using '/' is fine. I'd need to extract the path part off `rootUri` first before I could make use of the `/` operator from PathLib in order to append a sub path.

libsolutil/CommonIO.h
christianparpart: Using `recursive_directory_iterator` now. I also don't think that `BreathFirstSearch` is what I was looking for. :)

libsolidity/lsp/LanguageServer.cpp
christianparpart: It's a `Json::LogicError` that's being thrown.

`run()` cannot log invalid parameters, as it doesn't know what is right or wrong with regards to LSP functions that were being invoked from the client side with e.g. bad parameters. A `RequestError` is being thrown in the callees and that's being caught in `run()`. So thre rethrow here as a different type will make the error properly pass back to the client (caller) by reporting an error.

libsolidity/lsp/LanguageServer.cpp
christianparpart: I did add this check when the lsp.py didn't work nicely yet. No need for that, as LSP clients usually _always_ provide a project root (aka. base path in that context). I've dropped these lines. And CI is all working well.

libsolidity/lsp/LanguageServer.cpp
christianparpart: > What about files from include dirs?

That's a good question. Do we want diagnostics on them or able to run actions like symbolic-rename?

> And files accessible through symlinks
By default it wasn't. But reading you tells me that you'd like it to. So I've added it now. :)

libsolidity/lsp/LanguageServer.cpp
christianparpart: @cameel It's not a full compilation, just parse & analyze, however, I am aware of the fact that we could certainly improve on caching some bits here and there. This is well known and last time we talked about and agreed that we're fine in doing optimizations on that level at a later stage.

IMHO we could re-evaluate that on Q4 (Because until Q3 i have other more user-facing things in the the backlog)

libsolidity/lsp/LanguageServer.cpp
christianparpart: I disagree. At least from my point of view we either include the full project for analysis, or we don't. I'd prefer not to overengineer this area and just stick to what other LSPs are doing, too (which is analyzing all project files, but for testing purposes in Soliidty's LSP, it's "convenient" to only operate on opened & imported files.

test/lsp.py
christianparpart: @cameel I'm fine either way. I agree however, but maybe in some kind of cleanup-PR?

test/lsp.py
christianparpart: Why? (Just trying to understand)

test/lsp.py
christianparpart: That's done like 12 lines above. :)

test/lsp.py
christianparpart: > We should also have a test for the case of `analyze_all_files_in_project=False`. Overall, I think this is a bit undertested given how many corner cases there could be.

@cameel This is confusing. This is the default argument for the `setup_lsp()` And most of the test would actually fail if that would not be supported. It's used in all other tests but those that do explicitly state `analyze_all_files_in_project=True`. If that would not be working, tons of failures would arise when running lsp.py. What exactly is "a bit undertested" here?

> Can you add a more complex test, that would also have some include dirs, and some files that should not be detected?

You are asking for a more complex project and see if it'll analyze them all as expected. What files would you expect it should not load, such as .txt files I assume? 

> Preferably also symlinks but I guess it would not be a good idea to store these in the repo so you'd have to do that in a temp dir... Does the current testing framework support that?

It's supported if it's coded. So for now there is no framework for automatically spinning up some temp dir and populating it with test files as there was no need for that. Symlinks are actually followed (now), and I think it's not really a problem to add a symlink to the repo, it should all work/run just fine. The only concern I would have is how that works on Windows (NTFS), which does have symlinks, but that's via .lnk files, not sure how well that works with recursive directory traversal at all and if we're actually testing our impl for `boost::filesystem::recursive_directory_iterator` here.

test/lsp.py
christianparpart: `rootUri` (now `root_uri` due to your suggestion above) is simply a break-out as I could not easily express that inline anymore. It's having this name as that's the precise name of the parameter key that it is being assigned to right away and not used afterwards anymore. 
`project_root_uri_with_subdir` to me is misleading. it is in fact the project root uri, it's just that sometimes I tweak that to point to a child directory, but not always, and what name what that then be? `project_root_uri_or_deeper_inside_uri`? Hmmm....

> project_root_suffix -> project_root_subdir

done

libsolidity/lsp/LanguageServer.cpp
christianparpart: Summary from Call: We explicitly decided against including all files from include paths but leave the possibility open for a future PR to enable such a feature to be optionally enabled (default disabled).

libsolidity/lsp/LanguageServer.cpp
christianparpart: Call Summary: we leave analyze-all-files-in-project enabled by default. with respect to perf, I think iff that becomes a problem these users can still disable the functionality, and we eventually have perf opts built in (PR) anyways. But dealing with keeping watch dirs up to date goes way out of the scope of this PR. But a future PR makes sense.

test/lsp.py
christianparpart: Call Summary:

- [x] `analyze-all-files-in-project` = None
- [x] nested directories
- [x] another more complex test case containing nested dirs, with symlink(maybe?) & files importing others in its tree.
- [x] add test that also has include paths

test/lsp.py
christianparpart: Call summary: not worth it (for now)

Changelog.md
```diff
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+ * Language Server: Analyzes all files in a project by default (can be customized by setting ``'analyze-all-files-in-project'`` to ``false`` in LSP settings object).
```
```suggestion
 * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -147,6 +150,26 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
 
 void LanguageServer::changeConfiguration(Json::Value const& _settings)
 {
+	// The settings item: "file-load-strategy" (enum) defaults to "project-directory" if not (or not correctly) set.
+	// It can be overridden during client's handshake or at runtime, as usual.
+	//
+	// If this value is set to "project-root" (default), all .sol files within the project root will be subject to operations.
```
```suggestion
	// If this value is set to "project-directory" (default), all .sol files located inside the project directory or reachable through symbolic links will be subject to operations.
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -147,6 +150,26 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+	// those files being imported directly or indirectly will be included in operations.
+	if (_settings["file-load-strategy"])
+	{
+		auto const text = _settings["file-load-strategy"].asString();
+		if (text == "project-directory")
+			m_fileLoadStrategy = FileLoadStrategy::ProjectDirectory;
+		else if (text == "directly-opened-and-on-import")
+			m_fileLoadStrategy = FileLoadStrategy::DirectlyOpenedAndOnImported;
+		else
+			throw RequestError{ErrorCode::InvalidParams};
```
cameel: I think it would be nice to have an error message here. I see you're generally using `lspAssert()` for this (which should be renamed to something like `lspRequire()` but that's a separate matter):
```suggestion
			lspAssert(false, ErrorCode::InvalidParams, "Invalid file load strategy: " + text);
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -147,6 +150,26 @@ Json::Value LanguageServer::toJson(SourceLocation const& _location)
...
+	//
+	// If this value is set to "project-root" (default), all .sol files within the project root will be subject to operations.
+	//
+	// Operations include compiler analysis, but also finding all symbolic references or symbolic renaming.
+	//
+	// If this value is set to "directly-opened-and-on-import", then only currently directly opened files and
+	// those files being imported directly or indirectly will be included in operations.
+	if (_settings["file-load-strategy"])
+	{
+		auto const text = _settings["file-load-strategy"].asString();
```
cameel, line 180: I think this is missing a type validation. I.e. what if it's not a string?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,28 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+	namespace fs = boost::filesystem;
+
+	std::vector<fs::path> collectedPaths{};
+
+	// We explicitly decided against including all files from include paths but leave the possibility
+	// open for a future PR to enable such a feature to be optionally enabled (default disabled).
+	auto iterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
+	auto const iteratorEnd = fs::recursive_directory_iterator();
+
+	while (iterator != iteratorEnd)
```
cameel: I think that this iterator should work with a range loop:
```suggestion
	for (fs::directory_entry const& dirEntry: fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse))
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,28 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+	{
+		if (iterator->path().extension() == ".sol")
+			collectedPaths.push_back(iterator->path());
+
+		++iterator;
+	}
```
cameel: I think that this iterator works with a range loop:
```suggestion
    auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
	for (fs::directory_entry const& dirEntry: directoryIterator)
		if (dirEntry.path().extension() == ".sol")
			collectedPaths.push_back(dirEntry.path());
```

libsolidity/lsp/LanguageServer.h
```diff
@@ -36,6 +36,22 @@ namespace solidity::lsp
...
+ * Enum to mandate what files to take into consideration for source code analysis.
+ */
+enum class FileLoadStrategy
+{
+	/// Takes only those files into consideration that are explicitly opened and those
+	/// that have been directly or indirectly imported.
+	DirectlyOpenedAndOnImported = 0,
+
+	/// Takes all Solidity (.sol) files within the project root into account.
+	/// Symbolic links will be followed, regardless of their destination (no root escalation checks).
```
cameel: I think it's a good idea to mention allowed paths option by name because it's related. That way this place will come up if you search for it in the code.
```suggestion
	/// Symbolic links will be followed, even if they lead outside of the project directory
	/// (`--allowed-paths` is currently ignored by the LSP).
```

test/lsp.py
```diff
@@ -898,18 +898,28 @@ def main(self) -> int:
...
         params = {
             'processId': None,
-            'rootUri': self.project_root_uri,
+            'rootUri': project_root_dir_uri_with_maybe_subdir,
             # Enable traces to receive the amount of expected diagnostics before
             # actually receiving them.
             'trace': 'messages',
-            'initializationOptions': {},
+            'initializationOptions': {
+            },
```
cameel: Looks like an unnecessary change.
```suggestion
            'initializationOptions': {},
```

cameel (commented):
I went through the PR and I only have small suggestions. Note that that almost all of them have actual code snippets you can use so I hope they're really trivial to apply and won't take much time. Also none of them seems very controversial to me.

Other than that it's just the extra tests.

test/lsp.py
cameel: > it is in fact the project root uri

But it isn't. You have two vars that can have different values (`.project_root_uri` and `rootUri`) and only one of them can be the root.

test/lsp.py
```diff
@@ -898,18 +898,28 @@ def main(self) -> int:
...
+        lsp: JsonRpcProcess,
+        expose_project_root=True,
+        analyze_all_files_in_project=False,
+        project_root_subdir=None
+    ):
         """
         Prepares the solc LSP server by calling `initialize`,
         and `initialized` methods.
         """
+        project_root_dir_uri_with_maybe_subdir = self.project_root_uri
```
cameel: You can actually shorten that a little bit:
```suggestion
        project_root_uri_with_maybe_subdir = self.project_root_uri
```

Also, another possibility:
```suggestion
        project_root_uri_override = self.project_root_uri
```
Not great but at least it makes it clear why you have two similarly named things (you want to modify the base value for this particular test) and which you should use (the override unless you're interested in the original value for some reason).

test/lsp.py
cameel: To add some detail: it turned out that `rootUri` is a string, not `pathlib.Path` as I thought.

test/lsp.py
cameel: > The only concern I would have is how that works on Windows (NTFS), which does have symlinks, but that's via .lnk files

That's what I meant. I know git supports adding symlinks to the repo but they would not work as actual symlinks on Windows.

Also, `.lnk` files are not the equivalent of symlinks. Windows does have actual symlinks, though they're actually a pain to work with. The semantics are different in corner cases. For example there are separate symlink types for files and dirs and if you create the wrong type, you get very weird results.

test/lsp.py
cameel: Sorry, my bad. It was actually there.

test/lsp.py
cameel: About pathlib - sure. Could be a separate PR. Just noticed that and wanted to point that out.

But maybe I'm actually wrong - the other case turned out to be a string so maybe you're not really mixing them after all. Not sure.

test/lsp.py
christianparpart: @cameel so let's not support symlinks at all. If it works, it works (like on Unix), but if it isn't, that's simply then not supported. I do not see a major use of this.

libsolidity/lsp/LanguageServer.cpp
christianparpart: It'll be auto-converted to a string, and if impossible (i.e. it's an JSON object rather than a scalar type), then an `Json::LogicError` is thrown, which will be caught in the `run()` function that reports back to the caller with an error message, saying that the JSON input was invalid.

test/lsp.py
cameel: I'm fine with not testing them due to difficulties of setting that up, but I think they should still be supported - simply because the compiler does support them. Otherwise LSP won't be able to handle some project layouts that the compiler has no problems with.

One of the use cases where some people could have plausibly used symlinks is to work around the `@` imports on versions that did not have `--include-path`. This could have been done via remappings but links were a viable option as well.

libsolidity/lsp/LanguageServer.cpp
cameel: Ah, ok then. Sounds good.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,23 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+	namespace fs = boost::filesystem;
+
+	std::vector<fs::path> collectedPaths{};
+
+	// We explicitly decided against including all files from include paths but leave the possibility
+	// open for a future PR to enable such a feature to be optionally enabled (default disabled).
+
+	auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
+	for (fs::directory_entry const& dirEntry: directoryIterator)
+		if (dirEntry.path().extension() == ".sol")
```
aarlt: Not sure, but maybe this will also include directories (ending with `.sol`)?

christianparpart: we tested that and concluded that it'll yield all kinds of directory entries. So I added an additional check to that if-clause.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,28 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 	}
 }
 
+vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
+{
+	namespace fs = boost::filesystem;
+
+	std::vector<fs::path> collectedPaths{};
```
```suggestion
	vector<fs::path> collectedPaths{};
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,28 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+	// We explicitly decided against including all files from include paths but leave the possibility
+	// open for a future PR to enable such a feature to be optionally enabled (default disabled).
+
+	auto directoryIterator = fs::recursive_directory_iterator(m_fileRepository.basePath(), fs::symlink_option::recurse);
+	for (fs::directory_entry const& dirEntry: directoryIterator)
+	{
+		if (
+			dirEntry.status().type() == fs::file_type::regular_file &&
+			dirEntry.path().extension() == ".sol"
+		)
```
cameel: You should allow symbolic links to regular files too.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -173,6 +196,28 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+		collectedPaths.push_back(dirEntry.path());
+	}
```
```suggestion
	for (fs::directory_entry const& dirEntry: directoryIterator)
		if (
			dirEntry.status().type() == fs::file_type::regular_file &&
			dirEntry.path().extension() == ".sol"
		)
			collectedPaths.push_back(dirEntry.path());
```

test/lsp.py
```diff
@@ -537,6 +538,18 @@ def at_end(self):
...
+    Undefined = auto()
+    ProjectDirectory = auto()
+    DirectlyOpenedAndOnImport = auto()
+
+    def lsp_name(self):
+        if self == FileLoadStrategy.ProjectDirectory:
+            return 'project-directory'
+        elif self == FileLoadStrategy.DirectlyOpenedAndOnImport:
+            return 'directly-opened-and-on-import'
+        return None
```
cameel: You could do this instead:
```suggestion
class FileLoadStrategy(Enum):
    Undefined = None
    ProjectDirectory = 'project-directory'
    DirectlyOpenedAndOnImport = 'directly-opened-and-on-import'
```
You can then access the strings through the `.value` member.

test/lsp.py
```diff
@@ -898,18 +911,28 @@ def main(self) -> int:
...
+            project_root_uri_with_maybe_subdir = self.project_root_uri + '/' + project_root_subdir
         params = {
             'processId': None,
-            'rootUri': self.project_root_uri,
+            'rootUri': project_root_uri_with_maybe_subdir,
             # Enable traces to receive the amount of expected diagnostics before
             # actually receiving them.
             'trace': 'messages',
-            'initializationOptions': {},
+            # 'initializationOptions': {},
```
cameel: Why is this commented-out?

test/lsp.py
```diff
@@ -923,8 +946,19 @@ def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root=True):
                 }
             }
         }
+
+        if file_load_strategy != FileLoadStrategy.Undefined:
+            params['initializationOptions'] = {}
+            params['initializationOptions']['file-load-strategy'] = file_load_strategy.lsp_name()
```
```suggestion
            params['initializationOptions'] = {'file-load-strategy': file_load_strategy.lsp_name()}
```

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
     # }}}
 
     # {{{ actual tests
+    def test_analyze_all_project_files1(self, solc: JsonRpcProcess) -> None:
```
```suggestion
    def test_analyze_all_project_files_flat(self, solc: JsonRpcProcess) -> None:
```

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+        report = published_diagnostics[1]
+        self.expect_equal(report['uri'], self.get_test_file_uri('D', SUBDIR), "Correct file URI")
+        self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+        # E.sol
+        report = published_diagnostics[2]
+        self.expect_equal(report['uri'], self.get_test_file_uri('E', SUBDIR), "Correct file URI")
+        self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+    def test_analyze_all_project_files2(self, solc: JsonRpcProcess) -> None:
```
```suggestion
    def test_analyze_all_project_files_nested(self, solc: JsonRpcProcess) -> None:
```

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+        self.setup_lsp(
+            solc,
+            file_load_strategy=FileLoadStrategy.ProjectDirectory,
+            project_root_subdir=SUBDIR
+        )
+        published_diagnostics = self.wait_for_diagnostics(solc)
+        self.expect_equal(len(published_diagnostics), len(EXPECTED_FILES), "Test number of files analyzed.")
+        for report in published_diagnostics:
+            self.expect_true(report['uri'] in EXPECTED_URIS, "Correct file URI")
+            self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
```
cameel: This won't catch things like an URI being reported twice while another being missing. How about this?
```suggestion
        self.expect_equal({report['uri'] for report in published_diagnostics}, set(EXPECTED_URIS))
        self.expect_equal([len(report['uri']) for report in published_diagnostics], [0] * len(EXPECTED_URIS))
```
Also, you could just make `EXPECTED_FILES` and `EXPECTED_URIS` sets rather than lists.

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+        )
+
+        # All but the last report should be from expected files
+        for report in published_diagnostics[:-IMPLICITLY_LOADED_FILE_COUNT]:
+            self.expect_true(report['uri'] in EXPECTED_URIS, "Correct file URI")
+            self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+        # Check last report (should be the custom imported lib).
+        # This file is analyzed because it was imported via "A/B/C/foo.sol".
+        report = published_diagnostics[len(EXPECTED_URIS)]
```
```suggestion
        last_report = published_diagnostics[len(EXPECTED_URIS)]
```

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+            file_load_strategy=FileLoadStrategy.ProjectDirectory,
+            project_root_subdir=SUBDIR
+        )
+        published_diagnostics = self.wait_for_diagnostics(solc)
+        self.expect_equal(len(published_diagnostics), len(EXPECTED_FILES), "Test number of files analyzed.")
+        for report in published_diagnostics:
+            self.expect_true(report['uri'] in EXPECTED_URIS, "Correct file URI")
+            self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+    def test_analyze_all_project_files3(self, solc: JsonRpcProcess) -> None:
```
```suggestion
    def test_analyze_all_project_files_nested_with_include_paths(self, solc: JsonRpcProcess) -> None:
```

test/lsp.py
```diff
@@ -1295,6 +1337,99 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+
+        # All but the last report should be from expected files
+        for report in published_diagnostics[:-IMPLICITLY_LOADED_FILE_COUNT]:
+            self.expect_true(report['uri'] in EXPECTED_URIS, "Correct file URI")
+            self.expect_equal(len(report['diagnostics']), 0, "no diagnostics")
+
+        # Check last report (should be the custom imported lib).
+        # This file is analyzed because it was imported via "A/B/C/foo.sol".
+        report = published_diagnostics[len(EXPECTED_URIS)]
+        self.expect_equal(report['uri'], f"{self.project_root_uri}/other-include-dir/otherlib/second.sol", "Correct file URI")
```
cameel: I think you have a helper for this?
```suggestion
        self.expect_equal(report['uri'], self.get_test_file_path('otherlib/second.sol', 'other-include-dir'), "Correct file URI")
```

test/lsp.py
```diff
@@ -1408,6 +1543,43 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
         self.expect_equal(len(diagnostics), 1, "no diagnostics")
         self.expect_diagnostic(diagnostics[0], code=2018, lineNo=5, startEndColumns=(4, 62))
 
+    def test_custom_includes_with_full_project(self, solc: JsonRpcProcess) -> None:
+        """
+        Tests loading all all project files while having custom include directories configured.
```
```suggestion
        Tests loading all project files while having custom include directories configured.
```

test/lsp.py
```diff
@@ -1408,6 +1543,43 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
...
+        report = published_diagnostics[0]
+        self.expect_equal(report['uri'], self.get_test_file_uri('using-custom-includes', 'include-paths'))
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 0, "no diagnostics")
+
+        # imported file
+        report = published_diagnostics[1]
+        self.expect_equal(report['uri'], f"{self.project_root_uri}/other-include-dir/otherlib/otherlib.sol")
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 1, "no diagnostics")
```
cameel: It's not zero so it does have diagnostics :)
```suggestion
        self.expect_equal(len(diagnostics), 1)
```

cameel:
This is the last thing missing for approval: https://github.com/ethereum/solidity/pull/13150#discussion_r951688759, i.e. I think it won't handle symlinks. But that should be easy to fix. The other stuff are just simple tweaks and the new tests seem good enough for now.

libsolidity/lsp/LanguageServer.cpp
christianparpart: done

libsolidity/lsp/LanguageServer.h
```diff
@@ -36,6 +36,23 @@ namespace solidity::lsp
...
 enum class ErrorCode;
 
+/**
+ * Enum to mandate what files to take into consideration for source code analysis.
+ */
+enum class FileLoadStrategy
+{
+	/// Takes only those files into consideration that are explicitly opened and those
+	/// that have been directly or indirectly imported.
+	DirectlyOpenedAndOnImported = 0,
```
nikola-matic, line 46: Should this be `DirectlyOpenedAndOrImported`?

edit: ah never mind, looks like it references `directly-opened-and-on-import`. Still sounds a bit weird though.

test/lsp.py
```diff
@@ -923,8 +938,19 @@ def setup_lsp(self, lsp: JsonRpcProcess, expose_project_root=True):
                 }
             }
         }
+
+        if file_load_strategy != FileLoadStrategy.Undefined:
+            params['initializationOptions'] = {}
+            params['initializationOptions']['file-load-strategy'] = file_load_strategy.value
```
Marenz, line 944:
```suggestion
            params['initializationOptions'] = { 'file-load-strategy' : file_load_strategy.value }
```

I think that should work

Marenz (dismissed):
lgtm, found a minor optional thing

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -50,9 +53,24 @@ using namespace solidity::lsp;
...
 
+bool resolvesToRegularFile(boost::filesystem::path _path)
+{
+	fs::file_status fileStatus = fs::status(_path);
+
+	while (fileStatus.type() == fs::file_type::symlink_file)
+	{
+		_path = boost::filesystem::read_symlink(_path);
+		fileStatus = fs::status(_path);
+	}
```
cameel: Hmm... I was hoping this would be a simple check but yeah, I guess links can point at other links...

The problem with this is that you can fall into an infinite loop if you have two symlinks pointing at each other. Or a longer cycle.

Now that I think about it, what will the the whole directory scanning do if it finds a symlink loop?

libsolidity/lsp/LanguageServer.cpp
cameel: The indent of `collectedPaths` is still wrong. I guess it's not that visible on the suggestion.

test/lsp.py
cameel: Yeah, suggested that myself :) https://github.com/ethereum/solidity/pull/13150#discussion_r951702983

test/lsp.py
```diff
@@ -1295,6 +1329,98 @@ def user_interaction_failed_autoupdate(self, test, sub_dir):
...
+        Same as first test on that matter but with deeper nesting levels.
+        """
+        SUBDIR = 'include-paths-nested'
+        EXPECTED_FILES = [
+            "A/B/C/foo",
+            "A/B/foo",
+            "A/foo",
+            "foo",
+        ]
+        EXPECTED_URIS = [self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES]
```
cameel: This is what I meant by "make them sets" in https://github.com/ethereum/solidity/pull/13150#discussion_r951733484
```suggestion
        EXPECTED_FILES = {
            "A/B/C/foo",
            "A/B/foo",
            "A/foo",
            "foo",
        }
        EXPECTED_URIS = {self.get_test_file_uri(x, SUBDIR) for x in EXPECTED_FILES}
```

You can then just use `EXPECTED_URIS` insterad of `set(EXPECTED_URIS)` below.

cameel: Also, you have pretty much the same thing in the test below.

cameel (dismissed):
I still have some comments regarding the previous tweaks but all the important stuff has been resolved so I'm approving anyway.

Marenz (approved):

PR #13152, opened 2022-06-15
Add warning comment to dangling references docs.

ekpyron:
Should be fine to merge once one of the ``_cli`` jobs passes.

### DIFF
--- a/docs/types/reference-types.rst
+++ b/docs/types/reference-types.rst
@@ -555,6 +555,7 @@ to long layout in storage<bytes-and-string>`.
     // SPDX-License-Identifier: GPL-3.0
     pragma solidity >=0.8.0 <0.9.0;
 
+    // This will report a warning
     contract C {
         bytes x = "012345678901234567890123456789";
 


docs/types/reference-types.rst
```diff
@@ -555,6 +555,7 @@ to long layout in storage<bytes-and-string>`.
     // SPDX-License-Identifier: GPL-3.0
     pragma solidity >=0.8.0 <0.9.0;
 
+    // This will report a warning
```
cameel, line 558: I'd put it at the line that generates the warning (i.e. the one that does `push()`).

cameel (approved):
Approving anyway since it's not a big problem but I'd put the comment directly over the problematic line.

docs/types/reference-types.rst
ekpyron: All other instances have it up front, so I'll keep it there :-).

ekpyron:
The relevant parts of the CI runs just passed, so I'll merge.

docs/types/reference-types.rst
cameel: Oh, ok then. I thought we were using it more like an integral part of the surrounding text. Apparently not :)

PR #13154, opened 2022-06-15
Fix release regex and add Ubuntu Kinetic to PPA scripts

ekpyron:
For merging after the release.

### DIFF
--- a/scripts/release_ppa.sh
+++ b/scripts/release_ppa.sh
@@ -48,7 +48,7 @@ else
 fi
 
 is_release() {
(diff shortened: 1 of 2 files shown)


ekpyron:
We should probably also remove the logic for the now non-existent ``release`` branch.

cameel (dismissed):

scripts/release_ppa.sh
```diff
@@ -57,7 +57,7 @@ packagename=solc
 
 static_build_distribution=impish
 
-DISTRIBUTIONS="focal impish jammy"
+DISTRIBUTIONS="focal impish jammy kinetic"
```
cameel, line 60: Should we add it to deps-ppa too?

cameel (approved):

Issue #7665, opened 2019-11-08
Pretty print expected json output of command line tests.


ekpyron:
Came up in https://github.com/ethereum/solidity/pull/7589#discussion_r344259919.
Is there a nice and easy way to have a pretty printing pass on the output? That would make those tests way easier to review.

leonardoalt:
The `commandlineTests` could `jq` the output, but not sure we want to add that dependency.

cameel:
This should be very easy to do without extra dependencies now that #11583 is implemented. Just add the `--pretty-print` flag in `cmdlineTests.sh`, run it with `--update` and commit the result. I'm marking it as `good first issue`.

0xGeorgii:
@matheusaaguiar May I take it?

matheusaaguiar:
@GeorgePlotnikov yes, please! Let me know if you need any help.

0xGeorgii:
@matheusaaguiar let me clarify I'm moving in the right direction. There is a script `cmdlineTests.sh` that runs cmdlineTests. The script executes test with the `command_args="--standard-json "$(cat "${tdir}/args" 2>/dev/null || true)`.

The `--standard-json` arg which means I specify a test output in a json format, like here: [linking_standard_solidity/output.json](https://github.com/ethereum/solidity/blob/develop/test/cmdlineTests/linking_standard_solidity/output.json):

`{"contracts":{"A":{"C":{"evm":{"bytecode":{"linkReferences":{},"object":"<BYTECODE REMOVED>"}}}}},"sources":{"A":{"id":0}}}`

but we want to give a user opportunity to write an output in a `--pretty-json` format like that:

```
{
    "contracts":{
        "A":{
            "C":{
                "evm":{
                    "bytecode":{
                        "linkReferences":{
                            
                        },
                        "object":"<BYTECODE REMOVED>"
                    }
                }
            }
        }
    },
    "sources":{
        "A":{
            "id":0
        }
    }
}
```

My questions are:
1. Am I right that it is desirable to add a cmd line arg for the whole script like `cmdlineTests.sh --pretty-json`
2. If yes how the existing test cases should be managed:
  a. if there is no `--pretty-json` in the **args** or the output is not in a json format then it should be ignored?
  b. if the output not in the prettified json format then prettify it first, and compare afterwards?

matheusaaguiar:
[...]

PR #13162, opened 2022-06-17
cleaning up helpers around errors

### LINKED ISSUE #11959: Clean up the helpers around errors
For severity some places use strings, while others the enum.

Apparently it's also indirectly used by other places in the compiler, like CommandLineInterface.cpp, that do not adhere to the severity standard:
```
solc/CommandLineInterface.cpp:		formatter.printExceptionInformation(_exception, "Compiler error");
printExceptionInformation(SourceReferenceExtractor::extract(m_charStreamProvider, _exception, _severity));
```

_Originally posted by @leonardoalt in https://github.com/ethereum/solidity/pull/11606#discussion_r704815675_

nishant-sachdeva:
closes #11959 

Also need to close https://github.com/ethereum/solc-js/pull/643 for CI tests to pass here 
( Update :: closed. Decided we did not need this change in solc-js)


libsolidity/interface/StandardCompiler.cpp
```diff
@@ -125,7 +125,7 @@ Json::Value formatErrorWithException(
 	// TODO: consider enabling color
 	string formattedMessage = SourceReferenceFormatter::formatExceptionInformation(
 		_exception,
-		_type,
```
Marenz, line 128: Do we still need the variable at all? (`_type`), can the use in line 139 maybe also be replaced the same way?

nishant-sachdeva: The `_type` variable is used in line `Json::Value formatError` function. I'm not sure if it would be the right thing to constrain the `_type` variable there to just the 3 types allowed in the enum `Error::Severity`. 

If however, we are to do it, perhaps it would be better to make a different issue out of it ?

liblangutil/Exceptions.h
```diff
@@ -236,29 +237,26 @@ class Error: virtual public util::Exception
...
 		}
-		solAssert(false, "");
+		solAssert(false);
+	}
+
+	static std::string formatErrorSeverityLowercase(Error::Severity _severity)
+	{
+		std::string severityValue = formatErrorSeverity(_severity);
+		boost::algorithm::to_lower(severityValue);
+		return severityValue;
```
cameel, line 304:
```suggestion
		return boost::algorithm::to_lower(formatErrorSeverity(_severity));
```

liblangutil/Exceptions.h
```diff
@@ -236,29 +237,26 @@ class Error: virtual public util::Exception
...
-			return "warning";
+			return "Warning";
 		case Severity::Error:
-			solAssert(isError(_severity), "");
-			return "error";
+			solAssert(isError(_severity));
+			return "Error";
 		}
-		solAssert(false, "");
+		solAssert(false);
```
cameel: I could be wrong, but I think that we no longer need this assert since all cases are handled and the C++ compiler can detect that.

liblangutil/Exceptions.h
```diff
@@ -236,29 +237,26 @@ class Error: virtual public util::Exception
...
-		if (_severity == Severity::Info)
-			return "Info";
-		if (_severity == Severity::Warning)
-			return "Warning";
-		solAssert(isError(_severity), "");
-		return "Error";
-	}
-
-	static std::string formatErrorSeverityLowercase(Severity _severity)
+	static std::string formatErrorSeverity(Error::Severity _severity)
```
cameel:
```suggestion
	static std::string formatErrorSeverity(Severity _severity)
```

liblangutil/SourceReferenceExtractor.h
```diff
@@ -55,12 +55,12 @@ namespace SourceReferenceExtractor
 	struct Message
 	{
 		SourceReference primary;
-		std::string severity; // "Error", "Warning", "Info", ...
+		Error::Severity severity; // "Error", "Warning", "Info", ...
```
cameel:
```suggestion
		Error::Severity severity;
```

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -179,8 +179,13 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
...
 {
 	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
-	errorColored(severity) << _msg.severity;
+	optional<Error::Severity> severity = _msg.severity;
+	solAssert(
+		severity == Error::Severity::Error ||
+		severity == Error::Severity::Info ||
+		severity == Error::Severity::Warning
+	);
```
cameel: I don't think we need this assert. It's an enum so there are no other possible values and it can't be `nullopt` here.

And when we add new values, we don't need our attention drawn here. Only `errorColored()` will need an update the the compiler will catch that due to the `switch` used there.

cameel (changes requested):

liblangutil/Exceptions.h
nishant-sachdeva: Gives a compiler error. `error: control reaches end of non-void function [-Werror=return-type]`

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -125,7 +125,7 @@ Json::Value formatErrorWithException(
...
+		_severity,
```
cameel: I think this is wrong. You can see that even by looking at command-line tests - in many cases more specific error names got replaced by simply `Error` so we're losing some information. `_type` can have a lot of values that are not available in `Error::Severity`. All the ones from `Error::Type` and also some custom values that we set in `StandardCompiler.cpp`.

I think that these custom values should simply be added to `Error::Type` so that we can use this type for this field. I'm not sure why these custom values were not there in the first place - maybe because they're considered random failures rather than specific kinds of compiler errors - but then why have specific names for them at all?

nishant-sachdeva: You mean, they should be added to `Error::Severity` ?

liblangutil/Exceptions.h
nishant-sachdeva: This will raise a compile error, right ? 
`boost::algorithm::to_lower` returns a `void`, whereas this functions seeks to return a `string`

liblangutil/Exceptions.cpp
```diff
@@ -39,8 +39,23 @@ Error::Error(
 	m_errorId(_errorId),
 	m_type(_type)
 {
-	switch (m_type)
+	switch(m_type)
```
stackenbotten: Coding style error

liblangutil/Exceptions.cpp
```diff
@@ -62,7 +77,20 @@ Error::Error(
...
+		m_typeName = "UnimplementedFeatureError";
+		break;
+	case Type::YulException:
+		m_typeName = "YulException";
+		break;
+	case Type::SMTLogicException:
+		m_typeName = "SMTLogicException";
+		break;
+	default:
+		solAssert(_type == Type::Warning);
```
Marenz: Does it possibly make sense to just use a map here instead of a long switch? 

Either way, can't we use `formatErrorType` and reuse that logic instead of recreating it in two places?

cameel: `switch` is nice in cases when there's no `default`. The compiler can tell you that you need a new branch when stuff is added to the enum.

matheusaaguiar: > Either way, can't we use `formatErrorType` and reuse that logic instead of recreating it in two places?

reiterating that question:
The `switch case` statement here seems to be identical to the one in method `formatErrorType`. Couldn't we use it here instead?

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -179,10 +178,10 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg)
 {
 	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
-	errorColored(severity) << _msg.severity;
+	Error::Type type = _msg.type;
```
matheusaaguiar: Is this variable really needed? It is used in line 184, but not in 182... I guess we could use `_msg.type` directly.

liblangutil/Exceptions.cpp
nishant-sachdeva: Been thinking of doing that myself. Will implement that and push changes.

nishant-sachdeva: Been considering doing that myself. Will implement it in the next push

liblangutil/Exceptions.h
```diff
@@ -238,35 +247,72 @@ class Error: virtual public util::Exception
...
+		switch (_severity)
+		{
+		case Severity::Info:
 			return "Info";
-		if (_severity == Severity::Warning)
+		case Severity::Warning:
 			return "Warning";
-		solAssert(isError(_severity), "");
-		return "Error";
+		default:
```
cameel:
```suggestion
		case Severity::Error:
```

liblangutil/Exceptions.h
```diff
@@ -238,35 +247,72 @@ class Error: virtual public util::Exception
...
+			return "SyntaxError";
+		case Type::TypeError:
+			return "TypeError";
+		case Type::UnimplementedFeatureError:
+			return "UnimplementedFeatureError";
+		case Type::YulException:
+			return "YulException";
+		case Type::SMTLogicException:
+			return "SMTLogicException";
+		default:
```
cameel:
```suggestion
		case Type::Warning:
```
I'd also put `Warning` and `Info` next to each other.

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -178,11 +177,10 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 
 void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg)
 {
-	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
```
cameel, line 182: I think `severityFromString()` is unused now. I remember that it was created specifically because we needed it here.

solc/CommandLineInterface.cpp
```diff
@@ -765,7 +765,7 @@ void CommandLineInterface::compile()
 	catch (CompilerError const& _exception)
 	{
 		m_hasOutput = true;
-		formatter.printExceptionInformation(_exception, "Compiler error");
+		formatter.printExceptionInformation(_exception, Error::Type::CompilerError);
```
cameel: Would this work?
```suggestion
		formatter.printExceptionInformation(_exception, _exception.type());
```
If it does, and `CompilerError` inherits from `Error` as I suspect, we could actually just remove this `catch` clause since the one below will do the same thing.

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -178,11 +177,10 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
...
-	errorColored(severity) << _msg.severity;
+	Error::Type type = _msg.type;
+	errorColored(Error::errorSeverity(_msg.type)) << Error::formatErrorType(_msg.type);
```
cameel: This is not equivalent to the old version. Now it prints the exact type while it used to print severity. It should be this instead:
```suggestion
	errorColored(Error::errorSeverity(_msg.type)) << Error::formatErrorSeverity(Error::errorSeverity(_msg.type));
```

The result is that you get specific error types in all those command-line tests. They should not change. I think we only print severity in the CLI on purpose and I'd not change that in this PR.

liblangutil/SourceReferenceExtractor.h
```diff
@@ -55,12 +55,12 @@ namespace SourceReferenceExtractor
 	struct Message
 	{
 		SourceReference primary;
-		std::string severity; // "Error", "Warning", "Info", ...
+		Error::Type type;
```
cameel: I think this was actually meant to be just severity not the whole type:
```suggestion
		Error::Severity severity;
```
I.e. `Error` carries complete info about the error, including the type, but `Message` is just some subset of its info meant for display.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -1188,35 +1173,31 @@ Json::Value StandardCompiler::compileSolidity(StandardCompiler::InputsAndSetting
...
 			"general",
 			"Unknown exception during compilation: " + boost::diagnostic_information(_exception)
 		));
 	}
 	catch (...)
 	{
 		errors.append(formatError(
-			Error::Severity::Error,
-			"Exception",
+			Error::Type::InternalCompilerError,
```
cameel: It was `"Exception"` but you changed this to `InternalCompilerError`. Was that intentional?

It's not really wrong - it's debatable which one is more fitting here - but it changes current behavior. I think we should consider removing `Type::Exception` and just using the ICE; maybe also squash some other redundant types. I don't want to get into that now though so I'd leave this for a separate PR.

liblangutil/Exceptions.h
nishant-sachdeva: Doesn't work. Compiler gives an error `error: control reaches end of non-void function [-Werror=return-type]`

So, I'll add a `solAssert(false)` at the end of the function in addition to this change.

liblangutil/SourceReferenceExtractor.h
nishant-sachdeva: I'm not sure this should be done. 

Earlier, `severity` was a string value that was freely abused. Several values of `severity` were being passed. These were not all provided in `Error::Severity` enum. 

So, I converted this to `Error::Type`. This provides for a bigger coverage for the error types. 

Otherwise, there are cases right now, where `error types` are printed ( they were passed as strings before . So, even though the print function was supposed to accept `severity`, it compiled anyway ) . Those will be reduced to one of `Error::Severity` values. That will lead to some information loss.

As against now, where we are getting extra information in our error messages by printing the exact error type.

A workaround for this would be to expand `Error::Severity` the way we have expanded `Error::Type`

solc/CommandLineInterface.cpp
nishant-sachdeva: Nope, it didn't work. Keeping it as it is for now.

liblangutil/SourceReferenceExtractor.h
cameel: ok, I guess we could keep it as `Error::Type` here as long as https://github.com/ethereum/solidity/pull/13162#discussion_r939127367 is changed. I thought that that other place was removing all the extra info anyway and dumbing it down to `Error::Severity` but if not, we can keep it as is.

libsolidity/interface/StandardCompiler.cpp
nishant-sachdeva: Restored it to `Error::Exception` . 
EDIT :: Not yet restored. Changing it to Error::Type::Exception induces a CI error. Investigating that before restoring it.

We can look at removing these redundant error types in another PR. 
@cameel Should I make another issue out of it ?

liblangutil/SourceReferenceFormatter.cpp
nishant-sachdeva: Updated in the latest push.

liblangutil/SourceReferenceFormatter.cpp
nishant-sachdeva: Removed this code in the latest push.

liblangutil/SourceReferenceExtractor.h
nishant-sachdeva: If we go ahead with https://github.com/ethereum/solidity/pull/13162#discussion_r939127367, we get the following errors on the CI 

For eg. https://circleci.com/gh/ethereum/solidity/1131923?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link  


Here, the test apparently does expect a proper error Type. But since we've constrained it to only be from `Error::Severity` , it throws this error. 

Had made a PR to propagate this change to solc-js  ( https://github.com/ethereum/solc-js/pull/643) , but it was refused.

nishant-sachdeva: Also, indeed, you're right. Doing https://github.com/ethereum/solidity/pull/13162#discussion_r939127367  effectively dumbs it down to `Error::Severity` . 

Will push changes in a sperate commit till we figure out what to do with the error on solc-js side.

cameel: Can you track down the path of code that `SyntaxError` from solc-js goes through? I.e. I don't fully understand why it's turning into `Error` even though this is not on the CLI. Are we using `SourceReferenceFormatter` for more than just CLI?

I'd really prefer a solution that would keep the current output unchanged and only reshuffle things internally...

nishant-sachdeva: Trying to track the code at solc-js
`solc-js/test/cli.ts : 62 `

To be fair, the error does seem to be a genuine mismatch.  The file name is `cli.ts` . We do want our CLI tests to only show `severity` and not the full `type`. 

The functions seems to be sent for compilation at solc.ts:207
```
const output = JSON.parse(solc.compile(JSON.stringify(cliInput), callbacks));

```

During compilation, as expected, we get error as `Error:...` instead of `SyntaxError:...`

libsolidity/interface/StandardCompiler.cpp
cameel: I'd wait with a PR until we're done with this one.

Would be best if you could first propose which ones you would eliminate/merge. `Exception` is an obvious one but maybe we can get rid of some more. Then we can discuss on the channel whether to go on with this (and how breaking it is).

liblangutil/SourceReferenceExtractor.h
cameel: > To be fair, the error does seem to be a genuine mismatch.  The file name is `cli.ts` . We do want our CLI tests to only show `severity` and not the full `type`. 

It's complicated :) solc-js uses the Standard JSON interface and its CLI is just a thin wrapper over that. So it naturally gets full error type from JSON. Maybe it should convert that to severity internally but I don't think it's that important. It's not 1:1 equivalent to solc anyway.

For now just assume that in solc-js we want to keep it as `SyntaxError`.

liblangutil/Exceptions.h
cameel: Interesting. Looks like the compiler's right. This is still technically reachable because you could just cast some value to it, even if there's no corresponding enum value (["control reaches end of non-void function" with fully handled case switch over an enum type](https://stackoverflow.com/questions/33607284/control-reaches-end-of-non-void-function-with-fully-handled-case-switch-over-a)).

Ok then. Let's leave it as is.

liblangutil/SourceReferenceExtractor.h
cameel: Also, I see you still haven't changed this field back to `Error::Type`.

nishant-sachdeva: I don't think it's getting the full error type anymore. 

The expected value is `SyntaxError`. But the value it's getting is just `Error`. 
Perhaps we need to take another look at whether we really want to trim down error types into merely the severity values

matheusaaguiar: I tried to reproduce CI [ems_solcjs failing test ](https://app.circleci.com/pipelines/github/ethereum/solidity/25722/workflows/ad2d4a51-1a91-4884-9a0e-730ad5a83d42/jobs/1135117) locally, but I was not able to... any clues why this should be? Locally the node and npm versions are different from CI... 
Not that I think that node/npm versions are the problem here, but I got curious about this difference in behaviour when testing local and maybe this can help identify a solution here.

cameel: Probably because it was my proposal to change it to `Error::Severity` that was responsible for it. Now that it's back to `Error::Type`, there should be no error.

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -176,15 +175,18 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
...
 }
 
-void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg)
+void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg, bool jsonOutput)
 {
-	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
-	errorColored(severity) << _msg.severity;
+	errorColored(Error::errorSeverity(_msg.type)) << (
+		jsonOutput ?
```
stackenbotten: Coding style error

stackenbotten: Coding style error

libsolidity/interface/StandardCompiler.cpp
matheusaaguiar: > Restored it to Error::Exception .

Not sure what was decided here, but it was not changed.
Although, it does not seem to affect any test output....

matheusaaguiar (dismissed):
LGTM.
But maybe it still is necessary to make the change mentioned [here](https://github.com/ethereum/solidity/pull/13162/files#r940490980) (?).

libsolidity/interface/StandardCompiler.cpp
nishant-sachdeva: The change is resulting in an error on the CI. Will look into it and fix as required.

Will make another PR to remove redundant error types, once this is ready for merging.

liblangutil/Exceptions.h
```diff
@@ -238,35 +247,70 @@ class Error: virtual public util::Exception
...
+			return "YulException";
+		case Type::SMTLogicException:
+			return "SMTLogicException";
+		case Type::Warning:
+			return "Warning";
+		case Type::Info:
+			return "Info";
 		}
-		solAssert(false, "");
+		solAssert(false);
```
christianparpart:
```suggestion
		util::unreachable();
```
I think that makes it more clear what you mean by this.

christianparpart (commented):
I am generally for that, but I've to split my review into two sessions. Will be back in ~2h.

libsolidity/interface/StandardCompiler.cpp
cameel: > The change is resulting in an error on the CI.

Do you mean [this one in `t_ubu_soltest_all`](https://app.circleci.com/pipelines/github/ethereum/solidity/25863/workflows/23ed2022-7e2e-481f-b79a-541c5f592463/jobs/1141890/parallel-runs/4)? It looks completely unrelated. It's in one of the tests that create files/dirs on disk and it could be that running stuff in parallel is messing with it.

Is it reproducible?

nishant-sachdeva: Yes. It was indeed completely unrelated. 
A mere rerun ended up passing the test

liblangutil/Exceptions.h
```diff
@@ -211,11 +220,12 @@ class Error: virtual public util::Exception
...
-		if (_type == Type::Info)
-			return Severity::Info;
-		if (_type == Type::Warning)
-			return Severity::Warning;
-		return Severity::Error;
+		switch (_type)
+		{
+			case Type::Info: return Severity::Info;
+			case Type::Warning: return Severity::Warning;
+			default: return Severity::Error;
```
cameel, line 228: Please add `solAssert(isError(_type))`.

Actually I'm not sure why we did not have it here while we did in `formatErrorSeverity()` where it's useless.

liblangutil/Exceptions.h
cameel: Right. I missed that `to_lower()` works in-place.

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -176,15 +175,18 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 	}
 }
 
-void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg)
+void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg, bool jsonOutput)
```
cameel:
```suggestion
void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg, bool _jsonOutput)
```

Also in a few other places.

liblangutil/SourceReferenceFormatter.h
```diff
@@ -51,22 +51,24 @@ class SourceReferenceFormatter
 
 	/// Prints source location if it is given.
 	void printSourceLocation(SourceReference const& _ref);
-	void printExceptionInformation(SourceReferenceExtractor::Message const& _msg);
-	void printExceptionInformation(util::Exception const& _exception, std::string const& _severity);
+	void printExceptionInformation(SourceReferenceExtractor::Message const& _msg, bool jsonOutput=false);
+	void printExceptionInformation(util::Exception const& _exception, Error::Type _type, bool jsonOutput=false);
+	void printJSONExceptionInformation(util::Exception const& _exception, Error::Type _type);
```
cameel: I think this function is unused and has no definition :)

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -176,15 +175,18 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
...
 {
-	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
-	errorColored(severity) << _msg.severity;
+	errorColored(Error::errorSeverity(_msg.type)) << (
+		jsonOutput ?
+		Error::formatErrorType(_msg.type) :
+		Error::formatErrorSeverity(Error::errorSeverity(_msg.type))
+	);
```
cameel: Nice. That's a pretty elegant solution. I was afraid this was going to be uglier and require changes in multiple places but looks like this solves the problem in a very clean way.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -708,15 +704,14 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
 			{
 				/// If the import succeeded, let mark all the others as warnings, otherwise all of them are errors.
 				ret.errors.append(formatError(
-					found ? Error::Severity::Warning : Error::Severity::Error,
-					"IOError",
+					Error::Type::IOError,
```
cameel: Oh, this is an interesting corner case that I did not catch earlier. I think that changing this to an error breaks the [`urls` mechanism](https://docs.soliditylang.org/en/latest/path-resolution.html#initial-vfs-content-standard-json-with-import-callback). And we're apparently missing coverage since it did not fail. Please verify - add a command-line test case covering this (i.e. multiple urls of which only one is valid and is not first on the list). I think that the test case will fail.

To fix this keep the `found` variable and to replace `Error::Type::IOError` with `Error::Type::Warning`. This does change behavior slightly but only in this single situation and having a warning called `IOError` was not that great anyway. To solve this cleanly we would have to rework the whole PR to decouple "severity" from "type" - not worth the mess IMO :)

liblangutil/SourceReferenceExtractor.cpp
```diff
@@ -30,7 +30,7 @@ using namespace solidity::langutil;
 SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
 	CharStreamProvider const& _charStreamProvider,
 	util::Exception const& _exception,
-	string _severity
+	Error::Type _type
```
cameel: Boolean flags like `jsonOutput` are not great for readability and I wonder if we might be able to solve that better. What do you think about changing it to:
```suggestion
	variant<Error::Type, Error::Severity> _typeOrSeverity;
```
and then having separate overloads for each type: `printExceptionInformation(util::Exception const&, Error::Type)` and `printExceptionInformation(util::Exception const&, Error::Severity)`.

What do you think?

liblangutil/SourceReferenceFormatter.cpp
cameel: Also, I think this could have a more generic name since it does not have to be tied to JSON in general (even if now it is). For example:
```suggestion
void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg, bool _printFullType)
```

cameel (changes requested):
Found an important problem: one of the changes probably breaks `urls` and we don't have it covered so tests don't catch it.

Other than that, just minor stuff. I really like the solution you found in the end.

liblangutil/Exceptions.h
nishant-sachdeva: That's gonna be circular isn't it ? 
`isError(_type)`  calls the function `errorSeverity(_type)` itself.

Also, we've removed it from `formatErrorSeverity()` as well

liblangutil/SourceReferenceFormatter.h
nishant-sachdeva: Well noted. Will remove in the next code push.

liblangutil/Exceptions.h
cameel: > That's gonna be circular isn't it ?

Ah, good point. Forgot that one is defined in terms of the other. It's fine then.

> Also, we've removed it from `formatErrorSeverity()` as well

Yeah, I know. I was just saying that in that particular case it wasn't very useful. But maybe it's because the code changed around it. I remember that I suggested it originally and back then it did make sense.

liblangutil/SourceReferenceExtractor.cpp
nishant-sachdeva: At this point, all the locations calling the function `printExceptionInformation` send in the full `Error::Type` . 

As of now, only one case requires printing of the truncated `Error::Severity`. The `formatErrorWithException` function calls the `formatExceptionInformation` function  in `StandardCompiler.cpp:124`  which further makes a call to `printExceptionInformation`. 

Creating overloads for one function will also require creating overloads for these other functions that are involved in the chain. 

If we change it to `variant < ...`  , we'll have to propagate that change down to quite a few other functions.

 Leads to unnecessary code redundancy. The boolean variable `printFullType` seems like a reasonable way to get our work done imo.

liblangutil/SourceReferenceFormatter.h
```diff
@@ -51,22 +51,23 @@ class SourceReferenceFormatter
...
 
 	static std::string formatExceptionInformation(
 		util::Exception const& _exception,
-		std::string const& _name,
+		Error::Type _type,
 		CharStreamProvider const& _charStreamProvider,
 		bool _colored = false,
-		bool _withErrorIds = false
+		bool _withErrorIds = false,
+		bool jsonOutput = false
```
nikola-matic:
```suggestion
		bool _jsonOutput = false
```

cameel: Good catch, this also should have been renamed as the other instances of this parameter.
```suggestion
		bool _printFullType = false
```

liblangutil/SourceReferenceExtractor.cpp
cameel: I mean, I think it would not be *that* much in terms of extra code and having the overloads would improve readability.

But it's just a suggestion. I leave that up to your judgement. If you think it's better for it to stay as is, that's fine too.

liblangutil/Exceptions.h
```diff
@@ -238,35 +248,50 @@ class Error: virtual public util::Exception
...
+		case Type::CodeGenerationError: return "CodeGenerationError";
+		case Type::DeclarationError: return "DeclarationError";
+		case Type::DocstringParsingError: return "DocstringParsingError";
+		case Type::ParserError: return "ParserError";
+		case Type::SyntaxError: return "SyntaxError";
+		case Type::TypeError: return "TypeError";
+		case Type::UnimplementedFeatureError: return "UnimplementedFeatureError";
+		case Type::YulException: return "YulException";
+		case Type::SMTLogicException: return "SMTLogicException";
+		case Type::Warning: return "Warning";
```
aarlt, line 287: Why is "Warning" and "Info" also here? I thought this will then be managed by the severity?

liblangutil/SourceReferenceExtractor.cpp
aarlt: Hmm.. not sure, because I just scrolling down here.. I was wondering whether it could make sense to have both at the same time.. e.g. I saw in the top `Type::Warning` or `Type::Info` - I thought that this will be just removed and managed by severity.. if we have both at the same time we could just "warn" about specific type errors - but somehow it looks like that these errors are always fatal. So maybe it makes no sense to be able to treat some of those errors as just as a "warning" or "info".

aarlt: wouldn't it not make more sense to have "severity" and something like a "message" - where both are used to define error, warning or info "messages", where error messages are always fatal or something.. I was just thinking about this now - I didn't checked the whole context - sorry if it makes no sense at all :D

liblangutil/Exceptions.h
cameel: Because we don't really have different types of warnings and infos, like we do for errors. Perhaps it would be better to decouple severity from type but that's out of scope of the refactor here. We could discuss this in #13410 instead.

libsolidity/interface/StandardCompiler.cpp
nishant-sachdeva: Hi, so, after taking help from @matheusaaguiar  , a command line test has been added to possibly cover the case we are mentioning here. 

But it seems to be passing. Perhaps you could take a look and see if we missed anything on the test case ?

test/cmdlineTests/external_function_pointers_taking_calldata_args/output.json
```diff
@@ -0,0 +1 @@
+{"errors":[{"component":"general","formattedMessage":"Cannot import url (\"/projects/contracts.sol\"): File not found. Searched the following locations: \"\".","message":"Cannot import url (\"/projects/contracts.sol\"): File not found. Searched the following locations: \"\".","severity":"error","type":"IOError"},{"component":"general","formattedMessage":"Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".","message":"Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".","severity":"error","type":"IOError"},{"component":"general","formattedMessage":"Cannot import url (\"/tests/test.sol\"): File not found. Searched the following locations: \"\".","message":"Cannot import url (\"/tests/test.sol\"): File not found. Searched the following locations: \"\".","severity":"error","type":"IOError"}],"sources":{}}
```
cameel: Please add an `args` file with `--pretty-json --json-indent 4` to get this nicely formatted.

See how we do it in [`standard_debug_info_in_yul_and_evm_asm_print_all`](https://github.com/ethereum/solidity/blob/develop/test/cmdlineTests/standard_debug_info_in_yul_and_evm_asm_print_all) for example.

test/cmdlineTests/external_function_pointers_taking_calldata_args/input.json
```diff
@@ -0,0 +1,13 @@
+{
+    "language": "Solidity",
+    "sources": {
+        "url_not_found.sol": {
+            "urls": [
+                "/projects/contracts.sol",
+                "contract.sol",
+                "/tests/test.sol"
```
cameel: Please don't use arbitrary absolute paths in tests. They could exist on someone's machine.

Use paths relative to the current test dir, since we can assume that there are no stray files in it. For example:

```suggestion
                "external_function_pointers_taking_calldata_args/non-existent-contract-1.sol",
                "external_function_pointers_taking_calldata_args/contract.sol",
                "external_function_pointers_taking_calldata_args/non-existent-contract-2.sol"
```

test/cmdlineTests/external_function_pointers_taking_calldata_args/contract.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity >=0.7.0 <0.9.0;
+
+contract C {
+    function d() external {}
+}
```
cameel: We don't really need the contract here. The file existing will be enough for it to compile just fine.

Also, please don't hard-code solidity versions in these tests.
```suggestion
pragma solidity *;
```

And, once this test starts working, you might want to add the SPDX comment as well to avoid a warning.

test/cmdlineTests/external_function_pointers_taking_calldata_args/input.json
```diff
@@ -0,0 +1,13 @@
+{
+    "language": "Solidity",
+    "sources": {
+        "url_not_found.sol": {
+            "urls": [
+                "/projects/contracts.sol",
+                "contract.sol",
```
cameel: This is not the correct path. In external tests the working dir is `test/cmdlineTests/`, not the directory of the test itself (BTW, I think we should change that eventually). So you need to add the test dir name as well:
```suggestion
                "external_function_pointers_taking_calldata_args/contract.sol",
```

This is why you're getting `IOError`s for all three files.

libsolidity/interface/StandardCompiler.cpp
cameel: > But it seems to be passing. Perhaps you could take a look and see if we missed anything on the test case ?

The test is passing because the expectations are wrong. According to your `output.json` you expect the input to produce 3 `IOError`s (`"severity":"error","type":"IOError"`). But that's not what should happen. Here's what 0.8.16 outputs when `contract.sol` does exist:

```json
{
    "errors":
    [
        {
            "component": "general",
            "formattedMessage": "Cannot import url (\"/projects/contracts.sol\"): File not found. Searched the following locations: \"\".",
            "message": "Cannot import url (\"/projects/contracts.sol\"): File not found. Searched the following locations: \"\".",
            "severity": "warning",
            "type": "IOError"
        },
    ]
    "sources":
    {
        "url_not_found.sol":
        {
            "id": 0
        }
    }
}
```
Note that the compiler did not issue an error for the second file and the error for the first one was converted to a warning. Your test should be passing with this as the expectation.

test/cmdlineTests/external_function_pointers_taking_calldata_args/input.json
cameel: Please also add another test that has 3 files that do not exist. Looks like in that case we do emit 3 errors and we should have it covered too.

test/cmdlineTests/external_function_pointers_taking_calldata_args/input.json
```diff
@@ -0,0 +1,13 @@
+{
```
cameel: This test is not related at all to function pointers :)

My suggestion: `external_function_pointers_taking_calldata_args/` -> `standard_urls_existing_and_missing/`

libsolidity/interface/StandardCompiler.cpp
nishant-sachdeva: Changed the test case to resolve this comment.

test/cmdlineTests/external_function_pointers_taking_calldata_args/input.json
nishant-sachdeva: Adding test case in the latest change.

liblangutil/SourceReferenceExtractor.cpp
nishant-sachdeva: I have refactored the code to work around `std::variant<Error::Type, Error::Severity>`  in a separate commit.

What I find is that, creating overloaded functions doesn't really fully solve the problem. This is because, at a higher level of abstraction, a bool/Enum is required to check which overloaded function to call. 

@cameel Maybe we can take a look how the code has turned out, and see if any changes need to be made

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -210,5 +225,7 @@ void SourceReferenceFormatter::printErrorInformation(ErrorList const& _errors)
 
 void SourceReferenceFormatter::printErrorInformation(Error const& _error)
 {
-	printExceptionInformation(SourceReferenceExtractor::extract(m_charStreamProvider, _error));
+	printExceptionInformation(
+		SourceReferenceExtractor::extract(m_charStreamProvider, _error, Error::ErrorMessageDisclaimedWith::Severity)
```
cameel: You don't really need `ErrorMessageDisclaimedWith` here. You could just use the overload that accepts severity directly:

```suggestion
		SourceReferenceExtractor::extract(m_charStreamProvider, _error, Error::errorSeverity(_error.type()))
```

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -178,13 +177,24 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
...
-	errorColored(severity) << _msg.severity;
+	errorColored(
+		std::holds_alternative<Error::Type>(_msg._typeOrSeverity) ?
+		Error::errorSeverity(std::get<Error::Type>(_msg._typeOrSeverity)) :
+		std::get<Error::Severity>(_msg._typeOrSeverity)
+	) << (
+		std::holds_alternative<Error::Type>(_msg._typeOrSeverity) ?
+		Error::formatErrorType(std::get<Error::Type>(_msg._typeOrSeverity)) :
+		Error::formatErrorSeverity(std::get<Error::Severity>(_msg._typeOrSeverity))
+	);
```
cameel: For this I would define an overload of `errorSeverity()` that accepts the variant. I would also add `Error::formatTypeOrSeverity()` taking the variant as well.

liblangutil/SourceReferenceExtractor.cpp
```diff
@@ -49,11 +69,14 @@ SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
...
-	Error const& _error
+	Error const& _error,
+	Error::ErrorMessageDisclaimedWith _errorMessageDisclaimedWith
 )
 {
-	string severity = Error::formatErrorSeverity(Error::errorSeverity(_error.type()));
-	Message message = extract(_charStreamProvider, _error, severity);
+	Message message = (_errorMessageDisclaimedWith == Error::ErrorMessageDisclaimedWith::Type) ?
+							extract(_charStreamProvider, _error, _error.type()) :
+							extract(_charStreamProvider, _error, Error::errorSeverity(_error.type()));
```
cameel:
```suggestion
	Message message = (
		_errorMessageDisclaimedWith == Error::ErrorMessageDisclaimedWith::Type) ?
		extract(_charStreamProvider, _error, _error.type()) :
		extract(_charStreamProvider, _error, Error::errorSeverity(_error.type())
	);
```

liblangutil/SourceReferenceExtractor.cpp
```diff
@@ -30,7 +30,27 @@ using namespace solidity::langutil;
...
+			secondary.emplace_back(extract(_charStreamProvider, &info.second, info.first));
+
+	return Message{std::move(primary), _type, std::move(secondary), nullopt};
+}
+
+SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
+	CharStreamProvider const& _charStreamProvider,
+	util::Exception const& _exception,
+	Error::Severity _severity
 )
```
cameel: To avoid duplication I would merge this back with the other overload and accept variant as a parameter.

liblangutil/SourceReferenceExtractor.cpp
cameel: Well, it's not great but I think we can make it better - see my suggestions. With that it should actually be pretty concise and not require the extra bool/enum.

> wouldn't it not make more sense to have "severity" and something like a "message" - where both are used to define error, warning or info "messages", where error messages are always fatal or something.

@aarlt It would make sense but like I said in https://github.com/ethereum/solidity/pull/13162#discussion_r951385781, I don't think we should get into that in this PR. It would be better off as a separate change and overall it's a very low priority thing so best if we leave it for later.

cameel (commented):
The `urls` case is working and covered now.

Just a few changes needed if actually you want to go with the variant refactor (I thought you did not like the idea; I'm fine with doing it either way :))

test/cmdlineTests/standard_urls_existing_and_missing/output.json
```diff
@@ -0,0 +1,18 @@
+{
+    "errors":
+    [
+        {
+            "component": "general",
+            "formattedMessage": "Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".",
+            "message": "Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".",
+            "severity": "warning",
+            "type": "Warning"
+        }],
```
matheusaaguiar:
```suggestion
        }
    ],
```
It seems there should be a `linebreak` here in order to fix the failing test.

matheusaaguiar: Although, after running the `cmdlineTests.sh` locally, it does produce the output **without** the  `linebreak`.
Also, running `solc --standard-json --pretty-json --json-indent 4 <standard_urls_existing_and_missing/input.json` directly gives the output **with** the `linebreak`.

cameel: I wonder why `t_ubu_cli` did not fail because of this. Weird.

liblangutil/SourceReferenceExtractor.h
```diff
@@ -55,13 +56,18 @@ namespace SourceReferenceExtractor
...
 	};
 
-	Message extract(CharStreamProvider const& _charStreamProvider, util::Exception const& _exception, std::string _severity);
+	Message extract(
+		CharStreamProvider const& _charStreamProvider,
+		util::Exception const& _exception,
+		std::variant<Error::Type, Error::Severity> _typeOrSeverity
+	);
 	Message extract(CharStreamProvider const& _charStreamProvider, Error const& _error);
+	Message extract(CharStreamProvider const& _charStreamProvider, Error::Severity _severity, Error const& _error);
```
cameel: Why not make it the last arg so that `_error` is the second one in both cases? Seems more consistent like this.
```suggestion
	Message extract(CharStreamProvider const& _charStreamProvider, Error const& _error, Error::Severity _severity);
```

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -178,13 +178,12 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 
 void SourceReferenceFormatter::printExceptionInformation(SourceReferenceExtractor::Message const& _msg)
 {
-	// exception header line
-	optional<Error::Severity> severity = Error::severityFromString(_msg.severity);
-	errorColored(severity) << _msg.severity;
+	errorColored(Error::errorSeverityOrType(_msg._typeOrSeverity)) << (Error::formatTypeOrSeverity(_msg._typeOrSeverity));
```
cameel:
```suggestion
	errorColored(Error::errorSeverityOrType(_msg._typeOrSeverity)) << Error::formatTypeOrSeverity(_msg._typeOrSeverity);
```

cameel (dismissed):
Looks good apart from the failing test.

liblangutil/SourceReferenceExtractor.h
nishant-sachdeva: Ahh, I was trying to make it more so that the `_error` is the last argument in both.

Also, there seems to be a problem with putting `_error` as second.

We have 
`extract (_charStreamProvider, _error, _severity)`
Inside this function, we make a function call to 
`extract(_charStreamProvider, _error, _typeOrSeverity)`

This results in an infinite recursion error on the CI

When I put `_severity` as the second argument, things work fine. But in the order as specified here, I start getting seg faults.  I suspect the function polymorphism here results in the wrong function being called.

test/cmdlineTests/standard_urls_existing_and_missing/output.json
nishant-sachdeva: @matheusaaguiar  indeed, the local run of the tests does produce the output without the new line. 

I suspect, for now, I can add the line break on my own and let the tests run on the CI. But we might need to investigate as to why this is happening ?

cameel: So you mean that running with `--update` option updates the output in such a way? If so then yeah, might be a good idea to look into the `cmdlineTests.sh` script.

liblangutil/SourceReferenceExtractor.h
cameel: What if you make them both accept `_typeOrSeverity` as the last argument?

From what I can see the main difference is that one takes `util::Exception`, which does not include the error ID, and the other takes `langutil::Error` and is really just a convenience wrapper that adds the ID. It really looks like all the other args in them should be the same.

test/cmdlineTests/standard_urls_existing_and_missing/output.json
nishant-sachdeva: Should we do that in another issue @cameel , @matheusaaguiar  ? 
For now I've made the changes to get the test cases passed.

cameel: I would not create an issue out of it. Just play with the script a bit and see if you can figure out why it's doing this (i.e. is this done by the compiler? does it sometimes return one form and sometimes the other?). But don't go too deep. It's not **that** important and not something that will stop this PR. We're done here :)

Only if this starts breaking other PRs, then we'll look deeper into it.

liblangutil/SourceReferenceExtractor.h
nishant-sachdeva: I mean, doing that reduces the function to 
```
SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
	CharStreamProvider const& _charStreamProvider,
	Error const& _error,
	std::variant<Error::Type, Error::Severity> _typeOrSeverity
)
{
	solAssert(std::holds_alternative<Error::Severity>(_typeOrSeverity));
	Message message = extract(_charStreamProvider, _error, std::get<Error::Severity>(_typeOrSeverity));
	message.errorId = _error.errorId();
	return message;
}
```
Here, the function will be calling itself and creating an infinite recursion. 
The other variant of this function that uses `_exception` actually implements the creation of a `Message` object. I'll be emulating that code for this function too.

liblangutil/SourceReferenceExtractor.cpp
```diff
@@ -44,20 +45,39 @@ SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
...
+
+	string const* message = boost::get_error_info<util::errinfo_comment>(_error);
+	SourceReference primary = extract(_charStreamProvider, location, message ? *message : "");
+
+	std::vector<SourceReference> secondary;
+	auto secondaryLocation = boost::get_error_info<errinfo_secondarySourceLocation>(_error);
+	if (secondaryLocation && !secondaryLocation->infos.empty())
+		for (auto const& info: secondaryLocation->infos)
+			secondary.emplace_back(extract(_charStreamProvider, &info.second, info.first));
+
```
stackenbotten: Coding style error

liblangutil/SourceReferenceExtractor.h
cameel: What if you cast `_error` in the call to `util::Exception`?

nishant-sachdeva: Yep, doing that works. 

Have done that and removed the redundant code.

liblangutil/SourceReferenceExtractor.cpp
```diff
@@ -44,16 +45,15 @@ SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
...
 }
 
 SourceReferenceExtractor::Message SourceReferenceExtractor::extract(
 	CharStreamProvider const& _charStreamProvider,
 	Error const& _error
 )
 {
-	string severity = Error::formatErrorSeverity(Error::errorSeverity(_error.type()));
-	Message message = extract(_charStreamProvider, _error, severity);
+	Message message = extract(_charStreamProvider, util::Exception(_error), _error.type());
```
cameel:
```suggestion
	Message message = extract(_charStreamProvider, static_cast<util::Exception>(_error), _error.type());
```

liblangutil/SourceReferenceExtractor.h
```diff
@@ -55,13 +56,22 @@ namespace SourceReferenceExtractor
...
+		CharStreamProvider const& _charStreamProvider,
+		util::Exception const& _exception,
+		std::variant<Error::Type, Error::Severity> _typeOrSeverity
+	);
 	Message extract(CharStreamProvider const& _charStreamProvider, Error const& _error);
+	Message extract(
+		CharStreamProvider const& _charStreamProvider,
+		util::Exception const& _exception,
+		std::variant<Error::Type, Error::Severity> _typeOrSeverity
+	);
```
cameel: You have two identical declarations here.

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -210,5 +214,11 @@ void SourceReferenceFormatter::printErrorInformation(ErrorList const& _errors)
 
 void SourceReferenceFormatter::printErrorInformation(Error const& _error)
 {
-	printExceptionInformation(SourceReferenceExtractor::extract(m_charStreamProvider, _error));
+	SourceReferenceExtractor::Message message = SourceReferenceExtractor::extract(
+													m_charStreamProvider,
+													util::Exception(_error),
+													Error::errorSeverity(_error.type())
+												);
```
cameel:
```suggestion
	SourceReferenceExtractor::Message message = SourceReferenceExtractor::extract(
		m_charStreamProvider,
		static_cast<util::Exception>(_error),
		Error::errorSeverity(_error.type())
	);
```

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -210,5 +214,11 @@ void SourceReferenceFormatter::printErrorInformation(ErrorList const& _errors)
...
+	message.errorId = _error.errorId();
```
cameel: Why did you remove `_typeOrSeverity` arg from the second overload of `extract()`? I mean, the function is useless if you have to bypass it and move the ID over manually. I think it should have worked. If it does not then in the end changing the order of args would be better than this.

test/cmdlineTests/standard_urls_existing_and_missing/output.json
```diff
@@ -0,0 +1 @@
+{"errors":[{"component":"general","formattedMessage":"Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".","message":"Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".","severity":"warning","type":"Warning"}],"sources":{"url_not_found.sol":{"id":0}}}
```
cameel, line 1: Looks like the `args` files disappeared here and in the other JSON test so this reverted back to being unformattted.

nishant-sachdeva: Yea. That's what made the tests pass. Else we were having the issues brought up in the comment here https://github.com/ethereum/solidity/pull/13162#discussion_r967667609

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -210,5 +214,11 @@ void SourceReferenceFormatter::printErrorInformation(ErrorList const& _errors)
 
 void SourceReferenceFormatter::printErrorInformation(Error const& _error)
 {
-	printExceptionInformation(SourceReferenceExtractor::extract(m_charStreamProvider, _error));
+	SourceReferenceExtractor::Message message =
```
stackenbotten: Coding style error

test/cmdlineTests/standard_urls_existing_and_missing/output.json
cameel: But you can also make it pass by doing the small adjustment suggested there, right? This will only be a problem if this tests needs to be autoupdates which is not very often.

nishant-sachdeva: No, there are issues with the adjustments either way.

Whichever way the adjustment goes, one of 
https://app.circleci.com/pipelines/github/ethereum/solidity/26515/workflows/1537a81d-d798-4598-a7e3-4743160ca2c3/jobs/1174412 
and
https://app.circleci.com/pipelines/github/ethereum/solidity/26478/workflows/326ca1f8-e91e-4034-ba21-dfac16cd2b86/jobs/1172659

fails. 

So, far, removing the `args` file seems to be the only quick solution. Unless of course we go in depth and examine why this behaviour is happening.

nishant-sachdeva: For now, I'll proceed with removing the `args` file to enable the merge to go ahead.

cameel: Ok, right. I see where the problem is now. I thought it was consistently producing the same misformatted output but it just produces different output between release and pre-release builds. In that case I know what it it is. Reported as a separate issue: #13544.

Removing `--pretty-print` in this PR is fine then. We'll re-enable it in #7665.

liblangutil/SourceReferenceExtractor.cpp
cameel: I see you now reverted back to duplicating the code. Why? I just checked to make sure it compiles and it does:
```c++
	string severity = Error::formatErrorSeverity(Error::errorSeverity(_error.type()));
	Message message = extract(_charStreamProvider, static_cast<util::Exception>(_error), _typeOrSeverity);
	message.errorId = _error.errorId();
	return message;
```

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -210,5 +214,11 @@ void SourceReferenceFormatter::printErrorInformation(ErrorList const& _errors)
...
+		SourceReferenceExtractor::extract(
+			m_charStreamProvider,
+			_error,
+			Error::errorSeverity(_error.type())
+		);
```
cameel, line 222: This is fine with our style but a bit unusual :) We'd usually do it like this:
```suggestion
	SourceReferenceExtractor::Message message = SourceReferenceExtractor::extract(
		m_charStreamProvider,
		_error,
		Error::errorSeverity(_error.type())
	);
```

cameel (approved):

PR #13164, opened 2022-06-17
updated int assembly.rst (Sample program bug fix)

IMEF-FEMI:
fixed for loop bug in VectorSum.sumPureAsm function at line 42
# updated { data := add(dataElementLocation, 0x20) } to ----->  { dataElementLocation := add(dataElementLocation, 0x20) }
 in order to update dataElementLocation after every iteration. (Hence ending the loop)


ekpyron (approved):
Thanks, very much appreciated!

PR #13166, opened 2022-06-17
Fix for library name updates in isoltest semantics tests.

### LINKED ISSUE #13120: When updating the expected results of a test, `isoltest` removes the quotation marks from the source unit name of a library
## Description

When running `isoltest` with `accept-updates`, if a test has an obtained result different from the expected result, `isoltest` will automatically update the relevant lines of the test file and re-run the test. In the case there is a line defining a library name surrounded by quotation marks, `isoltest` will remove the quotation marks when updating the file.
If `--accept-updates` is **not** used, the same issue will happen if `u` (update) is pressed when prompted.
Also, when `optimize` is used, the issue **does not** happen.

## Environment

- Compiler version: 0.8.15-develop
- Operating system: GNU/Linux

## Steps to Reproduce
Download and compile the [solidity](https://github.com/ethereum/solidity) project, if you do not have it already.
Change the expected results for tests that have a library name defined with quotation marks surrounding it, for example, 
- `test/libsolidity/semanticTests/using/using_global_invisible.sol`
- `test/libsolidity/semanticTests/using/library_through_module.sol`
- `test/libsolidity/semanticTests/using/using_global_library.sol`

Run the `isoltest` with `accept-updates` and **do not** use `optimize`. 

```
# Download and compile the project if not already done and start from the solidity root dir
sed -i "29s/5 -> 5/5 -> 6/g" test/libsolidity/semanticTests/using/library_through_module.sol # or just manually change the file
./build/test/tools/isoltest --enforce-gas-cost --no-smt --accept-updates -t semanticTests/using/*
```

ekpyron:
Fixes https://github.com/ethereum/solidity/issues/13120.

However, this is again only a minimal hot fix for the issue.
What should really be done is to give those different kinds of function calls a proper structure instead of adding more most of the time unused fields and reusing fields with different meaning - e.g. it could be a variant.
And generally, the entire infrastructure here could probably be largely simplified.
Anyways, this adjustment here should fix the issue at hand.

### DIFF
--- a/test/libsolidity/SemanticTest.cpp
+++ b/test/libsolidity/SemanticTest.cpp
@@ -386,10 +386,7 @@ TestCase::TestResult SemanticTest::runTest(
 			// For convenience, in semantic tests we assume that an unqualified name like `L` is equivalent to one
 			// with an empty source unit name (`:L`). This is fine because the compiler never uses unqualified
 			// names in the Yul code it produces and does not allow `linkersymbol()` at all in inline assembly.
-			if (test.call().signature.find(':') == string::npos)
-				libraries[":" + test.call().signature] = m_contractAddress;
-			else
-				libraries[test.call().signature] = m_contractAddress;
+			libraries[test.call().libraryFile + ":" + test.call().signature] = m_contractAddress;
 			continue;
 		}
 		else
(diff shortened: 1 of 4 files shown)


matheusaaguiar (approved):
I confirmed that the issue is not reproducible anymore.
Approved.

Issue #10183, opened 2020-11-02
Solidity 0.5.17 installed with Homebrew produces a different output than the version downloaded from solc-bin.ethereum.org


alcuadrado:
Hi

@udwig, reported to us that he was unable to verify a contract on Etherscan, even when using the exact same solidity standard input json. We spent some time debugging it, and found this problem with solc or with its distribution:

If you install `solc` `0.5.17+commit.d19bba13` using Homebrew and downloading it from solc-bin.ethereum.org, you can get a different output for the same input.

Here's a script that can be used to reproduce it:

```bash
brew tap ethereum/ethereum
brew install ethereum/ethereum/solidity@5

wget --quiet https://gist.githubusercontent.com/alcuadrado/06c1ad0cf7e23758e17a32e829f26497/raw/b9f864cfda21d0cc9c5365fa1a1b435aec6285e8/input-to-get-different-outputs.json

wget --quiet https://solc-bin.ethereum.org/macosx-amd64/solc-macosx-amd64-v0.5.17+commit.d19bba13
chmod +x solc-macosx-amd64-v0.5.17+commit.d19bba13

echo '\n\n\n'
echo Brew solc
solc --version
cat input-to-get-different-outputs.json | solc --standard-json | md5sum

echo '\n\n\n'
echo Native solc
./solc-macosx-amd64-v0.5.17+commit.d19bba13 --version
cat input-to-get-different-outputs.json | ./solc-macosx-amd64-v0.5.17+commit.d19bba13 --standard-json | md5sum
```

It will output

```
Brew solc
solc, the solidity compiler commandline interface
Version: 0.5.17+commit.d19bba13.Darwin.appleclang
b2051480575328ea24cdfa7cbdc71cf5  -




Native solc
solc, the solidity compiler commandline interface
Version: 0.5.17+commit.d19bba13.mod.Darwin.appleclang
a13020de9abb553ca40d7171744a8735  -
```

Another interesting thing is that Etherscan seems to be running the same version than the one distributed by Homebrew. It may actually be solcjs. I got into [this other bug](https://github.com/ethereum/solc-js/issues/460) when trying to reproduce it with solcjs.

Any idea what's going on? Are the solc-bin.ethereum.org binaries somehow corrupted? This is really important to us, as it can lead to people deploying contracts that aren't verifiable on Etherscan.

cameel:
> Version: 0.5.17+commit.d19bba13.Darwin.appleclang
[...]

PR #13171, opened 2022-06-20
Add Dockerfile for Legacy builds and cli shim helper. 

kevzettler:
This PR adds a "Legacy Dockerfile" in `scripts/Dockerfile_legacy`. This Dockerfile is an attempt to reproduce legacy solidity environments to examine ancient versions of solidity. The Dockerfile can be used to build a container that includes a `solc` executable with:

`docker build -f ./scripts/Dockerfile_legacy -t solidity-legacy .`

the container can then be used to invoke the `solc` executable:

```
docker run solidity-legac --version
solc, the solidity compiler commandline interface
Version: 0.1.0-cd6e4fd4/Release-Linux/g++/int
```

This PR also includes a bash shim that can be moved into `$PATH` to act as a proxy to the Docker container.

Creation of this Dockerfile and reproduction of the legacy environment was part of a research exercise to audit some deployed legacy contracts. The audit was looking specifically at a contract compiled with `v0.1.2` of solidity. Its not clear that this Dockerfile 100% accurately reproduces the `v0.1.2` build but comparing some deployed bytecode with output from this executable did produce a match.

I'm sharing this work in hopes it will be useful to others and maybe prompt further collaboration Its not clear what the source of truth is for the legacy source code. It appears that code was manually duplicate copied between multiple repositories at the time like: 
* https://github.com/ethereumproject/cpp-ethereum 
* https://github.com/ethereum/aleth 
* https://github.com/ethereum/cpp-ethereum-cmake 
and
* https://github.com/ethereum/solidity 

This docker file attempts to build from the `ethereum/solidity` repo but I think it is actually building from `cpp-ethereum` they seem circular dependant. Hopefully this will be helpful for anyone else doing research.

### DIFF
--- a/scripts/Dockerfile_legacy
+++ b/scripts/Dockerfile_legacy
@@ -1,0 +1,90 @@
+FROM ubuntu:14.04
+
+WORKDIR "/home"
+RUN sudo apt-get -y update
+RUN sudo apt-get -y install language-pack-en-base wget unzip
+RUN sudo dpkg-reconfigure locales
+RUN sudo apt-get -y install software-properties-common
+RUN sudo add-apt-repository -y ppa:george-edison55/cmake-3.x
+
+### need to update apt to ignore gpg checks
+### ubuntu 14.04 is using some old TLS encryption that dosen't share a cyper with modern repositories
+### https://askubuntu.com/questions/74345/how-do-i-bypass-ignore-the-gpg-signature-checks-of-apt
+#RUN wget --no-check-certificate -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
+#RUN sudo add-apt-repository "deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.7 main"
+
+## these repos were used in legacy build instructions
+## and seem to no longer exist
+# RUN sudo add-apt-repository -y ppa:ethereum/ethereum-qt
+# RUN sudo add-apt-repository -y ppa:ethereum/ethereum
+# RUN sudo add-apt-repository -y ppa:ethereum/ethereum-dev
+
+
+RUN sudo apt-get -y update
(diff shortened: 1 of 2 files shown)


leonardoalt:
Thanks for the PR and sorry for the delay in reviewing it.
I think this is pretty cool! We need to decide whether we want to host it in the repo and review it.
Any team member, feel free to review.

ekpyron:
Hm... this is indeed nice in principle, but we should probably not take up too much maintenance burden here - we have precedent for similar legacy-scripts in the repo in ``scripts/wasm-rebuild``... maybe we should move all stuff like this into a ``scripts/legacy`` subfolder with a README clarifying that everything in there is there to help in dealing with old versions, but that we won't maintain it at all times?

nishant-sachdeva:
@kevzettler Are you still working on this ? Or shall we takeover

kevzettler:
@nishant-sachdeva I have moved on. Please takeover.

cameel:
I think that we shouldn't add this to the repo unless we can ensure that it actually builds the binaries correctly. And to do that, this should run the full test suite and the [bytecode comparison](https://github.com/ethereum/solc-bin/blob/gh-pages/.github/workflows/t-bytecode-compare.yml) on the binaries. Both of which are not that easy for such old versions. But I think it's really important because it's not hard to get binaries that seem to work but do not actually reproduce the bytecode. We had such problems in the past, e.g. #10183.

If we're sure the binaries built by this are correct, we could actually build them and add them to solc-bin. But if not, this would need a huge disclaimers that it may or may not built them correctly. And in that case it might be better off living somewhere outside of the official repo.

ekpyron:
Since we haven't made progress on deciding where to put these yet, but it doesn't look like we'll merge them in the main repo, I'm closing this PR - the scripts will still remain here for reference, resp. for once we find a better place for them.

PR #13173, opened 2022-06-20
Custom knowledge base

chriseth:
This implements a custom version of the knowledge base that does not make use of the simplification rules.

It speeds up the implementation a little, but the really big problem is not yet solved. Still investigating...

The problem is that there is a quadratic number of calls to the knowledge base - for each combination of memory writes.


chriseth:
This turned out not to be much faster.

PR #13182, opened 2022-06-21
Add all path declarations in IdentifierPath annotation

Marenz:
refs #13064

### DIFF
--- a/libsolidity/analysis/NameAndTypeResolver.h
+++ b/libsolidity/analysis/NameAndTypeResolver.h
@@ -95,7 +95,7 @@ public:
 	Declaration const* pathFromCurrentScope(std::vector<ASTString> const& _path) const;
 	/// Resolves a path starting from the "current" scope, but also searches parent scopes.
 	/// Should only be called during the initial resolving phase.
-	/// @note Returns an empty vector if any component in the path was not unique or not found otherwise all declarations along the path are returned.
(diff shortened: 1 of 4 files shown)


libsolidity/analysis/NameAndTypeResolver.h
```diff
@@ -93,6 +93,10 @@ class NameAndTypeResolver
 	/// Should only be called during the initial resolving phase.
 	/// @note Returns a null pointer if any component in the path was not unique or not found.
 	Declaration const* pathFromCurrentScope(std::vector<ASTString> const& _path) const;
+	/// Resolves a path starting from the "current" scope, but also searches parent scopes.
+	/// Should only be called during the initial resolving phase.
+	/// @note Returns an empty vector if any component in the path was not unique or not found otherwise all declarations along the path are returned.
```
wechman:
```suggestion
	/// @note Returns an empty vector if any component in the path was non-unique or not found. Otherwise, all declarations along the path are returned.
```

wechman (commented):
Are you able to deliver any tests at this stage of implementation?

Marenz:
> Are you able to deliver any tests at this stage of implementation?

This is at least partially tested as it replaces/modifies the existing code (and a mistake here would make existing tests fail), but there is no explicit test for the whole declarations array yet.
I thought a bit about further possible tests but I believe the best would be to rely on the tests of the (first) class that'll use this info, which will be my Rename lsp class..

wechman (approved):

PR #13185, opened 2022-06-21
Add location for parameter names in FunctionCall ASTNode

Marenz:
refs #13064


libsolidity/ast/ASTJsonExporter.cpp
```diff
@@ -128,6 +128,16 @@ string ASTJsonExporter::sourceLocationToString(SourceLocation const& _location)
 	return to_string(_location.start) + ":" + to_string(length) + ":" + (sourceIndexOpt.has_value() ? to_string(sourceIndexOpt.value()) : "-1");
 }
 
+Json::Value ASTJsonExporter::sourceLocationsToJson(vector<SourceLocation> const& _sourceLocations) const
+{
+	Json::Value locations = Json::arrayValue;
+
+	for (SourceLocation location: _sourceLocations)
```
wechman: `location` can be const reference

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -95,6 +95,24 @@ SourceLocation const ASTJsonImporter::createSourceLocation(Json::Value const& _n
 	return solidity::langutil::parseSourceLocation(_node["src"].asString(), m_sourceNames);
 }
 
+vector<SourceLocation> ASTJsonImporter::createSourceLocations(Json::Value const& _node, size_t _size) const
```
wechman: `_size` parameter and its usage look pretty weird here.  I think it should not be a part of the function. In consequence, the `astAssert` should be moved outside too.

libsolidity/parsing/Parser.h
```diff
@@ -153,8 +153,15 @@ class Parser: public langutil::ParserBase
 	);
 	ASTPointer<Expression> parsePrimaryExpression();
 	std::vector<ASTPointer<Expression>> parseFunctionCallListArguments();
-	std::pair<std::vector<ASTPointer<Expression>>, std::vector<ASTPointer<ASTString>>> parseFunctionCallArguments();
-	std::pair<std::vector<ASTPointer<Expression>>, std::vector<ASTPointer<ASTString>>> parseNamedArguments();
+
+	using FunctionCallArguments = std::tuple<
```
wechman: Does it make sense to define this as a structure to make further processing easier and less error prone?

Marenz:
Updated

libsolidity/ast/AST.h
```diff
@@ -2094,9 +2094,10 @@ class FunctionCall: public Expression
 		SourceLocation const& _location,
 		ASTPointer<Expression> _expression,
 		std::vector<ASTPointer<Expression>> _arguments,
-		std::vector<ASTPointer<ASTString>> _names
+		std::vector<ASTPointer<ASTString>> _names,
+		std::vector<SourceLocation> _nameLocations
 	):
-		Expression(_id, _location), m_expression(std::move(_expression)), m_arguments(std::move(_arguments)), m_names(std::move(_names)) {}
+		Expression(_id, _location), m_expression(std::move(_expression)), m_arguments(std::move(_arguments)), m_names(std::move(_names)), m_nameLocations(std::move(_nameLocations)) {}
```
christianparpart: Must `m_names.length()` always match `m_nameLocations.length()`? If so, it should probably asserted.

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -876,11 +890,17 @@ ASTPointer<FunctionCall> ASTJsonImporter::createFunctionCall(Json::Value const&
...
+
+	optional<vector<SourceLocation>> sourceLocations = createSourceLocations(_node);
+
 	return createASTNode<FunctionCall>(
 		_node,
 		convertJsonToASTNode<Expression>(member(_node, "expression")),
 		arguments,
-		names
+		names,
+		sourceLocations.has_value() ?
```
christianparpart:
```suggestion
		sourceLocations ?
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2074,23 +2084,29 @@ pair<vector<ASTPointer<Expression>>, vector<ASTPointer<ASTString>>> Parser::pars
...
+	FunctionCallArguments ret;
 
 	bool first = true;
 	while (m_scanner->currentToken() != Token::RBrace)
 	{
 		if (!first)
 			expectToken(Token::Comma);
 
-		ret.second.push_back(expectIdentifierToken());
+		auto identWithLocation = expectIdentifierWithLocation();
```
christianparpart:
```suggestion
		auto identifierWithLocation = expectIdentifierWithLocation();
```

libsolidity/parsing/Parser.cpp
```diff
@@ -2074,23 +2084,29 @@ pair<vector<ASTPointer<Expression>>, vector<ASTPointer<ASTString>>> Parser::pars
...
+
+		// Add name
+		ret.parameterNames.push_back(identWithLocation.first);
```
christianparpart:
```suggestion
		ret.parameterNames.emplace_back(move(identWithLocation.first));
```

libsolidity/parsing/Parser.h
```diff
@@ -77,6 +77,14 @@ class Parser: public langutil::ParserBase
 		std::vector<ASTPointer<ModifierInvocation>> modifiers;
 	};
 
+	// Struct to share parsed function call arguments
```
christianparpart:
```suggestion
	/// Struct to share parsed function call arguments.
```

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -95,6 +95,20 @@ SourceLocation const ASTJsonImporter::createSourceLocation(Json::Value const& _n
...
 }
 
+optional<vector<SourceLocation>> ASTJsonImporter::createSourceLocations(Json::Value const& _node) const
+{
+	vector<SourceLocation> locations;
+
+	if (_node.isMember("nameLocations") && _node["nameLocations"].isArray())
+	{
+		for (auto const& val: _node["nameLocations"])
+			locations.emplace_back(langutil::parseSourceLocation(val.asString(), m_sourceNames));
```
matheusaaguiar, line 105:
```suggestion
			locations.emplace_back(createSourceLocation(val, m_sourceNames));
```
Could use the existing method (It enforces an assert that maybe is important...) ? 
Edit: I guess that `val` would not have a `src` field and that's why `createSourceLocation` can't be used, right?

Marenz: Indeed, we have an array of locations-strings here that need to be parsed directly

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -893,11 +907,17 @@ ASTPointer<FunctionCall> ASTJsonImporter::createFunctionCall(Json::Value const&
...
+
+	optional<vector<SourceLocation>> sourceLocations = createSourceLocations(_node);
+
 	return createASTNode<FunctionCall>(
 		_node,
 		convertJsonToASTNode<Expression>(member(_node, "expression")),
 		arguments,
-		names
+		names,
+		sourceLocations ?
```
matheusaaguiar, line 918: It seems that doesn't matter if `sourceLocations` has values or not, but it must exist, as it is created when not present in the output of `createSourceLocations`. In that case, wouldn't it be better to just create the vector beforehand and then let it be filled (or not) by the method? Either passed as reference or instead its size being passed and then created inside the method...

Marenz: I actually did use the method with the passed size before and someone else complained about that :sweat_smile: 

About the other suggestion, it might be a bit faster, but I am not sure it would be easier to understand and maintain than a simple function returning an optional value..

matheusaaguiar: Ah, ok! 
I can agree that the optional is indeed more straightforward.

matheusaaguiar (approved):
Looks good to me.

PR #13189, opened 2022-06-21
Enabled external tests to use relative paths to solc binary.

### LINKED ISSUE #13188: [External tests] Using a relative path for solc causes compilation error message
## Description

Running `./test/externalTests.sh` or any of the individual external test scripts with a relative path to the solc binary causes the tests to fail with an obscure Hardhat message saying "The Solidity version pragma statement in these files doesn't match any of the configured compilers in your config. Change the pragma or configure additional compiler versions in your hardhat config."

At the beginning of the script output, you can see a message saying "test/externalTests/common.sh: line 100: ./build/solc/solc: No such file or directory" so it is clear that the current directory has changed and the relative path is not to the right location anymore.

There's actually a visible attempt to correct this: when configuring hardhat.config.js there is another error message saying "realpath: ./build/solc/solc: No such file or directory".

So the best solution in my opinion is to use realpath on the argument as early as possible.

## Environment

- Compiler version: 0.8.16-develop.2022.6.21+commit.c3ea8661.mod.Linux.g++
- Framework/IDE (e.g. Truffle or Remix): well, Hardhat
- Operating system: Linux

## Steps to Reproduce

`./tests/externalTests.sh native ./build/solc/solc`

matheusaaguiar:
Fixes #13188

### DIFF
--- a/test/externalTests/bleeps.sh
+++ b/test/externalTests/bleeps.sh
@@ -28,7 +28,7 @@ REPO_ROOT=$(realpath "$(dirname "$0")/../..")
 
 verify_input "$@"
 BINARY_TYPE="$1"
-BINARY_PATH="$2"
+BINARY_PATH="$(realpath "$2")"
 SELECTED_PRESETS="$3"
 
 function compile_fn { npm run compile; }
--- a/test/externalTests/brink.sh
+++ b/test/externalTests/brink.sh
@@ -28,7 +28,7 @@ REPO_ROOT=$(realpath "$(dirname "$0")/../..")
 
 verify_input "$@"
 BINARY_TYPE="$1"
-BINARY_PATH="$2"
+BINARY_PATH="$(realpath "$2")"
 SELECTED_PRESETS="$3"
 
 function compile_fn { yarn compile; }
--- a/test/externalTests/chainlink.sh
+++ b/test/externalTests/chainlink.sh
@@ -28,7 +28,7 @@ REPO_ROOT=$(realpath "$(dirname "$0")/../..")
 
 verify_input "$@"
 BINARY_TYPE="$1"
-BINARY_PATH="$2"
+BINARY_PATH="$(realpath "$2")"
 SELECTED_PRESETS="$3"
 
 function compile_fn { yarn compile; }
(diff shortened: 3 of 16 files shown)


test/externalTests.sh
```diff
@@ -35,6 +35,10 @@ REPO_ROOT=$(realpath "$(dirname "$0")/..")
 
 verify_input "$@"
```
jaa2: It would be better to modify test/externalTests/common.sh in case someone wants to run a single external test. For instance, the same error would happen if you tried running `test/externalTests/uniswap.sh`

jaa2 (changes requested):

test/externalTests.sh
jaa2: You can simply move your changes to the top of `test/externalTests/common.sh` for this to work.

matheusaaguiar: Initially, I thought of doing that inside `verify_input`. But then I thought it would not be possible to modify the args `$@` from inside the function. Is it possible?
As an alternative, I am thinking that the function could be changed to return an string (i.e, `echo`) with the desired format. Then all scripts would need to be adjusted for that

jaa2: I don't think it is possible to do it inside `verify_input`, but because the scripts use `source` on `common.sh` it works the way you implemented it if you move your changes to, say, line 26 of `common.sh`.

matheusaaguiar: Ah, right. I didn't think of that!

test/externalTests/common.sh
```diff
@@ -24,6 +24,13 @@ set -e
 
 CURRENT_EVM_VERSION=london
 
+if [[ -f "$2" ]]
+then
+    ARGS=("$@")
+    ARGS[1]="$(realpath "$2")"
+    set -- "${ARGS[@]}"
+fi
+
```
wechman: I am not sure about this solution. It works, but looks risky and more like a workaround than a proper solution. Can't we just use `realpath` to setup "BINARY_PATH" variable for each test suite (e.g. test/externalTests/gnosis.sh:31) like we do for "REPO_ROOT?

matheusaaguiar: I was having those concerns too and was waiting to discuss it further today. Nice that you already reviewed. I think we should go this way you suggest, unless someone else has a better proposal.

wechman (dismissed):
Now it looks good to me. Please squash commits before merge.

Marenz (dismissed):

Marenz (approved):

PR #13190, opened 2022-06-22
Base2 should inherit from Base1 

cuongquangnam:


docs/contracts/inheritance.rst
```diff
@@ -146,7 +146,7 @@ seen in the following example:
         function destroy() public virtual override { /* do cleanup 1 */ Destructible.destroy(); }
     }
 
-    contract Base2 is Destructible {
+    contract Base2 is Base1 {
```
hrkrshnn, line 149: What's wrong with the current one?

PR #13191, opened 2022-06-22
docs: Disallow decimal literals with trailing dots

aiman:
Currently the documentation suggests that a decimal literal can omit the fractional part [1]:

> Decimal fractional literals are formed by a `.` with at least one number on one side. Examples include `1.`, `.1` and `1.3`.

However, commit ac68710 (May 30, 2018) disallowed trailing dots that are not followed by a number [2].

Using decimal literals of the form `1.` will actually result in a `ParserError` and so the docs should not recommend this form.

[1] https://docs.soliditylang.org/en/v0.8.15/types.html#rational-and-integer-literals
[2] https://github.com/ethereum/solidity/commit/ac6871078940f0ae4a47380091fc46ed46a63d0a


hrkrshnn (approved):
Thanks!

PR #13196, opened 2022-06-22
event and error selectors are now allowed to be compile time constants

### LINKED ISSUE #13137: Event and error selectors are not constants
## Description
Function selectors can be used to initialize constants while error and event selectors (recently introduced in #12921) are rejected as non-constant. I think that all of them should work the same way in that regard.

## Steps to Reproduce

```solidity
contract C {
    event Ev();
    error Er();
    function f() external {}

    bytes4 constant functionSelector = this.f.selector;  // OK
    bytes4 constant errorSelector = Er.selector;         // Error: Initial value for constant variable has to be compile-time constant.
    bytes32 constant eventSelector = Ev.selector;        // Error: Initial value for constant variable has to be compile-time constant.
}
```
```
Error: Initial value for constant variable has to be compile-time constant.
 --> test.sol:7:37:
  |
7 |     bytes4 constant errorSelector = Er.selector;         // Error: Initial value for constant variable has to be compile-time constant.
  |                                     ^^^^^^^^^^^

Error: Initial value for constant variable has to be compile-time constant.
 --> test.sol:8:38:
  |
8 |     bytes32 constant eventSelector = Ev.selector;        // Error: Initial value for constant variable has to be compile-time constant.
  |                                      ^^^^^^^^^^^            
```

The test passes without errors if you remove `constant` from the state variables.

## Environment

- Compiler version: 0.8.15 (`develop`)

nishant-sachdeva:
closes #13137

### DIFF
--- a/test/libsolidity/syntaxTests/errors/error_selector_as_constant.sol
+++ b/test/libsolidity/syntaxTests/errors/error_selector_as_constant.sol
@@ -1,0 +1,7 @@
+contract C {
+    error Er();
+    function f() external {}
+    bytes4 constant errorSelector = Er.selector;
+}
+// ----
+// TypeError 8349: (95-106): Initial value for constant variable has to be compile-time constant.
--- a/test/libsolidity/syntaxTests/events/event_selector_as_constant.sol
+++ b/test/libsolidity/syntaxTests/events/event_selector_as_constant.sol
@@ -1,0 +1,8 @@
+contract C {
+    event Ev();
+    function f() external {}
+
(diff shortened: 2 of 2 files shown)


Marenz:
I am gonna close this for now as there is nothing to review but the tests. Feel free to reopen when there are updates <3

cameel:
Let's keep this closed until you're done with #13252.

And actually, please only reopen it after you have some code written for it. We don't want a draft with just some tests sitting around and inflating the PR count :)

PR #13205, opened 2022-06-24
Fix segmentation fault when  importing large AST from json file

### LINKED ISSUE #13179: solc: Stack overflow importing large AST from json file
## Description
Importing an AST with `--import-ast` using a large JSON file results in a stack overflow and segmentation fault with certain output options.

A quick run with memcheck shows that it stems from a regex match.

## Environment

- Compiler version: 0.8.16-develop.2022.6.21+commit.c3ea8661.mod.Linux.g++
- Operating system: Linux

## Steps to Reproduce

Create a large AST as JSON. Here is Uniswap V3 as an example (20.2 MB) from the external tests: [uniswap_ast.zip](https://github.com/ethereum/solidity/files/8946583/uniswap_ast.zip)

Try generating IR output - currently segfaults:
```
solc --import-ast ./164dbdb3af44e9eea766cd449cf8864c.json --ir
```

Try generating assembly from IR - currently segfaults:
```
solc --import-ast ./164dbdb3af44e9eea766cd449cf8864c.json --via-ir --asm
```

Regular assembly with the `--asm` flag only works as expected for this file.

matheusaaguiar:
Fixes #13179.

### DIFF
--- a/liblangutil/CharStream.h
+++ b/liblangutil/CharStream.h
@@ -73,9 +73,15 @@ public:
 	CharStream() = default;
 	CharStream(std::string _source, std::string _name):
 		m_source(std::move(_source)), m_name(std::move(_name)) {}
+	CharStream(std::string _source, std::string _name, bool _importedFromAST):
+		m_source(std::move(_source)),
+		m_name(std::move(_name)),
+		m_importedFromAST(_importedFromAST)
+	{ }
 
 	size_t position() const { return m_position; }
 	bool isPastEndOfInput(size_t _charsForward = 0) const { return (m_position + _charsForward) >= m_source.size(); }
+	bool isImportedFromAST() const { return m_importedFromAST; }
 
 	char get(size_t _charsForward = 0) const { return m_source[m_position + _charsForward]; }
 	char advanceAndGet(size_t _chars = 1);
@@ -138,6 +144,7 @@ public:
 private:
 	std::string m_source;
 	std::string m_name;
+	bool m_importedFromAST{false};
 	size_t m_position{0};
 };
 
--- a/libyul/AsmPrinter.cpp
+++ b/libyul/AsmPrinter.cpp
@@ -278,7 +278,11 @@ string AsmPrinter::formatSourceLocation(
 	{
 		sourceIndex = to_string(_nameToSourceIndex.at(*_location.sourceName));
 
-		if (_debugInfoSelection.snippet && _soliditySourceProvider)
+		if (
+			_debugInfoSelection.snippet &&
+			_soliditySourceProvider &&
+			!_soliditySourceProvider->charStream(*_location.sourceName).isImportedFromAST()
+		)
 		{
 			solidityCodeSnippet = escapeAndQuoteString(
 				_soliditySourceProvider->charStream(*_location.sourceName).singleLineSnippet(_location)
--- a/libsolidity/interface/CompilerStack.cpp
+++ b/libsolidity/interface/CompilerStack.cpp
@@ -404,7 +404,8 @@ void CompilerStack::importASTs(map<string, Json::Value> const& _sources)
 		source.ast = src.second;
 		source.charStream = make_shared<CharStream>(
 			util::jsonCompactPrint(m_sourceJsons[src.first]),
-			src.first
+			src.first,
+			true // imported from AST
 		);
 		m_sources[path] = move(source);
 	}


matheusaaguiar:
Ok, so the solution is not so simple as I expected.

libyul/AsmParser.cpp
```diff
@@ -152,6 +152,19 @@ void Parser::fetchDebugDataFromComment()
 	);
 
 	string_view commentLiteral = m_scanner->currentCommentLiteral();
+
+	// (ugly) workaround to avoid regex seg fault with too large strings
+	auto resizeIfGt = [&commentLiteral](size_t sizeLimit) {
+		if(commentLiteral.size() > sizeLimit)
```
stackenbotten: Coding style error

libyul/AsmParser.cpp
```diff
@@ -154,6 +154,14 @@ void Parser::fetchDebugDataFromComment()
 	string_view commentLiteral = m_scanner->currentCommentLiteral();
 	match_results<string_view::const_iterator> match;
 
+	// (ugly) workaround to avoid regex seg fault with too large strings
+	size_t sizeLimit = 32 * 1024;
+	if (commentLiteral.size() > sizeLimit)
+	{
+		size_t resizeIndex = commentLiteral.find_first_of('\"');
+		commentLiteral = commentLiteral.substr(0, resizeIndex);
```
wechman: Can we do that without information lost?

matheusaaguiar: Apparently, yes. 
Nevertheless, I don't think this should be the final solution. I thought about it a little and I guess it requires an alternative method to handle the special case of large strings that regex_search can't handle.

wechman: Then, maybe we can just limit the string size in all cases without a need of differentiation between long and short strings?

matheusaaguiar: The problem is that sometimes you have more info (src location, ast id) after the code snippet and they need to be searched for and parsed. So, I implemented this workaround to only limit the string size if it would cause a seg fault.

liblangutil/CharStream.h
```diff
@@ -138,6 +141,7 @@ class CharStream
 private:
 	std::string m_source;
 	std::string m_name;
+	bool m_imported_from_AST;
```
Marenz: `bool m_imported_from_AST = false;`
or 
`bool m_imported_from_AST{false};`

would be better here (similar to the var below)

libyul/AsmPrinter.cpp
```diff
@@ -278,7 +278,10 @@ string AsmPrinter::formatSourceLocation(
 	{
 		sourceIndex = to_string(_nameToSourceIndex.at(*_location.sourceName));
 
-		if (_debugInfoSelection.snippet && _soliditySourceProvider)
+		if (_debugInfoSelection.snippet &&
```
Marenz:
```suggestion
                if (
                    _debugInfoSelection.snippet &&
```

styling not quite correct

liblangutil/CharStream.h
```diff
@@ -72,10 +72,13 @@ class CharStream
 public:
 	CharStream() = default;
 	CharStream(std::string _source, std::string _name):
-		m_source(std::move(_source)), m_name(std::move(_name)) {}
+		m_source(std::move(_source)), m_name(std::move(_name)), m_imported_from_AST(false) {}
```
Marenz: maybe use line breaks here now, when it is getting so long?

Marenz (changes requested):
lgtm, only minor changes. Please merge the commits as well :)

libyul/AsmPrinter.cpp
Marenz: (ignore my fucked up space-based indentations, github makes it hard)

liblangutil/CharStream.h
```diff
@@ -138,6 +144,7 @@ class CharStream
 private:
 	std::string m_source;
 	std::string m_name;
+	bool m_imported_from_AST{false};
```
Marenz: I just realized you used snake-case here instead of camel case. I think we use camel case here (`m_importedFromAST`) (yes the m_ is the exception for member variables)

Marenz (changes requested):
just one minor style thing I noticed then we're good to go

liblangutil/CharStream.h
matheusaaguiar: Oh, yes, I sometimes forget that ... I guess old habits die hard lol

Marenz (approved):

PR #13210, opened 2022-06-24
Display human readable type name in error messages

### LINKED ISSUE #13106: Most error messages describe objects representing errors and events as function types
## Description
Continued from #12896  . Created the issue as requested in https://github.com/ethereum/solidity/pull/13005#discussion_r877162339

In custom errors like the following : 
```
interface MyInterface {
    error MyCustomError();
}
```

The compiler mistakenly considers `MyInterface.MyCustomError` to be of type `function() pure` instead of type `error`. The latter is what should be used.

Running `solc test.sol` for the following code 
```
interface MyInterface {
    error MyCustomError(uint256, bool);
}

contract Test {
    function test() public returns(bytes4) {
        return (MyInterface.MyCustomError);
    }
}
```
gives the following error
```
Error: Return argument type function (uint256,bool) pure is not implicitly convertible to expected type (type of first return variable) bytes4.
 --> test.sol:7:16:
  |
7 |         return (MyInterface.MyCustomError);
  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

timweri:
closes #13106
Error messages now display errors/events type as errors/events and not functions type. Test is added for each case where errors/events were previously described as function type.

For the sake of uniformity, `humanReadable()` is used (instead of `toString()`) on types for error messages even when errors/events cannot be the cause of the errors.

### DIFF
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -110,10 +110,10 @@ util::Result<TypePointers> transformParametersToExternal(TypePointers const& _pa
 	return transformed;
 }
 
-string toStringInParentheses(TypePointers const& _types, bool _short)
+string toStringInParentheses(TypePointers const& _types, bool _withoutDataLocation)
 {
 	return '(' + util::joinHumanReadable(
-		_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_short); }),
+		_types | ranges::views::transform([&](auto const* _type) { return _type->toString(_withoutDataLocation); }),
 		","
 	) + ')';
 }
@@ -1789,7 +1789,7 @@ vector<tuple<string, Type const*>> ArrayType::makeStackItems() const
 	solAssert(false, "");
 }
 
-string ArrayType::toString(bool _short) const
+string ArrayType::toString(bool _withoutDataLocation) const
 {
 	string ret;
 	if (isString())
@@ -1798,12 +1798,12 @@ string ArrayType::toString(bool _short) const
 		ret = "bytes";
 	else
 	{
-		ret = baseType()->toString(_short) + "[";
+		ret = baseType()->toString(_withoutDataLocation) + "[";
 		if (!isDynamicallySized())
 			ret += length().str();
 		ret += "]";
 	}
-	if (!_short)
+	if (!_withoutDataLocation)
 		ret += " " + stringForReferencePart();
 	return ret;
 }
@@ -2008,9 +2008,9 @@ bool ArraySliceType::operator==(Type const& _other) const
 	return false;
 }
 
-string ArraySliceType::toString(bool _short) const
+string ArraySliceType::toString(bool _withoutDataLocation) const
 {
-	return m_arrayType.toString(_short) + " slice";
+	return m_arrayType.toString(_withoutDataLocation) + " slice";
 }
 
 string ArraySliceType::humanReadableName() const
@@ -2279,10 +2279,10 @@ bool StructType::containsNestedMapping() const
 	return m_struct.annotation().containsNestedMapping.value();
 }
 
-string StructType::toString(bool _short) const
+string StructType::toString(bool _withoutDataLocation) const
 {
 	string ret = "struct " + *m_struct.annotation().canonicalName;
-	if (!_short)
+	if (!_withoutDataLocation)
 		ret += " " + stringForReferencePart();
 	return ret;
 }
@@ -2634,7 +2634,7 @@ bool UserDefinedValueType::operator==(Type const& _other) const
 	return other.definition() == definition();
 }
 
-string UserDefinedValueType::toString(bool /* _short */) const
+string UserDefinedValueType::toString(bool /* _withoutDataLocation */) const
 {
 	return *definition().annotation().canonicalName;
 }
@@ -2682,13 +2682,13 @@ bool TupleType::operator==(Type const& _other) const
 		return false;
 }
 
-string TupleType::toString(bool _short) const
+string TupleType::toString(bool _withoutDataLocation) const
 {
 	if (components().empty())
 		return "tuple()";
 	string str = "tuple(";
 	for (auto const& t: components())
-		str += (t ? t->toString(_short) : "") + ",";
+		str += (t ? t->toString(_withoutDataLocation) : "") + ",";
 	str.pop_back();
 	return str + ")";
 }
@@ -3137,15 +3137,15 @@ string FunctionType::humanReadableName() const
 	switch (m_kind)
 	{
 	case Kind::Error:
-		return "error " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _short */ true);
+		return "error " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _withoutDataLocation */ true);
 	case Kind::Event:
-		return "event " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _short */ true);
+		return "event " + m_declaration->name() + toStringInParentheses(m_parameterTypes, /* _withoutDataLocation */ true);
 	default:
-		return toString(/* _short */ false);
+		return toString(/* _withoutDataLocation */ false);
 	}
 }
 
-string FunctionType::toString(bool _short) const
+string FunctionType::toString(bool _withoutDataLocation) const
 {
 	string name = "function ";
 	if (m_kind == Kind::Declaration)
@@ -3156,7 +3156,7 @@ string FunctionType::toString(bool _short) const
 			name += *contract->annotation().canonicalName + ".";
 		name += functionDefinition->name();
 	}
-	name += toStringInParentheses(m_parameterTypes, _short);
+	name += toStringInParentheses(m_parameterTypes, _withoutDataLocation);
 	if (m_stateMutability != StateMutability::NonPayable)
 		name += " " + stateMutabilityToString(m_stateMutability);
 	if (m_kind == Kind::External)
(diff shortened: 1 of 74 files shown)


Marenz:
To get the tests working, please rebase on our latest `develop` branch

timweri:
Rebased!

Marenz (dismissed):

Marenz:
LGTM, but we should have at least one other reviewer.

chriseth:
I think this is a good change, but I don't think we should introduce the `bool short` parameter for `humanReadableName`.

What `short` does in `toString` is that it adds the data location if it's false and omits the data location if it's true.

What we should do for `humanReadableName` is always print it (if it is a reference type) for the outermost type and omit it for the inner types (because they should have the same data location anyway).

Marenz:
> What short does in toString is that it adds the data location if it's false and omits the data location if it's true.

May I suggest that instead of `_short` we call it `_withDataLocation` for `toString`? I can create an issue

timweri:
> I think this is a good change, but I don't think we should introduce the `bool short` parameter for `humanReadableName`.
> 
> What `short` does in `toString` is that it adds the data location if it's false and omits the data location if it's true.
> 
> What we should do for `humanReadableName` is always print it (if it is a reference type) for the outermost type and omit it for the inner types (because they should have the same data location anyway).

Right. I initially didn't want to change the logic of `humanReadableName` or `toString` but I ran into problems where replacing `toString(true)` with `humanReadableName()` would cause some established error messages to change. Since `humanReadableName()` calls `toString(false)` under the hood for some cases, the error message now contains data location.

If changing the data location in the message is more favourable than adding a `humanReadableBool(bool)`, I can update the PR to go that route.

> > What short does in toString is that it adds the data location if it's false and omits the data location if it's true.
> 
> May I suggest that instead of `_short` we call it `_withDataLocation` for `toString`? I can create an issue

As for this, I assume I can address the change and reference two issues in one PR?

Marenz:
> As for this, I assume I can address the change and reference two issues in one PR?

Sure, though do use an extra commit for the renaming

> where replacing toString(true) with humanReadableName() would cause some established error messages to change. 

That's alright if the new message is better, I'd say.

timweri:
> That's alright if the new message is better, I'd say.

I'm not sure if it's better so I'll include some examples here to see what your opinion is before going ahead with the change.

Also it might make the sentence confusing when the data location looks like a normal English word, and not part of the type name:
```
 Type struct Test.RecursiveStruct memory[1] memory is only valid in storage because it contains a (nested) mapping.
```
vs
```
Type struct Test.RecursiveStruct[1] is only valid in storage because it contains a (nested) mapping.
```

**Examples**:
```
syntaxTests/iceRegressionTests/recursive_struct_memory.sol: FAIL
  Contract:
    contract Test {
            struct RecursiveStruct {
                    address payable d ;
                    mapping ( uint => address payable ) c ;
                    mapping ( uint => address payable [ ] ) d ;
            }                                                                                                                                                                                                                                                                                                                                                 function func ( ) private pure {
                    RecursiveStruct [ 1 ] memory val ;                                                                                                                                                                                                                                                                                                                val ;
            }
    }

  Expected result:
    DeclarationError 2333: (157-198): Identifier already declared.
    TypeError 4061: (268-300): Type struct Test.RecursiveStruct[1] is only valid in storage because it contains a (nested) mapping.
  Obtained result:
    DeclarationError 2333: (157-198): Identifier already declared.
    TypeError 4061: (268-300): Type struct Test.RecursiveStruct memory[1] memory is only valid in storage because it contains a (nested) mapping.
```
and
```
syntaxTests/metaTypes/unsupported_arg_for_type.sol: FAIL
  Contract:
    contract Test {
        struct S { uint x; }
        function f() public pure {
            // Unsupported for now, but might be supported in the future
            type(S);
        }
    }
                                                                                                                                                                                                                                                                                                                                                    
  Expected result: 
    TypeError 4259: (154-155): Invalid type for argument in the function call. An enum type, contract type or an integer type is required, but type(struct Test.S) provided.                                                                                                                                                                        
  Obtained result: 
    TypeError 4259: (154-155): Invalid type for argument in the function call. An enum type, contract type or an integer type is required, but type(struct Test.S storage pointer) provided.
```
and
```
syntaxTests/abiEncoder/v1_call_to_v2_contract_function_pointer_accepting_struct.sol: FAIL
  Contract:
    ==== Source: A ====
    pragma abicoder               v2;

    contract C {
        struct Item {
            uint x;
        }

        function get(Item memory _item) external {}
    }
    ==== Source: B ====
    pragma abicoder v1;
    import "A";

    contract Test {
        function foo() public {
            C c = new C();
            function(C.Item memory) external ptr = c.get;
            ptr(C.Item(5));
        }
    }

  Expected result:
    TypeError 2443: (B:166-175): The type of this parameter, struct C.Item, is only supported in ABI coder v2. Use "pragma abicoder v2;" to enable the feature.
  Obtained result:
    TypeError 2443: (B:166-175): The type of this parameter, struct C.Item memory, is only supported in ABI coder v2. Use "pragma abicoder v2;" to enable the feature.
```

Marenz:
>  Type struct Test.RecursiveStruct memory[1] memory is only valid in storage because it contains a (nested) mapping.

How did you end up with `memory[1] memory` here? :)

Apart from that, I agree that they could be confused for normal words, but I think so far we've been using them this way. 
I personally think having a better way would be good.
Maybe we can think and discuss a syntax people are happy with. e.g. `memory(Test.recursive.Struct)` or ` Test.RecursiveStruct{m}` etc. But that's out of the scope of the PR I'd say.

timweri:
> > Type struct Test.RecursiveStruct memory[1] memory is only valid in storage because it contains a (nested) mapping.
> 
> How did you end up with `memory[1] memory` here? :)

I used the default `ArrayType::humanReadableName`, which gives that result. Through that I see what @chriseth meant about `ReferenceType` in his comment. I then overrode `ArrayType::humanReadableName` to only print the data location for the outer type and obtained:
```
Type struct Test.RecursiveStruct[1] memory is only valid in storage because it contains a (nested) mapping.
> 
```
```
TypeError 2428: (B:85-105): The type of return parameter 1, string[] memory, is only supported in ABI coder v2. Use "pragma abicoder v2;" to enable the feature.
```

I assume this is the desired result. I'll look for the remaining cases for `ReferenceType`.

> Apart from that, I agree that they could be confused for normal words, but I think so far we've been using them this way. I personally think having a better way would be good. Maybe we can think and discuss a syntax people are happy with. e.g. `memory(Test.recursive.Struct)` or ` Test.RecursiveStruct{m}` etc. But that's out of the scope of the PR I'd say.

I agree :).

Marenz:
That does look indeed like the desired result.
Oh and, I won't be creating an extra issue for the parameter rename if you do it right in this PR, just fyi :)

libsolidity/ast/Types.cpp
```diff
@@ -1808,6 +1808,23 @@ string ArrayType::toString(bool _short) const
 	return ret;
 }
 
+string ArrayType::humanReadableName() const {
```
Marenz: Style nitpick. We put the `{` on its own line

test/libsolidity/syntaxTests/errors/error_incompatible_binary_ops.sol
```diff
@@ -0,0 +1,49 @@
...
+        MyCustomError << MyCustomError;
+        MyCustomError >> MyCustomError;
+        MyCustomError ^ MyCustomError;
+        MyCustomError | MyCustomError;
+        MyCustomError & MyCustomError;
+
+        MyCustomError * MyCustomError;
+        MyCustomError / MyCustomError;
+        MyCustomError % MyCustomError;
+        MyCustomError * MyCustomError;
```
Marenz: `*` is twice

test/libsolidity/syntaxTests/events/event_incompatible_binary_ops.sol
```diff
@@ -0,0 +1,48 @@
...
+        MyCustomEvent << MyCustomEvent;
+        MyCustomEvent >> MyCustomEvent;
+        MyCustomEvent ^ MyCustomEvent;
+        MyCustomEvent | MyCustomEvent;
+        MyCustomEvent & MyCustomEvent;
+
+        MyCustomEvent * MyCustomEvent;
+        MyCustomEvent / MyCustomEvent;
+        MyCustomEvent % MyCustomEvent;
+        MyCustomEvent * MyCustomEvent;
```
Marenz: twice agian

Marenz (changes requested):
found only minor things. Nice PR!

timweri:
I choked a little on rebasing :D... I'll address the style comments in a bit.

test/libsolidity/syntaxTests/events/event_incompatible_binary_ops.sol
timweri: fixed!

test/libsolidity/syntaxTests/errors/error_incompatible_binary_ops.sol
timweri: fixed!

libsolidity/ast/Types.cpp
timweri: fixed!

Marenz (changes requested):
I believe the only thing missing now is a entry in the changelog, something like `Changed some errors and warnings to be more informative and readable` maybe?

timweri:
Yes I can add that to the change log

Marenz (approved):

timweri:
Thanks for your help @Marenz :)

Marenz:
And thank you for your contribution :)

PR #13215, opened 2022-06-25
Explained Checks-Effects-Interactions and added info on Checks-Effects-Events-Interactions

lukehutch:


lukehutch:
Small tweak: "makes" (twice in the first paragraph) should read "make"

Marenz:
I can't speak for the correctness of the content, but to get your CI tests fixed you need to rebase on our latest develop.
Further, please amend the commits into one (I see a merge commit)

docs/security-considerations.rst
```diff
@@ -116,10 +116,27 @@ outlined further below:
...
     }
 
-Note that re-entrancy is not only an effect of Ether transfer but of any
-function call on another contract. Furthermore, you also have to take
-multi-contract situations into account. A called contract could modify the
-state of another contract you depend on.
+The Checks-Effects-Interactions pattern ensures that all code paths through a contract complete all required verification
+of supplied parameters before modifying contract state (Checks), then makes any changes to contract state (Effects),
+then only makes calls to functions in other contracts *after* all planned changes to contract state have been written to
+storage (Interactions). This is the only completely foolproof way to prevent *reentrancy attacks*, where a call to an
```
leonardoalt, line 122: While I may be in favor of a little section on CEI, I disagree with this particular wording. This is **not** the only completely foolproof way to prevent reentrancy. Custom logic is often enough to guarantee that reentrancy is not harmful.

lukehutch:
@leonardoalt it's a dangerous precedent though to say to your users "you can use this, but you have to be smarter than your potential attackers at reasoning your way through a complex set of logic that proves than any state they can manipulate in the middle of a transaction will ultimately have no negative effect on any user". Far better to make a blanket statement that calling any contract before state is finalized is extremely likely to open you to security vulnerabilities, which it is, and that these security vulnerabilities have lost hundreds of millions of dollars for victims, which they have.

leonardoalt:
> @leonardoalt it's a dangerous precedent though to say to your users "you can use this, but you have to be smarter than your potential attackers at reasoning your way through a complex set of logic that proves than any state they can manipulate in the middle of a transaction will ultimately have no negative effect on any user". Far better to make a blanket statement that calling any contract before state is finalized is extremely likely to open you to security vulnerabilities, which it is, and that these security vulnerabilities have lost hundreds of millions of dollars for victims, which they have.

Sure, I haven't advocated for that at all. What I disagree with is the imprecise wording in the particular line I commented. I'm all for recommending CEIs, but it is simply *incorrect* to say that this is the **only** way.

lukehutch:
OK, I understand your point better now. Thanks for the feedback.

My main reasons for creating this PR is that the current wording is a bit too weak -- users need to know what a huge source of potential vulnerabilities it is to call external contracts before finalizing state -- and also to point out that events should also only be emitted after finalizing state, because otherwise it is possible to trick 3rd parties into believing false things about the final state of the contract after the transaction. (This latter point is raised almost nowhere.)

I'm happy for my suggestion to be completely rewritten, taking these points into account. Unfortunately I don't have bandwidth for this, I was just trying to be helpful by submitting a ready-made change.

leonardoalt:
Sounds good, thanks for it. If you don't mind then someone from the team will rewrite some of the wording and we'll bring it up in the call whether we want to have a section for CEIs.

lukehutch:
Please go ahead, thanks. You can close this PR whenever somebody writes something better!

leonardoalt:
We'll just try to keep this PR and adjust on top of it.

leonardoalt:
I tried pushing my changes directly to the given repo/branch, which is usually possible when external contributors make PRs to this repo, but this was not possible. So I made another branch/PR: https://github.com/ethereum/solidity/pull/13457 (keeping author's credit) and am closing this one.

PR #13219, opened 2022-06-27
Add whiskers test for weird case

axic:


test/libsolutil/Whiskers.cpp
```diff
@@ -85,6 +85,13 @@ BOOST_AUTO_TEST_CASE(conditional_with_else)
 	BOOST_CHECK_EQUAL(Whiskers(templ)("b", false).render(), "Y");
 }
 
+BOOST_AUTO_TEST_CASE(broken_conditional_with_else)
+{
+	string templ = "<?b>X<!bY</b>";
```
axic, line 90: This is `<?b> X <!b Y </b>`. The correct one is `<!b>`.

The problem is whiskers looks for `<..>` as opposed to tokenizing the input, so this is not really possible to detect easily, however we could have some heuristics:
 match for `<?`, `<!` and `</` and only allow identifiers which are terminated with `>`, otherwise it is invalid.

Marenz (approved):

PR #13227, opened 2022-06-29
functions taking calldata args should be assignable to function pointers of same type

### LINKED ISSUE #12778: Functions taking `calldata` arguments are not assignable to function pointers of the same type
## Description
The compiler seems to treat pointers to functions taking `calldata` arguments as if they had `memory` arguments instead. You cannot assign them, pass into a function of use in `abi.encodeCall()` even when the target type exactly matches the type of the function pointer.

### `calldata` is not accepted
```solidity
contract C {
    function f(function (string calldata) external) external {}
    function g(string calldata) external {}

    function main() external {
        function (string calldata) external ptr = this.g;
        abi.encodeCall(this.f, (this.g));
        this.f(this.g);
    }
}
```
```
Error: Type function (string memory) external is not implicitly convertible to expected type function (string calldata) external.
 --> test.sol:6:9:
  |
6 |         function (string calldata) external ptr = this.g;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Cannot implicitly convert component at position 0 from "function" to "function".
 --> test.sol:7:32:
  |
7 |         abi.encodeCall(this.f, (this.g));
  |                                ^^^^^^^^

Error: Invalid type for argument in function call. Invalid implicit conversion from function (string memory) external to function (string calldata) external requested.
 --> test.sol:8:16:
  |
8 |         this.f(this.g);
  |                ^^^^^^
```

### `memory` is accepted instead
On the other hand this compiles without errors:
```solidity
contract C {
    function g(bytes calldata b) external returns (bytes calldata) {
        return b[2:5];
    }

    function main() external returns (bytes memory) {
        function (bytes memory) external returns (bytes memory) ptr = this.g;
        bytes memory b = "123456789";
        return ptr(b);
    }
}
```

I checked in Remix and it even properly slices the `memory` string argument even though we only support slicing `calldata`.

This should probably be disallowed given that you can't normally pass in `memory` arguments to functions taking `calldata`. Also, when the functions are `internal`, the compiler does produce an error:
```
Error: Type function (bytes calldata) returns (bytes calldata) is not implicitly convertible to expected type function (bytes memory) returns (bytes memory).
 --> test.sol:7:9:
  |
7 |         function (bytes memory) internal returns (bytes memory) ptr = g;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Environment
- Compiler version: 0.8.12

nishant-sachdeva:
partially solves #12778 

- [x]  Document this case.
- [-] Improve the error message. ( postponing this for a future PR. Will be resolved once a decision is taken regarding https://github.com/ethereum/solidity/issues/12778#issuecomment-1212996300 )
- [x] Add semantic and syntax tests covering all possible location combinations here.

### DIFF
--- a/test/libsolidity/semanticTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
+++ b/test/libsolidity/semanticTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
@@ -1,0 +1,10 @@
+contract C {
+    function g(string calldata) external returns (bool) { return true; }
+
+    function main() external returns (bool) {
+        function (string memory) external returns (bool) ptr = this.g;
+        return ptr("testString");
+    }
+}
+// ----
+// main() -> true
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -789,6 +789,18 @@ This includes private, internal and public functions of both contracts and libra
 functions.
 External function types, on the other hand, are only compatible with public and external contract
 functions.
+
+.. note::
+    External functions with ``calldata`` parameters are incompatible with external function types with ``calldata`` parameters.
+    They are compatible with the corresponding types with ``memory`` parameters instead.
+    For example, there is no function that can be pointed at by a value of type ``function (string calldata) external`` while
+    ``function (string memory) external`` can point at both ``function f(string memory) external {}`` and
+    ``function g(string calldata) external {}``.
+    This is because for both locations the arguments are passed to the function in the same way.
+    The caller cannot pass its calldata directly to an external function and always ABI-encodes the arguments into memory.
+    Marking the parameters as ``calldata`` only affects the implementation of the external function and is
+    meaningless in a function pointer on the caller's side.
+
 Libraries are excluded because they require a ``delegatecall`` and use :ref:`a different ABI
 convention for their selectors <library-selectors>`.
 Functions declared in interfaces do not have definitions so pointing at them does not make sense either.
--- a/test/libsolidity/syntaxTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
+++ b/test/libsolidity/syntaxTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
@@ -1,0 +1,14 @@
+contract C {
+    function f(function (string calldata) external) external {}
+    function g(string calldata) external {}
+
+    function main() external {
+        function (string calldata) external ptr = this.g;
+        abi.encodeCall(this.f, (this.g));
+        this.f(this.g);
+    }
+}
+// ----
+// TypeError 9574: (161-209): Type function (string memory) external is not implicitly convertible to expected type function (string calldata) external.
+// TypeError 5407: (242-250): Cannot implicitly convert component at position 0 from "function (string memory) external" to "function (string calldata) external".
+// TypeError 9553: (268-274): Invalid type for argument in function call. Invalid implicit conversion from function (string memory) external to function (string calldata) external requested.
--- a/test/libsolidity/syntaxTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
+++ b/test/libsolidity/syntaxTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
@@ -1,0 +1,9 @@
+contract C {
+    function g(string calldata) external {}
+
+    function main() view external {
+        function (string memory) external ptr = this.g;
+        ptr;
+    }
+}
+// ----
--- a/test/libsolidity/syntaxTests/functionTypes/internal_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
+++ b/test/libsolidity/syntaxTests/functionTypes/internal_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
@@ -1,0 +1,9 @@
+contract C {
+    function g(bytes calldata b) pure internal {}
+
+    function main() pure external {
+        function (bytes calldata) internal ptr = g;
+        ptr;
+    }
+}
+// ----


libsolidity/ast/Types.cpp
```diff
@@ -3666,12 +3669,26 @@ FunctionTypePointer FunctionType::asExternallyCallableFunction(bool _inLibrary)
...
 	for (auto const& t: m_parameterTypes)
+	{
+		// TODO :: Add warning/error here for using calldata arguments
+		// solAssert(
+		// 	!TypeProvider::isReferenceWithLocation(t, DataLocation::CallData),
+		// 	"Calldata arguments not allowed for externally callable functions. Use memory arguments instead."
+		// );
 		if (TypeProvider::isReferenceWithLocation(t, DataLocation::CallData))
+		{
+			m_errorReporter.warning(
```
matheusaaguiar: I am not sure this is the right place to report the warning or "hint" to the user. 
Since this problem happens when converting from external functions with `calldata` arguments (which are altered here to `memory`), maybe it is necessary to identify the situation  in the context of `TypeChecker`.

docs/control-structures.rst
```diff
@@ -145,6 +145,41 @@ throws an exception or goes out of gas.
     external functions happen after any changes to state variables in your contract
     so your contract is not vulnerable to a reentrancy exploit.
 
+.. warning::
+    The type of this.g takes a memory argument, even if the function signature uses calldata.
+    This is because for external function pointers, you are only interested in the calling side and not in the called side.
+    For the caller, it does not make sense to even distinguish
+    function (string calldata) external and function (string memory) external.
```
matheusaaguiar: The text here can be rephrased to present a more informed situation to the reader. The lines here are [comments](https://github.com/ethereum/solidity/issues/12778#issuecomment-1169917941) from an ongoing discussion. 
A better approach maybe would be to first explain the error (the `calldata` arguments of external function pointers get converted internally to `memory` arguments) then explain the reasons (interest in calling side not called side...). After that, suggestion of how to amend the situation and examples of the problem can be showed to conclude.

matheusaaguiar (commented):
I think that the error reporting should be done in file `TypeChecker.cpp` at the relevant locations, i.e, places where there are conversions between `FunctionType` expressions. 
For example, assignment error will be reported inside the `if` block at [`TypeChecker.cpp:1383`](https://github.com/ethereum/solidity/blob/6a42da8d3579dc4b49fdd8679d7b1663658d70d0/libsolidity/analysis/TypeChecker.cpp#L1383), where the specific cause of the error could be checked to see if it is the one causing the problem being solved by this PR.

libsolidity/ast/Types.cpp
nishant-sachdeva: `asExternallyCallableFunction` is called from different places in the code for the different test cases that we're adding for this issue.

Instead of checking for `external` functions having `calldata` arguments around each of these instances, I thought it prudent to add the warning in this function itself. 

This way, we can be sure that, whenever we are trying to get an `external` function, that has had `calldata `arguments, which will now be converted to `memory` arguments, we can print out a warning.

nishant-sachdeva: Perhaps, we could even disallow it , as @chriseth  mentions in https://github.com/ethereum/solidity/issues/12778#issuecomment-1169917941 

@cameel  any thoughts ?

cameel: > This way, we can be sure that, whenever we are trying to get an `external` function, that has had `calldata `arguments, which will now be converted to `memory` arguments, we can print out a warning. 

The problem is that `asExternallyCallableFunction()` may be called in different contexts, not just during analysis. For example in the codegen where we don't even have access to an error reporter and nowhere really to print the error (you can print it to the console but what if the user requested StandardJSON?). I see that you actually added a reporter to `FunctionType` and I don't think that matches the current design of the type classes.

Another downside is that we can now get this warning multiple times - each time the function happens to be called anywhere. We only really want the warning when we are processing that part of the code during the analysis. Later we don't care. Best if we could have an assert that will somehow detect that we didn't catch this in the analysis but I don't think it's doable if it's just a warning and not an error.

docs/control-structures.rst
```diff
@@ -145,6 +145,41 @@ throws an exception or goes out of gas.
...
+
+    So, effectively, for external functions taking calldata arguments, the compiler treats them as if they had memory arguments instead.
+
+    You cannot assign them, pass into a function of use in abi.encodeCall() even when the target type exactly matches the
+    type of the function pointer.
+
+    For eg. All 3 lines of code in the following main() function will result in errors.
```
cameel:
```suggestion
    The type of ``this.g`` takes a memory argument, even if the function signature uses calldata.
    This is because for external function pointers, you are only interested in the calling side and not in the called side.
    For the caller, it does not make sense to even distinguish
    function (string calldata) external and function (string memory) external.

    So, effectively, for external functions taking calldata arguments, the compiler treats them as if they had memory arguments instead.

    You cannot assign them, pass into a function of use in ``abi.encodeCall()`` even when the target type exactly matches the
    type of the function pointer.

    All 3 lines of code in the following ``main()`` function will result in errors.
```

docs/control-structures.rst
```diff
@@ -145,6 +145,41 @@ throws an exception or goes out of gas.
...
+
+            function main() external {
+                function (string calldata) external ptr = this.g;
+                abi.encodeCall(this.f, (this.g));
+                this.f(this.g);
+            }
+
+        }
+
+    Here, we must observe that function (string calldata) external is a type that is not really useful.
```
cameel:
```suggestion
    Here, we must observe that ``function (string calldata) external`` is a type that is not really useful.
```

libsolidity/ast/Types.cpp
nishant-sachdeva: I've gone through the code. And by my judgement, it should be okay inserting a warning in `ContractLevelChecker.cpp::checkExternalTypeClashes(ContractDefinition const& _contract)`

The `asExternallyCallableFunction` call in L:428 https://github.com/ethereum/solidity/blob/477c40ff8cf0b240d158cff112c484658659b791/libsolidity/analysis/ContractLevelChecker.cpp#L428

here is being made for all 3 failing test cases. And the function itself seems to be part of the analysis phase. I've added a comment in a separate commit indicating the exact location until we figure out the solution fully.

@cameel  thoughts on this ?

matheusaaguiar: I consider this a better location to report the warning. Although I guess that some unrelated situations will get the warning too.
But I think that's fine and what we actually want here, right? better inform the user that external functions with calldata parameters should be instead declared with memory parameters

cameel: But how would that work? That code is processing all function definitions in all contracts. I don't think we want to warn about them having `calldata` arguments, do we? Or am I misunderstanding something here?

I think this check should go into some place that visits expressions like assignments, passing in parameters, etc.

matheusaaguiar: > But how would that work? That code is processing all function definitions in all contracts. I don't think we want to warn about them having `calldata` arguments, do we? Or am I misunderstanding something here?
> I think this check should go into some place that visits expressions like assignments, passing in parameters, etc.

I think that I was confused and not sure about the objective. Reading the discussion again, I think that we want to hint the user only when there is the type mismatch error involving pointers of function with `calldata` arguments, right? 
That is what I had understood initially and hence my [suggestion](https://github.com/ethereum/solidity/pull/13227#pullrequestreview-1065866454) earlier, pointing out to method`TypeChecker::visit(VariableDeclarationStatement const&)`.

What do you think about this @nishant-sachdeva ?

cameel: > I think that we want to hint the user only when there is the type mismatch error involving pointers of function with `calldata` arguments, right?

Yes, that's correct. `TypeChecker::visit(VariableDeclarationStatement const&)` might be one of the places where we want to do this but I don't think it's the only one. You can have assignments in expressions, you can pass parameters to functions, etc. and at all those places the conversion will happen.

Actually would be nice if we could do it in a more general way like @nishant-sachdeva tried to, but that still should be in the type checker, not in type classes. Maybe we have some helper we use for type checking conversions? If not, maybe we could create one. But hard to say if that's feasible without actually going through the code.

nishant-sachdeva: @matheusaaguiar yes, the problem was noticed when there was a type mismatch during function pointer assignment.

> I think that I was confused and not sure about the objective. Reading the discussion again, I think that we want to hint the user only when there is the type mismatch error involving pointers of function with calldata arguments, right?

From what I understand, this is a symptom of a larger issue that `calldata` arguments are not allowed for `external functions`. These are replaced with `memory` arguments at `FunctionType::asExternallyCallableFunction(bool _inLibrary) const`

Hence, I think the user should be warned about using `calldata` args in external functions because it can create problems other than function pointer assignments as well. 
>  You cannot assign them, pass into a function of use in `abi.encodeCall()` even when the target type exactly matches the type of the function pointer.

I was looking for a common place where all the external functions/ arguments are processed and add a warning there. 
>better inform the user that external functions with calldata parameters should be instead declared with memory parameters

@cameel Do you think one of `TypeChecker::visit(FunctionDefinition const& _function)`  or `void TypeChecker::typeCheckFunctionCall` would be a better place to add the check ? Something on the lines of 
```
if _function.isPartOfExternalInterface() 
{
     if  < code to check if parameters are of type calldata >
              < insert warning >
}
```

cameel: > I think the user should be warned about using calldata args in external functions because it can create problems other than function pointer assignments as well.

Ah, there seems to be a misunderstanding. You're probably thinking about `calldata` args in **function pointers** not **function definitions**? `calldata` args in external functions are actually recommended whenever possible. It's just the pointers where they might be an issue.

So far you've been pointing at places that process function definitions, not pointers. Same with `TypeChecker::visit(FunctionDefinition const& _function)` - it goes over definitions. You'd need to go through variable declarations instead. `TypeChecker::typeCheckFunctionCall` on the other hand does not seem relevant. We're interested in function pointers whether they're being called or not.

But overall that's a different direction than I thought we were going. Eventually we may want to even disallow such pointers (suggested that in https://github.com/ethereum/solidity/issues/12778#issuecomment-1169917941) but that discussion is not yet resolved. I actually did not consider simply warning about function pointers taking `calldata` args and this may be a good idea, but it's a design decision we should discuss first (you could add that to the agenda for today's call). For the time being I was assuming that the goal in this PR was just to adjust the error message, which happens when function pointers are passed around or assigned.

nishant-sachdeva: Indeed, the errors we are seeing in our test cases are at the variable assignment stage. 

But the `calldata` args of the `external function pointers` have already been converted to `memory` args before that. I'm trying to locate the place where that happens. I think, inserting a warning / error ( should we choose to disallow it completely ) there should be enough to close this issue.

matheusaaguiar:
Some ideas: 
- add a `bool` member to `FunctionType` class and then set it inside `FunctionType::asExternallyCallableFunction `, so we can track when the arguments were converted, or;
- If possible, compare parameter types as defined in the declaration of the function against the provided type for the function pointer value. I guess that this should also detect when the parameter types have been modified by `FunctionType::asExternallyCallableFunction ` and there is no need to modify `FunctionType`. Again, If possible. I am not sure this can be implemented...

I think that there is not a single point where this can be checked, but I might as well be wrong (very probable) since I am not familiar with the code. So, assuming that, I think that it is necessary to perform the check in the appropriate locations where there are assignments, parameter passing and other relevant manipulations of function pointers.

libsolidity/ast/Types.cpp
nishant-sachdeva: > We actually still have this to discuss in the agenda. Judging from what was said last time, I think it's likely that we will want to disallow such pointers after all. Or do something else that will make this warning irrelevant.
My suggestion would be to just keep the docs and test cases for now and leave the error message tweak out of the PR completely. Just post in the issue that the warning part was not done.

Closing this thread for now. Will resolve the other comments and look to wrap up this PR

stackenbotten:
There was an error when running `chk_coding_style` for commit `b94a3a77f436883f3509b295fe08f801bea20e7d`:
```
docs/control-structures.rst:153:    So, if you do supply calldata arguments to external function pointers, 

```

docs/control-structures.rst
```diff
@@ -145,6 +145,39 @@ throws an exception or goes out of gas.
     external functions happen after any changes to state variables in your contract
     so your contract is not vulnerable to a reentrancy exploit.
 
+.. warning::
```
matheusaaguiar:
```suggestion
.. warning::
 As stated previously, for an external call, all arguments have to be copied to memory. As a consequence of that, if an external function has ``calldata`` parameters, they will be internally (compulsorily ?) converted by the compiler to ``memory`` parameters. In practice, there is no actual difference between these two data locations for arguments of an external call. Thus, external function pointers cannot be assigned, passed as arguments or used in ``abi.encodeCall( )``. Attempts of these operations will result in type related errors. This can be seen in function  ``main()`` in the following example.

.. code-block:: solidity
// SPDX-License-Identifier: GPL-3.0
        pragma solidity >=0.6.2 <0.9.0;

        contract C {
            function f(function (string calldata) external) external {}

            function g(string calldata) external {}

            function main() external {
                function (string calldata) external ptr = this.g;
                abi.encodeCall(this.f, (this.g));
                this.f(this.g);
            }

        }

Although function ``g`` is declared to have ``calldata`` parameters, the function pointer ``this.g`` has its type converted to ``function (string memory) external`` and then the above contract will produce the following errors when compiled:

Error: Type function (string memory) external is not implicitly convertible to expected type function (string calldata) external.
 --> test.sol:6:9:
  |
6 |         function (string calldata) external ptr = this.g;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Cannot implicitly convert component at position 0 from "function" to "function".
 --> test.sol:7:32:
  |
7 |         abi.encodeCall(this.f, (this.g));
  |                                ^^^^^^^^

Error: Invalid type for argument in function call. Invalid implicit conversion from function (string memory) external to function (string calldata) external requested.
 --> test.sol:8:16:
  |
8 |         this.f(this.g);
  | 
                 ^^^^^^
In this case, it is recommended to use external functions with ``memory``parameters to circumvent the issue.  
```

docs/control-structures.rst
```diff
@@ -145,6 +145,39 @@ throws an exception or goes out of gas.
...
+    External function pointers only take memory arguments, even if the function signature uses calldata.
+    This is because for external function pointers, you are only interested in the calling side and not in the called side.
+    For the caller, it does not make sense to even distinguish `function (string calldata) external` and `function (string memory) external`.
```
cameel: It's RST, not Markdown. You need double backticks:
```suggestion
    For the caller, it does not make sense to even distinguish ``function (string calldata) external`` and ``function (string memory) external``.
```

docs/control-structures.rst
```diff
@@ -145,6 +145,39 @@ throws an exception or goes out of gas.
...
+    So, effectively, for external function pointers taking calldata arguments, the compiler treats them as if they had memory arguments instead.
+    So, if you do supply calldata arguments to external function pointers,
+    you cannot assign them, pass into a function of use in ``abi.encodeCall()`` even when the target type exactly matches the
```
cameel:
```suggestion
    you cannot assign them, pass into a function or use in ``abi.encodeCall()`` even when the target type exactly matches the
```

docs/control-structures.rst
```diff
@@ -145,6 +145,39 @@ throws an exception or goes out of gas.
...
+    type of the said function pointer.
+
+    For eg. All 3 lines of code in the following ``main()`` function will result in errors.
```
cameel:
```suggestion
    For example, all 3 lines of code in the following ``main()`` function will result in errors.
```

docs/control-structures.rst
cameel: @matheusaaguiar I think that the error messages make this a bit too verbose. Overall, this is just a corner case that's not very useful. Users should be made aware of it but it does not have to be prominent.

Maybe it does not even need an example. At least this test case is not a very good one. For example it uses `abi.encodeCall()` which is completely irrelevant and only makes it more complex. If I were to include an example, it would probably be something closer to this:
```solidity
pragma solidity >=0.6.2 <0.9.0;

contract C {
    function calldataFunction(string calldata) external {}
    function memoryFunction(string memory) external {}

    function assignPtr() external view {
        function (string calldata) external calldataPtr;
        function (string memory) external memoryPtr;

        // OK
        memoryPtr = this.calldataFunction;
        memoryPtr = this.memoryFunction;

        // Error: Type function (string memory) external is not implicitly
        // convertible to expected type function (string calldata) external.
        //calldataPtr = this.calldataFunction;
        //calldataPtr = this.memoryFunction;
    }
}
```

cameel (commented):
Here are some initial comments but I'll need to take a closer look later.

docs/control-structures.rst
matheusaaguiar: I thought that it was getting too long to be just a warning, but went ahead anyway because I find it easier to remove and trim than to add more text :)
Let's just remove the example and the error messages then.

docs/control-structures.rst
matheusaaguiar: Ok.

stackenbotten:
There was an error when running `chk_coding_style` for commit `b2d5bb837c64b1e1c879a194313543bf7085ebed`:
```
docs/control-structures.rst:155:    Hence, in such cases, it is recommended to use external functions with ``memory`` parameters to circumvent the issue. 

```

test/libsolidity/syntaxTests/functionCalls/external_functions_with_calldata_args_assgined_to_function_pointers_of_memory_type.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
+    function f(function (string calldata) external) external {}
```
matheusaaguiar: `function f` here can be removed. Or maybe we could pass `ptr` to it inside of `main` so we don't get the unrelated (to this PR) unused variable warning.

matheusaaguiar (commented):
It seems this is finished, just suggested a minor change.

test/libsolidity/syntaxTests/functionCalls/external_functions_with_calldata_args_assgined_to_function_pointers_of_memory_type.sol
nishant-sachdeva: No need for passing the `ptr` to function `f`. The warning can be removed otherwise as well.

Pushed changes for that

Also removing the function `f` because it doesn't serve any purpose at the moment.

docs/control-structures.rst
```diff
@@ -145,6 +145,15 @@ throws an exception or goes out of gas.
     external functions happen after any changes to state variables in your contract
     so your contract is not vulnerable to a reentrancy exploit.
 
+.. warning::
```
cameel: I think a note is enough. It's a detail that's good to have some understanding of but you don't have to be on guard and actively watch for it.
```suggestion
.. note::
```

cameel: Wouldn't this fit better under [Types > Function Types](https://docs.soliditylang.org/en/latest/types.html#function-types)? It even has a paragraph about assigning function types.

docs/control-structures.rst
```diff
@@ -145,6 +145,15 @@ throws an exception or goes out of gas.
...
+    As stated previously, for an external call, all arguments have to be copied to memory.
+    As a consequence of that, if an external function has ``calldata`` parameters,
+    they will be internally converted by the compiler to ``memory`` parameters.
+    In practice, there is no actual difference between these two data locations for arguments of an external call.
+    Thus, external function pointers cannot be assigned, passed as arguments or used in ``abi.encodeCall( )``.
+    Attempts of these operations will result in type related errors.
+    Hence, in such cases, it is recommended to use external functions with ``memory`` parameters to circumvent the issue.
```
cameel: I think that the explanation of why it's this way could be a bit clearer. I'd also put what happens first and only then explain why. How about this?

```suggestion
    External functions with ``calldata`` parameters are incompatible with external function types with ``calldata`` parameters.
    They are compatible with the corresponding types with ``memory`` parameters instead.
    For example, there is no function that can be pointed at by a value of type ``function (string calldata) external`` while
    ``function (string memory) external`` can point at both ``function f(string memory) external {}`` and
    ``function g(string calldata) external {}``.

    This is because for both locations the arguments are passed to the function in the same way.
    The caller cannot pass its calldata directly to an external function and always ABI-encodes the arguments into memory.
    Marking the parameters as ``calldata`` only affects the implementation of the external function and is
    meaningless in a function pointer on the caller's side.
```

test/libsolidity/syntaxTests/functionCalls/external_functions_with_calldata_args_assgined_to_function_pointers_of_memory_type.sol
```diff
@@ -0,0 +1,9 @@
+contract C {
```
cameel: Typo in test name: `assgined` -> `assigned`.

Also `function_pointers_of_memory_type` -> `function_pointer_with_memory_args`.

cameel: Since this works, we should also have at least one semantic test for such function pointers.

test/libsolidity/syntaxTests/functionCalls/internal_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
```diff
@@ -0,0 +1,11 @@
+contract C {
+    function g(bytes calldata b) pure internal returns (bytes calldata) {
+        return b[2:5];
+    }
```
cameel: I think that `syntaxTests/functionTypes/` would be a better location for these tests. It's more about pointers than about calls.

cameel: This bit would only be useful if this was a semantic test. In a syntax test we can keep it simple:
```suggestion
    function g(bytes calldata) pure internal returns (bytes calldata) {}
```

cameel (changes requested):

test/libsolidity/syntaxTests/functionCalls/internal_functions_with_calldata_args_assigned_to_function_pointers_of_same_type.sol
nishant-sachdeva: Done

cameel: Ah true. `b` is fine then.

nishant-sachdeva: Done.

docs/control-structures.rst
nishant-sachdeva: Changed the location of this `note` in the latest push.

docs/control-structures.rst
cameel: What about this one? It was marked as resolved but the text is still pretty much as it was before. I think it's a bit too vague about what's actually happening.

test/libsolidity/semanticTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
```diff
@@ -0,0 +1,10 @@
+contract C {
+    function g(string calldata) external {}
+
+    function main() view external returns (function (string memory) external) {
+        function (string memory) external ptr = this.g;
+        return ptr;
+    }
+}
+// ----
+// main() -> -28758981899283947592798738437886595599040542797332825194040240852133054775296
```
cameel: I have no idea if this result is correct or not and the point of a semantic test is to make sure the code calculates correct results. It would be better to add some return values and have `g` call the function it's getting and return its result (could be just the string it's getting as its argument). We could then check that result.

cameel (changes requested):

test/libsolidity/semanticTests/functionTypes/external_functions_with_calldata_args_assigned_to_function_pointers_with_memory_type.sol
nishant-sachdeva: Made relevant changes in the latest push.

docs/control-structures.rst
nishant-sachdeva: Made relevant changes in the latest push.

cameel (approved):

Issue #13224, opened 2022-06-28
ENS external test fails compilation via IR: `Could not create stack layout after 1000 iterations`


Marenz:
Our CI currently fails a required test "ens": https://app.circleci.com/pipelines/github/ethereum/solidity/25009/workflows/5155870a-20fb-421d-8e17-10c02edf8aa0/jobs/1101684

```
YulException: Could not create stack layout after 1000 iterations.


Error HH600: Compilation failed
```

Marenz:
Here is a coredump of the problem after it triggered the too deep error:

```
Catchpoint 1 (exception thrown), 0x00007fc7649e1672 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) up
#1  0x000055d442da4831 in boost::throw_exception<solidity::evmasm::StackTooDeepException> (e=...) at /usr/include/boost/throw_exception.hpp:70
70	    throw exception_detail::enable_both( e );
(gdb)
#2  0x000055d442da1a6e in solidity::evmasm::CSECodeGenerator::appendDup (this=0x7ffff6c50ab0, _fromPosition=-8, _location=...) at /home/marenz/eth/solidity/libevmasm/CommonSubexpressionEliminator.cpp:475
475	        assertThrow(instructionNum <= 16, StackTooDeepException, util::stackTooDeepString);
(gdb) bt
#0  0x00007fc7649e1672 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000055d442da4831 in boost::throw_exception<solidity::evmasm::StackTooDeepException> (e=...) at /usr/include/boost/throw_exception.hpp:70
#2  0x000055d442da1a6e in solidity::evmasm::CSECodeGenerator::appendDup (this=0x7ffff6c50ab0, _fromPosition=-8, _location=...) at /home/marenz/eth/solidity/libevmasm/CommonSubexpressionEliminator.cpp:475
#3  0x000055d442da0085 in solidity::evmasm::CSECodeGenerator::generateClassElement (this=0x7ffff6c50ab0, _c=16, _allowSequenced=false) at /home/marenz/eth/solidity/libevmasm/CommonSubexpressionEliminator.cpp:381
[...]

PR #13229, opened 2022-06-29
Fix external test ens

Marenz:
By using the last known working commit.
Only a temporary measure.
refs #13224


matheusaaguiar (approved):
Test is passing now in CI. Approved.

PR #13234, opened 2022-07-01
SMTComp changes by msooseth

msooseth:
* `Werror` removed (temporarily), because it fails to build with new gcc that way
* adding `m_` so it's more Solidity-like
* Removing conflict count from `cerr`
* adding (likely wrongly named) `STATICCOMPILE` flag for static compilation -- this may not work on Windows, actually...


chriseth (approved):

chriseth:
Rebased.

PR #13238, opened 2022-07-03
Improved stack too deep message when compiled without --optimize

### LINKED ISSUE #13237: > @shekhirin There should be a hint from 0.8.15 onwards.
> @shekhirin There should be a hint from 0.8.15 onwards.

Hmm, just updated to `0.8.15` but still don't have any hints:
```console
➜  ~ solc --version
solc, the solidity compiler commandline interface
Version: 0.8.15+commit.e14f2714.Darwin.appleclang

➜  ~ echo 'contract Test {
        event Log(address account);

        function test(address[] calldata accounts) public {
                for (uint256 i = 0; i < accounts.length; ) {
                        emit Log(accounts[i]);

                        unchecked { ++i; }
                }
        }
}' | solc --via-ir --asm -
Uncaught exception:
/solidity/libyul/backends/evm/EVMObjectCompiler.cpp(115): Throw in function void solidity::yul::EVMObjectCompiler::run(solidity::yul::Object &, bool)
Dynamic exception type: boost::wrapexcept<solidity::yul::StackTooDeepError>
std::exception::what: Variable var_i_11 is 1 slot(s) too deep inside the stack.
[solidity::util::tag_comment*] = Variable var_i_11 is 1 slot(s) too deep inside the stack.
```

_Originally posted by @shekhirin in https://github.com/ethereum/solidity/issues/13222#issuecomment-1172357653_

hrkrshnn:
Closes https://github.com/ethereum/solidity/issues/13237

### DIFF
--- a/libyul/backends/evm/EVMCodeTransform.cpp
+++ b/libyul/backends/evm/EVMCodeTransform.cpp
@@ -26,6 +26,7 @@
 #include <libyul/Utilities.h>
 
 #include <libsolutil/Visitor.h>
+#include <libsolutil/StackTooDeepString.h>
 
 #include <liblangutil/Exceptions.h>
 
@@ -785,7 +786,8 @@ size_t CodeTransform::variableHeightDiff(Scope::Variable const& _var, YulString 
 			_varName.str() +
 			" is " +
 			to_string(heightDiff - limit) +
-			" slot(s) too deep inside the stack."
+			" slot(s) too deep inside the stack. " +
+			stackTooDeepString
 		);
 		m_assembly.markAsInvalid();
 		return _forSwap ? 2 : 1;


libyul/backends/evm/EVMCodeTransform.cpp
```diff
@@ -785,7 +786,8 @@ size_t CodeTransform::variableHeightDiff(Scope::Variable const& _var, YulString
 			_varName.str() +
 			" is " +
 			to_string(heightDiff - limit) +
-			" slot(s) too deep inside the stack."
+			" slot(s) too deep inside the stack. " +
+			stackTooDeepString
```
wechman, line 790: Can we have a test case that covers this change?

hrkrshnn: It's an internal compiler error, we don't usually test for them (see: https://github.com/ethereum/solidity/pull/12983#issuecomment-1125949942).

cameel: Well, technically here it's just adding it to `m_stackErrors` which is exposed through `CodeTransform::stackErrors()` and that is sometimes used even when there's no ICE (mostly just to check if it's empty though).

But I guess this change is simple enough that I'm fine not having a test here. I suspect that we do have some tests covering the mechanism itself and it's just that we don't have anything comparing the message specifically.

cameel (approved):

PR #13239, opened 2022-07-03
Create placeholders.rst

The-Arbiter:
### Placeholder Documentation

**Why you made your change**

_When I recently read the Solidity docs, I felt that there wasn't enough information on placeholders used within modifiers._

This question has been [asked before](https://ethereum.stackexchange.com/questions/19171/what-does-underscore-do).

**What you did**

I wrote up a quick blurb on what the underscore (placeholder) operator does, with additional mentions that it is unrelated to _prefix or suffix_ uses of underscores. 

I also included a code example of a common _mutex_ modifier.

_Disclaimer: I am not experienced with .rst files, so if I have made a referencing mistake please let me know and I will fix it._

Thanks!

### DIFF
--- a/docs/contracts/placeholders.rst
+++ b/docs/contracts/placeholders.rst
@@ -6,7 +6,9 @@
 The Placeholder Operator
 ************************
 
-In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run. The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the function being modified should be run within the modifier.
+In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run.
+The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the
+function being modified should be run within the modifier.
 
 Let's take a look at how the placeholder is used within a mutex modifier:
 
--- a/docs/contracts/placeholders.rst
+++ b/docs/contracts/placeholders.rst
@@ -10,7 +10,7 @@ In function modifiers, it is necessary to tell Solidity when you want the functi
 The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the
 function being modified should be run within the modifier.
 
-Let's take a look at how the placeholder is used within a mutex modifier:
+Let's take a look at how the placeholder can be used within a ``mutex`` modifier:
 
 .. code-block:: solidity
 
--- a/docs/contracts/placeholders.rst
+++ b/docs/contracts/placeholders.rst
@@ -2,9 +2,9 @@
 
 .. _placeholders:
 
-******************
+************************
 The Placeholder Operator
-******************
+************************
 
 In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run. The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the function being modified should be run within the modifier.
 
--- a/docs/contracts/placeholders.rst
+++ b/docs/contracts/placeholders.rst
@@ -1,0 +1,25 @@
+.. index:: ! function;placeholder
+
+.. _placeholders:
+
+******************
+The Placeholder Operator
+******************
+
+In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run. The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the function being modified should be run within the modifier.
+
+Let's take a look at how the placeholder is used within a mutex modifier:
+
+.. code-block:: solidity
+
+	modifier mutex {  
+		require(!lock);  
+		lock = true;  
+		_; // This is where the function being modified will be run
+		lock = false;  
+	}
+
+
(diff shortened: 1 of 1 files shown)


The-Arbiter:
CircleCI thinks I'm writing C++. 

Let me know what I need to do and I'll do it.

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
+.. index:: ! function;placeholder
+
+.. _placeholders:
+
+******************
+The Placeholder Operator
+******************
```
cameel: In RST the frame must match the header length

```suggestion
************************
The Placeholder Operator
************************
```

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
+.. index:: ! function;placeholder
```
cameel:
```suggestion
.. index:: ! modifier;placeholder
```

cameel: To be part of the docs, this file has to be included from `docs/index.rst` or some other file.

But I don't think this really needs to be a separate page. I think it should be a part of `docs/contracts/function-modifiers.rst`.

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. _placeholders:
+
+******************
+The Placeholder Operator
+******************
+
+In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run. The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the function being modified should be run within the modifier.
```
cameel:
```suggestion
In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run.
The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the
function being modified should be run within the modifier.
```

cameel:
```suggestion
In function modifiers, it is necessary to specify when you want the function to which the modifier is applied to be run. The placeholder (denoted by an underscore character followed by a semicolon `_;` is used to denote where the function being modified should be run within the modifier.
```

cameel:
```suggestion
In function modifiers, it is necessary to tell Solidity when you want the function affected by the modifier to be run. The placeholder (denoted by a single underscore character `_`) is used to denote where the body of the function being modified should be inserted.
```

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Let's take a look at how the placeholder is used within a mutex modifier:
```
cameel:
```suggestion
Let's take a look at how the placeholder can be used within a ``mutex`` modifier:
```

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
+.. index:: ! function;placeholder
+
+.. _placeholders:
+
+******************
+The Placeholder Operator
```
cameel, line 6: It's not really an operator.
```suggestion
The Placeholder Statement
```

docs/contracts/placeholders.rst
```diff
@@ -0,0 +1,25 @@
...
+		require(!lock);  
+		lock = true;  
+		_; // This is where the function being modified will be run
+		lock = false;  
+	}
+
+
+A simple way of thinking of this modifier is as a function which wraps the original function, in which case the placeholder operator would represent where we call our original function. 
+
+Note that the placeholder operator is different from using underscores as leading or trailing characters in variable names, which is a stylistic choice.
```
cameel, line 27: I think it would be better off as a comment with your example. Then you could actually show a variable starting with an underscore and say that this does result in the function being executed.

cameel (changes requested):
Thanks for the PR! Here's some feedback.

I'm not sure if this is a good fit for the docs though and it's a bit repetitive. I think that it would be better to try to integrate it with what's already in `function-modifiers.rst`. That page does not even mention the word *placeholder* so it could be improved, but I don't think that the placeholder itself needs that much focus. It could be just a part of the bigger example that's already there.

cameel:
> CircleCI thinks I'm writing C++.
>
> Let me know what I need to do and I'll do it.

You probably mean the failing `chk_coding_style` job? It's exactly what it says, i.e. you seem to have left some whitespace past the end of line in a few places. It tells you on which lines it found it. This is flagged because randomly adding and removing this just adds noise to diffs and makes them harder to read. Please configure your editor to remove trailing whitespace automatically.

```
docs/contracts/placeholders.rst:15:	modifier mutex {  
docs/contracts/placeholders.rst:16:		require(!lock);  
docs/contracts/placeholders.rst:17:		lock = true;  
docs/contracts/placeholders.rst:19:		lock = false;  
docs/contracts/placeholders.rst:23:A simple way of thinking of this modifier is as a function which wraps the original function, in which case the placeholder operator would represent where we call our original function. 
```

The-Arbiter:
Thanks for this, I will close this one as I have created another PR for the function-modifiers.rst file.

Issue #4027, opened 2018-04-30
Run all tests on CircleCI on compiled as 32-bit Linux binary


axic:
This requires compiling Solidity to a 32-bit target first.

axic:
@ekpyron is Emscripten compiling to a 32-bit target or 64-bit? Couldn't find an answer in the docs. For the wasm output 64-bit makes sense, but for the asm.js output 32-bit makes more sense (given Javascript numbers are 53 bits).

ekpyron:
Afaik browser wasm doesn't supports 64-bit anywhere atm - and I think both the wasm and asm.js builds are 32-bits - one of the windows builds might actually be as well.

axic:
> Afaik browser wasm doesn't supports 64-bit anywhere atm - and I think both the wasm and asm.js builds are 32-bits - one of the windows builds might actually be as well.

Please note that wasm32 vs. wasm64 means 32-bit or 64-bit memory addressing. The instruction set is 64-bit in both cases.

PR #13240, opened 2022-07-03
Fix C++ compilation errors on 32-bit Raspbian due to `unsigned` -> `int` conversion

derekg:
Test cases were getting build errors with - simple problems with unsigned vs int - static_cast applied.  

- BOOST_LIB_VERSION "1_79"
- g++ (Raspbian 8.3.0-6+rpi1) 8.3.0
- 5.10.17-v7l+ armv7l GNU/Linux

### DIFF
--- a/test/Metadata.cpp
+++ b/test/Metadata.cpp
@@ -136,7 +136,7 @@ private:
 	}
 	bytes readBytes(unsigned length)
 	{
-		bytes ret{m_metadata.begin() + m_pos, m_metadata.begin() + m_pos + length};
+		bytes ret{m_metadata.begin() + static_cast<int>(m_pos), m_metadata.begin() + static_cast<int>(m_pos + length)};
 		m_pos += length;
 		return ret;
 	}
--- a/test/libevmasm/Optimiser.cpp
+++ b/test/libevmasm/Optimiser.cpp
@@ -141,7 +141,7 @@ namespace
 			ControlFlowGraph cfg(output);
 			AssemblyItems optItems;
 			for (BasicBlock const& block: cfg.optimisedBlocks())
-				copy(output.begin() + block.begin, output.begin() + block.end,
+				copy(output.begin() + static_cast<int>(block.begin), output.begin() + static_cast<int>(block.end),
 					 back_inserter(optItems));
 			output = move(optItems);
 		}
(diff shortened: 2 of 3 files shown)


cameel (approved):
I guess it should not break anything but I wonder why you're even running into this. I tried to reproduce the problem with various GCC versions (both ARM and x86) via godbolt.org and I'm getting no errors.

What is the exact error that you're getting?

derekg:
It is only a problem since `all warnings being treated as errors` 

`In file included from /home/pi/solidity/test/libsolidity/SemanticTest.h:17,
                 from /home/pi/solidity/test/InteractiveTests.h:27,
                 from /home/pi/solidity/test/soltest.cpp:40:
/home/pi/solidity/test/libsolidity/util/TestFileParser.h: In member function ‘void solidity::frontend::test::TestFileParser::Scanner::advance(unsigned int)’:
/home/pi/solidity/test/libsolidity/util/TestFileParser.h:101:31: error: conversion to ‘std::__iterator_traits<__gnu_cxx::__normal_iterator<const char*, std::__cxx11::basic_string<char> >, void>::difference_type’ {aka ‘int’} from ‘unsigned int’ may change the sign of the result [-Werror=sign-conversion]
    m_char = std::next(m_char, n);
                               ^
cc1plus: all warnings being treated as errors`

cameel:
So looks like this is because ARMv7l is 32-bit and the difference type is deduced to be `int` (and needs an explicit conversion from `unsigned`) while on x64 it probably ends up as a bigger type that can represent all values from both `int` and `unsigned`.

We should really have some 32-bit target in CI (apparently we even have an issue to do this: #4027) though I'm not sure if CircleCI even supports that. So thanks for submitting the PR, we would not have detected that otherwise. And nice that these 3 warnings were the only problems that actually were there.

BTW, I see that you closed this for some reason. I'm going to reopen and merge.

derekg:
I think actually some on 32bit envs  `-Werror=sign-conversion ` flag is getting set but it's not set on 64bit linux or other environments.   A simple test example 

`
#include <iostream>
using namespace std;

int main(int argc, char **argv) {
        unsigned x = 1;
        int y = 1;
        y = x + y;
        cout << x << " " << sizeof(unsigned) << endl;
        cout << y << " " << sizeof(int) << endl;
        return 0;
}
`
Compiles fine with -Wall but if -Werror=sign-conversion  is included on armv7l or **x86_64**  then it fails.   Hope this is helpful. I might investigate further.

Issue #12168, opened 2021-10-20
Wrong code snippets printed for errors in `--import-ast` mode


cameel:
## Description
If you save the AST using `solc --combined-json ast` and then load it with `solc --import-ast`, the code snippets included in errors printed to the console show bits of the JSON input rather than actual code.

Since the original source is not available in the AST import mode, the only solution seems to be to show only file name and line number.

## Environment

- Compiler version: 0.8.10 (`develop`)

## Steps to Reproduce
```bash
echo 'contract C { function foo(uint x) public { require(x > 0); } }' \
    solc --combined-json ast - | \
    solc --import-ast -
```
Output (first `solc` command):
```
Warning: Function state mutability can be restricted to pure
 --> <stdin>:1:14:
  |
1 | contract C { function foo(uint x) public { require(x > 0); } }
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Output (second `solc` command):
```
Warning: Function state mutability can be restricted to pure
 --> <stdin>:1:14:
  |
1 | {"absolutePath":"<stdin>","exportedSymbols":{"C":[13]},"id":14,"nodeType":"SourceUnit","nodes": ...
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

chriseth:
This is a known limitation, and I think it is better to preserve the debug info than removing it. Also note that json ast import is a very rarely used feature, so we should be careful not to complicate the debug info just because of this.

cameel:
Can't we just print the file name and line number without printing the snippet?

chriseth:
We can, but we first need to know about this situation.

cameel:
It was fixed by #13244.

PR #13244, opened 2022-07-04
Remove wrong code snippets from error message when in import-ast mode

matheusaaguiar:
This should fix the issue #12168.

### DIFF
--- a/liblangutil/SourceReferenceFormatter.cpp
+++ b/liblangutil/SourceReferenceFormatter.cpp
@@ -124,6 +124,9 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 
 	string_view text = _ref.text;
 
+	if (m_charStreamProvider.charStream(_ref.sourceName).isImportedFromAST())
+		return;
+
 	if (!_ref.multiline)
 	{
 		size_t const locationLength = static_cast<size_t>(_ref.endColumn - _ref.startColumn);


liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -124,6 +124,9 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 
 	string_view text = _ref.text;
 
+	if(m_charStreamProvider.charStream(_ref.sourceName).isImportedFromAST())
```
stackenbotten: Coding style error

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1286,7 +1286,7 @@ void CompilerStack::assemble(
 			"Contract code size is "s +
 			to_string(compiledContract.runtimeObject.bytecode.size()) +
 			" bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). "
-			"This contract may not be deployable on Mainnet. "
```
Marenz: Bad rebase?

Marenz (changes requested):

liblangutil/SourceReferenceFormatter.cpp
```diff
@@ -124,6 +124,9 @@ void SourceReferenceFormatter::printSourceLocation(SourceReference const& _ref)
 
 	string_view text = _ref.text;
 
+	if (m_charStreamProvider.charStream(_ref.sourceName).isImportedFromAST())
+		return;
```
Marenz, line 128: Could we have at least one test that shows the changes?

libsolidity/interface/CompilerStack.cpp
matheusaaguiar: Yeah, maybe, not sure what happened here ...

matheusaaguiar:
Replicating the same test used in issue [12168](https://github.com/ethereum/solidity/issues/12168):
```
echo 'contract C { function foo(uint x) public { require(x > 0); } }' | \
    solc --combined-json ast - | \
    solc --import-ast -
```
Output of first `solc` command:

```
Warning: Function state mutability can be restricted to pure
 --> <stdin>:1:14:
  |
1 | contract C { function foo(uint x) public { require(x > 0); } }
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
**Output of second `solc` command (Affected by the current PR):**
```
Warning: Function state mutability can be restricted to pure
 --> <stdin>:1:14:
```

matheusaaguiar:
Looking at it now, i think would be good to include a informative message letting the user know that the function code snippet cannot be shown because the source was imported from json.
edit: nevermind. `--import-ast` is rarely used and the user probably will figure this out.

Marenz (approved):

PR #13246, opened 2022-07-05
feat(parsing): allow trailing commas in some places

### LINKED ISSUE #2991: Re-enable trailing commas in function parameters, arrays, etc.
I don't know why this functionality was removed. Diffs are much cleaner if trailing comma are allowed.

EMCA supports trailing commas: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

shekhirin:
Partially resolves https://github.com/ethereum/solidity/issues/2991

Allowing trailing commas everywhere wouldn't be possible for the current state of Solidity syntax since it has tuples with empty elements that can be interpreted ambiguously with trailing commas:
```solidity
contract C {
	function f() public pure {
		(uint a,) = (uint(1),);
	}
}
```
Should we parse `(uint(1),)` as one element tuple or as two elements tuple with NULL as second element? It seems like a breaking change but I'd be happy to be wrong.

---

What isn't a breaking change is allowing trailing commas in:
1. Function/modifier/event/error/try/catch parameters, return parameters
2. Function call arguments
3. Enum definitions
4. Inline array definitions
5. Using directives
6. Override specifiers
7. Import aliases

We can start with it, making life easier for some refactorings and unifying this trailing comma rule across other languages that allow it.

### DIFF
--- a/docs/grammar/SolidityParser.g4
+++ b/docs/grammar/SolidityParser.g4
@@ -102,7 +102,7 @@ namedArgument: name=identifier Colon value=expression;
  * Arguments when calling a function or a similar callable object.
  * The arguments are either given as comma separated list or as map of named arguments.
  */
-callArgumentList: LParen ((expression (Comma expression)*)? | LBrace (namedArgument (Comma namedArgument)*)? RBrace) RParen;
+callArgumentList: LParen ((expression (Comma expression)* Comma?)? | LBrace (namedArgument (Comma namedArgument)* Comma?)? RBrace) RParen;
 /**
  * Qualified name.
  */
@@ -120,7 +120,7 @@ visibility: Internal | External | Private | Public;
 /**
  * A list of parameters, such as function arguments or return values.
  */
-parameterList: parameters+=parameterDeclaration (Comma parameters+=parameterDeclaration)*;
+parameterList: parameters+=parameterDeclaration (Comma parameters+=parameterDeclaration)* Comma?;
 //@doc:inline
 parameterDeclaration: type=typeName location=dataLocation? name=identifier?;
 /**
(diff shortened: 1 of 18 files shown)


leonardoalt:
Afaik `(a,)` is already considered a single element tuple (I think)

shekhirin:
> Afaik `(a,)` is already considered a single element tuple (I think)

This snippet doesn't compile with `Tuple component cannot be empty` error:
```solidity
contract Test {
	function test() public pure returns (uint256 a) {
		(a,) = (uint(1), );
	}
}
```

And this compiles fine:
```solidity
contract Test {
	function test() public pure returns (uint256 a) {
		(a,) = (uint(1), 1);
	}
}
```

leonardoalt:
Hi @shekhirin sorry for the delay on this.
We've just agreed in a call that it makes sense to add trailing commas to the language.
Your 7 points in the first comment make sense.

The main tricky part right now is tuples. A single-element tuple in Solidity already has a trailing comma: `(a,)` and can lead to ambiguity then. Rust solves that by requiring `_` or `..` for omitted tokens. I don't have a solution for that yet.

shekhirin:
@leonardoalt great, nice to hear!

Generally I like this idea https://github.com/ethereum/solidity/issues/2991#issuecomment-1175086265 for solving the tuples problem, but it does seem like a breaking change so imo it should definitely be released in 0.9.0.

I feel like we can already proceed with the current cases covered by my PR which include everything besides tuples, and add support for them in consequent PRs.

nishant-sachdeva (commented):
LGTM.

As for the ambiguity that @leonardoalt  mentions, https://github.com/ethereum/solidity/issues/2991#issuecomment-1175086265 does seem like a viable solution. Although that looks like a breaking change, so we may want to weigh the pros and cons there.

@cameel  Should we consider doing that in this PR only ? Or would that be too much for the scope of this PR and can perhaps be made into another PR ?

shekhirin:
@cameel pinging on this just in case you've missed!

cameel:
I think we definitely need to deal with that problem somehow. We can't just leave a hole in the language. Maybe we should start from the other side - let's figure out what breaking solution would be best and aim for that in 0.9.0. Then let's decide if we can do something sensible but non-breaking that would be a good enough lead-up to that.

cameel:
@shekhirin @leo So what would be your proposal for the ideal (but breaking solution) to the tuple problem?

NunoFilipeSantos:
Changing it back to draft as the PR is nowhere near ready.

ekpyron:
> Changing it back to draft as the PR is nowhere near ready.

I'll actually close it for now. We won't be doing cosmetics like this before our planned changes to the type system (including namable tuple types that will be affected by this). We can revisit the underlying issue afterwards.

PR #13256, opened 2022-07-08
Added build flag to disable pedantic builds

### LINKED ISSUE #13146: CMake option to disable pedantic builds.
We add a lot of compiler flags that make the builds as strict and pedantic as possible.

While that's very good for our development and CI, this can be annoying for end users and for dependent projects.
We should add a CMake option that disables all warning flags, treating warnings as errors and all other pedantic build flags in ``cmake/EthCompilerSettings.cmake``.

The flag should be off by default (i.e. the build should be strict and pedantic by default) and should maybe even explicitly emit a "use at your own risk" warning when switched on, but we should still have it.

YuviTz1:
Added build flag `DISABLE_WARNINGS` which is OFF by default and disables warnings.

Potentially closes #13146

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ endif()
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
 option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
+option(PEDANTIC "Enable extra warnings and pedantic build flags. Treat all warnings as errors." ON)
 
 # Setup cccache.
 include(EthCcache)
@@ -48,6 +49,9 @@ include_directories(SYSTEM ${JSONCPP_INCLUDE_DIR})
 
 find_package(Threads)
 
+if(NOT PEDANTIC)
+  message(WARNING "-- Pedantic build flags turned off. Warnings will not make compilation fail. This is NOT recommended in development builds.")
+endif()
 # Figure out what compiler and system are we using
 include(EthCompilerSettings)
 
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -23,7 +23,9 @@ if(NOT EMSCRIPTEN)
 	endif()
 endif()
 
-eth_add_cxx_compiler_flag_if_supported(-Wimplicit-fallthrough)
+if(PEDANTIC)
+	eth_add_cxx_compiler_flag_if_supported(-Wimplicit-fallthrough)
+endif()
 
 # Prevent the path of the source directory from ending up in the binary via __FILE__ macros.
 eth_add_cxx_compiler_flag_if_supported("-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=/solidity")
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ endif()
 option(SOLC_LINK_STATIC "Link solc executable statically on supported platforms" OFF)
 option(SOLC_STATIC_STDLIBS "Link solc against static versions of libgcc and libstdc++ on supported platforms" OFF)
 option(STRICT_Z3_VERSION "Use the latest version of Z3" ON)
+option(DISABLE_WARNINGS "Disable all warning flags, treating warnings as errors and all other pedantic build flags" OFF)
```
cameel: Technically this does not disable all warnings, just does not enable the extra warnings we normally want. What do you think about calling it `PEDANTIC` instead?

```suggestion
option(PEDANTIC "Enable extra warnings and pedantic build flags. Treat all warnings as errors." ON)
```

cmake/EthCompilerSettings.cmake
```diff
@@ -32,39 +34,49 @@ eth_add_cxx_compiler_flag_if_supported("-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=/
...
+		add_compile_options(-Wconversion)
+	endif()
 
 	check_cxx_compiler_flag(-Wextra-semi WEXTRA_SEMI)
 	if(WEXTRA_SEMI)
-		add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)
+		if(NOT DISABLE_WARNINGS)
+			add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>)
+		endif()
+	endif()
```
cameel: We don't really need the `WEXTRA_SEMI` flag when warnings are disabled. This the only place where we use it.

You can just put the new `if` around the whole part starting with `Enables all the warnings...` comment and ending with `-Wno-unknown-attributes)`.

cmake/EthCompilerSettings.cmake
```diff
@@ -32,39 +34,49 @@ eth_add_cxx_compiler_flag_if_supported("-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=/
...
-	eth_add_cxx_compiler_flag_if_supported(-Wfinal-dtor-non-final-class)
-	eth_add_cxx_compiler_flag_if_supported(-Wnewline-eof)
-	eth_add_cxx_compiler_flag_if_supported(-Wsuggest-destructor-override)
-	eth_add_cxx_compiler_flag_if_supported(-Wduplicated-cond)
-	eth_add_cxx_compiler_flag_if_supported(-Wduplicate-enum)
-	eth_add_cxx_compiler_flag_if_supported(-Wlogical-op)
-	eth_add_cxx_compiler_flag_if_supported(-Wno-unknown-attributes)
 
 	# Configuration-specific compiler settings.
 	set(CMAKE_CXX_FLAGS_DEBUG          "-O0 -g3 -DETH_DEBUG")
```
cameel, line 78: Can't comment on the line directly since it's too far below but for the PR to be complete, I think it should be disabling treating warnings as errors on MSVC too (the `/WX` flag).

CMakeLists.txt
cameel: I think we should add a note about this flag to the docs page about [building from source](https://docs.soliditylang.org/en/latest/installing-solidity.html#building-from-source).

cameel (dismissed):
Thanks for the PR!

There are some things I'd change but overall this is going in a good direction.

CMakeLists.txt
cameel: You have some trailing whitespace here:
```suggestion
option(DISABLE_WARNINGS "Disable all warning flags, treating warnings as errors and all other pedantic build flags" OFF)
```

That's why the `chk_coding_style` CI job is failing.

CMakeLists.txt
```diff
@@ -48,6 +49,9 @@ include_directories(SYSTEM ${JSONCPP_INCLUDE_DIR})
 
 find_package(Threads)
 
+if(DISABLE_WARNINGS)
+  message("-- Pedantic build flags turned off")
```
```suggestion
  message("-- Pedantic build flags turned off. Warnings will not make compilation fail. This is NOT recommended in development builds.")
```

YuviTz1:
Thank you for the detailed review!
I'll do the necessary changes.

CMakeLists.txt
```diff
@@ -48,6 +49,9 @@ include_directories(SYSTEM ${JSONCPP_INCLUDE_DIR})
 
 find_package(Threads)
 
+if(NOT PEDANTIC)
+  message("-- Pedantic build flags turned off. Warnings will not make compilation fail. This is NOT recommended in development builds.")
```
christianparpart: > "This is NOT recommended in development builds."

Shouldn't that NOT be recommended at all? It should probably say instead that this option should be only enabled when you know what you are doing - as in - "use with care!". What do you think?

cameel: IMO it really only matters for development. These are warnings, not errors and on top of that only those that don't happen in most configurations (or we'd catch them in CI)  so it's unlikely that they're harmful. The user building a release cannot do anything about them anyway.

Maybe we could just say that we recommend reporting warnings as issues in the repo?

cameel:
The PR looks fine to me in the current form but the last thing that is missing is the extra bit of docs (https://github.com/ethereum/solidity/pull/13256#discussion_r923560310). Other than that, we could merge it right away.

CMakeLists.txt
ekpyron: Recommending to report an issue to us sounds good. Also unconditionally stating "This is NOT recommended." is ok with me, but also fine as is.
But we should at least make it a cmake warning, i.e. ``message(WARNING "...``.

cameel: Changed to a warning.

cameel:
I added a bit of docs about the new flag. The PR is complete now.

leonardoalt (approved):

PR #13257, opened 2022-07-09
<address payable>.send warning correction

joejordan:


chriseth:
It fails if either run out of gas, but here, we want to stress the possibility that the recipient runs out of gas.

PR #13263, opened 2022-07-11
lsp: Fixes lsp.py to not bail out when no response is expected.

christianparpart:
This is required for another PR to support sending notifications as part of the I/O being documented as comments.

### DIFF
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -237,8 +237,10 @@ class JsonRpcProcess:
         self.process.stdin.write(rpc_message.encode("utf-8"))
         self.process.stdin.flush()
 
-    def call_method(self, method_name: str, params: Optional[dict]) -> Any:
+    def call_method(self, method_name: str, params: Optional[dict], expects_response: bool = True) -> Any:
         self.send_message(method_name, params)
+        if not expects_response:
+            return None
         return self.receive_message()
 
     def send_notification(self, name: str, params: Optional[dict] = None) -> None:
@@ -484,6 +486,8 @@ class TestParser:
             if self.at_end():
                 raise TestParserException(ret, "Request body not found")
 
+        if self.at_end():
+            return self.RequestAndResponse(**ret)
 
         # Parse response header
         if self.current_line().startswith(RESPONSE_START):
@@ -492,8 +496,6 @@ class TestParser:
                 raise TestParserException(ret, "Response header malformed")
             ret["response"] = self.current_line()[len(RESPONSE_START):] + "\n"
             ret["responseBegin"] = self.position()
-        else:
-            raise TestParserException(ret, "Response header not found")
 
         self.next_line()
 
@@ -568,7 +570,9 @@ class FileTestRunner:
 
             # Process diagnostics first
             self.expected_diagnostics = next(self.parsed_testcases)
-            assert isinstance(self.expected_diagnostics, TestParser.Diagnostics) is True
+            assert isinstance(self.expected_diagnostics, TestParser.Diagnostics)
+            if not self.expected_diagnostics.has_header:
+                return
 
             expected_diagnostics_per_file = self.expected_diagnostics.tests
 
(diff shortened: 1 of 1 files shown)


Marenz (changes requested):
LGTM, but `test/lsp.py:730:0: C0301: Line too long (132/130) (line-too-long)`

test/lsp.py
```diff
@@ -1445,8 +1452,11 @@ def test_generic(self, solc: JsonRpcProcess) -> None:
 
         for sub_dir in map(lambda filepath: filepath.name, sub_dirs):
             tests = map(
-                lambda filename, sd=sub_dir: sd + "/" + filename[:-len(".sol")],
-                os.listdir(f"{self.project_root_dir}/{sub_dir}")
+                lambda file_object, sd=sub_dir: sd + "/" + file_object.name[:-len(".sol")],
```
cameel, line 1459: Is the `sd` argument ever used? As far as I can tell from [the docs for `map()`](https://docs.python.org/3/library/functions.html#map), the function might get more than one arg only when `map()` gets more than one iterable and this does not seem to be the case here.

```suggestion
                lambda file_object: sub_dir + "/" + file_object.name[:-len(".sol")],
```

test/lsp.py
```diff
@@ -1445,8 +1452,11 @@ def test_generic(self, solc: JsonRpcProcess) -> None:
...
+                filter(
+                    lambda filepath: filepath.is_file() and filepath.name.endswith('.sol'),
+                    os.scandir(f"{self.project_root_dir}/{sub_dir}")
+                )
             )
```
cameel, line 1464: It seems to me like it would be a bit more straightforward as a list comprehension:
```python
tests = [
    sub_dir + "/" + dir_entry.name[:-len(".sol")]
    for dir_entry in os.scandir(f"{self.project_root_dir}/{sub_dir}")
    if dir_entry.is_file() and dir_entry.name.endswith('.sol')
]
```

I think that `map()` and `filter()` overall are falling out of use in modern Python because most of the time they can be replaced with list comprehensions.

test/lsp.py
```diff
@@ -569,6 +571,8 @@ def test_diagnostics(self):
             # Process diagnostics first
             self.expected_diagnostics = next(self.parsed_testcases)
             assert isinstance(self.expected_diagnostics, TestParser.Diagnostics) is True
```
```suggestion
            assert isinstance(self.expected_diagnostics, TestParser.Diagnostics)
```

cameel (dismissed):
Looks fine to me (apart from the failing test, but that change was already requested by @Marenz).

I just have some optional simplification suggestions.

test/lsp.py
Marenz: Your suggestion won't work because `sub_dir` at the time of the lambda execution might have a completely different value, so we use a capture here to make the right `sub_dir` value available. See also https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

test/lsp.py
Marenz: I never really warmed up to them because the syntax is so unintuitive :sweat_smile:

christianparpart: this code example isn't necessarily more readable, IMHO. To be fair, using filter()/map() isn't either. I was simply fixing a problem here. :-)

test/lsp.py
christianparpart: That line isn't part of the PR, but I generally agree for future touches :)

test/lsp.py
cameel: Ah, ok. That's something that at least warrants a comment though. It's not obvious.

test/lsp.py
cameel: I find it more readable because it's just a `for` loop with an `if` in it. It's just plain code without magic.

`map()` and `filter()` solution has more nesting, has multiple variables meaning the same thing (`filepath` vs `filename`), requires lambdas and comes with gotchas (like https://github.com/ethereum/solidity/pull/13263#discussion_r918389531).

cameel: But yeah. I know it's not the point of the PR. I approved it anyway. Just saying that this part could be simplified.

cameel: > I never really warmed up to them because the syntax is so unintuitive

I think it's a matter of getting used to them. Honestly, for me the `map()`/`filter()` version is something I have to decode step by step while the one with comprehensions seems just straightforward :)

cameel (approved):

Marenz (approved):

PR #13265, opened 2022-07-11
Add `--no-append-metadata` in CLI and `metadata.append` in JSON

### LINKED ISSUE #13233: Option to disable metadata
Currently the compiler will always include the metadata at the end of runtime + initcode. The metadata consists of version, hash, and potentially the experimental flag.

In the pipeline the only flag we have is to disable the inclusion of the hash, but the version field is always included.

While I really like having this there, removing it from the runtime code becomes quite challenging, because one needs to dissect the initcode. Why would someone remove the metadata? To pass code purity checks.

hrkrshnn:
Skips appending metadata to the binary

Closes https://github.com/ethereum/solidity/issues/13233


solc/CommandLineParser.cpp
```diff
@@ -739,6 +741,10 @@ General Information)").c_str(),
 
 	po::options_description metadataOptions("Metadata Options");
 	metadataOptions.add_options()
+		(
+			g_strNoAppendMetadata.c_str(),
+			"Do not append metadata to the end of the bytecode"
```
leonardoalt:
```suggestion
			"Do not append metadata to the end of the bytecode."
```

solc/CommandLineParser.cpp
```diff
@@ -1199,6 +1206,15 @@ void CommandLineParser::processArgs()
 			"--" + g_strYulDialect + " and --" + g_strMachine + " are only valid in assembly mode."
 		);
 
+	if (m_args.count(g_strNoAppendMetadata))
+	{
+		if (m_args.count(g_strMetadataHash))
+		{
+			// TODO Should we allow setting both of them at the same time?
```
leonardoalt: Probably not.

leonardoalt:
Missing tests.

test/cmdlineTests/no_append_metadata/output
```diff
@@ -0,0 +1,4 @@
+
+======= no_append_metadata/input.sol:C =======
+Binary:
+608080604052346013576035908160198239f35b600080fdfe6004361015600c57600080fd5b6000803560e01c6326121ff014602157600080fd5b346032578060031936011260325733ff5b80fd
```
hrkrshnn: Probably not the best way to test this as this could change in PRs that change the codegen some way. Hopefully, the contract is simple enough that this is not so likely in most PRs. The other option is to have like a boost cpp test, and check for the last byte or something similar.

test/cmdlineTests/metadata_append_standard_json/input.json
```diff
@@ -0,0 +1,33 @@
...
+	{
+		"metadata":
+		{
+			"append": false
+		},
+		"outputSelection":
+		{
+			"A":
+			{
+				"test": ["evm.bytecode", "metadata"]
```
hrkrshnn: TODO: need to remove "metadata" here. Otherwise, tests would likely be flaky.

libsolidity/interface/CompilerStack.h
```diff
@@ -349,6 +349,11 @@ class CompilerStack: public langutil::CharStreamProvider
 	/// caused by differences in metadata. Should only be used for testing.
```
matheusaaguiar:
```suggestion

```

The comment here (starting in the line before) probably should be removed now that the method is used [here](https://github.com/ethereum/solidity/blob/75999e4f4c5171648bea880da7d2250bea338885/libsolidity/interface/StandardCompiler.cpp#L1093) and [here](https://github.com/ethereum/solidity/blob/75999e4f4c5171648bea880da7d2250bea338885/solc/CommandLineInterface.cpp#L692).

test/cmdlineTests/metadata_append_standard_json/input.json
matheusaaguiar: While testing, I noticed that `compiler version` will change in `output` . I guess that will happen for every new commit/release. 

Expected output :
```
"metadata":"{\"compiler\":{\"version\":\"0.8.16-develop.2022.7.24+commit.b6a06b65.mod\"}
```
Obtained:
```
"metadata":"{\"compiler\":{\"version\":\"0.8.16-develop.2022.7.26+commit.75999e4f\"}
```
Maybe in `cmdLineTests.sh` we could fetch the right value and change before the tests are executed ?

libsolidity/interface/CompilerStack.h
hrkrshnn: Good catch. Going to change now.

test/cmdlineTests/metadata_append_standard_json/input.json
```diff
@@ -0,0 +1,33 @@
...
+						assert(s.x > 0);
+					}
+				}"
+		}
+	},
+	"settings":
+	{
+		"metadata":
+		{
+			"append": false
```
hrkrshnn: This is not really a good test because the testing framework is removing `object":"<BYTECODE REMOVED>"`. But it actually removed the bytecode at the end if you test it locally `./solc --standard-json ../../test/cmdlineTests/metadata_append_standard_json/input.json`.

```
 ./solc --standard-json ../../test/cmdlineTests/metadata_append_standard_json/input.json                                 no-append-metadata ✱ ◼
{"contracts":{"A":{"test":{"evm":{"bytecode":{"functionDebugData":{},"generatedSources":[],"linkReferences":{},"object":"608060405234801561001057600080fd5b50610120806100206000396000f3fe608060405234801561001057600080fd5b506004361061002b5760003560e01c806398c3a6c114610030575b600080fd5b61004a600480360381019061004591906100c4565b61004c565b005b8061005857600261005b565b60015b60ff1660008001600082825417925050819055506000806000015411610084576100836100f1565b5b50565b600080fd5b60008115159050919050565b6100a18161008c565b81146100ac57600080fd5b50565b6000813590506100be81610098565b92915050565b6000602082840312156100da576100d9610087565b5b60006100e8848285016100af565b91505092915050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052600160045260246000fd"
....
```
let me know if there is a better way to write a standard json test for this.

docs/using-the-compiler.rst
```diff
@@ -323,6 +323,8 @@ Input Description
         },
         // Metadata settings (optional)
         "metadata": {
+          // The metadata is appended at the end of the bytecode by default
```
axic:
```suggestion
          // The metadata is appended at the end of the bytecode by default.
          // Setting this to false omits the metadata from both runtime and deploy time code.
```

docs/using-the-compiler.rst
```diff
@@ -323,6 +323,8 @@ Input Description
...
+          "append": true,
```
axic: The `optimizer` section uses `enabled`. I also wonder how could these options reduced to a fewer, like an enum.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1548,6 +1549,9 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
 	if (m_revertStrings != RevertStrings::Default)
 		meta["settings"]["debug"]["revertStrings"] = revertStringsToString(m_revertStrings);
 
+	meta["settings"]["metadata"]["append"] =
```
axic: Hm?

axic: There's `MetadataFormat::NoMetadata` below, so perhaps this here would be way more readable as `m_metadataFormat != MetadataFormat::NoMetadata`.

hrkrshnn: Fixed.

test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol
```diff
@@ -39,9 +39,9 @@ contract test {
 
 // ----
 // constructor(), 20 wei ->
-// gas irOptimized: 262130
-// gas legacy: 402654
-// gas legacyOptimized: 274470
+// gas irOptimized: 303208
+// gas legacy: 443715
+// gas legacyOptimized: 315548
```
hrkrshnn: It's not clear to my why the gas increased for these two tests. They are both deploying a new contract inside a constructor. My initial theory was that in such cases in the past, metadata wasn't appended. But I tried to deploy contracts this way and that is not the case :(

test/cmdlineTests/metadata_append_standard_json/input.json
hrkrshnn: Fixed this by removing the field `metadata`.

docs/using-the-compiler.rst
hrkrshnn: Fixed this to "Setting this to false omits the metadata from the deployed code." which feels more accurate.

test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol
ekpyron: Hm... the semantics tests *should* have removed the metadata in general in order not to get flaky gas costs due to zeroes or non-zeroes and such...
```
	if (m_enforceGasCost)
	{
		m_compiler.setMetadataFormat(CompilerStack::MetadataFormat::NoMetadata);
		m_compiler.setMetadataHash(CompilerStack::MetadataHash::None);
	}
```
in ``test/libsolidity/SmenaticTest.cpp``....

Ah, I see now... ``SolidityExecutionFramework::multiSourceCompileContract`` later on calls ``m_compiler.reset()``, which earlier did *not* reset the metadata format, but now it does. That's the reason.

ekpyron: i.e. quick fix probably is to add a ``m_metadataFormat`` to ``SolidityExecutionFramework`` and set that instead of invoking ``m_compiler.setMetadataFormat`` in ``SemanticTest.cpp``...
``m_compiler.setMetadataHash`` in ``SemanticTest.cpp`` was silly anyways, since it just got overwritten :-).

At least if I saw correctly just now, only had a quick glance :-).

docs/using-the-compiler.rst
axic: It should remove it from both, not just deployed code.

libsolidity/interface/CompilerStack.cpp
axic: Also keeping in line with `m_metadataLiteralSources`, these fields are not present unless set to non-default values. Not sure whether that is a good approach or not, but mentioning it.

test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol
hrkrshnn: Thanks, that actually worked!

Changelog.md
```diff
@@ -9,6 +9,8 @@ Compiler Features:
  * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
  * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
  * Code Generator: More efficient code for checked addition and subtraction.
+ * Commandline Interface: Add `--no-append-metadata` that skips metadata from getting appended at the end of the deployed bytecode.
+ * Standard JSON: Add a boolean field `append` in the `metadata` settings that skips metadata from getting appended at the end of the deployed bytecode.
```
cameel: I think it's best to refer to these settings by their full "path":
```suggestion
 * Standard JSON: Add a boolean ``settings.metadata.append`` field that skips metadata from getting appended at the end of the deployed bytecode.
```

Changelog.md
```diff
@@ -9,6 +9,8 @@ Compiler Features:
  * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
  * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
  * Code Generator: More efficient code for checked addition and subtraction.
+ * Commandline Interface: Add `--no-append-metadata` that skips metadata from getting appended at the end of the deployed bytecode.
```
cameel: I think we should mention the new CLI option in https://docs.soliditylang.org/en/latest/metadata.html.

solc/CommandLineParser.h
```diff
@@ -215,6 +215,7 @@ struct CommandLineOptions
 
 	struct
 	{
+		CompilerStack::MetadataFormat format = CompilerStack::defaultMetadataFormat();
```
cameel, line 218: Can you add this field to tests in [`test/solc/CommandLineParser.cpp`](https://github.com/ethereum/solidity/blob/develop/test/solc/CommandLineParser.cpp)?

solc/CommandLineParser.cpp
```diff
@@ -1212,6 +1219,20 @@ void CommandLineParser::processArgs()
...
 
+	if (m_args.count(g_strNoAppendMetadata))
+	{
+		if (
+			m_args.count(g_strMetadataHash) &&
+			m_options.metadata.hash != CompilerStack::MetadataHash::None
+		)
+			solThrow(
+				CommandLineValidationError,
+				"Cannot specify a metadata-hash when --" + g_strNoAppendMetadata + " is set."
```
```suggestion
				"Cannot specify a metadata hashing method when --" + g_strNoAppendMetadata + " is set."
```

solc/CommandLineParser.cpp
```diff
@@ -739,6 +741,10 @@ General Information)").c_str(),
 
 	po::options_description metadataOptions("Metadata Options");
 	metadataOptions.add_options()
+		(
+			g_strNoAppendMetadata.c_str(),
+			"Do not append metadata to the end of the bytecode."
```
cameel: "Metadata" is somewhat ambiguous in this context. We have metadata at the end of the binary and then metadata stored in JSON. How about we invent a name for the binary one to be able to refer to it more precisely? For example what do you think about something like "metadata anchor". Option naming would also be easier - we could then call it `--no-metadata-anchor`.

```suggestion
			"Do not append metadata anchor at the end of the bytecode. The metadata anchor is a small bit of CBOR data that identifiers the compiler version and may include a hash of the full metadata."
```

If we go with this, the descriptions of the two options below should also be updated to match.

docs/using-the-compiler.rst
cameel: I like `enabled` more than `append` but it's also a little ambiguous. It sounds as if it would enable/disable metadata generation but just the bit at the end of bytecode. Metadata JSON is controlled separately via output selection. If we go with my suggestion to call that bit a "metadata anchor", we could use `"anchor": true` instead.

As for the other things, they seem like independent bits of information to me. You can choose literal sources regardless of whether you have the anchor or not. And though the hash does not make sense without the anchor, it's not just another state of it. You could have other bits of info that also depend on the anchor and they would not fit into an enum (e.g. how the version is stored).

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -908,6 +910,11 @@ std::variant<StandardCompiler::InputsAndSettings, Json::Value> StandardCompiler:
 	if (auto result = checkMetadataKeys(metadataSettings))
 		return *result;
 
+	ret.metadataFormat =
+		metadataSettings.get("append", Json::Value(true)).asBool() ?
+		CompilerStack::defaultMetadataFormat() :
+		CompilerStack::MetadataFormat::NoMetadata;
```
cameel: This will be broken if we ever change the default to `CompilerStack::MetadataFormat::NoMetadata`. Can you add an assert that this is not the case?

test/cmdlineTests/metadata_append_standard_json/output.json
```diff
@@ -0,0 +1 @@
+{"contracts":{"A":{"test":{"evm":{"bytecode":{"functionDebugData":{},"generatedSources":[],"linkReferences":{},"object":"<BYTECODE REMOVED>","opcodes":"<OPCODES REMOVED>","sourceMap":"<SOURCEMAP REMOVED>"}}}}},"sources":{"A":{"id":0}}}
```
cameel: Please add an `args` file with `--json-indent 4` to get this nicely formatted.

test/cmdlineTests/metadata_append_standard_json/input.json
```diff
@@ -0,0 +1,33 @@
...
+			"content": "// SPDX-License-Identifier: GPL-3.0\npragma solidity >=0.0;\n\ncontract test {
+					struct S {
+						uint x;
+					}
+					S s;
+					function f(bool b) public {
+						s.x |= b ? 1 : 2;
+						assert(s.x > 0);
+					}
+				}"
```
cameel: You can put this in a standalone solidity file and just put the path here. See for example https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/standard_debug_info_in_yul_and_evm_asm_print_all

But really, for the purpose of this test I'd just use something like `contract C {}` for simplicity because the actual content of the contract is not relevant.

test/cmdlineTests/no_append_incorrect_combination/input.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity >= 0.0.0;
+
+contract C {
+    function f() external {
+        selfdestruct(payable(msg.sender));
+    }
+}
```
cameel: Same here. Could be just `contract C {}` for simplicity.

test/cmdlineTests/no_append_metadata/input.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity >= 0.0.0;
```
cameel, line 2: To avoid the warning:
```suggestion
pragma solidity >= 0.0.0;
// SPDX-License-Identifier: GPL-3.0
```

test/cmdlineTests/metadata_append_standard_json/input.json
cameel: One way to test this properly would be to adjust the code that replaces the bytecode in https://github.com/ethereum/solidity/blob/develop/test/cmdlineTests.sh#L204-L238. It actually detects the metadata markers for the CLI tests. Not for JSON but it also could. Then, instead of replacing both bytecode and the metadata with a single `<BYTECODE REMOVED>`, we could have it replace just the metadata first with `<METADATA REMOVED>`, then detect bytecode based on the presence of that and only then insert `<BYTECODE REMOVED>`, this time not covering the metadata.

It would require a bit of fiddling with regexes and might be even more brittle than the current one. If anything, I'd do that in a separate PR.

But I'd accept this PR even without such a test. Just add some comments in the input file explaining why the metadata does not show up in the output.

test/cmdlineTests/no_append_metadata/output
cameel: Yeah, this is tricky because the bytecode detection works by finding the metadata :) But the upside is that the fact that it does not get replaced means that it works. I'd just add a comment in the input file that this binary blob is actually the expected result here and `<BYTECODE REMOVED>` means failure.

As for codegen changes, I think it's not a problem. We already have a lot of that churn in tests that produce Yul and we live with it. The biggest reason for removing the bytecode was actually the metadata because that would change not just with codegen changes but also with the environment (e.g. force release builds in CI have different compiler version format in the metadata). If it's gone, the problem is gone :)

test/cmdlineTests/metadata_append_standard_json/input.json
cameel: Right. `metadata` output is irrelevant to this test so think removing it is the right call.

docs/using-the-compiler.rst
cameel: It also removes it from the nested bytecode of any contract created with `new`. Not sure if that's worth mentioning here or not but that's the detail people often forget.

test/cmdlineTests/no_append_incorrect_combination/input.sol
cameel: Actually, if you want to have something here, I'd include contract creation with `new`. Or code that references `.creationBytecode` or `.runtimeBytecode`. That would show whether the metadata for nested contracts disappears as well (if it doesn't, the test script will detect it and replace bytecode with `<BYTECODE REMOVED>`).

test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol
cameel: Now tests seem to be failing with `Invalid or missing metadata in bytecode.` :)

test/cmdlineTests/no_append_metadata/output
ekpyron: Why not use ``--bin-runtime`` on an empty contract? That'd just be ``600080fd`` and probably never change...

cameel: Right. That's better better than what is there now but still not a great test overall - simply because if it changes when you update tests it might still be hard to realize that it changed the wrong way. On the other hand if you had something like `<BYTECODE REMOVED><METADATA REMOVED>` and the `<METADATA REMOVED>` part disappeared, it would be very clear.

test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol
hrkrshnn: Yeah, have to fix the new issue :(

test/cmdlineTests/no_append_incorrect_combination/input.sol
hrkrshnn: ~~I used selfdestruct so the opcode (`0xff`) can be a visible marker. But good point on nested contracts. I'll try to make it that way.~~ Ah, this one doesn't even compile in the end. Will change to a simpler contract.

test/cmdlineTests/metadata_append_standard_json/input.json
hrkrshnn: Actually copied it from an SMT checker test :), but can change to something simpler.

test/cmdlineTests/no_append_metadata/output
hrkrshnn: Going to use an empty contract, hopefully would be less flaky. Will probably have to use `--via-ir` so that there is none of those free memory initializations :)

test/cmdlineTests/metadata_append_standard_json/input.json
hrkrshnn: Funnily enough, if I add the extra args `--pretty-json --json-indent 4`, the bytecode ends up showing. Which solves the issue, actually.

test/cmdlineTests/metadata_append_standard_json/output.json
hrkrshnn: Nice, wasn't aware of the flag.

Changelog.md
hrkrshnn: Done

libsolidity/interface/StandardCompiler.cpp
hrkrshnn: Added an assert right above!

solc/CommandLineParser.cpp
hrkrshnn: I'm not sure about anchor here. It's a new terminology :)

cameel: I mean, would be great to have *any* clear term for this. Calling both metadata is not helpful.

We just had a discussion with @ekpyron today about how ambigous dynamically-sized and statically-sized really are...

docs/metadata.rst
```diff
@@ -163,6 +163,10 @@ Whereas release builds of solc use a 3 byte encoding of the version as shown
 above (one byte each for major, minor and patch version number), prerelease builds
 will instead use a complete version string including commit hash and build date.
 
+The commandline flag ``--no-append-metadata`` can be used to skip metadata
+from getting appended at the end of the deployed bytecode. Equivalently, the
+boolean field ``settings.metadata.append`` can be set to false.
```
cameel:
```suggestion
boolean field ``settings.metadata.append`` in Standard JSON input can be set to false.
```

test/cmdlineTests/metadata_append_standard_json/input.json
cameel: I think it's because the regex assumes no space after `:`. It's fine in this case since there's no metadata but we'll have to fix that because in other cases metadata is problematic. So in the long term I would not count on it staying this way and a comment would still be useful.

Or I guess, we could just start using `append: false` in tests :)

test/solc/CommandLineParser.cpp
```diff
@@ -224,6 +224,15 @@ BOOST_AUTO_TEST_CASE(cli_mode_options)
 	}
 }
 
+BOOST_AUTO_TEST_CASE(no_append_metadata)
+{
+	vector<string> commandLine = {"solc", "--no-append-metadata", "contract.sol"};
+	CommandLineOptions parsedOptions = parseCommandLine(commandLine);
+	bool assert = parsedOptions.metadata.format == CompilerStack::MetadataFormat::NoMetadata;
+
+	BOOST_TEST(assert);
```
cameel, line 233:
```suggestion
	BOOST_TEST(parsedOptions.metadata.format == CompilerStack::MetadataFormat::NoMetadata);
```

test/solc/CommandLineParser.cpp
```diff
@@ -224,6 +224,15 @@ BOOST_AUTO_TEST_CASE(cli_mode_options)
...
+}
```
cameel: This test is fine but what I meant in https://github.com/ethereum/solidity/pull/13265#discussion_r938088556 was that this file has tests for every input mode, listing all available options. And also a (still incomplete) test for options that are not allowed. Could you add the option to test for the compiler input mode?

test/cmdlineTests/metadata_append_standard_json/input.json
cameel: Yeah, most tests don't do that but I think we should be doing it more. I always do this in mine. When the snippet is nested formatting it is awkward and highlighting also breaks.

cameel:
I see that tests are now fixed (apart from zeppelin - is this just not rebased or did it break again?). So apart from trivial things the big question for me here is if we could have a better way to refer to this metadata blob than just "metadata", which is ambiguous.

test/solc/CommandLineParser.cpp
hrkrshnn: Surprisingly, this change leads to 10 pages of C++ errors. Some weird macro stuff going on here.

cameel: Ah, ok then. I think it's because the macro does some magic that expects the type to be printable with `<<`. There are several ways around that. One, I think, was to put it in an extra parenthesis. You could also declare the type as non-printable with some macro. Or define the operator. But overall probably not worth the effort just for this one line. Let's leave it as is.

test/solc/CommandLineParser.cpp
hrkrshnn: Just rebased and fixed the comment on `metadata.rst`. This is what's left now. Will try.

hrkrshnn: @cameel Is something like adding a `for (InputMode inputMode: {InputMode::Compiler, InputMode::CompilerWithASTImport})` etc what you are looking?

Maybe failures for `InputMode::Assembler` and `Linker`?

cameel: > Is something like adding a `for (InputMode inputMode: {InputMode::Compiler, InputMode::CompilerWithASTImport})` etc what you are looking?

Kinda, but not in a new test. Rather just add it in the test called `cli_mode_options`.

> Maybe failures for `InputMode::Assembler` and `Linker`?

And for that we have `invalid_options_input_modes_combinations`.

hrkrshnn: > Rather just add it in cli_mode_options.

This is a little bit annoying because you can't have `--no-cbor-metadata` and `--metadata-hash=swarm` together. Which breaks the stuff.

test/libsolidity/SolidityExecutionFramework.h
```diff
@@ -80,6 +86,7 @@ class SolidityExecutionFramework: public solidity::test::ExecutionFramework
 	bool m_compileViaYul = false;
 	bool m_compileToEwasm = false;
 	bool m_showMetadata = false;
+	bool m_metadataAppend = true;
```
hrkrshnn: Keeping this as it is :)

Changelog.md
```diff
@@ -9,6 +9,8 @@ Compiler Features:
  * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
  * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
  * Code Generator: More efficient code for checked addition and subtraction.
+ * Commandline Interface: Add `--no-cbor-metadata` that skips metadata from getting appended at the end of the deployed bytecode.
+ * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips metadata from getting appended at the end of the deployed bytecode.
```
```suggestion
 * Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the deployed bytecode.
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the deployed bytecode.
```

Should I use this? Feels off.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -503,14 +503,16 @@ std::optional<Json::Value> checkMetadataKeys(Json::Value const& _input)
 {
 	if (_input.isObject())
 	{
+		if (_input.isMember("appendCBOR") && !_input["appendCBOR"].isBool())
+			return formatFatalError("JSONError", "\"settings.metadata.append\" must be Boolean");
```
```suggestion
			return formatFatalError("JSONError", "\"settings.metadata.appendCBOR\" must be Boolean");
```
If we want to keep it at ``appendCBOR``... still hoping we can find something better...

hrkrshnn: Nice catch. Fixed.

test/solc/CommandLineParser.cpp
cameel: Ah, ok then. In that case a separate test makes sense.

Changelog.md
cameel: Why? Seems fine for me. @ekpyron what do you think?

ekpyron: Doesn't seem too off to me. ``CBOR-encoded`` would also work.

cameel: Yeah, but the point was that we agreed to call it "CBOR metadata". I.e. it's not just any piece of CBOR-encoded medata. It's a very specific one :)

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1548,6 +1549,9 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
 	if (m_revertStrings != RevertStrings::Default)
 		meta["settings"]["debug"]["revertStrings"] = revertStringsToString(m_revertStrings);
 
+	meta["settings"]["metadata"]["appendCBOR"] =
+		m_metadataFormat != MetadataFormat::NoMetadata;
```
ekpyron: This should really only be there, if it's not set to the default, shouldn't it? That's how all (or at least most of) the others work, isn't it?

docs/using-the-compiler.rst
```diff
@@ -323,6 +323,9 @@ Input Description
         },
         // Metadata settings (optional)
         "metadata": {
+          // The metadata is appended at the end of the bytecode by default.
```
cameel:
```suggestion
          // The CBOR metadata is appended at the end of the bytecode by default.
```

Changelog.md
hrkrshnn: Sure. Fixing now.

libsolidity/interface/CompilerStack.cpp
hrkrshnn: Ok, I'll fix this.

cameel: Yeah, this was pointed out before in one of the comments.

Personally, I ignored that concern because I'm not sure if it's a good thing. Defaults can change. But yeah, to be consistent with that we do currently, this should be omitted when it's set to the default.

solc/CommandLineParser.cpp
```diff
@@ -739,6 +741,10 @@ General Information)").c_str(),
 
 	po::options_description metadataOptions("Metadata Options");
 	metadataOptions.add_options()
+		(
+			g_strNoCBORMetadata.c_str(),
+			"Do not append metadata CBOR to the end of the bytecode."
```
```suggestion
			"Do not append CBOR metadata to the end of the bytecode."
```

test/solc/CommandLineParser.cpp
```diff
@@ -224,6 +224,15 @@ BOOST_AUTO_TEST_CASE(cli_mode_options)
 	}
 }
 
+BOOST_AUTO_TEST_CASE(no_append_metadata)
```
```suggestion
BOOST_AUTO_TEST_CASE(no_cbor_metadata)
```

libsolidity/interface/CompilerStack.cpp
axic: See also https://github.com/ethereum/solidity/pull/13265#discussion_r937774398

solc/CommandLineParser.cpp
hrkrshnn: Fixed.

test/solc/CommandLineParser.cpp
hrkrshnn: Fixed.

libsolidity/interface/CompilerStack.cpp
hrkrshnn: Fixed it.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1548,6 +1549,9 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
 	if (m_revertStrings != RevertStrings::Default)
 		meta["settings"]["debug"]["revertStrings"] = revertStringsToString(m_revertStrings);
 
+	if (m_metadataFormat != MetadataFormat::NoMetadata)
```
```suggestion
	if (m_metadataFormat != defaultMetadataFormat())
```

cameel (dismissed):

libsolidity/interface/CompilerStack.cpp
hrkrshnn: Ah, good catch. Will change to `==MetadataFormat::NoMetadata` since alex prefered it over `default`

Changelog.md
```diff
@@ -9,6 +9,8 @@ Compiler Features:
  * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
  * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
  * Code Generator: More efficient code for checked addition and subtraction.
+ * Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the deployed bytecode.
```
axic: Why was all this renamed to include cbor? :)

docs/using-the-compiler.rst
```diff
@@ -323,6 +323,9 @@ Input Description
         },
         // Metadata settings (optional)
         "metadata": {
+          // The CBOR metadata is appended at the end of the bytecode by default.
+          // Setting this to false omits the metadata from the deployed code.
```
axic: Still not sure why this is restricted to "deployed code", when in fact it isn't.

libsolidity/interface/CompilerStack.cpp
hrkrshnn: Just fixed.

axic: I'd actually oppose this, as it makes it way harder to understand the intent.

I mean oppose using the "default" helper.

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1548,6 +1549,9 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
 	if (m_revertStrings != RevertStrings::Default)
 		meta["settings"]["debug"]["revertStrings"] = revertStringsToString(m_revertStrings);
 
+	if (m_metadataFormat == MetadataFormat::NoMetadata)
+		meta["settings"]["metadata"]["appendCBOR"] = false;
```
hrkrshnn, line 1553: @axic @cameel this is the fix.

Changelog.md
ekpyron: Because ``metadata`` so far always refers to the json metadata file and we don't want to overload nomenclature any further. But we just decided not to rush this into the release after all and take until the end of this week to decide the final naming.

libsolidity/interface/CompilerStack.cpp
cameel: Isn't the intent to skip fields that have default values?

hrkrshnn: Yeah, that's right. But Alex had a preference to using `NoMetadata` over `default`.

docs/using-the-compiler.rst
hrkrshnn: What else?

Changelog.md
ekpyron: Well... "always" with the exception of standard-json's ``settings.metadata.bytecodeHash`` which kind of refers to both.

cameel: Haha, we just had a long discussion on the channel regarding the naming of this thing. We don't have a great name but calling it just "metadata" is ambiguous so the compromise was to at least call it "CBOR metadata" to distinguish it from the the full "JSON metadata".

@ekpyron was in favor of actually holding off this this PR so that we can find a better one. So I guess you'd also be in that camp? :)

docs/using-the-compiler.rst
hrkrshnn: The original suggestion by you was "runtime and deploy time code." Wouldn't deployed code capture both?

libsolidity/interface/CompilerStack.cpp
axic: > Isn't the intent to skip fields that have default values?

i think this is a confusing piece of code, as on the non-default case we include the an unrelated field with an unrelated value (compared to what the "default" says").

docs/using-the-compiler.rst
axic: initcode ("deploy time") and deployed ("runtime") code both contain it.

ekpyron:
Well, we said we wouldn't stall this longer than today due to naming and there wasn't that much more enlightening discussion on it...
https://github.com/ethereum/solidity/pull/13265/files#r940130991 needs to be addressed in any case.

As for naming our thinking so far was that in standard-json ``settings.metadata.bytecodeHash`` is a misnomer anyways, so we want to change that in a breaking release (accepting that it'll cause translation complexity in solc-js) - and in the process of that move the new option as well. Ideally we'd want a clear separation in the settings (and in terminology in general) between those referring to the json metadata file and those referring to the cbor encoded metadata in the bytecode.

leonardoalt:
ping @hrkrshnn

ekpyron:
For the record, this briefly came up in our meeting again today, and @chriseth mentioned that Sourceify calls the cbor metadata in the bytecode "aux data" these days, which is aligned with how we refer to it in Yul objects.

If we go down that road, we'd ideally have to refer to it like that everywhere, though.

Since ``metadata.bytecodeHash`` already exists anyways and doesn't align with that either, we might as well merge this as is, though, and then do a proper renaming of all of it in 0.9.

docs/using-the-compiler.rst
hrkrshnn: Fixed this comment.

hrkrshnn:
Fixed the last pending comment.

leonardoalt:
so does this still need changes before merging?

hrkrshnn:
> so does this still need changes before merging?

@leonardoalt I don't think so. I think I addressed all pending comments.

Changelog.md
```diff
@@ -42,8 +44,9 @@ Compiler Features:
  * Code Generator: More efficient code for checked addition and subtraction.
  * TypeChecker: Support using library constants in initializers of other constants.
  * Yul IR Code Generation: Improved copy routines for arrays with packed storage layout.
- * Yul Optimizer: Add rule to convert ``mod(add(X, Y), A)`` into ``addmod(X, Y, A)``, if ``A`` is a power of two.
- * Yul Optimizer: Add rule to convert ``mod(mul(X, Y), A)`` into ``mulmod(X, Y, A)``, if ``A`` is a power of two.
+ * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
+ * Yul Optimizer: Add rule to convert `mod(add(X, Y), A)` into `addmod(X, Y, A)`, if `A` is a power of two.
+ * Code Generator: More efficient code for checked addition and subtraction.
```
cameel: Looks like this reverts the change from #13357. Please remove from the PR.

Changelog.md
```diff
@@ -4,6 +4,8 @@ Language Features:
 
 
 Compiler Features:
+* Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the deployed bytecode.
+* Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the deployed bytecode.
```
cameel: This is now inconsistent with https://github.com/ethereum/solidity/pull/13265#discussion_r940130991, isn't it?
```suggestion
* Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the bytecode.
* Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
```

test/libsolidity/SolidityExecutionFramework.h
cameel: Why though? Wouldn't it make sense to change it to `m_appendCBORMetadata` here?

docs/using-the-compiler.rst
cameel: What about this? I think it would be a good idea to mention it.

Changelog.md
cameel: @axic Are you ok with the name or do you think we should use "auxdata" or something else?

cameel (commented):
Some small adjustments needed. But other than that I don't see big problems here and the naming issue was addressed. When these are addressed I'm going to approve.

Changelog.md
hrkrshnn: Oops. Don't know how it got in. Bad rebase.

hrkrshnn: Fixed

docs/using-the-compiler.rst
hrkrshnn: I feel that mentioning the nested part is redundant. As an interface, if there is an option to disable metadata, it should disable it everywhere. I can also add a note somewhere in the docs if you want to.

test/libsolidity/SolidityExecutionFramework.h
hrkrshnn: Fixed it.

docs/using-the-compiler.rst
cameel: Well, we could add a note on the page about metadata.

But I guess it's not that important. If you think it's not a good idea to put it here, let's just drop the idea and have the PR merged :)

cameel (dismissed):

cameel:
Looks like it needs a rebase on `develop`. I think we fixed the euler problem already.

hrkrshnn:
Just rebased. Edit: tests are failing now :( Will fix.

Fixed: this is the fix
```diff
-			return formatFatalError("JSONError", "\"settings.metadata.appendCBOR\" must be Boolean");
+			return formatFatalError(Error::Type::JSONError, "\"settings.metadata.appendCBOR\" must be Boolean");
```

cameel (approved):

PR #13270, opened 2022-07-13
add missing word for array length in docs storage layout

CJ42:
If I am understanding correctly storage layout for arrays, there is a missing word in the docs for the storage layout of arrays. See file changes.

### DIFF
--- a/docs/internals/layout_in_storage.rst
+++ b/docs/internals/layout_in_storage.rst
@@ -127,16 +127,20 @@ The type of the value is ``uint256``, so it uses a single slot.
 ``bytes`` and ``string``
 ------------------------
 
-``bytes`` and ``string`` are encoded identically.
-In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array length itself and
-a data area that is computed using a ``keccak256`` hash of that slot's position.
-However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
+``bytes`` and ``string`` are encoded identically. In general, the encoding is similar to ``bytes1[]`` in the
+ sense that the layout in storage depends on the ``.length``:
 
-In particular: if the data is at most ``31`` bytes long, the elements are stored
+- For short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
+
+If the data is at most ``31`` bytes long, the elements are stored
 in the higher-order bytes (left aligned) and the lowest-order byte stores the value ``length * 2``.
-For byte arrays that store data which is ``32`` or more bytes long, the main slot ``p`` stores ``length * 2 + 1`` and the data is
-stored as usual in ``keccak256(p)``. This means that you can distinguish a short array from a long array
-by checking if the lowest bit is set: short (not set) and long (set).
+
+- For long values (more than 32 bytes), the length is stored in one slot (which position is defined as per the [storage layout rules](#layout-of-state-variables-in-storage)), while the array elements are stored in a separately allocated data area whose location is computed using a ``keccak256`` hash of that slot's position.
+
+For byte arrays that store data which is ``32`` bytes long or more, the main slot ``p`` stores ``length * 2 + 1`` and the data is
+stored as usual in ``keccak256(p)``. 
+
+This means that you can distinguish a short array from a long array by checking if the lowest bit is set: short (not set) and long (set).
 
 .. note::
   Handling invalidly encoded slots is currently not supported but may be added in the future.
--- a/docs/internals/layout_in_storage.rst
+++ b/docs/internals/layout_in_storage.rst
@@ -128,7 +128,7 @@ The type of the value is ``uint256``, so it uses a single slot.
 ------------------------
 
 ``bytes`` and ``string`` are encoded identically.
-In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array itself and
+In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array length itself and
 a data area that is computed using a ``keccak256`` hash of that slot's position.
 However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
 


chriseth:
Thank you for your contribution! The slot represents the array. What is stored there is the length. So I would say the current wording is conceptually correct while the change adds a complication.

ghagevaibhav (approved):
looks good, thanks

ghagevaibhav (dismissed):

docs/internals/layout_in_storage.rst
```diff
@@ -128,7 +128,7 @@ The type of the value is ``uint256``, so it uses a single slot.
 ------------------------
 
 ``bytes`` and ``string`` are encoded identically.
-In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array itself and
+In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array length itself and
 a data area that is computed using a ``keccak256`` hash of that slot's position.
 However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
```
cameel: I agree with @chriseth that the change does not necessarily clarify things because the slot is not meant only for length.

If we were to expand this, this would be my proposal:
```
In general, the encoding is similar to ``bytes1[]``, in the sense that it consists of one fixed slot and
a separately allocated data area whose location is computed using a ``keccak256`` hash of that slot's position.
However, the fixed slot is not used exclusively to store the length.
For short values (shorter than 32 bytes) the array elements are stored together with the length in the fixed slot.
```

CJ42:
@cameel here is a new suggestion for this section.

Maybe breaking this down into bullet points could fit nicely, making it easier to read and in turns make the layout for these two variable types easier to understand.

Here is my proposal extended from yours:

```
``bytes`` and ``string`` are encoded identically. In general, the encoding is similar to ``bytes1[]``. The layout in storage for these types depends on the ``.length``:

- For short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
- For long values (more than 32 bytes), the length is stored in one slot (which position is defined as per the [storage layout rules](#layout-of-state-variables-in-storage)), while the array elements are stored in a separately allocated data area whose location is computed using a ``keccak256`` hash of that slot's position.
```

CJ42:
@cameel @chriseth @vaibhav4516 any updates? What do you think of the suggested changes in my last commit?

docs/internals/layout_in_storage.rst
```diff
@@ -127,16 +127,20 @@ The type of the value is ``uint256``, so it uses a single slot.
...
-In particular: if the data is at most ``31`` bytes long, the elements are stored
+- For short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
+
+If the data is at most ``31`` bytes long, the elements are stored
 in the higher-order bytes (left aligned) and the lowest-order byte stores the value ``length * 2``.
-For byte arrays that store data which is ``32`` or more bytes long, the main slot ``p`` stores ``length * 2 + 1`` and the data is
-stored as usual in ``keccak256(p)``. This means that you can distinguish a short array from a long array
-by checking if the lowest bit is set: short (not set) and long (set).
+
+- For long values (more than 32 bytes), the length is stored in one slot (which position is defined as per the [storage layout rules](#layout-of-state-variables-in-storage)), while the array elements are stored in a separately allocated data area whose location is computed using a ``keccak256`` hash of that slot's position.
```
nikola-matic, line 138:
```suggestion
- For long values (32 bytes or more), the length is stored in one slot (whose position is defined as per the [storage layout rules](#layout-of-state-variables-in-storage)), while the array elements are stored in a separately allocated data area whose location is computed using a ``keccak256`` hash of that slot's position.
```

docs/internals/layout_in_storage.rst
```diff
@@ -127,16 +127,20 @@ The type of the value is ``uint256``, so it uses a single slot.
...
 
-``bytes`` and ``string`` are encoded identically.
-In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array itself and
-a data area that is computed using a ``keccak256`` hash of that slot's position.
-However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
+``bytes`` and ``string`` are encoded identically. In general, the encoding is similar to ``bytes1[]`` in the
+ sense that the layout in storage depends on the ``.length``:
 
-In particular: if the data is at most ``31`` bytes long, the elements are stored
+- For short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
```
nikola-matic, line 133:
```suggestion
- For short values (less than 32 bytes) the array elements are stored together with the length in the same slot.
```

nikola-matic:
Coding style check is failing; you can either check the failed CI job, or even better, run `./scripts/check_style.sh` locally after making changes.

docs/internals/layout_in_storage.rst
```diff
@@ -127,16 +127,20 @@ The type of the value is ``uint256``, so it uses a single slot.
 ``bytes`` and ``string``
 ------------------------
 
-``bytes`` and ``string`` are encoded identically.
-In general, the encoding is similar to ``bytes1[]``, in the sense that there is a slot for the array itself and
-a data area that is computed using a ``keccak256`` hash of that slot's position.
-However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.
+``bytes`` and ``string`` are encoded identically. In general, the encoding is similar to ``bytes1[]`` in the
+ sense that the layout in storage depends on the ``.length``:
```
cameel, line 131: But this is not true. The layout of `bytes1[]` does not depend on `.length`. The similarity is that the long form of `bytes` and `bytes1[]` both have a length slot and a separate data area.

docs/internals/layout_in_storage.rst
cameel, line 138: This is not a markdown file. RST has different syntax for links (please check the text around).

docs/internals/layout_in_storage.rst
cameel, line 133: Maybe bullet points would work but in the current form you have two single-point lists, which is a bit weird.

leonardoalt:
The PR seems to change the docs in a way that the reviewers disagree, so I'm closing it for now. Reviewers please comment otherwise.

PR #13274, opened 2022-07-13
lsp: Fixes initialization phase if `include-paths` was not set at all then also no error should be generated.

christianparpart:
The reason why CI did not fail for that is because it's simply reported as message, and our CI test client does not strictly evaluate trace messages in that regard. I am also not sure that should be done or if that would be overkill. Nevertheless, that's fixing not having `include-paths` specified in the custom settings.


cameel (approved):

cameel:
As for traces, maybe they should be included as expectations in test files, the same way we include emitted events in compiler's semantic tests?

christianparpart:
> As for traces, maybe they should be included as expectations in test files, the same way we include emitted events in compiler's semantic tests?

we could have that as a separate-PR thing, with some feedback from @Marenz later. But thanks so far :)

PR #13163, opened 2022-06-17
formatNumberReadable now prints signed integers as well

Ruko97:
Fixes issue #9601

PR #13277, opened 2022-07-15
Improved output of formatNumberReadable

Ruko97:
Follow up to issue #9601 and pull request #13163.
Based on advice by @cameel in the conversation in issue.
Previously `formatNumberReadable(57896044618658097711785492504343953926634992332820282019728792003956564819968) `would equal to `0x80 * 2 ** 248`, but now it equals to `2 ** 255`.
Also, there was a bug in the test case `test_format_number_readable_signed` in `test/StringUtils.cpp`. `formatNumberReadable(-0xffffffff)` should be equal to `-2 ** 32 + 1` but previously it was tested to be equal to `-2 ** 32 - 1`. The new version of `formatNumberReadable` fixes this problem.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,8 @@ Compiler Features:
 Bugfixes:
  * Yul Optimizer: Hash hex and decimal literals according to their value instead of their representation, improving the detection of equivalent functions.
  * Solidity Upgrade Tool ``solidity-upgrade``: Fix the tool returning success code on uncaught exceptions.
+ * SMTChecker: Fix display error for negative integers that are one more than powers of two.
+ * SMTChecker: Improved readability for large integers that are powers of two or almost powers of two in error messages.
 
 
 ### 0.8.17 (2022-09-08)
--- a/libsolutil/StringUtils.h
+++ b/libsolutil/StringUtils.h
@@ -27,6 +27,8 @@
 #include <libsolutil/CommonData.h>
 #include <libsolutil/Numeric.h>
 
+#include <fmt/format.h>
+
 #include <algorithm>
 #include <limits>
 #include <locale>
@@ -101,110 +103,17 @@ std::string joinHumanReadablePrefixed
 		return _separator + joinHumanReadable(_list, _separator, _lastSeparator);
 }
 
-/// Same as @ref formatNumberReadable but only for unsigned numbers
-template <class T>
-inline std::string formatUnsignedNumberReadable (
-	T const& _value,
-	bool _useTruncation = false
-)
-{
-	static_assert(
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
-
-	// smaller numbers return as decimal
-	if (_value <= 0x1000000)
-		return _value.str();
-
-	HexCase hexcase = HexCase::Mixed;
-	HexPrefix prefix = HexPrefix::Add;
-
-	// when multiple trailing zero bytes, format as N * 2**x
-	int i = 0;
-	T v = _value;
-	for (; (v & 0xff) == 0; v >>= 8)
-		++i;
-	if (i > 2)
-	{
-		// 0x100 yields 2**8 (N is 1 and redundant)
-		if (v == 1)
-			return "2**" + std::to_string(i * 8);
-		return toHex(toCompactBigEndian(v), prefix, hexcase) +
-			" * 2**" +
-			std::to_string(i * 8);
-	}
-
-	// when multiple trailing FF bytes, format as N * 2**x - 1
-	i = 0;
-	for (v = _value; (v & 0xff) == 0xff; v >>= 8)
-		++i;
-	if (i > 2)
-	{
-		// 0xFF yields 2**8 - 1 (v is 0 in that case)
-		if (v == 0)
-			return "2**" + std::to_string(i * 8) + " - 1";
-		return toHex(toCompactBigEndian(T(v + 1)), prefix, hexcase) +
-			" * 2**" + std::to_string(i * 8) +
-			" - 1";
-	}
-
-	std::string str = toHex(toCompactBigEndian(_value), prefix, hexcase);
-	if (_useTruncation)
-	{
-		// return as interior-truncated hex.
-		size_t len = str.size();
-
-		if (len < 24)
-			return str;
-
-		size_t const initialChars = (prefix == HexPrefix::Add) ? 6 : 4;
-		size_t const finalChars = 4;
-		size_t numSkipped = len - initialChars - finalChars;
-
-		return str.substr(0, initialChars) +
-			"...{+" +
-			std::to_string(numSkipped) +
-			" more}..." +
-			str.substr(len-finalChars, len);
-	}
-
-	// otherwise, show whole value.
-	return str;
-}
-
 /// Formats large numbers to be easily readable by humans.
 /// Returns decimal representation for smaller numbers; hex for large numbers.
 /// "Special" numbers, powers-of-two and powers-of-two minus 1, are returned in
 /// formulaic form like 0x01 * 2**24 - 1.
-/// @a T can be any integer variable, will typically be u160, u256 or bigint.
+/// @a T can be any integer type, will typically be u160, u256 or bigint.
 /// @param _value to be formatted
 /// @param _useTruncation if true, internal truncation is also applied,
 /// like  0x5555...{+56 more}...5555
-/// @example formatNumberReadable((u256)0x7ffffff) = "0x08 * 2**24"
-/// @example formatNumberReadable(-57896044618658097711785492504343953926634992332820282019728792003956564819968) = -0x80 * 2**248
-template <class T>
-inline std::string formatNumberReadable(
-	T const& _value,
-	bool _useTruncation = false
-)
-{
-	static_assert(
-		std::numeric_limits<T>::is_integer,
-		"only integer numbers are supported"
-	);
-
-	if (_value >= 0)
-	{
-		bigint const _v = bigint(_value);
-		return formatUnsignedNumberReadable(_v, _useTruncation);
-	}
-	else
-	{
-		bigint const _abs_value = bigint(-1) * _value;
-		return "-" + formatUnsignedNumberReadable(_abs_value, _useTruncation);
-	}
-}
+/// @example formatNumberReadable((u256)0x7ffffff) = "2**27 - 1"
+/// @example formatNumberReadable(-57896044618658097711785492504343953926634992332820282019728792003956564819968) = -2**255
+std::string formatNumberReadable(bigint const& _value, bool _useTruncation = false);
 
 /// Safely converts an unsigned integer as string into an unsigned int type.
 ///
--- a/test/libsolutil/StringUtils.cpp
+++ b/test/libsolutil/StringUtils.cpp
@@ -107,17 +107,23 @@ BOOST_AUTO_TEST_CASE(test_human_readable_join)
 
 BOOST_AUTO_TEST_CASE(test_format_number_readable)
 {
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x8000000)), "0x08 * 2**24");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x80000000)), "0x80 * 2**24");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x800000000)), "0x08 * 2**32");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x8000000000)), "0x80 * 2**32");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x80000000000)), "0x08 * 2**40");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x10000000)), "2**28");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x100000000)), "2**32");
 
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffff)), "0x08 * 2**24 - 1");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7fffffff)), "0x80 * 2**24 - 1");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffffff)), "0x08 * 2**32 - 1");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7fffffffff)), "0x80 * 2**32 - 1");
-	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffffffff)), "0x08 * 2**40 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x8000000)), "2**27");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x80000000)), "2**31");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x800000000)), "2**35");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x8000000000)), "2**39");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x80000000000)), "2**43");
+
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffff)), "2**27 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7fffffff)), "2**31 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffffff)), "2**35 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7fffffffff)), "2**39 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x7ffffffffff)), "2**43 - 1");
+
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0xfffffff)), "2**28 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0xffffffff)), "2**32 - 1");
 
 	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x88000000)), "0x88 * 2**24");
 	BOOST_CHECK_EQUAL(formatNumberReadable(u256(0x8888888888000000)), "0x8888888888 * 2**24");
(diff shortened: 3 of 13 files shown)


libsolutil/StringUtils.h
```diff
@@ -101,109 +101,122 @@ std::string joinHumanReadablePrefixed
...
 {
 	static_assert(
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
+		std::numeric_limits<T>::is_integer,
+		"only integer numbers are supported"
+	);
+
+	bool _isNegative = false;
```
matheusaaguiar: Names starting with leading underscore are used for[ parameter variables](https://github.com/ethereum/solidity/blob/develop/CODING_STYLE.md#4-variable-prefixes). Please remove `_` from here and also `_sign` below so it is clear they are local.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,122 @@ std::string joinHumanReadablePrefixed
...
 		return _value.str();
 
+	if (_isNegative && bigint(-1 * _value) <= 0x1000000)
+		return "-" + (bigint(-1 * _value)).str();
+
 	HexCase hexcase = HexCase::Mixed;
-	HexPrefix prefix = HexPrefix::Add;
+	HexPrefix hexprefix = HexPrefix::Add;
+
+	std::string _sign = _isNegative ? "-" : "";
```
matheusaaguiar: Same as before regarding underscore. 
Also, maybe we could declare and define this before so it can be used in the previous `return` values.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,122 @@ std::string joinHumanReadablePrefixed
...
 	static_assert(
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
+		std::numeric_limits<T>::is_integer,
+		"only integer numbers are supported"
+	);
+
+	bool _isNegative = false;
+	bigint v;
```
matheusaaguiar:
```suggestion
	bigint signedValue = isNegative ? (bigint(-1) * _value) : bigint(_value);
	bigint v;
```
What do you think about using `v` as the working var used by the `for` loops? It can then be initialized with `signedValue` inside each `for` instead of being assigned the ternary operator result before the loops.
Also, `signedValue` can be used in the `return` values.

matheusaaguiar (changes requested):
Looks good!
Just some code style adjustments and readability suggestions.

libsolutil/StringUtils.h
Ruko97: I defined `sign` after the previous return values because the condition for entering the return statement was to have `isNegative` be true. So there wasn't any need to use the value of sign instead of just adding `"-"` before the value

libsolutil/StringUtils.h
Ruko97: Yes, I think that would be a great idea.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,117 @@ std::string joinHumanReadablePrefixed
...
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
+		std::numeric_limits<T>::is_integer,
+		"only integer numbers are supported"
+	);
+
+	bool isNegative = _value < 0;
+	bigint temp;
+	bigint absoluteValue = isNegative ? (bigint(-1) * _value) : bigint(_value);
```
matheusaaguiar: Maybe a little nitpick, but it seems to me that `absoluteValue` means how far from zero the number is and then it cannot be negative. What about `decimalValue` (or `signedValue` as suggested before)?

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,117 @@ std::string joinHumanReadablePrefixed
...
 
-	// when multiple trailing zero bytes, format as N * 2**x
+	// when multiple trailing zero bytes, format as N * 2**x (if positive)
+	// or -N * 2**x (if negative)
 	int i = 0;
-	T v = _value;
-	for (; (v & 0xff) == 0; v >>= 8)
+	temp = absoluteValue;
+
+	for (; (temp & 0xff) == 0; temp >>= 8)
```
matheusaaguiar:
```suggestion
	for (temp = absoluteValue; (temp & 0xff) == 0; temp >>= 8)
```
This is more compact and has the same readability.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,117 @@ std::string joinHumanReadablePrefixed
...
 	}
 
-	// when multiple trailing FF bytes, format as N * 2**x - 1
+	// when multiple trailing FF bytes, format as N * 2**x - 1 (if positive)
+	// or -N * 2**x + 1 (if negative)
 	i = 0;
-	for (v = _value; (v & 0xff) == 0xff; v >>= 8)
+	temp = absoluteValue;
+
+	for (; (temp & 0xff) == 0xff; temp >>= 8)
```
matheusaaguiar:
```suggestion
	for (temp = absoluteValue; (temp & 0xff) == 0xff; temp >>= 8)
```

libsolutil/StringUtils.h
Ruko97: I think signedValue would be better

matheusaaguiar:
I think this is in good shape. Please, rebase.

Ruko97:
Pushed a rebased version right now. Please check it.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
 		return _separator + joinHumanReadable(_list, _separator, _lastSeparator);
 }
 
-/// Same as @ref formatNumberReadable but only for unsigned numbers
+/// Formats large numbers to be easily readable by humans.
+/// Returns decimal representation for smaller numbers; hex for large numbers.
+/// "Special" numbers, powers-of-two and powers-of-two minus 1, are returned in
+/// formulaic form like 0x01 * 2**24 - 1.
+/// @a T can be any integer variable, will typically be u160, u256 or bigint.
```
matheusaaguiar:
```suggestion
/// @a T can be any integer type, will typically be u160, u256 or bigint.
```
Ignore upper case letter `A`, it should be `a`, but github is somehow tagging a user there... 
Just to be clear, suggestion is to change `variable` to `type`

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
+/// @param _value to be formatted
+/// @param _useTruncation if true, internal truncation is also applied,
+/// like  0x5555...{+56 more}...5555
+/// @example formatNumberReadable((u256)0x7ffffff) = "0x08 * 2**24"
```
matheusaaguiar:
```suggestion
/// @example formatNumberReadable((u256)0x7ffffff) = "2**27 - 1"
```
Shouldn't this have the same value as in the check at line 119 of `test/libsolutil/StringUtils.cpp` ?

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
 	T const& _value,
 	bool _useTruncation = false
 )
 {
 	static_assert(
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
+		std::numeric_limits<T>::is_integer,
+		"only integer numbers are supported"
```
matheusaaguiar:
```suggestion
		"only integer types are supported"
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
+	// or -N * 2**x + 1 (if negative)
 	i = 0;
-	for (v = _value; (v & 0xff) == 0xff; v >>= 8)
+	for (temp = signedValue; (temp & 0xff) == 0xff; temp >>= 8)
 		++i;
 	if (i > 2)
 	{
+		std::string suffix = isNegative ? " + 1" : " - 1";
+
 		// 0xFF yields 2**8 - 1 (v is 0 in that case)
```
matheusaaguiar:
```suggestion
		// 0xFF yields 2**8 - 1 (temp is 0 in that case)
```
Also, I guess we could remove `(temp is 0 in that case)`.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
+				" * 2**" + std::to_string(i * 8) +
+				suffix;
 	}
 
-	std::string str = toHex(toCompactBigEndian(_value), prefix, hexcase);
+	std::string str = toHex(toCompactBigEndian(signedValue), hexprefix, hexcase);
+
 	if (_useTruncation)
 	{
 		// return as interior-truncated hex.
```
matheusaaguiar: I'd say that the `if` line already makes this clear. I think the comment can be removed.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
-		return str.substr(0, initialChars) +
+		return sign +
+			str.substr(0, initialChars) +
 			"...{+" +
 			std::to_string(numSkipped) +
 			" more}..." +
 			str.substr(len-finalChars, len);
 	}
 
 	// otherwise, show whole value.
```
matheusaaguiar: I think this is redundant too.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,113 @@ std::string joinHumanReadablePrefixed
...
+	if (signedValue <= 0x1000000)
+		return sign + signedValue.str();
 
 	HexCase hexcase = HexCase::Mixed;
-	HexPrefix prefix = HexPrefix::Add;
+	HexPrefix hexprefix = HexPrefix::Add;
 
-	// when multiple trailing zero bytes, format as N * 2**x
+	// when multiple trailing zero bytes, format as N * 2**x (if positive)
+	// or -N * 2**x (if negative)
```
matheusaaguiar:
```suggestion
	// when multiple trailing zero bytes, format as N * 2**x
```
I think that distinction is handled by `sign` and only relevant at the `return`, so no need to have a multiple line comment.

matheusaaguiar (changes requested):
Sorry, I had another look and found some things that were glossed over before and could be improved.

leonardoalt:
Auto please rebase onto origin/develop to get rid of the CI errors.

libsolutil/StringUtils.h
Ruko97: It could help other programmers update the code.

libsolutil/StringUtils.h
Ruko97: I wanted to write `format as ±N * 2**x` in the comment here to indicate that the printed variable is shown with its sign, but I wasn't sure if Unicode is supported in comments.

Ruko97:
I have rebased and updated my commit based on the suggestions here. Sorry for being late.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,111 @@ std::string joinHumanReadablePrefixed
 		return _separator + joinHumanReadable(_list, _separator, _lastSeparator);
 }
 
-/// Same as @ref formatNumberReadable but only for unsigned numbers
+/// Formats large numbers to be easily readable by humans.
+/// Returns decimal representation for smaller numbers; hex for large numbers.
+/// "Special" numbers, powers-of-two and powers-of-two minus 1, are returned in
+/// formulaic form like 0x01 * 2**24 - 1.
+/// @a T can be any integer value, will typically be u160, u256 or bigint.
```
matheusaaguiar: I don't want to be pedantic but it looks to me that `T` refers to a type like `u160` or `bigint`. It is the function param variable   `_value` that stores the integer value of type `T`.
```suggestion
/// @a T can be any integer type, will typically be u160, u256 or bigint.
```

libsolutil/StringUtils.h
matheusaaguiar: Right, fair enough. I've noticed that also that you changed the templates to `N * 2**x - 1` and `-N * 2**x + 1` , so it seems worth having a comment about this.

libsolutil/StringUtils.h
matheusaaguiar: What about this and the next comment? I think that the code already makes it explicit, so we could remove this line.
```suggestion

```

libsolutil/StringUtils.h
matheusaaguiar:
```suggestion

```

libsolutil/StringUtils.h
Ruko97: Sorry, I accidentally might have forgot to add this edit to my code. My bad. It's fixed now.

matheusaaguiar: No problem :)

libsolutil/StringUtils.h
matheusaaguiar: @Ruko97 , what do you think about this?

libsolutil/StringUtils.h
Ruko97: Fixed now

libsolutil/StringUtils.h
Ruko97: This comment existed in the prior version of the function, I just left it as it is.

matheusaaguiar:
> I have rebased and updated my commit based on the suggestions here. Sorry for being late.

No need to apologize, you are not late at all. You are actually quite timely with your responses. Thanks for contributing to the project :)

matheusaaguiar (commented):
I consider it ready. But it needs another rebase on top of `develop` since we are having a release today.

matheusaaguiar (dismissed):
It is going to need another rebase, but I am approving so someone else from the team can merge it later.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,109 @@ std::string joinHumanReadablePrefixed
...
 	if (i > 2)
 	{
 		// 0x100 yields 2**8 (N is 1 and redundant)
-		if (v == 1)
-			return "2**" + std::to_string(i * 8);
-		return toHex(toCompactBigEndian(v), prefix, hexcase) +
-			" * 2**" +
-			std::to_string(i * 8);
+		if (temp == 1)
+			return sign + "2**" + std::to_string(i * 8);
```
nikola-matic: Hey, sorry for being a little late to the party. Can you please swap all instances of string concatenation with `fmt::format`. 
E.g., in this case, it would be something like this:
```
return fmt::format("{}2**{}", sign, std::to_string(i * 8);
```
Otherwise, looks good!

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,109 @@ std::string joinHumanReadablePrefixed
...
+			return sign + "2**" + std::to_string(i * 8 + j);
+		}
+		else
+			return sign +
+				toHex(toCompactBigEndian(temp), hexprefix, hexcase) +
+				" * 2**" +
+				std::to_string(i * 8);
 	}
 
 	// when multiple trailing FF bytes, format as N * 2**x - 1
```
nikola-matic: Shouldn't this now be commented as `2**x  - 1` instead of `N * 2**x  - 1`? Or at least that `N` is optional depending on whether the number is a power of 2?

test/libsolutil/StringUtils.cpp
```diff
@@ -190,7 +206,7 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
 
 	BOOST_CHECK_EQUAL(formatNumberReadable(b, true), "-0x5555...{+56 more}...5555");
 	BOOST_CHECK_EQUAL(formatNumberReadable(c, true), "-0x0BCD...{+56 more}...6789");
-	BOOST_CHECK_EQUAL(formatNumberReadable(d, true), "-0x5555555555555556 * 2**192 - 1");
+	BOOST_CHECK_EQUAL(formatNumberReadable(d, true), "-0x5555555555555556 * 2**192 + 1");
```
matheusaaguiar, line 265: Does this change has to do with the bug you mentioned, as in
> `formatNumberReadable(-0xffffffff)` should be equal to `-2 ** 32 + 1` but previously it was tested to be equal to `-2 ** 32 - 1`

?

Also, it is a bit difficult to track what these hex values translate to as decimals. How are you making sure that this is indeed the right value?

libsolutil/StringUtils.h
Ruko97: No, not always. `N` can be unequal to 1 if the variable is not exactly a power of 2 or one-less-than a power of 2. For example, `9 * 2**249` would be displayed as `18 * 2**248`.

test/libsolutil/StringUtils.cpp
Ruko97: Yes, the change is related to the bug I mentioned
Much of the test values is carried from previous values that were being tested. I only changed some of the variables to negative. I also used `isoltest` to verify the values.

libsolutil/StringUtils.h
Ruko97: Perhaps I should add a test regarding this.

libsolutil/StringUtils.h
Ruko97: ok, I am doing it now.

libsolutil/StringUtils.h
matheusaaguiar: That is a good idea. We can never have enough tests.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,109 @@ std::string joinHumanReadablePrefixed
...
 
 	HexCase hexcase = HexCase::Mixed;
-	HexPrefix prefix = HexPrefix::Add;
+	HexPrefix hexprefix = HexPrefix::Add;
 
 	// when multiple trailing zero bytes, format as N * 2**x
 	int i = 0;
-	T v = _value;
-	for (; (v & 0xff) == 0; v >>= 8)
+	for (temp = signedValue; (temp & 0xff) == 0; temp >>= 8)
```
axic: I think it may be better scoping `temp` here and in the next loop, as they are not used inbetween.

axic: Would also keep it named `v` as before.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,109 @@ std::string joinHumanReadablePrefixed
...
 
 	// smaller numbers return as decimal
-	if (_value <= 0x1000000)
-		return _value.str();
+	if (signedValue <= 0x1000000)
+		return sign + signedValue.str();
 
 	HexCase hexcase = HexCase::Mixed;
-	HexPrefix prefix = HexPrefix::Add;
+	HexPrefix hexprefix = HexPrefix::Add;
```
axic: I'd also name these `hexCase` and `hexPrefix`.

libsolutil/StringUtils.h
```diff
@@ -101,109 +101,109 @@ std::string joinHumanReadablePrefixed
...
-		std::is_same<bigint, T>::value || !std::numeric_limits<T>::is_signed,
-		"only unsigned types or bigint supported"
-	); //bigint does not carry sign bit on shift
+		std::numeric_limits<T>::is_integer,
+		"only integer types are supported"
+	);
+
+	bool isNegative = _value < 0;
+	bigint temp;
+	bigint signedValue = isNegative ? (bigint(-1) * _value) : bigint(_value);
```
axic: Since this is copied, perhaps make it const. And `sign` too.

libsolutil/StringUtils.h
Ruko97: wouldn't using `temp` here make it more clear what the role of the variable is? It took me some time to understand what role `v` played in the code; and I think other programmers would face a similar situation when looking at the code.

matheusaaguiar: I agree with naming the variable as `temp` and also with limiting its scope to only inside the loops.

Ruko97: ok, will do that

Ruko97:
The errors in CircleCI seem to be related to not being able to find `fmt::format`. Does CircleCI not use the fmt library?

cameel:
It's there but we added it as a dependency very recently so not all of our libs link to it yet. You have to add it to `target_link_libraries` in `libsolutil/CMakeLists.txt`. See how #13473 does it.

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+
+		// when multiple trailing zero bytes, format as N * 2**x
+		int i = 0;
+		for (temp = signedValue; (temp & 0xff) == 0; temp >>= 8)
+			++i;
+		if (i > 2)
+		{
+			// 0x100 yields 2**8 (N is 1 and redundant)
+			if (temp == 1)
+				return fmt::format("{}2**{}", sign, std::to_string(i * 8));
```
matheusaaguiar:
```suggestion
				return fmt::format("{}2**{}", sign, i * 8);
```
I guess we can remove the conversion to string now, since `fmt::format` should be able to handle it.

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+			else if ((temp & (temp-1)) == 0)
+			{
+				int j = 0;
+				for (; (temp & 0x1) == 0; temp >>= 1)
+					j++;
+				return fmt::format("{}2**{}", sign, std::to_string(i * 8 + j));
```
matheusaaguiar:
```suggestion
				return fmt::format("{}2**{}", sign, i * 8 + j);
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+			}
+			else
+				return fmt::format(
+					"{}{} * 2**{}",
+					sign,
+					toHex(toCompactBigEndian(temp), hexPrefix, hexCase),
+					std::to_string(i * 8)
```
matheusaaguiar:
```suggestion
					i * 8
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+		// when multiple trailing FF bytes, format as N * 2**x - 1
+		i = 0;
+		for (temp = signedValue; (temp & 0xff) == 0xff; temp >>= 8)
+			++i;
+		if (i > 2)
+		{
+			std::string suffix = isNegative ? " + 1" : " - 1";
+
+			if (temp == 0)
+				return fmt::format("{}2**{}{}", sign, std::to_string(i * 8), suffix);
```
matheusaaguiar:
```suggestion
				return fmt::format("{}2**{}{}", sign, i * 8, suffix);
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+			else if ((temp & (temp + 1)) == 0)
+			{
+				int j = 0;
+				for (; (temp & 0x1) != 0; temp >>= 1)
+					j++;
+				return fmt::format("{}2**{}{}", sign, std::to_string(i * 8 + j), suffix);
```
matheusaaguiar:
```suggestion
				return fmt::format("{}2**{}{}", sign, i * 8 + j, suffix);
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
+			}
+			else
+				return fmt::format(
+					"{}{} * 2**{}{}",
+					sign,
+					toHex(toCompactBigEndian(++temp), hexPrefix, hexCase),
+					std::to_string(i * 8),
```
matheusaaguiar:
```suggestion
					i * 8,
```

libsolutil/StringUtils.h
```diff
@@ -101,109 +103,119 @@ std::string joinHumanReadablePrefixed
...
-		return str.substr(0, initialChars) +
-			"...{+" +
-			std::to_string(numSkipped) +
-			" more}..." +
-			str.substr(len-finalChars, len);
+		return fmt::format(
+			"{}{}...{{+{} more}}...{}",
+			sign,
+			str.substr(0, initialChars),
+			std::to_string(numSkipped),
```
matheusaaguiar:
```suggestion
			numSkipped,
```

matheusaaguiar (commented):
Some adjustments and I wanted to confirm that you added the test case mentioned [here](https://github.com/ethereum/solidity/pull/13277#discussion_r978648348).
Also, @nikola-matic are you satisfied with the current replacements of string concatenation or do you think that they are needed for the simple cases at lines `133`, `203` and `218`?

nikola-matic:
> Some adjustments and I wanted to confirm that you added the test case mentioned [here](https://github.com/ethereum/solidity/pull/13277#discussion_r978648348). Also, @nikola-matic are you satisfied with the current replacements of string concatenation or do you think that they are needed for the simple cases at lines `133`, `203` and `218`?

The formatting looks good now, and no, no need for `fmt::format` for simple cases. Another rebase would be nice to get rid of the `gp2` failure.

Ruko97:
> Some adjustments and I wanted to confirm that you added the test case mentioned [here](https://github.com/ethereum/solidity/pull/13277#discussion_r978648348). Also, @nikola-matic are you satisfied with the current replacements of string concatenation or do you think that they are needed for the simple cases at lines `133`, `203` and `218`?

Yes, I added them at lines `185`, `186`, `285` and `286`. I added some more tests now just to make sure.

Ruko97:
I am also removing the `std::to_string` functions, and rebasing the code to the latest version.

libsolutil/StringUtils.h
matheusaaguiar: This is still pending.

libsolutil/StringUtils.h
matheusaaguiar: This is still pending.

libsolutil/StringUtils.h
Ruko97: fixed

libsolutil/StringUtils.h
Ruko97: fixed

nikola-matic:
@matheusaaguiar @ekpyron will this need a changelog entry?

ekpyron:
> @matheusaaguiar @ekpyron will this need a changelog entry?

I'd say, it's borderline. But strictly speaking it's user-facing, so we can just default to "yes".

matheusaaguiar:
@Ruko97 , please add a [Changelog](https://github.com/ethereum/solidity/blob/46a7ebd5b66c5638427295b9c3dbf15fc48c4ed2/Changelog.md) entry under the `Compiler Features` section, describing shortly the improvement on readability of large/hex numbers. Also, the fixed bug, under section `Bugfixes`.

Changelog.md
```diff
@@ -8,9 +8,11 @@ Compiler Features:
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 * Language Server: Add basic document hover support.
+* SMTChecker: Improved readability for large integers that are powers of two or almost powers of two
```
nikola-matic: @leonardoalt is this SMT checker related?

nikola-matic:
Rebased, resolved conflicts, and added periods to changelog entires. The build should now be fine, and hopefully closer to merging.

NunoFilipeSantos (approved):
LGTM

matheusaaguiar (approved):
LGTM

PR #13279, opened 2022-07-16
Fix out-of-range compiler crash error for unary tuples

### LINKED ISSUE #13208: [TypeChecker] Out-of-range exception when trying to assign to nested tuples
```
function f() {
  ((, ())) = (f, 3);
}
```

throws

https://github.com/ethereum/solidity/blob/3ed9a38abc9c9babf50a25f56bff2411da9cdfe7/libsolidity/analysis/TypeChecker.cpp#L168-L172

timweri:
closes #13208

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@ Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
  * Type Checker: Fix compiler crash when `abi.encodeCall` received a tuple expression instead of an inline tuple.
+ * Type Checker: Fix compiler crash on tuple assignments involving certain patterns with unary tuples on the left-hand side.
 
 
 ### 0.8.15 (2022-06-15)
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -112,8 +112,9 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
 	size_t storageByteAccesses = 0;
 	auto count = [&](TupleExpression const& _lhs, TupleType const& _rhs, auto _recurse) -> void {
 		TupleType const& lhsType = dynamic_cast<TupleType const&>(*type(_lhs));
+		TupleExpression const* lhsResolved = dynamic_cast<TupleExpression const*>(resolveOuterUnaryTuples(&_lhs));
 
-		if (lhsType.components().size() != _rhs.components().size())
+		if (lhsType.components().size() != _rhs.components().size() || lhsResolved->components().size() != _rhs.components().size())
 		{
 			solAssert(m_errorReporter.hasErrors(), "");
 			return;
@@ -134,7 +135,7 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
 			{
 				if (bytesType && bytesType->numBytes() == 1)
 				{
-					if (FunctionCall const* lhsCall = dynamic_cast<FunctionCall const*>(resolveOuterUnaryTuples(_lhs.components().at(index).get())))
+					if (FunctionCall const* lhsCall = dynamic_cast<FunctionCall const*>(resolveOuterUnaryTuples(lhsResolved->components().at(index).get())))
 					{
 						FunctionType const& callType = dynamic_cast<FunctionType const&>(*type(lhsCall->expression()));
 						if (callType.kind() == FunctionType::Kind::ArrayPush)
@@ -147,7 +148,7 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
 							}
 						}
 					}
-					else if (IndexAccess const* indexAccess = dynamic_cast<IndexAccess const*>(resolveOuterUnaryTuples(_lhs.components().at(index).get())))
+					else if (IndexAccess const* indexAccess = dynamic_cast<IndexAccess const*>(resolveOuterUnaryTuples(lhsResolved->components().at(index).get())))
 					{
 						if (ArrayType const* arrayType = dynamic_cast<ArrayType const*>(type(indexAccess->baseExpression())))
 							if (arrayType->isByteArray() && arrayType->dataStoredIn(DataLocation::Storage))
@@ -156,7 +157,7 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
 				}
 			}
 			else if (TupleType const* tupleType = dynamic_cast<TupleType const*>(componentType))
-				if (auto const* lhsNested = dynamic_cast<TupleExpression const*>(_lhs.components().at(index).get()))
+				if (auto const* lhsNested = dynamic_cast<TupleExpression const*>(lhsResolved->components().at(index).get()))
 					if (auto const* rhsNestedType = dynamic_cast<TupleType const*>(_rhs.components().at(index)))
 						_recurse(
 							*lhsNested,
--- a/test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
+++ b/test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
@@ -1,0 +1,35 @@
+function f() pure {
+    ((, ())) = (1, 3);
+}
+
+function g() pure {
+    ((, ((, ())))) = (1, 3);
+}
+
+function t() pure returns (int, int) {
+    return (4, 5);
+}
+
+function h() pure {
+    ((, ())) = t();
+}
+
+function ff() pure {
+    ((((, ())) , )) = ((1, 2), 3);
+}
+
+function fg() pure {
+    (((, ())) , ) = ((1, 2), 3);
+}
+
+// ----
+// TypeError 5547: (28-30): Empty tuple on the left hand side.
+// TypeError 7407: (35-41): Type tuple(int_const 1,int_const 3) is not implicitly convertible to expected type tuple(,tuple()).
+// TypeError 5547: (78-80): Empty tuple on the left hand side.
+// TypeError 7407: (87-93): Type tuple(int_const 1,int_const 3) is not implicitly convertible to expected type tuple(,tuple(,tuple())).
+// TypeError 5547: (187-189): Empty tuple on the left hand side.
+// TypeError 7407: (194-197): Type tuple(int256,int256) is not implicitly convertible to expected type tuple(,tuple()).
+// TypeError 5547: (233-235): Empty tuple on the left hand side.
+// TypeError 7407: (245-256): Type tuple(tuple(int_const 1,int_const 2),int_const 3) is not implicitly convertible to expected type tuple(tuple(,tuple()),).
+// TypeError 5547: (291-293): Empty tuple on the left hand side.
+// TypeError 7407: (302-313): Type tuple(tuple(int_const 1,int_const 2),int_const 3) is not implicitly convertible to expected type tuple(tuple(,tuple()),).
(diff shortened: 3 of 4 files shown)


Changelog.md
```diff
@@ -14,6 +14,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix out of range error nested tuples.
```
matheusaaguiar:
```suggestion
 * Type Checker: Fix out of range error for nested tuples.
```

matheusaaguiar (commented):
Looks good. The failing tests are a problem unrelated to this PR. Just need the minor amendment in `Changelog`.

Changelog.md
timweri: Fixed!

chriseth:
Can you explain why this is the proper fix?

timweri:
> Can you explain why this is the proper fix?

`_lhs` is the original `TupleExpression` that might be wrapped in outer unary tuples.
`index` in the for loop refers to the index of `lhsType` which is a `TupleType`.
So if `_lhs` is wrapped in outer unary tuples, `_lhs.components()` and `lhsType.components()` have different sizes.
I confirmed this by debug-printing out the sizes:
```
Warning: lhsType size: 2; _lhs size: 1; _rhs size: 2.
 --> test.sol:2:5:
  |
2 |     ((, ())) = (1, 2);
  |
```

I pushed a more proper (?) fix.

Changelog.md
```diff
@@ -14,6 +14,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix out of range error for nested tuples.
```
cameel: I think this should be a bit more specific. I'd need to take a closer look to give a proper suggestion but just at a quick glance, as a user, this does not tell me when I can run into this. this isn't caused by nesting alone, right? Does it specifically require an assignment or can it happen in other cases?

Also I'd say "compiler crash" rather than "out of range error". The specific type of an exception and even the fact that it's a C++ exception is just an unimportant implementation detail here in my opinion.

timweri: Right. I'll think of a better way to phrase it. But yes it's really about nesting outer unary tuples during assignment.

ekpyron (dismissed):

Changelog.md
```diff
@@ -14,6 +14,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix compiler crashing due to out-of-range error when left-hand side tuples are nested in outer unary tuple during tuple assignment.
```
matheusaaguiar:
```suggestion
 * Type Checker: Fix compiler crash when left-hand side tuples are nested in outer unary tuple during tuple assignment.
```
I think that we could remove the reason of the compiler crash since it is cpp related and not useful for end-user. 
Also, it seems that the problem does not happen when the nested tuple is the first element:
```
    // Ok
    function f() public pure {
        ((((, ())), )) = (1, 3);
    }
    // out of range crash
    function g() public pure {
        ((, ((, )))) = (1, 3);
    }
```
However, if we add an empty tuple to `function f()` above, we get the crash again:
```
     // Not ok anymore
     function f() public pure {
        ((((, ())), ())) = (1, 3);
    }
```
So, I am not sure if this is strictly related to tuples nested in outer unary tuple... I guess we could investigate it a little bit more to understand the exact situation which triggers the crash.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -112,8 +112,9 @@ void TypeChecker::checkDoubleStorageAssignment(Assignment const& _assignment)
 	size_t storageByteAccesses = 0;
 	auto count = [&](TupleExpression const& _lhs, TupleType const& _rhs, auto _recurse) -> void {
 		TupleType const& lhsType = dynamic_cast<TupleType const&>(*type(_lhs));
+		TupleExpression const* lhs = dynamic_cast<TupleExpression const*>(resolveOuterUnaryTuples(&_lhs));
```
matheusaaguiar: I think this var could have a more distinguishable name from the parameter var. What about `lhsResolved` or something like that?

test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
```diff
@@ -0,0 +1,12 @@
+function f() {
```
matheusaaguiar:
```suggestion
function f() public pure {
```

test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
```diff
@@ -0,0 +1,12 @@
...
+    ((, ())) = (1, 3);
+}
+
+function g() {
```
matheusaaguiar:
```suggestion
function g() public pure {
```

matheusaaguiar:
We could also add a test where the rhs tuple is returned from a function:
```
function g() public pure returns (int, int) {
        return (4, 5);
    }
function f() public pure {
        ((, ())) = g();
}

```

Changelog.md
timweri: I'll amend the changelog.

Here's my initial understanding of the cases you pointed out:
When the nested tuple is in the first element in the case below, the crash is normally due to the call `_lhs.components().at(index)`. Here, `index = 1`. This doesn't cause a crash because `_lhs.components()` here actually has . This is a lucky coincidence. 

But the element `index = 0` refers to is not `_lhs.components().at(0)` but rather `lhsType.components().at(0)`. When we retrieve the type, the extra meaningless nested unary tuples are already removed so these two elements are not the same. Recursively, this coincidence will peel away the outer unary tuples. This crash only reveals itself when there's a mismatch between the sizes of `_lhs` and `lhsType`.
```solidity
function f() public pure {
    ((((, ())), )) = (1, 3);
}
```

timweri: Seems like my answer isn't quite correct.

This is not fine:
```solidity
function f() pure {
    ((, ())) = (1,3);
}
```
But this is fine:
```solidity
function f() pure {
    ((,, ())) = (1,3);
}
```
I'll look into it more

ekpyron: I think the gist is that the bug could be triggered by any recursive unary tuple, not just an "outer" one (it does happen during a recursion into the tuples after all).

timweri: I think I understood it correctly. The reason why the below works is because it failed the size check at 

https://github.com/ethereum/solidity/blob/d5a78b18b3fd9e54b2839e9685127c6cdbddf614/libsolidity/analysis/TypeChecker.cpp#L116

```solidity
// This fails because lhsType.size() == _rhs.size()
function f() pure {
    ((, ())) = (1,3);
}

// This is ok because the size check caught size mismatch before the crash bug
// lhsType.size() == 3
// _rhs.size() == 2
function f() pure {
    ((,, ())) = (1,3);
}

// if we guarantee the sizes are the same, then crash happens
function f() pure {
    ((,, ())) = (1,3,4);
}
```

timweri: > I think the gist is that the bug could be triggered by any recursive unary tuple, not just an "outer" one (it does happen during a recursion into the tuples after all).

Right. I agree.

libsolidity/analysis/TypeChecker.cpp
timweri: Done

test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
timweri: I only added `pure`. Can't add `public` to free functions.

test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
timweri: I only added `pure`. Can't add `public` to free functions.

Changelog.md
timweri: I'm trying to understand why this happens:
```solidity
// works
function f() pure {
    ((((, ())) , )) = ((1, 2), 3);
}

// if the outer unary tuple is taken out, it crashes again
// this is the opposite of the simple cases where adding outer unary tuple causes the crash
function g() pure {
    (((, ())) , ) = ((1, 2), 3);
}
```

For the first case:
- Recursion 1: `index = 0` is accessed
   - `lhsExpression`: `((((, ())) , )) (size = 1)` becomes `(((, ())) , ) (size = 2)`
   - `lhsType`: `tuple(tuple(,tuple()),) (size = 2)` becomes `tuple(,tuple()) (size = 2)`
   - `rhsType`: `tuple(tuple(int_const 1,int_const 2),int_const 3) (size = 2)` becomes `tuple(int_const 1,int_const 2) (size = 2)`
- Recursion 2: `index=1` is accessed
   - `lhsExpression`:  `(((, ())) , ) (size = 2)` becomes `((, ())) (size = 1)`
   - `lhsType`: `tuple(,tuple()) (size = 2)` becomes `tuple() (size = 0)`
   - `rhsType`: `tuple(int_const 1,int_const 2) (size = 2)` becomes `int_const 2`
- Recursion terminates (no crash)

For the second case:
- Recursion 1: `index = 0` is accessed
   - `lhsExpression`: `(((, ())) , ) (size = 2)` becomes `((, ())) (size = 1)`
   - `lhsType`: `tuple(tuple(,tuple()),) (size = 2)` becomes `tuple(,tuple()) (size = 2)`
   - `rhsType`: `tuple(tuple(int_const 1,int_const 2),int_const 3) (size = 2)` becomes `tuple(int_const 1,int_const 2) (size = 2)`
- Recursion 2: `index = 1` is accessed. But `lhsExpression` has `size = 1` => crash.

timweri: So far the fix has addressed all the cases. But yea I changed the phrasing of the changelog to be more accurate.

ekpyron:
The CI failures should (hopefully) be fixed, if you rebase the PR to current ``develop``.

test/libsolidity/syntaxTests/tupleAssignments/empty_tuple_assignment.sol
matheusaaguiar: Oh yeah, sorry about that, i just copied the text from an open file I had at the time.

ekpyron:
[FYI you can ignore the ``t_native_test_ext_zeppelin`` CI failure - it's unrelated to the changes in this PR and will hopefully be fixed on ``develop`` soon]

Changelog.md
```diff
@@ -13,6 +13,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix compiler crashing because of nested tuple assignment.
```
matheusaaguiar:
```suggestion
 * Type Checker: Fix compiler crash on tuple assignments involving certain patterns with unary tuples on the left-hand side.
```
Don't want to be pedantic, but just so we can be more specific and precise here.
As discussed in the team chat and suggested by @ekpyron.
Also, a rebase on develop should fix external tests and then PR is good to go.

timweri: Haha don't worry. I do prefer the new changelog description.

Fixed and rebased!

ekpyron (approved):

PR #13280, opened 2022-07-17
Fix NULL dereference in abi.encodeCall check

### LINKED ISSUE #13266: `abi.encodeCall()` segfaults in a free function when passed an error/event
```
error Error(string);
error Panic(uint);

function panic(uint code)
{
  bytes memory data = abi.encodeCall(Panic, code);
  assembly "evmasm" ("memory-safe") {
    revert(data, mload(data))
  }
}
```

The compiler segfaults...

timweri:
closes #13266

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Compiler Features:
 
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
+ * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
 
 
 ### 0.8.15 (2022-06-15)
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -2195,14 +2195,28 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	)
 	{
 		string msg = "Expected regular external function type, or external view on public function.";
-		if (externalFunctionType->kind() == FunctionType::Kind::Internal)
-			msg += " Provided internal function.";
-		else if (externalFunctionType->kind() == FunctionType::Kind::DelegateCall)
-			msg += " Cannot use library functions for abi.encodeCall.";
-		else if (externalFunctionType->kind() == FunctionType::Kind::Creation)
-			msg += " Provided creation function.";
-		else
-			msg += " Cannot use special function.";
+
+		switch (externalFunctionType->kind())
+		{
+			case FunctionType::Kind::Internal:
+				msg += " Provided internal function.";
+				break;
+			case FunctionType::Kind::DelegateCall:
+				msg += " Cannot use library functions for abi.encodeCall.";
+				break;
+			case FunctionType::Kind::Creation:
+				msg += " Provided creation function.";
+				break;
+			case FunctionType::Kind::Event:
+				msg += " Cannot use events for abi.encodeCall.";
+				break;
+			case FunctionType::Kind::Error:
+				msg += " Cannot use errors for abi.encodeCall.";
+				break;
+			default:
+				msg += " Cannot use special function.";
+		}
+
 		SecondarySourceLocation ssl{};
 
 		if (externalFunctionType->hasDeclaration())
@@ -2213,10 +2227,14 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 				externalFunctionType->declaration().scope() == m_currentContract
 			)
 				msg += " Did you forget to prefix \"this.\"?";
-			else if (util::contains(
-				m_currentContract->annotation().linearizedBaseContracts,
-				externalFunctionType->declaration().scope()
-			) && externalFunctionType->declaration().scope() != m_currentContract)
+			else if (
+				m_currentContract &&
+				externalFunctionType->declaration().scope() != m_currentContract &&
+				util::contains(
+					m_currentContract->annotation().linearizedBaseContracts,
+					externalFunctionType->declaration().scope()
+				)
+			)
 				msg += " Functions from base contracts have to be external.";
 		}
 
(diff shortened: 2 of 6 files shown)


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2213,7 +2213,7 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 				externalFunctionType->declaration().scope() == m_currentContract
 			)
 				msg += " Did you forget to prefix \"this.\"?";
-			else if (util::contains(
+			else if (m_currentContract && util::contains(
 				m_currentContract->annotation().linearizedBaseContracts,
 				externalFunctionType->declaration().scope()
 			) && externalFunctionType->declaration().scope() != m_currentContract)
```
```suggestion
			else if (
				m_currentContract &&
				externalFunctionType->declaration().scope() != m_currentContract &&
				util::contains(
					m_currentContract->annotation().linearizedBaseContracts,
					externalFunctionType->declaration().scope()
				)
			)
```

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_error_free_function.sol
```diff
@@ -0,0 +1,7 @@
+error E(uint);
+
+function f() {
+    abi.encodeCall(E, (1));
+}
+// ----
+// TypeError 3509: (50-51): Expected regular external function type, or external view on public function. Cannot use special function.
```
cameel: I think we should have a separate message for errors and events. Not sure if technically we can classify them as "special functions" or not, but I think saying event/error would be much clearer for users anyway. Looks like it should be easy to change in `TypeChecker.cpp`, just above the change you already made.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2213,7 +2213,7 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 				externalFunctionType->declaration().scope() == m_currentContract
 			)
 				msg += " Did you forget to prefix \"this.\"?";
```
cameel, line 2229: From the implementation, looks like there should be more affected cases and indeed I found some. Could you add these as tests?

```solidity
function g(uint) {}

function f() {
    abi.encodeCall(g, (1));
}
```

```solidity
library L {
    function f() external {}
}

function f() {
    abi.encodeCall(L.f, (1));
}
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2213,7 +2213,7 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 				externalFunctionType->declaration().scope() == m_currentContract
```
cameel, line 2227: Can you check if `scope()` returns `nullptr` for free functions? If it does, we also need a check against `m_currentContract` being null here. When it's a free function, the compiler should not suggest adding `this.`

I think this part does not show up in syntax tests so please also test this by running compiled `solc` locally on one of the test cases. You should see the secondary message that way.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_error_free_function.sol
timweri: I added separate messages for events and errors. Let me know what you think.

libsolidity/analysis/TypeChecker.cpp
timweri: I did a `dynamic_cast` check and the `scope()` of a free function is not `nullptr`. It is the `SourceUnit`. Also, from my understanding, free functions would not pass the `visibility()` check above because free functions are not `public`.

libsolidity/analysis/TypeChecker.cpp
timweri: Confirmed that these cases are also affected. Added tests.

Marenz (dismissed):

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2195,14 +2195,28 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
+				msg += " Cannot use library functions for abi.encodeCall.";
+				break;
+			case FunctionType::Kind::Creation:
+				msg += " Provided creation function.";
+				break;
+			case FunctionType::Kind::Event:
+				msg += " Cannot use event function.";
+				break;
+			case FunctionType::Kind::Error:
+				msg += " Cannot use error function.";
```
cameel: Still needs a small tweak.
```suggestion
			case FunctionType::Kind::Event:
				msg += " Cannot use events for abi.encodeCall.";
				break;
			case FunctionType::Kind::Error:
				msg += " Cannot use errors for abi.encodeCall.";
```

libsolidity/analysis/TypeChecker.cpp
cameel: True. They'll fail the first check. Good.

cameel (commented):
I think the messages should not refer to errors and events as functions. Other than that everything seems fine.

timweri:
Fixed the error messages :)

cameel (dismissed):
You still need to update test output or tests will fail.

But anyway, I'm approving already. As soon as it's fixed, it's good to go.

timweri:
My bad. I updated the tests.

Marenz (dismissed):

timweri:
Seems like the CI is failing for something unrelated to my change. Other PRs are also failing the same test.

cameel:
Yeah, that looks like a false-positive unrelated to the PR. We'll need to disable these tests.

chriseth (dismissed):

chriseth:
Actually this is missing a changelog entry - could you add it please?

timweri:
@chriseth Added changelog

Changelog.md
```diff
@@ -13,6 +13,7 @@ Compiler Features:
 
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
+ * TypeChecker: Fix null dereference in `abi.encodeCall` type checking of free function.
```
```suggestion
 * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
```
(I know it is written like that above, but this is how it should be :) )

timweri: Fixed :)

chriseth (approved):

chriseth:
Thanks a lot!

PR #13282, opened 2022-07-19
Fix wrong cleanup when copying from calldata to memory.

### LINKED ISSUE #13249: Bad encoding for a dynamic array followed by a static calldata array
## Description

Inconsistent abi encode behavior with tuples containing both a dynamic array and fixed length array.

## Steps to Reproduce

```js
pragma solidity >=0.7.0 <0.9.0;

contract EncodeTest {
   
    struct Test {
        uint[] vals;
        uint[2] valsFixed;
    }

    function testEncode(Test calldata input) public pure returns (bytes memory) {
        return abi.encode(input);
    }

    function testDecode(bytes calldata input) public pure returns (Test memory) {
        return abi.decode(input, (Test));
    }
}
```

When calling `testEncode` with the following argument: `{ vals: [1, 2, 3], valsFixed: [1, 1] }` the following is received (split into chunks of 64 hex characters for readability): 
```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000000 // <- should be 0x03?
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000002
0000000000000000000000000000000000000000000000000000000000000003
```

When abi encoding using web3 or ethers the following is received:

```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000003
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000002
0000000000000000000000000000000000000000000000000000000000000003
```

<summary>web3/ethers code</summary>
<details>

```js
const { ethers } = require('ethers')
const Web3 = require('web3')

console.log(
  ethers.utils.defaultAbiCoder.encode(
    [`tuple(uint[] vals, uint[2] valFixed)`],
    [{vals: [1,2,3],valFixed: [1,1]}]
  )
)

const web3 = new Web3()
console.log(
  web3.eth.abi.encodeParameter(
    'tuple(uint[] vals, uint[2] valFixed)',
    [[1, 2, 3], [1, 1]]
  )
)
```

</details>

--

Furthermore, when using the `testDecode` function the first output does not decode the `vals` array, while the second output does. Is this expected or known behavior of `abi.encode`?

Marenz:
fixes #13249


Changelog.md
```diff
@@ -13,6 +13,7 @@ Compiler Features:
 
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
+ * Code Generation: Fix faulty cleanup for some situations when abi-encoding static calldata arrays together with other types.
```
cameel: It's a codegen issue destroying data. We will probably have to classify it as an important bugfix. What's your opinion on that?

If it's an important bug we'll also need a bug list entry, a blog post and all.

cameel: Also, I think the description should state that data is getting overwritten and lost. Currently it's not clear if it's something like that or just some unused space not being cleaned up (which would be much less serious).

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -73,8 +73,9 @@ class YulUtilFunctions
 
 	/// @returns a function that copies raw bytes of dynamic length from calldata
 	/// or memory to memory.
-	/// Pads with zeros and might write more than exactly length.
-	std::string copyToMemoryFunction(bool _fromCalldata);
+	/// @params _cleanup: If true, pads with zeros and might write more than
+	///         the exact length, else only the exact length will be written
+	std::string copyToMemoryFunction(bool _fromCalldata, bool _cleanup = true);
```
cameel: I would remove the default. `true` is dangerous if you're not careful. Better if you have to make that decision explicitly. I would also give the parameter name that does not sound that harmless :)
```suggestion
	std::string copyToMemoryFunction(bool _fromCalldata, bool _unsafeCleanup);
```

As a side effect it would also let us see every use of this function in the review.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -89,11 +89,14 @@ string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
 			return Whiskers(R"(
 				function <functionName>(src, dst, length) {
 					calldatacopy(dst, src, length)
+					<?cleanup>
 					// clear end
 					mstore(add(dst, length), 0)
+					</cleanup>
```
cameel: > else only the exact length will be written

Isn't this "else" part missing here?

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract C {
+    function f(uint256[] memory a, uint256[1] calldata b) public returns (bytes memory) {
+        return abi.encode(a, b);
+    }
+
+    function g(uint256[] memory a, uint256[1] calldata b) external returns (bytes memory) {
+        return f(a, b);
+    }
+}
```
cameel: I think we should cover more situations.

And we should also make sure that every use of `copyToMemoryFunction()` is covered by some test.

cameel (commented):
Just a quick initial pass. I haven't carefully checked all uses of `copyToMemoryFunction()` and that's a crucial part here.

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
Marenz: > And we should also make sure that every use of copyToMemoryFunction() is covered by some test.

Why, if this case is the only one that uses it without cleanup?
The rest wasn't changed..

libsolidity/codegen/YulUtilFunctions.cpp
Marenz: What else would you want to happen in the `else` part, other than not cleaning the end?

cameel: Ah, wait, I think I read too much into the docstring. I think it's correct and I was just thinking about the thing I suggested on the chat - i.e. that we should clean the exact number of bytes instead of not cleaning at all.

But yeah, I need to look at the usage to see if we actually need this. Maybe no cleanup is what we need after all...

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
cameel: Because the cleanup is what causes the issue. The other uses could be overwriting stuff too.

Marenz: I investigated the other usages of memcopy and couldn't find any scenario where its cleaning might cause problems.

We could add more tests that use different types of dyn and static arrays, but other than that I can't think of more variations to test here...

And I doubt they add much, either they take the same codepath and use the fixed memcopy or they use a completely different encoding and thus we're not testing this problem..

Changelog.md
Marenz: fair. And yes, I'd say it is an important bug

cameel: We usually add a section called `Important Bugfixes` for these. See how it looks for 0.8.15.

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -73,8 +73,10 @@ class YulUtilFunctions
 
 	/// @returns a function that copies raw bytes of dynamic length from calldata
 	/// or memory to memory.
-	/// Pads with zeros and might write more than exactly length.
-	std::string copyToMemoryFunction(bool _fromCalldata);
+	/// @params _unsafeCleanup: If true, pads with zeros and might write more
+	///         than the exact length, else only up until the exact length will
+	///         be written
```
cameel: Actually, does it ever **not** write more than necessary? I think that even if the length is a multiple of 32, it does the cleanup even though it's not necessary.

```suggestion
	/// @params _unsafeCleanup: If true, overwrites exactly 32 bytes past the end of the array with zeros.
	///         WARNING: The cleaned area always extends beyond the padding resulting from 32-byte alignment.
	///         If you use this value, you must ensure on your own that the data past the end of the array is
	///         preserved or is fine to overwrite.
```

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -498,7 +498,7 @@ string ABIFunctions::abiEncodingFunctionCalldataArrayWithoutCleanup(
 				);
 			templ("readableTypeNameFrom", _from.toString(true));
 			templ("readableTypeNameTo", _to.toString(true));
-			templ("copyFun", m_utils.copyToMemoryFunction(true));
+			templ("copyFun", m_utils.copyToMemoryFunction(true, false /*no cleanup*/));
```
cameel: I'm not sure if the `WithoutCleanup` part in the name really means there should be no cleanup past the array. At a face value this is what it seems to imply but after looking at its docstring I'm not so sure:
```c++
/// Part of @a abiEncodingFunction for array target type and given calldata array.
/// Uses calldatacopy and does not perform cleanup or validation and can therefore only
/// be used for byte arrays and arrays with the base type uint256 or bytes32.
std::string abiEncodingFunctionCalldataArrayWithoutCleanup(
```

It implies that the function is appropriate only for types that do not require cleanup. `uint` and `bytes32` arrays obviously don't since they fill whole slots. But it also mentions byte arrays. These do not have padding between elements but do have some at the end. `ABIFunctions::abiEncodingFunction()` which calls this function does not clean the padding past the end so I think that the function is only expected to skip the padding between elements but is still responsible for cleaning the padding at the end.

cameel: In any case, we're definitely missing coverage for `string` and `bytes` cleanup. Can you add some? I think this change changes compiler's behavior with respect to that padding and if we had these tests they would be failing now.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -80,7 +80,7 @@ string YulUtilFunctions::splitExternalFunctionIdFunction()
 	});
 }
 
-string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
+string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata, bool _cleanup)
```
cameel, line 83:
```suggestion
string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata, bool _unsafeCleanup)
```

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
```diff
@@ -0,0 +1,17 @@
+pragma abicoder               v2;
```
cameel:
```suggestion
pragma abicoder v2;
```

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -662,7 +662,7 @@ string ABIFunctions::abiEncodingFunctionMemoryByteArray(
 		templ("functionName", functionName);
 		templ("lengthFun", m_utils.arrayLengthFunction(_from));
 		templ("storeLength", arrayStoreLengthForEncodingFunction(_to, _options));
-		templ("copyFun", m_utils.copyToMemoryFunction(false));
+		templ("copyFun", m_utils.copyToMemoryFunction(false, true /* cleanup */));
```
cameel: I originally wanted to comment that this should be causing the same problem as the one in `abiEncodingFunctionCalldataArrayWithoutCleanup()`. I thought that something like this should trigger it:
```solidity
pragma abicoder v2;

contract C {
    function f(bytes memory a, bytes memory b) public returns (bytes memory) {
        return abi.encode(a, b);
    }
}
// ====
// EVMVersion: >homestead
// ----
// f(bytes,bytes): 0x40, 0x80, 32, "abcdefghijklmnopabcdefghijklmnop", 32, "abcdefghijklmnopabcdefghijklmnop" -> 0x20, 0xc0, 0x40, 0x80, 32, "abcdefghijklmnopabcdefghijklmnop", 32, "abcdefghijklmnopabcdefghijklmnop"
```

The cleanup of the first array should overwrite the size field of the second one. But it doesn't.

I think this is because the problem exists only in one specific corner case: when copying a static array that's the last element in the head area. The cleanup is perfectly safe if you only have dynamic arrays because they're copied in order and we can never write over bytes belonging to the previous array. Same if you have only static ones. But when you mix them, the cleanup past the last static array crosses the border between the head and the tail and overwrites the first slot of the tail.

`string` and `bytes` are always dynamic, so if I understood the problem correctly, this code can never overwrite anything after all. At least in the context of decoding. I think that the fact that it can write past the array should be listed as a warning in its docstring in case someone tries to reuse it for something else in the future.

I'd still add that test case though. And here's also another, similar but slightly more complex one:
```solidity
pragma abicoder v2;

struct S {
    bytes1 b;
    string s;
}

contract C {
    function f(S memory a, bytes memory b, S memory c) public returns (bytes memory) {
        return abi.encode(a, b, c);
    }
}
// ====
// EVMVersion: >homestead
// ----
// f((bytes1,string),bytes,(bytes1,string)): 0x60, 0xe0, 0x120, "x", 0x40, 4, "abcd", 4, "pqrs", "y", 0x40, 4, "efgh" -> 0x20, 0x1a0, 0x60, 0xe0, 0x120, "x", 0x40, 4, "abcd", 4, "pqrs", "y", 0x40, 4, "efgh"
```

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -1296,7 +1296,7 @@ string ABIFunctions::abiDecodingFunctionByteArrayAvailableLength(ArrayType const
 		templ("functionName", functionName);
 		templ("allocate", m_utils.allocationFunction());
 		templ("allocationSize", m_utils.arrayAllocationSizeFunction(_type));
-		templ("copyToMemFun", m_utils.copyToMemoryFunction(!_fromMemory));
+		templ("copyToMemFun", m_utils.copyToMemoryFunction(!_fromMemory, true /* cleanup */));
```
cameel: This one seems safe only because ABI decoding always allocates a new memory array for the decoded data **and** it allocates the next array only after the copying (and cleaning) of the previous one is already done.

I think this is not an obvious assumption and should be documented. It should at least be in the docstring but maybe adding a comment about it at the point of use would be a good idea too.

libsolidity/codegen/ABIFunctions.cpp
```diff
@@ -514,7 +514,7 @@ string ABIFunctions::abiEncodingFunctionCalldataArrayWithoutCleanup(
 			templ("functionName", functionName);
 			templ("readableTypeNameFrom", _from.toString(true));
 			templ("readableTypeNameTo", _to.toString(true));
-			templ("copyFun", m_utils.copyToMemoryFunction(true));
+			templ("copyFun", m_utils.copyToMemoryFunction(true, false /*no cleanup*/));
```
cameel: After looking at the 4 uses of this function, I think this is the only problematic one. This is the one that kicks in when you copy the static array. The other one above (for dynamic arrays) could probably be changed back to `true`. Or at least conditionally, just for byte arrays.

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
cameel: > We could add more tests that use different types of dyn and static arrays, but other than that I can't think of more variations to test here...

Some ideas:
- A case without an explicit `abi.encode()` call. The problem happens even when you simply return values from an external function.
- A case where the arrays are nested inside a struct (or even deeper). I think the issue happens at the border between the head and the tail at any nesting level.
    - Also a positive case: function taking multiple structs containing static arrays of `uint`. Each copy will overwrite stuff beyond its struct but it should not hurt as longs as the structs are copied in order - and they should be if they're all static.
- A library function call. These have slightly different ABI. I'm not even 100% sure if they go through the same set of functions of if we have some separate ones.
- Emitting an event or reverting with an error. These do encoding too. And again, they probably reuse the same functions but technically they don't have to and they're distinct enough that I think they should be covered separately.
- A case where something other than the size of a dynamic array gets overwritten. I think that you should be able to overwrite anything if you put it at the beginning of a struct that also contains a dynamic array.
- A case where the static calldata array is not the last argument. I think it must be the last one to cause this and would be nice to have a non-failing case confirming that.

> And I doubt they add much, either they take the same codepath and use the fixed memcopy or they use a completely different encoding and thus we're not testing this problem.

I think they do, and that's especially if they take a slightly different code path. The fact that this went undetected means that our tests are too weak. For example, as I wrote in another comment, looks like the string cleanup is not being tested at all. Each time we're fixing a bug like this, we should check which related cases are not covered and cover them.

And I mean cases that turn out not to be broken as well. These are still useful to make sure this stuff will not break in the future. At least those distinct enough from others. We don't need hundreds of variations of the same thing but I have the feeling that here even the straightforward ones are often not covered. Here are some things I'd test here:
- Array elements other than `uint`. Especially smaller than one slot, where there is something to clean.
- More arguments in the function. Especially dynamic and static ones in various orders.
- Different combinations of locations (`calldata` or `memory`).

docs/bugs.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "uid": "SOL-2022-6",
+        "name": "CorruptABIEncodingWithYul",
+        "summary": "Encoding a static calldata array followed by a dynamic array results in a corrupt length value for the dynamic array.",
+        "description": "When ABI-encoding a static calldata array, we used to clean the last 32 byte slot after the array data by overwriting it with ``0``. This is problematic if the data of the dynamic array was already written there. This is the case when encoding a dynamic array followed by a static calldata array. Then the length of the dynamic array is overwritten.",
+        "introduced": "0.5.8",
+        "fixed": "0.8.16",
+        "severity": "low"
```
cameel: I think the severity might be higher. We need to think which plausible use cases it could affect.

docs/bugs.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "uid": "SOL-2022-6",
+        "name": "CorruptABIEncodingWithYul",
```
cameel: How about this:
```suggestion
        "name": "ABIEncoderV2HeadOverflowWithStaticArrayCleanup",
```
It's kinda like a buffer overflow :)

 Also it's reproducible without Yul, isn't it?

docs/bugs.json
```diff
@@ -1,4 +1,13 @@
...
+        "summary": "Encoding a static calldata array followed by a dynamic array results in a corrupt length value for the dynamic array.",
+        "description": "When ABI-encoding a static calldata array, we used to clean the last 32 byte slot after the array data by overwriting it with ``0``. This is problematic if the data of the dynamic array was already written there. This is the case when encoding a dynamic array followed by a static calldata array. Then the length of the dynamic array is overwritten.",
+        "introduced": "0.5.8",
+        "fixed": "0.8.16",
+        "severity": "low"
+    },
```
cameel, line 13: Maybe we should add this?
```json
        "conditions": {
            "ABIEncoderV2": true
        }
```

Can it be reproduced with v1? I think anything dynamic requires v2 but I'm not sure.

And if it doesn't, we should have a test case that runs with v1 and fails.

docs/bugs.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "uid": "SOL-2022-6",
+        "name": "CorruptABIEncodingWithYul",
+        "summary": "Encoding a static calldata array followed by a dynamic array results in a corrupt length value for the dynamic array.",
+        "description": "When ABI-encoding a static calldata array, we used to clean the last 32 byte slot after the array data by overwriting it with ``0``. This is problematic if the data of the dynamic array was already written there. This is the case when encoding a dynamic array followed by a static calldata array. Then the length of the dynamic array is overwritten.",
```
cameel: I think that the problem is broader than just arrays. Here's how I would describe it after analyzing this PR:
```suggestion
        "summary": "ABI-encoding a tuple with a statically-sized calldata array in the last component would corrupt 32 leading bytes of its first dynamically-sized component.",
        "description": "When ABI-encoding a statically-sized calldata array, the compiler always pads the data area to a multiple of 32-bytes and ensures that the padding bytes are zeroed. In some cases, this cleanup used to be performed by always writing exactly 32 bytes, regardless of how many needed to be zeroed. This was done with the assumption that the data that would eventually occupy the area past the end of the array had not yet been written, because the encoder processes tuple components in the order they were given. While this assumption is mostly true, there is an important corner case: dynamically-sized tuple components are stored separately from the statically-sized ones in an area called the *tail* of the encoding and the tail immediately follows the *head*, which is where the statically-sized components are placed. The aforementioned cleanup, if performed for the last component of the head would cross into the tail and overwrite up to 32 bytes of the first component stored there with zeros. The only array type for which the cleanup could actually result in an overwrite were arrays with ``uint256`` or ``bytes32`` as the base element type and in this case the size of the corrupted area was always exactly 32 bytes. The problem affected tuples at any nesting level. This included also structs, which are encoded as tuples in the ABI. Note also that lists of parameters and return values of functions, events and errors are encoded as tuples."
```

docs/bugs.json
Marenz: I felt it was `low` because I thought: The worst effect is that the length of the array that was sent will be 0. The sender would just send corrupt data like that in the first place, so we're not really worse off than before in terms of "terrible things the sender can do"

cameel: ok, you have a point. But I think there's still potential for accidental data loss even if not malicious. Sending empty array when it was meant to contain stuff can still easily break things. We still include that in the severity assessment, right?

Changelog.md
```diff
@@ -1,5 +1,9 @@
 ### 0.8.16 (unreleased)
 
+Important Bugfixes:
+ * Code Generation: Fix a data corruption: When abi-encoding a dynamic array followed by a static calldata array the length of the dynamic array would get overwritten by ``0``.
```
chriseth: Can we make this more specific? Why has it not occurred earlier?

chriseth:
Thanks for the fix, this seems to have been rather tricky!
Is it an option to - instead of skipping the cleanup altogether - do a non-destructive cleanup, i.e. we first read, zero out some bits and then write again?

Marenz:
> Is it an option to - instead of skipping the cleanup altogether - do a non-destructive cleanup, i.e. we first read, zero out some bits and then write again?

I don't see the use of any clean up in this context to be honest. We are simply writing `0`s to the 32 bytes after copying the static calldata array.. I don't see why we do that or what benefit we have from that...

cameel:
We don't need cleanup for arrays of `uint256` and `bytes32` but for `bytes` and `string` I think we still do. And this function is handling both cases. Can you add the cleanup tests I mentioned in https://github.com/ethereum/solidity/pull/13282#discussion_r926969959? They should show that after the change byte arrays are no longer cleaned.

Marenz:
> They should show that after the change byte arrays are no longer cleaned.

I don't see why that should be the case. The calls in the `byte` related functions still use the `true` parameter for keeping the old behavior. 

And the calls that have changed behavior (only found in `string ABIFunctions::abiEncodingFunctionCalldataArrayWithoutCleanup(` mind you) have an assert for the correct size you mentioned in the beginning:

```
	solAssert(fromArrayType.location() == DataLocation::CallData, "");
	solAssert(
		fromArrayType.isByteArrayOrString() ||
		*fromArrayType.baseType() == *TypeProvider::uint256() ||
		*fromArrayType.baseType() == FixedBytesType(32),
		""
	);
```

Given this, I don't think I can construct any tests that would show it's not cleaned after the copy.. because nothing changed in that regard.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -89,11 +89,17 @@ string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
...
 				function <functionName>(src, dst, length) {
 					calldatacopy(dst, src, length)
-					// clear end
-					mstore(add(dst, length), 0)
+					// clear up to 32 byte boundary
+					let endPos := add(dst, length)
+					if mod(endPos, 0x20)
+					{
+						<zeroMemory>(endPos, sub(<roundUp>(endPos), endPos))
+					}
```
cameel: I was thinking about something like this (untested):
```yul
let spill := mod(endPos, 0x20)
if spill
{
    let mask := shr(not(0), sub(0x20, spill))
    mstore(endPos, and(mload(endPos), mask))
}
```
This is a single slot so using the zeroing function seems a bit heavyweight. You can instead load the current value of the slot, and clear the final bytes using a mask (all ones, shifted right by the amount of bytes we want to clear).

Also, I wonder if the optimizer is smart enough to replace `mod(endPos, 0x20)` with `and(endPos, 0x1f)` automatically. `mod` is more expensive than `and` (5 vs 3 gas) but also probably more readable so if the optimizer can do it, we might still want to use it.

Marenz: This is what the optimizer did:

```
function copy_calldata_to_memory(src, dst, length)
{
    calldatacopy(dst, src, length)
    let endPos := add(dst, length)
    if addmod(dst, length, 0x20)
    {
        calldatacopy(endPos, calldatasize(), sub(and(add(endPos, 31), not(31)), endPos))
    }
}
```

Changelog.md
Marenz: It existed for a while, my guess is the combination just wasn't used enough that people reported the problem?

I rephrased it to 
` * Code Generation: Fix a data corruption: When abi-encoding a dynamic type followed by a static calldata array, the cleanup of the static array would mistakenly zero out the first slot of the of tail-part of the dynamic type that follows it.`
now

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -73,7 +73,8 @@ class YulUtilFunctions
 
 	/// @returns a function that copies raw bytes of dynamic length from calldata
 	/// or memory to memory.
-	/// Pads with zeros and might write more than exactly length.
+	/// Pads with zeros up to the 32 byte boundary after the specified length
```
cameel:
```suggestion
	/// Pads with zeros up to the 32 byte boundary after the specified length.
```

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -89,11 +89,18 @@ string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
...
-					// clear end
-					mstore(add(dst, length), 0)
+					// clear up to 32 byte boundary
+					let endPos := add(dst, length)
+					let spill := and(endPos, 0x1F)
+					if spill
+					{
+						let mask := <shl>(not(0), sub(0x20, spill))
+						mstore(endPos, and(mload(endPos), mask))
+					}
```
cameel: I think that the `_unsafeCleanup` parameter (well, it's safe now so just `_cleanup` :)) was actually useful because this cleanup is not always needed. We only need it for the byte array case in `abiEncodingFunctionCalldataArrayWithoutCleanup()` (the one I mentioned in https://github.com/ethereum/solidity/pull/13282#discussion_r927056701). This would save us some gas.

BTW, if you do this, then the parameter should be included in the generated function name just like the `_fromCalldata` parameter. I think your previous version of this code might have had this problem too (not sure, can't check). I guess that the function with unnecessary type in its name you found the other day actually proved useful - I probably wouldn't even realize this otherwise :)

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
```diff
@@ -0,0 +1,34 @@
+pragma abicoder v2;
+
+contract C {
+    struct Data {
+        uint256[] a;
+        string b;
+    }
+
+    function f(Data calldata data, uint256[] calldata a) public pure returns (bytes memory) {
+        // Make memory dirty (numbers come from trial & err)
```
cameel:
```suggestion
        // Make memory dirty (numbers come from trial & error)
```

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
```diff
@@ -0,0 +1,34 @@
...
+        for (uint pos = 0x80; pos < 0x880; pos += 0x20)
```
cameel: I think it should be enough to clear between `mload(0x40)` (i.e. the free mem ptr) and `add(mload(0x40), 0x800)`.

That would be more future-proof, and would not break if we start reserving more memory at the beginning. I guess it could break even today if you happen to have some vars that need moving to memory.

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
```diff
@@ -0,0 +1,34 @@
...
+            assembly {
+                mstore(pos, not(0))
+            }
+
+        return abi.encode(data, a);
+    }
+
+    function g(Data memory data, uint256[] calldata a) public pure returns (bytes memory) {
+        // Make memory dirty for string (0x126 is where the string ends)
+        for (uint pos = 0x126; pos < 0x880; pos += 0x20)
```
cameel: Isn't this function redundant? Looks like the other one makes everything dirty and this one only after the string. But we're only interested in what happens after the string, so both will break if cleanup is broken and both will work if cleanup works.

So unless I'm missing something, I'd be for removing `g()` from this test, especially because it requires more specific hard-coding (you depend on the exact location of the string here). If you do decide to keep this, I think you should at least add an assert that positions 120-126 contain `"123456"`. This way it won't break silently.

Marenz: `g()` uses a memory tuple/struct so this one is testing the copying from memory to memory (where as the other one is for the calldata -> memory variant) 

The "dirty-making"  is so precise here because (I suspect) the memory is already filled before we want to make it dirty with the `data` parameter. When I used different numbers it just gave me a FAILED return..
I am actually not super confident that the dirty-ing is working correct here. If I start at 0x120 I get:

```
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,20]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,60]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,40]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,0]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,40]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,80]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,ff]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,6]
  [ff,ff,ff,ff,ff,ff,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0] // the ff's here should be the 123456 array.
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,2]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,af]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,bf]

```

Changelog.md
cameel: It's even more specific. Based on what I suggested earlier in `bugs.json` I'd describe it like this:
`* Code Generation: Fix corruption of the leading 32 bytes of the first dynamically-sized component of an ABI-encoded tuple when the last component contained a statically-sized calldata array.`

Or maybe this - more verbose but probably easier for users to figure out in what circumstances it happens:
`* Code Generation: Fix data corruption that affected ABI-encoding of calldata values represented by tuples: structs at any nesting level; argument lists of external functions, events and errors; return value lists of external functions. Up to 32 leading bytes of the first dynamically-sized value in the tuple would get zeroed when the last component contained a statically-sized array.`

libsolidity/codegen/YulUtilFunctions.cpp
cameel: ok, it used `addmod(dst, length, 0x20)` so basically same thing as `mod(endPos, 0x20)` (maybe saves some stack operations under the hood though). Apparently it does not optimize it.

Yeah, let's use `and(endPos, 0x1f)` explicitly.

Marenz: I feel it's even worse, because it's doing the add operation twice now (`endPos := add(dst, length)` and `addmod(dst, length, 0x20)`).. maybe we can add a rule for that or fix the rule that did that?

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
cameel: Ah, right. Missed the fact that it's using a `memory` arg.

> The "dirty-making" is so precise here because (I suspect) the memory is already filled before we want to make it dirty with the data parameter.

Hmm... in that case maybe it would be better to make it a local `memory` variable initialized in the function? Then you could do the dirtying before the variable declaration and be less precise.

In any case an `assert` would be a good validation that you found the right spot in memory.

libsolidity/codegen/YulUtilFunctions.cpp
cameel: Yeah. That too.

But it does not look like adjusting rules would fix it. It looks more like the optimizer should have reused a common expression but it could not because `add(dst, length)` went into the `addmod`.

We could make it not convert `add` + `mod` to `addmod` but I suspect it's not a good idea in a general case even if it would help here :)

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
Marenz: I couldn't go lower than 0x5E until the tests started to respond with "FAILURE"

docs/bugs.json
Marenz: I added a test case for v1, the problem couldn't be reproduced. Maybe that's why it didn't show up for so long?

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
Marenz: Ooh, I completely misunderstood you :laughing:

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -80,20 +80,35 @@ string YulUtilFunctions::splitExternalFunctionIdFunction()
...
-					mstore(add(dst, length), 0)
+					<?cleanup>
+					// clear up to 32 byte boundary
+					let endPos := add(dst, length)
+					let spill := and(endPos, 0x1F)
+					if spill
+					{
+						let mask := <shl>(not(0), sub(0x20, spill))
+						mstore(endPos, and(mload(endPos), mask))
+					}
```
ekpyron: Out of curiosity: Have you checked if something fanciful and branchless like
```suggestion
					// clear up to 32 byte boundary
					calldatacopy(add(dst, length), calldatasize(), and(sub(32, length), 31))
```
is cheaper?
I'd imagine this could tip either way in the ``spill = 0`` case and the branchless one should be cheaper otherwise, but I'm not sure and it's probably easier to just try :-).

ekpyron: [Resp. from our discussion in chat we might want to refert back to the original ``mstore(add(dst, length), 0)`` anyways]

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -105,14 +110,16 @@ string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
 					{
 						mstore(add(dst, i), mload(add(src, i)))
 					}
+					<?cleanup>
 					if gt(i, length)
 					{
```
ekpyron: Do we actually still need this check here?

Changelog.md
```diff
@@ -1,5 +1,8 @@
 ### 0.8.16 (unreleased)
 
+Important Bugfixes:
+ * Code Generation: Fix data corruption that affected ABI-encoding of calldata values represented by tuples: structs at any nesting level; argument lists of external functions, events and errors; return value lists of external functions. Up to 32 leading bytes of the first dynamically-sized value in the tuple would get zeroed when the last component contained a statically-sized array.
```
cameel: In the end I think we determined it's always exactly 32 bytes:

```suggestion
 * Code Generation: Fix data corruption that affected ABI-encoding of calldata values represented by tuples: structs at any nesting level; argument lists of external functions, events and errors; return value lists of external functions. 32 leading bytes of the first dynamically-sized value in the tuple would get zeroed when the last component contained a statically-sized array.
```

docs/bugs.json
cameel: Might explain < 0.8.0, but for later versions v2 is the default so I'd expect it to be reported earlier.

Maybe it's rather the fact that people rarely encode stuff directly from calldata and most of the time modifying it first in some way?

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -80,20 +80,25 @@ string YulUtilFunctions::splitExternalFunctionIdFunction()
...
 
-string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
+string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata, bool _cleanup)
 {
-	string functionName = "copy_" + string(_fromCalldata ? "calldata" : "memory") + "_to_memory";
+	string functionName =
+		"copy_" +
+		string(_fromCalldata ? "calldata" : "memory") +
+		"_to_memory" +
+		string(_cleanup ? "_cleanup" : "");
```
cameel:
```suggestion
		string(_cleanup ? "_with_cleanup" : "");
```

cameel:
```suggestion
		"copy_" +
		_fromCalldata ? "calldata"s : "memory"s +
		"_to_memory" +
		_cleanup ? "_cleanup"s : ""s;
```

libsolidity/codegen/YulUtilFunctions.cpp
cameel: I think we don't. When `length == i` cleanup is not necessary but should not be harmful.

test/libsolidity/gasTests/abiv2.sol
```diff
@@ -14,9 +14,9 @@ contract C {
 }
 // ----
 // creation:
-//   codeDepositCost: 1243000
-//   executionCost: 1295
-//   totalCost: 1244295
+//   codeDepositCost: 1256600
+//   executionCost: 1308
+//   totalCost: 1257908
```
cameel: I guess these are just not up to date? With the current version the cost should only decrease.

test/libsolidity/semanticTests/abiEncoderV1/memory_dynamic_array_and_calldata_bytes.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract C {
+    function f(uint256[] memory a, bytes calldata b) public returns (bytes memory) {
+        return abi.encode(a, b);
+    }
+
+    function g(uint256[] memory a, bytes calldata b) external returns (bytes memory) {
+        return f(a, b);
+    }
+}
```
cameel, line 11: > I added a test case for v1, the problem couldn't be reproduced. Maybe that's why it didn't show up for so long?

If you meant this case, then I I think it does not reproduce it simply because the static array is not the last argument.

This case actually makes it overwrite stuff:
```solidity
pragma abicoder v1;

contract C {
    function f(bool a, bytes calldata b, bytes32[2] calldata c)
        public
        returns (bool, bytes calldata, bytes32[2] calldata)
    {
        return (a, b, c);
    }
}
// ----
// f(bool,bytes,bytes32[2]): true, 0x80, "a", "b", 4, "abcd" -> true, 0x80, "a", "b", 4, "abcd"
```
```
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,80]
  [61,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
  [62,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
  [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
  [61,62,63,64,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```

@ekpyron Since these are dynamic types, is it actually going through v1 or is it just that v2 gets force-enabled in some way?

test/libsolidity/semanticTests/abiEncoderV1/memory_dynamic_array_and_calldata_bytes.sol
```diff
@@ -0,0 +1,17 @@
...
+
+// ====
+// EVMVersion: >homestead
```
cameel, line 14: With v1, does it actually require homestead?

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
```diff
@@ -0,0 +1,71 @@
...
+    {
+        makeDirty();
+
+        Data memory data;
+        data.a = new uint256[](1);
+        data.a[0] = 0xFF;
+        data.b = "123456";
+        uint256[] memory array = new uint256[](2);
+        array[0] = 0xAF;
+        array[1] = 0xBF;
```
cameel: Now that the memory and calldata cases are so similar, you could put this part in a helper function to make the test shorter.

test/libsolidity/semanticTests/abiEncoderV1/memory_dynamic_array_and_calldata_bytes.sol
ekpyron: abiencoder v1 does have dynamic types, just not nested ones, so this should work with v1.
From a quick look at it, running this through ``isoltest`` will still try to run it via-IR (which implies abiencoder v2) and that's what's failing - adding ``compileViaYul: false`` (i.e. having it really *only* run with abiencoder v1) has it pass.

Changelog.md
```suggestion
 * Code Generation: Fix data corruption that affected ABI-encoding of calldata values represented by tuples: structs at any nesting level; argument lists of external functions, events and errors; return value lists of external functions. Up to 32 leading bytes of the first dynamically encoded value in the tuple would get zeroed when the last component contained a statically-sized array.
```
In changelog and bug list we did use that phrasing before and it's clearer that way that also fixed-size-arrrays of dynamically-sized arrays are affected.

docs/bugs.json
```diff
@@ -1,4 +1,16 @@
...
...his cleanup used to be performed by always writing exactly 32 bytes, regardless of how many needed to be zeroed. This was done with the assumption that the data that would eventually occupy the area past the end of the array had not yet been written, because the encoder processes tuple components in the order they were given. While this assumption is mostly true, there is an important corner case: dynamically-sized tuple components are stored separately from the statically-sized ones in an area called the *tail* of the encoding and the tail immediately follows the *head*, which is where the statically-sized components are placed. The aforementioned cleanup, if performed for the last component of the head, would cross into the tail and overwrite exactly to 32 bytes of the first component stored there with zeros. The only array type for which the cleanup could actually result in an overwrite were arrays with ``uint256`` or ``bytes32`` as the base element type. The problem affected tuples at any nesting level. This included also structs, which are encoded as tuples in the ABI. Note also that lists of parameters and return values of functions, events and errors are encoded as tuples.",
```
ekpyron:
```suggestion
        "summary": "ABI-encoding a tuple with a statically-sized calldata array in the last component would corrupt 32 leading bytes of its first dynamically encoded component.",
        "description": "When ABI-encoding a statically-sized calldata array, the compiler always pads the data area to a multiple of 32-bytes and ensures that the padding bytes are zeroed. In some cases, this cleanup used to be performed by always writing exactly 32 bytes, regardless of how many needed to be zeroed. This was done with the assumption that the data that would eventually occupy the area past the end of the array had not yet been written, because the encoder processes tuple components in the order they were given. While this assumption is mostly true, there is an important corner case: dynamically encoded tuple components are stored separately from the statically-sized ones in an area called the *tail* of the encoding and the tail immediately follows the *head*, which is where the statically-sized components are placed. The aforementioned cleanup, if performed for the last component of the head, would cross into the tail and overwrite exactly to 32 bytes of the first component stored there with zeros. The only array type for which the cleanup could actually result in an overwrite were arrays with ``uint256`` or ``bytes32`` as the base element type. The problem affected tuples at any nesting level. This included also structs, which are encoded as tuples in the ABI. Note also that lists of parameters and return values of functions, events and errors are encoded as tuples.",
```

ekpyron: [FYI for me for some reason in the description github fails to highlight the change from ``dynamically sized`` to ``dynamically encoded``]

test/libsolidity/semanticTests/abiEncoderV1/memory_dynamic_array_and_calldata_bytes.sol
cameel: Right. Thanks.

In that case we should add this as a use case for v1 (with the extra `compileViaYul: false`) but fortunately v1 looks unaffected.

test/libsolidity/gasTests/abiv2.sol
Marenz: Oh true, I forgot the gastests

ekpyron (dismissed):
We could try to be even more nitpicky and precise about changelog entry and buglist entry (wrt. the base types and nested static arrays, etc.), but I'd be ok with it as is (pending a confirmation from @cameel ;-)).

docs/bugs.json
```diff
@@ -1,4 +1,16 @@
...
+        "introduced": "0.5.8",
+        "fixed": "0.8.16",
+        "severity": "low/medium",
```
cameel: I used "medium" in the blog post.
```suggestion
        "severity": "medium",
```

Marenz: fixed

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -80,20 +80,25 @@ string YulUtilFunctions::splitExternalFunctionIdFunction()
...
 
-string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
+string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata, bool _with_cleanup)
 {
-	string functionName = "copy_" + string(_fromCalldata ? "calldata" : "memory") + "_to_memory";
+	string functionName =
+		"copy_"s +
+		(_fromCalldata ? "calldata"s : "memory"s) +
+		"_to_memory"s +
+		(_with_cleanup ? "_cleanup"s : ""s);
```
cameel: Haha, My [previous suggestion](https://github.com/ethereum/solidity/pull/13282#discussion_r937004620) was about the function name, not the parameter :)
```suggestion
		(_with_cleanup ? "_with_cleanup"s : ""s);
```
Since we're not releasing today after all, we could still change that but on the other hand it's not such a big deal... I just thought that the name would be more readable that way.

libsolidity/codegen/YulUtilFunctions.h
```diff
@@ -73,8 +73,9 @@ class YulUtilFunctions
 
 	/// @returns a function that copies raw bytes of dynamic length from calldata
 	/// or memory to memory.
-	/// Pads with zeros and might write more than exactly length.
-	std::string copyToMemoryFunction(bool _fromCalldata);
+	/// @params _with_cleanup If true, pads with zeros up to the 32 byte boundary after the specified length
+	/// signature: (src, dst, length) ->
+	std::string copyToMemoryFunction(bool _fromCalldata, bool _with_cleanup);
```
cameel:
```suggestion
	std::string copyToMemoryFunction(bool _fromCalldata, bool _withCleanup);
```

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    function calldataTest(Data calldata data, uint256[] calldata a) public pure returns (bytes memory)
+    {
+        return abi.encode(data, a);
+    }
+
+    function memoryTest(Data memory data, uint256[] calldata a) public pure returns (bytes memory)
+    {
+        return abi.encode(data, a);
+    }
```
cameel: Bad news. This test no longer detects the missing cleanup :(

Just tested it locally by replacing `bytesOrString` in `ABIFunctions.cpp` with `false` and the test still passes. Isn't this because these functions are external so the dirtying does not affect what happens inside them?

ekpyron: With ``bytesOrString == false`` the new version *still* won't perform the old cleanup - that way you call it as if it was a ``uint[]`` or ``bytes32[]`` array, for which we generally removed the cleanup. That's the reason, isn't it? Or did I get it wrong what you mean :-)?

cameel (dismissed):
I'm tentatively approving since none of the stuff I found is a blocker for the release, but I think that the cleanup test is no longer effective.

Also I think we should have at least some of the tests I listed in https://github.com/ethereum/solidity/pull/13282#discussion_r927088574. For example we never tested if this really affects errors, events or library calls, I'm just assuming it must since it's probably using the same encoder code... This could be done in a separate PR, for now these could go into an issue.

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
ekpyron: Yeah, I think I misunderstood, let me have a closer look... (didn't really look at this test so far to be honest)

cameel: > With `bytesOrString == false` the new version still won't perform the old cleanup

That's the point. It's a separate thing from the bug. One of the early versions of this PR was basically equivalent to `bytesOrString == false`, which removes the cleanup in the `bytes` and `string` case. This means that byte arrays were not being cleaned. I noticed that and asked for a test covering that - because nothing failed which means we were missing coverage.

cameel:
Also, this needs a rebase now that the OZ fix has been merged.

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
cameel: For reference, here's my original comment where I explained that: https://github.com/ethereum/solidity/pull/13282#discussion_r926963789

ekpyron: The test gives me a headache :-). If it's just the cleanup when abi-encoding a string into previously dirty memory, this should be easier IMO:
```
contract C {
  function f(string calldata x) external returns (bytes memory r) {
        uint mptr;
        assembly {
                // dirty memory
                mptr := mload(0x40)
                for { let i := mptr } lt(i, add(mptr, 0x0100)) { i := add(i, 32) }
                {
                    mstore(i, sub(0,1))
                }
        }
        r = abi.encode(x);
        assembly {
                // assert that we dirtied the memory that was encoded to
                if iszero(eq(mptr, r)) { revert(0,0) }
        }
  }
  function test() external returns (bytes memory) {
        return this.f("abc");
  }
}
// ----
// test() -> 0x20, 0x60, 0x20, 3, "abc"
```

cameel:
Rebased to make the external tests pass.

test/libsolidity/semanticTests/abiEncoderV2/dirty_memory_dynamic_array_and_calldata_string.sol
cameel: I pushed a commit that adds @ekpyron's test and verified that it's effective. Not sure what to do about the existing one. If it's still broken, probably better to remove it.

test/libsolidity/semanticTests/abiEncoderV2/calldata_struct_dynamic_and_static.sol
cameel: I just pushed a commit that adds some of the more important tests from the list above: for errors, events and library functions - now I can confirm that all were affected. Also the more complex case from the blog post.

Also added some small tweaks to existing tests - renaming (they no longer contain structs), no homestead and abiv2 required (should still pass in all of these cases). Also added a function that returns the parameters directly, without explicit encoding.

Left the modifications unsquashed so that they're easy to review.

docs/bugs.json
```diff
@@ -1,4 +1,16 @@
 [
+    {
+        "uid": "SOL-2022-6",
+        "name": "ABIEncoderV2HeadOverflowWithStaticArrayCleanup",
```
ekpyron: Not actually sure, we'd have needed to emphasize "ABIEncoderV2" that much, given that it's long been the default now. But also fine.

ekpyron (dismissed):
Looks good to me now!
Not sure about https://github.com/ethereum/solidity/pull/13282/files#r938013916, but doesn't really hurt to keep in any case.

libsolidity/codegen/YulUtilFunctions.cpp
```diff
@@ -105,14 +110,11 @@ string YulUtilFunctions::copyToMemoryFunction(bool _fromCalldata)
 					{
 						mstore(add(dst, i), mload(add(src, i)))
 					}
-					if gt(i, length)
-					{
-						// clear end
-						mstore(add(dst, length), 0)
-					}
+					<?cleanup>mstore(add(dst, length), 0)</cleanup>
```
hrkrshnn, line 113: This line always made me uncomfortable: it ends up writing beyond what it's supposed to. It should only be managing memory from `[dst, dst + length)` at least from the signature of the function, but it ends up creeping into `[dst + length, dst + length + 32)` now.

ekpyron: The plan for after the release is to see if we can't just inline the function into all of its use cases (it's a weird helper function anyways and it only has four use cases, in which it should be apparent that the overwrite is fine).

docs/bugs.json
```diff
@@ -1,4 +1,16 @@
 [
+    {
+        "uid": "SOL-2022-6",
+        "name": "AbiReencodingHeadOverflowWithStaticArrayCleanup",
```
hrkrshnn, line 4:
```suggestion
        "name": "AbiReEncodingHeadOverflowWithStaticArrayCleanup",
```
Since the question of typos came up, shouldn't it be like this? The name is quite a mouthful :)

libsolidity/codegen/YulUtilFunctions.cpp
cameel: And if it stays, it and everything that uses it transitively should have scary comments and docstring pointing out the potential risk.

docs/bugs.json
ekpyron: ``"SOL-2022-2"`` was ``NestedCallataArrayAbiReencodingSizeValidation"``, that's where I took the ``AbiReencoding`` from... that being said, that's exactly the one with the typo, which we're going to change anyways, so...

cameel: Depends if you insist on putting a hyphen there or not :)
We used the same form in `NestedCallataArrayAbiReencodingSizeValidation` so I think I'd keep it as is for consistency.

> The name is quite a mouthful :)

Most of them are. Hard to keep them short and unambiguous. Unless we want to start doing the bug branding thing. How about "headbleed" :)

hrkrshnn: Auction off bug names? FTXArenaBug

cameel (approved):

PR #13283, opened 2022-07-19
Yul Debugger

Marenz:
This PR extends yulRunner with an interactive debugger interface.

This is not yet cleaned up, but the code is in a working state and was used for debugging


chriseth:
Is the yul mode for remix working for this case?

Marenz:
> Is the yul mode for remix working for this case?

I don't about any yul mode for remix and couldn't find any info after some short googling?

cameel:
There's [a dropdown that lets you select Yul](https://remix-ide.readthedocs.io/en/latest/compile.html#solidity-or-yul) in the compiler tab.

test/tools/yulInterpreter/Inspector.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace
+{
+
+void printVariable(YulString const& _name, u256 const& _value)
+{
+	cout << "\t" << _name.str() << " = " << _value.str();
+
+	if (_value != 0)
+		cout << " (" << toCompactHexWithPrefix(_value) << ")";
```
nikola-matic, line 41: nit: if the value is not zero, this will print two spaces after the equals sign.

test/tools/yulInterpreter/Inspector.cpp
```diff
@@ -0,0 +1,165 @@
...
+			<< "# ";
+
+		cout.flush();
+
+		getline(cin, input);
+		boost::algorithm::trim(input);
+
+		// Imitate GDB and repeat last cmd for empty string input.
+		if (input == "")
+			input = m_lastInput;
```
nikola-matic, line 91:
```suggestion
			input = m_lastInput;
		else
		    m_lastInput = input;
```

`m_lastInput` assignment could then be removed from following if/else if branches, no?

test/tools/yulInterpreter/Inspector.cpp
```diff
@@ -0,0 +1,165 @@
...
+		else if (input == "variables" || input == "v")
+		{
+			m_lastInput = input;
+			for (auto &&[yulStr, val]: _variables)
+				printVariable(yulStr, val);
+			cout << endl;
+		}
+		else if (
+			boost::starts_with(input, "print") ||
+			boost::starts_with(input, "p")
```
nikola-matic, line 109: Would `potato x` be valid input?

test/tools/yulInterpreter/Inspector.cpp
Marenz: That's not correct as we always output `_vaule.str()`. It looks like:
```
# p _7
_7 = 3 (0x03)
```

nikola-matic: Ah, you're right, my bad, didn't see `_value.str()` at the end.

test/tools/yulInterpreter/Inspector.cpp
Marenz: Good spot!

test/tools/yulInterpreter/Inspector.cpp
Marenz: Basically yeah. I didn't do the validation super strict here as I consider it more a dev-tool not a user-front-end tool.
In this case it would just interpret it as `print`...
I can make it more strict if we think this is desired here...

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +418,80 @@ void ExpressionEvaluator::incrementStep()
...
+	if (values()[1] != util::h160::Arith(m_state.address))
+		return;
+
+	Scope tmpScope;
+	InterpreterState tmpState;
+	tmpState.calldata = m_state.calldata;
+	tmpState.callvalue = callvalue;
+
+	// Create new interpreter for the called contract
+	shared_ptr<Interpreter> newInterpreter = makeInterpreterNew(tmpState, tmpScope);
```
nikola-matic: Can this be a `unique_ptr`?

test/tools/yulInterpreter/Interpreter.h
```diff
@@ -219,7 +232,30 @@ class ExpressionEvaluator: public ASTWalker
...
+			m_dialect,
+			m_scope,
+			m_enableExternalCalls,
+			m_disableMemoryTrace,
+			std::move(_variables)
+		);
+	}
+	virtual std::shared_ptr<Interpreter> makeInterpreterNew(InterpreterState& _state, Scope& _scope) const
+	{
+		return std::make_shared<Interpreter>(
```
nikola-matic: I guess not :laughing:

Marenz: I think I actually tried a unique pointer first :)

libevmasm/Instruction.h
```diff
@@ -188,6 +188,15 @@ enum class Instruction: uint8_t
 	SELFDESTRUCT = 0xff	///< halt execution and register account for later deletion
 };
 
+/// @returns true if the instruction is of the CALL opcode family
+inline bool isCallInstruction(Instruction _inst)
```
christianparpart:
```suggestion
constexpr bool isCallInstruction(Instruction _inst) noexcept
```

libevmasm/Instruction.h
```diff
@@ -188,6 +188,15 @@ enum class Instruction: uint8_t
...
+{
+	return Instruction::CALL == _inst ||
+		Instruction::CALLCODE == _inst ||
+		Instruction::DELEGATECALL == _inst ||
+		Instruction::STATICCALL == _inst;
```
christianparpart: Why "const COMP var" and not "var COMP const" and why not using `switch` statement instead?

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -72,17 +72,45 @@ void InterpreterState::dumpTraceAndState(ostream& _out, bool _disableMemoryTrace
...
+
+		_out << endl;
+	}
 }
 
 void Interpreter::run(
 	InterpreterState& _state,
 	Dialect const& _dialect,
 	Block const& _ast,
+	bool _enableExternalCalls,
```
christianparpart: This might lead to confusion. One boolean parameter asks for enabling the other right next for disabling.
Generally speaking, probably best would be to create `enum class`es both both of them and pass the intended value then explicitly. Should be much more readable.

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -288,7 +316,17 @@ void ExpressionEvaluator::operator()(FunctionCall const& _funCall)
 		if (BuiltinFunctionForEVM const* fun = dialect->builtin(_funCall.functionName.name))
 		{
 			EVMInstructionInterpreter interpreter(m_state, m_disableMemoryTrace);
-			setValue(interpreter.evalBuiltin(*fun, _funCall.arguments, values()));
+
+			u256 value = interpreter.evalBuiltin(*fun, _funCall.arguments, values());
```
christianparpart:
```suggestion
			u256 const value = interpreter.evalBuiltin(*fun, _funCall.arguments, values());
```

test/tools/yulInterpreter/Interpreter.h
```diff
@@ -219,7 +232,30 @@ class ExpressionEvaluator: public ASTWalker
 	/// Returns the list of values of the expression.
 	std::vector<u256> values() const { return m_values; }
 
-private:
+protected:
+	void runExternalCall(evmasm::Instruction _instruction);
+	virtual std::shared_ptr<Interpreter> makeInterpreterCopy(std::map<YulString, u256> _variables = {}) const
```
christianparpart: Why is this `shared_ptr` and not `unique_ptr`? Do we have multiple equally-relevant owners to the same resource, or can we rewrite that into a single ownership and use `unique_ptr` instead?

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +418,80 @@ void ExpressionEvaluator::incrementStep()
...
+
+	// Find file scope
+	while (abstractRootScope->parent)
+	{
+		fileScope = abstractRootScope;
+		abstractRootScope = abstractRootScope->parent;
+	}
+
+	// Get AST for file scope
+	for (auto &&[block, scope]: abstractRootScope->subScopes)
```
christianparpart:
```suggestion
	for (auto&& [block, scope]: abstractRootScope->subScopes)
```

christianparpart (dismissed):

libevmasm/Instruction.h
Marenz: Because like that not speaking me :D only Yoda.

Yah, I know the advantage, so we can't accidentally assign to var if we forget a `=`, but honestly, it reads so un-intuitive and shitty the const-first way in the majority of cases...

test/tools/yulInterpreter/Interpreter.cpp
Marenz: Probably, but I never intended this to be ready-for-production-user-facing code, so I didn't want to spent too much time on decorative details

test/libyul/yulInterpreterTests/external_call.yul
```diff
@@ -8,3 +8,5 @@
 // Memory dump:
 // Storage dump:
 //   0000000000000000000000000000000000000000000000000000000000000064: 0000000000000000000000000000000000000000000000000000000000000001
+// Calldata dump:
+//      0: 0000000000000000000000000000000000000000000000000000000000000000
```
aarlt: Just scrolled through this. Why is the call data here `0000000000000000000000000000000000000000000000000000000000000000`? Shouldn't it somehow reflect `0, 0x20, 0x30, 0x20`?

aarlt: ah ok, just read about the semantic of call. so you are providing 0x20 bytes from address 0, right? and this is just zero, because there are just zeroes, right?

aarlt: Wouldn't be a very similar test nice, with actually providing call data?

test/tools/yulInterpreter/Interpreter.h
```diff
@@ -47,6 +49,10 @@ class ExplicitlyTerminated: public InterpreterTerminatedGeneric
 {
 };
 
+class ExplicitlyTerminatedWithReturn: public ExplicitlyTerminated
```
aarlt, line 52: Somehow I'm not able to see where this is actually thrown.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -322,23 +326,24 @@ u256 EVMInstructionInterpreter::eval(
...
 	case Instruction::RETURN:
 	{
-		bytes data;
+		m_state.returndata = {};
 		if (accessMemory(arg[0], arg[1]))
-			data = readMemory(arg[0], arg[1]);
-		logTrace(_instruction, arg, data);
+			m_state.returndata = readMemory(arg[0], arg[1]);
+		logTrace(_instruction, arg, m_state.returndata);
 		BOOST_THROW_EXCEPTION(ExplicitlyTerminated());
```
aarlt: Not sure, but shouldn't this now be `ExplicitlyTerminatedWithReturn()`?

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +418,80 @@ void ExpressionEvaluator::incrementStep()
...
+			break;
+		}
+
+	yulAssert(ast);
+
+	try
+	{
+		(*newInterpreter)(*ast);
+	}
+	catch (ExplicitlyTerminatedWithReturn const&)
```
aarlt, line 496: This seem to be not thrown anywhere.

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +418,80 @@ void ExpressionEvaluator::incrementStep()
...
+	{
+		fileScope = abstractRootScope;
+		abstractRootScope = abstractRootScope->parent;
+	}
+
+	// Get AST for file scope
+	for (auto&& [block, scope]: abstractRootScope->subScopes)
+		if (scope.get() == fileScope)
+		{
+			ast = block;
```
aarlt, line 486: If I understand correctly, `ast` will be set to the block that is that will be interpreted (called), right? But somehow I don't understand how you exactly find that block that is called. Maybe it would be helpful - at least for me - to add an additional comment how you exactly find the block that need to get called.

aarlt: It reads like that you're iterating through all current scopes and if the scope is the file scope, you set the `ast`. Somehow I don't see that the found block, is really just the block of the function that need to be called. For me it looks like that it could also be another block. The problem is that I cannot understand the code correctly I guess.

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +418,80 @@ void ExpressionEvaluator::incrementStep()
...
+	)
+	{
+		memOutOffset = values()[4];
+		memOutSize = values()[5];
+	}
+	else
+		yulAssert(false);
+
+	// Don't execute external call if it isn't our own address
+	if (values()[1] != util::h160::Arith(m_state.address))
```
aarlt, line 457: wouldn't it be nice to exactly support this case?

test/tools/yulInterpreter/Interpreter.cpp
aarlt: Or is it better just to rename it to `_disableExternalCalls` - and of course also update its semantic.

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -288,7 +316,17 @@ void ExpressionEvaluator::operator()(FunctionCall const& _funCall)
 		if (BuiltinFunctionForEVM const* fun = dialect->builtin(_funCall.functionName.name))
 		{
 			EVMInstructionInterpreter interpreter(m_state, m_disableMemoryTrace);
-			setValue(interpreter.evalBuiltin(*fun, _funCall.arguments, values()));
+
+			u256 const value = interpreter.evalBuiltin(*fun, _funCall.arguments, values());
```
aarlt, line 321: If `fun` is an external function here, isn't it then already evaluated? E.g. didn't it ran already?

test/tools/yulrun.cpp
```diff
@@ -153,7 +161,7 @@ Allowed options)",
 		else
 			input = readUntilEnd(cin);
 
-		interpret(input);
+		interpret(input, arguments.count("interactive"), arguments.count("enable-external-calls"));
```
aarlt: maybe "interactive" and "enable-external-calls" could be extracted into a variable and used here and within the options some lines above.

test/tools/yulrun.cpp
```diff
@@ -74,7 +75,7 @@ pair<shared_ptr<Block>, shared_ptr<AsmAnalysisInfo>> parse(string const& _source
 	}
 }
 
-void interpret(string const& _source)
+void interpret(string const& _source, bool _inspect, bool _enableExternalCalls)
```
aarlt: What is actually happening if `_enableExternalCalls` is not set?

test/tools/yulInterpreter/Interpreter.h
```diff
@@ -243,6 +279,7 @@ class ExpressionEvaluator: public ASTWalker
 	std::vector<u256> m_values;
 	/// Current expression nesting level
 	unsigned m_nestingLevel = 0;
+	bool m_enableExternalCalls;
```
aarlt: As stated above, maybe its better to either have always "enable*" or always "disable*"

test/tools/yulInterpreter/Interpreter.cpp
Marenz: It couldn't be external, this branch is only entered for builtin functions...

test/tools/yulInterpreter/Interpreter.cpp
Marenz: Well, of course, but that needs a lot more work. Right now we can only call the same contract again externally.. which is enough for most debugging problems though..

test/tools/yulInterpreter/Interpreter.cpp
Marenz: There are two steps here:

First we find the scope that equals our whole file. This scope can be found directly under the "virtual" root scope which is just there to collect all the other scopes.
We can't access it directly, however as we only have our current scope, so we search upwards until we find the very end (when parent == nullptr), then our scope must be the root scope and the one before must be our file scope.

Once we have our file scope, we want the actual already parsed AST which is only available in a mapping from block -> scope, so we have to manually iterate through that mapping to find our AST.

In yul, the highest-level AST for files is of type block, that's why I called it that way here

test/tools/yulInterpreter/Interpreter.cpp
Marenz: Which is really weird, maybe it got lost while I was rebasing/rearranging commits, good find!

test/tools/yulrun.cpp
Marenz: Any external call with simply return with no data

Marenz:
While I appreciate all the comments and feedback, I also feel that all of you forget that this is not part of the compiler and I think it shouldn't be held to the same standard, simply because it's not worth the time and effort here. 

It's just a tool for us, after all.
I wrote this because I needed it and it provides value, but i am not sure we should invest so much time to make this "perfect"

test/tools/yulInterpreter/Interpreter.cpp
aarlt: Ah yes! True!

test/tools/yulrun.cpp
Marenz: hmm I just followed the existing pattern in the file, none of the parameters are handled that way there.. yet

ekpyron:
Pinging @bshastry about this one, just to confirm that the interpreter changes won't interfere with fuzzing (since most of it is conditional behind the flag, it shouldn't affect much, but still).

test/tools/yulInterpreter/Interpreter.h
```diff
@@ -243,6 +279,7 @@ class ExpressionEvaluator: public ASTWalker
 	std::vector<u256> m_values;
 	/// Current expression nesting level
 	unsigned m_nestingLevel = 0;
+	bool m_disableExternalCalls;
```
bshastry, line 296: Could you please add a comment as to what the new flag does?

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -322,24 +326,25 @@ u256 EVMInstructionInterpreter::eval(
 		return (0xdddddd + arg[1]) & u256("0xffffffffffffffffffffffffffffffffffffffff");
 	case Instruction::CALL:
 	case Instruction::CALLCODE:
-		// TODO assign returndata
 		accessMemory(arg[3], arg[4]);
 		accessMemory(arg[5], arg[6]);
+		m_state.calldata = readMemory(arg[3], arg[4]);
```
bshastry: Could we do this only if address is equal to self? Otherwise, we are simply populating call data without actually executing the call, and may be this gives the impression that the call took place although it actually didn't?

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -72,17 +72,45 @@ void InterpreterState::dumpTraceAndState(ostream& _out, bool _disableMemoryTrace
 	}
 	_out << "Storage dump:" << endl;
 	dumpStorage(_out);
+
+	if (!calldata.empty())
+	{
+		_out << "Calldata dump:";
+
+		for (size_t offset = 0; offset < calldata.size(); ++offset)
```
bshastry, line 80: Could you please dump only non-zero values like this?

https://github.com/ethereum/solidity/blob/f7fcd8ccaac2dadb14761198e980305de00d2baa/test/tools/yulInterpreter/Interpreter.cpp#L66-L71

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -322,24 +326,25 @@ u256 EVMInstructionInterpreter::eval(
...
 		logTrace(_instruction, arg);
 		return arg[0] & 1;
 	case Instruction::DELEGATECALL:
 	case Instruction::STATICCALL:
 		accessMemory(arg[2], arg[3]);
 		accessMemory(arg[4], arg[5]);
+		m_state.calldata = readMemory(arg[2], arg[3]);
```
bshastry: Same comment as above.

bshastry (commented):
Added a few comments for consistency/conciseness.

Mainly missing test cases for successful and unexecuted external calls via call/callcode/staticcall/externalcall

test/libyul/yulInterpreterTests/external_call_to_self.yul
```diff
@@ -0,0 +1,21 @@
+{
+	mstore(0x40, 1)
```
ekpyron: If you add something like
```suggestion
	mstore(0x40, 0x42)
	if calldataload(0) { sstore(0x0200, calldataload(0)) }

```
here, we'd have proof that the nested call actually works and gets the proper calldata... (as in the nested call should write 0x42 to storage slot 0x0200)

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -322,24 +326,23 @@ u256 EVMInstructionInterpreter::eval(
 		return (0xdddddd + arg[1]) & u256("0xffffffffffffffffffffffffffffffffffffffff");
 	case Instruction::CALL:
 	case Instruction::CALLCODE:
-		// TODO assign returndata
 		accessMemory(arg[3], arg[4]);
 		accessMemory(arg[5], arg[6]);
 		logTrace(_instruction, arg);
-		return arg[0] & 1;
+		return arg[0] > 0 ? 1 : 0;
```
ekpyron: I'm not entirely sure what the reasoning behind the old version here was...
But pinging @bshastry again about this: was this meant to basically simulate external calls pseudo-randomly failing or succeeding based on whether the passed in gas is odd or even?

We need to be careful about whether changing this makes fuzzing loose code paths after external calls.

test/tools/yulInterpreter/Interpreter.cpp
```diff
@@ -368,3 +419,89 @@ void ExpressionEvaluator::incrementStep()
...
+	catch (ExplicitlyTerminatedWithReturn const&)
+	{
+		// Copy return data to our memory
+		copyZeroExtended(
+			m_state.memory,
+			newInterpreter->returnData(),
+			memOutOffset.convert_to<size_t>(),
+			0,
+			memOutSize.convert_to<size_t>()
+		);
```
ekpyron, line 505: I guess, strictly speaking we should also set ``m_state.returndata = newInterpreter->returnData();`` here - that way we could also implement ``returndatasize`` and ``returndatacopy`` based on ``m_state.returndata``... but maybe it's fine skipping that for now.

ekpyron (commented):
Some more minor comments, but I'd generally be fine with this, considering it an internal testing tool to-be-extended-as-needed.
But before merging, I'd like to have another confirmation about the previous return values of the ``CALL`` instructions from @bshastry (see the respective review comment).

bshastry:
> Some more minor comments, but I'd generally be fine with this, considering it an internal testing tool to-be-extended-as-needed.
> But before merging, I'd like to have another confirmation about the previous return values of the ``CALL`` instructions from @bshastry (see the respective review comment).

I didn't quite follow. What do you mean by the previous return values of ``Call`` instructions

Marenz:
> I didn't quite follow. What do you mean by the previous return values of Call instructions

Well, the old code would return `gas() & 1` so it would randomly fail or succeed. 
We assumed it is deliberate but didn't know the exact motivation for it.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
bshastry: I don't remember doing this myself i.e., pseudo random return for call based on evenness of `arg[0]`

Since earlier we did not interpret call instructions, this might have been a way to simulate both success and failure events.

I don't quite understand the change. What is the semantics of the change proposed by the PR?

ekpyron (approved):
There's still a lot of stuff that could be done here (e.g. ``runExternalCall`` should return an exit status, overriding the hard-coded return value of the call instruction based on the evaluation of the nested call; returndatasize and returndatacopy could be handled correctly; too-deep call recursions could be graceful instead of exceptions; etc. pp) - and if we addressed more of those we could actually consider trying to set the flag to ``true`` for fuzzing...

But since we established that the current version should not interfere with fuzzing and apart from that it's so far mainly meant as a crude tool to help in investigating bugs, and we already spent more time on this than planned, I'm fine with merging.

PR #13286, opened 2022-07-21
Modify the metadata / hash document.

LCamel:


docs/metadata.rst
```diff
@@ -21,9 +21,11 @@ the :ref:`Standard JSON Interface<compiler-api>`.
 
 You have to publish the metadata file to IPFS, Swarm, or another service so
 that others can access it. You create the file by using the ``solc --metadata``
-command that generates a file called ``ContractName_meta.json``. It contains
+command that generates a JSON to the standart output. It contains
```
Marenz: We actually DO generate that file, but only when the `--output-dir` option is given. We should reflect that in the documentation though.

Marenz (changes requested):

docs/metadata.rst
LCamel: I have modified the sentence.

Ed023 (dismissed):
Nice

docs/metadata.rst
```diff
@@ -21,9 +21,12 @@ the :ref:`Standard JSON Interface<compiler-api>`.
 
 You have to publish the metadata file to IPFS, Swarm, or another service so
 that others can access it. You create the file by using the ``solc --metadata``
```
Marenz, line 23:
```suggestion
that others can access it. You create the file by using the ``solc --metadata`` command together with the ``--output-dir`` parameter. Without it, it will only be written to standard out.
```

I think it's important to mention both in the same sentence, otherwise ppl might try it and only read the ``--ouput-dir` part later.

Marenz (changes requested):

docs/metadata.rst
LCamel: Updated it in the same sentence.

Marenz (dismissed):

docs/metadata.rst
```diff
@@ -21,9 +21,12 @@ the :ref:`Standard JSON Interface<compiler-api>`.
...
-command that generates a file called ``ContractName_meta.json``. It contains
-IPFS and Swarm references to the source code, so you have to upload all source
-files and the metadata file.
+command together with the ``--output-dir`` parameter. Without the parameter,
+it will only be written to standard out.
+It contains IPFS and Swarm references to the source code, so you have to
+upload all source files and the metadata file. For IPFS, The hash contained
+in the CID returned by ``ipfs add`` (not the direct sha2-256 hash of the file)
+shall match with the one contained in the bytecode.
```
ekpyron:
```suggestion
the metadata will be written to standard output.
The metadata contains IPFS and Swarm references to the source code, so you have to
upload all source files in addition to the metadata file. For IPFS, the hash contained
in the CID returned by ``ipfs add`` (not the direct sha2-256 hash of the file)
shall match with the one contained in the bytecode.
```

LCamel: Thank you for the suggestion!

ekpyron (approved):

PR #13287, opened 2022-07-21
Document in ``solc --help`` usage of ``--metadata`` better.

Marenz:


solc/CommandLineParser.cpp
```diff
@@ -718,7 +718,7 @@ General Information)").c_str(),
 		(CompilerOutputs::componentName(&CompilerOutputs::signatureHashes).c_str(), "Function signature hashes of the contracts.")
 		(CompilerOutputs::componentName(&CompilerOutputs::natspecUser).c_str(), "Natspec user documentation of all contracts.")
 		(CompilerOutputs::componentName(&CompilerOutputs::natspecDev).c_str(), "Natspec developer documentation of all contracts.")
-		(CompilerOutputs::componentName(&CompilerOutputs::metadata).c_str(), "Combined Metadata JSON whose Swarm hash is stored on-chain.")
+		(CompilerOutputs::componentName(&CompilerOutputs::metadata).c_str(), "Request the combined Metadata JSON whose Swarm hash is stored on-chain. If used with --output-dir the metadata will be written in that directory, one file per contract each named \"<ContractName>_meta.json\", otherwise it is displayed on stdout.")
```
cameel: Now this is inconsistent with how all other components are described. They are all listed as things, but this one is now an action. I'd keep this part the way it was:
```suggestion
		(CompilerOutputs::componentName(&CompilerOutputs::metadata).c_str(), "Combined metadata JSON whose Swarm hash is stored on-chain. If used with --output-dir the metadata will be written in that directory, one file per contract each named \"<ContractName>_meta.json\", otherwise it is displayed on stdout.")
```

If you really want to change it, I think you should change it for all of them. But the way it is now is more succinct and still comprehensible so personally I'd keep it.

cameel: > If used with --output-dir the metadata will be written in that directory, one file per contract each named "<ContractName>_meta.json", otherwise it is displayed on stdout.

This is true but it's how pretty much every output option works (and some others too like `--combined-json`). Mentioning it only for `--metadata` will give an impression that it's the only one that works this way.

Maybe it would be better to expand the description of `--output-dir` instead? Or, alternatively we could have some text for the whole "Output Components" section. IIRC each section can have its own description.

Marenz: I think the last suggestion is better then, putting it in the section header

ekpyron: We should also change ``Swarm hash`` either to ``hash`` or ``IPFS hash`` or ``hash (see --metadata-hash)``...

leonardoalt:
bump reviews

ekpyron:
> bump reviews

To be clear: https://github.com/ethereum/solidity/pull/13287#discussion_r935307805 still needs addressing.

solc/CommandLineParser.cpp
Marenz: Turns out that section headers can't have more description than their caption.. at least I found no method to add text..

leonardoalt:
@Marenz I noticed that you pushed after the last comment. Was everything updated? If yes, can you maybe just ping the reviewers here?

Marenz:
I addressed Daniels comment and basically removed my original adjustment due to cameels valid points and instead opted for a tiny change to make it more clear that output options affect output components

leonardoalt:
Ok, thanks. I was just wondering whether it's ready for review again.

solc/CommandLineParser.cpp
```diff
@@ -718,7 +718,7 @@ General Information)").c_str(),
 		(CompilerOutputs::componentName(&CompilerOutputs::signatureHashes).c_str(), "Function signature hashes of the contracts.")
 		(CompilerOutputs::componentName(&CompilerOutputs::natspecUser).c_str(), "Natspec user documentation of all contracts.")
 		(CompilerOutputs::componentName(&CompilerOutputs::natspecDev).c_str(), "Natspec developer documentation of all contracts.")
-		(CompilerOutputs::componentName(&CompilerOutputs::metadata).c_str(), "Combined Metadata JSON whose Swarm hash is stored on-chain.")
+		(CompilerOutputs::componentName(&CompilerOutputs::metadata).c_str(), "Combined Metadata JSON whose IPFS hash is stored on-chain.")
```
aarlt, line 721: We should replace "stored on-chain" with "stored in contract binary".

cameel: How about "appended to the bytecode"? Or "stored in CBOR metadata at the end of the bytecode"?

solc/CommandLineParser.cpp
cameel: I see you changed it to `IPFS` now. Personally, I'd go with `hash` because otherwise we'll have to update it whenever we change the default.

cameel: Hmm... I guess we could hack that in by making the caption multi-line. Not sure if that would work out nice though.

cameel (approved):
I added more comments but overall, even in the current form, the PR is still an improvement so I wouldn't mind just merging it and being done with it :)

Unless you do want to polish it some more.

PR #13289, opened 2022-07-22
Calldata validation tests

### LINKED ISSUE #12922: Increase test coverage for Calldata Validation
We have comparably decent test coverage and fuzzing confirmation for data integrity on decoding valid calldata encodings.
However, we have only sparse coverage for calldata validation, leading to bugs like https://github.com/ethereum/solidity/pull/12907 going unnoticed.

While it is hard to test *everything* (this would mean checking all combinations between static and dynamic types up to three dimensions and in each verify the bounds checks for re-encoding, index accessing, copying to memory and copying to storage, resulting in an enormous test space), and we should also try to employ fuzzing for the validation part (see https://github.com/ethereum/solidity/issues/12914), at least *some* more coverage in the unit tests would be good as well.

wechman:
closes #12922


ekpyron:
Any reason for not adding ``// revertStrings: debug`` to all of them? If that doesn't cause issues, it may make it easy to properly check that this now catches all revert conditions.

wechman:
> Any reason for not adding `// revertStrings: debug` to all of them? If that doesn't cause issues, it may make it easy to properly check that this now catches all revert conditions.

@ekpyron  The reason is an error message that keeps changing in consecutive test runs. An example: 

`// f(uint256[][]): 0x20, 2, 0x40, 0xa0, 2, 5, 6, 3, 7, 8 -> FAILURE, 3963877391197344453575983046348115674221700746820753546331534351508065746944, 862718293348820473429344482784628181556388621521298319395315527974912, 0x2b414249206465636f64696e673a20696e76616c69642063616c6c6461, 52639898083992983106342913290719799829523823861698573317707643453664495927296, 0x4100000000000000c820b8edba5500000100000001000000
`

I guess we may want to fix that, but I am not sure if we should do that in scope of this ticket.

ekpyron:
Ah right, I vaguely remember that there's some legacy vs. via-IR differences with some of these cases - we should indeed try to fix that and align those eventually, but true, no need to burden this PR with that.

test/libsolidity/semanticTests/abiEncoderV2/calldata_overlapped_dynamic_arrays.sol
```diff
@@ -0,0 +1,43 @@
...
+    function f_memory(uint[] calldata a, uint[2] calldata b) public returns (uint[] memory, uint[2] memory) {
+        return (a, b);
+    }
+
+    function f_encode(uint[] calldata a, uint[2] calldata b) public returns (bytes memory) {
+        return abi.encode(a, b);
+    }
+
+    function f_which(uint[] calldata a, uint[2] calldata b, uint which) public returns (bytes memory) {
+	return abi.encode(a[which], b[1]);
```
ekpyron: indentation, also below

wechman: A fixup commit delivered. Also, I delivered another one that removes `compileViaYul: also` option, because meanwhile it became a default.

ekpyron (approved):

PR #13291, opened 2022-07-23
The word "in" was missed.

revug:
... so ``if (1) { ... }`` is *not* valid Solidity.


chriseth:
IMO the word is not needed, it's like "it's valid english" - you can say both.

revug:
OK, thank you very much!

PR #13292, opened 2022-07-23
Placeholder elaboration in function-modifiers.rst

The-Arbiter:
Hi,

This PR is a continuation of Create placeholders.rst #13239 

Thanks @cameel for the feedback, I have removed the mutex code example and cleaned things up a bit as well as implemented all of the changes in your review.

Looking back on it, I agree with you that placeholders didn't really deserve their own page in the docs. This version just adds two paragraphs to function-modifiers prior to the first mention of the `_` character in the section. This was the area where I feel it best fits within this page.

Please let me know if you have any other recommendations or changes.

Thanks,
Arby

### DIFF
--- a/docs/contracts/function-modifiers.rst
+++ b/docs/contracts/function-modifiers.rst
@@ -111,6 +111,12 @@ whitespace-separated list and are evaluated in the order presented.
 Modifiers cannot implicitly access or change the arguments and return values of functions they modify.
 Their values can only be passed to them explicitly at the point of invocation.
 
+In function modifiers, it is necessary to specify when you want the function to which the modifier is
+applied to be run. The placeholder statement (denoted by a single underscore character ``_``) is used to
+denote where the body of the function being modified should be inserted. Note that the
+placeholder operator is different from using underscores as leading or trailing characters in variable
+names, which is a stylistic choice.
+
 Explicit returns from a modifier or function body only leave the current
 modifier or function body. Return variables are assigned and
 control flow continues after the ``_`` in the preceding modifier.


docs/contracts/function-modifiers.rst
```diff
@@ -111,6 +111,15 @@ whitespace-separated list and are evaluated in the order presented.
 Modifiers cannot implicitly access or change the arguments and return values of functions they modify.
 Their values can only be passed to them explicitly at the point of invocation.
 
+In function modifiers, it is necessary to specify when you want the function to which the modifier is
```
chriseth, line 114: I don't think this is the right place to talk about the fundamental feature of modifiers.
The way they are used is explained in the source code comments of the example contract.
We should either not start with an example but with something like this paragraph instead and then provide a slimmed down example or leave it as it is.

The-Arbiter: Hi Chris,

Thanks for reviewing this.

In my previous PR I made my case for why adding this information adds value:

**Why you made your change**

_When I recently read the Solidity docs, I felt that there wasn't enough information on placeholders used within modifiers.

This question has been [asked before](https://ethereum.stackexchange.com/questions/19171/what-does-underscore-do)._

I largely learnt Solidity by going through the docs during my beginner phase and the modifier not being explicitly explained was the biggest hole in the docs IMO (and clearly by the SO thread others feel the same way). I've had a few friends in the space tell me they felt that this was a hole in the docs too; though how big of an issue it is varies person to person.

While I acknowledge it is in the code comments and everyone _should_ be reading that, not everyone does and thus I feel a little added clarity would be beneficial to beginners.

> We should either not start with an example but with something like this paragraph instead and then provide a slimmed down example or leave it as it is.

I do agree with your point here, perhaps it is too long - originally this was formatted for a separate .rst file and admittedly it can be slimmed down a little more. 

Could you please advise which aspects of the placeholder statement you deem to be most important (for example, usage in code, what it does, ability to use it multiple times, etc.) so that I can make appropriate changes? I will commit them and then tag you again in this PR.

Kind Regards,
Arbiter

chriseth: Thanks for your response! Please use your own judgment wrt what you think is most important.

The-Arbiter: Hey @chriseth I have cut it down to one paragraph as per your earlier guidance.

I feel that this is succinct enough while also offering context to confused readers - probably quite a good balance!

Let me know what you think.

Arby

The-Arbiter:
Shortened changes to one paragraph.

ekpyron (dismissed):
I'd say that it may be nicer to move the new paragraph up in the section, s.t. it becomes the second paragraph.
But I'd also be fine with merging where it is, so I'm approving.

leonardoalt (changes requested):
The commits contain a merge commit which should be removed. Please simply rebase your original commit onto origin/develop.

leonardoalt:
I fixed the merge commit and rebased it.

leonardoalt (approved):

PR #13299, opened 2022-07-25
Fix bad_cast when abi.encodeCall() receives a tuple expression

### LINKED ISSUE #13101: `bad_cast` when passing a tuple returned from a function to `abi.encodeCall()`
## Description
`abi.encodeCall()` expects a tuple as its second argument. Normally the tuple would be defined in place (e.g. `abi.encodeCall(f, (1, 2, 3))`) but it's also possible to give it a tuple that is the result of a function call. Doing so results in an unhandled `bad_cast` exception and the compiler crashes.

The same thing happens when you pass in the result of an error or event construction since technically these return an empty tuple.

The crash does not happen when the tuple has only one element.

Doing this with the other functions from the `abi.encodeX()` family results in a compilation error (`This type cannot be encoded`) so I think this situation should be a compilation error too.

## Steps to Reproduce
```solidity
contract C {
    function g0() internal {}
    function g1() internal returns (uint) {}
    function g2() internal returns (uint, uint) {}

    event Ev();
    error Er();

    function f() public pure {
        abi.encodeCall(this.f, g0());  // bad_cast
        abi.encodeCall(this.f, g1());  // OK
        abi.encodeCall(this.f, g2());  // bad_cast
        abi.encodeCall(this.f, Ev());  // bad_cast
        abi.encodeCall(this.f, Er());  // bad_cast
    }
}
```
```
Uncaught exception:
Dynamic exception type: std::bad_cast
std::exception::what: std::bad_cast
```
## Environment
- Compiler version: 0.8.11, 0.8.12, 0.8.13, 0.8.14

matheusaaguiar:
Fixes #13101.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix compiler crash when `abi.encodeCall` received a tuple expression instead of an inline tuple.
 
 
 ### 0.8.15 (2022-06-15)
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -2248,8 +2248,17 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
 	auto const* tupleType = dynamic_cast<TupleType const*>(type(*arguments[1]));
 	if (tupleType)
 	{
-		auto const& argumentTuple = dynamic_cast<TupleExpression const&>(*arguments[1].get());
-		callArguments = decltype(callArguments){argumentTuple.components().begin(), argumentTuple.components().end()};
+		if (TupleExpression const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get()))
+			callArguments = decltype(callArguments){argumentTuple->components().begin(), argumentTuple->components().end()};
+		else
+		{
+			m_errorReporter.typeError(
+				9062_error,
+				arguments[1]->location(),
+				"Expected an inline tuple, not an expression of a tuple type."
+			);
+			return;
+		}
 	}
 	else
 		callArguments.push_back(arguments[1]);
--- a/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_assignment_expression.sol
+++ b/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_assignment_expression.sol
@@ -1,0 +1,24 @@
+contract C {
+    function g0() internal pure {}
+    function g2() internal pure returns (uint, uint) { return (2, 3); }
+
+    function f0() public {}
+    function f2(uint, uint) public {}
+
+    function h() public view {
+        uint a;
+        uint b;
+
+        abi.encodeCall(this.f0, () = g0());
+        abi.encodeCall(this.f0, () = ());
+        abi.encodeCall(this.f2, (a, b) = g2());
+        abi.encodeCall(this.f2, (a, b) = (2, 3));
+    }
+}
+// ----
+// TypeError 5547: (284-286): Empty tuple on the left hand side.
+// TypeError 9062: (284-293): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 5547: (328-330): Empty tuple on the left hand side.
+// TypeError 9062: (328-335): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 9062: (370-383): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 9062: (418-433): Expected an inline tuple, not an expression of a tuple type.
--- a/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_returned_from_function.sol
+++ b/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_returned_from_function.sol
@@ -1,0 +1,22 @@
+contract C {
+    function g0() internal pure {}
+    function g2() internal pure returns (uint, uint) { return (2, 3); }
+
+    function f0() public {}
+    function f2(uint, uint) public {}
+
+    function h() public view {
+        abi.encodeCall(this.f0, g0());
+        abi.encodeCall(this.f2, g2());
+
+        abi.encodeCall(this.f0, (g0()));
+        abi.encodeCall(this.f2, (g2()));
+    }
+}
+// ----
+// TypeError 9062: (251-255): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 9062: (290-294): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 6473: (331-335): Tuple component cannot be empty.
+// TypeError 7788: (306-337): Expected 0 instead of 1 components for the tuple parameter.
+// TypeError 7788: (347-378): Expected 2 instead of 1 components for the tuple parameter.
+// TypeError 5407: (372-376): Cannot implicitly convert component at position 0 from "tuple(uint256,uint256)" to "uint256".
--- a/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_conditional_ternary_expression.sol
+++ b/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_conditional_ternary_expression.sol
@@ -1,0 +1,17 @@
+contract C {
+    function g0() internal pure {}
+    function g2() internal pure returns (uint, uint) { return (2, 3); }
+
+    function f0() public {}
+    function f2(uint, uint) public {}
+
+    function h() public view {
+        abi.encodeCall(this.f0, true ? g0() : g0());
+        abi.encodeCall(this.f2, true ? g2() : g2());
+        abi.encodeCall(this.f2, true ? (1, 2) : (3, 4));
+    }
+}
+// ----
+// TypeError 9062: (251-269): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 9062: (304-322): Expected an inline tuple, not an expression of a tuple type.
+// TypeError 9062: (357-379): Expected an inline tuple, not an expression of a tuple type.
--- a/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_error_event.sol
+++ b/test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_error_event.sol
@@ -1,0 +1,16 @@
+contract C {
+    event Ev();
+    error Er();
+
+    function f0() public {}
+
+    function h() public view {
+        abi.encodeCall(this.f0, Ev());
+        abi.encodeCall(this.f0, Er());
+        abi.encodeCall(this.f0, revert());
+    }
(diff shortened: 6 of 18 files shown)


stackenbotten:
There was an error when running `chk_coding_style` for commit `7d5296dc406373ad95511d90fc540b5d00a30d9b`:
```
Changelog.md:17: * Type Checker: Fix `bad_cast` exception when `abi.encodeCall` received a tuple returned from a function. 

```

test/libsolidity/syntaxTests/abiEncoder/abi_encode_call_tuple_from_function.sol
```diff
@@ -0,0 +1,15 @@
+contract C {
+    function g0() internal {}
+    function g1() internal returns (uint) {}
+    function g2() internal returns (uint, uint) {}
+
+    function f() public pure {
+        abi.encodeCall(this.f, g0());
+        abi.encodeCall(this.f, g1());
+        abi.encodeCall(this.f, g2());
```
cameel: Looking at the buggy code, seems that there are more cases affected by this bug. Could you add these to make sure your fix covers them as well?
```solidity
        uint a;
        uint b;
        abi.encodeCall(this.f, (a, b) = g2());
```

```solidity
        abi.encodeCall(this.f, true ? g2() : g2());
```

```solidity
        abi.encodeCall(this.f, (1, 1) + (2, 2));
```

```solidity
        abi.encodeCall(this.f, Ev() / Er());
```

```solidity
        abi.encodeCall(this.f, revert());
```

test/libsolidity/syntaxTests/abiEncoder/abi_encode_call_tuple_from_function.sol
```diff
@@ -0,0 +1,15 @@
+contract C {
+    function g0() internal {}
+    function g1() internal returns (uint) {}
+    function g2() internal returns (uint, uint) {}
+
+    function f() public pure {
```
cameel: I see I should have made it `view` in my original example. Otherwise this would not compile even if the type was right:
```suggestion
    function f() public view {
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -2248,8 +2248,17 @@ void TypeChecker::typeCheckABIEncodeCallFunction(FunctionCall const& _functionCa
...
-		auto const& argumentTuple = dynamic_cast<TupleExpression const&>(*arguments[1].get());
-		callArguments = decltype(callArguments){argumentTuple.components().begin(), argumentTuple.components().end()};
+		if (TupleExpression const* argumentTuple = dynamic_cast<TupleExpression const*>(arguments[1].get()))
+			callArguments = decltype(callArguments){argumentTuple->components().begin(), argumentTuple->components().end()};
+		else
+		{
+			m_errorReporter.typeError(
+				9062_error,
+				arguments[1]->location(),
+				"This type cannot be encoded."
```
cameel: The message is identical to `2056_error` and in that case I'd expect both to have the same error code.

But actually, the checks are not identical. Maybe we should just make the message here more specific?
```suggestion
				"Expected an inline tuple, not an expression of a tuple type."
```

Or maybe this. Might be clearer to users but is a bit informal. Not sure which one's better:
```suggestion
				"Expected an inline tuple embedded directly in the call, not an expression of a tuple type."
```

test/libsolidity/syntaxTests/abiEncoder/abi_encode_call_tuple_from_function.sol
cameel: I used `this.f` in my original example for brevity but in the test let's use proper functions that actually have the number or args matching number of returns in `g0()`, `g1()` and `g2()`. I.e. let's make it so that the error we're interested in is the only error in the snippet and it would compile otherwise.

cameel: Also, please add this one too unless we already have a test for it:
```
        abi.encodeCall(this.f, (g0()));
        abi.encodeCall(this.f, (g1()));
        abi.encodeCall(this.f, (g2()));
```

It will be a different error and does not cause `bad_cast` but we should just have it covered.

test/libsolidity/syntaxTests/abiEncoder/abi_encode_call_tuple_from_error_event.sol
```diff
@@ -0,0 +1,12 @@
+contract C {
+    event Ev();
+    error Er();
+
+    function f() public pure {
+        abi.encodeCall(this.f, Ev());  // bad_cast
+        abi.encodeCall(this.f, Er());  // bad_cast
```
cameel: This no longer causes `bad_cast` after the fix.
```suggestion
        abi.encodeCall(this.f, Ev());
        abi.encodeCall(this.f, Er());
```

Changelog.md
```diff
@@ -14,6 +14,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Type Checker: Fix `bad_cast` exception when `abi.encodeCall` received a tuple returned from a function.
```
cameel: Looks like this is more general than that. Happens even in cases without a function call.

Also, in cases where we have an uncaught exception, I'd call it "compiler crash" because from user's standpoint that's what it is. The fact that it's a C++ exception is just an unimportant implementation detail.
```suggestion
 * Type Checker: Fix compiler crash when `abi.encodeCall` received a tuple expression instead of an inline tuple.
```

cameel (changes requested):

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Right. I was not sure about the message so I just left it as in the other similar error.
About your suggestions, I think that the first one is good. The origin of the error will be highlighted by the compiler message, so I guess it should be obvious that the inline tuple is expected to be embedded inside the call:

```
Error: Expected an inline tuple, not an expression of a tuple type.
  --> /path/to/file.sol:x:y
   |
13 |         abi.encodeCall(this.f0, g0());
   |                                 ^^^^
```

stackenbotten:
There was an error when running `chk_coding_style` for commit `cd78e685c22a44b93252d41a54443f41ec53578d`:
```
test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_expression.sol:16:        

```

stackenbotten:
There was an error when running `chk_coding_style` for commit `22d3630282c3d041b1cac817c16a18e929d572bc`:
```
test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_expression.sol:16:    

```

matheusaaguiar:
Rebased.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_returned_from_function.sol
```diff
@@ -0,0 +1,26 @@
...
+    function g1() internal returns (uint) { return (1); }
+    function g2() internal returns (uint, uint) { return (2, 3); }
+
+    function f0() public {}
+    function f1(uint) public {}
+    function f2(uint, uint) public {}
+
+    function h() public view {
+        abi.encodeCall(this.f0, g0());
+        abi.encodeCall(this.f1, g1());
```
cameel:
```suggestion
        abi.encodeCall(this.f1, g1()); // OK
```

cameel: Actually, maybe we even should have a test covering this case with all `abi.` functions. I'm not sure we already have one.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_expression.sol
```diff
@@ -0,0 +1,44 @@
...
+
+        abi.encodeCall(this.f1, (1) + (2));
+        abi.encodeCall(this.f2, (1, 1) + (2, 2));
+
+        abi.encodeCall(this.f1, (a) = g1());
+        abi.encodeCall(this.f2, (a, b) = g2());
+
+        abi.encodeCall(this.f0, Ev() / Er());
+
+        abi.encodeCall(this.f0, !());
```
cameel: Here again, I'd add a short comment after each one that's supposed to work. And might be better to just put them in a separate test. Both to make it easier to match the errors to bad ones and because some errors may prevent other errors from showing up so the fact that here a specific line has no error does not 100% it works.

cameel (dismissed):
Looks great. There are some minor tweaks I'd like you to do but that's just cosmetics.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_expression.sol
matheusaaguiar: Ok. I separated them in more specific files. Also added a few more cases to some tests.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_returned_from_function.sol
matheusaaguiar: I created separated tests for each `abi.encode` function.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_assignment_expression.sol
```diff
@@ -0,0 +1,28 @@
...
+    function f2(uint, uint) public {}
+
+    function h() public view {
+        uint a;
+        uint b;
+
+        abi.encodeCall(this.f0, () = g0());
+        abi.encodeCall(this.f0, () = ());
+        abi.encodeCall(this.f1, (a) = g1()); // Ok
+        abi.encodeCall(this.f1, (a) = (1)); // Ok
```
cameel: Oh, it's actually weird that we allow that but I guess that's a consequence of the fact that a single-element tuple is not really a tuple...

cameel (dismissed):
Sorry, I think I haven't been really clear - I meant that the "ok" cases might be better off separated from the failing ones because then we would have both positive and negative cases. When anything in a test case fails you can't be 100% sure the not failing bits are OK because the analysis might have been cut short by the error. Or the case might pass analysis but fail in the codegen.

Overall, would be best to change that but it's not critical and I'd like to have this PR in the release so I'll merge it even if it remains as is. Just please squash your commits.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_assignment_expression.sol
matheusaaguiar: Yeah, in the end it becomes the type of the only element, right?

matheusaaguiar:
> Sorry, I think I haven't been really clear - I meant that the "ok" cases might be better off separated from the failing ones [...]

I can understand what you meant now and maybe I should have confirmed, but no problem :)
I am going to separate the "ok" tests, it should be quick.

test/libsolidity/syntaxTests/abiEncoder/abi_encodeCall_tuple_from_assignment_expression.sol
cameel: Yes. I mean, I think there is some AST annotation that lets you detect that it's wrapped in the parentheses. I remember @Marenz having to deal with that for weird cases where stuff could be wrapped (e.g. calls like `(L.f)()`).

But I'd say that dealing with that is out of scope of this PR. If we decided to do this, we'd want to change this for all the encoding functions. It would also be breaking because it would disallow some cases that now work fine.

matheusaaguiar:
Rebased and squashed commits.

cameel (approved):
Great. Thanks!

PR #13300, opened 2022-07-26
Docs — Minor grammar fixes in `abi-spec.rst`

anannyenaik:
## Category:

- [x] Documentation -> Grammar

## Description:

Minor grammar fixes in `abi-spec.rst`.

*Note: Please excuse the ambiguous branch name, I edited the file directly on GitHub; hence, it named and created the branch automatically.*

### DIFF
--- a/docs/abi-spec.rst
+++ b/docs/abi-spec.rst
@@ -13,7 +13,7 @@ The Contract Application Binary Interface (ABI) is the standard way to interact 
 from outside the blockchain and for contract-to-contract interaction. Data is encoded according to its type,
 as described in this specification. The encoding is not self describing and thus requires a schema in order to decode.
 
-We assume the interface functions of a contract are strongly typed, known at compilation time and static.
+We assume that the interface functions of a contract are strongly typed, known at compilation time and static.
 We assume that all contracts will have the interface definitions of any contracts they call available at compile-time.
 
 This specification does not address contracts whose interface is dynamic or otherwise known only at run-time.
@@ -29,7 +29,7 @@ first (left, high-order in big-endian) four bytes of the Keccak-256 hash of the 
 the function. The signature is defined as the canonical expression of the basic prototype without data
 location specifier, i.e.
 the function name with the parenthesised list of parameter types. Parameter types are split by a single
-comma - no spaces are used.
+comma — no spaces are used.
 
 .. note::
     The return type of a function is not part of this signature. In
@@ -133,7 +133,7 @@ The encoding is designed to have the following properties, which are especially 
    previous version of the ABI, the number of reads scaled linearly with the total number of dynamic
    parameters in the worst case.
 
-2. The data of a variable or array element is not interleaved with other data and it is
+2. The data of a variable or an array element is not interleaved with other data and it is
    relocatable, i.e. it only uses relative "addresses".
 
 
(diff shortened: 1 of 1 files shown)


docs/abi-spec.rst
```diff
@@ -646,18 +646,18 @@ would result in the JSON:
...
 The canonical type is determined until a tuple type is reached and the string description up
 to that point is stored in ``type`` prefix with the word ``tuple``, i.e. it will be ``tuple`` followed by
 a sequence of ``[]`` and ``[k]`` with
 integers ``k``. The components of the tuple are then stored in the member ``components``,
-which is of array type and has the same structure as the top-level object except that
+which is of an array type and has the same structure as the top-level object except that
 ``indexed`` is not allowed there.
 
-As an example, the code
+As an example, the code:
```
chriseth: I wouldn't use `:` here since the sentence is not yet finished at this point.

anannyenaik: Okay, I will edit it.

anannyenaik: Committed to the same branch -- removed the `:`.

chriseth:
Can you squash to a single commit, please?

anannyenaik:
@chriseth Sorry, how can do I that? Is it via squash merging into the `develop` branch of my fork?

chriseth:
No, squash-merging is a different thing. You have to rebase your own fork interactively (`git rebase -i origin/develop`) and then use "squash" on the second commit. After that you have to force-push to your branch (`git push --force`).

anannyenaik:
Got it, I will do it now.

By the way, in my case specifically does it matter if I just delete that extraneous commit altogether instead of squashing it?

Marenz:
You mean the commit that deletes the `:`? It is not "extraneous" as we want that change, right? But it should be part of the first commit.

anannyenaik:
@Marenz No no -- I meant the first one, that is the extraneous one.

Marenz:
I believe there is some kind of misunderstanding here.. but I am fairly certain what you want is what chris was describing. It is a bit better explained here (squash and fixup are basically the same): https://www.mikulskibartosz.name/git-fixup-explained/

anannyenaik:
@Marenz To clear the misunderstanding let's simplify. Ok, so there are two commits:

`Minor grammar fixes` and `Fixes :`

The one not needed is `Minor grammar fixes` -- so, I was just enquiring *if in my case* (after the interactive rebase): it makes any difference to just delete `Minor grammar fixes` than to squash it into one.

Marenz:
But `Minor grammar fixes` contains 99% of your changes and `Fixes :` only a correction to those previous changes, why do you think it is not needed?

If you delete the first, the second is just the removal of the `:` which by then doesn't exist anymore as the addition of it was in the first..

anannyenaik:
I assumed the latter commit incorporates the changes from that commit -- I guess I will just squash it then. Thank you

Marenz:
Squashing does exactly that. A commit is only a state in the repository if it is still attached to its previous commit. Without that it is just a set of changes.

So while the state is what we want at the `Fixes :` commit, the change-set in it is only the removal of `:`.

anannyenaik:
That clarified it, thank you. My rebase is done locally -- I also did `git push --force` but got the following message: "everything up-to-date".



*Side-note: Also, does the order precedence matter when squashing? For instance, `Commit B` before `Commit A` being squashed; would this yield the same result as `Commit B` after `Commit A` being squashed?*

ekpyron:
It does not look like the squashing actually worked - I still see two commits :-).
``git rebase -i origin/develop`` should have opened an editor to specify how to rebase and in particular, which commits to squash. Are you sure you requested the second commit to be squashed in there?

Also wrt to your question: you always squash a subsequent commit into the commit before it, i.e. if you have ``commit A`` followed by ``commit B``, you have to squash ``B`` into ``A``, not the other way round.

Apart from this, I'd say we can merge this.

Marenz:
@krakxn if you prefer we can take care of the rest from here on :)

anannyenaik:
@Marenz Is it okay now?

ekpyron (approved):

PR #13304, opened 2022-07-26
Disable zeppelin tests that detect wrong revert reason

Marenz:
Example of a failing test: https://app.circleci.com/pipelines/github/ethereum/solidity/25359/workflows/d917e214-0762-4caa-8f46-b7fba63e7e61/jobs/1117784
This will avoid the wrong revert reason detection like these:
* Received: Transaction reverted: function was called with incorrect parameters
* Expected: SafeCast: value doesn't fit in 248 bits

Related: https://github.com/NomicFoundation/hardhat/issues/2453

### DIFF
--- a/test/externalTests/zeppelin.sh
+++ b/test/externalTests/zeppelin.sh
@@ -70,6 +70,7 @@ function zeppelin_test
     sed -i "s|it(\('reverts \)|it.skip(\1|g" structs/EnumerableSet.behavior.js
     popd
 
+
     # In some cases Hardhat does not detect revert reasons properly via IR.
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2453 gets fixed.
     sed -i "s|it(\('reverts if the current value is 0'\)|it.skip(\1|g" test/utils/Counters.test.js
@@ -111,6 +112,11 @@ function zeppelin_test
     # Could result in an error in the future.
     sed -zi "s|it(\('deposit'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
     sed -zi "s|it(\('withdraw'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
+
+    # Here only the testToInt(248) and testToInt(256) cases fail so change the loop range to skip them
+    sed -i "s|range(8, 256, 8)\(.forEach(bits => testToInt(bits));\)|range(8, 240, 8)\1|" test/utils/math/SafeCast.test.js
+
+
 
     # TODO: Remove this when https://github.com/NomicFoundation/hardhat/issues/2115 gets fixed.
     sed -i "s|describe\(('Polygon-Child'\)|describe.skip\1|g" test/crosschain/CrossChainEnabled.test.js


Marenz:
I am basically excluding the `248` bit variant for this call: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/math/SafeCast.test.js#L132

Marenz:
Updated to exclude 256 bit as well (CI is failing)

test/externalTests/zeppelin.sh
```diff
@@ -111,6 +112,9 @@ function zeppelin_test
     # Could result in an error in the future.
     sed -zi "s|it(\('deposit'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
     sed -zi "s|it(\('withdraw'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
+    # Avoid the testToInt(248) and testToInt(256) case
+    sed -i "s|range(8, 256, 8).forEach(bits => testToInt(bits));|  range(8, 240, 8).forEach(bits => testToInt(bits));|" test/utils/math/SafeCast.test.js
```
cameel: The regex needs escaping:
```suggestion
    sed -i "s|range\(8, 256, 8\)\.forEach\(bits => testToInt\(bits\)\);|  range(8, 240, 8).forEach(bits => testToInt(bits));|" test/utils/math/SafeCast.test.js
```

cameel: Also, you don't have to repeat the whole thing if you use match groups:
```suggestion
    sed -i "s|range\(8, 256, 8\)(\.forEach\(bits => testToInt\(bits\)\);)|range(8, 240, 8)\1|" test/utils/math/SafeCast.test.js
```
This could even be further reduced to just the `256` part but I think including the whole `range()` part makes it more readable.

test/externalTests/zeppelin.sh
```diff
@@ -111,6 +112,9 @@ function zeppelin_test
     # Could result in an error in the future.
     sed -zi "s|it(\('deposit'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
     sed -zi "s|it(\('withdraw'\)|it.skip(\1|3" test/token/ERC20/extensions/ERC4626.test.js
+    # Avoid the testToInt(248) and testToInt(256) case
```
cameel: Even with this comment I had to look into the source to understand what's happening. Maybe we can tweak it to be clearer?
```suggestion
    # Here only the testToInt(248) and testToInt(256) cases fail so change the loop range to skip them
```
I think it wasn't clear to me mostly because I did not realize that `testToInt()` runs the test case. Though probably I should have guessed that from the name but I actually expected this to be a bit of code inside one of the test cases rather than something that runs test cases themselves...

cameel (commented):
Looks like it works, but it just needs small tweaks. Regex needs escaping for example.

Also, could you add a link to the failing CI job and also paste some of the the error messages into the description? It helps later to check if we have already seen a particular error and what we did.

test/externalTests/zeppelin.sh
Marenz: Looks like the escaping broke it. Only for match groups you need to escape...

cameel (dismissed):

cameel (approved):

PR #13305, opened 2022-07-26
CI: Add ARM64 builds

### LINKED ISSUE #11351: ARM builds for Linux
With Apple switching to ARMv8 releasing ARM builds of solc is increasingly important, especially linux ARM builds.

The reason for this is that if you use Docker with an ARM mac, you'll end up running your containers in an ARM-based Linux, so you can't run any native solc build.

In Mac the situation is different, you can still run x86_64 binaries using Rosetta2, and this process is 100% transparent.

Hardhat, and I guess all the other Node.js-based tools, automatically fallbacks to socjs, so it still works in ARM-based Docker containers. Other tools, like Brownie, probably break in that situation.

willianpaixao:
Closes #11351

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,6 +306,15 @@ defaults:
         TERM: xterm
         MAKEFLAGS: -j 10
 
+  - base_ubuntu2004_arm_large: &base_ubuntu2004_arm_large
+      <<: *base_ubuntu2004_arm_large
+      machine:
+        image: ubuntu-2004:2022.04.1
+      resource_class: arm.large
+      environment:
+        TERM: xterm
+        MAKEFLAGS: -j 5
+
   - base_buildpack_focal_small: &base_buildpack_focal_small
       docker:
         - image: buildpack-deps:focal
@@ -424,6 +433,7 @@ defaults:
       <<: *workflow_trigger_on_tags
       requires:
         - b_ubu_static
+        - b_ubu_arm_static
 
   - workflow_archlinux: &workflow_archlinux
       <<: *workflow_trigger_on_tags
@@ -629,6 +639,17 @@ jobs:
       - persist_to_workspace: *artifacts_executables
       - gitter_notify_failure_unless_pr
 
+  b_ubu_arm: &b_ubu_arm
+    <<: *base_ubuntu2004_arm_large
+    steps:
+      - checkout
+      - run: *run_build
+      - store_artifacts: *artifacts_solc
+      - store_artifacts: *artifact_solidity_upgrade
+      - store_artifacts: *artifact_yul_phaser
+      - persist_to_workspace: *artifacts_executables
+      - gitter_notify_failure_unless_pr
+
   # x64 ASAN build, for testing for memory related bugs
   b_ubu_asan: &b_ubu_asan
     # Runs slightly faster on large and xlarge but we only run it nightly so efficiency matters more.
@@ -703,6 +724,25 @@ jobs:
       CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
     steps:
       - checkout
+      - run: *run_build
+      - run:
+          name: strip binary
+          command: strip build/solc/solc
+      - store_artifacts: *artifacts_solc
+      - persist_to_workspace: *artifacts_executables
+      - gitter_notify_failure_unless_pr
+
+  b_ubu_arm_static:
+    <<: *base_ubuntu2004_arm_large
+    environment:
+      TERM: xterm
+      MAKEFLAGS: -j 5
+      CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
+    steps:
+      - checkout
+      - run:
+          name: "install dependencies"
+          command: sudo apt-get --quiet --assume-yes --no-install-recommends install libboost-all-dev z3 libz3-dev
       - run: *run_build
       - run:
           name: strip binary
@@ -908,6 +948,11 @@ jobs:
 
   t_ubu_soltest_all: &t_ubu_soltest_all
     <<: *base_ubuntu2004
+    parallelism: 50
+    <<: *steps_soltest_all
+
+  t_ubu_arm_soltest_all: &t_ubu_arm_soltest_all
+    <<: *base_ubuntu2004_arm_large
     parallelism: 50
     <<: *steps_soltest_all
 
@@ -1315,11 +1360,14 @@ workflows:
 
       # Static build
       - b_ubu_static: *workflow_trigger_on_tags
+      - b_ubu_arm_static: *workflow_trigger_on_tags
 
       # Ubuntu build and tests
       - b_ubu: *workflow_trigger_on_tags
+      - b_ubu_arm: *workflow_trigger_on_tags
       - t_ubu_cli: *workflow_ubuntu2004
       - t_ubu_soltest_all: *workflow_ubuntu2004
+      - t_ubu_arm_soltest_all: *workflow_ubuntu2004
       - t_ubu_soltest_enforce_yul: *workflow_ubuntu2004
       - b_ubu_clang: *workflow_trigger_on_tags
       - t_ubu_clang_soltest: *workflow_ubuntu2004_clang


willianpaixao:
@cameel following your instructions in https://github.com/ethereum/solidity/issues/11351#issuecomment-1072653767, note that at the moment CircleCI doesn't support docker in ARM, so you'll see the workflow is slightly different. Is this approach reasonable (if I finish the PR will be merged?).

leonardoalt:
@cameel question for you /\

.circleci/config.yml
```diff
@@ -711,6 +721,25 @@ jobs:
       - persist_to_workspace: *artifacts_executables
       - gitter_notify_failure_unless_pr
 
+  b_ubu_static_arm:
```
cameel:
```suggestion
  b_ubu_arm_static:
```

.circleci/config.yml
```diff
@@ -711,6 +721,25 @@ jobs:
...
+    <<: *base_ubuntu2004_arm_large
+    environment:
+      TERM: xterm
+      MAKEFLAGS: -j 5
+      CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DSTRICT_Z3_VERSION=OFF -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
```
cameel: In CI we always want to run with `-DSTRICT_Z3_VERSION=ON`
```suggestion
      CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
```

.circleci/config.yml
```diff
@@ -306,6 +306,15 @@ defaults:
         TERM: xterm
         MAKEFLAGS: -j 10
 
+  - base_ubuntu2004_arm_large: &base_ubuntu2004_arm_large
+      <<: *base_ubuntu2004_arm_large
+      machine:
+        image: ubuntu-2004:2022.04.1
```
cameel, line 312: Slightly offtopic but I think we should really have some checklist with the versions we have hard-coded in various places. There's ever more of them and we don't always remember to keep them up to date. I think that it will definitely be easy to forget about this one.

.circleci/config.yml
```diff
@@ -1315,6 +1344,7 @@ workflows:
 
       # Static build
       - b_ubu_static: *workflow_trigger_on_tags
+      - b_ubu_static_arm: *workflow_trigger_on_tags
```
cameel: I think this should have some test runs. We don't have one for `b_ubu_static` only because we already run a few different variations of soltest on standard Ubuntu. For ARM we should have at least one.

Also, I think we might want to have a debug build as well. We used to have them for macOS and Windows but they were slow and expensive. For ARM I see that [pricing is the same as for standard linux](https://circleci.com/product/features/resource-classes/#arm) so it's not an obstacle. As long as they are not too slow, we should have them.

So overall, please add:
- `b_ubu_arm`
- `t_ubu_arm_soltest_all`
- `t_ubu_arm_cli`
- `t_ubu_arm_static_soltest`

.circleci/config.yml
```diff
@@ -711,6 +721,25 @@ jobs:
       - persist_to_workspace: *artifacts_executables
       - gitter_notify_failure_unless_pr
 
+  b_ubu_static_arm:
+    <<: *base_ubuntu2004_arm_large
```
cameel, line 736: What's the running time for each size? We'd want the one that has the best time/cost ratio with a preference toward the faster one if ratios are close.

.circleci/config.yml
```diff
@@ -711,6 +721,25 @@ jobs:
...
+    environment:
+      TERM: xterm
+      MAKEFLAGS: -j 5
+      CMAKE_OPTIONS: -DCMAKE_BUILD_TYPE=Release -DUSE_Z3_DLOPEN=ON -DSTRICT_Z3_VERSION=OFF -DUSE_CVC4=OFF -DSOLC_STATIC_STDLIBS=ON
+    steps:
+      - checkout
+      - run:
+          name: "install dependencies"
+          command: sudo apt-get --quiet --assume-yes --no-install-recommends install libboost-all-dev z3 libz3-dev
```
cameel, line 745: I think we'll need more stuff from [`Dockerfile.ubuntu2004`](https://github.com/ethereum/solidity/blob/develop/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004) here.

For example Z3. We have our own PPA because Z3 in Ubuntu is usually outdated. That's why you needed `-DSTRICT_Z3_VERSION=OFF` but that's not something we want to keep in this PR. I guess this means we'll have to extend [`deps-ppa/static_z3.sh`](https://github.com/ethereum/solidity/blob/develop/scripts/deps-ppa/static_z3.sh) to cover ARM too.

You'll also need evmone to run soltest.

Other stuff like hera (needed for wasm), sphinx (docs) or other deps might actually be rendundant since the default Ubuntu image is used for a lot of different jobs and this one not necessarily.

.circleci/config.yml
cameel: A few more missing bits:
- We absolutely need `b_bytecode_ubu_arm` and that should be plugged into `t_bytecode_compare`.
- (optional) Maybe it would be worth having `t_ubu_arm_ext_hardhat` similar to the current `t_ems_ext_hardhat`?
- The new binary should be included in the archive created by `c_release_binaries`.
- The new platform must be mentioned in the [`ReleaseChecklist`](https://github.com/ethereum/solidity/blob/develop/ReleaseChecklist.md).
- The new platform has to be added to [solc-bin](https://github.com/ethereum/solc-bin/). This involves:
    - Creating a directory for the new platform with list stubs.
    - Adjusting the [`update`](https://github.com/ethereum/solc-bin/blob/gh-pages/update) script to process the binaries from it.
    - Adding the new platform to [`t-bytecode-compare.yml`](https://github.com/ethereum/solc-bin/blob/gh-pages/.github/workflows/t-bytecode-compare.yml) and creating a test PR to make sure it does work (this would be best done in a fork at first not to spam the repo with too many test PRs).
- This PR should have a changelog entry.

cameel (changes requested):
> following your instructions in https://github.com/ethereum/solidity/issues/11351#issuecomment-1072653767, note that at the moment CircleCI doesn't support docker in ARM, so you'll see the workflow is slightly different. Is this approach reasonable (if I finish the PR will be merged?).

Normally we'd add a new dockerfile to [`buildpack-deps/`](https://github.com/ethereum/solidity/tree/develop/scripts/docker/buildpack-deps) and plug it into the [`buildpack-deps.yml`](https://github.com/ethereum/solidity/blob/develop/.github/workflows/buildpack-deps.yml) workflow but if CircleCI does not allow custom images then I guess what you did here is the only thing we can really do. We'd ultimately do the same thing.

So yeah, we will not reject the PR just because of that :) But there are quite a few things missing and we'll only be willing to merge it if it's complete. Check my comments below, especially https://github.com/ethereum/solidity/pull/13305#discussion_r946786748.

In any case thanks for the work you already put into it. If we get this to completion this will be really appreciated by users.

> Closes #11351

I don't think the PR fixes the whole issue because we also need to provide builds of the older compiler versions and that's going to be a lot of separate work. So for now I'd just change it to something like `Part of #11351`.

.circleci/config.yml
willianpaixao: Thanks for the note, I'll mark this comment as resolved as we should open as a separated ticket.

.circleci/config.yml
willianpaixao: I do not know the average time as I have not access to CircleCI and have done no benchmark. That said, the _large_ instance is the fastest one in the free tier. https://circleci.com/docs/using-arm

.circleci/config.yml
willianpaixao: Applied.

.circleci/config.yml
willianpaixao: @cameel can you elaborate how these packages are built and pushed to the PPA?

.circleci/config.yml
cameel: The CI is public so you should be able to view the results (you may have to login with a github account though).

Anyway, yeah, if the CI checks in this PR are not running under the solidity org then the instance sizes may be limited. In that case we'll have to adjust it ourselves. I'm going to keep this comment open so that we don't forget.

.circleci/config.yml
cameel: The script I linked builds a `.deb` source package and then uploads to [`~ethereum/cpp-build-deps` PPA](https://launchpad.net/~ethereum/+archive/ubuntu/cpp-build-deps) on Launchpad. We run it locally whenever a new version of Z3 is released.

To be honest I'm only vaguely familiar with it. I know we run the script when we update Z3 but so far I never had to do it myself and I don't know much about how it's set up. In particular not sure if you could fit an ARM version into the same PPA or if we'd need a separate one.

Actually, I see that the source package has `Architecture: any` so it's possible that it might be as simple as selecting ARM from some list when the source package is uploaded. I remember that when @Marenz last ran the other PPA script (for solc) he had to choose an Ubuntu version. @leo Do you know how it works?

willianpaixao: I was initially compelled to simply change `Architecture: any` but in this case the source code must be compiled in the target architecture, then packed and uploaded. Meaning if I change and someone runs the script for ARM from an X86 the library files won't work.

As for the PPA, like most modern artifact storage, it can receive multiple packages from multiple architectures. Launchpad won't have any problem receiving packages for ARM.

cameel:
@willianpaixao Do you want to continue working on this? Need any help?

cameel:
I'm going to close this but feel free to reopen if you'd like to resume working on this (or anyone else for that matter). If not, we'll eventually do it ourselves, though at the moment there are just too many topics and this one is not the top priority so it has to wait for its turn.

PR #13306, opened 2022-07-26
Disable SMT tests on Arch Linux

cameel:
This should fix the failing `b_archlinux` job. This way we won't always have to keep our Z3 version in lockstep with Arch repos.


Marenz (approved):
The change looks good, let's see what the CI says

cameel:
Looks like all except OZ is passing so I'm force merging it.

PR #13309, opened 2022-07-27
Fix misspelling in bug name in `bugs.json` (`NestedCallataArrayAbiReencodingSizeValidation`)

yusufbenliii:
Misspelling fix

### DIFF
--- a/docs/bugs.json
+++ b/docs/bugs.json
@@ -60,7 +60,7 @@
     },
     {
         "uid": "SOL-2022-2",
-        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "name": "NestedCalldataArrayAbiReencodingSizeValidation",
         "summary": "ABI-reencoding of nested dynamic calldata arrays did not always perform proper size checks against the size of calldata and could read beyond ``calldatasize()``.",
         "description": "Calldata validation for nested dynamic types is deferred until the first access to the nested values. Such an access may for example be a copy to memory or an index or member access to the outer type. While in most such accesses calldata validation correctly checks that the data area of the nested array is completely contained in the passed calldata (i.e. in the range [0, calldatasize()]), this check may not be performed, when ABI encoding such nested types again directly from calldata. For instance, this can happen, if a value in calldata with a nested dynamic array is passed to an external call, used in ``abi.encode`` or emitted as event. In such cases, if the data area of the nested array extends beyond ``calldatasize()``, ABI encoding it did not revert, but continued reading values from beyond ``calldatasize()`` (i.e. zero values).",
         "link": "https://blog.soliditylang.org/2022/05/17/calldata-reencode-size-check-bug/",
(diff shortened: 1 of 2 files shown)


docs/bugs.json
```diff
@@ -34,7 +34,7 @@
     },
     {
         "uid": "SOL-2022-2",
-        "name": "NestedCallataArrayAbiReencodingSizeValidation",
+        "name": "NestedCalldataArrayAbiReencodingSizeValidation",
```
cameel, line 63: Note that this would have to be changed in more places. For example `bugs_by_version.json` refers to these identifiers.

cameel:
Thanks for the fix but I'm not really sure if we want to change that at this point. It's been there for a long time (~3 years) and I think some tools may not be prepared for changes like that. For example I know that Etherscan displays these identifiers and I'm not sure how it would actually behave.

Personally I'd be for leaving it as is, but we should get more opinions. @chriseth, @axic?

ekpyron:
> Thanks for the fix but I'm not really sure if we want to change that at this point. It's been there for a long time (~3 years) and I think some tools may not be prepared for changes like that. For example I know that Etherscan displays these identifiers and I'm not sure how it would actually behave.
> 
> Personally I'd be for leaving it as is, but we should get more opinions. @chriseth, @axic?

3 years? Hasn't this just been there since April this year?

cameel:
Ah, you're right :) I looked at when first affected version was released rather than when we fixed it. It's actually the reencode bug we fixed not so long ago.

Well, in that case maybe it makes sense to fix it, though I still wonder if some tools might be caching these IDs?

ekpyron:
> Ah, you're right :) I looked at when first affected version was released rather than when we fixed it. It's actually the reencode bug we fixed not so long ago.
> 
> Well, in that case maybe it makes sense to fix it, though I still wonder if some tools might be caching these IDs?

I'm not sure, maybe we should try to reach out - but generally, I'd tend towards fixing it.

cameel:
Yeah, if it's so fresh, I'm for fixing it too. I'll ask on the tooling channel just in case.

cameel (dismissed):

ekpyron (dismissed):

leonardoalt:
@cameel @ekpyron wait, I thought it was decided this wasn't going to be changed?

ekpyron:
> @cameel @ekpyron wait, I thought it was decided this wasn't going to be changed?

We actually decided that we *want* to change this, we just wanted to ping Etherscan to see if this may cause issues for them.

cameel:
The problem is how to reach them. I don't think their support form is the best way to do that for something like this.

cameel:
I got a response from Etherscan guys that fixing this is fine as far as they're concerned. They don't have this hard-coded anywhere.

On the other hand - apparently that's not the only typo we have :) There's a much older one that one's better left as is.

cameel (approved):

Issue #11889, opened 2021-09-03
Should private functions really clash ?


Amxx:
Let consider the following code:

```
contract A {
    function __myPrivateFunction() private {
        // Does something usefull
    }
    
    function callA() public {
        __myPrivateFunction();
    }
}

contract B {
    function __myPrivateFunction() private {
        // Does something usefull
    }
    
    function callB() public {
        __myPrivateFunction();
    }
}

contract AB is A, B {}
```

## What I am expecting

implementation of `__myPrivateFunction` are private, and they only make sense in the context of the contract that they are part of (`A` and `B`). They are not accessible from `AB`, so it is not like if a call from a function within `AB` wouldn't be resolvable.

IMO, this should compile, possibly with a warning, but not with an error.

## What happens

I get an error:

```TypeError: Derived contract must override function "__myPrivateFunction". Two or more base classes define function with same name and parameter types.```

## Question/Request

Is that behaviour wanted? needed? Whould it be possible to accept this kind of ghost-conflicts?

**Comment by chriseth** [comment]
The error message is of course bad because you cannot override the functions.

I'm fine with allowing it, but then without a warning.

**Comment by cameel** [comment]
I also think that such functions should be allowed and without a warning. If `private` functions cannot even be referenced from inherited contracts, they should not clash.

**Comment by chriseth** [comment]
We should be careful if non-private can be changed to private in inheritance.

**Comment by cameel** [comment]
Looks like it can't:
```solidity
contract C {
    function f() internal virtual {}
}

contract D is C {
    function f() private override {}
}
```
```
Error: Overriding function visibility differs.
 --> test.sol:6:5:
  |
6 |     function f() private override {}
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Note: Overridden function is here:
 --> test.sol:2:5:
  |
2 |     function f() internal virtual {}
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

cameel:
Also:
```solidity
contract C {
    function x() external virtual returns (uint) {}
    function y() public virtual returns (uint) {}
    function z() internal virtual returns (uint) {}
}

contract D is C {
    uint private override x;
    uint private override y;
    uint private override z;
}
```
```
Error: Identifier already declared.
 --> test.sol:9:5:
  |
9 |     uint private override y;
  |     ^^^^^^^^^^^^^^^^^^^^^^^
Note: The previous declaration is here:
[...]

PR #13328, opened 2022-07-28
Fix inheritedFunctions returns private functions in OverrideChecker

### LINKED ISSUE #12767: Private function names should not be checked for uniqueness when inherited by other contracts
Closes #11889.

**## Abstract**

Private functions are isolated to the specific contract they are defined in. This holds true during inheritance as well. A contract that inherits another contract cannot use or access private functions from the parent contract. In cases where a contract inherits two separate contracts that each contain the same private function name, there should be no compilation error due to naming conflicts. 

**## Motivation**

Because inheriting contracts cannot use or access the private functions of parent contracts, enforcing uniqueness of these private function names is unnecessarily restrictive.

An example of how this can happen is explained [here](https://github.com/ethereum/solidity/issues/11889#issue-987605750)

**## Specification**

1. Update the [ambiguous overrides checking function](https://github.com/ethereum/solidity/blob/c47da51eab8e5c4ebff0f6dd2ca748179aa21ab4/libsolidity/analysis/OverrideChecker.cpp#L727) in [OverrideChecker](https://github.com/ethereum/solidity/blob/c47da51eab8e5c4ebff0f6dd2ca748179aa21ab4/libsolidity/analysis/OverrideChecker.cpp#L1) to track the scope of functions when comparing them for uniqueness.

2.  Write tests for that case [here](https://github.com/ethereum/solidity/tree/69411436139acf5dbcfc5828446f18b9fcfee32c/test/libsolidity/syntaxTests/inheritance)  

**## Backwards Compatibility**

This change makes solidity compilation strictly more permissive and should not introduce any incompatibilities with existing contracts.

ghost:
Closes #12767 and #11889

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ Compiler Features:
 Bugfixes:
  * Commandline Interface: Disallow the following options outside of the compiler mode: ``--via-ir``,``--metadata-literal``, ``--metadata-hash``, ``--model-checker-show-unproved``, ``--model-checker-div-mod-no-slacks``, ``--model-checker-engine``, ``--model-checker-invariants``, ``--model-checker-solvers``, ``--model-checker-timeout``, ``--model-checker-contracts``, ``--model-checker-targets``.
  * Type Checker: Fix null dereference in `abi.encodeCall` type checking of free function.
+ * Override Checker: Fix `inheritedFunctions` returns private functions.
 
 
 ### 0.8.15 (2022-06-15)
--- a/libsolidity/analysis/OverrideChecker.cpp
+++ b/libsolidity/analysis/OverrideChecker.cpp
@@ -933,7 +933,7 @@ OverrideChecker::OverrideProxyBySignatureMultiSet const& OverrideChecker::inheri
 		{
 			set<OverrideProxy, OverrideProxy::CompareBySignature> functionsInBase;
 			for (FunctionDefinition const* fun: base->definedFunctions())
-				if (!fun->isConstructor())
+				if (!fun->isConstructor() && fun->visibility() != Visibility::Private)
 					functionsInBase.emplace(OverrideProxy{fun});
 			for (VariableDeclaration const* var: base->stateVariables())
 				if (var->isPublic())
--- a/test/libsolidity/semanticTests/scoping/private_function_vs_free_function.sol
+++ b/test/libsolidity/semanticTests/scoping/private_function_vs_free_function.sol
@@ -1,0 +1,10 @@
+function foo(uint256 value) pure returns (uint256) { return 1; }
+contract A {
+    function foo(uint256 value) private pure returns (uint256) { return 2; }
+}
+contract B is A {
+    using {foo} for uint256;
+    function test(uint256 value) public pure returns (uint256) { return value.foo(); }
+}
+// ----
+// test(uint256): 0 -> 1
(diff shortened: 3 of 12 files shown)


ekpyron:
Thank you for the PR!

This is mainly missing (a lot of) semantics tests. We have to make sure that, under any circumstances, the correct functions are called in the presence of multiple private functions in the inheritance graph. Since code generation was not designed with this in mind, we have to be very careful about that - it may *happen* to "just work", but there's no guarantee and we have to make sure with exhaustive tests.

ghost:
Sure. Let me know what other semantic tests you'd like and I'll work on them.

ekpyron:
> Sure. Let me know what other semantic tests you'd like and I'll work on them.

I just tried to :-). But let me try to be clearer :-): The PR so far has no semantic tests whatsoever, but only syntax tests.
You can see examples of semantic tests in ``test/libsolidity/semanticTests/*``. They involve calling some functions in a test contract and checking, if the call produces the desired result. Under the hood those tests are actually compiled down to bytecode and executed in a mock EVM.

We will need test cases like that, for any kind of situation, in which a call to a private function could theoretically pick up the wrong function. As one example:
```
contract A {
  function f() private pure returns (uint) { return 7; }
  function g() internal pure returns (uint) { return f(); }
}
contract B is A {
  function f() private pure returns (uint) { return 42; }
  function test() external pure returns (uint, uint) {
    return (g(), f());
  }
}
// ----
// test() -> 7, 42
```
That would, in the simplest inheritance situation, verify that the internal call to the base contract in fact calls the correct private function, even though the derived contract defines a function of the same name with different implementation. I'd expect this test case to pass, but there may be other cases to consider. The trickiest part in the underlying issue is to identify all distinct situations in which such tests are warranted.

ghost:
This PR includes one semantic test. Thanks for that test, I'll include it. I'll try to write some more semantic tests. I'll also join the Matrix chat tomorrow or Thursday, see if we can discuss what other scenarios need to be tested, which is what I initially did a few weeks ago. In fact, most of the tests included here were described by @cameel.

But as you say, it's tricky to identify all the different situations that need to be tested. The commit barely affects a line of code in `OverrideChecker.cpp`, yet it needs to test a lot of unmodified code.

ekpyron:
Ah, right, I must have missed that one semantics test, sorry about that.

But yeah, testing is the main issue with this, especially since, as you say, the effects of the change are "non-local".

Actually, one option to alleviate the need for exhaustive testing (at least somewhat) would be to add some safety assertion about this during code generation. In particular in

https://github.com/ethereum/solidity/blob/6b60524cfe4186eb7d22d80ca67b9554902d8fb3/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L952

we could assert that if the function definition has ``private`` visibility, it's annotated contract scope (i.e. ``functionDef->annotation().contract``) matches the current most derived contract (i.e. ``m_context.mostDerivedContract()``).
The main issue is that code generation so far is written under the assumption that functions are unique and do *not* have assertions like this - that's what makes this dangerous. We'd need a similar assertion for legacy code generation around https://github.com/ethereum/solidity/blob/6b60524cfe4186eb7d22d80ca67b9554902d8fb3/libsolidity/codegen/ExpressionCompiler.cpp#L627

ghost:
Yes, I like that. I'll fix the new tests and look into adding an assertion during code generation too.

matheusaaguiar:
If anyone is interested in taking over the PR, the changes are available at branch [fixInheritedReturnsPrivateFunction](https://github.com/ethereum/solidity/tree/fixInheritedReturnsPrivateFunction)

Issue #13214, opened 2022-06-25
Add unitialized variable warning when for-loops contain uninitialized variables


lukehutch:
## Description

The following compiles and runs fine in Solidity without any warning:

```solidity
for (uint256 i; i < 10; ++i) { ... }
```

Whereas the following should probably be required:

```solidity
for (uint256 i = 0; i < 10; ++i) { ... }
```

I believe all variables are zero by default in Solidity, so the current behavior is not surprising, but the first of the two examples above gives an uninitialized variable warning in Slither, which is probably appropriate. (Often uninitialized variables are symptomatic of bugs in code, so they should always be initialized.)

Related question -- in the second code example above, is `i = 0` duplicating the work of zeroing out `i`, since `uint256 i` already zeroes it out?

## Environment

- Compiler version: 0.8.14
- Target EVM version (as per compiler settings):
- Framework/IDE (e.g. Truffle or Remix):
- EVM execution environment / backend / blockchain client: Hardhat
- Operating system:  Linux

## Steps to Reproduce

See above

timweri:
I'm not sure what the decision is on implementing this but I would like to take a jab at it.

However, regarding the question, I believe `i = 0` does not zero `i` twice:

`explicit_zero.sol`:
```solidity
contract Test {
    uint256 num;
    function f() public {
        for (uint256 i = 0; i < 10; i++) {
            num++;
        }
    }
}
```

`default_zero.sol`:
```solidity
contract Test {
    uint256 num;
    function f() public {
        for (uint256 i; i < 10; i++) {
            num++;
        }
    }
}
```

Running
```sh
 diff --unified \
    <(solc explicit_zero.sol --optimize --ir-optimized --debug-info none) \
    <(solc default_zero.sol --optimize --ir-optimized --debug-info none)
```
gives
```diff
@@ -1,18 +1,18 @@
 Optimized IR:
-/// @use-src 0:"explicit_zero.sol"
-object "Test_22" {
+/// @use-src 0:"default_zero.sol"
+object "Test_21" {
     code {
         {
             let _1 := memoryguard(0x80)
             mstore(64, _1)
             if callvalue() { revert(0, 0) }
-            let _2 := datasize("Test_22_deployed")
-            codecopy(_1, dataoffset("Test_22_deployed"), _2)
+            let _2 := datasize("Test_21_deployed")
+            codecopy(_1, dataoffset("Test_21_deployed"), _2)
             return(_1, _2)
         }
     }
[...]

PR #13334, opened 2022-07-30
Issue warning for unitialized for-loop initialization expression

StrongerXi:
A simple check that addresses #13214.

New tests are added.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+ * Static Analyzer: Issue warning for uninitialized for-loop initialization expression.
  * TypeChecker: Support using library constants in initializers of other constants.
  * Yul IR Code Generation: Improved copy routines for arrays with packed storage layout.
  * Yul Optimizer: Add rule to convert `mod(mul(X, Y), A)` into `mulmod(X, Y, A)`, if `A` is a power of two.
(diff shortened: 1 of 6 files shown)


hrkrshnn (dismissed):
Thanks!

hrkrshnn:
Actually, can you add a changelog as well? @StrongerXi

StrongerXi:
> Actually, can you add a changelog as well? @StrongerXi

Totally, done.

hrkrshnn (approved):
I'm okay with this warning. However, it seems that we don't currently have warnings for uninitialized variable elsewhere (except for storage pointers), so it's a small deviation from how code works now.

Some users are vocal about warnings that they don't agree with. If that is a concern, we have something a bit gentler, which are info messages. We could convert this into an info message.

Anyway, I'll vote to merge this. But I'll wait for others in the team to chime in.

chriseth (changes requested):
Thank you for your contribution!
I'm not really convinced we need this warning - it is perfectly fine to just use `uint i`.
Also, what is special about loop variables?
In general, I think the compiler only warns about really bad situations and things that could be fine or could be bad should be warned about by additional tools like slither.

ekpyron:
We discussed this in our meeting today and decided that we don't want the underlying issue implemented, so I'm closing the PR.

PR #13340, opened 2022-08-02
[Docs] contract type to address conversion ambiguity fix

niko-renko:
An attempt to reduce ambiguity in the language for type conversions of contract types.

### DIFF
--- a/docs/types/value-types.rst
+++ b/docs/types/value-types.rst
@@ -225,7 +225,7 @@ Operators:
 .. note::
     The distinction between ``address`` and ``address payable`` was introduced with version 0.5.0.
     Also starting from that version, contracts do not derive from the address type, but can still be explicitly converted to
-    ``address`` or to ``address payable``, if they have a receive or payable fallback function.
+    ``address`` or to ``address payable``.
 
 .. _members-of-addresses:
 
@@ -349,13 +349,9 @@ Contract Types
 
 Every :ref:`contract<contracts>` defines its own type.
 You can implicitly convert contracts to contracts they inherit from.
-Contracts can be explicitly converted to and from the ``address`` type.
-
-Explicit conversion to and from the ``address payable`` type is only possible
-if the contract type has a receive or payable fallback function.  The conversion is still
-performed using ``address(x)``. If the contract type does not have a receive or payable
-fallback function, the conversion to ``address payable`` can be done using
-``payable(address(x))``.
+Contracts can be explicitly converted to and from the ``address`` type using ``address(x)``.
+Explicit conversion to and from the ``address payable`` can be done with ``payable(address(x))``, regardless 
+of whether the contract has a receive or a payable fallback function.
 You can find more information in the section about
 the :ref:`address type<address>`.
 


leonardoalt:
This PR will need a rebase after https://github.com/ethereum/solidity/pull/12954 is merged.

docs/types/value-types.rst
```diff
@@ -349,13 +349,9 @@ Contract Types
...
 You can implicitly convert contracts to contracts they inherit from.
-Contracts can be explicitly converted to and from the ``address`` type.
-
-Explicit conversion to and from the ``address payable`` type is only possible
-if the contract type has a receive or payable fallback function.  The conversion is still
-performed using ``address(x)``. If the contract type does not have a receive or payable
-fallback function, the conversion to ``address payable`` can be done using
-``payable(address(x))``.
+Contracts can be explicitly converted to and from the ``address`` type using ``address(x)``.
+Explicit conversion to and from the ``address payable`` can be done with ``payable(address(x))``, regardless
```
nikola-matic, line 353: Whilst this is technically true, from Solidity 0.8.0 and up, if you want to convert a contract into `address payable`, the `payable(x)` approach should be used as it's the safest way, i.e. it will only perform the conversion if the contract to be converted has a receive/fallback function implemented, otherwise, it will throw an error.

Using `payable(address(x))` does not offer such safety, although it does allow any contract to be converted into `address payable`, and should therefore be avoided.

nikola-matic:
Thanks for opening the PR! It looks like the docs in this case were a bit outdated (aside from being somewhat ambiguous). Please address the changes, and then as @leonardoalt mentioned, perform a rebase, and fix the coding style errors. The easiest way to check if you have any coding style errors is to run `./scripts/check_style.sh` locally.

nikola-matic (changes requested):

docs/types/value-types.rst
niko-renko:
```solidity
contract Test {
    uint public test;

    receive() external payable {}

    fallback() external payable {}
}

contract Test2 {
    address payable a = payable(Test);
}
```
Could you elaborate? I can't seem to successfully do such a conversion with just payable(x) as you mentioned. The contract above isn't compiling with solidity 0.8.16.

nikola-matic: The `Test` you're trying to convert via `address payable a = payable(Test);` is a type, not an instance of said type.

Try 
```solidity
contract Test2 {
    Test test = new Test();
    address payable a = payable(test);
}
```

niko-renko: I see, it works. Sorry for the confusion. But I also see that the documentation says the conversion is performed using address(x), and not payable(x)?

leonardoalt: Yea that's for the unsafe conversion. Is there nothing about the safe conversion that @nikola-matic was talking about?

nikola-matic:
Closing due to inactivity. @niko-renko if you feel like completing it in the future based on the change requests, please feel free to re-open.

PR #13342, opened 2022-08-03
Add full stops to error messages

nikola-matic:
This change moves the addition of full stops (periods, dots) from  #12362. So far, full stops were added to 4907, 9767, 7366 and 2271.

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -1601,7 +1601,8 @@ bool TypeChecker::visit(Assignment const& _assignment)
 				" not compatible with types " +
 				t->humanReadableName() +
 				" and " +
-				type(_assignment.rightHandSide())->humanReadableName()
+				type(_assignment.rightHandSide())->humanReadableName() +
+				"."
 			);
 	}
 	return false;
@@ -1723,9 +1724,7 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 	TypeResult result = type(_operation.subExpression())->unaryOperatorResult(op);
 	if (!result)
 	{
-		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->humanReadableName();
-		if (!result.message().empty())
-			description += ". " + result.message();
+		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->humanReadableName() + "." + (!result.message().empty() ? " " + result.message() : "");
 		if (modifying)
 			// Cannot just report the error, ignore the unary operator, and continue,
 			// because the sub-expression was already processed with requireLValue()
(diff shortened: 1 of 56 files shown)


nikola-matic:
@chriseth error 7366 also doesn't contain a full stop, however, this was not touched in your and @wechman 's PR. Should I update that one as well?

ekpyron:
> @chriseth error 7366 also doesn't contain a full stop, however, this was not touched in your and @wechman 's PR. Should I update that one as well?

I'd say yes!

Additionally, I'm actually wondering, if we should also generally make the error messages proper full sentences... i.e. for these that would mean  changing it to "Operator ... **is** not..." - any other opinions on that? Also fine with keeping that as is myself.

nikola-matic:
> > @chriseth error 7366 also doesn't contain a full stop, however, this was not touched in your and @wechman 's PR. Should I update that one as well?
> 
> I'd say yes!
> 
> Additionally, I'm actually wondering, if we should also generally make the error messages proper full sentences... i.e. for these that would mean changing it to "Operator ... **is** not..." - any other opinions on that? Also fine with keeping that as is myself.

I'll update the full stops for 7366 as well. As far as proper full sentences are concerned, most compilers tend to be quite terse (our current style), so this could perhaps be a discussion for another time? Although if @chriseth and yourself think it might improve things, I wouldn't mind implementing it.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1722,9 +1723,7 @@ bool TypeChecker::visit(UnaryOperation const& _operation)
 	TypeResult result = type(_operation.subExpression())->unaryOperatorResult(op);
 	if (!result)
 	{
-		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->humanReadableName();
-		if (!result.message().empty())
-			description += ". " + result.message();
+		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->humanReadableName() + (!result.message().empty() ? ". " + result.message() : ".");
```
wechman: How about making the first full stop a fixed part of the message and the second full stop a part of the `result.message()`?
```suggestion
		string description = "Unary operator " + string(TokenTraits::toString(op)) + " cannot be applied to type " + subExprType->humanReadableName() + "." + (!result.message().empty() ? " " + result.message() : "");
```

nikola-matic: Looks like pretty much the same thing; will this make it easier for you to rebase? :smile:

wechman: Looks similar but the difference is that the full stops do not depend on any logic. So you don't need to care about a full stop when you use `result.message()` somewhere else in code or when `result.message()` returns a string that already contains a full stop. Maybe, it would be a bit easier to rebase too :)

nikola-matic: I've made the changes in two places where it was applicable.

wechman (approved):
LGTM

PR #13343, opened 2022-08-03
update clang-format as it should now support all styles required

christianparpart:
### Checklist

* [x] update .clang-format
* [x] adapt CI to verify style by using clang-format (requires version 14 or above, Ubuntu 20.04 seems to have it)
* [x] update source files

I found a static built of clang-format [here](https://github.com/muttleyxd/clang-tools-static-binaries/releases).
clang-format file documentation is [here](https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html).

### Open question: broken up if-conditions

```cpp
if (
    overlyLongExpression
)
    then();
````

I could not find an equivalent of this in the clang-format documentation. Dangling closing `)` is definitely for other parts, but here I am uncertain.

### Conditionally breaking the rules

It is possible to break the clang-format defined style by surrounding the lines with `// clang-format off` and a `// clang-format on` comment. 
Because it is not clear to me if we want to update the sources respectively or what parts to force as-is (that need the comments as mentioned above), I did not update the sources yet. For sake of reviewing, I could add such a commit applying .clang-format on them.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,4 @@
+---
 # Formatting approximately used in Solidity's C++
 #
 # See https://clang.llvm.org/docs/ClangFormatStyleOptions.html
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .clang-format [hunk 1]
```diff
@@ -7,35 +8,196 @@
...
+AlignAfterOpenBracket: BlockIndent
+AlignArrayOfStructures: None
+AlignConsecutiveMacros: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: None
+AlignConsecutiveDeclarations: None
+AlignEscapedNewlines: Left
+# The old PR actually stated false (DontAlign), but seems like Align is correct.
+# i.e. having operators at EOL and the RHS on the next line.
+AlignOperands: Align
```
ekpyron, line 28: IIUC, ``DontAlign`` was correct. The option is about how far to indent after breaking, isn't it? And we don't want to have operands aligned, but the indent after the operator to be just the usual ``ContinuationIndentWidth``.

But yeah, I just had a quick look at this locally, and there's still a lot off, I'm not sure we can completely fix it.
For example the docs of ``AlignAfterOpenBracket`` in https://clang.llvm.org/docs/ClangFormatStyleOptions.html still state ``Warning Note: This currently only applies to parentheses.`` and we also need it brackets, only to name one...

So given that I'm afraid we'll probably have to postpone this once more.

christianparpart: I agree, we don't have to apply it to the sources and *require* the check. However, maybe just having the .clang-format updated as we've come closer.

ekpyron: I don't really mind updating the file, but then we should make it a strict improvement, and this line here for example was "more correct" before IIUC. But I'd also say, it's not really worth spending too much time on it, so as far as I'm concerned, I'd just leave things as they are (closing the PR) and revisit some other time.

cameel: I think it actually wouldn't be too bad to merge just the config without the CI part. Currently some contributors or new people do try to use it so having it closer to our style is still of some benefit, even if it does not match perfectly.

ekpyron: If we want to merge this, we at least have to change this setting back, though, because that was closer before.

leonardoalt:
so what's the status here? remove CI part? merge as is? close?

ekpyron:
> so what's the status here? remove CI part? merge as is? close?

Changing to ``AlignOperands: Align`` is *still* wrong...

.clang-format
leonardoalt: They've updated `AlignAfterOpenBracket` to `This applies to round brackets (parentheses), angle brackets and square brackets.`.

Issue #13338, opened 2022-08-02
Remove all references to "money"


fulldecent:
I don't believe the Ethereum Foundation has made any reference to classifying Ether as "money" and therefore neither should the Solidity project.

kriptonian1:
I would like to work on this issue

daredevil3435:
Is this issue still open?

NunoFilipeSantos:
Done.

There are only some references in actual solidity files as comments not on Docs.

NunoFilipeSantos:
<img width="1128" alt="Screenshot 2023-05-22 at 12 57 27" src="https://github.com/ethereum/solidity/assets/2582498/0e0f9fe3-b5d4-46a8-8f96-c9f6f3ff87d1">
<img width="1110" alt="Screenshot 2023-05-22 at 12 57 43" src="https://github.com/ethereum/solidity/assets/2582498/753bae74-0f0e-4343-b89d-8b449eb3bb7c">

fulldecent:
cool, thank you

PR #13344, opened 2022-08-03
Changed all instances of 'money' found in the docs

angusjoshi:
Tried to change to an appropriate alternative based on context.
Addressing #13338

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -21,9 +21,12 @@ the :ref:`Standard JSON Interface<compiler-api>`.
 
 You have to publish the metadata file to IPFS, Swarm, or another service so
 that others can access it. You create the file by using the ``solc --metadata``
-command that generates a file called ``ContractName_meta.json``. It contains
-IPFS and Swarm references to the source code, so you have to upload all source
-files and the metadata file.
+command together with the ``--output-dir`` parameter. Without the parameter,
+the metadata will be written to standard output.
+The metadata contains IPFS and Swarm references to the source code, so you have to
+upload all source files in addition to the metadata file. For IPFS, the hash contained
+in the CID returned by ``ipfs add`` (not the direct sha2-256 hash of the file)
+shall match with the one contained in the bytecode.
 
 The metadata file has the following format. The example below is presented in a
 human-readable way. Properly formatted metadata should use quotes correctly,


docs/examples/safe-remote.rst
```diff
@@ -6,18 +6,18 @@ Safe Remote Purchase
 
 Purchasing goods remotely currently requires multiple parties that need to trust each other.
 The simplest configuration involves a seller and a buyer. The buyer would like to receive
-an item from the seller and the seller would like to get money (or an equivalent)
+an item from the seller and the seller would like to get currency
```
ekpyron: Hm... I'm not sure "currency" is that much better than "money"... maybe we should go for "some compensation, e.g. Ether," in the first occurrence here and then move to "Ether" in the further occurrences below (those refer to the example in particular, in which it *is* Ether)...

ekpyron (commented):
Thank you very much! This looks good, except for the one case I commented about below.

docs/examples/safe-remote.rst
angusjoshi: Makes sense, just committed this change.

ekpyron: Can you also change the other occurrences of "currency"? (if I see it correctly "Ether" should fit for those)

angusjoshi1: Shall I change all instances of currency outwith the changes I made as well?

angusjoshi: done.

cameel:
Needs rebase on latest `develop` to have the OZ test pass. And the chainlink one looks just like a transient networking error.

angusjoshi:
done this rebase.

ekpyron:
This looks good now, but from the looks of it you *merged* develop instead of rebasing it :-).
If you can rebase it with ``git rebase -i develop`` and simultaneously "squash" the changes into a single commit, then we can merge (otherwise we can also take care of the rebase).

ekpyron:
Damn it, something went wrong here now...

ekpyron:
I tried to rebase this against the fork (which usually works), but apparently I lost the changes while doing so, I'm very sorry about that...

ekpyron:
@angusjoshi1 Do you happen to still have the changes locally? If so you can just push them again using ``git push -f`` and we can take it from there. Again sorry about my git accident here :-).

Issue #9735, opened 2020-09-03
Restrict free functions


chriseth:
@axic brought up that free function still carry around implicit "state" and should maybe be restricted more.

`this` is already disallowed and probably rightfully so.

Maybe it would make sense to also disallow them making external calls, or at least adding a keyword that tells if they can make external calls (and thus forward ether from the contract).

mudgen:
@chriseth @axic What implicit state?  Values such as msg.sender etc?

Why is `this` disallowed? In some small cases `this` might be useful in free functions.  

People don't have to use `this` in free functions if they don't want to.  I don't like restricting/removing capability on the basis that programmers are too stupid to do the right thing.  It can cause us to implement complicated workarounds to do what we are trying to do.

I really like free functions but if they are limited too much then I probably won't use them.  

But I think making things more explicit is good.

leonardoalt:
A free function naturally is not associated with any contract, therefore, in my opinion, conceptually it doesn't make any sense to allow `this` in free functions, and allowing it would be a hack. To me free functions should be as strict as possible.
As @axic pointed out, if `this` is so important for a free function it can just take it as an argument.

> People don't have to use this in free functions if they don't want to.

"People don't have to declare function mutability/visibility if they don't want to", I can go on with the list of things that were less strict and programmers still did the wrong thing, causing big problems.

mudgen:
I don't think we should assume how people will or should write code.  There may be surprising or unthought of uses and patterns that are very useful that people want to use and do.  

We are not programmers of every possible type of application so I don't think we can assume or know how every program should be written. In addition, different programmers have different styles and abilities and knowledge.   

@leonardoalt I don't think that function mutability/visibility limits much what a programmers does.  It enables a programmer to be explicit and enforce how he wants things to work.   I like a language I use to give me capability, choice, expressiveness, explicitness but not restrict me, unless I want that restriction ( this is choice here).

leonardoalt:
> I don't think that function mutability/visibility limits much what a programmers does. It enables a programmer to be explicit and enforce how he wants things to work.

That's not true, since it's mandatory to specify function visibility, for example (except for constructors). It precisely enforces the programmer to be explicit, it's not a choice.
Similarly, disallowing `this` in free functions and forcing the programmer to pass it via argument enforces explicitness.

> In addition, different programmers have different styles and abilities and knowledge.

[...]

PR #13345, opened 2022-08-04
Update description of `this` in free functions

minaminao:
The use of `this` in free functions is disallowed, but the document has not been updated.

Related Issue: #9735 (I could not immediately find when it was banned.)
>`this` is already disallowed and probably rightfully so.


cameel (dismissed):
Thanks! Looks good.

I rebased the PR to make it pass tests. I'll merge once it passes.

cameel (approved):

PR #13346, opened 2022-08-04
Update description of input parameter type and return type limitations

minaminao:
ABI coder V2 is now enabled by default. It allows us to do the following:
- Multidimensional arrays (including dynamic arrays) can be used as input for external functions.
- Multidimensional arrays (including dynamic arrays) and structs can be used as return values for external functions.

I changed the documentation based on this. (I have already experimented on my machine to make sure it works.)

However, `mapping` is not allowed as a return value before, so I left it as it is.

EDIT: I am not aware whether all multidimensional arrays of all value types are supported by ABI coder V2.

### DIFF
--- a/docs/contracts/functions.rst
+++ b/docs/contracts/functions.rst
@@ -72,16 +72,6 @@ with two integers, you would use something like the following:
 
 Function parameters can be used as any other local variable and they can also be assigned to.
 
-.. note::
-
-  Until version 0.6.0 it was not possible to use a multi-dimensional array or a struct
-  as an input for an :ref:`external function<external-function-calls>`.
-  ``abicoder v2`` made it possible and it's been enabled by default since version 0.8.0
-  (before that you had to enable it with ``pragma abicoder v2;``).
-
-  An :ref:`internal function<internal-function-calls>` can accept a
-  multi-dimensional array or a struct without any restrictions.
-
 .. index:: return array, return string, array, string, array of strings, dynamic array, variably sized array, return struct, struct
 
 Return Variables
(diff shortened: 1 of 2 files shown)


docs/contracts/functions.rst
```diff
@@ -139,12 +129,7 @@ If you use an early ``return`` to leave a function that has return variables,
 you must provide return values together with the return statement.
 
 .. note::
-    You cannot return some types from non-internal functions, notably
-    multi-dimensional dynamic arrays and structs. If you enable the
-    ABI coder v2 by adding ``pragma abicoder v2;``
-    to your source file then more types are available, but
-    ``mapping`` types are still limited to inside a single contract and you
-    cannot transfer them.
+    You cannot return some types from non-internal functions, notably ``mapping``.
```
ekpyron: Stricly speaking, ``mapping``s themselves and any type that recursively contains some mapping (as struct member or array base type). Also strictly speaking, you only can't return them from non-internal functions except library functions (for external library functions it'll work at least theoretically).
But since this wasn't explained in full detail here before anyways, I'd also be fine with the suggested change.

ekpyron (dismissed):
Yep, thanks, must have slipped through when we switched the default!

docs/contracts/functions.rst
minaminao: I'm glad to know the exact specifications. Thank you.

cameel:
The failing OpenZeppelin test is already fixed on latest `develop`. I rebased the PR and pushed to the original branch to make it pass so that we can merge this.

I also added some more tweaks while at it:
- Expanded the note according to @ekpyron's comment but with even more detail
- Removed another mention of v2 not being the default
- Added more consistent index entries for pragmas.

@minaminao Please check if you're fine with these changes.

minaminao:
Looks good. It's helpful to clearly understand the specifications.

cameel (dismissed):
Ok then. I'm approving but I'll also leave it unmerged yet so that @ekpyron can have the final look. I hope I did not miss any of the internal types on the list :)

docs/layout-of-source-files.rst
```diff
@@ -91,18 +91,19 @@ these follow the same syntax used by `npm <https://docs.npmjs.com/cli/v6/using-n
...
 select between the two implementations of the ABI encoder and decoder.
 
 The new ABI coder (v2) is able to encode and decode arbitrarily nested
-arrays and structs. It might produce less optimal code and has not
-received as much testing as the old encoder, but is considered
-non-experimental as of Solidity 0.6.0. You still have to explicitly
-activate it using ``pragma abicoder v2;``. Since it will be
-activated by default starting from Solidity 0.8.0, there is the option to select
+arrays and structs. It might produce less optimal code, but is considered
+non-experimental as of Solidity 0.6.0. It is activated by default starting
```
ekpyron:
```suggestion
arrays and structs. Apart from supporting more types, it involves more extensive validation and safety checks, which may result in higher gas costs, but also heightened security. It is considered
non-experimental as of Solidity 0.6.0 and is enabled by default starting
```

Or something like that... "It might produce less optimal code" is both vague and discouraging and we really don't want people to use v1 anymore.

leonardoalt (dismissed):

leonardoalt:
Rebased and applied suggestions from @ekpyron .

leonardoalt (approved):

PR #13347, opened 2022-08-04
Disable failing zeppelin tests

nikola-matic:


test/externalTests/zeppelin.sh
```diff
@@ -108,6 +108,15 @@ function zeppelin_test
     sed -i "s|it(\('other accounts cannot unpause'\)|it.skip(\1|g" test/token/ERC721/presets/ERC721PresetMinterPauserAutoId.test.js
     sed -i "s|it(\('prevents initialization'\)|it.skip(\1|g" test/proxy/utils/Initializable.test.js
     sed -i "s|it(\('divide by 0'\)|it.skip(\1|g" test/utils/math/Math.test.js
+    sed -i "s|describe(\('to a receiver contract that panics'\)|describe.skip(\1|g" test/token/ERC721/ERC721.behavior.js
```
nikola-matic: Do we have an open issue in zeppelin that I can reference in a comment here?

cameel: What was the error? If it's not similar to either https://github.com/NomicFoundation/hardhat/issues/2453 or https://github.com/NomicFoundation/hardhat/issues/2115 then we should really investigate the cause instead of just disabling it.

nikola-matic:
```
/tmp/ext-test-Zeppelin-JmPeox/ext/node_modules/hardhat/internal/hardhat-network/stack-traces/solidity-errors.js:108
    return new SolidityCallSite(sourceReference.sourceName, sourceReference.contract, sourceReference.function !== undefined
                                                ^

TypeError: Cannot read properties of undefined (reading 'sourceName')
```

This was the issue for all of the failing tests, which to me looks like https://github.com/NomicFoundation/hardhat/issues/2115

cameel: Yep, that's the one. The message changed slightly (it used to complain about `file`) but that's because of changes in Hardhat.

ekpyron (approved):

Marenz (approved):

Marenz:
Back-row approving :laughing: 
image

PR #13351, opened 2022-08-05
Fix stack shuffling oscillation.

ekpyron:
Part of investigating https://github.com/ethereum/solidity/issues/13224 and
to confirm that the issue is still there, since I'm having trouble reproducing it locally, and if the compiler injection works properly for current ENS master, which it doesn't seem to do for me locally.

### DIFF
--- a/libyul/backends/evm/StackHelpers.h
+++ b/libyul/backends/evm/StackHelpers.h
@@ -69,6 +69,8 @@ template<typename ShuffleOperations>
 concept ShuffleOperationConcept = requires(ShuffleOperations ops, size_t sourceOffset, size_t targetOffset, size_t depth) {
 	// Returns true, iff the current slot at sourceOffset in source layout is a suitable slot at targetOffset.
 	{ ops.isCompatible(sourceOffset, targetOffset) } -> std::convertible_to<bool>;
+	// Returns true, iff the slots at the given source offset is junk.
+	{ ops.sourceIsJunk(sourceOffset) } -> std::convertible_to<bool>;
 	// Returns true, iff the slots at the two given source offsets are identical.
 	{ ops.sourceIsSame(sourceOffset, sourceOffset) } -> std::convertible_to<bool>;
 	// Returns a positive integer n, if the slot at the given source offset needs n more copies.
@@ -186,6 +188,12 @@ private:
 	{
 		std::list<size_t> toVisit{_targetOffset};
 		std::set<size_t> visited;
+
+		if (_ops.targetIsArbitrary(_targetOffset))
+		{
+			_ops.pushOrDupTarget(_targetOffset);
+			return true;
+		}
 
 		while (!toVisit.empty())
 		{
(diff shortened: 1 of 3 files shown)


ekpyron:
Yep, looks like something off with the compiler injection: https://app.circleci.com/pipelines/github/ethereum/solidity/25600/workflows/db515520-4bf7-48e8-9b72-1961f8ffe511/jobs/1129043

cameel:
Weird. Could happen with some changes in Hardhat but then it would affect all tests. I wonder how changes in ENS alone could have done that...

One possibility is that the injection works fine and it's just that they added a dependency that ships artifacts of already compiled contracts. The solution then is then to remove them in the script. But here the file is in `artifacts/build-info/` so it does not seem to be the case. If it was a dependency, they'd be inside `node_modules/`.

ekpyron:
Looks like the repo now contains a typescript ``hardhat.config.ts`` that overwrites the ``.js`` file... proper way to fix this would probably be to have that supported in ``force_hardhat_compiler_binary`` and ``force_hardhat_compiler_settings`` (or have dedicated type-script versions of those)...

cameel:
It is supported :) They both accept name of the config file. Try putting the new config file name in `local config_file="hardhat.config.js"` at the top of the script.

These functions detect the type of the config based on the extension and support both JS and TS.

ekpyron:
I did that, and it failed, when I tried, but I can have another look :-).

ekpyron:
```
Assertion failed.

Stack trace:
    test/externalTests/common.sh:335 in function force_hardhat_compiler_settings()
        [[ $config_var_name != "" ]] || assertFail
    test/externalTests/ens.sh:65 in function ens_test()
        force_hardhat_compiler_settings "$config_file" "$(first_word "$SELECTED_PRESETS")"
    test/externalTests/common.sh:556 in function external_test()
        $main_fn
    test/externalTests/ens.sh:93 in function main()
        external_test ENS ens_test
```

So yeah, maybe non-standard or somehow non-uniform. I can have a look into fixing it myself, but I'll probably investigate the actual bug first (I got the case extracted into https://github.com/ethereum/solidity/issues/13224#issuecomment-1209571195 now with the crude ``rm`` workaround ;-)).

cameel:
Ah, that's because in TS the config exports a variable and getting the name of that automatically is tricky (and there seems to be no universal default; it differs wildly). So there's an extra variable (`config_var`) where you need to put it manually. I see that in this case it's called `config`.

So simply add `local config_var="config"` and add `$config_var` parameter where necessary (check e.g. [`yield-liquidator.sh`](https://github.com/ethereum/solidity/blob/develop/test/externalTests/yield-liquidator.sh) to see where it's needed)

libyul/backends/evm/StackHelpers.h
```diff
@@ -314,6 +322,22 @@ class Shuffler
 		// If we still need more slots, produce a suitable one.
 		if (ops.sourceSize() < ops.targetSize())
 		{
+			// If we do not need the current top and pushing another slot would make a position we need to fix unreachable,
+			// pop the current top.
```
ekpyron, line 326: This seems to be enough to fix the issue at hand - but I still need to verify that this can't cause any other oscillation.
And, tricky as it may be, we should probably extract these cases as unit tests...

ekpyron:
Linking this to https://github.com/ethereum/solidity/issues/13721, since this PR contains related work.

PR #13354, opened 2022-08-08
Place calldata to calldata coder sanity checks.

bshastry:
Fuzzer changes to find https://github.com/ethereum/solidity/issues/13249 that is fixed by https://github.com/ethereum/solidity/pull/13282

### DIFF
--- a/test/tools/ossfuzz/abiV2ProtoFuzzer.cpp
+++ b/test/tools/ossfuzz/abiV2ProtoFuzzer.cpp
@@ -66,9 +66,16 @@ DEFINE_PROTO_FUZZER(Contract const& _input)
 		// We don't care about EVM One failures other than EVMC_REVERT
 		solAssert(result->status_code != EVMC_REVERT, "Proto ABIv2 fuzzer: EVM One reverted");
 		if (result->status_code == EVMC_SUCCESS)
-			solAssert(
-				EvmoneUtility::zeroWord(result->output_data, result->output_size),
-				"Proto ABIv2 fuzzer: ABIv2 coding failure found"
-			);
+			if (!EvmoneUtility::zeroWord(result->output_data, result->output_size))
+			{
+				solidity::bytes resultAsBytes;
+				for (size_t i = 0; i < result->output_size; i++)
+					resultAsBytes.push_back(result->output_data[i]);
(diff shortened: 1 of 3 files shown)


test/tools/ossfuzz/abiV2ProtoFuzzer.cpp
```diff
@@ -66,9 +66,16 @@ DEFINE_PROTO_FUZZER(Contract const& _input)
 		// We don't care about EVM One failures other than EVMC_REVERT
 		solAssert(result->status_code != EVMC_REVERT, "Proto ABIv2 fuzzer: EVM One reverted");
 		if (result->status_code == EVMC_SUCCESS)
-			solAssert(
-				EvmoneUtility::zeroWord(result->output_data, result->output_size),
-				"Proto ABIv2 fuzzer: ABIv2 coding failure found"
-			);
+			if (!EvmoneUtility::zeroWord(result->output_data, result->output_size))
+			{
+				solidity::bytes res;
```
cameel:
```suggestion
				solidity::bytes resultAsBytes;
```

test/tools/ossfuzz/protoToAbiV2.h
```diff
@@ -390,6 +390,8 @@ class ProtoConverter
 	/// Contains typed parameter list to be passed to callee functions
 	std::ostringstream m_typedParamsExternal;
 	std::ostringstream m_typedParamsPublic;
+	/// Contains parameter list to be passed to callee functions
+	std::ostringstream m_paramsExternal;
```
cameel: Just for clarify, wouldn't it be better to explicitly call this "untyped"? I.e. currently it sounds like a more general version of `m_typedParamsExternal` but I guess it's not?
```suggestion
	std::ostringstream m_untypedParamsExternal;
```

Or maybe even more specific to suggest that these are parameters in a format suitable to pass into a call, though I have not good suggestion at hand.

cameel (dismissed):
Looks ok to me.

I have two suggestions that would probably have it made a tiny bit easier to me to understand what's happening but they're not very important overall.

ekpyron (approved):

cameel (approved):

PR #13355, opened 2022-08-08
Tidy up and update .gitignore

nikola-matic:
Removed superfluous ignores, added ignore policy for the entire `deps` directory, and added a leading slash `/` where necessary to indicate ignore directory that is found in the root of the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,13 @@
-commit_hash.txt
-prerelease.txt
+/commit_hash.txt
+/prerelease.txt
 
 # Compiled Object files
 *.slo
 *.lo
 *.o
 *.obj
+*.pyc
+__pycache__
 
 # Precompiled Headers
 *.gch
@@ -15,9 +17,6 @@ prerelease.txt
 *.so
 *.dylib
 *.dll
-
-# Fortran module files
-*.mod
 
 # Compiled Static libraries
 *.lai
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -33,14 +32,10 @@ prerelease.txt
...
-docs/_static/robots.txt
-__pycache__
-docs/utils/*.pyc
-/deps/downloads/
-deps/install
-deps/cache
-cmake-build-*/
+/docs/_build
+/docs/_static/robots.txt
+/docs/utils/
```
ekpyron: Did you see anything that might create this? If not we can just drop it.

nikola-matic: Yeah, `/docs/utils/` can probably be removed; `_build` and `_static` are generated via `spinx-build` (I guess) when `docs.sh` is run, so these would have to stay.

cameel: I think this one should not be ignored:
```suggestion
```

Checking history, [it used to contain the lexer](https://github.com/ethereum/solidity/tree/v0.4.26/docs/utils), which is now a standalone library: [pygments-lexer-solidity](https://gitlab.com/veox/pygments-lexer-solidity).

.gitignore
```diff
@@ -50,18 +45,15 @@ cmake-build-*/
 *~
 
 # IDE files
-.idea
-.vscode
-browse.VC.db
-CMakeLists.txt.user
+/.idea
+/.vscode
```
cameel:
```suggestion
/.idea/
/.vscode/
```

.gitignore
cameel: https://github.com/ethereum/solidity/pull/13355#discussion_r940186007

cameel (approved):
Approving, but please do not merge until we're done with the release.

PR #13360, opened 2022-08-08
`update_bugs_by_version.py`: don't fail when the list gets updated

cameel:
The script fails if the file gets updated, which is fine for testing but confusing when you run it manually. I changed the script so that the diffing happens only in CLI tests.

Also, refactored the script to use pathlib.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -355,7 +355,7 @@ function test_via_ir_equivalence()
 
     for yul_file in $(find . -name "${output_file_prefix}*.yul" | sort -V); do
         bin_output_two_stage+=$(
-	    msg_on_error --no-stderr "$SOLC" --strict-assembly --bin "${optimizer_flags[@]}" "$yul_file" |
+        msg_on_error --no-stderr "$SOLC" --strict-assembly --bin "${optimizer_flags[@]}" "$yul_file" |
                 sed '/^Binary representation:$/d' |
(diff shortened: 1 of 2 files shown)


test/cmdlineTests.sh
```diff
@@ -375,8 +375,13 @@ function test_via_ir_equivalence()
...
 ## RUN
 
-echo "Checking that the bug list is up to date..."
-"$REPO_ROOT"/scripts/update_bugs_by_version.py
+SOLTMPDIR=$(mktemp -d)
+printTask "Checking that the bug list is up to date..."
+cp "${REPO_ROOT}/docs/bugs_by_version.json" "${SOLTMPDIR}/original_bugs_by_version.json"
+"${REPO_ROOT}/scripts/update_bugs_by_version.py"
+diff --unified "${SOLTMPDIR}/original_bugs_by_version.json" "${REPO_ROOT}/docs/bugs_by_version.json" || \
+    fail "The bug list in bugs_by_version.json was out of date and has been updated. Please investigate and submit a bugfix if necessary."
```
wechman, line 383: Shouldn't it be: `A bug list in bugs_by_version.json was out of date and has been updated...`?

cameel: Depends. It would be `a` if you mean the list for a specific version but I by "bug list" I really mean the whole file.

wechman (approved):
LGTM

PR #13363, opened 2022-08-08
Release checklist update and tweaks after 0.8.16

cameel:
Some tweaks to the checklist based on our experience with this release.
- Docs get built automatically so looks like we no longer need separate steps for RTD. We should verify that with @chriseth though.
- In practice we never run full `tests.sh`. I updated that point to say only to run the buglist update script.
- Clarification regarding changelog and version commits.
- Warning on the release page and more detailed steps for blog posts.

PR #13370, opened 2022-08-09
Little enhancements to the ppa release script

Marenz:


scripts/release_ppa.sh
```diff
@@ -47,17 +51,34 @@ else
...
 is_release() {
     [[ "${branch}" =~ ^v[0-9]+(\.[0-9]+)*$ ]]
 }
 
-keyid=379F4801D622CDCF
-email=builds@ethereum.org
+# source keyid and email from .release_ppa_auth
+if [ -e .release_ppa_auth ]
+then
+    source .release_ppa_auth
```
ekpyron: Probably makes sense to add this to ``.gitignore`` then, doesn't it?

Marenz: good point

scripts/release_ppa.sh
```diff
@@ -47,17 +51,34 @@ else
     branch=$1
 fi
 
+# Sanity check
+if ! grep ethereum /etc/dput.cf -q; then
+    echo "Error: Missing required entries in /etc/dput.cf (check top comment in $0 for more information)."
+    exit 1
+fi
```
ekpyron: We can either keep this as a crude sanity check - or actually do it properly a few lines down :-).
If ``is_release`` you need ``[ethereum]`` and ``[ethereum-static]`` otherwise you need ``[ethereum-dev]``.

Marenz: updated :)

ekpyron (commented):
Script looks fine-to-merge to me - but also summoning @cameel as master-of-scripts to weigh in if he wants ;-).

But we should probably also adjust the release checklist accordingly - which may require some coordination with https://github.com/ethereum/solidity/pull/13363 :-).

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
-keyid=379F4801D622CDCF
-email=builds@ethereum.org
+# source keyid and email from .release_ppa_auth
+if [ -e .release_ppa_auth ]
+then
+    # shellcheck source=/dev/null
+    source .release_ppa_auth
+fi
+
+if [ -z "$LAUNCHPAD_KEYID" ] || [ -z "$LAUNCHPAD_EMAIL" ]
```
cameel:
```suggestion
if [[ -e .release_ppa_auth ]]
then
    # shellcheck source=/dev/null
    source .release_ppa_auth
fi

if [[ $LAUNCHPAD_KEYID == "" || $LAUNCHPAD_EMAIL == "" ]]
```

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
+fi
+
 packagename=solc
 
-static_build_distribution=impish
+# This needs to be a still active release
+static_build_distribution=focal
 
-DISTRIBUTIONS="focal impish jammy kinetic"
+DISTRIBUTIONS="focal jammy kinetic"
```
cameel, line 80: So we're also updating the distros? I think things like this should have a separate commit to be easily visible in history. They're more than just a refactor/enhancement.

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
+
+checkDputEntries () {
```
cameel:
```suggestion
function validateDputEntries {
```

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
+    if ! grep "${1}" /etc/dput.cf -q; then
```
cameel:
```suggestion
    if ! grep "${1}" /etc/dput.cf --quiet; then
```

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
+# source keyid and email from .release_ppa_auth
+if [ -e .release_ppa_auth ]
+then
+    # shellcheck source=/dev/null
+    source .release_ppa_auth
+fi
+
+if [ -z "$LAUNCHPAD_KEYID" ] || [ -z "$LAUNCHPAD_EMAIL" ]
+then
+    echo "Error: Couldn't find variables \$LAUNCHPAD_KEYID or \$LAUNCHPAD_EMAIL in sourced file .release_ppa_auth (check top comment in $0 for more information)."
```
cameel: This should go to stderr:
```suggestion
    >&2 echo "Error: Couldn't find variables \$LAUNCHPAD_KEYID or \$LAUNCHPAD_EMAIL in sourced file .release_ppa_auth (check top comment in $0 for more information)."
```
You could also import the `fail` function from `scripts/common.sh` and make this more concise:
```bash
[[ $LAUNCHPAD_KEYID != "" && $LAUNCHPAD_EMAIL != "" ]] || \
    fail "Error: Couldn't find variables \$LAUNCHPAD_KEYID or \$LAUNCHPAD_EMAIL in sourced file .release_ppa_auth (check top comment in $0 for more information)."
```

scripts/release_ppa.sh
cameel: I recommend always naming args for readability. Bash does not have named args but I usually simulate that with locals:
```suggestion
    local pattern="$1"

    if ! grep "$pattern" /etc/dput.cf -q; then
```

scripts/release_ppa.sh
```diff
@@ -51,17 +54,43 @@ is_release() {
...
-static_build_distribution=impish
+# This needs to be a still active release
+static_build_distribution=focal
 
-DISTRIBUTIONS="focal impish jammy kinetic"
+DISTRIBUTIONS="focal jammy kinetic"
+
+checkDputEntries () {
+    if ! grep "${1}" /etc/dput.cf -q; then
+        echo "Error: Missing ${1} section in /etc/dput.cf (check top comment in ${0} for more information)."
```
cameel:
```suggestion
        >&2 echo "Error: Missing ${1} section in /etc/dput.cf (check top comment in ${0} for more information)."
```

Also, with `scripts/common.sh` these checks might be concise enough to skip the function completely:
```
grep "\[ethereum\]" /etc/dput.cf --quiet || \
    fail "Error: Missing [ethereum] section in /etc/dput.cf (check top comm1ent in ${0} for more information)."
```
There's also `assertFail` you could use instead of `fail` but I guess this is not appropriate here since not having this section is still user error, right?

scripts/release_ppa.sh
```diff
@@ -9,13 +9,17 @@
...
 ##
-## The gnupg key for "builds@ethereum.org" has to be present in order to sign
-## the package.
-##
 ## It will clone the Solidity git from github, determine the version,
 ## create a source archive and push it to the ubuntu ppa servers.
 ##
-## This requires the following entries in /etc/dput.cf:
+## To interact with launchpad, you need to set the variables $LAUNCHPAD_EMAIL
+## and $LAUNCHPAD_KEYID in the file $PROJECTROOT/.release_ppa_auth to your launchpad email and pgp keyid.
```
cameel: We don't really have this variable so it may be a bit confusing. Maybe replace it with text?
```suggestion
## and $LAUNCHPAD_KEYID in the file .release_ppa_auth in the root directory of the project to your launchpad email and pgp keyid.
```

cameel (commented):
At your service. I humbly report that the pull request is 90% cromulent.

cameel:
As for the checklist, yeah, we should update it but it would be best to merge #13363 first.

scripts/release_ppa.sh
```diff
@@ -34,11 +39,13 @@
...
 ##  login			= anonymous
-
 ##
 ##############################################################################
 
-set -ev
+set -e
+
+# for the "fail" function
+source scripts/common.sh
```
cameel: Just realized that due to this (and the other `source`) this script **must** be called from repo root. This is what we do in other scripts to avoid this:
```bash
REPO_ROOT="$(dirname "$0")/.."

source "${REPO_ROOT}/scripts/common.sh"
```

Though not sure if the `source` is the only problem. Some other stuff in the file might need updating too to work when called from other dirs.

scripts/release_ppa.sh
```diff
@@ -51,17 +58,39 @@ is_release() {
...
-email=builds@ethereum.org
+# source keyid and email from .release_ppa_auth
+if [[ -e .release_ppa_auth ]]
+then
+    # shellcheck source=/dev/null
+    source .release_ppa_auth
+fi
+
+[[ "$LAUNCHPAD_KEYID" != "" && "$LAUNCHPAD_EMAIL" != "" ]] ||\
+    fail "Error: Couldn't find variables \$LAUNCHPAD_KEYID or \$LAUNCHPAD_EMAIL in sourced file .release_ppa_auth (check top comment in $0 for more information)."
```
cameel: It's a good habit to always put a space before `\`. Otherwise you might get two words fused into one without a space if you don't indent the second line :)
```suggestion
[[ "$LAUNCHPAD_KEYID" != "" && "$LAUNCHPAD_EMAIL" != "" ]] || \
    fail "Error: Couldn't find variables \$LAUNCHPAD_KEYID or \$LAUNCHPAD_EMAIL in sourced file .release_ppa_auth (check top comment in $0 for more information)."
```

scripts/release_ppa.sh
```diff
@@ -51,17 +58,39 @@ is_release() {
...
 packagename=solc
 
-static_build_distribution=impish
+# This needs to be a still active release
+static_build_distribution=focal
+
+DISTRIBUTIONS="focal jammy kinetic"
 
-DISTRIBUTIONS="focal impish jammy kinetic"
+function checkDputEntries () {
```
cameel: I'd stick to the `function f {}` syntax. Both for consistency (I was cleaning that up at some point: #12259) and because it's more universal (`function f() {}` does not work in `sh`).

```suggestion
function checkDputEntries {
```

cameel (commented):
A few more suggestions, though nothing critical. Not approving yet only because of the checklist.

scripts/release_ppa.sh
Marenz: In the loop for the distribution the script `cd`s into a tmp dir, so at least after that point nothing needs an update for this.
The only other possible thing I found is my sourcing of the ppa file, which I adjusted accordingly now

scripts/release_ppa.sh
```diff
@@ -51,17 +62,39 @@ is_release() {
     [[ "${branch}" =~ ^v[0-9]+(\.[0-9]+)*$ ]]
 }
 
-keyid=379F4801D622CDCF
-email=builds@ethereum.org
+# source keyid and email from .release_ppa_auth
+if [[ -e .release_ppa_auth ]]
+then
+    # shellcheck source=/dev/null
+    source ${REPO_ROOT}/.release_ppa_auth
```
ekpyron: Does this need quotes?

ekpyron (commented):
Still looks good to me :-).

scripts/release_ppa.sh
ekpyron: Ha, I saw it before the CI run complaining about it failed :-D.

leonardoalt:
@ekpyron this one seems good to go?

ekpyron:
> @ekpyron this one seems good to go?

It's waiting for https://github.com/ethereum/solidity/pull/13363, since it'll require to change the release checklist as well (just doesn't do that right away, since it'd only result in conflicts anyways).

That being said, since https://github.com/ethereum/solidity/pull/13363 is not merged yet and this can be, we can merge here and adjust there.

ekpyron (approved):

PR #13374, opened 2022-08-10
Update the default free memory pointer in Yul.rst

0xpranay:
The solidity docs and [Inline assembly memory management](https://docs.soliditylang.org/en/v0.8.15/assembly.html#memory-management) suggest the actual allocate-able memory starts from `0x80`, defaulting the free memory pointer to `0x80`. 

But the above yul example defaults the free memory pointer to `0x60` in initialisation cases which I'm not sure if it's intended.

### DIFF
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -1162,7 +1162,7 @@ An example Yul Object is shown below:
         code {
             function allocate(size) -> ptr {
                 ptr := mload(0x40)
-                if iszero(ptr) { ptr := 0x60 }
+                if iszero(ptr) { ptr := 0x80 }
                 mstore(0x40, add(ptr, size))
             }
 
--- a/docs/yul.rst
+++ b/docs/yul.rst
@@ -1162,7 +1162,8 @@ An example Yul Object is shown below:
         code {
             function allocate(size) -> ptr {
                 ptr := mload(0x40)
-                if iszero(ptr) { ptr := 0x80 }
+                // Note that Solidity generated IR code reserves memory offset ``0x60`` as well, but a pure Yul object is free to use memory as it chooses.
+                if iszero(ptr) { ptr := 0x60 }
                 mstore(0x40, add(ptr, size))
             }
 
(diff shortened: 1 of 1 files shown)


ekpyron:
I think this deviation may actually be intentional. The documentation of Yul objects describes *in general* how yul objects can be written, not what kind of Yul objects you can expect to be generated from Solidity code - and a pure Yul contract is in no way beholden to the Solidity memory layout. In that sense we could also have used ``0x20`` as offset *for* the free memory pointer here...

That being said, it may be right that this is more confusing than helpful, at least without saying something about it explicitly and I'm not sure that'd be overly beneficial and helpful...

ekpyron:
But yeah, in general this section of the documentation is decidedly *not* about inline assembly Yul, but about full Yul objects.

docs/yul.rst
```diff
@@ -1162,7 +1162,7 @@ An example Yul Object is shown below:
         code {
             function allocate(size) -> ptr {
                 ptr := mload(0x40)
-                if iszero(ptr) { ptr := 0x60 }
+                if iszero(ptr) { ptr := 0x80 }
```
wechman: If we decide to update the code snippet we also may want to have an update in line 1194.

0xpranay: @ekpyron do you think the proposed change is valid or should I close the PR?

ekpyron: I'd suggest to revert the change, i.e. change it back to ``0x60``, but instead add a comment like:
```
// Note that Solidity generated IR code reserves memory offset ``0x60`` as well, but a pure Yul object is free to use memory as it chooses.
```
or something like that. That may clear things up, and *explicitly* clarify what's happening and why.

leonardoalt: @0xpranay would you still like to apply the changes?

0xpranay: Yes. Will do soon.

0xpranay: @leonardoalt applied the changes.

leonardoalt (approved):

docs/yul.rst
leonardoalt: Great, thank you!

Issue #9627, opened 2020-08-14
Make the hard-coded parts of the optimiser sequence configurable


ekpyron:
The CSE run at the very end after the stack compressor actively causes problems in https://github.com/ethereum/solidity/issues/9622

So maybe we should think about making the hard-coded part of the sequence ``suite.runSequence("fDnTOc g", ast);``, resp. the earlier ``suite.runSequence("g", ast);`` configurable as well.

My idea would have been to add ``T``, the literal materializer to the end of the sequence to avoid the problem, but that's simply undone by the CSE without the possibility of preventing that :-).

I think the main reason for it being done like that is the stack compressor being rather special and needing special care or preservation using the phaser... not sure what's the best solution for this, but it'd be good if this allowed more freedom to configure things if needed.

cameel:
Here's the gist of my discussion with @ekpyron on Gitter:

The reason that the stack compressor and the `g` and `fDnTOc g` parts are hard-coded is that it's just final cleanup and we don't want them to disappear when we use a sequence produced by yul-phaser. This part is not meant to be evaluated using the same criteria as the rest of the sequence (i.e. just taking the size of the resulting code into account).

If Stack Compressor is safe to use at arbitrary points in the sequence (it does not have to be effective but must not crash the compiler), it could be assigned a letter and treated as any other step. I just didn't see any practical need for doing it like that and decided to keep it simple. It would need a tiny wrapper to make it accept the same parameters as other steps and the phaser would also need a special case to ignore the Stack Compressor step. Then the hard-coded part could be made a part of the main sequence.

We would just need to document that the cleanup part and the compressor should be included when using a custom sequence. Or, for better usability, it might be best to have two separate options: one to change the first part of the sequence (meant to optimize size) and the other to change the final cleanup. It would make it easier to override one without having to change the other. Having phaser actually treat it like two sequences and use different criteria is actually something that came up in my discussions with @chriseth while I was working on it.

So implementing it is not a problem, it's just a matter of deciding how configurable we want/need it to be.

cameel:
We discussed it today and decided that making stack compressor a step and the whole sequence configurable is a good idea. Also, we want to keep it simple and have just a single option for changing the sequence.

ekpyron:
We should prioritize this, since I think it will help to discuss the IR pipeline with debuggers, since only with this change we can really fine-tine exactly what the optimizer does and doesn't do without intransparent implicit default behaviour.

cameel:
[...]

PR #13376, opened 2022-08-10
Make hardcoded parts of the optimizer sequence configurable

nikola-matic:
Allows usage of colon `:` delimiter in the optimizer sequence to specify optimization steps (before delimiter), and clean up steps (after delimiter). Addresses #9627.


libyul/optimiser/Suite.cpp
```diff
@@ -310,6 +315,10 @@ void OptimiserSuite::validateSequence(string_view _stepAbbreviations)
 			nestingLevel--;
 			assertThrow(nestingLevel >= 0, OptimizerException, "Unbalanced brackets");
 			break;
+		case ':':
+			assertThrow(nestingLevel == 0, OptimizerException, "Cleanup delimiter may only be placed at nesting level zero");
```
nikola-matic: Are these messages OK?

libsolidity/interface/OptimiserSettings.h
```diff
@@ -59,6 +59,8 @@ struct OptimiserSettings
 		"]"
 		"jmul[jul] VcTOcul jmul";      // Make source short and pretty
 
+	static char constexpr DefaultYulOptimiserCleanupSteps[] = "fDnTOc";
```
nikola-matic, line 62: Since the above default optimization sequence is covered with appropriate comments, I'd like to comment this one as well. All I need is for someone to tell me what this step actually does :smile:

ekpyron (commented):
Without having looked at any details, the biggest point I'm noticing is that this is missing the analog change in the standard-json compiler interface.

Have a look at this command line test input:
https://github.com/ethereum/solidity/blob/develop/test/cmdlineTests/standard_optimizer_yulDetails_optimiserSteps_nested_brackets/input.json

That's standard json; see also: https://docs.soliditylang.org/en/develop/using-the-compiler.html#compiler-input-and-output-json-description

If I see it correctly, adding a colon in the sequence there, won't yet work.

This stuff is parsed in ``libsolidity/interface/StandardCompiler.cpp``, in particular ``checkOptimizerDetailSteps`` will need some adjustment.

nikola-matic:
@ekpyron Nice, I remember there was some mentioned of JSON, but I forgot to ask. Will fix.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -618,6 +655,8 @@ std::variant<OptimiserSettings, Json::Value> parseOptimizerSettings(Json::Value
 				return *error;
 			if (auto error = checkOptimizerDetailSteps(details["yulDetails"], "optimizerSteps", settings.yulOptimiserSteps))
 				return *error;
+			if (auto error = checkOptimizerDetailCleanupSteps(details["yulDetails"], "optimizerSteps", settings.yulOptimiserCleanupSteps))
+				return *error;
```
```suggestion
			if (auto error = checkOptimizerDetailSteps(details["yulDetails"], "optimizerSteps", settings.yulOptimiserSteps, settings.yulOptimiserCleanupSteps))
				return *error;
```

It'd probably be more straightforward and entail less duplication to handle both of them together, wouldn't it?

nikola-matic: Yup. Wasn't sure which approach to take; on one hand it would certainly be less redundant to simply extend the function prototype and accept two output strings (for settings), but on the other hand, all of these 'helper' functions handle one field, and so the current implementation 'fits' better with the rest of the codebase. I'm fine with either, so I'll either change it or leave it as is depending on the consensus. @cameel what do you think?

ekpyron: You can turn that around: all the other functions handle one json field entirely, while this, if split into two functions, handles only half of the json field :-). So it's *some* non-uniformity either way, hence I'd argue the less redundant one is better :-).

libsolidity/interface/OptimiserSettings.h
ekpyron: Hm... yeah... what *does* it actually do :-D... ``BlockFlattener``, ``DeadCodeEliminator``, ``ControlFlowSimplifier``, ``LiteralRematerialiser``, ``ForLoopConditionOutOfBody``, ``CommonSubexpressionEliminator``...

It basically cleans up the mess the ``StackCompressor`` leaves behind :-).
We should maybe actually pretty much say that:

```
// Cleanup after stack compression.
```

ekpyron: Unfortunately, that's only half-true, since only the "legacy code transform" pipeline actually runs the stack compressor before this (the newer "optimized code transform" gives better feedback to allow for a more precise, directed stack compression, which doesn't require any cleanup afterwards)...

ekpyron: Ah, well, it has a descriptive name, let's just skip the comment.
It's basically part of the ``Make source short and pretty`` sequence - we just couldn't keep it in there, since the stack compressor would destroy it again.

libyul/optimiser/Suite.cpp
ekpyron: Yep, sounds fine!

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -490,7 +490,16 @@ std::optional<Json::Value> checkOptimizerDetailSteps(Json::Value const& _details
...
+			if (
+				auto const delimiterPos = _details[_name].asString().find(":");
+				delimiterPos != string::npos
+			)
+			{
+				_optimiserSetting = _details[_name].asString().substr(0, delimiterPos);
+				_cleanupSetting = _details[_name].asString().substr(delimiterPos + 1);
+			}
+			else
+				_optimiserSetting = _details[_name].asString();
```
ekpyron: Maybe in this case, this is actually shorter and nicer :-):
```suggestion
			string fullSequence = _details[_name].asString();
			auto const delimiterPos = fullSequence.find(":");
			_optimiserSetting = fullSequence.substr(0, delimiterPos);
			if (delimiterPos != string::npos)
				_cleanupSetting = fullSequence.substr(delimiterPos + 1);
```
We could still eliminate one

ekpyron (commented):
Generally looks good to me - I'm wondering if there's a nicer way to test this (i.e. in which it's more obvious that the cleanup sequence is there or not), though...

Maybe
```
contract C {
        constructor() payable {
                assembly ("memory-safe") {
                        let a := 0
                        revert(0, a)
                }
        }
}
```

with ``--yul-optimizations "iDu"`` vs ``--yul-optimizations "iDu:"``... that should not contain that much noise, but should still show the difference (with the cleanup you should either get two zeroes or twice the same variable as arguments to ``revert``, while without the cleanup it should stay at one zero and one variable).

nikola-matic:
@ekpyron , regarding the testing suggestion: I generally do agree with the sentiment, but this should probably be done as part of a separate issue since it would require fiddling around with all of the `yul_optimizer_steps*` tests, as all of them just use an empty contract as input
```
pragma solidity >=0.0;

contract C
{
	function f() public pure {}
}
```
In fact, the vast majority of the cmdline tests that cover optimization sequence(s) are there mostly to verify the parsing of the sequence, and not so much to test the actual behaviour of the compiler when using a specific sequence; having said that, the tests did catch a bug of mine early on because the bytecode output didn't match, so I guess they're not that bad :)

ekpyron:
> @ekpyron , regarding the testing suggestion: I generally do agree with the sentiment, but this should probably be done as part of a separate issue since it would require fiddling around with all of the `yul_optimizer_steps*` tests, as all of them just use an empty contract as input
> 
> ```
> // SPDX-License-Identifier: GPL-3.0
> pragma solidity >=0.0;
> 
> contract C
> {
> 	function f() public pure {}
> }
> ```
> 
> In fact, the vast majority of the cmdline tests that cover optimization sequence(s) are there mostly to verify the parsing of the sequence, and not so much to test the actual behaviour of the compiler when using a specific sequence; having said that, the tests did catch a bug of mine early on because the bytecode output didn't match, so I guess they're not that bad :)

Ehm... no :-D?

Obviously, the tests should make it apparent that the added feature actually works, and the current ones do not. The preexisting optimizer step tests do that as well (note for example that ``yul_optimizer_steps`` specifically does not contain the inliner, but does contain the unused pruner, so if you know what to look for, the test makes it apparent on first sight that the functionality works).

So stop being lazy and add the proper tests ;-P!

ekpyron (changes requested):
Apart from the adjustment in the tests, this will also need a Changelog entry and documentation changes (at the very least in the standard-json option reference).

nikola-matic:
>so if you know what to look for, the test makes it apparent on first sight that the functionality works

Well, there's my problem :D. Alright, I'll change the tests, and ping back if I have some questions.

Changelog.md
```diff
@@ -7,7 +7,7 @@ Language Features:
 
 
 Compiler Features:
-
+ * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (`:`) in the sequence string
```
```suggestion
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (`:`) in the sequence string.
```
:-)

docs/yul.rst
```diff
@@ -1248,7 +1248,7 @@ You can override this sequence and supply your own using the ``--yul-optimizatio
 
 .. code-block:: sh
 
-    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul'
+    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul:fDnTOc'
```
ekpyron: I think we should either keep that as before and merely *mention* the delimiter below - or we need to rephrase the paragraph below.

If we keep the delimiter in the example, below should probably rather be something like:

The colon delimiter (``:``) in the example is used to specify a custom cleanup sequence, which is run after the main part of the sequence before the delimiter (and, when using the legacy EVM code transform, only *after* the stack compressor). If no such delimiter is present, the default cleanup sequence (``fDnTOc``) is run after the supplied sequence.


or so...

nikola-matic: I'd like to keep the delimiter in the example, since it's now a part of the optimization sequence. I don't really see what the rephrasing actually changes in comparison to the current wording; is it just for style?

ekpyron: Without the change the phrasing is contrary to the example...
Then the example *does* have the delimiter and *does not* run any kind of default sequence and only the case that is *not* shown, in which there was *no* delimiter would run a default sequence. That's the difference.
Having the delimiter in the example and still just starting off with the behaviour as if it *wasn't* there, is confusing.

test/cmdlineTests/yul_optimizer_steps_with_empty_cleanup_sequence/input.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+pragma abicoder v2;
+
+contract C {
+    constructor() payable {
+        assembly ("memory-safe") {
+            let a := 0
+            revert(0, a)
```
ekpyron: Since the intention of this may not be completely clear, we could consider adding some comments like:

```suggestion
            // Without the cleanup sequence this will not be simplified to ``revert(a, a)``.
            revert(0, a)
```
here and analogously
```
// The cleanup sequence will simplify this to ``revert(a,a)``
```
in the other cases.

But also fine without such comments.

ekpyron (dismissed):
Looks good to me now, but we can give @cameel a chance to comment before merging.

test/cmdlineTests/yul_optimizer_steps_with_empty_cleanup_sequence/input.sol
nikola-matic: Makes sense, will add these comments.

ekpyron:
Needs rebase.

ekpyron (dismissed):
Reapproving after rebase.

leonardoalt:
@ekpyron do you still wanna wait for @cameel here or should we merge?

docs/internals/optimizer.rst
```diff
@@ -325,11 +325,15 @@ the ``--yul-optimizations`` option:
 
 .. code-block:: bash
 
-    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul'
+    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul:fDnTOc'
 
 The sequence inside ``[...]`` will be applied multiple times in a loop until the Yul code
 remains unchanged or until the maximum number of rounds (currently 12) has been reached.
 
+The colon delimiter ``:`` is used to supply a custom clean up sequence in order to replace the
```
cameel:
```suggestion
The colon delimiter ``:`` is used to supply a custom clean-up sequence in order to replace the
```

docs/internals/optimizer.rst
```diff
@@ -325,11 +325,15 @@ the ``--yul-optimizations`` option:
...
+default (``fDnTOc``) one, which is run after the stack compressor when using the legacy EVM
```
cameel: Let's not specify the default here. We won't remember to update it when it changes.
```suggestion
default one, which is run after the stack compressor when using the legacy EVM
```
Same in ``yul.rst``.

docs/internals/optimizer.rst
```diff
@@ -325,11 +325,15 @@ the ``--yul-optimizations`` option:
...
+code transform.
```
cameel: It sounds a bit as if the cleanup sequence was only used for the legacy code transform.

docs/internals/optimizer.rst
```diff
@@ -325,11 +325,15 @@ the ``--yul-optimizations`` option:
 
 .. code-block:: bash
 
-    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul'
+    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul:fDnTOc'
 
 The sequence inside ``[...]`` will be applied multiple times in a loop until the Yul code
 remains unchanged or until the maximum number of rounds (currently 12) has been reached.
```
cameel: I see that we have two sections about the sequence (here and in `yul.rst`). We should combine them into one and have one place link to the other.

docs/internals/optimizer.rst
cameel: I think we should mention whether `:` is mandatory or not and whether omitting it is different than just leaving the part before/after it empty.

Also, I think we should mention that there are still some hard-coded steps before/after the sequences.

libsolidity/interface/OptimiserSettings.h
```diff
@@ -146,6 +148,11 @@ struct OptimiserSettings
 	/// them just by setting this to an empty string. Set @a runYulOptimiser to false if you want
 	/// no optimisations.
 	std::string yulOptimiserSteps = DefaultYulOptimiserSteps;
+	/// Sequence of clean up optimisation steps after the above (default or user supplied) set of
```
cameel: Better to refer to the field by name. The order might change.
```suggestion
	/// Sequence of clean-up optimisation steps to run after yulOptimiserSteps.
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -490,7 +490,12 @@ std::optional<Json::Value> checkOptimizerDetailSteps(Json::Value const& _details
 				);
 			}
 
-			_setting = _details[_name].asString();
+			std::string const fullSequence = _details[_name].asString();
```
cameel: This file has `using namespace std;`
```suggestion
			string const fullSequence = _details[_name].asString();
```

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -490,7 +490,12 @@ std::optional<Json::Value> checkOptimizerDetailSteps(Json::Value const& _details
...
+			auto const delimiterPos = fullSequence.find(":");
+			_optimiserSetting = fullSequence.substr(0, delimiterPos);
+
+			if (delimiterPos != string::npos)
+				_cleanupSetting = fullSequence.substr(delimiterPos + 1);
```
cameel, line 498: You're assuming that you're always dealing with a fresh settings object that has `_cleanupSetting` initialized with the default string. If you're not, omitting `:` will leave the previous cleanup string in the object instead of using the default.

The assumption is probably correct but I think you should have an assert to catch the problem if anyone tries to use this function without realizing that:
```c++
else
    solAssert(_cleanupSetting == DefaultYulOptimiserCleanupSteps);
```
Actually, that should probably even go to the beginning of the function as a precondition.

libyul/optimiser/Suite.cpp
```diff
@@ -139,7 +140,10 @@ void OptimiserSuite::run(
 			_optimizeStackAllocation,
 			stackCompressorMaxIterations
 		);
-	suite.runSequence("fDnTOc g", ast);
+
+	// Run the user supplied (otherwise default) clean up sequence
```
cameel: Well, from the point of view of this function it's always user-supplied :)
```suggestion
	// Run the user-supplied clean up sequence
```

libyul/optimiser/Suite.cpp
cameel:
```suggestion
			assertThrow(nestingLevel == 0, OptimizerException, "Colon delimiter may only be placed at nesting level zero");
```
Actually, "cleanup sequence delimiter" does not sound bad but then you should call it that consistently everywhere.

libyul/optimiser/Suite.cpp
```diff
@@ -310,6 +315,10 @@ void OptimiserSuite::validateSequence(string_view _stepAbbreviations)
 			nestingLevel--;
 			assertThrow(nestingLevel >= 0, OptimizerException, "Unbalanced brackets");
 			break;
+		case ':':
+			assertThrow(nestingLevel == 0, OptimizerException, "Cleanup delimiter may only be placed at nesting level zero");
+			assertThrow(++colonDelimiters <=1, OptimizerException, "Too many colon delimiters");
```
cameel: I'm really not a fan of having things like `++` or `=` in expressions. Can we do it this way like in the other cases?
```suggestion
			++colonDelimiters;
			assertThrow(colonDelimiters <= 1, OptimizerException, "Too many colon delimiters");
```
I think it's much more readable. You don't even have to think whether the value in the condition has already been incremented or not. It's also very easy to see that you're actually changing the value and not just checking a condition).

libyul/optimiser/Suite.cpp
cameel: Also, the current message is not bad, but I think that for the user it might be clearer to emphasize the brackets rather than the exact nesting level:
```suggestion
			assertThrow(nestingLevel == 0, OptimizerException, "Cleanup delimiter cannot be nested inside the brackets");
```

solc/CommandLineParser.cpp
```diff
@@ -266,7 +266,14 @@ OptimiserSettings CommandLineOptions::optimiserSettings() const
...
 	if (optimizer.yulSteps.has_value())
-		settings.yulOptimiserSteps = optimizer.yulSteps.value();
+	{
+		std::string const fullSequence = optimizer.yulSteps.value();
+		auto const delimiterPos = fullSequence.find(":");
+		settings.yulOptimiserSteps = fullSequence.substr(0, delimiterPos);
+
+		if (delimiterPos != string::npos)
+			settings.yulOptimiserCleanupSteps = fullSequence.substr(delimiterPos + 1);
+	}
```
cameel: This the same code as in `StandardCompiler.cpp` so my comments from there apply here too.

libsolidity/interface/StandardCompiler.cpp
```diff
@@ -472,7 +472,7 @@ std::optional<Json::Value> checkOptimizerDetail(Json::Value const& _details, std
 	return {};
 }
 
-std::optional<Json::Value> checkOptimizerDetailSteps(Json::Value const& _details, std::string const& _name, string& _setting)
+std::optional<Json::Value> checkOptimizerDetailSteps(Json::Value const& _details, std::string const& _name, string& _optimiserSetting, string& _cleanupSetting)
```
cameel, line 475: There's one very important bit missing: the cleanup string is not included in the metadata (see `CompilerStack::createMetadata()`). This is basically the reverse of what we have here. Here we decode and validate the optimizer settings and `createMetadata()` is supposed to serialize them back:
https://github.com/ethereum/solidity/blob/1cfd01c1065ea95468624a28cc57ef9c964bbd31/libsolidity/interface/CompilerStack.cpp#L1538-L1543

Without it being in the metadata, it would now no longer be possible to recreate the same exact bytecode just from metadata. Actually, we really should have some tests trying to do just that but that's out of the scope of this PR. I'll create an issue for that.

Please also grep the repo for `yulOptimiserSteps` and check if there are other places where the cleanup sequence should have been added and was not.

test/cmdlineTests/standard_optimizer_yulDetails_optimiserSteps_with_empty_cleanup_sequence/input.json
```diff
@@ -0,0 +1,16 @@
...
+    "language": "Solidity",
+    "sources": {
+        "A": {"urls": ["standard_optimizer_yulDetails_optimiserSteps_with_cleanup_sequence/in.sol"]}
+    },
+    "settings": {
+        "optimizer": {
+            "details": {
+                "yul": true,
+                "yulDetails": {
+                    "optimizerSteps": "dhfoDgvulfnTUtnIf\n[    xarrscLM\n]\njmuljuljul VcTOcul jmul:"
```
cameel, line 11: One thing that would be really nice to have covered by tests is the resulting sequence here. We want to be sure that this clears the cleanup sequence and does not use the default. Currently we won't spot that because these tests do not have any outputs selected that would depend on the sequence.

One way to do that would be to request the `metadata` output since that includes the parsed and reserialized sequence. The problem is that metadata contains the compiler version string, which changes based on build configuration so the test will be passing on some configurations in CI but not others. We do have some limited [metadata stripping in `cmdlineTests.sh`](https://github.com/ethereum/solidity/blob/v0.8.16/test/cmdlineTests.sh#L189-L241) exactly for that reason but that only applies to bytecode and strips the whole metadata anyway. Would be nice to extend that so that it can remove the version field - but that would be better off as a separate PR.

A better option would be to add some boost tests in [`test/solc/CommandLineParser.cpp`](https://github.com/ethereum/solidity/blob/develop/test/solc/CommandLineParser.cpp). There you can pass in parameters and inspect the resulting settings object.

cameel: Looking at the non-JSON tests I see it does get tested to some extent because there we have the optimized IR output. In that case we should probably request the optimized IR output here too. I'd make the input snippets identical to CLI ones as well.

Still, I think that having this tested in `test/solc/CommandLineParser.cpp` would be a good idea. It's much easier to notice that the sequence got mangled than that the IR output went through unexpected optimizer steps.

test/cmdlineTests/yul_optimizer_steps_with_empty_cleanup_sequence/input.sol
cameel: I think it would make more sense to put the comment above the line, not below it :)

test/cmdlineTests/yul_optimizer_steps_short_sequence/input.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+pragma abicoder v2;
+
+contract C {
+    constructor() payable {
+        assembly ("memory-safe") {
+            let a := 0
+            revert(0, a)
```
cameel, line 9:
```suggestion
            // With the default cleanup sequence we expect this to be simplified to ``revert(a, a)``.
            revert(0, a)
```

test/cmdlineTests/yul_optimizer_steps_with_cleanup_sequence/input.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.0;
+pragma abicoder v2;
+
+contract C
+{
+    constructor() {}
+}
```
cameel: Why not use the same code as in the other example so that we know exactly what to expect with that sequence?

Also, might be a good idea to use a cleanup sequence that's not the default so that the effects are different.

cameel (changes requested):
Here's my feedback. Mostly minor adjustments regarding docs/style/tests.

But I did also find one important problem (https://github.com/ethereum/solidity/pull/13376#discussion_r950412127): the cleanup sequence is not serialized into metadata and that will potentially break bytecode reproducibility for [Sourcify](https://sourcify.dev).

docs/internals/optimizer.rst
nikola-matic: I'll refer to it as `cleanup`, as that's what it's been referred to throughout the doc; even the section describing it is called `Cleanup`. If you want me to change all instances to `clean-up`, please ping. Whatever we choose, I'd prefer to keep it consistent.

docs/internals/optimizer.rst
nikola-matic: Describing everything in `optimizer.rst` and then linking from `yul.rst` would make the most sense to me in this case. Thoughts?

libsolidity/interface/StandardCompiler.cpp
nikola-matic: Force of habit

libsolidity/interface/StandardCompiler.cpp
nikola-matic: In this case, I would assume I'd be adding a new field (to match OptimiserSettings), i.e.
```c++
details["yulDetails"]["optimizerSteps"] = m_optimiserSettings.yulOptimiserCleanupSteps;
```
?

libsolidity/interface/StandardCompiler.cpp
nikola-matic: I believe this is incorrect, as we initialise the `OptimiserSettings` object one function up the call chain (`parseOptimizerSettings`), so it's always guaranteed to be default initialized. The assert would therefore never be triggered, so even though it wouldn't cause any harm, I'd prefer to omit it.

libyul/optimiser/Suite.cpp
nikola-matic: Cleanup sequence delimiter it is.

docs/internals/optimizer.rst
nikola-matic: True, I'll omit the mention of the stack compressor completely.

stackenbotten:
There was an error when running `chk_coding_style` for commit `9d9cd255d45f32ae3f096a906113714e35737295`:
```
docs/internals/optimizer.rst:322:Detailed information regrading the optimization sequence as well a list of abbreviations is 

```

test/cmdlineTests/yul_optimizer_steps_with_cleanup_sequence/input.sol
nikola-matic: This is tested in `yul_optimizer_steps_with_empty_optimization_sequence`, where we have an empty optimisation sequence (so just the hardcoded steps) with a specified and non-default clean-up sequence (`iDu`). This particular test was originally meant to be more of a sequence parsing tests than anything else, but now that I have those as unit tests (`CommandLineParser`), perhaps we can even delete this one?

Or can you suggest a specific optimisation sequence with a specific input that would make sense for this?

libsolidity/interface/CompilerStack.cpp
```diff
@@ -1540,6 +1540,7 @@ string CompilerStack::createMetadata(Contract const& _contract, bool _forIR) con
 			details["yulDetails"] = Json::objectValue;
 			details["yulDetails"]["stackAllocation"] = m_optimiserSettings.optimizeStackAllocation;
 			details["yulDetails"]["optimizerSteps"] = m_optimiserSettings.yulOptimiserSteps;
+			details["yulDetails"]["optimizerCleanupSteps"] = m_optimiserSettings.yulOptimiserCleanupSteps;
```
ekpyron: Yeah, damn, I really shouldn't have missed this earlier, shame on me :-)...

But wouldn't it be better to serialize this into ``optimizerSteps`` with a colon, though? That way consuming the metadata would stay compatible, wouldn't it?
If we add a new field, this needs specific support.

ekpyron: As for some more context: given some deployed bytecode, tools like sourcify fetch the metadata for the contract, which contains source references and compiler settings - then those sources are fetched and recompiled with the respective settings and it's confirmed that the resulting bytecode is identical to the one you started from (there's some additional pains involved in the process, but basically).

So if we add a new field here, such tools would need to become aware of it to correctly feed it back to the compiler for recompilation. If we just serialize it into ``optimizerSteps`` with the delimiter, though, the current logic of passing it to the compiler should just work - that's why I'd say that'd be preferable.

nikola-matic: Should be trivial, i.e. what you're suggesting is?
```c++
details["yulDetails"]["optimizerSteps"] = m_optimiserSettings.yulOptimiserSteps + ":" + m_optimiserSettings.yulOptimiserCleanupSteps;
```
Now, if I change it, will @cameel be OK with that?

ekpyron: That's what I'm suggesting, yeah :-).

nikola-matic: Sure thing

nikola-matic: Changed + added boost tests

libyul/optimiser/Suite.cpp
```diff
@@ -139,7 +140,10 @@ void OptimiserSuite::run(
 			_optimizeStackAllocation,
 			stackCompressorMaxIterations
 		);
-	suite.runSequence("fDnTOc g", ast);
+
+	// Run the user-supplied clean up sequence
+	suite.runSequence(_optimisationCleanupSequence, ast);
+	suite.runSequence("g", ast);
```
aarlt, line 149: Why not let the user decide whether `g` is part of the cleanup sequence? (did not check the semantic of `g` yet)

nikola-matic: Because @ekpyron said so :)

aarlt: I think in general there should be no implicit settings, e.g. if a cleanup sequence was defined.. I'm also wondering about the part that is happening from line `123` on `suite.runSequence("hgfo", ast);`. What if someone really wants to run a sequence, where e.g. `hgfo` should explicitly not be done before? I could imagine that this could be important for testing purposes, but probably its also not needed at all.

libyul/optimiser/Suite.cpp
```diff
@@ -310,6 +315,11 @@ void OptimiserSuite::validateSequence(string_view _stepAbbreviations)
 			nestingLevel--;
 			assertThrow(nestingLevel >= 0, OptimizerException, "Unbalanced brackets");
 			break;
+		case ':':
+			++colonDelimiters;
+			assertThrow(nestingLevel == 0, OptimizerException, "Cleanup sequence delimiter cannot be placed inside the brackets");
+			assertThrow(colonDelimiters <=1, OptimizerException, "Too many cleanup sequence delimiters");
```
aarlt: format `colonDelimiters <=1` -> `colonDelimiters <= 1`

libyul/optimiser/Suite.cpp
aarlt: E.g. these are hardcoded parts of the optimizer sequence configurable that are not configurable.

libyul/optimiser/Suite.cpp
nikola-matic: Good spot, shame `check_style` didn't catch this.

libyul/optimiser/Suite.cpp
aarlt: Maybe if an optimiser-sequences where given by the user - no implicit settings should be used.

ekpyron: ``g`` is the ``FunctionGrouper`` - which just creates some canonical form of the AST without really changing much. I.e. it brings the AST into the form ``{ { I... } F... }`` (with ``{ I... }`` being the main yul block and ``F...`` are all the functions). It has to remain hardcoded here, since some optimizer steps break that form, but the StackCompressor and StackLimitEvader rely on it (and throw otherwise).

But yeah, the step is extremely cheap, can be executed multiple times without anything bad happening and absolutely doesn't change anything substantial - it's just there to keep the other steps simpler.

Changelog.md
```diff
@@ -9,6 +9,7 @@ Language Features:
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (`:`) in the sequence string.
```
cameel: A single backtick would work fine but for whatever reason our convention is to use double backticks here. Maybe because they work the same way in markdown and rst, while in rst single backticks are for references.
```suggestion
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
```

cameel: Now that we released 0.8.17 this needs to be moved to 0.8.18.

docs/internals/optimizer.rst
```diff
@@ -323,18 +323,8 @@ on the individual steps and their sequence below.
...
-.. code-block:: bash
-
-    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul'
-
-The sequence inside ``[...]`` will be applied multiple times in a loop until the Yul code
-remains unchanged or until the maximum number of rounds (currently 12) has been reached.
-
-Available abbreviations are listed in the :ref:`Yul optimizer docs <optimization-step-sequence>`.
+Detailed information regrading the optimization sequence as well a list of abbreviations is
+available in the :ref:`Yul optimizer docs <optimization-step-sequence>`.
```
cameel: This is the optimizer docs :) `yul.rst` documents the assembly (i.e. Yul):
```suggestion
available in the :ref:`assembly documentation <optimization-step-sequence>`.
```

docs/yul.rst
```diff
@@ -1245,19 +1245,26 @@ In Solidity mode, the Yul optimizer is activated together with the regular optim
 Optimization Step Sequence
 --------------------------
 
-By default the Yul optimizer applies its predefined sequence of optimization steps to the generated assembly.
-You can override this sequence and supply your own using the ``--yul-optimizations`` option:
+By default the optimizer applies its predefined sequence of optimization steps to
+the generated assembly. You can override this sequence and supply your own using
+the ``--yul-optimizations`` option:
```
cameel: Best to break the line after the period:
```suggestion
By default the optimizer applies its predefined sequence of optimization steps to the generated assembly.
You can override this sequence and supply your own using the ``--yul-optimizations`` option:
```

docs/yul.rst
```diff
@@ -1245,19 +1245,26 @@ In Solidity mode, the Yul optimizer is activated together with the regular optim
...
 
-.. code-block:: sh
+.. code-block:: bash
+
+    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul:fDnTOc'
+
+The sequence inside ``[...]`` will be applied multiple times in a loop until the Yul code
+remains unchanged or until the maximum number of rounds (currently 12) has been reached.
+
+An important thing to note, is that there are hardcoded sequences that are run before and after the
```
cameel:
```suggestion
An important thing to note, is that there are some hardcoded steps that are always run before and after the
```

docs/yul.rst
```diff
@@ -1245,19 +1245,26 @@ In Solidity mode, the Yul optimizer is activated together with the regular optim
...
+user-supplied sequence, or the default sequence if one was not supplied by the user.
 
-    solc --optimize --ir-optimized --yul-optimizations 'dhfoD[xarrscLMcCTU]uljmul'
+The cleanup sequence delimiter ``:`` is optional, and is used to supply a custom cleanup sequence
+in order to replace the default one. If omitted, the optimizer will simply apply the default cleanup
+sequence. In addition, the delimiter may be placed at the beginning of the user-supplied sequence,
+which will be treated as an empty optimization sequence, whereas conversely, if placed at the end of
```
cameel:
```suggestion
which will result in the optimization sequence being empty, whereas conversely, if placed at the end of
```

docs/yul.rst
```diff
@@ -1245,19 +1245,26 @@ In Solidity mode, the Yul optimizer is activated together with the regular optim
...
+the sequence, will be treated as an empty cleanup sequence.
 
-The order of steps is significant and affects the quality of the output.
-Moreover, applying a step may uncover new optimization opportunities for others that were already
-applied so repeating steps is often beneficial.
-By enclosing part of the sequence in square brackets (``[]``) you tell the optimizer to repeatedly
-apply that part until it no longer improves the size of the resulting assembly.
-You can use brackets multiple times in a single sequence but they cannot be nested.
```
cameel, line 1260: Some of this info is not included in the new text. I guess it was omitted intentionally to simplify the description?

libsolidity/interface/StandardCompiler.cpp
cameel: > The assert would therefore never be triggered, so even though it wouldn't cause any harm, I'd prefer to omit it.

But that's exactly when you want an assert :) I.e. you don't use them to validate conditions that can get triggered in normal use. You use them on ones you assume will never fail as long as other things are coded correctly. This way you detect bugs that happen when someone comes and changes the code in a way that breaks your assumption.

Especially in this case the correctness of this code depends on some other thing that is done in another function. It's easy to change one without realizing that you need to change the other as well.

libyul/optimiser/Suite.cpp
cameel: Actually, can you add a distilled version of of what @ekpyron said as a comment? If this is not obvious to the reader, it should be explained.

test/libsolidity/Metadata.cpp
```diff
@@ -400,6 +400,55 @@ BOOST_AUTO_TEST_CASE(metadata_revert_strings)
...
+		}
+	)";
+
+	vector<tuple<string, string>> sequences =
+	{
+		// { "<optimizer sequence>", "<optimizer cleanup sequence>" }
+		{ "", "" },
+		{ "", "fDn" },
+		{ "dhfoDgvulfnTUtnIf", "" },
+		{ "dhfoDgvulfnTUtnIf", "fDn" }
```
cameel:
```suggestion
		// {"<optimizer sequence>", "<optimizer cleanup sequence>"}
		{"", ""},
		{"", "fDn"},
		{"dhfoDgvulfnTUtnIf", ""},
		{"dhfoDgvulfnTUtnIf", "fDn"}
```

test/libsolidity/Metadata.cpp
```diff
@@ -400,6 +400,55 @@ BOOST_AUTO_TEST_CASE(metadata_revert_strings)
...
+	};
+
+	auto check = [sourceCode](string const& optimizerSequence, string const& optimizerCleanupSequence)
```
cameel:
```suggestion
	auto check = [sourceCode](string const& _optimizerSequence, string const& _optimizerCleanupSequence)
```

test/libsolidity/Metadata.cpp
```diff
@@ -400,6 +400,55 @@ BOOST_AUTO_TEST_CASE(metadata_revert_strings)
...
+	{
+		OptimiserSettings optimizerSettings = OptimiserSettings::minimal();
+		optimizerSettings.runYulOptimiser = true;
+		optimizerSettings.yulOptimiserSteps = optimizerSequence;
+		optimizerSettings.yulOptimiserCleanupSteps = optimizerCleanupSequence;
+		CompilerStack compilerStack;
+		compilerStack.setSources({{"", std::string(sourceCode)}});
```
cameel, line 427:
```suggestion
		compilerStack.setSources({{"", string(sourceCode)}});
```
Also, won't an implicit conversion to `string` kick in here?

test/libsolidity/Metadata.cpp
```diff
@@ -400,6 +400,55 @@ BOOST_AUTO_TEST_CASE(metadata_revert_strings)
 	BOOST_CHECK_EQUAL(metadata["settings"]["debug"]["revertStrings"], "strip");
 }
 
+BOOST_AUTO_TEST_CASE(metadata_optimiser_sequence)
+{
+	char const* sourceCode = R"(
+		pragma solidity >=0.0;
+		contract test {
```
cameel: I'd call it `C` or `A` because then it stands out in calls below as a contract name :) `test` is a bit generic in this context.
```suggestion
		contract C {
```

test/libsolidity/Metadata.cpp
```diff
@@ -400,6 +400,55 @@ BOOST_AUTO_TEST_CASE(metadata_revert_strings)
...
+
+		string const metadataOptimizerSteps = metadata["settings"]["optimizer"]["details"]["yulDetails"]["optimizerSteps"].asString();
+		string const expectedMetadataOptimiserSteps = optimizerSequence + ":" + optimizerCleanupSequence;
+		BOOST_CHECK_EQUAL(metadataOptimizerSteps, expectedMetadataOptimiserSteps);
+	};
+
+	for (auto const& [sequence, cleanupSequence] : sequences)
+	{
+		check(sequence, cleanupSequence);
+	}
```
cameel:
```suggestion
	for (auto const& [sequence, cleanupSequence]: sequences)
		check(sequence, cleanupSequence);
```

test/libsolidity/Metadata.cpp
cameel: You're not testing any sequences without the delimiter.

test/libsolidity/StandardCompiler.cpp
```diff
@@ -1247,7 +1247,8 @@ BOOST_AUTO_TEST_CASE(optimizer_settings_details_different)
 		(set<string>{"stackAllocation", "optimizerSteps"})
 	);
 	BOOST_CHECK(optimizer["details"]["yulDetails"]["stackAllocation"].asBool() == true);
-	BOOST_CHECK(optimizer["details"]["yulDetails"]["optimizerSteps"].asString() == OptimiserSettings::DefaultYulOptimiserSteps);
+	BOOST_CHECK(optimizer["details"]["yulDetails"]["optimizerSteps"].asString() ==
+				string{OptimiserSettings::DefaultYulOptimiserSteps} + ":" + string{OptimiserSettings::DefaultYulOptimiserCleanupSteps});
```
cameel:
```suggestion
	BOOST_CHECK(
		optimizer["details"]["yulDetails"]["optimizerSteps"].asString() ==
		string{OptimiserSettings::DefaultYulOptimiserSteps} + ":" + string{OptimiserSettings::DefaultYulOptimiserCleanupSteps}
	);
```

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+
+BOOST_AUTO_TEST_CASE(valid_optimiser_sequences)
+{
+	vector<string> validSequenceInputs {
+		":", 						// Empty optimizaiton sequence and empty cleanup sequence
+		":fDn", 					// Empty optimization sequence and specified cleanup sequence
+		"dhfoDgvulfnTUtnIf:", 		// Specified optimizaiton sequence and empty cleanup sequence
+		"dhfoDgvulfnTUtnIf:fDn", 	// Specified optimization sequence and cleanup sequence
+		"dhfo[Dgvulfn]TUtnIf:f[D]n" // Specified and nested optimization and cleanup sequence
+	};
```
cameel: Please use tabs only for indents. Internal whitespace should be spaces.

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+	BOOST_CHECK_EQUAL(commandLineOptions.optimiserSettings().yulOptimiserSteps, OptimiserSettings::DefaultYulOptimiserSteps);
+	BOOST_CHECK_EQUAL(commandLineOptions.optimiserSettings().yulOptimiserCleanupSteps, OptimiserSettings::DefaultYulOptimiserCleanupSteps);
+}
+
+BOOST_AUTO_TEST_CASE(valid_optimiser_sequences)
+{
+	vector<string> validSequenceInputs {
+		":", 						// Empty optimizaiton sequence and empty cleanup sequence
+		":fDn", 					// Empty optimization sequence and specified cleanup sequence
+		"dhfoDgvulfnTUtnIf:", 		// Specified optimizaiton sequence and empty cleanup sequence
```
```suggestion
		":", 						// Empty optimization sequence and empty cleanup sequence
		":fDn", 					// Empty optimization sequence and specified cleanup sequence
		"dhfoDgvulfnTUtnIf:", 		// Specified optimization sequence and empty cleanup sequence
```

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
 		}
 }
 
+BOOST_AUTO_TEST_CASE(default_optimiser_sequence)
+{
+	auto const& commandLineOptions = parseCommandLine({"solc", "contract.sol", "--optimize"});
```
```suggestion
	CommandLineOptions const& commandLineOptions = parseCommandLine({"solc", "contract.sol", "--optimize"});
```

test/solc/CommandLineParser.cpp
cameel: Again, would be best to have some inputs without the delimiter.

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+		"dhfoDgvulfnTUtnIf:fDn", 	// Specified optimization sequence and cleanup sequence
+		"dhfo[Dgvulfn]TUtnIf:f[D]n" // Specified and nested optimization and cleanup sequence
+	};
+
+	vector<tuple<string, string>> expectedParsedSequences {
+		{ "", "" },
+		{ "", "fDn"},
+		{ "dhfoDgvulfnTUtnIf", ""},
+		{ "dhfoDgvulfnTUtnIf", "fDn"},
+		{ "dhfo[Dgvulfn]TUtnIf", "f[D]n"}
```
```suggestion
		{"", ""},
		{"", "fDn"},
		{"dhfoDgvulfnTUtnIf", ""},
		{"dhfoDgvulfnTUtnIf", "fDn"},
		{"dhfo[Dgvulfn]TUtnIf", "f[D]n"}
```

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+	};
+
+	BOOST_CHECK_EQUAL(validSequenceInputs.size(), expectedParsedSequences.size());
+
+	for (size_t i = 0; i < validSequenceInputs.size(); ++i)
```
cameel, line 456: Why not a range loop? Does not look like you're doing anything non-trivial with the loop counter.

cameel (dismissed):
I only found nitpicky stuff so tentatively approving. All the important things were fixed. It would be best to do those tweaks but none of it is so critical that it would really stop a merge if we badly needed this PR merged right away for some reason.

cameel:
Ah wait, the changelog thing is actually important. But I see it conflicts so the PR can't get merged without fixing that anyway :)

libsolidity/interface/CompilerStack.cpp
cameel: Yeah, looks good.

test/cmdlineTests/yul_optimizer_steps_with_cleanup_sequence/input.sol
cameel: Yeah, I guess you could remove some of the CLI tests that overlap with the unit tests you added. A lot of them were added to test parsing as well.

libsolidity/interface/StandardCompiler.cpp
cameel: As explained by @ekpyron elsewhere, best to just reserialize the sequence back into what the input was.

docs/internals/optimizer.rst
cameel: Seems like you did this the other way around :) But fine with me either way.

test/solc/CommandLineParser.cpp
nikola-matic: I think we've had this discussion before; all in all, I want to add these tests as part of a separate PR, because they were never there to begin with, and I feel they'd add a bit of noise to this PR.

test/solc/CommandLineParser.cpp
nikola-matic:
```
auto const& [expectedYulOptimiserSteps, expectedYulCleanupSteps] = expectedParsedSequences[i];
```
I'm using it to fetch a value at the same index from another array.

test/solc/CommandLineParser.cpp
nikola-matic: Don't really have a problem with this, but curious to know why? This seems like a textbook use of `auto` to me.

test/libsolidity/Metadata.cpp
nikola-matic: The test setup is a copy-paste from another test case, and looking at the file now, there's a ton of `std::string` usage, so not sure if I should include such changes in this PR. Also yes, a `char*` is implicitly convertible to a `string` :)

test/libsolidity/Metadata.cpp
nikola-matic: We have a few more `contract test` instances here as well. Will make a mental note to clean it up in another PR, together with the `std::string` stuff. I have changed this particular instance to `contract C` though.

docs/internals/optimizer.rst
nikola-matic: Not sure I get what you mean - was this wrong to begin with, because I've made no changes here? Also, this links to `yul.rst` `Optimization Step Sequence` anchor.

docs/yul.rst
nikola-matic: > I guess it was omitted intentionally to simplify the description?

No, I didn't cross check the descriptions in `optimizer.rst` and `yul.rst`, and ended up selecting the `optimizer.rst` description in lieu of the `yul.rst` one. I've added the missing bits.

test/solc/CommandLineParser.cpp
ekpyron: We have/had a lot of in-fighting about when to use ``auto`` :-). Glad to see that another reasonable position has joined us :-D (I've always advocated using ``auto`` extensively - but we also have people who prefer seeing the types of things locally and I gave into that...)
Or a different way to put it: It was Chris all along :-D.

test/solc/CommandLineParser.cpp
nikola-matic: Implemented in #13519

test/solc/CommandLineParser.cpp
ekpyron: But yeah, the vague policy we have in the code right now is to only use ``auto`` if the type appears in the same statement (like you have a ``dynamic_cast`` or the like on the rhs anyways mentioning the type), and, if the type is unnamable, and on structured binding (i.e. where you have to).

The coding style has ``9. Only use ``auto`` if the type is very long and rather irrelevant.`` which is also pretty vague, but yeah, what I just wrote above is how we ended up doing this in practice most of the time, even though we were never unanimous about it - and it's not like I don't see the point of that position at all, so...

nikola-matic: Yeah, I'll change it.

Changelog.md
```diff
@@ -21,7 +21,7 @@ Compiler Features:
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
 
 
-Bugfixes:
+Bugfixes:`````
```
ekpyron: Is there really no comment yet complaining about this or was I just blind :-)?

nikola-matic: Oh boy, looks like I was both fat fingered, and blind.

ekpyron: Happens :-).

Changelog.md
```diff
@@ -4,7 +4,7 @@ Language Features:
 
 
 Compiler Features:
-
+* Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (```:```) in the sequence string.
```
nikola-matic: Look @ekpyron, more blind :D (```:```)

docs/internals/optimizer.rst
cameel: True. Looks like it was wrong before. Or maybe not exactly wrong - I see now that Yul docs just have a small section about the optimizer so it's technically correct - but a bit misleading since the main optimizer docs are actually here, not there.

Looking at it now, I think it's a bit weird to have that section with content that's not present in the main optimizer docs. I think we should move most of its content to the main optimizer docs and leave it just as a quick stub with a link.

test/libsolidity/Metadata.cpp
cameel: ok. Fine then. I think we should fix individual cases we find but yeah, redoing the whole file would add too much noise. Not that important anyway.

test/libsolidity/Metadata.cpp
cameel: Yeah. I agree. Better to leave cleaning up the old stuff for later. I try to point this stuff out only when I see it in new code anyway because that's always an opportunity to do things better :)

test/solc/CommandLineParser.cpp
cameel: For me the thing is that `auto` helps mostly when writing, not when reading. When reading I almost always prefer to see the type, especially when it's short and using auto would not save much typing anyway.

This is a bit borderline case since the variable is already called `commandLineOptions` but names are not as reliable as just seeing the type.

test/solc/CommandLineParser.cpp
cameel: ok, sounds reasonable.

test/solc/CommandLineParser.cpp
cameel: Ah, ok, my bad. Was only skimming through it and did not notice.

Changelog.md
cameel: > Is there really no comment yet complaining about this or was I just blind :-)?

Just a race condition :P This got in there as a part of the fix, after my review :)

Changelog.md
```diff
@@ -4,7 +4,7 @@ Language Features:
 
 
 Compiler Features:
-
```
cameel, line 7: We usually keep two empty lines here, so if you delete it, we'll need to add one back when sorting the changelog anyway ;)

libyul/optimiser/Suite.cpp
```diff
@@ -139,7 +140,13 @@ void OptimiserSuite::run(
 			_optimizeStackAllocation,
 			stackCompressorMaxIterations
 		);
-	suite.runSequence("fDnTOc g", ast);
+
+	// Run the user-supplied clean up sequence
+	suite.runSequence(_optimisationCleanupSequence, ast);
+	// Hard-coded FunctionGrouper step used to bring the AST into a canonical form required by the StackCompressor
```
cameel: Without "is" this reads as past tense at first.
```suggestion
	// Hard-coded FunctionGrouper step is used to bring the AST into a canonical form required by the StackCompressor
```

solc/CommandLineParser.cpp
```diff
@@ -266,7 +266,14 @@ OptimiserSettings CommandLineOptions::optimiserSettings() const
...
 
 	if (optimizer.yulSteps.has_value())
-		settings.yulOptimiserSteps = optimizer.yulSteps.value();
+	{
+		string const fullSequence = optimizer.yulSteps.value();
+		auto const delimiterPos = fullSequence.find(":");
+		settings.yulOptimiserSteps = fullSequence.substr(0, delimiterPos);
+
+		if (delimiterPos != string::npos)
+			settings.yulOptimiserCleanupSteps = fullSequence.substr(delimiterPos + 1);
```
cameel, line 275: This also needs the assert I mentioned in https://github.com/ethereum/solidity/pull/13376/files#r950386578

test/libsolidity/StandardCompiler.cpp
```diff
@@ -1247,7 +1247,10 @@ BOOST_AUTO_TEST_CASE(optimizer_settings_details_different)
 		(set<string>{"stackAllocation", "optimizerSteps"})
 	);
 	BOOST_CHECK(optimizer["details"]["yulDetails"]["stackAllocation"].asBool() == true);
-	BOOST_CHECK(optimizer["details"]["yulDetails"]["optimizerSteps"].asString() == OptimiserSettings::DefaultYulOptimiserSteps);
+	BOOST_CHECK(
+		optimizer["details"]["yulDetails"]["optimizerSteps"].asString() ==
+				string{OptimiserSettings::DefaultYulOptimiserSteps} + ":" + string{OptimiserSettings::DefaultYulOptimiserCleanupSteps}
```
cameel:
```suggestion
		string{OptimiserSettings::DefaultYulOptimiserSteps} + ":" + string{OptimiserSettings::DefaultYulOptimiserCleanupSteps}
```

cameel:
```suggestion
				OptimiserSettings::DefaultYulOptimiserSteps + ":"s + OptimiserSettings::DefaultYulOptimiserCleanupSteps
```

test/solc/CommandLineParser.cpp
```diff
@@ -426,6 +426,58 @@ BOOST_AUTO_TEST_CASE(invalid_options_input_modes_combinations)
...
+}
+
+BOOST_AUTO_TEST_CASE(valid_optimiser_sequences)
+{
+	vector<string> validSequenceInputs {
+		":",					// Empty optimization sequence and empty cleanup sequence
+		":fDn",					// Empty optimization sequence and specified cleanup sequence
+		"dhfoDgvulfnTUtnIf:",			// Specified optimization sequence and empty cleanup sequence
+		"dhfoDgvulfnTUtnIf:fDn",		// Specified optimization sequence and cleanup sequence
+		"dhfo[Dgvulfn]TUtnIf:f[D]n"		// Specified and nested optimization and cleanup sequence
```
cameel: Please use tabs only for indentation. The inner whitespace should be spaces.

cameel (dismissed):

cameel (approved):

Issue #13308, opened 2022-07-27
Stop the war on for loops


fulldecent:
## Abstract

Add a specific overflow check optimization that will make many for-loop optimizations unnecessary.

## Motivation

Too many people are saying this is bad:

```solidity
for (uint256 mintCounter = 0; mintCounter < quantity; mintCounter++) {
    _mint(msg.sender, _assembleTokenID(dropID, mintCounter));
}
```

compared to this:

```solidity
for (uint256 mintCounter = 0; mintCounter < quantity;) {
    _mint(msg.sender, _assembleTokenID(dropID, mintCounter));
    unchecked{
        mintCounter++
    }
}       
```

### Note

For reference, a for-loop like this:

```
for (A; B; C) {D}
```

is implemented like this:

```basic
10 A

100 if !B GOTO 999
110 D
120 C
130 GOTO 100

999 EXIT
```

### New optimization

If inside a single code unit, if:

1. a variable has bit width X,
2. the variable is compared to be less than some quantity with bit width <=X,
3. the variable is not set until step 4 here, then
4. the variable is incremented;

then that incrementation in step 4 need not be safety checked. Or in $\LaTeX$:

$$x < a <= w \implies x+1 <= w$$

### Documentation

People like being fancy, so even if this is implemented, they will still use the `unchecked` "optimization" until it can be clearly explained that it is unnecessary.

So the documentation should be updated to specify that the gas cost of this code:

```solidity
contract C {
    event E;
    constructor(uint256 count) {
        for (uint256 i = 0; i < count; i++) {
            emit E(i);
        }
    }
}
```

shall not exceed that of this code:

```solidity
contract CUnchecked {
    event E;
    constructor(uint256 count) {
        for (uint256 i = 0; i < count;) {
            unchecked {
                emit E(i);
                i++;
            }
        }
    }
}
```

And this can be checked with a test case.

There is precedent in this with JavaScript where the specification requires that implementations implement tail-end recursion optimization.

## Backwards Compatibility

n/a

---

Inspired by the discussion with @FrankNFT-labs at https://github.com/LightArtists/light-smart-contracts/issues/2

fulldecent:
The second optimization is:

If inside a single code unit, if:

1. an unsigned variable is compared to be greater than some value,
2. the variable is not set until step 3 here, then
3. the variable is decremented,

then that decrementation in step 3 need not be safety checked. Or in $\LaTeX$:

$$x > a >= 0 \implies x-1 > = 0$$

[...]

PR #13378, opened 2022-08-10
unchecked increment for simple counter for loops

ekpyron:
I'm not saying we should do things this way. Not at all. But we said we explore all our options and want a PR draft, so here is a very minimal version.

Implements a reduced version of https://github.com/ethereum/solidity/issues/13308, restricted to for loop conditions and loop post expressions specifically only. (which gets around having to analyze control flow entirely compared to the general case)

If we decide to actually pursue this approach, we'd, of course, need extensive unit testing on this.
Might also even make sense to export the annotation in the JSON AST to give confirmation that the optimization kicked in and as an easy indicator if this were to turn out buggy.


ekpyron:
Our test suite are of course a gross misrepresentation of the effects of this, but still :-):

<details><summary>Click for a table of gas differences</summary>

| File name                                                                |   IR-optimized (%) |   Legacy-Optimized (%) |   Legacy (%) |
|--------------------------------------------------------------------------|--------------------|------------------------|--------------|
| events/event_indexed_string.sol                                          |        -0.944587   |             -3.12466   |   -5.69622   |
| externalContracts/snark.sol                                              |         0          |              0         |   -2.22538   |
| externalContracts/ramanujan_pi.sol                                       |        -0.255462   |             -1.64518   |   -2.5779    |
| externalContracts/prbmath_unsigned.sol                                   |        -0.348873   |             -0.530369  |   -0.752172  |
| externalContracts/deposit_contract.sol                                   |        -1.13166    |             -0.934168  |   -1.27186   |
| externalContracts/prbmath_signed.sol                                     |        -0.335524   |             -0.403214  |   -0.683038  |
| externalContracts/base64.sol                                             |        -2.15859    |             -3.015     |   -3.40462   |
| externalContracts/strings.sol                                            |        -0.157327   |             -0.241564  |   -0.103302  |
| abiEncoderV2/abi_encode_calldata_slice.sol                               |       -11.8235     |            -16.5388    |  -24.0856    |
| abiEncoderV2/calldata_array.sol                                          |        -5.65227    |              0         |  -27.6622    |
| userDefinedValueType/calldata.sol                                        |        -0.262964   |             -0.338576  |   -0.844212  |
| libraries/internal_types_in_library.sol                                  |        -0.453327   |             -1.10759   |   -1.8554    |
| abiEncoderV1/abi_encode_calldata_slice.sol                               |       -11.8235     |            -16.5388    |  -24.0856    |
| viaYul/array_memory_index_access.sol                                     |       -20.1241     |            -22.7966    |  -25.0549    |
| viaYul/array_storage_index_access.sol                                    |        -0.617372   |             -0.787358  |   -1.35619   |
| viaYul/array_storage_push_empty.sol                                      |        -0.874452   |             -2.40519   |   -4.07349   |
| viaYul/array_storage_index_zeroed_test.sol                               |        -0.714643   |             -0.892007  |   -1.51651   |
| array/function_array_cross_calls.sol                                     |        -0.213694   |             -0.238643  |   -0.265407  |
| array/byte_array_transitional_2.sol                                      |        -0.905852   |             -1.512     |   -2.20495   |
| array/dynamic_array_cleanup.sol                                          |        -0.0639088  |             -0.280256  |   -0.483144  |
| array/fixed_array_cleanup.sol                                            |        -0.107869   |             -0.269519  |   -0.509251  |
| array/delete/bytes_delete_element.sol                                    |        -0.768535   |             -2.34424   |   -4.70764   |
| array/push/byte_array_push_transition.sol                                |        -2.46718    |             -3.13525   |   -4.53342   |
| array/push/push_no_args_bytes.sol                                        |        -0.842108   |             -2.2654    |   -4.56991   |
| array/copying/array_copy_different_packing.sol                           |        -0.0908003  |             -0.236542  |   -0.486802  |
| array/copying/array_copy_target_simple_2.sol                             |        -0.0585701  |             -0.147416  |   -0.303951  |
| array/copying/memory_dyn_2d_bytes_to_storage.sol                         |        -0.534438   |             -0.634176  |   -1.19305   |
| array/copying/array_copy_including_array.sol                             |        -0.00797622 |             -0.0101895 |   -0.0186894 |
| array/copying/array_copy_target_simple.sol                               |        -0.0684556  |             -0.172879  |   -0.355701  |
| array/copying/array_copy_target_leftover.sol                             |        -2.87541    |             -2.68691   |   -4.77347   |
| array/copying/storage_memory_nested.sol                                  |        -0.0485472  |             -0.117279  |   -0.226081  |
| array/copying/storage_memory_packed_dyn.sol                              |        -0.569301   |             -1.85909   |   -3.13415   |
| array/copying/array_of_structs_containing_arrays_calldata_to_storage.sol |        -0.190361   |              0         |    0         |
| array/copying/storage_memory_nested_from_pointer.sol                     |        -0.0485472  |             -0.117276  |   -0.226076  |
| array/copying/bytes_storage_to_storage.sol                               |        -5.3558     |             -6.24933   |  -10.5266    |
| array/pop/byte_array_pop_masking_long.sol                                |        -0.580863   |             -1.60901   |   -3.24882   |
| array/pop/byte_array_pop_copy_long.sol                                   |        -1.37407    |             -1.77116   |   -3.50429   |
</details>

Keeping in mind that those results are not indicative of anything, since we never manually gas-optimized our tests for this issue.

External test benchmarks are of course much less impressive, since in real-world code people do work around this issue. However, surprisingly, it does still have *some* effect:

<details>


### `ir-no-optimize`
|          project |  bytecode_size | deployment_gas | method_gas |
|:----------------:|---------------:|---------------:|-----------:|
|           bleeps |                |                |            |
|            brink |                |                |            |
|        chainlink |                |                |            |
|           colony | **`-0.13% ✅`** |                |            |
|        elementfi |                |                |            |
|              ens |                |                |            |
|            euler |                |                |            |
|           gnosis |                |                |            |
|              gp2 |                |                |            |
|  perpetual-pools |                |                |            |
|    pool-together |                |                |            |
|         prb-math |                |                |            |
|          trident |                |                |            |
|          uniswap |                |                |            |
| yield_liquidator |                |                |            |
|         zeppelin |                |                |            |

### `ir-optimize-evm+yul`
|           bleeps | **`-0.24% ✅`** |           `0%` |          `-0%` |
|            brink |           `0%` |                |                |
|        chainlink | **`-0.09% ✅`** | **`-0.14% ✅`** | **`-0.07% ✅`** |
|           colony | **`-0.14% ✅`** |                |                |
|        elementfi | **`+0.22% ❌`** |                |                |
|              ens |           `0%` | **`-0.09% ✅`** |           `0%` |
|            euler | **`-0.27% ✅`** | **`-0.28% ✅`** | **`-0.01% ✅`** |
|           gnosis | **`-0.15% ✅`** | **`-0.14% ✅`** | **`-0.03% ✅`** |
|              gp2 | **`-0.35% ✅`** | **`-0.24% ✅`** | **`-0.04% ✅`** |
|  perpetual-pools |  **`-0.1% ✅`** |  **`-0.1% ✅`** | **`-0.01% ✅`** |
|    pool-together | **`+0.09% ❌`** | **`+0.08% ❌`** | **`-0.05% ✅`** |
|          trident | **`+0.07% ❌`** | **`-0.16% ✅`** | **`-0.13% ✅`** |
|          uniswap | **`-0.35% ✅`** | **`-0.35% ✅`** | **`-0.65% ✅`** |
| yield_liquidator | **`-0.18% ✅`** | **`-0.16% ✅`** | **`-0.05% ✅`** |
|         zeppelin | **`-0.08% ✅`** | **`-0.04% ✅`** |          `-0%` |

### `ir-optimize-evm-only`
|          project |  bytecode_size | deployment_gas | method_gas |
|:----------------:|---------------:|---------------:|-----------:|
|           bleeps |                |                |            |
|            brink |                |                |            |
|        chainlink |                |                |            |
|           colony | **`-0.13% ✅`** |                |            |
|        elementfi |                |                |            |
|              ens |                |                |            |
|            euler |                |                |            |
|           gnosis |                |                |            |
|              gp2 |                |                |            |
|  perpetual-pools |                |                |            |
|    pool-together |                |                |            |
|         prb-math |                |                |            |
|          trident |                |                |            |
|          uniswap |                |                |            |
| yield_liquidator |                |                |            |
|         zeppelin |                |                |            |

### `legacy-no-optimize`
|            brink |           `0%` |                |                |
|        chainlink | **`-0.35% ✅`** |                |                |
|           colony | **`-0.13% ✅`** |                |                |
|        elementfi | **`-0.42% ✅`** |                |                |
|            euler | **`-0.36% ✅`** | **`-0.39% ✅`** | **`-0.03% ✅`** |
|           gnosis | **`-0.14% ✅`** | **`-0.24% ✅`** | **`-0.08% ✅`** |
|              gp2 | **`-0.35% ✅`** |                |                |
|  perpetual-pools | **`-0.14% ✅`** | **`-0.15% ✅`** |          `+0%` |
|    pool-together |  **`-0.3% ✅`** | **`-0.38% ✅`** | **`-0.12% ✅`** |
|          trident | **`-0.27% ✅`** | **`-0.19% ✅`** | **`-0.11% ✅`** |
|          uniswap | **`-0.28% ✅`** | **`-0.24% ✅`** | **`-0.26% ✅`** |
| yield_liquidator | **`-0.38% ✅`** | **`-0.37% ✅`** | **`-0.11% ✅`** |
|         zeppelin | **`-0.25% ✅`** | **`-0.21% ✅`** | **`+0.02% ❌`** |

### `legacy-optimize-evm+yul`
|           bleeps | **`-0.32% ✅`** |           `0%` |           `0%` |
|            brink |           `0%` |                |                |
|        chainlink | **`-0.38% ✅`** | **`-0.37% ✅`** |  **`-0.2% ✅`** |
|           colony | **`-0.14% ✅`** |                |                |
|        elementfi |  **`-0.5% ✅`** |                |                |
|              ens |           `0%` | **`-0.13% ✅`** |           `0%` |
|            euler | **`-0.31% ✅`** | **`-0.34% ✅`** | **`-0.02% ✅`** |
|           gnosis |  **`-0.2% ✅`** | **`-0.18% ✅`** | **`-0.04% ✅`** |
|              gp2 | **`-0.37% ✅`** | **`-0.39% ✅`** | **`-0.04% ✅`** |
|  perpetual-pools | **`-0.14% ✅`** | **`-0.15% ✅`** | **`-0.01% ✅`** |
|    pool-together | **`-0.36% ✅`** | **`-0.42% ✅`** | **`-0.07% ✅`** |
|          trident |  **`-0.3% ✅`** | **`-0.28% ✅`** | **`-0.12% ✅`** |
|          uniswap |  **`-0.3% ✅`** | **`-0.26% ✅`** | **`-0.25% ✅`** |
| yield_liquidator | **`-0.39% ✅`** | **`-0.41% ✅`** | **`-0.06% ✅`** |
|         zeppelin | **`-0.25% ✅`** | **`-0.21% ✅`** | **`-0.01% ✅`** |

|            brink |           `0%` |                |                |
|        chainlink | **`-0.35% ✅`** | **`-0.34% ✅`** | **`-0.21% ✅`** |
|           colony | **`-0.13% ✅`** |                |                |
|        elementfi | **`-0.45% ✅`** |                |                |
|              ens |           `0%` | **`-0.12% ✅`** |           `0%` |
|            euler |  **`-0.3% ✅`** | **`-0.33% ✅`** | **`-0.02% ✅`** |
|           gnosis | **`-0.16% ✅`** | **`-0.17% ✅`** | **`-0.05% ✅`** |
|              gp2 | **`-0.34% ✅`** | **`-0.37% ✅`** | **`-0.04% ✅`** |
|  perpetual-pools | **`-0.12% ✅`** | **`-0.13% ✅`** | **`+0.01% ❌`** |
|    pool-together | **`-0.31% ✅`** | **`-0.37% ✅`** | **`-0.07% ✅`** |
|          trident | **`-0.26% ✅`** | **`-0.23% ✅`** | **`-0.12% ✅`** |
|          uniswap | **`-0.28% ✅`** | **`-0.23% ✅`** | **`-0.24% ✅`** |
| yield_liquidator | **`-0.37% ✅`** | **`-0.39% ✅`** | **`-0.06% ✅`** |
|         zeppelin | **`-0.22% ✅`** | **`-0.19% ✅`** | **`-0.01% ✅`** |


`!V` = version mismatch

</details>

But yeah, the goal here isn't actually saving gas directly, but rather allowing more readable code to be equally gas-efficient as hand-optimized ugly code.

fulldecent:
Wow, this is looking good for the test case gas changes.

But I'm surprised that ANY test cases are worse with this change.

ekpyron:
> Wow, this is looking good for the test case gas changes.
> 
> But I'm surprised that ANY test cases are worse with this change.

The gas reports on the external tests (i.e. hardhat's gas reports) are not 100% reliable, so very small changes < 0.05% may due to some non-deterministic factors. Apart from that, it seems to be a slight code size increase on occasion - there it might be that the change tipped the scales on the inlining heuristics in some cases. Might be worth a deeper look into, though, if we pursue this approach further.

Changelog.md
```diff
@@ -8,6 +8,7 @@ Language Features:
 
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
+ * Code Generator: Use unchecked arithmetic for the increment in specific for loops.
```
chriseth: If this is only an optimization, then we should phrase it as: Remove redundant overflow check for increment...

ekpyron: We'd probably need to make it some kind of optimizer setting then, though (i.e. it'd be easiest if it was a setting in `solidity::frontend::OptimiserSettings`)... doesn't need to be configurable in CLI, but maybe in standard-json... for that the crucial question would be if it should get a specific setting of its own or whether we add something generic like "optimized code generation" to be used for all similar cases like this...

ekpyron:
Adding the ``takeover`` label here, since I'm not sure when I'll find the time to get back to this myself.

The main thing missing here is a solution for testing this and similar future optimizations.

All of the properties resulting from the analysis of Solidity should, like here, end up in the AST, so technically our AST tests should do the trick - but they're too huge and unfocused for this purpose, so what needs to be done here is to define a nice way to test for specific AST properties and annotations in ``isoltest``.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,59 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+	SimpleCounterForLoopChecker(ErrorReporter& _errorReporter): Checker(_errorReporter) {}
+	bool visit(ForStatement const& _forStatement) override
+	{
+		_forStatement.annotation().isSimpleCounterLoop = isSimpleCounterLoop(_forStatement);
+		return true;
+	}
+	bool isSimpleCounterLoop(ForStatement const& _forStatement) const
+	{
+		auto const* cond = dynamic_cast<BinaryOperation const*>(_forStatement.condition());
+		if (!cond || cond->getOperator() != Token::LessThan)
```
chriseth: Just for safety we should check for both the `<`-operator here and the `++` operator below that they are not overwritten by a user-defined operator (i.e. they don't lead to a user function call).

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,59 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			return false;
+		if (!_forStatement.loopExpression())
+			return false;
+
+		auto const* post = dynamic_cast<UnaryOperation const*>(&_forStatement.loopExpression()->expression());
```
chriseth: Add a comment that this matches both `i++` and `++i`.

libsolidity/ast/ASTAnnotations.h
```diff
@@ -235,6 +235,7 @@ struct TryCatchClauseAnnotation: ASTAnnotation, ScopableAnnotation
 
 struct ForStatementAnnotation: StatementAnnotation, ScopableAnnotation
 {
+	util::SetOnce<bool> isSimpleCounterLoop;
```
chriseth, line 241: I think we should single those out into something like "optimizer hint annotations" to clearly separate them from type checker annotations.

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -1246,7 +1246,13 @@ bool ContractCompiler::visit(ForStatement const& _forStatement)
 
 	// for's loop expression if existing
 	if (_forStatement.loopExpression())
+	{
+		Arithmetic previousArithmetic = m_context.arithmetic();
+		if (*_forStatement.annotation().isSimpleCounterLoop)
```
chriseth: Maybe we should not set the annotation on the loop but rather on the `i++` expression?

chriseth: Also, as commented elsewhere, this should only be done if optimizations are activated.

ekpyron:
> This pull request is stale because it has been open for 14 days with no activity. It will be closed in 7 days unless the `stale` label is removed.

Not stale, just still waiting for someone to finish it.

Changelog.md
```diff
@@ -4,6 +4,7 @@ Language Features:
 
 
 Compiler Features:
+ * Code Generator: Remove redundant overflow checks in specific for loops.
```
fulldecent: Might be worth saying "specific popular for loops".

Or, if marketing speak is allowed... "most for loops"

fulldecent:
Unstale.

This one is important.

ekpyron:
Yeah, I had already un-staled it - I'll try to expedite finding someone to  finish it (it's mainly blocked by requiring more and better testing, which warrants some new testing infrastructure, since we'll likely do more optimizations like this in the future) - or otherwise try to find the time to finish it myself.

fulldecent:
@ekpyron Thank you. Maybe there is a cheat we can use to get this done with existing test tools. I assume the thing you are requiring is gas testing.

For example, with the existing `.call` syntax we can specify an amount of gas. And we can hard code in an expected amount of gas for the naive implementation (`for...`) and the deprecated over optimization (`for ... unchecked{...}`).

For some gas number both the naive and the deprecated should fail for out of gas. And for that number minus one, both should succeed.

ekpyron:
The issue is not gas tests, but tests for semantic correctness, i.e. tests ensuring that here exactly those loops that qualify get annotated as "simple counter loops", in general, that only the correct AST patterns get annotated. We need a good way to test these kinds of annotations, since we'll add more similar optimizations over time.

fulldecent:
It is possible to test this by making executing this optimization against all test suites and expecting NO CHANGE. Then manually reviewing and switching to CHANGE EXPECTED for the ones that are correct?

This means that as new tests are added the test suite will fail if they reach a situation where this optimization would change that.

NunoFilipeSantos:
There are 2 things currently missing:
- Handing down the optimize flag
- Lacks testing that checks single properties of the AST
  - Solidity code+Insert NatSpec comments (JSON path to the subfield name:subfield) on which the sub-property started (test expectations name:valueOfthefield)

ekpyron:
Quick hint: for rebasing this, it'll be easiest to just drop the gas changes in the tests and just regenerating them with isoltest.

The new kind of tests should roughly look like this:

```
contract C {
  function f() public pure {
     /// TestCase1: isSimpleCounterLoop
     for(uint i = 0; i < 42; ++i) {
     }
     /// TestCase2: isSimpleCounterLoop
     for(uint i = 1; i < 42; i = i * 2) {
     }
  }
}
/// TestCase1: true
/// TestCase2: false
```
However, feel free to adjust the design as you see fit when writing it, if a slightly different version makes this easier.
It should be possible for these tests to solely act on the JSON AST.

test/libsolidity/ASTPropertyTests/simpleLoopOptimization.sol
```diff
@@ -0,0 +1,30 @@
+contract C {
```
ekpyron: Nice! This very much looks like I'd have imagined this already - should split out the other stub test cases not connected to the counter loops into a separate file, but looks good!

test/libsolidity/ASTPropertyTest.h
cameel, line 1: On the last call we wanted this new system for testing AST to have a separate, quicker review process. Wouldn't it make more sense to extract it into a separate PR and here keep only the changes related to the loop counter optimization?

Pinging @ekpyron just in case he has objections to this, but I'd hope this is not controversial (though not sure because for some reason it was not even mentioned on the call).

ekpyron: Yes, technically, it'd be nicer to do it separately, I was wondering about asking to split out the testing myself - but the analysis of the counter loops is not overly complicated and not that much of additional code, so I'd have said we can skip the split here and just review it together.

ekpyron: Actually, that being said, since it won't be a huge problem to split the testing out either, it actually *does* make sense:

I just had a look at the rest and the loop optimizations are generally sound as they are, but we still need https://github.com/ethereum/solidity/pull/13378#discussion_r967026418, i.e. a finer-grained optimizer setting to control it at least for standard-json.

So @matheusaaguiar, are you up for extracting the AST property tests into a separate PR that we can then quickly merge first?

matheusaaguiar: @ekpyron , yes. I have created #14309 just now.

cameel:
@matheusaaguiar Can you rebase this on #14309?

test/libsolidity/ASTPropertyTest.h
cameel: Can you also rebase this PR on the new one?

matheusaaguiar: @cameel rebased.

Changelog.md
cameel: I agree that it wouldn't hurt if this was a little bit more concrete to give people a better idea of which loops get optimized this way.

We should also mention the new `isSimpleCounterLoop` annotation as an AST change.

cameel (commented):
Just a heads up - I rebased this on the latest state of `astPropertyTests` branch to be able to see only the changes from this PR in review.

I also pushed a small fixup commit with some tweaks I spotted so far that were not important enough to be worth posting as comments. E.g. the `m_declaration` pointer not having an initializer at the point of declaration, which is one of @ekpyron's pet peeves :)

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,60 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
 
+class LValueChecker: public ASTConstVisitor
+{
+public:
+	LValueChecker(Identifier const& _identifier): m_declaration(_identifier.annotation().referencedDeclaration) {}
+	bool willBeWrittenTo() const { return m_willBeWrittenTo; }
+	void endVisit(Identifier const& _identifier) override
+	{
+		if (
+			_identifier.annotation().referencedDeclaration == m_declaration &&
```
cameel: I don't think this code will work as expected if `referencedDeclaration` happens to be null. It's probably never supposed to be null, but in that case you should assert that here.

Then we could also safely use `ASTNode`'s `==` instead of comparing pointers:
```suggestion
			*_identifier.annotation().referencedDeclaration == *m_declaration &&
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,60 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+		auto const* lhsIdentifier = dynamic_cast<Identifier const*>(&cond->leftExpression());
+		auto const* lhsType = dynamic_cast<IntegerType const*>(cond->leftExpression().annotation().type);
+		auto const* commonType = dynamic_cast<IntegerType const*>(cond->annotation().commonType);
+
+		if (lhsIdentifier && lhsType && commonType && *lhsType == *commonType)
+		{
+			LValueChecker lValueChecker{*lhsIdentifier};
+			_forStatement.body().accept(lValueChecker);
+			if (!lValueChecker.willBeWrittenTo())
+				return true;
```
cameel: Is checking `willBeWrittenTo` enough to account for all possibilities where an integer variable could be modified? I think that inline assembly may be a case we're missing here. I don't think Yul assignments get Solidity `Expression` nodes with that annotation - which would mean that we're not checking such assignments. Perhaps we should skip loops containing inline assembly.

In any case, it would be nice to have a test for this mechanism with a loop containing inline assembly to see for sure what the behavior is.

cameel (changes requested):

libsolidity/analysis/PostTypeChecker.cpp
ekpyron: That's probably right, I don't think it currently accounts for that.

It should be easy to add a visit for ``InlineAssembly`` in the ``LValueChecker`` and just walk through the ``externalReferences`` in its annotation, checking if it refers to the declaration in question and consider any reference there as a write. That'll also include read-only access from assembly, but that's fine for now, we can still refine later.

test/libsolidity/ASTPropertyTest.cpp
nikola-matic, line 1: Wait, did we merge ASTPropertyTest PR into this branch, or into develop?

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,61 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+	{
+		auto const* simpleCondition = dynamic_cast<BinaryOperation const*>(_forStatement.condition());
+		if (!simpleCondition || simpleCondition->getOperator() != Token::LessThan || simpleCondition->userDefinedFunctionType())
+			return false;
+		if (!_forStatement.loopExpression())
+			return false;
+
+		auto const* simplePostExpression = dynamic_cast<UnaryOperation const*>(&_forStatement.loopExpression()->expression());
+		// This matches both operators ++i and i++
+		if (!simplePostExpression || simplePostExpression->getOperator() != Token::Inc || simplePostExpression->userDefinedFunctionType())
```
cameel, line 512: @nikola-matic's comment on the chat made me realize that `simpleCondition` + `simplePostExpression` is still a bit too weak as a check. A loop like this would pass that condition:

```solidity
for (uint i = type(uint).max - 5; i < 10; ++i) { ... }
```

even though this will pretty clearly overflow.

We need to check the initial value as well.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,61 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			return false;
+
+		auto const* lhsIdentifier = dynamic_cast<Identifier const*>(&simpleCondition->leftExpression());
+		auto const* lhsIntegerType = dynamic_cast<IntegerType const*>(simpleCondition->leftExpression().annotation().type);
```
cameel, line 516: I just realized that actually excludes loops where the upper limit is a literal:
```solidity
for (uint i = 0; i < 10; ++i) { ... }
```

I'd expect those to be pretty common (and safe to optimize).

libsolidity/analysis/PostTypeChecker.cpp
cameel: Ah, wait, disregard that. It would only overflow if the condition was `i != 10` (and we don't optimize that case). So it's fine after all.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -422,6 +433,61 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+	SimpleCounterForLoopChecker(ErrorReporter& _errorReporter): Checker(_errorReporter) {}
+	bool visit(ForStatement const& _forStatement) override
+	{
+		_forStatement.annotation().isSimpleCounterLoop = isSimpleCounterLoop(_forStatement);
+		return true;
+	}
+	bool isSimpleCounterLoop(ForStatement const& _forStatement) const
+	{
+		auto const* simpleCondition = dynamic_cast<BinaryOperation const*>(_forStatement.condition());
+		if (!simpleCondition || simpleCondition->getOperator() != Token::LessThan || simpleCondition->userDefinedFunctionType())
```
cameel, line 505: As pointed out by @nikola-matic, this covers `<` but not `<=`. It's pretty easy to support `<=` as well, so I think we should add it.

cameel: Ah, maybe not so trivial after all.

This case breaks things and will let `i` overflow:
```solidity
uint uintMax = type(uint).max;
for (uint i = uintMax; i <= uintMax; ++i) { ... }
```

In that case we can only really support `<`.

nikola-matic: Ah, of course.

test/libsolidity/ASTPropertyTest.cpp
cameel: We did not merge it yet. You see its commits here because this PR is based on it, but it diverged (i.e. the base PR was rebased and potentially changed) so github no longer recognizes these commits as a part of that other PR.

They will disappear once it's rebased.

libsolidity/analysis/PostTypeChecker.cpp
ekpyron: It checks the type of the **left hand side**, and the common type of the comparison, not the right-hand side, so it should *not* exclude literals as upper limit.

cameel: Ah, true. Great. This means this does not exclude things like array `.length` either.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Done.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
```diff
@@ -0,0 +1,29 @@
...
+     /// NegativeCase2: isSimpleCounterLoop
+     for(uint i = 42; i > 0; --i) {
+     }
+     /// NegativeCase3: isSimpleCounterLoop
+     for(uint i = 42; i > 0; i--) {
+     }
+     /// NegativeCase4: isSimpleCounterLoop
+     for(uint i = 1; i < 42; i = i * 2) {
+     }
+  }
```
ekpyron: There's further kinds of negative test cases we need here. At least:
- A proper ``for(uint i = 0; i < 42; ++i)``, but with several kinds of writes to the counter variable in the loop body (like ``i++`` or ``i = 43;``).
- Same as above, but with a write to the counter variable in inline assembly. Note that in the current state of the PR the result will be incorrect, so this still actually needs fixing.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
```diff
@@ -0,0 +1,29 @@
...
+}
+// ----
+// PositiveCase1: true
+// PositiveCase2: true
+// NegativeCase1: false
+// NegativeCase2: false
+// NegativeCase3: false
+// NegativeCase4: false
```
cameel: Having names for these cases makes the test *a lot* easier to process. This way, you can actually understand the expectations without having to jump back to the code to "decode" them as you read. This is already annoying with one case, will be much worse with lots of them.

```suggestion
// preIncrement: true
// postIncrement: true
// addOne: false
// preDecrement: false
// postDecrement: false
// multiplyByTwo: false
```

cameel: Since this is the first use of these tests since we introduce them I'm going to also go back to https://github.com/ethereum/solidity/pull/14309#discussion_r1237361089 for a moment. I still think that being able to use the property in the expectation would be vastly superior:

```solidity
// preIncrement.isSimpleCounterLoop: true
// postIncrement.isSimpleCounterLoop: true
// addOne.isSimpleCounterLoop: false
// preDecrement.isSimpleCounterLoop: false
// postDecrement.isSimpleCounterLoop: false
// multiplyByTwo.isSimpleCounterLoop: false
```

Even with names you have to jump to the code to see what the `false`/`true` refers to. Here you have the full condition in one spot and it actually reads like a condition. The process of reading the test becomes much more linear with even less jumping.

Changelog.md
cameel: Also, the entry needs to be moved to 0.8.22 now.

libsolidity/codegen/ir/IRGeneratorForStatements.cpp
```diff
@@ -3209,7 +3212,13 @@ void IRGeneratorForStatements::generateLoop(
 		_initExpression->accept(*this);
 	appendCode() << "} 1 {\n";
 	if (_loopExpression)
+	{
+		Arithmetic previousArithmetic = m_context.arithmetic();
+		if (_isSimpleCounterLoop)
+			m_context.setArithmetic(Arithmetic::Wrapping);
```
cameel: I guess this should be done only with optimizer activated, just like for the legacy pipeline, which @chriseth pointed out in https://github.com/ethereum/solidity/pull/13378#discussion_r1023778507?

libsolidity/codegen/ContractCompiler.cpp
```diff
@@ -1245,7 +1245,16 @@ bool ContractCompiler::visit(ForStatement const& _forStatement)
 
 	// for's loop expression if existing
 	if (_forStatement.loopExpression())
+	{
+		Arithmetic previousArithmetic = m_context.arithmetic();
+		if (
+			*_forStatement.annotation().isSimpleCounterLoop &&
+			m_optimiserSettings == OptimiserSettings::standard()
```
cameel: Hmm... making this dependent on having *all* the optimizations enabled seems suboptimal. And totally non-obvious - at the very least this fact should be documented. 

Not sure how to do it better though since the `enabled` flag does not exist internally. Should we maybe do this if any of the optimizer flags is enabled instead?

cameel (commented):
Here's more comments and some fixups instead of less important ones.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +432,61 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			return false;
+		if (!_forStatement.loopExpression())
+			return false;
+
+		auto const* simplePostExpression = dynamic_cast<UnaryOperation const*>(&_forStatement.loopExpression()->expression());
+		// This matches both operators ++i and i++
+		if (!simplePostExpression || simplePostExpression->getOperator() != Token::Inc || simplePostExpression->userDefinedFunctionType())
```
cameel, line 512: I just realized that we're not checking whether the same variable is used in the condition and in the increment.

This means that things like this could slip through and silently overflow:

```solidity
uint j = type(uint).max;
for (uint i = 0; i < 10; ++j) { ... }
```

We should have this as a test case by the way.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
cameel: We should really test all the corner cases which may disable the simple counter, these are really simple to test now with AST property tests:
- user-defined `++` or `<` operators.
- empty loop expression
- complex condition or condition with `<=`
- condition LHS not being a variable name or not an integer variable
- different common type in the condition
- non-integer counter

Also, would be good to make sure that some things indeed do not bypass the optimization:
- variable as a condition RHS
- array length as a condition RHS
- counter being accessed but not written to in the body (in Solidity and in inline assembly)

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
ekpyron: I still don't see that - if you put the path in the AST in the expectation, you need to jump from the expectation to the code to trace down the subpath, it's just a different kind of jumping. Especially, since the hook point for the AST paths are on statement level and in the future we may want to check properties on the expression level.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
matheusaaguiar: I added all suggested tests, although with some remarks for:
 * user-defined `++` or `<` operators. -> Operator `++` cannot be user-defined.
 * non-integer counter -> I don't see in the docs a non-integer type of counter that would have both `<` and `++` operator... I used `address` which has `<`, but not `++`.
 * Counter being written in inline assembly -> Currently assignments are detected. But I am not sure if there are other ways to write to the counter. I guess one could play with `mstore` for example?

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Added.

libsolidity/codegen/ContractCompiler.cpp
matheusaaguiar: After looking at how the optimizer settings are handled and configured in the compiler by `CommandLineParser` class, I changed it to check that `m_optimiserSettings != OptimiserSettings::none()`. I think that makes the compiler enable the `simpleCounterLoop` property (when adequate) if there is any optimization requested.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
```diff
@@ -0,0 +1,180 @@
...
+    uint z = 0;
+
+    function modifyStateVarZ() public returns (uint) {
+        z = type(uint).max;
+        return z;
+    }
+
+    function f() public {
+        // Positive Cases
+        /// SimplePreIncrement: isSimpleCounterLoop
```
matheusaaguiar: I renamed the test cases so it is more clear what they check for.
I guess it would also be nice to split them into two files, one containing the positive cases and another containing the negative ones.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
```diff
@@ -0,0 +1,180 @@
...
+        for(UINT i = UINT.wrap(0); i < UINT.wrap(10); i = i + UINT.wrap(1)) {
+        }
+        /// CounterAssignmentConditionRHS: isSimpleCounterLoop
+        for(uint i = 0; i < (i = i + 1); ++i) {
+        }
+        /// LiteralDifferentCommonTypeConditionRHS: isSimpleCounterLoop
+        for(uint8 i = 0; i < 257 ; ++i) {
+        }
+        /// StateVarCounterModifiedFunctionConditionRHS: isSimpleCounterLoop
+        for (z = 1; z < modifyStateVarZ(); ++z) {
```
matheusaaguiar: Needs fixing.
What would be the best approach here, regarding the rhs of the condition: (a) check forbidden/allowed cases or (b) restrict to only integer literals and variables?

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop.sol
```diff
@@ -0,0 +1,180 @@
...
+        }
+        /// StateVarCounterModifiedFunctionLoopBody: isSimpleCounterLoop
+        for (z = 1; z < 2048; ++z) {
+            modifyStateVarZ();
```
matheusaaguiar: I think `LValueChecker` needs to check if invoked functions are pure/view only.

ekpyron: Oh wow, yeah - I hadn't thought about state variables as loop variables actually :-). That won't happen in practice, but we indeed have to handle it properly nonetheless. But we can do something simpler: we can just require the loop counter to be a local variable (i.e. not a state variable) - that should be quick and painless to check and we won't loose any real-world cases. (That also means we won't need to account for ``sstore``s in assembly, which would be hell to reliably track - or downright impossible really)

libsolidity/codegen/ContractCompiler.cpp
ekpyron: About this: can you add a new optimizer setting for that instead?
It should just be enabled by default for an enabled optimizer (so for the CLI interface ``--optimize`` enables it and there doesn't have to be any finer control for disabling it) and for standard json, it could also be an implied default of ``optimizer { enabled: true }``, but can be more finely controled as ``optimizer { enabled: true, details: { simpleCounterLoopUncheckedIncrement: false } }``. That would make it behave very similarly to e.g. the peephole optimizer and should be quick to do and much cleaner.

This would involve:
- Extending the `OptimiserSettings` with a new field (defaulting to ``false``, but turned on in ``standard()`` and ``full()``).  (This should automatically produce the desired behaviour for CLI and standard-json without a optimizer ``details`` field)
- Parsing the `optimizer.details.simpleCounterLoopUncheckedIncrement` setting in ``StandardCompiler`` (search for ``"peephole"`` as a reference)
- Adding that field to the docs in https://docs.soliditylang.org/en/v0.8.21/using-the-compiler.html#input-description with a short comment explaining it (and mentioning it being enabled by default as for ``"peephole"``).
- Exposing the setting in the `else` branch within `CompilerStack::createMetadata` (to make sure it can be properly set for source verification)

ekpyron:
Just to make this visible more prominently: I think this mainly still needs https://github.com/ethereum/solidity/pull/13378#discussion_r1346218067 (exclude state variables) and I also just suggested https://github.com/ethereum/solidity/pull/13378#discussion_r1350718123 (i.e. make it a proper optimizer-detail setting for cleaner control than "do it whenever any optimization is enabled")

libsolidity/codegen/ContractCompiler.cpp
cameel: > Exposing the setting in the else branch within CompilerStack::createMetadata (to make sure it can be properly set for source verification)

As an aside here - it is very easy to forget about this bit if unaware and a common pitfall. I've seen this in multiple PRs, and sometimes been bitten by it myself, though I think in the end we always caught it. Still, maybe we should think about a mechanism that would make forgetting this more obvious. Either some automatic mechanism to include all settings by default (so that you can see it in tests and exclude if wrong) or some assert that will fail if a setting is not accounted for.

ekpyron: A nightly CI run similar to bytecode comparison that compares direct compilation with recompilation on the basis of metadata should already help with that... (of course nothing to do about this in this PR, just to be clear)

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +434,111 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			incExpressionIdentifier &&
+			incExpressionIdentifier->annotation().referencedDeclaration != lhsIdentifier->annotation().referencedDeclaration
+		)
+			return false;
+
+		if (
+			auto const* incVariableDeclaration = dynamic_cast<VariableDeclaration const*>(
+				incExpressionIdentifier->annotation().referencedDeclaration
+			);
+			!incVariableDeclaration->isLocalVariable()
```
nikola-matic, line 535: Shouldn't we check that `incVariableDeclaration` isn't `nullptr` before dereferencing?

libsolidity/ast/ASTJsonExporter.cpp
```diff
@@ -742,12 +742,18 @@ bool ASTJsonExporter::visit(WhileStatement const& _node)
...
+	std::vector<std::pair<std::string, Json::Value>> attributes = {
 		std::make_pair("initializationExpression", toJsonOrNull(_node.initializationExpression())),
 		std::make_pair("condition", toJsonOrNull(_node.condition())),
 		std::make_pair("loopExpression", toJsonOrNull(_node.loopExpression())),
 		std::make_pair("body", toJson(_node.body()))
-	});
+	};
+
+	if (_node.annotation().isSimpleCounterLoop.set())
+		attributes.emplace_back(std::make_pair("isSimpleCounterLoop", *_node.annotation().isSimpleCounterLoop));
```
nikola-matic: Does this need a `std::make_pair`? Wouldn't the following work just as fine (in fact, using `make_pair` here just turns the `emplace_back` into a `push_back`).

```suggestion
		attributes.emplace_back("isSimpleCounterLoop", *_node.annotation().isSimpleCounterLoop);
```

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
```diff
@@ -0,0 +1,134 @@
...
+
+contract C {
+    uint z = 0;
+
+    function modifyStateVarZ() public returns (uint) {
+        z = type(uint).max;
+        return z;
+    }
+
+    function f() public {
```
nikola-matic, line 33: We should probably add some tests using the shorthand notation for binary operations, e.g. `i += 1` in the post section of the loop statement.

Changelog.md
```diff
@@ -6,6 +6,7 @@ Language Features:
 
 Compiler Features:
  * Parser: Remove the experimental error recovery mode (``--error-recovery`` / ``settings.parserErrorRecovery``).
+ * Code Generator: Remove redundant overflow checks in specific for loops.
```
nikola-matic: Move this up one row.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: It is already checked. This is equivalent to 
```c++
auto const* incVariableDeclaration = ... ;
if (incVariableDeclaration && !incVariableDeclaration->isLocalVariable())
{ ... }
```

libsolidity/ast/ASTJsonExporter.cpp
matheusaaguiar: Indeed, `std::make_pair` is not necessary here and not coherent with `emplace_back`.
Fixed.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
matheusaaguiar: Added.

libsolidity/analysis/PostTypeChecker.cpp
nikola-matic: Ah, fair enough, I assumed that the assignment expression in the if statement would not be checked, but you are right.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +434,111 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+{
+public:
+	YulLValueChecker(ASTString const& _identifierName): m_identifierName(_identifierName) {}
+	bool willBeWrittenTo() { return m_willBeWrittenTo; }
+	using solidity::yul::ASTWalker::operator();
+	void operator()(solidity::yul::Assignment const& _assignment) override
+	{
+		for (auto const& yulIdentifier: _assignment.variableNames)
+			if (yulIdentifier.name.str() == m_identifierName)
+				m_willBeWrittenTo = true;
```
nikola-matic: How many `yulIdentifier`s can there be here? Wouldn't it make more sense to return as soon as we assign to `m_willBeWrittenTo`?

```suggestion
			if (yulIdentifier.name.str() == m_identifierName)
			{
				m_willBeWrittenTo = true;
				return;
			}
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +434,111 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
 	}
 };
 
+class YulLValueChecker : public solidity::yul::ASTWalker
+{
+public:
+	YulLValueChecker(ASTString const& _identifierName): m_identifierName(_identifierName) {}
+	bool willBeWrittenTo() { return m_willBeWrittenTo; }
```
nikola-matic:
```suggestion
	bool willBeWrittenTo() const { return m_willBeWrittenTo; }
```

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Yeah, makes more sense, no need to check further.

nikola-matic (dismissed):
I've no more complaints, so looks good to me. Hopefully @cameel gets enough time for another round tomorrow.

You'll need to rebase against develop to fix the changelog, and it might be a good time to start squashing some commits, since this is probably gonna get merged soon.

matheusaaguiar:
Rebased and squashed some commits.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -438,13 +438,16 @@ class YulLValueChecker : public solidity::yul::ASTWalker
 {
 public:
 	YulLValueChecker(ASTString const& _identifierName): m_identifierName(_identifierName) {}
-	bool willBeWrittenTo() { return m_willBeWrittenTo; }
+	bool willBeWrittenTo() const { return m_willBeWrittenTo; }
 	using solidity::yul::ASTWalker::operator();
 	void operator()(solidity::yul::Assignment const& _assignment) override
 	{
 		for (auto const& yulIdentifier: _assignment.variableNames)
```
nikola-matic: Huh, actually, now that I think about, this is just
```
if (std::any_of(...))
    m_willBeWrittenTo = true;
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,113 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+	YulLValueChecker(ASTString const& _identifierName): m_identifierName(_identifierName) {}
+	bool willBeWrittenTo() const { return m_willBeWrittenTo; }
+	using solidity::yul::ASTWalker::operator();
+	void operator()(solidity::yul::Assignment const& _assignment) override
+	{
+		if (ranges::any_of(
+			_assignment.variableNames,
+			[&](auto const& yulIdentifier) {return yulIdentifier.name.str() == m_identifierName; }
+		))
+		 	m_willBeWrittenTo = true;
```
stackenbotten, line 451: Coding style error

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Yeah, that's true!
Fixed, but I used `ranges::any_of` instead.

docs/using-the-compiler.rst
```diff
@@ -287,6 +287,9 @@ Input Description
             "cse": false,
             // Optimize representation of literal numbers and strings in code.
             "constantOptimizer": false,
+            // Allow unchecked arithmetic when incrementing the counter of for loops
```
cameel: "Allow" sounds odd in this context, as if it was just allowing you to use `unchecked`. IMO "use" is clearer.
```suggestion
            // Use unchecked arithmetic when incrementing the counter of for loops
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,113 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			*_identifier.annotation().referencedDeclaration == *m_declaration &&
+			_identifier.annotation().willBeWrittenTo
+		)
+			m_willBeWrittenTo = true;
+	}
+	void endVisit(InlineAssembly const& _inlineAssembly) override
+	{
+		YulLValueChecker yulChecker{m_declaration->name()};
+		yulChecker(_inlineAssembly.operations());
+		m_willBeWrittenTo = yulChecker.willBeWrittenTo();
```
cameel, line 487: Looks like an assembly block may reset the write flag set by anything that appears before it. Please check if this is correctly detected as a non-simple loop:

```c++
for (uint i = 0; i < 42; ++i) {
    i = type(uint).max;
    assembly {}
}
```

The correct code is probably this:
```suggestion
		m_willBeWrittenTo = m_willBeWrittenTo || yulChecker.willBeWrittenTo();
```

Or, better, only run `YulLValueChecker` when `m_willBeWrittenTo` is `false`, because otherwise there's no point - the answer is already known. You can do that in `operator()(yul::Assignment)` too to skip the `any_of` check.

libsolidity/analysis/PostTypeChecker.cpp
cameel: I don't think this is true. If you include the `;` then the declaration is not checked as a condition. At least that's what I always assumed.

I wasn't 100% sure myself since I rarely use that form, so I checked just in case with a small example and it seems to confirm that:

```c++
#include <iostream>

int main() {
    bool b = true;
    if (bool a = false; b)
        std::cout << "true" << std::endl;
    else
        std::cout << "false" << std::endl;
}
```

```
true
```

The result is "true" despite so the condition must be `b`, not `a && b`.

cameel: Which means that you do need a check against `incVariableDeclaration` being `nullptr`.

BTW, if this does not break anything in tests, you're missing some coverage. To get something without a declaration try things like `msg` or `block.number` or a function pointer. Or if you use a function name, it will give you `null` when you try to covert its definition to a variable. Could be impossible to trigger though - all of these will probably give you analysis errors before you even get to post type checker - in that case it'll be just an assert.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,113 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			LValueChecker rhsConditionChecker{*lhsIdentifier};
+			simpleCondition->rightExpression().accept(rhsConditionChecker);
+			if (rhsConditionChecker.willBeWrittenTo())
+				return false;
+		}
+
+		auto const* incExpressionIdentifier = dynamic_cast<Identifier const*>(&simplePostExpression->subExpression());
+		if (
+			incExpressionIdentifier &&
+			incExpressionIdentifier->annotation().referencedDeclaration != lhsIdentifier->annotation().referencedDeclaration
```
cameel: This check is skipped when the increment expression is an identifier. What if it's an expression?

```solidity
uint j = type(uint).max;
for (uint i = 0; i < 10; ++(j)) { ... }
```

Or member access?

```solidity
struct S {
    uint x;
}

...

S memory s = S(type(uint).max);
for (uint i = 0; i < 10; ++s.x) { ... }
```

Not sure if these will really bypass the check - both will probably fail on the next check since `(j)` has no referenced declaration and `s.x` might not count as a local variable. But there might be other situations I haven't thought about.

What you probably need here is to fail the check when the increment expression is not an identifier:

```suggestion
			!incExpressionIdentifier ||
			incExpressionIdentifier->annotation().referencedDeclaration != lhsIdentifier->annotation().referencedDeclaration
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,113 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+		)
+			return false;
+
+		if (
+			auto const* incVariableDeclaration = dynamic_cast<VariableDeclaration const*>(
+				incExpressionIdentifier->annotation().referencedDeclaration
+			);
+			!incVariableDeclaration->isLocalVariable()
```
cameel, line 535: Does this check allow function parameters or return values? Those might actually be fine, at least when stored on stack.

And what about members of local structs? Such structs are memory pointers under the hood so I guess they don't count as locals?

libsolidity/interface/OptimiserSettings.h
```diff
@@ -87,6 +87,7 @@ struct OptimiserSettings
 		s.runConstantOptimiser = true;
 		s.runYulOptimiser = true;
 		s.optimizeStackAllocation = true;
+		s.simpleCounterForLoopUncheckedIncrement = true;
```
cameel: Docs for this setting say "It is always on if no details are given." Currently it's not true because you don't set it to `true` in `OptimiserSettings::minimal()` above.

libsolidity/interface/OptimiserSettings.h
```diff
@@ -123,6 +124,11 @@ struct OptimiserSettings
 			expectedExecutionsPerDeployment == _other.expectedExecutionsPerDeployment;
```
cameel, line 126: You did not include the new setting in the definition of `operator==()`.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
```diff
@@ -0,0 +1,138 @@
...
+        uint x;
+        /// CounterIncrementRHSAssignment: isSimpleCounterLoop
+        for(uint i = 0; i < 10; ++i) {
+            x = i++;
+        }
+        /// NoEffectLoopExpression: isSimpleCounterLoop
+        for(uint i = 0; i < 42; i) {
+        }
+        /// EmptyLoopExpression: isSimpleCounterLoop
+        for(uint i = 0; i < 10; ) {
```
cameel, line 76: How about empty init or condition?

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
```diff
@@ -0,0 +1,138 @@
...
+        for(uint i = 0; h() < 100; ++i) {
+        }
+        /// FreeFunctionConditionDifferentCommonTypeLHS: isSimpleCounterLoop
+        for(uint8 i = 0; i < h(); ++i) {
+        }
+        /// NonIntegerTypeCondition: isSimpleCounterLoop
+        for(uint i = 0; address(this) < msg.sender; ++i) {
+        }
+        /// UDVTOperators: isSimpleCounterLoop
+        for(UINT i = UINT.wrap(0); i < UINT.wrap(10); i = i + UINT.wrap(1)) {
```
cameel: I'd use variables here to make it's as close to a simple loop as possible. Otherwise it's possible that it gets rejected for reasons other than the use of custom operators.

```suggestion
        UINT ZERO = UINT.wrap(0);
        UINT ONE = UINT.wrap(1);
        UINT TEN = UINT.wrap(10);
        for(UINT i = ZERO; i < TEN; i = i + ONE) {
```

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_positive_cases.sol
```diff
@@ -0,0 +1,55 @@
...
+        }
+        uint[8] memory array;
+        /// StaticArrayLengthConditionRHS: isSimpleCounterLoop
+        for(uint i = 0; i < array.length; i++) {
+        }
+        dynArray.push();
+        /// DynamicArrayLengthConditionRHS: isSimpleCounterLoop
+        for(uint i = 0; i < dynArray.length; i++) {
+            dynArray.push(i);
+        }
```
cameel, line 35: Oh, did not expect that one to count as a simple loop...

I guess it's safe because any expression not dependent on `i` will either become equal to `i` at some point and stop the loop or overflow before `i` does? And the fact that the length increases does not change that?

cameel (changes requested):

cameel:
I found some problems that look serious if left as is (one likely allows to bypass the check against modifying the counter in loop body), but they're all straightforward to fix, so I don't think that'll affect the release timing.

docs/using-the-compiler.rst
matheusaaguiar: Agreed.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: You are right. This will overwrite any previous value in the flag and also not needed when it is already true.
Fixed.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: You are right, it is not the same when you have the semicolon.
Sorry about that, especially to @nikola-matic who first spotted it.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Both of those cases, expressions and member access, were not covered in the tests. Hence, I hadn't gotten in the problem you point here, which is actually more severe, because, as you said they have no referenced declaration, leading to a seg fault.
Fixed.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: Function parameters and return values are allowed by this check.
members of local structs are not allowed.

libsolidity/analysis/PostTypeChecker.cpp
nikola-matic: > I don't think this is true. If you include the `;` then the declaration is not checked as a condition. At least that's what I always assumed.
> 
> I wasn't 100% sure myself since I rarely use that form, so I checked just in case with a small example and it seems to confirm that:
> 
> ```c++
> #include <iostream>
> 
> int main() {
>     bool b = true;
>     if (bool a = false; b)
>         std::cout << "true" << std::endl;
>     else
>         std::cout << "false" << std::endl;
> }
> ```
> 
> ```
> true
> ```
> 
> The result is "true" despite so the condition must be `b`, not `a && b`.

Ha, I actually did the same thing - the docs say that `init-statement` can be `nullptr`, but don't explicitly state how it affects the `if` as a whole, and didn't feel like digging through the standard, so ended up writing this example:

```c++
class C {
public:
	bool foo() { return true; }
};

int main()
{
	if (C *c = nullptr; c->foo()) {}
	return 0;
}
```
which did not in fact segfault, which is why I then assumed that Matheus was in fact right. So now that you posted your comment - I started extending the example a bit to see if I'm unable to write a simple program that will segfault (can write them by accident no problem though). Figured maybe it's because the actual `if` block is empty - so the compiler will just optimize it away, since `c` is in local scope anyway, and `foo()` (even though not explicitly `const`, is still likely treated as `const`, since it's pretty easy for the compiler to figure this out). So I added a `cout` to to the block, recompiled, ran it again, and still no segfault.
Then, I added another `nullptr` dereference by adding
```c++
C *a = nullptr;
c->foo();
```
before the `if` statement (so right in `main`), and still no segfault. Then, I figured, I have to add a member variable, and return *it* instead of just a `true` literal - and sure enough, this finally segfaulted. I tested it both with `gcc` and `clang`, and the behaviour is consistent between them, so even though dereferencing a `nullptr` is clearly undefined behaviour, the compiler implementers decided that it makes perfect sense to do the most unexpected thing when dereferencing a nullptr.

Thanks for coming to my TED talk, I sometime really hate this language.

libsolidity/interface/OptimiserSettings.h
matheusaaguiar: Fixed.

libsolidity/interface/OptimiserSettings.h
matheusaaguiar: Fixed.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
matheusaaguiar: Added.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
matheusaaguiar: Fixed.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_positive_cases.sol
matheusaaguiar: Yeah, that's what I am assuming that happens. The code checks if the common type of the expression is the same as the left hand side and that should guarantee that either `i` catches up or the array `length` overflows.
For example, if `i` had type `uint8`, then the flag `isSimpleCounterLoop` would be `false`.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,124 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+class LValueChecker: public ASTConstVisitor
+{
+public:
+	LValueChecker(Identifier const& _identifier):
+		m_declaration(_identifier.annotation().referencedDeclaration)
+	{}
+	bool willBeWrittenTo() const { return m_willBeWrittenTo; }
+	void endVisit(Identifier const& _identifier) override
+	{
+		if(m_willBeWrittenTo)
```
stackenbotten, line 470: Coding style error

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_negative_cases.sol
```diff
@@ -136,3 +161,7 @@ contract C {
 // StateVarCounterModifiedFunctionConditionRHS: false
 // StateVarCounterModifiedFunctionLoopBody: false
 // NonIntegerCounter: false
+// AssemblyAfterAssignmentToCounter: false
+// ExpressionIncrement: false
+// StructMemberAccessIncrement: false
+// DynamicArrayLengthConditionRHS: false
```
cameel: Tests are failing because this AST check is not defined. I'm pushing a fixup to remove the line.

cameel (dismissed):
All the issues I pointed out were addressed so I'm approving, but @ekpyron may want to take a final look just to make double sure I did not miss anything.

Also, needs some squashing and we should make sure that tests actually pass after the tweak I pushed to fix them.

cameel:
Looks like this needs an update to gas expectations.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_positive_cases.sol
ekpyron: The point in the logic for counting all these as simple loops is that the right-hand-side of the comparison does not matter at all. Including here, so yes, this is perfectly fine.

ekpyron: The right-hand-side can even contain `i` itself - doesn't change anything.

ekpyron: As in: the only situation in which an overflow-checked increment is relevant, is if `i` has the maximal value of its type before the increment is executed. Since it doesn't change in the body, it must have had that value in the condition already. But ``type(<type of i>).max`` *never* compares ``<`` to anything. So it should be fine even if the right-hand-side of the comparison reads `i` itself.

A weird fluke would be if the right-hand-side of the expression actually *changed* the value of `i`. For that soundness actually depends on the order of evaluation of the sides of the comparison, right?

I.e. do we currently allow
```
for(uint i = 0; i < ++i; i++) { }
```
?
That's fine, if the right-hand-side `++i` is evaluated before the left-hand-side, but wrong otherwise, so we should probably check.

cameel: > I.e. do we currently allow

We don't. We actually have a test against a very similar case:
```solidity
        /// CounterAssignmentConditionRHS: isSimpleCounterLoop
        for(uint i = 0; i < (i = i + 1); ++i) {
        }
```
and the check in the code is for `willBeWrittenTo` in the RHS, so it should catch `++i` as well.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,124 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			!dynamic_cast<Identifier const*>(&simpleCondition->rightExpression()) &&
+			!dynamic_cast<Literal const*>(&simpleCondition->rightExpression())
+		)
+		{
+			solAssert(lhsIdentifier);
+			LValueChecker rhsConditionChecker{*lhsIdentifier};
+			simpleCondition->rightExpression().accept(rhsConditionChecker);
+			if (rhsConditionChecker.willBeWrittenTo())
+				return false;
+		}
```
ekpyron: All of this can be removed. Instead the ``LValueChecker`` in the very end can just visit the right-hand-side as well, as in:
```
simpleCondition->rightExpression().accept(lhsLValueChecker);
```
in addition to just visiting the body.

test/libsolidity/astPropertyTests/forLoop/isSimpleCounterLoop_positive_cases.sol
ekpyron: Ah yeah, I didn't see it, since the check is written very weirdly, see https://github.com/ethereum/solidity/pull/13378/files#r1368592311

libsolidity/analysis/PostTypeChecker.cpp
ekpyron: It's not like the standard isn't perfectly clear in that
```
if (init; cond) ...
```
is equivalent to
```
{
  init;
  if (cond) ...
}
```
:-)

libsolidity/analysis/PostTypeChecker.cpp
cameel: Good point. We can just reuse the same `LValueChecker` instance for both. And the check for `Literal` and `Identifier` is redundant, because those will never modify the variable.

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,115 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+			auto const* incVariableDeclaration = dynamic_cast<VariableDeclaration const*>(
+				incExpressionIdentifier->annotation().referencedDeclaration
+			);
+			incVariableDeclaration &&
+			!incVariableDeclaration->isLocalVariable()
+		)
+			return false;
+
+		solAssert(lhsIdentifier);
+		LValueChecker lhsLValueChecker{*lhsIdentifier};
```
cameel:
```suggestion
		LValueChecker lValueChecker{*lhsIdentifier};
```

libsolidity/analysis/PostTypeChecker.cpp
```diff
@@ -421,6 +436,115 @@ struct ReservedErrorSelector: public PostTypeChecker::Checker
...
+		simpleCondition->rightExpression().accept(lhsLValueChecker);
+		_forStatement.body().accept(lhsLValueChecker);
+
+		return !lhsLValueChecker.willBeWrittenTo();
```
cameel: It would be more efficient to stop visiting as soon as we know the answer:

```suggestion
		simpleCondition->rightExpression().accept(lhsLValueChecker);
		if (!lhsLValueChecker.willBeWrittenTo())
			_forStatement.body().accept(lhsLValueChecker);

		return !lhsLValueChecker.willBeWrittenTo();
```

test/libsolidity/semanticTests/abiEncoderV1/abi_encode_calldata_slice.sol
```diff
@@ -59,10 +59,10 @@ contract C {
 // EVMVersion: >homestead
 // ----
 // test_bytes() ->
-// gas irOptimized: 360863
-// gas legacy: 411269
-// gas legacyOptimized: 317754
+// gas irOptimized: 321332
+// gas legacy: 319111
+// gas legacyOptimized: 269317
```
cameel, line 64: Looks like gains are sometimes non-negligible even in situations that aren't completely unrealistic. `test_bytes()` does 36 external calls (to the same contract, but still) and you'd think that would dominate the cost but apparently the cost of checked loop counter increments is still significant in comparison.

ekpyron: It's not without reason that people actually do the horrible ``unchecked``-block increment construction. But that also means, that in actual real-world situations, we likely won't see gas-improvements, but rather code-improvements (since the workaround is no longer necessary)

cameel: True.

libsolidity/analysis/PostTypeChecker.cpp
matheusaaguiar: RIght, Fixed.

matheusaaguiar:
All requests and suggestions have been addressed.
Rebased and squashed all commits.

nikola-matic (approved):

cameel (approved):

docs/using-the-compiler.rst
```diff
@@ -287,6 +287,9 @@ Input Description
             "cse": false,
             // Optimize representation of literal numbers and strings in code.
             "constantOptimizer": false,
+            // Use unchecked arithmetic when incrementing the counter of for loops
+            // under certain circumstances. It is always on if no details are given.
```
ekpyron, line 291: We should really have documented the precise conditions somewhere in the docs :-). But let's not hold things up due to that.

cameel: There's still time to do it in a follow-up PR :)

matheusaaguiar: I will open an issue, so we don't forget.

PR #13379, opened 2022-08-10
Make gas diff stats script executable.

ekpyron:
This has been annoying me ever since we added that script :-).


cameel (approved):

ekpyron:
I'm force-merging this over the current CI outage :-).

PR #13382, opened 2022-08-11
update: Assert statement used outside of tests

LucidSamuel:
Usage of assert statements in application logic is very much discouraged as It is recommended not to use assert in non-test files. Ideally, assert statement should be used only in tests.

### DIFF
--- a/docs/ext/remix_code_links.py
+++ b/docs/ext/remix_code_links.py
@@ -10,7 +10,8 @@ logger = logging.getLogger(__name__)
 
 
 def insert_node_before(child, new_sibling):
-    assert child in child.parent.children
+    if child not in child.parent.children:
+        raise AssertionError
 
     for position, node in enumerate(child.parent.children):
         if node == child:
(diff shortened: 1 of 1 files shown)


cameel:
Who discourages it? It's the first time I'm hearing of that and I've been coding in Python for a long time.

`assert` in Python predates the existence of unit testing frameworks and I think it was inspired by the `assert()` macro in languages like C/C++ where it has nothing to do with testing. It's a debugging tool. The [documentation for Python's `assert` statement](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) actually says just that:

> Assert statements are a convenient way to insert debugging assertions into a program:

It's exactly how we're using it here: to perform sanity checks. Function preconditions, postconditions, invariants, etc.

I know that frameworks like pytest have some magic that allows using the `assert` statement for checking expectations instead of dedicated functions like `assertEqual()` but to be honest I'd consider that more of an abuse of the statement than it's primary purpose. Yeah, may be convenient but it does not even work without the framework rewriting the code under the hood and the downside of that is that you then can't use `assert` for its intended purpose. Tests can have sanity checks as well and these are not the same as expectation checks. Unfortunately they both result in `AssertionError` but they still convey some semantic information: a failing `assert` means that there's something wrong in how you have written the test itself while failing `assertEqual()` means that the code under test did not satisfy the expectations.

In any case, thanks for the PR - I'm sure it's well intended - but we like it the way it is now. `assert` is a very useful tool and we use it this way all the time, not only in Python but also in C++ and Solidity.

PR #13383, opened 2022-08-11
Update: Assert statement used outside of tests

LucidSamuel:
Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.


cameel:
This works the way it was intended. Yes, these checks are no-ops in optimized code and that's the point of `assert`. They're sanity checks that we assume will never fail. If we thought it's possible to them to fail, they would not be assertions but actual validations. 

These checks are hopefully completely redundant but since bugs are always possible, we do want them in debug mode. They may sometimes be expensive which is why the optimized mode exists.

I'm going to close this for the same reason I stated in #13382.

PR #13384, opened 2022-08-11
Allow named parameters in mapping types

### LINKED ISSUE #11407: Allow specifying parameter name in public mappings
## Abstract

```solidity
contract Temp {
    mapping(address => uint256) public balanceOf;
}

// the above code produces the following ABI
[
  {
    inputs: [
      {
        internalType: 'address',
        name: '', // this is empty
        type: 'address',
      },
    ],
    name: 'balanceOf',
    outputs: [
      {
        internalType: 'uint256',
        name: '',
        type: 'uint256',
      },
    ],
    stateMutability: 'view',
    type: 'function',
  },
];
```

Since there is no param name, tools like typechain by default fill the param as `arg${i}`

```ts
// typescript
// an example of typing generated by typechain ethers v5 target
balanceOf(arg0: string, overrides?: CallOverrides): Promise<BigNumber>;
```

I understand it's not a big deal, but it'd be great to have a relevant param name like `userAddress` instead of `arg0` here. In cases where there are nested mappings (multiple inputs), parameter names can show up when hovered on the method, it can quickly help to understand which parameter is what.

```solidity
// solidity
mapping(address => mapping(address => uint256)) public allowance;

// typescript
// generated typing
allowance(arg0: string, arg1: string, overrides?: CallOverrides): Promise<BigNumber>;

// would be great to instead have
allowance(spender: string, owner: string, overrides?: CallOverrides): Promise<BigNumber>;
// edit: I just noticed I wrote the above order incorrectly xD
```

## Motivation

If solidity allows to somehow specify parameter names, through some syntax. For e.g.:

```solidity
mapping(address userAddress => uint256) public balanceOf;

// can also specify name for output
mapping(address userAddress => uint256 balance) public balanceOf;
```

This syntax is very similar to other elements in solidity that take input parameters:

```solidity
event Transfer(address, address, uint256);
function balanceOf(address) external view returns (uint256);

// with parameter names
event Transfer(address from, address to, uint256 value);
function balanceOf(address userAddress) external view returns (uint256 balance);
```

## Specification

In a mapping declaration, if there is a word between a data type and `=>` or `)`, then use that word as a name for that type.

## Backwards Compatibility

There is no backward compatibility as existing code will function the same. And adding parameters should not change bytecode as well, since it only fills the respective `name` parameters in the ABI.

zemse:
Closes #11407

- Allows providing an optional name/identifier just after the key type and the value type.
- Identifier defaults to empty string "".
- The optional identifier if provided is set to the "name" field in JSON ABI.
- Test cases are added for parser and ABI generator

Please let me know of any changes that are needed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ### 0.8.18 (unreleased)
 
 Language Features:
+* Allow named parameters in mapping types.
 
 
 Compiler Features:
--- a/docs/types/mapping-types.rst
+++ b/docs/types/mapping-types.rst
@@ -4,12 +4,13 @@
 Mapping Types
 =============
 
-Mapping types use the syntax ``mapping(KeyType => ValueType)`` and variables
-of mapping type are declared using the syntax ``mapping(KeyType => ValueType) VariableName``.
-The ``KeyType`` can be any
-built-in value type, ``bytes``, ``string``, or any contract or enum type. Other user-defined
-or complex types, such as mappings, structs or array types are not allowed.
-``ValueType`` can be any type, including mappings, arrays and structs.
+Mapping types use the syntax ``mapping(KeyType KeyName? => ValueType ValueName?)`` and variables of
+mapping type are declared using the syntax ``mapping(KeyType KeyName? => ValueType ValueName?)
+VariableName``. The ``KeyType`` can be any built-in value type, ``bytes``, ``string``, or any
+contract or enum type. Other user-defined or complex types, such as mappings, structs or array types
+are not allowed. ``ValueType`` can be any type, including mappings, arrays and structs. ``KeyName``
+and ``ValueName`` are optional (so ``mapping(KeyType => ValueType)`` works as well) and can be any
+valid identifier that is not a type.
 
 You can think of mappings as `hash tables <https://en.wikipedia.org/wiki/Hash_table>`_, which are virtually initialised
 such that every possible key exists and is mapped to a value whose
@@ -29,8 +30,10 @@ of contract functions that are publicly visible.
 These restrictions are also true for arrays and structs that contain mappings.
 
 You can mark state variables of mapping type as ``public`` and Solidity creates a
-:ref:`getter <visibility-and-getters>` for you. The ``KeyType`` becomes a parameter for the getter.
-If ``ValueType`` is a value type or a struct, the getter returns ``ValueType``.
+:ref:`getter <visibility-and-getters>` for you. The ``KeyType`` becomes a parameter
+with name ``KeyName`` (if specified) for the getter.
+If ``ValueType`` is a value type or a struct, the getter returns ``ValueType`` with
+name ``ValueName`` (if specified).
 If ``ValueType`` is an array or a mapping, the getter has one parameter for
 each ``KeyType``, recursively.
 
@@ -64,6 +67,25 @@ contract that returns the value at the specified address.
 The example below is a simplified version of an
 `ERC20 token <https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol>`_.
 ``_allowances`` is an example of a mapping type inside another mapping type.
+
+In the example below, the optional ``KeyName`` and ``ValueName`` are provided for the mapping.
+It does not affect any contract functionality or bytecode, it only sets the ``name`` field
+for the inputs and outputs in the ABI for the mapping's getter.
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity ^0.8.18;
+
+    contract MappingExampleWithNames {
+        mapping(address user => uint balance) public balances;
+
+        function update(uint newBalance) public {
+            balances[msg.sender] = newBalance;
+        }
+    }
+
+
 The example below uses ``_allowances`` to record the amount someone else is allowed to withdraw from your account.
 
 .. code-block:: solidity
--- a/libsolidity/analysis/DeclarationTypeChecker.cpp
+++ b/libsolidity/analysis/DeclarationTypeChecker.cpp
@@ -267,14 +267,59 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
 		solAssert(dynamic_cast<ElementaryTypeName const*>(&_mapping.keyType()), "");
 
 	Type const* keyType = _mapping.keyType().annotation().type;
+	ASTString keyName = _mapping.keyName();
+
 	Type const* valueType = _mapping.valueType().annotation().type;
+	ASTString valueName = _mapping.valueName();
 
 	// Convert key type to memory.
 	keyType = TypeProvider::withLocationIfReference(DataLocation::Memory, keyType);
 
 	// Convert value type to storage reference.
 	valueType = TypeProvider::withLocationIfReference(DataLocation::Storage, valueType);
-	_mapping.annotation().type = TypeProvider::mapping(keyType, valueType);
+	_mapping.annotation().type = TypeProvider::mapping(keyType, keyName, valueType, valueName);
+
+	// Check if parameter names are conflicting.
+	if (!keyName.empty())
+	{
+		auto childMappingType = dynamic_cast<MappingType const*>(valueType);
+		ASTString currentValueName = valueName;
+		bool loop = true;
+		while (loop)
+		{
+			bool isError = false;
+			// Value type is a mapping.
+			if (childMappingType)
+			{
+				// Compare top mapping's key name with child mapping's key name.
+				ASTString childKeyName = childMappingType->keyName();
+				if (keyName == childKeyName)
+					isError = true;
+
+				auto valueType = childMappingType->valueType();
+				currentValueName = childMappingType->valueName();
+				childMappingType = dynamic_cast<MappingType const*>(valueType);
+			}
+			else
+			{
+				// Compare top mapping's key name with the value name.
+				if (keyName == currentValueName)
+					isError = true;
+
+				loop = false; // We arrived at the end of mapping recursion.
+			}
+
+			// Report error.
+			if (isError)
+			{
+				m_errorReporter.declarationError(
+					1809_error,
+					_mapping.location(),
+					"Conflicting parameter name \"" + keyName + "\" in mapping."
+				);
+			}
+		}
+	}
 }
 
 void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
--- a/libsolidity/ast/AST.h
+++ b/libsolidity/ast/AST.h
@@ -1428,18 +1428,29 @@ public:
 		int64_t _id,
 		SourceLocation const& _location,
 		ASTPointer<TypeName> _keyType,
-		ASTPointer<TypeName> _valueType
+		ASTPointer<ASTString> _keyName,
+		ASTPointer<TypeName> _valueType,
+		ASTPointer<ASTString> _valueName
 	):
-		TypeName(_id, _location), m_keyType(std::move(_keyType)), m_valueType(std::move(_valueType)) {}
+		TypeName(_id, _location),
+		m_keyType(std::move(_keyType)),
+		m_keyName(std::move(_keyName)),
+		m_valueType(std::move(_valueType)),
+		m_valueName(std::move(_valueName))
+	{}
 	void accept(ASTVisitor& _visitor) override;
 	void accept(ASTConstVisitor& _visitor) const override;
 
 	TypeName const& keyType() const { return *m_keyType; }
+	ASTString keyName() const { return *m_keyName; }
 	TypeName const& valueType() const { return *m_valueType; }
+	ASTString valueName() const { return *m_valueName; }
 
 private:
 	ASTPointer<TypeName> m_keyType;
+	ASTPointer<ASTString> m_keyName;
 	ASTPointer<TypeName> m_valueType;
+	ASTPointer<ASTString> m_valueName;
 };
 
 /**
--- a/libsolidity/ast/ASTJsonExporter.cpp
+++ b/libsolidity/ast/ASTJsonExporter.cpp
@@ -598,7 +598,9 @@ bool ASTJsonExporter::visit(Mapping const& _node)
 {
 	setJsonNode(_node, "Mapping", {
 		make_pair("keyType", toJson(_node.keyType())),
+		make_pair("keyName", _node.keyName()),
 		make_pair("valueType", toJson(_node.valueType())),
+		make_pair("valueName", _node.valueName()),
 		make_pair("typeDescriptions", typePointerToJson(_node.annotation().type, true))
 	});
 	return false;
--- a/libsolidity/ast/ASTJsonImporter.cpp
+++ b/libsolidity/ast/ASTJsonImporter.cpp
@@ -648,7 +648,9 @@ ASTPointer<Mapping> ASTJsonImporter::createMapping(Json::Value const&  _node)
 	return createASTNode<Mapping>(
 		_node,
 		convertJsonToASTNode<TypeName>(member(_node, "keyType")),
-		convertJsonToASTNode<TypeName>(member(_node, "valueType"))
+		memberAsASTString(_node, "keyName"),
+		convertJsonToASTNode<TypeName>(member(_node, "valueType")),
+		memberAsASTString(_node, "valueName")
 	);
 }
 
--- a/libsolidity/parsing/Parser.cpp
+++ b/libsolidity/parsing/Parser.cpp
@@ -1187,11 +1187,21 @@ ASTPointer<Mapping> Parser::parseMapping()
 	}
 	else
 		fatalParserError(1005_error, "Expected elementary type name or identifier for mapping key type");
+	ASTPointer<ASTString> keyName;
+	if (m_scanner->currentToken() == Token::Identifier)
+		keyName = getLiteralAndAdvance();
+	else
+		keyName = make_shared<ASTString>("");
 	expectToken(Token::DoubleArrow);
 	ASTPointer<TypeName> valueType = parseTypeName();
+	ASTPointer<ASTString> valueName;
+	if (m_scanner->currentToken() == Token::Identifier)
+		valueName = getLiteralAndAdvance();
+	else
+		valueName = make_shared<ASTString>("");
 	nodeFactory.markEndPosition();
 	expectToken(Token::RParen);
-	return nodeFactory.createNode<Mapping>(keyType, valueType);
+	return nodeFactory.createNode<Mapping>(keyType, keyName, valueType, valueName);
 }
 
 ASTPointer<ParameterList> Parser::parseParameterList(
--- a/test/libsolidity/ABIJson/mapping.sol
+++ b/test/libsolidity/ABIJson/mapping.sol
@@ -1,0 +1,77 @@
+contract test {
+    mapping(address owner => mapping(address spender => uint value)) public allowance;
+    mapping(bytes32 => address sender) public commits;
+    mapping(bytes32 => bytes32) public something;
+}
+// ----
+//     :test
+// [
+//   {
+//     "inputs":
+//     [
+//       {
+//         "internalType": "address",
+//         "name": "owner",
+//         "type": "address"
+//       },
+//       {
+//         "internalType": "address",
+//         "name": "spender",
+//         "type": "address"
+//       }
+//     ],
+//     "name": "allowance",
+//     "outputs":
+//     [
+//       {
+//         "internalType": "uint256",
+//         "name": "value",
+//         "type": "uint256"
+//       }
+//     ],
+//     "stateMutability": "view",
+//     "type": "function"
+//   },
+//   {
+//     "inputs":
+//     [
+//       {
+//         "internalType": "bytes32",
+//         "name": "",
+//         "type": "bytes32"
+//       }
+//     ],
+//     "name": "commits",
+//     "outputs":
+//     [
+//       {
+//         "internalType": "address",
+//         "name": "sender",
+//         "type": "address"
+//       }
+//     ],
+//     "stateMutability": "view",
+//     "type": "function"
+//   },
+//   {
+//     "inputs":
+//     [
+//       {
+//         "internalType": "bytes32",
+//         "name": "",
+//         "type": "bytes32"
+//       }
+//     ],
+//     "name": "something",
+//     "outputs":
+//     [
+//       {
+//         "internalType": "bytes32",
+//         "name": "",
+//         "type": "bytes32"
+//       }
+//     ],
+//     "stateMutability": "view",
+//     "type": "function"
+//   }
+// ]
--- a/test/libsolidity/SolidityTypes.cpp
+++ b/test/libsolidity/SolidityTypes.cpp
@@ -97,12 +97,16 @@ BOOST_AUTO_TEST_CASE(storage_layout_mapping)
 		{"first", TypeProvider::fromElementaryTypeName("uint128")},
 		{"second", TypeProvider::mapping(
 			TypeProvider::fromElementaryTypeName("uint8"),
-			TypeProvider::fromElementaryTypeName("uint8")
+			"",
+			TypeProvider::fromElementaryTypeName("uint8"),
+			""
 		)},
 		{"third", TypeProvider::fromElementaryTypeName("uint16")},
 		{"final", TypeProvider::mapping(
 			TypeProvider::fromElementaryTypeName("uint8"),
-			TypeProvider::fromElementaryTypeName("uint8")
+			"",
+			TypeProvider::fromElementaryTypeName("uint8"),
+			""
 		)},
 	}));
 	BOOST_REQUIRE_EQUAL(u256(4), members.storageSize());
@@ -199,8 +203,8 @@ BOOST_AUTO_TEST_CASE(type_identifiers)
 	FunctionType metaFun(TypePointers{keccak256fun}, TypePointers{s.type()}, strings{""}, strings{""});
 	BOOST_CHECK_EQUAL(metaFun.identifier(), "t_function_internal_nonpayable$_t_function_keccak256_nonpayable$__$returns$__$_$returns$_t_type$_t_struct$_Struct_$3_storage_ptr_$_$");
 
-	Type const* m = TypeProvider::mapping(TypeProvider::fromElementaryTypeName("bytes32"), s.type());
-	MappingType m2(TypeProvider::fromElementaryTypeName("uint64"), m);
+	Type const* m = TypeProvider::mapping(TypeProvider::fromElementaryTypeName("bytes32"), "", s.type(), "");
+	MappingType m2(TypeProvider::fromElementaryTypeName("uint64"), "", m, "");
 	BOOST_CHECK_EQUAL(m2.identifier(), "t_mapping$_t_uint64_$_t_mapping$_t_bytes32_$_t_type$_t_struct$_Struct_$3_storage_ptr_$_$_$");
 
 	// TypeType is tested with contract
--- a/test/libsolidity/errorRecoveryTests/do_not_delete_at_error.sol
+++ b/test/libsolidity/errorRecoveryTests/do_not_delete_at_error.sol
@@ -10,4 +10,6 @@ contract Error2 {
 	mapping (address => uint balances; // missing ) before "balances"
 }
 // ----
-// ParserError 6635: (417-425): Expected ')' but got identifier
+// ParserError 6635: (425-426): Expected ')' but got ';'
+// ParserError 6635: (425-426): Expected identifier but got ';'
+// ParserError 6635: (458-459): Expected ';' but got '}'
--- a/test/libsolidity/syntaxTests/parsing/mapping_with_names_5.sol
+++ b/test/libsolidity/syntaxTests/parsing/mapping_with_names_5.sol
@@ -1,0 +1,4 @@
+contract test {
+    mapping(address owner => mapping(address spender => bytes32[] notes)) names;
+}
+// ----
--- a/test/libsolidity/syntaxTests/parsing/mapping_with_names_conflict_3.sol
+++ b/test/libsolidity/syntaxTests/parsing/mapping_with_names_conflict_3.sol
@@ -1,0 +1,7 @@
+contract test {
+    mapping(address owner => mapping(address owner => address owner)) owner;
+}
+// ----
+// DeclarationError 1809: (45-84): Conflicting parameter name "owner" in mapping.
+// DeclarationError 1809: (20-85): Conflicting parameter name "owner" in mapping.
+// DeclarationError 1809: (20-85): Conflicting parameter name "owner" in mapping.
--- a/test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_4.sol
+++ b/test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_4.sol
@@ -1,0 +1,6 @@
+contract test {
+    struct Person {
+        mapping(uint phone => mapping(uint call => uint time) callTimes) friends;
+    }
+}
+// ----
--- a/test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_6.sol
+++ b/test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_6.sol
@@ -1,0 +1,7 @@
+contract test {
+    struct Person {
+        mapping(uint nameSame => mapping(uint name1 => mapping(uint nameSame => uint name2) name3) name4) name5;
+    }
+}
+// ----
+// DeclarationError 1809: (44-141): Conflicting parameter name "nameSame" in mapping.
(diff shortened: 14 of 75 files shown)


test/libsolidity/errorRecoveryTests/do_not_delete_at_error.sol
```diff
@@ -10,4 +10,6 @@ contract Error2 {
 	mapping (address => uint balances; // missing ) before "balances"
 }
 // ----
-// ParserError 6635: (417-425): Expected ')' but got identifier
+// ParserError 6635: (425-426): Expected ')' but got ';'
```
zemse, line 13: Some existing error recovery tests change their behavior because a syntax that was illegal before (i.e. ability to place an identifier next to type in mapping) is legal now, hence the parser passes further and generates a different error.

libsolidity/ast/AST.h
```diff
@@ -1428,18 +1428,24 @@ class Mapping: public TypeName
 		int64_t _id,
 		SourceLocation const& _location,
 		ASTPointer<TypeName> _keyType,
-		ASTPointer<TypeName> _valueType
+		ASTPointer<TypeName> _valueType,
+		ASTPointer<ASTString> _keyName,
+		ASTPointer<ASTString> _valueName
 	):
-		TypeName(_id, _location), m_keyType(std::move(_keyType)), m_valueType(std::move(_valueType)) {}
+		TypeName(_id, _location), m_keyType(std::move(_keyType)), m_valueType(std::move(_valueType)), m_keyName(std::move(_keyName)), m_valueName(std::move(_valueName)) {}
```
hrkrshnn: Nit: can you make a new line?

libsolidity/parsing/Parser.cpp
```diff
@@ -1183,11 +1183,21 @@ ASTPointer<Mapping> Parser::parseMapping()
 	}
 	else
 		fatalParserError(1005_error, "Expected elementary type name or identifier for mapping key type");
+	ASTPointer<ASTString> keyName;
+	if(m_scanner->currentToken() == Token::Identifier)
```
hrkrshnn:
```suggestion
	if (m_scanner->currentToken() == Token::Identifier)
```

libsolidity/parsing/Parser.cpp
```diff
@@ -1183,11 +1183,21 @@ ASTPointer<Mapping> Parser::parseMapping()
...
+		keyName = getLiteralAndAdvance();
+	else
+		keyName = make_shared<ASTString>("");
 	expectToken(Token::DoubleArrow);
 	ASTPointer<TypeName> valueType = parseTypeName();
+	ASTPointer<ASTString> valueName;
+	if(m_scanner->currentToken() == Token::Identifier)
```
hrkrshnn:
```suggestion
	if (m_scanner->currentToken() == Token::Identifier)
```

libsolidity/parsing/Parser.cpp
```diff
@@ -1183,11 +1183,21 @@ ASTPointer<Mapping> Parser::parseMapping()
 	}
 	else
 		fatalParserError(1005_error, "Expected elementary type name or identifier for mapping key type");
+	ASTPointer<ASTString> keyName;
+	if(m_scanner->currentToken() == Token::Identifier)
+		keyName = getLiteralAndAdvance();
+	else
+		keyName = make_shared<ASTString>("");
```
hrkrshnn, line 1194: Can you add a test for something like `mapping (uint uint => uint)`. Should be a parser error.

test/libsolidity/ABIJson/mapping.sol
```diff
@@ -0,0 +1,33 @@
+contract test {
+    mapping(address owner => mapping(address spender => uint value)) public allowance;
```
hrkrshnn, line 2: Can you add one more mapping, but with some types that are not annotated?

hrkrshnn (commented):
Looks quite good! Some small comments.

We were wondering with the annotations, if it makes sense to have the function call options syntax for mappings. For example, for `mapping (address user => uint balance) balanceOf`, would the syntax `balanceOf[{user: 0xaddress}]` be allowed? It's not needed to implement it in this PR, but worth thinking about.

Also, it's missing a changelog entry.

test/libsolidity/syntaxTests/parsing/mapping_with_names_5.sol
```diff
@@ -0,0 +1,4 @@
+contract test {
+    mapping(address owner => mapping(address spender => bytes32[] notes)) names;
```
hrkrshnn, line 2: Can you add a test where these symbols conflict with an existing definition? As an example add a `address owner` as a state variable.

hrkrshnn:
Can you add a semantic test as well? https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/getters/mapping.sol here is an example

hrkrshnn:
Could you also add an example in the documentation?

zemse:
Thanks for the review! I've made the changes.

> function call options syntax for mappings

This might be tricky for nested mappings: `allowances[{owner: 0xAddr}][{spender: 0xAddr}]`. Somehow having it together could be good `allowances[{owner: 0xAddr, spender: 0xAddr}]`, however the names are optional hence this together syntax might not work. But yeah having it makes sense for improving code readability, does not affect any ABI or bytecode. I'd prefer adding it in a separate PR unless it is needed in this PR.

docs/types/mapping-types.rst
```diff
@@ -4,12 +4,13 @@
 Mapping Types
 =============
 
-Mapping types use the syntax ``mapping(KeyType => ValueType)`` and variables
-of mapping type are declared using the syntax ``mapping(KeyType => ValueType) VariableName``.
+Mapping types use the syntax ``mapping(KeyType KeyName => ValueType ValueName)`` and variables
```
Marenz: It might be nice to see both ways written out here, if one is not carefully reading, they might think this is not optional (the optional part is hidden in the full text rather than easily visible at a glance)

libsolidity/parsing/Parser.cpp
```diff
@@ -1183,11 +1183,21 @@ ASTPointer<Mapping> Parser::parseMapping()
 	}
 	else
 		fatalParserError(1005_error, "Expected elementary type name or identifier for mapping key type");
+	ASTPointer<ASTString> keyName;
+	if (m_scanner->currentToken() == Token::Identifier)
+		keyName = getLiteralAndAdvance();
+	else
+		keyName = make_shared<ASTString>("");
```
Marenz, line 1194: You create an `ASTString` here, but in most other places you use `std::string`. 
This is technically fine as `ASTString` is just an alias currently, but for consistency and to avoid future problems if we decide to change that alias, `ASTString` should be used everywhere :)

Marenz (changes requested):
already looking very good.
Additionally to the comments, I am wondering if it makes sense to make it an `optional<ASTPointer<ASTString>>`. 
Would love to hear @cameel and @christianparpart on that topic

cameel:
> Additionally to the comments, I am wondering if it makes sense to make it an `optional<ASTPointer<ASTString>>`.

Depends if we prefer to treat it as a missing name or an empty name. That will show up in the AST.

But overall `optional` would make it clearer that this is even possible. And since names need to be unique, having to treat `""` as an exception is not that great.

Also, I can't see any place in the AST where we'd have a single optional name but in a few cases we do have vectors of names and that's closer to `optional` than to having an empty string when the name is missing.

libsolidity/parsing/Parser.cpp
zemse: Thanks for pointing this out. I've replaced all `std::string` with `ASTString`.

docs/types/mapping-types.rst
zemse: I've added a commit which mentions that `mapping(KeyType => ValueType)` also works.

zemse:
I have just made the review changes requested and rebased over the latest develop branch.

> Depends if we prefer to treat it as a missing name or an empty name. 

I think if names are made optional, that'd need a small condition while generating ABI because if name is missing, then it has to create an empty string "" anyway for pushing it to param names array. But do lmk if this change is needed to be done in this PR.

zemse:
I have rebased over the latest `develop` branch to fix the conflicts.

Edit: I'm not able to request reviews from multiple people. cc: @hrkrshnn, @Marenz, @cameel, @nishant-sachdeva. (Sorry for tagging).

hrkrshnn:
Will rebase now

Changelog.md
```diff
@@ -1,6 +1,7 @@
 ### 0.8.18 (unreleased)
 
 Language Features:
+* Allow mapping key and value identifiers
```
```suggestion
* Allow named parameters in mapping types.
```

docs/types/mapping-types.rst
```diff
@@ -64,6 +67,25 @@ contract that returns the value at the specified address.
...
 ``_allowances`` is an example of a mapping type inside another mapping type.
+
+In the example below, the optional ``KeyName`` and ``ValueName`` are provided for the mapping. 
+It does not make affect any contract functionality or bytecode, it only sets the ``name`` field
+for the inputs and outputs in the ABI for the mapping's getter.
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity >=0.8.17;
```
hrkrshnn:
```suggestion
    pragma solidity ^0.8.17;
```
We like to restrict the pragmas in the documentation. Mainly because the documentation will still be available and there is no guarantee that the code with `>= 0.8.17` will compile on `0.9.0` for example.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -269,12 +269,15 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+	ASTString keyName = _mapping.keyName();
+	ASTString valueName = _mapping.valueName();
+
 	// Convert key type to memory.
 	keyType = TypeProvider::withLocationIfReference(DataLocation::Memory, keyType);
 
 	// Convert value type to storage reference.
 	valueType = TypeProvider::withLocationIfReference(DataLocation::Storage, valueType);
-	_mapping.annotation().type = TypeProvider::mapping(keyType, valueType);
+	_mapping.annotation().type = TypeProvider::mapping(keyType, valueType, keyName, valueName);
```
hrkrshnn: Nit: I would have preferred the order to be `keyType, KeyName, valueType, valueName`. But no strong argument on changing this.

libsolidity/ast/AST.h
```diff
@@ -1428,18 +1428,29 @@ class Mapping: public TypeName
...
+		ASTPointer<TypeName> _valueType,
+		ASTPointer<ASTString> _keyName,
+		ASTPointer<ASTString> _valueName
 	):
-		TypeName(_id, _location), m_keyType(std::move(_keyType)), m_valueType(std::move(_valueType)) {}
+		TypeName(_id, _location),
+		m_keyType(std::move(_keyType)),
+		m_valueType(std::move(_valueType)),
+		m_keyName(std::move(_keyName)),
+		m_valueName(std::move(_valueName))
```
hrkrshnn:
```suggestion
		m_valueName(std::move(_valueName))
```
:)

hrkrshnn: Kinda surprised the style checker didn't catch this. edit: it actually did.

libsolidity/ast/ASTJsonImporter.cpp
```diff
@@ -648,7 +648,9 @@ ASTPointer<Mapping> ASTJsonImporter::createMapping(Json::Value const&  _node)
 	return createASTNode<Mapping>(
 		_node,
 		convertJsonToASTNode<TypeName>(member(_node, "keyType")),
-		convertJsonToASTNode<TypeName>(member(_node, "valueType"))
+		convertJsonToASTNode<TypeName>(member(_node, "valueType")),
+		convertJsonToASTNode<ASTString>(member(_node, "keyName")),
+		convertJsonToASTNode<ASTString>(member(_node, "valueName"))
```
hrkrshnn: The equivalent code was missing in `ASTJsonExporter`. This is why the cli test was failing with
```
Looking at 2940 .sol files...
................................................................................................................................................ERROR: AST reimport failed for input file /home/hari/s/r2/test/libsolidity/syntaxTests/array/pop/storage_with_mapping_pop.sol

Compiler stderr:
Failed to import AST: JSON-Node needs to have 'nodeType' and 'id' fields.

Compiler stdout:

```
I'll fix this.

Kinda wish we had serde magic here :(

hrkrshnn: There is also a nasty bug here. `ASTString` is not an ASTNode. So this conversion would fail at runtime.

The correct way to do it would be:
```diff
-		convertJsonToASTNode<ASTString>(member(_node, "keyName")),
-		convertJsonToASTNode<ASTString>(member(_node, "valueName"))
+		memberAsASTString(_node, "keyName"),
+		memberAsASTString(_node, "valueName")
```

docs/types/mapping-types.rst
```diff
@@ -64,6 +67,25 @@ contract that returns the value at the specified address.
...
 ``_allowances`` is an example of a mapping type inside another mapping type.
+
+In the example below, the optional ``KeyName`` and ``ValueName`` are provided for the mapping.
+It does not make affect any contract functionality or bytecode, it only sets the ``name`` field
+for the inputs and outputs in the ABI for the mapping's getter.
+
+.. code-block:: solidity
+
+    // SPDX-License-Identifier: GPL-3.0
+    pragma solidity ^0.8.17;
```
hrkrshnn:
```suggestion
    pragma solidity ^0.8.18;
```

libsolidity/ast/ASTJsonExporter.cpp
```diff
@@ -599,6 +599,8 @@ bool ASTJsonExporter::visit(Mapping const& _node)
 	setJsonNode(_node, "Mapping", {
 		make_pair("keyType", toJson(_node.keyType())),
 		make_pair("valueType", toJson(_node.valueType())),
+		make_pair("keyName", _node.keyName()),
+		make_pair("valueName", _node.valueName()),
```
hrkrshnn: I wonder if this change would be considered breaking. cc @cameel @ekpyron

test/libsolidity/ABIJson/mapping.sol
hrkrshnn: I'll add this now

hrkrshnn:
Also, a fun test case would be the following
```solidity
type name is uint256;
contract C {
    mapping (name name => uint) public names;
}
```

The corresponding auto-generated function `function names(name name) returns (uint);` The compiler would complain that `Name has to refer to a struct, enum or contract.` which is a bit misleading.

Another example is
```solidity
mapping (uint x => uint x) public something;
```

The equivalent function would be `function f(uint x) returns (uint x)` which wouldn't compile because the name is defined twice in the same context.

I think we need to do some additional sanity checks on the names defined in this way. I think we should do this before merging it. @zemse do you feel like adding this in the implementation?

nikola-matic:
Heya @zemse, I've rebased against latest upstream develop to get the build in working order. Are you still interested in working on this PR, or should we have someone else takeover? Also, for future reference, please create a separate branch in your fork for features instead of working directly on `develop`.

docs/types/mapping-types.rst
```diff
@@ -64,6 +67,25 @@ contract that returns the value at the specified address.
 The example below is a simplified version of an
 `ERC20 token <https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol>`_.
 ``_allowances`` is an example of a mapping type inside another mapping type.
+
+In the example below, the optional ``KeyName`` and ``ValueName`` are provided for the mapping.
+It does not make affect any contract functionality or bytecode, it only sets the ``name`` field
```
nishant-sachdeva:
```suggestion
It does not affect any contract functionality or bytecode, it only sets the ``name`` field
```

docs/types/mapping-types.rst
```diff
@@ -4,12 +4,13 @@
...
 =============
 
-Mapping types use the syntax ``mapping(KeyType => ValueType)`` and variables
-of mapping type are declared using the syntax ``mapping(KeyType => ValueType) VariableName``.
-The ``KeyType`` can be any
-built-in value type, ``bytes``, ``string``, or any contract or enum type. Other user-defined
-or complex types, such as mappings, structs or array types are not allowed.
-``ValueType`` can be any type, including mappings, arrays and structs.
+Mapping types use the syntax ``mapping(KeyType KeyName => ValueType ValueName)`` and variables of
+mapping type are declared using the syntax ``mapping(KeyType KeyName => ValueType ValueName)
```
nishant-sachdeva:
```suggestion
Mapping types use the syntax ``mapping(KeyType KeyName? => ValueType ValueName?)`` and variables of
mapping type are declared using the syntax ``mapping(KeyType KeyName? => ValueType ValueName?)
```
Suggested this change to indicate that adding named parameters is optional

nishant-sachdeva (dismissed):

test/libsolidity/SolidityTypes.cpp
```diff
@@ -97,12 +97,16 @@ BOOST_AUTO_TEST_CASE(storage_layout_mapping)
 		{"first", TypeProvider::fromElementaryTypeName("uint128")},
 		{"second", TypeProvider::mapping(
 			TypeProvider::fromElementaryTypeName("uint8"),
-			TypeProvider::fromElementaryTypeName("uint8")
+			TypeProvider::fromElementaryTypeName("uint8"),
```
nikola-matic: These parameters are passed in the wrong order, i.e. they don't match the function declaration, which is why the build is failing.

nikola-matic (changes requested):
The grammar check is failing, since the grammar you've defined allows for named nested mappings, whereas your implementation does not. Given the other nested mapping tests, it seems like you don't want to allow named nested mappings. I'd be curious to know why, since this would then require a specific rule to handle nested mappings? I'll add it to our agenda for the next language design meeting, and we can discuss further there - if you want to join, it's the [Wednesday 15:00 CET meeting].(https://docs.soliditylang.org/en/latest/contributing.html#team-calls)

In any case, there were a few minor issues that I've fixed, which should now hopefully get the build into a workable state (excluding the failing grammar test).

zemse:
> it seems like you don't want to allow named nested mappings

```solidity
mapping(uint hello => mapping(uint world => uint whats) up) guys;

// getter for above mapping
function guys(uint hello, uint world) public returns (uint whats);
```

As you can see the mapping name "up" is not used at all. Since it means nothing, still allowing that code to be compiled will mislead people into expecting it to work. 

> since this would then require a specific rule to handle nested mappings?

If adding grammar for this case is complex then it makes sense to remove this requirement, since it's not really a big deal ig. And actually Hari helped me with the grammar previously, I added this check for named nested mappings after his help while adding additional test cases and was not aware I have to fix the language grammar as well (I'm not familiar with that, will need to figure it out). 

> I'll add it to our agenda for the next language design meeting

Thanks! I've added the call to my calendar.

> Are you still interested in working on this PR, or should we have someone else takeover? 

I am up for working on fixing minor issues in the PR. But if it's faster through someone from the solidity team, then we should do that probably. Specifically for the language grammar fix (due to the named nested mapping names), it would be great if I get some help.

> Also, for future reference, please create a separate branch in your fork for features instead of working directly on develop.

Sure, sorry about directly using the develop branch on my fork. Is there a way to change the fork's branch in this same PR? I think I will have to close this PR and create a new PR but that would lose the discussions in this thread. If that is okay, I can do it. And for future things, I'll make sure to not use default branches on fork.

nikola-matic:
I'm far from an expert so someone else will from the team will have to chime in.
>As you can see the mapping name "up" is not used at all. Since it means nothing, still allowing that code to be compiled will mislead people into expecting it to work.

Yup, looks like it. The only question I have is whether it's 'used', or rather, whether it shows up in the ABI/AST or not? The answer to this question will essentially dictate whether we alter the grammar, or the code.

>If adding grammar for this case is complex then it makes sense to remove this requirement, since it's not really a big deal ig. And actually Hari helped me with the grammar previously, I added this check for named nested mappings after his help while adding additional test cases and was not aware I have to fix the language grammar as well (I'm not familiar with that, will need to figure it out).

I'm sure this won't be a problem either way, it's just a matter of making a decision. We can discuss this on Wednesday.

>I am up for working on fixing minor issues in the PR. But if it's faster through someone from the solidity team, then we should do that probably. Specifically for the language grammar fix (due to the named nested mapping names), it would be great if I get some help.

Awesome - as long as you're responsive, it's not really a big deal. The only reason I asked was because the PR has been inactive for quite some time. Whatever decision we make, I'm guessing it won't require a lot of work; we'll help out with whatever you need.

>Sure, sorry about directly using the develop branch on my fork. Is there a way to change the fork's branch in this same PR? I think I will have to close this PR and create a new PR but that would lose the discussions in this thread. If that is okay, I can do it. 
And for future things, I'll make sure to not use default branches on fork.

No need to do anything regarding this now, I just mentioned it for future reference in case you decide to continue contributing.

chriseth:
Just randomly throwing in some important test cases we need to cover:
 - named mapping parameters for storage pointers of mapping types as local variables (and function parameters)
 - named mapping parameters for state variables of function type that takes a mapping as parameter
 - what about mappings as struct members?

Not saying anything about this should or should not work, but we should certainly have tests for these.

nikola-matic:
As suggested by @ekpyron in the language design meeting:
1. We should allow naming keys and values in nested mappings as well. Also it should be possible to name the value even when it’s a mapping or an array.
2. More tests

zemse:
Summary of the changes done:
1. Removed the commit ee0fee7f7613f45439143e42b37ed84ff6236a7b that disallowed param for mapping value types.
2. Added more tests in 4aedd7755ce215995cae080f644fb984f22e4522.
3. Rebased over upstream/develop branch. (For comparison of changes before & after the force push, I've pushed [backup branch](https://github.com/zemse/solidity/commits/develop-backup-19-nov))

test/libsolidity/syntaxTests/parsing/mapping_with_names_conflict_3.sol
```diff
@@ -0,0 +1,7 @@
+contract test {
+    mapping(address owner => mapping(address owner => address owner)) owner;
+}
+// ----
+// DeclarationError 5609: (45-84): Conflicting parameter name "owner" in mapping.
+// DeclarationError 1809: (20-85): Conflicting parameter name "owner" in recursive mapping.
```
nikola-matic: ~~Why does this one have a different error code, i.e. 1809 vs 5609?~~ Ah OK, it's a different error in case the nested type is a mapping, so you're differentiating between conflicting key -> key and key -> value names. The error message is the same though, which is perfectly fine, but in that case, the errors should have the same code.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,57 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+		while (true) {
+			// Value type is a mapping.
+			if (childMappingType)
+			{
+				// Compare top mapping's key name with child mapping's key name.
+				ASTString childKeyName = childMappingType->keyName();
+				if (keyName == childKeyName)
+				{
+					m_errorReporter.declarationError(
+						1809_error,
```
nikola-matic, line 296:
```suggestion
						5609_error,
```

test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_4.sol
```diff
@@ -0,0 +1,6 @@
+contract test {
+    struct Person {
+        mapping(uint phone => mapping(uint call => uint time) callTimes) friends;
```
nikola-matic, line 3: Adding another test case with an additional level of nesting would be nice, especially if it hits the conflicting parameter error between the root and leaf mapping, e.g.
```
mapping(uint phone => mapping(uint call => mapping(uint phone => something) leafMapping) callTimes) friends;
```

test/libsolidity/syntaxTests/parsing/mapping_with_names_conflict_3.sol
zemse: Since both these cases serve the same purpose, it might make sense to just have the same error and code in both cases.

> so you're differentiating between conflicting key -> key and key -> value names

I did that because the error logic was at different places in the code and I tend to write more descriptive error msg if possible. But here I agree the solidity dev that ends up on this error doesn't benefit much.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,57 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+						_mapping.location(),
+						"Conflicting parameter name \"" + keyName + "\" in recursive mapping."
```
nikola-matic: Oh yeah, one thing that kinda irks me is this being called a recursive mapping. Nested would be a more appropriate term here.

zemse: I've just combined the error into single one, removing the "recursive" word itself. So now it looks like: `Conflicting parameter name "nameSame" in mapping.`. Lmk if we have to change this into something better.

test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_4.sol
zemse: There was a typo in the logic and this test case caught that. I've updated the fix in commit [7438248](https://github.com/ethereum/solidity/pull/13384/commits/743824807e95d2ca282996b50ea14f945a79c9f3).

nikola-matic:
I've pushed an empty commit, squash and pushed in the hopes of restarting the build successfully, but unfortunately we seem to have CI problems, which will hopefully be resolved soon. In fact, as soon as that happens, I'll restart the build to see whether everything passes, and then I'll resume reviewing.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,56 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
 	// Convert key type to memory.
 	keyType = TypeProvider::withLocationIfReference(DataLocation::Memory, keyType);
 
 	// Convert value type to storage reference.
 	valueType = TypeProvider::withLocationIfReference(DataLocation::Storage, valueType);
-	_mapping.annotation().type = TypeProvider::mapping(keyType, valueType);
+	_mapping.annotation().type = TypeProvider::mapping(keyType, keyName, valueType, valueName);
+
+	// Check if parameter names are conflicting.
+	if (keyName != "")
```
nikola-matic:
```suggestion
	if (!keyName.empty())
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,56 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+		ASTString currentValueName = valueName;
+		while (true) {
+			// Value type is a mapping.
+			if (childMappingType)
+			{
+				// Compare top mapping's key name with child mapping's key name.
+				ASTString childKeyName = childMappingType->keyName();
+				if (keyName == childKeyName)
+				{
+					m_errorReporter.declarationError(
```
nikola-matic: The `chk_errorcodes` step is failing because you can't throw an error with the same code twice, i.e. this one, and the one on line 310; instead you'll have to use a boolean to indicate whether an error needs to be thrown, and then check at the end of the loop iteration and report error if true.

zemse: I've just pushed a change for this.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,59 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+				childMappingType = dynamic_cast<MappingType const*>(valueType);
+			}
+			else
+			{
+				// Compare top mapping's key name with the value name.
+				if (keyName == currentValueName)
+				{
+					isError = true;
+					break;
+				}
```
nikola-matic:
```suggestion
				if (keyName == currentValueName)
					isError = true;
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,59 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+			// Value type is a mapping.
+			if (childMappingType)
+			{
+				// Compare top mapping's key name with child mapping's key name.
+				ASTString childKeyName = childMappingType->keyName();
+				if (keyName == childKeyName)
+				{
+					isError = true;
+					break;
+				}
```
nikola-matic: This is semantically different than the previous approach (which was correct), as you're now break-ing from the loop as soon as you encounter an error instead of continuing to type check, which is ultimately why you had to remove one line form [this test](https://github.com/ethereum/solidity/pull/13384/files#diff-789fefb5e21fd40f374366380ff975e1bfdd6f2b2ed9af02a3c6982e9a449e07). All you need to do here is set the boolean indicator, and continue as per usual.

```suggestion
				if (keyName == childKeyName)
					isError = true;
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,59 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+			}
+		}
+		if (isError)
+		{
+			m_errorReporter.declarationError(
+				1809_error,
+				_mapping.location(),
+				"Conflicting parameter name \"" + keyName + "\" in mapping."
+			);
+		}
```
nikola-matic: Error checking should be done within the `for` loop after every iteration, otherwise you'll always miss one case.

```suggestion
			}
			if (isError)
			{
				m_errorReporter.declarationError(
					1809_error,
					_mapping.location(),
					"Conflicting parameter name \"" + keyName + "\" in mapping."
				);
			}
		}
```

test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_6.sol
```diff
@@ -0,0 +1,7 @@
+contract test {
+    struct Person {
+        mapping(uint nameSame => mapping(uint name1 => mapping(uint nameSame => uint name2) name3) name4) name5;
```
nikola-matic, line 3: Another three-level nested mapping test, but this time valid, please :)

nikola-matic (changes requested):

libsolidity/analysis/DeclarationTypeChecker.cpp
zemse: ~I will make this change but I still don't see the value since the error would still be reported once. Because in previous logic it was reported every time it was discovered and now only the last error would be reported.~

Okay, this makes sense, after the for-loop update it should be reporting all errors.

test/libsolidity/syntaxTests/parsing/mapping_with_names_struct_member_6.sol
zemse: I've added that, please review.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,63 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
+			{
+				m_errorReporter.declarationError(
+					1809_error,
+					_mapping.location(),
+					"Conflicting parameter name \"" + keyName + "\" in mapping."
+				);
+			}
+
+			// If we are at the end of the mapping recursion, we can break.
+			if (!loop) {
```
nikola-matic:
```suggestion
			if (!loop) 
			{
```

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,63 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
-	_mapping.annotation().type = TypeProvider::mapping(keyType, valueType);
+	_mapping.annotation().type = TypeProvider::mapping(keyType, keyName, valueType, valueName);
+
+	// Check if parameter names are conflicting.
+	if (!keyName.empty())
+	{
+		auto childMappingType = dynamic_cast<MappingType const*>(valueType);
+		ASTString currentValueName = valueName;
+		bool loop = true;
+		while (loop) {
```
nikola-matic: ~~This could be a `do { } while(childMappingType)` loop, which should hopefully make the logic simpler by not requiring the `loop` variable.~~
Tried it, and it of course doesn't work as we want to reach the case where the `childMappingType` is not a mapping, and then perform another iteration.

libsolidity/analysis/DeclarationTypeChecker.cpp
nikola-matic: Wait, do you even need this? If `loop` is false, it will not run the next iteration anyway.

docs/types/mapping-types.rst
nikola-matic: Agreed.

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -267,14 +267,58 @@ void DeclarationTypeChecker::endVisit(Mapping const& _mapping)
...
```
nikola-matic:
```suggestion
		while (loop)
		{
```

nikola-matic (approved):
I've addressed the remaining issues, rebased and squashed, and am satisfied with the state of the PR at the moment, and am thus approving. @cameel or @ekpyron, please take over for the final review.

PR #13386, opened 2022-08-13
Update emscripten Dockerfile to 3.1.18

### LINKED ISSUE #13122: Upgrade emscripten to 3.2.0+
Here's the yearly Emscripten update issue. Last update was in 0.8.10 in #11689.

Some cherry-picked updates from the [changelog](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md):

3.2.0:
> Emscripten now knows what minimum browser versions the WASM_BIGINT feature requires and will automatically set the defaults accordingly. (#17163)

Interesting, could check this out for solc-js. [Explainer](https://v8.dev/features/wasm-bigint).

3.1.11:
> compiler-rt updated to LLVM 14. (#16991)
> libc++ updated to LLVM 14. (#17000)
> libc++abi updated to LLVM 14. (#16993)

3.1.7:
> Added new compiler+linker option -sSHARED_MEMORY=1, which enables targeting a shared WebAssembly.Memory. (#16419)

3.1.5:
> Emscripten no longer uses the allocate() runtime function. For backwards compatabiliy with external JS code we still include this function by default but it will no longer be included in -sSTRICT mode. Usages of this function are generally best replaced with _malloc, stackAlloc or allocateUTF8.

3.1.3:
> New setjmp/longjmp support using Wasm EH instructions is added, which is faster and reduces code size. You need a browser that supports Wasm EH spec to use it. The new SjLj support is enabled by -sSUPPORT_LONGJMP=wasm. This can be used with Wasm exception support (-fwasm-exceptions), but not with Emscripten exception support (-fexceptions or -sDISABLE_EXCEPTION_CATCHING=0). When using Wasm EH with Wasm SjLj, there is one restriction that you cannot directly call setjmp within a catch clause. (Calling another function that calls setjmp is fine.) (#14976 and #16072)

3.1.0:
> Emscripten in starting to use ES6 features in its core libraries (at last!). For most users targeting the default set of browsers this is a code size win. For projects targeting older browsers (e.g. -sMIN_CHROME_VERSION=10), emscripten will now run closure compiler in WHITESPACE_ONLY mode in order to traspile any ES6 down to ES5. When this automatic transpilation is performed we generate a warning which can be disabled (using -Wno-transpile) or by explicitly opting in-to or out-of closure using --closure=1 or --closure=0. (#15763).

3.0.1:
> Deprecate EMMAKEN_CFLAGS is favor of EMCC_CFLAGS.

3.0.0:
> The version of musl libc used by emscripten was upgraded from v1.1.15 to v1.2.2. There could be some minor size regressions (or gains) due to changes in upstream musl code but we don't expect anything major. Since this is a fairly substantial change (at least internally) we are bumping the major version of Emscripten to 3. (#13006)

timweri:
closes #13122

Just following the steps mentioned in https://github.com/ethereum/solidity/pull/12241. This PR will build a docker image.

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.emscripten
+++ b/scripts/docker/buildpack-deps/Dockerfile.emscripten
@@ -32,8 +32,8 @@
 # apparently this currently breaks due to conflicting compatibility headers.
 # Using $(em-config CACHE)/sysroot/usr seems to work, though, and still has cmake find the
 # dependencies automatically.
-FROM emscripten/emsdk:2.0.33 AS base
-LABEL version="11"
+FROM emscripten/emsdk:3.1.18 AS base
+LABEL version="12"
 
 ADD emscripten.jam /usr/src
 RUN set -ex; \


timweri:
Nevermind. Docker image building is broken right now. The Docker build of `aarlt/comment-on-pr@v1.2.0` is failing because the Dockerfile didn't pin a version for `octokit`.
```
  Step 8/10 : RUN gem install octokit
   ---> Running in e5a7ffb004b1
  ERROR:  Error installing octokit:
  	The last version of octokit (>= 0) to support your Ruby & RubyGems was 4.25.1. Try installing it with `gem install octokit -v 4.25.1`
  	octokit requires Ruby version >= 2.7.0. The current ruby version is 2.6.0.0.
```

I guess this is blocked until `comment-on-pr` gets fixed.

timweri:
I don't have the right access to trigger buildpack jobs. Somebody should take over this.

Marenz:
I took over and recreated this PR in our repo https://github.com/ethereum/solidity/pull/13424

PR #13387, opened 2022-08-13
Add detail about limitation in voting contract to fix #12478

### LINKED ISSUE #12478: In "Solidity by Example --> Voting": will not report  winning proposal correctly if voteCount's are tied
## Page
https://docs.soliditylang.org/en/latest/solidity-by-example.html#voting
https://docs.soliditylang.org/en/latest/solidity-by-example.html#possible-improvements

## Abstract
Hi, in the voting example (at line 135 in Remix), picking the winning proposal will not work correctly if two (or more) proposals are tied in the voteCount (see screen shots below). 
Could we at least mention this as a limitation of the current implementation, and, perhaps include the handling of this special case as a "Possible Improvement"

Thank you!

image

image

taylorferran:
Fixes #12478

Just included some more detail on the winningProposal() function not handling a tie correctly

### DIFF
--- a/docs/examples/voting.rst
+++ b/docs/examples/voting.rst
@@ -193,5 +193,8 @@ of votes.
 Possible Improvements
 =====================
 
-Currently, many transactions are needed to assign the rights
-to vote to all participants. Can you think of a better way?
+Currently, many transactions are needed to
+assign the rights to vote to all participants.
+Moreover, if two or more proposals have the same
+number of votes, ``winningProposal()`` is not able
+to register a tie. Can you think of a way to fix these issues?


docs/examples/voting.rst
```diff
@@ -194,4 +194,6 @@ Possible Improvements
 =====================
 
 Currently, many transactions are needed to assign the rights
-to vote to all participants. Can you think of a better way?
+to vote to all participants, and ``winningProposal()`` will
+not work correctly if two or more proposals are tied in voteCount.
```
leonardoalt: not sure I agree with the wording "will not work correctly". It's still returning "a" winning proposal

taylorferran: Good point, will update it with something clearer

taylorferran:
Not so sure why this ci/circleci: t_native_test_ext_gnosis automation failed this time around

docs/examples/voting.rst
```diff
@@ -193,5 +193,9 @@ of votes.
...
 =====================
 
-Currently, many transactions are needed to assign the rights
-to vote to all participants. Can you think of a better way?
+Currently, many transactions are needed to
+assign the rights to vote to all participants,
+and ``winningProposal()`` will not register there
+is a tie if two or more proposals have the same
+number of votes. Can you think of a way to fix
+these issues?
```
matheusaaguiar:
```suggestion
assign the rights to vote to all participants.
Moreover, if two or more proposals have the same
number of votes, ``winningProposal()`` is not able 
to register a tie. Can you think of a way to fix these issues?
```
I think this is more straightforward. What do you think?

matheusaaguiar:
> Not so sure why this ci/circleci: t_native_test_ext_gnosis automation failed this time around

Not related to your PR. Just ignore it.

docs/examples/voting.rst
taylorferran: Yes I agree it's a bit easier to understand, will commit this suggestion

matheusaaguiar:
Looking good. Please, squash the commits and rebase.

leonardoalt:
@taylorferran you seem to have used `merge` instead of `rebase`, which cause a `merge commit` to be part of your commits. Please remove that and simply `git rebase` your branch on top of `origin/develop`.

taylorferran:
Sorry - still getting the hang of git, will do this and fix it based on what you've suggested

leonardoalt (dismissed):

leonardoalt:
I've squashed and rebased the original branch.

leonardoalt (approved):

taylorferran:
Thanks for the help

azflin:
Is there any viable solution to return a dynamic array of all tied proposals?

PR #13391, opened 2022-08-15
Add Natspec devdoc inheritance support for events

### LINKED ISSUE #13199: Event inheritance in natspec (documentation)
## Abstract
Events, in natspec documentation, are not inherited like the methods are.

## Motivation
I'm building a Typescript output for solidity. A little bit like Typechain. When inheriting ERC20 from openzeppelin for example I can generate TsDoc from solc's devdoc output for the methods, but not for the events.

Example of output I would like to achieved using the devdoc output of the ERC20: 
```typescript
  events: {
    /**
     * Emitted when the allowance of a `spender` for an `owner` is set by a call to {approve}. `value` is the new allowance.
     */
    Approval: (owner: string, spender: string, value: BigNumber) => void;
    /**
     * Emitted when when an amount of token is sent from one address to another
     * @param from the address the tokens are taken from
     * @param to the address the tokens are moved to
     * @param value the amount of tokens transfered
     */
    Transfer: (from: string, to: string, value: BigNumber) => void;
  };
```

## Specification
Best would be for the events to work like methods : https://docs.soliditylang.org/en/v0.8.13/natspec-format.html#inheritance-notes 

## Backwards Compatibility
Backwards  Compatible

## Related issues
In #8911, @Marenz mentioned why events are not inherited from the parent class. Though I think this would greatly help for tools that want to generate dev-related documentation from the contract.

timweri:
closes #13199

For some reason, event's Natspec is inherited in userdoc but not devdoc. Not sure if it's on purpose.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@ Language Features:
 Compiler Features:
 * Commandline Interface: Return exit code ``2`` on uncaught exceptions.
 * Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the bytecode.
+* Natspec: Add event Natspec inheritance for devdoc.
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 * Language Server: Add basic document hover support.
--- a/libsolidity/interface/Natspec.cpp
+++ b/libsolidity/interface/Natspec.cpp
@@ -168,7 +168,7 @@ Json::Value Natspec::devDocumentation(ContractDefinition const& _contractDef)
 			));
 	}
 
-	for (auto const& event: _contractDef.events())
+	for (auto const& event: _contractDef.definedInterfaceEvents())
 		if (auto devDoc = devDocumentation(event->annotation().docTags); !devDoc.empty())
 			doc["events"][event->functionType(true)->externalSignature()] = devDoc;
 
--- a/test/libsolidity/SolidityNatspecJSON.cpp
+++ b/test/libsolidity/SolidityNatspecJSON.cpp
@@ -485,6 +485,64 @@ BOOST_AUTO_TEST_CASE(event)
 	checkNatspec(sourceCode, "ERC20", userDoc, true);
 }
 
+BOOST_AUTO_TEST_CASE(event_inheritance)
+{
+	char const* sourceCode = R"(
+		contract ERC20 {
+			/// @notice This event is emitted when a transfer occurs.
+			/// @param from The source account.
+			/// @param to The destination account.
+			/// @param amount The amount.
+			/// @dev A test case!
+			event Transfer(address indexed from, address indexed to, uint amount);
+		}
+
+		contract A is ERC20 {
+		}
+
+		contract B is A {
+		}
+	)";
+
+	char const* devDoc = R"ABCDEF(
+	{
+		"events":
+		{
+			"Transfer(address,address,uint256)":
+			{
+				"details": "A test case!",
+				"params":
+				{
+					"amount": "The amount.",
+					"from": "The source account.",
+					"to": "The destination account."
+				}
+			}
+		},
+		"methods": {}
+	}
+	)ABCDEF";
+	checkNatspec(sourceCode, "ERC20", devDoc, false);
+	checkNatspec(sourceCode, "A", devDoc, false);
+	checkNatspec(sourceCode, "B", devDoc, false);
+
+	char const* userDoc = R"ABCDEF(
+	{
+		"events":
+		{
+			"Transfer(address,address,uint256)":
+			{
+				"notice": "This event is emitted when a transfer occurs."
+			}
+		},
+		"methods": {}
+	}
+	)ABCDEF";
+	checkNatspec(sourceCode, "ERC20", userDoc, true);
+	checkNatspec(sourceCode, "A", userDoc, true);
+	checkNatspec(sourceCode, "B", userDoc, true);
+}
+
 BOOST_AUTO_TEST_CASE(dev_desc_after_nl)
 {
 	char const* sourceCode = R"(


leonardoalt:
Thanks for the PR! The fix might be a bit more complex than that though, and may need to take into account annotations and conflicting parents.
https://github.com/ethereum/solidity/issues/13199#issuecomment-1231414366

timweri:
@leonardoalt I believe such conflict resolution is already implemented when I went through the code. Such inheritance has already been done for `events` in `userdoc`. I think the reason why `userdoc` and `devdoc` differs on `events` is because of this one line which might be an oversight.
https://github.com/ethereum/solidity/blob/31a25cd5b19412842fd17bfa81db84684a601e88/libsolidity/interface/Natspec.cpp#L171

axic (dismissed):
I think this looks good, but needs to be rebased. It is expected that no test would be affected as this adds a new functionality, but could imagine a few more test cases when it should not be inherited (can events be overriden anyhow?).

timweri:
I'll add some tests

timweri:
@axic I tried to override events but it seems like this is not allowed. Also inheriting two events with the same signature, each from a separate base contract, is not allowed.

I don't know if there's any other test cases I should include.

timweri:
@axic I accidentally uploaded a bunch of cache files and now I think the CI jobs are frozen

Changelog.md
```diff
@@ -27,6 +27,7 @@ Compiler Features:
 Bugfixes:
  * Type Checker: Fix internal compiler error on tuple assignments with invalid left-hand side.
  * Yul IR Code Generation: Fix internal compiler error when accessing the ``.slot`` member of a mapping through a storage reference in inline assembly.
+ * Natspec: Fix event's Natspec inheritance for devdoc.
```
nikola-matic: Looking at the original issue, I don't think this is a bug, rather a missing feature.

nikola-matic (commented):
@timweri can you rebase against develop, and force push - this will re-trigger the build. It happens quite often with forked repos that the CI will get stuck. You can also push and empty commit to the same effect.

nikola-matic:
@christianparpart can you take a look at this PR, it's seems like this is your niche :)

Changelog.md
timweri: I changed the words and put it under features. How does it sound now?

nikola-matic: Yup, that looks better, at least to me, although I'm not sure whether NatSpec is a compiler or a language feature. @christianparpart will have to provide that input.

nikola-matic:
@timweri can you please rebase against the latest upstream develop - some of the old jobs are failing.
@christianparpart another look pls, I think this one is close to being done (if not already).

libsolidity/interface/Natspec.cpp
```diff
@@ -168,7 +168,7 @@ Json::Value Natspec::devDocumentation(ContractDefinition const& _contractDef)
 			));
 	}
 
-	for (auto const& event: _contractDef.events())
+	for (auto const& event: _contractDef.definedInterfaceEvents())
```
christianparpart, line 171: Not actually related to this PR directly, but I had to lookup the implementation of this function to understand what it does, as the name wasn't really intuitively telling me that.

Maybe (not in here) in the future we might want to give this function a better name. This function returns a list of all events directly or indirectly (via inheritance) declared in this contract. So maybe having a name like `collectAllVisibleInterfaceEvents()` might improve this.

test/libsolidity/SolidityNatspecJSON.cpp
```diff
@@ -485,6 +485,62 @@ BOOST_AUTO_TEST_CASE(event)
...
+					"amount": "The amount.", "from": "The source account.", "to": "The destination account."
+				}
+			}
+		},
+		"methods": {}
+	}
+	)ABCDEF";
+	checkNatspec(sourceCode, "ERC20", devDoc, false);
+	checkNatspec(sourceCode, "A", devDoc, false);
+	checkNatspec(sourceCode, "B", devDoc, false);
```
christianparpart, line 527: TBH I think this test is sufficient. It's not actually ensuring if it adds up with events from the final contracts (like A or B) - but that would be simply unit-testing the call to the above's `_contractDef.definedInterfaceEvents()`.

christianparpart (dismissed):
I think that's sufficient, let's see what @cameel would add to that? ;-)

nikola-matic:
@timweri we've finally fixed our builds, so can you please rebase against upstream develop again?

timweri:
> @timweri we've finally fixed our builds, so can you please rebase against upstream develop again?

Done!

Changelog.md
```diff
@@ -25,6 +25,7 @@ Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Natspec: Add event Natspec inheritance for devdoc.
```
cameel: 0.8.17 has been released so this must be moved to the 0.8.18 section.

test/libsolidity/SolidityNatspecJSON.cpp
```diff
@@ -485,6 +485,62 @@ BOOST_AUTO_TEST_CASE(event)
...
+	char const* devDoc = R"ABCDEF(
+	{
+		"events":
+		{
+			"Transfer(address,address,uint256)":
+			{
+				"details": "A test case!",
+				"params":
+				{
+					"amount": "The amount.", "from": "The source account.", "to": "The destination account."
```
cameel: I see the other tests are doing this too, but I think this would be more readable and easier to inspect visually:
```suggestion
					"amount": "The amount.",
					"from": "The source account.",
					"to": "The destination account."
```

cameel (commented):
The code looks OK, but before merging, the changelog entry must be moved.

Changelog.md
cameel: > I'm not sure whether NatSpec is a compiler or a language feature.

In the changelog I see that it can be either but it's generally under "Language Features" when we touch the syntax and under "Compiler Features" when it's about the natspec output from the compiler. In this case I'd say it's the latter and there's actually one other inheritance related case that was classified like that.

> I don't think this is a bug, rather a missing feature.

You could also argue that it was a bug in #9338 because I don't see any good reason why it did not cover both. But I'm fine with classifying it as a missing feature too. It's kinda both.

BTW, that older PR did not even add a changelog entry.

cameel (approved):
I rebased this and fixed the small issues I mentioned. Now let's merge it.

timweri:
Thank you!

PR #13393, opened 2022-08-15
[WIP] Solidity UI - Would such tool be useful?

aarlt:
I think it would be nice to have somehow a tool that could visualize compiler internals. If once ready, the tool could be used to teach about how the compiler actually works and it will also be useful for debugging compiler bugs.

The idea would be to have at first a simple Solidity/Yul editor with syntax highlighting, after that all current provided debugging information and compiler outputs should be somehow visualized. Additionally it would be great to also visualize things, that we are currently not really exporting, e.g. optimiser code transformations, whiskers template instantiations (being able to see from where exactly (in our C++ code) a parameter of a template was set) etc.

I know that all this is quite a big effort, but I would say it would be very helpful in general. It will improve the code quality of the solidity compiler - because we need to see the compiler from a slightly different perspective. This tool will help us to communicate used algorithms and concepts, where it would also support us to investigate compiler bugs.

**I created this PR to collect possible feedback on such tool in general. What is your opinion?** 

The current code is just a hack - I just pushed the current code. It is not doing anything special yet. It will also right now only compile on macOS. The current code is just integrating [ImGui](https://github.com/ocornut/imgui) and some other components into our build system. I also added basic "project manager" code, and a first integration of the [Zep Editor](https://github.com/Rezonality/zep).

I had also a short talk with @ekpyron in our chat. He was critical on this, mainly because he don't see that such tool could be useful. He also sees the maintenance effort (in the future) critical.


ekpyron:
Sorry about that, but I don't think we will go for this and I don't think it makes sense to keep a PR open with this (we're trying to keep our PR count low enough to be manageable).

ekpyron:
Put differently: if *anyone* wants to build a UI on top of the compiler, it should interact with the compiler via the standard interfaces (like standard-json) and be developed separately from the compiler and not by us, categorically.
Any further "currently not exported" information should, if anything, become additional debug data to be consumed by arbitrary *external* tools.

aarlt:
Ah ok! No problem. But I will create an issue for that - just in the case anyone wants to add a comment on this.

ekpyron:
This is out of the scope of the compiler project, no matter what - an issue won't change that.

aarlt:
I understand your point. I would say that a compiler project could also create tools that will help to understand the compiler better.

PR #13399, opened 2022-08-15
Remove unnecessary "Callcode" in the heading

minaminao:
The heading contains "Callcode", but there is no description of callcode in this section, so it would be easier to understand if it were removed.

An alternative is to add a description of callcode, but since callcode is rarely used today, I think it is better to remove it.

### DIFF
--- a/docs/introduction-to-smart-contracts.rst
+++ b/docs/introduction-to-smart-contracts.rst
@@ -507,7 +507,7 @@ depth limit of a little less than 1000 in practice.
 .. index:: delegatecall, library
 
 Delegatecall and Libraries
-=====================================
+==========================
 
 There exists a special variant of a message call, named **delegatecall**
 which is identical to a message call apart from the fact that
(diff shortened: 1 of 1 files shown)


docs/introduction-to-smart-contracts.rst
```diff
@@ -504,10 +504,10 @@ operations, loops should be preferred over recursive calls. Furthermore,
 only 63/64th of the gas can be forwarded in a message call, which causes a
 depth limit of a little less than 1000 in practice.
 
-.. index:: delegatecall, callcode, library
```
leonardoalt, line 507: Did you check whether some other docs are using this as ref/anchor? If there are, it might be that those sites require changes too.

minaminao: I have done `make clean && make html` and confirmed that warnings do not appear :)
Are there any patterns that this does not detect?

leonardoalt: Sounds good. Weirdly enough, none of these seem to be used as references.

leonardoalt (approved):
Thanks!

PR #13396, opened 2022-08-15
Test "comment PR" job.

aarlt:

PR #13394, opened 2022-08-15
[buildpack] Switch to unsplash/comment-on-pr@v1.3.1.

aarlt:
- fixes #13392

PR #13400, opened 2022-08-15
[buildpack] Switch to unsplash/comment-on-pr@v1.3.0.

### LINKED ISSUE #13392: buildpack-deps jobs are failing due to `arrlt/comment-on-pr`
## Description

While trying to update emscripten (#13122), I ran into buildpack-deps jobs failing.
buildpack-deps rely on the Dockerfile `aarlt/comment-on-pr@v1.2.0`.

The problem is `aarlt/comment-on-pr@v1.2.0` doesn't pin the version of `octokit` it uses.
Since [v5.0.0](https://github.com/octokit/octokit.rb/releases/tag/v5.0.0), `octokit` requires Ruby of version `>=2.7.0`.
However, `aarlt/comment-on-pr@v1.2.0` uses `2.6.0`.
```
  Step 8/10 : RUN gem install octokit
   ---> Running in e5a7ffb004b1
  ERROR:  Error installing octokit:
  	The last version of octokit (>= 0) to support your Ruby & RubyGems was 4.25.1. Try installing it with `gem install octokit -v 4.25.1`
  	octokit requires Ruby version >= 2.7.0. The current ruby version is 2.6.0.0.
```

## Steps to Reproduce

Any run of buildpack-deps should trigger this. For example, https://github.com/ethereum/solidity/pull/13386

aarlt:
`unsplash/comment-on-pr@v1.3.1` seem to have some problems, `v1.3.0` seem to work (see #13396)
- fixes  #13392

### DIFF
--- a/.github/workflows/buildpack-deps.yml
+++ b/.github/workflows/buildpack-deps.yml
@@ -37,6 +37,9 @@ jobs:
 
       - name: comment PR
         if: "env.DOCKER_IMAGE"
-        uses: unsplash/comment-on-pr@b5610c6125a7197eaec80072ea35ef53e1fc6035 #v1.3.1
+        # NOTE: Can't update to v1.3.1 due to an error: `/entrypoint.sh:5:in 'require_relative': cannot load such file -- /lib/github (LoadError)`
+        uses: unsplash/comment-on-pr@ffe8f97ccc63ce12c3c23c6885b169db67958d3b #v1.3.0
         with:
           msg: "`${{ env.DOCKER_IMAGE }} ${{ env.DOCKER_REPO_DIGEST }}`."
+          check_for_duplicate_msg: false
+


.github/workflows/buildpack-deps.yml
```diff
@@ -37,6 +37,8 @@ jobs:
 
       - name: comment PR
         if: "env.DOCKER_IMAGE"
-        uses: unsplash/comment-on-pr@b5610c6125a7197eaec80072ea35ef53e1fc6035 #v1.3.1
+        uses: unsplash/comment-on-pr@ffe8f97ccc63ce12c3c23c6885b169db67958d3b #v1.3.0
         with:
-          msg: "`${{ env.DOCKER_IMAGE }} ${{ env.DOCKER_REPO_DIGEST }}`."
\ No newline at end of file
+          msg: "`${{ env.DOCKER_IMAGE }} ${{ env.DOCKER_REPO_DIGEST }}`."
+          check_for_duplicate_msg: false
```
cameel, line 44: Why don't we want this check? Won't the new hash be different anyway?

.github/workflows/buildpack-deps.yml
```diff
@@ -37,6 +37,8 @@ jobs:
 
       - name: comment PR
         if: "env.DOCKER_IMAGE"
-        uses: unsplash/comment-on-pr@b5610c6125a7197eaec80072ea35ef53e1fc6035 #v1.3.1
+        uses: unsplash/comment-on-pr@ffe8f97ccc63ce12c3c23c6885b169db67958d3b #v1.3.0
```
cameel, line 41: Would be best to have a TODO comment to update this and some explanation on why we can't do it right now.

cameel (commented):
Looks ok but I have some questions.

.github/workflows/buildpack-deps.yml
aarlt: I saw that yesterday the bot was only creating one comment (see https://github.com/ethereum/solidity/pull/13396), where it should have been created for. So I suspected that there is probably something wrong with this `check_for_duplicate_msg` - after changing this all comments showed up. But I did not do any further investigations on this.

.github/workflows/buildpack-deps.yml
aarlt: In `v1.3.1` they do a `require_relative` that seem not to work `/entrypoint.sh:5:in `require_relative': cannot load such file -- /lib/github (LoadError)` - so I just went one version down. Also here I did not do any further investigations, because the problem seem to be solved. 

Ah, the reason for this was that I had forked `unsplash/comment-on-pr` in the past, but that version is right now not working anymore, that's why we just changed to the original `unsplash` repository here https://github.com/ethereum/solidity/pull/13394. But this version was not working - so I downgraded that version.

aarlt: We merged https://github.com/ethereum/solidity/pull/13394 too fast without testing - testing was then later done here https://github.com/ethereum/solidity/pull/13396 - next time I will test the stuff already in the same PR.

.github/workflows/buildpack-deps.yml
cameel: OK. I guess the worst that can happen is that we'll get some duplicate messages and we'll look into it if it happens.

.github/workflows/buildpack-deps.yml
cameel: ok, I'd still add at least a short comment about that:
```suggestion
        # NOTE: Can't update to v1.3.1 due to an error: `/entrypoint.sh:5:in 'require_relative': cannot load such file -- /lib/github (LoadError)`
        uses: unsplash/comment-on-pr@ffe8f97ccc63ce12c3c23c6885b169db67958d3b # v1.3.0
```

cameel (dismissed):
Approving because it's good overall but we should really have the comment about 1.3.1.

.github/workflows/buildpack-deps.yml
aarlt: Ok! I added the comment.

cameel (approved):

PR #13401, opened 2022-08-16
Enable highlighting for more code blocks in the docs

cameel:
Fixes a few code blocks that were missing highlighting. There might be more, these are just those I noticed when preparing feedback on the new highlighter.


nishant-sachdeva (approved):
LGTM

PR #13402, opened 2022-08-16
Disable failing Gnosis tests

nikola-matic:


test/externalTests/gnosis.sh
```diff
@@ -87,6 +87,12 @@ function gnosis_safe_test
     sed -i "s|it\(('can only be called from Safe itself'\)|it.skip\1|g" test/libraries/Migration.spec.ts
     sed -i "s|it\(('should enforce delegatecall to MultiSend'\)|it.skip\1|g" test/libraries/MultiSend.spec.ts
 
+    sed -i "s|it\(('is not called when setting initially'\)|it.skip\1|g" test/core/GnosisSafe.GuardManager.spec.ts
```
cameel, line 90: This needs a comment explaining why we're disabling these tests and when the hack can be removed.

cameel (commented):
We could merge this if we don't manage to figure out anything better but I think that pinning Hardhat to a specific version should have prevented this in the first place. Why is it failing despite Hardhat being pinned at 2.9.4?

nikola-matic:
Closing this in favour of pegging the `hardhat-ethers` version which should prove more robust: https://github.com/ethereum/solidity/pull/13404

PR #13403, opened 2022-08-16
Updated a missing type

yalexis-eth:
##Proposed changes
Update docs


leonardoalt (commented):
Thanks for the PR! I think the previous wording was correct though.

nikola-matic:
Concur with @leonardoalt

cameel:
Yeah, I also think this was correct and we don't want to change this so I'm going to close the PR.

PR #13405, opened 2022-08-16
[libevmasm] Add Assembly::OptimiserSettings::translateSettings.

aarlt:
We had in the past some discussion about having such function.

- https://github.com/ethereum/solidity/pull/10286#issuecomment-827969786
- https://github.com/ethereum/solidity/pull/11550#discussion_r653680572
- https://github.com/ethereum/solidity/pull/12834#discussion_r869153232

I think, at least with the current source, such function is not introducing any additional dependencies. But maybe it is possible to clean up the dependencies. However, I created this PR so that we can discuss this.

@axic @cameel @chriseth

### DIFF
--- a/libevmasm/Assembly.cpp
+++ b/libevmasm/Assembly.cpp
@@ -792,3 +792,18 @@ Assembly const* Assembly::subAssemblyById(size_t _subId) const
 	assertThrow(currentAssembly != this, AssemblyException, "");
 	return currentAssembly;
 }
+
+Assembly::OptimiserSettings Assembly::OptimiserSettings::translateSettings(frontend::OptimiserSettings const& _settings, langutil::EVMVersion const& _evmVersion)
+{
+	// Constructing it this way so that we notice changes in the fields.
+	evmasm::Assembly::OptimiserSettings asmSettings{false,  false, false, false, false, false, _evmVersion, 0};
+	asmSettings.runInliner = _settings.runInliner;
+	asmSettings.runJumpdestRemover = _settings.runJumpdestRemover;
+	asmSettings.runPeephole = _settings.runPeephole;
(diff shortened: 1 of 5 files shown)


matheusaaguiar:
I read through the mentioned PRs and looked at the changes. The removal of the duplicated code is really nice and as you said, it doesn't introduce any additional dependencies.
Thinking about how to remove the already present dependencies, regarding `OptimiserSettings`, maybe it could be moved to a common place that is shared by `libyul`, `libsolidity` and `libewasm`.

cameel (commented):
For me it looks good but I would like to hear @axic's opinion because he objected last time we tried to do this.

axic:
My main concern was the cross-dependency we are introducing:
```mermaid
graph TD;
    evmasm-->frontend;
    yul-->evmasm;
    frontend-->evmasm;
    frontend-->yul;
```

The code deduplication is welcome. Perhaps one way to make this nicer is moving `interface/OptimiserSettings.h` to `libsolutil` or `liblangutil`?

I won't contest it however.

cameel:
> The code deduplication is welcome. Perhaps one way to make this nicer is moving `interface/OptimiserSettings.h` to `libsolutil` or `liblangutil`?

I guess we could do that, though it does seem a bit out of place there (it is not a general utility, it's really closely tied to compiler features, just in to separate libs).

IMO, we might be better off just merging this and untangling the dependency cycle at the point when we'll actually need to make `libevmasm` independent. As @aarlt said, the cyclical dependency already exists, even without this PR.

cameel:
I'm going to merge this and let's deal with severing the dependencies once it becomes an actual issue.

cameel (approved):

PR #13406, opened 2022-08-17
Fix `.slot` accessing via mapping reference in assembly

### LINKED ISSUE #13259: ICE in the codegen when using `.slot` on a mapping reference in assembly: `InternalCompilerError: Invalid stack item name: slot`
## Description

Sometimes compiling via IR results in the following error (with Foundry):
```
❯ forge build
[⠆] Compiling...
[⠆] Compiling 69 files with 0.8.15
[⠔] Solc 0.8.15 finished in 3.00s
Error: 
Compiler run failed
InternalCompilerError: Invalid stack item name: slot
```

As the error implies, it seems to be related to storing variables that point to storage slots.

## Environment

- Compiler version: reproduced with 0.8.15,0.8.14,0.8.13
- Framework/IDE (e.g. Truffle or Remix): Foundry
- EVM execution environment / backend / blockchain client: Foundry
- Operating system: macOS

## Steps to Reproduce

In my case, I was able to resolve this by changing the following code from this:

```solidity
        mapping(uint256 => address) storage tokenApprovalsPtr = _tokenApprovals;
        // The following is equivalent to `approvedAddress = _tokenApprovals[tokenId]`.
        assembly {
            // Compute the slot.
            mstore(0x00, tokenId)
            mstore(0x20, tokenApprovalsPtr.slot)
            approvedAddressSlot := keccak256(0x00, 0x40)
            // Load the slot's value from storage.
            approvedAddress := sload(approvedAddressSlot)
        }
```

to this:

```solidity
        // The following is equivalent to `approvedAddress = _tokenApprovals[tokenId]`.
        assembly {
            // Compute the slot.
            mstore(0x00, tokenId)
            mstore(0x20, _tokenApprovals.slot)
            approvedAddressSlot := keccak256(0x00, 0x40)
            // Load the slot's value from storage.
            approvedAddress := sload(approvedAddressSlot)
        }
```

Link to commit with issue present: https://github.com/jameswenzel/bound-layerable/blob/91b7096cfedd20bb422f5ed1f83c7e50261c2bcc/src/token/ERC721A.sol#L694
Link to commit with issue fixed: https://github.com/jameswenzel/bound-layerable/blob/770ded9df31d624f38e1db6c32d0d5c6f7e64132/src/token/ERC721A.sol#L694

While the code I changed is [copied from the ERC721A repo](https://github.com/chiru-labs/ERC721A/blob/7819fc341da2fece936688239c66b702f7c8add8/contracts/ERC721A.sol#L585), I'm able to build the code at that commit as-is without issues, with the IR pipeline enabled.

wechman:
fixes #13259

### DIFF
--- a/libsolidity/ast/Types.h
+++ b/libsolidity/ast/Types.h
@@ -1528,6 +1528,8 @@ public:
 	bool hasSimpleZeroValueInMemory() const override { solAssert(false, ""); }
 	bool nameable() const override { return true; }
 
+	std::vector<std::tuple<std::string, Type const*>> makeStackItems() const override;
+
 	Type const* keyType() const { return m_keyType; }
 	Type const* valueType() const { return m_valueType; }
 
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -3789,6 +3789,11 @@ TypeResult MappingType::interfaceType(bool _inLibrary) const
 	return this;
 }
 
+std::vector<std::tuple<std::string, Type const*>> MappingType::makeStackItems() const
+{
+	return {std::make_tuple("slot", TypeProvider::uint256())};
+}
+
 string TypeType::richIdentifier() const
 {
 	return "t_type" + identifierList(actualType());
--- a/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
+++ b/libsolidity/codegen/ir/IRGeneratorForStatements.cpp
@@ -173,10 +173,9 @@ private:
 		{
 			solAssert(suffix == "slot" || suffix == "offset");
 			solAssert(varDecl->isLocalVariable());
+			solAssert(!varDecl->type()->isValueType());
 			if (suffix == "slot")
 				value = IRVariable{*varDecl}.part("slot").name();
-			else if (varDecl->type()->isValueType())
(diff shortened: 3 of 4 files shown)


libsolidity/ast/Types.h
```diff
@@ -1528,6 +1528,8 @@ class MappingType: public CompositeType
 	bool hasSimpleZeroValueInMemory() const override { solAssert(false, ""); }
 	bool nameable() const override { return true; }
 
+	std::vector<std::tuple<std::string, Type const*>> makeStackItems() const override;
```
nikola-matic, line 1531: Where is `makeStackItems()` used?

Marenz: It's used in Types.h is one single usage :D

matheusaaguiar (approved):
I have looked at the original bug and from what I've understood, since `MappingType` did not had a `makeStackItems` overload, it defaulted to the `Type` base implementation that only returns a "blank" tuple. This in turn made method `part` or `IRVariable` never able to find a item named `slot` among the stack items of a `MappingType` var, which triggers a failing `solAssert`, i.e, the ICE from the original issue. 
So after all that explanation, that's why I consider this approved :)

Marenz (approved):

PR #13409, opened 2022-08-17
Reduce duplication in AST helpers for finding try/catch clauses

tcoyvwac:
* Using helper function: `checkForClause` in anonymous namespace, condensed some duplicate code.
* Preferred range-v3 function `tail`, to range-for-loop through data-array starting from index=`1`.

### DIFF
--- a/libsolidity/ast/AST.cpp
+++ b/libsolidity/ast/AST.cpp
@@ -29,6 +29,8 @@
 #include <libsolidity/ast/TypeProvider.h>
 #include <libsolutil/Keccak256.h>
 
+#include <range/v3/view/tail.hpp>
+
 #include <boost/algorithm/string.hpp>
 
 #include <functional>
@@ -37,6 +39,17 @@
 using namespace std;
 using namespace solidity;
 using namespace solidity::frontend;
+
+namespace
+{
+TryCatchClause const* findClause(vector<ASTPointer<TryCatchClause>> const& _clauses, optional<string> _errorName = {})
+{
+	for (auto const& clause: ranges::views::tail(_clauses))
+		if (_errorName.has_value() ? clause->errorName() == _errorName : clause->errorName().empty())
+			return clause.get();
+	return nullptr;
+}
+}
 
 ASTNode::ASTNode(int64_t _id, SourceLocation _location):
 	m_id(static_cast<size_t>(_id)),
@@ -981,26 +994,14 @@ TryCatchClause const* TryStatement::successClause() const
 	return m_clauses[0].get();
 }
 
-TryCatchClause const* TryStatement::panicClause() const
-{
-	for (size_t i = 1; i < m_clauses.size(); ++i)
-		if (m_clauses[i]->errorName() == "Panic")
-			return m_clauses[i].get();
-	return nullptr;
+TryCatchClause const* TryStatement::panicClause() const {
+	return findClause(m_clauses, "Panic");
 }
 
-TryCatchClause const* TryStatement::errorClause() const
-{
-	for (size_t i = 1; i < m_clauses.size(); ++i)
-		if (m_clauses[i]->errorName() == "Error")
-			return m_clauses[i].get();
-	return nullptr;
+TryCatchClause const* TryStatement::errorClause() const {
+	return findClause(m_clauses, "Error");
 }
 
-TryCatchClause const* TryStatement::fallbackClause() const
-{
-	for (size_t i = 1; i < m_clauses.size(); ++i)
-		if (m_clauses[i]->errorName().empty())
-			return m_clauses[i].get();
-	return nullptr;
+TryCatchClause const* TryStatement::fallbackClause() const {
+	return findClause(m_clauses);
 }


libsolidity/ast/AST.cpp
```diff
@@ -38,6 +40,20 @@ using namespace std;
 using namespace solidity;
 using namespace solidity::frontend;
 
+namespace
+{
+TryCatchClause const*
+checkForClause(vector<ASTPointer<TryCatchClause>> const& clausesList, optional<string> stringPred = {})
```
Marenz: We use a leading `_` for function parameters
```suggestion
checkForClause(vector<ASTPointer<TryCatchClause>> const& _clausesList, optional<string> _stringPred = {})
```

libsolidity/ast/AST.cpp
```diff
@@ -38,6 +40,20 @@ using namespace std;
...
+{
+	for (auto const& clause: ranges::views::tail(clausesList))
+	{
+		if ((stringPred) ? clause->errorName() == stringPred : std::empty(clause->errorName()))
+			return clause.get();
+	}
```
Marenz:
```suggestion
	for (auto const& clause: ranges::views::tail(clausesList))
		if ((stringPred) ? clause->errorName() == stringPred : std::empty(clause->errorName()))
			return clause.get();
```

Marenz (changes requested):
A nice PR, some style changes mainly.

I am not sure though whether it might be good to use std::move() for the string parameter. Someone else might want to pitch in on that?

Marenz:
Please rebase when you update this PR then the spelling CI test will succeed as well

libsolidity/ast/AST.cpp
nikola-matic: Do we need an optional here, we can default initialize a string just as well?

cameel:
```suggestion
TryCatchClause const* checkForClause(vector<ASTPointer<TryCatchClause>> const& clausesList, optional<string> stringPred = {})
```

cameel:
```suggestion
findClause(vector<ASTPointer<TryCatchClause>> const& _clauses, optional<string> _errorName = {})
```

cameel: We could use `""` for that because it's not a valid error name but personally I don't mind `optional` here. It makes the intent clear.

libsolidity/ast/AST.cpp
```diff
@@ -38,6 +40,20 @@ using namespace std;
...
 using namespace solidity::frontend;
 
+namespace
+{
+TryCatchClause const*
+checkForClause(vector<ASTPointer<TryCatchClause>> const& clausesList, optional<string> stringPred = {})
+{
+	for (auto const& clause: ranges::views::tail(clausesList))
+	{
+		if ((stringPred) ? clause->errorName() == stringPred : std::empty(clause->errorName()))
```
cameel:
```suggestion
		if (stringPred.has_value() ? clause->errorName() == stringPred : clause->errorName().empty())
```

tcoyvwac: This is an interesting suggestion and I liked discovering this suggested change. 

_(I personally wonder,)_ the challenge of understanding a codebase is "how much is _standardised_?". As your mentioned[ another comment, clear intent is a positive](https://github.com/ethereum/solidity/pull/13409#discussion_r949231577), and `std::empty` implies the datatype's `.empty()` follows `std::` standards.

However, for this project `solidity`, the `std::` namespace is name-masked, and having the code's text as `empty(clause->errorName())` is very awkward in understanding its intent... 

So, `clause->errorName().empty()`, is the only option one can take.

In no way saying anything negative about the suggestion, just more noting that I agree with the suggestion and that the suggestion is "interesting". :relaxed:

cameel: I mean, `empty(clause->errorName())` would be fine. It's just that `clause->errorName().empty()` is how we usually do it and I find it cleaner but it's subjective. I wouldn't really reject the PR if you insisted on the former but just wanted to show how I'd do it.

libsolidity/ast/AST.cpp
```diff
@@ -981,26 +994,8 @@ TryCatchClause const* TryStatement::successClause() const
...
 }
 
-TryCatchClause const* TryStatement::panicClause() const
-{
-	for (size_t i = 1; i < m_clauses.size(); ++i)
-		if (m_clauses[i]->errorName() == "Panic")
-			return m_clauses[i].get();
-	return nullptr;
-}
+TryCatchClause const* TryStatement::panicClause() const { return findClause(m_clauses, "Panic"); }
```
cameel: Since this is an implementation file, I'd wrap these properly:
```suggestion
TryCatchClause const* TryStatement::panicClause() const {
	return findClause(m_clauses, "Panic");
}
```

cameel (dismissed):
Just one more optional suggestion but it's already good so I'm approving.

leonardoalt (approved):

leonardoalt:
@tcoyvwac sorry for the delay here, could you rebase one more time to see if the CI problem goes away?

nikola-matic:
> @tcoyvwac sorry for the delay here, could you rebase one more time to see if the CI problem goes away?

A rebase would be nice and will make the failing `gnosis` step go away, but the issue here is @Marenz 's change request.

Marenz (approved):

tcoyvwac:
Thank you all for the feedback and merging btw! Very grateful! :cat2: :tada:

PR #13414, opened 2022-08-18
change to common config of 1 job to reuse between osx and ubuntu

chatchai-hub:
.circleci/config.yml


cameel:
This looks like a spam PR from a private account and with commits copied from #12923. Closing.

PR #13415, opened 2022-08-18
User defined literal suffix parallel work

matheusaaguiar:
Draft PR used to work in parallel with #12656 .  

- [x] Disallow suffixes with zero or more than one return value.
- [x] Disallow suffixes with storage and calldata return values.
- [x] Disallow signed exponent for decimal suffixes.
- [x] Fix: Suffixed values are not disallowed as array sizes even though they're not constants.
- [ ] Fix: Suffixes not detected as function calls by control flow graph (no "unreachable code" warning).
- [ ] More detailed error message for exponent/mantissa out of range.
- [ ] Include suffix definition as a secondary location in errors.
- [ ] Grammar update.
- [ ] Fix failing command line tests.
- [ ] When no suffix definition found, check if there are any non-suffix definitions and list them in error message.

### DIFF
--- a/libsolidity/analysis/TypeChecker.cpp
+++ b/libsolidity/analysis/TypeChecker.cpp
@@ -49,6 +49,8 @@
 #include <range/v3/view/zip.hpp>
 #include <range/v3/algorithm/any_of.hpp>
 
+#include <fmt/format.h>
+
 #include <memory>
 #include <vector>
 
@@ -3961,14 +3963,29 @@ bool TypeChecker::visit(Literal const& _literal)
 	if (_literal.isSuffixed() && !_literal.hasSubDenomination() && literalRationalType)
 	{
 		auto&& [mantissa, exponent] = literalRationalType->mantissaExponent();
-		solAssert((mantissa && exponent) || (!mantissa && !exponent));
-		if (!mantissa)
+
+		if (!mantissa || !exponent)
+		{
+			string mantissaOrExponentErrorMessage;
+
+			if (!mantissa && !exponent)
+				mantissaOrExponentErrorMessage = "Mantissa and exponent are";
+			else if (!exponent)
+				mantissaOrExponentErrorMessage = "Exponent is";
+			else
+				mantissaOrExponentErrorMessage = "Mantissa is";
+
 			m_errorReporter.typeError(
 				5503_error,
 				_literal.location(),
-				"This fractional number cannot be decomposed into a mantissa and decimal exponent "
-				"that fit the range of parameters of any possible suffix function."
+				fmt::format(
+					"This fractional number cannot be decomposed into a mantissa and decimal exponent "
+					"that fit the range of parameters of any possible suffix function. "
+					"{} out of range of the largest supported integer type.",
+					mantissaOrExponentErrorMessage
+				)
 			);
+		}
 	}
 
 	// NOTE: For suffixed literals this is not the final type yet. We will update it in endVisit()
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -1277,16 +1277,20 @@ pair<RationalNumberType const*, RationalNumberType const*> RationalNumberType::m
 	rational unsignedMantissa = abs(m_value);
 
 	if (unsignedMantissa > maxMantissa)
-		return {nullptr, nullptr};
+		return {nullptr, TypeProvider::rationalNumber(-exponent)};
 
 	while (unsignedMantissa.denominator() != 1)
 	{
 		unsignedMantissa *= 10;
 		--exponent;
 
-		// FIXME: What happens when exponent in scientific notation is max uint?
-		if (unsignedMantissa > maxMantissa || exponent > maxUint)
+		if (exponent > maxUint && unsignedMantissa > maxMantissa)
 			return {nullptr, nullptr};
+		// FIXME: What happens when exponent in scientific notation is max uint?
+		if (exponent > maxUint)
+			return {TypeProvider::rationalNumber(unsignedMantissa), nullptr};
+		if (unsignedMantissa > maxMantissa)
+			return {nullptr, TypeProvider::rationalNumber(-exponent)};
 	}
 
 	return {
--- a/libsolidity/ast/Types.h
+++ b/libsolidity/ast/Types.h
@@ -607,7 +607,8 @@ public:
 	/// base-10 exponent, such that the number is equal to `mantissa * 10**-exponent`.
 	/// @returns Pair of non-null pointers representing the types of the literals corresponding to
 	/// mantissa and exponent if the resulting mantissa and exponent both fit in 256 bits.
-	/// A pair of null pointers otherwise.
+	/// If either of these values does not fit in 256 bits, its value in the pair is null.
+	/// Otherwise, Pair of null pointers.
 	std::pair<RationalNumberType const*, RationalNumberType const*> mantissaExponent() const;
 
 	/// @returns true if the value is negative.
--- a/test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
+++ b/test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
@@ -19,10 +19,10 @@ contract C {
     }
 }
 // ----
-// TypeError 5503: (338-347): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (357-367): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (377-388): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (399-480): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (502-584): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (605-688): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (338-347): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (357-367): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (377-388): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (399-480): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (502-584): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (605-688): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (748-757): The type of the literal cannot be converted to the parameters of the suffix function.
--- a/test/libsolidity/syntaxTests/literalSuffixes/parameters/mantissa_out_of_range.sol
+++ b/test/libsolidity/syntaxTests/literalSuffixes/parameters/mantissa_out_of_range.sol
@@ -37,16 +37,16 @@ contract C {
 // TypeError 8838: (591-600): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (610-619): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (629-711): The type of the literal cannot be converted to the parameters of the suffix function.
-// TypeError 5503: (731-814): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (731-814): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (835-842): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (852-859): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (869-879): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (889-899): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (909-992): The type of the literal cannot be converted to the parameters of the suffix function.
-// TypeError 5503: (1022-1106): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (1022-1106): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (1137-1147): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (1157-1167): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (1177-1190): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (1200-1213): The type of the literal cannot be converted to the parameters of the suffix function.
-// TypeError 5503: (1223-1309): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
-// TypeError 5503: (1339-1426): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (1223-1309): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
+// TypeError 5503: (1339-1426): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
--- a/libsolidity/codegen/ExpressionCompiler.cpp
+++ b/libsolidity/codegen/ExpressionCompiler.cpp
@@ -2664,6 +2664,7 @@ void ExpressionCompiler::appendInternalFunctionCall(
 				assert(rationalNumberType);
 
 				auto&& [mantissa, exponent] = rationalNumberType->mantissaExponent();
+				solAssert(mantissa && exponent);
 				m_context << mantissa->literalValue(nullptr);
 				utils().convertType(*mantissa, *parameterTypes.at(0));
 				m_context << exponent->literalValue(nullptr);
--- a/test/libsolidity/syntaxTests/literalSuffixes/parameters/integer_out_of_range.sol
+++ b/test/libsolidity/syntaxTests/literalSuffixes/parameters/integer_out_of_range.sol
@@ -21,5 +21,5 @@ contract C {
 // TypeError 8838: (396-405): The type of the literal cannot be converted to the parameter of the suffix function.
 // TypeError 8838: (415-424): The type of the literal cannot be converted to the parameter of the suffix function.
 // TypeError 8838: (434-516): The type of the literal cannot be converted to the parameter of the suffix function.
-// TypeError 5503: (536-619): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (536-619): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (536-619): The type of the literal cannot be converted to the parameter of the suffix function.
--- a/test/libsolidity/syntaxTests/literalSuffixes/suffixableLiterals/invalid_suffixed_decimal_mantissa_too_large.sol
+++ b/test/libsolidity/syntaxTests/literalSuffixes/suffixableLiterals/invalid_suffixed_decimal_mantissa_too_large.sol
@@ -9,4 +9,4 @@ contract C {
     Decimal x = 0.115792089237316195423570985008687907853269984665640564039457584007913129639936 suffix; // 2**256 * 10**-78
 }
 // ----
-// TypeError 5503: (211-298): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (211-298): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
--- a/test/libsolidity/syntaxTests/literalSuffixes/suffixableLiterals/invalid_suffixed_integer_too_large.sol
+++ b/test/libsolidity/syntaxTests/literalSuffixes/suffixableLiterals/invalid_suffixed_integer_too_large.sol
@@ -5,7 +5,7 @@ contract C {
     uint y = 999999999999999999999999999999999999999999999999999999999999999999999999999999 suffix;
 }
 // ----
-// TypeError 5503: (92-106): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (92-106): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (92-106): The type of the literal cannot be converted to the parameter of the suffix function.
-// TypeError 5503: (121-206): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function.
+// TypeError 5503: (121-206): This fractional number cannot be decomposed into a mantissa and decimal exponent that fit the range of parameters of any possible suffix function. Mantissa is out of range of the largest supported integer type.
 // TypeError 8838: (121-206): The type of the literal cannot be converted to the parameter of the suffix function.


libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3827,7 +3827,7 @@ void TypeChecker::endVisit(Literal const& _literal)
 
 			optional<string> parameterTypeMessage;
 			if (parameterCountMessage.has_value())
-				m_errorReporter.typeError(4778_error, _literal.location(), parameterCountMessage.value());
+				m_errorReporter.typeError(9128_error, _literal.location(), parameterCountMessage.value());
```
cameel: Why a different number?

matheusaaguiar: This was the result of running `./scripts/error_codes.py --fix`, it identified that ID 4778 appears 2 times.
I ran the script to get a new error ID for the changes I introduced.

matheusaaguiar: Error 4778 was already present in [libyul/AsmAnalysis.cpp:L703](https://github.com/ethereum/solidity/blob/247df5509931d4ad747d2d926432a0955bb2e6f5/libyul/AsmAnalysis.cpp#L703).

cameel: Oh, interesting. This number was actually in the initial implementation I took over from Chris but it does look like a genuine duplicate.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3852,8 +3852,13 @@ void TypeChecker::endVisit(Literal const& _literal)
 			if (suffixFunctionType->returnParameterTypes().size() == 1)
 				_literal.annotation().type = suffixFunctionType->returnParameterTypes().front();
 			else
-				_literal.annotation().type = TypeProvider::tuple(suffixFunctionType->returnParameterTypes());
-
+			{
+				m_errorReporter.typeError(
+					7848_error,
+					_literal.location(),
+					"Function must return exactly one value to be used as literal suffix."
```
```suggestion
					"Literal suffix functions must return exactly one value."
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3852,8 +3852,13 @@ void TypeChecker::endVisit(Literal const& _literal)
 			if (suffixFunctionType->returnParameterTypes().size() == 1)
 				_literal.annotation().type = suffixFunctionType->returnParameterTypes().front();
 			else
-				_literal.annotation().type = TypeProvider::tuple(suffixFunctionType->returnParameterTypes());
-
+			{
+				m_errorReporter.typeError(
```
cameel: I think we might want to still keep this annotation. The error is not fatal and you can see that due to this some error messages do not make much sense. For example in `with_abi_encode_no_return_parameters.sol` you get this
```
// TypeError 7279: (225-237): Cannot perform packed encoding for a literal. Please convert it to an explicit type first.
```
That's because the annotation is not empty. In `visit(Literal)` we assign the type of the literal and here we update it based on the return type of the suffix. Now that you removed it, `abi.encodePacked()` thinks it's getting an integer literal. If you keep the assignment, we should be getting more sensible error messages.

cameel:
@matheusaaguiar once you're done with the tweaks I suggested, you could try the next point from the list in my PR: `Disallow suffixes with storage and calldata return values.`

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Done.

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Right. Thanks for the explanation! Already changed it.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3852,7 +3853,32 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+			}
+
+			for (Type const* returnType: suffixFunctionType->returnParameterTypes())
+			{
+				auto referenceType = dynamic_cast<ReferenceType const*>(returnType);
+				auto mappingType = dynamic_cast<MappingType const*>(returnType);
+				if (
+					mappingType ||
+					(referenceType && !referenceType->dataStoredIn(DataLocation::Memory))
+				  )
```
stackenbotten: Coding style error

cameel:
I pulled the first 3 commits into the base PR. Please rebase the remaining 2. I squashed some older fixups but the code did not change (yet) so just keeping the top 2 commits should let you avoid conflicts.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3844,13 +3846,23 @@ void TypeChecker::endVisit(Literal const& _literal)
 				)
 					// TODO: Is this triggered when the argument is out of range? Test.
 					parameterTypeMessage = "The type of the literal cannot be converted to the parameters of the suffix function.";
+
+				if (
+					auto expParam = dynamic_cast<IntegerType const*>(suffixFunctionType->parameterTypes().at(1));
```
```suggestion
					auto exponentType = dynamic_cast<IntegerType const*>(suffixFunctionType->parameterTypes().at(1));
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3844,13 +3846,23 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+
 			}
 			else if (!literalType->isImplicitlyConvertibleTo(*suffixFunctionType->parameterTypes().front()))
 				parameterTypeMessage = "The type of the literal cannot be converted to the parameter of the suffix function.";
 
 			if (parameterTypeMessage.has_value())
 				m_errorReporter.typeError(8838_error, _literal.location(), parameterTypeMessage.value());
 
+			if (exponentTypeMessage.has_value())
+				m_errorReporter.typeError(3123_error, _literal.location(), exponentTypeMessage.value());
```
cameel: You can report the error immediately, without introducing a variable for the message.

The only reason I added `parameterTypeMessage` is that I wanted to report the same error in two different places and that triggers the check for duplicate error codes.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3844,13 +3846,23 @@ void TypeChecker::endVisit(Literal const& _literal)
 				)
 					// TODO: Is this triggered when the argument is out of range? Test.
 					parameterTypeMessage = "The type of the literal cannot be converted to the parameters of the suffix function.";
+
+				if (
+					auto expParam = dynamic_cast<IntegerType const*>(suffixFunctionType->parameterTypes().at(1));
+					expParam && expParam->isSigned()
+				)
+					exponentTypeMessage = "The type of the exponent must be unsigned for decimal literal suffix functions.";
```
cameel: Maybe we could also explain why it's disallowed:
```suggestion
					exponentTypeMessage = "The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.";
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3862,6 +3874,25 @@ void TypeChecker::endVisit(Literal const& _literal)
 					"Literal suffix functions must return exactly one value."
 				);
 			}
+
+			if (ranges::any_of(
+					suffixFunctionType->returnParameterTypes(),
+					[](Type const* returnType) {
```
```suggestion
					[](Type const* _returnType) {
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3862,6 +3874,25 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+						auto referenceType = dynamic_cast<ReferenceType const*>(returnType);
+						auto mappingType = dynamic_cast<MappingType const*>(returnType);
+						return
+							mappingType ||
+							(referenceType && !referenceType->dataStoredIn(DataLocation::Memory))
+						;
+					}
+				)
+			)
```
cameel: This is how we'd usually wrap it:
```suggestion
			if (ranges::any_of(
				suffixFunctionType->returnParameterTypes(),
				[](Type const* returnType) {
					auto referenceType = dynamic_cast<ReferenceType const*>(returnType);
					auto mappingType = dynamic_cast<MappingType const*>(returnType);
					return
						mappingType ||
						(referenceType && !referenceType->dataStoredIn(DataLocation::Memory));
				}
			))
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3862,6 +3874,25 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+				m_errorReporter.typeError(
+					7251_error,
+					_literal.location(),
+					"Literal suffix functions must return memory value."
```
```suggestion
					"Literal suffix functions must return memory values."
```

test/libsolidity/syntaxTests/literalSuffixes/parameters/function_with_signed_exponent.sol
```diff
@@ -0,0 +1,14 @@
+struct Float {
```
cameel: To match the naming convention of other files, I'd call this one just `signed_exponent.sol`.

test/libsolidity/syntaxTests/literalSuffixes/parameters/function_with_signed_exponent.sol
```diff
@@ -0,0 +1,14 @@
...
+    uint mantissa;
+    int exponent;
+}
+
+function f(uint mantissa, int exponent) pure returns (Float memory) {
```
cameel: Can you test also smaller integer types. See e.g. how it's done in `exponent_out_of_range.sol`.

I mean, if possible - if the compiler stops at the first error then this test is fine. But the error is not fatal, so I think you should be able to test multiple functions in a single test case.

cameel (commented):
Disallowing of signed exponent should have made some existing tests fail. Please update their expectations.

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Yeah, I saw that and I thought this situation (same error in two different places) would be more visible avoiding the report immediately, albeit it may look a little bit weird. Do you think that makes sense or is it too much?

cameel: I think it's clearer if you report it immediately. I really wish I didn't have to do that with `parameterTypeMessage` either - I'd prefer to just have multiple `typeError()` calls.

matheusaaguiar: Ok, understood,  I'm going to change it

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Done.

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Done.

test/libsolidity/syntaxTests/literalSuffixes/parameters/function_with_signed_exponent.sol
matheusaaguiar: It is possible. Already done.

cameel:
I pushed some changes yesterday to the base branch - they added proper support for strings in IR codegen so that semantic tests now pass. I also added some new semantic tests. Please rebase on the new state of my branch.

Also, since I'm off now - maybe in the meantime you could also add a fix for `decimal_decomposition.sol` in your branch? I think I must have just forgotten the abi coder pragma.

matheusaaguiar:
> maybe in the meantime you could also add a fix for `decimal_decomposition.sol` in your branch? I think I must have just forgotten the abi coder pragma.

I don't get any errors in that test... should  `pragma abicoder v2;` be added?

cameel:
> I don't get any errors in that test... should `pragma abicoder v2;` be added?

There are errors in CI. See one of the `soltest_all` jobs. It didn't fail for me locally either but that's because by default soltest runs with abicoder v2. You have to give it the `--abiencoderv1` option to run with v1. In CI we run it both ways.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3845,6 +3846,17 @@ void TypeChecker::endVisit(Literal const& _literal)
 				)
 					// TODO: Is this triggered when the argument is out of range? Test.
 					parameterTypeMessage = "The type of the literal cannot be converted to the parameters of the suffix function.";
+
+				if (
+					auto exponentType = dynamic_cast<IntegerType const*>(suffixFunctionType->parameterTypes().at(1));
+					exponentType && exponentType->isSigned()
+				)
```
cameel: This should go into the `else` branch above. Otherwise you can't assume that `suffixFunctionType` is not null.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_in_range.sol
```diff
@@ -31,3 +31,12 @@ contract C {
...
+// TypeError 3123: (594-677): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
+// TypeError 3123: (812-896): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
+// TypeError 3123: (1030-1115): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
+// TypeError 3123: (1288-1297): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
+// TypeError 3123: (1347-34120): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
```
cameel: This test is supposed to test the range of the exponent. It should not be failing. Since signed exponent variants are no longer valid, I think it's best to just remove them. Now the `signed_exponent.sol` covers them.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
```diff
@@ -32,18 +32,26 @@ contract C {
...
 // TypeError 8838: (1006-1089): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (1109-1192): The type of the literal cannot be converted to the parameters of the suffix function.
+// TypeError 3123: (1252-1261): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
 // TypeError 8838: (1252-1261): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (1271-1280): The type of the literal cannot be converted to the parameters of the suffix function.
+// TypeError 3123: (1311-34085): The exponent parameter of a literal suffix function must be unsigned. Exponent is always either zero or a negative power of 10 but the parameter represents its absolute value.
 // TypeError 8838: (1311-34085): The type of the literal cannot be converted to the parameters of the suffix function.
```
cameel: Same here. Please remove signed variants.

test/libsolidity/syntaxTests/literalSuffixes/parameters/signed_exponent.sol
```diff
@@ -0,0 +1,18 @@
+function s8(uint mantissa, int8 exponent) pure returns (uint) { }
+function s64(uint mantissa, int64 exponent) pure returns (uint) { }
+function s160(uint mantissa, int160 exponent) pure returns (uint) { }
+function s256(uint mantissa, int256 exponent) pure returns (uint) { }
```
cameel: Can you add some variants with different mantissa types, especially signed ones?

cameel:
Some tweaks needed but overall looks good. I'll pull it into my branch when you apply them.

But you can already start working on some later points from the checklist. I updated the PR description above - choose one of those I added.

libsolidity/analysis/TypeChecker.cpp
matheusaaguiar: Sorry, I didn't understand this clearly. I assumed `suffixFunctionType` can't be null at this point because of a previous test at line `3802`.

cameel: That's only if it's a fatal error. This one is not.

matheusaaguiar: Ah, understood it now. Thanks.

matheusaaguiar: But isn't this inside the `else` branch of that test? I mean, if `suffixFunctionType` is null, then this block doesn't execute.
Maybe I'm missing something...

cameel: Damn, I think I got bamboozled by github's crappy rendering of tabs. For me it sometimes chooses with randomly for every nesting level and also, the width for some reason slightly shifts at transition between commentable and non-commentable code :|

Yeah, it seems fine now but I had to look closely and count indents to notice that...

matheusaaguiar: Yeah, sometimes the rendering of tabs mess with indents...it is really confusing.

test/libsolidity/syntaxTests/literalSuffixes/parameters/signed_exponent.sol
matheusaaguiar: Done.

libsolidity/ast/Types.cpp
```diff
@@ -1232,14 +1232,22 @@ pair<RationalNumberType const*, RationalNumberType const*> RationalNumberType::m
...
+		return {nullptr, TypeProvider::rationalNumber(-exponent)};
+
 	while (unsignedMantissa.denominator() != 1)
 	{
 		unsignedMantissa *= 10;
 		--exponent;
 
-		// FIXME: What happens when exponent in scientific notation is max uint?
-		if (unsignedMantissa > maxMantissa || exponent > maxUint)
+		if(exponent > maxUint && unsignedMantissa > maxMantissa)
```
stackenbotten: Coding style error

stackenbotten: Coding style error

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3768,14 +3769,22 @@ bool TypeChecker::visit(Literal const& _literal)
...
+			string mantissaOrExponentErrorMessage = "This fractional number cannot be decomposed into a ";
+
+			if (!mantissa && ! exponent)
+				mantissaOrExponentErrorMessage += "mantissa and exponent that fit ";
+			else if (!exponent)
+				mantissaOrExponentErrorMessage += "exponent that fits ";
+			else if (!mantissa)
+				mantissaOrExponentErrorMessage += "mantissa that fits ";
+
+			mantissaOrExponentErrorMessage += "the range of parameters of any possible suffix function.";
```
nikola-matic: This is a bit of a nit, but are we allowed to use `fmt::format` here, as it would make this a little bit less verbose?

matheusaaguiar: I guess we can since I've found uses of it on other parts of the code. I was not pleased by this `if ... else` form too. Thanks!

nikola-matic: You're still gonna need the `if else` though :D

matheusaaguiar: Ah, right!
I guess you are suggesting something like this:
```suggestion
	string mantissaOrExponentErrorMessage;

	if (!mantissa && ! exponent)
		mantissaOrExponentErrorMessage = "mantissa and exponent that fit";
	else if (!exponent)
		mantissaOrExponentErrorMessage = "exponent that fits";
	else if (!mantissa)
		mantissaOrExponentErrorMessage = "mantissa that fits";

	m_errorReporter.typeError(
		5503_error,
		_literal.location(),
		fmt::format(
			"This fractional number cannot be decomposed into a {} "
			"the range of parameters of any possible suffix function.",
			mantissaOrExponentErrorMessage
		)
	);
```

nikola-matic: Exactly, plus the last `else if (!mantissa)` can be just an else, since by the time you get to the branch, you're left with only one path.

matheusaaguiar: Yeah, good point!
Thanks! This feels cleaner now.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3863,6 +3891,23 @@ void TypeChecker::endVisit(Literal const& _literal)
 					"Literal suffix functions must return exactly one value."
 				);
 			}
+
+			if (ranges::any_of(
```
nikola-matic: What's the advantage of `ranges::any_of` instead of `std::any_of` in this case?

ekpyron: The advantage is that it works on ranges, not merely on iterator pairs, i.e. ``suffixFunctionType->returnParameterTypes()`` vs ``std::begin(suffixFunctionType->returnParameterTypes()), std::end(suffixFunctionType->returnParameterTypes())``.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3871,6 +3899,23 @@ void TypeChecker::endVisit(Literal const& _literal)
...
+					auto mappingType = dynamic_cast<MappingType const*>(_returnType);
+					return
+						mappingType ||
+						(referenceType && !referenceType->dataStoredIn(DataLocation::Memory));
+				}
+			))
+				m_errorReporter.typeError(
+					7251_error,
+					_literal.location(),
+					"Literal suffix functions must return memory values."
```
cameel: This is not really true. They can also return value types.

cameel: Fixed.

cameel:
I pulled the "Disallow suffixes with storage and calldata return values" part into the base PR. I'm still reviewing the rest of the PR.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3777,14 +3779,28 @@ bool TypeChecker::visit(Literal const& _literal)
...
 	{
 		auto&& [mantissa, exponent] = literalRationalType->mantissaExponent();
-		solAssert((mantissa && exponent) || (!mantissa && !exponent));
-		if (!mantissa)
+
+		if (!mantissa || !exponent)
+		{
+			string mantissaOrExponentErrorMessage;
+
+			if (!mantissa && ! exponent)
```
```suggestion
			if (!mantissa && !exponent)
```

libsolidity/ast/Types.cpp
```diff
@@ -1232,14 +1232,22 @@ pair<RationalNumberType const*, RationalNumberType const*> RationalNumberType::m
...
 
-		// FIXME: What happens when exponent in scientific notation is max uint?
-		if (unsignedMantissa > maxMantissa || exponent > maxUint)
+		if (exponent > maxUint && unsignedMantissa > maxMantissa)
 			return {nullptr, nullptr};
+		// FIXME: What happens when exponent in scientific notation is max uint?
+		if (exponent > maxUint)
+			return {TypeProvider::rationalNumber(unsignedMantissa), nullptr};
+		if (unsignedMantissa > maxMantissa)
+			return {nullptr, TypeProvider::rationalNumber(-exponent)};
```
cameel, line 1293:
```suggestion
		// FIXME: What happens when exponent in scientific notation is max uint?
		return {
			unsignedMantissa <= maxMantissa ? TypeProvider::rationalNumber(unsignedMantissa) : nullptr,
			exponent <= maxUint ? TypeProvider::rationalNumber(-exponent) : nullptr,
		};
```

libsolidity/ast/Types.cpp
```diff
@@ -1232,14 +1232,22 @@ pair<RationalNumberType const*, RationalNumberType const*> RationalNumberType::m
 
 	rational exponent = 0;
 	rational unsignedMantissa = abs(m_value);
+
+	if (unsignedMantissa > maxMantissa)
+		return {nullptr, TypeProvider::rationalNumber(-exponent)};
```
cameel: Oh, that's a good catch. So this is actually a bug in the base branch. I'm pulling it in as a fix.

libsolidity/ast/Types.h
```diff
@@ -592,7 +592,9 @@ class RationalNumberType: public Type
 	/// base-10 exponent, such that the number is equal to `mantissa * 10**-exponent`.
 	/// @returns Pair of non-null pointers representing the types of the literals corresponding to
 	/// mantissa and exponent if the resulting mantissa and exponent both fit in 256 bits.
-	/// A pair of null pointers otherwise.
+	/// If mantissa alone cannot fit in 256 bits, only the first pointer of the Pair is null.
+	/// If exponent alone cannot fit in 256 bits, only the second pointer of the Pair is null.
```
cameel:
```suggestion
	/// If either of these values does not fit in 256 bits, its value in the pair is null.
```

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3777,14 +3779,28 @@ bool TypeChecker::visit(Literal const& _literal)
...
 			m_errorReporter.typeError(
 				5503_error,
 				_literal.location(),
-				"This fractional number cannot be decomposed into a mantissa and decimal exponent "
-				"that fit the range of parameters of any possible suffix function."
+				fmt::format(
+					"This fractional number cannot be decomposed into a {} "
+					"the range of parameters of any possible suffix function.",
+					mantissaOrExponentErrorMessage
+				)
```
cameel: This new message does not make sense when you specify only mantissa or only exponent. Decomposition is always into both, it's just that sometimes one of them is out of range.

I think it would be better to keep the original message here and just append an extra message at the end, saying that one or both are out of range of the largest supported integer type.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_in_range.sol
```diff
@@ -1,33 +1,26 @@
...
...0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001 u16;
```
cameel: You need to increase this to `1e-65535`. 32k is not the maximum value of `uint16`. Though even for 32k it's slow (note for example that bytecode jobs are now timing out) so I wonder if it's worth it...

Also, this shouldn't have been moved to `min()`. The minimum value of the exponent is zero. 65k should be in `max()`.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_out_of_range.sol
```diff
@@ -1,49 +1,31 @@
...
...0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001 u16;
```
cameel: Same as above. This is not out of range of `uint16`.  It has to be `1e-65536` now.

Note that the error message for this line disappeared. It's easy to see because there was one with a very wide range (`1311-34085`) and now all of them are short. When modifying tests, please always make sure that the new error messages still make sense.

cameel (commented):
@matheusaaguiar I pulled what I could into the base branch. Here are are some comments for the remaining stuff. You can continue working on it now.

libsolidity/ast/Types.cpp
matheusaaguiar: But, in this way it will always return at the first loop iteration, no?

cameel: Ah. Right. Forgot about the `else` case :)

You could add an `if` to return only if both pair elements are non-empty.

matheusaaguiar: Sorry, could you explain this? I don't understand what you mean here.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_in_range.sol
matheusaaguiar: Yeah, it was already pretty slow... should we increase to 65k or just remove the test for `uint16` max exponent altogether?

libsolidity/analysis/DeclarationTypeChecker.cpp
```diff
@@ -291,6 +292,16 @@ void DeclarationTypeChecker::endVisit(ArrayTypeName const& _typeName)
...
 	if (Expression const* length = _typeName.length())
 	{
+		if (auto literal = dynamic_cast<Literal const*>(length))
+		{
+			if (!holds_alternative<Literal::SubDenomination>(literal->suffix()))
+				m_errorReporter.typeError(
+					4127_error,
+					literal->location(),
+					"Array length cannot contain a suffix function."
+				);
```
cameel: Can you merge this with error 5462 (`Invalid array length, expected integer literal or constant expression.`)? I think that both situations should report the same error ID. No point defining a special one just for suffixes.

If you really want, you could add something like this if you detect that it has a suffix: `A suffixed literal is not a constant expression unless the suffix is a denomination`. But it would be completely fine even without it.

test/libsolidity/syntaxTests/literalSuffixes/parameters/exponent_in_range.sol
cameel: For now let's increase it. We can decide later if it's worth keeping it.

libsolidity/ast/Types.cpp
cameel: I meant that you would instead of returning as I suggested, you'd assign the pair to a variable, then check if both pair elements are null and return only if they're not.

But, well, maybe no point. When I commented that, I thought the `?:` version was more readable, but now going back to this, yours seems fine. Let's keep it. It's just a tiny detail anyway.

test/libsolidity/syntaxTests/literalSuffixes/parameters/mantissa_out_of_range.sol
```diff
@@ -37,16 +37,16 @@ contract C {
 // TypeError 8838: (549-558): The type of the literal cannot be converted to the parameters of the suffix function.
 // TypeError 8838: (568-577): The type of the literal cannot be converted to the parameters of the suffix function.
```
cameel: As the TODO above says, our goal is to improve all of these errors. We should make it clear to the user why exactly we cannot convert the value. Right now the message is clear only for the extreme cases that exceed `uint256`. We need to do better.

Looks like these cases currently get rejected because there's no implicit conversion to the smaller type. I think that just indicating in the current message what we're trying to convert to what type would be good enough.

This is especially important for fractional numbers because then the value is different from what user sees in the literal - e.g. you should say that mantissa is x, exponent is y and x or y cannot be converted to some type T.

cameel (commented):
Sorry for not answering earlier. I've been trying to get more work done on operators done :) Here are some more answers and more pointers.

cameel:
> - [ ] Include suffix definition as a secondary location in errors.

Some hints on this since I just had to do this in the operators PR. To add a secondary location, you simply put an instance of `SecondarySourceLocation` into the third argument of helpers like `m_errorReporter.typeError()`. Here's an example from operators:

```c++
m_errorReporter.typeError(
	1884_error,
	functionDefinition.parameterList().location(),
	SecondarySourceLocation().append(
		"Function was used to implement an operator here:",
		path->location()
	),
	fmt::format(
		"Wrong parameters in operator definition. "
		"The function \"{}\" needs to have {} to be used for the operator {}.",
		joinHumanReadable(path->path(), "."),
		wrongParametersMessage.value(),
		TokenTraits::friendlyName(*operator_)
	)
)
```

You can add multiple secondary locations with multiple `append()` calls. I do that for example when there are multiple definitions of an operator to show them all.

Note that secondary locations do not show up test expectations. You have to compile the file with `solc` to see them.

libsolidity/analysis/DeclarationTypeChecker.cpp
matheusaaguiar: Merged and appended a message addressing the suffixed literal error case.

cameel:
I rebased this together with the base PR. Both are now on top of the operators PR that is currently in review.

cameel:
I'm pulling the commits for signed exponent check and array size check into the base branch.

cameel:
Also, looks like the tests with huge exponent massively increase the running time of `soltest_all` (from 5 min to 30+ min). Given that it's very parallelized (has 50 parallel jobs), that's a bit too much. I'm disabling those tests for now in the base branch.

cameel:
The syntax and type checking for suffix definitions is now done in the base branch. Now definitions need the `suffix` modifier. Some locations in error messages shifted due to this, which will likely cause conflicts when your rebase. Also note that I moved some of the type checks from `visit(Literal)` to `visit(FunctionDefinition)` because now we immediately know that a function is meant to be used as a suffix and we don't have to defer them until it's actually called.

cameel:
> - [ ] Fix: Suffixes not detected as function calls by control flow graph (no "unreachable code" warning).

Here's a hint about how to tackle this point. It refers to the TODO note in `syntaxTests/controlFlow/unreachableCode/revert_in_literal_suffix.sol`:

```solidity
function suffix(uint) pure suffix returns (uint) {
    revert();
}

contract C {
    function f() public pure {
        1 suffix;
        uint a = 0; // TODO: Should warn about unreachable code here
        a;
    }
}
```

The suffix unconditionally reverts so anything after `1 suffix` is unreachable and should trigger a warning. This does not happen. I'm pretty sure it's because the suffix is not treated as a function call by the control flow graph. Please take a look at how the operators PR (#13790) handled this in `libsolidity/analysis/ControlFlowBuilder.cpp`. We need to do something similar.

cameel:
Not sure what happened here, but this was supposed to be closed, not merged into the base branch.

matheusaaguiar:
Sorry about that. While trying to set `literal_suffix_functions` as the base branch for another PR, I must have done something wrong which resulted in this.

PR #13419, opened 2022-08-20
Update misleading information about OpenZeppelin's ECDSA library

rafales:
There were many changes to the ECDSA library in OpenZeppelin's library and referenced functions are no longer safe to use in this manner since 4.1 version of the library.

In short - two versions of the functions which accept signature as `bytes` accept both 65-byte regular signature and 64-byte EIP-2098 signature.

### DIFF
--- a/docs/units-and-global-variables.rst
+++ b/docs/units-and-global-variables.rst
@@ -231,8 +231,8 @@ Mathematical and Cryptographic Functions
     for _transaction_ signatures (see `EIP-2 <https://eips.ethereum.org/EIPS/eip-2#specification>`_), but
     the ecrecover function remained unchanged.
 
-    This is usually not a problem unless you require signatures to be unique or
-    use them to identify items. OpenZeppelin have a `ECDSA helper library <https://docs.openzeppelin.com/contracts/2.x/api/cryptography#ECDSA>`_ that you can use as a wrapper for ``ecrecover`` without this issue.
+    This is usually not a problem unless you require signatures to be unique or use them to identify items.
+    OpenZeppelin have a `ECDSA helper library <https://docs.openzeppelin.com/contracts/4.x/api/utils#ECDSA>`_ that you can use as a wrapper for ``ecrecover`` without this issue.
 
 .. note::
 


rafales:
It seems that this issue has been fixed in v4.7.3 released 10 days ago in OpenZeppelin's library, but the bug/feature was there for a long time and many people may be using outdated libraries. Should I change the wording to warn user about using older versions or just drop the PR altogether?

docs/units-and-global-variables.rst
```diff
@@ -232,7 +232,9 @@ Mathematical and Cryptographic Functions
     the ecrecover function remained unchanged.
 
     This is usually not a problem unless you require signatures to be unique or
-    use them to identify items. OpenZeppelin have a `ECDSA helper library <https://docs.openzeppelin.com/contracts/2.x/api/cryptography#ECDSA>`_ that you can use as a wrapper for ``ecrecover`` without this issue.
+    use them to identify items.
+    
+    OpenZeppelin have a `ECDSA helper library <https://docs.openzeppelin.com/contracts/4.x/api/utils#ECDSA>`_ that you can use as a wrapper for ``ecrecover``, but BE CAREFUL which helper you pick - both `tryRecover(hash, signature)` and `recover(hash, signature)` accept signatures in two forms (65 and 64 bytes), thus are not fixing anything. Use helpers which accept `r`, `v`, `s` arguments instead.
```
leonardoalt: I agree with the suggestion, but the current wording sounds rather confusing to me. "Why are they not fixing anything", a reader will likely ask. Is there an external link from OZ explaining those things that we can just link here?

rafales: Actually the fixed it ~2 weeks ago, the bug was present for 1 year and 4 months there. See my second comment: https://github.com/ethereum/solidity/pull/13419#issuecomment-1221418771

So i'm wondering if we should drop this PR or just warn users about the problem in older versions.

leonardoalt: We could keep the part of the PR that updates the OZ version in the link but remove the text part. Wdyt?

axic: Why not just remove the mention of OpenZeppelin?

nikola-matic: Whilst it's a bit weird having our docs point to a third party library, I don't think it's necessarily bad that we're mentioning OZ, since they're a major part of the ecosystem, so I'd side with @leonardoalt; simply update the OZ link to point to a later version, and remote the rest.

nikola-matic:
@rafales I've pushed the agreed upon changes do your fork's branch.
@leonardoalt I've removed some line breaks in order to conform with Kamil's suggestion, which is to try and keep a sentence on a single line as much as possible, in order to make reviews easier when future changes occur.

PR #13422, opened 2022-08-22
Cleanup static z3 script to work similar to release_ppa

Marenz:


scripts/common.sh
```diff
@@ -40,6 +40,26 @@ else
     function printLog { echo "$(tput setaf 3)$1$(tput sgr0)"; }
 fi
 
+function checkDputEntries
+{
+    local pattern="$1"
+    grep "${pattern}" /etc/dput.cf --quiet || \
+        fail "Error: Missing ${pattern//\\/} section in /etc/dput.cf (check top comment in ${0} for more information)."
```
cameel: I guess not every file where we might use this will have this comment.
```suggestion
        fail "Error: Missing ${pattern//\\/} section in /etc/dput.cf (check top comment in release_ppa.sh for more information)."
```

Or maybe it would be better to have something like `common_ppa.sh` and put the comments there?

cameel: Same for `sourcePPAConfig()`.

scripts/common.sh
```diff
@@ -40,6 +40,26 @@ else
...
+}
+
+function sourcePPAConfig
+{
+    # source keyid and email from .release_ppa_auth
```
cameel, line 54: I just noticed that this will accept variables exported from the shell - but only if `.release_ppa_auth` does not override them. Is that intentional? If not, maybe we should assert against that:

```suggestion
    [[ "$LAUNCHPAD_KEYID" == "" && "$LAUNCHPAD_EMAIL" == "" ]] || fail

    # source keyid and email from .release_ppa_auth
```

And if not, then it's actually not the way I'd expect. I think that normally env variables are used to override config settings and not the other way around.

scripts/release_ppa.sh
```diff
@@ -47,7 +47,7 @@ set -e
 
 REPO_ROOT="$(dirname "$0")/.."
 
-# for the "fail" function
+# Import "fail" and "checkDputEntries"
```
cameel:
```suggestion
# Import "fail", "checkDputEntries" and "sourcePPAConfig"
```
But I'm not sure if it's a good idea to have these listed here. The list will easily go out of sync.

scripts/deps-ppa/static_z3.sh
```diff
@@ -20,14 +20,28 @@
 ##
```
cameel, line 24: Comment above has hard-coded e-mail and needs update.

scripts/deps-ppa/static_z3.sh
```diff
@@ -20,14 +20,28 @@
...
 ##############################################################################
 
-set -ev
+set -e
 
-keyid=70D110489D66E2F6
-email=builds@ethereum.org
 packagename=z3-static
-version=4.8.17
+version=${1}
```
cameel:
```suggestion
version="$1"
```

scripts/deps-ppa/static_z3.sh
```diff
@@ -20,14 +20,28 @@
...
+then
+    fail "Usage: $0 v<version>";
+fi
+
+sourcePPAConfig
+
+# Sanity check
+checkDputEntries "\[cpp-build-deps\]"
+
+DISTRIBUTIONS="focal jammy kinetic"
```
cameel, line 44: As before, I'd put changes like this in a separate commit.

scripts/deps-ppa/static_z3.sh
```diff
@@ -20,14 +20,28 @@
...
+REPO_ROOT="$(dirname "$0")/../.."
+
+# Import "fail" and "checkDputEntries"
+# shellcheck source=/dev/null
+source "${REPO_ROOT}/scripts/common.sh"
+
+if [[ $version == "" ]]
+then
+    fail "Usage: $0 v<version>";
+fi
```
cameel:
```suggestion
[[ $version != "" ]] || fail "Usage: $0 v<version>"
```

scripts/deps-ppa/static_z3.sh
```diff
@@ -226,26 +240,26 @@ cd ..
...
+orig_sha1=$(sha1sum "$orig" | cut -d ' ' -f 1)
+orig_sha256=$(sha256sum "$orig" | cut -d ' ' -f 1)
+orig_md5=$(md5sum "$orig" | cut -d ' ' -f 1)
 
-if wget --quiet -O $orig-tmp "$ppafilesurl/$orig"
+if wget --quiet -O "$orig-tmp" "$ppafilesurl/$orig"
 then
     echo "[WARN] Original tarball found in Ubuntu archive, using it instead"
-    mv $orig-tmp $orig
+    mv "$orig-tmp" "$orig"
```
cameel: If you're adding quotes I'd also add braces. It looks a bit as if `orig-tmp` was the variable name.
```suggestion
    mv "${orig}-tmp" "$orig"
```

cameel (commented):
Some suggestions but nothing critical.

leonardoalt (approved):

PR #13423, opened 2022-08-22
allow setting remappings on lsp server init

chrisfarms:
### What

enables client's of the lsp server to configure remappings during the initialisation command.

we already allow configuring import-paths through the initialisation, so hopefully this is not controversial.

the format expected to be passed in from the initialise json is:

```
{ 
  "remappings": [
      {context: "", prefix: "foo", target: "bar"},
      ...
  ]
}
```

### Why

Because without being able to set remappings, the lsp-server throws errors for any project that requires them.

### Related

* https://github.com/ethereum/solidity/issues/12603
* https://github.com/ethereum/solidity/pull/12994

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ Compiler Features:
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 * Language Server: Add basic document hover support.
+* Language Server: Add configuration option to apply custom remappings via ``remappings`` in the LSP settings object.
 
 
 Bugfixes:
--- a/libsolidity/lsp/FileRepository.h
+++ b/libsolidity/lsp/FileRepository.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <libsolidity/interface/FileReader.h>
+#include <libsolidity/interface/ImportRemapper.h>
 #include <libsolutil/Result.h>
 
 #include <string>
@@ -29,10 +30,13 @@ namespace solidity::lsp
 class FileRepository
 {
 public:
-	FileRepository(boost::filesystem::path _basePath, std::vector<boost::filesystem::path> _includePaths);
+	FileRepository(boost::filesystem::path _basePath, std::vector<boost::filesystem::path> _includePaths, std::vector<frontend::ImportRemapper::Remapping> _remappings);
 
 	std::vector<boost::filesystem::path> const& includePaths() const noexcept { return m_includePaths; }
 	void setIncludePaths(std::vector<boost::filesystem::path> _paths);
+
+	std::vector<frontend::ImportRemapper::Remapping> const& remappings() const noexcept { return m_remappings; }
+	void setRemappings(std::vector<frontend::ImportRemapper::Remapping> _remappings);
 
 	boost::filesystem::path const& basePath() const { return m_basePath; }
 
@@ -63,6 +67,9 @@ private:
 
 	/// Additional directories used for resolving relative paths in imports.
 	std::vector<boost::filesystem::path> m_includePaths;
+
+	/// Remappings of imports.
+	std::vector<frontend::ImportRemapper::Remapping> m_remappings;
 
 	/// Mapping of source unit names to their URIs as understood by the client.
 	StringMap m_sourceUnitNamesToUri;
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -18,6 +18,7 @@
 #include <libsolidity/ast/AST.h>
 #include <libsolidity/ast/ASTUtils.h>
 #include <libsolidity/ast/ASTVisitor.h>
+#include <libsolidity/interface/ImportRemapper.h>
 #include <libsolidity/interface/ReadFile.h>
 #include <libsolidity/interface/StandardCompiler.h>
 #include <libsolidity/lsp/LanguageServer.h>
@@ -151,7 +152,7 @@ LanguageServer::LanguageServer(Transport& _transport):
 		{"textDocument/semanticTokens/full", bind(&LanguageServer::semanticTokensFull, this, _1, _2)},
 		{"workspace/didChangeConfiguration", bind(&LanguageServer::handleWorkspaceDidChangeConfiguration, this, _2)},
 	},
-	m_fileRepository("/" /* basePath */, {} /* no search paths */),
+	m_fileRepository("/" /* basePath */, {} /* no search paths */, {} /* no remappings */),
 	m_compilerStack{m_fileRepository.reader()}
 {
 }
@@ -212,6 +213,40 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 		if (typeFailureCount)
 			m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
 	}
+
+	Json::Value jsonRemappings = _settings["remappings"];
+
+	if (jsonRemappings)
+	{
+		bool typeFailures = false;
+		if (jsonRemappings.isArray())
+		{
+			vector<frontend::ImportRemapper::Remapping> remappings;
+			for (Json::Value const& jsonPath: jsonRemappings)
+			{
+				if (
+					jsonPath.isObject() &&
+					jsonPath.get("context", "").isString() &&
+					jsonPath.get("prefix", "").isString() &&
+					jsonPath.get("target", "").isString()
+				)
+					remappings.emplace_back(ImportRemapper::Remapping{
+						jsonPath.get("context", "").asString(),
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
+					});
+				else
+					typeFailures = true;
+			}
+			m_fileRepository.setRemappings(std::move(remappings));
+		}
+		else
+			typeFailures = true;
+
+		if (typeFailures)
+			m_client.trace("Invalid JSON configuration passed. \"remappings\" should be of form [{\"context\":\"\", \"prefix\":\"foo\", \"target\":\"bar\"}].");
+
+	}
 }
 
 vector<boost::filesystem::path> LanguageServer::allSolidityFilesFromProject() const
@@ -237,7 +272,7 @@ void LanguageServer::compile()
 	// For files that are not open, we have to take changes on disk into account,
 	// so we just remove all non-open files.
 
-	FileRepository oldRepository(m_fileRepository.basePath(), m_fileRepository.includePaths());
+	FileRepository oldRepository(m_fileRepository.basePath(), m_fileRepository.includePaths(), m_fileRepository.remappings());
 	swap(oldRepository, m_fileRepository);
 
 	// Load all solidity files from project.
@@ -261,6 +296,7 @@ void LanguageServer::compile()
 	// TODO: optimize! do not recompile if nothing has changed (file(s) not flagged dirty).
 
 	m_compilerStack.reset(false);
+	m_compilerStack.setRemappings(m_fileRepository.remappings());
 	m_compilerStack.setSources(m_fileRepository.sourceUnits());
 	m_compilerStack.compile(CompilerStack::State::AnalysisPerformed);
 }
@@ -404,7 +440,7 @@ void LanguageServer::handleInitialize(MessageID _id, Json::Value const& _args)
 	if (_args["trace"])
 		setTrace(_args["trace"]);
 
-	m_fileRepository = FileRepository(rootPath, {});
+	m_fileRepository = FileRepository(rootPath, {}, {});
 	if (_args["initializationOptions"].isObject())
 		changeConfiguration(_args["initializationOptions"]);
 
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -1085,6 +1085,13 @@ class SolidityLSPTestSuite: # {{{
         )
         return self.wait_for_diagnostics(solc_process)
 
+    def expect_report(self, published_diagnostics, uri):
+        for report in published_diagnostics:
+            if report['uri'] == uri:
+                return report
+        self.expect_true(False, f"expected ${uri} in published_diagnostics")
+        return None
+
     def expect_true(
         self,
         actual,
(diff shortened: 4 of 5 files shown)


christianparpart:
Many thanks for your contribution @chrisfarms. We will take a close look at your work. :)

libsolidity/lsp/FileRepository.h
```diff
@@ -64,6 +68,9 @@ class FileRepository
 	/// Additional directories used for resolving relative paths in imports.
 	std::vector<boost::filesystem::path> m_includePaths;
 
+	/// Remappings of imports
```
christianparpart:
```suggestion
	/// Remappings of imports.
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -20,6 +20,7 @@
 #include <libsolidity/ast/ASTVisitor.h>
 #include <libsolidity/interface/ReadFile.h>
 #include <libsolidity/interface/StandardCompiler.h>
+#include <libsolidity/interface/ImportRemapper.h>
```
christianparpart: includes are grouped by category and per category alphabetically sorted. I think this line should go 2 lines up, above ReadFile, if I'm not mistaken.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -171,14 +172,48 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+			for (Json::Value const& jsonPath: jsonRemappings)
+			{
+				if (jsonPath.isObject())
+				{
+					frontend::ImportRemapper::Remapping remapping = {
+						jsonPath.get("context", "").asString(),
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
+					};
+					remappings.emplace_back(remapping);
```
christianparpart:
```suggestion
					remappings.emplace_back(ImportRemapper::Remapping{
						jsonPath.get("context", "").asString(),
						jsonPath.get("prefix", "").asString(),
						jsonPath.get("target", "").asString(),
					});
```

`frontend` namespace was imported on top of the file already and  constructing the parameter directly as part of the emplace-call will not make you accidentally forget to "move" the temporary.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -171,14 +172,48 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+					typeFailureCount++;
+				}
+			}
+			m_fileRepository.setRemappings(move(remappings));
+		}
+		else
+			++typeFailureCount;
+
+		if (typeFailureCount)
+			m_client.trace("Invalid JSON configuration passed. \"remappings\" must be an array of objects.");
```
christianparpart: I think that's subjective, but maybe it's worth not just mentioning "array of objects" but rather say what that object should be like, especially since it's only three members that it must contain. But then it should be most useful to the reader.

christianparpart (changes requested):
I am sure you did your own interactive tests to make sure it all works as expected. 
But maybe we also have that automatically tested as part of the still young LSP CI tests.

For that you might want to look at `test/lsp.py` and its companion test dir `test/libsolidity/lsp/`. In case of concerns, @Marenz can be of great help with respect to that. :)

chrisfarms:
Thanks, addressed your review comments.

Will take a look at the lsp.py tests shortly (had not spotted those tests) 👍

chrisfarms:
added a very basic case to the python lsp tests

test/lsp.py
```diff
@@ -1408,6 +1408,38 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
...
+                        {
+                            'context': '',
+                            'prefix': '@other/',
+                            'target': f"{self.project_root_dir}/other-include-dir/otherlib/"
+                        }
+                    ]
+                }
+            }
+        )
+        published_diagnostics = self.open_file_and_wait_for_diagnostics(solc, 'include-paths/remapped-include')
```
Marenz: It looks to me like you need the content of this file only in this function and not for any of the "in file defined" test mechanics
Given that it serves no purpose other than in this function here, it would be nicer to just embed file test content directly like it is done in the function `test_textDocument_opening_two_new_files_edit_and_close`

You can still use the `otherlib.sol` for the import of course.

chrisfarms: 👍

Marenz:
I rebased the PR to the latest `develop` branch to fix the pylint CI failure

Marenz (dismissed):

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,39 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+					remappings.emplace_back(ImportRemapper::Remapping{
+						jsonPath.get("context", "").asString(),
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
+					});
+				}
+				else
+				{
+					typeFailureCount++;
+				}
```
leonardoalt: Nitpick: I think our style guide says no curly braces for single statement if/else blocks.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,39 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 		if (typeFailureCount)
 			m_client.trace("Invalid JSON configuration passed. \"include-paths\" must be an array of strings.");
 	}
+
+	Json::Value jsonRemappings = _settings["remappings"];
+
+	if (jsonRemappings)
+	{
+		int typeFailureCount = 0;
```
leonardoalt: This never goes negative so could/should be size_t/unsigned.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,39 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+				}
+				else
+				{
+					typeFailureCount++;
+				}
+			}
+			m_fileRepository.setRemappings(move(remappings));
+		}
+		else
+			++typeFailureCount;
```
leonardoalt: This number is not used to `break` the loop at any point and only used as a `bool` later, so I think it should be a bool instead.

leonardoalt (changes requested):

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,39 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
+					});
+				}
+				else
+				{
+					typeFailureCount++;
+				}
+			}
+			m_fileRepository.setRemappings(move(remappings));
```
leonardoalt:
```suggestion
			m_fileRepository.setRemappings(std::move(remappings));
```

christianparpart:
@chrisfarms I have been taking over this PR to finish the review, simply applying @leonardoalt's comments & adding a changelog item. I hope you don't mind.

Marenz (dismissed):

chrisfarms:
> @chrisfarms I have been taking over this PR to finish the review, simply applying @leonardoalt's comments & adding a changelog item. I hope you don't mind.

🙏 don't mind at all, thanks!

Changelog.md
```diff
@@ -9,6 +9,7 @@ Language Features:
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Language Server: Adding configuration option to apply custom remappings via ``remappings`` in the LSP settings object.
```
leonardoalt:
```suggestion
 * Language Server: Add configuration option to apply custom remappings via ``remappings`` in the LSP settings object.
```
Nitpick, but the gerund is not the standard in the changelog

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,35 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+		if (jsonRemappings.isArray())
+		{
+			vector<frontend::ImportRemapper::Remapping> remappings;
+			for (Json::Value const& jsonPath: jsonRemappings)
+			{
+				if (jsonPath.isObject())
+					remappings.emplace_back(ImportRemapper::Remapping{
+						jsonPath.get("context", "").asString(),
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
```
leonardoalt, line 236: What if these are not strings? Shouldn't this also imply `typeFailures = true`?

leonardoalt: It would also be nice to have tests for these.

leonardoalt (changes requested):
Need some clarification

christianparpart:
Adding @cameel for the final review.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -210,6 +211,40 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+		{
+			vector<frontend::ImportRemapper::Remapping> remappings;
+			for (Json::Value const& jsonPath: jsonRemappings)
+			{
+				if (
+					jsonPath.isObject() &&
+					jsonPath.get("context", "").isString() &&
+					jsonPath.get("prefix", "").isString() &&
+					jsonPath.get("target", "").isString()
+				)
```
cameel, line 232: Should we report an error on unexpected keys?

test/lsp.py
```diff
@@ -1534,6 +1541,70 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
         self.expect_equal(len(diagnostics), 1, "no diagnostics")
         self.expect_diagnostic(diagnostics[0], code=2018, lineNo=5, startEndColumns=(4, 62))
 
+    def test_remapping_wrong_paramateres(self, solc: JsonRpcProcess) -> None:
```
cameel, line 1544:
```suggestion
    def test_remapping_wrong_parameters(self, solc: JsonRpcProcess) -> None:
```
Also, something like `test_remapping_wrong_parameter_types` would be more precise.

libsolidity/lsp/LanguageServer.cpp
cameel, line 232: This is missing a validation that `target` is not empty. solc won't accept remappings with an empty target on the CLI.

test/lsp.py
```diff
@@ -1534,6 +1541,70 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
...
+        published_diagnostics = self.wait_for_diagnostics(solc)
+        self.expect_equal(len(published_diagnostics), 2, "expected reports for two files")
+
+        report = self.expect_report(published_diagnostics, file_with_remapped_import_uri)
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 0, "no diagnostics")
+
+        report = self.expect_report(published_diagnostics, f"{self.project_root_uri}/other-include-dir/otherlib/otherlib.sol")
+        diagnostics = report['diagnostics']
+        self.expect_equal(len(diagnostics), 1, "no diagnostics")
```
cameel, line 1605: I think that one of these `"no diagnostics"` messages does not match the number :)

test/lsp.py
```diff
@@ -1534,6 +1541,70 @@ def test_custom_includes(self, solc: JsonRpcProcess) -> None:
...
+                            'prefix': '@other/',
+                            'target': f"{self.project_root_dir}/other-include-dir/otherlib/"
+                        }
+                    ]
+                }
+            }
+        )
+
+        # test file
+        file_with_remapped_import_uri = 'file:///remapped-include.sol'
```
cameel, line 1583: It would be best to set this to something that would match the remapping if used in an import to show that these URIs are not remapped.

cameel (commented):
Only found minor stuff. Overall the feature seems sound.

cameel:
Does this fix #12603? If so, please add `#Fixes #12603` in the PR description.

Also needs rebase on latest `develop` and approvals from @christianparpart and @leonardoalt.

nikola-matic:
Rebased and resolved conflicts, should be in a more workable state now.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -212,6 +213,40 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+					remappings.emplace_back(ImportRemapper::Remapping{
+						jsonPath.get("context", "").asString(),
+						jsonPath.get("prefix", "").asString(),
+						jsonPath.get("target", "").asString(),
+					});
```
nikola-matic, line 237:
```suggestion
					remappings.emplace_back(
						jsonPath.get("context", "").asString(),
						jsonPath.get("prefix", "").asString(),
						jsonPath.get("target", "").asString(),
					);
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -212,6 +213,40 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+	{
+		bool typeFailures = false;
+		if (jsonRemappings.isArray())
+		{
+			vector<frontend::ImportRemapper::Remapping> remappings;
+			for (Json::Value const& jsonPath: jsonRemappings)
+			{
+				if (
+					jsonPath.isObject() &&
+					jsonPath.get("context", "").isString() &&
```
nikola-matic, line 229: New type checking helpers were added and recently merged in [this pull request](https://github.com/ethereum/solidity/pull/13579/files#diff-56f32f3c6823f15db6b0437642c7639a500cf3af3d7a86e16db2073dffca60ceR79), so these should be changed to use `isOfType<T>` or `isOfTypeIfExists<T>`, whichever one is more appropriate.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -212,6 +213,40 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+					typeFailures = true;
+			}
+			m_fileRepository.setRemappings(std::move(remappings));
+		}
+		else
+			typeFailures = true;
+
+		if (typeFailures)
+			m_client.trace("Invalid JSON configuration passed. \"remappings\" should be of form [{\"context\":\"\", \"prefix\":\"foo\", \"target\":\"bar\"}].");
+
```
nikola-matic, line 248:
```suggestion
```

libsolidity/lsp/LanguageServer.cpp
nikola-matic: By the way, these `get`s can also be replaced with `getOrDefault<string>(jsonPath["prefix"], "")`.

libsolidity/lsp/LanguageServer.cpp
cameel: I'd hold off a bit with using these helpers everywhere. I wanted to get them merged not to stall the asm import PR but in the end we may and up removing them (https://github.com/ethereum/solidity/pull/13579#issuecomment-1286803351). Depends on what we end up with after #13579. I mean, it's fine to use them but for now I'd also not force that.

NunoFilipeSantos:
Right now, we're choosing to close this PR as we won't allocate any time to this particular area.

PR #13424, opened 2022-08-22
Update emscripten Dockerfile to 3.1.19

Marenz:


scripts/ci/build_emscripten.sh
```diff
@@ -61,12 +59,20 @@ then
 	echo -n "$CIRCLE_SHA1" >commit_hash.txt
 fi
 
+# Disable warnings for unqualified "move()" calls, introduced and enabled by
+# default in clang-16 which is what the emscripten docker image uses.
+# Additionally, disable the warning for unknown warnings here, as this script is
+# also used with earlier clang versions.
+CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call -Wno-unknown-warning-option"
```
Marenz, line 69: I am not super happy about the `-Wno-unknown-warnings-option`-option as it might hide problems we don't see otherwise.

However, as it is only enabled for this particular script, it seemed okay to me.
Alternative approaches:

* Check if the current clang used supports the flag and only then use (and disable) it
  * Requires an extra file (empty `main()`) or other method to do that check
* Update all our `move()` calls to `std::move()` and if possible add this or a similar option in our normal compile-pipeline so that we don't accidentally add a `move()` without std
  * Not sure if such a flag exists before clang-16 and for gcc at all. If it doesn't we risk breaking b_ems .. which the CI would notice though...
* Update the image used by b_ems to use the latest clang as this seems to be the only other automated use of this script

I don't really know where the image for b_ems comes from (in the config it just says `- image: << pipeline.parameters.emscripten-docker-image >>`) so my last two points might not apply, depending how all those images are setup...

scripts/docker/buildpack-deps/Dockerfile.emscripten
```diff
@@ -32,11 +32,14 @@
...
-FROM emscripten/emsdk:2.0.33 AS base
-LABEL version="11"
+FROM emscripten/emsdk:3.1.19 AS base
+LABEL version="12"
 
 ADD emscripten.jam /usr/src
 RUN set -ex; \
+\
+	apt-get update && apt-get install lz4; \
+\
```
Marenz: Docker already warned about the "empty continuation" lines, so I fixed that as well while I was already touching the file..

scripts/ci/build_emscripten.sh
cameel: > I don't really know where the image for b_ems comes from

It's built from [`scripts/docker/buildpack-deps/Dockerfile.emscripten`](https://github.com/ethereum/solidity/blob/develop/scripts/docker/buildpack-deps/Dockerfile.emscripten), which currently extends `emscripten/emsdk:2.0.33`.

> Update the image used by b_ems to use the latest clang as this seems to be the only other automated use of this script

I guess we could add a clang update step to the dockerfile. But best if we don't have to. The distro it's based on might not have the package and we'd have to build it from source or find some release binaries and install them manually.

> Update all our `move()` calls to `std::move()` and if possible add this or a similar option in our normal compile-pipeline so that we don't accidentally add a `move()` without std 

This is probably the best long-term option. But we don't have to deal with right now. I'd just add a `good first issue` to do this eventually and for now having these extra flags seems fine as long as they're clearly commented as a temporary workaround that we need to get rid of.

scripts/docker/buildpack-deps/Dockerfile.emscripten
cameel: I'd add an indent here:
```suggestion
    \
```

Also, all these lines should really end with `&&` so that any error makes the whole set fail. Otherwise the return code depends only on the last command, which is `rm -r /usr/src/boost_1_75_0` here :)

scripts/docker/buildpack-deps/README.md
```diff
@@ -18,18 +18,20 @@ located in `develop`.
 ### Build, Test & Push
 
 Note that the whole workflow - including all defined strategies (image variants) - will be triggered,
-even if only a single Dockerfile was change. The full workflow will only gets executed, if the corresponding
+even if only a single Dockerfile was change. The full workflow will only get executed, if the corresponding
```
cameel:
```suggestion
even if only a single Dockerfile was changed. The full workflow will only get executed, if the corresponding
```

scripts/ci/build_emscripten.sh
Marenz: > It's built from scripts/docker/buildpack-deps/Dockerfile.emscripten, which currently extends emscripten/emsdk:2.0.33.

Which this PR modifies and updates... the whole reason for that flag is because this image does use the latest clang..
I only noticed that the b_ems failed earlier so I assumed it must be a different image as the buildpack stuff worked.
I suppose it simply didn't update the image yet at that point.
If that's the case we can just remove the `-Wno-unknown-warning-option` after the image is updated.

> I guess we could add a clang update step to the dockerfile.

Not required at all, the ems people use the master branch from clang in their image ;)

Marenz: See the last comment at https://github.com/emscripten-core/emscripten/pull/16386#issuecomment-1224298275

cameel: > I only noticed that the b_ems failed earlier so I assumed it must be a different image as the buildpack stuff worked.
> I suppose it simply didn't update the image yet at that point.

Yeah. We still have the old hash in `config.yml`. But thanks for pointing that out. I would completely forget that we also need to update the hash the in a separate PR - can you create such a PR?

cameel: > If that's the case we can just remove the `-Wno-unknown-warning-option` after the image is updated.

So can we remove it?

scripts/docker/buildpack-deps/Dockerfile.emscripten
```diff
@@ -32,16 +32,19 @@
...
 
 ADD emscripten.jam /usr/src
-RUN set -ex; \
-	cd /usr/src; \
-	git clone https://github.com/Z3Prover/z3.git -b z3-4.8.17 --depth 1 ; \
-	cd z3; \
-	mkdir build; \
-	cd build; \
+RUN set -ex && \
+\
```
cameel:
```suggestion
    \
```

scripts/docker/buildpack-deps/Dockerfile.emscripten
```diff
@@ -32,16 +32,19 @@
...
+	apt-get update && apt-get install lz4 && \
```
cameel:
```suggestion
	apt-get update && \
	apt-get install lz4 && \
```

cameel: I'd also try to do this:
```suggestion
	apt-get update && apt-get install lz4 --no-install-recommends && \
```
Less stuff gets installed that way and in images this stuff will likely not be used anyway.

scripts/docker/buildpack-deps/Dockerfile.emscripten
```diff
@@ -51,21 +54,21 @@ RUN set -ex; \
...
-	wget -q 'https://boostorg.jfrog.io/artifactory/main/release/1.75.0/source/boost_1_75_0.tar.bz2' -O boost.tar.bz2; \
-	test "$(sha256sum boost.tar.bz2)" = "953db31e016db7bb207f11432bef7df100516eeb746843fa0486a222e3fd49cb  boost.tar.bz2"; \
-	tar -xf boost.tar.bz2; \
-	rm boost.tar.bz2; \
-	cd boost_1_75_0; \
-	mv ../emscripten.jam .; \
-	./bootstrap.sh; \
-	echo "using emscripten : : em++ ;" >> project-config.jam ; \
+		.. && \
+	make && make install && \
```
cameel:
```suggestion
	make && \
	make install && \
```
Actually `make install` alone would probably be enough but if this works, let's keep it.

scripts/ci/build_emscripten.sh
cameel: Actually, since after update we'll get latest clang, does that mean we can remove both?

If not, there should be a TODO comment saying when we'll be able to remove the hack.

Marenz: Technically yes, but only after this PR updated the image.
I will just create another PR that will remove it right after this is merged

Marenz: Wait, both? Why both? `unqualified-std-cast-call` is enabled by default in the latest clang, that's where the whole issue came from

cameel: Ah, I assumed that we might not need it because emscripten itself adds `-Wno-unqualified-std-cast-call` (https://github.com/emscripten-core/emscripten/pull/16386). Just wasn't sure. So that applies only to their own code and we do need it too after all?

Marenz: Correct :)

scripts/ci/build_emscripten.sh
```diff
@@ -40,7 +40,7 @@ else
 	BUILD_DIR="$1"
 fi
 
-apt-get update && apt-get install lz4
+apt-get update && apt-get install lz4 --no-install-recommends
```
cameel: We got `set -e` so no need for `&&`:
```suggestion
apt-get update
apt-get install lz4 --no-install-recommends
```

scripts/ci/build_emscripten.sh
```diff
@@ -61,12 +62,20 @@ then
 	echo -n "$CIRCLE_SHA1" >commit_hash.txt
 fi
 
+# Disable warnings for unqualified "move()" calls, introduced and enabled by
+# default in clang-16 which is what the emscripten docker image uses.
+# Additionally, disable the warning for unknown warnings here, as this script is
+# also used with earlier clang versions.
+CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call -Wno-unknown-warning-option"
```
cameel, line 69: Can you add the TODO comment I mentioned earlier? I really think we need something that clearly indicates when the hack can be removed.

cameel (approved):
The only remaining problem is a missing TODO note so I'm approving but it would be best to have it IMO.

scripts/ci/build_emscripten.sh
Marenz: The hack can be removed literally right after this PR and I will do that once it's merged <3 so I don't think we need the comment

PR #13428, opened 2022-08-23
Add std:: qualifier to move() calls

Marenz:
Starting with clang-16, ``unqualified-std-cast-call`` will be enabled by default, which will warn for every unqualified ``move()`` call.

I wrote this code while trying to get the ems dockerimage update to work but ultimately decided for a different approach, namely adding the ``-Wno-unqualified-std-cast-call`` flag.

If we don't merge this PR we can only use clang-16 (and any later) with that flag.

So, if we want that, feel free to merge, if not, close.

### DIFF
--- a/libsolidity/analysis/NameAndTypeResolver.cpp
+++ b/libsolidity/analysis/NameAndTypeResolver.cpp
@@ -184,7 +184,7 @@ vector<Declaration const*> NameAndTypeResolver::nameFromCurrentScope(ASTString c
 	ResolvingSettings settings;
 	settings.recursive = true;
 	settings.alsoInvisible = _includeInvisibles;
-	return m_currentScope->resolveName(_name, move(settings));
+	return m_currentScope->resolveName(_name, std::move(settings));
 }
 
 Declaration const* NameAndTypeResolver::pathFromCurrentScope(vector<ASTString> const& _path) const
@@ -204,7 +204,7 @@ std::vector<Declaration const*> NameAndTypeResolver::pathFromCurrentScopeWithAll
 	settings.recursive = true;
 	settings.alsoInvisible = false;
 	settings.onlyVisibleAsUnqualifiedNames = true;
-	vector<Declaration const*> candidates = m_currentScope->resolveName(_path.front(), move(settings));
+	vector<Declaration const*> candidates = m_currentScope->resolveName(_path.front(), std::move(settings));
 
 	for (size_t i = 1; i < _path.size() && candidates.size() == 1; i++)
 	{
--- a/libsolidity/codegen/ir/IRGenerator.cpp
+++ b/libsolidity/codegen/ir/IRGenerator.cpp
@@ -113,7 +113,7 @@ pair<string, string> IRGenerator::run(
 	}
 	asmStack.optimize();
 
-	return {move(ir), asmStack.print(m_context.soliditySourceProvider())};
+	return {std::move(ir), asmStack.print(m_context.soliditySourceProvider())};
 }
 
 string IRGenerator::generate(
@@ -214,7 +214,7 @@ string IRGenerator::generate(
 	// NOTE: Function pointers can be passed from creation code via storage variables. We need to
 	// get all the functions they could point to into the dispatch functions even if they're never
 	// referenced by name in the deployed code.
-	m_context.initializeInternalDispatch(move(internalDispatchMap));
+	m_context.initializeInternalDispatch(std::move(internalDispatchMap));
 
 	// Do not register immutables to avoid assignment.
 	t("DeployedObject", IRNames::deployedObject(_contract));
@@ -236,8 +236,8 @@ string IRGenerator::generate(
 
 	solAssert(_contract.annotation().creationCallGraph->get() != nullptr, "");
 	solAssert(_contract.annotation().deployedCallGraph->get() != nullptr, "");
-	verifyCallGraph(collectReachableCallables(**_contract.annotation().creationCallGraph), move(creationFunctionList));
-	verifyCallGraph(collectReachableCallables(**_contract.annotation().deployedCallGraph), move(deployedFunctionList));
+	verifyCallGraph(collectReachableCallables(**_contract.annotation().creationCallGraph), std::move(creationFunctionList));
+	verifyCallGraph(collectReachableCallables(**_contract.annotation().deployedCallGraph), std::move(deployedFunctionList));
 
 	return t.render();
 }
@@ -317,7 +317,7 @@ InternalDispatchMap IRGenerator::generateInternalDispatchFunctions(ContractDefin
 				});
 			}
 
-			templ("cases", move(cases));
+			templ("cases", std::move(cases));
 			return templ.render();
 		});
 	}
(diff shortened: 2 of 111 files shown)


libsolidity/analysis/NameAndTypeResolver.cpp
```diff
@@ -450,7 +450,7 @@ vector<T const*> NameAndTypeResolver::cThreeMerge(list<list<T const*>>& _toMerge
 	{
 		for (auto it = _toMerge.begin(); it != _toMerge.end();)
 		{
-			it->remove(_candidate);
+			it->restd::move(_candidate);
```
cameel: Looks like your search&replace regex was a bit too weak :)
```suggestion
			it->remove(_candidate);
```

libsolidity/formal/SymbolicState.cpp
```diff
@@ -114,13 +114,13 @@ void SymbolicState::transfer(smtutil::Expression _from, smtutil::Expression _to,
...
 	addBalance(_from, 0 - _value);
-	addBalance(_to, move(_value));
+	addBalance(_to, std::move(_value));
 	unsigned indexAfter = m_state.index();
 	solAssert(indexAfter > indexBefore, "");
 	m_state.newVar();
 	/// Do not apply the transfer operation if _from == _to.
 	auto newState = smtutil::Expression::ite(
-		move(_from) == move(_to),
+		std::move(_from) == std::move(_to),
```
cameel, line 123: This is a weird use of `move()`. It would make sense if `==` was defined to take arguments by value rather than by reference - but why would one ever do that for `==`?

libyul/backends/evm/EVMDialect.cpp
```diff
@@ -82,7 +82,7 @@ pair<YulString, BuiltinFunctionForEVM> createEVMFunction(
 		_assembly.appendInstruction(_instruction);
 	};
 
-	return {f.name, move(f)};
+	return {f.name, std::move(f)};
```
cameel: This whole expression looks like it depends on the order of evaluation of arguments, which is undefined in C++.

scripts/ci/build_emscripten.sh
```diff
@@ -63,6 +63,7 @@ fi
 
 mkdir -p "$BUILD_DIR"
 cd "$BUILD_DIR"
+echo $BUILD_DIR
```
cameel: Looks unrelated to the PR.

libsolidity/formal/SymbolicState.cpp
Marenz: It seems it's because the parameters are moved into the returned object..
```
	friend Expression operator==(Expression _a, Expression _b)
	{
		smtAssert(_a.sort->kind == _b.sort->kind, "Trying to create an 'equal' expression with different sorts");
		return Expression("=", std::move(_a), std::move(_b), Kind::Bool);
	}
```

cameel: Hmm... interesting. I guess it's good then.

Marenz: I agree that it seems an unexpected way to use `operator==`

libsolidity/codegen/ir/IRGenerator.cpp
```diff
@@ -236,8 +236,8 @@ string IRGenerator::generate(
 
 	solAssert(_contract.annotation().creationCallGraph->get() != nullptr, "");
 	solAssert(_contract.annotation().deployedCallGraph->get() != nullptr, "");
-	verifyCallGraph(collectReachableCallables(**_contract.annotation().creationCallGraph), move(creationFunctionList));
-	verifyCallGraph(collectReachableCallables(**_contract.annotation().deployedCallGraph), move(deployedFunctionList));
+	verifyCallGraph(collectReachableCallables(**_contract.annotation().creationCallGraph), std::move(creationFunctionList));
+	verifyCallGraph(collectReachableCallables(**_contract.annotation().deployedCallGraph), std::move(deployedFunctionList));
```
nishant-sachdeva, line 240:
```suggestion
	verifyCallGraph(
	     collectReachableCallables(**_contract.annotation().creationCallGraph),
	     std::move(creationFunctionList)
	);
	verifyCallGraph(
	     collectReachableCallables(**_contract.annotation().deployedCallGraph), 
             std::move(deployedFunctionList)
        );
```
Just a coding style change.

libsolidity/formal/SMTEncoder.cpp
```diff
@@ -2991,7 +2991,7 @@ set<FunctionDefinition const*, ASTNode::CompareByID> const& SMTEncoder::contract
 			for (auto const* baseFun: base->definedFunctions())
 				allFunctions.insert(baseFun);
 
-		m_contractFunctionsWithoutVirtual.emplace(&_contract, move(allFunctions));
+		m_contractFunctionsWithoutVirtual.emplace(&_contract, std::move(allFunctions));
```
nishant-sachdeva, line 2994:
```suggestion
		m_contractFunctionsWithoutVirtual.emplace(
		     &_contract, 
		     std::move(allFunctions)
		);
```

libsolidity/formal/SymbolicState.cpp
```diff
@@ -94,12 +94,12 @@ smtutil::Expression SymbolicState::balance() const
...
 smtutil::Expression SymbolicState::balance(smtutil::Expression _address) const
 {
-	return smtutil::Expression::select(balances(), move(_address));
+	return smtutil::Expression::select(balances(), std::move(_address));
 }
 
 smtutil::Expression SymbolicState::blockhash(smtutil::Expression _blockNumber) const
 {
-	return smtutil::Expression::select(m_tx.member("blockhash"), move(_blockNumber));
+	return smtutil::Expression::select(m_tx.member("blockhash"), std::move(_blockNumber));
```
nishant-sachdeva, line 102:
```suggestion
	return smtutil::Expression::select(
	     m_tx.member("blockhash"), 
	     std::move(_blockNumber)
	);
```

nishant-sachdeva (commented):
Apart from the comments raised by @cameel , and a few minor coding style changes, LGTM

leonardoalt:
The style changes are independent from this PR, I think they should be handled separately.

leonardoalt:
My only question here is: why?

Marenz:
> My only question here is: why?

See PR description...

leonardoalt:
> > My only question here is: why?
> 
> See PR description...

:grimacing: thanks!

libyul/backends/evm/EVMDialect.cpp
Marenz: https://github.com/ethereum/solidity/pull/13449

leonardoalt (approved):

libsolidity/formal/SymbolicState.cpp
nikola-matic: I'd argue that we shouldn't change the formatting (at least not in this PR), especially if our `check_style` script doesn't complain. I'd much rather prefer to wait until `clang-format` starts supporting everything we want (which should be soon), and then we can just use that to format our code. In addition, it should significantly simplify our `check_style` script, as it should simply do a comparison between current file and `clang-format-X current-file`, and if there's a mismatch, fail the step.

nikola-matic (approved):

Issue #13421, opened 2022-08-22
Fix warning about `DOWNLOAD_EXTRACT_TIMESTAMP` on CMake 3.24


cameel:
I have just updated to CMake 3.24 and now I see following warning when running `cmake` (and sometimes `make`):
```
CMake Warning (dev) at /usr/share/cmake/Modules/ExternalProject.cmake:3071 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  /usr/share/cmake/Modules/ExternalProject.cmake:4167 (_ep_add_download_command)
  cmake/jsoncpp.cmake:43 (ExternalProject_Add)
  CMakeLists.txt:46 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Marenz:
Fixed in #13429

PR #13429, opened 2022-08-24
Fix warning about DOWNLOAD_EXTRACT_TIMESTAMP on CMake 3.24

Sotatek-ThanhTran:
This pull request is made to fix the issue https://github.com/ethereum/solidity/issues/13421

From CMake 3.24, the CMP0135 policy behavior should be specified when using ExternalProject_Add function.

This is the CMP0135
documentation: https://cmake.org/cmake/help/latest/policy/CMP0135.html#policy:CMP0135
> _New in version 3.24._
> 
> When using the URL download method with the [ExternalProject_Add()](https://cmake.org/cmake/help/latest/module/ExternalProject.html#command:externalproject_add) command, CMake 3.23 and below sets the timestamps of the extracted contents to the same as the timestamps in the archive. When the URL changes, the new archive is downloaded and extracted, but the timestamps of the extracted contents might not be newer than the previous contents. Anything that depends on the extracted contents might not be rebuilt, even though the contents may change.
> 
> CMake 3.24 and above prefers to set the timestamps of all extracted contents to the time of the extraction. This ensures that anything that depends on the extracted contents will be rebuilt whenever the URL changes.
> 
> The DOWNLOAD_EXTRACT_TIMESTAMP option to the [ExternalProject_Add()](https://cmake.org/cmake/help/latest/module/ExternalProject.html#command:externalproject_add) command can be used to explicitly specify how timestamps should be handled. When DOWNLOAD_EXTRACT_TIMESTAMP is not given, this policy controls the default behavior. The OLD behavior for this policy is to restore the timestamps from the archive. The NEW behavior sets the timestamps of extracted contents to the time of extraction.
> 
> This policy was introduced in CMake version 3.24. CMake version 3.24.1 warns when the policy is not set and uses OLD behavior. Use the [cmake_policy()](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) command to set it to OLD or NEW explicitly.

I think we should use NEW behavior for this policy to make sure the dependencies will be rebuilt whenever the URL changes.

### DIFF
--- a/cmake/EthPolicy.cmake
+++ b/cmake/EthPolicy.cmake
@@ -20,4 +20,9 @@ macro (eth_policy)
 		# Allow selecting MSVC runtime library using CMAKE_MSVC_RUNTIME_LIBRARY.
 		cmake_policy(SET CMP0091 NEW)
 	endif()
+
+	# Avoid warning about DOWNLOAD_EXTRACT_TIMESTAMP in CMake 3.24:
+	if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
+		cmake_policy(SET CMP0135 NEW)
+	endif()
 endmacro()


Sotatek-ThanhTran:
@cameel Please review this pull request.
I've tested this pull request with cmake 3.16.3, 3.23.3, and 3.24.1.
The test t_native_test_ext_gnosis has failed on CircleCI but I don't think it relates to this pull request.

matheusaaguiar (dismissed):
Confirmed that this prevents the warning message to be printed in the output when using CMake 3.24.1.

cmake/EthPolicy.cmake
```diff
@@ -20,4 +20,9 @@ macro (eth_policy)
 		# Allow selecting MSVC runtime library using CMAKE_MSVC_RUNTIME_LIBRARY.
 		cmake_policy(SET CMP0091 NEW)
 	endif()
+
+	# Avoid warnings in CMake 3.24.0:
```
cameel: Well, that's obvious :) It would be more useful to say which warning this is preventing.

Sotatek-ThanhTran: Sorry, I've force-pushed this pull request, but the code is not changed, just add information about the warning to the comment.

Marenz: I'd say if we change this to `Prevent warning message related to changed default behavior of DOWNLOAD_EXTRACT_TIMESTAMP` it's good to merge

Marenz (approved):

sgehrman:
just got this today: 

```
CMake Warning (dev) at C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.24/Modules/ExternalProject.cmake:3071 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
  C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.24/Modules/ExternalProject.cmake:4167 (_ep_add_download_command)
  CMakeLists.txt:9 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

ahmeedev:
How to solve this warning?

Nickrader:
> How to solve this warning?

In this case, he sets CMP0135 using cmake_policy. [click this page top, to the right where it says "Files Changed"]  Since this is a commit, you can see what he did.

Cmake Documentaion:
https://cmake.org/cmake/help/latest/policy/CMP0135.html

This talks about it but cause could be many things according to poster, gives example to suppress.
https://discourse.cmake.org/t/would-like-to-disable-cmake-dev-warning/2450

*Disclaimer: I have no problem here, this is where Google sent me when I copy/pasted error from unrelated project.  Putting it here for everyone else google sends.

My case it was caused by FetchContent (the warning said), I was using outdated Googletest instructions:
 https://google.github.io/googletest/quickstart-cmake.html

Warning went away when I used latest commit sha-1 for GIT_TAG, and did it this style [instead of having URL link]
https://cmake.org/cmake/help/latest/module/FetchContent.html#typical-case

MujibBasha:
> How to solve this warning?

how you solve it

Issue #13051, opened 2022-05-23
Remove workarounds for ethers 5.6.2+ and `getStorageAt()` from external tests


cameel:
https://github.com/ethers-io/ethers.js/discussions/2849 and https://github.com/NomicFoundation/hardhat/issues/2709 are now fixed. We added workarounds in #13027 and #13020. Now we should be able to revert them.

matheusaaguiar:
There are still 3 gnosis test cases failing in CI in every preset [1071808](https://app.circleci.com/pipelines/github/ethereum/solidity/24371/workflows/b68233fd-de80-4756-a47c-20a08ef21787/jobs/1071808). I was able to reproduce it locally, though I had to force hardhat version to 2.9.6 and apply the chainId workaround as discussed in [euler-xyz/euler-contracts#119](https://github.com/euler-xyz/euler-contracts/issues/119#issuecomment-1133367182). I guess it still has to do with the problem in [ethers-io/ethers.js#2849](https://github.com/ethers-io/ethers.js/discussions/2849).

cameel:
Does Uniswap pass? It failed for me when I tried removing the workaround in #13063.

matheusaaguiar:
> Does Uniswap pass? It failed for me when I tried removing the workaround in #13063.

I will test.

matheusaaguiar:
Uniswap fails if workaround is removed.

PR #13430, opened 2022-08-24
Fix inconsistent nested dependency in safe-contracts

nikola-matic:
The error causing CI failures is the following:
```
he types returned by 'provider.getFeeData()' are incompatible between these types.
      Type 'Promise<import("/tmp/ext-test-Gnosis-Safe-hbDXur/ext/node_modules/ethers/node_modules/@ethersproject/abstract-provider/lib/index").FeeData>' is not assignable to type 'Promise<import("/tmp/ext-test-Gnosis-Safe-hbDXur/ext/node_modules/@ethersproject/abstract-provider/lib/index").FeeData>'.
        Property 'lastBaseFeePerGas' is missing in type 'import("/tmp/ext-test-Gnosis-Safe-hbDXur/ext/node_modules/ethers/node_modules/@ethersproject/abstract-provider/lib/index").FeeData' but required in type 'import("/tmp/ext-test-Gnosis-Safe-hbDXur/ext/node_modules/@ethersproject/abstract-provider/lib/index").FeeData'.
```
The issue is that `@ethersproject/abstract-provider` has recently introduces a new member of `FeeData`, which unfortunately causes incompatibilities as we end up having different version of said dependency.

Unfortunately, the regular approach where we would peg the version by explicitly installing the specific version of the package does not work, as the package in question is also a child dependency of another dependency. I have therefore manually specified the overrides (feature available from `npm v8.3+`) for packages in question by altering the project's (`safe-contracts`) `package.json` once the project is cloned, but before we run `npm install`.

I don't know whether this approach is appropriate, nor whether the appropriate version of `npm` is available in our CI, but it fixed the tests for me locally.

The `jq` command will add the following section to the project's `package.json`:
```json
  "overrides": {
    "@ethersproject/abstract-provider": "5.6.0",
    "@ethersproject/abstract-signer@5.6.0": {
      "@ethersproject/abstract-provider": "5.6.0"
    }
  }
  ```

### DIFF
--- a/test/externalTests/gnosis.sh
+++ b/test/externalTests/gnosis.sh
@@ -31,8 +31,8 @@ BINARY_TYPE="$1"
 BINARY_PATH="$(realpath "$2")"
 SELECTED_PRESETS="$3"
 
-function compile_fn { npm run build; }
-function test_fn { npm test; }
+function compile_fn { npx npm run build; }
+function test_fn { npx npm test; }
 
 function gnosis_safe_test
 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nikola-matic** [comment]
```
ci/circleci: t_native_test_ext_gnosis — Your tests failed on CircleCI 
```
:sob:

File: test/externalTests/gnosis.sh [hunk 1]
```diff
@@ -87,6 +87,10 @@ function gnosis_safe_test
     sed -i "s|it\(('can only be called from Safe itself'\)|it.skip\1|g" test/libraries/Migration.spec.ts
     sed -i "s|it\(('should enforce delegatecall to MultiSend'\)|it.skip\1|g" test/libraries/MultiSend.spec.ts
 
+    jq '.overrides."@ethersproject/abstract-provider"="5.6.0" |
+        .overrides."@ethersproject/abstract-signer@5.6.0"
+                  ."@ethersproject/abstract-provider"="5.6.0"' package.json > package.json.tmp && mv package.json.tmp package.json
```
cameel: We're running with `set -e` so you don't really need the `&&` here. You can put it on a separate line.
```suggestion
    jq '.overrides."@ethersproject/abstract-provider"="5.6.0" |
        .overrides."@ethersproject/abstract-signer@5.6.0"."@ethersproject/abstract-provider"="5.6.0"' package.json > package.json.tmp
    mv package.json.tmp package.json
```
```

File: test/externalTests/gnosis.sh [hunk 2]
```diff
@@ -87,6 +87,10 @@ function gnosis_safe_test
     sed -i "s|it\(('can only be called from Safe itself'\)|it.skip\1|g" test/libraries/Migration.spec.ts
     sed -i "s|it\(('should enforce delegatecall to MultiSend'\)|it.skip\1|g" test/libraries/MultiSend.spec.ts
 
+    jq '.overrides."@ethersproject/abstract-provider"="5.6.0" |
```
cameel, line 94: Please add a comment explaining why we are doing this and when we can remove it.

cameel (commented):
The overall approach seems fine. Not much worse than other hacks we already have here.

So we can merge it if you can get it to work.

test/externalTests/gnosis.sh
nikola-matic: I would be fine with adding a comment as to why we're doing it, but realistically, is there any point in opening yet another issue that will not be addressed?

test/externalTests/gnosis.sh
```diff
@@ -87,33 +87,39 @@ function gnosis_safe_test
...
+                  ."@ethersproject/abstract-provider"="5.6.0"' package.json > package.json.tmp
+    mv package.json.tmp package.json
+
     neutralize_package_lock
     neutralize_package_json_hooks
     force_hardhat_compiler_binary "$config_file" "$BINARY_TYPE" "$BINARY_PATH"
     force_hardhat_compiler_settings "$config_file" "$(first_word "$SELECTED_PRESETS")" "$config_var"
-    npm install
-    npm install hardhat-gas-reporter
+    npm install npm@8.11.0
```
cameel: We don't need a specific version. We want the latest one.
```suggestion
    # We need the latest npm version for the `overrides` feature in package.json
    npm install npm
```

Or, if you're worried about the `overrides` feature:
```suggestion
    # We need an npm version that supports the `overrides` key in package.json
    npm install npm@>=8.3.0
```

test/externalTests/gnosis.sh
cameel: It does not look like an upstream issue to me so yeah, you should not report it. I think we only have this problem because we pinned version for `ethers`.

But please do add the comment. Each hack should be clearly commented to encourage people to try to remove it.

cameel: Actually, maybe you could try to remove the `ethers` hack and some of the others we have here? The original problem that made us pin `ethers` at 5.6.1 has been resolved several versions ago so we should have been able to remove it but there were some problems with it when @matheusaaguiar tried it last time. See #13051 and https://github.com/ethers-io/ethers.js/discussions/2849#discussioncomment-2812920 for some context.

If our current problem is caused by the old workarounds, removing them might help. Not having to pin version should also make it less likely that we'll hit another incompatibility soon.

nikola-matic: Unfortunately, removing any of the old hacks causes failures, so it looks like they'll have to stay for the time being.

nikola-matic:
@cameel, should be ready for merge, take a look when you can.

test/externalTests/gnosis.sh
cameel: But what kind of failures? Maybe it's something that has an easy solution?

cameel (approved):

test/externalTests/gnosis.sh
nikola-matic: Unchanged, i.e. the same issue with a mass of reverts expectations not being met (issue references in `TODO`).

Issue #12904, opened 2022-04-06
Improve building Solidity on Windows


christianparpart:
* While I personally think `vcpkg` should really be at least supported, I also struggled in getting `soltest.exe` linked against vcpkg's provided boost as it kept trying to link using the boost DLLs rather than static libraries.
* I was running `install_deps.ps1` from project root and have received error messages that come from the fact that  I wasn't executing the script one level deeper (d.g. ./build/ directory).
* Additional cmake flags are required, even when used in conjunction with the install-deps.ps1 provided dependencies. That should not be the case. Custom flags need to be passed to cmake are: `-DBoost_USE_STATIC_RUNTIME=ON -DBoost_DIR=${projectDir}\\deps\\boost\\lib\\cmake\\Boost-1.77.0 -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`
* The install-deps script could probably be made a little more fool-proof / robust, i.e. it should allow to be invoked more than once and not bail out with weird error messages.

ekpyron:
Furthermore:
- we should more gracefully handle privilege errors in the symlink tests.
- we may consider making the non-release windows build a debug build (that would have caught https://github.com/ethereum/solidity/pull/12902)
- we need to fix VS2022 builds (https://github.com/ethereum/solidity/pull/12802 contains a CI setup for that, but is still lacking the actual range-v3-related fix)
- Also (although this may not strictly related to windows builds), ideally, linking soltest to dynamic boost should also be possible without issues.

christianparpart:
I forgot what I did to work around the following error when executing `./scripts/install_deps.ps1`:

```
...
Building Boost.Build engine
LOCALAPPDATA=C:\Users\chris\AppData\Local
Found with vswhere C:\Program Files\Microsoft Visual Studio\2022\Community
###
### "Unknown toolset: vcunk"
...
```

which I am facing again, but probably also worth fixing for the common user.

nikola-matic:
> I forgot what I did to work around the following error when executing `./scripts/install_deps.ps1`:
> 
> ```
> ...
> Building Boost.Build engine
> LOCALAPPDATA=C:\Users\chris\AppData\Local
> Found with vswhere C:\Program Files\Microsoft Visual Studio\2022\Community
> ###
> ### "Unknown toolset: vcunk"
> ...
> ```
> 
> which I am facing again, but probably also worth fixing for the common user.

[...]

PR #13434, opened 2022-08-25
Bump cmake and boost versions to work with Visual Studio 2022

nikola-matic:
In an effort to address issues in #12904, `cmake` and `boost` versions are being bumped in order to support Visual Studio 2022. This should have no effect on users wishing to remain with pre-2022 versions of Visual Studio. 
`install_deps.ps1` script will now also clean up after itself by deleting any downloaded and extracted tarballs/directories.

### DIFF
--- a/scripts/install_deps.ps1
+++ b/scripts/install_deps.ps1
@@ -6,23 +6,26 @@ $progressPreference = "silentlyContinue"
 if ( -not (Test-Path "$PSScriptRoot\..\deps\boost") ) {
   New-Item -ItemType Directory -Force -Path "$PSScriptRoot\..\deps"
 
-  Invoke-WebRequest -URI "https://github.com/Kitware/CMake/releases/download/v3.18.2/cmake-3.18.2-win64-x64.zip" -OutFile cmake.zip
-  if ((Get-FileHash cmake.zip).Hash -ne "5f4ec834fbd9b62fbf73bc48ed459fa2ea6a86c403106c90fedc2ac76d51612d") {
+  Invoke-WebRequest -URI "https://github.com/Kitware/CMake/releases/download/v3.21.3/cmake-3.21.3-windows-x86_64.zip" -OutFile cmake.zip
+  if ((Get-FileHash cmake.zip).Hash -ne "e28178103fad901a99fb201fac04e4787d4cd4f664c5db88998c57eed68843b9") {
     throw 'Downloaded CMake source package has wrong checksum.'
   }
   tar -xf cmake.zip
-  mv cmake-3.18.2-win64-x64 "$PSScriptRoot\..\deps\cmake"
+  mv cmake-3.21.3-windows-x86_64 "$PSScriptRoot\..\deps\cmake"
+  Remove-Item cmake.zip
 
   # FIXME: The default user agent results in Artifactory treating Invoke-WebRequest as a browser
   # and serving it a page that requires JavaScript.
-  Invoke-WebRequest -URI "https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/boost_1_77_0.zip" -OutFile boost.zip -UserAgent ""
-  if ((Get-FileHash boost.zip).Hash -ne "d2886ceff60c35fc6dc9120e8faa960c1e9535f2d7ce447469eae9836110ea77") {
+  Invoke-WebRequest -URI "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.zip" -OutFile boost.zip -UserAgent ""
+  if ((Get-FileHash boost.zip).Hash -ne "f22143b5528e081123c3c5ed437e92f648fe69748e95fa6e2bd41484e2986cc3") {
     throw 'Downloaded Boost source package has wrong checksum.'
   }
   tar -xf boost.zip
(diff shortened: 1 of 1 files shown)


cameel:
Path tests are failing: https://github.com/ethereum/solidity/blob/f802eafc679541cc1d3ba0ca5bc7c12b4bdaf939/libsolidity/interface/FileReader.cpp#L332
Something must have changed in the newer version of `boost::filesystem` and stuff gets normalized differently... Let's take a look at the failing tests.

### `allow_path_should_work_with_various_path_forms`
https://github.com/ethereum/solidity/blob/f802eafc679541cc1d3ba0ca5bc7c12b4bdaf939/test/solc/CommandLineInterfaceAllowPaths.cpp#L268

`checkImport()` function here tries to compile the following snippet:
```solidity
// SPDX-License-Identifier: GPL-3.0"
pragma solidity >=0.0;
import '<m_portablePrefix>/a/b/c.sol';
```
by passing it to standard input of the following command:
```bash
solc - --no-color --error-codes --allow-paths "/<m_portablePrefix>"
```
`<m_portablePrefix>` should be replaced with the value calculated here:
https://github.com/ethereum/solidity/blob/f802eafc679541cc1d3ba0ca5bc7c12b4bdaf939/test/solc/CommandLineInterfaceAllowPaths.cpp#L128-L132
it represents the relative path to the temporary `code/` subdir used in the test, with forward slashes regardless of platform.

### `isPathPrefix_unc_path`
https://github.com/ethereum/solidity/blob/f802eafc679541cc1d3ba0ca5bc7c12b4bdaf939/test/libsolidity/interface/FileReader.cpp#L435

This one indicates that the way `//host/a/b/` is normalized must have changed. It used to be considered the normalized form. So I suspect that it might be getting replaced with `/host/a/b/` now? At least on Windows.

cameel:
We need to check if it's happening with latest boost on Linux as well.

In any case, it looks like [`FileReader::normalizeCLIPathForVFS()`](https://github.com/ethereum/solidity/blob/develop/libsolidity/interface/FileReader.cpp#L221-L303) will now have to be updated to handle paths starting with `//` the same way on both older and newer boost versions.

nikola-matic:
> ### `isPathPrefix_unc_path`
> 
> https://github.com/ethereum/solidity/blob/f802eafc679541cc1d3ba0ca5bc7c12b4bdaf939/test/libsolidity/interface/FileReader.cpp#L435
> 
> This one indicates that the way `//host/a/b/` is normalized must have changed. It used to be considered the normalized form. So I suspect that it might be getting replaced with `/host/a/b/` now? At least on Windows.

Uh, debugging this, and it seems a bit weirder than that; this is what I get when I run it through the debugger:

### prefix
```
_prefix.lexically_normal()
{m_pathname=L"\\\\host\\a\\b\\." }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"\\\\host\\a\\b\\."
_prefix
{m_pathname=L"//host/a/b/" }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"//host/a/b/"
```
### path
```
_path.lexically_normal()
{m_pathname=L"\\\\host\\a\\b\\." }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"\\\\host\\a\\b\\."
_path
{m_pathname=L"//host/a/b/" }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"//host/a/b/"
```

The assertion fails because obviously `//host/a/b/` is not equal to `\\\\host\\a\\b\\.`

Also, given the following snippet from the changelog docs for `boost 1.78.0`
```
v4: path::lexically_normal no longer produce a trailing dot (".") element and omits a directory separator after a trailing dot-dot ("..") element in the normalized paths.
```
I would have expected `lexically_normal` to not produce the trailing dot in `\\\\host\\a\\b\\.`

christianparpart:
@nikola-matic I think it might be best to split out the cmake version bump into its own PR as this could be right away merged and then this PR could focus on boost without holding the other part back. What do you think?

nikola-matic:
> @nikola-matic I think it might be best to split out the cmake version bump into its own PR as this could be right away merged and then this PR could focus on boost without holding the other part back. What do you think?
 
It could certainly be done that way, but I don't think it really brings that much benefit unless we bump boost together with it. I'd rather keep this in the same PR.

nikola-matic:
And the behaviour for the working version, i.e. the one using boost 1.77.0 is 
```
_prefix
{m_pathname=L"//host/a/b/" }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"//host/a/b/"
ln_prefix
{m_pathname=L"//host\\a\\b\\." }
    boost::filesystem::path_detail::path_constants<wchar_t,47,92,46>: {...}
    m_pathname: L"//host\\a\\b\\."
```
I've also tried it out on linux by comparing the same snippet with 1.77 and 1.78, and there's no difference there. So clearly a Windows issue.

nikola-matic:
Considering we do have a working configuration (boost 1.77 and VS 2019), I'm closing this PR as we have more pressing tasks.

PR #13436, opened 2022-08-25
Update docker images for emscripten.

Marenz:
Follow-up to #13424.

### DIFF
--- a/scripts/ci/build_emscripten.sh
+++ b/scripts/ci/build_emscripten.sh
@@ -62,11 +62,12 @@ then
 	echo -n "$CIRCLE_SHA1" >commit_hash.txt
 fi
 
-# Disable warnings for unqualified "move()" calls, introduced and enabled by
+# Disable warnings for unqualified `move()` calls, introduced and enabled by
 # default in clang-16 which is what the emscripten docker image uses.
 # Additionally, disable the warning for unknown warnings here, as this script is
 # also used with earlier clang versions.
-CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call -Wno-unknown-warning-option"
+# TODO: This can be removed if and when all usages of `move()` in our codebase use the `std::` qualifier.
+CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call"
 
 
 mkdir -p "$BUILD_DIR"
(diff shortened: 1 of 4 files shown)


scripts/ci/build_emscripten.sh
```diff
@@ -62,11 +62,12 @@ then
 	echo -n "$CIRCLE_SHA1" >commit_hash.txt
 fi
 
-# Disable warnings for unqualified "move()" calls, introduced and enabled by
+# Disable warnings for unqualified `move()` calls, introduced and enabled by
 # default in clang-16 which is what the emscripten docker image uses.
 # Additionally, disable the warning for unknown warnings here, as this script is
 # also used with earlier clang versions.
-CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call -Wno-unknown-warning-option"
+# TODO: This can be removed if and when all usages of `move()` use the `std::` qualifier.
```
cameel:
```suggestion
# TODO: This can be removed if and when all usages of `move()` in our codebase use the `std::` qualifier.
```
Also, let's create an issue for that.

scripts/ci/build_emscripten.sh
```diff
@@ -62,11 +62,12 @@ then
...
+CMAKE_CXX_FLAGS="-Wno-unqualified-std-cast-call"
```
cameel, line 70: This change requires another version bump or it won't get into the image.

I mean, things will still work but it's not great if the source you see does not match the image we're using. This can make for some very confused debugging.

Marenz: But this is not part of the image?

Marenz: To expand on that: This is called by the CI in the environment of the docker image, but it's not part of it.

Marenz: This can also be seen here: https://app.circleci.com/pipelines/github/ethereum/solidity/26062/workflows/a9734583-6920-4572-a62d-35efae37d0f7/jobs/1151467 (step "build", line 52)

cameel: Ah, good. One problem less :)

cameel (approved):

axic:
@cameel this closed #13122, right?

cameel:
Yes. #13386 linked to it but this one (and #13424, which did the actual update) didn't for some reason.

Also, now that you brought it up, I wonder why we went for 3.1.19 and not 3.2.0. @Marenz?

PR #13440, opened 2022-08-26
Meetings on Mondays and Wednesdays now happen at the same time

nishant-sachdeva:
As discussed in the meeting on 24/08/2022, the schedule update for the meetings is to be reflected in the docs

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -30,8 +30,7 @@ Team Calls
 If you have issues or pull requests to discuss, or are interested in hearing what
 the team and contributors are working on, you can join our public team calls:
 
-- Mondays at 3pm CET/CEST.
-- Wednesdays at 2pm CET/CEST.
+- Mondays and Wednesdays at 3pm CET/CEST.
 
 Both calls take place on `Jitsi <https://meet.ethereum.org/solidity>`_.
 


docs/contributing.rst
```diff
@@ -30,8 +30,7 @@ Team Calls
 If you have issues or pull requests to discuss, or are interested in hearing what
 the team and contributors are working on, you can join our public team calls:
 
-- Mondays at 3pm CET/CEST.
-- Wednesdays at 2pm CET/CEST.
+- Mondays and Wednesdays at 3pm CET/CEST.
 
 Both calls take place on `Jitsi <https://meet.ethereum.org/solidity>`_.
```
matheusaaguiar, line 35: I think the link should be updated too, otherwise, people will join an empty room.

docs/contributing.rst
```diff
@@ -30,10 +30,9 @@ Team Calls
 If you have issues or pull requests to discuss, or are interested in hearing what
 the team and contributors are working on, you can join our public team calls:
 
-- Mondays at 3pm CET/CEST.
-- Wednesdays at 2pm CET/CEST.
+- Mondays and Wednesdays at 3pm CET/CEST.
 
-Both calls take place on `Jitsi <https://meet.ethereum.org/solidity>`_.
+Both calls take place on `Jitsi <https://meet.ethdevops.io/solidity>`_.
```
nikola-matic: What have we decided regarding the link though? If I recall, this was only supposed to be a temporary measure until the `meet.ethereum` link was 'fixed'. Maybe we should ask @franzihei ?

franzihei: yes. please do not update this since I am not sure the other instance should be used "publicly" and they do intend to update meet.ethereum.org (hopefully soon)

leonardoalt (approved):

nikola-matic (approved):

PR #13443, opened 2022-08-28
Clarify effect of memory unsafe assembly

### LINKED ISSUE #13416: Clarify effect of memory unsafe assembly
>Therefore, moving stack variables to memory and additional memory optimizations are, by default, disabled in the presence of any inline assembly block that contains a memory operation or assigns to solidity variables in memory.

Can you clarify what "in the presence of" means exactly? Are these optimizations _globally_ disabled for the entire contract if there is a memory unsafe block anywhere in it? Are they only disabled in the scope of the function body?

frangio:
Fixes #13416. Pretty simple change but I think this gets the point across.


leonardoalt (dismissed):

leonardoalt:
Can you please rebase over origin/develop? It should get rid of the CI error

leonardoalt (approved):

PR #13449, opened 2022-08-29
[Tiny] Fix undefined order of evaluation

Marenz:


libyul/backends/evm/EVMDialect.cpp
```diff
@@ -82,7 +82,10 @@ pair<YulString, BuiltinFunctionForEVM> createEVMFunction(
 		_assembly.appendInstruction(_instruction);
 	};
 
-	return {f.name, move(f)};
+	pair<YulString, BuiltinFunctionForEVM> returnValue;
+	returnValue.first = f.name;
+	returnValue.second = std::move(f);
+	return returnValue;
```
leonardoalt: Is there really no better way of doing this? How about
```
auto name = f.name;
return {name, std::move(f)};
```

leonardoalt (approved):

Issue #13408, opened 2022-08-17
First batch of new labels: effort, impact, desirability, task type


cameel:
The first batch of the issue management changes from [Issue labeling guidelines (proposal)](https://notes.ethereum.org/@solidity/SkMG8szAq).

- [x] Add the labels we decided that we want for now:
    - Effort: `epic`, `high effort`, `medium effort`, `low effort`.
    - Impact: `high impact`, `medium impact`, `low impact`.
    - Desirability: `roadmap`, `must have`, `must have eventually`, `should have`, `nice to have`.
    - Task type: `refactor`, `bug`, `documentation`, `research`, `testing`.
- [x] Delete `feature` and `enhancement` labels (remove them from all issues).
- [x] Choose sensible colors. My suggestion would be to have the same base color for each category with a lighter shade for `low` and darker for `high`. Feel free to adjust some of the existing colors if they clash. There are some that are what they are for a good reason (`bug`=red, `codegen error`=black, etc.) but most of them were chosen ad-hoc.
- [x] Make sure that `solc-js` has these labels too. For now I'd not synchronize the ones that are not in one of the categories from the list above, just add the ones that are missing and adjust colors if they're not the same.
- [x] Remove `feature` label from the link in [`ISSUE_TEMPLATE/config.yml`](https://github.com/ethereum/solidity/blob/develop/.github/ISSUE_TEMPLATE/config.yml) (this is the config for [the template chooser](https://github.com/ethereum/solidity/issues/new/choose)).
- [x] Add an entry to [the template chooser](https://github.com/ethereum/solidity/issues/new/choose) for opening a new forum discussion in the right category. We decided not to have the `controversial` label, so I think we need to more strongly encourage people to post their wilder ideas on the forum first and only open feature requests for them after getting some feedback.
- [x]  Consider also tweaking the [`feature_request.md` template](https://github.com/ethereum/solidity/blob/develop/.github/ISSUE_TEMPLATE/feature_request.md) to mention the forum.

cameel:
Also, the "medium" labels should be added to the guidelines. Were the any other changes on Monday that we should document?

nikola-matic:
@cameel, when you say 
```
Remove feature label from the link in [ISSUE_TEMPLATE/config.yml](https://github.com/ethereum/solidity/blob/develop/.github/ISSUE_TEMPLATE/config.yml)
```
[...]

PR #13454, opened 2022-08-30
Update issue selector, and remove auto labeling

nikola-matic:
References #13408

Note: linking directly to URLs that would open a specific discussion doesn't seem possible, so I've only been able to link to the forum itself.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,14 +1,17 @@
 blank_issues_enabled: false
 contact_links:
   - name: Bug Report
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,14 +1,17 @@
 blank_issues_enabled: false
 contact_links:
   - name: Bug Report
-    url: https://github.com/ethereum/solidity/issues/new?template=bug_report.md&projects=ethereum/solidity/43&labels=bug+%3Abug%3A
```
nikola-matic, line 4: As per @cameel 's [comment](https://github.com/ethereum/solidity/issues/13408#issuecomment-1219568910), the labels have been removed as they're not auto-applied anyway, likely due to users not having appropriate permissions.

leonardoalt (approved):

PR #13456, opened 2022-08-30
solc: Adds logo image to --help / --license / --version output.

christianparpart:
Due to a topic we had in the office last week, I did create a very small patch adding a logo the the command line output
for the actions: `--help`, `--license`, `--version`.

image

Mind, while the Sixel standard is about 50 years old (coming from VT340), not every terminal does support that. However, non-supporting terminals are simply ignoring the sequence being sent through stdout to the terminal.

Supporting terminals are: xterm, mlterm, iTerm, wezterm, Contour, konsole (master), gnome-terminal (unreleased, all libvte based terminals will have that eventually), 

I'm not pressed to get that in, but it was really easy to do, and maybe of help with future concerns as @franzihei addressed last week. :)

### DIFF
--- a/solc/CommandLineInterface.cpp
+++ b/solc/CommandLineInterface.cpp
@@ -625,17 +625,39 @@ bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
 	return true;
 }
 
+void CommandLineInterface::printLogo()
+{
+	if (!m_tty)
+		return;
+
+	// Simply print out the logo (with white background) onto stdout.
+	// The format is Sixel, which is part of the VT340 standard, is using DCS (device control string)
+	// to submit the data through stdout.
+	// Non-supporting terminals will simply ignore the DCS sequence, resulting into a no-op.
+	//
+	// The logo was converted from the docs/logo.svg using:
+	//     inkscape -w 128 -h 128 docs/logo.svg -o solc/logo.png
+	//     img2sixel solc/logo.png solc/logo.sixel
+	// Since the DCS payload consists exclusively of ASCII printable characters, I decided to embedd it
+	// as a string directly.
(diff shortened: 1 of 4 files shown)


leonardoalt:
Was this decided as wanted?

christianparpart:
> Was this decided as wanted?

No decision. We can talk about it. This is just the proof how trivial it is to add. And if we as the team like it, we can merge it. :)

leonardoalt:
I agree with how easy it seems to be. I have no strong opinions on whether it should go in or not. My only concern is if it goes in, I think it should look generically nicer (for any terminal color scheme etc), which might just steal time from more important items.

Marenz:
Yeah, we should probably remove most of the white stuff and show mainly the logo :)

leonardoalt:
> Yeah, we should probably remove most of the white stuff and show mainly the logo :)

Sure, but then it's gonna look weird on a black bg, and completely different on a white bg.

Marenz:
Not saying completely remove it, just the thick white borders

christianparpart:
> Sure, but then it's gonna look weird on a black bg, and completely different on a white bg.

transparent background is possible, but then the logo should be a little bit more colorful such that it won't end up black on baclk.

EDIT: like this: https://soliditylang.org/images/SolBlueGradient.png (but white bg made transparent)

franzihei:
I'm generally supportive on having the logo in the --help section. I can elaborate on the reasons why in our next team meeting. :)

I think an ASCII version of it might be easier and more compatible for the reasons @leonardoalt mentioned above?

leonardoalt:
Yea agree with ASCII.

cameel:
I also think some ASCII would fit in better. I generally expect to be able to copy terminal output as text. Not that I would have much need to copy the logo but still...

aarlt:
> > Yeah, we should probably remove most of the white stuff and show mainly the logo :)
> 
> Sure, but then it's gonna look weird on a black bg, and completely different on a white bg.

@christianparpart can you query the background/foreground color? so that you basically generate the images with the correct background and foreground colours.. e.g. we have only the bit mask of solidity logo, where background and foreground color is filled accordingly.. but maybe its too complicated to do it like this..

aarlt:
ASCII is quite cool, but also Sixel is cool ;)

solc/CommandLineInterface.cpp
```diff
@@ -625,17 +625,39 @@ bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
...
+	sout() << sixelGraphics;
```
axic, line 644: I'd leave this on its own line and move text below.

```suggestion
	sout() << sixelGraphics << "\n";
```

axic:
I actually like the sixel, it is fun and unusual. As long as it doesn't disturb regular operation (non-support terminals; grabbing stdout for results, etc.) I'd be in favour.

solc/main.cpp
```diff
@@ -37,7 +37,8 @@ int main(int argc, char** argv)
 {
 	try
 	{
-		solidity::frontend::CommandLineInterface cli(cin, cout, cerr);
+		bool const tty = isatty(STDOUT_FILENO);
```
axic, line 40:
```
C:\Users\circleci\project\solc\main.cpp(40,27): error C2065: 'STDOUT_FILENO': undeclared identifier [C:\Users\circleci\project\build\solc\solc.vcxproj]
  FunctionSideEffects.cpp
C:\Users\circleci\project\solc\main.cpp(40,14): error C2737: 'tty': const object must be initialized [C:\Users\circleci\project\build\solc\solc.vcxproj]
```

Not supported on Windows.

solc/CommandLineInterface.cpp
```diff
@@ -625,17 +625,39 @@ bool CommandLineInterface::parseArguments(int _argc, char const* const* _argv)
...
+
+	// Simply print out the logo (with white background) onto stdout.
+	// The format is Sixel, which is part of the VT340 standard, is using DCS (device control string)
+	// to submit the data through stdout.
+	// Non-supporting terminals will simply ignore the DCS sequence, resulting into a no-op.
+	//
+	// The logo was converted from the docs/logo.svg using:
+	//     inkscape -w 128 -h 128 docs/logo.svg -o solc/logo.png
+	//     img2sixel solc/logo.png solc/logo.sixel
+	// Since the DCS payload consists exclusively of ASCII printable characters, I decided to embedd it
```
axic, line 641: Spellcheck complains:
```suggestion
	// Since the DCS payload consists exclusively of ASCII printable characters, I decided to embed it
```

axic:
Unfortunately the default terminal on macOS doesn't support Sixel and it doesn't even ignore the commands:

Screenshot

axic:
While I prefer the logo as opposed to ascii art (because it is fun an unusual), even if we can just disable this for macos (a hack), if unsupporting terminals are not ignoring it then it becomes annoying as people will get confused by the random characters.

nikola-matic:
Bumping for visibility since we've had this PR open for a long time, and based on @axic's investigation, it looks like Sixel support isn't great. Can we make a decision as to what to do with this PR - close it in lieu of another one that uses ASCII, or implement it via ASCII in this PR, or simply close it and not have a logo?

leonardoalt:
I think regardless of the decision of what to do next, unfortunately it looks like this isn't the way to go due to lack of support, so I think it makes sense to close the PR and separately make a decision on what to do.

nikola-matic:
Closing the PR as Sixel doesn't have wide enough support - looks like ASCII is the way to go.

Issue #13368, opened 2022-08-09
Parse ``immutable`` as data location.


ekpyron:
First step in https://github.com/ethereum/solidity/issues/13323.

Previously, we considered adding a ``code`` data location - however, since EOF will have an actual data section, ``code`` as keyword may actually be misleading. So we intend to stick with more semantic term ``immutable``, but change it from only a qualifier for state variables to a full data location as well.

ekpyron:
When closing this issue, https://github.com/ethereum/solidity/issues/13369 should be moved to the ``Design/Decide`` column of the Focus Board.

ekpyron:
Since we realized that address types have a ``.code`` member, this gets slightly more complicated.

In the call today, we decided to still keep both the address member and the data location called ``code`` for now. But that means that we need a parser hack during parsing member accesses:

In fact there is also such a parser hack in place for parsing the ``.address`` member.

In ``Parser::expectIdentifierTokenOrAddress``, we just need to do the same for ``Token::Code`` as we do for ``Token::Address``.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 1INCH (169.62 USD @ $0.57/1INCH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29399).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/29399).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $1,003,799,911.71 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 264 years, 1 month from now.
Please review their action plans below:


**1) [theosch](https://gitcoin.co/theosch) has been approved to start work.**

The plan is simple: 1) Learn where the data locations are parsed; 2) Make the `code` keyword be recognized and processed like it's requirefd (i.e. report a "not yet supported" error); that's all. As there is a lot of extra information in the discussion at GitHub, it's going to take a few hours only.
Thanks

[...]

PR #13458, opened 2022-08-30
adding parser support for code keyword for data location

nishant-sachdeva:
closes https://github.com/ethereum/solidity/issues/13368

### DIFF
--- a/libsolidity/parsing/Parser.cpp
+++ b/libsolidity/parsing/Parser.cpp
@@ -811,6 +811,9 @@ ASTPointer<VariableDeclaration> Parser::parseVariableDeclaration(
 					case Token::CallData:
 						location = VariableDeclaration::Location::CallData;
 						break;
+					case Token::Code:
+						solUnimplemented("Code keyword as a data location specifier has not been implemented yet.");
+						break;
 					default:
 						solAssert(false, "Unknown data location.");
 					}
--- a/docs/grammar/SolidityLexer.g4
+++ b/docs/grammar/SolidityLexer.g4
@@ -19,6 +19,7 @@ Bool: 'bool';
 Break: 'break';
 Bytes: 'bytes';
 Calldata: 'calldata';
+Code: 'code';
 Catch: 'catch';
 Constant: 'constant';
 Constructor: 'constructor';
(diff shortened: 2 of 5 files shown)


libsolidity/parsing/Parser.cpp
```diff
@@ -811,6 +811,14 @@ ASTPointer<VariableDeclaration> Parser::parseVariableDeclaration(
 					case Token::CallData:
 						location = VariableDeclaration::Location::CallData;
 						break;
+					case Token::Code:
+						location = VariableDeclaration::Location::Code;
+						syntaxError(
```
Marenz: I think an `solUnimplementedError` (or something like that, not sure what the exact name is) would be desired here

Marenz (changes requested):
Looking at the CI there seem to be a few more places that require changes, mostly switch cases that don't handle `code` yet.

Overall the approach looks fine.

Marenz:
Hmm there are a few CI failures that are fixed in `develop`, so we probably need to merge current `develop` into `breaking` after all.

matheusaaguiar:
This change is also affecting Yul. The following minimal example:
```
object "runtime" {
    code {

    }
}
```
compiled with `./solc --strict-assembly example.yul` will result in the error:
```Error: Expected keyword "code".```

File: libsolidity/parsing/Parser.cpp [hunk 2]
```diff
@@ -811,6 +811,9 @@ ASTPointer<VariableDeclaration> Parser::parseVariableDeclaration(
 					case Token::CallData:
 						location = VariableDeclaration::Location::CallData;
 						break;
+					case Token::Code:
+						solUnimplemented("Code keyword as a data location specifier has not been implemented yet.");
```
matheusaaguiar, line 815: I was checking and for this point to be reached, `TokenTraits::isLocationSpecifier` needs to be modified to recognize `Token::Code`. Also, this throws an exception and I understood from #13368 that it should be a syntax error.

Marenz: My bad advice then, I thought we wanted to use `solUnimplemented()`

ekpyron:
I'm actually not quite sure how to react to this... did any of you actually read the issue?
The task is to add the keyword **and** to actually parse it as a data location, while making any use of the keyword a syntax error. As in an error in ``SyntaxChecker``.
While, technically, this PR is an initial step towards that, this by no means solves the issue.
Also the interactions with Yul should have become apparent with a single local CI run (the issue can be fixed by a minor adjustment of ``ObjectParser::parseCode``).

I also generally don't think it's beneficial to keep draft PRs in a state like this open. So please make sure you understand the scope of the underlying issue, if necessary ask for clarification on it, and reopen a proper PR actually addressing the issue.

PR #13460, opened 2022-08-30
Set CMake policy `CMP0115` to `NEW`: "Source file extensions must be explicit"

Pospelove:


cmake/EthPolicy.cmake
```diff
@@ -25,4 +25,10 @@ macro (eth_policy)
 	if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
 		cmake_policy(SET CMP0135 NEW)
 	endif()
+
+	if(POLICY CMP0115)
+		# speedup of cmake command on Windows
```
cameel: Looking up these policies is tedious. Please make the comment a bit more detailed so that it's possible to understand what it's doing without having to follow the thread:
```suggestion
		# Require explicit extensions for source files, do not guess.
		# The extra calls to GetFileAttributesW significantly slow down cmake on Windows.
```

cameel (commented):
Looks good but please make the comment more detailed.

ekpyron (approved):

PR #13468, opened 2022-09-01
adding parser support for code keyword for data location

### LINKED ISSUE #13368: Parse ``immutable`` as data location.

nishant-sachdeva:
closes #13368

### DIFF
--- a/libsolidity/analysis/SyntaxChecker.cpp
+++ b/libsolidity/analysis/SyntaxChecker.cpp
@@ -496,3 +496,11 @@ bool SyntaxChecker::visit(StructDefinition const& _struct)
 
 	return true;
 }
+
+bool SyntaxChecker::visit(VariableDeclaration const& _node)
+{
+	if (_node.referenceLocation() == VariableDeclaration::Location::Code)
+		m_errorReporter.syntaxError(2397_error, _node.location(), "Usage of \"code\" as a data location is not yet supported.");
+
+	return true;
+}
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -1505,6 +1505,9 @@ TypeResult ReferenceType::unaryOperatorResult(Token _operator) const
 		return TypeProvider::emptyTuple();
 	case DataLocation::Storage:
 		return isPointer() ? nullptr : TypeProvider::emptyTuple();
+	case DataLocation::Code:
+		solUnimplemented("\"code\" as data location is not yet implemented");
+		break;
 	}
 	return nullptr;
 }
@@ -1532,6 +1535,8 @@ string ReferenceType::stringForReferencePart() const
 		return "calldata";
 	case DataLocation::Memory:
 		return "memory";
+	case DataLocation::Code:
+		return "code";
 	}
 	solAssert(false, "");
 	return "";
@@ -1550,6 +1555,9 @@ string ReferenceType::identifierLocationSuffix() const
 		break;
 	case DataLocation::CallData:
 		id += "_calldata";
+		break;
+	case DataLocation::Code:
+		id += "_code";
 		break;
 	}
 	if (isPointer())
@@ -1709,6 +1717,9 @@ BoolResult ArrayType::validForLocation(DataLocation _loc) const
 			if (storageSizeUpperBound() >= bigint(1) << 256)
 				return BoolResult::err("Type too large for storage.");
 			break;
+		case DataLocation::Code:
+			solUnimplemented("\"code\" as a data location is not yet implemented");
+			break;
 	}
 	return true;
 }
@@ -1789,6 +1800,9 @@ vector<tuple<string, Type const*>> ArrayType::makeStackItems() const
 		case DataLocation::Storage:
 			// byte offset inside storage value is omitted
 			return {std::make_tuple("slot", TypeProvider::uint256())};
+		case DataLocation::Code:
+			solUnimplemented("\"code\" as a data location is not yet implemented");
+			break;
 	}
 	solAssert(false, "");
 }
@@ -2529,6 +2543,9 @@ vector<tuple<string, Type const*>> StructType::makeStackItems() const
 			return {std::make_tuple("mpos", TypeProvider::uint256())};
 		case DataLocation::Storage:
 			return {std::make_tuple("slot", TypeProvider::uint256())};
+		case DataLocation::Code:
+			solUnimplemented("\"code\" as a data location is not yet implemented");
+			break;
 	}
 	solAssert(false, "");
 }
--- a/libsolidity/codegen/CompilerUtils.cpp
+++ b/libsolidity/codegen/CompilerUtils.cpp
@@ -1111,6 +1111,9 @@ void CompilerUtils::convertType(
 				"Invalid conversion to calldata type."
 			);
 			break;
+		case DataLocation::Code:
+			solUnimplemented("\"code\" as data location is not yet implemented");
+			break;
 		}
 		break;
 	}
@@ -1231,11 +1234,17 @@ void CompilerUtils::convertType(
 			case DataLocation::Memory:
 				// nothing to do
 				break;
+			case DataLocation::Code:
+				solUnimplemented("\"code\" as data location is not yet implemented");
+				break;
 			}
 			break;
 		case DataLocation::CallData:
 			solAssert(_typeOnStack == _targetType);
 			// nothing to do
+			break;
+		case DataLocation::Code:
+			solUnimplemented("\"code\" as data location is not yet implemented");
 			break;
 		}
 		break;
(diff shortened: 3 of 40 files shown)


ekpyron:
For the record: we need to decide how to deal with the ``.code`` member of ``address`` types to proceed here.

libsolidity/analysis/SyntaxChecker.cpp
```diff
@@ -496,3 +496,11 @@ bool SyntaxChecker::visit(StructDefinition const& _struct)
 
 	return true;
 }
+
+bool SyntaxChecker::visit(VariableDeclaration const& _node)
+{
+	if (_node.referenceLocation() == VariableDeclaration::Location::Code)
+		m_errorReporter.syntaxError(2397_error, _node.location(), "Usage of Code as a Data Location is not yet supported.");
```
wechman:
```suggestion
		m_errorReporter.syntaxError(2397_error, _node.location(), "Usage of \"code\" as a data location is not yet supported.");
```

libsolidity/ast/Types.cpp
```diff
@@ -1505,6 +1505,9 @@ TypeResult ReferenceType::unaryOperatorResult(Token _operator) const
 		return TypeProvider::emptyTuple();
 	case DataLocation::Storage:
 		return isPointer() ? nullptr : TypeProvider::emptyTuple();
+	case DataLocation::Code:
+		solUnimplemented("Code Keyword as Data Location is not yet implemented");
```
wechman:
```suggestion
		solUnimplemented("\"code\" keyword as a data location is not yet implemented");
```

wechman: Please fix all the occurences.

wechman:
> For the record: we need to decide how to deal with the `.code` member of `address` types to proceed here.

@ekpyron Why did we decide to use `code` for data location over `codedata`? To me, `code` in that context looks a bit weird :)

libsolidity/ast/Types.cpp
nishant-sachdeva: Fixed in the latest push.

ekpyron:
> > For the record: we need to decide how to deal with the `.code` member of `address` types to proceed here.
> 
> @ekpyron Why did we decide to use `code` for data location over `codedata`? To me, `code` in that context looks a bit weird :)

The choice is primarly motivated by the names of the opcodes on the EVM level. It's ``codecopy``, not ``codedatacopy`` :-).

ekpyron:
@nishant-sachdeva See https://github.com/ethereum/solidity/issues/13368#issuecomment-1239448558 for the solution to the issue with the ``.code`` member of address types. Let me know if you need more guidance with it than that comment, but since we already have a very similar workaround, it shouldn't cause too much issues.

libsolidity/codegen/CompilerUtils.cpp
```diff
@@ -1231,12 +1234,18 @@ void CompilerUtils::convertType(
...
+				solUnimplemented("\"code\" keyword as a data location is not yet implemented");
+				break;
 			}
 			break;
 		case DataLocation::CallData:
 			solAssert(_typeOnStack == _targetType);
 			// nothing to do
 			break;
+		case DataLocation::Code:
+			solUnimplemented("\"code\" keyword as a data location is not yet implemented");
```
aarlt: I think it might be better to just use `'code' as data location not yet implemented.`

libyul/ObjectParser.cpp
```diff
@@ -109,7 +109,7 @@ shared_ptr<Object> ObjectParser::parseObject(Object* _containingObject)
 
 shared_ptr<Block> ObjectParser::parseCode(optional<SourceNameMap> _sourceNames)
 {
-	if (currentToken() != Token::Identifier || currentLiteral() != "code")
+	if (currentToken() != Token::Code)
 		fatalParserError(4846_error, "Expected keyword \"code\".");
 	advance();
```
ekpyron: This is pretty much just ``expectToken(Token::Code);`` isn't it?

libyul/ObjectParser.cpp
```diff
@@ -109,10 +109,8 @@ shared_ptr<Object> ObjectParser::parseObject(Object* _containingObject)
 
 shared_ptr<Block> ObjectParser::parseCode(optional<SourceNameMap> _sourceNames)
 {
-	if (currentToken() != Token::Identifier || currentLiteral() != "code")
-		fatalParserError(4846_error, "Expected keyword \"code\".");
+	expectToken(Token::Code, false);
 	advance();
```
ekpyron: Isn't both of this just
```suggestion
	expectToken(Token::Code);
```
?

nikola-matic:
I feel as though this ought be covered more by the docs, but at the same time, this is only one part of the feature, so I'm assuming the docs will be covered in another PR?

libsolidity/ast/Types.cpp
```diff
@@ -1505,6 +1505,9 @@ TypeResult ReferenceType::unaryOperatorResult(Token _operator) const
 		return TypeProvider::emptyTuple();
 	case DataLocation::Storage:
 		return isPointer() ? nullptr : TypeProvider::emptyTuple();
+	case DataLocation::Code:
```
nikola-matic, line 1508: Should cases such as these be added to `TypeProvider.cpp` as well, or do we already throw an error by the time we're supposed to reach `TypeProvider` functions?

ekpyron:
> I feel as though this ought be covered more by the docs, but at the same time, this is only one part of the feature, so I'm assuming the docs will be covered in another PR?

Yeah, this PR is just the bare-bone that would allow us to actually implement the actual feature non-breakingly on top of it - so I'd document this only once it actually starts working with a subsequent PR.

libsolidity/ast/Types.cpp
nishant-sachdeva: At the moment, yes, we should have an error at `SyntaxChecker` itself. But if required, I can add the skeletal code for `DataLocation::Code` as well. 

But it won't make any difference on the user end or any of the added test cases.

test/libsolidity/syntaxTests/code/usage_as_identifier/usage_as_int_variable.sol
```diff
@@ -0,0 +1,8 @@
+contract X {
+    function codeAsString() external {
+        uint32 calldata code;
+    }
+}
+// ----
+// ParserError 3548: (76-80): Location already specified.
```
nikola-matic, line 7: Once this gets released, I would expect something like this to be the most frequent misuse of the `code` keyword, and this error (or rather the combination of them) is not as descriptive as I would like. Sure, if you know about the new `code` location, it's easy to decipher, but if you don't, it's a little bit too cryptic.

Now, C++ and Java use our current approach, where there's no mention of a reserved keyword, but Rust for instance produces a very nice error, e.g.
```rust
fn main() {
    let break = 10;
}
```
will produce the following output when you attempt to compile
```
error: expected identifier, found keyword `break`
 --> src/main.rs:2:9
  |
2 |     let break = 10;
  |         ^^^^^ expected identifier, found keyword
  |
```
My point here is a bit of digression, and is not really in the scope of this PR, but just felt like mentioning it.

matheusaaguiar (approved):
lgtm

NunoFilipeSantos:
@nishant-sachdeva 13 tests are failing.

nikola-matic:
> @nishant-sachdeva 13 tests are failing.

These are all external tests, and the reason they're failing is that somewhere in these external repositories, `code` is likely used as a variable name, or a function name, whereas with @nishant-sachdeva's changes, `code` is now a reserved keyword, causing these tests to fail (compilation failures).
Also note that the destination branch for this is `breaking` and not `develop`, so I guess the merging procedure is different, but someone else will have to chime in on this since I have no idea :smile: 
The one thing that was supposed to be done was to either fix theses tests in their respective repositories, or to fix them in our own forks and then change the repo address in our bash scripts that run these external tests. In any case, I forgot what the outcome of this discussion was as it was a while ago.

NunoFilipeSantos:
@nikola-matic Thank you for the explanation and the headache. 😅

nishant-sachdeva:
@nikola-matic @NunoFilipeSantos 

So, about the CI tests. As Nikola mentioned some tests had this issue where they were using the `code` as an identifier. 

However, all of that has been fixed. I've made the changes in our forks in the Solidity External Tests repo. At the moment, most of these tests that are failing are showing `wrong version detected` as the error. 

For fixing that, I think, to begin with, we need to merge `develop` into `breaking` and then rebase this branch on top of `breaking`. Once we've done that, we'll get a much better estimation of where the changes need to be made.

NunoFilipeSantos:
Changing this PR to draft.

NunoFilipeSantos:
I am closing this for now.

Issue #12825, opened 2022-03-21
LSP throwing an error on each and every character I type in the editor


Gilgames000:
## Description

The LSP throws an error whenever I'm editing a .sol file. The error gets triggered every time I type a character. Restarting the LSP or reinstalling it doesn't fix the problem.

## Environment

- Compiler version: 0.8.13
- Editor: NeoVim 0.7.0
- LSP client: NeoVim built-in
- Operating system: Arch Linux 

## Steps to Reproduce

Open any .sol file and start typing.

Error message:
```
LSP[solc]: Error INVALID_SERVER_MESSAGE: {
  error = {
    code = -32603,
    message = "Unhandled exception: /solidity/libsolidity/lsp/HandlerBase.cpp(48): Throw in function Json::Value solidity::lsp::HandlerBase::toJson(const solidity::langutil::SourceLocatio
n&) const\nDynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>\nstd::exception::what: Solidity assertion failed\n[solidity::util::tag_comment*] = Solidity
 assertion failed\n"
  },
  jsonrpc = "2.0"
```

christianparpart:
Hey @Gilgames000. Thanks for your report. I am using coc.nvim along with NeoVIM / VIM. Can you maybe tell me your vim-native LSP configuration such that I can reproduce your local environment?

Gilgames000:
Sure @christianparpart! I use the built-in LSP client, the official [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) configuration for it, [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) and an auto-completion engine that uses LSPs as completion sources, and [nvim-lsp-installer](https://github.com/williamboman/nvim-lsp-installer) to easily install and manage LSP servers.

You can install all of this using [vim-plug](https://github.com/junegunn/vim-plug)
```
Plug 'neovim/nvim-lspconfig'
Plug 'williamboman/nvim-lsp-installer'
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-nvim-lsp'
```

[...]

PR #13469, opened 2022-09-01
Adds LSP trace-log-file configuration option.

christianparpart:
This PR is to aid future investigations like in #12825 to easily ask for an I/O trace log.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,7 @@ Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Language Server: Adds configuration option ``trace-log-file`` to enable trace logging to a local file.
 
 
 Bugfixes:
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -186,6 +186,16 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 			lspAssert(false, ErrorCode::InvalidParams, "Invalid file load strategy: " + text);
 	}
 
+	if (auto traceLogFileNode = _settings["trace-log-file"])
+	{
+		if (traceLogFileNode.isString())
+			m_client.setTraceLogFile(boost::filesystem::path(_settings["trace-log-file"].asString()));
+		if (traceLogFileNode.isNull())
+			m_client.setTraceLogFile(nullopt);
+		else
+			lspAssert(false, ErrorCode::InvalidParams, "Invalid trace-log-file value. Must be a path to a file.");
+	}
+
 	m_settingsObject = _settings;
 	Json::Value jsonIncludePaths = _settings["include-paths"];
 
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -189,11 +189,11 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 	if (auto traceLogFileNode = _settings["trace-log-file"])
 	{
 		if (traceLogFileNode.isString())
-			m_client.setTraceLogFile(boost::filesystem::path(_settings["trace-log-file"].asString()));
-		if (traceLogFileNode.isNull())
+			m_client.setTraceLogFile(boost::filesystem::path(traceLogFileNode.asString()));
+		else if (traceLogFileNode.isNull())
 			m_client.setTraceLogFile(nullopt);
 		else
-			lspAssert(false, ErrorCode::InvalidParams, "Invalid trace-log-file value. Must be a path to a file.");
+			lspAssert(false, ErrorCode::InvalidParams, "Invalid trace-log-file value. Must be a path to a file. "s + std::to_string(traceLogFileNode.type()));
 	}
 
 	m_settingsObject = _settings;
--- a/libsolidity/lsp/Transport.cpp
+++ b/libsolidity/lsp/Transport.cpp
@@ -27,6 +27,7 @@
 
 #include <boost/algorithm/string.hpp>
 
+#include <fstream>
 #include <iostream>
 #include <sstream>
 #include <string>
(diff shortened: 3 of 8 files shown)


libsolidity/lsp/Transport.cpp
```diff
@@ -58,6 +59,12 @@ optional<Json::Value> Transport::receive()
 
 	string const data = readBytes(stoul(headers->at("content-length")));
 
+	if (m_traceLogFilePath)
+	{
+		ofstream traceLogger(m_traceLogFilePath.value().generic_string(), ios::app);
```
aarlt: Hmm.. just wondering.. if `generic_string` is basically not the native path format - isn't it possible that a path returned here is not compatible with `ofstream`? so that `.string()` should be used here - `returns the path in native pathname format converted to a string`. Not sure though. @cameel?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -186,6 +186,16 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
 			lspAssert(false, ErrorCode::InvalidParams, "Invalid file load strategy: " + text);
 	}
 
+	if (auto traceLogFileNode = _settings["trace-log-file"])
+	{
+		if (traceLogFileNode.isString())
+			m_client.setTraceLogFile(boost::filesystem::path(traceLogFileNode.asString()));
```
nikola-matic, line 192: Looking at the test with `solc.log` passed as the parameter - will this only accept file names, or are fully qualified paths OK as well? If yes, what will happen if we pass an invalid path file here; or rather, if we do, boost will certainly complain, likely via exception (?), so is this handled appropriately?

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -186,6 +186,16 @@ void LanguageServer::changeConfiguration(Json::Value const& _settings)
...
+		else if (traceLogFileNode.isNull())
+			m_client.setTraceLogFile(nullopt);
```
nikola-matic, line 194: Is this necessary? I.e. can `nullopt` simply be a default value for `m_traceLogFilePath`?

cameel: I guess it's needed in case the client sends `changeConfiguration()` multiple times. If we set the default, it will only be `nullopt` the first time.

libsolidity/lsp/Transport.cpp
```diff
@@ -41,6 +42,12 @@ using namespace std;
 using namespace solidity::lsp;
 
 // {{{ Transport
+ofstream Transport::traceLogFileStream() const
+{
+	lspAssert(m_traceLogFilePath, ErrorCode::InternalError, "");
```
cameel, line 47:
```suggestion
	lspAssert(m_traceLogFilePath.has_value(), ErrorCode::InternalError, "");
```

Actually, I wanted to suggest `.empty()` at first but then noticed it's an `optional`. Don't you also need a check against an empty path?

cameel: Also, if the client sets the file path to empty, this will fail. I don't think you can assert here (or even have a validation against this). You just have to handle the case of not having a file name set.

libsolidity/lsp/Transport.cpp
```diff
@@ -41,6 +42,12 @@ using namespace std;
...
+	return ofstream(m_traceLogFilePath.value().generic_string(), ios::app);
```
cameel, line 48: You're not handling errors here. Is the intention that in case of things like missing permissions or being unable to create a file this will throw and `receive()` will log the error and continue?

solc/CommandLineParser.cpp
```diff
@@ -649,6 +650,16 @@ General Information)").c_str(),
 	;
 	desc.add(alternativeInputModes);
 
+	po::options_description lspModeOptions("LSP Mode Options");
+	lspModeOptions.add_options()
+		(
+			g_strLSPTrace.c_str(),
+			po::value<string>()->value_name(""),
+			"Enables trace I/O logging to a given file."
```
cameel, line 658: Other options are described in the imperative:
```suggestion
			"Enable trace I/O logging to a given file."
```

cameel: Also, I think this might be a bit clearer to users:

```suggestion
			"Log all JSON messages exchanged with an LSP client to a file."
```

solc/CommandLineParser.cpp
```diff
@@ -942,7 +953,12 @@ void CommandLineParser::processArgs()
 		);
 
 	if (m_options.input.mode == InputMode::LanguageServer)
+	{
+		if (m_args.count("lsp-trace"))
+			m_options.languageServer.traceLogFile = boost::filesystem::path(m_args.at("lsp-trace").as<string>());
```
cameel, line 958: Since this option is not valid in other input modes, please make them report it as an error.

You can add this by adding it to `validOptionInputModeCombinations`. Don't forget to update `CommandLineParser` tests for that part too.

test/lsp.py
```diff
@@ -948,6 +948,11 @@ def setup_lsp(
                 params['initializationOptions'] = {}
             params['initializationOptions']['include-paths'] = custom_include_paths
 
+        if self.trace_io:
```
cameel, line 951: Shouldn't the type of `trace_io` be declared along with other class members at the top of the class definition?

test/lsp.py
```diff
@@ -948,6 +948,11 @@ def setup_lsp(
...
+            if params['initializationOptions'] is None:
+                params['initializationOptions'] = {}
+            params['initializationOptions']['trace-log-file'] = "solc.log"
```
cameel, line 954: Would be nice if was possible to immediately tell at a glance that this is specifically the log from the LSP and not something else `solc` might do:
```suggestion
            params['initializationOptions']['trace-log-file'] = "solc-lsp.log"
```

Changelog.md
```diff
@@ -19,6 +19,7 @@ Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Language Server: Adds configuration option ``trace-log-file`` to enable trace logging to a local file.
```
cameel, line 22:
```suggestion
 * Language Server: Add configuration option ``trace-log-file`` to enable trace logging to a local file.
```

cameel (changes requested):

PR #13473, opened 2022-09-02
LSP: Fixes URL decoding incoming file names.

### LINKED ISSUE #13035: LSP: `textDocument/definition` always returns "result":[]
## Environment

- Compiler version: Version: 0.8.14+commit.80d49f37.Windows.msvc
- Framework/IDE (e.g. Truffle or Remix): Notepad++ 8.3.3. 64bit & https://github.com/Ekopalypse/NppLspClient V0.0.9 64 bits
- EVM execution environment / backend / blockchain client: default
- Operating system: Windows

### setup
Install `solc-windows.exe` in `C:\tmp\solidity\`
Install `Notepad++`
Setup a language definition for Solidity in `Notepad++`  (for example via `https://github.com/alin1popa/solidity-notepadplusplus-userdefined-highlight`)
Download `https://github.com/Ekopalypse/NppLspClient/releases/download/v.0.0.9-alpha/NppLspClient_x64.zip`
Install `NppLspClient.dll` in `C:\Program Files\Notepad++\plugins\NppLspClient`
Edit config file via menu `open configuration file` (`NppLspClientConfig.toml`) to contain:
```
[lspservers.solidity]
executable = 'C:\tmp\solidity\solc-windows.exe'
args = '--lsp'
mode = "io"
auto_start_server = true
```
Have some sample soldity files in  `C:\tmp\solidity\`
test.sol:
```solidity
import "test2.sol";

contract test() {
   uint x=1;
   uint y=x;
}   
```
test2.sol
```solidity
```

### test in Notepad++
Open file `test.sol` in `Notepad++`
In menu option `Plugins/NppLspClient_x64` choose `Toggle Console` to see debug output.
Put the cursor on `import`
In menu option `Plugins/NppLspClient_x64` choose `Goto Definition`
Now nothing happens in the editor main screen.
The LSP output console shows: `{"id":4,"jsonrpc":"2.0","result":[]}`
So somehow the lookup doesn't work.
The same happens with anything else I select.

Note: the detection of syntax errors does work and is shown in the editor, so the LSP interaction does work.

### full log
```
checking current lexer
checking language server status: solidity
  trying to start C:\tmp\solidity\solc-windows.exe
  running
on_initialize: 15908
write_to_stdin: Content-Length: 2128

{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"processId":15908,"clientInfo":{"name":"NppLspClient","version":"0.0.1"},"rootUri":"file:///C%3A/tmp/solidity","initializationOptions":{},"capabilities":{"workspace":{"applyEdit":false,"workspaceEdit":{"documentChanges":false},"didChangeConfiguration":{"dynamicRegistration":false},"didChangeWatchedFiles":{"dynamicRegistration":false},"symbol":{"dynamicRegistration":false,"symbolKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]}},"executeCommand":{"dynamicRegistration":false},"configuration":false,"workspaceFolders":false},"textDocument":{"publishDiagnostics":{"relatedInformation":false},"synchronization":{"dynamicRegistration":false,"willSave":false,"willSaveWaitUntil":false,"didSave":true},"completion":{"dynamicRegistration":false,"contextSupport":false,"completionItem":{"snippetSupport":false,"commitCharactersSupport":false,"documentationFormat":["plaintext"],"deprecatedSupport":false},"completionItemKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25]}},"hover":{"dynamicRegistration":false,"contentFormat":["plaintext"]},"signatureHelp":{"dynamicRegistration":false,"signatureInformation":{"documentationFormat":["plaintext"]}},"definition":{"dynamicRegistration":false},"references":{"dynamicRegistration":false},"documentHighlight":{"dynamicRegistration":false},"documentSymbol":{"dynamicRegistration":false,"symbolKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]}},"codeAction":{"dynamicRegistration":false},"codeLens":{"dynamicRegistration":false},"formatting":{"dynamicRegistration":false},"rangeFormatting":{"dynamicRegistration":false},"onTypeFormatting":{"dynamicRegistration":false},"rename":{"dynamicRegistration":false},"documentLink":{"dynamicRegistration":false},"typeDefinition":{"dynamicRegistration":false},"implementation":{"dynamicRegistration":false},"colorProvider":{"dynamicRegistration":false},"foldingRange":{"dynamicRegistration":false,"rangeLimit":100,"lineFoldingOnly":true}}},"trace":"off","workspaceFolders":null}}
on_message_received: Content-Length: 205

{"id":0,"jsonrpc":"2.0","result":{"capabilities":{"definitionProvider":true,"implementationProvider":true,"textDocumentSync":{"change":2,"openClose":true}},"serverInfo":{"name":"solc","version":"0.8.14"}}}
    initialized response received
write_to_stdin: Content-Length: 52

{"jsonrpc":"2.0","method":"initialized","params":{}}
on_file_opened:: solidity: C:\tmp\solidity\test.sol
on_file_opened: initialized=true
write_to_stdin: Content-Length: 291

{"jsonrpc":"2.0","method":"textDocument/didOpen","params":{"textDocument":{"uri":"file:///C%3A/tmp/solidity/test.sol","languageId":"solidity","version":0,"text":"// SPDX-License-Identifier: MIT\r\nimport \"test2.sol\";\r\n\r\ncontract test() {\r\n   uint x=1;\r\n   uint y=x;\r\n}   \r\n"}}}
on_message_received: Content-Length: 299

{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"diagnostics":[{"code":2314,"message":"ParserError: Expected '{' but got '('","range":{"end":{"character":14,"line":3},"start":{"character":13,"line":3}},"severity":1,"source":"solc"}],"uri":"file:///C%3A/tmp/solidity/test.sol"}}
write_to_stdin: Content-Length: 166

{"jsonrpc":"2.0","id":1,"method":"textDocument/definition","params":{"textDocument":{"uri":"file:///C%3A/tmp/solidity/test.sol"},"position":{"character":0,"line":1}}}
on_message_received: Content-Length: 36

{"id":1,"jsonrpc":"2.0","result":[]}

```

christianparpart:
This fixes #13035. 

On Windows platform, it is much more likely to hit, as we did not URL decode the incoming URL, and thus, the files (nor the base path, if specified as `rootUri`) could not be opened nor traversed.

### Checklist

- [x] fix URL decoding/encoding
- [x] Changelog entry
- [x] tests

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ Compiler Features:
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 * Language Server: Add basic document hover support.
+* Language Server: Fixes URL decoding of incoming file names.
 
 
 Bugfixes:
@@ -23,6 +24,7 @@ Important Bugfixes:
 
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
+ * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
  * Language Server: Analyze all files in a project by default (can be customized by setting ``'file-load-strategy'`` to ``'directly-opened-and-on-import'`` in LSP settings object).
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
 
--- a/libsolutil/CMakeLists.txt
+++ b/libsolutil/CMakeLists.txt
@@ -43,7 +43,7 @@ set(sources
 )
 
 add_library(solutil ${sources})
-target_link_libraries(solutil PUBLIC jsoncpp Boost::boost Boost::filesystem Boost::system range-v3)
+target_link_libraries(solutil PUBLIC jsoncpp Boost::boost Boost::filesystem Boost::system range-v3 fmt::fmt-header-only)
 target_include_directories(solutil PUBLIC "${CMAKE_SOURCE_DIR}")
 add_dependencies(solutil solidity_BuildInfo.h)
 
--- a/libsolutil/StringUtils.cpp
+++ b/libsolutil/StringUtils.cpp
@@ -25,8 +25,10 @@
 #include <libsolutil/StringUtils.h>
 #include <string>
 #include <vector>
+#include <fmt/format.h>
 
 using namespace std;
+using namespace std::string_view_literals;
 using namespace solidity::util;
 
 bool solidity::util::stringWithinDistance(string const& _str1, string const& _str2, size_t _maxDistance, size_t _lenThreshold)
@@ -113,3 +115,50 @@ string solidity::util::suffixedVariableNameList(string const& _baseName, size_t 
 	}
 	return result;
 }
+
+string solidity::util::decodeURI(string const& _uri)
+{
+	string decodedString;
+	for (size_t i = 0; i < _uri.size(); ++i)
+	{
+		char const ch = _uri[i];
+		if (ch != '%')
+			decodedString.push_back(ch);
+		else if (i + 2 < _uri.size())
+		{
+			char const buf[3] = { _uri[i + 1], _uri[i + 2], '\0' };
+			char* end = nullptr;
+			auto const numericValue = std::strtol(buf, &end, 16);
+			if (end == buf + 2)
+			{
+				decodedString.push_back(static_cast<char>(numericValue));
+				i += 2;
+			}
+		}
+	}
+	return decodedString;
+}
+
+string solidity::util::encodeURI(string const& _uri)
+{
+	string encodedString;
+	for (char const ch: _uri)
+	{
+		// see https://en.wikipedia.org/wiki/Percent-encoding#Types_of_URI_characters
+		static auto constexpr allowedCharacters = string_view(
+			// unreserved characters
+			"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+			"abcdefghijklmnopqrstuvwxyz"
+			"0123456789"
+			"-_.~"
+			// reserved characters (special meaning depending on URI component)
+			"/:"
+		);
+		if (allowedCharacters.find(ch) != string_view::npos)
+			encodedString += ch;
+		else
+			encodedString += fmt::format("%{:02X}", static_cast<unsigned>(ch));
+	}
+	return encodedString;
+}
+
--- a/libsolutil/StringUtils.h
+++ b/libsolutil/StringUtils.h
@@ -50,6 +50,13 @@ std::string quotedAlternativesList(std::vector<std::string> const& suggestions);
 /// If @a _startSuffix == @a _endSuffix, the empty string is returned.
 std::string suffixedVariableNameList(std::string const& _baseName, size_t _startSuffix, size_t _endSuffix);
 
+/// Decodes a URI with respect to %XX notation.
+/// No URI-validity verification is performed but simply the URI decoded into non-escaping characters.
+std::string decodeURI(std::string const& _uri);
+
+/// Encodes a string into a URI conform notation.
+std::string encodeURI(std::string const& _uri);
+
 /// Joins collection of strings into one string with separators between, last separator can be different.
 /// @param _list collection of strings to join
 /// @param _separator defaults to ", "
--- a/test/libsolutil/StringUtils.cpp
+++ b/test/libsolutil/StringUtils.cpp
@@ -210,6 +210,38 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
 	);
 }
 
+BOOST_AUTO_TEST_CASE(test_encodeURI)
+{
+	BOOST_CHECK_EQUAL(util::encodeURI(""), "");
+	BOOST_CHECK_EQUAL(util::encodeURI(" "), "%20");
+	BOOST_CHECK_EQUAL(util::encodeURI("%"), "%25");
+	BOOST_CHECK_EQUAL(util::encodeURI("Hello World."), "Hello%20World.");
+	BOOST_CHECK_EQUAL(util::encodeURI("\x01\x02\x7F"), "%01%02%7F");
+	BOOST_CHECK_EQUAL(util::encodeURI("C:\\readme.md"), "C:%5Creadme.md");
+	BOOST_CHECK_EQUAL(util::encodeURI("C:/readme.md"), "C:/readme.md");
+	BOOST_CHECK_EQUAL(util::encodeURI("/var/log"), "/var/log");
+}
+
+BOOST_AUTO_TEST_CASE(test_decodeURI)
+{
+	BOOST_CHECK_EQUAL(util::decodeURI(""), "");
+	BOOST_CHECK_EQUAL(util::decodeURI(""), "");
+	BOOST_CHECK_EQUAL(util::decodeURI(" ")," ");
+	BOOST_CHECK_EQUAL(util::decodeURI("%25"), "%");
+	BOOST_CHECK_EQUAL(util::decodeURI("Hello%20World."), "Hello World.");
+	BOOST_CHECK_EQUAL(util::decodeURI("Hello World."), "Hello World.");
+	BOOST_CHECK_EQUAL(util::decodeURI("%01%02%7F"), "\x01\x02\x7F");
+	BOOST_CHECK_EQUAL(util::decodeURI("C%3A%5Creadme.md"), "C:\\readme.md");
+	BOOST_CHECK_EQUAL(util::decodeURI("C:/readme.md"), "C:/readme.md");
+
+	// Decoding failure cases (there's not really a standard for that).
+	BOOST_CHECK_EQUAL(util::decodeURI("%"), "");
+	BOOST_CHECK_EQUAL(util::decodeURI("%ZZ"), "ZZ");
+	BOOST_CHECK_EQUAL(util::decodeURI("%7Ge"), "7Ge");
+	BOOST_CHECK_EQUAL(util::decodeURI("%2F%2%2F"), "/2/");
+	BOOST_CHECK_EQUAL(util::decodeURI("%1G/%7G/%FG"), "1G/7G/FG");
+}
+
 BOOST_AUTO_TEST_SUITE_END()
 
 }
--- a/test/lsp.py
+++ b/test/lsp.py
@@ -11,6 +11,7 @@ import re
 import subprocess
 import sys
 import traceback
+import urllib.parse
 from collections import namedtuple
 from copy import deepcopy
 from enum import Enum, auto
@@ -358,6 +359,10 @@ def extendEnd(marker, amount=1):
     newMarker["end"]["character"] += amount
     return newMarker
 
+# Convenience method to intuitively decode a URI.
+def decodeURI(text: str) -> str:
+    return urllib.parse.unquote(text)
+
 class TestParserException(Exception):
     def __init__(self, incompleteResult, msg: str):
         self.result = incompleteResult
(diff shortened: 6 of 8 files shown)


libsolutil/StringUtils.cpp
```diff
@@ -113,3 +115,47 @@ string solidity::util::suffixedVariableNameList(string const& _baseName, size_t
 	}
 	return result;
 }
+
+string solidity::util::decodeURI(string const& _uri)
+{
+	string decodedString;
+	for (size_t i = 0; i < _uri.size(); ++i)
+	{
+		char const ch = _uri.at(i);
```
nikola-matic: Why are you using `at` access here, since I don't see a way for this to throw an out of bounds exception?

christianparpart: @nikola-matic me neither. But if I remember correctly, I was corrected to prefer .at() as good practice. I check the source code for references. Otherwise, yes, you're right. I think it's actually obvious here in this case that this can't go OOB.

nikola-matic: I don't actually mind the usage of `at` so much, as it is definitely safer, plus in the case of an eventual failure, it's definitely easier to debug an out of bounds error vs a segfault, it's just that you're already doing the bounds checking in `if` and `for` here, so it's kinda unnecessary.

In any case, it's just a nit, don't feel pressured to change if you're fine with it as it is now :)

libsolutil/StringUtils.cpp
```diff
@@ -113,3 +115,47 @@ string solidity::util::suffixedVariableNameList(string const& _baseName, size_t
...
+			"abcdefghijklmnopqrstuvwxyz"
+			"0123456789"
+			"-_.~"
+			// reserved characters (special meaning depending on URI component)
+			"/"
+		);
+		if (allowedCharacters.find(ch) != string_view::npos)
+			encodedString += ch;
+		else
+			encodedString += fmt::format("%{:02X}", static_cast<unsigned>(ch));
```
aarlt, line 160: It would be nice to add some tests for `decodeURI` & `encodeURI` to `test/libsolutil/StringUtils.cpp`.

nikola-matic: He has a task open for tests :)

aarlt: ah yeah.. you mean that checkbox? I missed that :D

Marenz (changes requested):
I'd like to see at least one test with a full url that is also using `/`

Marenz:
Actually, maybe also a windows-style path in the tests would be good

Changelog.md
```diff
@@ -9,6 +9,7 @@ Language Features:
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
  * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
+ * Language Server: Fixes URL decoding incoming file names.
```
nikola-matic:
```suggestion
 * Language Server: Fixes URL decoding of incoming file names.
```

test/libsolutil/StringUtils.cpp
```diff
@@ -210,6 +210,33 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
 	);
 }
 
+BOOST_AUTO_TEST_CASE(test_encodeURI)
```
nikola-matic, line 213: It would be nice to add tests for all of the 'special' characters, i.e. non alpha-numeric from `allowedCharacters` list. Also, can you add some realistic examples here in the context of the LSP (both for encoding and decoding)?

nikola-matic: Ah, I see @Marenz already left a similar top level comment.

christianparpart: done

libsolutil/StringUtils.cpp
cameel: I usually use it like an assert. So in contexts where I'd not feel the need to assert or I have already asserted I use `[]`, otherwise it's `at()`.

Changelog.md
```diff
@@ -20,6 +21,7 @@ Important Bugfixes:
 
 Compiler Features:
  * Code Generator: More efficient overflow checks for multiplication.
+ * Yul Optimizer: Simplify the starting offset of zero-length operations to zero.
```
cameel, line 27: This seems out of place in this PR.

Changelog.md
```diff
@@ -7,6 +7,7 @@ Compiler Features:
 * Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
+* Language Server: Fixes URL decoding of incoming file names.
```
cameel, line 12: I was initially confused as to why file names would need URL-decoding. I think it would be better to say that we're decoding the `file://` URLs.
```suggestion
* Language Server: Properly URL-decode the `file://` URIs coming from the client and encode the URIs coming out of by the server.
```
I'd also say it's a bugfix rather than a feature :)

libsolutil/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ set(sources
 )
 
 add_library(solutil ${sources})
-target_link_libraries(solutil PUBLIC jsoncpp Boost::boost Boost::filesystem Boost::system range-v3)
+target_link_libraries(solutil PUBLIC jsoncpp Boost::boost Boost::filesystem Boost::system range-v3 fmt::fmt-header-only)
```
cameel, line 46: This and the URL-encoding helpers are things I'd put in a separate commit. They should stand out as its own thing in history.

libsolutil/StringUtils.h
```diff
@@ -50,6 +50,13 @@ std::string quotedAlternativesList(std::vector<std::string> const& suggestions);
 /// If @a _startSuffix == @a _endSuffix, the empty string is returned.
 std::string suffixedVariableNameList(std::string const& _baseName, size_t _startSuffix, size_t _endSuffix);
 
+/// Decodes a URI with respect to %XX notation.
+/// No URI-validity verification is performed but simply the URI decoded into non-escaping characters.
+std::string decodeURI(std::string const& _uri);
+
+/// Encodes a string into a URI conform notation.
+std::string encodeURI(std::string const& _uri);
```
cameel, line 58: Given that these don't have to be URIs (it works for arbitrary strings), I'd call it something like `urlEncode()` and `urlDecode()` (or even `urlencode()` and `urldecode()`). I associate the term "urlencoding" specifically with this kind of encoding while "encoding an URI" could mean any encoding. I would probably look for the former I wanted to find it.

cameel, line 58:
```suggestion
/// Decodes a percent-encoded string (i.e. the %XX notation).
/// No URI-validity verification is performed but simply the URI decoded into non-escaping characters.
std::string decodeURI(std::string const& _uri);

/// Percent-encodes a string.
/// Note that `:` and `/` are preserved, which means that the function will not mangle the `protocol://`
/// prefix in URLs.
std::string encodeURI(std::string const& _uri);
```

libsolutil/StringUtils.cpp
```diff
@@ -113,3 +115,50 @@ string solidity::util::suffixedVariableNameList(string const& _baseName, size_t
...
+	return decodedString;
+}
+
+string solidity::util::encodeURI(string const& _uri)
+{
+	string encodedString;
+	for (char const ch: _uri)
+	{
+		// see https://en.wikipedia.org/wiki/Percent-encoding#Types_of_URI_characters
+		static auto constexpr allowedCharacters = string_view(
```
cameel, line 148: Wouldn't a `set` be more appropriate here? You'd avoid linear search in a string.

test/libsolutil/StringUtils.cpp
```diff
@@ -210,6 +210,38 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
...
+{
+	BOOST_CHECK_EQUAL(util::encodeURI(""), "");
+	BOOST_CHECK_EQUAL(util::encodeURI(" "), "%20");
+	BOOST_CHECK_EQUAL(util::encodeURI("%"), "%25");
+	BOOST_CHECK_EQUAL(util::encodeURI("Hello World."), "Hello%20World.");
+	BOOST_CHECK_EQUAL(util::encodeURI("\x01\x02\x7F"), "%01%02%7F");
+	BOOST_CHECK_EQUAL(util::encodeURI("C:\\readme.md"), "C:%5Creadme.md");
+	BOOST_CHECK_EQUAL(util::encodeURI("C:/readme.md"), "C:/readme.md");
+	BOOST_CHECK_EQUAL(util::encodeURI("/var/log"), "/var/log");
```
cameel, line 222: Could you add a few more cases?
- A Windows path with a `file://` prefix.
- `😃`
- `%%`.

test/libsolutil/StringUtils.cpp
```diff
@@ -210,6 +210,38 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
...
+}
+
+BOOST_AUTO_TEST_CASE(test_decodeURI)
+{
+	BOOST_CHECK_EQUAL(util::decodeURI(""), "");
+	BOOST_CHECK_EQUAL(util::decodeURI(""), "");
```
cameel, line 228: Duplicate test.

test/libsolutil/StringUtils.cpp
```diff
@@ -210,6 +210,38 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
...
+	BOOST_CHECK_EQUAL(util::decodeURI(" ")," ");
```
cameel, line 229:
```suggestion
	BOOST_CHECK_EQUAL(util::decodeURI(" "), " ");
```

test/libsolutil/StringUtils.cpp
```diff
@@ -210,6 +210,38 @@ BOOST_AUTO_TEST_CASE(test_format_number_readable_signed)
...
+	BOOST_CHECK_EQUAL(util::decodeURI("%01%02%7F"), "\x01\x02\x7F");
+	BOOST_CHECK_EQUAL(util::decodeURI("C%3A%5Creadme.md"), "C:\\readme.md");
+	BOOST_CHECK_EQUAL(util::decodeURI("C:/readme.md"), "C:/readme.md");
+
+	// Decoding failure cases (there's not really a standard for that).
+	BOOST_CHECK_EQUAL(util::decodeURI("%"), "");
+	BOOST_CHECK_EQUAL(util::decodeURI("%ZZ"), "ZZ");
+	BOOST_CHECK_EQUAL(util::decodeURI("%7Ge"), "7Ge");
+	BOOST_CHECK_EQUAL(util::decodeURI("%2F%2%2F"), "/2/");
+	BOOST_CHECK_EQUAL(util::decodeURI("%1G/%7G/%FG"), "1G/7G/FG");
```
cameel, line 242: Wouldn't it be better to handle these as errors? Having `%` silently disappear from the end of a string when you pass in something not encoded by accident does not sound like the nicest behavior.

cameel, line 242: Can you add a few more?
- A Windows path with a `file://` prefix.
- `%😃`
- `%Z`

test/lsp.py
```diff
@@ -747,7 +752,9 @@ def run_testcase(self, testcase: TestParser.RequestAndResponse):
             if isinstance(actualResponseJson["result"], list):
                 for result in actualResponseJson["result"]:
                     if "uri" in result:
-                        result["uri"] = result["uri"].replace(self.suite.project_root_uri + "/" + self.sub_dir + "/", "")
+                        result["uri"] = decodeURI(result["uri"]).replace(
+                            self.suite.project_root_uri + "/" + self.sub_dir + "/", ""
```
cameel, line 756:
```suggestion
                            self.suite.project_root_uri + "/" + self.sub_dir + "/",
                            ""
```

cameel (commented):
Looks good overall, but could use some small tweaks.

PR #13474, opened 2022-09-02
Add note about peephole optimizer in docs

matheusaaguiar:


docs/internals/optimizer.rst
```diff
@@ -26,6 +26,10 @@ One can use ``solc --ir-optimized --optimize`` to produce an
 optimized Yul IR for a Solidity source. Similarly, one can use ``solc --strict-assembly --optimize``
 for a stand-alone Yul mode.
 
+.. note::
+    The PeepholeOptimizer is always enabled by default and can only be turned off via the
```
nikola-matic: Would it make sense to mention what the peephole optimizer is, since I don't think we have it documented anywhere?

matheusaaguiar: From what I gather, `Peephole` refers to a common optimization method, so I guess that's why it does not have its own doc section. Maybe we could add an external link ([wikipedia article](https://en.wikipedia.org/wiki/Peephole_optimization)) as a reference to the technique itself.

matheusaaguiar:
```suggestion
    The `PeepholeOptimizer <https://en.wikipedia.org/wiki/Peephole_optimization>`_ is always enabled by default and can only be turned off via the
```

nikola-matic:
Pull `develop` and rebase against it to get the build to a merge-able state pls :)

nikola-matic:
By the way, did you check if the linking to the JSON input output works properly with `docs/docs.sh`? Otherwise, LGTM.

nikola-matic (dismissed):

matheusaaguiar:
> By the way, did you check if the linking to the JSON input output works properly with `docs/docs.sh`? Otherwise, LGTM.

Yeah, checked it.
Rebased on develop now.

nikola-matic (dismissed):

nikola-matic:
> > By the way, did you check if the linking to the JSON input output works properly with `docs/docs.sh`? Otherwise, LGTM.
> 
> Yeah, checked it. Rebased on develop now.

Approved again, although you might wanna wait for @ekpyron, @chriseth or @cameel as per the new policy?

matheusaaguiar:
I guess it is not really needed, in this case. Although, I do prefer the merge is done by someone else rather than PR author.
Anyway, this is not urgent :)

docs/internals/optimizer.rst
```diff
@@ -26,6 +26,10 @@ One can use ``solc --ir-optimized --optimize`` to produce an
 optimized Yul IR for a Solidity source. Similarly, one can use ``solc --strict-assembly --optimize``
 for a stand-alone Yul mode.
 
+.. note::
+    The `PeepholeOptimizer <https://en.wikipedia.org/wiki/Peephole_optimization>`_ is always
+    enabled by default and can only be turned off via the :ref:`JSON-input-output interface <compiler-api>`.
```
cameel: Peephole optimizer is not the only such step. The other one is the inliner:
>        // The inliner is always on if no details are given,
>        // use details to switch it off.
>        "inliner": true,

Can you mention it too?

matheusaaguiar: Done.

docs/internals/optimizer.rst
```diff
@@ -26,6 +26,10 @@ One can use ``solc --ir-optimized --optimize`` to produce an
 optimized Yul IR for a Solidity source. Similarly, one can use ``solc --strict-assembly --optimize``
 for a stand-alone Yul mode.
 
+.. note::
+    The `PeepholeOptimizer <https://en.wikipedia.org/wiki/Peephole_optimization>`_ and the inliner are always
```
cameel: Are we referring to optimizer steps or just using generic names? I'd make that consistent one way or another.
```suggestion
    The `peephole optimizer <https://en.wikipedia.org/wiki/Peephole_optimization>`_ and the inliner are always
```

docs/internals/optimizer.rst
```diff
@@ -26,6 +26,10 @@ One can use ``solc --ir-optimized --optimize`` to produce an
...
+    enabled by default and can only be turned off via the :ref:`JSON-input-output interface <compiler-api>`.
```
cameel: We usually refer to it as Standard JSON:
```suggestion
    enabled by default and can only be turned off via the :ref:`Standard JSON <compiler-api>`.
```

cameel (dismissed):
I noticed two tiny details that might be good to change but overall the PR is good.

docs/internals/optimizer.rst
matheusaaguiar: The other occurrences of the word are as you suggested, so I am changing it.

docs/internals/optimizer.rst
matheusaaguiar: Ok. Done.

cameel (approved):

PR #13479, opened 2022-09-05
Fix of unused store remover storage bug.

### LINKED ISSUE #13478: [Yul optimizer] RSE incorrectly removes sstore
The following Yul program

```
{
  function f(){
    sstore(0, 1)
    g()
  }
  function g() {
    switch mload(0)
    case 2 { leave }
    return(0, 0)
  }
  f()
  f()
}
```

on optimization, produces

```
{
        {                                                                      
            g()                                                                
            sstore(0, 1)                                                       
            g()
        }                              
        function g()      
        {                              
            if eq(2, mload(0)) { leave }
            return(0, 0)               
        }
}
```

It can happen that `mload(0)` not equal to `2` in which case, the optimized program will not write `1` to storage slot `0` but the unoptimized program will. The unoptimized program semantics is correct.

I think the bug is that the RSE removes the first sstore under the assumption that the second sstore will always happen. But it may not if `g()` returns.

Note that the bug affects only IR pipeline since legacy would not invoke RSE due to not having global scope i.e., in the following semantic test, all combinations except `optimize + ir` pass.

```
contract C {
  uint public s;
  function f() external {
    assembly {
      function g() {
        sstore(s.slot, 1)
        h()
      }
      function h() {
        switch mload(0)
        case 2 { leave }
        return(0, 0)
      }
      g()
      g()
    }
  }
}
// ====
// compileViaYul: also
// ----
// f() ->
// s() -> 1
```

ekpyron:
Fixes https://github.com/ethereum/solidity/issues/13478

### DIFF
--- a/docs/bugs.json
+++ b/docs/bugs.json
@@ -1,5 +1,18 @@
 [
     {
+        "uid": "SOL-2022-7",
+        "name": "StorageWriteRemovalBeforeConditionalTermination",
+        "summary": "Calling functions that conditionally terminate the external EVM call using the assembly statements ``return(...)`` or ``stop()`` may result in incorrect removals of prior storage writes.",
+        "description": "A call to a Yul function that conditionally terminates the external EVM call could result in prior storage writes being incorrectly removed by the Yul optimizer. This used to happen in cases in which it would have been valid to remove the store, if the Yul function in question never actually terminated the external call, and the control flow always returned back to the caller instead. Conditional termination within the same Yul block instead of within a called function was not affected. In Solidity with optimized via-IR code generation, any storage write before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may have been incorrectly removed, whenever it would have been valid to remove the write without the ``return(...)`` or ``stop()``. In optimized legacy code generation, only inline assembly that did not refer to any Solidity variables and that involved conditionally-terminating user-defined assembly functions could be affected.",
+        "link": "https://blog.soliditylang.org/2022/09/08/storage-write-removal-before-conditional-termination/",
+        "introduced": "0.8.13",
+        "fixed": "0.8.17",
+        "severity": "medium/high",
+        "conditions": {
+            "yulOptimizer": true
+        }
+    },
+    {
         "uid": "SOL-2022-6",
         "name": "AbiReencodingHeadOverflowWithStaticArrayCleanup",
         "summary": "ABI-encoding a tuple with a statically-sized calldata array in the last component would corrupt 32 leading bytes of its first dynamically encoded component.",
--- a/libyul/optimiser/UnusedStoreEliminator.cpp
+++ b/libyul/optimiser/UnusedStoreEliminator.cpp
@@ -105,10 +105,13 @@ void UnusedStoreEliminator::operator()(FunctionCall const& _functionCall)
 	else
 		sideEffects = m_controlFlowSideEffects.at(_functionCall.functionName.name);
 
+	if (sideEffects.canTerminate)
+		changeUndecidedTo(State::Used, Location::Storage);
 	if (!sideEffects.canContinue)
 	{
 		changeUndecidedTo(State::Unused, Location::Memory);
-		changeUndecidedTo(sideEffects.canTerminate ? State::Used : State::Unused, Location::Storage);
+		if (!sideEffects.canTerminate)
+			changeUndecidedTo(State::Unused, Location::Storage);
 	}
 }
 
--- a/test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
+++ b/test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
@@ -1,0 +1,15 @@
+contract C {
+	uint public x;
+	function f() public {
+		x = 1; // This write used to be removed by the Yul optimizer due to the StorageWriteRemovalBeforeConditionalTermination bug.
+		g();
+		x = 2;
+	}
+	function g() internal {
+		if (msg.data.length > 4) return;
+		assembly { return(0, 0) }
+	}
+}
+// ----
+// f() ->
+// x() -> 1
--- a/test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_3.yul
+++ b/test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_3.yul
@@ -1,0 +1,27 @@
+{
+    function conditionallyStop() {
+        if calldataload(0) { leave }
+        return(0, 0)
+    }
+    let x := 0
+    let y := 1
+    sstore(x, y)
+    conditionallyStop()
+    sstore(x, y)
+}
+// ----
+// step: unusedStoreEliminator
+//
+// {
+//     {
+//         let x := 0
+//         let y := 1
+//         conditionallyStop()
+//         sstore(x, y)
+//     }
+//     function conditionallyStop()
+//     {
+//         if calldataload(0) { leave }
+//         return(0, 0)
+//     }
+// }
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ### 0.8.17 (unreleased)
 
 Important Bugfixes:
+ * Yul Optimizer: Prevent the incorrect removal of storage writes before calls to Yul functions that conditionally terminate the external EVM call.
 
 
 Language Features:
--- a/docs/bugs_by_version.json
+++ b/docs/bugs_by_version.json
@@ -1737,6 +1737,7 @@
     },
     "0.8.13": {
         "bugs": [
+            "StorageWriteRemovalBeforeConditionalTermination",
             "AbiReencodingHeadOverflowWithStaticArrayCleanup",
             "DirtyBytesArrayToStorage",
             "InlineAssemblyMemorySideEffects",
@@ -1747,6 +1748,7 @@
     },
     "0.8.14": {
         "bugs": [
+            "StorageWriteRemovalBeforeConditionalTermination",
             "AbiReencodingHeadOverflowWithStaticArrayCleanup",
             "DirtyBytesArrayToStorage",
             "InlineAssemblyMemorySideEffects"
@@ -1755,12 +1757,15 @@
     },
     "0.8.15": {
         "bugs": [
+            "StorageWriteRemovalBeforeConditionalTermination",
             "AbiReencodingHeadOverflowWithStaticArrayCleanup"
         ],
         "released": "2022-06-15"
     },
     "0.8.16": {
-        "bugs": [],
+        "bugs": [
+            "StorageWriteRemovalBeforeConditionalTermination"
+        ],
         "released": "2022-08-08"
     },
     "0.8.2": {
--- a/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call.yul
+++ b/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call.yul
@@ -1,0 +1,28 @@
+{
+    function conditionallyStop() {
+        if calldataload(0) { leave }
+        return(0, 0)
+    }
+    let x := 0
+    let y := 1
+    sstore(x, y) // used to be removed due to the StorageWriteRemovalBeforeConditionalTermination
+    conditionallyStop()
+    sstore(x, y)
+}
+// ----
+// step: unusedStoreEliminator
+//
+// {
+//     {
+//         let x := 0
+//         let y := 1
+//         sstore(x, y)
+//         conditionallyStop()
+//         sstore(x, y)
+//     }
+//     function conditionallyStop()
+//     {
+//         if calldataload(0) { leave }
+//         return(0, 0)
+//     }
+// }
--- a/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call_complex_control_flow.yul
+++ b/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call_complex_control_flow.yul
@@ -1,0 +1,60 @@
+{
+    function conditionallyStop() {
+        if calldataload(0) { leave }
+        return(0, 0)
+    }
+    function g() {
+        let a := 0
+        let b := 1
+        sstore(a, b)
+    }
+    let x := 0
+    let y := 1
+    let z := 2
+    switch calldataload(64)
+    case 0 {
+        sstore(z, x)
+        g()
+    }
+    default {
+        sstore(x, z) // used to be removed due to the StorageWriteRemovalBeforeConditionalTermination
+    }
+    conditionallyStop()
+    switch calldataload(32)
+    case 0 {
+        revert(0, 0)
+    }
+    default {
+        sstore(x, z)
+    }
+}
+// ----
+// step: unusedStoreEliminator
+//
+// {
+//     {
+//         let x := 0
+//         let y := 1
+//         let z := 2
+//         switch calldataload(64)
+//         case 0 {
+//             sstore(z, x)
+//             g()
+//         }
+//         default { sstore(x, z) }
+//         conditionallyStop()
+//         switch calldataload(32)
+//         case 0 { revert(0, 0) }
+//         default { sstore(x, z) }
+//     }
+//     function conditionallyStop()
+//     {
+//         if calldataload(0) { leave }
+//         return(0, 0)
+//     }
+//     function g()
+//     {
+//         let a := 0
+//         sstore(a, 1)
+//     }
+// }
--- a/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call_revert.yul
+++ b/test/libyul/yulOptimizerTests/unusedStoreEliminator/store_before_conditionally_terminating_function_call_revert.yul
@@ -1,0 +1,27 @@
+{
+    function conditionallyStop() {
+        if calldataload(0) { leave }
+        return(0, 0)
+    }
+    let x := 0
+    let y := 1
+    sstore(x, y) // used to be removed due to the to the StorageWriteRemovalBeforeConditionalTermination bug
+    conditionallyStop()
+    revert(0,0)
+}
+// ----
+// step: unusedStoreEliminator
+//
+// {
+//     {
+//         let x := 0
+//         sstore(x, 1)
+//         conditionallyStop()
+//         revert(0, 0)
+//     }
+//     function conditionallyStop()
+//     {
+//         if calldataload(0) { leave }
+//         return(0, 0)
+//     }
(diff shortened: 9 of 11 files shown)


test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
```diff
@@ -0,0 +1,15 @@
+contract C {
+	uint public x;
+	function f() public {
+		x = 1; // This write used to be removed by the yul optimizer.
+		g();
+		x = 1;
```
bshastry:
```suggestion
		x = 2;
```

bshastry: Just so that the buggy removal is obvious.

chriseth (commented):
Looks good!

bshastry (commented):
Minor comment.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -105,10 +105,13 @@ void UnusedStoreEliminator::operator()(FunctionCall const& _functionCall)
...
 		sideEffects = m_controlFlowSideEffects.at(_functionCall.functionName.name);
 
+	if (sideEffects.canTerminate)
+		changeUndecidedTo(State::Used, Location::Storage);
 	if (!sideEffects.canContinue)
 	{
 		changeUndecidedTo(State::Unused, Location::Memory);
-		changeUndecidedTo(sideEffects.canTerminate ? State::Used : State::Unused, Location::Storage);
+		if (!sideEffects.canTerminate)
+			changeUndecidedTo(State::Unused, Location::Storage);
```
bshastry, line 114: When is this branch hit exactly? i.e., can neither continue nor terminate

test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
bshastry: Nevermind. The assembly `return(0, 0)` implies `x = 0` so retaining the `1` is fine. I just felt the redundancy could be made more clear by writing a different value but I see that it is not required for understanding the bug.

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: ``revert(0,0)`` :-). "canTerminate" means "can terminate the transaction successfully, i.e. while keeping state".

If the function can't continue and can't terminate, it always reverts - then all state changes that are so far undecided will uncondtionally revert and can rightly be marked as "Unused" here.

bshastry: Is `canContinue` implied by `!canTerminate`? Or when can the transaction not continue if it is guaranteed to not terminate?

ekpyron:
```
canContinue == true && canTerminate == true && canRevert == false

function f() {
  if calldataload(0) { return(0,0) }
}
```

```
canContinue == true && canTerminate == false && canRevert == true

function f() {
  if calldataload(0) { revert(0,0) }
}
```


```
canContinue == false && canTerminate == false && canRevert == true

function f() {
  revert(0,0)
}
```

```
canContinue == false && canTerminate == true && canRevert == false

function f() {
  return(0,0)
}
```


```
canContinue == true && canTerminate == false && canRevert == false

function f() {
}
```

Does that make it clear :-)?

ekpyron: ``!canContinue && !canTerminate`` "mostly" implies ``canRevert``. It's defined in ``libyul/ControlFlowSideEffects.h``.
There's one exception in which all three are false - that's infinite recursion (which effectively also reverts).

docs/bugs.json
```diff
@@ -1,4 +1,17 @@
 [
+    {
+        "uid": "SOL-2022-7",
+        "name": "StorageWriteRemovalBeforeConditionalTermination",
+        "summary": "Calling Yul functions that conditionally terminate the transaction using ``return(...)`` or ``stop()`` may result in incorrect removals of storage writes.",
+        "description": "A call to a Yul function that conditionally terminates the transaction did not correctly mark all prior changes to storage as 'used', which may result in storage writes to be incorrectly removed. This happens in cases in which it had been valid to remove the store, if the Yul function in question never actually terminated the transaction. Conditional termination of the transaction within the same Yul function is not affected. In Solidity with via-IR code generation, any state change before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may be incorrectly removed, whenever it had been valid to remove the change without the ``return(...)`` or ``stop()``. Since the bug only occurs on calls to Yul functions, inlining may prevent the bug in some cases.",
```
```suggestion
        "description": "A call to a Yul function that conditionally terminates the transaction did not correctly mark all prior changes to storage as *used*, which may result in storage writes being incorrectly removed. This happens in cases in which it would have been valid to remove the store, if the Yul function in question never actually terminated the transaction. Conditional termination of the transaction within the same Yul function is not affected. In Solidity with via-IR code generation, any state change before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may be incorrectly removed, whenever it would have been valid to remove the change without the ``return(...)`` or ``stop()``. Since the bug only occurs on calls to Yul functions, inlining may prevent the bug in some cases.",
```

cameel: Maybe we should add some context on how the step works? What do you think about adding this before your text?

```
Yul optimizer has a step that detects and removes storage operations which write values that get overwritten before they are ever read. The step tracks all storage writes performed within a Yul function definition or the top-level Yul block and marks each of them as *used*, *unused* or *undecided*. When the control flow jumps to a different function, *undecided* operations should be changed to *used* unless the function is known to always revert. After the analysis, all operations storing unused values can be safely removed without altering the behavior of the contract. For the purpose of the analysis, for each function call it has to be determined whether it can revert, successfully terminate the whole transaction or return back to the caller and continue. More than one of these can be true due to conditionals.
```

cameel: Maybe we should add more context on how the optimizer step works?
```suggestion
        "description": "Yul optimizer has a step that detects and removes storage operations which write values that get overwritten before they are ever read. The step tracks all storage writes performed within a Yul function definition or the top-level Yul block and marks each of them as *used*, *unused* or *undecided*. When the control flow jumps to a different function, *undecided* operations should be changed to *used* unless the function is known to always revert. After the analysis, all operations storing unused values can be safely removed without altering the behavior of the contract. For the purpose of the analysis, for each function call it has to be determined whether it can revert, successfully terminate the whole transaction or return back to the caller and continue. More than one of these can be true due to conditionals. A call to a Yul function that conditionally terminates the transaction did not correctly mark all prior changes to storage as 'used', which may result in storage writes to be incorrectly removed. This happens in cases in which it had been valid to remove the store, if the Yul function in question never actually terminated the transaction. Conditional termination of the transaction within the same Yul function is not affected. In Solidity with via-IR code generation, any state change before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may be incorrectly removed, whenever it had been valid to remove the change without the ``return(...)`` or ``stop()``. Since the bug only occurs on calls to Yul functions, inlining may prevent the bug in some cases.",
```
Note that here I'm just adding text before your description, not modifying it.

libyul/optimiser/UnusedStoreEliminator.cpp
```diff
@@ -105,10 +105,13 @@ void UnusedStoreEliminator::operator()(FunctionCall const& _functionCall)
 	else
 		sideEffects = m_controlFlowSideEffects.at(_functionCall.functionName.name);
 
+	if (sideEffects.canTerminate)
+		changeUndecidedTo(State::Used, Location::Storage);
```
cameel, line 109: So this now catches the `canContinue && canTerminate` case that was not handled before. I'm not sure I understand why in this case we were ending up with unused. I see in the file that there are only two situations in which we mark a storage write unused:
1. The `!sideEffects.canTerminate` case below.
2. `knownCovered()` in `applyOperation()`.

And I don't think that it was the second.

Were we just leaving it undecided until we encounter another function call? But then what if that's the only function call?

test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
```diff
@@ -0,0 +1,15 @@
...
+	}
+	function g() internal {
+		if (msg.data.length > 4) return;
+		assembly { return(0,0) }
```
cameel:
```suggestion
		assembly { return(0, 0) }
```

test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
cameel: I think it would still be a bit clearer to make it `2` so that it's completely obvious which writes were executed.

test/libsolidity/semanticTests/unused_store_storage_removal_bug.sol
```diff
@@ -0,0 +1,15 @@
+contract C {
+	uint public x;
+	function f() public {
+		x = 1; // This write used to be removed by the yul optimizer.
```
cameel:
```suggestion
		x = 1; // This write used to be removed by the yul optimizer due to StorageWriteRemovalBeforeConditionalTermination bug.
```
And I'd put that in the comment in the other two test cases too.

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_3.yul
```diff
@@ -0,0 +1,28 @@
+{
+    function conditionallyStop() {
+        if calldataload(0) { leave }
+        return(0, 0)
+    }
+    let x := 0
+    let y := 1
+    sstore(x, y) // used to be removed due to a bug
```
cameel: Could you also add an opposite case that shows that the line does get removed when the termination is unconditional?

libyul/optimiser/UnusedStoreEliminator.cpp
ekpyron: It just remained undecided. Let me track the state of the write ``x = 1`` in the following, maybe that helps:
```
x = 1;
// undecided
f();
// still undecided (buggy! since f() may successfully terminate,
// it should already be marked as *used* after this call!)
x = 2;
// Now it changed from undecided to unused - the value was overwritten,
// we don't need the write above. That's of course wrong if ``f()`` can terminate,
// but that's what was missing - we assume that control flow *always* continues after ``f()``.
...
function f() { if (<sth>) { return; } assembly { return(0,0) } }
```
So the problem is that the call to ``f`` *doesn't* fix the write as "used", but leaves it undecided - not that it itself marks it as "unused". If the write is left undecided, a subsequent write may overwrite it (``knownCovered``) and change it to ``unused``, even though it should already have been fixed as used.

ekpyron: Probably the same will happen with
```
x = 1;
// undecided
f();
// still undecided
revert();
// unused, since we assume the transaction always reverts anyways
```

That may actually even be the more common and more dangerous case - we'll need to check that. But yeah, then the ``!canContinue && !canTerminate`` should kick in and mark it as unused, once we hit the ``revert();``

docs/bugs.json
ekpyron: Maybe that's rather for the blog post to explain in more detail? I'd keep the description here as concise as possible, while conveying which cases are affected as precisely as possible... but yeah, maybe you're right that it helps for the understanding of the situation here to explain it briefly as well...

cameel: Well, it's just that I don't expect people using this list to be that familiar with internals so I think we should either add some necessary context to understand it or make it even simpler and redirect people to the blog post.

Or maybe it would be better if these descriptions focused on how to tell if you are affected by the bug and what bad can happen and leave the details completely to the blog post. That's the info I'd be interested in if I were browsing the list. I would not expect to be able to understand what's happening just from this.

libyul/optimiser/UnusedStoreEliminator.cpp
cameel: Thanks.

I was worried about the `canContinue && canRevert` case, but looks like that's not a problem after all. On the other hand yours with `revert()` does look bad.

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_3.yul
ekpyron: It only *should* get removed, if the function unconditionally *doesn't* return... I can add that as test, though...

docs/bugs.json
ekpyron: "before they are ever read" doesn't really fit for the storage case, I think... it's rather "if they can never remain visible after the end of the current EVM call" or so...

docs/bugs.json
```diff
@@ -1,4 +1,17 @@
...
+        "name": "StorageWriteRemovalBeforeConditionalTermination",
+        "summary": "Calling Yul functions that conditionally terminate the external EVM call using ``return(...)`` or ``stop()`` may result in incorrect removals of storage writes.",
+        "description": "A call to a Yul function that conditionally terminates the external EVM call did not correctly mark all prior changes to storage as *used*, which may result in storage writes being incorrectly removed. This happens in cases in which it would have been valid to remove the store, if the Yul function in question never actually terminated, but control flow always returned back to the caller. Conditional termination within the same Yul block is not affected. In Solidity with via-IR code generation, any state change before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may be incorrectly removed, whenever it would have been valid to remove the change without the ``return(...)`` or ``stop()``. In legacy code generation, only inline assembly that does not refer any Solidity variables and that involves conditionally-terminating user-defined assembly functions can be affected.",
```
cameel: Past tense?
```suggestion
        "description": "A call to a Yul function that conditionally terminates the external EVM call did not correctly mark all prior changes to storage as *used*, which may result in storage writes being incorrectly removed. This used to happen in cases in which it would have been valid to remove the store, if the Yul function in question never actually terminated, and the control flow always returned back to the caller. Conditional termination within the same Yul block was not affected. In Solidity with via-IR code generation, any state change before a function conditionally calling ``return(...)`` or ``stop()`` in inline assembly, may have been incorrectly removed, whenever it would have been valid to remove the change without the ``return(...)`` or ``stop()``. In legacy code generation, only inline assembly that did not refer to any Solidity variables and that involved conditionally-terminating user-defined assembly functions could be affected."
```

docs/bugs.json
```diff
@@ -1,4 +1,17 @@
...
+        "link": "TODO",
```
```suggestion
        "link": "https://blog.soliditylang.org/2022/09/08/storage-write-removal-before-conditional-termination/",
```

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_5.yul
```diff
@@ -0,0 +1,60 @@
...
+       sstore(x,z) // used to be removed due to the StorageWriteRemovalBeforeConditionalTermination
+    }
+    conditionallyStop()
+    switch calldataload(32)
+    case 0 {
+      revert(0,0)
+    }
+    default {
+      sstore(x,z)
+    }
```
cameel:
```suggestion
    case 0 {
        sstore(z, x)
        g()
    }
    default {
        sstore(x, z) // used to be removed due to the StorageWriteRemovalBeforeConditionalTermination
    }
    conditionallyStop()
    switch calldataload(32)
    case 0 {
        revert(0, 0)
    }
    default {
        sstore(x, z)
    }
```

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_6.yul
```diff
@@ -0,0 +1,27 @@
+{
```
cameel: Would be nice to have more meaningful names for these tests now that their number is growing. Numbers are not very self-explanatory :)

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_3.yul
cameel: Ah right. ok then, probably not that important. We have many test cases already.

cameel:
Overall the PR seems to be ready in terms of functionality and testing. Just needs some minor tweaks.

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_6.yul
ekpyron: In my defense, it wasn't me who started with those numbers, they were already there :-). But yeah, I can try to think of some better names tomorrow morning :-).

ekpyron:
> Overall the PR seems to be ready in terms of functionality and testing. Just needs some minor tweaks.

Not entirely sure about the bug list description yet myself - might try to improve that a bit further tomorrow morning as well.

libyul/optimiser/UnusedStoreEliminator.cpp
bshastry: I summarized the expected behavior for (C)ontinue, (R)evert, (T)erminate, and (I)nfinite recursion (function summary between storage writes to same slot) from the discussion so far, as follows: (0 = false, 1 = true; `?` means I'm not sure). Is this a reasonable summary?

| C  | R | T | I | Outcome
| ------------- | ------------- | --- | ---- | --- |
| 0  | 0  | 0 | 0 | Impossible |
| 0  | 0  | 0 | 1 | Both stores removed |
| 0  | 0  | 1 | 0 | Second store removed, first retained |
| 0  | 0  | 1 | 1 | Second store removed, first retained |
| 0  | 1  | 0 | 0 | Both stores removed|
| 0  | 1 | 0 | 1 | Both stores removed|
| 0  | 1  | 1 | 0 | Second store removed |
| 0  | 1  | 1 | 1 | Second store removed |
| 1  | 0  | 0 | 0 | First store removed |
| 1  | 0  | 0 | 1 | First store removed |
| 1 | 0  | 1 | 0 | Neither store removed |
| 1  | 0  | 1 | 1 | Neither store removed |
| 1  | 1  | 0 | 0 | First store removed |
| 1  | 1  | 0 | 1 | First store removed |
| 1  | 1  | 1 | 0 | Neither store removed |
| 1  | 1  | 1 | 1 | Neither store removed |

test/libyul/yulOptimizerTests/unusedStoreEliminator/function_side_effects_6.yul
cameel: Yeah, I know. Just following an existing convention is always acceptable. It's just not a great convention here so I thought we should change it anyway :)

cameel (approved):
> Not entirely sure about the bug list description yet myself - might try to improve that a bit further tomorrow morning as well.

Well, I think it's good enough for what it is. I think we should maybe change the way we write these descriptions overall like I suggested in https://github.com/ethereum/solidity/pull/13479#discussion_r964118704, but looks like there's no time left for that before this release.

PR #13485, opened 2022-09-05
Pin hardhat version in GP2 external tests

nikola-matic:
New hardhat release breaks the GP2 tests as follows:

```
bench/trace/gas.ts:76:38 - error TS2339: Property 'addn' does not exist on type 'bigint'.

76         cumulativeGas: trace.gasUsed.addn(transactionGas),
                                        ~~~~
```
Since [GP2 repository](https://github.com/gnosis/gp-v2-contracts.git) has been archived, I'm pinning the hardhat version to the previous stable one.

### DIFF
--- a/test/externalTests/gp2.sh
+++ b/test/externalTests/gp2.sh
@@ -69,6 +69,8 @@ function gp2_test
     force_hardhat_unlimited_contract_size "$config_file" "$config_var"
     npm install
 
+    npm install hardhat@2.10.2
+
     # Some dependencies come with pre-built artifacts. We want to build from scratch.
     rm -r node_modules/@gnosis.pm/safe-contracts/build/
 


ekpyron (commented):
"Soft-approval" in the sense of that we can merge this if it fixes the non-required gp2 external test run.

ekpyron (approved):
Ah, the run already passed. Approving then.

test/externalTests/gp2.sh
```diff
@@ -69,6 +69,8 @@ function gp2_test
     force_hardhat_unlimited_contract_size "$config_file" "$config_var"
     npm install
 
+    npm install hardhat@2.10.2
```
cameel, line 72: This should really have some comment on why we're doing this.

nikola-matic: Lol, knew I forgot something, even described the issue in the opening PR comment. Will update tomorrow.

PR #13488, opened 2022-09-06
PeepholeOptimizer: TruthyBooleanComparison

### LINKED ISSUE #13442: Missing optimizer rules for `if (b == true)` to `if (b)`
Here's the example: https://godbolt.org/z/Mbs9bd5nr

Similarly for `if (b == false)`  to `if(!b)`

reubenr0d:
Closes #13442 

TODO:
- [x] if(a == true) -> if (a) => iszero iszero push1 eq -> ()
- [x]  if(true == a) -> if (a)  => iszero iszero push1 eq -> ()
- [x] if(a == false) -> if (!a) &&  if(false == a) -> if (!a) - already optimized
- [x] if(a != true) -> if(!a) => push1 swap1 iszero iszero sub pushtag jumpi -> iszero pushtag jumpi
- [x] if(true != a) -> if(!a) => iszero iszero push1 sub pushtag jumpi -> iszero pushtag jumpi
- [ ] a == true -> a => complex
- [ ] a != true -> !a =>  complex
- [ ] if (a.send(1) == true) -> if(a.send(1)) => complex
- [x] unit tests
- [ ] Standard JSON tests
- [ ] Update changelog

### DIFF
--- a/libevmasm/PeepholeOptimiser.cpp
+++ b/libevmasm/PeepholeOptimiser.cpp
@@ -428,6 +428,92 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
 	}
 };
 
+// if(a == true) -> if (a)
+struct TruthyEqualComparison: SimplePeepholeOptimizerMethod<TruthyEqualComparison>{
+	static bool applySimple(
+		AssemblyItem const& _iszero,
+		AssemblyItem const& _iszero1,
+		AssemblyItem const& _push1,
+		AssemblyItem const& _eq,
+		std::back_insert_iterator<AssemblyItems>
+	)
+	{
+		if (
+			_iszero == Instruction::ISZERO &&
+			_iszero1 == Instruction::ISZERO &&
+			_push1.type() == Push && _push1.data() == 1 &&
+			_eq == Instruction::EQ
+		)
+		{
+			return true;
+		} else
+			return false;
+	}
+};
+
+// if(a != true) -> if(!a)
+struct TruthyInequalComparison: SimplePeepholeOptimizerMethod<TruthyInequalComparison>{
+	static bool applySimple(
+		AssemblyItem const& _iszero,
+		AssemblyItem const& _iszero1,
+		AssemblyItem const& _push1,
+		AssemblyItem const& _sub,
+		AssemblyItem const& _pushTag,
+		AssemblyItem const& _jumpi,
+		std::back_insert_iterator<AssemblyItems> _out
+	)
+	{
+		if (
+			_iszero == Instruction::ISZERO &&
+			_iszero1 == Instruction::ISZERO &&
+			_push1.type() == Push && _push1.data() == 1 &&
+			_sub == Instruction::SUB &&
+			_pushTag.type() == PushTag &&
+			_jumpi == Instruction::JUMPI
+		)
+		{
+			*_out = _iszero;
+			*_out = _pushTag;
+			*_out = _jumpi;
+			return true;
+		} else
+			return false;
+	}
+};
+
+// if(true != a) -> if(!a)
+struct ReverseTruthyInequalComparison: SimplePeepholeOptimizerMethod<ReverseTruthyInequalComparison>{
+	static bool applySimple(
+		AssemblyItem const& _push1,
+		AssemblyItem const& _swap1,
+		AssemblyItem const& _iszero,
+		AssemblyItem const& _iszero1,
+		AssemblyItem const& _sub,
+		AssemblyItem const& _pushTag,
+		AssemblyItem const& _jumpi,
+		std::back_insert_iterator<AssemblyItems> _out
+	)
+	{
+		if (
+			_push1.type() == Push && _push1.data() == 1 &&
+			SemanticInformation::isSwapInstruction(_swap1) &&
+			getSwapNumber(_swap1.instruction()) == 1 &&
+			_iszero == Instruction::ISZERO &&
+			_iszero1 == Instruction::ISZERO &&
+			_sub == Instruction::SUB &&
+			_pushTag.type() == PushTag &&
+			_jumpi == Instruction::JUMPI
+		)
+		{
+			*_out = _iszero;
+			*_out = _pushTag;
+			*_out = _jumpi;
+			return true;
+		} else
+			return false;
+	}
+};
+
 /// Removes everything after a JUMP (or similar) until the next JUMPDEST.
 struct UnreachableCode
 {
@@ -488,7 +574,7 @@ bool PeepholeOptimiser::optimise()
 	while (state.i < m_items.size())
 		applyMethods(
 			state,
-			PushPop(), OpPop(), OpStop(), OpReturnRevert(), DoublePush(), DoubleSwap(), CommutativeSwap(), SwapComparison(),
+			TruthyEqualComparison(), TruthyInequalComparison(), ReverseTruthyInequalComparison(), PushPop(), OpPop(), OpStop(), OpReturnRevert(), DoublePush(), DoubleSwap(), CommutativeSwap(), SwapComparison(),
 			DupSwap(), IsZeroIsZeroJumpI(), EqIsZeroJumpI(), DoubleJump(), JumpToNext(), UnreachableCode(),
 			TagConjunctions(), TruthyAnd(), Identity()
 		);
--- a/test/libevmasm/Optimiser.cpp
+++ b/test/libevmasm/Optimiser.cpp
@@ -975,6 +975,70 @@ BOOST_AUTO_TEST_CASE(block_deduplicator_loops)
 	BOOST_CHECK_EQUAL(pushTags.size(), 1);
 }
 
+BOOST_AUTO_TEST_CASE(peephole_truthy_equal_comparison)
+{
+	AssemblyItems items{
+		Instruction::ISZERO,
+		Instruction::ISZERO,
+		AssemblyItem(Push, 1),
+		Instruction::EQ,
+	};
+	AssemblyItems expectation{};
+	PeepholeOptimiser peepOpt(items);
+	BOOST_REQUIRE(peepOpt.optimise());
+	BOOST_CHECK_EQUAL_COLLECTIONS(
+		items.begin(), items.end(),
+		expectation.begin(), expectation.end()
+	);
+}
+
+BOOST_AUTO_TEST_CASE(peephole_truthy_inequal_comparison)
+{
+	AssemblyItems items{
+		Instruction::ISZERO,
+		Instruction::ISZERO,
+		AssemblyItem(Push, 1),
+		Instruction::SUB,
+		AssemblyItem(PushTag, 1),
+		Instruction::JUMPI
+	};
+	AssemblyItems expectation{
+		Instruction::ISZERO,
+		AssemblyItem(PushTag, 1),
+		Instruction::JUMPI
+	};
+	PeepholeOptimiser peepOpt(items);
+	BOOST_REQUIRE(peepOpt.optimise());
+	BOOST_CHECK_EQUAL_COLLECTIONS(
+		items.begin(), items.end(),
+		expectation.begin(), expectation.end()
+	);
+}
+
+BOOST_AUTO_TEST_CASE(peephole_reverse_truthy_inequal_comparison)
+{
+	AssemblyItems items{
+		AssemblyItem(Push, 1),
+		Instruction::SWAP1,
+		Instruction::ISZERO,
+		Instruction::ISZERO,
+		Instruction::SUB,
+		AssemblyItem(PushTag, 1),
+		Instruction::JUMPI
+	};
+	AssemblyItems expectation{
+		Instruction::ISZERO,
+		AssemblyItem(PushTag, 1),
+		Instruction::JUMPI
+	};
+	PeepholeOptimiser peepOpt(items);
+	BOOST_REQUIRE(peepOpt.optimise());
+	BOOST_CHECK_EQUAL_COLLECTIONS(
+		items.begin(), items.end(),
+		expectation.begin(), expectation.end()
+	);
+}
+
 BOOST_AUTO_TEST_CASE(clear_unreachable_code)
 {
 	AssemblyItems items{


nikola-matic:
Consider adding a `cmdLineTest` or two, e.g. https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/dup_opt_peephole. Ideally we would have two tests - one with, and one without peephole optimization, but peephole optimization can only be turned off via the JSON interface, as we have no command supported to do so via command line.

You'll also need to add an entry to `Changelog.md`.

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -428,6 +428,36 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
 	}
 };
 
+// if(a == true) -> if(a)
```
nikola-matic: Will this work the other way around, i.e. `true == a`?

nikola-matic:
By the way, thanks for contributing with this PR! Also, feel free to either re-run the failed jobs (the `ext_*`) or even better, rebase against fresh `develop` from time to time; these are external tests that were broken by the recent hardhat release, and we're actively fixing them, i.e. they're not failing because of your changes :)

libevmasm/PeepholeOptimiser.cpp
reubenr0d: Oh good spot, this indeed won't work for `true == a`, also need to add similar rules for `false` checks. Thanks for the other suggestions as well. Will get these done, pushed the draft to make sure I'm headed in the right direction :)

nikola-matic: No worries. In any case, @ekpyron and @chriseth are experts in this domain, so they'll likely provide a more detailed review. Be patient though, since they're quite busy. Just make sure to rebase against fresh develop from time to time to keep the PR up to date.

test/cmdlineTests/truthy_bool_comparison_peephole/args
```diff
@@ -0,0 +1 @@
+--asm
```
nikola-matic: Gotta put newlines at the end of files. Before pushing, run `scripts/check_style.sh` to see if you need to make further changes.

reubenr0d:
One thing I noticed is that the equality is checked within an if condition is with `SUB` (since it expects 0 for `JUMPI`?), but if we simply do ` return a == true;` the equality here is checked with `EQ`. 

I guess that means we have to optimize for both these cases (https://github.com/ethereum/solidity/pull/13488/commits/922be5b834cfd157b68ea740da34cb9bd595652f), or am I misunderstanding something here? 

Also, not sure if analysing unoptimzed code and then running this rule first is the correct approach to this. Would like some guidance on the same.

test/cmdlineTests/truthy_bool_comparison_peephole/args
reubenr0d: Oh yea, did try to run it but was getting this, which I still need to look into:
``` 
$ scripts/check_style.sh 
grep: repetition-operator operand invalid

$ sysctl kern.version
kern.version: Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:41 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T6000
$ grep -V
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
```

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -428,6 +428,47 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
 	}
 };
 
+// (a == true) -> a and (true == a) -> a
+struct TruthyBooleanComparison: SimplePeepholeOptimizerMethod<TruthyBooleanComparison>{
+	static bool applySimple(
+		AssemblyItem const& _pushOrDupN,
+		AssemblyItem const& _iszero,
+		AssemblyItem const& _iszero1,
```
cameel: I think we already have a rule that turns this `PUSH 1 ISZERO ISZERO` into `PUSH 1` (not sure where but I see in the output that it does get replaced) so this should still work if you just expect `PUSH 1`. Same for `false` with `PUSH 0`. But please test that.

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -428,6 +428,47 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
...
+		AssemblyItem const& _pushOrDupN1,
+		AssemblyItem const& _iszero2,
+		AssemblyItem const& _iszero3,
+		AssemblyItem const& _sub,
```
cameel: Note that this catches only the simplest `a == true` case. Any even slightly more complex expression will slip through. E.g. `(1 + 1 == 2) == true` looks like this:
```
      0x01
      iszero
      iszero
      0x02
      dup1
      eq
      iszero
      iszero
      sub
      tag_11
      jumpi
```

Not sure if we can do anything about that. And maybe it's not worth it. Even `a == true` is a bit of an anti-pattern and I'd expect more complex expressions to be rare.

cameel, line 460: > One thing I noticed is that the equality is checked within an if condition is with `SUB` (since it expects 0 for `JUMPI`?), but if we simply do ` return a == true;` the equality here is checked with `EQ`. 
>
> I guess that means we have to optimize for both these cases, or am I misunderstanding something here?

The code generated for equality is `EQ` and for inequality `EQ ISZERO`: https://github.com/ethereum/solidity/blob/b4b5f7b355d46e8f286926d9b0da613ecc7a09dc/libsolidity/codegen/ExpressionCompiler.cpp#L2336-L2339

The thing is that the condition `JUMPI` checks is actually the `else` condition. I.e. we just continue past `JUMPI` when the condition is satisfied and we jump when it's not. So `EQ` for equality gets negated into `EQ ISZERO` and that gets turned into `SUB` by the `EqIsZeroJumpI` rule: https://github.com/ethereum/solidity/blob/b4b5f7b355d46e8f286926d9b0da613ecc7a09dc/libevmasm/PeepholeOptimiser.cpp#L304-L315

The rule is part of the peephole optimizer so I think you can just assume it will always get applied and only handle the code after its application. So not need to handle `EQ` separately if you only want to handle `if`.

On the other hand, it may still appear in other constructs. For example:
```solidity
    function id(bool a) external returns (bool) {
        return a == true;
    }
```
looks like this unoptimized:
```
      0x01
      iszero
      iszero
      dup3
      iszero
      iszero
      eq
      swap1
      pop
      swap2
      swap1
      pop
      jump
```

though currently it gets optimized into this eventually:
```
      iszero
      iszero
      0x01
      eq
      swap1
      jump
```

cameel: One of the more complex expressions which I could see actually happening is `send()`:
```solidity
    function id(address payable a) external returns (bool) {
        if (a.send(1) == true)
            return true;
        return false;
    }
```
But that looks way to complex for peephole optimizer;
```
      0x01
      iszero
      iszero
      dup3
      0xffffffffffffffffffffffffffffffffffffffff
      and
      0x08fc
      0x01
      swap1
      dup2
      iszero
      mul
      swap1
      mload(0x40)
      0x00
      mload(0x40)
      dup1
      dup4
      sub
      dup2
      dup6
      dup9
      dup9
      call
      swap4
      pop
      pop
      pop
      pop
      iszero
      iszero
      sub
      tag_12
      jumpi
```

test/libevmasm/Optimiser.cpp
```diff
@@ -975,6 +975,52 @@ BOOST_AUTO_TEST_CASE(block_deduplicator_loops)
...
+		Instruction::ISZERO,
+		Instruction::DUP3,
+		Instruction::ISZERO,
+		Instruction::ISZERO,
+		Instruction::SUB
+	};
+	AssemblyItems expectation{
+		Instruction::DUP2,
+		Instruction::ISZERO
+	};
```
cameel: One problem here is that these will not really give the same exact result. E.g. `1 0 SUB` is `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` but you'll never get that from `ISZERO`. The values are only equivalent when we know they're used as an input for some conditional instruction like `JUMPI`. So you must check that `JUMPI` is a part of the input sequence.

This is important because libevmasm is meant to work correctly with arbitrary asm code, not just the output the compiler produces. While you cannot just run arbitrary asm through the compiler right now, @aarlt is working on #11787 which will allow that.

test/cmdlineTests/truthy_bool_comparison_peephole/input.sol
```diff
@@ -0,0 +1,24 @@
...
+        return true;
+    }
+
+    function returnTrueCheck(bool a) external pure returns (bool) {
+       return a == true;
+    }
+
+    function returnReverseTrueCheck(bool a) external pure returns (bool) {
+       return true == a;
+    }
```
cameel: Maybe you could add the `a.send(1) == true` case I mentioned in another comment as well?

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -428,6 +428,51 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
...
+			if (
+				_pushOrDupN.type() == Push && _pushOrDupN.data() == 1 &&
+				SemanticInformation::isDupInstruction(_pushOrDupN1)
+			){
+				*_out = AssemblyItem(dupInstruction(getDupNumber(_pushOrDupN1.instruction()) - 1));
+			} else if (
+				SemanticInformation::isDupInstruction(_pushOrDupN) &&
+				_pushOrDupN1.type() == Push && _pushOrDupN1.data() == 1
+			){
+				*_out = _pushOrDupN;
```
cameel: Might be better to split it into two separate rules since the `true == a` case can be more generic. In that case `PUSH 1` is just before the jump so you could not assume anything about the other argument and this way be able to handle even complex expressions. I.e. you could catch this pattern:

```
      iszero
      iszero
      0x01
      sub
      tag_11
      jumpi
```

libevmasm/PeepholeOptimiser.cpp
```diff
@@ -428,6 +428,51 @@ struct TruthyAnd: SimplePeepholeOptimizerMethod<TruthyAnd>
...
+			if (_subOrEq == Instruction::SUB)
+			{
+				*_out = _iszero;
+				return true;
+			} else if (_subOrEq == Instruction::EQ)
+				return true;
+			else
+				return false;
+		} else
+			return false;
```
cameel: By the way, this is the style we use:

```suggestion
			if (_subOrEq == Instruction::SUB)
			{
				*_out = _iszero;
				return true;
			}
			else if (_subOrEq == Instruction::EQ)
				return true;
			else
				return false;
		} 
		else
			return false;
```

test/cmdlineTests/truthy_bool_comparison_peephole/args
cameel: Please also add a test with `--via-ir`. The godbolt example posted in the issue is actually using that. It uses a different codegen, which might produce different patterns.

cameel: As pointed out by @nikola-matic in https://github.com/ethereum/solidity/pull/13488#issuecomment-1237975229, you can only disable the peephole optimizer via Standard JSON. For that reason maybe it would be better to turn this into a Standard JSON test? See [`standard_optimizer_yul`](https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/standard_optimizer_yul) for an example of such a test. For your case you'd just need to add `evm.assembly` output (see [docs for Standard JSON input](https://docs.soliditylang.org/en/latest/using-the-compiler.html#input-description)).

cameel:
> Also, not sure if analysing unoptimzed code and then running this rule first is the correct approach to this. Would like some guidance on the same.

I you should look at the optimized code instead. Peephole optimizer is executed in `Assembly::optimiseInternal()` and it does not run just once: https://github.com/ethereum/solidity/blob/b4b5f7b355d46e8f286926d9b0da613ecc7a09dc/libevmasm/Assembly.cpp#L342-L483

It's executed multiple times until the output stabilizes. Also, if any other optimizer step changes the output, the whole sequence is reexecuted.

In case of `--via-ir` we also run the Yul optimizer, which is a separate thing. It runs before EVM asm optimizer and may alter the input so if you look at the unoptimized output, you might be targetting a sequence that the peephole optimizer will not actually see :)

test/cmdlineTests/truthy_bool_comparison_peephole/args
cameel: Hmm... maybe the script is broken on macOS. I'd expect someone to report it if that was the case but on the other hand we rarely run it locally. It's executed by CI anyway and once you get the hang of the style, it's rare to trigger it in practice.

You could report it as a separate issue and we'll fix it.

reubenr0d: https://github.com/ethereum/solidity/issues/13492

nikola-matic: Looks like this was broken for ages; left a comment on the issue.

reubenr0d:
> you should look at the optimized code instead

This was crucial, with that knowledge I've given it another shot. Still not sure if everything is entirely correct. Still learning how to decode opcodes :D

Also, thanks @cameel for leaving the very detailed explanations/reviews. They really helped.

nishant-sachdeva:
@reubenr0d Are you still working on this ? Or shall we takeover

cameel:
@nishant-sachdeva This was actually updated after my last review and is waiting for another review round from our side.

Now that we significantly trimmed our pile of PRs, I think I'll finally be able to get back to this after the Berlin on-site. But feel free to leave your review too.

reubenr0d:
Yes, that's correct. Still have to add tests, but wanted a quick review to make sure I'm on track. Keen on finishing this up :)

chriseth:
It would be nice to also have tests that show that the rules have the desired effect on gas usage. Furthermore, I think we should simplify some of them by moving at least part of the optimization to the RuleList.h. I commented in the original issue: https://github.com/ethereum/solidity/issues/13442

ekpyron:
We'd prefer doing this on the Yul level rather than in the peephole optimizer, so I'm closing the PR.

PR #13489, opened 2022-09-06
Pin hardhat version in yield-liquidator tests

nikola-matic:
See https://github.com/yieldprotocol/yield-liquidator-v2/issues/34


wechman (approved):
LGTM

PR #10996, opened 2021-02-23
Export all events

chriseth:
Fixes #9765.
Fixes #13086.

PR #13493, opened 2022-09-07
Abi emited events

### LINKED ISSUE #13086: Include all emitted events in the contract ABI and userdoc in a non-breaking way by ignoring conflicting Natspec

wechman:
closes #13086


test/cmdlineTests/recovery_ast_constructor/output
```diff
@@ -34,6 +34,7 @@ JSON AST (compact format):
       "canonicalName": "Error1",
       "contractDependencies": [],
       "contractKind": "contract",
+      "emittedEvents": [],
```
aarlt, line 37: Maybe I did not understand the semantic correctly. But I would have expected here one event for `emit L.e1(5);`

aarlt: Ah ok! Looks like that `emittedEvents` is just not yet implemented. I was not able to find any test that defines any `emittedEvents`. So I guess this PR is still in progress.

wechman:
I moved all changes to #10996.

PR #13499, opened 2022-09-08
Add workflow to label and comment stale issues and PRs

r0qs:
This PR address some of the concerns of the issue https://github.com/ethereum/solidity/issues/8969.

It adds a workflow to label an issue or PR as stale after 30 days of inactivity, comments on them, and if they remain inactive for more 7 days it closes them. If an update occurs, the label is removed, and the timer restarted. The action can also be prevented by using the label `keep`.

It is currently set for debugging only, and can be trigger manually in the workflow dashboard, so we can check the expected behavior in the workflow logs without affecting the current open issues and PRs.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,14 +16,18 @@ jobs:
       - uses: actions/stale@v5
         with:
           debug-only: true
-          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
+          stale-issue-message: |
+            This issue is stale because it has been open for 30 days with no activity.
+            It will be closed in 7 days unless the `stale` label is removed.
           stale-issue-label: stale
-          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
+          stale-pr-message: |
+            This pull request is stale because it has been open for 30 days with no activity.
+            It will be closed in 7 days unless the `stale` label is removed.
           stale-pr-label: stale
           days-before-stale: 30
           days-before-close: 7
           remove-stale-when-updated: true
-          exempt-issue-labels: 'keep'
+          exempt-issue-labels: 'roadmap,must have,must have eventually,should have,nice to have'
           exempt-draft-pr: true
+          exempt-all-milestones: true
           operations-per-run: 60
-
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,18 +16,18 @@ jobs:
       - uses: actions/stale@v5
         with:
           debug-only: true
-          stale-issue-message: |
-            This issue is stale because it has been open for 30 days with no activity.
-            It will be closed in 7 days unless the `stale` label is removed.
-          stale-issue-label: stale
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
           stale-pr-message: |
             This pull request is stale because it has been open for 30 days with no activity.
             It will be closed in 7 days unless the `stale` label is removed.
           stale-pr-label: stale
-          days-before-stale: 30
-          days-before-close: 7
-          remove-stale-when-updated: true
-          exempt-issue-labels: 'roadmap,must have,must have eventually,should have,nice to have'
+          days-before-pr-stale: 30
+          days-before-pr-close: 7
+          exempt-pr-labels: 'roadmap,must have,must have eventually,should have,nice to have,external contribution'
           exempt-draft-pr: true
           exempt-all-milestones: true
+          remove-stale-when-updated: true
           operations-per-run: 60
+          start-date: "2022-09-13"
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -5,6 +5,10 @@ on:
   schedule:
     - cron: '0 12 * * *'
 
+permissions:
+  issues: write
+  pull-requests: write
+
 jobs:
   stale:
     runs-on: ubuntu-latest
@@ -12,7 +16,6 @@ jobs:
       - uses: actions/stale@v5
         with:
           debug-only: true
-          repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
           stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
           stale-issue-label: stale
           stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,9 +13,9 @@ jobs:
         with:
           debug-only: true
           repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
-          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
           stale-issue-label: stale
-          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
           stale-pr-label: stale
           days-before-stale: 30
           days-before-close: 7
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: Check stale issues and pull requests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 12 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
+          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-issue-label: stale
+          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 12 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
```
cameel, line 16: I think we'd probably want `exempt-all-milestones`? If we assigned a milestone, we likely do not want it closed.

cameel, line 16: Is there a way to exempt untriaged issues? I don't think we want to autoclose issues we never even looked at.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          debug-only: true
+          repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
```
cameel: What permissions does this token need?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-issue-label: stale
+          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
```
cameel: Does it support markdown? This just gets pasted into a comment, right?
```suggestion
          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
          stale-issue-label: stale
          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the `stale` label is removed.'
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+          stale-pr-label: stale
+          days-before-stale: 30
+          days-before-close: 7
+          remove-stale-when-updated: true
+          exempt-issue-labels: 'keep'
```
cameel: How about exempting [desirability labels](https://notes.ethereum.org/@solidity/SkMG8szAq#Desirability) instead? I think if the issue has one, we don't want it closed. The issues that we want people (and us) to be bugged about are those that have already been triaged but now need a decision - either to close them or to place them somewhere on our roadmap.

What do you think @ekpyron?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
+          stale-issue-message: 'This issue is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-issue-label: stale
+          stale-pr-message: 'This pull request is stale because it has been open for 30 days with no activity. It will be closed in 7 days unless the stale label is removed.'
+          stale-pr-label: stale
+          days-before-stale: 30
```
cameel: When is it counted from on first run? Will the bot just mass-message almost all of existing issues when we run it for the first time? :)

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
+name: Check stale issues and pull requests
+
+on:
+  workflow_dispatch:
```
cameel, line 4: Are we going to be triggering it manually?

cameel (commented):
Here's some feedback from me.

Did you test the action in your fork?

.github/workflows/stale.yml
cameel: Ah, I see:

> It is currently set for debugging only, and can be trigger manually in the workflow dashboard, so we can check the expected behavior in the workflow logs without affecting the current open issues and PRs.

.github/workflows/stale.yml
r0qs: We can pass the option `any-of-labels` with the labels that we want to consider. So unlabeled issues/PR will never be checked.

.github/workflows/stale.yml
r0qs: Issues and pull-requests write permissions

.github/workflows/stale.yml
r0qs: Yes it does. Right.

.github/workflows/stale.yml
r0qs: Good question. I could not see mention of it in their docs. But yes, I suppose that is when you commit the config to your repo, so the workflow gets registered. As the PR is currently a draft, it will not appear there.

Honestly, I do not know how they keep the counter for each issue/PR and how efficient it is to run it for a big project. But it would be nice to find out, indeed. I know that they have an option to limit the number of operations: [operation-per-run](https://github.com/actions/stale#operations-per-run), so it can indeed dispatch a bunch of actions, and we need to limit it to our needs to not exceed the the github rate limit.

I do not know precisely how it will perform in a repository with many issues. However, we can test it safely while the debug is true. It will not label or close any issue or PR, just display in the workflow logs of what actions it would take and statistics.

Here is the output of a test that I run:
```
Starting the stale action process...
Processing the batch of issues  #1  containing  5  issues...
[#20] Pull request #20
[#9] Issue #9
[#8] Issue #8
[#4] Issue #4
[#2] Issue #2
Batch  #1  processed.
No more issues found to process. Exiting...
Statistics:
Processed items: 5
├── Processed issues: 4
└── Processed PRs   : 1
New stale PRs: 1
Added PRs labels: 1
Added PRs comments: 1
Fetched items: 5
Fetched items events: 5
Fetched items comments: 5
Fetched pull requests: 1
Operations performed: 15
```

r0qs:
> Here's some feedback from me.
> 
> Did you test the action in your fork?

No, I tested in a private repo only with a few issues/PR (around 20).

.github/workflows/stale.yml
ekpyron: Makes sense at least for now - if we eventually have too many ``nice to have`` issues going stale, we can still reconsider and change this :-).

.github/workflows/stale.yml
cameel: I guess might be good enough. We'd just need a rule not to partially label the issues. All or none.

.github/workflows/stale.yml
cameel: > I suppose that is when you commit the config to your repo, so the workflow gets registered.

That might be good but won't it just look at when an issue was last modified instead?

Well, I guess mass tagging as stale would not be bad as long as we are sure the rules are set up correctly and it would only touch issues we do consider stale. We definitely need that dry run. Too bad that log does not have issue titles. This will make it quite tedious to review.

cameel: The downside is that we'll need to wait with enabling it until we're done with labeling our issue backlog. Currently most of our issues do not have any desirability label and would be marked stale and then closed. Maybe we'd need some label to keep them open after all... or maybe we should have some minimum date?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+  schedule:
+    - cron: '0 12 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
```
cameel, line 10: Nice. Much better than having to use a custom token.

.github/workflows/stale.yml
cameel: Let's do this.

cameel:
We talked with @ekpyron about this PR today. Some preliminary decisions:
- We do want to do this for the PRs.
    - We'll exempt the `external contribution` label. These PRs often go unreviewed for some time and we don't want to drive away contributors because of that.
- For issues we don't want to enable it yet until we figure out how to exempt issues that already exist. Best if we can avoid having to put some label on all of them. Something based on creation date might be better.

.github/workflows/stale.yml
r0qs: > 
> That might be good but won't it just look at when an issue was last modified instead?
> 
> Well, I guess mass tagging as stale would not be bad as long as we are sure the rules are set up correctly and it would only touch issues we do consider stale. We definitely need that dry run. Too bad that log does not have issue titles. This will make it quite tedious to review.

Sorry, I guess I express myself wrong. The counter is per issue/PR; yes, it checks when they were last modified. I meant that the workflow would start the counters when we deploy it (changing it from draft to normal PR will already trigger the execution - but it is with debugging enabled, so no worries). Then it will display the statistics according to our configuration.
 
However, I could not understand one thing in their source code: where they store the counters. It seems they only keep the last states in memory and iterate over all issues/PR updating or removing the counter accordingly. But I guess this level of reliability is fine for this type of application. In the worst-case scenario, you would only wait for more to close something. But I have my doubts about the performance.

> Too bad that log does not have issue titles. This will make it quite tedious to review.

About having the issue titles, we could fork the original project and adapt it. But you will know the issues for sure when they get tagged. They are only not shown in the statistics. And we can have another automation to move tagged `stale` issues to some column in the board (like the icebox). However, the apps that I found available that can do this do not have a dry-run mode.

> The downside is that we'll need to wait with enabling it until we're done with labeling our issue backlog. Currently most of our issues do not have any desirability label and would be marked stale and then closed. Maybe we'd need some label to keep them open after all... or maybe we should have some minimum date?

I will enable it in the dry run, it will not harm, and we can already see some statistics. We can first use it only for some labeled tags (e.g., bug) and see how it plays out. For that, we can use the configuration `only-labels`.

r0qs:
> * For issues we don't want to enable it yet until we figure out how to exempt issues that already exist.

We can ignore them by using the option [start-date](https://github.com/actions/stale#start-date). So only dates after the starting date will count (i.e. new issues/PRs).

r0qs:
I initially had opened the PR from my fork. Please consider PR https://github.com/ethereum/solidity/pull/13522 instead.

.github/workflows/stale.yml
cameel: > It seems they only keep the last states in memory

Is it a github app? I thought it runs within a github action and in that case memory is not persistent.

> About having the issue titles, we could fork the original project and adapt it.

We shouldn't do that. We don't want the burden of having to maintain the fork. If anything submitting a PR would be much better.

> But you will know the issues for sure when they get tagged.

Well, the point was to see that before we allow them to get tagged.

Anyway, we can probably work around that somehow. For example I see that in the `#<number>` format so maybe just pasting the output into github comment preview would give us the titles...

PR #13500, opened 2022-09-08
Fix dynamic array spec

supermassive:
Fixes dynamic array spec. 
Elements should be encoded as tuple so everything after `enc(k)` should be same as for fixed size array above

### DIFF
--- a/docs/abi-spec.rst
+++ b/docs/abi-spec.rst
@@ -193,7 +193,7 @@ on the type of ``X`` being
 
   ``enc(X) = enc(k) enc((X[0], ..., X[k-1]))``
 
-  i.e. it is encoded as if it were an array of static size ``k``, prefixed with
+  i.e. it is encoded as if it were a tuple with ``k`` elements of the same type (resp. an array of static size ``k``), prefixed with
   the number of elements.
 
 - ``bytes``, of length ``k`` (which is assumed to be of type ``uint256``):
--- a/docs/abi-spec.rst
+++ b/docs/abi-spec.rst
@@ -191,7 +191,7 @@ on the type of ``X`` being
 
 - ``T[]`` where ``X`` has ``k`` elements (``k`` is assumed to be of type ``uint256``):
 
-  ``enc(X) = enc(k) enc([X[0], ..., X[k-1]])``
+  ``enc(X) = enc(k) enc(([X[0], ..., X[k-1]]))``
 
   i.e. it is encoded as if it were an array of static size ``k``, prefixed with
   the number of elements.
--- a/docs/abi-spec.rst
+++ b/docs/abi-spec.rst
@@ -191,7 +191,7 @@ on the type of ``X`` being
 
 - ``T[]`` where ``X`` has ``k`` elements (``k`` is assumed to be of type ``uint256``):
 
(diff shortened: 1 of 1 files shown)


docs/abi-spec.rst
```diff
@@ -191,7 +191,7 @@ on the type of ``X`` being
 
 - ``T[]`` where ``X`` has ``k`` elements (``k`` is assumed to be of type ``uint256``):
 
-  ``enc(X) = enc(k) enc([X[0], ..., X[k-1]])``
+  ``enc(X) = enc(k) enc(([X[0], ..., X[k-1]]))``
```
ekpyron: Without looking at it and thinking through it in all detail, I actually think the previous version is correct and the change *makes* this incorrect.
``[X[0], ..., X[k-1]]`` is meant to be a fixed size array, which will then recurse back to the previous clause, ultimately resulting in ``(X[0], ..., X[k-1])``.

A single-element tuple of a statically-sized array, i.e. the version after the change, can actually be encoded differently (e.g. in case the base type is dynamic).

ekpyron (changes requested):

docs/abi-spec.rst
ekpyron: It would be an option to change this to ``enc((X[0], ..., X[k-1]))`` (note the missing ``[ ]`` compared to the version in the PR).
But if we did that, we'd also need to change the comment below.

Here one could argue either way - to me it seems nicer to reduce dynamic arrays to static arrays first, and only then to tuples, rather than directly to tuples - but we don't actually formally introduce the ``[...]`` notation for static arrays, do we? If we don't formally introduce that notation, directly moving to a tuple may indeed be simpler.

docs/abi-spec.rst
```diff
@@ -191,7 +191,7 @@ on the type of ``X`` being
 
 - ``T[]`` where ``X`` has ``k`` elements (``k`` is assumed to be of type ``uint256``):
 
-  ``enc(X) = enc(k) enc([X[0], ..., X[k-1]])``
+  ``enc(X) = enc(k) enc((X[0], ..., X[k-1]))``
 
   i.e. it is encoded as if it were an array of static size ``k``, prefixed with
   the number of elements.
```
ekpyron: If we change the formula to a tuple, I'd suggest something like

```suggestion
  i.e. it is encoded as if it were a tuple with ``k`` elements of the same type (resp. an array of static size ``k``), prefixed with
  the number of elements.
```

here

docs/abi-spec.rst
supermassive: My bad. I meant it to match `enc((X[0], ..., X[k-1]))` above but somehow failed to remove `[]` brackets.

supermassive: @ekpyron PTAL

ekpyron (approved):
Looks good to me now, thank you very much!

supermassive:
@ekpyron what is the process of making PRs merged to dev branch? I have no write access.

ekpyron:
I was mainly waiting for CI to pass and whether anyone else wants to weigh in on this - but CI has passed and the change is good, so I'll merge now :-).

PR #13501, opened 2022-09-08
Document UnusedStoreEliminator

### LINKED ISSUE #13497: Document UnusedStoreEliminator
## Page

https://docs.soliditylang.org/en/latest/internals/optimizer.html#optimizer-steps
https://docs.soliditylang.org/en/latest/yul.html#optimization-step-sequence.

## Abstract

The `UnusedStoreEliminator` optimizer step is not listed in the documentation. Neither #12672 nor #12311 provided any docs for it.

When fixing, please also check if all the other [steps listed in `Suite.cpp`](https://github.com/ethereum/solidity/blob/develop/libyul/optimiser/Suite.cpp#L241-L287) are present in the docs.

nikola-matic:
Fixes #13497

- [x] Update code comment in `UnusedStoreEliminator.h`
- [x] Check other undocumented optimizer steps

### DIFF
--- a/docs/internals/optimizer.rst
+++ b/docs/internals/optimizer.rst
@@ -322,6 +322,7 @@ Abbreviation Full name
 ``a``        :ref:`SSA-transform`
 ``t``        :ref:`structural-simplifier`
 ``p``        :ref:`unused-function-parameter-pruner`
+``S``        :ref:`unused-store-eliminator`
 ``u``        :ref:`unused-pruner`
 ``d``        :ref:`var-decl-initializer`
 ============ ===============================
@@ -958,7 +959,7 @@ DeadCodeEliminator
 This optimization stage removes unreachable code.
 
 Unreachable code is any code within a block which is preceded by a
-leave, return, invalid, break, continue, selfdestruct or revert.
+leave, return, invalid, break, continue, selfdestruct, revert or by a call to a user-defined function that recurses infinitely.
 
 Function definitions are retained as they might be called by earlier
 code and thus are considered reachable.
@@ -1120,6 +1121,52 @@ Prerequisites: Disambiguator, FunctionHoister, LiteralRematerialiser.
 The step LiteralRematerialiser is not required for correctness. It helps deal with cases such as:
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
+
+.. index:: ! unused store eliminator
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes redundant ``sstore`` and memory store statements.
+In case of an ``sstore``, if all outgoing code paths revert (due to an explicit ``revert()``, ``invalid()``, or infinite recursion) or
+lead to another ``sstore`` for which the optimizer can tell that it will overwrite the first store, the statement will be removed.
+However, if there is a read operation between the initial ``sstore`` and the revert, or the overwriting ``sstore``, the statement
+will not be removed.
+Such read operations include: external calls, user-defined functions with any storage access, and ``sload`` of a slot that cannot be
+proven to differ from the slot written by the initial ``sstore``.
+
+For example, the following code
+
+.. code-block:: yul
+
+    {
+        let c := calldataload(0)
+        sstore(c, 1)
+        if c {
+            sstore(c, 2)
+        }
+        sstore(c, 3)
+    }
+
+will be transformed into the code below after the Unused Store Eliminator step is run
+
+.. code-block:: yul
+
+    {
+        let c := calldataload(0)
+        if c { }
+        sstore(c, 3)
+    }
+
+For memory store operations, things are generally simpler, at least in the outermost yul block as all such
+statements will be removed if they are never read from in any code path.
+At function analysis level, however, the approach is similar to ``sstore``, as we do not know whether the memory location will
+be read once we leave the function's scope, so the statement will be removed only if all code code paths lead to a memory overwrite.
+
+Best run in SSA form.
+
+Prerequisites: Disambiguator, ForLoopInitRewriter.
 
 .. _equivalent-function-combiner:
 
--- a/libyul/optimiser/UnusedStoreEliminator.h
+++ b/libyul/optimiser/UnusedStoreEliminator.h
@@ -39,8 +39,15 @@ struct Dialect;
 struct AssignedValue;
 
 /**
- * Optimizer component that removes sstore statements if they
- * are overwritten in all code paths or never read from.
+ * Optimizer component that removes sstore and memory store statements if conditions are met for their removal.
+ * In case of an sstore, if all outgoing code paths revert (due to an explicit revert(), invalid(),
+ * or infinite recursion) or lead to another ``sstore`` for which the optimizer can tell that it will overwrite the first store,
+ * the statement will be removed.
+ *
+ * For memory store operations, things are generally simpler, at least in the outermost yul block as all such statements
+ * will be removed if they are never read from in any code path. At function analysis level however, the approach is similar
+ * to sstore, as we don't know whether the memory location will be read once we leave the function's scope,
+ * so the statement will be removed only if all code code paths lead to a memory overwrite.
  *
  * The m_store member of UnusedStoreBase is only used with the empty yul string
  * as key in the first dimension.


docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,46 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes ``sstore`` and ``mstore`` statements if conditions are met for said removal.
+In case of ``sstore``, if all code paths lead to a revert, or if a previous ``sstore`` is overwritten,
+the statement will be removed. For example, the following code
```
ekpyron: Maybe rather something like
```suggestion
In case of an ``sstore``, if all outgoing code paths lead to a revert or to another ``sstore`` for which the optimizer can tell that it will overwrite the first store, the statement will be removed. For example, the following code
```
?
Not sure I'm following the "or if a previous ``sstore`` is overwritten" bit.

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,46 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+        {
+            let c := calldataload(0)
+            let _2 := 1
+            if c { let _3 := 2 }
+            sstore(c, 3)
+        }
+    }
+
+For ``mstore``, things are generally simpler, as all ``mstore`` statements will be removed if they are never
+read from in all code paths.
```
ekpyron: Yeah, but that's only half of the truth... that's true for ``mstore``s in the main outermost yul block - if you're analyzing a function body on the other hand, it works more like the storage case: then you don't know if the memory may be read, *after* returning from the function, so anything that's not overwritten in the function itself, will be retained.

Also it's not just ``mstore``s, but any memory-writing operation like ``calldatacopy`` as well for example (in which case you have to deal with arbitrary ranges in memory, for which "overwriting" or "not reading from" is actually rather hard to determine in general, but we don't have to get into that part in the docs anyways).

docs/internals/optimizer.rst
nikola-matic: Maybe it's because I wrote it, but it makes sense to me, especially with the example straight after it. I find your suggestion equally understandable, so am not against changing it.

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes ``sstore`` and ``mstore`` statements if conditions are met for said removal.
+In case of an ``sstore``, if all outgoing code paths lead to a revert or to another ``sstore`` for which the optimizer
```
cameel: We might want to point out that it does not necessarily have to be an explicit `revert()`. Infinite recursion is also treated as a revert (because it will eventually lead to out of gas).

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+    {
+        let c := calldataload(0)
+        sstore(c, 1)
+        if c {
+            sstore(c, 2)
+        }
+        sstore(c, 3)
+    }
+
+will produce the code below after the unused store eliminator step is run
```
cameel:
```suggestion
will be transformed into the code below after the Unused Store Eliminator step is run
```

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+.. code-block:: yul
+
+    {
+        {
+            let c := calldataload(0)
+            let _2 := 1
+            if c { let _3 := 2 }
+            sstore(c, 3)
+        }
+    }
```
cameel: I don't think the step actually add the extra block. I looks like the effect of some of the initial required steps. Same with the extra assignment. When I run just this step in `yulopti` I get this:
```yul
{
    let c := calldataload(0)
    if c { }
    sstore(c, 3)
}
```

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+.. code-block:: yul
+
+    {
+        let c := calldataload(0)
+        sstore(c, 1)
+        if c {
+            sstore(c, 2)
+        }
+        sstore(c, 3)
+    }
```
cameel, line 1150: How about adding another `sstore` that is not redundant?

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+        {
+            let c := calldataload(0)
+            let _2 := 1
+            if c { let _3 := 2 }
+            sstore(c, 3)
+        }
+    }
+
+For ``mstore``, things are generally simpler, at least in the outermost yul block as all ``mstore``
+statements will be removed if they are never read from in all code paths. At function analysis level however,
```
cameel:
```suggestion
statements will be removed if they are never read from in any code path. At function analysis level however,
```

cameel: It's best to break lines at the end of the sentence. RST will join lines anyway and this way you do not have to rewrap the whole paragraph when you modify just one sentence. Diffs are also cleaner:
```suggestion
statements will be removed if they are never read from in all code paths.
At function analysis level however, the approach is similar to ``sstore``, as we don't know
```

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes ``sstore`` and ``mstore`` statements if conditions are met for said removal.
```
cameel: For memory it's not just `mstore`. There are many opcodes that read and write memory: https://github.com/ethereum/solidity/blob/8df45f5f8632da4817bc7ceb81497518f298d290/libevmasm/SemanticInformation.cpp#L359-L392

docs/internals/optimizer.rst
```diff
@@ -1096,6 +1097,48 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
```
cameel, line 1128: I'd add an index entry for it. We should really have index entries for all of them but I guess that's out of scope of this PR.

docs/internals/optimizer.rst
nikola-matic: Should all of these be mentioned explicitly, or just use something like 'memory store' instead of `mstore`?

docs/internals/optimizer.rst
nikola-matic: I've copied this from `if_overwrite_all_branches.yul` test. I would assume that one only uses the `USE` step; does the same go for your `yulopti` run?

docs/internals/optimizer.rst
nikola-matic: This isn't mentioning an explicit `revert()` though, it's just saying revert, whatever the cause of said revert may be is kind of irrelevant here, no?

docs/internals/optimizer.rst
nikola-matic: How and where do I do that?

cameel: Check on other pages how the `index` directive is used. It's often next to headers.

Also see the "Keyword Index" section in the docs menu to see how the index looks like.

docs/internals/optimizer.rst
cameel: Probably the latter. We don't want to have to update the list. It just should not say only `mstore`.

docs/internals/optimizer.rst
cameel: That test probably runs multiple steps. `yulopti` lets you run steps one by one, interactively. It's a simple utility you can use in your local build. Check out [`yulopti.cpp`](https://github.com/ethereum/solidity/blob/develop/test/tools/yulopti.cpp).

docs/internals/optimizer.rst
cameel: I know, but it could be taken that way. I think it's not obvious that the optimizer would actually detect infinite recursion and treat it as a revert so I think it's worth mentioning.

nikola-matic: I disagree with this, as I think we'd have to go through the entire optimizer doc and mention infinite recursion detection and the subsequent `revert` due to it in every instance; I feel as though it's too much detail for something that should give a broad explanation of an optimizer step. @ekpyron @chriseth, any of you want to be the tie breaker?

docs/internals/optimizer.rst
nikola-matic: Another `sstore` that writes to `c`, or to a different variable?

docs/internals/optimizer.rst
nikola-matic: Yup, you're right. Thanks for pointing out `yulopti`.

docs/internals/optimizer.rst
nikola-matic: > It's best to break lines at the end of the sentence.
Sure, I'll re-work it, but I'm assuming this doesn't apply to longer sentences.

docs/internals/optimizer.rst
ekpyron: I mean, it doesn't say ``revert``, but revert :-D...
But yeah, if we want to go with @nikola-matic's argumentation I'd rather say ``if all outgoing code paths revert`` not ``lead to a revert``. 

But yeah, it then doesn't hurt to make it

```
In case of an ``sstore``, if all outgoing code paths revert (due to an explicit ``revert(...)`` or ``invalid()`` or due to infinite recursion) or lead to another ``sstore`` for which the optimizer
```

As for checking the other optimizer steps: we actually *should* do that :-). In ``DeadCodeEliminator`` the recursion case is, for example, also missing (and there ``revert`` is used in a list of statements, so clearly refers to the builtin call only)...

ekpyron: On the other hand I also think the ``DeadCodeEliminator`` is also the *only* case that needs adjustment, so it's not too bad.

docs/internals/optimizer.rst
nikola-matic: On the other hand, I think the examples should be as simple (minimal) as possible, so not sure what adding another non-redundant `sstore` brings, since the last `sstore(c, 3)` is already not redundant and is hence not removed after the step?

docs/internals/optimizer.rst
```diff
@@ -933,7 +934,7 @@ DeadCodeEliminator
 This optimization stage removes unreachable code.
 
 Unreachable code is any code within a block which is preceded by a
-leave, return, invalid, break, continue, selfdestruct or revert.
+leave, return, invalid, break, continue, selfdestruct or revert (whether it be an explicit ``revert()``, or caused by infinite recursion).
```
ekpyron: I'd actually not try to overload ``revert`` here. There's several builtin functions in the list already, so it's weird not to have this revert also refer to the builtin.

I'd just make it

```suggestion
leave, return, invalid, break, continue, selfdestruct or revert or by a call to a user-defined function that infinitely recurses.
```
or something like that

ekpyron: Also note that ``invalid()`` earlier in the list also is a case of a revert (we may also want to add that to the UnusedStoreEliminator docs)

nikola-matic: That's why I was more for the option of just using revert as a blanket statement, instead of mentioning specific reasons that a revert may occur. Aside from `invalid()` (which is explicit I guess?), is there anything else that may cause a revert?

ekpyron: You can reverse the point, though - since here you have an explicit list of statements and builtins causing reverts, "revert" specifically *doesn't* cover ``invalid`` nor infinite recursion. If ``revert`` was meant to cover it, ``invalid`` wouldn't be in the list.
But no, there's no other cases. Well, theoretically, infinite loops, but the optimizer doesn't handle them specially.

nikola-matic: Yes, I understand the point, and I do somewhat agree with it, but my understanding is that we have a blanket concept of a revert, which may be caused by explicit calls, such as `revert()` and `invalid()`, and by ones generated implicitly through an optimizer step, such as the infinite recursion case. The unfortunate thing here I guess, is the fact that a `revert()` is a subset of revert, so a `revert()` is a revert, but a revert isn't a `revert()` only :D 
And I guess it's a little late to introduce a term like rollback.

docs/internals/optimizer.rst
```diff
@@ -958,7 +959,7 @@ DeadCodeEliminator
 This optimization stage removes unreachable code.
 
 Unreachable code is any code within a block which is preceded by a
-leave, return, invalid, break, continue, selfdestruct or revert.
+leave, return, invalid, break, continue, selfdestruct, revert or by a call to a user-defined function that infinitely recurses.
```
ekpyron:
```suggestion
leave, return, invalid, break, continue, selfdestruct, revert or by a call to a user-defined function that recurses infinitely.
```
I may have suggested this in the first place, but the other way round is more idiomatic, isn't it? Not too important in any case (in the end it's a rather silly special case and technicality anyways).

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,46 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes ``sstore`` and memory store statements if conditions are met for said removal.
+In case of an ``sstore``, if all outgoing code paths revert (due to an explicit ``revert()``, ``invalid()``, or infinite recursion) or
+lead to another ``sstore`` for which the optimizer can tell that it will overwrite the first store, the statement will be removed.
```
ekpyron, line 1133: We should probably filll in a bit more detail here. It may be obvious, but between the initial ``sstore`` and the revert or the overwriting ``sstore`` there also cannot be any statement that could possibly read the value back that was initially written.

That includes external calls, user-defined functions with *any* storage access and ``sload``s of a slot that can't be proven to differ from the slot written by the initial ``sstore``.

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,46 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes ``sstore`` and memory store statements if conditions are met for said removal.
```
ekpyron:
```suggestion
Optimizer component that removes redundant ``sstore`` and memory store statements.
```

``if conditions are met`` is rather redundant itself, isn't it? It of course doesn't do it unconditionally for all of them :-).

ekpyron (dismissed):
Some minor comments, but we should especially be clearer about the fact that the written values possibly being read back of course prevents the removal (even though it's a bit obvious).

docs/internals/optimizer.rst
ekpyron: We could also phrase it more abstractly as something like ``removes statements with side-effects on storage or memory in cases in which it can prove that those side-effects can never be observed``... although I'm not sure about that - might make things worse...

ekpyron: Strictly speaking that'd be more accurate, though... one example of a weird side-effect of memory is that ``msize()`` can change if you remove memory stores, so the step actually also doesn't remove any memory store, if there is any ``msize()`` in the code... maybe we don't need to mention that detail specifically here, but yeah.

nikola-matic: I honestly prefer the first suggestion as `redundant` is explicit, and pretty simple to decipher for someone reading the docs, where as `statements with side-effects`, and `side-effects can never be observed` is a little too abstract for my tastes :)

docs/internals/optimizer.rst
nikola-matic: If it's a special case, I suggest even more special wording - `ad infinitum`.
Jokes aside, yeah, your suggestion does sound better.

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,51 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+.. _unused-store-eliminator:
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes redundant ``sstore`` and memory store statements.
+In case of an ``sstore``, if all outgoing code paths revert (due to an explicit ``revert()``, ``invalid()``, or infinite recursion) or
+lead to another ``sstore`` for which the optimizer can tell that it will overwrite the first store, the statement will be removed.
+However, if a there is read operation between the initial ``sstore`` and the revert,or the overwriting ``sstore``, the statement
+will not be removed.
```
ekpyron: Spacing:
```suggestion
However, if a there is read operation between the initial ``sstore`` and the revert, or the overwriting ``sstore``, the statement
will not be removed.
```

nikola-matic:
@ekpyron can you remove the change request so some one else can take over the review, as I'd rather not have this sit around for another couple of weeks :)

cameel:
You can just dismiss the review :)

But in any case, I'm now back from Berlin and catching up so I'm going to go back to this pretty soon.

docs/internals/optimizer.rst
cameel: For longer sentences I'd just break them in the middle (around 100 chars). The whole sentence is then still a complete unit that can be modified or removed without having to rewrap stuff around it.

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,51 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
 ``function f(x) -> y { revert(y, y} }`` where the literal ``y`` will be replaced by its value ``0``,
 allowing us to rewrite the function.
 
+.. _unused-store-eliminator:
```
cameel, line 1126:
```suggestion
.. index:: ! unused store eliminator
.. _unused-store-eliminator:
```

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,51 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+
+UnusedStoreEliminator
+^^^^^^^^^^^^^^^^^^^^^
+
+Optimizer component that removes redundant ``sstore`` and memory store statements.
+In case of an ``sstore``, if all outgoing code paths revert (due to an explicit ``revert()``, ``invalid()``, or infinite recursion) or
+lead to another ``sstore`` for which the optimizer can tell that it will overwrite the first store, the statement will be removed.
+However, if a there is read operation between the initial ``sstore`` and the revert, or the overwriting ``sstore``, the statement
```
cameel:
```suggestion
However, if a there is a read operation between the initial ``sstore`` and the revert, or the overwriting ``sstore``, the statement
```

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,51 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+will not be removed.
+Such a read operation includes external calls, user-defined functions with any storage access, and ``sload`` of a slot that can't be
```
cameel:
```suggestion
Such read operations include: external calls, user-defined functions with any storage access, and ``sload`` of a slot that can't be
```

cameel: The general style in the docs is not to use contractions.
```suggestion
Such a read operation includes external calls, user-defined functions with any storage access, and ``sload`` of a slot that cannot be
```

docs/internals/optimizer.rst
```diff
@@ -1121,6 +1122,51 @@ The step LiteralRematerialiser is not required for correctness. It helps deal wi
...
+
+    {
+        let c := calldataload(0)
+        if c { }
+        sstore(c, 3)
+    }
+
+For memory store operations, things are generally simpler, at least in the outermost yul block as all such
+statements will be removed if they are never read from in any code path.
+At function analysis level however, the approach is similar to ``sstore``, as we don't know whether the memory location will
```
cameel:
```suggestion
At function analysis level, however, the approach is similar to ``sstore``, as we do not know whether the memory location will
```

cameel (commented):
Looks good in terms of content. Just needs some wording tweaks and the index entry I mentioned.

cameel:
> Addresses #13497

Does it fix the issue completely?

nikola-matic:
> > Addresses #13497
> 
> Does it fix the issue completely?

I believe so, yes.

docs/internals/optimizer.rst
nikola-matic: Why though, as I've already put it up top in the `index` section?

docs/internals/optimizer.rst
nikola-matic: Can I suggest a change to your change :D 
`However, if a there is a read...`

cameel:
OK, changed to `Fixes #13497` to have github close it automatically.

docs/internals/optimizer.rst
cameel: Sure.... but isn't what you wrote identical to the initial part of my suggestion?

cameel: Ah wait. It is but just noticed that you probably meant that extra `a` I missed xD

docs/internals/optimizer.rst
cameel: Yeah, but `!` means it's the main entry for the topic (there can be many index entries for the same keyword).

Also, putting it at the top is not great IMO and I'd remove it. When you go to the index, you want to be taken to the section that's about your keyword, not the beginning of a big page that has that somewhere :)

nikola-matic: I'd prefer to do this in a separate PR (we can even open an issue for this), since none of anchors (or indices) in `optimizer.rst` use this approach.

cameel (dismissed):
Tentatively approving, but please squash the review fixes and I still think the index entry would be better off pointing at the section that is about the keyword, not the whole page

nikola-matic:
> Tentatively approving, but please squash the review fixes and I still think the index entry would be better off pointing at the section that is about the keyword, not the whole page

My issue with this is that there's no uniformity regarding this in `optimizer.rst` if I make this change - our current top level `index::` doesn't have the vast majority of the optimizer steps in there anyway, and there's not a single usage of a `!` main entry topic either. I do agree that your suggestion would make things better, but only if applied to the whole document. I'm gonna squash the commits, and open a separate issue for this.

cameel (dismissed):
`!` is used lot in other places so I don't think it's very out of line. And the index entries are a mess in general so I'm trying to at least remind about cleaning them up when docs are added or updated.

If you really want to make another PR to introduce proper index entries then fine. But otherwise, I still think that adding a proper entry here as a first step and just leaving it for later would be good enough, especially given that we want to focus on getting the PR count down.

cameel (approved):

PR #13510, opened 2022-09-08
Release checklist: recommend checking if all PRs have changelog entries

cameel:


ekpyron (approved):

ekpyron:
I'm setting this to auto-merge - but this means that we will need to release frequently enough, s.t. checking this actually remains feasible :-).

cameel:
I see it more as a recommendation. We can always skip it if it's too overwhelming and we should always make sure in the PRs that the entry is there but does not hurt to be reminded about it on the checklist.

PR #7508, opened 2019-10-02
Prevent build path from ending up in binaries.

ekpyron:
Where supported this change will replace the source path in ``__FILE__`` macros (implicitly used during our assertions) with ``/solidity``, whenever this is supported by the compiler.

Almost fixes, i.e. refs https://github.com/ethereum/solidity/issues/6230
In a stripped binary there's still some references to the source path remaining caused by json-cpp, which is built as an external subproject, so we'd need more involved checks for compiler support (resp. force a specific compiler for the subproject).

This is a step towards reproducible binaries and will reduce binary size.

Issue #6228, opened 2019-03-07
Reproducible Builds: Store details about the build environment in the compiler.


ekpyron:
We should store all relevant details about the build environment in solc and output it on ``--version`` (or with dedicated new flag; should also be exported in solc-js).
This should for example include:
- Compiler version.
- Boost version.
- For soljson.js Emscripten version.
- Z3 and CVC4 versions plus maybe even their dependencies CLN and GMP in turn.
- All compiler and linker flags affecting the build (i.e. ideally no warning flags)
- All other cmake flags/options affecting the build, if any.
- cmake version (might not be needed)
~~- CPU architecture (might not be needed)~~ (bad idea probably)
~~- OS, OS Version, distribution, if applicable (on linux might attempt to use ``lsb_release`` and to read ``/etc/os-release`` or otherwise report "unknown") (might not be needed)~~ (bad idea probably)
- ...

This would be a step towards reproducible binaries, would help in bug reports and might help us in creating better patched older versions in cases of stuff like the soljson.js-Firefox-rule-list-stack-height-drama.

chriseth:
Wouldn't most of the items make it harder to reproduce the binary? You have to find a matching system...

ekpyron:
With mismatching compiler and dependency versions or flags you will probably get a mismatching binary anyways - if the binary can tell you its build environment, then you at least have a chance to match it... but yes - it's something to think about. Compiler flags regarding warnings e.g. shouldn't affect the binary.

Another option would be to store all this information outside the binary in some metadata file - but if one does that, I'm pretty sure that nobody will keep this metadata file around in the end...

Maybe there's some trick to include this information in an elf section that can be disregarded when comparing binaries... probably too involved, though...

ekpyron:
We just got a Z3 assertion containing `` File: /build/libz3-static-dev-q0RqJn/libz3-static-dev-4.8.8/src/qe/qe_arith.cpp`` in CI - which looks like the static z3 PPA build chooses some weird build directory name and that ends up in our binaries in the end, so if we ever want really nicely reproducible builds, this will come up as a problem.

PR #13513, opened 2022-09-09
Run emscripten builder as current user

### LINKED ISSUE #13506: Nightly builds in solc-bin broken due to git's `safe.directory` check
Nightly builds in solc-bin seem to have been failing for 2 weeks now: [8239563561](https://github.com/ethereum/solc-bin/runs/8239563561).

```
fatal: detected dubious ownership in repository at '/root/project'
To add an exception for this directory, call:

	git config --global --add safe.directory /root/project
```

This seems to be due to the checks introduced recently by git to address a security vulnerability: [Git security vulnerability announced](https://github.blog/2022-04-12-git-security-vulnerability-announced/).

r0qs:
Fixes https://github.com/ethereum/solidity/issues/13506.

The PR changes the current behavior of running the `build_emscripten` as root to be run as the current user. It, in turn, also fixes the git error when traversing files with different ownership without the need to change the default [safe.directory](https://github.blog/2022-04-12-git-security-vulnerability-announced/) configuration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1016,15 +1016,15 @@ jobs:
       - store_artifacts: *artifacts_test_results
       - gitter_notify_failure_unless_pr
 
-  # NOTE: work around to git unsafe repository error by adding /root/project as a safe.directory
-  # See https://github.com/CircleCI-Public/cimg-base/issues/170
   b_ems:
     <<: *base_ems_large
     environment:
       TERM: xterm
       MAKEFLAGS: -j 10
     steps:
-      - run: git config --global --add safe.directory /root/project
+      # NOTE: work around to git unsafe repository error by adding /project as a safe.directory
+      # See https://github.com/CircleCI-Public/cimg-base/issues/170
+      - run: git config --global --add safe.directory /project
       - checkout
       - run:
           name: Build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1016,12 +1016,15 @@ jobs:
       - store_artifacts: *artifacts_test_results
       - gitter_notify_failure_unless_pr
 
+  # NOTE: work around to git unsafe repository error by adding /root/project as a safe.directory
+  # See https://github.com/CircleCI-Public/cimg-base/issues/170
   b_ems:
     <<: *base_ems_large
     environment:
       TERM: xterm
       MAKEFLAGS: -j 10
     steps:
+      - run: git config --global --add safe.directory /root/project
       - checkout
       - run:
           name: Build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,12 @@
 #     - ubu: ubuntu
 #     - ems: Emscripten
 version: 2.1
+
+aliases:
+  - &container_config
+    # default working_directory for all containers.
+    working_directory: &working_directory ~/project
+
 parameters:
   ubuntu-2004-docker-image:
     type: string
@@ -288,6 +294,7 @@ defaults:
   # Base Image Templates
 
   - base_ubuntu1604_clang: &base_ubuntu1604_clang
+      <<: *container_config
       docker:
         - image: << pipeline.parameters.ubuntu-1604-clang-ossfuzz-docker-image >>
       environment:
@@ -302,6 +309,7 @@ defaults:
         MAKEFLAGS: -j 2
 
   - base_ubuntu2004_clang: &base_ubuntu2004_clang
+      <<: *container_config
       docker:
         - image: << pipeline.parameters.ubuntu-2004-clang-docker-image >>
       environment:
@@ -329,6 +337,7 @@ defaults:
         MAKEFLAGS: -j 5
 
   - base_ubuntu2004: &base_ubuntu2004
+      <<: *container_config
       docker:
         - image: << pipeline.parameters.ubuntu-2004-docker-image >>
       environment:
@@ -357,6 +366,7 @@ defaults:
         MAKEFLAGS: -j 10
 
   - base_buildpack_focal_small: &base_buildpack_focal_small
+      <<: *container_config
       docker:
         - image: buildpack-deps:focal
       resource_class: small
@@ -365,6 +375,7 @@ defaults:
         MAKEFLAGS: -j 2
 
   - base_buildpack_latest_small: &base_buildpack_latest_small
+      <<: *container_config
       docker:
         - image: buildpack-deps:latest
       resource_class: small
@@ -373,6 +384,7 @@ defaults:
         MAKEFLAGS: -j 2
 
   - base_archlinux: &base_archlinux
+      <<: *container_config
       docker:
         - image: archlinux:base
       environment:
@@ -380,6 +392,7 @@ defaults:
         MAKEFLAGS: -j 3
 
   - base_archlinux_large: &base_archlinux_large
+      <<: *container_config
       docker:
         - image: archlinux:base
       resource_class: large
@@ -388,22 +401,26 @@ defaults:
         MAKEFLAGS: -j 5
 
   - base_win_powershell: &base_win_powershell
+      <<: *container_config
       executor:
         name: win/default
         shell: powershell.exe
 
   - base_win_powershell_large: &base_win_powershell_large
+      <<: *container_config
       executor:
         name: win/default
         shell: powershell.exe
         size: large
 
   - base_win_cmd: &base_win_cmd
+      <<: *container_config
       executor:
         name: win/default
         shell: cmd.exe
 
   - base_osx: &base_osx
+      <<: *container_config
       macos:
         xcode: "13.2.0"
       environment:
@@ -411,6 +428,7 @@ defaults:
         MAKEFLAGS: -j5
 
   - base_osx_large: &base_osx_large
+      <<: *container_config
       macos:
         xcode: "13.2.0"
       resource_class: large
@@ -419,6 +437,7 @@ defaults:
         MAKEFLAGS: -j10
 
   - base_ems_large: &base_ems_large
+      <<: *container_config
       docker:
         - image: << pipeline.parameters.emscripten-docker-image >>
       resource_class: large
@@ -427,6 +446,7 @@ defaults:
         MAKEFLAGS: -j 5
 
   - base_python_small: &base_python_small
+      <<: *container_config
       docker:
         - image: circleci/python:3.6
       resource_class: small
@@ -435,6 +455,7 @@ defaults:
         MAKEFLAGS: -j 2
 
   - base_node_small: &base_node_small
+      <<: *container_config
       docker:
         - image: circleci/node
       resource_class: small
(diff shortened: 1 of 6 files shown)


scripts/build_emscripten.sh
```diff
@@ -35,6 +35,7 @@ else
 fi
 
 # solbuildpackpusher/solidity-buildpack-deps:emscripten-13
-docker run -v "$(pwd):/root/project" -w /root/project \
+docker run -v "$(pwd):/emsdk/project" -w /emsdk/project \
+    --user "$(id -u):$(id -g)" \
```
ekpyron:
```suggestion
    --user "1000:1000" \
```
?

That's the user-id and group-id of the ``emscripten`` user in the docker image, so this way the user actually exists... and it's less prone to non-determinism (not sure there's any way the user/group ID could actually affect anything, but no need to introduce the potential for variance if there's no need).

scripts/ci/build_emscripten.sh
```diff
@@ -40,10 +40,7 @@ else
 	BUILD_DIR="$1"
 fi
 
-apt-get update
-apt-get install lz4 --no-install-recommends
-
-WORKSPACE=/root/project
+WORKSPACE=/emsdk/project
```
cameel: Tests are failing because of this. You probably just need to create the dir. Either inside or outside the image. Or is docker supposed to create it automatically?

scripts/build_emscripten.sh
```diff
@@ -35,6 +35,7 @@ else
 fi
 
 # solbuildpackpusher/solidity-buildpack-deps:emscripten-13
-docker run -v "$(pwd):/root/project" -w /root/project \
+docker run -v "$(pwd):/emsdk/project" -w /emsdk/project \
```
cameel: While at it, let's also switch to long options. We should prefer them for readability unless there are portability concerns - and with younger tools like docker I don't think there are any.
```suggestion
docker run --volume "$(pwd):/emsdk/project" --workdir /emsdk/project \
```

scripts/ci/build_emscripten.sh
r0qs: I thought that docker run would create the volume if it does not exists in the [container](https://docs.docker.com/storage/volumes/#start-a-container-with-a-volume). Maybe this script is being called directly by another script I did not see. I will check.

r0qs:
It seems that the "recommended" way to fix the issue in the CI is indeed to add the directory as safe in the git config: https://github.com/CircleCI-Public/cimg-base/issues/170#issuecomment-1134844852

So I added the option in the CI but kept the non-root configuration in the local build script to fix the permissions issues when building from the source.

ekpyron:
Have you checked if the resulting ``soljson.js`` after running ``scripts/build_emscripten.sh`` and the ``soljson.js`` CI artifact of the ``b_ems`` build on the same commit are still binary-identical with these changes?

.circleci/config.yml
```diff
@@ -1016,12 +1016,15 @@ jobs:
       - store_artifacts: *artifacts_test_results
       - gitter_notify_failure_unless_pr
 
+  # NOTE: work around to git unsafe repository error by adding /root/project as a safe.directory
+  # See https://github.com/CircleCI-Public/cimg-base/issues/170
```
cameel: Can you put it directly above the workaround? Otherwise it looks as if the whole job was the workaround :P

scripts/ci/build_emscripten.sh
```diff
@@ -40,12 +40,9 @@ else
 	BUILD_DIR="$1"
 fi
 
-apt-get update
-apt-get install lz4 --no-install-recommends
+WORKSPACE="${2:-/root/project}"
```
cameel: I'd just remove the default. The directory is pretty arbitrary so if we have to specify it for one case, I think it's cleaner to specify it in all cases.

r0qs: The `/root/project` folder is used by the CI jobs, so I tried to keep the consistency between the CI configs. But the CI runs as the root user in the containers. However, this is unnecessary locally and messes up with the permissions in the solidity folder when developing.
Maybe we could just specify `/project` for all then, and only the permissions will it be different depending on the environment.
Alternatively, we can run the CI with a default user (e.g., 1000), assuming all base images have a user or that we created a builder user for this case.

.circleci/config.yml
r0qs: > Have you checked if the resulting `soljson.js` after running `scripts/build_emscripten.sh` and the `soljson.js` CI artifact of the `b_ems` build on the same commit are still binary-identical with these changes?

No, but I just did it now, and they differ :0
```
diff -q soljson.js soljson-ci.js 
Files soljson.js and soljson-ci.js differ
```
Need to investigate where now, because the diff is pretty long.

r0qs:
> Have you checked if the resulting `soljson.js` after running `scripts/build_emscripten.sh` and the `soljson.js` CI artifact of the `b_ems` build on the same commit are still binary-identical with these changes?

No, but I just did it now, and they differ :0
```
diff -q soljson.js soljson-ci.js 
Files soljson.js and soljson-ci.js differ
```

```
Local build                                                   CI

return uncompress(base64DecToArr(source), uncompressedSize);}	return uncompress(base64DecToArr(source), uncompressedSize);}
"Q/gVAPAPAGFzbQEAAAABswyiAWABfwBgAX8Bf2ACf38AYAJ/CwAgA38MACED |	"8/cVAPAPAGFzbQEAAAABswyiAWABfwBgAX8Bf2ACf38AYAJ/CwAgA38MACED
20802184);var Module=typeof Module!="undefined"?Module:{};var |	20802489);var Module=typeof Module!="undefined"?Module:{};var
Module["wasmBinary"] = undefined;				Module["wasmBinary"] = undefined;
```

scripts/ci/build_emscripten.sh
r0qs: I changed everything to ~`/project`~ `$HOME/project` and I'll update the checkout step in the CI to use it as the working directory.

ekpyron:
For reference: this is the issue for moving towards reproducible builds: https://github.com/ethereum/solidity/issues/6228 - even though I thought it contained more helpful information on this :-). Previously https://github.com/ethereum/solidity/pull/7508 added the use of ``-fmacro-prefix-map`` with the hope that this'd catch most of the path dependencies, but apparently it didn't.

ekpyron:
@r0qs Actually, before I'm sending you on a wild goose-chase with the reproducibility... it may just be that the CI builds will use a different version string (e.g. ``0.8.18-ci.*`` vs ``0.8.18-develop.*``) - which can manually be overridden by providing a ``prerelease.txt`` file. So might make sense to check the ``solcjs --version`` output of the CI version and only compare against a local version with a matching prerelease.txt file...

ekpyron:
(I think in the past I mainly checked that builds are binary-identical on release builds that don't suffer from those version string mismatches)

r0qs:
Yeah, the binaries are still different in this build.

About the `prerelease.txt` file, the local `build_emscripten.sh` script calls the same build script used by the CI, which executes this line: https://github.com/ethereum/solidity/blob/develop/scripts/ci/build_emscripten.sh#L58 and set the `prerelease.txt` to `ci.2022.9.12`. So the solcjs version of both matches:
```
CI version: 0.8.18-ci.2022.9.12+commit.dc5d8f77.Emscripten.clang
Local version: 0.8.18-ci.2022.9.12+commit.dc5d8f77.Emscripten.clang
```

However, the binaries still differ in the first bytes:
```
Local build                                                   CI
return uncompress(base64DecToArr(source), uncompressedSize);}	return uncompress(base64DecToArr(source), uncompressedSize);}
"x/cVAPAPAGFzbQEAAAABswyiAWABfwBgAX8Bf2ACf38AYAJ/CwAgA38MACED |	"8PcVAPAPAGFzbQEAAAABswyiAWABfwBgAX8Bf2ACf38AYAJ/CwAgA38MACED
20803149);var Module=typeof Module!="undefined"?Module:{};var |	20802489);var Module=typeof Module!="undefined"?Module:{};var
Module["wasmBinary"] = undefined;				Module["wasmBinary"] = undefined;
```

ekpyron:
Yeah, ok, then it's probably easiest to just keep the build directories aligned between CI and the local docker builds (that way binaries *should* match). Although, eventually, we should try to figure out what exactly is still affected by the build directory.

r0qs:
> Yeah, ok, then it's probably easiest to just keep the build directories aligned between CI and the local docker builds (that way binaries _should_ match). Although, eventually, we should try to figure out what exactly is still affected by the build directory.

Right, so I will create another PR for https://github.com/ethereum/solidity/issues/13506, adding the `safe.directory` option and keeping the `/root/project`, until we figure out a solution. Should we close this PR? Because we still with the problem of local emscripten builds with wrong permissions.

Maybe we could use something like [rebuilderd](https://github.com/kpcyrd/rebuilderd) to monitor the reproducibility of our builds. The Archlinux distro uses it to verify if packages can be reproduced from the published source code: https://reproducible.archlinux.org/

ekpyron:
> Maybe we could use something like [rebuilderd](https://github.com/kpcyrd/rebuilderd) to monitor the reproducibility of our builds. The Archlinux distro uses it to verify if packages can be reproduced from the published source code: https://reproducible.archlinux.org/

Ah, nice! I didn't know that one. But yeah, might be worth looking into eventually - so far (as demonstrated by the issues we had here) our builds aren't *properly* reproducible, though, so I guess we should first get to a point in which this at least mostly works :-).

ekpyron:
> Right, so I will create another PR for https://github.com/ethereum/solidity/issues/13506, adding the safe.directory option and keeping the /root/project, until we figure out a solution. Should we close this PR? Because we still with the problem of local emscripten builds with wrong permissions.

Actually, this wouldn't be fixed by this PR ever since https://github.com/ethereum/solidity/pull/13513#discussion_r966996860, would it? Hadn't thought about that, but it's true, I've been annoyed by those build directories ending up owned by root in the past... I guess there's no way to have proper user-id mapping in cases like this? (I.e. map ``emscripten:emscripten``, resp. 1000:1000 in the docker image to the current user outside docker?) That'd be ideal... but in any case, while that's annoying and we should try to fix it eventually, I'd not put it high up on our agenda.

In general, I hate it that those docker builds have to be run as root, but afaik there's no *easy* way around that :-).

r0qs:
> Actually, this wouldn't be fixed by this PR ever since [#13513 (comment)](https://github.com/ethereum/solidity/pull/13513#discussion_r966996860), would it?

Yes, it would fix the git repository error without setting the `safe.directory` option since both the CI and the local build would use the user id, 1000, so git would detect no permission changes. However, it would indeed cause other permission problems because it would depend on the id and gid of the current users. The build directory would not be created as root for local builds, but it would have the permissions of the user id 1000 in the host machine, whatever the user is.

And there is also the reproducibility build problem caused by changing the building path.

> I guess there's no way to have proper user-id mapping in cases like this?

Correct. I am also not aware of a proper fix for that. Even if we create a builder user in the docker images, the permissions in the host machine would still differ.

r0qs:
Closing I am this in favor of https://github.com/ethereum/solidity/pull/13527. Feel free to reopen if needed.

PR #13514, opened 2022-09-09
Fix create2 memory access in yul interpreter.

bshastry:
Fixes bug in yul interpreter that assumed the offset and size arguments in the `create2` builtin are the last two parameters instead of the second and third parameters.


test/libyul/yulInterpreterTests/create2.yul
```diff
@@ -0,0 +1,16 @@
+{
+    pop(create2(0, 0, 1, 0))
+    sstore(0, msize())
+    pop(create2(0, 0, 2, 128))
```
bshastry: This would return `msize() = 160` before bug fix.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -317,7 +317,7 @@ u256 EVMInstructionInterpreter::eval(
 		logTrace(_instruction, arg);
 		return (0xcccccc + arg[1]) & u256("0xffffffffffffffffffffffffffffffffffffffff");
 	case Instruction::CREATE2:
-		accessMemory(arg[2], arg[3]);
+		accessMemory(arg[1], arg[2]);
```
axic, line 320: From [EIP-1014](https://eips.ethereum.org/EIPS/eip-1014):
> Adds a new opcode (CREATE2) at 0xf5, which takes 4 stack arguments: endowment, memory_start, memory_length, salt.

So `arg[1], arg[2]` seems correct.

test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -317,7 +317,7 @@ u256 EVMInstructionInterpreter::eval(
...
 		logTrace(_instruction, arg);
 		return (0xdddddd + arg[1]) & u256("0xffffffffffffffffffffffffffffffffffffffff");
```
axic, line 322: Shouldn't this use `arg[3]` aka salt? Or a both `arg[1] + arg[3]` for added resilience?

bshastry: Since the optimizer does not affect salt, status quo should be fine. Actually, the problem is zero-length ranges that will be addressed in an upcoming PR.

Marenz (approved):

PR #13516, opened 2022-09-12
Fix memory expansion semantics in the yul interpreter.

bshastry:
Earlier, zero length but non-zero offset range operations would expand memory. Now they don't.

### DIFF
--- a/test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
+++ b/test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
@@ -499,18 +499,29 @@ u256 EVMInstructionInterpreter::evalBuiltin(
 
 bool EVMInstructionInterpreter::accessMemory(u256 const& _offset, u256 const& _size)
 {
-	if (((_offset + _size) >= _offset) && ((_offset + _size + 0x1f) >= (_offset + _size)))
+	if (_size == 0)
+		return true;
+	else if (((_offset + _size) >= _offset) && ((_offset + _size + 0x1f) >= (_offset + _size)))
 	{
 		u256 newSize = (_offset + _size + 0x1f) & ~u256(0x1f);
 		m_state.msize = max(m_state.msize, newSize);
-		// We only record accesses to contiguous memory chunks that are at most 0xffff bytes
-		// in size and at an offset of at most numeric_limits<size_t>::max() - 0xffff
-		return _size <= 0xffff && _offset <= u256(numeric_limits<size_t>::max() - 0xffff);
+		// We only record accesses to contiguous memory chunks that are at most s_maxRangeSize bytes
+		// in size and at an offset of at most numeric_limits<size_t>::max() - s_maxRangeSize
+		return _size <= s_maxRangeSize && _offset <= u256(numeric_limits<size_t>::max() - s_maxRangeSize);
 	}
 	else
 		m_state.msize = u256(-1);
 
 	return false;
+}
+
+bytes EVMInstructionInterpreter::readMemory(u256 const& _offset, u256 const& _size)
+{
+	yulAssert(_size <= s_maxRangeSize, "Too large read.");
+	bytes data(size_t(_size), uint8_t(0));
+	for (size_t i = 0; i < data.size(); ++i)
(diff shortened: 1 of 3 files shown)


test/tools/yulInterpreter/EVMInstructionInterpreter.cpp
```diff
@@ -483,7 +483,10 @@ u256 EVMInstructionInterpreter::evalBuiltin(
 
 bool EVMInstructionInterpreter::accessMemory(u256 const& _offset, u256 const& _size)
 {
-	if (((_offset + _size) >= _offset) && ((_offset + _size + 0x1f) >= (_offset + _size)))
+	// Don't expand memory if range length is zero.
+	if (_size == 0)
+		return false;
```
ekpyron: Why ``false``?

bshastry: Good point. Initially, I was of the impression that accessMemory returns a flag that states whether memory is actually accessed and hence set this line to `false`.

The API docs states that it is false only if size is greater than `0xffff` but I'm wondering if we should change this to `return false if size greater than 0xffff or size equals zero`

Also, `0xffff` is currently hard-coded. Perhaps we should move this into a constexpr inside the `EVMInstructionInterpreter` class? I can then refer to it in the fuzzer generator so that out-of-bounds accesses are not generated.

ekpyron: Hm... yeah... I didn't have a closer look myself, but I always thought of as returning ``false`` indicating that we've hit some resource constraints and don't want to proceed with the access, even though it should actually happen.
I.e. returning ``false`` being a graceful error condition.

But I guess one could also think of it as returning ``true``, if an actual memory access is supposed to happen instead - a quick look at the uses of ``accessMemory`` (without checking in detail) seems like both would work...

So yeah, in that sense it's fine with me either way, but we should probably adjust the comment to be clearer about this :-).

Making the ``0xffff`` a static constexpr member or something like that sounds good in any case, especially if you can make use of it in the fuzzer.

axic: Looking at the callsites for `accessMemory`, the boolean indicates if it is safe to access the memory, i.e. access/expansion was successful.

Probably need to return true **and** strengthen all callsites to handle `offset !=0 && size == 0`.

Looked at it briefly, and I think `readMemory*`/`copyZeroExtended` is safe, but `MSTORE*` and `LOG*` may not be.

bshastry: > Probably need to return true and strengthen all callsites to handle offset !=0 && size == 0.

Could you briefly explain the reasoning for handling this at callsite instead of the `accessMemory` helper? I summarized the current logic as a comment here, could you please take a look to check if it is sound? https://github.com/ethereum/solidity/pull/13516/files#diff-b55fc83988292caec0f11b2c61f28c28da9c0aea18e4f847fea67cac91fac605R83

axic: The call sites do not necessarily work with `offset != 0`, as they may try to access it, even though `accessMemory` would ignore it.

ekpyron: So to reiterate on this: this should really return ``true`` here and a memory access of size zero should be considered a "valid" memory access.
Consider for example the ``Instruction::KECCAK256`` case: if this here returns ``false`` it will return a bogus hash. However, there is no reason for that at all - it can just compute a perfectly fine hash from a zero-length memory region.

So the return here should be changed to ``true`` and the comment in the header adjusted accordingly.

Furthermore, we should have a quick look at all use cases of ``accessMemory`` to check if it correctly handles the zero-size case, but I expect that it should all be fine just as is actually and the *only* thing that needs to be done is changing the return value to ``true`` here.

bshastry: Addressed this comment by returning true.

ekpyron (approved):

PR #13519, opened 2022-09-13
Boost tests for non-delimited optimizer sequence

nikola-matic:
These tests are already covered with `cmdLineTests`, but should be more at home as boost tests.
This PR depends on #13376 being merged, i.e. destination branch will be changed to develop once #13376 is merged.


nikola-matic:
The question that remains is, should `cmdlineTests` that are now covered here be deleted?

ekpyron (approved):

PR #13522, opened 2022-09-13
Workflow to label and comment stale pull requests

r0qs:
This PR supersedes https://github.com/ethereum/solidity/pull/13499 and address some of the concerns of the issue https://github.com/ethereum/solidity/issues/8969.

It adds a workflow to label PRs as stale after 14 days of inactivity, comments on them, and if they remain inactive for more than 7 days, it closes them. If an update occurs, the label is removed, and the timer restarted.

PRs labeled as `external contribution`, milestones, and draft PRs are ignored.
Issues are ignored (i.e., set to `-1`).
It runs every day at noon.

It is currently set for debugging only and can be triggered manually in the workflow dashboard, so we can check the expected behavior in the workflow logs without affecting the current open PRs.

### DIFF
--- a/.github/workflows/stale-pull-requests.yml
+++ b/.github/workflows/stale-pull-requests.yml
@@ -1,0 +1,39 @@
+name: Check stale pull requests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 12 * * *'
+
+permissions:
+  issues: read
+  pull-requests: write
+
+env:
+  BEFORE_STALE: 14
+  BEFORE_CLOSE: 7
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for ${{ env.BEFORE_STALE }} days with no activity.
+            It will be closed in ${{ env.BEFORE_CLOSE }} days unless the `stale` label is removed.
+          close-pr-message: |
+            This pull request was closed due to a lack of activity for ${{ env.BEFORE_CLOSE }} days after it was stale.
+          stale-pr-label: stale
+          close-pr-label: closed-due-inactivity
+          days-before-pr-stale: ${{ env.BEFORE_STALE }}
+          days-before-pr-close: ${{ env.BEFORE_CLOSE }}
+          exempt-pr-labels: 'external contribution'
+          exempt-draft-pr: true
+          exempt-all-milestones: true
+          remove-stale-when-updated: true
+          operations-per-run: 60


.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          debug-only: true
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for 30 days with no activity.
+            It will be closed in 7 days unless the `stale` label is removed.
+          stale-pr-label: stale
+          days-before-pr-stale: 30
```
ekpyron: I'd actually even be more aggressive here and set this to something like 14.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+          days-before-pr-close: 7
+          exempt-pr-labels: 'roadmap,must have,must have eventually,should have,nice to have,external contribution'
```
ekpyron: While for issues, it may make sense to exempt everything that has a "desirability" label, I'd actually also be more aggressive for PRs here. If a roadmap PR is open for more than a few weeks without activity, there is something wrong with it and we should still close it (it can always be recreated).

So I'd actually *only* exempt ``external contribution`` here.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for 30 days with no activity.
```
ekpyron:
```suggestion
            This pull request is stale because it has been open for 14 days with no activity.
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+            It will be closed in 7 days unless the `stale` label is removed.
+          stale-pr-label: stale
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          exempt-pr-labels: 'external contribution'
+          exempt-draft-pr: true
+          exempt-all-milestones: true
+          remove-stale-when-updated: true
+          operations-per-run: 60
+          start-date: "2022-09-14"
```
ekpyron: Actually, for PRs, we don't need this - it'd be nice to have this run on all PRs including the old ones :-).

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for 14 days with no activity.
```
nikola-matic: So, take this with a grain of salt, because I've only ever written one GH workflow, and I tested it out by mercilessly pushing commits until I got it working :smile: 
It says up top in the `name` section that it checks stale issues and pull requests, but the message here refers only to pull requests, and I assume the below 
```yml
         # disable issues
          days-before-issue-stale: -1
          days-before-issue-close: -1
```
doesn't apply this workflow to issues then, in which case the `name` portions is wrong?
Now, I see you've described all of this in the opening comment of the PR, but does that mean we're going to add support for issues later in this exact workflow, or we'll have a separate one for issues?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          debug-only: true
```
nikola-matic: I looked this up, on from my understanding, it will run the workflow, but will only log potential changes without actually making them, right?

nikola-matic (commented):
If my assumptions are correct, then this looks good to me, and if the debugging assumption specifically is correct, then I see no problem with merging this and trying it out.

.github/workflows/stale.yml
r0qs: Yes it is wrong indeed. It is because initially, I configured it for both (issues and PRs), but after some discussion, we decided that we will have it activated only for PRs (see here: https://github.com/ethereum/solidity/pull/13499#issuecomment-1245600327).

As I understood, for PRs the idea is to consider all currently open PRs (https://github.com/ethereum/solidity/pull/13522#discussion_r970773088), label them as stale, and comment (after 14 days of inactivity), and eventually close them after more 7 days.

For issues, the idea is to exempt issues that already exist and only consider new ones based on their creation/update time. They should not label as stale but commented. They should not be closed but we could move them to an `Icebox` column on the project board.
This could be done using the same GH action that I am using here but adding a new job that is only performed for issues or checking the stale status manually. I started to experiment a bit with the latter here: https://github.com/ethereum/solidity/commit/131d6e53df270f3c8c04e572fd4efef3c8674236

.github/workflows/stale.yml
r0qs: Yes, it is correct. It will only print the output in the workflow logs in the actions tab.

nikola-matic: Then fix the `name` field, and let's merge this in. We'll get far more info regarding the correctness of the job from the logs than we will from leaving the PR up for weeks and going over 30-40 lines of yml :)

r0qs: I also added a message and a label for when the PR is closed by the workflow :)

.github/workflows/stale-pull-requests.yml
```diff
@@ -0,0 +1,34 @@
...
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for 14 days with no activity.
+            It will be closed in 7 days unless the `stale` label is removed.
+          close-pr-message: |
+            This pull request was closed due to a lack of activity for 7 days after it was stale.
+          stale-pr-label: stale
+          close-pr-label: closed-due-inactivity
+          days-before-pr-stale: 14
```
nikola-matic: Now that I think about it, do you think it would make sense to use environment variables to define `days-before-pr-stale` and `days-before-pr-close`. E.g.
```
env:
  BEFORE_STALE: 14
  BEFORE_CLOSE: 7
```
And then define
```
days-before-pr-stale: $BEFORE_STALE
days-before-pr-close: $BEFORE_CLOSE
```
and also the messages could look like this
```
This pull request was closed due to a lack of activity for $BEFORE_CLOSE days after it was stale.
```
It would make it neater, with the added benefit that if we decide to change these values, we only have to do it in one place. Of course, I'm not sure if this is a proper approach :man_shrugging: 

Found the docs [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables)

nikola-matic:
Also, please rebase against develop, the `euler` external test is fixed. I know it's not required, but I really like those :heavy_check_mark: builds :)

.github/workflows/stale-pull-requests.yml
r0qs: Yes, definitely.

nikola-matic (approved):
Looks like euler is broken again, so don't worry about it.

nikola-matic:
> Looks like euler is broken again, so don't worry about it.

Or not :smile:

.github/workflows/stale-pull-requests.yml
```diff
@@ -0,0 +1,39 @@
...
+          stale-pr-message: |
+            This pull request is stale because it has been open for ${{ env.BEFORE_STALE }} days with no activity.
+            It will be closed in ${{ env.BEFORE_CLOSE }} days unless the `stale` label is removed.
+          close-pr-message: |
+            This pull request was closed due to a lack of activity for ${{ env.BEFORE_CLOSE }} days after it was stale.
+          stale-pr-label: stale
+          close-pr-label: closed-due-inactivity
+          days-before-pr-stale: ${{ env.BEFORE_STALE }}
+          days-before-pr-close: ${{ env.BEFORE_CLOSE }}
+          exempt-pr-labels: 'external contribution'
```
cameel, line 35: This label actually has an emoji in it :) It should be `external contribution :star:`.

.github/workflows/stale-pull-requests.yml
```diff
@@ -0,0 +1,39 @@
...
+          # disable issues
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: |
+            This pull request is stale because it has been open for ${{ env.BEFORE_STALE }} days with no activity.
+            It will be closed in ${{ env.BEFORE_CLOSE }} days unless the `stale` label is removed.
+          close-pr-message: |
+            This pull request was closed due to a lack of activity for ${{ env.BEFORE_CLOSE }} days after it was stale.
+          stale-pr-label: stale
+          close-pr-label: closed-due-inactivity
```
cameel, line 32: Please remember to actually create these labels before we run the bot. Github would create them automatically but without colors or description.

.github/workflows/stale-pull-requests.yml
```diff
@@ -0,0 +1,39 @@
...
+
+env:
+  BEFORE_STALE: 14
+  BEFORE_CLOSE: 7
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
```
cameel, line 20: The action is producing this warning:
```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

Do they have a newer version that does not use `set-output` already?

.github/workflows/stale-pull-requests.yml
```diff
@@ -0,0 +1,39 @@
...
+            This pull request was closed due to a lack of activity for ${{ env.BEFORE_CLOSE }} days after it was stale.
+          stale-pr-label: stale
+          close-pr-label: closed-due-inactivity
+          days-before-pr-stale: ${{ env.BEFORE_STALE }}
+          days-before-pr-close: ${{ env.BEFORE_CLOSE }}
+          exempt-pr-labels: 'external contribution'
+          exempt-draft-pr: true
+          exempt-all-milestones: true
+          remove-stale-when-updated: true
+          operations-per-run: 60
```
cameel, line 39: I'd increase this to 100 so that we can process all our current PRs in one go.

.github/workflows/stale-pull-requests.yml
r0qs: Yes, they released v6 some days ago which already upgrade to the new format: https://github.com/actions/stale/pull/839

.github/workflows/stale-pull-requests.yml
r0qs: Yes. But for some reason, this action is parsing issues even though we explicitly disable it. I'm thinking of using another one that I did in my tests that call the graphql api directly for the PRs instead of use this `actions/stale`, but it is still a bit incomplete (it does not do everything that the `actions/stale` does yet).

cameel: Right. It probably fetches from the API first and filters only then. Might be worth a feature request or something.

I'd keep it simple for now and stick to this action but yeah, in the longer term we might actually want to switch to the other one (or build a custom one) because I think that just marking as stale is not what we want in some cases. E.g. for unreviewed PRs we might want to instead assign someone randomly or ping the channel or something.

PR #13526, opened 2022-09-14
Small Addition to Structure Of A Contract

yanukadeneth99:
- Added sentence to functions regarding multiple return types
- Link - https://docs.soliditylang.org/en/v0.8.17/structure-of-a-contract.html#functions


nikola-matic:
Hi, and thanks for contributing!

This behaviour is already documented [here](https://docs.soliditylang.org/en/v0.8.17/contracts.html#returning-multiple-values), and the link you posted actually links to it; see [here](https://docs.soliditylang.org/en/v0.8.17/contracts.html#function-parameters-return-variables).

I would be in favour of closing this, but let's see what the others think.

yanukadeneth99:
Oh sure. Thanks for the comment! Just thought maybe stating it would be great.

PR #13527, opened 2022-09-14
Add git `safe.directory` workaround in emscripten build script

### LINKED ISSUE #13506: Nightly builds in solc-bin broken due to git's `safe.directory` check

r0qs:
Fixes https://github.com/ethereum/solidity/issues/13506 by adding [safe.directory](https://github.blog/2022-04-12-git-security-vulnerability-announced/) git configuration to prevent errors when git traverses files with different ownership.

The binaries match since no changes were made in the building paths.
```
diff soljson.js soljson-ci.js -s    
Files soljson.js and soljson-ci.js are identical
```
```
CI version: 0.8.18-ci.2022.9.14+commit.f0f40cf2.Emscripten.clang
Local version: 0.8.18-ci.2022.9.14+commit.f0f40cf2.Emscripten.clang
```

We still need to investigate the feasibility of the `add-nightly-and-push` action in the solc-bin workflow with branch protection enabled, as mentioned here: https://github.com/ethereum/solidity/issues/13506#issuecomment-1245668362

### DIFF
--- a/scripts/ci/build_emscripten.sh
+++ b/scripts/ci/build_emscripten.sh
@@ -40,12 +40,9 @@ else
 	BUILD_DIR="$1"
 fi
 
-apt-get update
-apt-get install lz4 --no-install-recommends
-
 WORKSPACE=/root/project
 
-cd $WORKSPACE
+cd "$WORKSPACE"
 
 # shellcheck disable=SC2166
 if [[ "$CIRCLE_BRANCH" = release || -n "$CIRCLE_TAG" || -n "$FORCE_RELEASE" || "$(git tag --points-at HEAD 2>/dev/null)" == v* ]]
--- a/scripts/build_emscripten.sh
+++ b/scripts/build_emscripten.sh
@@ -35,6 +35,8 @@ else
 fi
 
 # solbuildpackpusher/solidity-buildpack-deps:emscripten-13
+# NOTE: Without `safe.directory` git would assume it's not safe to operate on /root/project since it's owned by a different user.
+# See https://github.blog/2022-04-12-git-security-vulnerability-announced/
 docker run -v "$(pwd):/root/project" -w /root/project \
     solbuildpackpusher/solidity-buildpack-deps@sha256:f1c13f3450d1f2e53ea18ac1ac1a17e932573cb9a5ccd0fd9ef6dd44f6402fa9 \
-    ./scripts/ci/build_emscripten.sh "$BUILD_DIR"
+    /bin/bash -c "git config --global --add safe.directory /root/project && ./scripts/ci/build_emscripten.sh $BUILD_DIR"


scripts/ci/build_emscripten.sh
```diff
@@ -40,13 +40,14 @@ else
...
-apt-get update
-apt-get install lz4 --no-install-recommends
-
 WORKSPACE=/root/project
 
 cd $WORKSPACE
 
+# NOTE: workaround to git unsafe repository error by adding /root/project as a safe.directory
+# See https://github.blog/2022-04-12-git-security-vulnerability-announced/
+git config --global --add safe.directory $WORKSPACE
```
cameel: I'd rather keep global config changes directly in `config.yml`. There we know 100% that we're setting up the environment and have full control over it. This script is technically meant only for CI but it's still nice to be able to run scripts and not mess up your system by accident.

If it really has to be in the script, I'd set the option only for the repo, not globally.this for the working f

cameel:
```suggestion
cd "$WORKSPACE"

# NOTE: workaround to git unsafe repository error by adding /root/project as a safe.directory
# See https://github.blog/2022-04-12-git-security-vulnerability-announced/
git config --global --add safe.directory "$WORKSPACE"
```

scripts/ci/build_emscripten.sh
```diff
@@ -40,13 +40,14 @@ else
...
 
-apt-get update
-apt-get install lz4 --no-install-recommends
-
 WORKSPACE=/root/project
 
 cd $WORKSPACE
 
+# NOTE: workaround to git unsafe repository error by adding /root/project as a safe.directory
+# See https://github.blog/2022-04-12-git-security-vulnerability-announced/
```
cameel: Would be better if the comment explained why we need to do this:
```suggestion
# NOTE: Without `safe.directory` git would assume it's not safe to operate on /root/project since it's owned by a different user.
# See https://github.blog/2022-04-12-git-security-vulnerability-announced/
```

scripts/ci/build_emscripten.sh
r0qs: The problem is that the script `scripts/ci/build_emscripten.sh` is called by the CI but also by the `scripts/build_emscripten.sh` when building emscripten for development using docker. So setting the config in the yml would not help in the latter.

Setting it locally (with `--local`) also does not work because when the volume is mounted in the container as root user under `/root/project` it will contain a different worktree with different permissions from the original repo in the host machine, and thus git will refuse to traverse the directory path.

See the example below:
```
[in the host machine]
git worktree list                                    
/home/cromagnon/src/solidity  f0f40cf24 [git-safe-directory-workaround]

[in the container after running the build_emscripten and setting the global safe.directory flag]
git worktree list
/root/project  f0f40cf24 [git-safe-directory-workaround]
```
The second command is only possible in the build script after setting the `safe.directory` using the global flag. Otherwise, the error that we are trying to workaround happens: `fatal: detected dubious ownership in repository at '/root/project'`

The global setting does not set the configuration in the host machine, only in the docker container. Of course that this assumes that the user did not run the ci script directly but used the provided `scripts/build_emscripten.sh` script.
We can also clean this configuration after the script finish, using: `git config --global --unset safe.directory`.

And if it is indispensable to avoid the case of someone running the CI script without docker or CI, we can remove the `sed -e` e and add a test to check if the `safe.directory` exists before attempting to overwrite it.
This can be done using the following: `git config --global --get safe.directory`, which returns an error if it is not set (exit 1), so we can safely set it. Otherwise, we should bail out and instruct the user to unset it or build using our scripts.

cameel: > `git config --global --unset safe.directory`

That's still destructive. The value on the host is not guaranteed to be empty. I guess it would be more acceptable if you saved and restored the original value but even then if the script fails, you're left with modifications in your git config.

> The problem is that the script `scripts/ci/build_emscripten.sh` is called by the CI but also by the `scripts/build_emscripten.sh` when building emscripten for development using docker. So setting the config in the yml would not help in the latter.

Then how about making `scripts/build_emscripten.sh` run it too? I.e. something like:

``` bash
docker run ...
    'git config --global --add safe.directory $WORKSPACE && ./scripts/ci/build_emscripten.sh "$BUILD_DIR"'
```
(not tested; likely needs some quoting adjustments to work properly).

> And if it is indispensable to avoid the case of someone running the CI script without docker or CI, we can remove the sed `-e` e and add a test to check if the `safe.directory` exists before attempting to overwrite it.

I think our scripts *should* be built to be safely runnable locally because that makes debugging them easier. But in the end it's not like we can't live without it. The script already runs `apt-get` commands, which would modify your system (or fail on many distros) so in these terms the problem already exists and the current state of the PR does not change that situation, just does not improve it :)

Would be best if you could try my suggestion with modifying `scripts/build_emscripten.sh` above. But what matters here the most to get the nightlies back up and running so if it does not work or you have some good reasons why it would not be a good solution, just go with the current version (but please remove the `git config --global --unset safe.directory` line).

cameel:
Pinging @r0qs. After you're done with fixing the hardhat test, would be best to finalize this. Just needs a small tweak at this point.

scripts/ci/build_emscripten.sh
r0qs: This would not solve the problem because the [git show](https://github.com/ethereum/solidity/blob/develop/scripts/ci/build_emscripten.sh#L58) command which is the one that triggers the git permission issue is in the CI script and not in the docker run script. So the `git config --global --add safe.directory $WORKSPACE` need to be called from the CI script otherwise the CI tests will continue failing (i.e. the `scripts/ci/build_emscripten.sh` is called directly by the CI `b_ems` job).

As you pointed out, the CI script already contains commands that may break the host machine. They are not intended to be run locally by the user (only through the docker run scripts).

cameel: > This would not solve the problem because the [git show](https://github.com/ethereum/solidity/blob/develop/scripts/ci/build_emscripten.sh#L58) command which is the one that triggers the git permission issue is in the CI script and not in the docker run script.

But I mean adding it to the `docker run` command, which would make it run in the container and would affect its environment.

r0qs: Oh, ok, my bad. Now I guess I understand what you mean.

I thought that the solidity emscripten CI job would fail since it does not use the docker run but the CI script directly (i.e. [scripts/ci/build_emscripten.sh](https://github.com/ethereum/solidity/blob/develop/.circleci/config.yml#L1029))
But in their case, the directory will have the same permissions as the CI runner. The problem indeed only happens when the `scripts/build_emscripten.sh` is called (by the solc-bin nightly build CI or by the user).

I applied your suggestion.

cameel: Actually, I expected that you'd add the `git config` command in the CI file as well. But if it works without it, even better :)

cameel (dismissed):

cameel:
Oh, please don't forget to squash review fixes before merging.

scripts/ci/build_emscripten.sh
r0qs: I think it is not needed, since it will run with the CI runner user. But I may be wrong haha

cameel (approved):

PR #13528, opened 2022-09-15
Move new issues to triage

r0qs:
This PR adds a simple workflow to move new issues to the [old project board](https://github.com/ethereum/solidity/projects/43) as mentioned here: https://github.com/ethereum/solidity/issues/8969#issuecomment-1245619135

The workflow adds any new issue not labeled yet to the Triage column. It requires repository, project and org permissions and only works for github `classic` project board. However, it is easy to adjust it to the `projectV2` api when we move to the new GH boards.
It also has a `dry-run` flag, so nothing will be mutated in the boards until we turn it to false.
The `COLUMN_NAME` is only for debugging and will be removed.

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,0 +1,61 @@
+name: Add new issues to triage column
+
+on:
+  issues:
+    types:
+      - opened
+
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  ORGANIZATION: ethereum
+  REPOSITORY: solidity
+  PROJECT_NUMBER: 43
+  DRY_RUN: true
+
+jobs:
+  triage_issues:
+    runs-on: ubuntu-latest
+    if: join(github.event.issue.labels) == ''
+    steps:
+      - name: Retrieve the content of all columns on the board
+        run: |
+          gh api graphql \
+            --raw-field owner="$ORGANIZATION" \
+            --field project_number="$PROJECT_NUMBER" \
+            --raw-field repository_name="$REPOSITORY" \
+            --raw-field query='
+              query($owner: String!, $repository_name: String!, $project_number: Int!) {
+                repository(owner: $owner, name: $repository_name) {
+                  project(number: $project_number) {
+                    columns(first: 10) {
+                      nodes {
+                        id,
+                        name
+                      }
+                    }
+                  }
+                }
+              }' > project_columns.json
+          echo 'COLUMN_ID='$(jq '.data.repository.project.columns.nodes[] | select(.name == "Triage") | .id' project_columns.json) >> $GITHUB_ENV
+          echo 'COLUMN_NAME='$(jq '.data.repository.project.columns.nodes[] | select(.name == "Triage") | .name' project_columns.json) >> $GITHUB_ENV
+
+      - name: Add issue#${{ github.event.issue.number }} to Triage column
+        env:
+          ISSUE_ID: ${{ github.event.issue.node_id }}
+        run: |
+          echo "Adding issue: ${{ github.event.issue.number }} to column $COLUMN_NAME in project $PROJECT_NUMBER"
+          if [[ $DRY_RUN == "false" ]]; then
+            gh api graphql \
+              --silent \
+              --raw-field column=$COLUMN_ID \
+              --raw-field issue=$ISSUE_ID \
+              --raw-field query='
+                mutation($column: ID!, $issue: ID!) {
+                  addProjectCard(input: {
+                    projectColumnId: $column,
+                    contentId: $issue
+                  }) {
+                    clientMutationId
+                  }
+                }'
+          fi


.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
+name: Triage new issues
```
cameel:
```suggestion
name: Add new issues to triage column
```

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  issues:
+    types: [ opened ]
```
cameel:
```suggestion
    types:
      - opened
```

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+
+env:
+  GITHUB_TOKEN: ${{ secrets.PROJECT_BOARD_AUTOMATION }}
```
cameel: So it still needs a token? We won't get away with the built-in token for managing boards?

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+                      id,
+                      name
+                    }
+                  }
+                }
+              }
+            }' --raw-field org=$ORGANIZATION --field number=$PROJECT_NUMBER >> project_data.json
+          echo 'COLUMN_ID='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .id' project_data.json) >> $GITHUB_ENV
+          echo 'COLUMN_NAME='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .name' project_data.json) >> $GITHUB_ENV
+      - name: Add issue#${{ github.event.issue.number }} to Triage column
```
cameel: I'd add an empty line between steps for readability.

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  triage_issues:
+    runs-on: ubuntu-latest
+    if: join(github.event.issue.labels) == ''
+    steps:
+      - name: Retrieve project data
+        env:
+          ORGANIZATION: ${{ env.ORGANIZATION }}
+          PROJECT_NUMBER: ${{ env.PROJECT_NUMBER }}
```
cameel: Aren't these already in the env? Do you have to repeat them?

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+                project(number: $number) {
+                  columns(first: 10) {
+                    nodes {
+                      id,
+                      name
+                    }
+                  }
+                }
+              }
+            }' --raw-field org=$ORGANIZATION --field number=$PROJECT_NUMBER >> project_data.json
```
cameel:
```suggestion
            }' --raw-field organization="$ORGANIZATION" --field project_number="$PROJECT_NUMBER" >> project_data.json
```

cameel: This looked hard to decipher at first with the big query in the middle of it. What do you think about formatting it this way?
```suggestion
          gh api graphql \
            --raw-field org=$ORGANIZATION \
            --field number=$PROJECT_NUMBER \
            --raw-field query='
              query($org: String!, $number: Int!) {
                organization(login: $org) {
                  project(number: $number) {
                    columns(first: 10) {
                      nodes {
                        id,
                        name
                      }
                    }
                  }
                }
              }' >> project_data.json
```

Same with the other query below.

cameel:
```suggestion
            }' --raw-field org=$ORGANIZATION --field number=$PROJECT_NUMBER > project_columns.json
```

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+          echo 'COLUMN_ID='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .id' project_data.json) >> $GITHUB_ENV
+          echo 'COLUMN_NAME='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .name' project_data.json) >> $GITHUB_ENV
+      - name: Add issue#${{ github.event.issue.number }} to Triage column
+        env:
+          ISSUE_ID: ${{ github.event.issue.node_id }}
+          PROJECT_NUMBER: ${{ env.PROJECT_NUMBER }}
+        run: |
+          if [ ${{ env.DRY_RUN }} ]; then
+            echo "Adding issue: ${{ github.event.issue.number }} to column $COLUMN_NAME in project $PROJECT_NUMBER"
```
cameel: Why not print this even if it's not a dry run? Or does `gh api graphql` already print something? From `--silent` I'm guessing it does not?

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+          else
+            gh api graphql --raw-field query='
+              mutation($column: ID!, $issue: ID!) {
+                addProjectCard(input: {
+                  projectColumnId: $column,
+                  contentId: $issue
+                }) { clientMutationId }
+              }' --raw-field column=$COLUMN_ID --raw-field issue=$ISSUE_ID --silent
```
cameel: What will happen if the issue is already on the board? Either in Triage or in another column? Will this fail or just remove it from the other column?

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+              }
+            }' --raw-field org=$ORGANIZATION --field number=$PROJECT_NUMBER >> project_data.json
+          echo 'COLUMN_ID='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .id' project_data.json) >> $GITHUB_ENV
+          echo 'COLUMN_NAME='$(jq '.data.organization.project.columns.nodes[] | select(.name == "Triage") | .name' project_data.json) >> $GITHUB_ENV
+      - name: Add issue#${{ github.event.issue.number }} to Triage column
+        env:
+          ISSUE_ID: ${{ github.event.issue.node_id }}
+          PROJECT_NUMBER: ${{ env.PROJECT_NUMBER }}
+        run: |
+          if [ ${{ env.DRY_RUN }} ]; then
```
cameel: You should compare with "true". Otherwise the condition will still be true when you change `DRY_RUN` to `false`.

Also, this is not a bash variable so it needs quotes even in `[[ ]]` in case it contains spaces.
```suggestion
          if [[ "${{ env.DRY_RUN }}" == true ]]; then
```

.github/workflows/triage.yml
```diff
@@ -0,0 +1,53 @@
...
+  ORGANIZATION: ethereum
+  PROJECT_NUMBER: 43
+  DRY_RUN: true
+
+jobs:
+  triage_issues:
+    runs-on: ubuntu-latest
+    if: join(github.event.issue.labels) == ''
+    steps:
+      - name: Retrieve project data
```
cameel:
```suggestion
      - name: Retrieve the content of all columns on the board
```

cameel (commented):
Needs some tweaks but the overall direction is good. Once these are fixed we can merge and see if it does the job.

.github/workflows/triage.yml
r0qs: Actually no, we do not need it. I tested in a private repo under a organization that I created to play with the configurations using classic projects, but now I realized that I should have used the [repository](https://docs.github.com/en/graphql/reference/objects#repository) API instead of the organization, since the current Solidity project board is in the repo scope and not in the org. I will update the query, so the org permissions will not be necessary.

.github/workflows/triage.yml
r0qs: Nope, they can be removed.

.github/workflows/triage.yml
r0qs: Indeed. The true should also be between `"` I believe.

.github/workflows/triage.yml
r0qs: The `--silent` only prevents the api from printing the response body, which is useless in this case. But yes, we can print the same message regardless.

.github/workflows/triage.yml
r0qs: Hum... good question. I believe that it will attempt to move it anyway. But the action is only triggered when the issue is open (and has no labels), so in theory, it can only be triggered once per issue.

.github/workflows/triage.yml
r0qs: Definitely, much better! :)

.github/workflows/triage.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  triage_issues:
+    runs-on: ubuntu-latest
+    if: join(github.event.issue.labels) == ''
+    steps:
+      - name: Retrieve the content of all columns on the board
+        run: |
+          gh api graphql \
+            --raw-field owner="$ORGANIZATION" \
```
nikola-matic, line 23: I'd change this to `${{ env.ORGANIZATION }}`. Same for other variable references below.

nikola-matic:
@r0qs can you rebase, and let's ping @cameel; if you've addressed all of the changes, we may as well merge this to test it out :)

.github/workflows/triage.yml
cameel: Why though? These are just environment variables. I think it's fine to use them like this (and it looks cleaner :)).

I'd actually do the opposite and replace `"${{ env.DRY_RUN }}"` below with `$DRY_RUN` :)

.github/workflows/triage.yml
```diff
@@ -0,0 +1,61 @@
+name: Add new issues to triage column
+
+on:
+  issues:
+    types:
+      - opened
+
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
cameel, line 9: This seems unused.

cameel (dismissed):
Looks good apart from some cosmetic things.

And in the end we need to merge it and see how it'll be have in practice.

cameel:
Please squash the commits before merging.

.github/workflows/triage.yml
cameel: ok, I guess we'll see how it behaves when we merge. Some weird effects will likely be easy to notice and fix.

.github/workflows/triage.yml
cameel: Technically yes (without quotes it's a regex) but since it does not have any special chars, it's safe not to quote it here.

.github/workflows/triage.yml
nikola-matic: Because that's the style we used in #13522. I personally prefer this instead of `$VAR` as it's make it super obvious at a glance that it's a variable reference, but I'm not really super opposed to your suggestion either. I would just prefer to have a consistent style across all of the scripts.

r0qs: I replaced the `"${{ env.DRY_RUN }}"` because it was indeed wrong.
Just to clarify the use of `env` in the github actions, it is used to define the scope of the environment variable in the workflow, and it can be accessed directly, e.g. `$SOMETHING`, using the appropriate syntax of the runner environment (i.e. Ubuntu with bash, in our case), or accessed through the equivalent context property, e.g. `${{ env.SOMETHING }}`. According to the docs, the latter is designed to be used in the parts of the workflow that is not processed by the runner but by the GitHub Actions itself. So, using the `DRY-RUN` through context was indeed not the recommended way.

.github/workflows/triage.yml
r0qs: No, it is required by the GitHub cli: https://github.com/cli/cli/blob/e7102f9d84dae8135aca13edffe4c7d0049bdde4/cmd/gh/main.go#L317
I set it to the entire workflow because both steps needed it. But it can be scoped for each step.

.github/workflows/triage.yml
r0qs: As the job only runs when an issue is opened and has no labels, I believe we will not have such a problem. However, if we also trigger the job for reopening and remove the checking for no labels, we will have a problem when attempting to add an already existing issue in the board because of the use of the `addProjectCard` method.

I tested here, and when an issue is reopened, the issue can be moved automatically to a column if that column has set the `Reopened` option in the board settings. And if that is the case, and we also allow the workflow to be triggered by reopening, the action will fail.

As a reopened issue may require a new evaluation of the task, it may be good to change that command and support adding new issues to triage or move them to triage if they are reopened. This can be added as a separated PR. It will basically be the removal of the empty label requirement and the addition of a condition in the last step to decide whether add or move (using `moveProjectCard`) the issue.

cameel: Right, this should run also on reopening as long as the issue is still untriaged. It does happen that a someone closes something by accident and then reopens it. Having it drop from Triage in that case is undesirable. But that's something for a followup PR. We can always fix these things manually when we notice.

cameel (approved):

nikola-matic (approved):

.github/workflows/triage.yml
r0qs: Indeed. I will open another PR that modifies it to add another step for reopening ;)

PR #13529, opened 2022-09-15
optimize and_shl distribution operation

### LINKED ISSUE #9119: Optimize AND/SHL
``and(shl(A, X), shl(A, Y))``
should be optimized to
``shl(A, and(X, Y))``

or even

``and(shl(Z, X), shl(Z, Y))``
should be optimized to
``shl(Z, and(X, Y))``

nishant-sachdeva:
added optimization. shl comes out from being distributed over and

closes https://github.com/ethereum/solidity/issues/9119


libevmasm/RuleList.h
```diff
@@ -438,7 +438,7 @@ template <class Pattern>
 std::vector<SimplificationRule<Pattern>> simplificationRuleListPart7(
 	Pattern A,
 	Pattern B,
-	Pattern,
+	Pattern Z,
```
ekpyron: Since this is called as ``rules += simplificationRuleListPart7(A, B, C, W, X);``, the third parameter will match constants - if you need another pattern ranging over arbitrary expressions, you need to additionally pass ``Y`` or ``Z`` in here on the call site instead.

libevmasm/RuleList.h
```diff
@@ -629,6 +630,18 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart7(
 		feasibilityFunction
 	});
 
+	rules.push_back({
+		// AND(SHL(A,X), SHL(A,Y)) -> SHL(A, AND(X,Y))
+		Builtins::AND(Builtins::SHL(A,X), Builtins::SHL(A,Y)),
```
chriseth: style: `,` -> `, `

libevmasm/RuleList.h
```diff
@@ -629,6 +630,18 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart7(
...
+		[=]() -> Pattern { return Builtins::SHL(A, Builtins::AND(X,Y)); }
+	});
+
+	rules.push_back({
+		// AND(SHL(Z,X), SHL(Z,Y)) -> SHL(Z, AND(X,Y))
```
chriseth: Please note that there is a huge difference between A B C and X Y Z. A B C are always known constants, X Y Z are arbitrary complex expressions that don't need to be constant.

Please replace `A` above by `Z` and compare gas costs.

libevmasm/RuleList.h
chriseth: style (space after comma)

chriseth (changes requested):
see comments

libevmasm/RuleList.h
nishant-sachdeva: If I replace `A` with `Z`, the rule at `RuleList.h: 633` becomes exactly the same as the rule at `RuleList.h:639`

test/formal/and_distributed_over_shl.py
```diff
@@ -0,0 +1,25 @@
...
+from z3 import BitVec
+
+"""
+Rule:
+AND(SHL(A,X), SHL(A,Y)) -> SHL(A, AND(X,Y))
+"""
+
+rule = Rule()
+
+n_bits = 128
```
matheusaaguiar, line 12: why not 256?

nishant-sachdeva: A Shift Left Operation being performed below. Having a 128 helps avoid overflow and keeps it simple. 

But if I've missed something and its required, I can change it to 256 bits as well.

matheusaaguiar: Right. I was just curious about it.

matheusaaguiar (dismissed):
Looks good to me.

libevmasm/RuleList.h
```diff
@@ -629,6 +630,12 @@ std::vector<SimplificationRule<Pattern>> simplificationRuleListPart7(
 		feasibilityFunction
 	});
 
+	rules.push_back({
```
nikola-matic, line 633: As @ekpyron [mentioned](https://github.com/ethereum/solidity/issues/9119#issuecomment-1028060749) in the original issue, `A, B, C...` is for constants, and `X, Y, Z, ...` for arbitrary values - so will this rule then cover the `and(shl(A, X), shl(A, Y))` as well?

chriseth: Yes it will. constants are part of arbitrary values :)

test/formal/and_distributed_over_shl.py
```diff
@@ -0,0 +1,25 @@
+from opcodes import AND, SHL
+from rule import Rule
+from z3 import BitVec
+
+"""
+Rule:
+AND(SHL(A,X), SHL(A,Y)) -> SHL(A, AND(X,Y))
```
chriseth: Just for clarity: Please update to also use X, Y, Z here.

nishant-sachdeva: Done

test/formal/and_distributed_over_shl.py
ekpyron: If it works using full 256 bits, i.e. it actually proves it, then it should *always* be 256 bits.
Even more so since using a lower bit width exactly might hide overflow cases or special edge cases. Usually the proven properties are monotone in bit-width and at times the solvers can't prove them at full 256-bits, but it significantly increases the review effort not to use 256 bits, so: if possible always keep at 256 bits.

ekpyron: Just as a general remark. I just checked it here: the rule works for 256 bits, but is very slow to prove - and it's rather easy to see that it's monotonic in bit width, so it's fine to keep it at 128.

chriseth (dismissed):

chriseth (dismissed):

chriseth (approved):

PR #13531, opened 2022-09-15
Add VM Version Paris and new built in function ``prevrandao``

### LINKED ISSUE #13512: feat: introduce `block.prevrandao` as alias for `block.difficulty`
## Summary

[EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) renames the `DIFFICULTY (0x44)` opcode to `PREVRANDAO (0x44)`. The return value of the `DIFFICULTY (0x44)` instruction after this change is the output of the randomness beacon provided by the beacon chain.

For reference, I implemented this feature in Vyper [here](https://github.com/vyperlang/vyper/pull/3085).

### LINKED ISSUE #13517: Add Paris as EVM version
Came up with https://github.com/ethereum/solidity/issues/13512.
We should aim for having this in the next release regardless of how we decide to deal with that issue.

https://github.com/ethereum/solidity/issues/13512 should be the only relevant change for us to consider for paris, though.

Marenz:
fixes #13512 
fixes #13517


nikola-matic:
I've done a bit of grepping, and it seems like you're missing a case in `libevmasm/SemanticInformation.cpp` in `invalidInPureFunction(Instruction _instructioN)`. If `DIFFICULTY` was invalid there, then surely `PREVRANDAO` is?

Also in `test/tools/yulInterpreter/EVMInstructionInterpreter.cpp`, and `test/tools/ossfuzz/protoToYul.cpp`, but those are tools, so maybe not relevant to this PR.

Marenz:
> seems like you're missing a case in libevmasm/SemanticInformation.cpp

They both have the same value, so it can only contain either `DIFFICULTY` or `PREVRANDAO`, in that switch case i decided to leave the original one.

> ...

same with the other cases, all switches just use the number and the semantics didn't really change. Whether we want to update the yul interpreter m_state.difficulty part is a different question, but nothing breaks if we keep it like this.

Marenz:
Not quiet sure why the grammar test fails:
```
[1907/2991] Testing /root/project/test/libsolidity/semanticTests/inlineAssembly/prevrandao_pre-paris_function.sol FAILED
line 4:16 no viable alternative at input 'letprevrandao'
line 4:27 mismatched input ':=' expecting YulLParen
line 6:8 mismatched input '}' expecting YulLParen
line 10:21 mismatched input 'prevrandao' expecting YulIdentifier
line 10:34 extraneous input '->' expecting {YulBreak, YulContinue, YulFor, YulFunction, YulIf, 'leave', 'let', 'switch', YulEVMBuiltin, YulLBrace, YulRBrace, YulIdentifier}

Exited with code exit status 1

```

Marenz:
found an exception in the grammar test for `let basefee...` and added one as well for `let prevrandao`

ekpyron (commented):
Nice, that was quick :-)!

I'm missing two things:

evmasm output (so ``solc --asm`` and related outputs) will contain opcode names - in those ``0x44`` should appear as ``difficulty`` pre-paris and as ``prevrandao`` with post-paris set as evm-version - I don't see that part accounted for in code or tests yet... ``test/libyul/evmCodeTransform`` tests can be abused for this, and/or ``test/libevmasm/Assembler.cpp``.

There might also be some related complication in ``solidity::evmasm::c_instructions`` and ``c_instructionInfo`` getting initialized with the same key twice. This will only keep one version, won't it? We'll probably have to decide for ``PREVRANDAO`` only there and replace accesses to those two maps by a function taking an evm-version as argument, which adjusts accordingly.

The second thing is:
Ideally, the use of ``difficulty()`` and ``prevrandao()`` in inline assembly and yul would also cause warnings similar to the ``block`` members on mismatching evm versions. We usually don't warn about such things on the Yul level, so there may be some complication there and we may have to weigh whether the warning is worth adding additional infrastructure there, if needed, but ideally those should be warnings.

Changelog.md
```diff
@@ -1,10 +1,11 @@
 ### 0.8.18 (unreleased)
 
 Language Features:
+ * Add support for ``prevrandao()`` which was introduced in EVM version to "Paris", supplanting ``difficulty`` in the process. Both can be used interchangeably.
```
axic, line 4: As mentioned on the issue I think this is a bad idea, because the two outputs are expected to be **vastly different** and hence expect different use cases.

libyul/backends/evm/EVMDialect.cpp
```diff
@@ -139,6 +146,17 @@ set<YulString> createReservedIdentifiers(langutil::EVMVersion _evmVersion)
 
 map<YulString, BuiltinFunctionForEVM> createBuiltins(langutil::EVMVersion _evmVersion, bool _objectAccess)
 {
+	auto hasOpCode = [&](evmasm::Instruction _instr, string const& _instrName) -> bool
+	{
+		if (
+			_instr == evmasm::Instruction::PREVRANDAO &&
+			_instrName == "prevrandao"
+		)
+			return _evmVersion.hasPrevRandAO();
```
axic: What is this weird capitalisation, `RandAO`?

test/evmc/evmc.h
```diff
@@ -832,12 +832,19 @@ enum evmc_revision
      */
     EVMC_LONDON = 9,
 
+    /**
+     * The Paris revision.
+     *
+     * https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md
+     */
+    EVMC_PARIS = 10,
```
axic, line 840: This is *fine*, but this PR can't be tested without newer evmone, so I'd suggest to take #13532 first.

cameel:
This issue solves #13517 as well, right? Should be added to the top comment.

libyul/backends/evm/EVMDialect.cpp
Marenz: I dunno, I tried my best to make sense of the weird name...

Changelog.md
Marenz: Well, this is exactly implemented according to the summary after the discussion, so I'm not sure what to tell you here...
https://github.com/ethereum/solidity/issues/13512#issuecomment-1246792479

libyul/backends/evm/EVMDialect.cpp
axic: It comes from
> A DAO (decentralised autonomous organisation) that anyone can participate in, and the random number is generated by all participants together!

So the names used are RANDAO, randao, Randao -- in this order in the `consensus-specs.` Likely `hasPrevRandao` is the best choice here.

docs/yul.rst
```diff
@@ -935,6 +935,8 @@ the ``dup`` and ``swap`` instructions as well as ``jump`` instructions, labels a
 +-------------------------+-----+---+-----------------------------------------------------------------+
 | difficulty()            |     | F | difficulty of the current block                                 |
 +-------------------------+-----+---+-----------------------------------------------------------------+
+| prevrandao()            |     | P | randomness provided by the beacon chain                         |
```
axic: I think this documentation piece makes it look like that both opcodes exists and return a different value.

test/libsolidity/syntaxTests/inlineAssembly/prevrandao_warn_pre-paris.sol
```diff
@@ -0,0 +1,7 @@
+function f() view returns (uint) {
+	return block.prevrandao;
```
hrkrshnn: Nit: indent and file name. Similarly, for other files as well. Somewhat

test/libsolidity/syntaxTests/types/magic_block_istanbul.sol
```diff
@@ -2,9 +2,6 @@ contract C {
     function f() public view returns (address payable) {
         return block.coinbase;
     }
-    function g() public view returns (uint) {
-        return block.difficulty;
-    }
```
hrkrshnn, line 7: Is it possible to keep this and change the EVM version in lines 15..17?

libevmasm/Instruction.h
```diff
@@ -84,7 +84,8 @@ enum class Instruction: uint8_t
 	TIMESTAMP,			///< get the block's timestamp
 	NUMBER,				///< get the block's number
 	DIFFICULTY,			///< get the block's difficulty
-	GASLIMIT,			///< get the block's gas limit
+	PREVRANDAO = DIFFICULTY,	///< get randomness provided by the beacon chain
+	GASLIMIT = 0x45,	///< get the block's gas limit
```
hrkrshnn, line 262: Hmm, didn't know this works.

test/evmc/evmc.h
hrkrshnn: So Paris is the 10th revision? Interesting coincidence.

hrkrshnn (commented):
We also need a test for the view-pure checker.

hrkrshnn:
> This issue solves https://github.com/ethereum/solidity/issues/13517 as well, right? Should be added to the top comment.

It should. But just to note, this PR hasn't made Paris the default.

Marenz:
> It should. But just to note, this PR hasn't made Paris the default.

If that is desired I can change that, no problem :)

docs/yul.rst
Marenz: Hm what would be a good way to present it here, with the one letter column allowing very limited space

test/libsolidity/syntaxTests/types/magic_block_istanbul.sol
Marenz: I moved the test to another file.. would you prefer to keep this one here?

test/libsolidity/syntaxTests/inlineAssembly/prevrandao_warn_pre-paris.sol
Marenz: As far as I know we use both tabs/spaces in test files, just not mixed in the same file.
About the name.. you don't like the `pre-` part? I guess I could just remove it

cameel: > As far as I know we use both tabs/spaces in test files

We do, but I think that's not on purpose. We should settle on one version.

test/libsolidity/syntaxTests/types/magic_block_istanbul.sol
hrkrshnn: Ah true, nevermind.

stackenbotten:
There was an error when running `chk_coding_style` for commit `1228b043036d68149a94d1e2547f7cc5b09ad626`:
```
Coding style error:
libevmasm/Instruction.h:295:    InstructionInfo(
libevmasm/Instruction.h:296:        std::string _name,
libevmasm/Instruction.h:297:        int _additional,
libevmasm/Instruction.h:298:        int _args,
libevmasm/Instruction.h:299:        int _ret,
libevmasm/Instruction.h:300:        bool _sideEffects,
libevmasm/Instruction.h:301:        Tier _gasPriceTier
libevmasm/Instruction.h:302:    ):
libevmasm/Instruction.h:303:        m_name(_name),
libevmasm/Instruction.h:304:        additional(_additional),
libevmasm/Instruction.h:305:        args(_args),
libevmasm/Instruction.h:306:        ret(_ret),
libevmasm/Instruction.h:307:        sideEffects(_sideEffects),
libevmasm/Instruction.h:308:        gasPriceTier(_gasPriceTier)
libevmasm/Instruction.h:309:    {}
libsolidity/interface/CompilerStack.cpp:227:    if (m_evmVersion < langutil::EVMVersion::paris())
libsolidity/interface/CompilerStack.cpp:228:        solidity::evmasm::usePreParisOpcodeNames();

```

libsolidity/interface/CompilerStack.cpp
```diff
@@ -223,6 +223,9 @@ void CompilerStack::setEVMVersion(langutil::EVMVersion _version)
 	if (m_stackState >= ParsedAndImported)
 		solThrow(CompilerError, "Must set EVM version before parsing.");
 	m_evmVersion = _version;
+
+    if (m_evmVersion < langutil::EVMVersion::paris())
```
Marenz: this is from an old attempt. ignore.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3286,18 +3286,33 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
-				"\"basefee\" is not supported by the VM version."
-			);
+			else if (memberName == "basefee" && !m_evmVersion.hasBaseFee())
+				m_errorReporter.typeError(
+					5921_error,
+					_memberAccess.location(),
+					"\"basefee\" is not supported by the VM version."
+				);
+			else if (memberName == "difficulty" && m_evmVersion.hasPrevRandao())
+				m_errorReporter.warning(
```
aarlt, line 3304: I would say this should be an error - because the semantic of `difficulty` vs. `prevrandao` is different. There is no case where `difficulty` in `Paris` means `difficulty` - that's why I would argue that compilation need to fail here.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3286,18 +3286,33 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
+					8417_error,
+					_memberAccess.location(),
+					"\"difficulty\" was renamed and supplanted by \"prevrandao\" in the VM version paris."
+				);
+			else if (memberName == "prevrandao" && !m_evmVersion.hasPrevRandao())
+				m_errorReporter.warning(
```
aarlt, line 3310: Also this should be an error from my point of view.

libsolidity/analysis/TypeChecker.cpp
Marenz: A bit late to [the discussion](https://github.com/ethereum/solidity/issues/13512#issuecomment-1246792479) I'd say :)
I don't really mind either way, but as I understand it, this all already has been discussed

test/libsolidity/syntaxTests/inlineAssembly/difficulty_reserved.sol
```diff
@@ -0,0 +1,18 @@
...
+        assembly {
+            function difficulty() -> r {
+                r := 1000
+            }
+            ret := difficulty()
+        }
+    }
+}
+// ----
+// ParserError 5568: (98-108): Cannot use builtin function name "difficulty" as identifier name.
```
aarlt, line 18: Maybe also `EVMVersion: <paris` here? Is a `difficulty` identifier allowed again for >=paris?

test/libsolidity/syntaxTests/inlineAssembly/difficulty_warn_paris.sol
```diff
@@ -0,0 +1,7 @@
+function f() view returns (uint) {
+	return block.difficulty;
+}
+// ====
+// EVMVersion: >=paris
+// ----
+// Warning 8417: (43-59): "difficulty" was renamed and supplanted by "prevrandao" in the VM version paris.
```
aarlt: As written above, I think it's better to create an error instead of a warning.

test/libsolidity/syntaxTests/inlineAssembly/prevrandao_warn_pre-paris.sol
```diff
@@ -0,0 +1,7 @@
...
+}
+// ====
+// EVMVersion: <paris
+// ----
+// Warning 9432: (43-59): "prevrandao" is not supported by the VM version and will be treated like "difficulty".
```
aarlt: Error?

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_view.sol
```diff
@@ -73,7 +73,7 @@ contract C {
             pop(coinbase())
             pop(timestamp())
             pop(number())
-            pop(difficulty())
+            pop(prevrandao())
```
aarlt, line 76: Maybe it's better to have two tests here, one `<paris` and one `>= paris`?

test/libsolidity/syntaxTests/inlineAssembly/difficulty_reserved.sol
aarlt: If yes, I think the test for this is missing.

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -3286,18 +3286,33 @@ bool TypeChecker::visit(MemberAccess const& _memberAccess)
...
+				m_errorReporter.typeError(
+					5921_error,
+					_memberAccess.location(),
+					"\"basefee\" is not supported by the VM version."
+				);
+			else if (memberName == "difficulty" && m_evmVersion.hasPrevRandao())
+				m_errorReporter.warning(
+					8417_error,
+					_memberAccess.location(),
+					"\"difficulty\" was renamed and supplanted by \"prevrandao\" in the VM version paris."
```
aarlt, line 3307: Why `renamed and supplanted` and not just `replaced`?

libevmasm/Instruction.cpp
```diff
@@ -223,6 +224,8 @@ static std::map<Instruction, InstructionInfo> const c_instructionInfo =
 	{ Instruction::COINBASE,	{ "COINBASE",		0, 0, 1, false, Tier::Base } },
 	{ Instruction::TIMESTAMP,	{ "TIMESTAMP",		0, 0, 1, false, Tier::Base } },
 	{ Instruction::NUMBER,		{ "NUMBER",			0, 0, 1, false, Tier::Base } },
+//	{ Instruction::PREVRANDAO,	{ "PREVRANDAO",		0, 0, 1, false, Tier::Base } },
+// TODO Replace with PREVRANDAO in the next breaking release
 	{ Instruction::DIFFICULTY,	{ "DIFFICULTY",		0, 0, 1, false, Tier::Base } },
```
ekpyron: IMHO, this doesn't make sense.
It's a question of the EVM version, not with source compatibility - with the next breaking release compiling to <paris *still* should result in ``DIFFICULTY`` and already now, it's weird to emit a non-existent opcode name for >=paris.

So should really do this right away, as already described in https://github.com/ethereum/solidity/pull/13531#pullrequestreview-1109684090.

The EIP just changes the fact that there is a constant mapping from the binary opcode to instruction infos - we have to account for that in the code one way or another (and it's not like dragging an EVM version dependency through here is that much of a hassle).

ekpyron (changes requested):

libsolidity/analysis/TypeChecker.cpp
ekpyron: The current plan is to make it a warning, since making it an error right away is borderline breaking (even though one can argue about that). With 0.9 it should definitely become an error.

libsolidity/analysis/TypeChecker.cpp
Marenz: It's the terminology they used in the EIP and I believe makes it most clear that it's the same opcode behind it

test/libsolidity/syntaxTests/viewPureChecker/inline_assembly_instructions_allowed_view.sol
Marenz: I introduced another test [inline_assembly_instructions_allowed_view_london.sol](https://github.com/ethereum/solidity/pull/13531/files#diff-11182a2286915423c63070802c2a16488b8f2e4f155dc69243d2567eca086a3a) for london that contains only the changed opcode

stackenbotten:
There was an error when running `chk_coding_style` for commit `513d5a32658127eb4afa44b45cff526a36e87e30`:
```
Coding style error:
libevmasm/Disassemble.cpp:19:#include "libevmasm/Instruction.h"
libevmasm/Disassemble.cpp:20:#include "liblangutil/EVMVersion.h"
libevmasm/Disassemble.h:21:#include "liblangutil/EVMVersion.h"
libsolidity/interface/StandardCompiler.cpp:24:#include "libevmasm/Instruction.h"
libyul/AsmAnalysis.cpp:22:#include "libevmasm/Instruction.h"

```

libevmasm/Disassemble.cpp
```diff
@@ -16,6 +16,8 @@
 */
 // SPDX-License-Identifier: GPL-3.0
 
+#include "libevmasm/Instruction.h"
```
stackenbotten, line 19: Coding style error

libevmasm/Disassemble.cpp
```diff
@@ -16,6 +16,8 @@
...
+#include "liblangutil/EVMVersion.h"
```
stackenbotten, line 20: Coding style error

Marenz:
I tried to keep my automatic [Search & Replace](https://github.com/ethereum/solidity/pull/13531/commits/513d5a32658127eb4afa44b45cff526a36e87e30) in that commit.

Tests & style are not yet updated, probably some broken due to the rename.

libevmasm/Disassemble.cpp
```diff
@@ -57,10 +59,10 @@ void solidity::evmasm::eachInstruction(
...
 }
 
-string solidity::evmasm::disassemble(bytes const& _mem, string const& _delimiter)
+string solidity::evmasm::disassemble(bytes const& _mem, langutil::EVMVersion _evmVersion, string const& _delimiter)
 {
 	stringstream ret;
-	eachInstruction(_mem, [&](Instruction _instr, u256 const& _data) {
+	eachInstruction(_mem, _evmVersion, [&](InternalInstruction _instr, u256 const& _data) {
 		if (!isValidInstruction(_instr))
 			ret << "0x" << std::uppercase << std::hex << static_cast<int>(_instr) << _delimiter;
```
ekpyron, line 67: This doesn't look right - in case the byte wasn't a valid instruction, it should output the byte directly, not the internal enum value... so both this and ``eachInstruction`` will probably need to change slightly.

libevmasm/Instruction.h
```diff
@@ -26,11 +26,185 @@
...
+	REVERT,				///< halt execution, revert state and return output data
+	INVALID,			///< invalid instruction for expressing runtime errors (e.g., division-by-zero)
+	SELFDESTRUCT,		///< halt execution and register account for later deletion
+						///
+	MAX_INTERNAL_INSTRUCTION ///< Marker for largest internal instruction value
+};
+
 /// Virtual machine bytecode instruction.
-enum class Instruction: uint8_t
+enum class InstructionOpCode: uint8_t
```
ekpyron, line 207: Do we really still need this one? The instruction info array can just contain the plain bytecode and the traits like ``inline bool isPushInstruction`` can all be easily implemented on top of the other enum, can't they? Not sure I've seen all remaining uses of this yet, though.

Marenz: Oh, good thought. Yeah, it is very very rarely used, I think only in Disassembly.

Marenz: The inline functions are already implemented on top of the internal one, true

libyul/AsmAnalysis.cpp
```diff
@@ -308,31 +309,48 @@ vector<YulString> AsmAnalyzer::operator()(FunctionCall const& _funCall)
...
+			},
+			[&](Scope::Function const& _fun)
+			{
+				parameterTypes = &_fun.arguments;
+				returnTypes = &_fun.returns;
+			}
+		});
+	};
+
+	// TODO remove this in 0.9.0. We allow creating functions or identifiers in Yul with the name
```
nikola-matic, line 331: Do you think it would be a good idea to open an issue preemptively, so we don't forget, or is this something that will actively cause problems once we bump to 0.9.0. and will have to be removed anyway?

libsolidity/analysis/TypeChecker.cpp
aarlt: @ekpyron yeah that makes sense! @Marenz is that tracked somewhere somehow, so we will not forget it?

libsolidity/analysis/TypeChecker.cpp
aarlt: Yeah, I saw that the EIP is using the exact same words - but I needed to check it's translation ;) that's why I thought it could make sense to use an easier wording

Changelog.md
ekpyron: The main point is that we should not *suggest* to use them interchangeably. Supporting both (with a warning on mismatch) is a compromise for not being breaking for hypothetical libraries that want to use it as a source of randomness in a way that can rely on either version - but in general you should still very much use the correct one and not use them interchangeably.

.circleci/soltest_all.sh
```diff
@@ -31,7 +31,7 @@ REPODIR="$(realpath "$(dirname "$0")"/..)"
 # shellcheck source=scripts/common.sh
 source "${REPODIR}/scripts/common.sh"
 
-EVM_VALUES=(homestead byzantium constantinople petersburg istanbul berlin london)
+EVM_VALUES=(homestead byzantium constantinople petersburg istanbul berlin london paris)
```
axic, line 34: @ekpyron @cameel I think in order to make this PR consumable, it would be useful creating a simple one which adds the Paris version in `EVMVersion` and `EVMHost` and runs the test on that version too. **Without** making any changes to `prevrandao`.

Then this PR remains the `prevrandao` saga.

axic: Did this in #13698.

cameel: Sure. As long as we're sure that the prevandao will actually land, having the new EVM version in a separate PR does make the review easier.

axic: This was done & merged.

ekpyron:
Closing this as it will be superseded by https://github.com/ethereum/solidity/pull/13759.
Will keep the branch around as reference for now.

PR #11087, opened 2021-03-11
EVMHost: charge gas for precompiles

axic:
Closes #11057.

PR #13532, opened 2022-09-16
Upgrade to evmc10 + evmone9

axic:
Depends on #13537

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,16 +9,16 @@ version: 2.1
 parameters:
   ubuntu-2004-docker-image:
     type: string
-    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004-14
-    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:d1ef23849db4c5462b248d89c111da4009b153cbd5002cb8755b0580312be581"
+    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004-15
+    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:e4f83457bf1d6475c3189e9013da77289793a5ecd6a0e15dbec9411880b11e22"
   ubuntu-2004-clang-docker-image:
     type: string
-    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004.clang-14
-    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:beb8c91998ec0df99a488900b3723a06f1122f0954fc73786b6c53fd73a6408d"
+    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004.clang-15
+    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:8dda4fdae312f840fbb4e25b9ef01ad3209e9014e49e4564ab0f0d2510225131"
   ubuntu-1604-clang-ossfuzz-docker-image:
     type: string
-    # solbuildpackpusher/solidity-buildpack-deps:ubuntu1604.clang.ossfuzz-19
-    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:8c9bf1813c261d781f4c65fceed2dfb3ecf5be9ecf49bddbd250b570a7f3baea"
+    # solbuildpackpusher/solidity-buildpack-deps:ubuntu1604.clang.ossfuzz-20
+    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:7b3ccaed3b5d37dc2cc4bbe1a1e40949266292dcdbc3ad015271ab4e5e6f5038"
   emscripten-docker-image:
     type: string
     # solbuildpackpusher/solidity-buildpack-deps:emscripten-13
--- a/test/Common.h
+++ b/test/Common.h
@@ -33,19 +33,19 @@ namespace solidity::test
 
 #ifdef _WIN32
 static constexpr auto evmoneFilename = "evmone.dll";
-static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.8.0/evmone-0.8.0-windows-amd64.zip";
+static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.9.1/evmone-0.9.1-windows-amd64.zip";
 static constexpr auto heraFilename = "hera.dll";
-static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/archive/v0.3.2-evmc8.tar.gz";
+static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/archive/v0.6.0.tar.gz";
 #elif defined(__APPLE__)
 static constexpr auto evmoneFilename = "libevmone.dylib";
-static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.8.0/evmone-0.8.0-darwin-x86_64.tar.gz";
+static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.9.1/evmone-0.9.1-darwin-x86_64.tar.gz";
 static constexpr auto heraFilename = "libhera.dylib";
-static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/releases/download/v0.5.0/hera-0.5.0-darwin-x86_64.tar.gz";
+static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/releases/download/v0.6.0/hera-0.6.0-darwin-x86_64.tar.gz";
 #else
 static constexpr auto evmoneFilename = "libevmone.so";
-static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.8.0/evmone-0.8.0-linux-x86_64.tar.gz";
+static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.9.1/evmone-0.9.1-linux-x86_64.tar.gz";
 static constexpr auto heraFilename = "libhera.so";
-static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/releases/download/v0.5.0/hera-0.5.0-linux-x86_64.tar.gz";
+static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/releases/download/v0.6.0/hera-0.6.0-linux-x86_64.tar.gz";
 #endif
 
 struct ConfigException: public util::Exception {};
--- a/test/EVMHost.cpp
+++ b/test/EVMHost.cpp
@@ -125,7 +125,7 @@ EVMHost::EVMHost(langutil::EVMVersion _evmVersion, evmc::VM& _vm):
 	else
 		assertThrow(false, Exception, "Unsupported EVM version");
 
-	tx_context.block_difficulty = evmc::uint256be{200000000};
+	tx_context.block_prev_randao = evmc::uint256be{200000000}; // TODO: should make it >2**64 for >Paris
 	tx_context.block_gas_limit = 20000000;
 	tx_context.block_coinbase = 0x7878787878787878787878787878787878787878_address;
 	tx_context.tx_gas_price = evmc::uint256be{3000000000};
@@ -145,7 +145,6 @@ EVMHost::EVMHost(langutil::EVMVersion _evmVersion, evmc::VM& _vm):
 void EVMHost::reset()
 {
 	accounts.clear();
-	m_currentAddress = {};
 	// Clear self destruct records
 	recorded_selfdestructs.clear();
 	// Clear call records
@@ -194,14 +193,14 @@ void EVMHost::transfer(evmc::MockedAccount& _sender, evmc::MockedAccount& _recip
 	_recipient.balance = convertToEVMC(u256(convertFromEVMC(_recipient.balance)) + _value);
 }
 
-void EVMHost::selfdestruct(const evmc::address& _addr, const evmc::address& _beneficiary) noexcept
+bool EVMHost::selfdestruct(const evmc::address& _addr, const evmc::address& _beneficiary) noexcept
 {
 	// TODO actual selfdestruct is even more complicated.
 
 	transfer(accounts[_addr], accounts[_beneficiary], convertFromEVMC(accounts[_addr].balance));
 
 	// Record self destructs. Clearing will be done in newTransactionFrame().
-	MockedHost::selfdestruct(_addr, _beneficiary);
+	return MockedHost::selfdestruct(_addr, _beneficiary);
 }
 
 void EVMHost::recordCalls(evmc_message const& _message) noexcept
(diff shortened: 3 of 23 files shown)


axic:
Need to pull out the docker changes, but first need to cut a new Hera release too.

test/Common.h
```diff
@@ -33,17 +33,17 @@ namespace solidity::test
...
 #elif defined(__APPLE__)
 static constexpr auto evmoneFilename = "libevmone.dylib";
-static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.8.0/evmone-0.8.0-darwin-x86_64.tar.gz";
+static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.9.1/evmone-0.9.1-darwin-x86_64.tar.gz";
 static constexpr auto heraFilename = "libhera.dylib";
 static constexpr auto heraDownloadLink = "https://github.com/ewasm/hera/releases/download/v0.5.0/hera-0.5.0-darwin-x86_64.tar.gz";
 #else
 static constexpr auto evmoneFilename = "libevmone.so";
-static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.8.0/evmone-0.8.0-linux-x86_64.tar.gz";
+static constexpr auto evmoneDownloadLink = "https://github.com/ethereum/evmone/releases/download/v0.9.1/evmone-0.9.1-linux-x86_64.tar.gz";
```
axic, line 46: These binaries are not created yet, cc @chfast.

test/EVMHost.cpp
```diff
@@ -127,7 +125,10 @@ EVMHost::EVMHost(langutil::EVMVersion _evmVersion, evmc::VM& _vm):
 	else
 		assertThrow(false, Exception, "Unsupported EVM version");
 
-	tx_context.block_difficulty = evmc::uint256be{200000000};
+	if (m_evmRevision >= EVMC_PARIS)
+		tx_context.block_prev_randao = 0xa86c2e601b6c44eb4848f7d23d9df3113fbcac42041c49cbed5000cb4f118777_bytes32;
```
axic, line 131: This is the value from the merge block.

axic:
Since evmc10 supports refund accounting, after this change the tests could include that.

test/Common.h
chfast: They should be there now.

axic:
Pulled out buildpack to #13537

test/EVMHost.cpp
cameel: It would probably be better to add that as a comment in the code.

test/EVMHost.cpp
```diff
@@ -147,7 +148,6 @@ EVMHost::EVMHost(langutil::EVMVersion _evmVersion, evmc::VM& _vm):
 void EVMHost::reset()
 {
 	accounts.clear();
-	m_currentAddress = {};
```
chfast, line 148: This is not needed any more because EVMC provides additional "code address" in message.

test/EVMHost.cpp
```diff
@@ -189,14 +189,17 @@ void EVMHost::transfer(evmc::MockedAccount& _sender, evmc::MockedAccount& _recip
...
 
 	transfer(accounts[_addr], accounts[_beneficiary], convertFromEVMC(accounts[_addr].balance));
 	accounts.erase(_addr);
-	// Record self destructs
-	recorded_selfdestructs.push_back({_addr, _beneficiary});
+
+	// Record self destructs.
+	auto& beneficiaries = recorded_selfdestructs[_addr];
+	beneficiaries.emplace_back(_beneficiary);
+	return beneficiaries.size() == 1;
```
chfast: You can reuse MockedHost.

```suggestion
	return MockedHost::selfdestruct(_addr, _beneficiary);
```

test/EVMHost.cpp
```diff
@@ -205,24 +208,26 @@ void EVMHost::recordCalls(evmc_message const& _message) noexcept
...
 }
 
-evmc::result EVMHost::call(evmc_message const& _message) noexcept
+// NOTE: this is used for both internal and external calls.
+// External calls are triggered from ExecutionFramework and contain only EVMC_CREATE or EVMC_CALL.
+evmc::Result EVMHost::call(evmc_message const& _message) noexcept
 {
 	recordCalls(_message);
-	if (_message.destination == 0x0000000000000000000000000000000000000001_address)
+	if (_message.recipient == 0x0000000000000000000000000000000000000001_address)
```
chfast, line 222: Not sure how much this is worth here, but you can do `0x01_address` now.

test/EVMHost.cpp
```diff
@@ -240,7 +245,7 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
 			message.gas -= message.input_data[i] == 0 ? evmasm::GasCosts::txDataZeroGas : evmasm::GasCosts::txDataNonZeroGas(m_evmVersion);
 		if (message.gas < 0)
 		{
-			evmc::result result({});
+			evmc::Result result({});
```
chfast:
```suggestion
			evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -266,13 +271,13 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
-			accounts[message.destination].nonce > 0 ||
-			!accounts[message.destination].code.empty()
+		message.recipient = convertToEVMC(createAddress);
+		if (accounts.count(message.recipient) && (
+			accounts[message.recipient].nonce > 0 ||
+			!accounts[message.recipient].code.empty()
 		))
 		{
-			evmc::result result({});
+			evmc::Result result({});
```
chfast:
```suggestion
			evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -282,19 +287,20 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
 	}
 	else if (message.kind == EVMC_DELEGATECALL || message.kind == EVMC_CALLCODE)
 	{
-		code = accounts[message.destination].code;
-		message.destination = m_currentAddress;
+		code = accounts[message.code_address].code;
```
chfast: This should work for any call now (if I'm not wrong) so the `else` branch can be eliminated.

test/EVMHost.cpp
```diff
@@ -282,19 +287,20 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
 
-	auto& destination = accounts[message.destination];
+	auto& destination = accounts[message.recipient];
 
 	if (value != 0 && message.kind != EVMC_DELEGATECALL && message.kind != EVMC_CALLCODE)
 	{
 		if (value > convertFromEVMC(sender.balance))
 		{
-			evmc::result result({});
+			evmc::Result result({});
```
chfast:
```suggestion
			evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -407,14 +410,14 @@ evmc::result EVMHost::precompileSha256(evmc_message const& _message) noexcept
 		_message.input_data + _message.input_size
 	));
 
-	evmc::result result({});
+	evmc::Result result({});
```
chfast:
```suggestion
	evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -480,27 +483,27 @@ evmc::result EVMHost::precompileRipeMD160(evmc_message const& _message) noexcept
...
 }
 
-evmc::result EVMHost::precompileIdentity(evmc_message const& _message) noexcept
+evmc::Result EVMHost::precompileIdentity(evmc_message const& _message) noexcept
 {
 	// static data so that we do not need a release routine...
 	bytes static data;
 	data = bytes(_message.input_data, _message.input_data + _message.input_size);
-	evmc::result result({});
+	evmc::Result result({});
```
chfast:
```suggestion
	evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -480,27 +483,27 @@ evmc::result EVMHost::precompileRipeMD160(evmc_message const& _message) noexcept
...
 	result.output_size = data.size();
 	return result;
 }
 
-evmc::result EVMHost::precompileModExp(evmc_message const&) noexcept
+evmc::Result EVMHost::precompileModExp(evmc_message const&) noexcept
 {
 	// TODO implement
-	evmc::result result({});
+	evmc::Result result({});
```
chfast:
```suggestion
	evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -787,18 +790,18 @@ evmc::result EVMHost::precompileGeneric(
 		return resultWithGas(_message, _inOut.at(input));
 	else
 	{
-		evmc::result result({});
+		evmc::Result result({});
```
chfast:
```suggestion
		evmc::Result result;
```

test/EVMHost.cpp
```diff
@@ -787,18 +790,18 @@ evmc::result EVMHost::precompileGeneric(
...
 }
 
-evmc::result EVMHost::resultWithGas(
+evmc::Result EVMHost::resultWithGas(
 	evmc_message const& _message,
 	bytes const& _data
 ) noexcept
 {
-	evmc::result result({});
+	evmc::Result result({});
```
chfast:
```suggestion
	evmc::Result result;
```

test/EVMHost.cpp
axic: It doesn't actually matter what it is.

axic: Added.

test/EVMHost.cpp
axic: I think explicit conversion from integer would be more useful here (i.e. `== 1)`, but otherwise probably fine to just leave as is imho.

axic:
Let's see if #13594 simplifies this too.

cameel:
What's the status here? Is it ready to be merged along with #13537?

axic:
> What's the status here?

I'm still unclear about some of the gas changes, but if we don't want to investigate further it is good to merge (after some squashing).

cameel:
Ah, ok, I do see the huge drops in some tests. I think it would be better to investigate. We're sometimes basing our optimizer decisions on these costs so better to have some confidence in them.

ekpyron:
Marking this as "priority review" for investigating the gas changes, since this will is and will even more so become a blocker for other work.

test/EVMHost.cpp
```diff
@@ -174,10 +178,14 @@ void EVMHost::newTransactionFrame()
...
 	// Clear EIP-2929 account access indicator
 	recorded_account_accesses.clear();
-	// Clear EIP-2929 storage access indicator
 	for (auto& [address, account]: accounts)
 		for (auto& [slot, value]: account.storage)
+		{
+			// Clear EIP-2929 storage access indicator
 			value.access_status = EVMC_ACCESS_COLD;
+			// Mark storage values for proper refund calculation.
+			value.original = value.current;
```
axic: This is an important change needed for correct calculation. The new system has three variables: original (value prior to the transaction), previous (... value in the transaction), current (... value in the transaction).

test/EVMHost.cpp
```diff
@@ -295,35 +303,32 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
-			evmc::result result({});
+			evmc::Result result;
 			result.status_code = EVMC_OUT_OF_GAS;
 			accounts = stateBackup;
 			return result;
 		}
 
 		code = evmc::bytes(message.input_data, message.input_data + message.input_size);
 	}
-	else if (message.kind == EVMC_DELEGATECALL || message.kind == EVMC_CALLCODE)
```
axic, line 330: This is an important change, in the new API there's a distinction between `recipient` and `code_address`, no need to manually track it.

test/ExecutionFramework.cpp
```diff
@@ -174,29 +175,36 @@ void ExecutionFramework::sendMessage(bytes const& _data, bool _isCreation, u256
...
 
-	evmc::result result = m_evmcHost->call(message);
+	evmc::Result result = m_evmcHost->call(message);
 
 	m_output = bytes(result.output_data, result.output_data + result.output_size);
 	if (_isCreation)
 		m_contractAddress = EVMHost::convertFromEVMC(result.create_address);
 
 	m_gasUsed = InitialGas - result.gas_left;
+	unsigned const refundRatio = (m_evmVersion >= langutil::EVMVersion::london() ? 5 : 2);
```
axic: This change adds support for calculating refunds, which was not the case prior.

test/ExecutionFramework.cpp
```diff
@@ -210,11 +218,12 @@ void ExecutionFramework::sendEther(h160 const& _addr, u256 const& _amount)
...
 			cout << " value: " << _amount << endl;
 	}
-	evmc_message message = {};
+	evmc_message message{};
 	message.sender = EVMHost::convertToEVMC(m_sender);
 	message.value = EVMHost::convertToEVMC(_amount);
 	message.kind = EVMC_CALL;
-	message.destination = EVMHost::convertToEVMC(_addr);
+	message.recipient = EVMHost::convertToEVMC(_addr);
+	message.code_address = message.recipient;
```
axic, line 221: `code_address` is the same as this is a regular call. It only differs for delegatecall/callcode.

test/ExecutionFramework.cpp
ekpyron: Can we decouple this change from the actual evmc/evmone upgrade? I mean, we should definitely fix this, but mixing it together here makes it rather unclear what causes what in the PR.

axic: It is a separate commit, so already decoupled, the differences exists before and after it.

axic:
Rebased after #11087, but tests not updated yet.

.circleci/config.yml
```diff
@@ -9,16 +9,16 @@ version: 2.1
 parameters:
   ubuntu-2004-docker-image:
     type: string
-    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004-14
-    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:d1ef23849db4c5462b248d89c111da4009b153cbd5002cb8755b0580312be581"
+    # solbuildpackpusher/solidity-buildpack-deps:ubuntu2004-15
+    default: "solbuildpackpusher/solidity-buildpack-deps@sha256:f609a27be985e3dae160116139b471252a34cd0d3dc949a8a86be84b8c106f7b"
```
axic: Need to update hashes prior to merging.

test/ExecutionFramework.cpp
axic: Moved to #13695.

axic:
Ready for final review.

test/EVMHost.cpp
```diff
@@ -313,35 +317,32 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
+	else
 	{
```
ekpyron: Minor coding style thing: no need for brackets anymore here.

chfast (dismissed):

test/EVMHost.cpp
```diff
@@ -449,11 +447,11 @@ evmc::result EVMHost::precompileECRecover(evmc_message const& _message) noexcept
 			}
 		}
 	};
-	evmc::result result = precompileGeneric(_message, inputOutput);
+	evmc::Result result = precompileGeneric(_message, inputOutput);
 	// ECRecover will return success with empty response in case of failure
 	if (result.status_code != EVMC_SUCCESS && result.status_code != EVMC_OUT_OF_GAS)
-//		return resultWithGas(_message.gas, gas_cost, {});
 	{
+		// Could use resultWithGas(_message.gas, gas_cost, {}) here, but it expects a reference.
```
ekpyron: Looking at this here now, I actually see that ``resultWithGas`` *doesn't* copy, but stores a pointer in ``result.output_data``...
Is evmc robust against having "garbage" in ``result.output_data`` as long as ``result.output_size`` is zero?
If so, this is still fine. If not, we may actually want to set ``result.output_data`` to ``nullptr`` in ``EVMHost::resultWithGas`` if ``_data`` is empty.

But yeah, if too much hassle, we can also leave it at this comment, if need be ;-).

test/EVMHost.cpp
```diff
@@ -449,11 +447,11 @@ evmc::result EVMHost::precompileECRecover(evmc_message const& _message) noexcept
...
 		result.status_code = EVMC_SUCCESS;
 		result.gas_left = _message.gas - gas_cost;
 		result.output_data = {};
```
ekpyron: I'd actually have made this
```suggestion
		result.output_data = nullptr;
```
to make it clearer what this is, independently of https://github.com/ethereum/solidity/pull/13532/files#r1017828619

test/EVMHost.cpp
axic: https://github.com/ethereum/solidity/blob/develop/test/evmc/evmc.h#L357-L375 says that `output_data` MAY be null, and if it is, `output_size` MUST be null. Doesn't state that if `output_size` is null then `output_data` MUST NOT be accessed. @chfast do you have any comments?

test/EVMHost.cpp
axic: Fixed.

test/EVMHost.cpp
axic: Fixed.

test/EVMHost.cpp
axic: Could also change `resultWithGas` such that
```diff
        result.output_data = _data.data();
        result.output_size = _data.size();
+       if (result.output_size == 0)
+               result.output_data = nullptr;
```

axic: Could also have an overload of `resultWithGas` without a `_data` field 🤓

ekpyron: > Could also change `resultWithGas` such that
> 
> ```diff
>         result.output_data = _data.data();
>         result.output_size = _data.size();
> +       if (result.output_size == 0)
> +               result.output_data = nullptr;
> ```

Or
```
result.output_data = _data.empty() ? nullptr : _data.data();
result.output_size = _data.size();
```
That's what I meant above. Actually unfortunate in the C++ standard that ``.data()`` specs as "If size() is ​0​, data() may or may not return a null pointer.", but what can one do...

chfast: EVM will handle non-null pointer without problem so `_data.data(), _data.size()` should always be fine.

axic: > EVM will handle non-null pointer without problem so _data.data(), _data.size() should always be fine.

Can we put the requirement in writing in `evmc.h` not to dereference `output_data` if `output_size` is 0?

chfast: Yes, you can put it there.

axic: Added in https://github.com/ethereum/evmc/pull/673.

@ekpyron do you still want to add the change or can we merge?

ekpyron: Given https://github.com/ethereum/evmc/pull/673, I'd say let's switch to the currently commented version here, but with that we can consider it ready to merge.

axic: Updated, @ekpyron please review.

ekpyron (approved):

Issue #11593, opened 2021-06-29
Remove abi.encodePacked


axic:
`abi.encodePacked` was introduced in 0.4.22 as a backwards compatibility measure for the irregular "packed" encoding.

Since then we have introduced `bytes.concat` (#10903) and learned that most projects (rightly) do not rely on the packed encoding, but only concatenation. I suggest we drop `abi.encodePacked` in the next breaking release.

(We could consider introducing a message with #11508 if we want to test the waters 😅 , but would rather refrain from this.)

chriseth:
Which types can be used with `encodePacked` that cannot be used with `bytes.concat`?

charles-cooper:
> Which types can be used with `encodePacked` that cannot be used with `bytes.concat`?

Arrays. https://docs.soliditylang.org/en/v0.8.6/abi-spec.html.

What is the effect of this proposal on event index encoding?

chriseth:
Event index encoding will not change, since we only plan to remove `abi.encodePacked` from the language.

Is packed encoding of arrays used in the real world?

e-nikolov:
> Is packed encoding of arrays used in the real world?

One place where it can be used is for EIP-712 hashes - https://eips.ethereum.org/EIPS/eip-712#definition-of-encodedata.

For example if there is a struct that contains an array:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

contract MultiTransferDemo {
    struct MultiTransfer {
        uint256 id;
        Transfer[] transfers;
    }

    struct Transfer {
        uint256 amount;
        address recepient;
    }

    function hash(MultiTransfer memory mt) public pure returns (bytes32) {
        return
            keccak256(
                abi.encode(
                    keccak256(
                        "MultiTransfer(uint256 id,Transfer[] transfers)Transfer(uint256 amount,address recepient)"
                    ),
                    mt.id,
                    hashTransfers(mt.transfers)
                )
            );
    }

    function hashTransfers(Transfer[] memory transfers)
        public
        pure
        returns (bytes32)
    {
        bytes32[] memory hashedTransfers = new bytes32[](transfers.length);
        for (uint256 i = 0; i < transfers.length; i++) {
            hashedTransfers[i] = hashTransfer(transfers[i]);
        }
        return keccak256(abi.encodePacked(hashedTransfers));
    }

    function hashTransfer(Transfer memory transfer)
        public
        pure
        returns (bytes32)
    {
        return
[...]

PR #13533, opened 2022-09-17
EVMHost: Calculate address for CREATE properly

### LINKED ISSUE #12737: Wrong address computed for CREATE2 in semantic tests
## Description
The example we have in our documentation for computing the address that `create2` would deploy a contract to ([Salted contract creations / create2](https://docs.soliditylang.org/en/latest/control-structures.html#salted-contract-creations-create2)) does not pass when added as a semantic test. The addresses are different and the `require` fails.

I think that the code is correct and it fails due to a bug either in isoltest or in evmone.

## Steps to Reproduce
```solidity
contract D {
    uint public x;
    constructor(uint a) {
        x = a;
    }
}

contract C {
    function createDSalted(bytes32 salt, uint arg) public {
        address predictedAddress = address(uint160(uint(keccak256(abi.encodePacked(
            bytes1(0xff),
            address(this),
            salt,
            keccak256(abi.encodePacked(
                type(D).creationCode,
                arg
            ))
        )))));

        D d = new D{salt: salt}(arg);
        require(address(d) == predictedAddress, "Address mismatch.");
    }
}
// ====
// compileViaYul: also
// ----
// createDSalted(bytes32,uint256): 42, 64 ->

```
```
Running 3 test cases...
/solidity/test/boostTest.cpp(111): error: in "semanticTests/salted_create/salted_create_deterministic_address": Test expectation mismatch.
Expected result:
// createDSalted(bytes32,uint256): 42, 64 ->

Obtained result:
// createDSalted(bytes32,uint256): 42, 64 -> FAILURE, hex"08c379a0", 0x20, 0x11, "Address mismatch."
Warning: The call to "createDSalted(bytes32,uint256)" returned
[8,c3,79,a0]
[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,20]
[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,11]
[41,64,64,72,65,73,73,20,6d,69,73,6d,61,74,63,68,2e,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]

Attention: Updates on the test will apply the detected format displayed.

Note that the test also has to pass via Yul.


*** 1 failure is detected in the test module "SolidityTests"
```

axic:
Fixes #12737.


test/EVMHost.cpp
```diff
@@ -249,13 +249,33 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
 
 	if (message.kind == EVMC_CREATE)
 	{
-		// TODO this is not the right formula
 		// TODO is the nonce incremented on failure, too?
```
axic, line 253: So this is actually complex and not worth implementing here potentially.

test/EVMHost.cpp
```diff
@@ -249,13 +249,33 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
+		sender.nonce++;
+
+		auto encodeRlpInteger = [](int value) -> bytes {
+			if (value == 0) {
+				return bytes(1, 128);
```
chfast:
```suggestion
				return bytes{128};
```
You can create it like this.

test/EVMHost.cpp
```diff
@@ -249,13 +249,33 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
+			} else if (value <= 127) {
+				return bytes(1, static_cast<uint8_t>(value));
+			} else if (value <= 0xff) {
+				return bytes(1, 128 + 1) + bytes(1, static_cast<uint8_t>(value));
+			} else if (value <= 0xffff) {
+				return bytes(1, 128 + 55 + 2) + bytes(1, static_cast<uint8_t>(value >> 8)) + bytes(1, static_cast<uint8_t>(value));
```
chfast:
```suggestion
				return bytes{128 + 55 + 2, static_cast<uint8_t>(value >> 8)), static_cast<uint8_t>(value)};
```

test/EVMHost.cpp
```diff
@@ -249,13 +249,33 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
 
 	if (message.kind == EVMC_CREATE)
 	{
-		// TODO this is not the right formula
 		// TODO is the nonce incremented on failure, too?
+		sender.nonce++;
```
hrkrshnn, line 256: Shouldn't this start at 0? So the increment should be after, right?

test/EVMHost.cpp
hrkrshnn: Are you sure about this?

```
>>> cast --to-rlp '["0xf467baad569be981fb1a781f640de36059cd774a", "00"]'
0xd694f467baad569be981fb1a781f640de36059cd774a00
```

Seems to be mapping to 0?

test/libsolidity/semanticTests/various/create_random.sol
```diff
@@ -0,0 +1,23 @@
...
+        assembly {
+            a1 := create(0, add(creation, 32), mload(creation))
+            a2 := create2(0, add(creation, 32), mload(creation), address())
+        }
+    }
+}
+
+// ----
+// addr() -> 0xbe9efd422ee479f010753a6bc06afe3a8444fc00
+// test() -> 0xf467baad569be981fb1a781f640de36059cd774a, 0x4e76c5b8db09c0f1be7f40921d7fec817e358ab7
```
hrkrshnn: The values don't quite match
```
>>> cast compute-address 0xbe9efd422ee479f010753a6bc06afe3a8444fc00 --nonce 0
Computed Address: 0xC1397e3D7f722ba7F2BBA94054a8f2D3D66E296A
>>> cast compute-address 0xbe9efd422ee479f010753a6bc06afe3a8444fc00 --nonce 1
Computed Address: 0x640DE36059cD774a4fd957bf6Cf8EEF4491E431c
>>> cast compute-address 0xbe9efd422ee479f010753a6bc06afe3a8444fc00 --nonce 2
Computed Address: 0x5e01F63D4C8ea2060e389A2bC5c408CFCC02e421
```

test/EVMHost.cpp
chfast: Yes. The address should be computed before the sender's nonce is bumped. E.g. https://github.com/ethereum/evmone/pull/484/files#diff-a6fe16f4d5493d757459aa2ee53b83d1e7761259fe171a0aed72ea950d2393d5R155-R159

test/EVMHost.cpp
chfast: This is right. Usually when encoding integers the leading zero bytes are eliminated before byte-string encoding. In case of 0 you end up with empty bytes `""` encoded as `0x80`.

test/EVMHost.cpp
hrkrshnn: @axic told me this is different when the sender is a contract and that the nonces for it starts at 1. But didn't check it though.

test/EVMHost.cpp
hrkrshnn: Okay, that adds up
```
>>> cast --to-rlp '["0xf467baad569be981fb1a781f640de36059cd774a", ""]'
0xd694f467baad569be981fb1a781f640de36059cd774a80
```

test/EVMHost.cpp
chfast: This is a bit different matter. When contract is created it has nonce 1.

test/EVMHost.cpp
axic: Done.

test/libsolidity/semanticTests/various/create_random.sol
axic:
```
create from: 1212121212121212121212121212120000000012 nonce: 1 -> be9efd422ee479f010753a6bc06afe3a8444fc00
create from: be9efd422ee479f010753a6bc06afe3a8444fc00 nonce: 1 -> f467baad569be981fb1a781f640de36059cd774a
```

So it turns out the problem is the conversion from h256 -> h160 is truncated from the wrong side, left-aligned vs. right-aligned.

test/libsolidity/semanticTests/various/create_random.sol
```diff
@@ -0,0 +1,25 @@
...
+            a1 := create(0, add(creation, 32), mload(creation))
+            a2 := create2(0, add(creation, 32), mload(creation), address())
+        }
+    }
+}
+
+// ====
+// compileViaYul: false
+// ----
+// addr() -> 0xc06afe3a8444fc0004668591e8306bfb9968e79e
```
hrkrshnn, line 37:
```
cast compute-address 0x1212121212121212121212121212120000000012 --nonce 1
Computed Address: 0xc06afE3A8444fc0004668591e8306bfB9968E79E
```
Should be really `nonce 0`.

hrkrshnn: Don't remember how the testing framework works. I'm assuming that `contract A` is not deployed directly. 
edit: yeah, this matches with other tests.

axic: Just read the code, there's a TODO for that: https://github.com/ethereum/solidity/pull/13533/files#diff-0b947a700f82b453dfef24ae4301f8e1f511428310cd15130bb90758fcde08c8R255

axic: There are two ways to do it:
1) check if the account has code
2) when creating an account set the nonce to 1

test/libsolidity/semanticTests/various/create_random.sol
```diff
@@ -0,0 +1,20 @@
+contract C {
+    function addr() external returns (address) {
+        return address(this);
+    }
+
+    function test() external returns (address a1, address a2) {
+        assembly {
+            // This is `return(0, 1)`. We are using a simplified/fixed initcode to avoid
+            // instability due to metadata changes.
+            let initcode := hex"60016000f3"
```
axic, line 10: Changed initcode so the test is not depending on metadata.

test/EVMHost.cpp
```diff
@@ -250,23 +250,48 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
 		message.destination = convertToEVMC(createAddress);
+		assertThrow(accounts.count(message.destination) == 0, Exception, "Account cannot exist");
+
 		code = evmc::bytes(message.input_data, message.input_data + message.input_size);
 	}
 	else if (message.kind == EVMC_CREATE2)
 	{
 		h160 createAddress(keccak256(
-			bytes(1, 0xff) +
+			bytes{0xff} +
```
cameel, line 289: Is this correct? I think that `bytes(1, 0xff)` is a single-element vector containing `0xff` while `bytes{0xff}` is a vector of 255 zeros.

cameel: Ah, wait. It's ok. It's an initializer list now.

test/EVMHost.cpp
```diff
@@ -250,23 +250,48 @@ evmc::result EVMHost::call(evmc_message const& _message) noexcept
...
 			bytes(begin(message.sender.bytes), end(message.sender.bytes)) +
 			bytes(begin(message.create2_salt.bytes), end(message.create2_salt.bytes)) +
 			keccak256(bytes(message.input_data, message.input_data + message.input_size)).asBytes()
-		));
+		), h160::AlignRight);
```
cameel: Is this calculation correct? Looks like the changes here should not affect the result, which means it must have been correct already. But when I tried to do `CREATE2` address calculation in the past I was not getting the expected result: #12737.

test/libsolidity/semanticTests/various/create_random.sol
```diff
@@ -0,0 +1,22 @@
...
+            mstore(0, initcode)
+
+            a1 := create(0, 0, 5)
+            a2 := create2(0, 0, 5, address())
```
cameel, line 14: Wouldn't it be safer to use free mem ptr here? There's nothing explicitly using memory in the test, but still, do we guarantee that the compiler will never to anything that allocates memory on its own?

cameel (dismissed):
I haven't compared the calculation with the spec, but since @hrkrshnn checked the addresses using foundry, it must be correct.

test/libsolidity/semanticTests/various/create_random.sol
axic: `0` is the scratch space: https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_memory.html

If we change this assumption it will be good to have many breaking tests, because many people/contracts actually depend on this.

test/EVMHost.cpp
axic: I haven't verified it against `create2`, but wanted to. Will do it next. However the change of alignment likely fixes it, because the input looked correct.

axic: Here's a forge test verifying it is correct now:
```
    function calculateCreate2(address creator, bytes32 codehash, bytes32 salt) private pure returns (address) {
        return address(uint160(uint256(keccak256(abi.encodePacked(bytes1(0xff), creator, salt, codehash)))));
    }

    function testcreate2() public {
       assertEq(calculateCreate2(0xc06afE3A8444fc0004668591e8306bfB9968E79E, keccak256(hex"60016000f3"), bytes32(uint256(uint160(0xc06afE3A8444fc0004668591e8306bfB9968E79E)))),0x2C1C30623DDD93e0b765A6CAACA0C859eeb0644d);
    }
```

cameel: Cool. Can you add the repro from the issue as a test here? It's basically the example from the docs.

test/libsolidity/semanticTests/various/create_random.sol
cameel: Ah, my bad. I know about the scratch space but I just realized that the code is writing just 5 bytes not 5 full slots. So yeah, it won't overwrite the pointer.

test/EVMHost.cpp
axic: Will add this into the test.

axic: Done.

test/libsolidity/semanticTests/salted_create/prediction_example.sol
```diff
@@ -0,0 +1,29 @@
...
+            ))
+        )))));
+
+        D d = new D{salt: salt}(arg);
+        require(address(d) == predictedAddress, "Address mismatch.");
+    }
+}
+// ====
+// EVMVersion: >=constantinople
+// compileViaYul: also
```
cameel, line 26: Actually, we don't need to set `compileViaYul: also` explicitly anymore. This is now the default. I think the example in the issue comes from before the time when we did changed it.

test/libsolidity/semanticTests/various/create_random.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    function testCalc() external returns (address a1, address a2) {
+        a1 = calculateCreate(address(this), 1);
+        a2 = calculateCreate2(address(this), keccak256(hex"60016000f3"), bytes32(uint256(uint160(address(this)))));
+    }
+
+    function calculateCreate(address from, uint256 nonce) private pure returns (address) {
+        assert(nonce <= 127);
+        bytes memory data =
+            bytes.concat(hex"d694", bytes20(uint160(from)), nonce == 0 ? bytes1(hex"80") : bytes1(uint8(nonce)));
```
cameel, line 26: What do you think about allowing `address` in `bytes.concat()` and `string.concat()` without conversion? https://github.com/ethereum/solidity/issues/11593#issuecomment-1109688253

I mean, in general, not in this PR of course.

cameel (approved):

test/libsolidity/semanticTests/various/create_random.sol
axic: Left a comment there.

test/libsolidity/semanticTests/salted_create/prediction_example.sol
axic: I just copy pasted from the issue 😅

Do you want it to be changed?

cameel: I know, not complaining, just pointing it out :)

I guess would be better to change it but it's very minor so feel free to also just merge it if you don't want to bother.

axic: Would be nice to avoid another 1hr CI run.

hrkrshnn:
The address calculation for EOA isn't correct right? The problem is that this assumes nonces start at 0. Would be nice to at least open an issue about it.

PR #13535, opened 2022-09-17
Yul optimizer: replace add with sub when it would save gas

### LINKED ISSUE #6765: [Yul] Re-introduce ``sub`` opcode
During CSE, every `sub` by a constant is replaced by an `add` to get uniform expressions. At the end of the optimization phase, after constant-reintroduction, this should be reversed if the cost of the modified constant is better.

The reason to do this at the end: SUB and ADD are more or less equivalent, the main difference is that ADD is commutative, so it is preferred. Only having ADD instead of ADD and SUB also simplifies the rules.

gumb0:
Trying to fix https://github.com/ethereum/solidity/issues/6765.

This all feels very ad-hoc and hacky, I'm not sure at all if this makes sense and is at the reasonable place.

But so far some small example works for me, and some tests are broken.

### DIFF
--- a/libyul/optimiser/ExpressionSimplifier.cpp
+++ b/libyul/optimiser/ExpressionSimplifier.cpp
@@ -21,11 +21,13 @@
 
 #include <libyul/optimiser/ExpressionSimplifier.h>
 
-#include <libyul/optimiser/SimplificationRules.h>
-#include <libyul/optimiser/OptimiserStep.h>
-#include <libyul/optimiser/OptimizerUtilities.h>
 #include <libyul/AST.h>
 #include <libyul/Utilities.h>
+#include <libyul/backends/evm/EVMDialect.h>
+#include <libyul/backends/evm/EVMMetrics.h>
+#include <libyul/optimiser/OptimiserStep.h>
+#include <libyul/optimiser/OptimizerUtilities.h>
+#include <libyul/optimiser/SimplificationRules.h>
 
 #include <libevmasm/SemanticInformation.h>
 
@@ -35,7 +37,7 @@ using namespace solidity::yul;
 
 void ExpressionSimplifier::run(OptimiserStepContext& _context, Block& _ast)
 {
-	ExpressionSimplifier{_context.dialect}(_ast);
+	ExpressionSimplifier{_context.dialect, _context.expectedExecutionsPerDeployment}(_ast);
 }
 
 void ExpressionSimplifier::visit(Expression& _expression)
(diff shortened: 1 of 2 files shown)


gumb0:
image

libyul/optimiser/ExpressionSimplifier.cpp
```diff
@@ -63,6 +71,33 @@ void ExpressionSimplifier::visit(Expression& _expression)
 					)
 						startArgument = Literal{debugDataOf(startArgument), LiteralKind::Number, "0"_yulstring, {}};
 				}
+
+			// Replace add(X, A) with sub(X, -A) if it would save gas.
```
gumb0, line 75: This seemed like a good place for this, because `ExpressionSimplifier` is where `sub -> add` replacements mentioned in the issue are done.

And this is outside of main `SimplificationRules` loop, because it has to be done after all simplifications, and also to be able to estimate gas savings.

libyul/optimiser/ExpressionSimplifier.cpp
```diff
@@ -63,6 +71,33 @@ void ExpressionSimplifier::visit(Expression& _expression)
...
+			{
+				Expression const& arg1 = functionCall->arguments.at(0);
+				Expression& arg2 = functionCall->arguments.at(1);
+				if (auto const* arg2Literal = get_if<Literal>(&arg2))
+				{
+					// Estimate gas savings.
+					bigint const costOfAdd = gasMeter.costs(_expression);
+					Identifier const identifierSub = Identifier{
+						debugDataOf(functionCall->functionName),
+						dynamic_cast<EVMDialect const&>(m_dialect).builtin("sub"_yulstring)->name};
```
gumb0, line 86: There probably should be some less awkward way to get builtin's identifier.

hrkrshnn:
Left some comments in https://github.com/ethereum/solidity/issues/6765#issuecomment-1258687946

nishant-sachdeva:
@gumb0  Are you still working on this ? Or shall we take it over

gumb0:
> @gumb0 Are you still working on this ? Or shall we take it over

Not at the moment, feel free to take it over.

chriseth:
Shouldn't this be a peephole optimizer step?

chriseth:
Thanks a lot for your contribution, @gumb0 . Unfortunately, I don't think we can use it like this because the optimization is still done during the main optimizer loop, where the goal is to simplify expressions (and also unify them) instead of optimizing gas usage. We should add that at the very end somehow, maybe in the peephole optimizer.

PR #13537, opened 2022-09-18
Upgrade to evmone 0.9.1 and hera 0.6.0 in buildpack

axic:
Pulled out from #13532. Should be merged together with it.

### DIFF
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
 #------------------------------------------------------------------------------
 FROM buildpack-deps:focal AS base
-LABEL version="14"
+LABEL version="15"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
@@ -48,7 +48,7 @@ FROM base AS libraries
 # EVMONE
 RUN set -ex; \
 	cd /usr/src; \
-	git clone --branch="v0.8.0" --recurse-submodules https://github.com/ethereum/evmone.git; \
+	git clone --branch="v0.9.1" --recurse-submodules https://github.com/ethereum/evmone.git; \
 	cd evmone; \
 	mkdir build; \
 	cd build; \
@@ -60,7 +60,7 @@ RUN set -ex; \
 # HERA
 RUN set -ex; \
 	cd /usr/src; \
-	git clone --branch="v0.5.0" --depth 1 --recurse-submodules https://github.com/ewasm/hera.git; \
+	git clone --branch="v0.6.0" --depth 1 --recurse-submodules https://github.com/ewasm/hera.git; \
 	cd hera; \
 	mkdir build; \
 	cd build; \
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu1604.clang.ossfuzz
@@ -22,7 +22,7 @@
 # (c) 2016-2021 solidity contributors.
 #------------------------------------------------------------------------------
 FROM gcr.io/oss-fuzz-base/base-clang:latest as base
-LABEL version="19"
+LABEL version="20"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
@@ -93,7 +93,7 @@ RUN set -ex; \
 # EVMONE
 RUN set -ex; \
 	cd /usr/src; \
-	git clone --branch="v0.8.2" --recurse-submodules https://github.com/ethereum/evmone.git; \
+	git clone --branch="v0.9.1" --recurse-submodules https://github.com/ethereum/evmone.git; \
 	cd evmone; \
 	mkdir build; \
 	cd build; \
--- a/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004.clang
+++ b/scripts/docker/buildpack-deps/Dockerfile.ubuntu2004.clang
@@ -22,7 +22,7 @@
 # (c) 2016-2019 solidity contributors.
 #------------------------------------------------------------------------------
 FROM buildpack-deps:focal AS base
-LABEL version="14"
+LABEL version="15"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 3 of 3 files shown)


axic:
Depends on https://github.com/ewasm/hera/pull/572 getting finished.

axic:
Added the Hera image.

scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
```diff
@@ -48,7 +48,7 @@ FROM base AS libraries
 # EVMONE
 RUN set -ex; \
 	cd /usr/src; \
-	git clone --branch="v0.8.0" --recurse-submodules https://github.com/ethereum/evmone.git; \
+	git clone --branch="v0.9.1" --recurse-submodules https://github.com/ethereum/evmone.git; \
```
cameel, line 51: What about the other two dockerfiles? They still have 0.8.2 hard-coded.

cameel (commented):
Is this PR independent of other changes? I.e. can it be merged without bumping the EVMC API version?

axic:
> Is this PR independent of other changes? I.e. can it be merged without bumping the EVMC API version?

Well it is independent assuming you don't use these buildpack versions. The EVMC ABI is breaking so they can't be used without the other PR.

scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
axic: Which other two?

I think ossfuzz was missed, but that's not even using it? Will fix that.

bshastry: ossfuzz uses it to fuzz the abi coder (because coding correctness cannot be interpreted via the yul interpreter and needs a standalone evm client).

bshastry (commented):
LGTM. Since the buildpack versions created by this PR will only be used after https://github.com/ethereum/solidity/pull/13532 is merged, it is better to keep this PR open until then to tally buildpack versions used in that PR.

nikola-matic:
@axic `install_evmone.ps1` powershell script for Windows also pulls the `0.8.0` evmone, although this script doesn't seem to be used in any of the CI steps, so my assumption is that it's just a convenience script for building the compiler on Windows.
Should this also be bumped? @cameel

cameel:
@nikola-matic The script is used by `t_win_soltest`.

But yeah, I think that the version should be bumped everywhere.

scripts/docker/buildpack-deps/Dockerfile.ubuntu2004
cameel: Yeah, I meant ossfuzz. I think we should update all images as long as they're in the repo.

Not sure where I see the other one. I must have mistakenly looked at one of the images already updated here.

I see this was addressed so marking this as resolved.

cameel (dismissed):
@nikola-matic The scripts should be updated, but that's addressed by #13532. Here we're only concerned with rebuilting the Docker images.

nikola-matic:
I've rebased against fresh develop to get this into a merge-able state. The PR is approved, however I see @bshastry would prefer for this to stay open until #13532 is close to being merged. With our current push to trim our PR queue, I'd be fine with merging this right away. @bshastry, @cameel, any counter points?

axic:
> I've rebased against fresh develop to get this into a merge-able state.

Oh, that means now need to change the hashes in the other PR. These two PRs should only be merged/rebased at the same time, they are unfortunately not independent. The reason they are split is due to the way buildpack works.

nikola-matic:
> > I've rebased against fresh develop to get this into a merge-able state.
> 
> Oh, that means now need to change the hashes in the other PR. These two PRs should only be merged/rebased at the same time, they are unfortunately not independent. The reason they are split is due to the way buildpack works.

Sure thing, although if this isn't merge-able, then the PR should be converted to a draft so it's clear to everyone. At least the rebase + force push dismissed @cameel's approval.

nikola-matic:
I've converted this to a draft as it should be merged together with https://github.com/ethereum/solidity/pull/13532. @axic feel free to revert this to a review-ready PR if needed.

cameel:
Well, it's not mergeable but it's ready so we don't want people to skip reviewing it. So perhaps `has dependencies` label would be better here. These two PRs have a cyclic dependency on each other :)

nikola-matic:
> Well, it's not mergeable but it's ready so we don't want people to skip reviewing it. So perhaps `has dependencies` label would be better here. These two PRs have a cyclic dependency on each other :)

Yup, I though about adding the `has dependencies` label, but at the same time the circular dependency made me think otherwise. The PR that does depend on this however, is in fact a draft, so really no idea what to think about this :)

cameel:
I mean, I don't consider it a draft because there's nothing more to do in it. It's done.

cameel (approved):

axic:
@cameel the only risk is merging this and then due to some other package update the hashes are changing in circle will cause a failure due to incompatible EVMC ABIs. If we expect the other PR to be merged very soon then it is safe to merge this, otherwise not.

nikola-matic:
> @cameel the only risk is merging this and then due to some other package update the hashes are changing in circle will cause a failure due to incompatible EVMC ABIs. If we expect the other PR to be merged very soon then it is safe to merge this, otherwise not.

So that's kinda the reason I converted it to a draft - mainly so that it blocks merging before the other PR is ready as well. Not sure what to do know. You decide @axic, you're the boss.

cameel:
> If we expect the other PR to be merged very soon then it is safe to merge this, otherwise not.

Should be pretty soon - the only blocker was this gas cost change to investigate. I see you pushed new commits there. So is that resolved?

But I'd still keep this open and merge them both at the same time.

> So that's kinda the reason I converted it to a draft - mainly so that it blocks merging before the other PR is ready as well. Not sure what to do know. You decide @axic, you're the boss.

At this point it's moot since this PR is already approved so either way achieves the goal. But my point was more in general - if you want reviews on such a PR it's probably better not to have it marked as draft so `has dependencies` would work better :) That label is a warning that the PR cannot be "just" merged.

ekpyron:
For the record, I'd also tend to keep these things drafts to prevent accidental merging, until both PRs are ready :-). Draft meaning "cannot be merged" here, which is true for either of them, regardless of reviews, unless both of them are ready.

axic:
Ready for final review.

PR #13540, opened 2022-09-19
renames lspAssert() to lspRequire + documentation

### LINKED ISSUE #13539: Rename LSP's lspAssert() function into something more meaningful.
"""
Yes, please rename it. And yeah, not in this PR, I'd create a separate one - though it very likely will go into `develop` before this one anyway.

As for recoverability - I think that asserts are fatal enough that they should not be recoverable. And the other way around - if you want to recover from it, then is it really a solid assumption that is guaranteed to be always true in a bug-free program? There are other kinds of errors the LSP will not be able to recover from, like memory violations or out of memory. I consider it this kind of problem. If another, unexpected kind of node appears out of nowhere, the program is just coded wrong, best to shut down ASAP to limit potential damage and just fix the bug.
"""

_Originally posted by @cameel in https://github.com/ethereum/solidity/pull/12790#discussion_r968270187_

christianparpart:
Closes #13539. Requires lsp-find-all-refs to be merged first.

### DIFF
--- a/libsolidity/lsp/FileRepository.cpp
+++ b/libsolidity/lsp/FileRepository.cpp
@@ -88,7 +88,7 @@ string FileRepository::sourceUnitNameToUri(string const& _sourceUnitName) const
 
 string FileRepository::uriToSourceUnitName(string const& _path) const
 {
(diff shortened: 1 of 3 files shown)


nishant-sachdeva (approved):
LGTM

Issue #13253, opened 2022-07-06
LSP: Allow disabling/enabling of individual features


Marenz:
This is a request by [vscode-solidity](vscode-solidity) author @juanfranblanco to make integration easier as it allows to use only the implemented and working features of our lsp.

christianparpart:
Hey @joshuahannan !

Do you have any preferred way of providing that information?
IMHO we could receive that information as part of the LSP handshake. Thanksfully the client initiates the initialize call and there one could simply say what it supports (or in other words(!): what it wants us to support), using the capabilities field. If the client only advertises semanticTokens, for example, we could respone the initialize handshake with a capabilities field that also only advertises semantic tokens support - effectively making sure that all the other features are in fact responding with "Method not supported" (forgot the actual code naming, but you get the point).

What do you think?

EDIT: Forgot to ask how that can help you on the vscode-solidity side. I am curious :)

juanfranblanco:
Hey @christianparpart,

Somehow I am guessing you meant to tag me!! Yeah.. that is mainly it. It helps so features that might not be working or in progress then can be disabled, also easier to experiment. Although the end, end goal, remember is to load the solcjs and proxy the calls depending on the compiler version, and what is supported.

christianparpart:
> to load the solcjs and proxy the calls depending on the compiler version, and what is supported.

Agreed. If we get the native passing through first, I'd highly like that, and then I hope we can focus on solcjs again. Having solcjs exposing LSP is pretty high on the table. I just hope I'll find someone knowledgeable here to pair with me on that in one way or another. :)

christianparpart:
@juanfranblanco Would it be sufficient to simply obey to what the client is advertising as feature? Say the client supports feature set (A, B), but the server supports (B, C), then only (C) is being enabled? Would this suffer [vscode-solidity](https://github.com/juanfranblanco/vscode-solidity)?

christianparpart:
For reference for myself. This is an example init payload for its client capabilities

```json
    "capabilities":
    {
      "textDocument": {
          "callHierarchy": { "dynamicRegistration": true },
[...]

PR #12790, opened 2022-03-14
LSP: Implements find-all-references.

christianparpart:
### Checklist

- [x] find-all-references logic revived.
- [x] I've included documentHighlight (semantically highlighting symbols matching the current word) because it's actually the same backend code.
- [x] tests added
- [x] Changelog entry

PR #13541, opened 2022-09-19
Only advertise and handle features that were reported in client capabilities.

christianparpart:
References: #13253
Based on top of: #12790 

NB: This does not apply to text edit feature, as this should always be present in order to get notified about text updates.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@ Compiler Features:
 * Commandline Interface: Add `--no-cbor-metadata` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Standard JSON: Add a boolean field `settings.metadata.appendCBOR` that skips CBOR metadata from getting appended at the end of the bytecode.
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
+* Language Server: Constrain server feature set based on advertised client capabilities.
 
 
 Bugfixes:
--- a/libsolidity/lsp/GotoDefinition.cpp
+++ b/libsolidity/lsp/GotoDefinition.cpp
@@ -16,6 +16,7 @@
 */
 // SPDX-License-Identifier: GPL-3.0
 #include <libsolidity/lsp/GotoDefinition.h>
+#include <libsolidity/lsp/LanguageServer.h>
 #include <libsolidity/lsp/Transport.h> // for RequestError
 #include <libsolidity/lsp/Utils.h>
 #include <libsolidity/ast/AST.h>
@@ -31,6 +32,23 @@ using namespace solidity::frontend;
 using namespace solidity::langutil;
 using namespace solidity::lsp;
 using namespace std;
+
+Json::Value GotoDefinition::onReportCapabilities(Json::Value const& _clientCapabilities)
+{
+	Json::Value replyCapabilities = Json::objectValue;
+	if (_clientCapabilities["textDocument"]["definition"])
+	{
+		replyCapabilities["definitionProvider"] = true;
+		m_server.registerHandler("textDocument/definition", *this);
+	}
+
+	if (_clientCapabilities["textDocument"]["implementation"])
+	{
+		replyCapabilities["implementationProvider"] = true;
+		m_server.registerHandler("textDocument/implementation", *this);
+	}
+	return replyCapabilities;
+}
 
 void GotoDefinition::operator()(MessageID _id, Json::Value const& _args)
 {
--- a/libsolidity/lsp/HandlerBase.h
+++ b/libsolidity/lsp/HandlerBase.h
@@ -18,10 +18,11 @@
 #pragma once
 
 #include <libsolidity/lsp/FileRepository.h>
-#include <libsolidity/lsp/LanguageServer.h>
 
 #include <liblangutil/SourceLocation.h>
 #include <liblangutil/CharStreamProvider.h>
+
+#include <libsolutil/JSON.h>
 
 #include <optional>
 
@@ -29,6 +30,7 @@ namespace solidity::lsp
 {
 
 class Transport;
+class LanguageServer;
 
 /**
  * Helper base class for implementing handlers.
@@ -37,6 +39,16 @@ class HandlerBase
 {
 public:
 	explicit HandlerBase(LanguageServer& _server): m_server{_server} {}
+	virtual ~HandlerBase() = default;
+
+	/// Callback to be invoked on every custom handler that can be used to decide whether to enable
+	/// or disable this feature on the server side.
+	///
+	/// @param _clientCapabilities JSON node to the root of the client advertised capabilities to be used
+	///                            to decide if the implemented feature should be enabled or not.
+	/// @returns JSON object, with mapping to the capabilities to reply back with.
+	///          Use this to write the capabilities-reply with respect the implementation.
+	virtual Json::Value onReportCapabilities(Json::Value const& /*_clientCapabilities*/) { return Json::objectValue; }
 
 	Json::Value toRange(langutil::SourceLocation const& _location) const;
 	Json::Value toJson(langutil::SourceLocation const& _location) const;
@@ -45,9 +57,9 @@ public:
 	/// from the JSON-RPC parameters.
 	std::pair<std::string, langutil::LineColumn> extractSourceUnitNameAndLineColumn(Json::Value const& _params) const;
 
-	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.compilerStack(); }
-	FileRepository& fileRepository() const noexcept { return m_server.fileRepository(); }
-	Transport& client() const noexcept { return m_server.client(); }
+	langutil::CharStreamProvider const& charStreamProvider() const noexcept;
+	FileRepository& fileRepository() const noexcept;
+	Transport& client() const noexcept;
 
 protected:
 	LanguageServer& m_server;
--- a/libsolidity/lsp/LanguageServer.cpp
+++ b/libsolidity/lsp/LanguageServer.cpp
@@ -128,11 +128,57 @@ Json::Value semanticTokensLegend()
 	return legend;
 }
 
+class SemanticTokensHandler: public HandlerBase
+{
+public:
+	using HandlerBase::HandlerBase;
+	Json::Value onReportCapabilities(Json::Value const& _clientCapabilities) override;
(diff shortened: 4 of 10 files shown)


Marenz (dismissed):

libsolidity/lsp/HandlerBase.h
```diff
@@ -37,6 +39,9 @@ class HandlerBase
 {
 public:
 	explicit HandlerBase(LanguageServer& _server): m_server{_server} {}
+	virtual ~HandlerBase() = default;
+
+	virtual void initialize(Json::Value const& /*_clientCapabilities*/, Json::Value& /*_replyCapabilities*/) {}
```
cameel:
```suggestion
	virtual void initialize(Json::Value const& /*_clientCapabilities*/, Json::Value& /*_serverCapabilities*/) {}
```

cameel: This does not look like generic initialization. Looks like hook for modifying the capability list before it's reported to the client:

```suggestion
	virtual void onReportCapabilities(Json::Value const& /*_clientCapabilities*/, Json::Value& /*_replyCapabilities*/) {}
```

Also, is `_replyCapabilities` always empty or pre-initialized? This should be specified in the docstring.

Finally, I think it would make more sense for the function to only return the list of *extra* capabilities the handler provides instead of being able to modify the whole list. Should one handler really be able to remove capabilities added by another handler?

libsolidity/lsp/HandlerBase.h
```diff
@@ -45,9 +50,9 @@ class HandlerBase
 	/// from the JSON-RPC parameters.
 	std::pair<std::string, langutil::LineColumn> extractSourceUnitNameAndLineColumn(Json::Value const& _params) const;
 
-	langutil::CharStreamProvider const& charStreamProvider() const noexcept { return m_server.compilerStack(); }
-	FileRepository& fileRepository() const noexcept { return m_server.fileRepository(); }
-	Transport& client() const noexcept { return m_server.client(); }
+	langutil::CharStreamProvider const& charStreamProvider() const noexcept;
+	FileRepository& fileRepository() const noexcept;
+	Transport& client() const noexcept;
```
cameel, line 62: Why?

libsolidity/lsp/GotoDefinition.cpp
```diff
@@ -32,6 +33,21 @@ using namespace solidity::langutil;
...
+	{
+		_replyCapabilities["definitionProvider"] = true;
+		m_server.registerHandler("textDocument/definition", *this);
+	}
+
+	if (_clientCapabilities["textDocument"]["implementation"])
+	{
+		_replyCapabilities["implementationProvider"] = true;
+		m_server.registerHandler("textDocument/implementation", *this);
+	}
```
cameel: The spec says this:
> Client capabilities got introduced with version 3.0 of the protocol. They therefore only describe capabilities that got introduced in 3.x or later. Capabilities that existed in the 2.x version of the protocol are still mandatory for clients. Clients cannot opt out of providing them. So even if a client omits the `ClientCapabilities.textDocument.synchronization` it is still required that the client provides text document synchronization (e.g. open, changed and close notifications).

`textDocument/definition` seems to be one of those 2.x capabilities so older clients might omit it even if they support it.

Might be similar with some other capabilities, I checked only those 2.

Changelog.md
```diff
@@ -6,6 +6,7 @@ Language Features:
 Compiler Features:
 * Yul Optimizer: Allow replacing the previously hard-coded cleanup sequence by specifying custom steps after a colon delimiter (``:``) in the sequence string.
 * Language Server: Implements finding all references as well as semantic highlighting of symbols.
+* Language Server: Constrain server feature set based on advertised client capabilities.
```
cameel, line 10: Does the LSP spec actually say that the server should trim its capabilities based on what the client requests? I did a quick check and did not see anything about it in the section about the initialization request, though maybe it's buried somewhere deeper.

But just from a common sense perspective - why would the server have different capabilities depending on what the client reports? These things seem completely independent and it seems useful for the client to know about all capabilities even if it does not support them (they could be e.g. displayed to the user as debug info; if would be frustrating for me as a user if it changed based on what the client requests).

libsolidity/lsp/LanguageServer.h
```diff
@@ -115,6 +118,11 @@ class LanguageServer
 	enum class State { Started, Initialized, ShutdownRequested, ExitRequested, ExitWithoutShutdown };
 	State m_state = State::Started;
 
+	// Contains the list of features that are only advertised if advertised by the client
+	// during initialization stage.
+	std::array<std::unique_ptr<HandlerBase>, 5> m_availableHandlers;
```
cameel:
```suggestion
	std::array<std::unique_ptr<HandlerBase>, 5> m_onDemandHandlers;
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -130,34 +130,75 @@ Json::Value semanticTokensLegend()
...
+	SourceUnit const& ast = m_server.compilerStack().ast(sourceName);
+	charStreamProvider().charStream(sourceName);
+
+	Json::Value data = SemanticTokensBuilder().build(ast, charStreamProvider().charStream(sourceName));
+
+	Json::Value reply = Json::objectValue;
+	reply["data"] = data;
+
+	client().reply(_id, std::move(reply));
+}
```
cameel: The other handlers are not much bigger than this. If they're in separate files, I'd put this one in a separate file too.

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -130,34 +130,75 @@ Json::Value semanticTokensLegend()
 	return legend;
 }
 
+class SemanticTokens: public HandlerBase
```
```suggestion
class SemanticTokensHandler: public HandlerBase
```

libsolidity/lsp/LanguageServer.cpp
```diff
@@ -130,34 +130,75 @@ Json::Value semanticTokensLegend()
...
+{
+public:
+	explicit SemanticTokens(LanguageServer& _languageServer): HandlerBase(_languageServer) {}
```
```suggestion
	using HandlerBase::HandlerBase;
```

libsolidity/lsp/HandlerBase.h
christianparpart: If I keep that implementation in the header, then it'll not compile.

Changelog.md
christianparpart: Generally speaking, this is PR 1 of 2 (to be followed). this one constrains the functionality as advertised by the LSP client. While it certainly would (in theory) not hurt if the server can respond to more than what the client advertises to support it is also useless if the server provides support for, say, `textDocument/definition`, when the LSP client did not advertise support for it anyways.

Long story short, the VScode solidity extension should in the long run use this LSP server to provide solidity LSP features. But for a smoother transition, the author was requesting to explicitly whitelist/blacklist features that it want's solc to have enabled (that's PR 2 of 2, it's not part of this PR). So this PR is generally also setting the grounding building blocks for the follow-up.

libsolidity/lsp/HandlerBase.h
christianparpart: done

libsolidity/lsp/GotoDefinition.cpp
christianparpart: We're currently at version 3.17. Version 3.1 came out Feb. 2017, I do not think it makes sense to support anything significantly older in this context.

libsolidity/lsp/LanguageServer.cpp
christianparpart: @cameel Do you want me to do that in exactly this PR? (Because that's a different topic you're addressing here).

libsolidity/lsp/GotoDefinition.cpp
cameel: I mean, older clients is just the justification but regardless of the reason, the spec basically says that `textDocument/definition` is always supported, even if the client tries to say it's not.

Changelog.md
cameel: > Long story short, the VScode solidity extension should in the long run use this LSP server to provide solidity LSP features. But for a smoother transition, the author was requesting to explicitly whitelist/blacklist features that it want's solc to have enabled (that's PR 2 of 2, it's not part of this PR). So this PR is generally also setting the grounding building blocks for the follow-up.

Oh. That's a very important piece of context that I was missing when I was reviewing this for the first time. I only skimmed the description but I now see that this was discussed in #13253.

This should really be in some comment in the code as well. It's not obvious at all and without this justification the whole mechanism seems just unnecessary and weird.

Also, wouldn't [`ServerCapabilities::experimental`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities) be a better (and really self-explanatory) way to do this? Dynamically enabling/disabling capabilities seems just like round-about way to go at it.

libsolidity/lsp/LanguageServer.cpp
cameel: Not sure I understand. Why is it a different topic? It's this PR that adds the handler so it could just as well add it as a separate file. It's not modifying an existing handler. Unless I'm missing some reason that prevented you from doing this?

libsolidity/lsp/LanguageServer.h
```diff
@@ -113,6 +116,11 @@ class LanguageServer
 	enum class State { Started, Initialized, ShutdownRequested, ExitRequested, ExitWithoutShutdown };
 	State m_state = State::Started;
 
+	// Contains the list of features that are only advertised if advertised by the client
+	// during initialization stage.
+	std::array<std::unique_ptr<HandlerBase>, 3> m_onDemandHandlers;
```
cameel, line 121: Now that I know the context, I think that `experimentalHandlers` would actually be a much better name:
```suggestion
	std::array<std::unique_ptr<HandlerBase>, 3> m_experimentalHandlers;
```

Also, the comment should provide this context and say why we're doing this in the first place. It's probably not a typical LSP behavior.

libsolidity/lsp/HandlerBase.h
cameel: But why? It did compile until now and at a glance the implementation looks unchanged. Am I missing some detail here?

libsolidity/lsp/HandlerBase.h
```diff
@@ -45,9 +57,9 @@ class HandlerBase
...
```
cameel, line 62: Just noticed that these return non-const references. Should the functions really be `const` in that case?

Issue #13543, opened 2022-09-20
Semantic test for copying calldata structs placed in the wrong subdirectory


matheusaaguiar:
It seems that [test/libsolidity/semanticTests/structs/copy_from_calldata.sol](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/semanticTests/structs/copy_from_calldata.sol) is misplaced, since it does not involve structs at all, but rather just dynamic arrays. Shouldn't it be inside `array` subdirectory? 
Noticed the possible issue [here](https://github.com/ethereum/solidity/pull/13464#issuecomment-1239819462).

cameel:
True. We should move it and any other non-struct tests from that dir. But I think we should also add an actual test for copying calldata structs.

KatyaPrusakova:
Hi, I would like to contribute, it would be my ever [good first issue] , so I hope it works :)

cameel:
Sure! Feel free to create a PR.

nikola-matic:
> True. We should move it and any other non-struct tests from that dir. But I think we should also add an actual test for copying calldata structs.

We already have these:
```
copy_struct_with_nested_array_from_calldata_to_memory.sol
copy_struct_with_nested_array_from_calldata_to_storage.sol
```

PR #13546, opened 2022-09-21
test copy_from_calldata.sol folder location update

KatyaPrusakova:
#13543 Semantic test for copying calldata structs placed in the wrong subdirectory
@cameel  could you help to see if I got the task correct?


cameel:
1) Have you checked if there are any other non-struct tests in that dir?
2) The issue says to also add a proper struct-based test for this case.
3) There are many tests for copying calldata in the `array/copying/` dir so we should give this test a more specific name. How about `calldata_1d_array_into_2d_memory_array_element.sol`

nikola-matic:
@KatyaPrusakova, bumping since you've been in active, and the PR would technically be considered stale? Do you plan on implementing the test @cameel suggested? I've checked, and `copy_from_calldata.sol` is the only tests in the `structs` directory that doesn't actually use a struct, so no other tests need to be moved.

nikola-matic:
I've renamed the test as per @cameel's suggestion. The tests he requested however (copy calldata structs) already exist, so it's a no-op.

cameel:
I'd squash the commits (no need to rename it twice :)) but other than that it looks fine.

nikola-matic (approved):

cameel (approved):

PR #13549, opened 2022-09-21
Fix stripping in CLI tests does not produce correct JSON formatting

### LINKED ISSUE #13544: Pre-release warning stripping in CLI tests does not always produce correct JSON formatting when combined with `--pretty-print`
This came up in https://github.com/ethereum/solidity/pull/13162#pullrequestreview-1101249411.

Non-release builds of the compiler always output an extra warning, which results in a different output between release and debug builds in CI:
```
Warning: This is a pre-release compiler version, please do not use it in production.
```

For this reason `cmdlineTests.sh` script strips the warning from compile output. For Standard JSON this is more tricky, since the error is a part of an array and we have to match the JSON formatting of the expectations. The stripping code seems to have a small bug because in some cases the formatting it produces after stripping the warning differs from what we get when the warning is not there at all.

We need to adjust the stripping code to handle this corner case. It would be best if we could also solve the problem more generally by making the CLI tests simply ignore the JSON formatting.

Fixing this is likely a prerequisite for #7665, since enabling pretty printing may uncover this problem in other existing tests.

### How to reproduce
Let's look at the prettified Standard JSON output for this `input.json` file:
```json
{
    "language": "Solidity",
    "sources": {
        "url_not_found.sol": {
            "urls": ["contract.sol"]
        }
    }
}
```
```bash
solc --standard-json input.json --pretty-json --json-indent 4
```
This is what the output of a pre-release binary looks like unstripped:
```json
{
    "errors":
    [
        {
            "component": "general",
            "formattedMessage": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "message": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "severity": "error",
            "type": "IOError"
        },
        {
            "component": "general",
            "errorCode": "3805",
            "formattedMessage": "Warning: This is a pre-release compiler version, please do not use it in production.\n\n",
            "message": "This is a pre-release compiler version, please do not use it in production.",
            "severity": "warning",
            "type": "Warning"
        }
    ],
    "sources": {}
}
```
The stripping in `cmdlineTests.sh` removes the pre-release warning, which results in this output:
```json
{
    "errors":
    [
        {
            "component": "general",
            "formattedMessage": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "message": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "severity": "error",
            "type": "IOError"
        }],
    "sources": {}
}
```
Note the `}],` sequence in the output. A release binary, that does not output the warning produces a slightly different formatting in that spot:
```json
{
    "errors":
    [
        {
            "component": "general",
            "formattedMessage": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "message": "Cannot import url (\"contract.sol\"): File not found. Searched the following locations: \"\".",
            "severity": "error",
            "type": "IOError"
        }
    ],
    "sources": {}
}
```

As a result, it's impossible to write a test with nicely formatted JSON output that matches both release and pre-release build output. For example in #13162 [`t_ubu_release` fails if we expect `}],`](https://app.circleci.com/pipelines/github/ethereum/solidity/26515/workflows/1537a81d-d798-4598-a7e3-4743160ca2c3/jobs/1174412) and [`t_ubu_cli` fails if we expect the other variant](https://app.circleci.com/pipelines/github/ethereum/solidity/26478/workflows/326ca1f8-e91e-4034-ba21-dfac16cd2b86/jobs/1172659).

### The warning stripping code
You can find the code responsible for stripping this warning here:
https://github.com/ethereum/solidity/blob/1fbee8259aca10ffe04efa47a21f31d7dee1baa9/test/cmdlineTests.sh#L189-L199

miles170:
Closes #13544.

### DIFF
--- a/test/cmdlineTests.sh
+++ b/test/cmdlineTests.sh
@@ -192,9 +192,9 @@ function test_solc_behaviour
 import re, sys
 json = open("$stdout_path", "r").read()
 json = re.sub(r"{[^{}]*Warning: This is a pre-release compiler version[^{}]*},?", "", json)
-json = re.sub(r"},\s*]", "}]", json)                  # },] -> }]
-json = re.sub(r"\"errors\":\s*\[\s*\],?\s*","",json)  # Remove "errors" array if it's not empty
-json = re.sub("\n\\s+\n", "\n\n", json)               # Remove any leftover trailing whitespace
+json = re.sub(r"\"errors\":\s*\[\s*\],?\s*","",json)    # Remove "errors" array if it's not empty
+json = re.sub("\n\\s+\n", "\n\n", json)                 # Remove trailing whitespace
+json = re.sub(r"},(\n{0,1})\n*(\s*])", r"}\1\2", json)  # },] -> }]
 open("$stdout_path", "w").write(json)
 EOF
         sed -i.bak -E -e 's/ Consider adding \\"pragma solidity \^[0-9.]*;\\"//g' "$stdout_path"
--- a/test/cmdlineTests/standard_urls_existing_and_missing/args
+++ b/test/cmdlineTests/standard_urls_existing_and_missing/args
@@ -1,0 +1,1 @@
+--pretty-json --json-indent 4
--- a/test/cmdlineTests/standard_urls_existing_and_missing/output.json
+++ b/test/cmdlineTests/standard_urls_existing_and_missing/output.json
@@ -1,1 +1,19 @@
-{"errors":[{"component":"general","formattedMessage":"Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".","message":"Cannot import url (\"standard_urls_existing_and_missing/non-existent-contract-1.sol\"): File not found. Searched the following locations: \"\".","severity":"warning","type":"Warning"}],"sources":{"url_not_found.sol":{"id":0}}}
+{
+    "errors":
+    [
+        {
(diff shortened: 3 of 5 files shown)


nishant-sachdeva (dismissed):
LGTM

cameel:
Can you add `--pretty-print --json-indent 4` in `args` to (`test/cmdlineTests/standard_urls_existing_and_missing/`) like we attempted in https://github.com/ethereum/solidity/pull/13162#pullrequestreview-1101249411? This is the real test that this change actually works :)

If that passes, we can merge.

nishant-sachdeva:
> Can you add --pretty-print --json-indent 4 in args to (test/cmdlineTests/standard_urls_existing_and_missing/) like we attempted in https://github.com/ethereum/solidity/pull/13162#pullrequestreview-1101249411? This is the real test that this change actually works :)

In that case , we should add it to an `args` file in `test/cmdlineTests/standard_urls_missing/args` as well. Both of these were giving issues in https://github.com/ethereum/solidity/pull/13162#pullrequestreview-1101249411

miles170:
CI failed because error not matched, I will fix it asap.

r0qs (dismissed):
Looks good to me as well.

nishant-sachdeva (approved):

r0qs (approved):

PR #13553, opened 2022-09-22
Review checklist

cameel:
I tried to create a checklist for reviewers and contributors to make our reviews a bit more consistent.

Marking it as a draft since I want first to get opinions and feedback from the team. Both about the form and content. Feel free to suggest things you would change, add or remove.

I wanted it to be concise but it's hard to do that while keeping the reasoning clear. I used bold to mark the important parts and make it still somewhat usable as a checklist.

PR #13554, opened 2022-09-23
Add code style rule for std:: qualifier

nikola-matic:
See https://github.com/ethereum/solidity/pull/13553#discussion_r977920167

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -55,6 +55,7 @@ To set indentation and tab width settings uniformly, the repository contains an 
 3. All symbols should be declared in a namespace except for final applications.
 4. Use anonymous namespaces for helpers whose scope is a cpp file only.
 5. Preprocessor symbols should be prefixed with the namespace in all-caps and an underscore.
+6. Do not use `std::` qualifier in cpp files (see 2.), except for `std::move`, which will otherwise cause the `check_style` step to fail.
 
 Only in the header:
 ```cpp


CODING_STYLE.md
```diff
@@ -55,6 +55,7 @@ To set indentation and tab width settings uniformly, the repository contains an
 3. All symbols should be declared in a namespace except for final applications.
 4. Use anonymous namespaces for helpers whose scope is a cpp file only.
 5. Preprocessor symbols should be prefixed with the namespace in all-caps and an underscore.
+6. Do not use `std::` qualifier in cpp files (see 2.), except for `std::move`, which will otherwise cause the `check_style` step to fail.
```
nishant-sachdeva, line 58:
```suggestion
6. Do not use `std::` qualifier in cpp files (see 2.), except for `std::move` (this is an exception to the rule, failing which  will otherwise cause the `check_style` step to fail ).
```

nikola-matic: I don't see this making it any clearer, just more verbose.

cameel: I agree. The current wording is pretty clear (it does say `except for`).

We might add something about *why* `std::move` is an exception but IMO it's also fine without it.

cameel, line 58: The rule is actually a bit more general. We just do not add any namespace qualifier when we already have `using namespace` in the file. The `std` is just the most prominent example because we have it everywhere and it's the easiest one to spot in reviews :)

cameel (approved):
We could make the rule a bit more general but even in the current form it's an improvement so approving.

PR #13560, opened 2022-09-25
Updated Whiskers for checking invalid tags

### LINKED ISSUE #13226: Strengthen Whiskers' matching
Whiskers' regex looks for `<..>` as opposed to tokenizing the input, so it is not able to detect truncated/mistyped input.

For example it can't detect that `<!b` is not terminated in`<?b> X <!b Y </b>`.

With some heuristics it could be improved: match for `<?`, `<!` and `</` and only allow identifiers which are terminated with `>`, otherwise it is invalid.

_Originally posted by @axic in https://github.com/ethereum/solidity/pull/13219#discussion_r907476574_

rajgaur98:
For Example, raise an error if it contains invalid tags like <?ba<!b</b

Closes #13226

### DIFF
--- a/libsolutil/Whiskers.cpp
+++ b/libsolutil/Whiskers.cpp
@@ -34,6 +34,7 @@ using namespace solidity::util;
 Whiskers::Whiskers(string _template):
 	m_template(std::move(_template))
 {
+	checkTemplateValid();
 }
 
 Whiskers& Whiskers::operator()(string _parameter, string _value)
@@ -72,6 +73,17 @@ Whiskers& Whiskers::operator()(
 string Whiskers::render() const
 {
 	return replace(m_template, m_parameters, m_conditions, m_listParameters);
+}
+
+void Whiskers::checkTemplateValid() const
+{
+	regex validTemplate("<[#?!\\/]\\+{0,1}[a-zA-Z0-9_$-]+(?:[^a-zA-Z0-9_$>-]|$)");
+	smatch match;
+	assertThrow(
+		!regex_search(m_template, match, validTemplate),
+		WhiskersError,
+		"Template contains an invalid/unclosed tag " + match.str()
+	);
 }
 
 void Whiskers::checkParameterValid(string const& _parameter) const
--- a/test/libsolutil/Whiskers.cpp
+++ b/test/libsolutil/Whiskers.cpp
@@ -88,8 +88,34 @@ BOOST_AUTO_TEST_CASE(conditional_with_else)
 BOOST_AUTO_TEST_CASE(broken_conditional_with_else)
 {
 	string templ = "<?b>X<!bY</b>";
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", true).render(), "X<!bY");
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", false).render(), "");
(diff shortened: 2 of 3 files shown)


test/libsolutil/Whiskers.cpp
```diff
@@ -88,8 +88,13 @@ BOOST_AUTO_TEST_CASE(conditional_with_else)
...
 	string templ = "<?b>X<!bY</b>";
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", true).render(), "X<!bY");
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", false).render(), "");
+	BOOST_CHECK_THROW(Whiskers _(templ), WhiskersError);
+
+	templ = "<?bX<!b>Y</b>";
+	BOOST_CHECK_THROW(Whiskers _(templ), WhiskersError);
+
+	templ = "<?b>X<!b>Y</b";
+	BOOST_CHECK_THROW(Whiskers _(templ), WhiskersError);
```
leonardoalt:
```suggestion
	BOOST_CHECK_THROW(Whiskers{templ}, WhiskersError);
```
how about this?

test/libsolutil/Whiskers.cpp
```diff
@@ -88,8 +88,13 @@ BOOST_AUTO_TEST_CASE(conditional_with_else)
...
 	string templ = "<?b>X<!bY</b>";
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", true).render(), "X<!bY");
-	BOOST_CHECK_EQUAL(Whiskers(templ)("b", false).render(), "");
+	BOOST_CHECK_THROW(Whiskers{templ}, WhiskersError);
+
+	templ = "<?bX<!b>Y</b>";
+	BOOST_CHECK_THROW(Whiskers{templ}, WhiskersError);
+
+	templ = "<?b>X<!b>Y</b";
+	BOOST_CHECK_THROW(Whiskers{templ}, WhiskersError);
```
axic, line 97: Would be nice if we could inspect the exception message, but didn't find such a feature in boost after a quick look.

axic (dismissed):
Looks good to me.

libsolutil/Whiskers.cpp
```diff
@@ -74,6 +75,17 @@ string Whiskers::render() const
 	return replace(m_template, m_parameters, m_conditions, m_listParameters);
 }
 
+void Whiskers::checkTemplateValid() const
+{
+	regex validTemplate("<[?!\\/][a-zA-Z0-9_$-]+(?:[^a-zA-Z0-9_$>-]|$)");
```
axic: > 	/// Sets a list parameter, <#listName> </listName>.

There are also list parameters which could be truncated. Perhaps should support them here.

axic: While it looks ugly, probably should use `paramRegex()` for identifier part, as that's what other regexes in this file do.

Maybe this needs a refactoring in a separate PR, let's keep it as-is but leave a comment that it must match `paramRegex`.

rajgaur98: Sure, I will include the check for list params

rajgaur98: Ahh I see, totally forgot about `paramRegex`. I can refactor this in another PR

axic (approved):

PR #13562, opened 2022-09-26
EVMHost: Simplify some code and document functions

axic:
These are some cleanups benefitting both current version and #13532.

### DIFF
--- a/test/EVMHost.h
+++ b/test/EVMHost.h
@@ -40,8 +40,25 @@ using StorageMap = std::map<evmc::bytes32, evmc::storage_value>;
 class EVMHost: public evmc::MockedHost
 {
 public:
-	using MockedHost::get_code_size;
+	// Verbatim features of MockedHost.
+	using MockedHost::account_exists;
+	using MockedHost::get_storage;
+	using MockedHost::set_storage;
 	using MockedHost::get_balance;
+	using MockedHost::get_code_size;
+	using MockedHost::get_code_hash;
+	using MockedHost::copy_code;
+	using MockedHost::get_tx_context;
+	using MockedHost::emit_log;
+	using MockedHost::access_account;
+	using MockedHost::access_storage;
+
+	// Modified features of MockedHost.
+	void selfdestruct(evmc::address const& _addr, evmc::address const& _beneficiary) noexcept final;
+	evmc::result call(evmc_message const& _message) noexcept final;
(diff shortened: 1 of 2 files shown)


test/EVMHost.h
```diff
@@ -35,12 +35,30 @@
...
+	using MockedHost::copy_code;
+	using MockedHost::get_tx_context;
+	using MockedHost::emit_log;
+	using MockedHost::access_account;
+	using MockedHost::access_storage;
+
+	// Modified features of MockedHost.
+	void selfdestruct(evmc::address const& _addr, evmc::address const& _beneficiary) noexcept final;
+	evmc::result call(evmc_message const& _message) noexcept final;
+	evmc::bytes32 get_block_hash(int64_t number) const noexcept final;
```
leonardoalt, line 59:
```suggestion
	evmc::bytes32 blockHash(int64_t number) const noexcept final;
```

test/EVMHost.h
```diff
@@ -67,35 +86,21 @@ class EVMHost: public evmc::MockedHost
...
-	{
-		return evmc::MockedHost::account_exists(_addr);
-	}
-
-	void selfdestruct(evmc::address const& _addr, evmc::address const& _beneficiary) noexcept final;
-
-	evmc::result call(evmc_message const& _message) noexcept final;
-
-	evmc::bytes32 get_block_hash(int64_t number) const noexcept final;
+	StorageMap const& get_address_storage(evmc::address const& _addr);
```
leonardoalt, line 89:
```suggestion
	StorageMap const& addressStorage(evmc::address const& _addr);
```

test/EVMHost.h
axic: > 	// Modified features of MockedHost.

i.e. `get_block_hash` is the name.

leonardoalt: ah nvmd, just noticed it was just moved up

test/EVMHost.h
axic: Note that this is a refactoring PR, but sure we can rename these helpers.

leonardoalt: yea better not to

axic: This one may make more sense to rename, because the Solidity extensions use camel case, but EVMC uses snake case, so it is more distinguishable.

cameel (approved):
Everything looks fine to me here.

PR #13565, opened 2022-09-26
EVMHost: a few more simplifications

axic:
Forgot in #13562


hrkrshnn (approved):
Tested the inputs in https://gist.github.com/hrkrshnn/aa2237ba58d8a8c44798df324030e6a1 against revm / foundry.

test/EVMHost.cpp
```diff
@@ -771,14 +771,141 @@ evmc::result EVMHost::precompileALTBN128G1Mul(evmc_message const& _message) noex
...
+				"090689d0585ff075ec9e99ad690c3395bc4b313370b38ef355acdadcd122975b"
+				"12c85ea5db8c6deb4aab71808dcb408fe3d1e7690c43d37b4ce6cc0166fa7daa"
+				"0000000000000000000000000000000000000000000000000000000000000001"
+				"30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd45"
+				"0a09ccf561b55fd99d1c1208dee1162457b57ac5af3759d50671e510e428b2a1"
+				"2e539c423b302d13f4e5773c603948eaf5db5df8ae8a9a9113708390a06410d8"
+				"19b763513924a736e4eebd0d78c91c1bc1d657fee4214057d21414011cfcc763"
+				"2f8d9f9ab83727c77a2fec063cb7b6e5eb23044ccf535ad49d46d394fb6f6bf6"
+			),
+			fromHex("0000000000000000000000000000000000000000000000000000000000000001")
```
cameel, line 791: Do we have any tests that actually use these outputs?

cameel (approved):

test/EVMHost.cpp
axic: Yes, the snarks test is used by these. In fact it was generated by running the snarks test with having this function empty and gradually filling it out.

cameel: Ah right. Somehow I thought it must have been failing without these but now I realized that we were just accepting any input. In that case if the CI passes, it must be fine.

PR #13572, opened 2022-09-27
[Docs] Clarify relative path normalization rules

cameel:
Replaces and closes #12879.

I don't think that PR clarifies things well, but after reviewing it, I think I know where it's coming from. The current description is indeed a bit unclear about normalization.

### DIFF
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -197,7 +197,7 @@ source unit name.
 
     A source unit name is just an identifier and even if its value happens to look like a path, it
     is not subject to the normalization rules you would typically expect in a shell.
-    Any ``/./`` or ``/../`` seguments or sequences of multiple slashes remain a part of it.
+    Any ``/./`` or ``/../`` segments or sequences of multiple slashes remain a part of it.
     When the source is provided via Standard JSON interface it is entirely possible to associate
     different content with source unit names that would refer to the same file on disk.
 
--- a/docs/path-resolution.rst
+++ b/docs/path-resolution.rst
@@ -248,19 +248,14 @@ and is bounded by two path separators.
 A separator is a forward slash or the beginning/end of the string.
 For example in ``./abc/..//`` there are three path segments: ``.``, ``abc`` and ``..``.
 
-The compiler computes a source unit name from the import path in the following way:
+The compiler resolves the import into a source unit name based on the import path, in the following way:
 
-1. First a prefix is computed
-
-    - Prefix is initialized with the source unit name of the importing source unit.
-    - The last path segment with preceding slashes is removed from the prefix.
-    - Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
-      characters is considered.
-      For every ``..`` segment found in this part the last path segment with preceding slashes is
-      removed from the prefix.
-
-2. Then the prefix is prepended to the normalized import path.
-   If the prefix is non-empty, a single slash is inserted between it and the import path.
+#. We start with the source unit name of the importing source unit.
+#. The last path segment with preceding slashes is removed from the resolved name.
+#. Then, for every segment in the import path, starting from the leftmost one:
+    - If the segment is ``.``, it is skipped.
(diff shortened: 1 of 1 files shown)


docs/path-resolution.rst
```diff
@@ -250,32 +250,23 @@ For example in ``./abc/..//`` there are three path segments: ``.``, ``abc`` and
...
-      removed from the prefix.
-
-2. Then the prefix is prepended to the normalized import path.
-   If the prefix is non-empty, a single slash is inserted between it and the import path.
+#. The result is initialized with the source unit name of the importing source unit.
+#. The last path segment with preceding slashes is removed from the result.
+#. Then, for every segment in the import path, starting from the leftmost one:
+    - If the segment is ``.``, it is skipped.
+    - If the segment is ``..``, the last path segment with preceding slashes is removed from the result.
+    - Otherwise, the segment, (preceded by a single slash if the result is not empty), is appended to the result.
```
cameel: This part was a bit convoluted. By "normalized import path" I meant a path where `.` and `..` segments are resolved as much as possible. I.e. `.././../a//b/..//c/./d.sol` would be normalized to `../../a/c/d.sol`, which has no consecutive slashes and removes the internal `..` and `.` segments but still can have several `..` leading segments because those can't be matched with anything.

And yeah, it did not mention that these leading `..` segments should be removed before joining the prefix and the normalized path.

I replaced the whole thing with something that I think is simpler and more closely matches the way this is actually implemented: https://github.com/ethereum/solidity/blob/311b2054afae02bfaa8521fd94475a06e0669025/libsolutil/CommonIO.cpp#L141-L159

docs/path-resolution.rst
```diff
@@ -250,32 +250,23 @@ For example in ``./abc/..//`` there are three path segments: ``.``, ``abc`` and
...
-    - Prefix is initialized with the source unit name of the importing source unit.
-    - The last path segment with preceding slashes is removed from the prefix.
-    - Then, the leading part of the normalized import path, consisting only of ``/`` and ``.``
-      characters is considered.
-      For every ``..`` segment found in this part the last path segment with preceding slashes is
-      removed from the prefix.
-
-2. Then the prefix is prepended to the normalized import path.
-   If the prefix is non-empty, a single slash is inserted between it and the import path.
+#. The result is initialized with the source unit name of the importing source unit.
```
r0qs: What is the `result` is not very clear to me here. Why not keep it as prefix?

docs/path-resolution.rst
```diff
@@ -250,32 +250,23 @@ For example in ``./abc/..//`` there are three path segments: ``.``, ``abc`` and
...
 
-The normalization rules are the same as for UNIX paths, namely:
-
-- All the internal ``.`` segments are removed.
-- Every internal ``..`` segment backtracks one level up in the hierarchy.
-- Multiple slashes are squashed into a single one.
-
-Note that normalization is performed only on the import path.
-The source unit name of the importing module that is used for the prefix remains unnormalized.
+Note that the process normalizes the part of the source unit name that comes from the import according
```
r0qs:
```suggestion
Note that the process normalizes the part of the source unit name that comes from the import path according
```

docs/path-resolution.rst
```diff
@@ -250,32 +250,23 @@ For example in ``./abc/..//`` there are three path segments: ``.``, ``abc`` and
...
+to the usual rules for UNIX paths, i.e. all ``.`` and ``..`` are removed and multiple slashes are
+squashed into a single slash.
```
r0qs:
```suggestion
squashed into a single one.
```

r0qs (commented):
Hey, here is my 2 cents :)

docs/path-resolution.rst
cameel: Because the prefix is just the prefix, not the complete result. It made more sense when the final step that combines prefix with the import path was still here.

I changed it to "resolved name". Check if it's better now.

r0qs: Ah, I see. LGTM now :D

r0qs (approved):

PR #13573, opened 2022-09-27
Drop deprecated UndefMacros file

axic:
This file was added at a time when the token names were upper case only (such as [37e4919fcb0dd9a491f34cc54a04136bbcdb1f40](https://github.com/ethereum/solidity/blob/37e4919fcb0dd9a491f34cc54a04136bbcdb1f40/Token.h)).

Currently it is not useful and serves no purpose.


cameel (approved):
Should be fine if tests pass.

